PR #2434, opened 2023-02-09
ReplacementHandler: guard against StackOverflow if LKI gets used for Moved

tool4ever:
There's a small risk that when it takes the RE from LKI another effect calls copyLastState after, in that case it will recreate it from the original and the `hasRun()` will be false again.
Now (if the RE could apply again) it will loop.

This would only affect scenarios where a full replacement is needed, luckily most are already done with Updated result.

### DIFF
--- a/forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
+++ b/forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
@@ -200,7 +200,7 @@ public class ReplacementHandler {
                         }
                     }
 
-                    if (!replacementEffect.hasRun()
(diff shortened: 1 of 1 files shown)


Hanmac (approved):
👍

Hanmac:
i don't know if we need both "hasRun" or if we just can rely on the Attribute in the ReplacementHandler

but that can be merged for now

tool4ever:
Yea I also saw that when it creates RE for LKI it makes sure to keep the same ID, so it might be outdated

something to keep in mind for refactor

Hanmac:
see the part about `_ReplacedTimestamp` that was added because of Riot what might not be needed anymore

PR #2436, opened 2023-02-09
fix:'Draconic Destiny' cast cost and activated abilities cost.

CCTV-1:


forge-gui/res/cardsfolder/d/draconic_destiny.txt
```diff
@@ -2,9 +2,9 @@ Name:Draconic Destiny
 ManaCost:1 R R
 Types:Enchantment Aura
 K:Enchant creature
-A:SP$ Attach | Cost$ R | ValidTgts$ Creature | AILogic$ Pump
+A:SP$ Attach | Cost$ 1 R R | ValidTgts$ Creature | AILogic$ Pump
```
Hanmac, line 5: you could just remove the Cost and then it copies the ManaCost

Hanmac (approved):
👍

PR #2440, opened 2023-02-09
messing with some "this and that" stuff

Northmoc:


forge-gui/res/cardsfolder/c/ceta_sanctuary.txt
```diff
@@ -1,14 +1,12 @@
 Name:Ceta Sanctuary
 ManaCost:2 U
 Types:Enchantment
-T:Mode$ Phase | Phase$ Upkeep | ValidPlayer$ You | TriggerZones$ Battlefield | IsPresent$ Permanent.YouCtrl+Red,Permanent.YouCtrl+Green | Execute$ TrigDraw | TriggerDescription$ At the beginning of your upkeep, if you control a red or green permanent, draw a card, then discard a card. If you control a red permanent and a green permanent, instead draw two cards, then discard a card.
-SVar:TrigDraw:DB$ Draw | NumCards$ C | SubAbility$ DBDiscard
+T:Mode$ Phase | Phase$ Upkeep | ValidPlayer$ You | TriggerZones$ Battlefield | IsPresent$ Permanent.YouCtrl+Red | IsPresent2$ Permanent.YouCtrl+Green | PresentCompare2$ EQ0 | Execute$ TrigDraw | TriggerDescription$ At the beginning of your upkeep, if you control a red or green permanent, draw a card, then discard a card. If you control a red permanent and a green permanent, instead draw two cards, then discard a card.
```
tool4ever, line 4: Hmn, I'm not sure saving the SVars by splitting the trigger into three is a good improvement :/

Northmoc: I also wasn't sure that was "better" 😅

forge-gui/res/cardsfolder/c/ceta_sanctuary.txt
```diff
@@ -9,6 +9,5 @@ SVar:Z:Count$Valid Permanent.Green+YouCtrl/LimitMax.1
 SVar:X:SVar$Y/Plus.Z
 SVar:C:Count$Compare X EQ2.2.1
```
Hanmac, line 10: Shouldn't that C not be needed?

The Trigger (because its an Intervening IF Clause should only resolve when you get Red or Green
meaning that if you take Y and Z that each can be either 0 or 1
and Plus them with X, then it only can't be 1 or 2
because if its 0, the trigger should not be able to resolve

meaning you don't need to compare it with 2?

Hanmac:
we should move the Checks from SpellCondition and SpellRestriction into the Parent Class, but that is an Issue for later

The Checks should be the same, just the variables are different which already got handled

Hanmac (approved):

PR #2441, opened 2023-02-10
messing with StackDescription building

Northmoc:
I've been noticing that a lot of SpellDescriptions would be fine for StackDescription with:
 - a few simple substitutions
 - removal of reminder text

This allows you to substitute X new text for Y existing text in a SpellDescription, then use that as a StackDescription

image

Also tidied up the Targeting add-on to only work if there are none of these new replacements (since tokens will probably remove the need for it) and to use joinHomogenous

image


paulsnoops (approved):

Hanmac:
i don't like it that some of the parts is hardcoded for LangEnglish :(

Northmoc:
> i don't like it that some of the parts is hardcoded for LangEnglish :(

I just didn't want to break descriptions in other languages. I figured since the REP and other info on the card will be in English, it would have to be just done in English for now.

PR #2449, opened 2023-02-11
Flashback: use castKeyword

Hanmac:
This makes Flashback use castKeyword

meaning if a spell has multiple instances of the keyword, the Exile Replacement should only activate once

Also casting a card with Flashback doesn't add a second instance onto it if its intrinsic

Keyword instances now have HostCard and Intrinsic available (its mostly knowing when the SpellAbility is created in GameActionUtil for Split Cards)

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -268,7 +268,7 @@ public class GameAction {
                 // when moving to stack, copy changed card information
                 copied.setChangedCardColors(c.getChangedCardColorsTable());
                 copied.setChangedCardColorsCharacterDefining(c.getChangedCardColorsCharacterDefiningTable());
-                copied.setChangedCardKeywords(c.getChangedCardKeywords());
+                copied.setChangedCardKeywords(c.getChangedCardKeywords(), false);
                 copied.setChangedCardTypes(c.getChangedCardTypesTable());
                 copied.setChangedCardTypesCharacterDefining(c.getChangedCardTypesCharacterDefiningTable());
                 copied.setChangedCardNames(c.getChangedCardNames());
@@ -622,7 +622,9 @@ public class GameAction {
         if (zoneTo.is(ZoneType.Stack) && cause != null && cause.isSpell() && !cause.isIntrinsic() && c.equals(cause.getHostCard())) {
             if (cause.getKeyword() != null) {
                 if (!copied.getKeywords().contains(cause.getKeyword())) {
-                    copied.addChangedCardKeywordsInternal(ImmutableList.of(cause.getKeyword()), null, false, game.getTimestamp(), 0, false);
+                    copied.addChangedCardKeywordsInternal(ImmutableList.of(cause.getKeyword()), null, false, game.getNextTimestamp(), 0, false);
+                    // update Keyword Cache
+                    copied.updateKeywords();
                 }
             }
         }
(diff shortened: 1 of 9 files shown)


Agetian (dismissed):

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -619,7 +619,7 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
         checkStaticAbilities();
 
         // 400.7g try adding keyword back into card if it doesn't already have it
-        if (zoneTo.is(ZoneType.Stack) && cause != null && cause.isSpell() && c.equals(cause.getHostCard())) {
+        if (zoneTo.is(ZoneType.Stack) && cause != null && cause.isSpell() && !cause.isIntrinsic() && c.equals(cause.getHostCard())) {
```
tool4ever, line 622: Sorry, still found a way to break it:

When it gains two flashback (e.g. Flicker Snapcaster after resolving) and then cast it via the _first_ altcost it won't be exiled

Works with any spell

Hanmac: you don't need to flicker it, just cast a second Snapcaster

i just don't understand it why it gets two times the keywords when moving to stack, i need to debug it more

Hanmac: @tool4ever found the reason: getReplacementEffects returns an FCollection meaning it does filter duplicates

meaning Line 270+ with `copied.setChangedCardKeywords(c.getChangedCardKeywords());` already added the Flashkeyword to stack, but as a copy not as direct value (keywords doesn't have ID yet)
i made it so it doesn't copy when moving to stack


Can we try to get back why this was done?
a624cf565a6adf3fd12d968311acecd99cbdd4c9

tool4ever: From the timing I'd say we had the same issue then :P

image

Hanmac: the Old Code for Flashback isn't needed anymore now that the new code especially can handle the CastKeyword

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -265,25 +265,13 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
...
-                copied.setChangedCardTypesCharacterDefining(c.getChangedCardTypesCharacterDefiningTable());
-                copied.setChangedCardNames(c.getChangedCardNames());
-                copied.setChangedCardTraits(c.getChangedCardTraits());
-                copied.setDrawnThisTurn(c.getDrawnThisTurn());
-
-                copied.copyChangedTextFrom(c);
-
-                // clean up changes that come from its own static abilities
-                copied.cleanupCopiedChangesFrom(c);
+                // try not to copy changed stats when moving to stack
```
tool4ever, line 268: Seems all Bestowed now disappear from Game on resolve 💩

Hanmac: i moved setSplitStateToPlayAbility up a bit, so i could skip a extra call to `setSplitStateToPlayAbility`

tool4ever (dismissed):
Couldn't break it again 👍

Hanmac:
i found a way to refactor the Riot checks without making it crash into StackOverflow

tool4ever (approved):

PR #2461, opened 2023-02-12
Delayed triggers vs. phased out

tool4ever:
This fixes phasing affecting delayed triggers.

E.g. activating _Angel of Condemnation_ first ability, then phasing him out would only let the return trigger fire after phase in again.


Hanmac (approved):
👍

PR #2478, opened 2023-02-15
CardFactoryUtil: move Disturb Keyword

Hanmac:
This is just a bit code cleanup

No Card can gain Disturb and it doesn't work with Split cards, so we can make CardFactoryUtil generate it which is a bit cleaner


Agetian (approved):

tool4ever:
👍

PR #2490, opened 2023-02-17
Prowl & support

tool4ever:
I tried to do it even cleaner but this seems the best possible, should be ok for now.


Agetian:
:+1:

kevlahnota (approved):

PR #2493, opened 2023-02-18
Adventure updates

jjayers99:
Adventure items can now spawn cards in command zone via startBattleWithCardInCommandZone - Updated Mana Shard based items to do so.

Added new map objects:  Mana Shard rewards, generic collision zone, intended for use in upcoming map updates.

Fixed mismatched wall tile in green castle.


forge-gui/res/adventure/Shandalar/custom_cards/cursed_treasure.txt
```diff
@@ -3,6 +3,7 @@ ManaCost:no cost
 Types:Artifact
 S:Mode$ Continuous | Description$ Provided by Cursed Treasure (Equipped Item - Right)
 StackDescription$ Create a Treasure token. You lose 2 life. | SpellDescription$ Create a Treasure token. You lose 2 life.
-A:AB$ Token | Cost$ PayShards<1> Sac<1/CARDNAME> | TokenScript$ c_a_treasure_sac | SubAbility$ DBLoseLife2 | SpellDescription$ Create a Treasure token.
+A:AB$ Token | Cost$ PayShards<1> T | ActivationZone$ Command |  TokenScript$ c_a_treasure_sac | SubAbility$ DBLoseLife2 | SubAbility$ Eject | SpellDescription$ Create a Treasure token.
```
kevlahnota, line 6: tap cost in command zone?

forge-gui/res/adventure/Shandalar/custom_cards/cursed_treasure.txt
```diff
@@ -3,6 +3,7 @@ ManaCost:no cost
...
 SVar:DBLoseLife2:DB$ LoseLife | LifeAmount$ 2 | Defined$ You
-Oracle: Provided by Cursed Treasure. Pay {M}, sacrifice Cursed Treasure: Create a Treasure token. You lose 2 life.
\ No newline at end of file
+Oracle:Pay {M}, exile Cursed Treasure: Create a Treasure token. You lose 2 life.
```
kevlahnota, line 8: exile is not part of the cost. it's part of the effect, so it should be like {M}: create treasure token. Ypu lose 2 life. Exile Cursed Treasure.

forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
```diff
@@ -20,6 +20,7 @@
 import forge.assets.FSkin;
 import forge.deck.Deck;
 import forge.deck.DeckProxy;
+import forge.deck.DeckSection;
```
kevlahnota, line 23: image

Agetian:
Some changes are needed there for the merge, and one more MR is waiting on this merge request. Is the original contributor available to make the necessary modifications?

forge-gui/res/adventure/Shandalar/custom_cards/cursed_treasure.txt
jjayers99: Intentional: Tap cost in command zone prevents the single use card from being triggered again in response to something else being placed on the stack. Card is placed in a tapped state (and will not ever untap without an ability specifically doing so) even though it doesn't appear that way on the UI, and thus that cost cannot be paid again.

{M}, Exile Example Item: Target creature gets a +1/+1 token.
Example item actual script: {M} {T}: Target creature gets a +1/+1 token. Exile Example Item.

Player activates, example item targeting a 3/3.
Opponent interrupts by targeting the 3/3 with a lightning bolt for 3 damage.
Player cannot spend an extra mana shard to put another copy of this ability on top of the stack and have the creature live, and it's the tap cost that prevents it.

forge-gui/res/adventure/Shandalar/custom_cards/cursed_treasure.txt
jjayers99: This was also intentional, although I understand if it needs to be changed.

{M}, Exile Example Item: Target creature gets a +1/+1 token.
Example item actual script: {M} {T}: Target creature gets a +1/+1 token. Exile Example Item.

Player activates, example item targeting a 3/3.
Opponent interrupts by targeting the 3/3 with a lightning bolt for 3 damage.
Player cannot spend an extra mana shard to put another copy of this ability on top of the stack and have the creature live, and it's the tap cost that prevents it.

In this case, having Exile (which needs to happen anyway to clean up the board) as a part of the cost explains to the player that this needs to be paid for the ability to go on the stack. It's not entirely accurate, but it's a lot shorter than adding to every card done in this style. "You may only activate this ability once per game and only if no other abilities of this card have already been activated."

kevlahnota (approved):

PR #2497, opened 2023-02-18
Cleanup some remembering + Specialize fix

tool4ever:


Hanmac:
wouldn't it be easier to use `Defined$ ExiledWithSource` instead of `Valid$ Card.ExiledWithSource`?

forge-gui/res/cardsfolder/w/wizards_spellbook.txt
```diff
@@ -1,11 +1,11 @@
...
+A:AB$ ChangeZone | Cost$ T | SorcerySpeed$ True | Origin$ Graveyard | Destination$ Exile | Mandatory$ True | ValidTgts$ Instant,Sorcery | TgtPrompt$ Select target instant or sorcery card in a graveyard | RememberChanged$ True | SubAbility$ DBRollDice | SpellDescription$ Exile target instant or sorcery card from a graveyard. Roll a d20. Activate only as a sorcery.
 SVar:DBRollDice:DB$ RollDice | Sides$ 20 | ResultSubAbilities$ 1-9:PlayCopy,10-19:PlayCost1,20:PlayAny
 SVar:PlayCopy:DB$ Play | Valid$ Card.IsRemembered+ExiledWithSource | ValidSA$ Spell | ValidZone$ Exile | Amount$ All | CopyCard$ True | Optional$ True | SubAbility$ DBCleanup | SpellDescription$ 1—9 VERT Copy that card. You may cast the copy.
 SVar:PlayCost1:DB$ Play | Valid$ Card.IsRemembered+ExiledWithSource | ValidSA$ Spell | ValidZone$ Exile | Amount$ All | CopyCard$ True | Optional$ True | PlayCost$ 1 | SubAbility$ DBCleanup | SpellDescription$ 10—19 VERT Copy that card. You may cast the copy by paying {1} rather than paying its mana cost.
```
Hanmac, line 7: 1-19 doesn't say `exiled with`  like 20 does
they just say "that card" so it should be enough to use Remember for them?

tool4ever:
> wouldn't it be easier to use `Defined$ ExiledWith` instead of `Valid$ Card.ExiledWithSource`?

Maybe, but it's not trustable enough yet because no links get verified as it just uses the exiled from host

Making that possible will be another work, but I still need to decide how exactly the class fields should be organized

Hanmac:
in theory, accessing ExiledWith from the Source Card should be faster than iterating over Cards in a Zone and check if their ExiledWith matches

Hanmac:
what happens when you haven't spezialized the card? you still can't cast it?
also i think after you spezialized, you shouldn't be able to play the normal side again?

tool4ever:
Yea both works
You can only cast the CardState that's currently enabled

kevlahnota (approved):

PR #2509, opened 2023-02-20
Fix color word usage

### LINKED ISSUE #134: Update cards checking for color words without oracle
cards like `Opal Titan` needs to be rewritten so they don't use Color Words in their effect because their oracle doesn't use them.

that causes them to be affected by Card Text Changes when they should not

maybe some investigation if there are more cards like this

### LINKED ISSUE #2511: Use Color Words on cards that can gain abilities
- `GainsAbilitiesActivateIgnoreColor$`
- `ManaConvert` keyword

tool4ever:
- [x] Fix XColor$

Closes #134
Closes #2511


Hanmac:
For XColor you need to be careful 

* Crimson Hellkite: `Spend only **red** mana on X.`
* Leyline Tyrant: `you may pay any amount of {R}` 
* Goblin Polka Band should be something different: `This ability costs {R} more to activate for each target.`
This is more of a RaiseCost Effect that should be defined on the activated SpellAbility?

tool4ever:
_Swirl the Mists_ already works fine with _Crypt Rats_ now.

Spells like _Drain Life_ however are unaffected, maybe a missing call somewhere. Will try to figure it out before I convert to followup

Hanmac:
> _Swirl the Mists_ already works fine with _Crypt Rats_ now.
> 
> Spells like _Drain Life_ however are unaffected, maybe a missing call somewhere. Will try to figure it out before I convert to followup

sounds like the changed color word stuff isn't copied/applied for spells
or i broke it again

tool4ever:
The changedWords got applied correctly, but the problem was it then only modified the Stack card - but the SA you're playing is the hand instance

Should be fixed

I don't think I'll do more stuff as part of this

kevlahnota (approved):

PR #2521, opened 2023-02-21
LandEvaluator for AI

kevlahnota:


forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -476,6 +477,8 @@ private Card chooseBestLandToPlay(CardCollection landList) {
             // If nothing is done here, proceeds to the default land picking strategy
         }
 
+        return ComputerUtilCard.getBestLandToPlayAI(landList);
```
tool4ever, line 480: I don't see such arbitrary mixing of heuristics and simulation as an improvement :/

the code you commented out provides (among other things) logic to play one drops or try to fix its colors

Whereas simulation will just score the card "blindly" and only makes sense if an evaluator mechanism is then able to check the results from that further.

If this is the way we want to evolve it imo it would be preferable to keep the superior logic intact and only call this instead if `landList.get(0)` would be reached

Agetian: I agree with this observation.

kevlahnota: thanks I update it and add it instead of landlist.get(0). thanks for the inputs.

tool4ever (approved):

PR #2524, opened 2023-02-21
Allow Defined and Targets to coexist on one effect

Northmoc:
I'm not in love with the Param name but can't think of anything better at the moment 😅 

A few targeting oopsies on these cards, too...


tool4ever (approved):
I would probably omit the "This" but that can be cleaned up later too

Hanmac:
@Northmoc another way would be to use like this:
`Defined$ Targeted,Self | ValidTgts$ Card.Other`

because most effects should use Defined over Targets if set?

Northmoc:
> @Northmoc another way would be to use like this: `Defined$ Targeted,Self | ValidTgts$ Card.Other`
> 
> because most effects should use Defined over Targets if set? Edit: ChangeZoneEffect is not one of them

Yeah, I've wished Defined$ could be split with `,` or something before

Thankfully this is fairly small number of cards and this band-aid will work until someone wants to do the definitive fix

PR #2527, opened 2023-02-21
ValidAfterStack for MayPlay

### LINKED ISSUE #2121: Serra Paragon allows casting X spells with MV>3
I had Serra Paragon and the game allowed me to cast Meathook Massacre from my graveyard for X=2 (MV 4), which shouldn't be allowed.

tool4ever:
Closes #2121


Hanmac:
Doesn't Lurrus needs the same piece of Code?

tool4ever:
Right, forgot all about him

Hanmac:
the ValidAfterStack should probably respect the Type too, like for Lurrus it needs to be a `Permanent+nonLand` i think it is currently checked by MayPlay, i just don't know how stable that is.

i probably need to touch it anyway when refactoring MayPlay

tool4ever:
Yea I checked, the last time Affected$ is checked is pretty close before Announcing

I don't think anything should change types between those points

Otherwise we'd need to try and parse the Affected$ of all MayPlay into a "ValidAfterStack$ Spell.*" param...? :/

Hanmac:
Currently the Affected of the MayPlay is checked there again:
https://github.com/Card-Forge/forge/blob/master/forge-game/src/main/java/forge/game/spellability/SpellAbilityRestriction.java#L276

may not be the cleanest way, but it works for now

its just that after MayPlay gets reworked we might think of something different

kevlahnota (approved):

PR #2540, opened 2023-02-23
GameAction: cleanup card on the battlefield gaining Keyword from castSA

### LINKED ISSUE #2488: Refactor Game Action Stack to Battlefield Keep Keyword
https://github.com/Card-Forge/forge/blob/751a213d6b77c9e25da83eba03f5c3995a6b4883/forge-game/src/main/java/forge/game/GameAction.java#L490-L516

this section should be refactored now that the Card on the Stack knows about CastSA#getKeyword ?
no need to extra look for the keyword in the changed keywords Table when we already got the instance ?

Hanmac:
closes #2488

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -32,7 +32,6 @@ import forge.game.card.*;
 import forge.game.event.*;
 import forge.game.keyword.Keyword;
 import forge.game.keyword.KeywordInterface;
-import forge.game.keyword.KeywordsChange;
 import forge.game.mulligan.MulliganService;
 import forge.game.player.GameLossReason;
 import forge.game.player.Player;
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -481,33 +480,10 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
...
-                        if (keepKeyword) {
-                            ki.setHostCard(copied);
-                            newKw.add(ki);
-                        }
-                    }
-                }
-                if (!newKw.isEmpty()) {
-                    copied.addChangedCardKeywordsInternal(newKw, null, false, copied.getTimestamp(), 0, true);
-                }
+            if (zoneFrom.is(ZoneType.Stack) && toBattlefield && c.getCastSA() != null && c.getCastSA().getKeyword() != null) {
```
tool4ever, line 483: It gets messed up somehow and will have Blitz twice :/

grafik

Hanmac: i make it only for non intrinsic ones (that should be Henzi)

tool4ever (approved):

PR #2578, opened 2023-02-27
Update custom card pics, NPE prevention on cardart

kevlahnota:


Simisays:
@kevlahnota These look awesome !  ButI've also buffed Jace's Signature Hoodie with a new effect. `Whenever you draw a card, each opponent mills a card.` Thought I might let you know

PR #2590, opened 2023-02-28
Seek effect + SeekAll trigger for new Arena stuff

Northmoc:
@Hanmac sorry for the duplicate work – dunno if it will be better for you to make comments on this or just copy stuff that your branch is missing?

Just need to refactor the rest of the old cards, I think...

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/SeekEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/SeekEffect.java
@@ -74,9 +74,14 @@ public class SeekEffect extends SpellAbilityEffect {
                     moveParams.put(AbilityKey.LastStateBattlefield, lastStateBattlefield);
                     moveParams.put(AbilityKey.LastStateGraveyard, lastStateGraveyard);
                     Card movedCard = game.getAction().moveToHand(c, sa, moveParams);
-                    soughtCards.add(movedCard);
+                    ZoneType resultZone = movedCard.getZone().getZoneType();
+                    if (!resultZone.equals(ZoneType.Library)) { // as long as it moved we add to triggerList
+                        triggerList.put(ZoneType.Library, movedCard.getZone().getZoneType(), movedCard);
+                    }
+                    if (resultZone.equals(ZoneType.Hand)) { // if it went to hand as planned, consider it "sought"
+                        soughtCards.add(movedCard);
+                    }
 
-                    triggerList.put(ZoneType.Library, movedCard.getZone().getZoneType(), movedCard);
                 }
             }
             if (!soughtCards.isEmpty()) {
--- a/forge-game/src/main/java/forge/game/ability/effects/SeekEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/SeekEffect.java
@@ -34,15 +34,8 @@ public class SeekEffect extends SpellAbilityEffect {
     @Override
     public void resolve(SpellAbility sa) {
         final Card source = sa.getHostCard();
-        List<Player> seekers;
-
-        if (sa.hasParam("Defined")) {
-            seekers = AbilityUtils.getDefinedPlayers(source, sa.getParam("Defined"), sa);
-        } else {
-            seekers = Lists.newArrayList(sa.getActivatingPlayer());
-        }
-
         final Game game = source.getGame();
+
         List<String> seekTypes = Lists.newArrayList();
         if (sa.hasParam("Types")) {
             seekTypes.addAll(Arrays.asList(sa.getParam("Types").split(",")));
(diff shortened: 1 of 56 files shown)


Hanmac:
i will give you my code for the Seek Effect later

as for your trigger, it doesn't check if Seek Num is 0 (or if the library is empty)

as for Defined Cards, is there one that doesn't use a ValidCard syntax but requires you to seek defined cards?

Hanmac:
Also my Code doesn't support multiple types like this yet: `Types$ Card.Land,Card.nonLand`

Northmoc:
> i will give you my code for the Seek Effect later
> 
> as for your trigger, it doesn't check if Seek Num is 0 (or if the library is empty)
> 
> as for Defined Cards, is there one that doesn't use a ValidCard syntax but requires you to seek defined cards? Ah the Stupid Trove Mage!

Sounds good – sorry again for the mixup 😬 

Good call on ways to insulate that trigger

Unsure about the DefinedCards case – I don't think so because Seek is random but I could definitely be wrong
May be better to just use a different param name on that?

forge-game/src/main/java/forge/game/ability/effects/SeekEffect.java
```diff
@@ -0,0 +1,85 @@
...
+            CardCollectionView lastStateBattlefield = game.copyLastStateBattlefield();
+            CardCollectionView lastStateGraveyard = game.copyLastStateGraveyard();
+            for (String seekType : seekTypes) {
+                CardCollection pool;
+                if (sa.hasParam("DefinedCards")) {
+                    pool = AbilityUtils.getDefinedCards(source, sa.getParam("DefinedCards"), sa);
+                } else {
+                    pool = new CardCollection(seeker.getCardsIn(ZoneType.Library));
+                }
+                pool = (CardCollection)AbilityUtils.filterListByType(pool, seekType, sa);
```
Hanmac, line 57: i personally hate `filterListByType` and prefer the real `CardLists.getValidCards`

i think filterListByType is from a time where we aren't that flexible and most stuff gets to be hardcoded

Northmoc: Sounds good to me – as you can probably tell, I just went through ChangeZoneEffect and copied the needed bits

Northmoc:
> Also my Code doesn't support multiple types like this yet: `Types$ Card.Land,Card.nonLand`

OK – we'll just patch together the best parts

forge-gui/res/cardsfolder/g/gitrog_horror_of_zhava.txt
```diff
@@ -6,8 +6,9 @@ K:Menace
 T:Mode$ Phase | Phase$ BeginCombat | TriggerZones$ Battlefield | Execute$ TrigSac | IsPresent$ Card.Self+untapped | TriggerDescription$ At the beginning of each combat, if CARDNAME is untapped, any opponent may sacrifice a nontoken creature. If they do, tap CARDNAME, then seek a land card and put it onto the battlefield tapped.
 SVar:TrigSac:DB$ Sacrifice | Defined$ Opponent | Amount$ 1 | SacValid$ Creature.nonToken | RememberSacrificed$ True | Optional$ True | AILogic$ DesecrationDemon | SubAbility$ DBTap
 SVar:DBTap:DB$ Tap | Defined$ Self | ConditionCheckSVar$ X | ConditionSVarCompare$ GE1 | SubAbility$ DBSeek
```
Hanmac, line 8: Branch Effect might be better for this?

Hanmac:
my start:

```java

    @Override
        final Game game = host.getGame();
        String changeType = sa.getParamOrDefault("ChangeType", "Card");

        CardCollectionView lastStateBattlefield = game.copyLastStateBattlefield();
        CardCollectionView lastStateGraveyard = game.copyLastStateGraveyard();

        int num = AbilityUtils.calculateAmount(host, sa.getParamOrDefault("Num", "1"), sa);

        if (num <= 0) {
            return;
        }
        
        final CardZoneTable triggerList = new CardZoneTable();
        
        for (Player p : getTargetPlayers(sa)) {
            if (!p.isInGame()) {
                continue;
            }
            CardCollectionView lib = p.getCardsIn(ZoneType.Library);
            if (lib.isEmpty()) {
                continue;
            }
            // only filter if changeType isn't Card
            if (!changeType.equals("Card")) {
                lib = CardLists.getValidCards(lib, changeType, sa.getActivatingPlayer(), host, sa);
            }

            // random n cards from library to hand
            for (Card c : Aggregates.random(lib, num)) {

                Map<AbilityKey, Object> moveParams = AbilityKey.newMap();
                moveParams.put(AbilityKey.LastStateBattlefield, lastStateBattlefield);
                moveParams.put(AbilityKey.LastStateGraveyard, lastStateGraveyard);
                Card movedCard = game.getAction().moveToHand(c, sa, moveParams);

                triggerList.put(ZoneType.Library, movedCard.getZone().getZoneType(), movedCard);

                if (sa.hasParam("RememberChanged")) {
                    host.addRemembered(movedCard);
                }
            }
        }

        triggerList.triggerChangesZoneAll(game, sa);
```

forge-game/src/main/java/forge/game/ability/effects/SeekEffect.java
```diff
@@ -0,0 +1,105 @@
...
+                }
+
+                CardCollection found = new CardCollection();
+                for (int i = 0; i < seekNum; i++) {
+                    Card c = Aggregates.random(pool);
+                    pool.remove(c);
+                    found.add(c);
+                }
+
+                for (final Card c : found) {
```
Hanmac, line 80: `for (final Card c : Aggregates.random(pool, seekNum)) {`
with that you don't need to manually get N random cards

Hanmac:
does the Seeked Card needs to be revealed to the player who is seeking? or is it enough to add the card to hand?

Northmoc:
> does the Seeked Card needs to be revealed to the player who is seeking? or is it enough to add the card to hand?

I think adding to hand is enough... when we used ChangeZone we didn't do a reveal, either.

forge-game/src/main/java/forge/game/ability/effects/SeekEffect.java
```diff
@@ -0,0 +1,94 @@
...
+
+                for (final Card c : Aggregates.random(pool, seekNum)) {
+
+                    Map<AbilityKey, Object> moveParams = AbilityKey.newMap();
+                    moveParams.put(AbilityKey.LastStateBattlefield, lastStateBattlefield);
+                    moveParams.put(AbilityKey.LastStateGraveyard, lastStateGraveyard);
+                    Card movedCard = game.getAction().moveToHand(c, sa, moveParams);
+                    soughtCards.add(movedCard);
+
+                    triggerList.put(ZoneType.Library, movedCard.getZone().getZoneType(), movedCard);
```
Hanmac, line 79: what if in a Commander Game, the commander is put into libary (and the player kept it there) and then it got seeked but THEN the player choose to put the Commander into Command Zone instead? then it shouldn't be put to hand?

>903.9b If a commander would be put into its owner’s hand or library from anywhere, its owner may put it into the command zone instead. This replacement effect may apply more than once to the same event. This is an exception to rule 614.5.

Hanmac: @Northmoc you need to take care if the card is not moved to hand because of some kind of replacement effect

Hanmac (approved):
👍 

@Agetian we might tweak AI later

PR #2605, opened 2023-03-02
Fix Revival Experiment via handlePaid cleanup

tool4ever:


Hanmac:
whats with RememberedAmount and why aren't the Players counted too?

was that a breaking change before?

tool4ever:
We still have `Count$RememberedSize` if we need to know about both

Otherwise it makes sense to have it only look at Cards
(For players we still have `PlayerCountRemembered)
`
Because the syntax was wrong on scripts like _Molten Psyche_ the calculation didn't get messed up

Hanmac:
i mean it more a long time ago it might have been done different, that's why the scripts needed Minus1?

tool4ever:
Yea, if it wasn't a plain bug a long time ago at least

kevlahnota (approved):

PR #2609, opened 2023-03-03
Playable Phyrexians Update

EldritchBimbo:
Adds 20 portraits, 2 animation sets for a playable Phyrexian race. (Parity with existing ones.) Adds lines to heroes.json and avatar.atlas to accommodate them. Edits avatar.atlas to reflect new avatar.png site.


kevlahnota:
I tested them and it works. It seems the attack animation is not very noticeable. since they look like a miniboss or something, I think a special attack animation like fireball, lightning, spark etc or anything to make the attack noticeable (don't know if any particle effects can be reused though and just trigger the effect when attacking).

kevlahnota:
> Good to know! If nothing else, i can look into tweaking the animation some to make it more visible.
> 
> Was the attack animation an issue in both cases, or only a particular one?

Both sprites needs tweaking, if the new sprites has some sort of weapon (wand or anything), that would be better visually.

EldritchBimbo:
Good to know, thanks for the feedback ^^
(And the advice, it's given me a direction to take things.)

kevlahnota:
Don't know why you close this since its only extra animation. If you don't have any extra animation available, you can call any particle effect upon attacking, but it will require some internal code and limit the use for phyrexian model for player.

PR #2611, opened 2023-03-04
Fix static trigger messing up LTB trigger

tool4ever:
image

- SBA for lethal damage -> ``runWaitingTriggers``
- the static trigger to remove the Effect card for the Dies RE would run
- LTB trigger would be lost

@Hanmac
here's the original commit:
https://github.com/Card-Forge/forge/commit/f990f1

even without knowing the full case I doubt you wanted to do it for static ones?


Hanmac (approved):
:+1:

PR #2620, opened 2023-03-05
GameFormat: filter Rebalanced cards

Hanmac:
This makes "A-" cards not legal in other formats other Arena ones,
while also stopping the Replaced Cards being legal in Arena

I'm unsure about the naming, or if i should put a cache around the checks?

do we need to update any format files? or did we forget to ban the cards respectively?

Also we might need to update the LDA data? I mean it does slightly check for the format, but maybe not always?

### DIFF
--- a/forge-core/src/main/java/forge/StaticData.java
+++ b/forge-core/src/main/java/forge/StaticData.java
@@ -900,4 +900,17 @@ public class StaticData {
         this.enableSmartCardArtSelection = isEnabled;
     }
 
+    public boolean isRebalanced(String name)
+    {
+        if (!name.startsWith("A-")) {
+            return false;
+        }
+        for(PaperCard pc : this.getCommonCards().getAllCards(name)) {
+            CardEdition e = this.editions.get(pc.getEdition());
+            if (e != null && e.isRebalanced(name)) {
+                return true;
+            }
+        }
+        return false;
+    }
 }
--- a/forge-core/src/main/java/forge/card/CardEdition.java
+++ b/forge-core/src/main/java/forge/card/CardEdition.java
@@ -375,7 +375,8 @@ public final class CardEdition implements Comparable<CardEdition> {
     public String getSheetReplaceCardFromSheet2() { return sheetReplaceCardFromSheet2; }
     public String[] getChaosDraftThemes() { return chaosDraftThemes; }
 
-    public List<CardInSet> getCards() { return cardMap.get("cards"); }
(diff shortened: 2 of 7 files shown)


paulsnoops:
No rebalanced cards were banned in any formats so we shouldn't need to change anything there.

Only ones that would need changing are Archived Arena Standard and Archived Explorer as although they use Subtype:Arena they don't use rebalanced/alchemy cards.

Maybe they can instead use:
(Explorer is based on Pioneer) Subtype:Pioneer
(Arena Standard) Subtype:Standard

Would that work?

kevlahnota:
can we filter all Alchemy cards as an option in the settings so they don't appear since some users don't like them... does Alchemy only legal in Magic Arena @paulsnoops ? Forge is not Arena (it just supports those cards). 
image

Hanmac:
also its currently checking the exact same set
so for reprints we might need to update the other sets

paulsnoops:
@kevlahnota yeah the only formats that rebalanced are legal in are Historic (which we support), Alchemy (which we support via Archived format filters only), and Historic Brawl (which we don't support). The only other Arena formats are the aforementioned Arena Standard and Explorer (both of which are Archived formats only)

kevlahnota:
> to get all Editions from for a CardName, to check if it is rebalanced in one of them? (i don't want to check for "A-" if able?)
> 
> should i try to use `CardDb.getAllCards(name)` to get all the PaperCard and then get the Editions or should i store the cardname => editions multimap somewhere else?

can we just update the editionid for alchemy sets, just add a special symbol if it's a unique alchemy card so you don't need to check all other editions? and just compare it by id? @paulsnoops maybe add "★" in the edition id .

Hanmac:
@kevlahnota that probably wouldn't work

my bigger problem is to check if a card is NOT rebalanced

tehdiplomat:
If the script itself is fresh can't we just have a special param and then we don't have to look inside edition files?

Hanmac:
> If the script itself is fresh can't we just have a special param and then we don't have to look inside edition files?

this doesn't have anything to do with the card scripts? or what do you mean?

paulsnoops:
On testing this branch it all seems to be working well at least for the deck editor using the format filters, can't speak for any other areas as I've not tested there but all looks good to me, thank you @Hanmac

paulsnoops (approved):

kevlahnota:
@Hanmac  is there anything left to do or this is ok to merge?

PR #2628, opened 2023-03-06
Add alchemy draft formats

pfirpfel:


pfirpfel:
A bit more information about the format:

For standard expansions, WotC seems to release a set of alchemy cards after a while on Arena. These consist of 5 mythics, 15 rares and 10 uncommons. At some point there is a alchemy version of that draft format, where a common gets replaced by a card from that alchemy sheet. That alchemy slot seems to follow rarity at least a little bit. From my observation, uncommons appear indeed most often.
For that slot, I added the new UncommonRareMythic slot. From my test drafts, it behaves more or less realistic.

| Rarity | # in set | factor | # in generated sheet |
| ------------- | ------------- | ------------- | ------------- |
| Mythic | 5 | 1 | 5 |
| Rare | 15 | 2 | 30 |
| Uncommon | 10 | 4 | 40 |

Agetian:
:+1:

paulsnoops (approved):

paulsnoops:
Looks good, thanks!

PR #2640, opened 2023-03-07
Fix to avoid incompatibility if compile with JDK9+

kevlahnota:


tool4ever:
@kevlahnota 
why this fix instead of making the right setting in the POM?

This will easier break again if they get used without this class somewhere else...

kevlahnota:
> @kevlahnota why this fix instead of making the right setting in the POM?
> 
> This will easier break again if they get used without this class somewhere else...

Laziness I guess, and this is called on Adventure mode only. Any devs that compiles Android and Mobile port will automatically set this to JDK8 or you'll fail producing an apk.

kevlahnota:
@tool4ever  If you have better solution than this feel free to revert and put a correct fix. Just looking at others how they did it though: https://jira.mongodb.org/browse/JAVA-2559 and some others here: https://issues.apache.org/jira/browse/MRESOLVER-85

PR #2644, opened 2023-03-08
Fix Make an Example

### LINKED ISSUE #2641: Make an example not working
**Describe the bug**
Make an example does not cause AI opponents to sacrifice their creatures.

**To Reproduce**
1. Play commander game
2. Have AI opponents play creatures
3. Play Make an Example
4. Select a pile, no matter which pile you select, AI opponents sacrifice NO creatures.

**Expected behavior**
The creatures in the chosen pile should be sacrificed

 - Device: Google Pixel 4a
 - OS: Android 13
 - Version 1.6.56-SNAPSHOT-03.06

tool4ever:
- Closes #2641

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
@@ -994,7 +994,7 @@ public class CountersPutAi extends CountersAi {
             types.add((CounterType)params.get("CounterType"));
         } else {
             for (String s : sa.getParam("CounterType").split(",")) {
-                CounterType.getType(s);
+                types.add(CounterType.getType(s));
             }
         }
 
--- a/forge-core/src/main/java/forge/StaticData.java
+++ b/forge-core/src/main/java/forge/StaticData.java
@@ -225,7 +225,7 @@ public class StaticData {
         return foil ? card.getFoiled() : card;
     }
 
-    public void attemptToLoadCard(String cardName){
+    public void attemptToLoadCard(String cardName) {
         this.attemptToLoadCard(cardName, null);
     }
     public void attemptToLoadCard(String cardName, String setCode) {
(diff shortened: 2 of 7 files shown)


forge-gui/src/main/java/forge/gui/card/CardDetailUtil.java
```diff
@@ -626,6 +617,12 @@ public static String composeCardText(final CardStateView state, final GameView g
                 area.append("\n");
             }
             area.append("Exiled until this leaves the battlefield: ").append(card.getUntilLeavesBattlefield());
+        } else if (card.getExiledCards() != null) {
```
Hanmac, line 620: that might be two different lists, the Card may have more than one way to exile cards,
while some of them might use UntilLeavesBattlefield, while some does not

> Angel of Condemnation {2}{W}{W}
> {2}{W}, {T}: Exile another target creature. Return that card to the battlefield under its owner’s control at the beginning of the next end step.

This is in ExiledCards:

> {2}{W}, {T}, Exert Angel of Condemnation: Exile another target creature until Angel of Condemnation leaves the battlefield. (An exerted creature won’t untap during your next untap step.)

This one would show the Until Leaves Battlefield

tool4ever: Hmn good point (though it also uses Remembered)

I didn't like the doubled text for cards like _Banishing Light_ but maybe it's too tricky to determine exactly

Will revert that part I guess

Hanmac: Bishop of Binding for example

The ETB Trigger uses UntilLeavesBattlefield, while the Attack Trigger uses ExiledCards info

Hanmac:
does it still Closes #2641  ? if yes add it to the description

kevlahnota (approved):

PR #2659, opened 2023-03-11
Maven: relocate platform-specific dependency

tool4ever:
Now GDX will be excluded from Desktop JAR.

### DIFF
--- a/forge-gui-desktop/pom.xml
+++ b/forge-gui-desktop/pom.xml
@@ -118,7 +118,7 @@
                         </goals>
                         <configuration>
                             <!-- TODO: insert placeholder for latest version tag -->
-                            <fromRef>forge-1.6.56</fromRef>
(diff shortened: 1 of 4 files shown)


forge-gui-desktop/pom.xml
```diff
@@ -118,7 +118,7 @@
                         </goals>
                         <configuration>
                             <!-- TODO: insert placeholder for latest version tag -->
-                            <fromRef>forge-1.6.56</fromRef>
+                            <fromRef>forge-1.6.55</fromRef>
```
Hanmac, line 121: was this change on purpose?

tool4ever: yes, that's the latest tag we have available so it needs to start from there

kevlahnota (approved):

PR #2670, opened 2023-03-13
Fix using old target from previous checks

tool4ever:
grafik

I think the conditional ``resetTargets`` is just too risky as shown by this case...

@Agetian

do you remember why you fixed it like that:

https://github.com/Card-Forge/forge/commit/0ca045

At that time "PowerDmg" was not supported by ``CountersPutAi`` but that is now taken care of, so during my tests this seems to be played correctly?

Also made path for _Crater's Claws_ work again.


Agetian:
Hmm, I don't remember the details of this particular instance, but generally speaking, in cases similar to this one the AI had issues targeting with the card (e.g. couldn't or didn't target with the subability), something along those lines. As long as it's tested to work at the moment and nothing seems off, I think it's a good call! :+1:

Agetian (approved):

PR #2677, opened 2023-03-14
Speed up London Mulligan when you just want to mulligan again.

tehdiplomat:
Selext first N cards for London Mulligan. 

This does not automatically mulligan again, you need to click Auto to select N cards. 
Then click Mulligan to roll the next round of mulligans.


Hanmac:
@tehdiplomat is the speedup still in the mulligan rules?

tehdiplomat:
@Hanmac It just chooses the first N cards to bottom on the bottom as mentioned in the description. That way if you mulled 3 times, you don't have to click on three cards and then click mulligan. You can just click auto to have 3 cards chosen for you, then you click mulligan.

Hanmac:
the N cards that gets tucked under the library and between you draw N-1 new Cards, it will be shuffled anyway?

Also the N cards should be the same amount in hand anyway, right?

PR #2683, opened 2023-03-14
LTR: Mount Doom + 2 cards

MikeS-NZ:
LTR: Add card scripts for spoiled LTR cards Mount Doom, Reprieve, and You Cannot Pass!

- Mount Doom
- Reprieve
- You Cannot Pass!


forge-gui/res/cardsfolder/upcoming/mount_doom.txt
```diff
@@ -0,0 +1,9 @@
+Name:Mount Doom
+ManaCost:no cost
+Types:Legendary Land
+A:AB$ Mana | Cost$ T PayLife<1> | Produced$ Combo B R | SpellDescription$ Add {B} or {R}.
+A:AB$ DealDamage | Cost$ 1 B R T | Defined$ Opponent | NumDmg$ 1 | SpellDescription$ CARDNAME deals 1 damage to each opponent.
+A:AB$ ChooseCard | Cost$ 5 B R T Sac<1/CARDNAME> Sac<1/Artifact.Legendary/legendary artifact> | Defined$ You | Amount$ 2 | Choices$ Creature | ChoiceTitle$ Choose up to two creatures | SorcerySpeed$ True | AILogic$ Duneblast | SubAbility$ DBDestroyAll | StackDescription$ {p:You} chooses up to two creatures. Destroy the rest. | SpellDescription$ Choose up to two creatures, then destroy the rest. Activate only as a sorcery.
+SVar:DBDestroyAll:DB$ DestroyAll | ValidCards$ Creature.nonChosenCard | StackDescription$ None
+DeckNeeds:Type$Artifact.Legendary
```
tool4ever, line 8: Unfortunately proper valid is not supported yet because this gets parsed at lower level, only `CoreType.SubType`

You could write it as printed in single string though that will miss some like "Legendary Enchantment Artifact"

MikeS-NZ: Thank you for feedback. I'm not sure exactly how to interpret the single string suggestion, so I've copied the OR operator convention with DeckNeeds tags from other card scripts. Let me know whether or not this is what you meant.

tool4ever: Hmn, no still wrong / inaccurate on the other

I think I'll add support for it

tool4ever (approved):

PR #2685, opened 2023-03-14
Remove some obsolete stuff

tool4ever:
This makes sure when canceling a SA the partial mana isn't sent back to pool in opposite order (which could lead to misplays when it has riders etc)


Hanmac:
about CanCounter, we should probably make it in a StaticAbility

tool4ever:
I gave it an effect so players can actually confirm in GUI that it was used 👍 

other KW can be refactored later

PR #2696, opened 2023-03-17
CardFactoryUtil: Equip shouldn't care if it is a creature

Hanmac:
found this while i was also doing the Unearth cost

rules wise an Artifact Equipment Creature with Equip can target itself with the Equip Ability
but of cause it doesn't has any effect (except you activated an ability)

@Agetian do we need more logic for AI so an Equipment Creature doesn't try to target itself?


tool4ever (approved):
It seems AI will check canBeAttached there anyway, so it should be safe

Agetian:
Yes, should be safe :)

PR #2699, opened 2023-03-17
CardFactory: remove redundant copying

tool4ever:
Both of these are already done recursively by copy() so we can save just a few calls here.

https://github.com/Card-Forge/forge/blob/395eab4d59a54a26988b71b011452da171486a4f/forge-game/src/main/java/forge/game/spellability/SpellAbility.java#L1156


Hanmac (approved):
:+1: 
its also copied for getCopiedTriggeredAbility

PR #2703, opened 2023-03-19
CardFactoryUtil: Read Ahead use DisableTriggers

### LINKED ISSUE #1489: Read Ahead Refactoring
The first change is that it doesn't just affect etb counter.

> 702.155a Read ahead is a keyword found on some Saga cards and represents two static abilities. The first ability is “As this Saga enters the battlefield, choose a number from one to this Saga’s final chapter number. This Saga enters the battlefield with the chosen number of lore counters on it.” **The second ability is “Chapter abilities of this Saga can’t trigger the turn it entered the battlefield unless it has exactly the number of lore counters on it specified in the chapter symbol of that ability.”** See rule 715, “Saga Cards.”

The Second Ability is something that stops Trigger from Triggering.

The easy way would be to add some parameter into the Chapter Trigger when it is from a Read ahead one. (we can't just use the Read Ahead value, because of Copy Effects. also we might not need that value anymore)

The more complex way would be to define a Static Ability that can block triggers.
With that we could refactor `noCreatureETBTriggers` and `noCreatureDyingTriggers` values. it probably needs to Look Back?

Also we might need some AI logic which Counter it should use if possible

Hanmac:
Closes #1489

I first thought about adding isValid to Trigger but then i thought it might be too complicated

the ReadAhead on PutCounter i keep for now, it might not be needed because you can get the Keyword from the SpellAbility anyway now, but it might be useful for AI to not choose TOO high or TOO low in case of counter doubler like Vorinclex

AI might need a bit more tweaking?

### DIFF
--- a/forge-game/src/main/java/forge/game/trigger/Trigger.java
+++ b/forge-game/src/main/java/forge/game/trigger/Trigger.java
@@ -613,4 +613,10 @@ public abstract class Trigger extends TriggerReplacementBase {
     public SpellAbility ensureAbility() {
         return ensureAbility(this);
     }
+
+    @Override
+    public void setOverridingAbility(SpellAbility overridingAbility0) {
+        super.setOverridingAbility(overridingAbility0);
+        overridingAbility0.setTrigger(this);
+    }
 }
--- a/forge-game/src/main/java/forge/game/ForgeScript.java
+++ b/forge-game/src/main/java/forge/game/ForgeScript.java
@@ -9,6 +9,7 @@ import forge.game.ability.AbilityUtils;
 import forge.game.ability.ApiType;
 import forge.game.card.Card;
 import forge.game.card.CardState;
+import forge.game.card.CounterEnumType;
 import forge.game.cost.Cost;
 import forge.game.mana.Mana;
 import forge.game.mana.ManaCostBeingPaid;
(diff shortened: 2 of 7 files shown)


forge-game/src/main/java/forge/game/trigger/Trigger.java
```diff
@@ -613,4 +613,10 @@ public SpellAbility ensureAbility(final IHasSVars sVarHolder) {
     public SpellAbility ensureAbility() {
         return ensureAbility(this);
     }
+
+    @Override
+    public void setOverridingAbility(SpellAbility overridingAbility0) {
+        super.setOverridingAbility(overridingAbility0);
+        overridingAbility0.setTrigger(this);
```
tool4ever, line 620: It kind of makes sense but I'm wondering if it could cause problems here:
https://github.com/Card-Forge/forge/blob/1186d4ee221730f1d6b12f60bde74e84b3cef6ba/forge-game/src/main/java/forge/game/card/ActivationTable.java#L27

I didn't think of a good test yet but does the root now lead to the copied Trigger so it can't find the original?

tool4ever (approved):
Logic seems fine and cleans it up nicely

I think for now I can offer these AI improvements as follow-up:

1. For Read Ahead always make AI choose first chapter instead (currently always defaults to last)

2. Perform check on first chapter before playing

With 1. I think the way the current cards are done it should be a pretty safe change to increase the overall value from playing it. I don't see trying to compare all chapters against each other in some generic way feasible considering we have no good solution to calculate their individual score (at that point)

For 2. this is arguably slightly more muddy because sometimes the other chapters might still make it worth to play, but generally speaking in most cases I'd say it should have better alternatives than not getting the trigger because of missing targets. Or even worse playing something like _The War in Heaven_ with empty library and losing on the spot :P

forge-game/src/main/java/forge/game/trigger/Trigger.java
Hanmac: imo that shouldn't have any effect i think

Trigger.setOverridingAbility is only called when creating the trigger

for example: Tested with `Omnath, Locus of Creation` and `Strionic Resonator` the ability still works as it should

PR #2704, opened 2023-03-19
Booster Draft: SIR

paulsnoops:
A note on the SIR draft format: As you can see on the link below, Arena has an evolving draft format that includes different cards from SIS on a weekly basis. As not all cards have rotated in yet there are no draft rankings available. Once all cards have a ranking I will probably come back to this and include them but with the following difference: because Forge's draft doesn't support the way Arena handles it, the draft will pick a SIS card at random to include in a pack; keep it simple.

https://magic.wizards.com/en/news/mtg-arena/shadows-of-the-past-card-list-and-schedule


Northmoc (approved):
Doesn't bother me if it isn't the same as current Arena – this is more for an archive.

If people want to play Arena they should just play Arena ;)

PR #2708, opened 2023-03-19
Music downloader

Simisays:
link to files (since I can't add them myself) https://www.dropbox.com/sh/zj6oe23a1c7ac5e/AABecDyE1WV-1U53_KithEdka?dl=0
I'm not sure if Forge will be able to move the downloaded files to the correct subfolder ?


kevlahnota:
why the desktop interface has the music downloader? Desktop launches seperately from mobile port.

forge-gui/res/languages/en-US.properties
```diff
@@ -2965,4 +2965,6 @@ lblEmblem=Emblem
 lblBoon=Boon
 lblExitToWoldMap=Exit to the World Map?
 lblWouldYouLikeDestroy=Would you like to destroy {0}?
-lblAdventureDescription=Adventure mode is where you explore the ever-changing landscape of Shandalar, and duel creatures to gain gold, shards, equipment and new cards to become the best and collect them all!
\ No newline at end of file
+lblAdventureDescription=Adventure mode is where you explore the ever-changing landscape of Shandalar, and duel creatures to gain gold, shards, equipment and new cards to become the best and collect them all!
+lblDownloadMusic=Download the custom Forge soundtrack.
+btnDownloadMusic=Download Custom Forge Soundtrack.
```
kevlahnota, line 2970: Cstom music and sounds resides in the cache, they need different paths and own labels since forge custom music sets and sounds are selectable, adventure has a single BGM set and there's no customization implemented (unless the user overwrite the actual music files or create a specific cache folder for adventure custom sounds)

Simisays: Thanks for the feedback. I'll just limit this PR to match and battle tracks for both Desktop and Mobile and will later try to get it wired up for adventure (mobile exclusive)

forge-gui/src/main/java/forge/localinstance/properties/ForgeConstants.java
```diff
@@ -364,6 +365,7 @@ public final class ForgeConstants {
     public static final String CACHE_SKINS_DIR               = CACHE_DIR + "skins" + PATH_SEPARATOR;
     public static final String CACHE_SOUND_DIR               = CACHE_DIR + "sound" + PATH_SEPARATOR;
     public static final String CACHE_MUSIC_DIR               = CACHE_DIR + "music" + PATH_SEPARATOR;
+    public static final String CACHE_MUSIC_ADVENTURE_DIR     = CACHE_DIR + "adventure" + PATH_SEPARATOR;
```
kevlahnota, line 368: I think there's no method to override the default directory for the adventure music, so even you put it in the cache for adventure directory (also it should be cache/adventure/music) it will not be played.

PR #2723, opened 2023-03-21
Adventure lootupgrade

Simisays:


forge-gui/res/adventure/Shandalar/world/enemies.json
```diff
@@ -5087,7 +5115,11 @@
 			"type": "deckCard",
 			"probability": 1,
 			"count": 2,
-			"addMaxCount": 4
+			"addMaxCount": 8,
+			"rarity": [
+				"common",
+				"uncommon"
+			]
```
jjayers99, line 5122: Combined with the 50% shot at 4 rares, this still feels a touch high, but I'm assuming this is intended to be one of the harder fights in its' zone.

jjayers99 (approved):
Looks a lot more consistent, and I like that there's generally a shift to focus on commons/uncommons.

forge-gui/res/adventure/Shandalar/world/enemies.json
Simisays: Yeah it'a a dungeon only enemy. I think there are only two of these in total

kevlahnota (approved):

PR #2726, opened 2023-03-22
CountersPutEffect: remove MaxFromEffect, let the CounterTable handle it

Hanmac:
in case of opponents Vorinclex, you could choose to put MORE counters on it, so when it is halved, you still get the amount you wanted

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
@@ -176,7 +176,6 @@ public class CountersPutEffect extends SpellAbilityEffect {
         final Player activator = sa.getActivatingPlayer();
         final PlayerController pc = activator.getController();
         final boolean etbcounter = sa.hasParam("ETB");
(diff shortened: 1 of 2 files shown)


tool4ever:
I also found this, either for here or as follow-up:
> If a number of dream counters would be put on Rasputin Dreamweaver for any reason that would bring the total number of dream counters on it to eight or more, the total number of dream counters on it is brought to seven instead.

Currently it's only handled as SBA, but that means it should also be checked inside `replaceCounterEffect` just like in the Clockworks case imo

Hanmac:
hm i need to think about Dreamweaver and the Dream counters. 
i would prefer to not have it as text keyword, but i will see what i can do.

> the total number of dream counters on it is brought to seven instead.

This is not part of the comprehensive rules, the rules only mention SBA, but i will check that more

Hanmac:
i also would like for the CantPutCounter Static Ability to use CantHappen Replacement, but it might be too tricky

tool4ever (approved):

PR #2733, opened 2023-03-23
A bunch of new enemies + removal of giant crab + fix Kiora Island (squash merge)

Simisays:
Should be good now, feel free to squash merge


forge-gui/res/adventure/Shandalar/sprites/dungeon/slimefoot.png
kevlahnota, line 1: can you update the sprite looking like a boss? it looks a recolored fungus though...

forge-gui/res/adventure/Shandalar/maps/map/teferi.tmx
kevlahnota, line 1: you removed the giant crab, does it cause crashes on old saves?

forge-gui/res/adventure/Shandalar/sprites/dungeon/slimefoot.png
Simisays: Well, Slimefoot is a fungus .... Here's a pic how it looks ingame 
image

Simisays: I can revert it back to the old sprite though I still think long term to move away from these custom sprites

Simisays:
I revered Slimefoot back, I'll think about a new sprite for the future

Simisays:
@kevlahnota Should be good to merge now ?

kevlahnota:
> @kevlahnota Should be good to merge now ?

@jjayers99 can you check this.. looks fine for me

jjayers99 (commented):
Rewards and spawn locations look fine. Decks look thematic. Sprite atlases are random enough that I need to review those further when possible, and I would like to do a visual review of the tile maps as well. Give me a bit to load these up locally and I'll get back to you on them.

Simisays:
> Sprite atlas issues:
> 
> Badger tail is clipping on both attack and death animation.
> 
> Dog has only idle and walk behaviors defined in atlas, plenty of frames in the sprite for the others. Clips slightly on the right side as it moves forward in the animation.
> 
> Fungus is clipping on right side during death animation and left side during attack, also includes a blank frame.
> 
> Jackal Warrior is clipping on both sides during idle, and I'm guessing you're cutting a frame off the top of what you've selected for the avatar
> 
> Jellyfish is clipping on bottom and left during death animation. Would recommend looking at using a frame where the tentacles aren't completely vertical for the avatar.
> 
> Mummy's atlas just needs to be redone from scratch, the animations don't line up on any of them and they all clip a lot. Would recommend looking at adding a couple pixels to the avatar size to show a couple lines of the torso, where it stops now looks awkward to me.
> 
> Octopus clips on bottom in one attack frame. Avatar could also stand to be larger vertically or have the water surface effect from the idle animation.
> 
> Owl is pretty good, but again I don't know that I like the avatar dimensions. Just looking at the avatar directly, that could just as easily be a spider as an owl. I almost feel like with some of these animals you might be better to try a full frame.
> 
> Plant is weird, but fine I guess. Again, would experiment with different avatars.
> 
> Raven walk / idle / avatar look great, death and attack both clip on right side.
> 
> Rat Ninja clips pretty badly on death. Idle looks like it's a couple pixels off from attack vertically which might cause it to look odd transitioning between the two. Same story again on the avatar, would experiment with showing a few more pixels vertically.
> 
> Scorpion clips on bottom in idle and death, on left during attack.
> 
> Stray cat clips on bottom during walk. Avatar looks... plain? For comedic effect I would look at the top of frame two of the attack animation, or maybe include a paw up like the top right area of attack frame 3.
> 
> Walking brain clips on the left during walk, and on both sides during death and attack. the walk clipping happens because you shift your indexing by two extra pixels between frames 2 & 3, which also makes the animation look like it's walking backwards.
> 
> Wild rat clips in every animation. Walk looks like it moves backwards, but it doesn't. Maybe it looks better when actually moving as opposed to animating in place. Walk looks lke it might be a little higher than attack, which is a pixel or two above idle, which is a pixel above death.

Thanks for your long review. I'll take them into account. Ideally we can just use 32x 32x for the size to remove all the clipping problems but that creates problems with collision range

jjayers99 (commented):
Don't know if these are existing issues, haven't looked at the maps before and tmx files don't do useful diffs.

Booster pack on teferi map is awarding 3 rare / mythic rare instead of 1, I doubt that's intentional.

Not a huge deal so long as the map functions as intended, but layers on the teferi and kiora map are a mess to the point where they would be a significant pain to edit if changes needed to be made.

jjayers99:
> Thanks for your long review. I'll take them into account. Ideally we can just use 32x 32x for the size to remove all the clipping problems but that creates problems with collision range

I honestly haven't experimented with it. But if 32x32 does create collision range issues, are some of what we're seeing as-is already from using whatever dimensions you're cropping all of these to?

Simisays:
> Don't know if these are existing issues, haven't looked at the maps before and tmx files don't do useful diffs.
> 
> Booster pack on teferi map is awarding 3 rare / mythic rare instead of 1, I doubt that's intentional.
> 
> Not a huge deal so long as the map functions as intended, but layers on the teferi and kiora map are a mess to the point where they would be a significant pain to edit if changes needed to be made.

Yeah that were some of my first dungeons before I got the hang of it how to make collision work. But the maps themselves should be fine

Simisays:
@jjayers99 updated the sprites according to your comments. I've checked them in game and they looked fine for me

kevlahnota:
@Simisays @jjayers99 is this ready?

Simisays:
@kevlahnota From my end it should be good to go. Sprites have been tweaked to prevent clipping (thanks to TTG comment)

jjayers99 (approved):

PR #2746, opened 2023-03-25
Gitaxian Dungeon, 2 Enemies

EldritchBimbo:
Adds:
1 dungeon based on the Progress Engine, tiles for such, and associated infrastructure. 
2 new enemies based on the same faction, along with deck .jsons and spritesheet atlases. 
Changes:
Edits basilica dungeon - replaces 1 of the treasure chests with a themed booster pack. 
Edit Phyrexian Buildings atlas and png to add new overworld sprite for new dungeon. 
Edits Phyrexiantiles to add requested extra tiles that were not covered in previous upload. 
Adds to blue, enemies, and POI jsons to make dungeon, enemies, and enemy loot pools exist and correctly function.


kevlahnota:
I think you need to rebase to resolve conflicting files

EldritchBimbo:
As far as I can tell, it should be good now - 2 commits were adding new enemies. I resolved it manually, so let me know if that wasn't the right thing to do, this is my first time dealing with a conflict. Thanks!

forge-gui/res/adventure/Shandalar/world/points_of_interest.json
```diff
@@ -1850,6 +1850,15 @@
 	"sprite": "BasilicaSmall",
 	"map": "maps/map/phyrexian_w1.tmx"
 },
+{
+	"name": "GitaxianLab",
+	"type": "dungeon",
+	"count": 2
```
Simisays, line 1856: missing comma

forge-gui/res/adventure/Shandalar/world/points_of_interest.json
```diff
@@ -1850,6 +1850,15 @@
 	"sprite": "BasilicaSmall",
 	"map": "maps/map/phyrexian_w1.tmx"
 },
+{
+	"name": "GitaxianLab",
```
Simisays, line 1854: Crash when entering the dungeon on my end. This is the error 
image

EldritchBimbo: Should be fixed - it was looking for booster.tx in my test environment instead of the filesystem itself for some reason.

forge-gui/res/adventure/Shandalar/world/points_of_interest.json
EldritchBimbo: Fixed, thanks!

Simisays:
Tested both dungeons and they look good imo! I really like your use of Booster packs being color restricted, makes the rewards you get feel fitting ! Though perhaps two little pieces of advice

1. Colission seems a bit off in both dungeons. Sometimes there is collision where there shoudn't be and the walls sometimes can be transgressed like this
image
2. This is just my opinion but I feel like custom .dck would be better in these kind of dungeon to make them fit to the specific phyrexian praeter. The decks of the enemies felt rather weak and a mishmash of various Phyrexian mechanics with little coherence. Perhaps a combination of Json and .dck using "RandomizeDeck" ?

EldritchBimbo:
Thanks for the heads-up. Did some tweaking, and at least wasn't able to reproduce that error on my end.

Glad people like the "themed booster" idea!  I feel like they're a good idea to expand some, and in different directions (similar to shops.) - If we expand the ways for players to semi-targetedly farm what they want (packs from specific sets, packs using specific mechanics, etc), players will be more likely to spend their time hunting down the dungeons that will give them a better chance, instead of farming tournaments for the spellsmith or rerolling shops over and over.

As for the decks, that's fair. (Admittedly, these are also the smaller and easier dungeons compared to planned later ones, I just decided to work my way up to that.) I'm loathe to _completely_ de-randomize them, since these kinds of cards are in the very strange place of being a playstyle that's easy to start in (2 of the 5 standard deck starting options feature phyrexian cards pretty heavily at minimum) but hard to get parts for after that, and I wanted to allow for some degree of semi-targeted farming. This is the first I'm hearing of the randomize option for .dck files - how exactly does that work? I'd would give them *some* core cards they always had, and the rest would fill out randomly?

jjayers99:
Check the Challenger enemy types in enemies.json for a deck randomization
example. Define multiple decks, and set the randomize decks attribute

EldritchBimbo:
Gotcha gotcha, thanks!
That doesn't quite do what I'd wanted, admittedly.
Hmm...I'd been intending to (assuming the schedule I'm aiming for holds) have time to get feedback on the enemies for each, and then add the lower tier enemies to the spawn list for their respective area around April 21st. (To be concurrent with the irl release of the set focusing around the invasion.) What would people think of leaving them randomized (maybe with a rarity buff to one or both?) until then, and once that happens, since the weaker enemies will be respawning at that point, reworking the stronger ones to be more like minibosses with their own deck.
Other options include:
. Just add the weaker enemies to the spawn pool and make the miniboss decks now.
. Something else.

Simisays:
> Gotcha gotcha, thanks! That doesn't quite do what I'd wanted, admittedly. Hmm...I'd been intending to (assuming the schedule I'm aiming for holds) have time to get feedback on the enemies for each, and then add the lower tier enemies to the spawn list for their respective area around April 21st. (To be concurrent with the irl release of the set focusing around the invasion.) What would people think of leaving them randomized (maybe with a rarity buff to one or both?) until then, and once that happens, since the weaker enemies will be respawning at that point, reworking the stronger ones to be more like minibosses with their own deck. Other options include: . Just add the weaker enemies to the spawn pool and make the miniboss decks now. . Something else.

Feel free to design your dungeons as you want, it was just a suggestion ! You can aways go back and change the decks if you have a change of heart

EldritchBimbo:
Fair enough!
I think, barring any other suggestions, I'll stick with the "enemies go into the overworld all at once when MOM drops, along with the dungeon bosses getting defined decks now that the reason not to is gone" idea. Thanks for the feedback!

kevlahnota:
@Simisays @jjayers99 is this ready?

Simisays:
@kevlahnota Looks good to me !

jjayers99 (approved):
Not having looked at the maps in much detail, I do wonder about the elements of GitaxianTilesheet, it doesn't look like they would line up quite as intended, leaving some corners missing or looking as though one wall is above another that it is supposed to adjoin to smoothly.

I appreciate the thought going in to timing the release of the enemies into the overworld. I also agree completely with the life totals of these enemies being a little higher than precedent for similar enemies, and think we should eventually go back and adjust _some_ of the older enemies to similar values.

PR #2748, opened 2023-03-25
further improvements to tibalt dungeon

Simisays:
better pathfinder for enemy patrols
tweaks to Tibalt
impoved teleporters between floor 1 and 2


forge-gui/res/adventure/Shandalar/world/enemies.json
```diff
@@ -10029,7 +10029,7 @@
 		"decks/torturer.dck"
 	],
 	"ai": "",
-	"flying": false,
+	"flying": true,
```
kevlahnota, line 10032: why it is flying, seems there's only single animation?

Simisays: It's basically to prevent the sprite getting stuck by clutter objects. The sprite also looks like it's floating so it shoudn't be that  strange ?

kevlahnota (approved):

PR #2767, opened 2023-03-27
Fix script

tool4ever:


forge-gui/res/cardsfolder/c/cemetery_prowler.txt
```diff
@@ -7,6 +7,6 @@ T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.S
 T:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigExile | Secondary$ True | TriggerDescription$ Whenever CARDNAME enters the battlefield or attacks, exile a card from a graveyard.
 SVar:TrigExile:DB$ ChangeZone | Origin$ Graveyard | Destination$ Exile | ChangeType$ Card | ChangeNum$ 1 | SelectPrompt$ Select a card from a graveyard | Mandatory$ True | Hidden$ True | AILogic$ ExilePreference:MostProminentNonLandNonExiledOwnType
 S:Mode$ ReduceCost | ValidCard$ Card | Type$ Spell | Amount$ AffectedX | Activator$ You | Description$ Spells you cast cost {1} less to cast for each card type they share with cards exiled with CARDNAME.
-SVar:AffectedX:Count$TypesSharedWith Card.ExiledWithSource
+SVar:AffectedX:Count$TypesSharedWith ValidExile Card.ExiledWithSource
```
Hanmac, line 10: @tool4ever wouldn't the Defined "ExiledWith" better than ValidExile ?

tool4ever: That still needs more rework to use all the logic for linked imo
So I try to avoid updating scripts with it for now

PR #2768, opened 2023-03-27
Magetower maps Revamp

Simisays:
- reorganised POI and enemies.json based alphabetically (this might help @jjayers99 tagging for quests)
- regarding the revamps i've tried to keep the layout the same (except with Mazetower 8 and 14) but just add more enemies rewards and exploration
- New item cycle based on the "staff of the X" cycle. I've made it so that in each dungeon there is 20% chance to drop them making it so that normally during every run one of these should drop
- Some enemy rebalance including doppelganger farming nerf


forge-gui/res/adventure/Shandalar/maps/main.tiled-session
jjayers99, line 1: No biggie, still not needed

Simisays: Yeah I'm still pretty new to Github and stuff. I added it to gitignore using the "Github for Desktop" app but it doesn't seem to listen ...

forge-gui/res/adventure/Shandalar/world/items.json
jjayers99, line 1: I strongly dislike all of the Staff of the X Magus items; specifically their drawbacks.

The Phyrexian Arena effect can be combined with Death Ring as an alternative to the Flame Sword that is about to be reduced in power.

I don't know that the AI will make use of Fervor without creature cards being explicitly tagged as PlayMain1. If so, this might be the closest to a balanced and usable item of the five, but I still don't love it.

Memory Erosion can be a very powerful aid to certain blue decks that want to amass instants/sorceries in their own graveyard - or a godsend for black decks looking to reanimate creatures from their graveyard.

Beastmaster Ascension at least feels like an actual drawback, but also likely far too much of one for players to want to use unless they have an explicit way up their sleeves to benefit from the +5/+5 boost such as Ensnaring Bridge or Meekstone.

Luminarch Ascension is a hard no as its effects that would be beneficial to the opponent can only be used if they have access to white mana.


I would much rather give out just the physical card to be included in decks, or a watered down version of the item that required Mana Shards rather than granting the opponent a card in play.

forge-gui/res/adventure/Shandalar/world/red.json
jjayers99, line 1: Is it intentional that we're removing Factory, Factory2, and Factory3 from the biome? That feels slightly more appropriate than some of the other removals, I just want to make sure it's intentional.

forge-gui/res/adventure/Shandalar/world/waste.json
jjayers99, line 1: Again, not disagreeing outright with the exclusion, but MageTowerC-C5 were intentionally placed in the waste originally (C representing colorless) - Is there a specific reason behind their removal?

forge-gui/res/adventure/Shandalar/world/enemies.json
jjayers99, line 1: Can you provide a summary of what you're changing here in addition to re-ordering the file? The diff view makes it almost impossible to follow, but I do see some attributes having been modified.

forge-gui/res/adventure/Shandalar/maps/tileset/main.png
jjayers99, line 1: Appreciate what you're doing here, but this is not the right home for the hexagram sprite. If you want to fix it now, just remove the sprite from the map that has an issue, we can replace it once we have a non-colliding version of buildings.tmx

forge-gui/res/adventure/Shandalar/world/waste.json
Simisays: The idea is that there are 5 "staf of the x" mage tower (one for each biome)
and then 3 or so other magetower dungeons, these can probably be in the waste zone

.gitignore
```diff
@@ -82,3 +82,5 @@ forge-gui/tools/EditionTrackingResults
 forge-gui/tools/PerSetTrackingResults
 
 *.tiled-session
+*.tiled-session
```
tool4ever, line 85: more of the same lines are useless

forge-gui/res/adventure/Shandalar/world/items.json
Simisays: I personally really like the mechanics where the item requires risk/reward thinking like how I've made them now. (you get a strong starting card but the enemy also). It's just cool design space that hasn't been explored yet

Though if you really don't like them I got no problem changing them to where you can conjure the staffs unto the battlefield by paying crystals

forge-gui/res/adventure/Shandalar/world/blue.json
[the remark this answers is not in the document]
Simisays: MageTowerU6 refers to a nonexistent magetower7_tmx
MageTower UD refers to magetower14_tmx. I've made this dungeon part of magetower 8

PR #2771, opened 2023-03-28
When importing/buying a Commander deck, add Commander to quest owned …

tehdiplomat:
…cards


kevlahnota:
Does this happens only on desktop version? Seems they have the same code.

tool4ever (approved):

PR #2777, opened 2023-03-29
tom_bombadil.txt + TriggerAbilityResolves

Northmoc:


tool4ever (approved):
Really wish we could avoid such a broad trigger all the time, but should be ok for now

Issue #2244, opened 2023-01-15
Cost merging needs additional logic


tool4ever:
If _Drought_ affects a spell with multiple black shards the resulting sacrifices should happen simultaneously (e.g. a single ChangesZoneAll trigger for _Nefarious Imp_).

 - But what about a spell like _Corpse Cobble_ that has another sacrifice cost? Their types aren't compatible, will that lead to two triggers instead then?
 - In either case sacrificing a manland for the Drought part should probably not increase the total power of the new token

Hanmac:
* _Rescue from the Underworld_ for example is something where the additional cost matters

Hanmac:
One idea: have the outer Cost class itself handle the bigger Trigger Lists

so the CostParts doesn't need to be merged, the All Trigger still works
and we could differ with the Paid lists if the Costs was paid using the cards own abilities (like the AddionalCost from cards like Rescue of the Underworld) or from other Cards like Drought

tool4ever:
A lot of the merging is for exotic cumulative upkeep.
I think I'll try to refactor that so it creates the full cost directly.

tool4ever:
Some AI parts also expect the merging to make better decisions.
E.g. multiple _Cultist of the Absolute_ causing Ward life payments

Hanmac:
hm maybe we need two different functions for merging costs? an extra one for ai?

tool4ever:
I found this:
![grafik](https://user-images.githubusercontent.com/8506892/228782517-82f89dd3-55a1-4c07-8ea9-c63f19c282d0.png)

So following that reasoning nothing seems to contradict these results:
- {B} or {BB} SA with Drought -> 1 ChangesZoneAll trigger
- Corpse Cobble with Drought -> 2 ChangesZoneAll

Therefore I am considering to close this after Part 2.

Issue #2472, opened 2023-02-14
Spell Refactoring Additional Cost


Hanmac:
Currently:
Spells that has "As an additional cost to cast this spell" like Discard a Card like `Tormenting Voice` use this as part of their cost for the spell

```
Name:Tormenting Voice
ManaCost:1 R
Types:Sorcery
A:SP$ Draw | Cost$ 1 R Discard<1/Card> | CostDesc$ As an additional cost to cast this spell, discard a card. | NumCards$ 2 | Defined$ You | SpellDescription$ Draw two cards.
```

Maybe better?
```
Name:Tormenting Voice
ManaCost:1 R
Types:Sorcery
S:Mode$ RaiseCost | ValidCard$ Card.Self | Activator$ You | Type$ Spell | Cost$ Discard<1/Card> | EffectZone$ All | Description$ As an additional cost to cast this spell, discard a card.
A:SP$ Draw | NumCards$ 2 | Defined$ You | SpellDescription$ Draw two cards.
```

it might help us for coping cards if the "intrinsic" cost of the spell is only its mana cost?

then we might better differ from which static ability the additional cost is coming from in case of "linked abilities"?

Hanmac:
A bit tricky might be a possible refactoring for optional additional costs like Dragon Presence:

```
Draconic Roar {1}{R}

Instant

As an additional cost to cast this spell, you may reveal a Dragon card from your hand.

Draconic Roar deals 3 damage to target creature. If you revealed a Dragon card or controlled a Dragon as you cast this spell, Draconic Roar deals 3 damage to that creature’s controller.
```

maybe a way to make RaiseCost Optional there?

tool4ever:
We especially need optional because there are 7 cards with "cost up to" (while it's ReduceCost it should probably work the same)

tool4ever:
@Hanmac
We also need to think about if we need extra static for this.
Technically "cost increase" and "additional cost" are not the same right?
But I'm not sure if there's a corner case where it would matter 🤔

Hanmac:
> @Hanmac We also need to think about if we need extra static for this. Technically "cost increase" and "additional cost" are not the same right? But I'm not sure if there's a corner case where it would matter 🤔


i think "cost increase" and "additional cost" should be the same. `S:Mode$ RaiseCost` should work for them

The corner case that might matter if that additional cost does something intrinsic to the card

Hanmac:
[...]

PR #2253, opened 2023-01-17
Cost tweaks

tool4ever:
This is a transition PR towards #2244.

It cleans up some parts and not considering AI checks either disables merging of costs not supported by the rules ~~or skips it for performance by setting the amount directly~~.

That does already fix the _Drought_ case causing sacrifices to be split into different payments.

Outside of CostAdjustment there could be other modifications happening where further undesired interactions might occur, but I'll investigate more after this one.

PR #2779, opened 2023-03-29
PaidLists now store if the CostPart source is linked to host

tool4ever:
Part for #2244:
Casting _Corbse Cobble_ via _Maestros Ascendancy_ now correctly only uses the stats from the original cost for the token P/T.

Some bonus cleanup especially in _CardUtil_, the removed parts turned redundant to ``SpellAbility.canTarget`` checks.

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardUtil.java
+++ b/forge-game/src/main/java/forge/game/card/CardUtil.java
@@ -35,7 +35,6 @@ import forge.card.MagicColor;
 import forge.game.CardTraitBase;
 import forge.game.Game;
 import forge.game.GameEntity;
-import forge.game.GameObject;
 import forge.game.ability.AbilityKey;
 import forge.game.ability.AbilityUtils;
 import forge.game.ability.ApiType;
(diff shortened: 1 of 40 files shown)


forge-game/src/main/java/forge/game/card/CardUtil.java
```diff
@@ -550,66 +549,21 @@ public static List<Card> getValidCardsToTarget(TargetRestrictions tgt, SpellAbil
...
                 choices.remove(activatingCard);
             }
         }
-        List<GameObject> targetedObjects = ability.getUniqueTargets();
 
         // Remove cards already targeted
         final List<Card> targeted = Lists.newArrayList(ability.getTargets().getTargetCards());
         choices.removeAll(targeted);
 
-        // Remove cards exceeding total CMC
```
Hanmac, line 569: why is that removed? i think that needs to be checked there too

Hanmac:
cost parts that was marked intrinsic and costs that aren't shouldn't be merged?

tool4ever:
There is one exception by the rules which is for _Carth the Lion_

but yes other than that ``CostAdjustment`` uses ``Cost.mergeTo()`` which I already implemented in #2253
It should make sure that all different CostParts are kept apart

(Unfortunately this temporarily slighty increased cheating potential via GUI canceling a partial payment after costs that can't refund but that needs to be fixed better in a different place anyway)

Hanmac:
For example: `Rescue from the Underworld`
> As an additional cost to cast this spell, sacrifice a creature.

i want this part to be a Static Ability on the card instead of `Cost$` 

then `Drought`
> Spells cost an additional “Sacrifice a Swamp” to cast for each black mana symbol in their mana costs.

so BOTH of them should be done as Static Abilities, but only the first one should be treated as intrinsic, the second should not.

so i mean this costs should not be merged (even if the valid part would be the same)

tool4ever:
For the refactoring we have this: #2472

But I'm not sure it's high priority because even without it's already working now

Part 1 (#2253) stopped the merging so they are paid as different cost parts like it should be
Part 2 (this) turns the paidLists into Table so linked abilities can filter for the right ones

kevlahnota (approved):

PR #2799, opened 2023-04-01
MOM: Backup keyword and cards

### LINKED ISSUE #2774: MOM: Backup Mechanic

Northmoc:
- closes #2774


forge-game/src/main/java/forge/game/keyword/Keyword.java
```diff
@@ -28,6 +28,7 @@ public enum Keyword {
     ASSIST("Assist", SimpleKeyword.class, true, "Another player can pay up to %s of this spell's cost."),
     AURA_SWAP("Aura swap", KeywordWithCost.class, false, "%s: You may exchange this Aura with an Aura card in your hand."),
     AWAKEN("Awaken", KeywordWithCostAndAmount.class, false, "If you cast this spell for %s, also put {%d:+1/+1 counter} on target land you control and it becomes a 0/0 Elemental creature with haste. It's still a land."),
+    BACKUP("Backup", KeywordWithAmountAndType.class, false, "When this creature enters the battlefield, put {%1$d:+1/+1 counter} on target creature. If that's another creature, it gains the following ability until end of turn."),
```
Hanmac, line 31: shouldn't it be just withAmount or does it mess up the description?

Northmoc: Crashes with `KeywordWithAmount` and Backup:2:**THIS**

Hanmac: Hmm I don't know If that's okay for now

I would prefer the Type ones only for the ones with types

@tehdiplomat your opinion?

forge-gui/res/cardsfolder/upcoming/archpriest_of_shadows.txt
```diff
@@ -0,0 +1,13 @@
+Name:Archpriest of Shadows
+ManaCost:3 B B
+Types:Creature Human Warlock
+PT:4/4
+K:Backup:BackupAbilities
```
tehdiplomat, line 5: Missing the number here

forge-game/src/main/java/forge/game/keyword/Keyword.java
tehdiplomat: Yea I feel like we'd need an extra Keyword class for this maybe `KeywordWithAmountAndOptions` or something like that, where we can just pass "extra things" that might get handled a bit different depending on the keyword.

Northmoc: There was already some support for "Extra" on KeywordWithAmount, so I expanded that

forge-gui/res/cardsfolder/upcoming/archpriest_of_shadows.txt
```diff
@@ -0,0 +1,14 @@
+Name:Archpriest of Shadows
+ManaCost:3 B B
+Types:Creature Human Warlock
+PT:4/4
+K:Backup:2:BackupAbilities
```
tool4ever, line 5: should still be 1?

Hanmac:
For `PlayMain1`: @Agetian should try to add AI logic that Ai should only play a Backup card as Main 1 if it has creatures that could attack?

Northmoc:
@Agetian I took a try at implementing the logic Hanmac suggested... see what you think 😄

Agetian:
Yep, should work nice :) :+1:

Agetian (approved):

PR #2814, opened 2023-04-02
DamageEachEffect refactor + MOM Tandem Takedown / MOC Great Aerie

Northmoc:


forge-game/src/main/java/forge/game/ability/effects/DamageEachEffect.java
```diff
@@ -39,12 +41,7 @@ protected String getStackDescription(SpellAbility sa) {
             sb.append(sa.getParam("StackDescription"));
```
tool4ever, line 41: This first if shouldn't be needed...
StackDescription would always override anyway?

forge-game/src/main/java/forge/game/ability/effects/DamageEachEffect.java
```diff
@@ -58,10 +55,20 @@ protected String getStackDescription(SpellAbility sa) {
     public void resolve(SpellAbility sa) {
         final Card card = sa.getHostCard();
         final Game game = card.getGame();
+        final String num = sa.getParamOrDefault("NumDmg", "X");
 
         FCollectionView<Card> sources = game.getCardsIn(ZoneType.Battlefield);
         if (sa.hasParam("ValidCards")) {
             sources = CardLists.getValidCards(sources, sa.getParam("ValidCards"), sa.getActivatingPlayer(), card, sa);
+        } else if (sa.hasParam("DefinedDamagers")) {
+            List<Card> defined = AbilityUtils.getDefinedCards(card, sa.getParam("DefinedDamagers"), sa);
```
Hanmac, line 64: wouldn't that just be sources = Defined? or just checking if Defined is inZoneBattlefield?

there is no need to get all cards in Battlefield, and then overwrite them from DefinedDamagers anyway

forge-gui/res/cardsfolder/b/band_together.txt
```diff
@@ -1,10 +1,6 @@
...
-SVar:DBEachDamage:DB$ EachDamage | ValidCards$ Creature.IsImprinted | NumDmg$ X | DamageDesc$ damage equal to its power | DefinedCards$ Remembered | SubAbility$ DBCleanup | StackDescription$ None
-#NumDmg isn't really used here. It is left for clarity. The AF pulls Damage straight from "X" hardcoded.
-SVar:X:Count$CardPower
-SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True | ClearImprinted$ True
+A:SP$ Pump | TargetMin$ 0 | TargetMax$ 2 | ValidTgts$ Creature.YouCtrl | TgtPrompt$ Select up to two target creatures you control | AILogic$ PowerDmg | SubAbility$ DBEachDamage | StackDescription$ {c:ThisTargetedCard} | SpellDescription$ Up to two target creatures you control
+SVar:DBEachDamage:DB$ EachDamage | DefinedDamagers$ ParentTarget | ValidTgts$ Creature | TgtPrompt$ Select another target creature | TargetUnique$ True | NumDmg$ Count$CardPower | StackDescription$ REP another target creature_{c:ThisTargetedCard} | SpellDescription$ each deal damage equal to their power to another target creature.
```
tool4ever, line 5: DamageDesc$ can be kept?

forge-gui/res/cardsfolder/l/living_inferno.txt
```diff
@@ -2,12 +2,7 @@ Name:Living Inferno
...
-SVar:Retribution:DB$ EachDamage | ValidCards$ Creature.IsRemembered | ValidDescription$ of those creatures | NumDmg$ X | DamageDesc$ damage equal to its power | DefinedPlayers$ Self | SubAbility$ DBCleanup | StackDescription$ then each of those creatures deals damage equal to its power to CARDNAME
-#NumDmg isn't really used here. It is left for clarity. The AF pulls Damage straight from "X" hardcoded.
-SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
-SVar:X:Count$CardPower
-SVar:FirePower:Count$CardPower
-SVar:Min:SVar$FirePower/LimitMax.1
+A:AB$ DealDamage | Cost$ T | ValidTgts$ Creature | TgtPrompt$ Select any number of target creatures | NumDmg$ Count$CardPower | TargetMin$ Min | TargetMax$ Count$CardPower | DividedAsYouChoose$ Count$CardPower | SubAbility$ Retribution | SpellDescription$ CARDNAME deals damage equal to its power divided as you choose among any number of target creatures.
```
tool4ever, line 5: Min broken now

forge-gui/res/cardsfolder/upcoming/tandem_takedown.txt
```diff
@@ -0,0 +1,6 @@
+Name:Tandem Takedown
+ManaCost:1 G G
+Types:Instant
+A:SP$ Pump | TargetMin$ 0 | TargetMax$ 2 | ValidTgts$ Creature.YouCtrl | TgtPrompt$ Select up to two target creatures you control | NumAtt$ +1 | ImprintCards$ ThisTargetedCard | AILogic$ PowerDmg | SubAbility$ DBEachDamage | StackDescription$ REP Up to two target creatures you control_{c:ThisTargetedCard} | SpellDescription$ Up to two target creatures you control each get +1/+0 until end of turn.
```
tool4ever, line 4: Imprint unneeded?

forge-gui/res/cardsfolder/l/living_inferno.txt
tool4ever, line 5: Is this handled?

> The amount of damage Living Inferno will deal is locked in as you activate its ability, since you’re dividing damage. If Living Inferno’s power changes after its ability is activated but before it resolves (via Giant Growth, for example), the amount of damage Living Inferno deals won’t change.

tool4ever (approved):

Hanmac:
is that ready to be merged or should we wait for the MOC card "the Great Aerie" with:
> each of these creatures deals damage equal to its toughness to the other.

tool4ever:
> is that ready to be merged or should we wait for the MOC card "the Great Aerie" with:
> 
> > each of these creatures deals damage equal to its toughness to the other.
> 
> or should we just use a Damage map?

Seems like a rare ability but at least _Grim Contest_ already needs it too.

I don't think the second damage effect could find the first target cleanly, so it would probably need remember too:

Targeting Pump -> DBDamage -> DBDamage -> DamageResolve -> DBCleanup

That does look a bit messy, so more engine support might still be worth for it

Hanmac:
ugh yeah `Grim Contest` needs to be rewritten like that, because it isn't an `Fight` for cards that care about fights

also then there is no `FightWithToughness` so that can be removed too

Northmoc:
Interesting... yeah, I'll try to get Grim Contest and the Great Aerie added to this PR

tool4ever (approved):
nice work

PR #2818, opened 2023-04-03
Fix incomplete replacement caching leading to StackOverflow

### LINKED ISSUE #2749: StackOverflow from Master Chef Replacement
Because the static is not connected to a KW the card will always get a new RE (with different ID)

1. Reverting https://github.com/Card-Forge/forge/pull/2449/commits/4734b02c0 works

2. another way would be doing ``newCopy.setStoredReplacementEffects(in.getStoredReplacementEffects())`` in `getLKICopy` and then also later when the Affected is updated

tool4ever:
Closes #2749

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -4401,6 +4401,17 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
         return result;
     }
 
+    public void setStoredReplacements(Table<StaticAbility, String, ReplacementEffect> table) {
+        storedReplacementEffect.clear();
+        for (Table.Cell<StaticAbility, String, ReplacementEffect> c : table.cellSet()) {
+            storedReplacementEffect.put(c.getRowKey(), c.getColumnKey(), c.getValue().copy(this, true));
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4571,7 +4582,28 @@ public Table<Long, String, KeywordInterface> getStoredKeywords() {
...
+            if (lki) {
+                for (Table.Cell<Long, Long, KeywordsChange> kwCell : changedCardKeywords.cellSet()) {
+                    // same static
+                    if (kwCell.getColumnKey().equals(c.getRowKey())) {
+                        for (KeywordInterface kw : kwCell.getValue().getKeywords()) {
+                            // same kw
+                            if (kw.getOriginal().equals(c.getValue().getOriginal())) {
+                                stored = kw;
+                            }
+                        }
```
Hanmac, line 4597: i don't think that is always true what if the card would give multiple instances of the same keyword?

tool4ever: L4591 does check that the static IDs are same too

I mean otherwise ``getKeywordForStaticAbility()`` would have a problem too?

kevlahnota (approved):

PR #2827, opened 2023-04-05
PumpEffect: make DefinedKW better Keywords

Hanmac:
The first Reason was for Skrelv and similar making better Hexproof from Color effects
the second one was cleaning up some of the Protection from Player stuff

The Effects might need better Stack Description

And Skrelv still needs better AI


forge-gui/res/cardsfolder/n/noble_heritage.txt
```diff
@@ -9,6 +9,6 @@ SVar:DBChoice:DB$ ChooseCard | Defined$ Player.IsRemembered | Amount$ 1 | MinAmo
 SVar:DBPutCounter:DB$ PutCounterAll | ValidCards$ Creature.IsRemembered | CounterType$ P1P1 | CounterNum$ 2 | SubAbility$ DBRememberPlayer
 SVar:DBRememberPlayer:DB$ Pump | RememberObjects$ RememberedController | SubAbility$ DBRepeatEach
 SVar:DBRepeatEach:DB$ RepeatEach | RepeatPlayers$ Opponent.IsRemembered | RepeatSubAbility$ DBPump | SubAbility$ DBCleanup
```
tool4ever, line 11: Imo `RepeatPlayers$ RememberedController` should work so you can at least get rid of the Pump above there

Nice work on getting rid of TiedToKeyword, I was suspecting it was no longer needed

Hanmac: For this, PumpEffect should probably be extended to turn one Keyword into multiple extra ones like StaticAbility can

Hanmac:
I'm going to cleanup `DefinedKW` even more, there are only 4 stages:
* ChosenType
* ChosenPlayer
* ChosenColor
* anything else is definedPlayer (so i don't need to check for "you" or ends with "Player")

tool4ever (approved):
looks good

PR #2836, opened 2023-04-05
Playable Werewolf Update

EldritchBimbo:
Additions:

.Adds 20 portraits, 2 animation sets for playable werewolf race, as well as infrastructure.

Changes:

.Modifies avatar.png to add portraits.
.Modifies avatar.atlas and heroes.json to accommodate new options.


jjayers99 (commented):
No obvious structural problems. Will review animation when able tomorrow.

The avatars are fine in terms of  design and the ability to have avatars of either state is a nice touch. There is a reasonably noticable difference in levels of shading compared to other races - It doesn't have to be changed now, but that's the only place I can see where I would want any change at all (within current framework, see below).

For the movement sprite, the contrast between the red color clothing in the human state with the blue / green eyes in wolf form is a particularly nice touch. 

I especially like the transformations included in the female sprite. It's not available in our current framework, but In a long term sense I would like to see this (along with the npc vampire sprite) have an active transition phase between idle and walking states. Or, alternatively, to integrate the state of the transformation with the day/night code.

jjayers99 (approved):
Seeing the sprites in the game makes me want a transition phase much more. I'm not in love with it, but there's nothing particularly wrong with it either. No reason not to approve.

EldritchBimbo:
Yeah, I'd definitely agree with a transition phase being better - I'd experimented with the first frame of the walking animation being an in-between (similar to one of the frames in the female defeat animation) but that resulted in her partially turning back every couple frames, so it just looked bad.

kevlahnota (approved):

PR #2845, opened 2023-04-06
Fix Gorex

### LINKED ISSUE #2842: Gorex, the Tombshell not pulling cards from Exile
Gorex, the Tombshell's attack/death trigger does not bring one of the exiled cards from his ability back to the owner's hand, they just stay exiled.

tool4ever:
Closes #2842


Agetian:
Looks good to me, but I'm not the best at in-depth mechanics, so I'm not sure if any side effects are possible here... ^^;

Agetian (approved):

PR #2849, opened 2023-04-07
More Incubate

Northmoc:


forge-game/src/main/java/forge/game/ability/effects/IncubateEffect.java
```diff
@@ -43,10 +43,11 @@ protected String getStackDescription(SpellAbility sa) {
 
     @Override
     public void resolve(SpellAbility sa) {
-        final Card card = sa.getHostCard();
-        final Game game = card.getGame();
-        final Player activator = sa.getActivatingPlayer();
-        final int times = AbilityUtils.calculateAmount(card, sa.getParamOrDefault("Times", "1"), sa);
+        final Card host = sa.getHostCard();
+        final Game game = host.getGame();
+        final Player activator = sa.hasParam("Defined") ?
```
Hanmac, line 48: Use getTargetPlayers for a loop?

forge-gui/res/cardsfolder/upcoming/corruption_of_towashi.txt
```diff
@@ -0,0 +1,14 @@
+Name:Corruption of Towashi
+ManaCost:4 U
+Types:Enchantment
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigIncubate | TriggerDescription$ When CARDNAME enters the battlefield, incubate 4. (Create an Incubator token with four +1/+1 counters on it and "{2}: Transform this artifact." It transforms into a 0/0 Phyrexian artifact creature.)
+SVar:TrigIncubate:DB$ Incubate | Amount$ 4
+T:Mode$ Transformed | ValidCard$ Permanent.YouCtrl | Execute$ TrigDraw | TriggerZones$ Battlefield | CheckSVar$ X | SVarCompare$ LT1 | OptionalDecider$ You | TriggerDescription$ Whenever a permanent you control transforms or a permanent enters the battlefield under your control transformed, you may draw a card. Do this only once each turn.
```
tool4ever, line 6: StoreSVar shouldn't be needed, we have `ResolvedLimit$ 1` for it

forge-gui/res/cardsfolder/upcoming/elvish_vatkeeper.txt
```diff
@@ -0,0 +1,11 @@
+Name:Elvish Vatkeeper
+ManaCost:1 B G
+Types:Creature Phyrexian Elf
+PT:3/3
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigIncubate | TriggerDescription$ When CARDNAME enters the battlefield, incubate 2. (Create an Incubator token with two +1/+1 counters on it and "{2}: Transform this artifact." It transforms into a 0/0 Phyrexian artifact creature.)
+SVar:TrigIncubate:DB$ Incubate | Amount$ 2
+A:AB$ SetState | Cost$ 5 | TgtPrompt$ Select target Incubator token you control | ValidTgts$ Incubator.token+YouCtrl | Mode$ Transform | SubAbility$ DBDouble | SpellDescription$ Transform target Incubator token you control.
+SVar:DBDouble:DB$ MultiplyCounter | Defined$ Targeted | CounterType$ P1P1 | StackDescSpellDescription$ Double the number of +1/+1 counters on it.
```
tool4ever, line 8: StackDescSpellDescription isn't a thing right?

forge-gui/res/cardsfolder/upcoming/excise_the_imperfect.txt
```diff
@@ -0,0 +1,9 @@
+Name:Excise the Imperfect
+ManaCost:1 W W
+Types:Instant
+A:SP$ ChangeZone | Origin$ Battlefield | Destination$ Exile | ValidTgts$ Permanent.nonLand | TgtPrompt$ Select target nonland permanent | RememberLKI$ True | SubAbility$ DBIncubate | SpellDescription$ Exile target nonland permanent.
+SVar:DBIncubate:DB$ Incubate | Defined$ RememberedController | Amount$ X | SubAbility$ DBCleanup | SpellDescription$ Its controller incubates X, where X is its mana value. (They create an Incubator token with X +1/+1 counters on it and "{2}: Transform this artifact." It transforms into a 0/0 Phyrexian artifact creature.)
```
tool4ever, line 5: I think you need `RememberedLKIController` or you might still get the wrong one

forge-gui/res/cardsfolder/upcoming/corruption_of_towashi.txt
Northmoc: Ah, since the overriding ability is the same for both triggers – NICE

Hanmac: I'm not 100% sure this is fool proof

I mean yeah it's the same SVar, but when the trigger are built, isn't it creating two SpellAbility objects out of it?

tool4ever: Ah good point, now I'm not so sure
It might be that even if the script uses the same SVar the triggers will not share it :/

tool4ever: I guess we can't do it

We could teach CardFactory to reuse the SA - but some other cards will have two triggers that are separate abilities which might still work with the same SVar

So it would require manual tagging and that seems too fiddly imo

Sorry about that

Northmoc: Hey no worries – I actually meant to ask about this as I wasn't sure the SVars were good enough

Maybe there's some way to  link triggers of different trigger types for this sort of thing – not this PR though

Northmoc: It did seem to be working in testing... I was doing stuff with the new Elesh Norn so lots of opportunities for both of the triggers to go off.

tool4ever: You're right it works

Need to debug why :D

tool4ever: The problem is the Transformed property fails because for some reason the card is still in OriginalState

So your ChangesZone never triggers but your Transformed actually does for Elesh Norn too because of missing `+inZoneBattlefield` most likely

Northmoc: Oops, I see, it's all fixing Corruption (as you know)

tool4ever (approved):

PR #2859, opened 2023-04-07
see_double.txt and support

Northmoc:


tool4ever (approved):
Possibly CopySpell Replacement could be reused but I'm not sure it's worth the refactoring trouble, shouldn't be a showstopper for now

PR #2860, opened 2023-04-07
Dynamically-updating search post fix

bbp9857:
Search posts weren't being updated except for when loading the game.  They now are updated dynamically.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/stage/MapSprite.java
+++ b/forge-gui-mobile/src/forge/adventure/stage/MapSprite.java
@@ -8,6 +8,7 @@ import com.badlogic.gdx.utils.Array;
 import com.github.tommyettinger.textra.TextraLabel;
 import forge.adventure.data.BiomeSpriteData;
 import forge.adventure.pointofintrest.PointOfInterest;
+import forge.adventure.pointofintrest.PointOfInterestChanges;
 import forge.adventure.util.Controls;
 import forge.adventure.world.WorldSave;
 import org.apache.commons.lang3.tuple.Pair;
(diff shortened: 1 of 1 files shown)


kevlahnota:
@bbp9857  this was designed passively so it will perform fast on android devices, did you try to test this on entry-tier/mid-tier android since you move the check on draw (the check always run on every draw render). I don't think you check other usages for checkOut() method since it is called elsewhere.

forge-gui-mobile/src/forge/adventure/stage/MapSprite.java
```diff
@@ -53,13 +50,7 @@ public static Array<Actor> GetMapSprites(int chunkX, int chunkY) {
...
         List<Pair<Vector2, Integer>> objects = WorldSave.getCurrentSave().getWorld().GetMapObjects(chunkX, chunkY);
 
-        for (Pair<Vector2, Integer> entry : objects) {
-            BiomeSpriteData data = WorldSave.getCurrentSave().getWorld().getObject(entry.getValue());
-            if (data.layer != SpriteLayer)
-                continue;
-            Actor sprite = new MapSprite(entry.getKey(), WorldSave.getCurrentSave().getWorld().getData().GetBiomeSprites().getSprite(data.name, (int) entry.getKey().x + (int) entry.getKey().y * 11483), null);
-            actorGroup.add(sprite);
-        }
+        actorGroup.addAll(GetMapSpritesBackground(chunkX, chunkY));
```
kevlahnota, line 53: spritelayer is different from background layer, why you remmove this? if you want to remove duplicate code, just update the method to differentiate spritelayer and backgroundlayer

bbp9857:
All of the calls are o(1), but there is admittedly the cost of the additional stack frames.  I did verify that checkOut was unused.

forge-gui-mobile/src/forge/adventure/stage/MapSprite.java
bbp9857: All of the calls are o(1), but there is admittedly the cost of the additional stack frames. I did verify that checkOut was unused.

forge-gui-mobile/src/forge/adventure/stage/MapSprite.java
```diff
@@ -82,9 +73,12 @@ public void draw(Batch batch, float parentAlpha) {
         if (texture == null)
             return;
         batch.draw(texture, getX(), getY());
-        if (isCaveDungeon && !isOldorVisited) {
-            searchPost.setPosition(getX() - 7, getY() + 7);
-            searchPost.draw(batch, parentAlpha);
+        if (isCaveDungeon && point.getData().map != null && point.getID() != null){
```
kevlahnota, line 76: the update should be either on leave or entry of Poi on worldstage not here

forge-gui-mobile/src/forge/adventure/stage/MapSprite.java
kevlahnota: my bad the worldstage is not updated on master but was updated on snapshot version which it is used.. gonna commit it now
image

PR #2861, opened 2023-04-08
Dross Pit Content Update

EldritchBimbo:
Additions:
.2 new enemies, based on the Dross Pit Phyrexians, plus infrastructure. (Both use .json deck generation, for now) 
.Additional level tiles based on Dross Pit aesthetic, plus infrastructure. 
.Additional overworld tile for new dungeon, plus infrastructure. 
.1 new dungeon, utilizing new enemies and tiles.
Changes:
.Additions to Phyrexian_Buildings .atlas and .png files to accommodate new overworld tile. .Addition to black.json to allow new dungeon to spawn. 
.Addition to points_of_interest.json to create new dungeon entry.
.Addition to enemies.json to create enemy entries for new enemies.


Simisays:
@EldritchBimbo Cool dungeon ! though make sure the collision is working as intended 
image

As a general aside, are you intending for every dungeon to have a unique tileset file ? Maybe it might be best to merge these in a general "Phyrexian" tileset to prevent having too much little tilesets, in that way it would allow other contributers to make new phyrexian dungeon using the same tileset. I'm actually in the process of creating two new tilesets (based on the main one) consisting of "Farm and Crops" and "Dungeon props and tiles" for easier acces for creating dungeons. Just FWY

jjayers99:
> As a general aside, are you intending for every dungeon to have a unique tileset file ? Maybe it might be best to merge these in a general "Phyrexian" tileset to prevent having too much little tilesets, in that way it would allow other contributers to make new phyrexian dungeon using the same tileset.

The separate tilesheets were my suggestion. There was concern over creating conflicting versions of tileset sprites as we weren't sure exactly what the timing was going to be on creating these, plus I knew that the images would probably take several iterations when being built from scratch. I volunteered to merge these all into the master tileset and update the dungeons to use them once EldritchBimbo was satisfied and felt that they were complete.

EldritchBimbo:
Appreciate the heads-up on the collision. That didn't happen in my testing, I'll have to double -check things.

EldritchBimbo:
Was able to reproduce the collision issue once i knew what to look for, should be mostly hammered out. (The player is still able to look like they're standing partially in the Necrogen on the bridge tiles, which isn't intended, but I haven't been able to find a collision tile that blocks just that part of their movement off. )
The patrols are a separate problem, and as far as I can tell not one I have power over - when you exit a patrolling enemy's threat radius, they try to head directly to the next waypoint they _would_ have gone to, and if they can't reach it they'll just walk into a wall forever trying.

jjayers99:
There is a timeout system for patrolling enemies, where any patrolling
enemy will give up on its target waypoint as unreachable and move on to
attempting the next one. It's not perfect, but it should allow for recovery
from basic collisions close to the original path if given sufficient time
to cycle through options.

I do think this can be improved upon in the future without much overhead by
adding a sort of exploration behavior to look for reachable waypoints, but
as with much of what I have added so far, mobile enemies in their current
state were intended only as a prototype / proof of concept.

EldritchBimbo:
Oh, that's good to know! Is that an automatic behaviour, or something I'd have to enable?

jjayers99:
That's automatic, and currently not configurable. There had been some
preliminary discussion of implementing a far better pathfinding system, so
I didn't put a huge amount of effort into cleaning up this version yet.

kevlahnota:
is this ready?

EldritchBimbo:
From my perspective at least, yes-I've acted on the feedback that needed it, as far as I'm aware.

kevlahnota (approved):

PR #2865, opened 2023-04-08
CounterAddedOnce add FirstTime

### LINKED ISSUE #2784: Botanical Brawler
> Whenever one or more +1/+1 counters are put on another permanent you control, if it's the first time +1/+1 counters have been put on that permanent this turn, put a +1/+1 counter on Botanical Brawler. 

try to use `getGame().getCounterAddedThisTurn` for this to check if it would be the first time counters where added on this card.

might use a different function if `Card.StrictlySelf` doesn't work for valid String

Hanmac:
Closes #2784

This adds `FirstTime` to the AddedOnce Trigger

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/botanical_brawler.txt
+++ b/forge-gui/res/cardsfolder/upcoming/botanical_brawler.txt
@@ -1,0 +1,9 @@
+Name:Botanical Brawler
+ManaCost:G W
+Types:Creature Elemental Warrior
+PT:0/0
+K:Trample
+K:etbCounter:P1P1:2
+T:Mode$ CounterAddedOnce | ValidCard$ Card.Other+inZoneBattlefield | TriggerZones$ Battlefield | CounterType$ P1P1 | FirstTime$ True | Execute$ TrigPutCounter | TriggerDescription$ Whenever one or more +1/+1 counters are put on another permanent you control, if it’s the first time +1/+1 counters have been put on that permanent this turn, put a +1/+1 counter on CARDNAME.
+SVar:TrigPutCounter:DB$ PutCounter | CounterType$ P1P1 | CounterNum$ 1
+Oracle:Trample\nBotanical Brawler enters the battlefield with two +1/+1 counters on it.\nWhenever one or more +1/+1 counters are put on another permanent you control, if it’s the first time +1/+1 counters have been put on that permanent this turn, put a +1/+1 counter on Botanical Brawler.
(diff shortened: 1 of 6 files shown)


tool4ever (changes requested):
Code looks fine - but I got Brawler to trigger from the ETB counters of _Fertilid_ and then also a second time on putting a counter on it after with _Simic Ascendancy_ (really shouldn't matter what gets used there)

Not a third time though

forge-gui/res/cardsfolder/upcoming/botanical_brawler.txt
```diff
@@ -0,0 +1,9 @@
+Name:Botanical Brawler
+ManaCost:G W
+Types:Creature Elemental Warrior
+PT:0/0
+K:Trample
+K:etbCounter:P1P1:2
+T:Mode$ CounterAddedOnce | ValidCard$ Card.Other+YouCtrl+inZoneBattlefield | TriggerZones$ Battlefield | CounterType$ P1P1 | FirstTime$ True | Execute$ TrigPutCounter | TriggerDescription$ Whenever one or more +1/+1 counters are put on another permanent you control, if it’s the first time +1/+1 counters have been put on that permanent this turn, put a +1/+1 counter on CARDNAME.
+SVar:TrigPutCounter:DB$ PutCounter | CounterType$ P1P1 | CounterNum$ 1
+Oracle:Trample\nBotanical Brawler enters the battlefield with two +1/+1 counters on it.\nWhenever one or more +1/+1 counters are put on another permanent you control, if it’s the first time +1/+1 counters have been put on that permanent this turn, put a +1/+1 counter on Botanical Brawler.
```
Agetian, line 9: DeckHas:Ability$Counters
DeckHints:Ability$Counters

Hanmac:
> Code looks fine - but I got Brawler to trigger from the ETB counters of _Fertilid_ and then also a second time on putting a counter on it after with _Simic Ascendancy_ (really shouldn't matter what gets used there)
> 
> Not a third time though

was it the right trigger? i tried:
- cast Brawler
- cast Fertilid (it enters with +1/+1 counters, Brawler triggers)
- cast Simic Ascendancy
- use its ability to put another +1/+1 counter on Fertilid, this time its the Simic Ascendancy that triggers and not the Brawler

tool4ever:
Ugh it seems to only happen if it ETB again, maybe some timestamp or lki gets messed up?

Here's the chain that shows it:
image

Hanmac:
> Ugh it seems to only happen if it ETB again, maybe some timestamp or lki gets messed up?
> 
> Here's the chain that shows it: image

the reason was ChangesZoneEffect was resetting the Timestamp, i removed it and it shouldn't break anything for now

tool4ever:
1. it looks like it was done for World rule, see https://github.com/card-forge/forge/commit/3e1626
so while it breaks that part it was always a bad fix and we have an issue for it already

2. that means you can remove even more code I added here: https://github.com/card-forge/forge/commit/33e8dc
it should no longer be needed

tool4ever (approved):

PR #2874, opened 2023-04-09
Draft: flitting_guerrilla.txt

Northmoc:
Seems like we shouldn't need line 8 and be able to just use `ExileAnyGrave<1/Card.TriggeredNewCard>` in line 9 – but that doesn't work.

All other cards like that are AB$ with Cost$ but this kind of has to be DB$ with UnlessCost$ because it's after the Mill ☹️ 

Made this a separate PR in case the engine should be updated

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardProperty.java
+++ b/forge-game/src/main/java/forge/game/card/CardProperty.java
@@ -1962,7 +1962,8 @@ public class CardProperty {
         } else if (property.startsWith("Triggered")) {
             if (spellAbility instanceof SpellAbility) {
                 final String key = property.substring(9);
-                Object o = ((SpellAbility)spellAbility).getTriggeringObject(AbilityKey.fromString(key));
+                SpellAbility sa = (SpellAbility) spellAbility;
+                Object o = sa.getRootAbility().getTriggeringObject(AbilityKey.fromString(key));
                 boolean found = false;
                 if (o != null) {
                     if (o instanceof CardCollection) {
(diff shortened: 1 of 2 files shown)


forge-gui/res/cardsfolder/upcoming/flitting_guerrilla.txt
```diff
@@ -0,0 +1,13 @@
+Name:Flitting Guerrilla
+ManaCost:2 B
+Types:Creature Faerie Rogue
+PT:2/2
+K:Flying
+T:Mode$ ChangesZone | Origin$ Battlefield | Destination$ Graveyard | ValidCard$ Card.Self | Execute$ TrigMill | TriggerDescription$ When CARDNAME dies, each player mills two cards. Then you may exile CARDNAME. When you do, put target creature or battle card from your graveyard on top of your library. (To mill two cards, a player puts the top two cards of their library into their graveyard.)
+SVar:TrigMill:DB$ Mill | Defined$ Player | NumCards$ 2 | SubAbility$ DBRemember
+SVar:DBRemember:DB$ Pump | RememberObjects$ TriggeredNewCardLKICopy | SubAbility$ DBImmediateTrig
```
tool4ever, line 8: https://github.com/Card-Forge/forge/blob/96ec9d201241ad050c2f0048f39f596a06089b9f/forge-game/src/main/java/forge/game/card/CardProperty.java#L1964

Insert getRootAbility() call there + for NotTriggered then no Remember needed

forge-game/src/main/java/forge/game/card/CardProperty.java
```diff
@@ -1980,7 +1981,10 @@ else if (property.equals("blocked")) {
         } else if (property.startsWith("NotTriggered")) {
             final String key = property.substring("NotTriggered".length());
             if (spellAbility instanceof SpellAbility) {
-                if (card.equals(((SpellAbility)spellAbility).getTriggeringObject(AbilityKey.fromString(key)))) {
+                SpellAbility sa = (SpellAbility) spellAbility;
+                Object o = sa.getRootAbility().getTriggeringObject(AbilityKey.fromString(key));
```
tool4ever, line 1985: Still need to use o there

tool4ever (approved):

PR #2877, opened 2023-04-09
cut_short.txt and support

Northmoc:


forge-game/src/main/java/forge/game/card/CardProperty.java
```diff
@@ -1067,6 +1067,10 @@ public static boolean cardHasProperty(Card card, String property, Player sourceC
             if (!property.startsWith("without") && !card.hasStartOfUnHiddenKeyword(property.substring(4))) {
                 return false;
             }
+        } else if (property.startsWith("activated")) {
```
tool4ever, line 1070: > A planeswalker was "activated" during a turn if one of its abilities was activated that turn. This includes its loyalty abilities and **any other activated abilities it may have**.

Try just checking numberTurnActivations -> isEmpty()

Northmoc: The problem with this is that if you cast a planeswalker spell, that populates the table. So currently Forge considers a PW "activated" on the turn it is cast even if it hasn't used any activated ability. I couldn't come up with a good way to get around it 😦 
Possibly any kind of Permanent CastSA shouldn't go into the activated table? Not *quite* sure how to wire that up...
image

tool4ever: Yea that doesn't make much sense

Ideally you can just add ``ability.isAbility()`` here:
 https://github.com/Card-Forge/forge/blob/5dc5bc3bfb9ee0ec11d54df1d139f0ec90174985/forge-game/src/main/java/forge/game/zone/MagicStack.java#L138

@Agetian
is that safe? Some (though not all) AI api have code like this but I'm not sure if it matters for spells or is still needed at all?
```
// prevent run-away activations - first time will always return true
boolean chance = MyRandom.getRandom().nextFloat() <= Math.pow(.6667, sa.getActivationsThisTurn());
```

Otherwise we need to do more work and filter Spells out of the table

Northmoc: I've got this version in the PR now... seems to work!

Northmoc: With that tweak things seem to be proper...

tool4ever (approved):

forge-game/src/main/java/forge/game/card/CardProperty.java
kevlahnota: is this ready?

Northmoc: I think so

PR #2879, opened 2023-04-09
StrictlyOther property

### LINKED ISSUE #2435: Zevlor: target choice happens too late
The script is a bit wrong, currently you can't choose illegal targets.

> As you make your choices for each other opponent, you can choose that opponent or any permanent they control, but if the chosen player or permanent isn't a legal target for the spell, the copy won't be created.

### LINKED ISSUE #2871: Other property missing timestamp check
With _Charmed Stray_ ETB trigger on the stack, flicker it. Because it is now a new object it should get a counter.

There are a lots of cards like that.

@Hanmac
I don't think we can make `equalsWithTimestamp` default or it will break other cards that use it when finding cards during zone change.

So probably need to update scripts with a `StrictlyOther` or you got a better idea?

tool4ever:
Closes #2871
Closes #2435


forge-gui/res/cardsfolder/a/agrus_kos_eternal_soldier.txt
```diff
@@ -4,6 +4,6 @@ Types:Legendary Creature Spirit Soldier
 PT:3/4
 K:Vigilance
 T:Mode$ BecomesTarget | ValidTarget$ Card.Self | ValidSource$ Ability.numTargets EQ1 | TriggerZones$ Battlefield | Execute$ TrigCopy | TriggerDescription$ Whenever CARDNAME becomes the target of an ability that targets only it, you may pay {1}{R/W}. If you do, copy that ability for each other creature you control that ability could target. Each copy targets a different one of those creatures. ({R/W} can be paid with either {R} or {W}.)
-SVar:TrigCopy:AB$ CopySpellAbility | Cost$ 1 RW | Defined$ TriggeredStackInstance | CopyForEachCanTarget$ Creature.YouCtrl+Other
+SVar:TrigCopy:AB$ CopySpellAbility | Cost$ 1 RW | Defined$ TriggeredStackInstance | CopyForEachCanTarget$ Creature.YouCtrl
```
Northmoc, line 7: should this have the new property too?

tool4ever: All these abilities are worded "for each other" so the engine already prevents that and I removed it for consistency

kevlahnota (approved):

PR #2899, opened 2023-04-12
Effect cleanup can prevent any LTB trigger

tool4ever:
Example: Target _Solemn Simulacrum_ with _Rogue's Passage_ and then deal lethal dmg to it

This is because the Static ExileOnMoved$ trigger will resolve first and the registered trigger from the LKI before death will be lost.

Notably it will fail with any combination of such cards but only if it dies from SBA, otherwise MagicStack handles it correctly.

I think a full call to ``resetActiveTriggers`` is unneeded / would be wrong during SBA.


Hanmac (approved):
👍

PR #2910, opened 2023-04-13
Added copy and delete deck

bbp9857:
Also adds regional language, though I am not fluent in any of these beyond English.


kevlahnota:
portrait mode needs testing @jjayers99

jjayers99:
Functionality is good. Does need to be added to portrait layout (deck_selector_portrait.json).

image

Logical approach is to expand to a second row of buttons at the bottom of the view. (Spacing on a touch screen interface could be an issue, so may need to be revisited later.)

Delete -- Copy -- Rename (33%/33%/33%)
Back ---- Edit (50%/50%)

However, this can be handled separately if needed, as portrait layout simply retains its current functionality without the additions.

kevlahnota (approved):

PR #2913, opened 2023-04-13
Battle: lazy build SpellPermanent on TDFC

tool4ever:
> If a permanent that is represented by a transforming double-faced card becomes a copy of a Siege, it will be exiled as that Siege's triggered ability resolves, then it will be cast transformed. Note that this applies only to transforming double-faced cards, not to modal double-faced cards that can normally be played using either face.

grafik

This will now correctly cast the back of _Aetherblade Agent_.

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -2939,6 +2939,12 @@ public class AbilityUtils {
                 la.setCardState(original);
                 list.add(la);
             }
+            if (state == CardStateName.Transformed && tgtCard.isPermanent() && !tgtCard.isAura()) {
+                // casting defeated battle
+                Spell sp = new SpellPermanent(tgtCard, original);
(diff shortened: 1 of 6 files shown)


forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -2939,6 +2939,12 @@ public static final List<SpellAbility> getBasicSpellsFromPlayEffect(final Card t
                 la.setCardState(original);
                 list.add(la);
             }
+            if (state == CardStateName.Transformed && tgtCard.isPermanent() && !tgtCard.isAura()) {
+                // casting defeated battle
+                Spell sp = new SpellPermanent(tgtCard);
```
Hanmac, line 2944: Use this constructor:

https://github.com/Card-Forge/forge/blob/master/forge-game/src/main/java/forge/game/spellability/SpellPermanent.java#L53

Northmoc (approved):
Looks great!

PR #2929, opened 2023-04-14
Furnace Host Update

EldritchBimbo:
Additions:
.Added two enemies based on the Furnace Host along with accompanying infrastructure. 
.Added level based on the Furnace Host.
.Added tileset based on the Furnace Host along with accompanying infrastructure.

Changes:
.Added new building to Phyrexian Buildings for furnace dungeon, edited atlas file to detect it. 
.Added new dungeon to POIs and Red jsons.
.Added new enemies to enemies json.
.Added proper quest tags to Dross faction enemies and dungeon.


forge-gui/res/adventure/Shandalar/world/enemies.json
```diff
@@ -4669,6 +4669,12 @@
 		}
 	],
 	"colors": "B"
```
Simisays, line 4671: missing comma

forge-gui/res/adventure/Shandalar/world/enemies.json
```diff
@@ -4741,6 +4747,14 @@
 		}
 	],
 	"colors": "BU"
```
Simisays, line 4749: Missing Comma

forge-gui/res/adventure/Shandalar/decks/furnacetormentor.json
```diff
@@ -0,0 +1,12 @@
+{                                    
+"name":"Furnace Tormentor",
+  "template":
+  {
+   "count":60,
+   "colors":["Red, Green"],
```
Simisays, line 6: Hmm I've played against this enemy and it had a colorless deck 
image

Simisays:
Cool dungeon, really like the custom design. One thing that might improve it a bit is giving the enemies a "Threatrange" of like 10 or 20 since now it's very easy to avoid all the patroling enemies (don't know if that's intended)

forge-gui/res/adventure/Shandalar/decks/furnacetormentor.json
jjayers99: Missing quotes in colors.
"Colors": ["Red", "Green"]

EldritchBimbo:
Appreciate the feedback and y'all catching those! (Did I do the rest of the things with quests right? I wasn't sure if there was a masterlist I had to define quest tags in, but things don't seem to be having issues coming from me adding new ones.)

The lack of a threat range was originally intended for thematic reasons, the ease wasn't. Ended up adding one based on this feedback - gameplay has to beat story in this case I think, and it makes the difficulty a lot more where I'd originally wanted it.

jjayers99:
There's no master list for quest tags, which means that consistency and
casing are important, but the worst that would happen is that the item
doesn't show up in quests.

Once I get it fixed up a little bit more, doing tags through the editor UI
will be the way to go, as it already pulls in a list of all existing tags
(helping make sure we don't end up with GrayDwarfs, greyDwarves, and 15
other variations thereof), and I hope to eventually show a list on a given
enemy / PoI as to what quests it will appear in based on the tags selected.

kevlahnota (approved):

PR #2931, opened 2023-04-15
Adventure quest updates

jjayers99:
Arena quest stage objectives now available, one quest added. 
Removed extra trigger for accepting quest in "A Vision of Destruction" offer dialog. 
Minor dialog enhancements.


kevlahnota (dismissed):

jjayers99 (commented):
Good catch. Didn't think about having potential for a null reference when moving to String representation of the UUID. Thanks!

PR #2943, opened 2023-04-17
Quest Worlds overhaul

icy-:
(re-made pull request as sourced from a branch to keep my fork trunk synced)
A project of passion to be able to delve into a particular block or expac "season" and duel against real decks at the time.  See the discord  #quest-modes post from April 1 for more info.
I've been maintaining and testing duels locally when I've been suggested to create a formal Pull Request, so here we are ^_^

I'll try to list out what changed to keep it easier to review.

**Renamed items:**
- Directory renaming.  Attached is a visual showing the dash system between some years and months, where present, to instead use periods.   This is to keep the dash available to specify a spread of years encompassing the bigger spans.  
- items within should be intact, just in a new directory now

**Added items:**
- thousands of new duels, mostly from Standard, mostly sourced from mtgtop8, where I was able to obtain and convert some deck lists with metadata such as author, tournament name, deck archetype, sideboard
- custom_icons directory.  This is a placeholder for icons of player-uploaded worlds that I'd initially found on SlightlyMagic forums such as Alara or Llorwyn.  As far as I can tell, these icons don't exist in a committable repository and need to be manually moved to `%localappdata%/Forge/Cache/pics/icons `  ,  so hence the folder and the accompanying readme.  I saw recently another player was talking about making their own custom 100x100 icons, so it might be a good idea to have a place to commit these to too.

**Modified items:**
- worlds.txt  .  Attached is another visual image.  (As a side note, if you set the syntax of this file to "Ruby", one of my fav langs, it's a happy coincidence that this file becomes much more readable.  The special syntax in the reader to split each line by pipes and then look for "Keyword:" labels looks nice on Ruby cause it's similar to that language's Symbols notation)
- items within are organized for the most part by Year.Month , ascending, and at the bottom you'll find the Random worlds just like before

Can DM or ping me on Forge discord for further questions -- i'm icy`   on there, and is fastest way to reach me.

overhaul_visual
overhaul_visual2


tehdiplomat:
So if you are in one of these worlds, what happens by you changing the name of them?

icy-:
> So if you are in one of these worlds, what happens by you changing the name of them?

Can you elaborate?  You mean to an older savefile?  It should be fine, IIRC was just showing blank duels until you "Travel" out to an existing one.
I've only seen problematic savefile errors when Edition elements are missing from the actual edition folder files -- in such cases Forge could startup crash upon trying to load cards in saves that it thought didnt exist or w/e.  But world stuff should be crash-free.

JustinLovinger:
A couple of suggestions after briefly testing, mostly with Phyrexia: All Will Be One.

First, I think it would be less jarring if duels used the name of the deck's creator for their duel title instead of a random unrelated character name or vague description. E.x. "Kevin Burroughs" instead of "Lupin the Third".

Second, as far as I can tell, opponents in a world use cards that were Standard at the time of the sets release, but players only get cards from that one set. Gameplay would be more symmetrical if the world for a given set included sets Standard at the time, instead of just the one set. It would also better fit the idea of "playing through MTG history".

Related to the second point, I suggest removing Commander sets, as those are not Standard. I assume opponents are not using Commander cards.

icy-:
> A couple of suggestions after briefly testing, mostly with Phyrexia: All Will Be One.
> 
> First, I think it would be less jarring if duels used the name of the deck's creator for their duel title instead of a random unrelated character name or vague description. E.x. "Kevin Burroughs" instead of "Lupin the Third".
> 
> Second, as far as I can tell, opponents in a world use cards that were Standard at the time of the sets release, but players only get cards from that one set. Gameplay would be more symmetrical if the world for a given set included sets Standard at the time, instead of just the one set. It would also better fit the idea of "playing through MTG history".
> 
> Related to the second point, I suggest removing Commander sets, as those are not Standard. I assume opponents are not using Commander cards.

For the duel names in duel selection screen, I had just looked at my icons directory and randomly picked a token avatar from what I had -- those names are just the names of the icons.  I saw from the existing worlds, this looked like a colorful common thing to do, like, hey, you're fighting Rambo!  Yeah maybe a little jarring, but you still have the duel description underneath in Italics as you're picking which duel to do.

For the Challenges, I had been trying to set the actual player's name, as there's an `OpponentName=`  line.  I'm not sure if can use this line within normal duel metadata too or not  
image

Yeah, the nearby Commander cards was probably overkill on my part if sticking with mostly Standard -- was partially me wanting to chance related things.  I was exploring what people had built as Vintage concepts not too long ago and found a recent Initiative deck with 4x White Plume Adventurer and 4x Seasoned Dungeoneer.   Both of these are only from Commander Legends: Battle for Baldur's Gate (CLB), so I figured I'd try to farm these while dueling and it kinda went from there ^_^

PR #2944, opened 2023-04-17
upcoming card scripting for Scrollshift[MOM]

icy-:
https://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=607051  

Confirmed that it doesnt draw card if the target gets instantly shrouded on the stack.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/scrollshift.txt
+++ b/forge-gui/res/cardsfolder/upcoming/scrollshift.txt
@@ -1,0 +1,8 @@
+Name:Scrollshift
+ManaCost:2 W
+Types:Instant
+A:SP$ ChangeZone | ValidTgts$ Artifact.YouCtrl,Creature.YouCtrl,Enchantment.YouCtrl | Origin$ Battlefield | Destination$ Exile | TargetMin$ 0 | TargetMax$ 1 | TgtPrompt$ Select up to one target artifact, creature, or enchantment you control | RememberTargets$ True | SubAbility$ DBReturn | StackDescription$ Exile {c:Targeted}, | SpellDescription$ Exile up to one target artifact, creature, or enchantment you control, then return it to the battlefield under its owner's control. Draw a card.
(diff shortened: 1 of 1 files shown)


forge-gui/res/cardsfolder/upcoming/scrollshift.txt
```diff
@@ -0,0 +1,8 @@
+Name:Scrollshift
+ManaCost:2 W
+Types:Instant
+A:SP$ ChangeZone | ValidTgts$ Artifact.YouCtrl,Creature.YouCtrl,Enchantment.YouCtrl | Origin$ Battlefield | Destination$ Exile | TargetMin$ 0 | TargetMax$ 1 | TgtPrompt$ Select up to one target artifact, creature, or enchantment you control | RememberTargets$ True | SubAbility$ DBReturn | StackDescription$ Exile {c:Targeted}, | SpellDescription$ Exile up to one target artifact, creature, or enchantment you control, then return it to the battlefield under its owner's control. Draw a card.
```
Northmoc, line 4: RememberChanged$ might be better than RememberTargets$

Other cards seem inconsistent... @tool4ever thoughts?

tool4ever: Eh that's a very rules specific choice

If you remember the target you're obviously getting the old object from its previous zone instead of the new one

As long as you're not using "Remembered**LKI**" after that it doesn't matter because you will fetch the newest card state anyway

For this card everything happens in a single chain and to my current knowledge no replacement effects exist that could mess it up

So it should be safe here, but one could argue that it's not the ideal default and might lead to bad practice for other scripters when they copy/paste because there are certainly variants of this effect where it's important that the timestamps get compared

Northmoc: Since there are other cards with the same issue and this is a bit murky let's merge this PR and make an issue for later

Northmoc (approved):
👍

PR #2947, opened 2023-04-18
fix to allow Serum Powder to draw cards after exiling hand

icy-:
It was previously just exiling the hand and not drawing cards

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/match/input/InputConfirmMulligan.java
+++ b/forge-gui/src/main/java/forge/gamemodes/match/input/InputConfirmMulligan.java
@@ -102,32 +102,34 @@ public class InputConfirmMulligan extends InputSyncronizedBase {
 
     volatile boolean cardSelectLocked = false;
 
+    /**
+     * When a card selected at the time of mulligan (currently affects just Serum Powder).
+     */
     @Override
     protected boolean onCardSelected(final Card c0, final List<Card> otherCardsToSelect, final ITriggerEvent triggerEvent) { // the only place that would cause troubles - input is supposed only to confirm, not to fire abilities
+        if (cardSelectLocked) { return false; }
         final boolean fromHand = player.getZone(ZoneType.Hand).contains(c0);
         final boolean isSerumPowder = c0.getName().equals("Serum Powder");
-        final boolean isLegalChoice = fromHand && (isSerumPowder);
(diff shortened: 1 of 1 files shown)


forge-gui/src/main/java/forge/gamemodes/match/input/InputConfirmMulligan.java
```diff
@@ -117,7 +117,7 @@ protected boolean onCardSelected(final Card c0, final List<Card> otherCardsToSel
             cardSelectLocked = true;
             ThreadUtil.invokeInGameThread(new Runnable() {
                 @Override public void run() {
-                    final CardCollectionView hand = c0.getController().getCardsIn(ZoneType.Hand);
+                    final CardCollection hand = new CardCollection(c0.getController().getCardsIn(ZoneType.Hand));
```
kevlahnota, line 120: this should work rather than instantiating new objects
```
                    final CardCollectionView hand = c0.getController().getCardsIn(ZoneType.Hand);
                    final int size = hand.size();
                    for (final Card c : hand.threadSafeIterable()) {
                        player.getGame().getAction().exile(c, null);
                    }
                    c0.getController().drawCards(size);
                    cardSelectLocked = false;
```

kevlahnota: hmm, so whats the best method to fix this @Hanmac

Hanmac: imo threadSafeIterable might be better than full CardCollection

tool4ever: yea that just makes it clearer to the reader what the intent is

icy-: ty; implemented this suggestion.   I was looking to simplify for readability, and was wondering if we even need the "fromHand" check?  At this step, it'll always be from hand.
  1) Game starts  2) Serum Powder checks  3) mulligans 4) possible jump back to second step  5) Hand secured 6) interactions with Leylines/Gemstone Caverns/Chancellors/Providence/Sphinx of Foresight

kevlahnota (approved):

PR #2949, opened 2023-04-18
Add Blood Moon adventure item

bbp9857:
This adds a new item for the adventurer to discover; using it will reset the nearest cave or dungeon so that the player can relive the experience.


forge-gui/res/adventure/Shandalar/world/items.json
```diff
@@ -1179,5 +1179,20 @@
         "Slobad's Iron Boots"
       ]
 }
+},
```
jjayers99, line 1182: The line before this would need to be removed, otherwise this creates invalid JSON.

jjayers99:
This is a neat idea, but I do feel like it has some overlap with existing functionality and future plans. 

Specific dungeons already respawn enemies automatically. Quests that send you in to a dungeon now reset that dungeon. And there are plans to potentially randomize some dungeon content, which this may interact with in an unwanted manner.

I'm not saying no, more that I need to thoroughly think through the interactions this will have before saying yes.

forge-gui/res/adventure/Shandalar/world/items.json
bbp9857: This is an artifact of how the rest of the file is formatted.  I kept the whitespace positioning in line with the other blocks.  I can assure you that the json is valid, otherwise I wouldn't have been able to test this at all.

jjayers99: Got it. This is what I get for reviewing changes half asleep.

PR #2954, opened 2023-04-18
Update ReplaceDamage with Battle checks

Hanmac:
extra check for ReplaceDamage to redirect damage to Battle


Hanmac:
@tehdiplomat @tool4ever current tests are failing, we need to fix them

kevlahnota (approved):

kevlahnota:
> @tehdiplomat @tool4ever current tests are failing, we need to fix them

it was the check I added but commented it out.

PR #2960, opened 2023-04-18
Non-planeswalker card ultimate achievements (for MOM sagas)

Agetian:
This achievement framework extension allows any card ability to serve as an ultimate in case it's specified as "Ultimate$ True". If it's a non-planeswalker ability, it'll be registered as a card ability ultimate, with the achievements awarded per the list in special-card-achievements.txt.

Currently used for the new MOM sagas. Achievement texts were made by Marek14.

Some of the saga cards aren't yet in upstream master branch. Once they're in, their scripts will need to be updated with "Ultimate$ True" for the new achievements to work for them.


kevlahnota (approved):

Northmoc:
Cool!

PR #2963, opened 2023-04-19
Update target restrictions to account for "Any"

tehdiplomat:


Northmoc (approved):
Looks good to me

Hanmac:
i would added a canTgtAny function, but that is fine too

tehdiplomat:
@Hanmac Yea I think I might just refactor this portion of the class. I'd rather loop through the ValidTgts once, and track what GameEntity types are targetable in a set, then have each of these functions check the set.

PR #2976, opened 2023-04-21
Reduce inventory clutter for unique items

bbp9857:
Non-consumable quest items cannot be deleted. Multiples clutter inventory.
This change allows some items to be marked as unique such that only one
instance of a unique item may reside in the player's inventory. Note that
quest items can still have multiples if not explicitly marked as unique.


jjayers99:
This has been on my to-do list for a while, but I'm not sure that this is the proper approach. I feel as though the items should become sellable rather than just not getting them.

PR #2979, opened 2023-04-21
MOC: Chaos Ensues!!

### LINKED ISSUE #2829: Planescape Chaos Ensues
* "roll for Chaos" needs to be changed into a Chaos (Ensues) Trigger
* Chaos also needs to be an Effect so that it can be called by SpellAbility (better make it a function in Game?)
* Rolling Chaos on the PlanarDie makes it ensue Chaos (no extra Trigger, should be in Game)

`Stairs to Infinity` Updated Oracle:
> Whenever you roll the planar die, draw a card.
> Whenever chaos ensues, reveal the top card of your planar deck. You may put it on the bottom of your planar deck

We still need `TriggerType.PlanarDice` for this,

Northmoc:
- closes #2829


kevlahnota (approved):

forge-game/src/main/java/forge/game/trigger/TriggerChaosEnsues.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /* (non-Javadoc)
+     * @see forge.card.trigger.Trigger#performTest(java.util.Map)
+     */
+    @Override
+    public boolean performTest(Map<AbilityKey, Object> runParams) {
+        if (!matchesValidParam("ValidPlayer", runParams.get(AbilityKey.Player))) {
+            return false;
+        }
+        if (runParams.containsKey(AbilityKey.Affected)) {
```
Hanmac, line 36: This can be cleaned up more with the valid functions

Hanmac:
The Trigger Zones for the other Planes might be problematic when getting hit by `The Fertile Lands of Saulvinia`

Hanmac:
@Northmoc have you tested the trigger zones for the special "The Fertile Lands of Saulvinia" case?

Northmoc:
> @Northmoc have you tested the trigger zones for the special "The Fertile Lands of Saulvinia" case?

Yep, they seem to be working. I put some code in the "Defined" part of ChaosEnsuesEffect for this

Hanmac:
@Northmoc ah now i found your solution:
https://github.com/Card-Forge/forge/pull/2979/files#diff-859933e432a48b305d730a6ead94271c86143ca1a0140ea6b1e5f3e47b9f7402R41-R54

imo its a bit ugly, but i think that works for now

tool4ever:
I think it might still break the triggers because when it adds the extra zone it reuses the same Collection later used for restoring?

PR #2994, opened 2023-04-24
Addition of MOM Jumpstart packs, adding into Adventure start options

jjayers99:


forge-gui/res/adventure/Shandalar/config.json
```diff
@@ -203,6 +203,7 @@
 	"BRO",
 	"J22",
 	"ONE",
+    "MOM",
```
paulsnoops, line 206: Can this be tabbed in one more?

PR #3001, opened 2023-04-24
Fix RE messing up Triggers

### LINKED ISSUE #2229: Flagstones of Trokair don't trigger when sacrificed via Lotus Vale
To reproduce, have Flagstones of Trokair and Plains in play, then Play Lotus Vale and sacrifice Flagstones of Trokair and Plains. Flagstones of Trokair should trigger its ability. Original report found here https://www.slightlymagic.net/forum/viewtopic.php?f=52&t=6333&p=254273#p254273
Screenshot_20230111_152326

tool4ever:
- Fixed Valakut not triggering when ETB last together with enough mountains:
Got broken by https://github.com/Card-Forge/forge/commit/fb317f

- Closes #2229:
I went back to the original commit (https://github.com/Card-Forge/forge/commit/9c7871) and after testing consider it safe to remove completely because `clearActiveTriggers` now already takes care of avoiding a misfire

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -1453,7 +1453,6 @@ public class AbilityUtils {
 
         // Needed - Equip an untapped creature with Sword of the Paruns then cast Deadshot on it. Should deal 2 more damage.
         game.getAction().checkStaticAbilities(); // this will refresh continuous abilities for players and permanents.
-        game.getTriggerHandler().resetActiveTriggers(!sa.isReplacementAbility());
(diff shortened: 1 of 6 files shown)


tool4ever:
Seems the AI doesn't try to unfreeze in such cases at all. But unfortunately I'm not really sure what other bug it could prevent and if this discrepancy is actually supposed to exist for the future...

tool4ever:
For effect cleanup I brought over Pump entry from timestamp branch already.

Mostly because it felt very game breaking to Flicker some creature that had a regeneration shield which would still apply to it after.

kevlahnota (approved):

kevlahnota:
this looks good, gonna merge this.

PR #3014, opened 2023-04-26
MAT: tazri_stalwart_survivor.txt + support

Northmoc:


forge-gui/res/cardsfolder/upcoming/tazri_stalwart_survivor.txt
```diff
@@ -0,0 +1,12 @@
...
+ManaCost:2 W
+Types:Legendary Creature Human Warrior
+PT:3/3
+S:Mode$ Continuous | Affected$ Creature.YouCtrl | AddAbility$ Mana | Description$ Each creature you control has "{T}:  Add one mana of any of this creature's colors. Spend this mana only to activate an ability of a creature. Activate only if this creature has another activated ability."
+SVar:Mana:AB$ ManaReflected | Cost$ T | Valid$ Defined.Self | ColorOrType$ Color | ReflectProperty$ Is | RestrictValid$ Activated.Creature+inZoneBattlefield | IsPresent$ Card.Self+hasOtherActivatedAbility | SpellDescription$ Add one mana of any of this creature's colors. Spend this mana only to activate an ability of a creature. Activate only if this creature has another activated ability.
+A:AB$ Mill | Cost$ W U R B G | NumCards$ 5 | RememberMilled$ True | SubAbility$ DBChangeZone | SpellDescription$ Mill five cards.
+SVar:DBChangeZone:DB$ ChangeZoneAll | ChangeType$ Creature.IsRemembered+hasNonManaActivatedAbility | Origin$ Graveyard | Destination$ Hand | SubAbility$ DBCleanup | SpellDescription$ Put all creature cards with activated abilities that aren't mana abilities from among the milled cards into your hand.
```
tool4ever, line 8: We usually add Exile because of rulings like this:
> If a replacement effect causes a player to exile the top four cards of their library instead of putting them into their graveyard as Gyruda’s triggered ability resolves, the creature card you choose may be one of those cards in exile.

Northmoc: Ah, because they're still technically "milled" regardless of zone they go into...

Hmm, might `Origin$ All` be better futureproofing then?

tool4ever: I'd have to look up the rule but I think it only works with public zones

Northmoc: Fair enough... so maybe a bigger refactor later if needed.

tool4ever (approved):

forge-gui/res/cardsfolder/upcoming/tazri_stalwart_survivor.txt
```diff
@@ -0,0 +1,12 @@
...
+A:AB$ Mill | Cost$ W U R B G | NumCards$ 5 | RememberMilled$ True | SubAbility$ DBChangeZone | SpellDescription$ Mill five cards.
+SVar:DBChangeZone:DB$ ChangeZoneAll | ChangeType$ Creature.IsRemembered+hasNonManaActivatedAbility | Origin$ Graveyard,Exile | Destination$ Hand | SubAbility$ DBCleanup | SpellDescription$ Put all creature cards with activated abilities that aren't mana abilities from among the milled cards into your hand.
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+DeckHas:Ability$Mill
+AI:RemoveDeck:Random
+Oracle:Each creature you control has "{T}: Add one mana of any of this creature's colors. Spend this mana only to activate an ability of a creature. Activate only if this creature has another activated ability."\n{W}{U}{R}{B}{G}: Mill five cards. Put all creature cards with activated abilities that aren't mana abilities from among the milled cards into your hand.
```
squee1968, line 12: Isn't the order supposed to wubrg?

PR #3016, opened 2023-04-26
MAT: spark_rupture.txt

Northmoc:
There was some discussion regarding this one in the Discord, but it seems to work fine without engine changes... making it its own PR in case it gets hairy with corner cases or something.

It seems to work properly on Planeswalkers that are already on the battlefield when it is cast, and on any other Planeswalkers that ETB afterward.

Planeswalkers also seem to revert properly after it is removed.


Hanmac:
yeah i just wanted to point out that we need to check that it only applies to Planeswalker that are on the Battlefield, not while they are entering

tool4ever (approved):

PR #3019, opened 2023-04-27
Fix failed trigger keeping the stack frozen

tool4ever:
So I think this bug (which only affects human) has been hiding and potentially causing confusion for quite a while because it's 1. a bit hard to notice and 2. also requires specific board states to occur which are usually not obvious either.

As you can see in the reported Screenshot Ward triggers way too late:
grafik

This happens because the Spell cast by the AP also triggers a Magecraft ability which however fails targeting but leaves the stack frozen for the following (NAP) Ward trigger.


Hanmac (approved):
👍

PR #3023, opened 2023-04-27
MAT: ob_nixilis_captive_kingpin.txt + support

Northmoc:
I'm pretty happy with this after some testing

One interesting situation I'm not sure about:
Opp 1 has 2 lands
Opp 2 has 1 land
I play `Cleansing` and both opps pay 1 life to save each land. New Ob Nix currently triggers 3 times. Is that right?


Hanmac:
Yeah we probably need to think about multiple UnlessCost :/

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -2429,6 +2430,9 @@ public void dealDamage(final boolean isCombat, final CardDamageMap damageMap, fi
 
                 e.setValue(Integer.valueOf(e.getKey().addDamageAfterPrevention(e.getValue(), sourceLKI, isCombat, counterTable)));
                 sum += e.getValue();
+                if (!isCombat && e.getKey() instanceof Player && e.getValue() > 0) {
+                    lossMap.put((Player) e.getKey(), e.getValue());
```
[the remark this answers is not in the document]
Northmoc: Good catch

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -455,7 +455,14 @@ public final boolean setLife(final int newLife, final SpellAbility sa) {
         boolean change = false;
         // rule 119.5
         if (life > newLife) {
-            change = loseLife(life - newLife, false, false) > 0;
+            final Map<Player, Integer> lossMap = Maps.newHashMap();
```
tool4ever, line 458: Shouldn't this be in LifeSetEffect too?

forge-game/src/main/java/forge/game/trigger/TriggerLifeLostAll.java
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * 4/27/2023 - this trigger is written for only Ob Nixilis, Captive Kingpin – any future uses will probably need
+ * additional logic
+ *
+ * @author Forge (Northmoc)
+ * @version $Id$
+ */
+public class TriggerLifeLostAll extends Trigger {
+
+    private Map<Player, Integer> triggerMap = Maps.newHashMap();
```
tool4ever, line 43: Hmn, I think there might be corner cases hiding there where it would get overwritten if it triggers more than once during some other effect

Imo just filter it again, shouldn't really matter for performance

forge-game/src/main/java/forge/game/player/Player.java
Hanmac: Yes

Northmoc: Hmm, I thought I got that effect... I'll check

Northmoc: Ah, I see what I did wrong now – thanks!

Northmoc: While I was in there, I noticed that Reverse the Sands didn't really give you a choice in choosing players... so I tried to straighten that out as well.

Northmoc:
> Yeah we probably need to think about multiple UnlessCost :/

Do you two think we should make this an issue for follow-up or that is needs to be handled in this PR? Seems like we may need rules clarifications on the new trigger type... my dig through the rules didn't give me a conclusive answer.

forge-game/src/main/java/forge/game/ability/effects/LifeSetEffect.java
```diff
@@ -33,23 +67,37 @@ public void resolve(SpellAbility sa) {
...
 
+        final Map<Player, Integer> lossMap = Maps.newHashMap();
         for (final Player p : players.threadSafeIterable()) {
             if (!p.isInGame()) {
                 continue;
             }
             if (!redistribute) {
+                final int preLife = p.getLife();
                 p.setLife(lifeAmount, sa);
+                final int diff = preLife - p.getLife();
```
tool4ever, line 78: looks like you could save these 4 lines if you'd just call them at loop ending

forge-game/src/main/java/forge/game/ability/effects/LifeLoseEffect.java
```diff
@@ -41,13 +46,19 @@ public void resolve(SpellAbility sa) {
...
         final int lifeAmount = AbilityUtils.calculateAmount(sa.getHostCard(), sa.getParam("LifeAmount"), sa);
 
+        final Map<Player, Integer> lossMap = Maps.newHashMap();
         for (final Player p : getTargetPlayers(sa)) {
             if (!p.isInGame()) {
                 continue;
             }
-            lifeLost += p.loseLife(lifeAmount, false, false);
+            final int lost = p.loseLife(lifeAmount, false, false);
+            lossMap.put(p, lost);
```
tool4ever, line 55: We should really avoid running triggers if nothing happened which doesn't seem to be the case in all places

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -543,6 +543,10 @@ public final boolean canGainLife() {
     }
 
     public final int loseLife(int toLose, final boolean damage, final boolean manaBurn) {
+        return loseLife(toLose, damage, manaBurn, false);
```
tool4ever, line 546: is this extra param really needed / useful?
Seems to me the logic is when `damage = true` GameAction needs to do it

Hanmac: unless Cost with Player losing Life

forge-game/src/main/java/forge/game/phase/PhaseHandler.java
```diff
@@ -465,13 +465,19 @@ private void onPhaseEnd() {
...
         }
 
+        final Map<Player, Integer> lossMap = Maps.newHashMap();
         for (Player p : game.getPlayers()) {
             int burn = p.getManaPool().clearPool(true).size();
 
             if (p.getManaPool().hasBurn()) {
-                p.loseLife(burn, false, true);
+                final int lost = p.loseLife(burn, false, false);
+                lossMap.put(p, lost);
```
tool4ever, line 474: Looks like this could still fill the map with 0 values

tool4ever:
I asked judges:
something like _Sarkhan the Mad_ Ult will result in a single life loss no matter how many dragons deal damage to you, so _Vampire Scrivener_ should never trigger more than once.

That means I can clean it up some more

forge-game/src/main/java/forge/game/ability/effects/LifeLoseEffect.java
squee1968: > Yeah we probably need to think about multiple UnlessCost :/

Same for Lion's Eye Diamond

tool4ever:
I think it might be good enough for now?

Hanmac (approved):

PR #3027, opened 2023-04-28
Fix crash

tool4ever:


forge-game/src/main/java/forge/game/ability/effects/CopySpellAbilityEffect.java
```diff
@@ -72,7 +72,11 @@ public void resolve(SpellAbility sa) {
 
         List<SpellAbility> tgtSpells = getTargetSpells(sa);
 
-        tgtSpells.removeIf(tgtSA -> StaticAbilityCantBeCopied.cantBeCopied(tgtSA.getHostCard()));
```
Hanmac, line 75: wouldn't it be better to get removeIf working for us?

tool4ever: good point

kevlahnota (approved):

Hanmac (approved):

PR #3031, opened 2023-04-28
MAT: tranquil_frillback.txt + support/related

Northmoc:
I have some ideas for teaching the AI to actually use the abilities Tranquil Frillback / Adversary cycle, but that is for another PR. For now they get the `RemoveDeck:All` because the AI will never use the abilities.

On the human side, I think I've finally got Cemetery Desecrator and Guile, Sonic Soldier displaying well:

image

image

image

### DIFF
--- a/forge-game/src/main/java/forge/game/trigger/Trigger.java
+++ b/forge-game/src/main/java/forge/game/trigger/Trigger.java
@@ -191,15 +191,13 @@ public abstract class Trigger extends TriggerReplacementBase {
                         break;
                     }
                     if (ApiType.ImmediateTrigger.equals(api) || ApiType.DelayedTrigger.equals(api)) {
-                        if (sa.hasAdditionalAbility("Execute")) {
-                            SpellAbility trigSA = sa.getAdditionalAbility("Execute");
-                            while (trigSA != null) {
-                                if (ApiType.Charm.equals(trigSA.getApi())) {
-                                    saDesc = CharmEffect.makeFormatedDescription(trigSA, !forStack);
-                                    break;
-                                }
-                                trigSA = trigSA.getSubAbility();
+                        SpellAbility trigSA = sa.getAdditionalAbility("Execute");
+                        while (trigSA != null) {
+                            if (ApiType.Charm.equals(trigSA.getApi())) {
+                                saDesc = CharmEffect.makeFormatedDescription(trigSA, !forStack);
+                                break;
                             }
+                            trigSA = trigSA.getSubAbility();
                         }
                         break;
                     }
--- a/forge-game/src/main/java/forge/game/trigger/Trigger.java
+++ b/forge-game/src/main/java/forge/game/trigger/Trigger.java
@@ -17,23 +17,10 @@
  */
 package forge.game.trigger;
 
-import java.util.ArrayList;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
-
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
-
-import forge.game.Game;
-import forge.game.GameEntity;
-import forge.game.GameStage;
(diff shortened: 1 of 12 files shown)


forge-game/src/main/java/forge/game/trigger/Trigger.java
```diff
@@ -184,17 +177,68 @@ public final String replaceAbilityText(final String desc, SpellAbility sa) {
...
+                                }
+                            }
+                        }
+                        subSa = subSa.getSubAbility();
+                    }
                 }
             } else if (ApiType.Charm.equals(sa.getApi())) {
                 // use special formating, can be used in Card Description
                 saDesc = CharmEffect.makeFormatedDescription(sa);
+            } else if (ApiType.ImmediateTrigger.equals(sa.getApi())) {
```
tool4ever, line 215: I think with some refactoring and maybe a single flag you would only need both of these new blocks once

forge-game/src/main/java/forge/game/trigger/Trigger.java
```diff
@@ -177,23 +170,43 @@ public final String replaceAbilityText(final String desc, SpellAbility sa) {
...
+            if (digMore) { // if ABILITY is used, there is probably Charm somewhere
+                while (sa != null) {
+                    ApiType api = sa.getApi();
+                    if (ApiType.Charm.equals(api)) {
+                        saDesc = CharmEffect.makeFormatedDescription(sa, !forStack);
+                        break;
+                    }
+                    if (ApiType.ImmediateTrigger.equals(api) || ApiType.DelayedTrigger.equals(api)) {
+                        if (sa.hasAdditionalAbility("Execute")) {
+                            SpellAbility trigSA = sa.getAdditionalAbility("Execute");
```
tool4ever, line 195: `getAdditionalAbility` already calls `hasAdditionalAbility` so you can save that if-block too
Should be good after

forge-gui/res/cardsfolder/i/intrepid_adversary.txt
```diff
@@ -10,4 +10,5 @@ SVar:X:Count$TriggerRememberAmount
 S:Mode$ Continuous | Affected$ Creature.YouCtrl | AddPower$ Z | AddToughness$ Z | Description$ Creatures you control get +1/+1 for each valor counter on CARDNAME.
 SVar:Z:Count$CardCounters.VALOR
 DeckHas:Ability$LifeGain|Counters
+AI:RemoveDeck:All
```
Hanmac, line 13: imo the AI can use this good, even if it would pay them for the max, it is still okay if the AI already has a board of creatures

Hanmac: I didn't tested it recently but I saw it happen on Android once

Northmoc: I can reproduce it... huh...

Northmoc: I can't figure out why the AI will dump mana into Intrepid Adversary but not Bloodthirsty Adversary or the Tranquil Frillback 😖

forge-gui/res/cardsfolder/b/bloodthirsty_adversary.txt
```diff
@@ -11,4 +11,5 @@ SVar:DBCopyCast:DB$ Play | Valid$ Card.IsRemembered | ValidZone$ Exile | Control
 SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
 SVar:X:Count$TriggerRememberAmount
 DeckHas:Ability$Counters
+AI:RemoveDeck:All
```
Hanmac, line 14: This one needs two thoughts for the AI:
* First check for Instance and Sorcery with cmc 3 or less that are okay with being cast for free
* Then Check if the Extra Mana to get more counters are worth it (is 1 counter for 2 R enough to be lethal?)

Northmoc:
Flagged some AI unplayable cards for review later - for now at least these can be added to engine

PR #3040, opened 2023-04-30
MAT: tyvar_the_bellicose.txt + support

Northmoc:


forge-game/src/main/java/forge/game/zone/MagicStack.java
```diff
@@ -255,12 +255,13 @@ public final void add(final SpellAbility sp) {
             runParams.put(AbilityKey.Cost, sp.getPayCosts());
             runParams.put(AbilityKey.Activator, activator);
             runParams.put(AbilityKey.CastSA, sp);
+            runParams.put(AbilityKey.SpellAbility, sp);
```
tool4ever, line 258: Just use CastSA...?

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -1846,7 +1816,12 @@ public static int xCount(final Card c, final String s, final CardTraitBase ctb)
                     }
                     return count;
                 }
+                // Count$ManaProduced
+                if (sq[0].startsWith("AmountManaProduced")) {
+                    final SpellAbility root = sa.getRootAbility();
```
tool4ever, line 1821: sometimes the manaPart is a sub so this might crash
use `getAllManaParts()`

tool4ever (changes requested):
Probably needs `|| regtrig instanceof TriggerAbilityResolves` here:
https://github.com/Card-Forge/forge/blob/0427c9af71cfd7e418edb43903dfb6cbeea58c00/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java#L489

tool4ever:
Currently TapsForMana abilities like that of _Vorinclex, Voice of Hunger_ can't get it to trigger:

Probably need to add it here too:
https://github.com/Card-Forge/forge/blob/0427c9af71cfd7e418edb43903dfb6cbeea58c00/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java#L571

Though it feels a bit ugly 🤷‍♂️

Northmoc:
> Currently TapsForMana abilities like that of _Vorinclex, Voice of Hunger_ can't get it to trigger:
> 
> Probably need to add it here too:
> 
> https://github.com/Card-Forge/forge/blob/0427c9af71cfd7e418edb43903dfb6cbeea58c00/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java#L571
> 
> Though it feels a bit ugly 🤷‍♂️

Hmm, that seems right to me though... Vorinclex doesn't modify the ability on the land, it is its own trigger and is not itself a mana ability... right?

This is getting into judge / rules lawyer territory ha ha

tool4ever:
It is one:

> 605.1b. A triggered ability is a mana ability if it meets all of the following criteria: it doesn't require a target (see rule 115.6), it triggers from the activation or resolution of an activated mana ability (see rule 605.1a) or from mana being added to a player's mana pool, and it could add mana to a player's mana pool when it resolves.

Northmoc:
> Probably needs `|| regtrig instanceof TriggerAbilityResolves` here:
> 
> https://github.com/Card-Forge/forge/blob/0427c9af71cfd7e418edb43903dfb6cbeea58c00/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java#L489

Looks like you added this in a different PR

forge-game/src/main/java/forge/game/trigger/TriggerAbilityResolves.java
```diff
@@ -55,11 +55,12 @@ public final boolean performTest(final Map<AbilityKey, Object> runParams) {
     /** {@inheritDoc} */
     @Override
     public final void setTriggeringObjects(final SpellAbility sa, Map<AbilityKey, Object> runParams) {
-        final SpellAbility triggeredSA = (SpellAbility) runParams.get(AbilityKey.SpellAbility);
+        final SpellAbility triggeredSA = (SpellAbility) runParams.get(AbilityKey.CastSA);
         sa.setTriggeringObject(AbilityKey.Source, triggeredSA.getHostCard());
+        sa.setTriggeringObject(AbilityKey.SpellAbility, triggeredSA);
```
Hanmac, line 60: maybe update the triggers to use CastSA instead of SpellAbility, so we don't need to use SpellAbility at all?

Northmoc: Hmm, refactoring out SpellAbility here causes a bit of a tangle because a few cards are already wired up with `SVar:X:TriggeredSpellAbility$`...

If you think it would be worthwhile that might be a good issue for a future PR?

tool4ever: I think I will do the opposite and just get rid of `AbilityKey.CastSA`
it's rarely used and doesn't really fit for all these trigger variants anyway

Hanmac: hm CastSA -> SpellAbility is fine by me too i think

tool4ever (approved):

PR #3042, opened 2023-05-01
Release 1.6.57

tehdiplomat:


leriomaggio:
Leaving a comment here as a general note (also as a reminder for myself):
there is the startup script for macOS that needs revision - I have it done already and I am planning to file in a PR in the next couple of days :)

tehdiplomat:
@leriomaggio if we need a revision can we hook it up with maven so it auto-updates?

PR #3055, opened 2023-05-02
Kor Replacement Update

EldritchBimbo:
Changes:
.Replaced Kor Warrior sprite with more canon-compliant one. 
.Replaced Kor player models and portraits with more canon-compliant new ones. 
.Archived old sprites, portraits.

Additional Comments:
.This was originally discussed in the spirit of replacing the existing sprites, so that's what I designed it as. That said, (specifically in the case of the player sprite), we have a couple other options if we want to keep the old sprites. (This would also prevent existing saves from having their appearance changed.) Our options in total are:
1. Just replace the sprites.
2. Keep the current sprites as the Kor, and rename the new ones to "Kor (Zendikar)" or similar.
3. (My personal preference.) The old sprites don't much look like Kor, but honestly they'd make decent Orcs. Rename the existing sprites accordingly, and implement the new ones as a new option.


Simisays:
Looks good to me, thanks for your work ! Feel free to change the Kor enemy sprite with your new one, since like you said the old one doesn't really look like a Kor.

kevlahnota:
@EldritchBimbo can you make the face darker so the avatar dont look flat? Something like this for reference:
image

EldritchBimbo:
Yea, I can do that! ^^
Is that for specific ones, or does that apply to all of them?

kevlahnota:
> Yea, I can do that! ^^ Is that for specific ones, or does that apply to all of them?

I just searched google for kor card images so it can be varied since some have black hair... 
image

EldritchBimbo:
Seems like they're bluer, too. I think that means they're Rathi?
I can retool some of these to be that kind, and shade the faces more in general.

kevlahnota:
> > Yea, I can do that! ^^ Is that for specific ones, or does that apply to all of them?
> 
> I just searched google for kor card images so it can be varied since some have black hair... image

maybe, i look at the nomads enkor and shaman enkor (I think they are kor types) and also have black hairs too.. maybe hybrids...

kevlahnota:
> Seems like they're bluer, too. I think that means they're Rathi? I can retool some of these to be that kind, and shade the faces more in general.

maybe, I look at shaman enkor and nomads en kor and they have black hairs (they are kor types too, maybe hybrids I think...)

kevlahnota (approved):

PR #3067, opened 2023-05-06
MayPlayLimit is not copiable

tool4ever:
image


forge-game/src/main/java/forge/game/staticability/StaticAbility.java
```diff
@@ -604,6 +604,8 @@ public StaticAbility copy(Card host, final boolean lki) {
             // reset to force refresh if needed
             clone.payingTrigSA = null;
 
+            clone.mayPlayTurn = 0;
```
Hanmac, line 607: maybe only if lki is false?

Hanmac (approved):
:+1:

PR #3079, opened 2023-05-08
QHD/UHD resolution support in Adventure Mode

Agetian:
- Add 1440p and 2160p to the list of supported modes in Adventure Mode.


forge-gui-mobile/src/forge/screens/settings/SettingsPage.java
```diff
@@ -154,7 +154,7 @@ public void select() {
             lstSettings.addItem(new CustomSelectSetting(FPref.UI_VIDEO_MODE,
                 Forge.getLocalizer().getMessage("lblVideoMode"),
                 Forge.getLocalizer().getMessage("nlVideoMode"),
-                new String[]{"720p", "768p", "900p", "1080p"}) {
+                new String[]{"720p", "768p", "900p", "1080p", "1440p", "2160p"}) {
```
tool4ever, line 157: Should probably have this as a static string that both classes can use, maybe even try and combine those identical calls below into something callable?

Agetian: Good idea! I wonder if Adventure Mode actually has a certain class that stores static constants, hmm... I'll look into it :)

kevlahnota (dismissed):

forge-gui-mobile/src/forge/screens/settings/SettingsPage.java
Agetian: @kevlahnota I'm looking into ways of consolidating the video resolution modes into a single static array that would be referenced from the two places in SettingsPage and SettingsScene. However, I'm having trouble identifying the best place where to consolidate stuff like that (as well as possibly make a static method that would set the exact width and height dimensions depending on the chosen video mode). It doesn't look like forge-gui has a defined place for static stuff like that, or I don't see it... My current ideas would be to either move it over to Graphics (though it seems slightly clunky as that class doesn't really have any public static class and isn't otherwise used that way), add it to the Forge class (which does have some public static stuff in it, but not quite along the same vein as I want), or create a separate class (such as VideoModes) somewhere and implement the relevant stuff there. Maybe, however, I don't see something obvious since I haven't worked much with the mobile UI code base. What would you recommend here? Do you have any pointers for what would be the optimal place for this? Thanks a lot in advance!

kevlahnota: > @kevlahnota I'm looking into ways of consolidating the video resolution modes into a single static array that would be referenced from the two places in SettingsPage and SettingsScene. However, I'm having trouble identifying the best place where to consolidate stuff like that (as well as possibly make a static method that would set the exact width and height dimensions depending on the chosen video mode). It doesn't look like forge-gui has a defined place for static stuff like that, or I don't see it... My current ideas would be to either move it over to Graphics (though it seems slightly clunky as that class doesn't really have any public static class and isn't otherwise used that way), add it to the Forge class (which does have some public static stuff in it, but not quite along the same vein as I want), or create a separate class (such as VideoModes) somewhere and implement the relevant stuff there. Maybe, however, I don't see something obvious since I haven't worked much with the mobile UI code base. What would you recommend here? Do you have any pointers for what would be the optimal place for this? Thanks a lot in advance!

I was planning to put them on ForgeConstants before, and just get the defined supported resolution from it.  Anything that will look clean and remove those duplicated code should do good.

Agetian: Attempted to do some code juggling and refactoring along the lines defined above.

tool4ever (approved):

PR #3087, opened 2023-05-09
Phyrexian Shops Update

EldritchBimbo:
Changes:
.Adds 6 shops for Phyrexian theme (1 per region.)
.Adds sign + infrastructure to indicate Phyrexian shops. 
.Adds Phyrexian shops to towns - mythic rarity in capital and generic, rare rarity in tribal.


jjayers99:
Would recommend using the "Incubate" keyword instead of "Incubator". Even if these cards are appearing as expected now, there's no guarantee that the keyword will always have reminder text.

Agetian:
@jjayers99 It looks like the shop definition already has both "incubate" and "incubator" listed as keywords, or were you referring to some other instance?

jjayers99:
I've really got to stop looking over these on my phone. In that case, I
think "incubator" should be removed, as it would only add a couple false
positives.

https://scryfall.com/search?as=grid&order=name&q=Incubator+%28game%3Apaper%29

Agetian:
Yeah, good call :+1:

kevlahnota (approved):

PR #3089, opened 2023-05-10
Stored kw idx

### LINKED ISSUE #3077: Zhulodok, Void Gorger
Static Abilities currently can't handle multiple keyword instances of the same String

meaning a Static Ability giving "cascade; cascade" currently can't differ between both strings
so we probably need to use an int index for this (and may need to be stored in the Keyword Instance?)

Hanmac:
When Multiple Keywords are added via StaticAbility, they all get an idx to makes duplicates possible

Closes #3077

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -4658,7 +4658,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
 
     public final KeywordInterface getKeywordForStaticAbility(String kw, final long staticId, final long idx) {
         KeywordInterface result;
-        if (staticId < 1 || !storedKeywords.contains(staticId, kw)) {
+        if (staticId < 1 || !storedKeywords.contains(kw, ImmutablePair.of(staticId, idx))) {
             result = Keyword.getInstance(kw);
             result.setStaticId(staticId);
             result.createTraits(this, false);
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -4658,15 +4658,16 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
 
     public final KeywordInterface getKeywordForStaticAbility(String kw, final long staticId, final long idx) {
         KeywordInterface result;
-        if (staticId < 1 || !storedKeywords.contains(kw, ImmutablePair.of(staticId, idx))) {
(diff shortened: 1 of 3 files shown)


Agetian (dismissed):

Agetian:
Looks good to me! :+1:

forge-gui/res/cardsfolder/upcoming/zhulodok_void_gorger.txt
```diff
@@ -0,0 +1,6 @@
+Name:Zhulodok, Void Gorger
+ManaCost:5 C
+Types:Legendary Creature Eldrazi
+PT:7/4
+S:Mode$ Continuous | Affected$ Card.Colorless+YouCtrl+YouOwn+wasCastFromHand+cmcGE7 | AffectedZone$ Stack | AddKeyword$ Cascade & Cascade | Description$ Spells you cast with mana value 6 or greater have cascade.
+Oracle:Colorless spells you cast from your hand with mana value 7 or greater have “Cascade, cascade.” (When you cast one, exile cards from the top of your library until you exile a nonland card that costs less. You may cast it without paying its mana cost. Put the exiled cards on the bottom of your library in a random order. Then do it again.)
```
Simisays, line 6: Maybe DeckHints:Type$Eldrazi|Construct|Golem ?

Hanmac:
i need to see if `getCopyForStoredKeyword´ works the way i want, but i don't know a case where i can test that or where it matters

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4653,53 +4656,54 @@ public final void addChangedCardKeywords(final List<String> keywords, final List
...
-            result = storedKeywords.get(staticId, kw);
+            result = storedKeywords.get(triple);
         }
         return result;
     }
 
     public final void addKeywordForStaticAbility(KeywordInterface kw) {
         if (kw.getStaticId() > 0) {
-            storedKeywords.put(kw.getStaticId(), kw.getOriginal(), kw);
+            storedKeywords.put(Triple.of(kw.getOriginal(), kw.getStaticId(), 0l), kw);
```
Hanmac, line 4677: @tool4ever can you test if that is good enough? like with our friend Henzi Toolbox?

tool4ever: Anything specific? He seems to still work from what I can tell

Hanmac: because i didn't copy the idx, i thought it might get duplicated entities (and resulting in two Blitz trigger?)

PR #3092, opened 2023-05-10
Clean up

tool4ever:


forge-game/src/main/java/forge/game/trigger/TriggerRevealed.java
```diff
@@ -7,6 +7,7 @@
 import forge.game.spellability.SpellAbility;
 import forge.util.Localizer;
 
+@Deprecated
```
Hanmac, line 10: Why is the Trigger Deprecated when its still in use of TriggerType?

tool4ever: There‘s no card that makes use of it

Hanmac: yeah its for old Miracle stuff, i think the Trigger can be removed

PR #3095, opened 2023-05-11
Update CardFactory CloneStates: fix SetPower and setToughness

Hanmac:
fix around SetPower and setToughness

@tool4ever this should fix the possible NPE, or do you still want to debug why originalState might be null?

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/CopyPermanentEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/CopyPermanentEffect.java
@@ -264,19 +264,17 @@ public class CopyPermanentEffect extends TokenEffectBase {
     public static Card getProtoType(final SpellAbility sa, final Card original, final Player newOwner) {
         final Card host = sa.getHostCard();
         int id = newOwner == null ? 0 : newOwner.getGame().nextCardId();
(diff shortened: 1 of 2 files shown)


Agetian (dismissed):

tool4ever:
I don't know if it's a problem yet 🤔 
If _Scarab God_ makes a token of _Arcee, Sharpshooter_ should both sides be 4/4 or is it correct if the Vehicle back stays 2/2?

Hanmac:
> I don't know if it's a problem yet thinking If _Scarab God_ makes a token of _Arcee, Sharpshooter_ should both sides be 4/4 or is it correct if the Vehicle back stays 2/2?

i think i need to ask judges for this, imo the Vehicle probably needs to be a 4/4 too

Hanmac:
@tool4ever 
> Hydra17: the modification would apply to both faces of the TDFC, as far as circumstances allow. in the case of the Vehicle, it is black, and will be a 4/4 while animated. but it is not a Zombie

Hanmac:
i needed the Original Card States from `CardFactory.getCard`

Agetian (approved):

tool4ever (approved):
Seems to work 👍

tool4ever:
Copied permanent spells don't seem to be able to transform at all, not sure why yet

PR #3107, opened 2023-05-13
Draft: Battle AI support [highly sketchy]

Agetian:
Warning: Here be dragons. Or, rather, highly sketchy, ugly test code that sort of, kind of works in certain scenarios and is meant more as a demonstration of what's necessary to do to support Battle AI than serve as the definitive implementation.

This is the first attempt at a very basic sketchy draft of what could become the foundation of the future AI support for Battles.

What it currently does:
- It modifies the AI attack controller in a way that allows it to overcome the limitations of the current implementation and determine the Battles that can be attacked and scenarios in which it's possible to attack them. For the most part, this works, but it's very basic and there is no randomness or chance variation at which this will happen. Thus, the AI *will* attack its own Battles given the opportunity (within the current constraints).
- It modifies the choice of the defending opponent when a Battle is the defender card so that the AI chooses the opposing player (e.g. the defender of the said Battle) if the "defender" is set to a Battle. Before this was done, the AI attacked itself with the remaining creatures that it didn't want to assign to attack a Battle (rather than attacking the opponent).
- It adds an [ugly and hacky] way for the AI to play the flip side of the Battle once it loses all its counters. Note that this is done as a workaround for now - the proper way to do it would be to modify the code in PlayAi and probably elsewhere, too, to properly determine that the flip side of a DFC card is being queried about. I tried a few things but couldn't determine a way to do it - neither method that returns spells or SAs seems to work for getState(CardStateName.Transformed). Help needed here in the form of code contributions.

What it currently DOESN'T do:
- It doesn't yet allow the AI to defend Battles. In fact, there seems to be no way at all to defend a Battle at the moment, even for a human player (see below), so this will most likely need a mechanics update too.
- It doesn't yet introduce any form of advanced parametrization, heuristics, or profile options for attacking/defending Battles, so the AI support is rudimentary and highly predictable.
- It doesn't introduce the cleanest AI support possible. Most certainly this can be done in a multitude of better ways, the point of this PR for now is to highlight the possibilities and the potential issues so that better code can be written and contributed.

What problems it highlights in the Battle implementation as a whole:
- At the very least, it seems like blocking mechanics isn't updated to account for Battles in any way. During the Declare Blockers step, the declareBlockers call isn't made for the player who is the defender of the Battle, which makes it impossible for both the human and the AI to actually defend a Battle.
- The AI may need better hooks in the attack/block controllers for cleaner support code (e.g. better integrated with the code that defines creature lists and chooses defenders and defending opponents), but this can be done later once the basic support and all the mechanics are figured out.

Please note that I would welcome any and all help in the form of code contributions. Feel free to improve this PR and submit code improvements to it. Unfortunately, I don't have enough spare time at the moment in order to complete this project on my own, and I don't have the necessary knowledge to improve the game mechanics to support e.g. defending Battles. Thanks a lot in advance for help!

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -22,7 +22,6 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 
-import forge.card.CardType;
 import forge.game.staticability.StaticAbility;
 import forge.game.staticability.StaticAbilityAssignCombatDamageAsUnblocked;
 import org.apache.commons.lang3.tuple.Pair;
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -22,6 +22,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 
+import forge.card.CardType;
 import forge.game.staticability.StaticAbility;
 import forge.game.staticability.StaticAbilityAssignCombatDamageAsUnblocked;
 import org.apache.commons.lang3.tuple.Pair;
@@ -120,6 +121,10 @@ public class AiAttackController {
 
     private void refreshCombatants(GameEntity defender) {
         this.oppList = getOpponentCreatures(defendingOpponent);
+        if (defender instanceof Card && ((Card) defender).isBattle()) {
+            this.oppList.clear();
+            this.oppList.addAll(getOpponentCreatures(((Card) defender).getProtectingPlayer()));
+        }
         this.attackers = new ArrayList<>();
         for (Card c : myList) {
             if (canAttackWrapper(c, defender)) {
@@ -722,9 +727,21 @@ public class AiAttackController {
             return pwNearUlti != null ? pwNearUlti : ComputerUtilCard.getBestPlaneswalkerAI(pwDefending);
         }
 
-        List<Card> battleDefending = c.getDefendingBattles();
+        List<Card> battleDefending = CardLists.filter(c.getDefendingBattles(),
+                Predicates.or(CardPredicates.isControlledByAnyOf(ai.getAllies()), CardPredicates.isController(ai)));
+        Card bestBattle = null;
         if (!battleDefending.isEmpty()) {
-            // TODO filter for team ones
+            // Get the battle that is the closest to triggering
+            int defCounters = Integer.MAX_VALUE;
+            for (Card battle : battleDefending) {
+                if (battle.getCounters(CounterEnumType.DEFENSE) < defCounters) {
+                    defCounters = battle.getCounters(CounterEnumType.DEFENSE);
+                    bestBattle = battle;
+                }
+            }
+            if (bestBattle != null) {
+                return bestBattle;
+            }
         }
 
         return prefDefender;
@@ -756,7 +773,17 @@ public class AiAttackController {
         // decided to attack another defender so related lists need to be updated
         // (though usually rather try to avoid this situation for performance reasons)
         if (defender != defendingOpponent) {
-            defendingOpponent = defender instanceof Player ? (Player) defender : ((Card)defender).getController();
+            if (defender instanceof Player) {
+                defendingOpponent = (Player) defender;
+            } else if (defender instanceof Card) {
+                Card defCard = (Card) defender;
+                if (defCard.isBattle()) {
+                    defendingOpponent = defCard.getProtectingPlayer();
+                } else {
+                    // TODO: assume Planeswalker for now, may need to be updated later if more unique mechanics appear like Battle
+                    defendingOpponent = defCard.getController();
+                }
+            }
             refreshCombatants(defender);
         }
         if (this.attackers.isEmpty()) {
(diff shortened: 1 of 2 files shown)


Agetian:
Made several modifications to this PR today which allows Battles to be defended (both by the human and by the AI player) and added some rudimentary support to the AI blocking controller so that the AI actually defends Battles it's protecting.

Agetian:
I tried some shenanigans to try and get the AI to meaningfully evaluate the transformed side of the Battle in PlayAi, e.g.:

```
                    List<SpellAbility> sas = AbilityUtils.getBasicSpellsFromPlayEffect(c, c.getController(), CardStateName.Transformed);
                    SpellAbility saTransformed = null;
                    for (SpellAbility ab : sas) {
                        if (ab.getCardStateName() == CardStateName.Transformed) {
                            saTransformed = ab;
                            break;
                        }
                    }
                    if (saTransformed != null) {
                        Spell sp = (Spell) saTransformed;
                        if (AiPlayDecision.WillPlay == ((PlayerControllerAi)ai.getController()).getAi().canPlayFromEffectAI(sp, !(isOptional || sa.hasParam("Optional")), true)) {
                            return sp.isTargetNumberValid(); // if targeting failed, don't choose the card
                        }
                    }

```

However, this doesn't work - the AI still evaluates the resulting spell ability as having "bad ETB effects" :/

tool4ever:
Should work now, you were close as you can see

Just don't have time to test it fully now

Agetian:
Thanks a lot for help, @tool4ever ! I'll test it a little soon.

Agetian:
Seems to work great for Battles, @tool4ever :+1: 
Any non-Battle related scenarios we would want to test here? Nothing comes to mind immediately :/

Agetian:
At CombatUtil.java:85, you mean the allies of playerWhoAttacks? Something along the lines of the following?

```
        for (Player p : playerWhoAttacks.getAllies()) {
            defenders.addAll(p.getBattlesInPlay());
        }

```

Hanmac:
Are battles that lost all counters because of combat, getting exiled and cast to battlefield again removed from combat yet? The arrows are still pointing in my older snapshot

Agetian:
No, doesn't look like they are removed.

Agetian:
It will be removed from combat if we update GameAction.java circa line 455 in the following way, but I'm not sure if it's the desired solution:

```
            if (fromBattlefield && (c.isCreature() || c.isBattle()) && game.getCombat() != null) {
                if (!toBattlefield) {
                    game.getCombat().saveLKI(lastKnownInfo);
                }
                game.getCombat().removeFromCombat(c);
            }

```

tool4ever:
> At CombatUtil.java:85, you mean the allies of playerWhoAttacks? Something along the lines of the following?
> 
> ```
>         for (Player p : playerWhoAttacks.getAllies()) {
>             defenders.addAll(p.getBattlesInPlay());
>         }
> ```

We need to check which cases are allowed, e. g. currently you could attack a battle an ally protects?

Maybe the whole logic can be simplified to: "battle in play protected by opponent"

Agetian:
No, that function currently doesn't work for Battles - the Battle card doesn't end up anywhere in the iterated collections so it's not picked up. It's possible to clean it up from there if we get the key of attackedBands and see if it's a card and it's no longer present on the battlefield, e.g.:

```
            GameEntity attacked = ee.getKey();
            if (attacked instanceof Card) {
                Card attackedCard = (Card) attacked;
                if (!attackedCard.isInPlay()) {
                    missingCombatants.add(attackedCard);
                }
            }

```

tool4ever:
Ok thanks I went with what felt more reasonable

Agetian:
No prob! Thanks for your help! I agree with your choice.

forge-game/src/main/java/forge/game/combat/Combat.java
```diff
@@ -653,8 +653,7 @@ public final boolean removeAbsentCombatants() {
         CardCollection missingCombatants = new CardCollection();
         for (Entry<GameEntity, AttackingBand> ee : attackedByBands.entries()) {
             CardCollectionView atk = ee.getValue().getAttackers();
-            for (int i = atk.size() - 1; i >= 0; i--) { // might remove items from collection, so no iterators
-                Card c = atk.get(i);
+            for (Card c : atk) {
```
Hanmac, line 656: How about this:

`for (Card c : ee.getValue().getAttackers()) {`

Agetian: Yep, agreed :)

Agetian:
> We need to check which cases are allowed, e. g. currently you could attack a battle an ally protects?
> Maybe the whole logic can be simplified to: "battle in play protected by opponent"

I can suggest something like this, but maybe it can be written in a cleaner manner:

```
        final Game game = playerWhoAttacks.getGame();
        final CardCollection battles = CardLists.filter(game.getCardsIn(ZoneType.Battlefield), CardPredicates.Presets.BATTLES);
        for (Card battle : battles) {
            if (battle.getType().hasSubtype("Siege") && battle.getProtectingPlayer().isOpponentOf(playerWhoAttacks)) {
                defenders.add(battle);
            }
        }
```

Agetian:
Alrighty, I think this is as far as I can take this particular PR. I suggest that the further work on the mechanics is carried on in followup PRs.

Agetian:
That's a good point, I'll see if I can take a look at that. In fact, the decision to attack a certain Battle (whether own or opponent's) is pretty complex, and ideally we'll have to introduce some interesting diversity and strategic choices into the mix to make the AI smarter and less predictable, but imo this should be reserved for a followup PR, since it's a pretty complex task in its own right.

tool4ever (approved):
Nice work 🧙‍♂️

PR #3116, opened 2023-05-15
Some script and engine fixes

### LINKED ISSUE #3115: Marchesa, Resolute Monarch (Invasion of Fiora) requires target, when it should be "up to one"
**Describe the bug**
When Marchesa, Resolute Monarch attacks, triggered ability triggers, "Whenever Marchesa, Resolute Monarch attacks, remove all counters from up to one target permanent." When this trigger hits the stack, you should be able to select no targets, but currently it requires selection of a target.

**To Reproduce**
1. Have Marchesa, Resolute Monarch on the battlefield and attack with it
2. Observe that target selection cannot be skipped

**Expected behavior**
Should allow a selection of a target permanent OR no permanents since it says "up to one"

 - OS: Windows
 - Version: 1.6.57-SNAPSHOT-05.15

tool4ever:
@Hanmac 
New _Niv-Mizzet, Supreme_ got future-broken by https://github.com/Card-Forge/forge/commit/9d2ecf
There's probably another way it could be fixed but I'm not sure how it could be done with castKeyword and the `OptionalCost.Jumpstart` that modifies the spell?

Closes #3115

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -591,9 +591,7 @@ public class GameAction {
         // 400.7g try adding keyword back into card if it doesn't already have it
         if (zoneTo.is(ZoneType.Stack) && cause != null && cause.isSpell() && !cause.isIntrinsic() && c.equals(cause.getHostCard())) {
             if (cause.getKeyword() != null && !copied.getKeywords().contains(cause.getKeyword())) {
-                copied.addChangedCardKeywordsInternal(ImmutableList.of(cause.getKeyword()), null, false, game.getNextTimestamp(), 0, false);
-                // update Keyword Cache
-                copied.updateKeywords();
+                copied.addChangedCardKeywordsInternal(ImmutableList.of(cause.getKeyword()), null, false, game.getNextTimestamp(), 0, true);
             }
         }
 
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -264,7 +264,6 @@ public class GameAction {
 
             if (zoneTo.is(ZoneType.Stack)) {
                 // try not to copy changed stats when moving to stack
-                copied.setChangedCardKeywords(c.getChangedCardKeywords());
 
                 // copy exiled properties when adding to stack
                 // will be cleanup later in MagicStack
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -264,6 +264,7 @@ public class GameAction {
 
             if (zoneTo.is(ZoneType.Stack)) {
                 // try not to copy changed stats when moving to stack
+                copied.setChangedCardKeywords(c.getChangedCardKeywords());
 
                 // copy exiled properties when adding to stack
                 // will be cleanup later in MagicStack
--- a/forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
@@ -636,10 +636,8 @@ public abstract class SpellAbilityEffect {
             combat.initConstraints();
             if (sa.hasParam("ChoosePlayerOrPlaneswalker")) {
                 PlayerCollection defendingPlayers = AbilityUtils.getDefinedPlayers(sa.hasParam("ForEach") ? c : host, attacking, sa);
-                defs = new FCollection<>();
-                for (Player p : defendingPlayers) {
-                    defs.addAll(combat.getDefendersControlledBy(p));
-                }
+                defs = new FCollection<>(defendingPlayers);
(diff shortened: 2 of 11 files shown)


Hanmac:
technically Jumpstart is not a cast Keyword, the Keyword there is explicit not an ability that allows you to cast the Keyword (i think) it is not an Alternative Cost Keyword like Flashback, but an Additional Cost Keyword

and the Paid Keyword sadly doesn't know about their keyword instance (because i think some of them needs to be Serialized for GUI)

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -264,6 +264,7 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
 
             if (zoneTo.is(ZoneType.Stack)) {
                 // try not to copy changed stats when moving to stack
+                copied.setChangedCardKeywords(c.getChangedCardKeywords());
```
Hanmac, line 267: i still don't know if that would cause other problems, i think there was a reason why i tried to remove it

tool4ever: Yea I agree but I haven't found a good alternative yet

Hanmac: does it cause problems with other cards like Henzi or others what have a CastBy Keyword?

tool4ever: Still unsure but it looks like your suspicion was right and #3089 did in fact introduce a regression for Henzie:
image

This is on the stack, when it ETB it seems to get cleaned up again (confirmed it's not a View problem)

It doesn't always happen, but usually when you cast it directly from extrinsic Blitz

How to fix?

Hanmac: it wasn't directly broken by this one i think, for the fix, i just need to store the idx inside the Keyword like i did with the static Ability id

Hanmac: This part should be fixed, right?

tool4ever: yea Henzie is ok

but you were right still need another solution for Niv-Mizzet
this one causes problems when combining extrinsic and intrinsic flashback

Hanmac: You should try to move the other fixes into a separate MR

tool4ever: Fixed it via script for now instead, there should be nothing where it matters
Another approach for an engine fix can be attempted later

forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
```diff
@@ -636,10 +636,8 @@ protected static boolean addToCombat(Card c, Player controller, SpellAbility sa,
             combat.initConstraints();
             if (sa.hasParam("ChoosePlayerOrPlaneswalker")) {
                 PlayerCollection defendingPlayers = AbilityUtils.getDefinedPlayers(sa.hasParam("ForEach") ? c : host, attacking, sa);
-                defs = new FCollection<>();
-                for (Player p : defendingPlayers) {
-                    defs.addAll(combat.getDefendersControlledBy(p));
-                }
+                defs = new FCollection<>(defendingPlayers);
+                defs.addAll(Iterables.filter(combat.getDefendingPlaneswalkers(), CardPredicates.isControlledByAnyOf(defendingPlayers)));
```
Hanmac, line 640: isn't isControlledByAnyOf wrong? shouldn't that be Protector?

also shouldn't you make new MR?

tool4ever: Myriad and the few other cards that use it should not interact with battles...
>attacking that player or a planeswalker they control

kevlahnota (approved):

forge-gui/res/cardsfolder/n/niv_mizzet_supreme.txt
```diff
@@ -4,7 +4,7 @@ Types:Legendary Creature Dragon Avatar
 PT:5/5
 K:Flying
 K:Hexproof:Card.MonoColor:monocolored
-S:Mode$ Continuous | Affected$ Instant.YouCtrl+numColorsEQ2,Sorcery.YouCtrl+numColorsEQ2 | AffectedZone$ Graveyard | AddKeyword$ Jump-start | Description$ Each instant and sorcery card in your graveyard that's exactly two colors has jump-start.
+S:Mode$ Continuous | Affected$ Instant.YouCtrl+numColorsEQ2,Sorcery.YouCtrl+numColorsEQ2 | AffectedZone$ Graveyard,Stack | AddKeyword$ Jump-start | Description$ Each instant and sorcery card in your graveyard that's exactly two colors has jump-start.
```
Hanmac, line 7: I think this needs a `#TODO` so it should be fixed sooner than later

kevlahnota (approved):

PR #3119, opened 2023-05-15
AI: Fix counters calculation for Wildborn Preserver activation

Agetian:
Experimental, needs testing.


forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
```diff
@@ -777,38 +756,6 @@ protected boolean doTriggerAINoCost(Player ai, SpellAbility sa, boolean mandator
             // No target. So must be defined
             list = AbilityUtils.getDefinedCards(source, sa.getParam("Defined"), sa);
 
-            if (amountStr.equals("X")
```
tool4ever, line 780: Uhh no, it's not just for calculation but also important for preparing the right amount on "standard" triggers

So without this X will always be 0 on cards like _Hero of Leina Tower_

Agetian: Hmm, interestingly, Hero of Leina Tower seems to work fine for the AI without this code, at keast as far as X specification goes.

If the code is indeed still useful though, do you have a suggestion for how to make it work with e.g. Wildborn Prrserver?

tool4ever: Huh that's rather weird, it certainly doesn't for me and I really can't think of another place where the AI might obtain the right X for it
Can you check again?

For fixing the Preserver I'd have to debug a bit more later

Agetian: Sure , I'll test a bit more tomorrow and see if maybe I missed something.

For Wildborn Preserver, this code overrides X and writes -2 into it 😅

Agetian: Ok, retested and indeed , Hero of Leina Tower doesn't work without it.

Looks like we need to keep the code but somehow make it play ball with the likes of Wildborn Preserver.

The problem there is that X is already correctly set up by the immediate trigger , but this code doesn't discern that case and overwrites the X definition with the wrong number ( -2 ). I haven't yet found a way to effectively discern between these cases.

Agetian: Alright, tried a more conservative fix where "amount == 0" is tested so that if something already set the amount before, it's not done again. Seems to work with all the cards the original code was meant for and also for Wildborn Preserver, but I don't know if maybe this also breaks something :/

Hanmac: we may check if the ability is inside a DelayedTrigger or ImmediateTrigger

Agetian: How can we do it? I haven't figured out a way to test that :/ Would appreciate a hint, hehe :)

Hanmac: i was thinking about setSpawningAbility but this might not be set when AI is thinking about it

Agetian: Hmm, couldn't find a way to get to getSpawningAbility from there... The best I was able to grab so far is the fact that the SA is coming from a trigger (sa.isTrigger() == true) and it's possible to get the trigger itself via sa.getTrigger(), which doesn't have any indication by itself as to whether it's immediate/delayed/etc. Not sure what this gives us, unfortunately :/

Hanmac: like i said, IT ISN'T SET yet when the AI is thinking about it

Agetian: Yep, and sadly, I don't see other alternative means of getting this info either

tool4ever: Ok so I also tried to figure out if all that calculation should happen and agree that it seems pretty redundant
- `ComputerUtilCost.getMaxXValue` is already returning the right amount that can be used based on other shards present
- X for ETB counters doesn't really get decided there either, that's done by `PermanentAi` shouldn't be

I will add my cleanup shortly, it does seems like we might be able to reuse it for _Wildborn Preserver_
- When it resolves the "standard trigger" `CountersPutAi` can find the parent and set the X
- When `ImmediateTrigger` resolves the parent will be removed and everything should stay untouched

Agetian: Very nice job on the cleaner implementation, @tool4ever , thank you for help! :)
I tested it and it looks like it works for both cases (things like Hero of Leina Tower and Endless One, and for things like Wildborn Preserver).

tool4ever (approved):

PR #3124, opened 2023-05-16
Fix the file name of the swamp tribal town.

Agetian:
Fixes entering these towns when running Forge on case sensitive file systems (Linux, MacOS).


Simisays (approved):
I'll keep this in mind, not using lower case file names

PR #3126, opened 2023-05-17
Fix AI not choosing to fight when putting counters from a trigger

### LINKED ISSUE #3125: Invasion of Muraganda - AI doesn't fight (A)
**Describe the bug**
After the ETB trigger on the Invasion of Muraganda, the AI never chooses to fight after giving its creature the +1/+1 counter, even if it would have favorable results.

**To Reproduce**
1. Give AI a vanilla 3/3 and Invasion of Muraganda in hand. Give human player a vanilla 2/2.
2. After casting Invasion of Muraganda, AI will give +1/+1 counter to its 3/3, but not choose to fight the 2/2.

**Expected behavior**
Would expect AI to target a creature for favorable fight 

 - OS: Windows
 - Version 1.6.57-SNAPSHOT-05.16

Agetian:
Fixes #3125 

This fixes the immediate issue, but it also brings up a bigger issue for a potential rework later - there are lots of logic-related checks in checkApiLogic, none of which are tested from doTriggerAINoCost or chkAIDrawback. I thought that unifying them elsewhere (in checkAiLogic, or maybe even as a separate method called from all the checks) would be straightforward enough, but it's not as easy as it seems at first - some of those logic checks depend on certain things being set (such as the counter amount or having targets set), some aren't; some require additional variables to be set in the calling code (e.g. playAggro), some don't need it. So, there's a very high chance of breaking stuff in trying to extract this part of the method. Most likely a target for a potential future rework. For now, I merely copied the Fight logic into doTriggerAINoCost so that it works for the Invasion of Muraganda trigger.


forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
```diff
@@ -877,6 +877,18 @@ protected boolean doTriggerAINoCost(Player ai, SpellAbility sa, boolean mandator
                 }
             }
         }
+
+        // TODO: Unify the AI logic checking code elsewhere so that it can be checked similarly
+        // for AB, DB, or trigger PutCounter alike.
+        String logic = sa.getParam("AILogic");
```
tool4ever, line 883: You should move this at the beginning of the relevant targeting if (L791)
Because the trigger is mandatory, so even if there's no favorable Fight (false returned) you still need to target (and at least getting a counter out of it is obviously also desirable)

Agetian: :+1: Good call! Done :)

tool4ever: You can only return if canFight() is true (which indicates targeting was successful)

Agetian: Ah, true :)

tool4ever (approved):

PR #3130, opened 2023-05-17
Use current shard when trying to pay with combo mana

tool4ever:
Needs some testing...

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -1507,7 +1507,7 @@ public class AiAttackController {
                 }
                 if (sa.usesTargeting()) {
                     sa.setActivatingPlayer(c.getController(), true);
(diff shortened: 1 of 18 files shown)


Agetian:
I ran some tests and this seems to work fine for the AI.
Do we need an additional "guard" test along the lines of the following? Not sure if it can matter at some point, couldn't think of a test case where this would fail, but just thought that maybe it can be a good precaution against the AI somehow finding a way to cheat with this? What do you think?

```
            if (m.canProduce(MagicColor.toShortString(toPay.getColorMask()), sa)) {
                m.setExpressChoice(ColorSet.fromMask(toPay.getColorMask()));
            }

```

Agetian (dismissed):

tool4ever:
Already checked by `getComboManaChoice`

Agetian:
Alrighty then, I don't think I see an immediate way of breaking it and it works in whatever I tried, so... I guess let's integrate it and see what happens? :)

PR #3133, opened 2023-05-18
AI: Evaluate creature spells when deciding between two spells with the same CMC

Agetian:
Pretty sure this used to work before, but something must have changed since then.
Right now the AI manifests somewhat strange priorities - if e.g. it has four CMC 3 creature spells, it'll just play the first one that it has, instead of playing the best one. The current decision chain doesn't feature evaluateCreature in any capacity in order to decide between these spells, so this PR adds a simple priority system where a creature spell that evaluates to a higher score will evaluate slightly higher than a creature spell that evaluates to a lower score.


Agetian:
Maybe ".getCardState().getCard()" is more appropriate if it allows to grab the specific side of the card in question?

forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -1082,6 +1082,17 @@ public int compare(final SpellAbility a, final SpellAbility b) {
                 }
             }
 
+            // If both are permanent creature spells, prefer the one that evaluates higher
+            if (a1 == b1 && a.getApi() == ApiType.PermanentCreature && b.getApi() == ApiType.PermanentCreature) {
+                int evalA = ComputerUtilCard.evaluateCreature(a.getCardState().getCard());
```
tool4ever, line 1087: Nice observation, I would have also thought this would have been wired up but couldn't find a trace of it either

The problem here is `getCardState().getCard()` does not return another switched version of the host

Currently I only know as fix a temporary manual switch of the state (so you could probably add helper `evaluateCreature(SA)`)

Similar to how I did it here:
https://github.com/Card-Forge/forge/blob/6de7dd6bcc4f6be5357ff0cd32f3426caa4a65de/forge-ai/src/main/java/forge/ai/AiController.java#L737

tool4ever: Or try and rewrite `evaluateCreature` to query CardState from param directly

Agetian: Oh, nice, thanks a lot for the hint! I'll see if I can figure it out later today :)

Agetian: Tried something similar to what you demonstrated, please take a look when able :)

tool4ever (approved):

PR #3137, opened 2023-05-19
Update ghostfire.txt

Klisz:
Fixes a bug where Ghostfire is considered to have colorless *identity* in addition to being colorless itself, and thus is allowed to appear in Commander, Oathbreaker, etc decks that shouldn't be allowed to have it due to lacking red in their color identity.


Hanmac (approved):
It's more in line with the oracle text

Issue #2866, opened 2023-04-08
Transforming Copy Token


Hanmac:
> This change will also apply to tokens that are created as copies of other transforming double-faced permanents or as copies of transforming double-faced cards in zones other than the battlefield. For example, if you control a Flame Channeler and an effect creates a token that's a copy of it, the resulting permanent is a transforming double-faced token that is a copy of Flame Channeler. It enters the battlefield with its front face up, and it can transform. If it does, it will become an Embodiment of Flame, the back face of Flame Channeler. Similarly, if you controlled an Embodiment of Flame and created a token that is a copy of it, the token would be the same transforming double-faced token, but it would enter with its back face up.

That means ONLY for CopyPermanent should the Transforming Backside be copied.
For TDFC Clones entering as Copy of another TDFC, the Backside should not be copied.

not sure how that is for Spells that enter as token of TDFC Spells (when the spell is copied on the stack)

Hanmac:
some of it got accidentally commited into master

@tool4ever you probably should use this ticket to document errors you find with this

tool4ever:
Sure, I'll test some more later

Here's the crash after https://github.com/Card-Forge/forge/commit/bb3709ebad509a316df0ccaffec4546082764d67:
```
Game-3 > java.lang.NullPointerException: Cannot invoke "forge.game.card.CardState.getStateName()" because "originalState" is null
    at forge.game.card.CardFactory.getCloneStates(CardFactory.java:812)
    at forge.game.ability.effects.CopyPermanentEffect.getProtoType(CopyPermanentEffect.java:285)
    at forge.game.ability.effects.CopyPermanentEffect.resolve(CopyPermanentEffect.java:247)
```

tool4ever:
Bug: I defeated _Invasion of Amonkhet_ and copied the backside spell with _Lithoform Engine_

The new token looked ok but somehow still got 4 defense counters

(Not affected by #3095)

tool4ever:
Copied permanent spells don't seem to be able to transform at all, not sure why yet

Hanmac:
> Copied permanent spells don't seem to be able to transform at all, not sure why yet

i did only Copy Token yet, not Copy Spell

Hanmac:
@tool4ever @Agetian i need to look at `copySpellHost` and `copyCopiableCharacteristics`
Specially `copyCopiableCharacteristics` is only used on one location, and might need to be cleaned up too

Hanmac:
[...]

PR #3139, opened 2023-05-20
CardFactory: copySpellHost uses getCloneStates

Hanmac:
Handles the SpellCopy part of #2866

- [ ] remove copyCopiableCharacteristics use in GameCopier
- [ ] maybe remove copyStats and remove copyState

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardFactory.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactory.java
@@ -124,62 +124,49 @@ public class CardFactory {
         final Card source = sourceSA.getHostCard();
         final Card original = targetSA.getHostCard();
         final Game game = source.getGame();
-        final Card c = new Card(game.nextCardId(), original.getPaperCard(), game);
-        copyCopiableCharacteristics(original, c, sourceSA, targetSA);
+        int id = game.nextCardId();
 
-        if (sourceSA.hasParam("NonLegendary")) {
-            c.removeType(CardType.Supertype.Legendary);
-        }
+        // need to create a physical card first, i need the original card faces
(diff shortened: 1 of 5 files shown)


Agetian (approved):

Hanmac:
@Agetian @tehdiplomat @tool4ever should we do the removal of copyCopiableCharacteristics in this MR or in a later one?

forge-game/src/main/java/forge/game/card/CardFactory.java
```diff
@@ -124,62 +124,49 @@ private final static Card copySpellHost(final SpellAbility sourceSA, final Spell
         final Card source = sourceSA.getHostCard();
         final Card original = targetSA.getHostCard();
         final Game game = source.getGame();
-        final Card c = new Card(game.nextCardId(), original.getPaperCard(), game);
-        copyCopiableCharacteristics(original, c, sourceSA, targetSA);
+        int id = game.nextCardId();
```
tool4ever, line 127: Wasn't the old method faster because we didn't have to rebuild the old card completely? (at least for single faced cards)

So while the last call is only done in simulation for tokens it means tests still use it so we might not see if it broke something? 🤔

Hanmac: We don't "rebuild" the whole card for copySpellHost, but just enough so I can apply copy states

We need to add more tests for this one

Hanmac:
As for the remove of copyCharacteristics, we still might need to rethink it over, even if it's just used for tests

forge-game/src/main/java/forge/game/card/CardFactory.java
kevlahnota: gonna merge this, tests can be added on a later PR

PR #3140, opened 2023-05-20
Improve AI for Tempered Veteran

### LINKED ISSUE #2563: AI using Tempered Veteran's more expensive ability even though creature already had +1/+1 counters
image

Could have spent just W and had another pump from Anointed Chorister itself (not that it matterd much thanks to Kor Haven from me), but instead it just dumped all its mana for that one counter.

Win 10, 1.6.56-SNAPSHOT-02.23, Insane Difficulty Adventure Mode

Agetian:
It'll now prefer the cheaper ability if the target card already has +1/+1 counters.

Fixes #2563

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
@@ -471,6 +471,13 @@ public class CountersPutAi extends CountersAi {
                     if (sacSelf && c.equals(source)) {
                         return false;
                     }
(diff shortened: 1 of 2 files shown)


tool4ever (approved):

Hanmac:
imo the AI should recognize that both are the same effect, but one is cheaper cost, but i think that works for now

Agetian:
Yeah, I also thought about it, but haven't found a straightforward and clean enough way to do that :/ If I think of anything, I'll make a followup PR.

PR #3141, opened 2023-05-20
Adventure Mode: some minor text correction/clarification

Agetian:


jjayers99 (approved):
Never noticed these before,  even when I was creating copies of the gates. Thanks for the cleanup!

PR #3143, opened 2023-05-21
fix NPE on CommanderGauntlet

kevlahnota:


tool4ever:
Why can it be null?
E. g. does it mean user is missing data files where we could print an error message?

kevlahnota:
> Why can it be null? E. g. does it mean user is missing data files where we could print an error message?

Because of this:
https://github.com/Card-Forge/forge/blob/ea62081c00e569d2ca0d90a7fb0e26224234b157/forge-gui/src/main/java/forge/deck/DeckgenUtil.java#L439

tool4ever (approved):

PR #3144, opened 2023-05-21
AI creature spell priority tweaks

Agetian:
A followup tweak from the previous implementation: if a creature is preferred due to evaluation and no other prior priority considerations apply, actually make sure that the best creature evaluates to one higher than the worse one.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -1083,7 +1083,7 @@ public class AiController {
             }
 
             // If both are permanent creature spells, prefer the one that evaluates higher
-            if (a1 == b1 && a.getApi() == ApiType.PermanentCreature && b.getApi() == ApiType.PermanentCreature) {
+            if (a.getApi() == ApiType.PermanentCreature && b.getApi() == ApiType.PermanentCreature) {
                 int evalA = ComputerUtilCard.evaluateCreature(a);
                 int evalB = ComputerUtilCard.evaluateCreature(b);
                 if (evalA > evalB) {
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -1087,9 +1087,9 @@ public class AiController {
                 int evalA = ComputerUtilCard.evaluateCreature(a);
                 int evalB = ComputerUtilCard.evaluateCreature(b);
                 if (evalA > evalB) {
-                    a1 += Math.max(1, Math.ceil(evalA / 100.0f));
+                    a1 += Math.max(1, Math.round(evalA / 100.0f));
                 } else if (evalB > evalA) {
-                    b1 += Math.max(1, Math.ceil(evalB / 100.0f));
+                    b1 += Math.max(1, Math.round(evalB / 100.0f));
                 }
             }
 
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -1087,9 +1087,9 @@ public class AiController {
                 int evalA = ComputerUtilCard.evaluateCreature(a);
                 int evalB = ComputerUtilCard.evaluateCreature(b);
                 if (evalA > evalB) {
-                    a1++;
+                    a1 = b1 + 1;
                 } else if (evalB > evalA) {
-                    b1++;
+                    b1 = a1 + 1;
                 }
             }
 
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -1087,9 +1087,9 @@ public class AiController {
                 int evalA = ComputerUtilCard.evaluateCreature(a);
                 int evalB = ComputerUtilCard.evaluateCreature(b);
                 if (evalA > evalB) {
-                    a1 = b1 + 1;
+                    a1 += Math.max(1, Math.ceil(evalA / 100.0f));
                 } else if (evalB > evalA) {
-                    b1 = a1 + 1;
+                    b1 += Math.max(1, Math.ceil(evalB / 100.0f));
                 }
             }
 
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -655,10 +655,11 @@ public class AiController {
         List<SpellAbility> all = ComputerUtilAbility.getSpellAbilities(cards, player);
 
         try {
-            Collections.sort(all, saComparator); // put best spells first
+            Collections.sort(all, ComputerUtilAbility.saEvaluator); // put best spells first
+            ComputerUtilAbility.sortCreatureSpells(all);
         } catch (IllegalArgumentException ex) {
             System.err.println(ex.getMessage());
-            String assertex = ComparatorUtil.verifyTransitivity(saComparator, all);
+            String assertex = ComparatorUtil.verifyTransitivity(ComputerUtilAbility.saEvaluator, all);
             Sentry.captureMessage(ex.getMessage() + "\nAssertionError [verifyTransitivity]: " + assertex);
         }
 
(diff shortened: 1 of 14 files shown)


forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -1087,9 +1087,9 @@ public int compare(final SpellAbility a, final SpellAbility b) {
                 int evalA = ComputerUtilCard.evaluateCreature(a);
                 int evalB = ComputerUtilCard.evaluateCreature(b);
                 if (evalA > evalB) {
-                    a1++;
+                    a1 = b1 + 1;
```
tool4ever, line 1090: We need to be careful here, pretty sure this can open up more potential for `AssertionError`
Need to think on some tests for this

Agetian: We can also set them to fixed values instead of making them depend on each other. I was also thinking about just adding some number dependent on the result of the evaluation, but couldn't figure out the formula.

Agetian: Tried something to make things a little bit more interesting depending on how big of a difference the evaluation comes to, can still be improved but better than just assuming +1 no matter if the difference is minor or huge.

tool4ever: The logic for transitivity means this must always be true:
`A > B & B > C <=> A > C`

Let's see:

- A = creature spell with 5 cmc and some easy keywords that we check for to raise the value
- B = creature spell with 7 cmc which is very valueable but we can't recognize (e.g. _Elesh Norn, Grand Cenobite_)
- C = random instant that costs 6 cmc

Now you might get A > B but when comparing against the instant B > C but **C > A** ⚠️

(the evaluator also considers cmc slightly but it's still not a real safeguard)

I'm sure you can find some combination of cards that can fulfill this. Then you will have a crash 💥

Agetian: Hmm, do you have an idea for what might work then?
Currently, this entire branch will only fire when comparing two creature permanent spells, so it'll never trigger for situations when comparing a creature against an instant, for example.

Agetian: Alternatively, if it's an issue in the "big picture" when all the spells are sorted later (not immediately when two are compared against each other), we can avoid modifying a1/b1 in any way and just return "1" or "-1" depending on the evaluation results (though this will screw up AI-specific priority a little bit, which only matters for a couple creatures IIRC). Do you think that'll work better?

Agetian: Btw, on a side note, the more I deal with this, the more I wonder how this got screwed up in the first place and how exactly it worked before, haha :) Because I know for a fact that the AI used to prioritize creatures properly a while back, but I guess some code got changed eventually and that part got lost to time.

tool4ever: Comparators can be tricky imo, here it's all part of the same so it will compare all available SA against each other with the potential for a logic breakdown like I tried to outline above

I think we also had an issue for it on the old Gitlab and I still see it sometimes pop up on sentry but luckily it's rare that's why I'm concerned, e. g. https://sentry.cardforge.org/organizations/forge/issues/25926/?project=3&query=is%3Aunresolved+release%3A%2A1.6.57%2A&referrer=issue-stream&statsPeriod=14d&stream_index=9

I might have to sleep on this, maybe even a few nights and research some more to come up with a safe approach that still keeps the desired Evaluator input too :/

Agetian: Actually, I think just returning -1 and 1 from there isn't the best option since getSpellAbilityPriority has more work to do depending on various things :/

Agetian: Yeah, not sure how to do this better for now, if I think of something I'll try to implement it as well :)
So far I haven't seen it crash but I understand what you mean, there might indeed be cases where it may matter and go south

tool4ever: Returning 1 earlier doesn't change the underlying problem which is you're sometimes using a different scoring function for the inputs

Agetian: Hmm, so if we do, for instance:

            int a1 = a.getPayCosts().getTotalMana().getCMC() + <evaluateCreature on a>;
            int b1 = b.getPayCosts().getTotalMana().getCMC() + <evaluateCreature on b>;

And make it always check evaluateCreature for everything, returning 0 for non-creatures and some kind of a value for creatures, it won't crash, right?

This will, however, have an unwanted side effect of always prioritizing creatures higher than everything else though, since this value will only ever be meaningful (and non-zero) for creature spells :/

tool4ever: > Btw, on a side note, the more I deal with this, the more I wonder how this got screwed up in the first place and how exactly it worked before, haha :) Because I know for a fact that the AI used to prioritize creatures properly a while back, but I guess some code got changed eventually and that part got lost to time.

I think that might have been before I joined then, I mean we also still have `PermanentCreatureAi` which also prevents some misplays (though kind of differently)

Agetian: Also, getSpellAbilityPriority makes quite a lot of modifications to the scoring of each ability (a and b) before the values are compared, is that as dangerous to do as what we're trying here? Or if not, what makes it less dangerous? (since that function's been in for years and I haven't ever encountered a single crash coming from it)

Agetian: Sadly, PermanentCreatureAi probably won't be of much help here since we're dealing with relative priority at the moment the SA is chosen... At least for now I don't see a way to integrate it there...

tool4ever: > Hmm, so if we do, for instance:
> 
> ```
>         int a1 = a.getPayCosts().getTotalMana().getCMC() + <evaluateCreature on a>;
>         int b1 = b.getPayCosts().getTotalMana().getCMC() + <evaluateCreature on b>;
> ```
> 
> And make it always check evaluateCreature for everything, returning 0 for non-creatures and some kind of a value for creatures, it won't crash, right?
> 
> This will, however, have an unwanted side effect of always prioritizing creatures higher than everything else though, since this value will only ever be meaningful (and non-zero) for creature spells :/

Yes probably but like you said with some rather unwanted behaviour adjustment

Maybe there's some elegant workaround but currently I'm thinking a two-step process where you first sort the generic way and then check if a creature is first place and then there's another creature within a certain threshold you can compare those against each other
(which probably means storing the priorities in a Map or something)

tool4ever: > Sadly, PermanentCreatureAi probably won't be of much help here since we're dealing with relative priority at the moment the SA is chosen... At least for now I don't see a way to integrate it there...

I just meant you might think it got deprioritized when in reality AI just declined playing it and went for the next instead

Agetian: Yeah, a two-step process might be viable :+1:  Maybe looping over the result of the comparator and grabbing the creature that evaluates the highest if there's a creature on top of the loop would work, I'll see if I can try it.

tool4ever: > Also, getSpellAbilityPriority makes quite a lot of modifications to the scoring of each ability (a and b) before the values are compared, is that as dangerous to do as what we're trying here? Or if not, what makes it less dangerous? (since that function's been in for years and I haven't ever encountered a single crash coming from it)

It's somewhat safer because it treats all inputs the same

Agetian: Hmm, alright. So far, this is probably beyond my capabilities :) I checked how the comparator is used - it's used before the SAs are actually tested for playability, so there's no way to even see if the AI can pay for whatever is on top of the list. Therefore, the second iteration over the comparator should somehow preserve everything while changing the relative priority of only a few things (e.g. creatures that evaluate to a higher value).

Agetian: Btw, is this alright then? (it's been done for years too and I don't think it treats all inputs as the same, as not only does it look specifically for planeswalker abilities but it also sets the priority differently depending on their converted cost amount)

```
            // sort planeswalker abilities with most costly first
            if (sa.isPwAbility()) {
                final CostPart cost = sa.getPayCosts().getCostParts().get(0);
                if (cost instanceof CostRemoveCounter) {
                    p += cost.convertAmount() == null ? 1 : cost.convertAmount();
                } else if (cost instanceof CostPutCounter) {
                    p -= cost.convertAmount();
                }
                if (sa.hasParam("Ultimate")) {
                    p += 9;
                }
            }

```

tool4ever: I think so since it's not based on a dependency between A and B

I might have one solution, will commit shortly

Agetian: Oh, nice! Feel free to have at it, since I'm almost out of ideas for now :)

The only thing I could come up with so far, if it helps avoid dependency between A and B, is something along the lines of:
```
            // If both are permanent creature spells, prefer the one that evaluates higher
            if (a.getApi() == ApiType.PermanentCreature && b.getApi() == ApiType.PermanentCreature) {
                a1 += Math.max(1, Math.round(ComputerUtilCard.evaluateCreature(a) / 100.0f));
                b1 += Math.max(1, Math.round(ComputerUtilCard.evaluateCreature(b) / 100.0f));
            }
```

tool4ever: I'm trying (ignoring special SVar and the diff margin etc.) if  to get any input list
(where the CMC numbers with "c" are creatures)

|from|1. transformed to|2. and then maybe to something like
|-|-|-
|6|7|7
|4|6|6
|3c_1|5c|3c_2
|7|3c_1|3c_1
|5c|4|4
|2|3c_2|5c
|3c_2|2|2

so in 2. it keeps the general priority order but can switch around some or even all of the creatures

But it's tricky if some non creatures are between there

Agetian: A very interesting approach, thanks a lot for help! I'll test a few things a bit later to see how it works :) :+1:

Agetian: In all my test cases the AI works well with the changes you implemented, I'd say this is a lot better than having no priority for creatures whatsoever, so I'd vote for integrating this. I'll test a bit more during the day to see if anything is off, otherwise it does the job for sure :) (unless you have followup ideas for other situations you outlined, but even that can be reserved for a followup PR)

tool4ever: So to summarize:
1. with roughly the same CMC creature spells can still get reordered if their permanent value is considered to be "too unequal" (the factor 100f tries to align the scales and might still be lowered if it's not giving us the desired result in the majority of cases)

2. ideally we could also recognize another common variant of power creep in which creatures A and B are very close in card value, but A got printed first much earlier than B - so A's CMC is actually 1-2 higher and therefore worse overall (though there are exceptions to this like the P9 but luckily they aren't creatures...)

tool4ever: I think currently getting 2. right more often than wrong might not be possible as there are too many cases the Evaluator can't recognize:
E.g. _Toxrill_ will also look vanilla and would then be seen as less cost efficient to something like _Fusion Elemental_

Agetian: Yeah, (2) is tricky, and tbh I'm not sure what the ideal solution would be there, it's quite difficult to formalize and generalize the difference :/

Hanmac:
maybe related to this, but AI needs to get some logic when to NOT pay for extra optional costs like kicker

currently AI thinks "More is Better" so it tries to pay with all optional costs if possible

Agetian:
As for this particular PR in its current state, I've been playing many games today with different cards and decks and stuff, and I believe it works well :)

kevlahnota (approved):

forge-ai/src/main/java/forge/ai/AiController.java
kevlahnota: Gonna merge this. A followup PR can be done if available.

PR #3146, opened 2023-05-23
LKICopy Transform and flip

### LINKED ISSUE #3201: NPE when trying to cast Split by Melek
Sentry Issue: [DESKTOP-PWR](https://sentry.cardforge.org/organizations/forge/issues/28145/?referrer=github_integration)

```
NullPointerException: None
    at forge.game.card.CardView.updateState(CardView.java:997)
    at forge.game.card.Card.setState(Card.java:517)
    at forge.game.card.CardFactory.copySpellHost(CardFactory.java:149)
    at forge.game.card.CardFactory.copySpellAbilityAndPossiblyHost(CardFactory.java:182)
    at forge.game.ability.effects.CopySpellAbilityEffect.resolve(CopySpellAbilityEffect.java:155)
...
(16 additional frame(s) were not displayed)
```

Hanmac:
Closes #3201

need to fix the Tests, don't have time right now

The Main Reason for this is to have cards like Ratadravik to be able to copy Transforming Double faced cards (and their transformed state)

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityFactory.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityFactory.java
@@ -511,6 +511,7 @@ public final class AbilityFactory {
         totalCost.add(parseAbilityCost(rightState, rightMap, rightType));
 
         final SpellAbility left = getAbility(leftType, leftApi, leftMap, totalCost, leftState, leftState);
+        left.setCardState(card.getState(CardStateName.Original));
         final AbilitySub right = (AbilitySub) getAbility(AbilityRecordType.SubAbility, rightApi, rightMap, null, rightState, rightState);
         left.appendSubAbility(right);
         return left;
--- a/forge-game/src/main/java/forge/game/card/CardFactory.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactory.java
@@ -753,6 +753,19 @@ public class CardFactory {
             final CardState ret2 = new CardState(out, CardStateName.Transformed);
             ret2.copyFrom(in.getState(CardStateName.Transformed), false, sa);
             result.put(CardStateName.Transformed, ret2);
+        } else if (in.isSplitCard()) {
+            // for split cards, copy all three states
+            final CardState ret1 = new CardState(out, CardStateName.Original);
+            ret1.copyFrom(in.getState(CardStateName.Original), false, sa);
+            result.put(CardStateName.Original, ret1);
+
+            final CardState ret2 = new CardState(out, CardStateName.LeftSplit);
+            ret2.copyFrom(in.getState(CardStateName.LeftSplit), false, sa);
+            result.put(CardStateName.LeftSplit, ret2);
+
+            final CardState ret3 = new CardState(out, CardStateName.RightSplit);
+            ret3.copyFrom(in.getState(CardStateName.RightSplit), false, sa);
+            result.put(CardStateName.RightSplit, ret3);
         } else {
             // in all other cases just copy the current state to original
             final CardState ret = new CardState(out, CardStateName.Original);
--- a/forge-game/src/main/java/forge/game/card/CardUtil.java
+++ b/forge-game/src/main/java/forge/game/card/CardUtil.java
@@ -216,6 +216,12 @@ public final class CardUtil {
             newCopy.getState(CardStateName.Original).copyFrom(in.getState(CardStateName.Original), true);
             newCopy.addAlternateState(CardStateName.Adventure, false);
             newCopy.getState(CardStateName.Adventure).copyFrom(in.getState(CardStateName.Adventure), true);
+        } else if (in.isSplitCard()) {
+            newCopy.getState(CardStateName.Original).copyFrom(in.getState(CardStateName.Original), true);
+            newCopy.addAlternateState(CardStateName.LeftSplit, false);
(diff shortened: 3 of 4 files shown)


Hanmac:
errors down to one failed test

Clone by Cytoshape shouldn't make it transformable 🤔

kevlahnota:
here's a patch I used seems to work but didn't test all others, it just transforms the copied card...
```
Index: forge-game/src/main/java/forge/game/card/CardUtil.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/forge-game/src/main/java/forge/game/card/CardUtil.java b/forge-game/src/main/java/forge/game/card/CardUtil.java
--- a/forge-game/src/main/java/forge/game/card/CardUtil.java	(revision 29317d6b83341f73b25b664bd82a33570e80444a)
+++ b/forge-game/src/main/java/forge/game/card/CardUtil.java	(date 1684826522874)
@@ -206,11 +206,9 @@
             newCopy.setType(new CardType(in.getFaceDownState().getType()));
             // prevent StackDescription from revealing face
             newCopy.updateStateForView();
-        }
-
-        if (in.isAdventureCard() && in.getFaceupCardStateName().equals(CardStateName.Original)) {
-            newCopy.addAlternateState(CardStateName.Adventure, false);
-            newCopy.getState(CardStateName.Adventure).copyFrom(in.getState(CardStateName.Adventure), true);
+        } else if (in.hasAlternateState() && in.getFaceupCardStateName().equals(CardStateName.Original)) {
+            newCopy.addAlternateState(in.getAlternateStateName(), false);
+            newCopy.getState(in.getAlternateStateName()).copyFrom(in.getAlternateState(), true);
         }
 
         /*
Index: forge-game/src/main/java/forge/game/card/CardFactory.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/forge-game/src/main/java/forge/game/card/CardFactory.java b/forge-game/src/main/java/forge/game/card/CardFactory.java
--- a/forge-game/src/main/java/forge/game/card/CardFactory.java	(revision 29317d6b83341f73b25b664bd82a33570e80444a)
+++ b/forge-game/src/main/java/forge/game/card/CardFactory.java	(date 1684828791699)
@@ -741,7 +741,7 @@
             final CardState ret2 = new CardState(out, CardStateName.Adventure);
             ret2.copyFrom(in.getState(CardStateName.Adventure), false, sa);
             result.put(CardStateName.Adventure, ret2);
-        } else if (in.isTransformable() && sa instanceof SpellAbility && (
+        } else if (in.hasState(CardStateName.Transformed) && sa instanceof SpellAbility && (
                 ApiType.CopyPermanent.equals(((SpellAbility)sa).getApi()) ||
                 ApiType.CopySpellAbility.equals(((SpellAbility)sa).getApi())
                 )) {

```

Hanmac:
@kevlahnota 
the first part of your patch i need to check if that's good or not
the lki part needs to copy also the state in which the original is in (transformed or not)

as for the second part, it shouldn't matter for Cytoshape because it should be neither CopyPermanent nor CopySpellAbility

tool4ever:
When `in.getState(CardStateName.Transformed)` is called the original card gets a new state here:
https://github.com/Card-Forge/forge/blob/29317d6b83341f73b25b664bd82a33570e80444a/forge-game/src/main/java/forge/game/card/CardCloneStates.java#L49

Hanmac:
> When `in.getState(CardStateName.Transformed)` is called the original card gets a new state here:
> 
> https://github.com/Card-Forge/forge/blob/29317d6b83341f73b25b664bd82a33570e80444a/forge-game/src/main/java/forge/game/card/CardCloneStates.java#L49

I need to debug at which point it causes a problem
Is it:
* A) when it makes the LKI?
* B) when it makes the Clone States?

Hanmac:
@kevlahnota can you debug why Fused Spell like "Alive // Well" has the wrong color/state when put on the stack? even when the effect is correct?

Hanmac:
I tested Ratadravik + Transforming Legendary Permanent (Phyrexian) (i tested with new Heliod)
* if the Legendary dies before it transforms, than the token will be front side and transformable (and it will transform into backside)
* if the legendary dies while transformed than the token will be backside too.

@Agetian i think we just need to turn that into TestCases

Hanmac:
I added a Test Case for this to make a Transformed Token from LKI

@tool4ever @Agetian is otherwise this PR ready?

tool4ever (approved):
I tested the case of 712.13a:
> A player controls both Mycosynth Lattice and March of the Machines, the combined effects of which make all permanents artifact creatures in addition to their other types. They also control a Clone on the battlefield that is a copy of Bird Admirer, a creature with daybound. It is currently night, but that permanent can't transform because it isn't represented by a double-faced card. Its controller casts Mystic Reflection targeting it, then casts Invasion of Kylem, a Siege battle whose back face is a sorcery. Because Invasion of Kylem is entering the battlefield as a creature, Mystic Reflection's replacement effect applies to it and it tries to enter the battlefield as a copy of Bird Admirer. Since it is night, the daybound ability would normally cause it to enter with its back face up. However, since its back face is a sorcery, it is instead put into its owner's graveyard.

I still got a new Daybound copy. (So I don't think it's a problem with missing layers dependency)

But that can be checked later

PR #3149, opened 2023-05-24
Add "Base" and "Relief" terrain sets

schnautzr:
The terrain set named "Base" contains all base terrain tiles from Main. It can be used to speed up the creation of maps.


schnautzr:
It doesn't include layer sprites yet, so forests still require manual work.

schnautzr:
Relief terrain set added, which includes forests, mountains, fungus, and coral.

jjayers99 (approved):

PR #3151, opened 2023-05-25
prevent NPE

kevlahnota:


forge-game/src/main/java/forge/game/trigger/Trigger.java
```diff
@@ -205,7 +205,8 @@ public final String replaceAbilityText(final String desc, SpellAbility sa, boole
                 }
             }
             if (saDesc.equals("")) { // in case we haven't found anything better
-                saDesc = sa.toString();
+                if (sa != null)
```
Hanmac, line 208: This can't happen, look at line 172 it is already inside a ` if (sa != null)` ?

kevlahnota: Can you explain why this happened?
Screenshot_2023-05-25-17-35-17-02_40deb401b9ffe8e1df2f1cc5ba480b12

Hanmac: Because digMore was messing with sa variable

```                        
            if (digMore) { // if ABILITY is used, there is probably Charm somewhere
                while (sa != null) {
                    ApiType api = sa.getApi();
                    if (ApiType.Charm.equals(api)) {
                        saDesc = CharmEffect.makeFormatedDescription(sa, !forStack);
                        break;
                    }
                    if (ApiType.ImmediateTrigger.equals(api) || ApiType.DelayedTrigger.equals(api)) {
                        SpellAbility trigSA = sa.getAdditionalAbility("Execute");
                        while (trigSA != null) {
                            if (ApiType.Charm.equals(trigSA.getApi())) {
                                saDesc = CharmEffect.makeFormatedDescription(trigSA, !forStack);
                                break;
                            }
                            trigSA = trigSA.getSubAbility();
                        }
                        break;
                    }
                    sa = sa.getSubAbility();
                }
            }
```

PR #3154, opened 2023-05-25
Sprite and deck reorganisation + "wiz" enemies rework

Simisays:
1. Five new stores: Insects,Shapeshifters, Gods and demigods, Eldrazi,small cats
2. Reorganisation of the decks and sprites into groups to make it easier to find the sprites or decks you might need
3. Rework of the "wiz" enemies, new decks and loottables
4. new enemies: 4 new kobolds, Simic Wizard, Runelord


kevlahnota:
@Simisays Did you try this with old saves? does the changes in enemy names allows loading of save files? and does not interfere anything on older saves?

jjayers99:
So long as all instances of the name have been replaced, I suspect that the
only disruption would be in the win / loss record on the player statistics
page, as those are stored by name.

Simisays:
Table Top General is right, only difference is that your win/loss records of the "wiz" enemies are lost. I haven't found other problems problems with older saves

kevlahnota (approved):

PR #3158, opened 2023-05-26
Fix devout invocation never creating any tokens

Ampersandnz:
Needs review - I'm not 100% sure this is correct.
I copied from [Siege Striker](https://github.com/Card-Forge/forge/blob/c46cab158eb94524b75e71c786d317481aadfc81/forge-gui/res/cardsfolder/s/siege_striker.txt#L4), which is working

### DIFF
--- a/forge-gui/res/cardsfolder/d/devout_invocation.txt
+++ b/forge-gui/res/cardsfolder/d/devout_invocation.txt
@@ -4,7 +4,7 @@ Types:Sorcery
 A:SP$ Tap | CardChoices$ Creature.YouCtrl+untapped | AnyNumber$ True | ChoiceAmount$ Count$Valid Creature.YouCtrl+untapped | RememberTapped$ True | SubAbility$ DBToken | SpellDescription$ Tap any number of untapped creatures you control. Create a 4/4 white Angel creature token with flying for each creature tapped this way. | StackDescription$ SpellDescription
 SVar:DBToken:DB$ Token | TokenAmount$ Y | TokenScript$ w_4_4_angel_flying | TokenOwner$ You | SubAbility$ DBCleanup
 SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
-SVar:Y:Remembered$Amount
+SVar:Y:Count$RememberedSize
 SVar:NeedsToPlay:Creature.YouCtrl+untapped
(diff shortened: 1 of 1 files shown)


forge-gui/res/cardsfolder/d/devout_invocation.txt
```diff
@@ -4,7 +4,7 @@ Types:Sorcery
 A:SP$ Tap | CardChoices$ Creature.YouCtrl+untapped | AnyNumber$ True | ChoiceAmount$ Count$Valid Creature.YouCtrl+untapped | RememberTapped$ True | SubAbility$ DBToken | SpellDescription$ Tap any number of untapped creatures you control. Create a 4/4 white Angel creature token with flying for each creature tapped this way. | StackDescription$ SpellDescription
 SVar:DBToken:DB$ Token | TokenAmount$ Y | TokenScript$ w_4_4_angel_flying | TokenOwner$ You | SubAbility$ DBCleanup
 SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
-SVar:Y:Remembered$Amount
+SVar:Y:Count$RememberedSize
```
tool4ever, line 7: This should not lead to any other results

The card works fine for me, did you test this in a clean setting (e.g. only basics in all zones)?

Ampersandnz: I tested 1v1 vs ai with a deck that was all land + fast mana + 0 mana creatures
See in the log: striker attacks for 4 on previous turn, this turn full tap but no tokens
image

```
Deck: test

Main: 
4 Black Lotus
4 Crimson Kobolds
4 Crookshank Kobolds
4 Devout Invocation
4 Kobolds of Kher Keep
4 Mana Crypt
4 Mox Pearl
4 Ornithopter
20 Plains
4 Siege Striker
4 Sol Ring
```

Ampersandnz: I'm on version 1.6.56

tool4ever: In that case you need to compare your local version to the repository

You'll notice it already got a fix after that release was done

Ampersandnz: Gotcha, thanks!

PR #3161, opened 2023-05-27
Fix Containment Priest exiling itself

tool4ever:
@Hanmac
I checked with past versions of ReplacementHandler and it seems like we never had code to handle this.

So I just took the logic you used for _Mowu_ and made it a helper.


Hanmac (approved):
:+

PR #3208, opened 2023-06-05
Improve AI playability for Nissa, Genesis Mage

Agetian:


forge-ai/src/main/java/forge/ai/ability/UntapAi.java
```diff
@@ -190,6 +191,12 @@ private static boolean untapPrefTargeting(final Player ai, final SpellAbility sa
                         && ai.getGame().getPhaseHandler().getPhase().isBefore(PhaseType.COMBAT_DECLARE_ATTACKERS)) {
                     choice = ComputerUtilCard.getWorstPermanentAI(list, false, false, false, false);
                 } else if (!sa.isMinTargetChosen() || sa.isZeroTargets()) {
+                    // for Planeswalker +Loyalty abilities, activate them even if no good targets are present
+                    if (sa.getRootAbility().isPwAbility()
```
tool4ever, line 195: Just call `ComputerUtil.activateForCost`? :)

Agetian: Nice suggestion! Interestingly, for some reason it doesn't work the same way - in particular, the AI will try to activate it in Main 2 but will fail to target anything (e.g. the stack will merely say "Untap Untap", and nothing will get untapped). I'll try to debug that a bit more and see why it's behaving that way in M2 :)

Agetian: Ok, made some tweaks to make it work in both Main1 and Main2. It looks like activateForCost only checked the current SA for planeswalker costs, while it should really check the root ability, not the subs. Also, it looks like resetting the targets late in UntapAi doesn't always work, making the targets linger over from the previous iteration, and that makes the whole thing fail to work in subsequent checks (the AI mistakenly assumes that it had already targeted something before).

PR #3226, opened 2023-06-09
filter out invalid option for BoosterDraft

kevlahnota:


forge-gui/src/main/java/forge/gamemodes/limited/LimitedPoolType.java
```diff
@@ -5,16 +5,45 @@
...
         displayName = name;
     }
 
+    public static LimitedPoolType[] values(boolean draftable) {
+        if (!draftable) {
+            return values();
+        }
+
+        int n = 0;
+        for (LimitedPoolType lpt : values()) {
```
tool4ever, line 31: this loop doesn't do anything?

kevlahnota: @tehdiplomat can you check, i change this PR to reflect the one you suggested (removed the streams and filter).

tehdiplomat: No he's right this loop is later down in line 38. This one is duplicate

tehdiplomat: This loop is counting the size for the be est array

kevlahnota: Ok so It counts all draftable then set the size of array and fill the array with valid enum values.

tool4ever (approved):

PR #3228, opened 2023-06-09
Add several popular cubes from Cube Cobra

Agetian:


Agetian:
Need to check if all the cards are supported by Forge.

kevlahnota (approved):

PR #3232, opened 2023-06-10
AI: Account for transformable cards (e.g. Incubator tokens) when considering defenders

### LINKED ISSUE #3061: AI will always attack into non-transformed incubate tokens with open mana
**Describe the bug**
AI treats an incubate token with 2 open mana as if it's not there

**To Reproduce**
1. Give AI a 2/2. Create a 5/5 incubate token and two open mana.
2. AI will ignore the incubate token and attack into the open board.

**Expected behavior**
With 2 open mana AI ought to treat an incubate token as if it were transformed.

 - OS: Windows
 - Version 1.6.57-SNAPSHOT-05.01

Agetian:
Fixes #3061 .

The code at AiAttackController.java:140-141 originally skipped all checks for tokens in this manner (probably for speed?) I currently made an exclusion to consider transformable tokens such as Incubator stuff, is that a viable solution? :/

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -153,7 +153,9 @@ public class AiAttackController {
                         && sa.getRestrictions().checkOtherRestrictions(c, sa, defender)) {
                     Card transformedCopy = CardUtil.getLKICopy(c);
                     transformedCopy.setState(CardStateName.Transformed, false);
-                    defenders.add(transformedCopy);
+                    if (transformedCopy.isCreature()) {
+                        defenders.add(transformedCopy);
+                    }
                 }
             }
         }
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -17,28 +17,17 @@
  */
 package forge.ai;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.List;
-
-import forge.card.CardStateName;
-import forge.game.card.*;
-import forge.game.staticability.StaticAbility;
-import forge.game.staticability.StaticAbilityAssignCombatDamageAsUnblocked;
-import org.apache.commons.lang3.tuple.Pair;
-
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
-
 import forge.ai.ability.AnimateAi;
 import forge.card.CardTypeView;
 import forge.game.GameEntity;
 import forge.game.ability.AbilityUtils;
 import forge.game.ability.ApiType;
 import forge.game.ability.effects.ProtectEffect;
+import forge.game.card.*;
 import forge.game.combat.Combat;
 import forge.game.combat.CombatUtil;
 import forge.game.combat.GlobalAttackRestrictions;
@@ -48,6 +37,8 @@ import forge.game.player.Player;
 import forge.game.player.PlayerCollection;
 import forge.game.spellability.SpellAbility;
 import forge.game.spellability.SpellAbilityPredicates;
+import forge.game.staticability.StaticAbility;
+import forge.game.staticability.StaticAbilityAssignCombatDamageAsUnblocked;
 import forge.game.trigger.Trigger;
 import forge.game.trigger.TriggerType;
 import forge.game.zone.Zone;
@@ -57,6 +48,12 @@ import forge.util.Expressions;
 import forge.util.MyRandom;
 import forge.util.collect.FCollection;
 import forge.util.collect.FCollectionView;
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
 
 
 /**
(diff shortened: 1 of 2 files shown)


Agetian:
Will need to add a check if the transformed card is actually a creature , will do that after work.

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -152,6 +147,15 @@ public boolean apply(Card c) {
                     defenders.add(animatedCopy);
                 }
             }
+            // Transform (e.g. Incubator tokens)
+            for (SpellAbility sa : Iterables.filter(c.getSpellAbilities(), SpellAbilityPredicates.isApi(ApiType.SetState))) {
+                if ("Transform".equals(sa.getParam("Mode")) && ComputerUtilCost.canPayCost(sa, defender, false)
+                        && sa.getRestrictions().checkOtherRestrictions(c, sa, defender)) {
+                    Card transformedCopy = CardUtil.getLKICopy(c);
+                    transformedCopy.setState(CardStateName.Transformed, false);
+                    defenders.add(transformedCopy);
```
Hanmac, line 156: might check if the transformed Copy is a creature?

Agetian: Yup, going to add that after work 👍

Agetian: Done :)

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -152,6 +147,17 @@ public boolean apply(Card c) {
                     defenders.add(animatedCopy);
                 }
             }
+            // Transform (e.g. Incubator tokens)
+            for (SpellAbility sa : Iterables.filter(c.getSpellAbilities(), SpellAbilityPredicates.isApi(ApiType.SetState))) {
```
tool4ever, line 151: Looks like this code is pretty much `ComputerUtilCombat..canTransform?`

Agetian: More or less, yes, but sadly, without iterating through spell abilities and grabbing the transform sa (with SetStateAPI) somehow, it'd be impossible to check if the AI can pay the cost for it :/ Do you have any ideas for how to implement that part seamlessly while simplifying the loop (e.g. by changing it into a canTransform check)?

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -142,7 +137,7 @@ public boolean apply(Card c) {
             }
         };
         for (Card c : CardLists.filter(defender.getCardsIn(ZoneType.Battlefield), canAnimate)) {
-            if (c.isToken() && c.getCopiedPermanent() == null) {
+            if (c.isToken() && c.getCopiedPermanent() == null && !c.canTransform(null)) {
```
tool4ever, line 140: we might check if this logic is a bit outdated?
e.g. _Lita, Mechanical Engineer_ can create Vehicle token

forge-ai/src/main/java/forge/ai/AiAttackController.java
Agetian: I mean, we can check canTransform, and if it returns true, then loop over the SAs and find the SetState one and check the cost, but I'm not sure if that's much better than what it's currently doing :)

forge-ai/src/main/java/forge/ai/AiAttackController.java
Agetian: Yeah, I think maybe it's possible to remove it altogether?.. It was probably done for speed reasons originally though (checking it on hundreds of tokens in a token deck might be costly, I guess :) )

forge-ai/src/main/java/forge/ai/AiAttackController.java
tool4ever: It looks like it will check `canPayCost` too...?
Otherwise just returns the untransformed

tool4ever: The only problem might be we don't add up the costs so the AI might expect a lot of defenders that are actually not possible

Agetian: Yes, true, it's probably also true for other situations similar to this one where the AI weighs every case individually. Not sure how to make it account for all of them together off the top of my head tbh.

Agetian: Hmm, maybe some trickery with remembering the mana sources necessary to pay every processed cost would be possible, but I'm not sure

forge-ai/src/main/java/forge/ai/AiAttackController.java
Agetian: Commented out this code for now, will test with it disabled from now on.

forge-ai/src/main/java/forge/ai/AiAttackController.java
Agetian: Checked the code for remembering and reserving mana sources, that's only implemented on the AI side (in the AI controller) and hooked up to AI-specific routines... Don't see a way to simulate that for the human opponent in order to determine how much can be paid :( There's probably a roundabout way where we can remember how much mana is already potentially spent (during the loop iteration) and then we bail out of adding defenders like that if we're above the total limit, it's going to be imprecise though.

Agetian: Attempted what I mentioned above, it's imprecise but it sort of works for simple cases like Incubator where only colorless mana is used, I guess. I welcome any better implementation if you can think of one... This is prolly the best I can do for now. I wanted to also hook this up to Animate above, but I'm afraid that'll be a little bit too imprecise since many animate costs depend on particular color shards, and that'll make the thing fail more often than work vs. multicolor decks :/

Agetian: I guess it'll also need to check if the cost actually has a mana part to avoid a NPE?

Agetian: Maybe it can actually also be done for Animate... It'll miscount sometimes, I guess, but won't be as passive when not due? Prolly needs testing though.

Agetian: After some tests, added this to Animate as well and added some NPE prevention. Probably the best I can do for now, I don't have any better ideas, sadly :)

forge-ai/src/main/java/forge/ai/AiAttackController.java
Agetian: So far, testing doesn't indicate any issues with this code commented out.

tool4ever (approved):

PR #3242, opened 2023-06-11
AI: simple Reveal logic, example implementation for Delver of Secrets

Agetian:
Makes the AI skip the reveal if it sees a non-Instant/non-Sorcery card on Delver of Secrets, for which it's optional.

Can most likely be used for other similar effects and possibly other stuff too.
Currently remembers only the last reveal (will clean up the revealed cards the next time something is revealed). If this is to be extended to allow some complex logic depending on what was revealed during the turn, this will need to be changed (the cleanup in the reveal method will need to be removed).

The overload of the "reveal" method with the CardView parameter is tricky because it doesn't allow to get to the card directly, so I'm currently using a "bruteforce" iteration over all the cards in the game to find the source. Can this be done faster/easier somehow? Luckily, this method isn't used as often, but still.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/PeekAndRevealAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/PeekAndRevealAi.java
@@ -1,10 +1,6 @@
 package forge.ai.ability;
 
-import java.util.Map;
-import java.util.Set;
-
 import forge.ai.AiAttackController;
-import forge.ai.AiCardMemory;
 import forge.ai.SpellAbilityAi;
 import forge.ai.SpellApiToAi;
 import forge.game.card.Card;
@@ -17,6 +13,8 @@ import forge.game.spellability.AbilityStatic;
 import forge.game.spellability.AbilitySub;
 import forge.game.spellability.SpellAbility;
 import forge.game.zone.ZoneType;
+
+import java.util.Map;
 
 /** 
  * TODO: Write javadoc for this type.
--- a/forge-ai/src/main/java/forge/ai/ability/PeekAndRevealAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/PeekAndRevealAi.java
@@ -8,6 +8,7 @@ import forge.ai.AiCardMemory;
 import forge.ai.SpellAbilityAi;
 import forge.ai.SpellApiToAi;
 import forge.game.card.Card;
+import forge.game.card.CardCollection;
 import forge.game.phase.PhaseHandler;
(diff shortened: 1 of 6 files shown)


forge-ai/src/main/java/forge/ai/ability/PeekAndRevealAi.java
```diff
@@ -73,6 +75,15 @@ protected boolean canPlayAI(Player aiPlayer, SpellAbility sa) {
      */
     @Override
     public boolean confirmAction(Player player, SpellAbility sa, PlayerActionConfirmMode mode, String message, Map<String, Object> params) {
+        if ("InstantOrSorcery".equals(sa.getParam("AILogic"))) {
+            Set<Card> revealed = AiCardMemory.getMemorySet(player, AiCardMemory.MemorySet.REVEALED_CARDS);
```
tool4ever, line 79: Hmn, it would certainly be nice for the AI to change it's playstyle if it gains information about opponents hidden zones

But for this case it seems a bit roundabout to me as just modifying `PeekAndRevealEffect` to pass those cards via the params like in other cases would be easier? 🤷‍♂️

Agetian: Yeah, it just seemed like a good opportunity to implement the stubbed out reveal for possible future cases :) I might switch the Delver implementation into a params based one but keep the reveal stuff for other uses later. What do you think?

tool4ever: Sure, maybe I'll see if we can get some quick wins with it in places like `hasAFogEffect` next

There is also `Game.findByView`

Agetian: Hmm, I tried a few things but I can't quite figure out how we can utilize a SA param for this... I mean, the SA param map maps from a String to a String, while the contents of the things to peek/reveal will either be a single Card object or even a collection of Card objects, if multiple are revealed at the same time. How can this be stored in the PeekAndRevealEffect in order to be grabbed later from confirmAction in the AI code?.. I'll probably need a few pointers here, sadly...

Agetian: Ah, that makes sense! :) Didn't see that one somehow, haha. Thanks!! :)

Agetian: Done, I think :)

tool4ever (approved):
seems good

PR #3251, opened 2023-06-12
Consider fog effects ruining assaults

tool4ever:
While there are probably some other places that could benefit from this function for performance I've tried catching the most devastating counterattack scenario for starters...

AI now correctly decides to skip attacking since it would lose without a blocker (that is if you give it another strong enough wall it will still commit to bait it out):
image

It will obviously encounter memory loss on next turn but that's a problem we can think on some more.


Agetian (approved):

Agetian:
👍

PR #3255, opened 2023-06-12
AI: Support for RevealOrChoose (Dragon's Fire)

### LINKED ISSUE #2128: AI's Dragon's Fire cast keeps dealing 0 damage
I'm in a HBG draft, and the AI cast Dragon's Fire which resolved as dealing 0 damage. I tried casting it myself in a few ways and couldn't replicate the effect, but if I give the AI the card it keeps casting it that way.

[state.txt](https://github.com/Card-Forge/forge/files/10251931/state.txt)

Agetian:
Fixes #2128

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiCostDecision.java
+++ b/forge-ai/src/main/java/forge/ai/AiCostDecision.java
@@ -2,10 +2,7 @@ package forge.ai;
 
 import static forge.ai.ComputerUtilCard.getBestCreatureAI;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import forge.card.MagicColor;
 import forge.game.cost.*;
@@ -479,6 +476,14 @@ public class AiCostDecision extends CostDecisionMakerBase {
 
         if (cost.getType().equals("Hand")) {
             return PaymentDecision.card(hand);
+        }
+
+        if (cost.getRevealFrom().containsAll(Arrays.asList(ZoneType.Hand, ZoneType.Battlefield))) { // RevealOrChoose
+            CardCollection battlefieldOrHand = CardLists.getValidCards(player.getCardsIn(ZoneType.Battlefield),
+                    type.split(";"), player, source, ability);
+            hand = CardLists.getValidCards(hand, type.split(";"), player, source, ability);
+            battlefieldOrHand.addAll(hand);
+            return PaymentDecision.card(getBestCreatureAI(battlefieldOrHand));
         }
 
         if (cost.getType().equals("SameColor")) {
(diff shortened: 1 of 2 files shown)


Hanmac:
@Agetian For `Dragon's Fire` also AI should not reveal if Dragon Power is < 3

Agetian:
Done :)

forge-ai/src/main/java/forge/ai/AiCostDecision.java
```diff
@@ -481,6 +470,20 @@ public PaymentDecision visit(CostReveal cost) {
             return PaymentDecision.card(hand);
         }
 
+        if (cost.getRevealFrom().size() == 2 && cost.getRevealFrom().containsAll(Arrays.asList(ZoneType.Hand, ZoneType.Battlefield))) { // RevealOrChoose
+            String aiLogic = ability.getParamOrDefault("AILogic", "");
+            CardCollection battlefieldOrHand = CardLists.getValidCards(player.getCardsIn(ZoneType.Battlefield),
+                    type.split(";"), player, source, ability);
+            battlefieldOrHand.addAll(CardLists.getValidCards(hand, type.split(";"), player, source, ability));
```
tool4ever, line 477: getRevealFrom() is `List<ZoneType>` so both should already be there?

forge-ai/src/main/java/forge/ai/AiCostDecision.java
```diff
@@ -481,6 +470,20 @@ public PaymentDecision visit(CostReveal cost) {
...
+
+            if (aiLogic.startsWith("PowerAtLeast.")) {
```
tool4ever, line 479: does this really work? I thought the decision to pay kicker was done earlier

Agetian:
So,what hapoens now is if the AI has a weak dragon, it'll agree to pay Kicker, but then decide not to pay the entire cost, thus missing an opportunity to pay the spell with no kicker. We need to find where the kicker decision is made and rig it to obey the said logic.

Hanmac:
> So,what happens now is if the AI has a weak dragon, it'll agree to pay Kicker, but then decide not to pay the entire cost, thus missing an opportunity to pay the spell with no kicker. We need to find where the kicker decision is made and rig it to obey the said logic. May need some help on this one.

That's a general problem
We need to teach the AI that optional cost aren't always better

If possible, depending on the Effects

Agetian:
Yeah, that's a much harder goal, I tried considering it a few times but couldn't figure out how to best do it :(
I think this is the best I can do for this particular PR, the rest can be delegated to followup work... This works better than it did before, at least, and doesn't zero out the damage too :)

tool4ever (approved):
Agreed :)

PR #3263, opened 2023-06-14
Support common adventure resources, relocate resources to common directory, sorting, and groundwork for Amonkhet

schnautzr:
Add support for common resources for adventure mode. Particle effects, skin, sprites, ui, and most of the world files are now common resources. A file in "Shandalar" (or whatever plane you are using) will take precedence over a file with the same name in "common".

Further work is required to allow common audio, custom cards/pics, decks, maps, and music.


kevlahnota:
looks fine to me, only some testing if the moved files are working as intended since they're quite a lot.. maybe a unit test perhaps to read the json and get all the resources files and check if they all exists hence passes the test.

Simisays:
Also looks good to me. Though I wonder why it isn't necessary to change enemies.json to make the spritelocations match

schnautzr:
Sprites and decks have a new directory structure. Added Amonkhet files.

PR #3276, opened 2023-06-15
Aura without Enchant shouldnt't stay attached

tool4ever:
> If you target an Aura with Immovable Rod's last ability, it will lose its enchant ability and be put into its owner's graveyard because it can't legally enchant anything.

Also partial tweak for #2229


Agetian (approved):

Hanmac:
for the Aura thing i need to fully rework Enchant keyword because it might be possible to have a card with multiple enchant keywords

PR #3282, opened 2023-06-17
Support common adventure resources, relocate resources to common directory, sorting, and groundwork for Amonkhet

schnautzr:
Add support for common resources for adventure mode. Particle effects, skin, sprites, ui, and most of the world files are now common resources. A file in "Shandalar" (or whatever plane you are using) will take precedence over a file with the same name in "common".

Further work is required to allow common audio, custom cards/pics, decks, maps, and music.

Sprites and decks have a new, more intuitive directory structure. Sprites are sorted now according to creature type at top level, followed by race, then class. (The old structure had inconsistencies that confused creature type, class, race, and even alignment as top-level directories.)

Deck sorting is now reduced to a minimal directory structure.

Amonkhet, an upcoming plane, is added. This plane is a WIP. In order to use the plane, you must change the plane to "Amonkhet" in your settings file and back to "Shandalar" when finished.


kevlahnota:
Just looking at the configuration files, should there be a common configuration and specific plane configuration? Also the custom card and pics, seems they should also be in common folder.. only specific configs and files should be seperated per plane I think so there will no duplicates.

schnautzr:
> Just looking at the configuration files, should there be a common configuration and specific plane configuration? Also the custom card and pics, seems they should also be in common folder.. only specific configs and files should be seperated per plane I think so there will no duplicates.

Unfortunately I haven't figured out yet how to get the custom card/pics to work in the common directory. But I agree.

schnautzr:
All sprite paths and deck paths are now resolved.

forge-gui-mobile/src/forge/adventure/util/Config.java
kevlahnota, line 1: on line 61, you need a method to filter only the plane folders (Shandalar, Amonkhet) or the common folder will be included in the list of planes which is not a plane. On Line 108, you need to update the path of the config.json to the commons folder, move the config.json to commons folder so this will be the main config.json, and prevent duplicated resources.

schnautzr: Not sure I understand what line 61 is doing, but I've taken care of line 108 and the corresponding move as requested. Initially it was suggested we use the config file to do plane-specific things, but I haven't done so yet.

kevlahnota: > Not sure I understand what line 61 is doing, but I've taken care of line 108 and the corresponding move as requested. Initially it was suggested we use the config file to do plane-specific things, but I haven't done so yet.

Line 61 gets all the files and directories from the defined path. So you must filter it or those unwanted files and folders like the common folder and other files that exists there that will also be included and treated as "user planes", try enabling the code for the combo box for the plane selection in settingscene, it should only load shandalar and amonkhet folders (and maybe other future user planes).

kevlahnota: also for line 280, and line 281, you can update it ie: Currentconfig.getcommonprefix() + paths.CUSTOM_CARDS and Currentconfig.getcommonprefix() + paths.CUSTOM_CARDS_PICS if you will move the custom cards and custom card pics to common folder

schnautzr: image
Done.

forge-gui-mobile/src/forge/adventure/util/Config.java
```diff
@@ -58,7 +56,17 @@ static public Config instance()
     private Config() {
 
         String path= resPath();
-         adventures = new File(GuiBase.isAndroid() ? ForgeConstants.ADVENTURE_DIR : path + "/res/adventure").list();
+        FilenameFilter planesFilter = new FilenameFilter() {
+            @Override
+            public boolean accept(File file, String s) {
+                if (!s.contains(".") && !s.equals(commonDirectoryName)) {
```
kevlahnota, line 62: Does file.isdirectory() works?

forge-gui-mobile/src/forge/adventure/util/Config.java
```diff
@@ -98,7 +106,7 @@ private Config() {
 
         //prefix = "forge-gui/res/adventure/Shandalar/";
         prefix = getPlanePath(settingsData.plane);
-        commonPrefix = getPlanePath("common");
+        commonPrefix = getPlanePath(commonDirectoryName);
```
kevlahnota, line 109: i still think common directory is not planepath but common resource folder for the planes, can the adventure directory have planes directory and put the amonkhet and shandalar there and just read there? I think this will eliminate the file namefilter since anything you put there are folder for planes.

forge-gui-mobile/src/forge/adventure/util/Config.java
schnautzr: No, I tried that. Here, "file" is actually referring to the adventure directory, and "s" is a string from the list of filenames.

forge-gui-mobile/src/forge/adventure/util/Config.java
schnautzr: If it makes you feel better we can duplicate getPlanePath() with a different name... I don't think it makes sense for the common folder to be located anywhere other than with Amonkhet and Shandalar, since each is technically a collection of resources, not a plane.

schnautzr: Updated so it's not using getPlanePath for that.

kevlahnota: I thought you want to organize this, so under adventure directory there is common folder and planes folder, and inside planes folder, amonkhet and shandalar reside. Doesn't that makes better?

schnautzr: Usually a common folder shares the same home as the directories that override it. I've never seen an implementation that put the overrides all in a subdirectory.

kevlahnota:
@jjayers99 @Simisays what do you think still needs to be done for this PR? for me I think the custom card and custom pics should be in common folder. Can you provide any suggestions or any review? thanks.

TrueFuFLeaderG:
Could you please avoid using something like this
```java
      int x=0;
        if(Cache.containsKey(fullPath)) x=1;
        else if(Cache.containsKey(commonPath)) x=2;
        switch(x) {
            case 2:
                return Cache.get(commonPath);
            case 0:
```
Please use a enum or static finial variable and hide the "magic numbers"

```java
static final int NotInCache = 0;
static final int FullPathInCache= 1;
static final int CommonPathInCache= 2;
```
but you can just avoid this whole check by always saving it under the FullPath, it is not really important where the resource came from.
```java
                } else if (Files.exists(Paths.get(fullPath))) {
                    Cache.put(fullPath, new FileHandle(fullPath));
                } else if (Files.exists(Paths.get(commonPath))) {
                    Cache.put(fullPath, new FileHandle(commonPath));
```
Also you could save a tiny bit of performance if you make the commonPath only when it did not find the file in the full path.


For the custom cards you should be able to just load also the common path
```java
    public void loadResources() {
        RewardData.getAllCards();//initialize before loading custom cards
        final CardRules.Reader rulesReader = new CardRules.Reader();
        ImageKeys.ADVENTURE_CARD_PICS_DIR=Config.currentConfig.getFilePath(forge.adventure.util.Paths.CUSTOM_CARDS_PICS);// not the cleanest solution
        for(File cardFile:new File(getFilePath(forge.adventure.util.Paths.CUSTOM_CARDS)).listFiles())
        {

            FileInputStream fileInputStream;
            try {
                fileInputStream = new FileInputStream(cardFile);
                rulesReader.reset();
                final List<String> lines =  FileUtil.readAllLines(new InputStreamReader(fileInputStream, Charset.forName(CardStorageReader.DEFAULT_CHARSET_NAME)), true);
                CardRules rules=  rulesReader.readCard(lines, com.google.common.io.Files.getNameWithoutExtension(cardFile.getName()));
                rules.setCustom();
                PaperCard card=new PaperCard(rules, CardEdition.UNKNOWN.getCode(), CardRarity.Special){
                    @Override
                    public String getImageKey(boolean altState) {
                        return ImageKeys.ADVENTURECARD_PREFIX + getName();
                    }
                };
                FModel.getMagicDb().getCommonCards().addCard(card);
            } catch (FileNotFoundException e) {
                throw new RuntimeException(e);
            }
        }
```
But for the images you probably have to add a second search location like ImageKeys.ADVENTURE_CARD_PICS_DIR but i also do not really see a problem with only loading cards from the common folder besides a little bit inconsistent

schnautzr:
Thanks for the insight! I can definitely relate to the cryptic variables, having worked with a QBASIC project that actually reused existing single-letter variables for multiple purposes without even changing their value within proximity of their use! I'll have another pass at this and return with changes.

schnautzr:
@TrueFuFLeaderG Made requested changes and optimized getFile. Can we go ahead and pull this and do cards/images as a separate pull in the future?

TrueFuFLeaderG (commented):
Looks fine for me.

Delaying the custom cards is fine for me.
If we want custom cards in the shop/reward random pools then we do not want them in all planes anyway (currently they should not appear there, at least that was my intention)

TrueFuFLeaderG (approved):

Issue #3262, opened 2023-06-13
AI Logic and Optional Cost extract into SpellAbilityAI


Hanmac:
`PlayerControllerAi#chooseOptionalCosts` needs to respect the API of the SpellAbility
to make the optional cost depend of the the Effect

Also need better logic when NOT trying to pay the Optional Cost

PR #3284, opened 2023-06-17
Several AI improvements

Agetian:
- Improved Blitz AI support (will cast in Main 1 when viable, will avoid casting in Main 2).
- Hold for Ransom AI hint to actually use the ability when able.
- Basic AI support for Grisly Sigil so that at least it doesn't fizzle.
- Basic BranchAi framework, currently will mostly just return true, but may be used for special logic when needed.

Current issues, most likely for the future PR:
- Grisly Sigil is very basic. Casualty payment is currently disabled for it (the AI will always pay for it, and when it does, it'll also fail to target something else, choosing the same target as for the original). Probably needs #3262 in order to work better, as well as some improvement to the targeting system for copied spells.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/SpecialCardAi.java
+++ b/forge-ai/src/main/java/forge/ai/SpecialCardAi.java
@@ -632,7 +632,6 @@ public class SpecialCardAi {
             CardCollection potentialTgts = CardLists.filterControlledBy(CardUtil.getValidCardsToTarget(sa), ai.getOpponents());
 
             for (Card c : potentialTgts) {
-
                 int potentialDamage = c.getAssignedDamage(false, null) > 0 ? 3 : 1; // TODO: account for damage reduction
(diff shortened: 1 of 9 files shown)


forge-ai/src/main/java/forge/ai/SpecialCardAi.java
```diff
@@ -625,6 +625,34 @@ public static boolean consider(final Player ai, final SpellAbility sa) {
         }
     }
 
+    // Grisly Sigil
+    public static class GrislySigil {
+        public static boolean consider(final Player ai, final SpellAbility sa) {
+            // TODO: improve targeting support for Casualty 1
+            List<Card> validTgts = CardUtil.getValidCardsToTarget(sa);
+            CardCollection potentialTgts = new CardCollection();
```
tool4ever, line 633: AI might target its own cards? 🤔

tool4ever (approved):

PR #3287, opened 2023-06-17
Fix Green Castle crash on non-case insensitive file systems, fix Ghalta's Presence effect.

Agetian:
- Fix Green Castle crash on any non-Windows operating system with a case sensitive file system.
- Fix Ghalta's Presence increasing the cost of everything (most likely should be non-creature and non-battle for a single card, otherwise it increases the costs of [non-battle] creatures and [non-creature] battles.


Simisays (approved):
Thanks :thumbsup:

PR #3293, opened 2023-06-18
AI hint for Traverse the Outlands

### LINKED ISSUE #3288: KI use traverse the Outlands incorrect.
KI use Card incorrect. Card is traverse of the Outlands.

**To Reproduce**
1. Play Commander Game against several KI ( KI play a Deck including Traverse the Oulands)
2. Wait & play a few Turns 
3. KI plays the card in two games Traverse the Oulands without a creature on the board ( invest 5 Mana for 0 Lands)

**Expected behavior**
Traverse the Oulands should only be played, when the controller habe a creature on the board. (best is big creature with high power)

**Screenshots**

Agetian:
Fixes #3288 

Is PowerGE2 alright here or should it be buffed higher? :/


tool4ever (approved):
I feel like this could be something the API check should recognize by itself 🤔

Agetian:
Yeah, that can definitely be figured out as well :) i'll think if I can find a way to do that

PR #3295, opened 2023-06-18
ChangeZone AI: bail if ChangeNum is a variable and will be 0

Agetian:
A more general case where the AI shouldn't activate AF ChangeZone if no cards will change zone because ChangeNum is 0.
Probably could use some testing since ChangeZone AI is very tangled and easy to break.


tool4ever (approved):
after a brief look at the affected scripts I'd say it will be ok

PR #3301, opened 2023-06-20
Enforce explicit plugin versions

Tillerino:
Hi!

I recently bumped the Maven version in Nixpgks (NixOS/nixpkgs#238746) and ran into some small but annoying manual effort w.r.t. the `forge-mtg` package. The issue is that `forge-mtg` is packaged using the [double invocation](https://ryantm.github.io/nixpkgs/languages-frameworks/maven/#double-invocation) method which unfortunately breaks if not all maven plugin versions are set explicitly.

We're extending the documentation to prevent this (NixOS/nixpkgs#238774), but I'm reaching out to fix the packages which have already been packaged. In this PR, we explicitly set the versions of the only two plugins which hadn't been set explicitly before and add a check with the enforcer plugin to make sure this won't accidentally break in the future.

I took the liberty of deduplicating the configuration of the enforcer plugin, since this was already causing inconsistencies (`requireMavenVersion`).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,26 +96,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-enforcer-plugin</artifactId>
-                    <version>3.0.0-M2</version>
-                    <executions>
-                        <execution>
-                            <id>enforce-versions</id>
-                            <phase>validate</phase>
-                            <goals>
-                                <goal>enforce</goal>
-                            </goals>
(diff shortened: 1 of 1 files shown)


Agetian:
This change seems to work fine, I tested it on my Linux dev workflow (running Linux Mint 21.1). Haven't noticed anything off when compiling via Maven. Anyone else willing to run some tests? (esp. on other systems)

kevlahnota (approved):

Issue #2219, opened 2023-01-08
Lithoform Engine can't copy its own ability


Samthere:
I used Lithoform Engine's first ability, then untapped it and used it again with the first instance on the stack. The game would not allow me to target the first ability (presumably erroneously as part of a check preventing an instance of the ability from targeting itself). Not sure if there's a situation where it matters, but still!

Hanmac:
@tool4ever @Northmoc this is probably very tricky because Activated Abilities aren't copied when going on the Stack

and we probably can't use SpellAbilityStackInstance either. (I'm still indifferent about this class, i don't know if we should extend the use or remove it)

tool4ever:
Agreed

I'll see if I can find a clean bypass for the time but the current navigation from SA->SI might make that too hard without refactor 🤷‍♂️

Hanmac:
@tehdiplomat i ping you also, this is probably something bigger that we need to discuss in the developers channels

as for StackInstance, i don't know how much refactor is needed ... or what the right direction would be

* like for Trigger, the Ability inside is currently copied, but if StackInstance would work better, that might not be needed anymore
* or if Activated Abilities would be copied, then we might not need StackInstance anymore

tehdiplomat:
Well the whole point of StackInstance is we had some major issues with targeting resetting especially in regards to the AI testing abilities.

tool4ever:
Huh such cheating is certainly a risk to keep in mind but hopefully it is now also covered by other parts
Otherwise we can try to find another fix as I'm currently leaning more towards getting rid of it if we can't find a better reason

Imo some abstraction is helpful but usually things get ugly when we move too far from the rules and having this additional layer on the stack always felt clumsy

tehdiplomat:
What do we have that covers this same space? A SpellAbility that was attached to a Card (Something that you can activate and pick targets), was the exact same thing that was used while it was on the stack. So while its on the stack, you can still activate it again, and what is in place to prevent the SA on the Stack from being overwritten? We're not creating a new instance of the SpellAbility each time it hits the stack, or replacing the old SA with a "clean" one at that point right? 

If we have a good way of doing it cleaner that's fine, but I'm not convinced anyone has written anything that covers this ground?

tool4ever:
No, it would lead to having to copy activated abilities

But seeing how we're already doing that for triggers and when copying spells anyway this would also allow us further cleanup of the branching cases for it

tehdiplomat:
[...]

Issue #1179, opened 2022-07-23
Information about illegal targets must still be provided by the engine in some cases


tool4ever:
When you cast _Searing Blaze_ targeting opponent + his creature, if he responds with _Blossoming Calm_, the creature should still be damaged after.
Currently the `Creature.ControlledBy ParentTarget` will also turn illegal because the player got removed.

https://twitter.com/Dunkatog/status/1439013991330680833

Hanmac:
We probably should do illegal Targets as extra property so the original Targets arent't touched

But then we might need to check all other places where getTargets are returned ...
This might also only be set by the MagicStack, any other checks by the AI for maybe illegal targets should not alter them.

someone else has some ideas for this?

tool4ever:
It could probably be fixed the same way I did _Yosei, the Morning Star_ with `RememberOriginalTargets$`

tool4ever:
_Biomantic Mastery_ is another case

PR #3306, opened 2023-06-21
Remove outdated fizzle code

tool4ever:
This old (https://github.com/Card-Forge/forge/commit/25f0e0) timestamp-related code seems very problematic:

- technically it shouldn't even fizzle untargeted ones
- it breaks other cards that don't want to affect the target directly, e.g. _Bellowing Fiend_
- the card mentioned in the original commit is already fixed for LKI usage anyway

So we rip it out asap.

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -699,11 +699,6 @@ public class GameAction {
                 copied.getDamageHistory().setNotBlockedSinceLastUpkeepOf(p);
                 copied.getDamageHistory().setNotBeenBlockedSinceLastUpkeepOf(p);
             }
-            if (zoneFrom.is(ZoneType.Graveyard)) {
-                // fizzle all "damage done" triggers for cards returning to battlefield from graveyard
-                game.getStack().fizzleTriggersOnStackTargeting(copied, TriggerType.DamageDone);
-                game.getStack().fizzleTriggersOnStackTargeting(copied, TriggerType.DamageDoneOnce);
-            }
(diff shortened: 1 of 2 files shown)


Hanmac:
@tool4ever for Magic Stack, i want to refactor the hasFizzled check

currently: it checks on each target, and if the target is invalid, it gets removed
what i want: check each target, and if the target is invalid, put them in a list, and remove them later

with that, it wouldn't invalidate stuff like "target player and creature they control" if the player gets hexproof

tool4ever:
Yes that should help for #1179

But it doesn't depend on keeping this bad logic :P

Hanmac:
> Yes that should help for #1179
> 
> But it doesn't depend on keeping this bad logic :P

yeah, i just thought you could do that in the same try

tool4ever:
Probably not, currently I'm mostly working on `SpellAbilityStackInstance` problems though progress is slow

Hanmac:
> Probably not, currently I'm mostly working on `SpellAbilityStackInstance` problems though progress is slow

i don't know if we already make an Issue for this, (or something similar?)
but we should try to use SaSi on the Stack instead of SpellAbility while on the stack

so it would be the StackInstance that is getting checked for canTarget and isValid checks instead of SpellAbility

the main reason is that Activated Abilities (other than Spells or Triggers) aren't copied when going on the Stack

so we either need to
* copy ActivatedAbilties when going to the stack
* or use StackInstance instead of (Activated)SpellAbilties for isValid

tool4ever:
You probably mean this one: #2219

There are other even more annoying cases where the wrong Instance might be affected

Unless something causes more problems your approaches are also my endgame

PR #3331, opened 2023-06-24
Common music

schnautzr:
Move all music to common resources.
Add support for common music, allowing Shandalar/music to override.
Add colorless biome music folder (populated with menu1.mp3).
Add support for colorless biome music.

Requiring further work: Shandalar will override regardless of the plane in use. Need to make it get the plane name and use the appropriate plane's override resources.
Colorless contains the menus1.mp3 that it's been using to date. Need a different track for this biome.


forge-gui-mobile/src/forge/adventure/stage/GameHUD.java
```diff
@@ -734,7 +736,7 @@ public void updateMusic() {
                 changeBGM(MusicPlaylist.WHITE);
                 break;
             case "waste":
-                changeBGM(MusicPlaylist.MENUS);
+                changeBGM(MusicPlaylist.COLORLESS);
```
kevlahnota, line 739: why not call MusicPlaylist.MENUS to play menu1.mp3 if that's the default music for waste?

schnautzr: Because I want to be able to put music in a folder for colorless that's not the same as the menu music. Of course, the music I'm putting there isn't music I can upload to an open project like this, but having the distinction at least allows users to do this. I don't know where you guys get the music from, but it'd be a good idea to find a track to put here.

Simisays:
There seem to be no music playing in the waste zone ? Otherwise looks good to me

schnautzr:
> There seem to be no music playing in the waste zone ? Otherwise looks good to me

Should be pulling from the colorless directory.

kevlahnota (approved):

PR #3347, opened 2023-06-27
Fix wrong decider when copying trigger

### LINKED ISSUE #3344: Possible bug with how additional/replacement tokens are created
**Describe the bug**
I have Gala Greeters on field alongside Peregrin Took. I cast a creature and create a tapped treasure token, as well as a food token, but the food token also comes in tapped. Is this the correct behavior? I would've assumed that the food token would come in untapped.

tool4ever:
Closes #3344

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
@@ -621,7 +621,7 @@ public abstract class SpellAbilityEffect {
     protected static boolean addToCombat(Card c, Player controller, SpellAbility sa, String attackingParam, String blockingParam) {
         final Card host = sa.getHostCard();
         final Game game = controller.getGame();
-        if (!game.getPhaseHandler().inCombat()) {
+        if (!c.isCreature() || !game.getPhaseHandler().inCombat()) {
             return false;
         }
         boolean combatChanged = false;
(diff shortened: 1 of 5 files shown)


forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
```diff
@@ -621,7 +621,7 @@ public void run() {
     protected static boolean addToCombat(Card c, Player controller, SpellAbility sa, String attackingParam, String blockingParam) {
         final Card host = sa.getHostCard();
         final Game game = controller.getGame();
-        if (!game.getPhaseHandler().inCombat()) {
+        if (!c.isCreature() || !game.getPhaseHandler().inCombat()) {
```
Hanmac, line 624: the creature check only for attacking

for blocking it might be a planeswalker or battle

tool4ever: no, this is not for defender stuff like _Portal Mage_

c is always meant as attacker or blocker which means only creature is legal

> 509.7. If a **creature** is put onto the battlefield blocking...

Hanmac: hm ok fine

Agetian (approved):

PR #3349, opened 2023-06-27
Support scripted switching to different trigger context

tool4ever:
With this language extension manual remembering on triggers can be reduced.

This ends up being equivalent because by doing ``setSpawningAbility(sa.copy(...`` we are basically already storing everything that the SpellAbility holds safely anyway.


Hanmac:
otherwise, would it make sense to always get the value from the spawning ability, or would that be too problematic?

tool4ever:
I think there are some cards where the second trigger needs its own triggeredObjects and not the parent ones
So it would have to be coded as implicit fallback to the spawningAbility if it can't find in the current scope:
- scripters would need to know which AbilityKey can collide and use remember to keep the right one
- the way AbilityUtils is structured this fallback would have to be injected into multiple branches

🤔

tool4ever:
@Hanmac 
I think it can be merged?

Hanmac (approved):
i think this is okay

if not it can be changed later

Issue #2356, opened 2023-01-31
PutCounter + Existing + Can't Effect


Hanmac:
* I control _Animation Module_ (Put existing Counter on Perm)
* Opponent has Creature with -1/-1 counter and another (like Oil)
* Opponent also has _Melira, Sylvok Outcast_ (-1/-1 counters can't be put on creatures)
* I Activate Animation Modules ability targeting the creature with Multiple Counters.

now i should be able to choose the -1/-1 counter and do nothing (this might matters in case opponent flashes in Melira)

PR #3354, opened 2023-06-28
Goldberry support

tool4ever:


Hanmac:
@tool4ever other checks for canRecrive might be removed too, like this one:
https://github.com/Card-Forge/forge/blob/83a23d054a6db2d80f10f5ae562c46d47a57623c/forge-game/src/main/java/forge/game/ability/effects/CountersMoveEffect.java#L290

like for https://github.com/Card-Forge/forge/issues/2356

tool4ever:
I'm not sure...

_Animation Module_:
> Choose a counter on target permanent or player. Give that permanent or player another counter of that kind.

This is Action A followed by Action B.

But _Leech Bonder_ is just a single action:
> Move a counter from target creature onto a second target creature.

In other situations like this there's been rulings that when it's a single action your choices are limited under CR 608.2d:
> The player can't choose an option that's illegal or impossible...

➡️ since moving -1/-1 on your _Melira, Sylvok Outcast_ becomes impossible you can't choose that type

forge-game/src/main/java/forge/game/ability/effects/CountersMoveEffect.java
```diff
@@ -300,10 +303,6 @@ public void resolve(SpellAbility sa) {
 
                         removeCounter(sa, source, cur, chosenType, counterNum, countersToAdd);
                     } else {
-                        if (!cur.canReceiveCounters(cType)) {
-                            continue;
-                        }
```
Hanmac, line 305: then you shouldn't remove this part?

tool4ever: that check was redundant because `removeCounter` does the same too

Agetian (approved):

PR #3365, opened 2023-06-29
Fix logic

tool4ever:


forge-ai/src/main/java/forge/ai/ability/AttachAi.java
```diff
@@ -1260,7 +1261,7 @@ public boolean apply(final Card c) {
         // TODO Somehow test for definitive advantage (e.g. opponent low on health, AI is attacking)
         // to be able to deal the final blow with an enchanted vehicle like that
         boolean canOnlyTargetCreatures = true;
-        for (String valid : attachSource.getFirstAttachSpell().getTargetRestrictions().getValidTgts()) {
+        for (String valid : ObjectUtils.firstNonNull(attachSource.getFirstAttachSpell(), sa).getTargetRestrictions().getValidTgts()) {
```
Hanmac, line 1264: Shouldn't it better check if this is an Aura first?

PR #3373, opened 2023-06-30
Fix collision issues in Shandalar

schnautzr:
Many sprites in the overworld converted to structures so they have collision. New structure models created. Biomes reworked to include the new structures.


kevlahnota:
Does older save will continue to work without problems?

schnautzr:
> Does older save will continue to work without problems?

I didn't think to check; I've always hit "New Game+" when using newer revisions.

schnautzr:
> Does older save will continue to work without problems?

Yes. Just tested.

kevlahnota (approved):

PR #3385, opened 2023-07-01
Regenerate rework

### LINKED ISSUE #3304: Regenerate API Changes and RegenerationTrigger
* `HIDDEN CARDNAME can't be regenerated` as Static Ability (need some AI changes?)
* Replace `RegenerationTrigger` with a `RegenerationSubAbility` using AdditionalSubAbility. No Play with SVars
* RegenerateBaseEffect needs to transfer this ability into the inner `Regeneration` Effect
* then the Inner Regeneration Effect should use that for an `ImmediateTriggerEffect` like thing or the outer effect should add a Subability
* The `Regeneration` Effect doesn't need to check for `if (!c.canBeShielded() || !c.isInPlay()) {` because the `ReplaceDestory` should already check it? so Subability for the ImmediateTriggerEffect shouldn't be a problem?
* Then we might remove Regenerated Trigger?

Some special cases:
* `Debt of Loyalty` doesn't say "When/Whenever" , so it shouldn't use the stack and be static?
  > Regenerate target creature. You gain control of that creature if it regenerates this way.
* `Soldevi Sentry` probably should remember the targeted player in a different way?
  > Choose target opponent. Regenerate Soldevi Sentry. When it regenerates this way, that player may draw a card.

Hanmac:
Closes #3304

Might need some better AI logic for Effect + CantRegenerate


forge-game/src/main/java/forge/game/ability/effects/RegenerateBaseEffect.java
```diff
@@ -50,26 +58,14 @@ public void createRegenerationEffect(SpellAbility sa, final Iterable<Card> list)
             eff.addRemembered(AbilityUtils.getDefinedObjects(hostCard, sa.getParam("RememberObjects"), sa));
         }
 
-        if (sa.hasParam("RegenerationTrigger")) {
-            final String str = sa.getSVar(sa.getParam("RegenerationTrigger"));
-
-            SpellAbility trigSA = AbilityFactory.getAbility(str, eff);
-
-            final String trigStr = "Mode$ Regenerated | ValidCause$ Effect.Self | TriggerZones$ Command "
```
tool4ever, line 58: Remove TriggerRegenerated from engine?

Hanmac: yeah, i forgot this one

forge-ai/src/main/java/forge/ai/ability/PumpAiBase.java
```diff
@@ -166,12 +166,6 @@ public boolean apply(final Card c) {
                 return false;
             }
             return ph.isPlayerTurn(ai) || (combat != null && combat.isAttacking(card) && card.getNetCombatDamage() > 0);
-        } else if (keyword.endsWith("CARDNAME can't be regenerated.")) {
```
Agetian, line 169: How should this part be implemented right now? It's a pretty extensive PR, pretty difficult to tell what exactly needs doing in order to make sure the AI doesn't break for a bunch of cards...

Agetian (changes requested):

forge-ai/src/main/java/forge/ai/ability/PumpAiBase.java
Hanmac: i probably need to code that in EffectAi, i need to think about it what would be the best logic for this

it probably needs to check if there is already an effect that tires to regenerate the Card
and if the card still can regenerate

Agetian: Alrighty, please assist with this if able :)

Hanmac: Effect AI Cases:

* As SubAbility of DealDamage like `Carbonize`, just treat them as bonus
* as Activated Ability with Target from `Furnace Brood`, probably should choose a creature that already has an RegenerationShield or another Ability with `Regeneration$ True` , when using ReplacementList, it already filters if the creature can regenerate or not, just check for Replace Destroy Effects that doesn't use Regeneration, like Totem Armor
* Spell with Target but with Subability like `Gravebind`, but it should just care about the draw one
* Nimbus Cards, like `Knight of the Holy Nimbus`, AI should try to activate it at least once? should check if it would be destroyed by damage or other effects on the stack?

Hanmac:
@Agetian i started logic for CantRegenerate

right now, it should only use it against stuff that has a Regeneration Shield (or similar)

what we might need now is to check if the creature would be destroyed by something on the stack
(or if it would be damaged enough + check for damage prevention)

Agetian:
> @Agetian i started logic for CantRegenerate
> right now, it should only use it against stuff that has a Regeneration Shield (or similar)
> what we might need now is to check if the creature would be destroyed by something on the stack (or if it would be damaged enough + check for damage prevention)

Nice! Yeah, that should be possible to do, there's some similar code checking the top of the stack in ChangeTargetsAi and CounterAi, and there's some code that checks for whether something can be killed in dealDamageChooseTgtC in DealDamageAi, not sure if that would be adaptable to your needs?

Hanmac:
yeah, first I'm doing to check for Destroy Effect on the Stack that doesn't has `NoRegen` so the Spell/Ability would try to Regen it

as Second Part i'm going to check for Damage to see if the creature could be dealt damage until its deadly (assuming the damage source doesn't have wither or similar)


Or i could go the other way around it, and check if the source of the Regeneration is an Effect or an ability like from Nimbus creatures.
so in case of Regeneration Effect in Command/Effect Zone, it probably has a reason why its trying to regenerate. (so i make the AI logic play against them)

and i only need to check for Destory/Damage on the other cases like the Nimbus ones

Hanmac:
@Agetian i added some logic to Peek for Destory/DestroyAll on the Nimbus Creatures
the ones with inbuilt regenerate effect with activated ability to disable it (by opponent)

i want opponent only to activate it when there is something on the stack that would destroy them
either by direct destroy effect, or by damage

i think the first part with destroy check is done, (i didn't tested it yet)
and the second part with damage need to predict the damage amount and if it would be lethal

forge-ai/src/main/java/forge/ai/ability/EffectAi.java
```diff
@@ -256,6 +271,171 @@ public boolean apply(final Card c) {
...
+                        return false;
+                    }
+                    // TODO check Stack for Effects that would destroy host, either direct or indirect
+                    SpellAbility stackSa = stack.peekAbility();
+                    if (stackSa == null) {
+                        return false;
+                    }
+
+                    // need to get the wrapped sub abilities
+                    if (stackSa.isWrapper()) {
```
tool4ever, line 329: imo all calls below should be handled by WrappedAbility calls
so no need to extra unwrap?

Hanmac: hm maybe, you could remove it if you want

Agetian:
> @Agetian i added some logic to Peek for Destory/DestroyAll on the Nimbus Creatures the ones with inbuilt regenerate effect with activated ability to disable it (by opponent)
> 
> i want opponent only to activate it when there is something on the stack that would destroy them either by direct destroy effect, or by damage
> 
> i think the first part with destroy check is done, (i didn't tested it yet) and the second part with damage need to predict the damage amount and if it would be lethal

Very nice! Thanks for seeing this through! I'm currently out of town, so can't test stuff, but once I'm back (next week) I'll be able to take a look unless someone does it faster :)

forge-ai/src/main/java/forge/ai/ability/PumpAiBase.java
Hanmac: i did some logic for the Nimbus cards, for the other cards that uses Target, i just let AI handle the other part

Agetian (dismissed):

forge-gui/res/cardsfolder/r/runesword.txt
```diff
@@ -2,11 +2,10 @@ Name:Runesword
 ManaCost:6
 Types:Artifact
 A:AB$ Pump | Cost$ 3 T | NumAtt$ +2 | ValidTgts$ Creature.attacking | TgtPrompt$ Select target attacking creature | SubAbility$ RuneswordEffect | SpellDescription$ Target attacking creature gets +2/+0 until end of turn. When that creature leaves the battlefield this turn, sacrifice CARDNAME. If the creature deals damage to a creature this turn, the creature dealt damage can't be regenerated this turn. If a creature dealt damage by the targeted creature would die this turn, exile that creature instead.
-SVar:RuneswordEffect:DB$ Effect | Name$ Runesword Effect | Triggers$ TrigSacRunesword,TrigNoregen | ReplacementEffects$ RuneswordRep | RememberObjects$ Targeted | ImprintCards$ Self
+SVar:RuneswordEffect:DB$ Effect | Name$ Runesword Effect | Triggers$ TrigSacRunesword | ReplacementEffects$ RuneswordRep | StaticAbilities$ NoRegen | RememberObjects$ Targeted | ImprintCards$ Self | ForgetOnMoved$ Battlefield | AILogic$ CantRegenerate
```
tool4ever, line 5: this effect needs to stay even if the creature leaves?

forge-gui/res/cardsfolder/b/bone_shaman.txt
```diff
@@ -2,7 +2,7 @@ Name:Bone Shaman
 ManaCost:2 R R
 Types:Creature Giant Shaman
 PT:3/3
-A:AB$ Animate | Cost$ B | Defined$ Self | staticAbilities$ BoneShamanStatic | StackDescription$ CARDNAME gains "Creatures dealt damage by CARDNAME this turn can't be regenerated this turn." | SpellDescription$ Until end of turn, CARDNAME gains "Creatures dealt damage by CARDNAME this turn can't be regenerated this turn."
-SVar:BoneShamanStatic:Mode$ Continuous | Affected$ Creature.DamagedBy | AddHiddenKeyword$ CARDNAME can't be regenerated. | Description$ Creatures dealt damage by CARDNAME this turn can't be regenerated this turn.
+A:AB$ Animate | Cost$ B | Defined$ Self | staticAbilities$ BoneShamanStatic | AILogic$ BoneManCantRegenerate | StackDescription$ CARDNAME gains "Creatures dealt damage by CARDNAME this turn can't be regenerated this turn." | SpellDescription$ Until end of turn, CARDNAME gains "Creatures dealt damage by CARDNAME this turn can't be regenerated this turn."
```
tool4ever, line 5: no logic exists

forge-gui/res/cardsfolder/j/jaya_ballard_task_mage.txt
```diff
@@ -5,5 +5,6 @@ PT:2/2
 A:AB$ Destroy | Cost$ R T Discard<1/Card> | ValidTgts$ Permanent.Blue | TgtPrompt$ Select target blue permanent | SpellDescription$ Destroy target blue permanent.
 A:AB$ DealDamage | Cost$ 1 R T Discard<1/Card> | ValidTgts$ Any | NumDmg$ 3 | SubAbility$ DBNoRegen | SpellDescription$ CARDNAME deals 3 damage to target creature or player. A creature dealt damage this way can't be regenerated this turn.
```
tool4ever, line 6: remembering missing

forge-gui/res/cardsfolder/r/runesword.txt
Hanmac: The Static and Replacement Effect probably want to have ForgetOnMoved

But the leave the Battlefield trigger interferes with it

can you help me to code the Leave The Battlefield trigger as Delayed Trigger? then it doesn't need to be part of the effect

Hanmac: i moved the leave the battlefield trigger into a Delayed one

but i still removed ForgetOnMoved

tool4ever (approved):

MorihiroS:
Is it actual after rework?

PR #3386, opened 2023-07-01
Exiled card becomes new object when it's exiled

tool4ever:
This fixes a missing opportunity to "reduce" some of the interactions of an exiled card.

For testing have _Rest in Peace_ and try to remove exiled cards with _Pull from Eternity_:
- with the activated ability of _Torrent Elemental_ on the stack it should be prevented from ETB
- ExiledWith links should break
- MayPlay stuff should end
- etc.

Because the rules say it doesn't change zones this would also stop effect cleanup, but luckily we have another trigger for this.

- [x] Fix _Imminent Doom_

### DIFF
--- a/forge-game/src/main/java/forge/game/trigger/TriggerSpellAbilityCastOrCopy.java
+++ b/forge-game/src/main/java/forge/game/trigger/TriggerSpellAbilityCastOrCopy.java
@@ -279,11 +279,12 @@ public class TriggerSpellAbilityCastOrCopy extends Trigger {
         sa.setTriggeringObject(AbilityKey.LifeAmount, castSA.getAmountLifePaid());
         sa.setTriggeringObjectsFrom(
                 runParams,
-            AbilityKey.Player,
-            AbilityKey.Activator,
-            AbilityKey.CurrentStormCount,
-            AbilityKey.CurrentCastSpells
-        );
+                AbilityKey.CardLKI,
+                AbilityKey.Player,
+                AbilityKey.Activator,
+                AbilityKey.CurrentStormCount,
+                AbilityKey.CurrentCastSpells
+                );
     }
 
     @Override
--- a/forge-game/src/main/java/forge/game/zone/MagicStack.java
+++ b/forge-game/src/main/java/forge/game/zone/MagicStack.java
@@ -320,6 +320,14 @@ public class MagicStack /* extends MyObservable */ implements Iterable<SpellAbil
 
         // Copied spells aren't cast per se so triggers shouldn't run for them.
         Map<AbilityKey, Object> runParams = AbilityKey.mapFromPlayer(sp.getHostCard().getController());
+
+        if (sp.isSpell() && !sp.isCopied()) {
+            final Card lki = CardUtil.getLKICopy(sp.getHostCard());
+            runParams.put(AbilityKey.CardLKI, lki);
+            thisTurnCast.add(lki);
+            sp.getActivatingPlayer().addSpellCastThisTurn();
+        }
+
         runParams.put(AbilityKey.Cost, sp.getPayCosts());
         runParams.put(AbilityKey.Activator, sp.getActivatingPlayer());
         runParams.put(AbilityKey.CastSA, si.getSpellAbility(true));
@@ -461,10 +469,6 @@ public class MagicStack /* extends MyObservable */ implements Iterable<SpellAbil
 
         GameActionUtil.checkStaticAfterPaying(sp.getHostCard());
 
-        if (sp.isSpell() && !sp.isCopied()) {
-            thisTurnCast.add(CardUtil.getLKICopy(sp.getHostCard()));
-            sp.getActivatingPlayer().addSpellCastThisTurn();
-        }
         if (sp.isActivatedAbility() && sp.isPwAbility()) {
             sp.getActivatingPlayer().setActivateLoyaltyAbilityThisTurn(true);
         }
--- a/forge-game/src/main/java/forge/game/zone/MagicStack.java
+++ b/forge-game/src/main/java/forge/game/zone/MagicStack.java
@@ -44,7 +44,6 @@ import forge.game.ability.AbilityKey;
 import forge.game.ability.AbilityUtils;
 import forge.game.ability.ApiType;
 import forge.game.card.Card;
-import forge.game.card.CardCollection;
 import forge.game.card.CardUtil;
 import forge.game.event.EventValueChangeType;
 import forge.game.event.GameEventCardStatsChanged;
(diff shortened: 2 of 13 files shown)


Hanmac:
we might look into all the other cases where `suppress` could happen?

are there any other ways that would make a card try to change into the Zone it is already in?

maybe we can remove this extra check anyway?

tool4ever:
Ok I'm leaning towards removing the first part for cleanup...

Milling vs. _Wheel of Sun and Moon_ should also produce new objects:
> 701.16c. If cards in a player's library are shuffled or otherwise reordered, any revealed cards that are reordered stop being revealed and become new objects.

To make that interaction work I've removed the calls I found weren't doing anything like I mentioned on Discord but the problem remains that because of
https://github.com/Card-Forge/forge/blame/754882f546987da2d1f6873320311d57c8e9314d/forge-game/src/main/java/forge/game/ability/effects/MillEffect.java#L43

image

this order will still make the replacement fail (because TriggerHandler fills the LastState).

tool4ever:
I think maybe TriggerHandler shouldn't prefill its LastStates and just let ReplacementHandler decide which is the valid one there? (E. g. some like ChangeZoneEffect can initialize it in in their runParams for better results)
https://github.com/Card-Forge/forge/blame/754882f546987da2d1f6873320311d57c8e9314d/forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java#L319

Hanmac:
Hm if multiple cards cause multiple replacement effects to apply, the last known states needs to be before everything, that's why changes zones sets the run params
And the trigger needs to be set from this too

For your change of removing the lines that sets both LastState to Empty
That was because of memory leak

You need to check if you didn't break it again

tool4ever:
Everything seems to work during testing

Any specific scenarios I should try?

I will work on the remaining task after you think the rest is ok

Hanmac:
i was wondering if we could remove `suppress` but it seems to be not jet

tool4ever:
Yea I don't think so
scrying to bottom is just no real zone change

forge-gui/res/cardsfolder/i/imminent_doom.txt
```diff
@@ -6,5 +6,5 @@ T:Mode$ SpellCast | ValidCard$ Card.cmcEQX | ValidActivatingPlayer$ You | Trigge
 SVar:TrigDealDamage:DB$ DealDamage | ValidTgts$ Any | NumDmg$ Y | SubAbility$ DBPutCounter
 SVar:DBPutCounter:DB$ PutCounter | Defined$ Self | CounterType$ DOOM | CounterNum$ 1
 SVar:X:Count$CardCounters.DOOM
-SVar:Y:LastStateBattlefield$CardCounters.DOOM
+SVar:Y:ThisTurnCast$Card.TriggeredCard$CardManaCost
```
Hanmac, line 9: Can't you get the cmc from the Triggered Spell or something?
`SVar:Y:TriggeredStackInstance$CardManaCostLKI`
There is no need for extra This Turn Cast?

Hanmac: @tool4ever ☝️

tool4ever: @Hanmac 
not really I think
this card needs to always look back:
- so if I use `TriggeredSpellAbility`
- it's X spell
- there is _Unbound Flourishing_

it might deal the wrong damage :/

tool4ever: > The amount of damage Imminent Doom’s triggered ability deals is the number of counters it had on it as the ability triggered.

Hanmac: > > The amount of damage Imminent Doom’s triggered ability deals is the number of counters it had on it as the ability triggered.

But that amount should be cmc when the ability triggered

tool4ever: - yes and if I cast _Endless One_ with X=1
- then resolve _Unbound Flourishing_ first
- `TriggeredStackInstance$CardManaCostLKI` is 2

I wrote it explicitly to only use LKI if the zone changed because other cards like _Vial Smasher_ should see the current value if X gets changed by that special card

Hanmac: yeah, its CastSACMC that got removed with this:
https://github.com/Card-Forge/forge/commit/d011c219b9dc142fd56aaf060cd764e74c566cca

Trigger sets these three values:
```java
        sa.setTriggeringObject(AbilityKey.Card, castSA.getHostCard());
        sa.setTriggeringObject(AbilityKey.SpellAbility, castSA.copy(castSA.getHostCard(), true));
        sa.setTriggeringObject(AbilityKey.StackInstance, si);
```

Then Unbound Flourishing updates X for The HostCard and the StackInstance?
https://github.com/Card-Forge/forge/blob/master/forge-game/src/main/java/forge/game/ability/effects/ChangeXEffect.java

so: `AbilityKey.SpellAbility` shouldn't be touched, but `AbilityKey.Card` gets updated, right?

maybe that should be an LKI object too?

forge-game/src/main/java/forge/game/trigger/TriggerSpellAbilityCastOrCopy.java
```diff
@@ -265,8 +265,14 @@ public final void setTriggeringObjects(final SpellAbility sa, Map<AbilityKey, Ob
...
         final SpellAbilityStackInstance si = sa.getHostCard().getGame().getStack().getInstanceFromSpellAbility(castSA);
         final SpellAbility saForTargets = si != null ? si.getSpellAbility(true) : castSA;
-        sa.setTriggeringObject(AbilityKey.Card, castSA.getHostCard());
-        sa.setTriggeringObject(AbilityKey.SpellAbility, castSA.isWrapper() ? castSA : castSA.copy(castSA.getHostCard(), true));
+        final Card host = castSA.getHostCard();
+        sa.setTriggeringObject(AbilityKey.Card, host);
+        if (castSA.isSpell()) {
+            @SuppressWarnings("unchecked")
+            List<Card> lkiHosts = ((List<Card>) runParams.get(AbilityKey.CurrentCastSpells));
+            sa.setTriggeringObject(AbilityKey.CardLKI, lkiHosts.get(lkiHosts.lastIndexOf(host)));
```
Hanmac, line 273: the CardLKI could be set by MagicStack?

tool4ever: ok done

forge-game/src/main/java/forge/game/zone/MagicStack.java
```diff
@@ -462,10 +469,6 @@ private SpellAbilityStackInstance push(final SpellAbility sp) {
 
         GameActionUtil.checkStaticAfterPaying(sp.getHostCard());
```
Hanmac, line 470: you should check if the LKI needs to be done after this StaticCheck?

tool4ever: This is still the case because push() jumps back to add() and nothing interesting happens between

Though probably nothing would care about either order

tool4ever:
@Hanmac 
anything else needed?

Hanmac (approved):
👍

PR #3394, opened 2023-07-02
Basic AI logic for Banding / Bands with Other

Agetian:
Very basic logic, currently suboptimal in many cases, but should work in the typical scenarios.
Code contributions are welcome to improve it.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -533,7 +533,7 @@ public class AiAttackController {
             CardCollection attackers = combat.getAttackers();
             Card bestAttacker = ComputerUtilCard.getBestCreatureAI(attackers);
             for (Card c : bandingCreatures) {
-                Card bestBand = null;
+                Card bestBand;
 
                 if (c.hasKeyword("Bands with Other Legendary Creatures")) {
                     bestBand = ComputerUtilCard.getBestCreatureAI(CardLists.getType(attackers, "Legendary"));
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -357,35 +357,41 @@ public class AiAttackController {
     }
 
     // this checks to make sure that the computer player doesn't lose when the human player attacks
-    public final List<Card> notNeededAsBlockers(final List<Card> attackers) {
+    public final List<Card> notNeededAsBlockers(final List<Card> currentAttackers, final List<Card> potentialAttackers) {
         //check for time walks
         if (ai.getGame().getPhaseHandler().getNextTurn().equals(ai)) {
-            return attackers;
+            return potentialAttackers;
         }
         // no need to block (already holding mana to cast fog next turn)
         if (!AiCardMemory.isMemorySetEmpty(ai, AiCardMemory.MemorySet.CHOSEN_FOG_EFFECT)) {
             // Don't send the card that'll do the fog effect to attack, it's unsafe!
 
             List<Card> toRemove = Lists.newArrayList();
-            for (Card c : attackers) {
+            for (Card c : potentialAttackers) {
                 if (AiCardMemory.isRememberedCard(ai, c, AiCardMemory.MemorySet.CHOSEN_FOG_EFFECT)) {
                     toRemove.add(c);
                 }
             }
-            attackers.removeAll(toRemove);
-            return attackers;
+            potentialAttackers.removeAll(toRemove);
+            return potentialAttackers;
         }
 
+        if (ai.isCardInPlay("Masako the Humorless")) {
+            // "Tapped creatures you control can block as though they were untapped."
+            return potentialAttackers;
+        }
+
+        final CardCollection notNeededAsBlockers = new CardCollection(potentialAttackers);
+
         final List<Card> vigilantes = new ArrayList<>();
-        for (final Card c : myList) {
-            if (c.getName().equals("Masako the Humorless")) {
-                // "Tapped creatures you control can block as though they were untapped."
-                return attackers;
-            }
+        for (final Card c : Iterables.concat(currentAttackers, potentialAttackers)) {
             // no need to block if an effect is in play which untaps all creatures
             // (pseudo-Vigilance akin to Awakening or Prophet of Kruphix)
             if (c.hasKeyword(Keyword.VIGILANCE) || ComputerUtilCard.willUntap(ai, c)) {
                 vigilantes.add(c);
+            } else if (currentAttackers.contains(c)) {
+                // already attacking so can't block
+                notNeededAsBlockers.add(c);
             }
         }
         // reduce the search space
@@ -399,10 +405,8 @@ public class AiAttackController {
             }
         });
 
-        final CardCollection notNeededAsBlockers = new CardCollection(attackers);
-
         // don't hold back creatures that can't block any of the human creatures
-        final List<Card> blockers = getPossibleBlockers(attackers, opponentsAttackers, true);
+        final List<Card> blockers = getPossibleBlockers(potentialAttackers, opponentsAttackers, true);
 
         if (!blockers.isEmpty()) {
             notNeededAsBlockers.removeAll(blockers);
@@ -473,12 +477,15 @@ public class AiAttackController {
         // these creatures will be available to block anyway
         notNeededAsBlockers.addAll(vigilantes);
 
+        // remove those that were only included to ensure a full picture for the baseline
+        notNeededAsBlockers.removeAll(currentAttackers);
+
         // Increase the total number of blockers needed by 1 if Finest Hour in play
         // (human will get an extra first attack with a creature that untaps)
         // In addition, if the computer guesses it needs no blockers, make sure
         // that it won't be surprised by Exalted
         final int humanExaltedBonus = defendingOpponent.countExaltedBonus();
-        int blockersNeeded = attackers.size() - notNeededAsBlockers.size();
+        int blockersNeeded = potentialAttackers.size() - notNeededAsBlockers.size();
 
         if (humanExaltedBonus > 0) {
             final boolean finestHour = defendingOpponent.isCardInPlay("Finest Hour");
@@ -511,24 +518,28 @@ public class AiAttackController {
     public void reinforceWithBanding(final Player player, final Combat combat) {
         reinforceWithBanding(player, combat, null);
     }
-
     public void reinforceWithBanding(final Player player, final Combat combat, final Card test) {
+        CardCollection attackers = combat.getAttackers();
+        if (attackers.isEmpty()) {
+            return;
+        }
+
         List<String> bandsWithString = Arrays.asList("Bands with Other Legendary Creatures",
                 "Bands with Other Creatures named Wolves of the Hunt",
                 "Bands with Other Dinosaurs");
 
-        List<Card> bandingCreatures = new CardCollection();
-
+        List<Card> bandingCreatures = null;
         if (test == null) {
-            bandingCreatures.addAll(notNeededAsBlockers(player.getCreaturesInPlay()));
-            bandingCreatures = CardLists.filter(bandingCreatures, card -> card.hasKeyword(Keyword.BANDING) || card.hasAnyKeyword(bandsWithString));
+            bandingCreatures = CardLists.filter(myList, card -> card.hasKeyword(Keyword.BANDING) || card.hasAnyKeyword(bandsWithString));
 
             // filter out anything that can't legally attack or is already declared as an attacker
             bandingCreatures = CardLists.filter(bandingCreatures, card -> !combat.isAttacking(card) && CombatUtil.canAttack(card));
+
+            bandingCreatures = notNeededAsBlockers(attackers, bandingCreatures);
         } else {
             // Test a specific creature for Banding
             if (test.hasKeyword(Keyword.BANDING) || test.hasAnyKeyword(bandsWithString)) {
-                bandingCreatures.add(test);
+                bandingCreatures = new CardCollection(test);
             }
         }
 
@@ -539,12 +550,11 @@ public class AiAttackController {
             return;
         }
 
-        if (!bandingCreatures.isEmpty()) {
+        if (bandingCreatures != null) {
             List<String> evasionKeywords = Arrays.asList("Flying", "Horsemanship", "Shadow", "Plainswalk", "Islandwalk",
                     "Forestwalk", "Mountainwalk", "Swampwalk");
 
             // TODO: Assign to band with the best attacker for now, but needs better logic.
-            CardCollection attackers = combat.getAttackers();
             for (Card c : bandingCreatures) {
                 Card bestBand;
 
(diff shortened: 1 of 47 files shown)


forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -1304,6 +1304,57 @@ public void declareAttackers(Player attacker, Combat combat) {
         AiAttackController aiAtk = new AiAttackController(attacker); 
         lastAttackAggression = aiAtk.declareAttackers(combat);
 
+        // Check if we can reinforce with Banding creatures
+        if (!combat.getAttackers().isEmpty()) {
+            List<String> bandsWithString = Arrays.asList("Bands with Other Legendary Creatures",
+                    "Bands with Other Creatures named Wolves of the Hunt",
+                    "Bands with Other Dinosaurs");
+            List<Card> bandingCreatures = new CardCollection(aiAtk.notNeededAsBlockers(attacker.getCreaturesInPlay()));
```
tool4ever, line 1312: It's tempting to try and check all available creatures here so you can also catch cases where only a banded attack is acceptable
But this might lead to cheating because the banding creature could have Defender or something?

So probably need to start from the actually selected attackers and respect already made defender decisions
(that is if A attacks PW and B attacks opp don't reroute either into a band as that might just lead to an illegal constraint violation)

Agetian: Good call! I tried to implement a filter where the creatures that are already attacking and/or can't legally attack won't be routed into bands. The addAttacker call already ensures that the target of the banded creature is the same as the target of whatever it's being banded into, so the creatures in the same band shouldn't end up being assigned to different defenders, unless there's a loophole somewhere I don't see?..

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
```diff
@@ -2031,16 +2034,31 @@ public static boolean canDestroyBlocker(Player ai, Card blocker, Card attacker,
...
-    public static Map<Card, Integer> distributeAIDamage(final Card attacker, final CardCollectionView block, final CardCollectionView remaining, int dmgCanDeal, GameEntity defender, boolean overrideOrder) {
-        // TODO: Distribute defensive Damage (AI controls how damage is dealt to own cards) for Banding and Defensive Formation
+    public static Map<Card, Integer> distributeAIDamage(final Player self, final Card attacker, final CardCollectionView block, final CardCollectionView remaining, int dmgCanDeal, GameEntity defender, boolean overrideOrder) {
         Map<Card, Integer> damageMap = Maps.newHashMap();
         Combat combat = attacker.getGame().getCombat();
 
         boolean isAttacking = defender != null;
 
+        // Check for Banding, Defensive Formation
+        boolean isAttackingMe = isAttacking && combat.getDefenderPlayerByAttacker(attacker).equals(self);
```
tool4ever, line 2044: Defensive Formation is a bit different than banding
because it should only affect **your** attackers (not PW/battle)

Agetian:
Hmm, for Defensive Formation, do you know a good enough way to determine if the effect that allows us to distribute damage is a Defensive Formation? I guess something like: the card has no Banding keyword but the Formation static gives it the ability?

tool4ever:
I think the engine should already make sure it calls the functions on the right controller based on the combat declarations

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
```diff
@@ -2031,16 +2034,31 @@ public static boolean canDestroyBlocker(Player ai, Card blocker, Card attacker,
...
+        boolean isBlockingMyBand = false;
+        if (attacker.getController().isOpponentOf(self)) {
```
tool4ever, line 2046: isn't this just `AttackingBand.isValidBand`?

Hanmac:
i might do `Bands with Other` as better keyword, even if its a dead one

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
tool4ever: I think just checking that the attacker(blocker) controller isn't AI should be enough to determine that it should switch from "good" assignments to "bad"?

Hanmac: > I think just checking that the attacker(blocker) controller isn't AI should be enough to determine that it should switch from "good" assignments to "bad"?

attacker might be teammate of ai

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
```diff
@@ -2078,31 +2096,46 @@ public static Map<Card, Integer> distributeAIDamage(final Card attacker, final C
...
+                        damageMap.put(lastBlocker, dmgCanDeal + damageMap.get(lastBlocker));
+                    }
+                }
+            } else {
+                // In the event of Banding or Defensive Formation, assign max damage to the blocker who
+                // can tank all the damage or to the worst blocker to lose as little as possible
+                for (final Card b : block) {
+                    final int dmgToKill = getEnoughDamageToKill(b, dmgCanDeal, attacker, true);
+                    if (dmgToKill > dmgCanDeal) {
+                        damageMap.put(b, dmgToKill);
```
tool4ever, line 2133: I think you would need to put dmgCanDeal or you might assign more than legal?

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
Agetian: > isn't this just `AttackingBand.isValidBand`?

Pretty much, yeah, will convert to this call.

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
```diff
@@ -2044,14 +2045,8 @@ public static Map<Card, Integer> distributeAIDamage(final Player self, final Car
...
         if (attacker.getController().isOpponentOf(self)) {
-            List<String> bandsWithString = Arrays.asList("Bands with Other Legendary Creatures",
-                    "Bands with Other Creatures named Wolves of the Hunt",
-                    "Bands with Other Dinosaurs");
-            for (final Card c : block) {
-                if (c.hasKeyword(Keyword.BANDING) || c.hasAnyKeyword(bandsWithString)) {
-                    isBlockingMyBand = true;
-                    break;
-                }
+            if (AttackingBand.isValidBand((List<Card>)block, true)) {
```
Hanmac: I'm unsure if we can trust this cast or if we should make a new Array List?

Agetian: I was also thinking about it :) Debugging that place shows that it converts cleanly, but I dunno :) I guess let's make a new ArrayList instead. I was thinking if there's a cleaner/nicer way to get a list from a CardCollection, but couldn't come up with anything except for block.subList(0, block.size()) (which works too, I guess)

Hanmac: Or make isValidBand get a java Collection too , it shouldn't try to alter it anyway so it shouldn't be a problem?

Agetian: Yeah, good idea. Done :)

Agetian:
Alright, probably the best I can do with this mechanic for now, lol :) The rest can be tweaked in separate PRs when necessary.

tool4ever:
It's a great PR
I should have time for a full review tonight!

Agetian:
Sure thing! Thanks for your feedback and help, as always! :) Much appreciated!

forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
```diff
@@ -1406,6 +1406,39 @@ && doesCreatureAttackAI(ai, c)) {
...
+            if (phase.is(PhaseType.COMBAT_BEGIN) && phase.isPlayerTurn(ai)) {
+                // will this card participate in an attacking band?
+                Card bandingCard = getPumpedCreature(ai, sa, c, toughness, power, keywords);
+                // TODO: It may be possible to use AiController.getPredictedCombat here, but that makes it difficult to
+                // use reinforceWithBanding through the attack controller, especially with the extra card parameter in mind
+                AiAttackController aiAtk = new AiAttackController(ai);
+                Combat predicted = new Combat(ai);
+                aiAtk.declareAttackers(predicted);
+                aiAtk.reinforceWithBanding(ai, predicted, bandingCard);
+                if (predicted.isAttacking(bandingCard) && predicted.getBandOfAttacker(bandingCard).getAttackers().size() > 1) {
```
tool4ever, line 1419: Hmn, what if there is only attacker without banding -
Would the AI think it could then band with the virtual copy of itself?

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -511,6 +508,82 @@ public boolean apply(final Card c) {
...
+        } else {
+            // Test a specific creature for Banding
+            if (test.hasKeyword(Keyword.BANDING) || test.hasAnyKeyword(bandsWithString)) {
+                bandingCreatures.add(test);
+            }
+        }
+
+        // respect global attack constraints
+        GlobalAttackRestrictions restrict = GlobalAttackRestrictions.getGlobalRestrictions(ai, combat.getDefenders());
+        int attackMax = restrict.getMax();
```
tool4ever, line 537: if attackMax already reached bail out already

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -511,6 +508,82 @@ public boolean apply(final Card c) {
...
+
+    public void reinforceWithBanding(final Player player, final Combat combat, final Card test) {
+        List<String> bandsWithString = Arrays.asList("Bands with Other Legendary Creatures",
+                "Bands with Other Creatures named Wolves of the Hunt",
+                "Bands with Other Dinosaurs");
+
+        List<Card> bandingCreatures = new CardCollection();
+
+        if (test == null) {
+            bandingCreatures.addAll(notNeededAsBlockers(player.getCreaturesInPlay()));
```
tool4ever, line 523: Should reorder the calls here a bit by their computation costs:

- KW first
- switch isAttacking and canAttack
- `notNeededAsBlockers` probably last as most expensive

Agetian: Done :)
I also wanted to do notNeededAsBlockers last, but sadly, that method returns all the creatures that aren't needed as blockers, effectively negating any filtering...

forge-ai/src/main/java/forge/ai/AiAttackController.java
Agetian: Done :)

forge-ai/src/main/java/forge/ai/AiAttackController.java
tool4ever: The way it's currently called is probably problematic:
it expects to be passed the potential attackers BUT at this point any number of them are already declared and unavailable for blocking anyway
so AI could greatly overestimate its defensive power when it grants the banders

tool4ever: or maybe it will get filtered somewhere deeper
I'll investigate that as followup I think

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -511,6 +508,85 @@ public boolean apply(final Card c) {
...
+                }
+
+                Card bestAttacker = ComputerUtilCard.getBestCreatureAI(attackers);
+                if (c.hasKeyword("Bands with Other Legendary Creatures")) {
+                    bestBand = ComputerUtilCard.getBestCreatureAI(CardLists.getType(attackers, "Legendary"));
+                } else if (c.hasKeyword("Bands with Other Dinosaurs")) {
+                    bestBand = ComputerUtilCard.getBestCreatureAI(CardLists.getType(attackers, "Dinosaur"));
+                } else if (c.hasKeyword("Bands with Other Creatures named Wolves of the Hunt")) {
+                    bestBand = ComputerUtilCard.getBestCreatureAI(CardLists.filter(attackers, CardPredicates.nameEquals("Wolves of the Hunt")));
+                } else if (bestAttacker != null && !c.hasAnyKeyword(evasionKeywords) && bestAttacker.hasAnyKeyword(evasionKeywords)) {
```
tool4ever, line 563: When would bestAttacker be null - only if there were no attackers at all right?

So this should just be a guard clause at the top to avoid tons of calls 🤔

forge-ai/src/main/java/forge/ai/AiAttackController.java
tool4ever: Ok it's been a while since I wrote that function and it's rather complex but I think it should behave correctly now
Would need more time for sanity checks later though 😴

Agetian: Nice, thanks a lot for help here! I played a few games and things seemed normal, but feel free to test additionally as well when able :)

forge-ai/src/main/java/forge/ai/AiAttackController.java
Agetian: Yep, removed the null check since you added a guard above :+1:

Agetian:
Played a few more games, things seem pretty normal to me :)
The only other thing I wish I could have done is I wish I could teach the AI to see preemptively that it can attack and band something up to get higher power (currently it won't consider cumulative power when declaring attacks), but it seems really tricky, and I don't have a good enough idea at the moment for how (and if) to do that, hehe. So, I'll probably stop here feature-wise for this PR in particular (unless I come up with something all of a sudden again, lol).

Agetian:
Hmm, for some reason, the AI won't declare an attack with its Giver of Runes here though it seems profitable, maybe it's because of the changes to notNeededAsBlockers?.. Might need to investigate :/

[ai_declareattacks.txt](https://github.com/Card-Forge/forge/files/11954081/ai_declareattacks.txt)

tool4ever:
Funny I see you added three more commits after mine but only one is shown in this PR?

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -511,6 +515,88 @@ public boolean apply(final Card c) {
...
+        if (bandingCreatures != null) {
+            List<String> evasionKeywords = Arrays.asList("Flying", "Horsemanship", "Shadow", "Plainswalk", "Islandwalk",
+                    "Forestwalk", "Mountainwalk", "Swampwalk");
+
+            // TODO: Assign to band with the best attacker for now, but needs better logic.
+            for (Card c : bandingCreatures) {
+                Card bestBand;
+
+                if (c.getNetPower() <= 0) {
+                    // Don't band a zero power creature if there's already a banding creature in a band
```
tool4ever, line 562: I think I still see a corner case here:
- if you band creature A
- next iteration you have B with power = 0
- if no other attacker you will now filter it empty
- now could crash again?

Agetian: Ah, indeed, fair enough :) I restored the NPE guard for bestAttacker

tool4ever (approved):
👏

PR #3399, opened 2023-07-04
Update DeckImport.java

kevlahnota:
fix importing deck with accented characters


Hanmac:
how do we handle if the card has accents in the card script like `Lim-Dûl` but the user can't input the character? should we try find the card name by guessing?

i think yours is only in the direction if the user-input has accents but the card has not?

kevlahnota:
> 

I think all card names are normalized in the database. if the textfield has changed listener, maybe it could replace the character upon input

Hanmac:
For this i probably should make another MR that makes the CardNames be normalized when reading the card scripts into CardDb so the cardnames can have `û` again

Hanmac:
i probably need to do a different approach when importing the Card Files into the CardDb

PR #3403, opened 2023-07-04
fix elephant and dragonfly tags

schnautzr:
* Fix elephant's color identity to show as GW instead of W.
* Fix dragonfly's hunting hints to say it can be found in the blue biome (not the black, green, or colorless biomes).


Simisays:
👍

kevlahnota (approved):

PR #3405, opened 2023-07-04
flag cards AI is consistently bad at.

schnautzr:
Elderscale Wurm, Grothama, Tempting Wurm, and Endless Swarm are all consistently problematic for AI. Adding flags with comments.

### DIFF
--- a/forge-gui/res/cardsfolder/e/elderscale_wurm.txt
+++ b/forge-gui/res/cardsfolder/e/elderscale_wurm.txt
@@ -9,4 +9,6 @@ R:Event$ LifeReduced | ActiveZones$ Battlefield | ValidPlayer$ You | Result$ LT7
 SVar:ReduceLoss:DB$ ReplaceEffect | VarName$ Result | VarValue$ X
 SVar:X:ReplaceCount$Result/NMinus.7
 SVar:ElderscaleCondition:Count$YourLifeTotal
+#NOTE: If the opponent controls the wurm, AI won't take into consideration that combat damage may be ineffective.
+AI:RemoveDeck:All
 Oracle:Trample\nWhen Elderscale Wurm enters the battlefield, if your life total is less than 7, your life total becomes 7.\nAs long as you have 7 or more life, damage that would reduce your life total to less than 7 reduces it to 7 instead.
--- a/forge-gui/res/cardsfolder/e/endless_swarm.txt
+++ b/forge-gui/res/cardsfolder/e/endless_swarm.txt
@@ -6,4 +6,6 @@ A:SP$ Token | Cost$ 5 G G G | TokenAmount$ X | TokenScript$ g_1_1_snake | TokenO
 SVar:X:Count$InYourHand
 #NOTE: The AI will not stop making land drops after casting this spell (so, random decks may also utilize manlands and other cards with activation cost already on the battlefield to the maximum effect)
 AI:RemoveDeck:Random
+#NOTE: Memory leak.
+AI:RemoveDeck:All
 Oracle:Create a 1/1 green Snake creature token for each card in your hand.\nEpic (For the rest of the game, you can't cast spells. At the beginning of each of your upkeeps, copy this spell except for its epic ability.)
--- a/forge-gui/res/cardsfolder/g/grothama_all_devouring.txt
+++ b/forge-gui/res/cardsfolder/g/grothama_all_devouring.txt
@@ -10,4 +10,6 @@ SVar:TrigRepeat:DB$ RepeatEach | RepeatPlayers$ Player | RepeatSubAbility$ TrigD
 SVar:TrigDraw:DB$ Draw | Defined$ Remembered | NumCards$ X
 SVar:X:TriggeredCard$TotalDamageThisTurn Card.RememberedPlayerCtrl
 SVar:HasAttackEffect:TRUE
+#NOTE: AI will always choose to activate the "fight Grothama" ability if one of its creatures attacks.
+AI:RemoveDeck:All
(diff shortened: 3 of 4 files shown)


forge-gui/res/cardsfolder/t/tempting_wurm.txt
```diff
@@ -6,4 +6,6 @@ T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.S
 SVar:EachOpponent:DB$ RepeatEach | RepeatPlayers$ Player.Opponent | RepeatSubAbility$ TemptingChange
 SVar:TemptingChange:DB$ ChangeZone | Origin$ Hand | Destination$ Battlefield | ChangeType$ Artifact,Creature,Enchantment,Land | DefinedPlayer$ Remembered | ChangeNum$ X
 SVar:X:Count$ValidHand Artifact.RememberedPlayerCtrl,Creature.RememberedPlayerCtrl,Enchantment.RememberedPlayerCtrl,Land.RememberedPlayerCtrl
+#NOTE: AI will play Tempting Wurm on turn two when the opponent still has a full hand.
+AI:RemoveDeck:All
```
Agetian, line 10: Better turned into a NeedsToPlay/NeedsToPlayVar hint so that this is not an issue.

forge-gui/res/cardsfolder/e/endless_swarm.txt
```diff
@@ -6,4 +6,6 @@ A:SP$ Token | Cost$ 5 G G G | TokenAmount$ X | TokenScript$ g_1_1_snake | TokenO
 SVar:X:Count$InYourHand
 #NOTE: The AI will not stop making land drops after casting this spell (so, random decks may also utilize manlands and other cards with activation cost already on the battlefield to the maximum effect)
 AI:RemoveDeck:Random
+#NOTE: Memory leak.
```
Agetian, line 9: What exactly is meant here? What happens exactly?

forge-gui/res/cardsfolder/e/elderscale_wurm.txt
```diff
@@ -9,4 +9,6 @@ R:Event$ LifeReduced | ActiveZones$ Battlefield | ValidPlayer$ You | Result$ LT7
 SVar:ReduceLoss:DB$ ReplaceEffect | VarName$ Result | VarValue$ X
 SVar:X:ReplaceCount$Result/NMinus.7
 SVar:ElderscaleCondition:Count$YourLifeTotal
+#NOTE: If the opponent controls the wurm, AI won't take into consideration that combat damage may be ineffective.
```
Agetian, line 12: AI:RemoveDeck:All flag is only for the AI's own decks, not when the opponent has one in his deck. This won't stop human players from having it and abusing it vs. the AI. Thus, the flag isn't the best here. If there's a concern that the AI misplays it or plays it at the wrong time, then other hints may be added as needed.

forge-gui/res/cardsfolder/g/grothama_all_devouring.txt
```diff
@@ -10,4 +10,6 @@ SVar:TrigRepeat:DB$ RepeatEach | RepeatPlayers$ Player | RepeatSubAbility$ TrigD
 SVar:TrigDraw:DB$ Draw | Defined$ Remembered | NumCards$ X
 SVar:X:TriggeredCard$TotalDamageThisTurn Card.RememberedPlayerCtrl
 SVar:HasAttackEffect:TRUE
+#NOTE: AI will always choose to activate the "fight Grothama" ability if one of its creatures attacks.
+AI:RemoveDeck:All
```
Agetian, line 14: This is a pretty universal issue, not only for the AI deck - if the human plays one from his own deck, it'll also be very abuse-worthy. I'll think about how to improve the AI handling of this card so that it doesn't work as badly as it does right now :)

forge-gui/res/cardsfolder/e/endless_swarm.txt
Agetian: Tested it and it seems like it worked correctly for me. It does produce a lot of tokens so low-memory systems may run out of memory, yes, but it's not necessarily a problem with the AI handling of the card. Increasing the Java memory heap size may help on systems that have a lot of RAM. Not sure if maybe you meant something else though?

Agetian:
I'm going to look at these on a case by case basis after I complete my Banding AI project. Please keep this PR open for now but don't merge. I feel like these deserve a better solution than a Remove flag for the AI, especially considering that most of these issues aren't exclusive to the AI player having these cards in a deck.

forge-gui/res/cardsfolder/e/endless_swarm.txt
schnautzr: image

The only abilities the opponent had here were mana abilities, and it was slowing down every single step, not just the combat steps. That tells me the AI was *considering* what to do when it had *nothing* it could do.

forge-gui/res/cardsfolder/e/elderscale_wurm.txt
schnautzr: Fair enough.

forge-gui/res/cardsfolder/t/tempting_wurm.txt
schnautzr: Yeah, I don't know how to program the AI logic; I'll let you do your magic here. ;)

schnautzr:
> Please keep this PR open for now but don't merge.

Would it make more sense to open four issues and assign you to them?

Agetian:
Yeah, feel free to do that, sure :)

schnautzr:
Opened issues #3406 #3407 #3408 #3410

PR #3424, opened 2023-07-07
AI shouldn't miss Grothama's triggers even when it doesn't want to activate them

Agetian:


forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
```diff
@@ -256,10 +256,15 @@ public boolean confirmStaticApplication(Card hostCard, GameEntity affected, Stri
     @Override
     public boolean confirmTrigger(WrappedAbility wrapper) {
         final SpellAbility sa = wrapper.getWrappedAbility();
+        final String aiLogic = sa.getParamOrDefault("AILogic", "");
         //final Trigger regtrig = wrapper.getTrigger();
+
         if (ComputerUtilAbility.getAbilitySourceName(sa).equals("Deathmist Raptor")) {
             return true;
+        } else if (aiLogic.equals("Grothama")) {
```
tool4ever, line 264: I don't think we should "pollute" PlayerControllerAi like that:
- like all other API `FightAi.doTriggerAINoCost` on the first call (mandatory=true) should just always return true if it's not targeted
- if it's non-mandatory there will be a second call from here anyway which then should lead to the consider

Agetian: Good call! I wasn't aware that it'll call the trigger routine the second time around :)

tool4ever (approved):

PR #3426, opened 2023-07-07
Fix Praetor's Grasp

tool4ever:


Agetian (approved):

Hanmac (approved):
👍 

for Deathmist Raptor i might do a more generic approach or more AI

PR #3438, opened 2023-07-09
Basic AI for Chalice of the Void, minor DamageDealAi improvement

### LINKED ISSUE #3442: AI vs opponent with Elderscale Wurm
When an opponent sits at 7 life with an Elderscale Wurm, the AI activates Ember Hauler's ability to deal 2 damage to the opponent, which has no effect.

Agetian:
Not the smartest thing in the world, but better than nothing.
Can be improved if the AI looks at the library/hand contents instead of all cards (assuming that it knows the deck lists *and* looks at the graveyard and exile for what already left the game), but it's a bit more like outright cheating, so I don't know, lol :)
I'm not sure if the AI should really go above X=2 or X=3 for this card or so, it's usually played for low X in specific decks against specific (usually low-cost) threats, so... Not sure too, haha.

Also fixes #3442, improving damage prediction when deciding to deal damage to the opponent with AF DealDamage.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/DamageDealAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/DamageDealAi.java
@@ -735,7 +735,7 @@ public class DamageDealAi extends DamageAiBase {
                         || (SpellAbilityAi.isSorcerySpeed(sa, ai) && phase.is(PhaseType.MAIN2))
                         || immediately) {
                     boolean pingAfterAttack = "PingAfterAttack".equals(logic) && phase.getPhase().isAfter(PhaseType.COMBAT_DECLARE_ATTACKERS) && phase.isPlayerTurn(ai);
-                    if (pingAfterAttack || shouldTgtP(ai, sa, dmg, noPrevention)) {
+                    if ((pingAfterAttack && !avoidTargetP(ai, sa)) || shouldTgtP(ai, sa, dmg, noPrevention)) {
                         tcs.add(enemy);
                         if (divided) {
                             sa.addDividedAllocation(enemy, dmg);
--- a/forge-ai/src/main/java/forge/ai/ability/DamageDealAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/DamageDealAi.java
@@ -630,7 +630,7 @@ public class DamageDealAi extends DamageAiBase {
             if (tgt.canTgtPlaneswalker()) {
                 // We can damage planeswalkers with this, consider targeting.
                 Card c = dealDamageChooseTgtPW(ai, sa, dmg, noPrevention, enemy, false);
-                if (c != null && !shouldTgtP(ai, sa, dmg, noPrevention, true)) {
+                if (c != null && !shouldTgtP(ai, sa, dmg, noPrevention)) {
                     tcs.add(c);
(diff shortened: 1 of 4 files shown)


forge-ai/src/main/java/forge/ai/ability/DamageDealAi.java
```diff
@@ -746,17 +746,17 @@ private boolean damageChoosingTargets(final Player ai, final SpellAbility sa, fi
...
-                        || ("PingAfterAttack".equals(logic) && phase.getPhase().isAfter(PhaseType.COMBAT_DECLARE_ATTACKERS) && phase.isPlayerTurn(ai))
-                        || immediately || shouldTgtP(ai, sa, dmg, noPrevention)) &&
-                        (!avoidTargetP(ai, sa))) {
-                    tcs.add(enemy);
-                    if (divided) {
-                        sa.addDividedAllocation(enemy, dmg);
-                        break;
+                        || immediately) {
+                    boolean pingAfterAttack = "PingAfterAttack".equals(logic) && phase.getPhase().isAfter(PhaseType.COMBAT_DECLARE_ATTACKERS) && phase.isPlayerTurn(ai);
+                    if ((pingAfterAttack || shouldTgtP(ai, sa, dmg, noPrevention)) && !avoidTargetP(ai,sa)) {
```
tool4ever, line 754: should be rewritten slightly:
 - shouldTgtP already has its own call to avoidTargetP
 - it calls the "bad" version with PW redirection which could lead to wrong result, probably better to get rid of that already

Agetian: Done :) Yeah, Planeswalker redirection hasn't been a thing for a while, I stripped it completely from the AI code

forge-ai/src/main/java/forge/ai/ability/DamageDealAi.java
```diff
@@ -746,17 +730,17 @@ private boolean damageChoosingTargets(final Player ai, final SpellAbility sa, fi
...
-                        || ("PingAfterAttack".equals(logic) && phase.getPhase().isAfter(PhaseType.COMBAT_DECLARE_ATTACKERS) && phase.isPlayerTurn(ai))
-                        || immediately || shouldTgtP(ai, sa, dmg, noPrevention)) &&
-                        (!avoidTargetP(ai, sa))) {
-                    tcs.add(enemy);
-                    if (divided) {
-                        sa.addDividedAllocation(enemy, dmg);
-                        break;
+                        || immediately) {
+                    boolean pingAfterAttack = "PingAfterAttack".equals(logic) && phase.getPhase().isAfter(PhaseType.COMBAT_DECLARE_ATTACKERS) && phase.isPlayerTurn(ai);
+                    if (pingAfterAttack || shouldTgtP(ai, sa, dmg, noPrevention)) {
```
tool4ever, line 738: logic should be:
`((pingAfterAttack && !avoidTargetP(ai, sa)) || shouldTgtP(ai, sa, dmg, noPrevention))`

Agetian: Good catch! :) Done

tool4ever (approved):

PR #3444, opened 2023-07-10
change default card back for mobile to MtG official card back, move purple mana version to first empty slot.

schnautzr:
No reason for "Magic the Gathering Six-Color Card Back", © 2016 [LordNyriox](https://www.deviantart.com/lordnyriox/gallery) to be default. Moving it to the first empty sleeve slot and adding the correct default card back.


Agetian:
:+1:

Agetian (approved):

PR #3469, opened 2023-07-14
fix CardView NPE

kevlahnota:


forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1431,7 +1431,12 @@ public Object vote(final SpellAbility sa, final String prompt, final List<Object
     public boolean confirmReplacementEffect(final ReplacementEffect replacementEffect, final SpellAbility effectSA,
                                             GameEntity affected, final String question) {
         if (GuiBase.getInterface().isLibgdxPort()) {
-            return this.getGui().confirm(effectSA.getView().getHostCard(), question.replaceAll("\n", " "));
+            CardView cardView;
+            if (effectSA.getView() != null) //updated view
+                cardView = effectSA.getView().getHostCard();
+            else //fallback
+                cardView = effectSA.getCardView();
```
Hanmac, line 1438: could this be part of the SpellAbility method?
like getCardView is always trying to do that?

PR #3470, opened 2023-07-14
InputSelectTargets: able to remove Target

### LINKED ISSUE #3420: Feature Request: The ability to deselect a spell's/trigger's targets
I don't know how hard this would be, or maybe it already exists and I'm just stupid, but if it doesn't already exist, would it be at all possible to add to Forge the ability to "deselect" a target when you're selecting targets? There have been plenty of times where I either clicked a creature, or some other permanent/player by accident or just changed my mind, but couldn't remove that selection. If I clicked it again it would just add the thing to the targeted list once more, whereas normally one would assume that clicking it another time would remove it from the targets list.

I feel like the list of targets shouldn't be final until I click "ok."

As an example I just had an instance where Lathiel, the Bounteous Dawn triggered and I was selecting the creatures. I initially clicked one of my creatures, but then changed my mind and didn't want to distribute counters to that particular target, but alas I was not allowed to change my mind. Forge said: "*You made your bed, now lie in it.*"

Hanmac:
Closes #3420

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/match/input/InputSelectTargets.java
+++ b/forge-gui/src/main/java/forge/gamemodes/match/input/InputSelectTargets.java
@@ -2,13 +2,14 @@ package forge.gamemodes.match.input;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
+
 import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
+import java.util.Set;
 
 import com.google.common.base.Predicate;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
 
 import forge.game.GameEntity;
 import forge.game.GameObject;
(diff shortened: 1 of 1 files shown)


forge-gui/src/main/java/forge/gamemodes/match/input/InputSelectTargets.java
```diff
@@ -360,8 +360,7 @@ private void addTarget(final GameEntity ge) {
         else if (ge instanceof Player) {
             getController().getGui().setHighlighted(PlayerView.get((Player) ge), true);
         }
-        final Integer val = targetDepth.get(ge);
-        targetDepth.put(ge, val == null ? Integer.valueOf(1) : Integer.valueOf(val.intValue() + 1) );
+        targets.add(ge);
 
         if (hasAllTargets()) {
```
kevlahnota, line 365: can the hasAllTargets() check popup a dialog to ask human player if the target is final, if not remove the selected target otherwise select the target and finish...

Hanmac: om i have no idea to how code that right now? you are free to extend this MR if you want

kevlahnota: It can be added in a later MR since the additional confirmation is only needed if the sa can be paid for free (since you can cancel if theres inputpayment)

Agetian (approved):

PR #3473, opened 2023-07-15
CMB1 land tokens fix

schnautzr:
- Forest land tokens created by Generated Horizons have "{T}: Add {G}" per rule 305.6.
- Tokens created by Waste Land are land tokens with the land subtype Wastes, which doesn't appear anywhere else yet.

### DIFF
--- a/forge-gui/res/tokenscripts/c_l_forest.txt
+++ b/forge-gui/res/tokenscripts/c_l_forest.txt
@@ -1,4 +1,6 @@
 Name:Forest Token
 ManaCost:no cost
 Types:Land Forest
-Oracle:
+A:AB$ Mana | Cost$ T | Produced$ G | SpellDescription$ Add {G}.
+DeckHas:Ability$Mana.Green
+Oracle:{T}: Add {G}.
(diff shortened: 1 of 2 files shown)


Hanmac:
@schnautzr "intrinsic ability" for the forest token means the ability is generated by the basic land type

schnautzr:
See rulings for dual lands like Plateau. It has basic land types, but it's not a basic land. Simply having land types that are defined as basic land types gives it the intrinsic abilities.

Hanmac:
> See rulings for dual lands like Plateau. It has basic land types, but it's not a basic land. Simply having land types that are defined as basic land types gives it the intrinsic abilities.

the intrinsic abilities are done by the CardType you don't need to define them via `A:`

schnautzr:
Ah, misunderstood you then. I'll check and see if this is already happening when I get back to my computer.

Hanmac:
as for the Wastes token, Wastes isn't a Land Subtype

schnautzr:
It says "Wastes land token", not "land token named 'Wastes'".

Hanmac:
thats why Testcards are so problematic :P

schnautzr:
Yeah it needs an Oracle revision!

PR #3490, opened 2023-07-19
Remove special case of StackInstances sharing the same SpellAbility object

### LINKED ISSUE #2219: Lithoform Engine can't copy its own ability

tool4ever:
- Closes #2219
- I think we no longer need extra LKI in _TriggerSpellAbilityCastOrCopy_
  - this means if there's a trigger on the stack to copy `TriggeredSpellAbility`
  - if we now first change the targets of that with another effect
  - when the trigger resolves the copy will start with the updated targets
- > If you spent both {G} and {W} to cast Dawnglow Infusion, you’ll gain (X + X) life. An ability that triggers whenever you gain life will trigger just once.
- I probably plan to wait for one or two snapshots feedback before cleaning up `SpellAbilityStackInstance` as follow-up


forge-gui/res/cardsfolder/d/dawnglow_infusion.txt
```diff
@@ -1,7 +1,9 @@
 Name:Dawnglow Infusion
 ManaCost:X GW
 Types:Sorcery
-A:SP$ GainLife | Cost$ X GW | LifeAmount$ X | ConditionManaSpent$ G | SubAbility$ WPaid | SpellDescription$ You gain X life if {G} was spent to cast this spell and X life if {W} was spent to cast this spell. (Do both if {G}{W} was spent.)
-SVar:WPaid:DB$ GainLife | LifeAmount$ X | ConditionManaSpent$ W
+A:SP$ GainLife | Cost$ X GW | LifeAmount$ Y | SpellDescription$ You gain X life if {G} was spent to cast this spell and X life if {W} was spent to cast this spell. (Do both if {G}{W} was spent.)
```
Hanmac, line 4: @Agetian when Condition Mana Spent is removed, then it probably needs some AI logic?

Hanmac:
do your stuff still works with Count Activations and Count Resolves for Activated abilities?

forge-gui/res/cardsfolder/d/dawnglow_infusion.txt
Agetian: Most likely, yes, but I don't remember where exactly off the top of my head... (still out of town, so can't take a look, sadly)

forge-gui/res/cardsfolder/d/dawnglow_infusion.txt
```diff
@@ -1,7 +1,9 @@
...
 SVar:X:Count$xPaid
+SVar:Y:SVar$W/Plus.G
+SVar:W:Count$Adamant_1.White.X.0
+SVar:G:Count$Adamant_1.Green.X.0
```
Hanmac, line 8: @tool4ever wrong words

Adamant is affected by Color TextChange,
these use the Mana Symbols and should not be affected by TextChange

tool4ever: Ugh right

Hanmac:
one thing we need to think about is if we still need the StackInstances, but that can be stuff for later

Hanmac:
the use of OriginalAbility might be wrong when something uses "gain Activated Abilities" like from Quicksilver Elemental

then each gained ability should count differently?

forge-gui/res/cardsfolder/d/dawnglow_infusion.txt
tool4ever: Actually AI would not even play the old script, fixed now
Not sure teaching it more is worth it at this point

Agetian: Most likely not, but even if needed, we can do it as a followup PR later :)

Agetian: Thanks for looking into this!

tool4ever:
> the use of OriginalAbility might be wrong when something uses "gain Activated Abilities" like from Quicksilver Elemental
> 
> then each gained ability should count differently?

In that case they will have different `grantorStatic` as columnKey

Hanmac:
for Drawglow Infusion, one of the AI things, would be nice if you can get it to pay G and W if possible

also such spells should be marked as `not copyable` because they have no benefit

Hanmac:
> > the use of OriginalAbility might be wrong when something uses "gain Activated Abilities" like from Quicksilver Elemental
> > then each gained ability should count differently?
> 
> In that case they will have different `grantorStatic` as columnKey

I want to refactor `grantorStatic` anyway, but that can be done later, i think that had something to do with mayPlay or something

Agetian:
> for Drawglow Infusion, one of the AI things, would be nice if you can get it to pay G and W if possible
> 
> also such spells should be marked as `not copyable` because they have no benefit

Yes, the AI used to detect this in CopySpellAbilityAi via the presence of the now removed ConditionManaSpent param. Maybe it now needs a special AI flag?

tool4ever:
Hmn, I will need to think about if it should really use `canPlayFromEffectAI` to continue checking in the future...
maybe some `checkApiLogic` would be smart enough already to decline the manaless copy

For now I made a small tweak so it doesn't try to play it as mandatory trigger 🤖 
image
(my dmg spell, their copy spell, no other targets)

Hanmac (approved):

PR #3502, opened 2023-07-21
Card: Wither damage as property, use Static for Everlasting Torment

### LINKED ISSUE #3491: Damage as Wither
Everlasting Torment has:
> All damage is dealt as though its source had wither.

Currently it is done as Global Rule, but i think it would be better to use StaticAbility for this (which might be more extendable)

Asking Judges, it does not follow LKI, so for the following Scenario:

- Creature activates ability that would deal damage while Everlasting Torment is on the field
- while Everlasting Torment continues to be on the field, the Creature is removed while its ability is still on the stack
- then Everlasting Torment is removed too
- now, when the damage is dealt, it is not affected by the Static Ability of Everlasting Torment anymore

what probably can't be done is, make Wither and Infect also use the ability internally, because Wither and Infect should still use LKI

- Creature enchanted by Fists of the Demigod (it gets +1/+1 and has wither) activates its ability
- while Ability on the stack, the Creature is removed
- the Ability uses LKI and the Creature still had Wither and so does Wither damage

Hanmac:
Closes #3491 

it also allows for a more generic `isWitherDamage`

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -5820,7 +5820,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
             subtractCounter(CounterType.get(CounterEnumType.DEFENSE), damageIn, true);
         }
         if (isCreature()) {
-            if (isWitherDamage()) { // 120.3d
+            if (source.isWitherDamage()) { // 120.3d
                 addCounter(CounterEnumType.M1M1, damageIn, source.getController(), counterTable);
                 damageType = DamageType.M1M1Counters;
             }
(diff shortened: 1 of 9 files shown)


Hanmac:
@tool4ever @Agetian can someone debug the failed test? i didn't touched anything there?

Agetian:
Sadly, can't help at the moment, still don't have access to my dev PC (out of town)...

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -5821,10 +5820,7 @@ public final int addDamageAfterPrevention(final int damageIn, final Card source,
             subtractCounter(CounterType.get(CounterEnumType.DEFENSE), damageIn, true);
         }
         if (isCreature()) {
-            boolean wither = game.getStaticEffects().getGlobalRuleChange(GlobalRuleChange.alwaysWither)
-                    || source.hasKeyword(Keyword.WITHER) || source.hasKeyword(Keyword.INFECT);
-
-            if (wither) { // 120.3d
+            if (isWitherDamage()) { // 120.3d
```
tool4ever, line 5823: **source**.isWitherDamage()

tool4ever (approved):

PR #3515, opened 2023-07-22
Card: store created CardTraits by Text Change: Volrath’s Shapeshifter

### LINKED ISSUE #3414: Phyrexian Devourer cloned via Volrath's Shapeshifter causes infinite Sacrifice triggers
When Phyrexian Devourer reaches >=7 power, Forge stacks infinite instances of the sacrifice trigger.

According to the current Oracle rulings, the sacrifice ability only triggers once.

The current handling of this card makes it impossible to execute any devourer combos within Forge.  This is a fairly popular win condition in Premodern and it would be nice to actually be able to resolve it in this program.

Hanmac:
Closes  #3414

@tool4ever can you test this?

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -4560,7 +4560,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
         Trigger result = storedTriggerByText.get(stAb, tr);
         if (result == null) {
             result = tr.copy(this, false);
-            result.setIntrinsic(false); // needs to be changed by CardTextChanges
+            result.setIntrinsic(true); // needs to be changed by CardTextChanges
             storedTriggerByText.put(stAb, tr, result);
         }
         return result;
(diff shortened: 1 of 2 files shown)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4525,6 +4533,71 @@ public final StaticAbility getStaticAbilityForStaticAbility(final String str, fi
...
+            storedTriggerByText.put(stAb, tr, result);
+        }
+        return result;
+    }
+
+    public final ReplacementEffect getReplacementEffectForStaticAbilityByText(final ReplacementEffect re, final StaticAbility stAb) {
+        ReplacementEffect result = storedReplacementEffectByText.get(stAb, re);
+        if (result == null) {
+            result = re.copy(this, false);
+            result.setIntrinsic(false); // needs to be changed by CardTextChanges
```
tool4ever, line 4573: why are these now set to false? this breaks it but the comment says it should work?

tool4ever (approved):

PR #3524, opened 2023-07-24
remove translation

kevlahnota:
-should fix ai effect logic
-translation should be done somewhere


forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
```diff
@@ -498,28 +498,7 @@ protected static Card createEffect(final SpellAbility sa, final Player controlle
         final Game game = hostCard.getGame();
         final Card eff = new Card(game.nextCardId(), game);
         String finalname = name.replaceAll("\\([^()]*\\)", "");
```
Hanmac, line 500: should this be removed too?

kevlahnota: I dont remember why I add that before since it removes enclosed in (), is there such card name like that?

Hanmac: some UN Cards use "(" https://scryfall.com/search?q=%22%28%22

PR #3540, opened 2023-07-27
Remove some obsolete code

tool4ever:
First cleanup run for StackInstances.

OptionalCosts don't need to be stored in the Card and will just be checked via castSA.

### DIFF
--- a/forge-gui/res/cardsfolder/o/of_one_mind.txt
+++ b/forge-gui/res/cardsfolder/o/of_one_mind.txt
@@ -1,7 +1,7 @@
 Name:Of One Mind
 ManaCost:2 U
 Types:Sorcery
-S:Mode$ ReduceCost | ValidCard$ Card.Self | Type$ Spell | Amount$ 2 | EffectZone$ All | IsPresent$ Creature.Human+YouCtrl | IsPresent2$ Creature.nonHuman+YouCtrl | Description$ CARDNAME costs {2} less to cast if you control a Human creature and a non-Human creature.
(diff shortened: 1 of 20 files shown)


Hanmac:
does that still work when you copy a kicked spell for example?

tool4ever:
yes, castSA gets copied but is otherwise left alone (no clearing or anything when you counter the original etc.)

Hanmac:
important is that manaSpent should not be referenced by CastSA because these are not copied

forge-gui/res/cardsfolder/o/of_one_mind.txt
```diff
@@ -1,7 +1,7 @@
 Name:Of One Mind
 ManaCost:2 U
 Types:Sorcery
-S:Mode$ ReduceCost | ValidCard$ Card.Self | Type$ Spell | Amount$ 2 | EffectZone$ All | IsPresent$ Creature.Human+YouCtrl | IsPresent2$ Creature.nonHuman+YouCtrl | Description$ CARDNAME costs {2} less to cast if you control a Human creature and a non-Human creature.
+S:Mode$ ReduceCost | ValidCard$ Card.Self | Type$ Spell | Amount$ 2 | EffectZone$ All | IsPresent$ Creature.Human+YouCtrl | CheckSVar$ Count$Valid Creature.nonHuman+YouCtrl | Description$ CARDNAME costs {2} less to cast if you control a Human creature and a non-Human creature.
```
Hanmac, line 4: Why Change this?

tool4ever: fixes the card for now because `StaticAbility.checkConditions` does not support IsPresent2
we could probably try and merge it into CardTraitBase checks but that is a bigger refactor

tool4ever:
> important is that manaSpent should not be referenced by CastSA because these are not copied

that's already done correctly

Hanmac:
> > important is that manaSpent should not be referenced by CastSA because these are not copied
> 
> that's already done correctly

yeah just the reminder that it needs to be kept that way

PR #3549, opened 2023-07-28
Clean up train continues

### LINKED ISSUE #3548: Lae'zel, Githyanki Warrior (HBG) self-added ability only lasts until end of turn
Current snapshot, Windows. Lae'zel says 
> When you cast this spell, it gains "When this creature becomes the target of a spell or ability an opponent controls, you may exile this creature, then return it to the battlefield under its owner's control."

This triggers and works correctly, except the ability is only gained for that turn, and goes away after.

tool4ever:
Notably with `addAndUnfreeze` we no longer work with a throw-away StackInstance for the frozenStack only to recreate it again when unfreezing.

Closes #3548

### DIFF
--- a/forge-game/src/main/java/forge/game/spellability/SpellAbilityStackInstance.java
+++ b/forge-game/src/main/java/forge/game/spellability/SpellAbilityStackInstance.java
@@ -20,7 +20,6 @@ package forge.game.spellability;
 import java.util.Map;
 import java.util.Set;
 
-import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 
 import forge.game.GameObject;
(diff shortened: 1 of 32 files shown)


forge-game/src/main/java/forge/game/spellability/SpellAbilityStackInstance.java
```diff
@@ -172,33 +112,12 @@ public int getId() {
     //      Perhaps lets move the refresh logic to a separate function called only when necessary
     public final SpellAbility getSpellAbility(boolean refresh) {
         if (refresh) {
-            ability.setTargets(tc);
             ability.setActivatingPlayer(activatingPlayer);
```
Hanmac, line 115: is that the only thing getSpellAbility(true) does now? reset the Activating Player?

tool4ever: yep! In theory it can be ripped out completely now, I will check

Agetian (approved):

kevlahnota (approved):

PR #3550, opened 2023-07-28
TriggerVote improve with new cards

Northmoc:
Should all of these (including Grudge Keeper) be RemoveDeck:Random?

Thought it would be nice to add a record of voting to game Log
image


forge-game/src/main/java/forge/game/trigger/TriggerVote.java
```diff
@@ -66,29 +66,54 @@ public final boolean performTest(final Map<AbilityKey, Object> runParams) {
...
+                true, true
+        );
+        sa.setTriggeringObject(AbilityKey.OpponentVotedSame, oppVotedSame);
     }
 
     @Override
     public String getImportantStackObjects(SpellAbility sa) {
         StringBuilder sb = new StringBuilder();
-        sb.append(Localizer.getInstance().getMessage("lblVoters")).append(": ").append(sa.getTriggeringObject(AbilityKey.OtherVoters));
+        if (hasParam("List")) {
```
kevlahnota, line 87: if the sa don't have List$ param defined, it's ok to return empty string? Cards like Capital Punishment, Coercive Portal?

forge-game/src/main/java/forge/game/trigger/TriggerVote.java
```diff
@@ -66,29 +66,54 @@ public final boolean performTest(final Map<AbilityKey, Object> runParams) {
...
 
     private static FCollection<Player> getVoters(final Player player,
             final ListMultimap<Object, Player> votes,
-            final boolean isOpponent, final boolean votedOtherchoice) {
+            final boolean isOpponent, final boolean votedSame) {
         final FCollection<Player> voters = new FCollection<>();
         for (final Object voteType : votes.keySet()) {
             final List<Player> players = votes.get(voteType);
-            if (votedOtherchoice ^ players.contains(player)) {
+            if (!votedSame ^ players.contains(player)) {
```
tool4ever, line 114: I don't think any changes should be needed here?
the XOR operator should already give you either result

forge-game/src/main/java/forge/game/trigger/TriggerVote.java
Northmoc: Hmm, all the `T:Mode$ Votes` do have the List param so I think this OK for now?

kevlahnota: are you sure its `$ Votes` 
image

not `$ Vote`
image

Northmoc: Ah sorry for the typo.

TriggerVote which corresponds to `T:Mode$ Vote` is the only thing that looks for `List` param.

VoteEffect which would be `SP$ Vote` or `DB$ Vote` doesn't access the TriggerVote code

tool4ever (approved):

PR #3554, opened 2023-07-29
Update card translations

klaxnek:
Updated spanish card translations from scryfall database. Added manually patched new cards.


tool4ever:
thanks
could you run it for the other languages too?

klaxnek:
Yep. No problem I can update it with other language, except French. I've seen that the french translation is manually patched directly in the final file (cardnames-fr-FR.txt) and is not using the patch file (cardnames-fr-FR-patch.txt)

klaxnek:
Finished with all language card translations ;)

tool4ever (approved):

PR #3589, opened 2023-08-01
Seeker of Slaanesh

tool4ever:
Probably corner cases left, though that's more related to the overall implemenation of attack requirements.

But the AI will actually be less suicidal against it, compared to before:
image

### DIFF
--- a/forge-game/src/main/java/forge/game/combat/AttackConstraints.java
+++ b/forge-game/src/main/java/forge/game/combat/AttackConstraints.java
@@ -29,6 +29,7 @@ import forge.game.card.CardCollectionView;
 import forge.game.card.CardLists;
 import forge.game.card.CardPredicates;
 import forge.game.card.CounterEnumType;
+import forge.game.staticability.StaticAbilityMustAttack;
 import forge.game.zone.ZoneType;
 import forge.util.collect.FCollection;
 import forge.util.collect.FCollectionView;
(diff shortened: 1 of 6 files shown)


Northmoc:
Looks great to me, but there are a few things you're doing here that I'm not quite familiar with yet!

forge-game/src/main/java/forge/game/combat/AttackConstraints.java
```diff
@@ -44,12 +45,14 @@ public class AttackConstraints {
...
 
     public AttackConstraints(final Combat combat) {
         final Game game = combat.getAttackingPlayer().getGame();
         possibleAttackers = combat.getAttackingPlayer().getCreaturesInPlay();
         possibleDefenders = combat.getDefenders();
         globalRestrictions = GlobalAttackRestrictions.getGlobalRestrictions(combat.getAttackingPlayer(), possibleDefenders);
+        playerRequirements = StaticAbilityMustAttack.mustAttackSpecific(combat.getAttackingPlayer(), possibleDefenders);
 
         // Number of "must attack" constraints on each creature with a magnet counter (equal to the number of permanents requiring that constraint).
         int nMagnetRequirements = 0;
```
Hanmac, line 58: for later MR, can we turn the following Keywords into Static Abilities?

Northmoc (approved):
👍 👍

PR #3591, opened 2023-08-02
 Remove some fake kicker costs 

tool4ever:


Hanmac:
@tool4ever checkout `K:Presence:Dragon` to be removed

tool4ever:
> Does Dragon's Fire work for the AI after the change? Can't test atm, just remember that card being quite finicky for the AI and sensitive to implementation changes:)

It works just like before (=semi-optimally) because the logic for adding kicker is also closely related to any other OptionalCostValue

tool4ever:
Tested text change and it worked fine

Agetian (approved):

forge-game/src/main/java/forge/game/cost/CostPartMana.java
```diff
@@ -132,6 +132,7 @@ public ManaCost getManaCostFor(SpellAbility sa) {
             return sa.getPaidList(CostExile.HashLKIListKey, true).get(0).getManaCost();
         }
         if (isEnchantedCreatureCost() && sa.getHostCard().isEnchantingCard()) {
+            // TODO human can still activate on TDFC backside
             return sa.getHostCard().getEnchantingCard().getManaCost();
```
Hanmac, line 136: @tool4ever did you test what happen when it gets the "no cost" from a TDFC?

tool4ever: For some reason it will allow payment, will debug it more in another PR

Hanmac (approved):

PR #3595, opened 2023-08-02
fix issue 3447

### LINKED ISSUE #3447: Absorb Energy (YMID) perpetual effect isn't working
Current snapshot, Windows. I countered Sky-Eel School with Absorb Energy. It created a reminder effect in the command zone for the perpetual effect. However, it doesn't note anything about which objects it effects, the creature cards in my hand don't have the new perpetual ability, and they're not receiving the cost reduction when I try to cast them.

Northmoc:
- closes #3447 

Seems like a lot of important things in the Effect were whiffing because the target object that gets remembered is a Spell, not a Card


forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -227,6 +227,9 @@ else if (defined.equals("Remembered") || defined.equals("RememberedCard")) {
             for (final Object o : hostCard.getRemembered()) {
                 if (o instanceof Card) {
                     cards.add((Card) o);
+                } else if (o instanceof Spell) {
```
tool4ever, line 230: I'm not sure if that doesn't cause problems on some script if it finds more than it might expect...
Why not just RememberForCounter$?

tool4ever (approved):

PR #3599, opened 2023-08-02
Endless Whispers and similar

### LINKED ISSUE #1233: EffectOwner$ use is sometimes problematic
An example would be the delayed trigger from _Endless Whispers_:
_Nimble Obstructionist_ allows you to counter it, but you should still be the controller

(Currently you might not be able to target it, but that is a different problem: #1229)

Also it's missing `OneOff$` to prevent it from triggering again after it gets countered. [^1]

For these reasons it might be cleaner to rewrite such cards into DelayedTrigger directly?

[^1]: _Teferi's Time Twist_ also has that problem (if it wouldn't be missing `Duration$ Permanent` anyway), but that param can't be used there because the RE needs to stay

Northmoc:
- closes #1233


forge-gui/res/cardsfolder/q/quenchable_fire.txt
```diff
@@ -1,13 +1,11 @@
...
+SVar:DBEffect:DB$ Effect | EffectOwner$ TargetedOrController | Duration$ Permanent | Abilities$ PayUp | ImprintOnHost$ True | SubAbility$ DBDelayTrig | SpellDescription$ It deals an additional 3 damage to that player or planeswalker at the beginning of your next upkeep step unless that player or that planeswalker's controller pays {U} before that step.
+SVar:PayUp:ST$ ChangeZone | Cost$ U | Defined$ Self | Origin$ Command | Destination$ Exile | ActivationZone$ Command | AILogic$ Always | SpellDescription$ Pay {U} to avoid another 3 damage.
+SVar:DBDelayTrig:DB$ DelayedTrigger | Mode$ Phase | Phase$ Upkeep | ValidPlayer$ You | RememberObjects$ Targeted,Imprinted | Execute$ Bleed | SubAbility$ DBCleanup | StackDescription$ None | TriggerDescription$ It deals an additional 3 damage to that player or planeswalker at the beginning of your next upkeep step unless that player or that planeswalker's controller pays {U} before that step.
+SVar:Bleed:DB$ DealDamage | ConditionDefined$ DelayTriggerRemembered | ConditionPresent$ Effect | ConditionZone$ Command | Defined$ DelayTriggerRemembered | NumDmg$ 3 | SubAbility$ ExileEffect
```
tool4ever, line 8: > If the player has performed the action by then, the “deal an additional 3 damage” ability won’t trigger.

forge-game/src/main/java/forge/game/card/CardProperty.java
```diff
@@ -173,7 +174,16 @@ public static boolean cardHasProperty(Card card, String property, Player sourceC
...
+            List<Object> trigRems = spellAbility.getTriggerRemembered();
+            if (trigRems.size() == 0) { // try wrapped
+                SpellAbility sa = (SpellAbility) spellAbility;
+                SpellAbility root = sa.getRootAbility();
+                if (root.isWrapper()) {
+                    root = ((WrappedAbility) root).getWrappedAbility();
+                }
+                trigRems = root.getTriggerRemembered();
+            }
+            for (Object o : trigRems) {
```
Hanmac, line 186: wouldn't it be better if this part would have been handled by WrappedAbility already?

forge-gui/res/cardsfolder/q/quenchable_fire.txt
```diff
@@ -1,13 +1,11 @@
...
+A:SP$ DealDamage | ValidTgts$ Player,Planeswalker | TgtPrompt$ Select target player or planeswalker | NumDmg$ 3 | SubAbility$ DBEffect | SpellDescription$ CARDNAME deals 3 damage to target player or planeswalker.
+SVar:DBEffect:DB$ Effect | EffectOwner$ TargetedOrController | Duration$ Permanent | Abilities$ PayUp | ImprintOnHost$ True | SubAbility$ DBDelayTrig | SpellDescription$ It deals an additional 3 damage to that player or planeswalker at the beginning of your next upkeep step unless that player or that planeswalker's controller pays {U} before that step.
+SVar:PayUp:ST$ ChangeZone | Cost$ U | Defined$ Self | Origin$ Command | Destination$ Exile | ActivationZone$ Command | AILogic$ Always | SpellDescription$ Pay {U} to avoid another 3 damage.
+SVar:DBDelayTrig:DB$ DelayedTrigger | Mode$ Phase | Phase$ Upkeep | ValidPlayer$ You | RememberObjects$ Targeted,Imprinted | Execute$ Bleed | SubAbility$ DBCleanup | StackDescription$ None | TriggerDescription$ It deals an additional 3 damage to that player or planeswalker at the beginning of your next upkeep step unless that player or that planeswalker's controller pays {U} before that step.
```
tool4ever, line 7: if the trigger gets countered the effect to pay shouldn't stay around

forge-game/src/main/java/forge/game/card/CardProperty.java
tool4ever: - WrappedAbility actually already has a call for it
- even so, an AbilitySub usually shouldn't have WrappedAbility as parent

so I don't know why it was added

forge-gui/res/cardsfolder/q/quenchable_fire.txt
```diff
@@ -1,13 +1,9 @@
...
-SVar:PayUp:ST$ ChangeZone | Cost$ U | Defined$ Self | Origin$ Command | Destination$ Exile | ActivationZone$ Command | AILogic$ Always | SpellDescription$ Pay {U} to remove this effect.
-SVar:UpkeepTrig2:Mode$ Phase | Phase$ Upkeep | ValidPlayer$ Player.IsRemembered | TriggerZones$ Command | Execute$ Bleed2 | TriggerDescription$ EFFECTSOURCE deals an additional 3 damage to that planeswalker at the beginning of that player's next upkeep step unless you pay {U} before that step.
-SVar:Bleed2:DB$ DamageAll | ValidCards$ Card.IsRemembered | NumDmg$ 3 | DamageSource$ EffectSource | SubAbility$ ExileEffect
+A:SP$ DealDamage | ValidTgts$ Player,Planeswalker | TgtPrompt$ Select target player or planeswalker | NumDmg$ 3 | SubAbility$ DBEffect | SpellDescription$ CARDNAME deals 3 damage to target player or planeswalker.
+SVar:DBEffect:DB$ Effect | OneOff$ True | Duration$ Permanent | Abilities$ PayUp | Triggers$ DamageTrig | RememberObjects$ Targeted | SpellDescription$ It deals an additional 3 damage to that player or planeswalker at the beginning of your next upkeep step unless that player or that planeswalker's controller pays {U} before that step.
```
tool4ever, line 5: can't you just use TargetedOrController to avoid the extra code?

tool4ever:
Fixed AI, also for _Xanathar_

Agetian (approved):

PR #3606, opened 2023-08-05
improve Placer logic for some PutCounterAll cards

### LINKED ISSUE #1217: CountersPutAll needs smarter Placer$ support
If you control _Orzhov Advokist_ while your opponent has _Vorinclex, Monstrous Raider_:

When resolving the trigger for counters each player chooses their creature, they both only get 1.

But that's wrong, for this card the putter is the creature controller, so Vorinclex should end up with 4 instead.

Northmoc:
- closes #1217


Agetian (approved):

Agetian:
The AI logic can probably be renamed since it's not exclusive to Orzhov Advokist (maybe something like AILogic$ EachPlayer?), but otherwise looks good to me :)

PR #3619, opened 2023-08-11
Adventure updates / bugfixes

jjayers99:
Not Adventure specific:
Prevent softlock on app startup when Forge doesn't have focus at end of loading sequence
LTR jumpstart block definition for meta-drafts

Adventure specific:
Update of collision handling, more built in collision mappings in source files
Bandit map fixes
Improved controller support for shop screens
Freeze all enemies until player moves again with reward scenes and dialogs
Additional default fields on some tiled objects (full support still pending for a few)
Support for including specific named cards in union reward nodes


Simisays:
Great find on the banditcamp fixes, but I'm wondering how you found that out, didn't remember hearing about that in the Discord

jjayers99:
> Great find on the banditcamp fixes, but I'm wondering how you found that out, didn't remember hearing about that in the Discord

Sometimes it's simple. I walked in to the maps and saw the items failing to load in my error log.

PR #3621, opened 2023-08-12
Turbo Fog ai cards

tehdiplomat:


Hanmac:
i think the ones that are not just generic Fog Effects should have a more finer granulated AI

Also imo we should/could remove the outdated Fog Effect and replace them with Effect too

tehdiplomat:
Sure but thats not really what this PR is. I think I'd be fine leaving the "Fog" API but just having it create an effect instead of the hardcoded "Damage prevented" flag. Whoever did that work would just need to make sure theres an obvious way to not cast multiple fogs in a turn. 

Right now the AI was never casting Fog unless they were about to die, which is fine if you have like 4 in your deck, but doesn't work for this deck archetype.

Agetian:
Really nice! ☺️👍

Agetian (dismissed):

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
```diff
@@ -223,7 +223,7 @@ public static int poisonIfUnblocked(final Card attacker, final Player attacked)
         int damage = attacker.getNetCombatDamage();
         int poison = 0;
         damage += predictPowerBonusOfAttacker(attacker, null, null, false);
-        if (attacker.hasKeyword(Keyword.INFECT)) {
+        if (attacker.hasKeyword(Keyword.INFECT) || attacker.hasKeyword(Keyword.TOXIC)) {
```
tool4ever, line 226: this seems wrong, Toxic has its own magnitude
`predictPoisonFromTriggers` should already consider it

tehdiplomat: Oh right. I'll rename that function since I didn't realize it already had something in it.

tool4ever (approved):

PR #3627, opened 2023-08-13
Vanishing fix

tool4ever:


Hanmac:
i think Vanishing might need extra class for different reminder text without amount

Agetian (approved):

tool4ever:
> i think Vanishing might need extra class for different reminder text without amount

You already added code for it here: https://github.com/Card-Forge/forge/commit/9d66df
so GUI wise it looks fine

Hanmac:
> > i think Vanishing might need extra class for different reminder text without amount
> 
> You already added code for it here: [9d66df](https://github.com/Card-Forge/forge/commit/9d66df) so GUI wise it looks fine

Might not be the best, but I think I can make that better later

PR #3630, opened 2023-08-14
Fix missing cleanup step priority when SBA are performed

tool4ever:


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -1462,7 +1466,7 @@ else if (c.hasBeenDealtDeathtouchDamage() || (c.getDamage() > 0 && c.getLethal()
         }
 
         if (game.getAge() != GameStage.Play) {
-            return;
+            return performedSBA;
```
tehdiplomat, line 1469: Should this just return false if our GameStage isn't Play?

tool4ever: Most likely it will not provide any advantage

Agetian (approved):

PR #3637, opened 2023-08-15
Fix creature haunting

### LINKED ISSUE #3631: Haunt does not work
**Describe the bug**
'Haunt' mechanics does not work. When a creature with Haunt dies you do get to choose a target for haunt, however after that the dead creature with haunt does go to the Graveyard and when the supposedly haunted creature dies, haunt effect does not trigger. 

**To Reproduce**
Example scenario:

1. Blind Hunter dies
2. You choose say Avacynian Priest for the haunt
3. Blind Hunter goes to the graveyard in contrary to what Haunt description says
4. When Avacynian Priest dies, "lose 2 life / gain 2 life" of Blind Hunter does not trigger.

note this is just an example, it's not just Blind Hunter that does not work I've tested it with a few different Haunt cards.

**Expected behavior**
When a creature with Haunt dies, it should be removed from the game "haunting" another creature as per description. When the haunted creature dies, haunt should trigger.

 - OS: Windows 10
 - Version Forge 1.6.57

**Additional context**
Tested in adventure mode against adventure opponents, not sure if its relevant.

tool4ever:
@Hanmac 
looks like this was broken for a while:
the LTB trigger would get hosted by LKI and then fail timestamp check

Closes #3631

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/HauntEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/HauntEffect.java
@@ -1,6 +1,7 @@
 package forge.game.ability.effects;
 
 import forge.game.Game;
+import forge.game.ability.AbilityKey;
 import forge.game.ability.SpellAbilityEffect;
 import forge.game.card.Card;
 import forge.game.spellability.SpellAbility;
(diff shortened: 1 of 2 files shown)


Hanmac:
what happens if the creature is removed from gy before the hunt trigger resolves?

with your changes does this still fizzle as it should?

tool4ever:
yes, then timestamps will be unequal again

Hanmac (approved):
👍

Agetian (approved):

PR #3640, opened 2023-08-16
WOE: Bargain

Hanmac:
First Bargain card

others might need more with Condition Params


Agetian (dismissed):

tool4ever (dismissed):

Hanmac:
i will add the Bargain Condition and the other spoiled Bargain cards (especially the spells) later today

Hanmac:
@tool4ever you might look into updating "Torch the Tower" Description later

tool4ever (approved):
I don't like it but for now `,,,` would give you a linebreak

PR #3644, opened 2023-08-16
Update scripts

### LINKED ISSUE #3617: Buff sharing doesn't seem to share first strike
**Describe the bug**
Cards that share buffs among creatures, like Concerted Effort and Odric, Lunarch Marshal, seem to be ignoring first strike when it should be included. There may be others that don't get shared, but I can confirm that vigilance, lifelink, indestructible, and trample are shared correctly.

**To Reproduce**
Have at least two creatures in play under your control with at least one lacking first strike when the effect of either Concerted Effort or Odric, Lunarch Marshal are triggered.

**Expected behavior**
If you control a creature with first strike, then all other creatures you control will have first strike when the effect is triggered.

**Screenshots**
Screenshot_20230809-153637_Forge

 - Device: Galaxy Note 10+
 - OS: Android 12
 - App version 1.6.58-SNAPSHOT-08.04

tool4ever:
Closes #3617


forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -556,7 +556,7 @@ public static List<String> sharedKeywords(final Iterable<String> kw, final Strin
                 } else if (k.startsWith("Trample")) {
                     tramplekw.add(k);
                 }
-                allkw.add(k);
+                allkw.add(k.toLowerCase());
```
Hanmac, line 559: i don't like this one, are you sure that doesn't bite us in the butt?

tool4ever: I'd say KW equalness should't depend on the case?
I mean as long as we generate instances using `ignoreCase` it certainly feels weird to categorize them with it now, so I wanted to future-proof against a hard to notice bug.
(+ we still return the unchanged KW from the script param)

Hanmac: in very most of the cases, but for some specific Protection or Hexproof it could be problematic when the kw changes into lowercase

Hanmac: would it be better if `allkw.add(k.toLowerCase());` was inside an `else` ?

tool4ever: Sure, that should make it slightly cleaner

Hanmac (approved):
👍

PR #3665, opened 2023-08-20
TriggerTaps: adds Cause for Tapping by ReplacementEffect

### LINKED ISSUE #3647: WOE: Whenever you tap an untapped creature an opponent controls
The Trigger for Taps needs to be extended to have a Cause/Player that causes the Card to be tapped

 Sharae of Numbing Depths even has this:
> Whenever you tap one or more untapped creatures your opponents control

which requires a whole new Trigger

The tricky part right there are Regeneration, because this can cause the opponents card to be tapped
but the trigger should only apply when YOU are causing the destroy
meaning:
* Using Destroy Creature Effect and Regenerate => Trigger applies
* Dealing lethal damage to it, and SBA tries to destroy it => Trigger doesn't apply

and then there is the problem with Multiple Destroy Effects at once
meaning: if i destroy multiple creatures, and all of them regenerate and become tapped, Sharae should only trigger once

Hanmac:
Closes #3647 

Tests cases i tested:

- Tapping by Ability
- Tapping by Regeneration (either controlled by me or controlled by opponent)
- The Regeneration either by Destroy Effect (that should trigger the Taps) or by StateBasedAction (this should not trigger)

 Sharae of Numbing Depths with one or more might be a bit tricky, but @tool4ever said just test it with a Limit on the Trigger?


Agetian (approved):

forge-game/src/main/java/forge/game/ability/effects/TapEffect.java
```diff
@@ -48,7 +48,7 @@ public void resolve(SpellAbility sa) {
                 if (tgtC.isUntapped() && remTapped || alwaysRem) {
                     card.addRemembered(tgtC);
                 }
-                tgtC.tap(true);
+                tgtC.tap(true, sa);
```
tool4ever, line 51: what about funky cards like _Raiding Party_?
While you control the SA you're not always the tapper... :/

Hanmac: for such cases we probably need to wait for CR to fix that later

Hanmac:
https://twitter.com/WotC_Matt/status/1693756748887359770?s=20

>Solitary Sanctuary cares that you’re being asked to tap the creature an opponent controls. If your spell/ability causes an opponent to do it (e.g. Reservoir Kraken) or the game rules do it (e.g. the creature regenerates), the last ability won’t trigger.

for this we need to rewrite who is tapping the creature
or maybe if it is tapped as cost or not

`Raiding Party` might need to be reworked as UnlessCost ?

tool4ever:
We could just use `AbilityKey.Causer` with activatingPlayer as default
and manually tag the few cards with Tapper$ or something?

Hanmac:
> We could just use `AbilityKey.Causer` with activatingPlayer as default and manually tag the few cards with Tapper$ or something?

don't know if Causer is the right Key for this, i used player for now

i updated the Cost ones, as for stuff like Raiding Party, they probably needs to be rewritten

Hanmac:
https://magic.wizards.com/en/news/feature/wilds-of-eldraine-release-notes
>Hylda of the Icy Crown's ability will trigger only when an effect instructs you to tap an opponent's creature. It won't trigger if a spell or ability you control instructs an opponent to tap a creature they control. For example, if you control Tangle Wire and an opponent taps an untapped creature they control as part of the resolution of Tangle Wire's triggered ability, Hylda's ability won't trigger.

 Tangle Wire 
>At the beginning of each player’s upkeep, that player taps an untapped artifact, creature, or land they control for each fade counter on Tangle Wire.

means we do need more Who Taps logic for Tap Effects ... time to dig up all cards that tap

forge-game/src/main/java/forge/game/cost/CostTap.java
```diff
@@ -66,7 +66,7 @@ public final boolean canPay(final SpellAbility ability, final Player payer, fina
 
     @Override
     public boolean payAsDecided(Player ai, PaymentDecision decision, SpellAbility ability, final boolean effect) {
-        ability.getHostCard().tap(true);
+        ability.getHostCard().tap(true, ability, ability.getHostCard().getController());
```
tool4ever, line 69: imo this needs to be the Player param because of _Piracy_ corner case allowing you to pay opponents land costs

tool4ever (approved):

PR #3673, opened 2023-08-22
SLX card scripts (Walking Dead reprints)

paulsnoops:
These are the 6 Walking Dead cards reprinted "In-Universe". Just copied and pasted the scripts and changed some names/oracles. I assume original scripts worked fine so these should too.


Hanmac:
The problem is that rules wise they need to share the name

paulsnoops:
> The problem is that rules wise they need to share the name

Yeah I know but so far no-one has a solution so adding these as a temporary workaround like the others. Once a solution is implemented the scripts for all SLX cards can be merged.

PR #3675, opened 2023-08-22
Adventure - Pathfinding update

jjayers99:
A* implementation for dungeon pathfinding - mobs should now route themselves around obstacles when patrolling or chasing player - flee rework and new behaviors will follow in a future update. Some waypoints may need to be moved away from obstacles by the slightest bit (i.e. x: 100 -> x: 100.01) if mobs fail to navigate to them, and this will optionally allow removal of some waypoints, but no other changes should be necessary for the new behavior to take effect.

New mob waypoint behavior: random waypoints - When specifying a list of waypoints for a mob, the following format is now valid: "1,2,r3-4-5,6" - any number of waypoints can be specified as the character 'r' followed by a dash delimited list of waypoint IDs. The mob will randomly select one of these points each time they loop through their route.

Draft versions of Tiled map template and upcoming dungeons

Reward scene restock and done buttons will now be selectable via controller navigation


jjayers99:
@kevlahnota - I suspect that an appropriate version of gdx-box2d-platform will need to be added to forge-gui-android for this, but will leave that for you to handle rather than doing it incorrectly and then you having to fix it anyway.

kevlahnota:
> @kevlahnota - I suspect that an appropriate version of gdx-box2d-platform will need to be added to forge-gui-android for this, but will leave that for you to handle rather than doing it incorrectly and then you having to fix it anyway.

try to compile on android if this works though

kevlahnota (approved):

PR #3688, opened 2023-08-25
Update GameAction.java

### LINKED ISSUE #3687: Bug: "Damage isn't removed at the end step" and indestructible creatures with excess damage causes freeze on cleanup
image

kevlahnota:
- closes #3687

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -33,6 +33,7 @@ import forge.game.event.*;
 import forge.game.keyword.Keyword;
 import forge.game.keyword.KeywordInterface;
 import forge.game.mulligan.MulliganService;
+import forge.game.phase.PhaseType;
 import forge.game.player.GameLossReason;
 import forge.game.player.Player;
 import forge.game.player.PlayerActionConfirmMode;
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -1316,7 +1317,11 @@ else if (c.hasBeenDealtDeathtouchDamage() || (c.getDamage() > 0 && c.getLethal()
                         }
                         desCreats.add(c);
                         c.setHasBeenDealtDeathtouchDamage(false);
-                        checkAgainCard = true;
+                        boolean check = true;
+                        if (game.getPhaseHandler() != null && c.hasKeyword("Damage isn't removed from CARDNAME during cleanup steps.")) {
+                            check = PhaseType.CLEANUP != game.getPhaseHandler().getPhase();
+                        }
+                        checkAgainCard = check;
```
tool4ever, line 1324: Hmn, I‘m not sure force setting this to false couldn‘t cause issues with other cards of that run rules wise
Gonna debug tonight

tehdiplomat: I would suggest having an else if line on 1302 that does nothing if the creature has indestructible. Then move on to see if the creature was destroed from damage-y things.

tool4ever: yes good call:
> 702.12b. A permanent with indestructible can't be destroyed. Such permanents aren't destroyed by lethal damage, **and they ignore the state-based action that checks for lethal damage**

This was actually pretty bad for performance because it meant indestructibles with lethal would cause the loop to always run the full 9 iterations on each SBA step (until cleanup)

tehdiplomat (approved):

PR #3695, opened 2023-08-26
Role: add StateBasedAction

### LINKED ISSUE #3685: WOE: Role unattach
just to be clear, i will handle the unattach of the Roles later when we get more clarification on what happens when multiple roles are attached at the same time.

if they are attached + unattached in Sequence
or if they are attached at once and the others unattached at once
or if they are just unattached as SBA

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/issues/3636#issuecomment-1691219344_
            
for this we need answers from Rules Manager and/or CR update

Hanmac:
Closes #3685 

the Send to Graveyard should only be needed as StateBasedAction

as per rules, it looks like they are directly send to GY instead of getting unattached first? That might matter?
I still unattach them first for better trigger

As for the Cards that makes the Roles, @Agetian AI might need some logic to not add Roles to cards that already has Roles? (except the Cursed one?)

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/TokenAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/TokenAi.java
@@ -1,5 +1,6 @@
 package forge.ai.ability;
 
+import java.util.List;
 import java.util.Map;
 
 import com.google.common.base.Predicate;
@@ -23,6 +24,7 @@ import forge.game.card.Card;
 import forge.game.card.CardCollection;
 import forge.game.card.CardLists;
 import forge.game.card.CardPredicates;
+import forge.game.card.CardUtil;
 import forge.game.card.token.TokenInfo;
 import forge.game.combat.Combat;
 import forge.game.cost.CostPart;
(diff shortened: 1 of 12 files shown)


tool4ever:
should be good besides AI

Hanmac:
@Agetian can that be merged yet?

i think the general AI about Role Auras can be polished later

tool4ever:
I've added a initial part so at least it doesn't cheat with stuff like _Spiteful Hexmage_

forge-ai/src/main/java/forge/ai/ability/TokenAi.java
```diff
@@ -359,4 +372,42 @@ public static Card spawnToken(Player ai, SpellAbility sa) {
         return result;
     }
 
+    private boolean tgtRoleAura(final Player ai, final SpellAbility sa, final Card tok, final boolean mandatory) {
+        boolean isCurse = tok.getFirstAttachSpell().getParamOrDefault("AILogic", "").equals("Curse");
```
Hanmac, line 376: The SpellAbility that creates the token also should have AILogic Curse

imo that should be checked first

Hanmac:
i think that AI is good for now

about the FristAttachSpell i will probably touch that to a later day when changing EnchantKeyword

Issue #309, opened 2022-05-09
Refactor SpellAbility Payment


Hanmac:
Rules wise it should be like this:

1. > 601.2b If the spell is modal, the player announces the mode choice (see rule 700.2). If the player wishes to splice any cards onto the spell (see rule 702.47), they reveal those cards in their hand. If the spell has alternative or additional costs that will be paid as it’s being cast such as buyback or kicker costs (see rules 118.8 and 118.9), the player announces their intentions to pay any or all of those costs (see rule 601.2f). A player can’t apply two alternative methods of casting or two alternative costs to a single spell. If the spell has a variable cost that will be paid as it’s being cast (such as an {X} in its mana cost; see rule 107.3), the player announces the value of that variable. If the value of that variable is defined in the text of the spell by a choice that player would make later in the announcement or resolution of the spell, that player makes that choice at this time instead of that later time. If a cost that will be paid as the spell is being cast includes hybrid mana symbols, the player announces the nonhybrid equivalent cost they intend to pay. If a cost that will be paid as the spell is being cast includes Phyrexian mana symbols, the player announces whether they intend to pay 2 life or a corresponding colored mana cost for each of those symbols. Previously made choices (such as choosing to cast a spell with flashback from a graveyard or choosing to cast a creature with morph face down) may restrict the player’s options when making these choices.    
    * First choose how to pay Hybrid and Phyrexian Mana, after that the ManaCost should only have normalized Mana Symbols now
3. > 601.2f The player determines the total cost of the spell. Usually this is just the mana cost. Some spells have additional or alternative costs. Some effects may increase or reduce the cost to pay, or may provide other alternative costs. Costs may include paying mana, tapping permanents, sacrificing permanents, discarding cards, and so on. The total cost is the mana cost or alternative cost (as determined in rule 601.2b), plus all additional costs and cost increases, and minus all cost reductions. If multiple cost reductions apply, the player may apply them in any order. If the mana component of the total cost is reduced to nothing by cost reduction effects, it is considered to be {0}. It can’t be reduced to less than {0}. Once the total cost is determined, any effects that directly affect the total cost are applied. Then the resulting total cost becomes “locked in.” If effects would change the total cost after this time, they have no effect.
    * Apply CostReduction and CostIncrease Effects. Previous parts that were looking for Hybrid Mana can be removed. Apply CostSet from TriniSphere. This Also means the AI would need logic to not Choose pay life for Phyrexian Mana if a TriniSphere applies.
[...]

PR #3718, opened 2023-08-31
Mana convert refactor

### LINKED ISSUE #3679: Refactor ManaConvert and similar into Static Abilities
* `K:ManaConvert` for Activated Abilities, currently a Keyword, but should be a Hidden keyword if any?
* "May spend mana as though it were mana of any type to cast CARDNAME" for Spells
* "May spend mana as though it were mana of any color to cast CARDNAME" for Spells

* ManaConversion Param on StaticAbility Continuous 
* "You may spend mana as though it were mana of any color to cast a spell this turn." for `North Star` but this should not use Keywords on Player?

Most of them are Allow More Effects, while for example False Dawn also is a Disallow Effect

Hanmac:
Closes #3679 

TODO: 

- [ ] AI needs better checks, it still looks for only MayPlay, not the other parts
- [ ] the Continuous need to be refactored, but they don't pass through HumanPlaySpellAbility


forge-gui/res/cardsfolder/q/quicksilver_elemental.txt
```diff
@@ -2,7 +2,7 @@ Name:Quicksilver Elemental
 ManaCost:3 U U
 Types:Creature Elemental
 PT:3/4
-K:ManaConvert:Blue->AnyColor:You may spend blue mana as though it were mana of any color to pay the activation costs of CARDNAME's abilities.
+S:Mode$ ManaConvert | ValidPlayer$ You | ValidCard$ Card.Self | ValidSa$ Activated | ManaConversion$ Blue->AnyColor | Description$ You may spend blue mana as though it were mana of any color to pay the activation costs of CARDNAME's abilities.
 A:AB$ Effect | Cost$ U | ValidTgts$ Creature | TgtZone$ Battlefield | TgtPrompt$ Select target creature card | StaticAbilities$ STSteal | RememberLKI$ Targeted | Duration$ UntilHostLeavesPlayOrEOT | SpellDescription$ CARDNAME gains all activated abilities of target creature until end of turn.
 SVar:STSteal:Mode$ Continuous | Affected$ Card.EffectSource | EffectZone$ Command | GainsAbilitiesOfDefined$ RememberedLKI | Description$ Quicksilver Elemental gains all activated abilities of that card until end of turn.
 AI:RemoveDeck:All
```
squee1968, line 8: Ha! I remember scripting this card years ago.

tool4ever:
I'm not sure the Continuous can be refactored better, I mean it probably still needs timestamp ordering?

Hanmac:
> I'm not sure the Continuous can be refactored better, I mean it probably still needs timestamp ordering?

its not about timestamp i just need to move the Continous ones into all places where you can pay costs, but don't use CostPayment (mostly trigger and unless cost)

Hanmac:
what i need to debug: how does (current) and updated System works when you have:

ignore ManaType/Color to Cast this Spell, and while you do cast it, you try to activate ` Agent of Stromgald ` which is:
> {R}: Add {B}.

does the `You can spend mana of any type to cast creature spells.` try to mess with the Mana Ability?

tool4ever:
Ugh, that's indeed a problem but on master and this branch :/
The Agents mana ability can be activated with any color while casting a creature spell.

Somehow the timing for applying/restoring payment ConversionMatrix on the manapool needs to be improved?

tool4ever:
The statics should update but this check here might prevent refresh:
https://github.com/Card-Forge/forge/blob/45cf328a60b26cd79bf1652a6141da86e14a2545/forge-game/src/main/java/forge/game/GameAction.java#L1081

Without it I tested sacrificing _Chromatic Orrery_ during _Flash_ unless cost payment and was then no longer able to spend any color for the rest.

Hanmac:
> Ugh, that's indeed a problem but on master and this branch :/ The Agents mana ability can be activated with any color while casting a creature spell.
> 
> Somehow the timing for applying/restoring payment ConversionMatrix on the manapool needs to be improved?



> The statics should update but this check here might prevent refresh:
> 
> https://github.com/Card-Forge/forge/blob/45cf328a60b26cd79bf1652a6141da86e14a2545/forge-game/src/main/java/forge/game/GameAction.java#L1081
> 
> Without it I tested sacrificing _Chromatic Orrery_ during _Flash_ unless cost payment and was then no longer able to spend any color for the rest.

Wrong card, you need to test `Vizier of the Menagerie` that should only affect creature spells
Does this affect abilities too WHILE paying for a Spell?

tool4ever:
Yes, those were two different tests for different things:
1. the ConversionMatrix from the Vizier will get applied to the ManaPool here:
https://github.com/Card-Forge/forge/blob/86f62a69f3a654b07a56d9291c6c225f3a6ebd66/forge-gui/src/main/java/forge/gamemodes/match/input/InputPayManaOfCostPayment.java#L26C16-L26C16
I don't think there is another reset (``restoreColorReplacements()``) for it, so it **will** affect abilities during payment

- From my understanding the Matrix from CostPayment + CostPartMana are only for transfer values
- During payment only the resulting Matrix inside ManaPool is checked

2. The way it's setup up currently sacrificing either _Vizier_ or _Orrery_ during payment does not undo their Matrix changes, e.g. you can continue to spend any color. Maybe that's correct?

forge-gui/res/cardsfolder/v/vizier_of_the_menagerie.txt
```diff
@@ -4,5 +4,5 @@ Types:Creature Naga Cleric
 PT:3/4
 S:Mode$ Continuous | Affected$ Card.TopLibrary+YouCtrl | AffectedZone$ Library | MayLookAt$ You | Description$ You may look at the top card of your library any time.
 S:Mode$ Continuous | Affected$ Creature.TopLibrary+YouCtrl+nonLand | AffectedZone$ Library | MayPlay$ True | Description$ You may cast creature spells from the top of your library.
-S:Mode$ Continuous | Affected$ Creature.YouCtrl | AffectedZone$ Stack | AddHiddenKeyword$ May spend mana as though it were mana of any type to cast CARDNAME | Description$ You may spend mana as though it were mana of any type to cast creature spells.
+S:Mode$ ManaConvert | ValidPlayer$ You | ValidCard$ Creature.YouCtrl | ValidSa$ Spell | ManaConversion$ AnyType->AnyType | Description$ You may spend mana as though it were mana of any type to cast creature spells.
```
tool4ever, line 7: there is a typo error on these scripts: engine checks for "ValidSA"

but that alone is not enough to fix the problems

Hanmac: I also asked you to check how it was on old code before my changes

tool4ever: "Ugh, that's indeed a problem but on master and this branch :/"

tool4ever: We need to decide if this card should have different interaction outcomes together with Continuous:

- A. Play _Celestial Dawn_ first, then _Vizier_ -> Vizier would allow any color
- B. Play _Vizier_ before _Celestial Dawn_ -> the disallow timestamp is later so nonWhite colors get restricted

Currently the ordering would always lead to result of A, but I think supporting B too would make more sense?
Which would mean they should all just use ManaConvert static (with timestamp ordering support)

- CardPlayOption from MayPlay should probably also be added from its timestamp?
- And finally SA params like `ActivateIgnoreColor` get applied last? 🤔 

With that it would be cleaner to setup the current ConversionMatrix on payment chains:
- Start casting Spell A
- call `StaticAbilityManaConvert.manaConvert`
  - Start activating mana ability B
  - reset ManaPool matrix
  - call `StaticAbilityManaConvert.manaConvert` to calculate matrix for the mana ability
- after finishing paying B either restore previous matrix *OR* (which I think makes more sense) recalculate a new matrix again because statics might have changed?

Hanmac: As far as I think it's that Celestial Dawn's restriction always overwrites the allow from Vizier

But I ask judges

Hanmac: If I'm right and TImestamp doesn't matter, then we can just apply them in any order and the additive and restrictive would just apply 

as for the Payment Chain, i need to see if i convert all of them then it should be easier to reset them back again

Hanmac: Judges so far says Can't beats Can

Hanmac:
For Celestial Dawn and: 
```
S:Mode$ Continuous | Affected$ You | ManaConversion$ White->AnyColor | Description$ You may spend white mana as though it were mana of any color.
S:Mode$ Continuous | Affected$ You | ManaConversion$ UBRG<-C | Description$ You may spend other mana only as though it were colorless mana.
```

* First i want to write this as One Line if possible, it should support a split at `" "`
* Second, i probably need to rethink the OTHER mana part, probably something like NonWhite, to make it better affected by ColorWordChanges

Hanmac:
I need more tests, but for Human Play it seems to work

@tool4ever AI Test Case:
* Plains
* Sunglasses of Urza (Allow to spent White for Red)
* Shock

as Human i can spent the White Mana for Shock, but AI Part fails for both Auto Pay and Tapping of the ManaAbility directly

for ManaAbility Directly, i probably need some more cases where i apply the ManaMatrix to the Pool

forge-gui/src/main/java/forge/gamemodes/match/input/InputPayMana.java
```diff
@@ -345,7 +346,12 @@ else if (colorMatches.size() < abilitiesMap.size()) {
         game.getAction().invoke(new Runnable() {
             @Override
             public void run() {
+                ManaConversionMatrix old = new ManaConversionMatrix();
```
tool4ever, line 349: not ideal
we might just have to do a full recalculation for new boardstate
the only problem is carrying over playerManaConversion? 🤔

forge-game/src/main/java/forge/game/combat/CombatUtil.java
```diff
@@ -288,8 +290,14 @@ public static boolean checkPropagandaEffects(final Game game, final Card attacke
         // If there's a better way of handling this somewhere deeper in the code, feel free to remove
         final SpellAbility fakeSA = new SpellAbility.EmptySa(attacker, attacker.getController());
         fakeSA.setCardState(attacker.getCurrentState());
-        return attacker.getController().getController().payManaOptional(attacker, attackCost, fakeSA,
-                "Pay additional cost to declare " + attacker + " an attacker", ManaPaymentPurpose.DeclareAttacker);
+
+        attacker.getController().getManaPool().restoreColorReplacements();
```
tool4ever, line 294: the problem is it might be needed for human
but for AI both `payManaOptional` and `payCostToPreventEffect` will lead to `ComputerUtilMana.payManaCost` anyway

so we either need to pass the matrix a lot deeper so it knows it can avoid duplicate work or move this somehow?

forge-gui/src/main/java/forge/gamemodes/match/input/InputPayMana.java
Hanmac: I think the spent mana stuff might be locked-in when starting to cast/activate abilities

Otherwise I'm going to ask Judges

forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
```diff
@@ -666,47 +666,42 @@ private static boolean payManaCost(final ManaCostBeingPaid cost, final SpellAbil
...
         ManaCostShard toPay = null;
         List<SpellAbility> saExcludeList = new ArrayList<>();
 
         // Loop over mana needed
         while (!cost.isPaid()) {
             while (!cost.isPaid() && !manapool.isEmpty()) {
                 boolean found = false;
-
-                // Apply the color/type conversion matrix if necessary
-                final CostPayment pay = new CostPayment(sa.getPayCosts(), sa);
```
Hanmac, line 702: @tool4ever I don't know if the removal of CostPayment is okay?

tool4ever: Well I thought if all that happens is creating a fresh Matrix only to copy it later onto the reset ManaPool it's not needed

Currently everything seems to work but this needs more tests later

forge-game/src/main/java/forge/game/combat/CombatUtil.java
Hanmac: i'm trying to check where we would need to apply them deeper for AI or not

or if we should change that for Human

@tehdiplomat your opinions?

forge-gui/src/main/java/forge/gamemodes/match/input/InputPayMana.java
Hanmac: so far this example:

* Vizier of the Menagerie on the field, 
* cast Bayou Groff, sacrifice Vizier as additional Cost

Now the problem, technically you should be able to pay the costs in any order, but Forge doesn't support that

then IF you sacrifice Vizier before paying the Mana, then the apply would/should be lost

forge-game/src/main/java/forge/game/combat/CombatUtil.java
tool4ever: Well I think I cleaned it up, also less code needed this way

forge-gui/src/main/java/forge/player/HumanPlay.java
```diff
@@ -623,6 +624,13 @@ else if (ability.getXManaCostPaid() != null) { //ensure pre-announced X value re
             emerge = ability.getSacrificedAsEmerge();
         }
         if (!toPay.isPaid()) {
+            // if matrix still null it's effect payment
+            if (matrix == null) {
+                matrix = new ManaConversionMatrix();
+                matrix.restoreColorReplacements();
```
Hanmac, line 630: you probably mean restore for manapool? the matrix new created should have restored already

forge-gui/src/main/java/forge/player/HumanPlay.java
```diff
@@ -623,6 +624,13 @@ else if (ability.getXManaCostPaid() != null) { //ensure pre-announced X value re
...
+                StaticAbilityManaConvert.manaConvert(matrix, activator, ability.getHostCard(), ability);
```
Hanmac, line 631: IF its effect payment, then ability should be null (otherwise it 's problematic with ValidSA)

forge-gui/src/main/java/forge/gamemodes/match/input/InputPayMana.java
tool4ever: I guess then we can worry more about it when #309 has progress

forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
```diff
@@ -666,47 +666,42 @@ private static boolean payManaCost(final ManaCostBeingPaid cost, final SpellAbil
...
+            } else if (mayPlay.isIgnoreManaCostColor()) {
+                ignoreColor = true;
             }
-        } else if (sa.hasParam("ActivateIgnoreColor")) {
+            mayPlay.applyManaConvert(manapool);
+        } else if (sa.isActivatedAbility() && sa.getGrantorStatic() != null && sa.getGrantorStatic().hasParam("ManaConversion")) {
+            AbilityUtils.applyManaColorConversion(manapool, sa.getGrantorStatic().getParam("ManaConversion"));
             ignoreColor = true;
         }
+        StaticAbilityManaConvert.manaConvert(manapool, ai, sa.getHostCard(), sa);
```
Hanmac, line 686: for these we need to check if effect is false (and mayPlay should probably check for isSpell too)

the last line with `StaticAbilityManaConvert.manaConvert(manapool, ai, sa.getHostCard(), sa);` should use `null` if effect is true

forge-gui/src/main/java/forge/player/HumanPlay.java
tool4ever: no, the problem is new object starts with {0} matrix on `colorRestriction` so nothing can be paid instead of default

maybe we should make that part of the constructor?

forge-gui/src/main/java/forge/gamemodes/match/input/InputPayMana.java
Hanmac: yeah, i think it might be better to allow that for now

Hanmac:
currently MayPlay only has two different ManaConvert, either IgnoreColor or IgnoreType (and that one Special Case about Snow)

i think we should keep if that way for now, and i might do it to a more flexible parameter once i try to refactor MayPlay in general

tool4ever:
**Test results**

Improved over master:
1. _Vizier_ does not affect payment matrix of mana ability while casting a creature spell ✔️ 
2. AI is able to use _Vizier_ matrix to pay with other colors ✔️

Regression tests:

3. mixing multiple e.g. _Sunglasses_ does not mess up the final matrix ✔️
4. AI can't cheat with allow/disallow from _Celestial Dawn_ ✔️
5. AI can use MayPlay conversions ⚠️

@Hanmac
This one got broken because of refactor ``sa.getHostCard().mayPlay(sa.getMayPlay())``
- First AI calls ``payManaCost`` with test=true and it's ok
- Next call with test=false for real payment
> ComputerUtilMana: payManaCost() cost was not paid for Abbey Griffin - Creature 2 / 2 by Hurl Through Hell (122) (may spend mana as though it were mana of any color to cast it) (Abbey Griffin). Didn't find what to pay for {W}
- it probably happens because the hostCard gets moved to stack, so its MayPlay values are gone (for human it seems to work because it grabs the CardPlayOption before)

Hanmac:
> **Test results**
> @Hanmac This one got broken because of refactor `sa.getHostCard().mayPlay(sa.getMayPlay())`
> 
>     * First AI calls `payManaCost` with test=true and it's ok
> 
>     * Next call with test=false for real payment
> 
> 
> > ComputerUtilMana: payManaCost() cost was not paid for Abbey Griffin - Creature 2 / 2 by Hurl Through Hell (122) (may spend mana as though it were mana of any color to cast it) (Abbey Griffin). Didn't find what to pay for {W}
> 
>     * it probably happens because the hostCard gets moved to stack, so its MayPlay values are gone (for human it seems to work because it grabs the CardPlayOption before)

hm could we store the CardPlayOption on addition to the MayPlay in the SpellAbility ? (until i try to refactor it?)

tool4ever:
fixed

forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
```diff
@@ -666,47 +666,45 @@ private static boolean payManaCost(final ManaCostBeingPaid cost, final SpellAbil
...
-            } else if (mayPlay.hasParam("MayPlayIgnoreType")) {
-                ignoreType = true;
-            }
-        } else if (sa.hasParam("ActivateIgnoreColor")) {
-            ignoreColor = true;
-        }
         boolean hasConverge = sa.getHostCard().hasConverge();
+        // TODO use matrix instead of ignoreColor, ignoreType
         ListMultimap<ManaCostShard, SpellAbility> sourcesForShards = getSourcesForShards(cost, sa, ai, test,
                 checkPlayable, hasConverge, ignoreColor, ignoreType);
```
Hanmac, line 698: @tool4ever can we try to check the manaPool instead of ignoreColor/ignoreType to check what colors AI needs for what shards?

imo that would makes it more flexible

tool4ever: Need to test a few things, so far it seems like that part might just be obsolete now

Hanmac: ignoreColor/ignoreType is used on some places later too, feel free to try to remove it

tool4ever: Done - I could not find any regressions from it :)

And this also fixes another AI payment case, in this scenario it will now cast my exiled creature:
image

Hanmac:
@Agetian @tehdiplomat you guys have some suggestions?

i think this MR is clear for now?

tehdiplomat:
I would love to see some system tests to prove that this isn't breaking anything that already exists, or potentially fixing problems that were already broken.

Agetian:
Good point! :+1:

Hanmac:
@Agetian @tool4ever that one Test for `Agent of Stromgald` would be interesting because that's one of the things that should be broken in master

tool4ever:
that one was human only (in the way that AI can't pay with mana abilities that have mana costs anyway)
so we really can't test for it :/

it would be nice if AI could pay with filter lands but I think that's another project complexity wise

tool4ever (approved):
LGTM

squee1968:
Helluva job!!! A lot of hard work, but well worth the effort and greatly appreciated!

PR #3723, opened 2023-09-01
CardDb: support rebranded versions

### LINKED ISSUE #1927: Commander not respecting Universes Within/Universes Beyond
When deck building in Commander (and I assume other formats as well) the game doesn't seem to be able to tell that the Universes Within cards are the same as the Universes Beyond ones for name purposes. Example: Chief Jim Hopper and Sophina, Spearsage Deserter are considered the same card, but the game allows both to be in the same Commander deck.

tool4ever:
Closes #1927 by avoiding script duplicates.

I decided we can't share the CardFace with the rebranded version because of different Oracle...
but we still avoid parsing it again so the bigger String data is the same object!

I thought about mapping this via edition files but supporting this functionality just via a single file change felt more appealing in the end - especially since I extended the previously unused "AltName" field which already provided some of the needed support logic to make this work as smooth as possible.

Also engine & deck editor should recognize they're to be treated as identical.

I'd like to leave the dirty work of script cleanup to another PR or someone else ;)


paulsnoops:
I'll clean up the rest of the scripts once this has been merged :)

tool4ever:
Yea I think that makes most sense since that script will get written first, so then later only 1/2 lines need to be added

paulsnoops (approved):

PR #3738, opened 2023-09-06
Fix LastChapter logic

tool4ever:


Hanmac (approved):

forge-core/src/main/java/forge/deck/DeckFormat.java
```diff
@@ -341,7 +341,7 @@ public String getDeckConformanceProblem(Deck deck) {
         final CardPool allCards = deck.getAllCardsInASinglePool(hasCommander());
 
         // should group all cards by name, so that different editions of same card are really counted as the same card
-        for (final Entry<String, Integer> cp : Aggregates.groupSumBy(allCards, PaperCard.FN_GET_NAME)) {
+        for (final Entry<String, Integer> cp : Aggregates.groupSumBy(allCards, pc -> StaticData.instance().getCommonCards().getName(pc.getName(), true))) {
```
Hanmac, line 344: I liked the FN a bit more, but that's okay too I think

PR #3758, opened 2023-09-10
Support for Imodane, the Pyrohammer

kevlahnota:


forge-game/src/main/java/forge/game/trigger/TriggerDamageDoneOnce.java
```diff
@@ -36,6 +37,21 @@ public boolean performTest(Map<AbilityKey, Object> runParams) {
             return false;
         }
 
+        if (hasParam("SingleTarget")) {
+            Object target = runParams.get(AbilityKey.DamageTarget);
+            final Map<Card, Integer> damageMap = (Map<Card, Integer>) runParams.get(AbilityKey.DamageMap);
+            for (Card source : getDamageSources(damageMap)) {
```
tool4ever, line 43: if you use DamageDone trigger instead you wouldn't need to iterate over all sources?

forge-game/src/main/java/forge/game/trigger/TriggerDamageDoneOnce.java
```diff
@@ -36,6 +37,21 @@ public boolean performTest(Map<AbilityKey, Object> runParams) {
             return false;
         }
 
+        if (hasParam("SingleTarget")) {
```
tool4ever, line 40: clearer name like "SourceDamagedTarget"?

Hanmac:
i got a different aproach, going to commit soon, but i'm taking the card script

Hanmac:
i think in most cases, the castSA should work?

i mean is there a card that can cause an Instant and Sorcery to deal damage (while not being the Spell itself?)

kevlahnota:
closes in favor of #3759

PR #3761, opened 2023-09-10
Fix Chatterfang + Role token

tool4ever:


forge-game/src/main/java/forge/game/ability/effects/TokenEffectBase.java
```diff
@@ -134,7 +134,10 @@ protected TokenCreateTable makeTokenTable(TokenCreateTable tokenTable, final boo
                 }
 
                 if (!sa.hasParam("AttachAfter") && sa.hasParam("AttachedTo") && !attachTokenTo(tok, sa)) {
-                    continue;
+                    // CR 303.4i
+                    if (tok.isAura()) {
```
Hanmac, line 138: Add this to the If Conditions above at the end?

Hanmac (approved):

PR #3770, opened 2023-09-13
Player: first step for Ashiok

Hanmac:
i still have some questions about the Cause Param or if the exile function should get more move Params

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
@@ -3018,7 +3018,7 @@ public class CardFactoryUtil {
                 @Override
                 public void resolve() {
                     final Game game = getHostCard().getGame();
-                    final Card c = game.getAction().exile(getHostCard(), this);
+                    final Card c = game.getAction().exile(getHostCard(), this, null);
                     c.setForetold(true);
                     game.getTriggerHandler().runTrigger(TriggerType.IsForetold, AbilityKey.mapFromCard(c), false);
                     c.setForetoldThisTurn(true);
@@ -3434,7 +3434,7 @@ public class CardFactoryUtil {
                 @Override
                 public void resolve() {
                     final Game game = this.getHostCard().getGame();
-                    final Card c = game.getAction().exile(this.getHostCard(), this);
+                    final Card c = game.getAction().exile(this.getHostCard(), this, null);
 
                     int counters = AbilityUtils.calculateAmount(c, k[1], this);
                     GameEntityCounterTable table = new GameEntityCounterTable();
--- a/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
@@ -3018,7 +3018,7 @@ public class CardFactoryUtil {
                 @Override
                 public void resolve() {
                     final Game game = getHostCard().getGame();
-                    final Card c = game.getAction().exile(getHostCard(), this, null);
+                    final Card c = game.getAction().exile(new CardCollection(getHostCard()), this, null).get(0);
                     c.setForetold(true);
                     game.getTriggerHandler().runTrigger(TriggerType.IsForetold, AbilityKey.mapFromCard(c), false);
                     c.setForetoldThisTurn(true);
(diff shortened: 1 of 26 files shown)


forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -625,6 +625,13 @@ public final boolean payLife(final int lifePayment, final SpellAbility cause, fi
             return false;
         }
 
+        // Ashiok Exile instead of paying life
+        if (hasKeyword("Exile library instead of pay life") && lifePayment <= getZone(ZoneType.Library).size()) {
```
tool4ever, line 629: while technically you can't get different results when controlling multiple Ashiok it should probably still be done as Replacement to avoid problems later?

Hanmac: Ashiok, Wicked Manipulator doesn't have "exiled with"
But probably yeah 

I will think about that

Of cause that makes passing the params even more complicated

Hanmac:
example on how the values need to be passed through:

* Multiple Lands with PayLife ETB enter the battlefield from the Library
* First, the Player need to select them via GUI
* Then they need to be removed from Library and placed in a Limbo like place (for this we probably need better way than int Libary position for this)
  * So they can't be exiled for itself, but what about other cards?
* Then each of them tries to move onto the battlefield
  * while respecting CantHappen for `Lands can't enter the battlefield` stuff, but we might move this check extra above the Limbo place
  * When the ETB -> PayLife does happen, the New Replacement Effect should apply, replacing them with Exile N Cards
  * These N cards should be exiled separately, causing Laelia to trigger once for each Exile N Cards part
* for the general CardZoneTable, currently i think the Exile N Cards happens to trigger their own ChangesZoneAll

Hanmac:
i'm currently reworking it into a ReplacePayLife

i was looking at the `can't pay Life` Static that might be changed into a CantHappen Replacement

and while that i was looking at `CantChangeLife` and i was asking Judges and Rules Manager if Ashiok might bypass the `your life can't change` restriction from `Platinum Emperion` or if it still prevents you from paying life

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -3018,7 +3018,7 @@ public boolean isForetelling() {
                 @Override
                 public void resolve() {
                     final Game game = getHostCard().getGame();
-                    final Card c = game.getAction().exile(getHostCard(), this);
+                    final Card c = game.getAction().exile(getHostCard(), this, null);
```
tool4ever, line 3021: Suspend + Foretell should also be done with CardCollection so Ranar ChangesZoneAll can trigger for it

Hanmac: Hm OK it should use the CardCollection one?

tool4ever: that seems reasonable

tool4ever:
maybe we could have all effects that can change zones store their ETB cards in Game for checking or something 🤔 
Since the rule says they are still in library at that point other effects would miscount if they moved to limbo place first?

Hanmac:
> maybe we could have all effects that can change zones store their ETB cards in Game for checking or something 🤔 Since the rule says they are still in library at that point other effects would miscount if they moved to limbo place first?

imo that can be done later, you should try to have a first version of Ashiok

forge-gui/res/cardsfolder/upcoming/ashiok_wicked_manipulator.txt
```diff
@@ -0,0 +1,14 @@
+Name:Ashiok, Wicked Manipulator
+ManaCost:3 B B
+Types:Legendary Planeswalker Ashiok
+Loyalty:5
+R:Event$ PayLife | ActiveZones$ Battlefield | ValidPlayer$ You | ReplaceWith$ ExileTop | Amount$ GT0 | Description$ If you would pay life while your library has at least that many cards in it, exile that many cards from the top of your library instead.
```
kevlahnota, line 5: @Hanmac how to limit the condition for the library check?

Hanmac: @kevlahnota LEN where where N = Library Size

kevlahnota: Doesn't work with the amount comparison if I change the condition to Amount$ LEY (SVar Y is declared but crashes since it can't parse the Y variable)

Hanmac: fixed

Hanmac:
@tehdiplomat @tool4ever @Agetian should we merge that soon ?

tehdiplomat (approved):

PR #3772, opened 2023-09-13
Adventure updates

jjayers99:
Large sprite collision fix
Jumpstart reward fix
Enemy instance specific deck override capability
Selectable reward framework
Work in progress for new storyline dungeons
Removal of negative reputation from declining some difficult quests


forge-gui-mobile/src/forge/adventure/scene/RewardScene.java
```diff
@@ -476,6 +491,13 @@ else if (fW / fH >= 2f)
                     stage.addActor(buyCardButton);
                     addToSelectable(buyCardButton);
                 }
+            } else if (type == Type.RewardChoice){
+                if (currentRow != ((i + 1) / numberOfColumns))
+                    yOff += doneButton.getHeight();
```
kevlahnota, line 496: Does the new button support portrait mode?

jjayers99: It should work for portrait, as this is mostly a copy of buy buttons from the shop version of the screen. This has not been tested in portrait however, so I will verify before putting in any rewards of this type beyond demos.

kevlahnota (approved):

PR #3787, opened 2023-09-17
Fix Ashiok

tool4ever:
@Hanmac 
Current implementation can lead to card duplication:

- Cast _Genesis Ultimatum_ with 3 _Breeding Pool_  on top
- Choose top two for ETB
- Pay 2 Life for both to replace with _Ashiok_
- End Result: 4 Pools ⚠️  (1 Exile, 1 Hand, 2 in play)

Here's my idea
- [x] need to update other effects

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/ashiok_wicked_manipulator.txt
+++ b/forge-gui/res/cardsfolder/upcoming/ashiok_wicked_manipulator.txt
@@ -3,7 +3,7 @@ ManaCost:3 B B
 Types:Legendary Planeswalker Ashiok
 Loyalty:5
 R:Event$ PayLife | ActiveZones$ Battlefield | ValidPlayer$ You | ReplaceWith$ ExileTop | Amount$ LEY | Description$ If you would pay life while your library has at least that many cards in it, exile that many cards from the top of your library instead.
-SVar:ExileTop:DB$ DigUntil | Amount$ X | Valid$ Card.NotDefinedReplacedSimultaneousETB | NoMoveRevealed$ True | FoundDestination$ Exile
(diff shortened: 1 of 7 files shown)


Hanmac (dismissed):

Hanmac:
for this we need to check if the cards are moved as one event (Dig) or if they are moved in sequence (like Cascade / DigUntil)

forge-gui/res/cardsfolder/upcoming/ashiok_wicked_manipulator.txt
```diff
@@ -3,12 +3,12 @@ ManaCost:3 B B
 Types:Legendary Planeswalker Ashiok
 Loyalty:5
 R:Event$ PayLife | ActiveZones$ Battlefield | ValidPlayer$ You | ReplaceWith$ ExileTop | Amount$ LEY | Description$ If you would pay life while your library has at least that many cards in it, exile that many cards from the top of your library instead.
-SVar:ExileTop:DB$ Dig | DigNum$ X | ChangeNum$ All | DestinationZone$ Exile
+SVar:ExileTop:DB$ DigUntil | Amount$ X | Valid$ Card.NotDefinedReplacedSimultaneousETB | NoMoveRevealed$ True | FoundDestination$ Exile
```
Hanmac, line 6: You need to check if your DigUntil change does alter how the Cards are Exiled

Ashiok should still exile them at once, not in sequence

tool4ever: yea that part is ok
only if RevealedDestination$ was also Exile DigUntil logic would think it's doing "Exile Until"

kevlahnota (approved):

PR #3809, opened 2023-09-23
Alchemy scryfall downloads need a hyphen in their collector "number"

tehdiplomat:


forge-gui/src/main/java/forge/util/ImageFetcher.java
```diff
@@ -205,9 +205,10 @@ public void fetchImage(final String imageKey, final Callback callback) {
             }
             final String cardCollectorNumber = paperCard.getCollectorNumber();
             if (!cardCollectorNumber.equals(IPaperCard.NO_COLLECTOR_NUMBER)) {
-                final String scryfallURL = this.getScryfallDownloadURL(paperCard, face, useArtCrop, hasSetLookup, filename, downloadUrls);
-                if (scryfallURL != null && !hasSetLookup)
-                    downloadUrls.add(scryfallURL);
+                if (!hasSetLookup) {
```
tool4ever, line 208: didn't you change the logic here?
when SetLookup=true no URL will be added now 🤔

forge-gui/src/main/java/forge/util/ImageFetcher.java
```diff
@@ -65,13 +65,13 @@ private String getScryfallDownloadURL(PaperCard c, String face, boolean useArtCr
                     }
                 }
             }
-            return null;
         } else {
             String setCode = edition.getScryfallCode();
             String langCode = edition.getCardsLangCode();
-            return ForgeConstants.URL_PIC_SCRYFALL_DOWNLOAD +
-                    ImageUtil.getScryfallDownloadUrl(c, face, setCode, langCode, useArtCrop);
+            downloadUrls.add(ForgeConstants.URL_PIC_SCRYFALL_DOWNLOAD + ImageUtil.getScryfallDownloadUrl(c, face, setCode, langCode, useArtCrop));
```
tool4ever, line 71: For most cards you would just add the same URL twice...
Meaning two threads would try the same download, which seems pretty nasty ⚠️

forge-gui/src/main/java/forge/util/ImageFetcher.java
tehdiplomat: Yea. I think when i made this change I didn't realize that this function updated downloadUrls and never came back to it.

forge-gui/src/main/java/forge/util/ImageFetcher.java
tehdiplomat: Alright that should be fixed now.

tool4ever (approved):

PR #3824, opened 2023-09-27
prevent NPE

kevlahnota:
- mobileGUI confirm handles null cardview


tool4ever:
It feels weird that effect is even null there
do you know which card was involved there?

kevlahnota:
> It feels weird that effect is even null there do you know which card was involved there?

found here on sentry: https://sentry.cardforge.org/organizations/forge/issues/29930/?project=2&project=3&query=is%3Aunresolved&referrer=issue-stream&stream_index=21
image

PR #3827, opened 2023-09-28
Fix puzzle priority

marthinwurer:
We'll see if I implemented this right. The first puzzle ends up at the main phase.

### DIFF
--- a/forge-game/src/main/java/forge/game/phase/PhaseHandler.java
+++ b/forge-game/src/main/java/forge/game/phase/PhaseHandler.java
@@ -101,13 +101,13 @@ public class PhaseHandler implements java.io.Serializable {
     private transient Player pPlayerPriority = null;
     private transient Player pFirstPriority = null;
     private transient Combat combat = null;
+    private boolean combatDamageAssigned = false;
     private boolean bRepeatCleanup = false;
 
     private transient Player playerDeclaresBlockers = null;
     private transient Player playerDeclaresAttackers = null;
 
     /** The need to next phase. */
-    private boolean givePriorityToPlayer = false;
 
     private final transient Game game;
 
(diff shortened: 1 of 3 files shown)


forge-game/src/main/java/forge/game/phase/PhaseHandler.java
```diff
@@ -264,20 +264,83 @@ private boolean isSkippingPhase(final PhaseType phase) {
         }
     }
 
+    public boolean givePriorityToPlayer() {
+        if (isSkippingPhase(phase)) {
+            return true;
```
tool4ever, line 269: why would this return true when it gets skipped? 😕

forge-game/src/main/java/forge/game/phase/PhaseHandler.java
```diff
@@ -1281,7 +1311,7 @@ public String debugPrintState(boolean hasPriority) {
 
     // just to avoid exposing variable to outer classes
     public void onStackResolved() {
-        givePriorityToPlayer = true;
+//        givePriorityToPlayer = true;
```
tool4ever, line 1314: was it already obsolete or why this change?
Seems to satisfy 117.3b

Agetian:
What's the exact issue that you are trying to solve?

marthinwurer:
This one reported on the discord: https://discord.com/channels/267367946135928833/1156369694958293063

forge-game/src/main/java/forge/game/phase/PhaseHandler.java
marthinwurer: Good catch, it should be false.

kevlahnota:
I tried this fix but it produces NPE when using Ninjutsu ability, to reproduce declare an attacker ie raging goblin, then on declare blockers step, activate ninjutsu ability targeting raging goblin then let the ability resolve, NPE is thrown.. 
```
java.lang.NullPointerException: null
    at forge.game.ability.effects.ChangeZoneEffect.changeKnownOriginResolve(ChangeZoneEffect.java:702)
    at forge.game.ability.effects.ChangeZoneEffect.resolve(ChangeZoneEffect.java:438)
    at forge.game.ability.AbilityApiBased.resolve(AbilityApiBased.java:41)
    at forge.game.ability.AbilityUtils.resolveApiAbility(AbilityUtils.java:1427)
    at forge.game.ability.AbilityUtils.resolve(AbilityUtils.java:1395)
    at forge.game.zone.MagicStack.resolveStack(MagicStack.java:547)
    at forge.game.phase.PhaseHandler.mainLoopStep(PhaseHandler.java:1257)
    at forge.game.phase.PhaseHandler.startFirstTurn(PhaseHandler.java:1142)
    at forge.game.GameAction.startGame(GameAction.java:2108)
    at forge.game.Match.startGame(Match.java:105)
    at forge.gamemodes.match.HostedMatch$2.run(HostedMatch.java:255)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
    at java.lang.Thread.run(Thread.java:834)
```

forge-game/src/main/java/forge/game/phase/PhaseHandler.java
kevlahnota: @Agetian It works fine here, Don't really know if this covers 117.3b  correctly. 
image

forge-game/src/main/java/forge/game/phase/PhaseHandler.java
tool4ever: why resolve, still seems wrong...?

forge-game/src/main/java/forge/game/phase/PhaseHandler.java
tool4ever: Well my question was did it have to be commented out now to make this fix work at all?

kevlahnota:
Karn Liberated fails restarting the game since the restartgameeffect relies on the old phasehandler fallback code.

Agetian:
Is the original contributor still available to make the necessary followup changes to make this fix work without breaking other aspects of the game?
If it proves to be too difficult or too time-consuming to devise a working fix, I suggest that we either (a) roll back the original change that caused the Puzzle Mode to break, until a working solution is found *or* (b) implement a simpler "workaround" fix in devAdvancePhase (giving priority to the player in it) that unbreaks Puzzle Mode while keeping the originally contributed MR intact. This seems to work in my local testing, though I haven't tested all the possible corner cases. Whether we go for (b) or not, of course, also depends on whether anything else is broken (except Puzzle Mode) without a followup fix. 
What do you think?

tool4ever:
Well seeing that @kevlahnota discovered another problem already existing after #3669 I'm leaning towards reverting that now before it gets too difficult (unless that's another easy fix)

PR #3844, opened 2023-10-02
Some tweaks related to KW costs

tool4ever:
- Fix missing ChangesZoneAll trigger for Offering/Emerge/Aura swap
- Fix so spell copies can find objects that convoked original


tool4ever:
While Offering might still be improved in the future this should help for now

Agetian (approved):

PR #3856, opened 2023-10-06
Adventure - Graveyard map tweaks

jjayers99:
Minor visual updates, blocked off area that looked as though it could have been the map exit, opened blocked pathway to the top left section of map.


Simisays:
👍

PR #3887, opened 2023-10-11
Add back faced cards to writeDeckHtml

karlek:
First: super cool project!

When printing a deck from forge-desktop-gui, cards with back faces were not fully represented. This simple PR, simply adds back faced cards to the output of `DeckHtmlSerializer.writeDeckHtml`.

/ Henry


tool4ever (approved):
thanks

It'd be interesting to know if this feature actually still works for you, seems like the URLs end up slightly wrong?

In either case this seems good to have

karlek:
Yeah, they don't work right away. But I usually patch the output html with a few sed replacements:

```bash
sed -i 's#https://downloads.cardforge.org/images#http://localhost:8000#g' proxy.html
sed -i 's#.full.#.fullborder.#g' proxy.html
cd ~.cache/forge/pics && python3 -m http.server
```

PR #3890, opened 2023-10-12
day_of_the_moon.txt and support/refactor in NameCard

Northmoc:


tool4ever:
Meh, not too thrilled this got done (meaning foremost the card, not the script)

If only they templated it more in the "Note ..." kind of style, especially with _The Toymaker's Trap_ needing the same kind of treatment for ChosenNumber... 🤹 

The extra cleanup sure feels awkward, though I don't have a better idea at this point

Agetian (approved):

Issue #3790, opened 2023-09-18
AI makes Mite token in response but they can't block


Hanmac:
i noticed multiple times that AI was trying to create Mite token in response to an Attack

but totally didn't notice that these Token can't block

Agetian:
I'm currently investigating the issue with the AI creating a Phyrexian Mite token in response to an attack. It looks like CombatUtil.cantBlock doesn't account for the possibility of a static ability implementation of this, and I don't see an easy way to check for its presence... What would be the preferred way to test for a CantBlockBy static? Should I iterate over all possible attackers and test whether the static applies? That might be a bit expensive performance-wise. Another issue is that this kind of check doesn't seem to work for hypothetical conditions (e.g. a card object that doesn't exist on the battlefield yet), which complicates matters further, since we want to test the token before it's spawned on the battlefield. Assistance is needed with this, as it prevents from using the CantBlockBy / CantBlock static more extensively without breaking the AI.

Hanmac:
@Agetian possible easy fix:
add attacker and blocker to the cards checked by 
https://github.com/Card-Forge/forge/blob/7c303d2fb9ce7439a4f1600789a6d63f9b2835c0/forge-game/src/main/java/forge/game/staticability/StaticAbilityCantAttackBlock.java#L147

with that, it should be able for the AI to use LKI of the possible token to check against?

Agetian:
Umm, how exactly? I tried looking at that spot before but haven't found a working way to integrate it there... :/

Hanmac:
Like this

https://github.com/Card-Forge/forge/blob/7c303d2fb9ce7439a4f1600789a6d63f9b2835c0/forge-game/src/main/java/forge/game/staticability/StaticAbilityCantBeCast.java#L44-47

Agetian:
Yes, pretty much, I did a similar thing there.

Hanmac:
[...]

PR #3891, opened 2023-10-13
Cleanup fixes

### LINKED ISSUE #3886: Immobilizer Eldrazi rulings not correctly applied
**Describe the bug**
According to the [rulings](https://gatherer.wizards.com/pages/card/Details.aspx?multiverseid=407607) `Immobilizer Eldrazi`'s ability should check for power and toughness when you declare blockers however it seems to check it as the ability resolves.

**To Reproduce**
1. Begin the puzzle `Perplexing Chimera (GatheringMagic.com) 012616 - Sucker Punch`
2. You should be in the attack phase of your opponent
3. During the attack phase of your opponent do the following:
      a. Cast `Sweep Away` using `1UR` targetting opponent's attacking `Mountain` and choosing to put it into their library.
      b. Cast `Dimensional Infiltrator` using `U`. Before it resolves cast `Abstruse Interference` using `U` and `Hedron Crawler`'s ability targetting the `Dimensional Infiltrator`.
      c. Let `Kozilek's Sentinel` trigger resolve so it becomes a `2/4`.
      d. Let `Abstruse Interference` resolve but pay `R` so `Dimensional Infiltrator` doesn't get counter.
      e. Let `Kozilek's Sentinel` trigger resolve so it becomes a `3/4`.
      f. Finally let `Dimensional Infiltrator` resolve.
      g. Using `R` and `Cultivator Drone`'s ability, activate `Dimensional Infiltrator`'s ability. The card on top of your opponent's library is a `Mountain` and we can return `Dimensional Infiltrator` to our hand.
      h. Finally, cast `Dimensional Infiltrator` again and let everything resolve. Your `Kozilek's Sentinel` should now be a `4/4`.
4. Move to the `Declare Blockers` phase and try to block any creature with `Kozilek's Sentinel`.
5. `Kozilek's Sentinel` is not able to block any creature.

**Expected behavior**
`Kozilek's Sentinel` is now a `4/4` and should be able to block even though it was a `1/4` when `Immobilizer Eldrazi`'s ability was activated.

**Desktop**
 - OS: MacOs
 - Browser: chrome (not sure why this matters)
 - Version: mtg forge 1.6.59

**Additional context**
I assume there are many ways to hit this issue, I just found it during the puzzle.

tool4ever:
- Closes #3886

- Fix _Vodalian War Machine_ trigger not counting Merfolk if its abilities get countered


forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -448,11 +448,12 @@ public Integer announceRequirements(final SpellAbility ability, final String ann
         boolean canChooseZero = true;
         Cost cost = ability.getPayCosts();
 
+        if (ability.hasParam("XMaxLimit")) {
```
Hanmac, line 451: XMaxLimit should only apply when announce = X ?

forge-gui/res/cardsfolder/i/immobilizer_eldrazi.txt
```diff
@@ -3,6 +3,7 @@ ManaCost:1 R
 Types:Creature Eldrazi Drone
 PT:2/1
 K:Devoid
-A:AB$ PumpAll | Cost$ 2 C | ValidCards$ Creature.powerLTtoughness | KW$ HIDDEN CARDNAME can't block. | SpellDescription$ Each creature with toughness greater than its power can't block this turn.
```
Hanmac, line 6: `HIDDEN CARDNAME can't block.` needs to be removed from other cards too

tool4ever: I know but too many for now especially since AI logic depends on it

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
tool4ever: Hmn, I kind of wanted to avoid the extra code for _Tranquil Frillback_
Don't think it's possible for another Announce to get mixed there? 🤔

Hanmac: it might be confusing to use XMaxLimit for something not named X

forge-gui/res/cardsfolder/e/elvish_vatkeeper.txt
```diff
@@ -5,7 +5,7 @@ PT:3/3
 T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigIncubate | TriggerDescription$ When CARDNAME enters the battlefield, incubate 2. (Create an Incubator token with two +1/+1 counters on it and "{2}: Transform this artifact." It transforms into a 0/0 Phyrexian artifact creature.)
 SVar:TrigIncubate:DB$ Incubate | Amount$ 2
 A:AB$ SetState | Cost$ 5 | TgtPrompt$ Select target Incubator token you control | ValidTgts$ Incubator.token+YouCtrl | Mode$ Transform | SubAbility$ DBDouble | SpellDescription$ Transform target Incubator token you control.
-SVar:DBDouble:DB$ MultiplyCounter | Defined$ Targeted | CounterType$ P1P1 | SpellDescription$ SpellDescription | SpellDescription$ Double the number of +1/+1 counters on it.
+SVar:DBDouble:DB$ MultiplyCounter | Defined$ Targeted | CounterType$ P1P1 | SpellDescription$ Double the number of +1/+1 counters on it.
```
Northmoc, line 8: probably `StackDescription$ SpellDescription` was what it was supposed to be

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
Northmoc: Yes, seems this would work. The param name is just a bit unintuitive

forge-gui/res/cardsfolder/upcoming/tempest_hart_scan_the_clouds.txt
```diff
@@ -14,6 +14,6 @@ ALTERNATE
 Name:Scan the Clouds
 ManaCost:1 U
 Types:Instant  Adventure
```
Northmoc, line 16: looks like a spacing issue here too

forge-gui/res/cardsfolder/v/vodalian_war_machine.txt
```diff
@@ -3,9 +3,11 @@ ManaCost:1 U U
 Types:Creature Wall
 PT:0/4
 K:Defender
-A:AB$ Effect | Cost$ tapXType<1/Merfolk> | RememberCostCards$ True | StaticAbilities$ CanAttack | SpellDescription$ CARDNAME can attack this turn as though it didn't have defender.
+T:Mode$ AbilityCast | ValidCard$ Card.Self | Static$ True | Execute$ TrigStore
+SVar:TrigStore:DB$ Pump | RememberObjects$ TriggeredSpellAbility>TappedCards
```
Northmoc, line 7: shouldn't it also forget at end of turn?

Say you activate some abilities on your turn, then on opponent's turn they kill the VWM... in this case it would destroy the tapped merfolk even though they weren't tapped "this turn", right?

forge-gui/res/cardsfolder/e/elvish_vatkeeper.txt
tool4ever: I checked and the autobuilt looked fine

forge-gui/res/cardsfolder/v/vodalian_war_machine.txt
tool4ever: it has static phase cleanup trigger (still problematic but it is what it is)

Northmoc: heh, I just missed that in the soup

forge-gui/res/cardsfolder/i/immobilizer_eldrazi.txt
Agetian: Yeah, CARDNAME can't block is a tricky case (as outlined in #3790) - the AI can only detect inability to block through this pseudo-keyword. The (proper?) static ability application of can't block currently causes issues for the AI that I haven't yet found a way to resolve. Changing all the cards to a CantBlock static will require a certain rework of the AI since otherwise it'll break AI determining whether something can or can't block completely.

Agetian: (Addendum: assistance is definitely needed with that, as I've tried a few different things in the past but none of them did the trick :/ )

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
tool4ever: Alright, guess I'll just switch back to the previous solution to be on the safe side
no big deal anyway

Agetian (approved):

PR #3898, opened 2023-10-14
Shard / Wedge Order Updates

edg444:
Changed order of mana symbols for some cards that had the outdated order.


Agetian:
These cards should not be in "upcoming",

edg444:
Closed to place cards in proper folders.

PR #3899, opened 2023-10-14
Card cleanup

edg444:
Reordered wedge and shard mana symbols for 23 cards. Updated oracle text for Thunderhawk Gunship.


tool4ever (approved):
thanks

Issue #1922, opened 2022-11-18
Source information can be invalidated when static abilities involved


tool4ever:
1. Cast _World Breaker_ with a target for its cast trigger
2. Counter the spell
3. Observe the ability turns green:
![grafik](https://user-images.githubusercontent.com/8506892/202693310-47d19e03-06d1-44b2-a722-2516905487ad.png)
4. The Ability can now be retargeted with e.g. _Bolt Bend_ onto _Gaea's Revenge_ even though it has "can't be the target of nongreen spells or abilities from nongreen sources" (as long as it has one of the required card types)

The problem is the Devoid static gets removed from the old zone card.

Hanmac:
@tool4ever i have seen something similar with Jadzi the MDFC when using her trigger to cast a second version of the same card from the libary, but as a Spell Side

tool4ever:
@Hanmac 
I don't know a better fix than the one I suggested here: #1771

Would also save us a decent amount of remembering across different cards imo

Hanmac:
> @Hanmac I don't know a better fix than the one I suggested here: #1771
> 
> Would also save us a decent amount of remembering across different cards imo

1. i don't understand why it can become green, shouldn't Devoid stop that?
2. i think the trigger on the stack should use LKI for the HostCard, so it doesn't change when the Card is removed from stack?
3. i think what would help here is if i finally have time to implement game-timestamp so we better keep track on if the card should be treated as new object or not

tool4ever:
Since Devoid is done as static if the Card leaves the stack `StaticEffect.remove` will remove colorless and then it won't be applied to the old card again

The trigger can't use LKI by default because it should still see changes to its source while it's on the stack

Hanmac:
@tool4ever then maybe when the card is removed from the stack, it should set the host of the remaining trigger to LKI before the StaticEffect is removed?

tool4ever:
Ugh it's problematic (multiple tests/cards breaking) because lots of effects store their `getHostCard` in a variable at the start.
Now if `GameAction.changeZone` switches host to LKI when returning to effect code any Remember$ params will use the old host and 💣 

So that approach would probably need a pretty big refactor and then still leave a lingering risk of wrong object reusage...

@friarsol
do you have a good idea?

tehdiplomat:
This sounds similar to other things I've seen where the card really needs to be a Card + Timestamp. And if Card + Timestamp isn't actively in a zone, then we need to go fetch it in LKI.

PR #3912, opened 2023-10-16
AbilityUtils: add Emerged for Adipose Offering

### LINKED ISSUE #3888: Adipose Offspring
Emerge Cost needs to be stored in the Ability like it was done for Convoked cards

Hanmac:
Closes #3888 

Does AI need some tests or Logic for this?

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtil.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtil.java
@@ -923,7 +923,7 @@ public class ComputerUtil {
         }
 
         for (int i = 0; i < max; i++) {
-            Card c = chooseCardToSacrifice(remaining, ai, destroy);
+            Card c = chooseCardToSacrifice(source, remaining, ai, destroy);
             remaining.remove(c);
             if (c != null) {
                 sacrificed.add(c);
@@ -938,7 +938,7 @@ public class ComputerUtil {
     }
 
     // Precondition it wants: remaining are reverse-sorted by CMC
-    private static Card chooseCardToSacrifice(final CardCollection remaining, final Player ai, final boolean destroy) {
+    private static Card chooseCardToSacrifice(final SpellAbility source, CardCollection remaining, final Player ai, final boolean destroy) {
         // If somehow ("Drop of Honey") they suggest to destroy opponent's card - use the chance!
         for (Card c : remaining) { // first compare is fast, second is precise
             if (ai.isOpponentOf(c.getController()))
@@ -951,6 +951,7 @@ public class ComputerUtil {
                 return indestructibles.get(0);
             }
         }
+
         for (int ip = 0; ip < 6; ip++) { // priority 0 is the lowest, priority 5 the highest
             final int priority = 6 - ip;
             for (Card card : remaining) {
@@ -958,6 +959,11 @@ public class ComputerUtil {
                     return card;
                 }
             }
+        }
+
+        if (source.isEmerge() || source.isOffering()) {
+            // don't sac when cost wouldn't be reduced
+            remaining = CardLists.filter(remaining, CardPredicates.greaterCMC(1));
         }
 
         Card c = null;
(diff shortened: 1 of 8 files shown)


forge-game/src/main/java/forge/game/cost/CostPayment.java
```diff
@@ -336,16 +336,18 @@ public static void handleOfferings(final SpellAbility sa, boolean test, boolean
             offering.setUsedToPay(false);
             if (costIsPaid && !test) {
                 sa.getHostCard().getGame().getAction().sacrifice(offering, sa, false, table, null);
```
tool4ever, line 338: X value will be wrong if toughness was changed by static

- A: still my favorite - don't undo statics on old card when it changes zone (#1922)
- B: check if `changeZoneLKIInfo` can provide it
- C: somehow reuse LKI from ChangesZone trigger
- D: switch value to extra LKI here which feels excessive?

Hanmac: I'm going to try to use B, and try to reuse the LKI from ChangeZone

Agetian:
Hmm, not sure off the top of my head if the AI will need some adaptation, probably yes, I'll need to experiment with the card and see how it works and what needs doing for the AI. Sadly, might need to get back to it closer to the weekend since the working week has started and it's pretty crazy :)

Hanmac:
i think it might not need much AI
i mean i would prefer it to sacrifice a a creature with 2 or more toughness

and Emerge might need AI to not sacrifice cards without mana values?

forge-game/src/main/java/forge/game/cost/CostPayment.java
Hanmac: i used changeZoneLKIInfo

i tried to use ability.getLastStateBattlefield but it wasn't set at the time
and passing LKI to the function would have been reset later anyway

or your opinion?

tool4ever: should be safe 🤷‍♂️

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -960,6 +962,11 @@ private static Card chooseCardToSacrifice(final CardCollection remaining, final
             }
         }
 
+        if (source instanceof SpellPermanent && (source.getHostCard().hasKeyword(Keyword.EMERGE) || source.getHostCard().hasKeyword(Keyword.OFFERING))) {
```
Hanmac, line 965: shouldn't the check be source.isEmerge or something because it should only apply to the Emerge Spell?

tool4ever: yea somehow I thought they didn't exist

Hanmac:
would `SVar:AIPreference:SacCost$Creature.toughnessGE2` work for Emerge Cost? so for Adipose Offering, it should only sacrifice a creature where the Toughness is 2 or more?

Hanmac:
> 1. that gets parsed by `getCardPreference` which seems to be only used in `chooseSacrificeType` currently
> 
>     2. it seems the way `chooseCardToSacrifice` is hooked up the AI does always expect it to return a choice. which means it would now cheat and cast emerged without sacrifice (when it only has 0 cmc)
>        I don't know if I can support 1. easily but I need to fix 2. at least

yeah, that is probably something that @Agetian needs to look into

Agetian:
If you provide me with a test case where this is relevant, I can try to take a look (maybe closer to the weekend) unless one of you guys is faster :D

Hanmac:
> If you provide me with a test case where this is relevant, I can try to take a look (maybe closer to the weekend) unless one of you guys is faster :D

@Agetian the AI should think this:
* what is the difference between the Emerge Cost and the normal Mana Cost?
  * the Eldrazi ones where 1 cheaper to Emerge, but with more Colors
  * for the Aidipose Offering, the difference is 2 with the same colors
* now the AI would like to do sacrifice a creature with ManaValue >= 2 so the discount matters
* Now the AI should look at Toughness, it should be at Minimum 2 or more, otherwise it would be better to just cast it for its normal cost

or maybe check for a `Creature.cmcGE2+toughnessGE2` before trying to Emerge anyway?

tool4ever (approved):
should be ok for future iteration

PR #3916, opened 2023-10-16
MagicStack: don't leave old targets on card ability

tool4ever:
image


forge-gui/res/cardsfolder/upcoming/clockwork_droid.txt
```diff
@@ -3,7 +3,7 @@ ManaCost:2
 Types:Artifact Creature Robot
 PT:3/1
 S:Mode$ OptionalAttackCost | ValidCard$ Card.Self | Trigger$ TrigEffect | Cost$ Exert<1/CARDNAME> | Description$ You may exert CARDNAME as it attacks. When you do, it can't be blocked this turn and you scry 1. (An exerted creature won't untap during your next untap step. To scry 1, look at the top card of your library. You may put that card on the bottom.)
-SVar:TrigEffect:DB$ Effect | StaticAbilities$ Unblockable | SubAbility$ DBScry
+SVar:TrigEffect:DB$ Effect | StaticAbilities$ Unblockable | SubAbility$ DBScry | SpellDescription$ When you do, it can't be blocked this turn and you scry 1.
```
Hanmac, line 6: Isn't this more like a TriggerDescription?

tool4ever: Hmn, this is the way it gets picked up currently
Unfortunately I don't quite remember if there was a special reason why I wrote it like that in `CostPartWithTrigger` back then :/

Hanmac (approved):

PR #3921, opened 2023-10-18
YWOE high_fae_prankster.txt and support

Northmoc:


Hanmac:
The Exchange probably need something different because  of how Exchange Normally works

Northmoc:
> The Exchange probably need something different because of how Exchange Normally works

I wondered about that. Perpetual just makes everything uglier. Hmm

Agetian:
Is any more work needed on this PR?

tool4ever:
It's hard to tell with Alchemy
But one thing I like less now compared to the previous version is that it will add two newPT entries to each of the cards (effect + static) 🤔

Northmoc:
OK, here's an implementation where perpetual might move away from statics living on effects... I kind of like it
image

Northmoc:
So proof of concept at least works for refactoring Perpetual out of being Effects... I don't think adding statics, etc. will be out of reach

Surely some details missing yet

@Simisays here are some examples of how we might re-script

tool4ever:
Ugh I'll try to think about this with my best (or rather neutral) judgement

But honestly if this is mostly for cosmetic reasons there should be easier/cleaner options to explore that let the GUI handle the responsibility (like just filtering out perpetual effects from command display) 😕

Northmoc:
In my mind, the refactor would absolutely need to have more benefit that just cleaning up the command zone... that was just the best way to prove concept with a screenshot.

Other benefits I see here are:
 - card scripts become tidier and more logical - a pump is a pump, just with a perpetual duration (at the expense of more code) - [High Fae Prankster](https://scryfall.com/card/ywoe/20/high-fae-prankster) is the main example here so far as there's no good clean way to handle the power exchange without two statics, lots of remembering, etc.
 - future shenanigans in paper that might parallel Alchemy designs will already have support in the appropriate effects instead of all being handled by statics on Effects (good example is "base power" exchange from High Fae Prankster...)
 - similarly, actually using appropriate effects (as opposed to statics) leads to improvements in those effects. PumpZone in PumpEffect and PumpAllEffect probably needed to be cleaned up anyway, but [Accident-Prone Apprentice](https://scryfall.com/card/ywoe/16/accident-prone-apprentice-amphibian-accident) ability being handled by a static on an Effect meant PumpEffect didn't get updated.

I don't think refactoring in stages would break existing perpetual EffectEffect scripts, so work could be done in batches, making review easier.

Also worth noting is that while I don't think most of us care for Alchemy much, it doesn't seem Alchemy or perpetual are going away any time soon.

Issues I see with a refactor:
 - 100+ card scripts that would need a look
 - multiple effects that would need new code

I'm cautiously in favor of the refactor at this point, but I'd love someone to talk me out of it 😆

Northmoc:
Another benefit to removing Effects is that we would no longer need to hide or update lists of remembered cards in hidden zones

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4327,6 +4327,42 @@ public final boolean hasIntensity() {
...
+        }
+    }
+
+    public final void setPerpetual(Map<String, List<Object>> map) {
+        for (String key : map.keySet()) {
+            addPerpetual(key, map.get(key));
+        }
+    }
+
+    public final void generatePerpetual(String type, Integer power, Integer toughness, long timestamp) {
```
tool4ever, line 4357: I think `String type, long timestamp, Object... params` so it can be reused by all effects

tool4ever:
Thanks for the detailed explanation and sorry it took so long
I think we could move forward with the approach soon, should also help with #4135

forge-game/src/main/java/forge/game/card/Card.java
Northmoc: I reworked this a fair amount

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
```diff
@@ -79,16 +84,38 @@ public static void doAnimate(final Card c, final SpellAbility sa, final Integer
         }
 
         if (!addType.isEmpty() || !removeType.isEmpty() || addAllCreatureTypes || !remove.isEmpty()) {
```
Northmoc, line 86: Right now this check always passes, so if perpetual = true it builds a useless Pair

I think it's because addType always isn't empty because of the incomplete CardType?

Northmoc: Is it the `!` that should be removed?
image

tool4ever: yea I think so

Northmoc: Upon removing the `!` it is behaving more as I would expect it to 😁

forge-gui/res/cardsfolder/f/flames_of_moradin.txt
```diff
@@ -1,11 +1,12 @@
...
-SVar:TrigSac:Mode$ Phase | ValidPlayer$ You | Phase$ End of Turn | Execute$ TrigSacrifice | TriggerZones$ Battlefield | TriggerDescription$ Sacrifice CARDNAME at you end step.
+A:SP$ Destroy | ValidTgts$ Artifact | TargetMin$ 0 | TargetMax$ 3 | TgtPrompt$ Select up to three target artifacts | SubAbility$ DBConjure | SpellDescription$ Destroy up to three target artifacts.
+SVar:DBConjure:DB$ MakeCard | Conjure$ True | DefinedName$ Targeted.nonToken | Zone$ Hand | RememberMade$ True | SubAbility$ DBAnimate | SpellDescription$ Conjure a duplicate of each nontoken artifact destroyed this way into your hand.
+SVar:DBAnimate:DB$ Animate | Duration$ Perpetual | Defined$ Remembered | staticAbilities$ PerpAltCost | Triggers$ PhaseTrig | SubAbility$ DBCleanup | StackDescription$ SpellDescription | SpellDescription$ The duplicates perpetually gain "You may pay {R} rather than pay this spell's mana cost" and "at the beginning of your end step, sacrifice this artifact."
```
Northmoc, line 6: This is the most confusing thing – PhaseTrig will NOT go off. Test never even seems to run. Any ideas?

The other one with `Triggers$` – "Antique Collector" works just fine. It is AnimateAll instead of Animate but the important business is happening in AnimateEffectBase and Card anyway, so I don't think that's the issue.

Any insight? I've pored over the lines mulitple times and attempted debugging and can't find why the trigger seems to be a dud.

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4386,6 +4386,52 @@ public final boolean hasIntensity() {
...
+    public final boolean hasPerpetual() {
+        return !perpetual.isEmpty();
+    }
+    public final List<Pair<String, Map<String, Object>>> getPerpetual() {
+        return perpetual;
+    }
+
+    public final void addPerpetual(Pair<String, Map<String, Object>> p) {
+        perpetual.add(p);
+        executePerpetual(p);
```
tool4ever, line 4399: if you don't execute it here you don't need all the "else" in the effects?

forge-game/src/main/java/forge/game/ability/effects/PumpEffect.java
```diff
@@ -70,12 +74,25 @@ private static void applyPump(final SpellAbility sa, final Card applyTo,
...
+                gameCard.addPerpetual(Pair.of("PTBoost", params));
+            } else gameCard.addPTBoost(a, d, timestamp, 0);
             redrawPT = true;
         }
 
         if (!kws.isEmpty()) {
-            gameCard.addChangedCardKeywords(kws, Lists.newArrayList(), false, timestamp, 0);
+            if (perpetual) {
+                Map <String, Object> params = new HashMap<>();
+                params.put("AddKeywords", keywords);
```
tool4ever, line 90: kws

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
```diff
@@ -158,6 +158,7 @@ public static void doAnimate(final Card c, final SpellAbility sa, final Integer
         final List<Trigger> addedTriggers = Lists.newArrayList();
         for (final String s : triggers) {
             final Trigger parsedTrigger = TriggerHandler.parseTrigger(AbilityUtils.getSVar(sa, s), c, false, sa);
+            if (perpetual) parsedTrigger.setSpawningAbility(sa);
```
Northmoc, line 161: With this line the previously broken EOT triggers on Goblin Trapfinder and Flames of Moradin will trigger when expected...

BUT the Sacrifice always whiffs. Debugging in SacrificeEffect shows that the zone of the card is the Stack somehow. The GUI still shows the card on the Battlefield.

Does the trigger going to the stack somehow move the host card of the trigger to the stack as well??

tool4ever: I don't think you can do that (making them look like delayed triggers to the engine)
they're gonna trigger from any zone now...

forge-gui/res/cardsfolder/f/flames_of_moradin.txt
```diff
@@ -4,7 +4,7 @@ Types:Sorcery
...
 SVar:DBConjure:DB$ MakeCard | Conjure$ True | DefinedName$ Targeted.nonToken | Zone$ Hand | RememberMade$ True | SubAbility$ DBAnimate | SpellDescription$ Conjure a duplicate of each nontoken artifact destroyed this way into your hand.
 SVar:DBAnimate:DB$ Animate | Duration$ Perpetual | Defined$ Remembered | staticAbilities$ PerpAltCost | Triggers$ PhaseTrig | SubAbility$ DBCleanup | StackDescription$ SpellDescription | SpellDescription$ The duplicates perpetually gain "You may pay {R} rather than pay this spell's mana cost" and "at the beginning of your end step, sacrifice this artifact."
-SVar:PerpAltCost:Mode$ Continuous | EffectZone$ All | MayPlay$ True | MayPlayAltManaCost$ R | MayPlayDontGrantZonePermissions$ True | Affected$ Card.Self | AffectedZone$ Hand,Graveyard,Library,Exile,Command | Description$ You may pay {R} rather than pay this spell's mana cost.
+SVar:PerpAltCost:Mode$ Continuous | CharacteristicDefining$ True | AddKeyword$ Alternative Cost:R | Description$ You may pay {R} rather than pay this spell's mana cost.
```
tool4ever, line 7: MayPlay would be better

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
Northmoc: OK, back to the drawing board

forge-gui/res/cardsfolder/f/flames_of_moradin.txt
Northmoc: This has been bugging me... we have two different ways of doing essentially the same thing?

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
Northmoc: Probably the delayed trigger thing is the reason the engine assumed the source was not on the battlefield anyway

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
```diff
@@ -132,6 +159,7 @@ public static void doAnimate(final Card c, final SpellAbility sa, final Integer
         final List<Trigger> addedTriggers = Lists.newArrayList();
         for (final String s : triggers) {
             final Trigger parsedTrigger = TriggerHandler.parseTrigger(AbilityUtils.getSVar(sa, s), c, false, sa);
+            if (perpetual) parsedTrigger.setCardState(c.getCurrentState());
```
Northmoc, line 162: I'm not sure if this is necessary? Flames of Moradin and Goblin Trapfinder EOT phase triggers still don't work with it

Without it, cardstate at this point is Flames/Trapfinder and NOT the card being animated to.

But again, not sure it really matters

forge-gui/res/cardsfolder/b/bind_to_secrecy.txt
```diff
@@ -1,15 +1,12 @@
...
-SVar:DBConjure:DB$ MakeCard | Conjure$ True | TgtPrompt$ Select target creature card in an opponent's graveyard | ValidTgts$ Creature.OppOwn+inZoneGraveyard | TgtZone$ Graveyard | DefinedName$ Targeted | Zone$ Hand | RememberMade$ True | SubAbility$ DBEffect | SpellDescription$ Conjure a duplicate of target creature card in an opponent's graveyard into your hand. It perpetually gains "You may spend mana as though it were mana of any color to cast this spell."
-SVar:DBEffect:DB$ Effect | RememberObjects$ Remembered | StaticAbilities$ PerpetualAbility | Duration$ Permanent | Triggers$ Update | Name$ Bind to Secrecy's Perpetual Effect | SubAbility$ DBCleanup | SpellDescription$ It perpetually gains "You may spend mana as though it were mana of any color to cast this spell."
-SVar:PerpetualAbility:Mode$ Continuous | AddStaticAbility$ SpendAnyMana | Affected$ Card.IsRemembered | EffectZone$ Command | AffectedZone$ Battlefield,Hand,Graveyard,Exile,Stack,Library,Command | Description$ The conjured card perpetually gains "You may spend mana as though it were mana of any color to cast this spell."
+A:SP$ Charm | Choices$ CounterNonCreature,DBConjure | CharmNum$ 1 | SubAbility$ DBDraft
```
Northmoc, line 4: feels weird to send this to SubAbility just for the CardDetail and then to also need to send it from each Choice

I think this is the only MtG card that does a Charm and then chains another effect so it may not be worth the hassle of monkeying with the wiring for now

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
```diff
@@ -183,11 +210,17 @@ public String getDescription() {
         if (removeAll || removeNonManaAbilities
                 || !addedAbilities.isEmpty() || !removedAbilities.isEmpty() || !addedTriggers.isEmpty()
                 || !addedReplacements.isEmpty() || !addedStaticAbilities.isEmpty()) {
-            c.addChangedCardTraits(addedAbilities, removedAbilities, addedTriggers, addedReplacements,
-                    addedStaticAbilities, removeAll, removeNonManaAbilities, timestamp, 0);
+            final CardTraitChanges ctc = new CardTraitChanges(addedAbilities, removedAbilities, addedTriggers, 
+                addedReplacements, addedStaticAbilities, removeAll, removeNonManaAbilities);
+            c.addChangedCardTraits(ctc, timestamp, 0);
```
tool4ever, line 215: it's debatable if storing the ctc too would improve accuracy
otherwise this can be cleaned up again

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4386,6 +4385,52 @@ public final boolean hasIntensity() {
...
+            addChangedCardTraits((CardTraitChanges) params.get("ChangedTraits"), 
+                (long) params.get("Timestamp"), (long) 0);
+        }
+    }
+
+    public final void setPerpetual(final Card oldCard) {
+        final List<Pair<String, Map<String, Object>>> perp = oldCard.getPerpetual();
+        perpetual = perp;
+        for (Pair <String, Map<String, Object>> p : perp) {
+            if (p.getKey().equals("Abilities")) {
```
tool4ever, line 4426: not sure why you added this here now while leaving the older part?

also you're still copying them with the old host, that's just wrong

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -2154,6 +2154,10 @@ public static int xCount(final Card c, final String s, final CardTraitBase ctb)
             return doXMath(c.getRememberedCount(), expr, c, ctb);
         }
 
+        if (sq[0].startsWith("ChosenSize")) {
```
Northmoc, line 2157: I will make use of this in Spara's Bodyguard

image

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
```diff
@@ -79,15 +84,36 @@ public static void doAnimate(final Card c, final SpellAbility sa, final Integer
         }
 
         if (!addType.isEmpty() || !removeType.isEmpty() || addAllCreatureTypes || !remove.isEmpty()) {
+            if (perpetual) {
+                Map <String, Object> params = new HashMap<>();
+                params.put("AddTypes", addType);
+                params.put("RemoveTypes", removeType);
+                params.put("RemoveXTypes", remove);
+                params.put("Timestamp", timestamp);
+                c.addPerpetual(Pair.of("Types", params));
```
tool4ever, line 93: for readability there's no point in wrapping them in pairs on all these calls
doing it in `addPerpetual` instead would be even easier

forge-gui/res/cardsfolder/a/arming_gala.txt
```diff
@@ -1,9 +1,6 @@
...
-SVar:PerpetualP1P1:Mode$ Continuous | Affected$ Card.IsRemembered | AddPower$ 1 | AddToughness$ 1 | EffectZone$ Command | AffectedZone$ Battlefield,Hand,Graveyard,Exile,Stack,Library,Command | Description$ At the beginning of your end step, creatures you control and creature cards in your hand, library, and graveyard perpetually get +1/+1.
-SVar:Update:Mode$ ChangesZone | Origin$ Any | Destination$ Any | Static$ True | ValidCard$ Card.IsRemembered | Execute$ DBUpdate
-SVar:DBUpdate:DB$ UpdateRemember
-Oracle:At the beginning of your end step, creatures you control and creature cards in your hand, library, and graveyard perpetually get +1/+1.
\ No newline at end of file
+T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You | TriggerZones$ Battlefield | Execute$ TrigPerpetualP1P1 | TriggerDescription$ At the beginning of your end step, creatures you control and creature cards in your hand, library, and graveyard perpetually get +1/+1.
+SVar:TrigPerpetualP1P1:DB$ PumpAll | ValidCards$ Creature.YouCtrl,Creature.YouOwn | PumpZone$ Battlefield,Hand,Library,Graveyard | NumAtt$ 1 | NumDef$ 1 | Duration$ Perpetual
```
tool4ever, line 5: bad valid?

forge-game/src/main/java/forge/game/trigger/TriggerHandler.java
```diff
@@ -566,6 +567,19 @@ else if (sa instanceof AbilitySub || !sa.hasParam("Cost") || (sa.getPayCosts() !
         if (regtrig.hasParam("OneOff") && host.isImmutable()) {
             host.getController().getZone(ZoneType.Command).remove(host);
         }
+        if (regtrig.hasParam("LosePerpetual")) {
```
tool4ever, line 570: Umh...why would you put this in TriggerHandler?

this is something that happens as part of the effect resolving...

Northmoc: I just had "OneOff" in my mind from recent memory. 😅 

I'll find a better place for it

tool4ever (approved):

Issue #2289, opened 2023-01-20
AttachEffect QoL


tool4ever:
i checked AttachEffect again, while it works, we should probably Enhance AttachEffect

* when giving list of `Object$` and `Optional$ True` and maybe `Defined$` is used (it does fall back to Targeted anyway when using `ValidTgts`)  Filter and Pick the Aura and Equipments via Multiselect, and then skip the later optional check.
* with that it would be A faster for Player to select which ones to attach to target
* You don't get asked for stuff you can't equip to target like Equipment to a Player

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/pull/2273#discussion_r1082312467_

Hanmac:
* Case A: Multiple Equipment, Not Optional, Attach to itself: Vulshok Battlemaster
  >  If an Equipment can’t equip Vulshok Battlemaster, it isn’t attached to the Battlemaster, and it doesn’t become unattached (if it’s attached to a creature). 
* Case B: Multiple Equipment and Auras, Not Optional, Attach to Choice: Fumble
  >  The Auras and Equipment that you gain control of must all be attached to one creature that they can all legally be attached to. If they can’t all legally be attached to one creature, they all remain on the battlefield unattached (and Auras among them will be put into their owners’ graveyards). 

* Case C: Multiple Equipment, Any Number Optional, Attach to Target: Ardenn, Intrepid Archaeologist
  > No Ruling for this. but this should instead of asking each Equipment individually, it should use a Select Entities and filter the ones that can be attached to Target

* Case D: Multiple Equipment, Any Number Optional, Attach to Choice: Armed and Armored
  > No Ruling for this, i don't know if we can Keep it as One Effect or if it should be two, or if i can flag it with something that the Dwarf is chosen first before the Equipment?

* Case E: Multiple Equipment, Any Target Number only FULL Optional, Attach to itself: Armory Automaton
  > You can choose to target no Equipment if you don’t want to attach any to Armory Automaton. Because the ability is optional, you also don’t have to attach the Equipment it targets as you resolve the ability. However, you must either attach all of the target Equipment or attach none of them. (2016-11-08)
  >
  > because for this we can't use Optional on the Effect, but need to use OptionalTrigger?

PR #3927, opened 2023-10-19
YWOE last few scripts + various tweaks

Northmoc:
bring Nahiri the Lithomancer in line with the other Nahiri with the same ability (to catch corner case of doubling tokens)
help DamageAll stack description output be more obvious when missing params, help punctuation

### DIFF
--- a/forge-gui/res/cardsfolder/n/nahiri_the_lithomancer.txt
+++ b/forge-gui/res/cardsfolder/n/nahiri_the_lithomancer.txt
@@ -3,10 +3,9 @@ ManaCost:3 W W
 Types:Legendary Planeswalker Nahiri
 Loyalty:3
 Text:CARDNAME can be your commander.
-A:AB$ Token | Cost$ AddCounter<2/LOYALTY> | Planeswalker$ True | TokenScript$ w_1_1_kor_warrior | RememberTokens$ True | SubAbility$ DBChooseToken | SpellDescription$ Create a 1/1 white Kor Warrior creature token. You may attach an Equipment you control to it.
-SVar:DBChooseToken:DB$ ChooseCard | DefinedCards$ Remembered | Mandatory$ True | ChoiceTitle$ Choose a token | SubAbility$ DBAttach | StackDescription$ None
-SVar:DBAttach:DB$ Attach | Optional$ True | Choices$ Equipment.YouCtrl | ChoiceTitle$ Choose an Equipment you control | Defined$ ChosenCard | SubAbility$ DBCleanup | StackDescription$ None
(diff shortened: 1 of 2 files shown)


forge-gui/res/cardsfolder/n/nahiri_the_lithomancer.txt
```diff
@@ -3,9 +3,10 @@ ManaCost:3 W W
 Types:Legendary Planeswalker Nahiri
 Loyalty:3
 Text:CARDNAME can be your commander.
-A:AB$ Token | Cost$ AddCounter<2/LOYALTY> | Planeswalker$ True | TokenAmount$ 1 | TokenScript$ w_1_1_kor_soldier | TokenOwner$ You | RememberTokens$ True | SubAbility$ DBAttach | SpellDescription$ Create a 1/1 white Kor Soldier creature token. You may attach an Equipment you control to it.
-SVar:DBAttach:DB$ Attach | Optional$ True | Choices$ Equipment.YouCtrl | ChoiceTitle$ Choose an Equipment you control | Defined$ Remembered | SubAbility$ DBCleanup | StackDescription$ None
-SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+A:AB$ Token | Cost$ AddCounter<2/LOYALTY> | Planeswalker$ True | TokenScript$ w_1_1_kor_warrior | RememberTokens$ True | SubAbility$ DBChooseToken | SpellDescription$ Create a 1/1 white Kor Warrior creature token. You may attach an Equipment you control to it.
+SVar:DBChooseToken:DB$ ChooseCard | DefinedCards$ Remembered | Mandatory$ True | ChoiceTitle$ Choose a token | SubAbility$ DBAttach | StackDescription$ None
```
tool4ever, line 7: what does the corner case look like?
this seems to have worked already before

Northmoc: I think it was in the case of Doubling Season and similar, the player should have ability to choose which token gets an attachment. @Hanmac may remember better

Not sure why we didn't bring Nahiri, Heir of Ancients and this to parity before 🤷‍♂️ they definitely should be the same one way or the other

tool4ever: yea and when I tested that with the old version it already lets you choose the token

Hanmac: Imo it would be better to have the choice inside the Attach

Hanmac: There is my thoughts about refactoring 
https://github.com/Card-Forge/forge/issues/2289#issuecomment-1399565687

tool4ever (approved):

PR #3934, opened 2023-10-20
CantHaveKeyword can prevent effect creation

tool4ever:
> For example, if you control Archetype of Courage, a spell cast by an opponent that gives creatures they control first strike wouldn’t cause the creatures to have first strike, even if later in the turn Archetype of Courage left the battlefield.

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -4683,7 +4683,17 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
         if (keywords != null) {
             long idx = 1;
             for (String kw : keywords) {
-                kws.add(getKeywordForStaticAbility(kw, staticId, idx));
+                // CR 113.11
+                boolean canHave = true;
+                for (Keyword cantKW : getCantHaveKeyword()) {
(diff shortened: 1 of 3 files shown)


Hanmac:
You should move the CanGainKeyword check into a function on Card


Or filter inside addChangedCardKeywords

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4683,8 +4683,18 @@ public final void addChangedCardKeywords(final List<String> keywords, final List
...
+                boolean canHave = true;
+                for (Keyword cantKW : getCantHaveKeyword()) {
+                    if (kw.startsWith(cantKW.toString())) {
+                        canHave = false;
+                        break;
+                    }
+                }
+                if (canHave) {
+                    kws.add(getKeywordForStaticAbility(kw, staticId, idx));
+                    idx++;
```
Hanmac, line 4696: i think the idx should still be go up

for example if something would give 3 Keywords,
and the first one is "can't have", then the following keywords would get different ids than if the first one is not negated.

that could cause more objects that not used

tool4ever: good point

Hanmac (approved):

PR #3935, opened 2023-10-20
add AI forced playing Living End decks

kevlahnota:
https://github.com/Card-Forge/forge/assets/9781539/22af49ac-1543-49cb-b910-1d1900db3f18

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -1572,6 +1572,8 @@ public class AiController {
                 return spellAbility instanceof LandAbility || (spellAbility.getHostCard() != null && ComputerUtilCard.isCardRemAIDeck(spellAbility.getHostCard()));
             }
         });
+        //update LivingEndPlayer
+        player.setHasLivingEnd(CardLists.filter(player.getZone(ZoneType.Library).getCards(), c-> "Living End".equalsIgnoreCase(c.getName())).size() > 0);
 
         SpellAbility chosenSa = chooseSpellAbilityToPlayFromList(saList, true);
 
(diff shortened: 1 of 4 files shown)


forge-game/src/main/java/forge/game/Game.java
```diff
@@ -265,6 +265,7 @@ public Game(Iterable<RegisteredPlayer> players0, GameRules rules0, Match match0,
         for (RegisteredPlayer psc : players0) {
             IGameEntitiesFactory factory = (IGameEntitiesFactory)psc.getPlayer();
             Player pl = factory.createIngamePlayer(this, plId++);
+            pl.setHasLivingEnd(psc.getDeck().getMain().countByName("Living End", true) > 0);
```
tool4ever, line 268: Should probably be part of AiController instead

tehdiplomat, line 268: Would love to see this defined in the Deck file so it could potentially be reused for different deck strategies. But maybe we do that in the next deck.

kevlahnota: I think we cant do that on LDA generated decks since they are not saved, gonna add the Living End check on AIConroller I guess (like @tool4ever suggested)

Agetian: Agreed

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -168,6 +168,7 @@ public class Player extends GameEntity implements Comparable<Player> {
     private int maxHandSize = 7;
     private int startingHandSize = 7;
     private boolean unlimitedHandSize = false;
+    private boolean hasLivingEnd = false;
```
tool4ever, line 171: move this to AiController too to keep engine code clean?

forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -1572,6 +1572,8 @@ public boolean apply(final SpellAbility spellAbility) { //don't include removedA
                 return spellAbility instanceof LandAbility || (spellAbility.getHostCard() != null && ComputerUtilCard.isCardRemAIDeck(spellAbility.getHostCard()));
             }
         });
+        //update LivingEndPlayer
+        player.setHasLivingEnd(CardLists.filter(player.getZone(ZoneType.Library).getCards(), c-> "Living End".equalsIgnoreCase(c.getName())).size() > 0);
```
tool4ever, line 1576: use `Iterables.any` for performance so it can terminate earlier instead of filtering each card by name first

tool4ever (approved):

PR #3937, opened 2023-10-20
Support for Zygon Infiltrator

tool4ever:


forge-gui/res/cardsfolder/d/dermotaxi.txt
```diff
@@ -3,7 +3,7 @@ ManaCost:2
 Types:Artifact Vehicle
 PT:0/0
 K:ETBReplacement:Other:Imprint
-SVar:Imprint:DB$ ChangeZone | Imprint$ True | ChangeType$ Creature | ChangeNum$ 1 | Origin$ Graveyard | Destination$ Exile | Mandatory$ True | Hidden$ True | Chooser$ You | SpellDescription$ Imprint - As CARDNAME enters the battlefield, exile a creature card from a graveyard.
+SVar:Imprint:DB$ ChangeZone | Imprint$ True | ChangeType$ Creature.NotDefinedReplacedSimultaneousETB | ChangeNum$ 1 | Origin$ Graveyard | Destination$ Exile | Mandatory$ True | Hidden$ True | Chooser$ You | SpellDescription$ Imprint - As CARDNAME enters the battlefield, exile a creature card from a graveyard.
```
Northmoc, line 6: Seems unfortunate to have to add that ungainly string to each of these... I guess there isn't another way?

tool4ever: It could/should probably be handled by the engine instead and I'm about to add that for Mill 🧙 
For the other cases I'd like to take a rain check to think about a way that keeps the overhead in the effects to a minimum

Northmoc (approved):
Looks good to me!

Issue #3948, opened 2023-10-23
AI uses abilities of two Roiling Vortex at the same time


MorihiroS:
**Describe the bug**
AI uses the abilities of two Roiling Vortex at the same time, though it is enough to use the ability of one

**To Reproduce**
1. AI casts 2 Roiling Vortex
2. Player uses, for example, Bubbling Cauldron
3. AI uses abilities of two Roiling Vortex 

**Expected behavior**
AI should only use the ability of one of the Roiling Vortex

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-10.19

**Screenshots**

![vortex](https://github.com/Card-Forge/forge/assets/115870280/d2c1a15f-a2cd-4f1c-94f5-e4117c7ba2d0)

Agetian:
Can't reproduce in the latest snapshot (23.10), the AI only uses one Roiling Vortex in this instance. Please retry in the latest version, and if it's still an issue, please dump the game state (using the Dev panel "Save Game State" feature), if possible.

MorihiroS:
@Agetian, tried it on the latest version - you're right, it works as it should.

PR #3952, opened 2023-10-23
Improve several AI routines (Reflection of Kiki-Jiki; Roiling Vortex; Karn, the Great Creator; Ashad, the Lone Cyberman)

### LINKED ISSUE #3814: AI doesn't use the second ability of Karn, the Great Creator
**Describe the bug**
AI doesn't use the second ability of Karn, the Great Creator, only the first, even if it doesn't make sense to do so

**To Reproduce**
1. AI cast Karn, the Great Creator
2. AI uses only the first ability of Karn, the Great Creator

**Expected behavior**
AI use the second ability of Karn, the Great Creator, even in the first place

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.58-SNAPSHOT-09.23

**Additional context**
This is characteristic of all four types of AI

**Screenshots**
Karn

### LINKED ISSUE #3938: AI incorrectly uses 2 Reflection of Kiki-Jiki
**Describe the bug**
AI copies one Reflection of Kiki-Jiki with another and so on for as long as AI has mana. The AI ends up with no mana and with one untapped Reflection of Kiki-Jiki.

**To Reproduce**
1. AI casts 2 Fable of the Mirror-Breaker
2. Both Fable of the Mirror-Breaker return to the battlefield transformed
3. AI uses all mana to copy Reflection of Kiki-Jiki with another Reflection of Kiki-Jiki.

**Expected behavior**
AI shouldn't copy one Reflection of Kiki-Jiki with another one (unless the opponent is trying to destroy/exile another Reflection of Kiki-Jiki).

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-10.19

**Screenshots**
kiki1
kiki2

### LINKED ISSUE #3949: AI doesn't use Roiling Vortex's ability when there are multiple healing triggers
**Describe the bug**
AI doesn't use Roiling Vortex's ability when player has multiple 1 hp healing triggers 

**To Reproduce**
1. AI casts Roiling Vortex
2. Player casts some Blood Artist and/or Zulaport Cutthroat
3. A creature under player control dies
4. AI doesn't use Roiling Vortex's ability

**Expected behavior**
AI uses Roiling Vortex's ability when there are some 1 healing triggers (or even for 1 healing trigger

**Additional context**
AI type - experimental

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-10.19

**Screenshots**

vortex

### LINKED ISSUE #3955: AI doesn't use Roiling Vortex ability in response of using the second ability of Bubbling Cauldron
**Describe the bug**
AI doesn't use Roiling Vortex ability in response of using the second ability of Bubbling Cauldron, although it heals player. But AI uses Roiling Vortex ability in response of using the first ability of Bubbling Cauldron

**To Reproduce**
1. AI casts Roiling Vortex
2. Player casts Bubbling Cauldron and Festering Newt
3. Player activates the second ability of Bubbling Cauldron
4. AI doesn't use Roiling Vortex

**Expected behavior**
AI uses Roiling Vortex ability in response of using the second ability of Bubbling Cauldron

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-10.23

**Screenshots**

vortex

### LINKED ISSUE #3961: AI will readily sacrifice Ashad to his own ability for no benefit whatsoever
**Describe the bug**
The AI will consistently sacrifice Ashad to his own ability if there are no other eligible creatures in play. It consistently does this even if he is the commander, and the artifact it is trying to duplicate is a cheap artifact creature or mana rock. Even worse, since Ashad's removal from play causes the cast artifact to lose Casualty, it's not even duplicated. The AI, in effect, is consistently willing to remove its own commander for no reason.

**To Reproduce**
1. Make a deck with Ashad in it. Make him the commander for the guarantee of him being cast. Ensure the deck has lots of cheap artifacts in it.
2. Put the AI in charge of the deck.
3. Continue playing against the AI until it casts an artifact spell while Ashad is the only creature it has in play.

**Expected behavior**
The AI should never treat any creature that grants Casualty to cast spells as eligible fuel for its own ability. It'd be a waste even if the spell were actually copied. (The spell not being copied because of the sacrifice of the Casualty-granting creature is in line with the rules as intended.)

- OS: macOS Monterey Version 12.5.1
- Browser: Chrome
- Version: 1.6.59-SNAPSHOT-10.19

**Additional context**
I suspect the only reason why Anhelo doesn't have the same problem is that, while he also grants Casualty 2, Anhelo's power is 1. I haven't tested Anhelo, but it'd be interesting to see if he'd have the same issue when accompanied by some kind of non-creature anthem.

Agetian:
Fixes #3814.
Fixes #3938.
Fixes #3949.
Fixes #3955.
Fixes #3961.
Can't reproduce #3948 in the latest snapshot, considering fixed for now (unless proven otherwise).


forge-ai/src/main/java/forge/ai/ability/EffectAi.java
```diff
@@ -211,9 +211,19 @@ public boolean apply(final Card c) {
...
-                return tgtPlayer.isOpponentOf(ai) && tgtPlayer.canGainLife() && topStack.getApi() == ApiType.GainLife;
+                if (tgtPlayer.isOpponentOf(ai) && tgtPlayer.canGainLife()) {
+                    while (topStack != null) {
+                        if (topStack.getApi() == ApiType.GainLife) {
+                            if ("You".equals(topStack.getParam("Defined")) || topStack.isTargeting(tgtPlayer)) {
+                                return true;
+                            }
+                        }
+                        topStack = topStack.getSubAbility();
+                    }
```
Hanmac, line 224: would `findSubAbilityByType` be good enough for this? or is it not good for the corner cases?

you also could check for damage if topStack source has lifelink

Agetian: Ah, a good idea :) Also, checking for combat lifelink is a good idea too.
Not sure if findSubAbilityByType would help make it cleaner here tbh, since we also need to check the primary SA as well... Do you have an idea for how to write it in a clean enough manner checking both the main SA and the subs?

forge-ai/src/main/java/forge/ai/ability/EffectAi.java
```diff
@@ -208,12 +202,40 @@ public boolean apply(final Card c) {
...
+                if (!game.getStack().isEmpty()) {
+                    SpellAbility topStack = game.getStack().peekAbility();
+                    final Player activator = topStack.getActivatingPlayer();
+                    if (activator.isOpponentOf(ai) && activator.canGainLife()) {
+                        while (topStack != null) {
+                            if (topStack.getApi() == ApiType.GainLife) {
+                                if ("You".equals(topStack.getParam("Defined")) || topStack.isTargeting(activator)) {
+                                    return true;
+                                }
+                            } else if (topStack.getApi() == ApiType.DealDamage && topStack.getHostCard().hasKeyword(Keyword.LIFELINK)) {
```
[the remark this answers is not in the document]
Agetian: I guess something like this is possible:
```
                                for (GameEntity target : topStack.getTargets().getTargetEntities()) {
                                    if (ComputerUtilCombat.predictDamageTo(target,
                                            AbilityUtils.calculateAmount(host, topStack.getParam("NumDmg"), topStack), host, false) > 0) {
                                        return true;
                                    }
                                }

```

forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
```diff
@@ -598,6 +598,20 @@ public boolean apply(Card card) {
             }
         }
 
+        // Bail early on Casualty in case there are no cards that would make sense to pay with
+        if (sa.getHostCard().hasKeyword(Keyword.CASUALTY)) {
```
tool4ever, line 602: there is already some logic in `PlayerControllerAi.chooseNumberForKeywordCost`, so that might be a better placement

Agetian: I tried that, but couldn't make it work :/ If I return 0 from there, it happens too late and the AI has already made the decision to go for the spell ability with the Casualty cost applied, which will result in a "AI failed to play..." misplay.

tool4ever: hmn ok :/

PR #3954, opened 2023-10-23
WHO - Doctor's companion and support

### LINKED ISSUE #3504: DoctorWho: Doctor's Companion
> You can have two commanders if the other is the Doctor.

Partner variant

Northmoc:
- closes #3504

### DIFF
--- a/forge-core/src/main/java/forge/card/CardRules.java
+++ b/forge-core/src/main/java/forge/card/CardRules.java
@@ -275,16 +275,51 @@ public final class CardRules implements ICardCharacteristics {
         return false;
     }
 
+    public boolean canBePartnerCommanders(CardRules b) {
+        if (!(canBePartnerCommander() && b.canBePartnerCommander())) {
+            return false;
+        }
+        boolean legal = false;
+        if (hasKeyword("Partner") && b.hasKeyword("Partner")) {
+            legal = true; // normal partner commander
+        }
+        if (getName().equals(b.getPartnerWith()) && b.getName().equals(getPartnerWith())) {
+            legal = true; // paired partner commander
+        }
+        if (hasKeyword("Friends forever") && b.hasKeyword("Friends forever")) {
+            legal = true; // Stranger Things Secret Lair gimmick partner commander
+        }
+        if (hasKeyword("Choose a Background") && b.canBeBackground()
+                || b.hasKeyword("Choose a Background") && canBeBackground()) {
+            legal = true; // commander with background
+        }
+        if (isDoctor() && b.hasKeyword("Doctor's companion")
+                || hasKeyword("Doctor's companion") && b.isDoctor()) {
+            legal = true; // Doctor Who partner commander
+        }
+        return legal;
+    }
+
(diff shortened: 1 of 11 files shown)


Hanmac (approved):

forge-gui/res/cardsfolder/upcoming/donna_noble.txt
Glorax, line 1: "TrigToken" needs to be chanced to "TrigDamage". The script doesn't have any "TrigToken" to reference, but it does have "TrigDamage".

Glorax: Actually testing the card reveals that the fix appears functional. It seems to be acting as intended.

forge-core/src/main/java/forge/card/CardRules.java
```diff
@@ -275,18 +275,53 @@ public boolean canBeCommander() {
...
+        }
+        if (hasKeyword("Friends forever") && b.hasKeyword("Friends forever")) {
+            legal = true; // Stranger Things Secret Lair gimmick partner commander
+        }
+        if (hasKeyword("Choose a Background") && b.canBeBackground()
+                || b.hasKeyword("Choose a Background") && canBeBackground()) {
+            legal = true; // commander with background
+        }
+        if (isDoctor() && b.hasKeyword("Doctor's companion")
+                || hasKeyword("Doctor's companion") && b.isDoctor()) {
```
kevlahnota, line 297: IIRC @tool4ever updated to use `haskeyword(Keyword.key)`, maybe just use `hasKeyword(Keyword.DOCTORS_COMPANION)` and maybe update others too..

Hanmac: Doesn't work
Only Card has `hasKeyword(KeywordEnum)`

CardRules only has `hasKeyword(String)`

Also the KeywordEnum only exist in Forge-Game, not Forge-Core

Hanmac: @tehdiplomat we need to think about it if we really still need Forge-Core or if we should merge this with Forge-Game

forge-gui/res/cardsfolder/upcoming/sarah_jane_smith.txt
```diff
@@ -0,0 +1,10 @@
+Name:Sarah Jane Smith
+ManaCost:1 W
+Types:Legendary Creature Human Detective
+PT:2/1
+T:Mode$ SpellCast | ValidCard$ Card.Historic | ValidActivatingPlayer$ You | TriggerZones$ Battlefield | Execute$ TrigInvestigate | TriggerDescription$ Whenever you cast a historic spell, investigate. This ability triggers only once each turn. (Artifacts, legendaries, and Sagas are historic. To investigate, create a Clue token. It's an artifact with "{2}, Sacrifice this artifact: Draw a card.")
```
tool4ever, line 5: "Only once" not handled

Hanmac (approved):

PR #3963, opened 2023-10-24
WHO: flesh_duplicate.txt and support

Northmoc:


forge-game/src/main/java/forge/game/card/CardFactory.java
```diff
@@ -691,6 +692,23 @@ public static CardCloneStates getCloneStates(final Card in, final Card out, fina
             keywords.addAll(Arrays.asList(sa.getParam("AddKeywords").split(" & ")));
         }
 
+        if (sa.hasParam("AddKeywordsIfNew")) {
+            List<KeywordInterface> inKW = in.getKeywords();
+            for (String k : sa.getParam("AddKeywordsIfNew").split(" & ")) {
```
tool4ever, line 697: This touches a group of corner cases that aren't handlet yet: #2222

Though luckily this is probably the easiest of them
- if a card **only gained** Vanishing this effect should also add to the copy
- Seems only _Maelstrom Djinn_ can do it?

For this you can probably do your check a bit lower inside the state modifications

forge-game/src/main/java/forge/game/card/CardFactory.java
```diff
@@ -688,7 +690,11 @@ public static CardCloneStates getCloneStates(final Card in, final Card out, fina
         }
 
         if (sa.hasParam("AddKeywords")) {
-            keywords.addAll(Arrays.asList(sa.getParam("AddKeywords").split(" & ")));
+            String kwString = sa.getParam("AddKeywords");
+            if (kwString.startsWith("IfNew"))
+                KWifNew = true;
+                kwString = kwString.substring(6);
```
tool4ever, line 696: it would always cut off something

tool4ever (approved):

PR #3977, opened 2023-10-25
Full Control

### LINKED ISSUE #1519: Order of cost reductions can sometimes matter
If you control _Tezzeret, Betrayer of Flesh_ + _Heartstone_ you should be able to activate _Burnished Hart_ for {0}.

Currently it can still be {1} if the Stone ETB after Tezzeret.

But rules say:
> If multiple cost reductions apply, the player may apply them in any order.

@Hanmac 
Always choosing order might be annoying? What you think about just sorting them so those with `MinMana$` are always done first?

tool4ever:
One of the few good ideas I've seen from some other experimental project 😜 

While I do suck at GUI stuff I feel like clicking/tapping your player portrait actually makes it work quite smoothly: hidden enough to not confuse new players and still quick to toggle for power user... 🤹‍♂️ 

Closes #1519

### DIFF
--- a/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
+++ b/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
@@ -36,6 +36,7 @@ import forge.game.phase.PhaseType;
 import forge.game.player.*;
 import forge.game.replacement.ReplacementEffect;
 import forge.game.spellability.*;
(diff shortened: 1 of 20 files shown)


kevlahnota (approved):

Hanmac:
the UpTo Cost Reduce is interesting, because sometimes you want to spent more mana on it

PR #3980, opened 2023-10-26
LCI: Finality counter

Hanmac:
Might need Different Color

and need some Card examples


Agetian (approved):

Hanmac:
@kevlahnota maybe in the future, have the colors of the counters be defined in a theme-depending file?
So we don't need to have this data in the non Gui parts?

kevlahnota:
> @kevlahnota maybe in the future, have the colors of the counters be defined in a theme-depending file? So we don't need to have this data in the non Gui parts?

that would be better and easier to customize, just default to white color if not defined I guess.

Hanmac:
it also might help to define CounterTypes via a Config File (if they aren't used in the Engine like Shield/Stun/Finality)

PR #3982, opened 2023-10-26
remove UI_CLONE_MODE_SOURCE preference

kevlahnota:
- clone effect was updated last 2019 and this isn't used anymore, free up some space for options

### DIFF
--- a/forge-gui-mobile/src/forge/screens/settings/SettingsPage.java
+++ b/forge-gui-mobile/src/forge/screens/settings/SettingsPage.java
@@ -233,10 +233,6 @@ public class SettingsPage extends TabPage<SettingsScreen> {
                 Forge.getLocalizer().getMessage("cbFilteredHands"),
                 Forge.getLocalizer().getMessage("nlFilteredHands")),
                 1);
-        lstSettings.addItem(new BooleanSetting(FPref.UI_CLONE_MODE_SOURCE,
-                Forge.getLocalizer().getMessage("cbCloneImgSource"),
-                Forge.getLocalizer().getMessage("nlCloneImgSource")),
-                1);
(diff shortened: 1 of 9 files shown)


forge-gui-mobile/src/forge/screens/settings/SettingsPage.java
```diff
@@ -233,10 +233,6 @@ public void valueChanged(String newValue) {
                 Forge.getLocalizer().getMessage("cbFilteredHands"),
                 Forge.getLocalizer().getMessage("nlFilteredHands")),
                 1);
-        lstSettings.addItem(new BooleanSetting(FPref.UI_CLONE_MODE_SOURCE,
-                Forge.getLocalizer().getMessage("cbCloneImgSource"),
-                Forge.getLocalizer().getMessage("nlCloneImgSource")),
```
tool4ever, line 238: we could remove it from languages too?
but honestly I would rather make it work again?

kevlahnota: no idea if the previous method will not be in conflict on other cardviews since it overrides the imagekey directly, it should be done in a different way somehow leaving the original imagekey intact, and the option shouldn't be in the gamerules

tool4ever: Hmn I'll see if I can make it work again, I think the idea behind this feature is still sound

kevlahnota: I think it should be inline on the card art preference as a global settings, but it seems it's lacking some views (like if the player only want earlier cards to be shown to persist, etc) when it was refactored by @leriomaggio

PR #3984, opened 2023-10-26
Reimplement UI_CLONE_MODE_SOURCE

### LINKED ISSUE #3983: Reimplement UI_CLONE_MODE_SOURCE
I think it should be inline on the card art preference as a global settings, but it seems it's lacking some views (like if the player only want earlier cards to be shown to persist, etc) when it was refactored by @leriomaggio

_Originally posted by @kevlahnota in https://github.com/Card-Forge/forge/pull/3982#discussion_r1372848274_

tool4ever:
Only affects the cloned state.

Closes #3983


kevlahnota:
I think the better way to do this is in cardview (like the tokenkey for facedown) except when getimagekey is called by cloner, check  StaticData.instance().getCardArtPreference() if it's using original, return the original art imagekey instead overriding the preferred art.

Hanmac:
like i said in the linked issue before, i think the logic should be flipped, and only special cards should set a different tokenImage

tool4ever:
> like i said in the linked issue before, i think the logic should be flipped, and only special cards should set a different tokenImage

not sure I fully understand:
- from my POV the goal of this preference is to have more visual help in quickly identifying the original cards when playing with mass clone decks
- most likely not all corner cases are worth accounting for

E. g. say my opponent casts _Mirrorweave_ on his best creature:
grafik

and then you happen on a targeted removal trigger - it's very annoying having to compare creatures or IDs individually to find the original one so it would feel a lot more comfortable to have the option of still retaining the above display:
grafik

tehdiplomat (approved):

PR #3986, opened 2023-10-26
Fix Cyber Conversion turning face-down card face-up again

tool4ever:


forge-gui/res/cardsfolder/b/backslide.txt
```diff
@@ -1,7 +1,7 @@
 Name:Backslide
 ManaCost:1 U
 Types:Instant
-A:SP$ SetState | Cost$ 1 U | ValidTgts$ Creature.faceUp+withMorph,Creature.faceUp+withMegamorph | TgtPrompt$ Select target creature with morph. | Mode$ TurnFace | SpellDescription$ Turn target creature with a morph ability face down.
+A:SP$ SetState | Cost$ 1 U | ValidTgts$ Creature.withMorph,Creature.withMegamorph | TgtPrompt$ Select target creature with morph. | Mode$ TurnFaceDown | SpellDescription$ Turn target creature with a morph ability face down.
```
Hanmac, line 4: rules wise this still should only be target face up cards?

forge-gui/res/cardsfolder/m/master_of_the_veil.txt
```diff
@@ -4,6 +4,6 @@ Types:Creature Human Wizard
 PT:2/3
 K:Morph:2 U
 T:Mode$ TurnFaceUp | ValidCard$ Card.Self | Execute$ TrigSetState | OptionalDecider$ You | TriggerZones$ Battlefield | TriggerDescription$ When CARDNAME is turned face up, you may turn target creature with morph face down.
-SVar:TrigSetState:DB$ SetState | ValidTgts$ Creature.faceUp+withMorph,Creature.faceUp+withMegamorph | TgtPrompt$ Select target creature with morph. | Mode$ TurnFace
+SVar:TrigSetState:DB$ SetState | ValidTgts$ Creature.withMorph,Creature.withMegamorph | TgtPrompt$ Select target creature with morph. | Mode$ TurnFaceDown
```
Hanmac, line 7: this should still only target face up cards?

forge-gui/res/cardsfolder/upcoming/cyber_conversion.txt
```diff
@@ -1,6 +1,6 @@
 Name:Cyber Conversion
 ManaCost:U U
 Types:Instant
-A:SP$ SetState | ValidTgts$ Creature | Mode$ TurnFace | FaceDownSetType$ Artifact & Creature & Cyberman | FaceDownPower$ 2 | FaceDownToughness$ 2 | SpellDescription$ Turn target creature face down. It's a 2/2 Cyberman artifact creature.
+A:SP$ SetState | ValidTgts$ Creature | Mode$ TurnFaceDown | FaceDownSetType$ Artifact & Creature & Cyberman | FaceDownPower$ 2 | FaceDownToughness$ 2 | SpellDescription$ Turn target creature face down. It's a 2/2 Cyberman artifact creature.
```
Hanmac, line 4: @Agetian does this maybe need AI logic to not target an already face down card? or is that already handled in SetStateAI ?

forge-gui/res/cardsfolder/b/backslide.txt
tool4ever: they only say "with morph ability" so face-down copy of one should be allowed imo (for all the good that it would do)

Hanmac: my question is more of a rules thing, like does a facedown card count as "with morph ability" ?

forge-gui/res/cardsfolder/upcoming/cyber_conversion.txt
tool4ever: added sanity checks

forge-gui/res/cardsfolder/b/backslide.txt
tool4ever: I could not find any information to indicate that it should let you target a normal face-down morph, so no further changes for now

Hanmac: does it mean you keep it as `Creature.faceUp+withMorph` for now?

tool4ever: I think I should still remove it because the card doesn't say so
this does not change any behaviour ingame

Hanmac: its the difference if the spell is allowed to target a face down card or not

tool4ever: "with a morph ability" just means "with morph" (702.1d)
so just checking for KW seems right
and if the facedown could somehow gain it, it should be allowed

Hanmac: hm it might be from a time when we had morph copied to face down state.

have you tried to target a face down card with this?

tool4ever: - it still can't target face-down morph
- when it targets face-up morph it can be unmorphed fine

Hanmac (approved):

PR #3992, opened 2023-10-26
LCI – Discover mechanic implementation

Northmoc:
might want to figure out a way to trim reminder text out of corner dialogue prompts? It is a little much

### DIFF
--- a/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
+++ b/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
@@ -762,7 +762,9 @@ public class PlayerControllerAi extends PlayerController {
     @Override
     public boolean chooseBinary(SpellAbility sa, String question, BinaryChoiceType kindOfChoice, Boolean defaultVal) {
         switch (kindOfChoice) {
-            case TapOrUntap: return true;
+            case TapOrUntap:
+            case CastOrHand:
+                return true;
             case UntapOrLeaveTapped:
                 Card source = sa.getHostCard();
                 if (source != null && source.hasSVar("AIUntapPreference")) {
--- a/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
+++ b/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
@@ -242,7 +242,7 @@ public class PlayerControllerAi extends PlayerController {
     }
 
     @Override
-    public boolean confirmAction(SpellAbility sa, PlayerActionConfirmMode mode, String message, Card cardToShow, Map<String, Object> params) {
+    public boolean confirmAction(SpellAbility sa, PlayerActionConfirmMode mode, String message, List<String> options, Card cardToShow, Map<String, Object> params) {
         return getAi().confirmAction(sa, mode, message, params);
     }
 
@@ -762,9 +762,7 @@ public class PlayerControllerAi extends PlayerController {
     @Override
     public boolean chooseBinary(SpellAbility sa, String question, BinaryChoiceType kindOfChoice, Boolean defaultVal) {
         switch (kindOfChoice) {
-            case TapOrUntap:
-            case CastOrHand:
-                return true;
+            case TapOrUntap: return true;
             case UntapOrLeaveTapped:
                 Card source = sa.getHostCard();
                 if (source != null && source.hasSVar("AIUntapPreference")) {
--- a/forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
@@ -14,6 +14,7 @@ import forge.game.cost.CostDiscard;
 import forge.game.cost.CostPart;
 import forge.game.cost.CostReveal;
 import forge.game.player.Player;
+import forge.game.player.PlayerCollection;
 import forge.game.spellability.LandAbility;
 import forge.game.spellability.SpellAbility;
 import forge.game.trigger.TriggerType;
@@ -21,6 +22,7 @@ import forge.game.zone.PlayerZone;
 import forge.game.zone.Zone;
 import forge.game.zone.ZoneType;
 import forge.util.CardTranslation;
+import forge.util.Lang;
 import forge.util.Localizer;
 
 import org.apache.commons.lang3.StringUtils;
@@ -33,118 +35,121 @@ public class DiscoverEffect extends SpellAbilityEffect {
 
     @Override
     protected String getStackDescription(SpellAbility sa) {
+        final PlayerCollection players = getDefinedPlayersOrTargeted(sa);
+        final String verb = players.size() == 1 ? " discovers " : " discover ";
 
-        return sa.getActivatingPlayer() + " discovers " + sa.getParamOrDefault("Num", "1") + ".";
+        return Lang.joinHomogenous(players) + verb + sa.getParamOrDefault("Num", "1") + ".";
     }
 
     @Override
     public void resolve(SpellAbility sa) {
         final Card host = sa.getHostCard();
         final Game game = host.getGame();
-        final Player p = sa.getActivatingPlayer();
+        final PlayerCollection players = getDefinedPlayersOrTargeted(sa);
 
         // Exile cards from the top of your library until you exile a nonland card with <N> mana value or less.
         final int num = AbilityUtils.calculateAmount(host, sa.getParamOrDefault("Num", "1"), sa);
 
-        if (p == null || !p.isInGame()) return;
+        for (final Player p : players) {
+            if (p == null || !p.isInGame()) return;
 
-        Card found = null;
-        CardCollection exiled = new CardCollection();
-        CardCollection rest = new CardCollection();
+            Card found = null;
+            CardCollection exiled = new CardCollection();
+            CardCollection rest = new CardCollection();
 
-        final PlayerZone library = p.getZone(ZoneType.Library);
+            final PlayerZone library = p.getZone(ZoneType.Library);
 
-        for (final Card c : library) {
-            exiled.add(c);
-            if (!c.isLand() && c.getCMC() <= num) {
-                found = c;
-                if (sa.hasParam("RememberDiscovered"))
-                    host.addRemembered(c);
-                break;
-            } else {
-                rest.add(c);
-            }
-        }
-
-        if (exiled.size() > 0) {
-            game.getAction().reveal(exiled, p, false);
-        }
-
-        changeZone(exiled, ZoneType.Exile, game, sa);
-
-        // Cast it without paying its mana cost or put it into your hand.
-        if (found != null) {
-            String prompt = Localizer.getInstance().getMessage("lblDiscoverChoice",
-                            CardTranslation.getTranslatedName(found.getName()));
-            final Zone origin = found.getZone();
-            List<String> options =
-                    Arrays.asList(StringUtils.capitalize(Localizer.getInstance().getMessage("lblCast")),
-                            StringUtils.capitalize(Localizer.getInstance().getMessage("lblHandZone")));
-            final boolean play = p.getController().confirmAction(sa, null, prompt, options, found, null);
-            boolean cancel = false;
-
-            if (play) {
-                // get basic spells (no flashback, etc.)
-                List<SpellAbility> sas = AbilityUtils.getBasicSpellsFromPlayEffect(found, p);
-
-                // filter out land abilities due to MDFC or similar
-                Iterables.removeIf(sas, Predicates.instanceOf(LandAbility.class));
-                // the spell must also have a mana value equal to or less than the discover number
-                sas.removeIf(sp -> sp.getPayCosts().getTotalMana().getCMC() > num);
-
-                if (sas.isEmpty()) { // shouldn't happen!
-                    System.err.println("DiscoverEffect Error: " + host + " found " + found + " but couldn't play sa");
+            for (final Card c : library) {
+                exiled.add(c);
+                if (!c.isLand() && c.getCMC() <= num) {
+                    found = c;
+                    if (sa.hasParam("RememberDiscovered"))
+                        host.addRemembered(c);
+                    break;
                 } else {
-                    SpellAbility tgtSA = p.getController().getAbilityToPlay(found, sas);
+                    rest.add(c);
+                }
+            }
 
-                    if (tgtSA == null) { // in case player canceled from choice dialog
-                        cancel = true;
+            if (exiled.size() > 0) {
+                game.getAction().reveal(exiled, p, false);
+            }
+
+            changeZone(exiled, ZoneType.Exile, game, sa);
+
+            // Cast it without paying its mana cost or put it into your hand.
+            if (found != null) {
+                String prompt = Localizer.getInstance().getMessage("lblDiscoverChoice",
+                        CardTranslation.getTranslatedName(found.getName()));
+                final Zone origin = found.getZone();
+                List<String> options =
+                        Arrays.asList(StringUtils.capitalize(Localizer.getInstance().getMessage("lblCast")),
+                                StringUtils.capitalize(Localizer.getInstance().getMessage("lblHandZone")));
+                final boolean play = p.getController().confirmAction(sa, null, prompt, options, found, null);
+                boolean cancel = false;
+
+                if (play) {
+                    // get basic spells (no flashback, etc.)
+                    List<SpellAbility> sas = AbilityUtils.getBasicSpellsFromPlayEffect(found, p);
+
+                    // filter out land abilities due to MDFC or similar
+                    Iterables.removeIf(sas, Predicates.instanceOf(LandAbility.class));
+                    // the spell must also have a mana value equal to or less than the discover number
+                    sas.removeIf(sp -> sp.getPayCosts().getTotalMana().getCMC() > num);
+
+                    if (sas.isEmpty()) { // shouldn't happen!
+                        System.err.println("DiscoverEffect Error: " + host + " found " + found + " but couldn't play sa");
                     } else {
-                        tgtSA = tgtSA.copyWithNoManaCost();
+                        SpellAbility tgtSA = p.getController().getAbilityToPlay(found, sas);
 
-                        // 118.8c
-                        boolean optional = false;
-                        for (CostPart cost : tgtSA.getPayCosts().getCostParts()) {
-                            if ((cost instanceof CostDiscard || cost instanceof CostReveal)
-                                    && !cost.getType().equals("Card") && !cost.getType().equals("Random")) {
-                                optional = true;
-                                break;
+                        if (tgtSA == null) { // in case player canceled from choice dialog
+                            cancel = true;
+                        } else {
+                            tgtSA = tgtSA.copyWithNoManaCost();
+
+                            // 118.8c
+                            boolean optional = false;
+                            for (CostPart cost : tgtSA.getPayCosts().getCostParts()) {
+                                if ((cost instanceof CostDiscard || cost instanceof CostReveal)
+                                        && !cost.getType().equals("Card") && !cost.getType().equals("Random")) {
+                                    optional = true;
+                                    break;
+                                }
                             }
-                        }
-                        if (!optional) {
-                            tgtSA.getPayCosts().setMandatory(true);
-                        }
+                            if (!optional) {
+                                tgtSA.getPayCosts().setMandatory(true);
+                            }
 
-                        if (tgtSA.usesTargeting() && !optional) {
-                            tgtSA.getTargetRestrictions().setMandatory(true);
-                        }
+                            if (tgtSA.usesTargeting() && !optional) {
+                                tgtSA.getTargetRestrictions().setMandatory(true);
+                            }
 
-                        tgtSA.setSVar("IsCastFromPlayEffect", "True");
+                            tgtSA.setSVar("IsCastFromPlayEffect", "True");
 
-                        if (p.getController().playSaFromPlayEffect(tgtSA)) {
-                            final Card played = tgtSA.getHostCard();
-                            // add remember successfully played here if ever needed
-                            final Zone zone = game.getCardState(played).getZone();
-                            if (!origin.equals(zone)) {
-                                CardZoneTable trigList = new CardZoneTable();
-                                trigList.put(origin.getZoneType(), zone.getZoneType(), game.getCardState(found));
-                                trigList.triggerChangesZoneAll(game, sa);
+                            if (p.getController().playSaFromPlayEffect(tgtSA)) {
+                                final Card played = tgtSA.getHostCard();
+                                // add remember successfully played here if ever needed
+                                final Zone zone = game.getCardState(played).getZone();
+                                if (!origin.equals(zone)) {
+                                    CardZoneTable trigList = new CardZoneTable();
+                                    trigList.put(origin.getZoneType(), zone.getZoneType(), game.getCardState(found));
+                                    trigList.triggerChangesZoneAll(game, sa);
+                                }
                             }
                         }
                     }
                 }
+                if (!play || cancel) changeZone(new CardCollection(found), ZoneType.Hand, game, sa);
             }
-            if (!play || cancel) changeZone(new CardCollection(found), ZoneType.Hand, game, sa);
+
+            // Put the rest on the bottom in a random order.
+            changeZone(rest, ZoneType.Library, game, sa);
+
+            // Run discover triggers
+            final Map<AbilityKey, Object> runParams = AbilityKey.mapFromPlayer(p);
+            runParams.put(AbilityKey.Amount, num);
+            game.getTriggerHandler().runTrigger(TriggerType.Discover, runParams, false);
         }
-
-        // Put the rest on the bottom in a random order.
-        changeZone(rest, ZoneType.Library, game, sa);
-
-        // Run discover triggers
-        final Map<AbilityKey, Object> runParams = AbilityKey.mapFromPlayer(p);
-        runParams.put(AbilityKey.Amount, num);
-        game.getTriggerHandler().runTrigger(TriggerType.Discover, runParams, false);
-
     }
 
     private void changeZone(CardCollection cards, ZoneType zone, Game game, SpellAbility sa) {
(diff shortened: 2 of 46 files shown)


forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,176 @@
...
+            if (found != null) {
+                String prompt = Localizer.getInstance().getMessage("lblDiscoverChoice",
+                                CardTranslation.getTranslatedName(found.getName()));
+                final boolean play = p.getController().chooseBinary(sa, prompt,
+                        PlayerController.BinaryChoiceType.CastOrHand);
+
+                final Zone origin = found.getZone();
+                if (play) {
+                    // get basic spells (no flashback, etc.)
+                    List<SpellAbility> sas = AbilityUtils.getBasicSpellsFromPlayEffect(found, p, CardStateName.Original);
```
Hanmac, line 85: Similar to Cascade, you should only be able to select Spells that are LE to Num

Hanmac: > You can either *cast* that card without paying its mana cost or you can put it into your hand.

> Remember that the spell must also have a mana value equal to or less than the discover number.

So the Spell should be NonLand with cmc LE Num

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,169 @@
...
+
+    @Override
+    public void resolve(SpellAbility sa) {
+        final Card host = sa.getHostCard();
+        final Game game = host.getGame();
+
+        // Exile cards from the top of your library until you exile a nonland card with <N> mana value or less.
+        String type = "Card.nonLand+cmcLE" + (sa.getParamOrDefault("Num", "1"));
+
+        for (final Player p : getTargetPlayers(sa)) {
```
Hanmac, line 47: target players is not needed, Discover only allows you to do it

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,169 @@
...
+            }
+
+            changeZone(exiled, ZoneType.Exile, game, sa);
+
+            // Cast it without paying its mana cost or put it into your hand.
+            if (found != null) {
+                String prompt = Localizer.getInstance().getMessage("lblDiscoverChoice",
+                                CardTranslation.getTranslatedName(found.getName()));
+                final boolean play = p.getController().chooseBinary(sa, prompt,
+                        PlayerController.BinaryChoiceType.CastOrHand);
```
Hanmac, line 80: i don't know if need another Binary Choice or if we just do "Yes OR No" 

@Agetian your opinion?

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,169 @@
...
+                    List<SpellAbility> sas = AbilityUtils.getBasicSpellsFromPlayEffect(found, p, CardStateName.Original);
+
+                    // filter land abilities due to MDFC or similar
+                    List <SpellAbility> landAbilities = Lists.newArrayList();
+                    for (SpellAbility s : sas) {
+                        if (s instanceof LandAbility) {
+                            landAbilities.add(s);
+                        }
+                    }
+                    sas.removeAll(landAbilities);
```
Hanmac, line 94: use Predicates.[instanceOf](https://guava.dev/releases/21.0/api/docs/com/google/common/base/Predicates.html#instanceOf-java.lang.Class-)(LandAbility)
with
sas.removeIf or Iterables.removeIf(sas,)

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,169 @@
...
+                                final Zone zone = game.getCardState(played).getZone();
+                                if (!origin.equals(zone)) {
+                                    CardZoneTable trigList = new CardZoneTable();
+                                    trigList.put(origin.getZoneType(), zone.getZoneType(), game.getCardState(found));
+                                    trigList.triggerChangesZoneAll(game, sa);
+                                }
+                            }
+                        }
+                    }
+                } else {
```
Hanmac, line 134: in case the dialog is canceled, the card should be put in hand too?

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
Northmoc: That's fair... could make it work either way.

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
Northmoc: Sure, would be easy enough to add it back if/when some pushed card allows multiple players to do it

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
Northmoc: Think I've got this sorted

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,170 @@
...
+            String prompt = Localizer.getInstance().getMessage("lblDiscoverChoice",
+                            CardTranslation.getTranslatedName(found.getName()));
+            final Zone origin = found.getZone();
+            final boolean play = p.getController().chooseBinary(sa, prompt,
+                    PlayerController.BinaryChoiceType.CastOrHand);
+            boolean cancel = false;
+
+            if (play) {
+                // get basic spells (no flashback, etc.)
+                List<SpellAbility> sas = AbilityUtils.getBasicSpellsFromPlayEffect(found, p, CardStateName.Original);
```
Hanmac, line 89: isn't the third parameter wrong? Doesn't this function only allows two?

i can't find an example for this?

Northmoc: Hmm, that would have been copied from PlayEffect... looks like third param is OK, but in this case it's the default anyway so might as well trim it
image

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,170 @@
...
+            } else {
+                rest.add(c);
+            }
+        }
+
+        if (exiled.size() > 0) {
+            game.getAction().reveal(exiled, p, false);
+        }
+
+        changeZone(exiled, ZoneType.Exile, game, sa);
```
tool4ever, line 76: this is still not sequential?

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
Northmoc: It was a copy/paste of line 305 in PlayEffect that I must have modified without checking the AbilityUtils options

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,170 @@
...
+        for (Card c : cards) {
+            final ZoneType origin = c.getZone().getZoneType();
+
+            Card m = game.getAction().moveTo(zone, c, pos, sa, moveParams);
+
+            if (m != null && !origin.equals(m.getZone().getZoneType())) {
+                table.put(origin, m.getZone().getZoneType(), m);
+            }
+        }
+        table.triggerChangesZoneAll(game, sa);
```
tool4ever, line 168: only the exiling should be sequential
the move rest is not

forge-gui/res/cardsfolder/upcoming/quintorius_kand.txt
```diff
@@ -0,0 +1,18 @@
...
+SVar:TrigDamage:DB$ DealDamage | Defined$ Opponent | NumDmg$ 2 | SubAbility$ DBGainLife
+SVar:DBGainLife:DB$ GainLife | LifeAmount$ 2
+A:AB$ Token | Cost$ AddCounter<1/LOYALTY> | Planeswalker$ True | TokenScript$ rw_3_2_spirit | SpellDescription$ Create a 3/2 red and white Spirit creature token.
+A:AB$ Discover | Cost$ SubCounter<3/LOYALTY> | Planeswalker$ True | Num$ 4 | SpellDescription$ Discover 4.
+A:AB$ ChangeZone | Cost$ SubCounter<6/LOYALTY> | Planeswalker$ True | Ultimate$ True | Origin$ Graveyard | Destination$ Exile | TargetMin$ 0 | TargetMax$ Yard | TgtPrompt$ Select any number of target cards in your graveyard | ValidTgts$ Card.YouOwn | RememberChanged$ True | StackDescription$ SpellDescription | SpellDescription$ Exile any number of target cards from your graveyard. Add {R} for each card exiled this way. You may play those cards this turn.
+SVar:DBMana:DB$ Mana | Produced$ R | Amount$ X | SubAbility$ DBEffect
+SVar:DBEffect:DB$ Effect | RememberObjects$ Remembered | StaticAbilities$ STMayPlay | Duration$ UntilYourNextTurn | SubAbility$ DBCleanup
```
tool4ever, line 12: wrong Duration

forge-gui/res/cardsfolder/upcoming/quintorius_kand.txt
```diff
@@ -0,0 +1,18 @@
...
+SVar:STMayPlay:Mode$ Continuous | MayPlay$ True | EffectZone$ Command | Affected$ Card.IsRemembered | AffectedZone$ Exile | Description$ You may play those cards.
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True | SubAbility$ DBSpace
```
tool4ever, line 14: bad sub

Northmoc: 👀

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
Northmoc: I think I have this sorted –
I also thought about making line 60 `changeZone(new CardCollection(c), ZoneType.Exile, game, sa)` and removing the CardCollection exiled... but it seems like that would be a bit less computationally efficient?
image

Northmoc:
Here's a script for the Curator (using MTGoldfish translation) if anyone wants to play around with the trigger. We can add the official card once we have the exact text.
```
Name:Curator of Sun's Creation
ManaCost:3 R
Types:Creature Human Artificer
PT:3/3
T:Mode$ Discover | ValidPlayer$ You | TriggerZone$ Battlefield | Execute$ TrigDiscover | ActivationLimit$ 1 | TriggerDescription$ Whenever you discover, discover again for the same value. This ability triggers only once per turn.
SVar:TrigDiscover:DB$ Discover | Num$ X
SVar:X:TriggerCount$Amount
Oracle:Whenever you discover, discover again for the same value. This ability triggers only once per turn.
```

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,179 @@
...
+        Card found = null;
+        CardCollection exiled = new CardCollection();
+        CardCollection rest = new CardCollection();
+
+        final PlayerZone library = p.getZone(ZoneType.Library);
+
+        for (int i = 0; i < library.size(); i++) {
+            final Card c = library.get(i);
+            exiled.add(c);
+            if (c.isValid("Card.nonLand+cmcLE" + val, sa.getActivatingPlayer(), host, sa)) {
```
Hanmac, line 62: i don't know if i like the IsValid string.

when there is nothing that would be changed by text stuff, or is coming from different parameters,

wouldn't it better to just hard code the nonLand and cmc checks?

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,179 @@
...
+            boolean cancel = false;
+
+            if (play) {
+                // get basic spells (no flashback, etc.)
+                List<SpellAbility> sas = AbilityUtils.getBasicSpellsFromPlayEffect(found, p);
+
+                // filter out land abilities due to MDFC or similar
+                Iterables.removeIf(sas, Predicates.instanceOf(LandAbility.class));
+                // the spell must also have a mana value equal to or less than the discover number
+                sas.removeIf(sp -> !sp.isValid("Spell.cmcLE" + val, p , host, sa));
```
Hanmac, line 94: same with there, the SpellCMC should be filtered out better than using isValid?

have you tested it with Adventure spells? if the check really gives the right cmc?

forge-gui/res/cardsfolder/upcoming/brasss_tunnel_grinder_tecutlan_the_searing_rift.txt
```diff
@@ -0,0 +1,26 @@
+Name:Brass's Tunnel-Grinder
+ManaCost:2 R
+Types:Legendary Artifact
+T:Mode$ ChangesZone | ValidCard$ Card.Self | Origin$ Any | Destination$ Battlefield | Execute$ TrigDiscard | TriggerDescription$ When CARDNAME enters the battlefield, discard any number of cards, then draw that many cards plus one.
+SVar:TrigDiscard:DB$ Discard | AnyNumber$ True | Optional$ True | Mode$ TgtChoose | RememberDiscarded$ True | SubAbility$ DBDraw
+SVar:DBDraw:DB$ Draw | Defined$ You | NumCards$ Remembered$Amount/Plus.1 | SubAbility$ DBCleanup
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You.descended | TriggerZones$ Battlefield | Execute$ TrigPutCounter | TriggerDescription$ At the beginning of your end step, if you descended this turn, put a bore counter on Brass's Tunnel-Grinder. Then if there are three or more bore counters on it, remove those counters and transform it. (You descended if a permanent card was put into your graveyard from anywhere.)
```
tool4ever, line 8: CARDNAME

forge-gui/res/cardsfolder/upcoming/brasss_tunnel_grinder_tecutlan_the_searing_rift.txt
```diff
@@ -0,0 +1,26 @@
+Name:Brass's Tunnel-Grinder
+ManaCost:2 R
+Types:Legendary Artifact
+T:Mode$ ChangesZone | ValidCard$ Card.Self | Origin$ Any | Destination$ Battlefield | Execute$ TrigDiscard | TriggerDescription$ When CARDNAME enters the battlefield, discard any number of cards, then draw that many cards plus one.
+SVar:TrigDiscard:DB$ Discard | AnyNumber$ True | Optional$ True | Mode$ TgtChoose | RememberDiscarded$ True | SubAbility$ DBDraw
+SVar:DBDraw:DB$ Draw | Defined$ You | NumCards$ Remembered$Amount/Plus.1 | SubAbility$ DBCleanup
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You.descended | TriggerZones$ Battlefield | Execute$ TrigPutCounter | TriggerDescription$ At the beginning of your end step, if you descended this turn, put a bore counter on CARDNAME. Then if there are three or more bore counters on it, remove those counters and transform it. (You descended if a permanent card was put into your graveyard from anywhere.)
+SVar:TrigPutCounter:DB$ PutCounter | CounterType$ BORE | SubAbility$ DBBranch
+SVar:DBBranch:DB$ Branch | BranchConditionSVar$ Count$Valid Card.Self+counters_GE3_BORE | TrueSubAbility$ DBRemoveCounter
```
tool4ever, line 10: `BranchConditionSVar$ Count$CardCounters.BORE`

forge-gui/res/cardsfolder/upcoming/brasss_tunnel_grinder_tecutlan_the_searing_rift.txt
```diff
@@ -0,0 +1,26 @@
...
+AlternateMode:DoubleFaced
+Oracle:When Brass's Tunnel-Grinder enters the battlefield, discard any number of cards, then draw that many cards plus one.\nAt the beginning of your end step, if you descended this turn, put a bore counter on Brass's Tunnel-Grinder. Then if there are three or more bore counters on it, remove those counters and transform it. (You descended if a permanent card was put into your graveyard from anywhere.)
+
+ALTERNATE
+
+Name:Tecutlan, the Searing Rift
+ManaCost:no cost
+Types:Legendary Land Cave
+A:AB$ Mana | Cost$ T | Produced$ R | TriggersWhenSpent$ SpentTrigger | SubAbility$ DBSpace | SpellDescription$ Add {R}.
+SVar:DBSpace:DB$ BlankLine | SpellDescription$ ,,,,,,Whenever you cast a permanent spell using mana produced by CARDNAME, discover X, where X is that spell's mana value.
```
tool4ever, line 22: I think this needs more support, this is not the normal TriggersWhenSpent that's tied to its mana ability as delayed trigger

This is a trigger on its own with the consequence it can work for any other mana ability the card woud have available

forge-gui/res/cardsfolder/upcoming/quintorius_kand.txt
```diff
@@ -0,0 +1,18 @@
...
+SVar:TrigDamage:DB$ DealDamage | Defined$ Opponent | NumDmg$ 2 | SubAbility$ DBGainLife
+SVar:DBGainLife:DB$ GainLife | LifeAmount$ 2
+A:AB$ Token | Cost$ AddCounter<1/LOYALTY> | Planeswalker$ True | TokenScript$ rw_3_2_spirit | SpellDescription$ Create a 3/2 red and white Spirit creature token.
+A:AB$ Discover | Cost$ SubCounter<3/LOYALTY> | Planeswalker$ True | Num$ 4 | SpellDescription$ Discover 4.
+A:AB$ ChangeZone | Cost$ SubCounter<6/LOYALTY> | Planeswalker$ True | Ultimate$ True | Origin$ Graveyard | Destination$ Exile | TargetMin$ 0 | TargetMax$ Yard | TgtPrompt$ Select any number of target cards in your graveyard | ValidTgts$ Card.YouOwn | RememberChanged$ True | StackDescription$ SpellDescription | SpellDescription$ Exile any number of target cards from your graveyard. Add {R} for each card exiled this way. You may play those cards this turn.
+SVar:DBMana:DB$ Mana | Produced$ R | Amount$ X | SubAbility$ DBEffect
+SVar:DBEffect:DB$ Effect | RememberObjects$ Remembered | StaticAbilities$ STMayPlay | SubAbility$ DBCleanup
```
tool4ever, line 12: missing ForgetOnMoved

forge-gui/res/cardsfolder/upcoming/brasss_tunnel_grinder_tecutlan_the_searing_rift.txt
Northmoc: Seems to work in basic testing

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,175 @@
...
+
+        if (p == null || !p.isInGame()) return;
+
+        Card found = null;
+        CardCollection exiled = new CardCollection();
+        CardCollection rest = new CardCollection();
+
+        final PlayerZone library = p.getZone(ZoneType.Library);
+
+        for (int i = 0; i < library.size(); i++) {
```
tool4ever, line 55: just use for-each

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        changeZone(exiled, ZoneType.Exile, game, sa);
+
+        // Cast it without paying its mana cost or put it into your hand.
+        if (found != null) {
+            String prompt = Localizer.getInstance().getMessage("lblDiscoverChoice",
+                            CardTranslation.getTranslatedName(found.getName()));
+            final Zone origin = found.getZone();
+            final boolean play = p.getController().chooseBinary(sa, prompt,
```
Hanmac, line 79: I still don't know if Binary Choice is a good idea there

wouldn't it be better to use Confirm Style for Casting?

@Agetian your opinion for this? right now the AI would try to cast it and then maybe fail later?

forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
```diff
@@ -762,7 +762,7 @@ public void notifyOfValue(SpellAbility saSource, GameObject realtedTarget, Strin
     @Override
     public boolean chooseBinary(SpellAbility sa, String question, BinaryChoiceType kindOfChoice, Boolean defaultVal) {
         switch (kindOfChoice) {
-            case TapOrUntap: return true;
+            case TapOrUntap:
             case UntapOrLeaveTapped:
```
Hanmac, line 766: i don't think they can be merged for AI this way 

the return value is different for both of them

Northmoc: This was just a bad rollback, will fix

Northmoc:
The last card I did (Zoyowa's Justice) required a little tweak to DiscoverEffect

forge-game/src/main/java/forge/game/ability/effects/DiscoverEffect.java
```diff
@@ -0,0 +1,178 @@
...
+    protected String getStackDescription(SpellAbility sa) {
+
+        return sa.getActivatingPlayer() + " discovers " + sa.getParamOrDefault("Num", "1") + ".";
+    }
+
+    @Override
+    public void resolve(SpellAbility sa) {
+        final Card host = sa.getHostCard();
+        final Game game = host.getGame();
+        final Player p = getDefinedPlayersOrTargeted(sa).get(0);
```
Hanmac, line 44: you could put this in a loop, so we could handle if multiple players should discover at the same time
(it that is ever needed)

Hanmac (dismissed):

forge-gui/res/cardsfolder/upcoming/brasss_tunnel_grinder_tecutlan_the_searing_rift.txt
```diff
@@ -0,0 +1,25 @@
...
+ManaCost:2 R
+Types:Legendary Artifact
+T:Mode$ ChangesZone | ValidCard$ Card.Self | Origin$ Any | Destination$ Battlefield | Execute$ TrigDiscard | TriggerDescription$ When CARDNAME enters the battlefield, discard any number of cards, then draw that many cards plus one.
+SVar:TrigDiscard:DB$ Discard | AnyNumber$ True | Optional$ True | Mode$ TgtChoose | RememberDiscarded$ True | SubAbility$ DBDraw
+SVar:DBDraw:DB$ Draw | Defined$ You | NumCards$ Remembered$Amount/Plus.1 | SubAbility$ DBCleanup
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You.descended | TriggerZones$ Battlefield | Execute$ TrigPutCounter | TriggerDescription$ At the beginning of your end step, if you descended this turn, put a bore counter on CARDNAME. Then if there are three or more bore counters on it, remove those counters and transform it. (You descended if a permanent card was put into your graveyard from anywhere.)
+SVar:TrigPutCounter:DB$ PutCounter | CounterType$ BORE | SubAbility$ DBBranch
+SVar:DBBranch:DB$ Branch | BranchConditionSVar$ Count$Valid Count$CardCounters.BORE | BranchConditionSVarCompare$ GE3 | TrueSubAbility$ DBRemoveCounter
```
tool4ever, line 10: `BranchConditionSVar$ Count$CardCounters.BORE`

forge-gui/res/cardsfolder/upcoming/brasss_tunnel_grinder_tecutlan_the_searing_rift.txt
```diff
@@ -0,0 +1,25 @@
...
+
+ALTERNATE
+
+Name:Tecutlan, the Searing Rift
+ManaCost:no cost
+Types:Legendary Land Cave
+A:AB$ Mana | Cost$ T | Produced$ R | SpellDescription$ Add {R}.
+T:Mode$ SpellCast | ValidCard$ Permanent | ValidSA$ Spell.ManaFromCard.StrictlySelf | ValidActivatingPlayer$ You | Execute$ TrigDiscover | TriggerDescription$ Whenever you cast a permanent spell using mana produced by CARDNAME, discover X, where X is that spell's mana value.
+SVar:TrigDiscover:DB$ Discover | Num$ TriggeredStackInstance$CardManaCostLKI
+SVar:X:TriggeredCard$CardToughness
```
tool4ever, line 24: unused

forge-gui/res/cardsfolder/upcoming/caparocti_sunborn.txt
```diff
@@ -0,0 +1,8 @@
+Name:Caparocti Sunborn
+ManaCost:2 R W
+Types:Legendary Creature Human Soldier
+PT:4/4
+T:Mode$ Attacks | ValidCard$ Card.Self | TriggerZones$ Battlefield | Execute$ TrigDiscover | TriggerDescription$ Whenever CARDNAME attacks, you may tap two other untapped creatures you control. If you do, discover 3. (Exile cards from the top of your library until you exile a nonland card with mana value 3 or less. Cast it without paying its mana cost or put it into your hand. Put the rest on the bottom in a random order.)
```
tool4ever, line 5: Description is slighly wrong

tool4ever (approved):
:clap:

Hanmac (approved):

PR #4006, opened 2023-10-27
Macro system based on recording clicks rather than typing

tehdiplomat:


squee1968:
All hope for success in all your endeavors, especially this one. Could you possibly tell where I can find out how to use the current system?

tehdiplomat:
@squee1968 I have no idea how to use the current one thats why I started trying to track it by clicks and I've been playing a lot of combo decks lately.

forge-game/src/main/java/forge/game/player/actions/PlayerAction.java
```diff
@@ -0,0 +1,22 @@
...
+
+public abstract class PlayerAction {
+    protected String name;
+    protected GameEntityView gameEntityView = null;
+
+    public PlayerAction(GameEntityView cardView) {
+        gameEntityView = cardView;
+    }
+
+    public void run(PlayerController controller) {
```
tehdiplomat, line 14: Do I still need this? Should we move the manager actions into these functions instead?

Agetian:
This is really awesome! Thanks for contributing! Not sure if it needs anything else tbh, I guess if our GUI experts don't say anything about what may be required, it should be good to implement upstream?

forge-gui/src/main/java/forge/player/RecordActionsMacroSystem.java
```diff
@@ -0,0 +1,179 @@
...
+
+        if (playbackActions.isEmpty()) {
+            playbackActions.addAll(actions);
+        } else if (actions.size() != playbackActions.size()) {
+            // Not at the beginning of the loop
+            return;
+        }
+
+        while(!playbackActions.isEmpty()) {
+            runFirstAction();
```
tehdiplomat, line 104: This isn't hooked up yet, but we need to wait until Input is ready to receive before running the next action.

forge-game/src/main/java/forge/game/player/actions/PlayerAction.java
tehdiplomat: This didn't work since they were in different modules

forge-gui/src/main/java/forge/player/RecordActionsMacroSystem.java
tehdiplomat: Will look into this in the next one.

tehdiplomat:
@Agetian You want to approve so we can merge?

Agetian (approved):

PR #4007, opened 2023-10-28
LCI – Player.descended support and cards

Northmoc:


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -220,6 +220,9 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
         if (fromBattlefield && !toBattlefield) {
             c.getController().setRevolt(true);
         }
+        if (zoneTo.is(ZoneType.Graveyard) && c.isPermanent()) {
```
Hanmac, line 223: also non Token

@tool4ever  @Agetian does this needs to be checked for Merged card parts?

forge-gui/res/cardsfolder/upcoming/broodrage_mycoid.txt
```diff
@@ -0,0 +1,9 @@
+Name:Broodrage Mycoid
+ManaCost:3 B
+Types:Creature Fungus
+PT:4/3
+T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You | TriggerZones$ Battlefield | CheckDefinedPlayer$ You.descended | Execute$ TrigToken | TriggerDescription$ At the beginning of your end step, if you descended this turn, create a 1/1 black Fungus creature token with "This creature can't block." (You descended if a permanent card was put into your graveyard from anywhere.)
```
Hanmac, line 5: wouldn't it be enough to do `ValidPlayer$ You.decended` ?

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -220,6 +220,9 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
...
+            c.getOwner().descend();
```
tool4ever, line 224: this is way too early for this, the card might still be replaced to move somewhere else

forge-game/src/main/java/forge/game/GameAction.java
tool4ever: yea it would have to be
perhaps it would be cleanest to do this in `Zone.add`

Northmoc: I moved it there!

forge-gui/res/cardsfolder/upcoming/broodrage_mycoid.txt
Northmoc: Hmm, some triggers with `CheckDefinedPlayer$ You.isMonarch` might be refactored too

Hanmac: I think the difference are Intervening If -Clause 

While you can lose the Monarch in response, you can't lose "descended this turn"

tool4ever:
might want to rebase to get rid of the extra stuff

Northmoc:
> might want to rebase to get rid of the extra stuff

yeah I screwed something up 😬

tool4ever (approved):

Hanmac:
@Northmoc when you merge such PR with Reverted stuff
Try to Rebase it first and remove them or Squash it, so it isn't changed in the final Merge

PR #4014, opened 2023-10-28
AI framework to improve sacrificing endangered cards + several AI hints (Stoneforge Mystic, Atog, others) and improvements.

### LINKED ISSUE #4005: AI doesn't sacrifice his own creatures when they are in danger
**Describe the bug**
AI doesn't use abilities of Bubbling Cauldron or Lampad of Death's Vigil to sacrifice his own creatures when they are in danger.

**To Reproduce**
1. AI casts Lampad of Death's Vigil / Bubbling Cauldron and creature.
2. AI has free mana.
3. Player casts spell to destroy/exile Lampad of Death's Vigil or others creatures.
4. AI allows its creatures to die without being helpful to itself.

**Expected behavior**
AI should uses abilities of Bubbling Cauldron or Lampad of Death's Vigil to sacrifice his own creatures when they are in danger.

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-10.27

**Screenshots**

bubbling

### LINKED ISSUE #4022: AI prefers to attack with Stoneforge Mystic rather than activating its ability
**Describe the bug**
AI prefers to attack with Stoneforge Mystic (if the latter is not threatened) rather than activating its ability, even if it has Kaldra Compleat in its hand and activating the ability would deal more damage.

**To Reproduce**
1. AI casts Stoneforge Mystic and finds Kaldra Compleat
2. Player doesn't have any big creatures on the table, he passes the turn
3. AI attacks with its Stoneforge Mystic

**Expected behavior**
AI should activate Stoneforge Mystic ability as often as possible, priority should be higher

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-10.28

**Screenshots**

mystic1
mystic2
mystic3
mystic4

### LINKED ISSUE #4023: AI ineffectively activates Stoneforge Mystic ability
**Describe the bug**
AI activates Stoneforge Mystic ability after the combat phase, so Phyrexian Germ equipped with Kaldra Compleat doesn't go to attack, losing a whole turn.

**To Reproduce**
1. AI casts Stoneforge Mystic and finds Kaldra Compleat
2. On the next turn, AI activates the Stoneforge Mystic ability after the combat phase

**Expected behavior**
AI should activate the Stoneforge Mystic ability either before the combat phase (with Kaldra Compleat in hand or other gear that can be attached immediately) or during the opponent's turn (e.g., with Batterskull in hand to use it as a surprise blocker).

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-10.28

**Screenshots**

kaldra
kaldra2

### LINKED ISSUE #4036: AI doesn't overload Winds of Abandon
**Describe the bug**
AI doesn't overload Winds of Abandon, only targets creatures

**To Reproduce**
1. Player casts many creatures
2. AI has enough mana to overload Winds of Abandon
3. AI casts Winds of Abandon and targets a creature

**Expected behavior**
AI should overload Winds of Abandon in some cases if it has enough mana to overload Winds of Abandon

**Additional context**
Probably all spells with "overload" have a similar problem

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-10.28

**Screenshots**

winds
winds2
winds3
winds4

### LINKED ISSUE #4037: AI doesn't cast Hallowed Moonlight
**Describe the bug**
AI doesn't cast Hallowed Moonlight when player puts a creature from his hand on the table (e.g., from Elvish Piper ability).

**To Reproduce**
1. Player casts Elvish Piper
2. Player activates ability of Elvish Piper
3. AI has Hallowed Moonlight in hand
4. AI doesn't cast Hallowed Moonlight

**Expected behavior**
AI should casts Hallowed Moonlight in response to spells or abilities that can put a creature on the table.

**Additional context**
Problem like #3707. 
I played with AI with a Martyr Life deck until AI had 11 cards left in the deck. And I activated Elvish Piper every turn. I'm guessing that at least one Hallowed Moonlight was in AI's hand.

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-10.28

### LINKED ISSUE #4038: AI kills itself with The One Ring
**Describe the bug**
AI casts The One Ring and uses its ability with a low amount of life.

**To Reproduce**
1. AI has a low amount of life.
2. AI casts The One Ring
3. AI uses ability of The One Ring
4. AI is dead

**Expected behavior**
The AI shouldn't use The One Ring ability with a low amount of life.

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-10.28

**Screenshots**

ring1
ring2
ring3

### LINKED ISSUE #4085: AI sacrifices Lotus Bloom and casts nothing
**Describe the bug**
AI sacrifices Lotus Bloom after casting it without paying its mana cost and casts nothing for it's mana.

**To Reproduce**
1. AI suspends Lotus Bloom
2. When the last time counter is removed, AI casts Lotus Bloom without paying its mana cost
3. AI sacrifices Lotus Bloom at the second main phase and take black mana
4. AI casts nothing and pass the turn

**Expected behavior**
AI should sacrifice Lotus Bloom to cast something for 3+ mana

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-11.02

**Screenshots**

lotus
lotus2

### LINKED ISSUE #4088: AI - uses whitewater naiads on my hexproof angel...
the AI has Hydra Omnivore and Whitewater Naiads, but somehow does not make the Hydra unblockable but tries it on my Akroma with Lightning Greaves attached.

This happened twice in a row now. Happened while the AI played the Estrid Adaptive Enchantment Precon.

thank you for your efforts.

### LINKED ISSUE #4102: AI fails to properly cast Spiteful Banditry and it deals 0 damage no matter the X choice
**Describe the bug**
Each time the AI tries to cast Spiteful Banditry, it correctly chooses and announces the X value, but the card deals 0 damage.
The problem is that once the card enters the battlefield, the information about the X payment is lost to the AI, and the game assumes that 0 was chosen.

**To Reproduce**
1. Give the AI Spiteful Banditry and some mana to cast it, and give yourself a few 1/1s or 2/2s or whatever.
2. Wait for the AI to cast Spiteful Banditry.

**Expected behavior**
The spell should deal the expected amount of damage when the AI casts it.

I tried debugging this and I can clearly see that it's precisely because the X information is lost between casting the spell and the resolution of the ETB trigger, but I'm not sure what a good approach would be to fixing this... Any ideas, @tool4ever @Hanmac ?

Agetian:
Aims to improve / partially fix #4005.
Unfortunately, the decision to sacrifice or not to sacrifice a certain object is closely tied to each particular AF, so for now, this will only (mostly) reliably work for the common AFs GainLife, LoseLife, Pump, PutCounter, and Draw. Other AFs will work only if there is no restrictive code in place (based on phase or conditions) that prevents or overrides the implemented higher level checks.

Completely fixing this issue requires a nearly complete rewrite of the whole system (and most AFs), which is way more than I can handle at the moment :)

Also:
Fixes #4022.
Fixes #4023.
Fixes #4036.
Fixes #4037.
Fixes #4038.
Fixes #4102.
Fixes #4088.
Fixes #4085 (at least improves it by handling the Lotus cards similar to other mana ritual cards as far as ManaEffectAi is concerned).


tool4ever:
Might take me a bit to review, sorry 😵‍💫

Agetian:
Hehe, no worries at all, I'm still testing and tweaking as well :D

forge-ai/src/main/java/forge/ai/CreatureEvaluator.java
```diff
@@ -247,6 +247,14 @@ else if (c.hasKeyword(Keyword.WITHER)) {
             value -= subValue(20, "upkeep-dmg");
         } 
 
+        // evaluate abilities
+        // TODO: expand this for other scenarios where an ability may influence the priority
+        for (SpellAbility sa : c.getSpellAbilities()) {
```
tool4ever, line 252: I don't think we can/should tweak the Evaluator like that...
You might sometimes get better results on removal decisions, but at the same time you've now taught the AI it should rather play a vanilla 1/1 than _Sakura-Tribe Elder_ when given the choice

Agetian: Ah, indeed! Hmm, maybe this should then be limited to cards that are already on the battlefield and are controlled by the opponent? Then again, there will indeed be cases when there will be a need to prioritize it higher with something... Maybe dropping the idea is better then? 🤔 Do you have any ideas?

tool4ever: you could override `evaluateCreature` with a boolean or API param so you can check the context
or just another helper function that can filter, though finding all the right places to apply it might be a bit unpleasant

Agetian: Gotcha! Yeah, I'll revert the current approach for now and I'll think a bit on how to implement it better, possibly as a separate PR since this one is getting pretty fat by now XD

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -433,12 +427,40 @@ public static Card getCardPreference(final Player ai, final Card activate, final
...
+                                if (card.isCreature() && ComputerUtilCard.evaluateCreature(card) > maxCreatureEval) {
+                                    return false;
+                                }
+
+                                return (allowTokens && card.isToken())
+                                        || (card.getCMC() >= minCMC && card.getCMC() <= maxCMC && !dontSac.contains(card.getName()));
+                            }
+                        });
+                        if (!allowList.isEmpty()) {
+                            sacMeList.addAll(typeList);
```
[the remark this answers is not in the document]
Agetian: Ah, should be allowList, not typeList there ;)

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -3257,5 +3305,12 @@ public static boolean isETBprevented(Card c) {
         List<ReplacementEffect> list = c.getGame().getReplacementHandler().getReplacementList(ReplacementType.Moved, repParams, ReplacementLayer.CantHappen);
         return !list.isEmpty();
     }
-    
+
+    public static boolean shouldSacrificeThreatenedCard(Player ai, Card c, SpellAbility sa) {
+        Game game = ai.getGame();
+        Combat combat = game.getCombat();
+        return (c.isCreature() && ComputerUtil.predictCreatureWillDieThisTurn(ai, c, sa, false)
+                && (game.getCombat() == null || !ComputerUtilCombat.willOpposingCreatureDieInCombat(ai, c, combat)))
```
tool4ever, line 3313: since you added null check to willOpposingCreatureDieInCombat you don't need to check it here additionally

forge-ai/src/main/java/forge/ai/SpecialCardAi.java
```diff
@@ -1610,6 +1610,22 @@ public boolean apply(Card card) {
...
+    public static class TheOneRing {
+        public static boolean consider(final Player ai, final SpellAbility sa) {
+            if (!ai.canLoseLife() || ai.cantLoseForZeroOrLessLife()) {
+                return true;
+            }
+
+            AiController aic = ((PlayerControllerAi) ai.getController()).getAi();
+            int lifeInDanger = aic.getIntProperty(AiProps.AI_IN_DANGER_THRESHOLD);
+            int numCtrs = sa.getHostCard().getCounters(CounterEnumType.BURDEN);
+            boolean notDangerous = !ai.canLoseLife() || ai.cantLoseForZeroOrLessLife() || (ai.getLife() > numCtrs + 1 && ai.getLife() > lifeInDanger);
```
tool4ever, line 1623: aren't the first two conditions already checked above anyway?

Agetian: Good catch! Done :)

forge-ai/src/main/java/forge/ai/ComputerUtil.java
Agetian: Nice, thanks! :) Done.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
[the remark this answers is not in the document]
Agetian: Then I don't get it ^^;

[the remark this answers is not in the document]
Agetian: Ok, let's try a simpler approach, haha :) I think there's at least one array copy that can be avoided

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -383,7 +377,8 @@ public static Card getCardPreference(final Player ai, final Card activate, final
                     }
 
                     for (String validItem : prefValid[1].split(",")) {
-                        final CardCollection prefList = CardLists.getValidCards(typeList, validItem, activate.getController(), activate, null);
+                        CardCollection prefList = CardLists.getValidCards(typeList, validItem, activate.getController(), activate, null);
+                        prefList = CardLists.filter(prefList, card -> !card.hasSVar("AIDontSacrifice"));
```
tool4ever, line 381: Not sure this new SVar makes sense?
I mean the P9 are that mostly for their cheap costs, but if you already played them they should be ok to sac (compared to other artifacts anyway)?

Agetian: The AI may be too eager to sacrifice them at the first available opportunity otherwise... Like, let's say it plays a turn 1 land and Mox, then on turn 2, it plays something and sacs the Mox even though it could have kept it for the early game mana advantage (having 3 mana sources on turn 2 instead of having 2 mana sources), something like that. Think it makes sense? Or doesn't matter much?

Agetian: If it makes sense to keep this at all, maybe we can change it so that the AI sacrifices them if it has enough mana sources or something like that, what do you think?

tool4ever: right, I just don't like extra checks like that if they only improve one scenario type while worsening others

tool4ever: from a logic point, it would make sense to have some kind of counterpart to SacCost which instead of such source can be put on potential victims?
maybe some logic from NeedsToPlay can be reused

forge-ai/src/main/java/forge/ai/ability/LifeLoseAi.java
```diff
@@ -149,6 +157,10 @@ protected boolean checkApiLogic(Player ai, SpellAbility sa) {
             return true;
         }
 
+        if (sa.getPayCosts() != null && sa.getPayCosts().hasSpecificCostType(CostSacrifice.class)) {
```
tool4ever, line 160: these parts seem like another thing that's waaaay too broad
makes it sacrifice stuff like _Bile Urchin_ first chance after casting

forge-ai/src/main/java/forge/ai/ComputerUtil.java
Agetian: Yeah, I agree with the reasoning, I guess let's keep this part out of this particular MR, I need to think a bit more on how to best implement this... I'll keep the logic in the "default auto-sac preference" part for the time being, as it's a bit too risky to leave with the Moxen freely sacrifice-able :)

forge-ai/src/main/java/forge/ai/ability/LifeLoseAi.java
Agetian: I think we can try limiting this to activating it in a threatened card scenario for now.

tool4ever: this could misplay because you might be checking the wrong sacrifice target
but that aside, this should already be checked by the `activateForCost` call above anyway?
what about the other cases?

Agetian: Yeah, I think this is superfluous and risky, I'll remove it. I believe you're right it's handled by the other checks, I tested a couple scenarios and they still work. Do you have any particular other misplay situations you'd like me to check out? I'll probably do a better job figuring out how to wire it up properly if I know a specific situation where things go south :)

tool4ever: for LifeGain it might be ok because it uses custom `willPayCosts`
but for which case does this help? 🤔

Agetian: Hmm, not sure off the top of my head either :/

forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
```diff
@@ -358,8 +358,10 @@ public static boolean checkSacrificeCost(final Player ai, final Cost cost, final
                     }
                     if (source.isCreature()) {
                         // e.g. Sakura Tribe-Elder
+                        final Combat combat = ai.getGame().getCombat();
                         final boolean beforeNextTurn = ai.getGame().getPhaseHandler().is(PhaseType.END_OF_TURN) && ai.getGame().getPhaseHandler().getNextTurn().equals(ai);
-                        final boolean creatureInDanger = ComputerUtil.predictCreatureWillDieThisTurn(ai, source, sourceAbility);
+                        final boolean creatureInDanger = ComputerUtil.predictCreatureWillDieThisTurn(ai, source, sourceAbility, false)
+                                && (combat == null || !ComputerUtilCombat.willOpposingCreatureDieInCombat(ai, source, combat));
```
tool4ever, line 364: don't need null check again

Agetian: :+1:

forge-ai/src/main/java/forge/ai/ability/LifeLoseAi.java
Agetian: Ugh, can't really think of anything that would need a similar routine in LifeGain, I think we keep things as is for now in this part?...

kevlahnota:
I think this is ok, before it gets really big..

kevlahnota (approved):

tool4ever (changes requested):
check discord

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -433,12 +427,39 @@ public static Card getCardPreference(final Player ai, final Card activate, final
...
+                                || (priority == 1 && shouldSacrificeThreatenedCard(ai, c, sa));
                     }
                 });
                 if (!sacMeList.isEmpty()) {
                     CardLists.shuffle(sacMeList);
-                    return sacMeList.get(0);
+                    return sacMeList.getFirst();
+                } else {
+                    // empty sacMeList, so get some viable average preference if the option is enabled
+                    AiController aic = ((PlayerControllerAi) ai.getController()).getAi();
```
kevlahnota, line 439: @tool4ever @Agetian seems the classcast and maybe others needs checking if the controller is human?

Agetian: Fixed, segments like that only need to run for the AI decisions anyway, so the .isAI check is mandatory there. Good and timely catch, thanks! :)

kevlahnota (dismissed):

tool4ever (changes requested):
Sorry, here's another new one:
```
java.lang.StackOverflowError: null
    at java.util.concurrent.ConcurrentHashMap.putVal(ConcurrentHashMap.java:1012)
    at forge.ai.ComputerUtilCombat.attackerWouldBeDestroyed(ComputerUtilCombat.java:711)
    at forge.ai.ComputerUtilCombat.combatantWouldBeDestroyed(ComputerUtilCombat.java:687)
    at forge.ai.ComputerUtilCombat.willOpposingCreatureDieInCombat(ComputerUtilCombat.java:2581)
    at forge.ai.ComputerUtil.shouldSacrificeThreatenedCard(ComputerUtil.java:3265)
    at forge.ai.ComputerUtil.lambda$getCardPreference$0(ComputerUtil.java:428)
    at com.google.common.collect.Iterators$5.computeNext(Iterators.java:673)
    at com.google.common.collect.AbstractIterator.tryToComputeNext(AbstractIterator.java:145)
    at com.google.common.collect.AbstractIterator.hasNext(AbstractIterator.java:140)
    at forge.util.collect.FCollection.addAll(FCollection.java:339)
    at forge.util.collect.FCollection.<init>(FCollection.java:105)
    at forge.game.card.CardCollection.<init>(CardCollection.java:115)
    at forge.game.card.CardLists.filter(CardLists.java:318)
    at forge.ai.ComputerUtil.getCardPreference(ComputerUtil.java:428)
    at forge.ai.ComputerUtilCost.checkSacrificeCost(ComputerUtilCost.java:383)
    at forge.ai.ComputerUtilCost.checkSacrificeCost(ComputerUtilCost.java:406)
    at forge.ai.ComputerUtil.canRegenerate(ComputerUtil.java:1040)
    at forge.ai.ComputerUtilCombat.canDestroyBlockerBeforeFirstStrike(ComputerUtilCombat.java:1823)
    at forge.ai.ComputerUtilCombat.canDestroyAttacker(ComputerUtilCombat.java:1669)
    at forge.ai.ComputerUtilCombat.canDestroyAttacker(ComputerUtilCombat.java:1656)
    at forge.ai.ComputerUtilCombat.attackerWouldBeDestroyed(ComputerUtilCombat.java:711)
    at forge.ai.ComputerUtilCombat.combatantWouldBeDestroyed(ComputerUtilCombat.java:687)
    at forge.ai.ComputerUtilCombat.willOpposingCreatureDieInCombat(ComputerUtilCombat.java:2581)
    at forge.ai.ComputerUtil.shouldSacrificeThreatenedCard(ComputerUtil.java:3265)
    at forge.ai.ComputerUtil.lambda$getCardPreference$0(ComputerUtil.java:428)
    at com.google.common.collect.Iterators$5.computeNext(Iterators.java:673)
    at com.google.common.collect.AbstractIterator.tryToComputeNext(AbstractIterator.java:145)
    at com.google.common.collect.AbstractIterator.hasNext(AbstractIterator.java:140)
    at forge.util.collect.FCollection.addAll(FCollection.java:339)
    at forge.util.collect.FCollection.<init>(FCollection.java:105)
    at forge.game.card.CardCollection.<init>(CardCollection.java:115)
    at forge.game.card.CardLists.filter(CardLists.java:318)
    at forge.ai.ComputerUtil.getCardPreference(ComputerUtil.java:428)
```

Agetian:
Hmm, a stack overflow error, I wonder how that occurred, must have run into some kind of an infinite loop, will need to think as to how exactly that could have happened

Agetian:
Do you have a board state or do you remember anything about what could have been happening when this occurred?

Agetian:
By the way, this is strange, but the line numbers don't seem to correspond to the current state of the branch... E.g. ComputerUtilCombat.java:2581 is a closing brace inside checkAttackerLifelinkDamage (instead of willOpposingCreatureDieInCombat), ComputerUtilCombat.java:1823 shows a call to get net toughness of a blocker for Flanking :/

tool4ever:
you can check Sentry for details, the line numbers might happen because snapshots with at least some parts(?) are already uploaded unfortunately so it's hard to tell what else exactly might be included
seems like it was testing playability of _Corrupted Harvester_, hopefully you can reproduce it first to devise some logic that catches the case that's not super hardcoded but still allows what you want at other times

Agetian:
Ok, I'll see if I can reproduce it with Corrupted Harvester :)

tool4ever:
since you guys seems highly eager to get this merged asap the quick solution might be to eliminate the possibility fully by not calling `shouldSacrificeThreatenedCard` from `getCardPreference`?

Agetian:
I'm currently experimenting with Corrupted Harvester but I can't get it into an infinite loop yet... I found something silly it can do though, I'll see if I can fix that as a part of patching this up - it can try to sacrifice itself in an attempt to regenerate the creature, lol, but even that doesn't make it smash the stack :) Still looking for that stack-smashing condition.

tool4ever:
the trace implies it happened during a non-trivial combat scenario (so all those checks have a reason to continue)

Agetian:
So far, can't reproduce it in combat, unfortunately, either... Tried a few things with a variety of blocks, but can't recreate a situation where it would start to loop :/ Will keep trying. I don't like the idea of just removing that call, it'll basically make 75% of situations (even the basic ones such as sacrificing a creature to a Bubbling Cauldron in order to gain life fail to work. We need to somehow figure out a better solution, I just wish I knew how or at least could repro the crash so that I can experiment with it and trap things in the debugger :)

Agetian:
The worst part of it is that I have no idea if this is still a possible infinite loop since I'm not sure what state the branch was in when the snapshot went live... I made quite a lot of subtle modifications one way or the other in the last few days, no idea if something still lurks out there or maybe it's a wild goose chase :( At any rate, for now I'll assume that it's still the case that this is reproducible and possible in some weird scenario.

Agetian:
No worries, I actually think I recreated it ;) Not sure if it's the same one, but it is indeed a stack overflow. I'll see if it's consistent, and if it is, I'll play around in the debugger, see what I see, and think how to best fix it hopefully with minimal reverts to the logic itself :)

Agetian:
Hmm, so far, I only know how to make it work for non-combat predictions... Combat predictions run into a loop, no matter what I try. Basically, if we only check the non-combat conditions in shouldSacrificeThreatenedCard, it avoids the issue, as no other calls can enter a recursive loop. I'm trying to find a better place for the combat prediction (maybe moving the sacrifice cost check that seems to be the culprit - at ComputerUtil.java:431 - but I don't see a good place for it :/ )
Here's the game state to test if you're interested - basically, let the AI attack and then block the Corrupted Harvester with a 3/3 and the other creature with the flying creature). I wish you could help with this one, tbh.
[ctd_corrupted_harvester.txt](https://github.com/Card-Forge/forge/files/13229993/ctd_corrupted_harvester.txt)

Agetian:
It looks like the loop happens because of the combination of the threatened card sacrifice and the regenerate check in particular - the latter causes another call to the former, which later gets to the former again through the combat routines :/

Agetian:
Ok, used a trick similar to the one used to suppress combat regen check in ComputerUtilCombat.java , it's sort of hackish but seems to do the job :/

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -433,12 +427,41 @@ public static Card getCardPreference(final Player ai, final Card activate, final
                 final CardCollection sacMeList = CardLists.filter(typeList, new Predicate<Card>() {
                     @Override
                     public boolean apply(final Card c) {
-                        return c.hasSVar("SacMe") && Integer.parseInt(c.getSVar("SacMe")) == priority;
+                        return (c.hasSVar("SacMe") && Integer.parseInt(c.getSVar("SacMe")) == priority)
+                                || (priority == 1 && shouldSacrificeThreatenedCard(ai, c, sa));
```
kevlahnota, line 431: sa.getAPI() NPE on 
```
shouldSacrificeThreatenedCard(ai, c, sa)
```
because of 
```
getCardPreference(ai, activate, pref, typeList, null)
```
 line above... hence NPE is thrown.. 
 
image

Agetian: Thanks! :) Fixed.

Agetian:
Playtested some more in drafts and vs. lots of different constructed decks, don't see anything else immediately wrong with it. Think anything else needs doing here, guys?

kevlahnota:
> Playtested some more in drafts and vs. lots of different constructed decks, don't see anything else immediately wrong with it. Think anything else needs doing here, guys?

Classcast... I think the method should always check if the controller is AI by default..
Screenshot_2023-11-03-17-47-03-91_40deb401b9ffe8e1df2f1cc5ba480b12

forge-gui/res/cardsfolder/a/anticognition.txt
```diff
@@ -1,7 +1,7 @@
 Name:Anticognition
 ManaCost:1 U
 Types:Instant
-A:SP$ Pump | Cost$ 1 U | IsCurse$ True | TargetType$ Spell | TgtZone$ Stack | TgtPrompt$ Select target creature or planeswalker spell | ValidTgts$ Creature,Planeswalker | SubAbility$ DBBranch | StackDescription$ SpellDescription | SpellDescription$ Counter target creature or planeswalker spell unless its controller pays {2}. If an opponent has eight or more cards in their graveyard, instead counter that spell, then scry 2.
+A:SP$ Branch | BranchConditionSVar$ X | TargetType$ Spell | TgtZone$ Stack | ValidTgts$ Creature,Planeswalker | BranchConditionSVarCompare$ GE8 | TrueSubAbility$ CounterScry | FalseSubAbility$ CounterUnless | AILogic$ BranchCounter | SpellDescription$ Counter target creature or planeswalker spell unless its controller pays {2}. If an opponent has eight or more cards in their graveyard, instead counter that spell, then scry 2.
 SVar:DBBranch:DB$ Branch | BranchConditionSVar$ X | BranchConditionSVarCompare$ GE8 | TrueSubAbility$ CounterScry | FalseSubAbility$ CounterUnless | StackDescription$ None
```
tool4ever, line 5: this line is no longer used

Agetian: Thanks! Removed :)

Agetian:
> > Playtested some more in drafts and vs. lots of different constructed decks, don't see anything else immediately wrong with it. Think anything else needs doing here, guys?
> 
> Classcast... I think the method should always check if the controller is AI by default.. 

Yep, thanks! Fixed. Made one method only run for the actual AI and also added a guard in the called method just in case some other routine ever requests (now or in the future) the AI profile flag for the human player.

tool4ever:
Seeing how 4 not so obvious crashes were still lurking I'd recommend we wait another 24-48 hours for a new round of sentry snapshots (since kev seems willing to give them new updates to this branch) 🤔 
That should also give me enough time for some more manual testing

kevlahnota:
> Seeing how 4 not so obvious crashes were still lurking I'd recommend we wait another 24-48 hours for a new round of sentry snapshots (since kev seems willing to give them new updates to this branch) 🤔 That should also give me enough time for some more manual testing

That's the point of adding it on snaps.. it looks fine for me but some bugs will appear if there's any.. 3 people testing is not enough I guess

Agetian:
Yeah, sure, let's wait for a couple more days and see if anything unexpected shows up - it's the weekend, too, so I should have time to do the fixing when and if needed ;)

forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
```diff
@@ -358,11 +366,13 @@ public static boolean checkSacrificeCost(final Player ai, final Cost cost, final
                     }
                     if (source.isCreature()) {
                         // e.g. Sakura Tribe-Elder
+                        final Combat combat = ai.getGame().getCombat();
                         final boolean beforeNextTurn = ai.getGame().getPhaseHandler().is(PhaseType.END_OF_TURN) && ai.getGame().getPhaseHandler().getNextTurn().equals(ai);
-                        final boolean creatureInDanger = ComputerUtil.predictCreatureWillDieThisTurn(ai, source, sourceAbility);
+                        final boolean creatureInDanger = ComputerUtil.predictCreatureWillDieThisTurn(ai, source, sourceAbility, false)
+                                && !ComputerUtilCombat.willOpposingCreatureDieInCombat(ai, source, combat);
                         final int lifeThreshold = (((PlayerControllerAi) ai.getController()).getAi().getIntProperty(AiProps.AI_IN_DANGER_THRESHOLD));
```
tool4ever, line 373: this here is another unsafe cast which will crash it on certain paths
I think we really need to come up with some clean solution to this as follow-up, this is turning into quite the patchwork with very difficult to debug outcomes

kevlahnota: so computerutil is used by automation/prediction even the controller is not AI? there should be a check on the controller on every method that casts PlayerControllerAI to prevent unsafe cast

tool4ever: E. g. those static booleans are rather problematic because they're obviously shared across the whole application

this has the side effect of "breaking contract" for independent game threads since AI thinking in Game 1 could toggle the field for AI in Game 2 at an unfortunate time
- now most users might not have two games active at the same time but how do tournaments get played out in the background?
- I'm also not 100% sure such race conditions couldn't already happen in a single game with more than one AiController

kevlahnota: hmm I thought the method call is per Game, so AIController is shared? that's not good since it will affect the other game at least for desktop version.

tool4ever: `PlayerControllerAi + AiController` are exclusive objects for each Game and AI player
but you can always add static variables which by definition exist only once and then behave as a poor version of information sharing between classes

Agetian: Yeah, agreed, though solving this globally goes beyond the scope of this MR, I'm afraid (and may require quite a bit of planning and assistance, too). I'll add a safety check here and we can think as to how to solve the bigger issue as a follow-up task :)

Agetian: Done :)

tool4ever (approved):
Alright, I hope the worst problems got sorted out...

Should also try and keep the next one a bit smaller or less of a mix imo (the BranchAi stuff alone is a bit debatable to me though it's probably not worth poking too hard at it since it's contained behind logic param)

Agetian:
Yeah, the BranchAi implementation is very specific to those two particular cards and may break if the implementation is changed (and will definitely need rework if it's to be made more generic for anything else). I don't have any bright ideas as to how to do it better at the moment though, sadly... it needs to read very specific things from both the root and the subability to understand what's going on, and it needs to run the entire CounterAi logic from the Branch, basically... I might give it some more thought later. For now, it's alright as is I guess, at least for this particular scenario and implementation.

And yeah, this MR definitely got too fat for its own good, haha :) I generally tend to keep to smaller ones, yeah :)

PR #4028, opened 2023-10-29
A temporary endgame tweak for a little bit more "flow" until the real endgame is committed after the revamp.

Agetian:


Simisays (approved):
👍 thanks !

PR #4029, opened 2023-10-29
WHO: become_the_pilot.txt (and some tidying)

Northmoc:


forge-gui/res/cardsfolder/t/tromokratis.txt
```diff
@@ -3,5 +3,5 @@ ManaCost:5 U U
 Types:Legendary Creature Kraken
 PT:8/8
 S:Mode$ Continuous | Affected$ Card.Self+notattacking+notblocking | AddKeyword$ Hexproof | Description$ CARDNAME has hexproof unless it's attacking or blocking.
-K:CARDNAME can't be blocked unless all creatures defending player controls block it.
+S:Mode$ MinMaxBlocker | ValidCard$ Card.Self | Min$ Count$Valid Creature.DefenderCtrl | Secondary$ True | Description$ CARDNAME can't be blocked unless all creatures defending player controls block it. (If any creature that player controls doesn't block this creature, it can't be blocked.)
```
tool4ever, line 6: why secondary?

tool4ever, line 6: @Northmoc 
somehow I don't think this can work in all cases, because DefenderCtrl needs it to be an declared attacker already
but this check is usually run preemptively...

forge-gui/res/cardsfolder/upcoming/become_the_pilot.txt
```diff
@@ -0,0 +1,9 @@
+Name:Become the Pilot
+ManaCost:3 U U
+Types:Enchantment Aura
+K:Enchant noncommander creature
+A:SP$ Attach | ValidTgts$ Creature.IsNotCommander | TgtPrompt$ Select target noncommander creature | AILogic$ GainControl
+S:Mode$ Continuous | Affected$ Creature.EnchantedBy | GainControl$ You | Description$ You control enchanted creature.
+S:Mode$ Continuous | Affected$ Creature.EnchantedBy | AddPower$ 2 | AddToughness$ 2 | Description$ Enchanted creature gets +2/+2 and can't be blocked unless it's attacking its owner or a permanent its owner controls.
+S:Mode$ CantBlockBy | ValidAttacker$ Creature.EnchantedBy | ValidBlocker$ Creature.!ControlledBy EnchantedOwner | Secondary$ True
```
tool4ever, line 8: don't need secondary without text?

forge-gui/res/cardsfolder/upcoming/become_the_pilot.txt
```diff
@@ -0,0 +1,9 @@
+Name:Become the Pilot
+ManaCost:3 U U
+Types:Enchantment Aura
+K:Enchant noncommander creature
+A:SP$ Attach | ValidTgts$ Creature.IsNotCommander | TgtPrompt$ Select target noncommander creature | AILogic$ GainControl
+S:Mode$ Continuous | Affected$ Creature.EnchantedBy | GainControl$ You | Description$ You control enchanted creature.
```
tool4ever, line 6: @Hanmac
the GainControl$ could be merged with the other or does it matter?

Hanmac: From the oracle it's two abilities

forge-gui/res/cardsfolder/t/tromokratis.txt
Northmoc: The check does run a whole bunch, too

Is there a case I should check?

tool4ever: this call in `AiAttackController.doAssault()` is done before attackers get declared:
image

so maybe it miscalculates, though arguably it isn't that big of a deal

Northmoc: This restores the old code without going back to the keyword... and now Player defender is actually used in applyMinMaxBlockerAbility, so that's nice

Seems like it should be no worse than before now?

tool4ever: seems good :)

tool4ever (approved):

PR #4043, opened 2023-10-31
LCI: deeproot_pilgrimage.txt and support (TapAllTrigger)

Northmoc:


forge-game/src/main/java/forge/game/ability/effects/RegenerationEffect.java
```diff
@@ -25,6 +30,7 @@ public void resolve(SpellAbility sa) {
             c.setDamage(0);
             c.setHasBeenDealtDeathtouchDamage(false);
             c.tap(true, cause, c.getController());
```
Hanmac, line 32: Shouldn't tap return bool to check if it has changed?

Northmoc: Currently Card.tap is void – want me to refactor it to boolean?

Hanmac: Yeah boolean would be better to return if has really changed

Northmoc: Oops thought I pushed this update

forge-game/src/main/java/forge/game/cost/CostPartWithList.java
```diff
@@ -136,9 +140,17 @@ protected boolean executePayment(Player payer, SpellAbility ability, CardCollect
             }
             cardList.addAll(doListPayment(payer, ability, targetCards, effect));
         } else {
+            CardCollection tapped = new CardCollection();
```
tool4ever, line 143: you could try to implement `canPayListAtOnce()` for `CostTapType`
then this can moved there for cleaner structure

Northmoc: Seems to work!

Hanmac (approved):

PR #4045, opened 2023-10-31
WHO 5 cards

Glorax:


forge-gui/res/cardsfolder/upcoming/the_flood_of_mars.txt
```diff
@@ -0,0 +1,12 @@
+Name:The Flood of Mars
+ManaCost:2 U U
+Types:Creature Alien Zombie Horror
+PT:3/3
+K:Islandwalk
+T:Mode$ Attacks | ValidCard$ Card.Self | TriggerZones$ Battlefield | Execute$ DBCounter | TriggerDescription$ Water Always Wins — Whenever CARDNAME attacks, put a flood counter on another target creature or land. If it's a creature, it becomes a copy of The Flood of Mars. If it's a land, it becomes an Island in addition to its other types.
+SVar:DBCounter:DB$ PutCounter | ValidTgts$ Creature.Other,Land.Other | AITgts$ Creature.OppCtrl,Land.OppCtrl | TgtPrompt$ Select target creature or land | CounterType$ FLOOD | SubAbility$ TrigCopy
+SVar:TrigCopy:DB$ Clone | ConditionDefined$ ParentTarget | ConditionPresent$ Card.Creature | Defined$ TriggeredSourceLKICopy | CloneZone$ Battlefield | CloneTarget$ ParentTarget | SubAbility$ TrigAnimate
```
Glorax, line 8: Currently, everything that comes after placing the counter isn't working.

forge-gui/res/cardsfolder/upcoming/regenerations_restored.txt
```diff
@@ -0,0 +1,11 @@
+Name:Regenerations Restored
+ManaCost:W U
+Types:Enchantment
+K:Vanishing:12
+T:Mode$ CounterRemovedOnce | ValidCard$ Card.Self | CounterType$ TIME | TriggerZones$ Battlefield | Execute$ TrigScry | TriggerDescription$ Whenever one or more time counters are removed from CARDNAME, scry 1 and you gain 1 life. Then if CARDNAME has no time counters on it, exile it. When you do, take an extra turn after this one.
+SVar:TrigScry:DB$ Scry | ScryNum$ 1 | SubAbility$ DBGainLife
+SVar:DBGainLife:DB$ GainLife | LifeAmount$ 1 | SubAbility$ DBExile
+SVar:DBExile:DB$ ChangeZone | Origin$ Battlefield | Destination$ Exile | Defined$ ReplacedCard | ConditionPresent$ Card.Self+counters_GE0_TIME | SubAbility$ DBExtraTurn
+SVar:DBExtraTurn:DB$ AddTurn | Defined$ You | NumTurns$ 1 | ConditionPresent$ Card.Self+counters_GE0_TIME | SpellDescription$ Take an extra turn after this one.
```
Glorax, line 9: Is there a more efficient way to write this without including the condition twice?

forge-gui/res/cardsfolder/upcoming/martha_jones.txt
```diff
@@ -0,0 +1,14 @@
+Name:Martha Jones
+ManaCost:2 U
+Types:Legendary Creature Human Cleric
+PT:3/2
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigInvestigate | TriggerDescription$ Woman Who Walked the Earth — When CARDNAME enters the battlefield, investigate. (Create a Clue token. It's an artifact with "{2}, Sacrifice this artifact: Draw a card.")
+SVar:TrigInvestigate:DB$ Investigate
+T:Mode$ Sacrificed | ValidPlayer$ You | ValidCard$ Clue | TriggerZones$ Battlefield | Execute$ DBUnblockable | TriggerDescription$ Whenever you sacrifice a Clue, CARDNAME and up to one other target creature can't be blocked this turn.
+SVar:DBUnblockable:DB$ Effect | RememberObjects$ Self | ForgetOnMoved$ Battlefield | StaticAbilities$ Unblockable | SubAbility$ TrigPump
```
tool4ever, line 8: should still combine the effects by remembering both

forge-gui/res/cardsfolder/upcoming/regenerations_restored.txt
tool4ever: ReplacedCard doesn't make sense outside of replacement effects
"when you do" is always reflexive trigger

forge-gui/res/cardsfolder/upcoming/the_flood_of_mars.txt
tool4ever: `Defined$ Self` is good enough

forge-gui/res/cardsfolder/upcoming/regenerations_restored.txt
Northmoc: In other cases you could use BranchEffect but in this case (as TRT mentioned) you'd use ImmediateTrigger

forge-gui/res/cardsfolder/upcoming/reverse_the_polarity.txt
```diff
@@ -0,0 +1,9 @@
+Name:Reverse the Polarity
+ManaCost:1 U U
+Types:Instant
+A:SP$ Charm | Cost$ 1 U U | Choices$ DBCounter,DBSwitch,DBUnblockable
```
Northmoc, line 4: No need for Cost param unless it is different than ManaCost

forge-gui/res/cardsfolder/upcoming/martha_jones.txt
Glorax: Is there an example script that does this that I could use as a reference?

forge-gui/res/cardsfolder/upcoming/regenerations_restored.txt
Glorax: I updated the script last night to use an ImmediateTrigger. However, I'm running into an issue where it gets sacrificed to Vanishing before the exile effect is able to take effect. Reordering the effects manually works, but if you just click "auto" you don't get the extra turn. Is there any way to get it to stay true to the rules as written while making it so that it prioritizes exiling it before sacrificing it by default?

(I'll be able to get the updated script loaded on here in an hour or so.)

forge-gui/res/cardsfolder/upcoming/reverse_the_polarity.txt
Glorax: It's weird then, because I've seen a lot of scripts that are written exactly like this.

forge-gui/res/cardsfolder/upcoming/the_flood_of_mars.txt
Glorax: Thanks! I'll try that.

forge-gui/res/cardsfolder/upcoming/regenerations_restored.txt
tool4ever: no, that's how it works by the rules

forge-gui/res/cardsfolder/upcoming/martha_jones.txt
tool4ever: just do `RememberObjects$ Self,Targeted` on L9 and get rid of 8

forge-gui/res/cardsfolder/upcoming/reverse_the_polarity.txt
tool4ever: that's exactly it
just nobody bothered enough to make that big cleanup yet

forge-gui/res/cardsfolder/upcoming/regenerations_restored.txt
Glorax: Are you saying that what I've described isn't how it should work, or that it's exactly how it should work?

tool4ever: Both CounterRemoved* events here happen while you're resolving the upkeep trigger and then
> 603.3b. If multiple abilities have triggered since the last time a player received priority, the abilities are placed on the stack in a two-part process. First, each player, in APNAP order, puts each triggered ability they control with a trigger condition that isn't another ability triggering on the stack in any order they choose.

I think the GUI supports remembering of the last order though

When you choose to sacrifice it first you will have a new object in your graveyard that the remaining trigger from the battlefield version is by no means directly involved with and therefore the event for the reflexive trigger doesn't happen
It requires explicit exceptions when that is supposed to work (see 400.7)

forge-gui/res/cardsfolder/upcoming/the_flood_of_mars.txt
```diff
@@ -0,0 +1,12 @@
+Name:The Flood of Mars
+ManaCost:2 U U
+Types:Creature Alien Zombie Horror
+PT:3/3
+K:Islandwalk
+T:Mode$ Attacks | ValidCard$ Card.Self | TriggerZones$ Battlefield | Execute$ DBCounter | TriggerDescription$ Water Always Wins — Whenever CARDNAME attacks, put a flood counter on another target creature or land. If it's a creature, it becomes a copy of The Flood of Mars. If it's a land, it becomes an Island in addition to its other types.
+SVar:DBCounter:DB$ PutCounter | ValidTgts$ Creature.Other,Land.Other | AITgts$ Creature.OppCtrl+notnamedThe Flood of Mars,Land.OppCtrl+nonIsland | TgtPrompt$ Select target creature or land | CounterType$ FLOOD | SubAbility$ TrigCopy
+SVar:TrigCopy:DB$ Clone | ConditionDefined$ ParentTarget | ConditionPresent$ Creature | Defined$ Self | CloneZone$ Battlefield | CloneTarget$ ParentTarget | Execute$ TrigAnimate
+SVar:TrigAnimate:DB$ Animate | ConditionDefined$ ParentTarget | ConditionPresent$ Land | Type$ Island | Duration$ Permanent | IsCurse$ True
```
Glorax, line 9: Now the copying is working, but the TrigAnimate isn't working.

forge-gui/res/cardsfolder/upcoming/regenerations_restored.txt
Glorax, line 1: Requires confirmation that this is actually compliant with the rules.

forge-gui/res/cardsfolder/upcoming/the_flood_of_mars.txt
tool4ever: well you changed it to Execute$ there which doesn't fit

forge-gui/res/cardsfolder/upcoming/the_flood_of_mars.txt
```diff
@@ -0,0 +1,12 @@
+Name:The Flood of Mars
+ManaCost:2 U U
+Types:Creature Alien Zombie Horror
+PT:3/3
+K:Islandwalk
+T:Mode$ Attacks | ValidCard$ Card.Self | TriggerZones$ Battlefield | Execute$ DBCounter | TriggerDescription$ Water Always Wins — Whenever CARDNAME attacks, put a flood counter on another target creature or land. If it's a creature, it becomes a copy of The Flood of Mars. If it's a land, it becomes an Island in addition to its other types.
```
tool4ever, line 6: CARDNAME

forge-gui/res/cardsfolder/upcoming/the_flood_of_mars.txt
Glorax: Not working with SubAbility$ either.

tool4ever: you're also missing Defined$ to tell it which object to affect

forge-gui/res/cardsfolder/upcoming/the_flood_of_mars.txt
```diff
@@ -0,0 +1,12 @@
+Name:The Flood of Mars
+ManaCost:2 U U
+Types:Creature Alien Zombie Horror
+PT:3/3
+K:Islandwalk
+T:Mode$ Attacks | ValidCard$ Card.Self | TriggerZones$ Battlefield | Execute$ DBCounter | TriggerDescription$ Water Always Wins — Whenever CARDNAME attacks, put a flood counter on another target creature or land. If it's a creature, it becomes a copy of CARDNAME. If it's a land, it becomes an Island in addition to its other types.
+SVar:DBCounter:DB$ PutCounter | ValidTgts$ Creature.Other,Land.Other | AITgts$ Creature.OppCtrl+notnamedThe Flood of Mars,Land.OppCtrl+nonIsland | TgtPrompt$ Select target creature or land | CounterType$ FLOOD | SubAbility$ DBCopy
+SVar:DBCopy:DB$ Clone | ConditionDefined$ ParentTarget | ConditionPresent$ Creature | Defined$ Self | CloneZone$ Battlefield | CloneTarget$ ParentTarget | SubAbility$ DBAnimate
+SVar:DBAnimate:DB$ Animate | ConditionDefined$ ParentTarget | ConditionPresent$ Land | Defined$ ParentTarget | Type$ Island | Duration$ Permanent | IsCurse$ True
```
Glorax, line 9: I've tried so many things by now, but for whatever reason, this effect refuses to turn the targeted land into an Island. I think this may be beyond my power to resolve.

tool4ever: don't worry, fixed the typo for you

Glorax: What was the issue?

tool4ever: check last commit to this branch for diff views

Glorax: Well, that was embarrassing.

Northmoc: The s or not before the `$` can be tough to catch

When you start to actually debug it becomes easier to narrow down where the mistakes must be

Hang in there, it's great to have another contributor

tool4ever (approved):

PR #4047, opened 2023-10-31
Landfall updates

edg444:
Updated 17 cards that were retroactively given the Landfall ability word. I've tested the changes I made and believe they are the only ones necessary, but if more needs to be done. let me know and I can do another request, or someone else can handle it.


Northmoc (approved):
👍

PR #4072, opened 2023-11-01
Adventure - Garruk river cleanup

jjayers99:
Fixed collision under bridge, moved all river tiles to layer "Lowest Background"


Simisays (approved):
👍

PR #4076, opened 2023-11-01
Panharmonicon: improve filtering of cause for complex trigger

### LINKED ISSUE #4109: NPE when Esix tries to replace token with TDFC backside
```
 java.lang.NullPointerException: Cannot invoke "forge.game.card.CardState.getName()" because "source" is null
	at forge.game.card.CardState.copyFrom(CardState.java:570)
	at forge.game.card.CardState.copyFrom(CardState.java:566)
	at forge.game.card.CardUtil.getLKICopy(CardUtil.java:213)
	at forge.game.card.CardUtil.getLKICopy(CardUtil.java:171)
	at forge.game.ability.effects.TokenEffectBase.makeTokenTable(TokenEffectBase.java:160)
	at forge.game.ability.effects.TokenEffectBase.makeTokenTable(TokenEffectBase.java:75)
	at forge.game.ability.effects.TokenEffect.resolve(TokenEffect.java:126)
```

@Hanmac 
the problem is it skips the transformed state
should I just try adding ReplaceToken API here?

https://github.com/Card-Forge/forge/blob/3a879c35d23d7614e1ee7342954448e6daccb8bd/forge-game/src/main/java/forge/game/card/CardFactory.java#L749

tool4ever:
Judges seemed pretty sure:
https://www.reddit.com/r/askajudge/comments/17lmope/can_naban_dean_of_iteration_double_a_trigger_with/

Closes #4109

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardFactory.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactory.java
@@ -746,7 +746,8 @@ public class CardFactory {
             result.put(CardStateName.Adventure, ret2);
         } else if (in.isTransformable() && sa instanceof SpellAbility && (
                 ApiType.CopyPermanent.equals(((SpellAbility)sa).getApi()) ||
-                ApiType.CopySpellAbility.equals(((SpellAbility)sa).getApi())
+                ApiType.CopySpellAbility.equals(((SpellAbility)sa).getApi()) ||
+                ApiType.ReplaceToken.equals(((SpellAbility)sa).getApi())
                 )) {
             // CopyPermanent can copy token
             final CardState ret1 = new CardState(out, CardStateName.Original);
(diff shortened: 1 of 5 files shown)


Hanmac:
This probably needs such checks too for the damage trigger?

IMG_0590-1

tool4ever:
I guess so
_Breeches, Brazen Plunderer_ could deal damage to opponent + itself and it shouldn't double that trigger

Currently Damage trigger support isn't available yet anyway, but it's something to keep in mind for that cards PR

tool4ever:
I fixed support for triggered mana abilities:
image

Hanmac:
Since all triggers are copied now, we should probably add the other LTB trigger too, like Exploited, Sacrificed or Destroyed

https://github.com/Card-Forge/forge/blob/3a879c35d23d7614e1ee7342954448e6daccb8bd/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java#L242

tool4ever:
@Hanmac 
I don't understand, none of them should matter as Static$...?

Hanmac:
> @Hanmac I don't understand, none of them should matter as Static$...?

None of them are static, but they should apply for the LTB check below

Hanmac (approved):

PR #4090, opened 2023-11-02
WHO City of the Daleks

Glorax:


forge-gui/res/cardsfolder/upcoming/city_of_the_daleks.txt
```diff
@@ -0,0 +1,12 @@
+Name:City of the Daleks
+ManaCost:no cost
+Types:Plane Skaro
+T:Mode$ AttackersDeclared | AttackingPlayer$ You | Execute$ TrigLoseLife | TriggerZones$ Command | TriggerDescription$ Whenever your team attacks with exactly two creatures, those creatures gain double strike until end of turn.
```
tool4ever, line 4: wrong description

forge-gui/res/tokenscripts/b_3_3_a_dalek_menace_designated.txt
Simisays, line 1: @paulsnoops I was thinking more of b_3_3_a_dalek_menace_mustattack 

Thoughts ?

Glorax: I was following the example of the Archenemy cards that have similar effects.

forge-gui/res/cardsfolder/upcoming/city_of_the_daleks.txt
```diff
@@ -0,0 +1,12 @@
+Name:City of the Daleks
+ManaCost:no cost
+Types:Plane Skaro
+T:Mode$ AttackersDeclared | AttackingPlayer$ You | Execute$ TrigLoseLife | TriggerZones$ Command | TriggerDescription$ Whenever your team attacks with exactly two creatures, those creatures gain double strike until end of turn.
+SVar:TrigLoseLife:DB$ LoseLife | LifeAmount$ X | ValidTgts$ Opponent | TgtPrompt$ Select target opponent
+T:Mode$ ChaosEnsues | TriggerZones$ Command | Execute$ TrigRepeat | TriggerDescription$ Whenever chaos ensues, for each opponent, you create a 3/3 black Dalek artifact creature token with menace that attacks that opponent this turn if able. Those tokens gain haste. Sacrifice them at the beginning of the next end step.
```
tool4ever, line 6: these old scheme cards are doing it wrong
the "attacks that opponent this turn if able" is not an ability of the token
it's an effect that gets applied to the game rules

you could look at the Encore implementation inside CardFactory as a better approach for educational purpose, but since this is also only another variant card I don't see this as high priority

Northmoc:
@Glorax invited you to join contributors – then you can claim cards in the project view to avoid any confusion

forge-gui/res/tokenscripts/b_3_3_a_dalek_menace_designated.txt
paulsnoops: yeah designated works fine

paulsnoops: Actually, can you remove the token from the PR and in the script use the token we already have: b_3_3_a_dalek_menace

I know the card says "that attacks that opponent this turn if able" but the token you have made doesn't actually exist so we can't use it.

https://scryfall.com/search?as=grid&order=name&q=dalek+type%3Atoken+%28game%3Apaper%29

Glorax: @paulsnoops So I should find a way to integrate that effect into the card then, as opposed to the token.

forge-gui/res/cardsfolder/upcoming/city_of_the_daleks.txt
tool4ever: should be fixed but still needs testing

Simisays: Seems to work in my testing 
image

tool4ever (approved):

PR #4091, opened 2023-11-03
WHO: idris_soul_of_the_tardis.txt + support

Northmoc:


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4515,6 +4515,17 @@ public final Trigger getTriggerForStaticAbility(final String str, final StaticAb
         return result;
     }
 
+    public final Trigger addTriggerForStaticAbility(final Trigger trig, final StaticAbility stAb) {
+        String str = trig.toString();
+        Trigger result = storedTrigger.get(stAb, str);
+        if (result == null) {
+            result = trig.copy(this, false);
+            addTrigger(result);
```
tool4ever, line 4523: you shouldn't need to add the trigger here manually, that the job for the static

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4515,6 +4515,17 @@ public final Trigger getTriggerForStaticAbility(final String str, final StaticAb
         return result;
     }
 
+    public final Trigger addTriggerForStaticAbility(final Trigger trig, final StaticAbility stAb) {
```
Hanmac, line 4518: Probably should add Trigger Id as key too, in case of multiple Trigger with the same Trigger are exiled

tool4ever (approved):

PR #4093, opened 2023-11-03
WHO: caught_in_a_parallel_universe.txt + support

Northmoc:


tool4ever:
Still aware of this...just haven't had the time yet to think about if there's a cleaner way to handle this (without another extra Card field) 🧐

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -112,6 +112,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
     private Map<Long, CardCollection> mustBlockCards = Maps.newHashMap();
     private List<Card> blockedThisTurn = Lists.newArrayList();
     private List<Card> blockedByThisTurn = Lists.newArrayList();
+    private Map<Player, CardCollection> chosenMap = Maps.newHashMap();
```
tool4ever, line 115: We need to think if this can just replace chosenCards but if it needs to store List<Card> instead for _Human-Time Lord Meta-Crisis_ to work correctly

forge-gui/res/cardsfolder/g/grenzos_rebuttal.txt
```diff
@@ -1,12 +1,12 @@
 Name:Grenzo's Rebuttal
 ManaCost:4 R R
 Types:Sorcery
-A:SP$ Token | Cost$ 4 R R | TokenAmount$ 1 | TokenScript$ r_4_4_ogre | TokenOwner$ You | SubAbility$ DBRepeat | SpellDescription$ Create a 4/4 red Ogre creature token. Starting with you, each player chooses an artifact, a creature, and a land from among the permanents controlled by the player to their left. Destroy each permanent chosen this way.
+A:SP$ Token | TokenScript$ r_4_4_ogre | SubAbility$ DBRepeat | SpellDescription$ Create a 4/4 red Ogre creature token. Starting with you, each player chooses an artifact, a creature, and a land from among the permanents controlled by the player to their left. Destroy each permanent chosen this way.
 SVar:DBRepeat:DB$ RepeatEach | RepeatPlayers$ Player | StartingWithActivator$ True | RepeatSubAbility$ DBChooseArtifact | SubAbility$ DBDestroy
```
tool4ever, line 5: Based on this refactor we don't really need to embed it into RepeatEach either?

tool4ever (approved):
🤞

PR #4105, opened 2023-11-04
UNF: magar_of_the_magic_strings.txt and support

Northmoc:


forge-gui/res/cardsfolder/upcoming/magar_of_the_magic_strings.txt
```diff
@@ -0,0 +1,11 @@
+Name:Magar of the Magic Strings
```
tool4ever, line 1: Need to fix it in SetStateEffect or something:
> There are no cards in this set that would turn a face-down instant or sorcery card on the battlefield face up, but some older cards can try to do this. If something tries to turn a face-down instant or sorcery card on the battlefield face up, reveal that card to show all players it’s an instant or sorcery card. The permanent remains on the battlefield face down. Abilities that trigger when a permanent turns face up won’t trigger, because even though you revealed the card, it never turned face up. 

Currently they just disappear forever 😱

forge-gui/src/main/java/forge/gui/card/CardDetailUtil.java
```diff
@@ -446,6 +446,14 @@ public static String composeCardText(final CardStateView state, final GameView g
             area.append(")");
         }
 
+        // noted name
+        if (!card.getNotedName().isEmpty()) {
```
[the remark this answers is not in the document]
tool4ever: or actually rules say:
> 707.14. One card (Magar of the Magic Strings) instructs a player to note the name of a particular card in a graveyard and create a copy of the card with the noted name. To do so, use the characteristics of that card as it last existed in the graveyard to determine the copiable values of the copy.
this means you should rather just Remember the LKI (the "last existed" part would be quite tricky and probably involve static trigger workaround but I don't see it mattering much here as copiable values in graveyard aren't really going to be modified anyway 🤔 )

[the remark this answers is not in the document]
Northmoc: Think this is sorted now! 😅

forge-gui/res/cardsfolder/upcoming/magar_of_the_magic_strings.txt
```diff
@@ -0,0 +1,11 @@
...
+SVar:DBAnimate:DB$ Animate | RememberObjects$ RememberedLKI | Defined$ Remembered | Triggers$ DealsCDTrig | Replacements$ ReplaceDies | Duration$ Permanent
+SVar:DealsCDTrig:Mode$ DamageDone | ValidSource$ Card.Self | ValidTarget$ Player | CombatDamage$ True | Execute$ TrigPlay | TriggerZones$ Battlefield | TriggerDescription$ Whenever this creature deals combat damage to a player, you may create a copy of the card with the noted name. You may cast the copy without paying its mana cost.
+SVar:TrigPlay:DB$ Play | CopyFromRemembered$ True | WithoutManaCost$ True | Optional$ True
```
tool4ever, line 8: - You should still use Imprinted too so the RememberObjects$ only grabs one card?
- Why add any extra code when CopyCard$ already exists...
- Also ValidSA$ missing
- Cleanup missing

Northmoc: RememberObjects does find / remember one card... probably a bug because of things being turned face-down? I checked it in the debugger

Similar issue with the Cleanup... nothing is remembered on Magar anyway. Another / same bug?
image

`CopyCard$` was flipping the face-down card and putting the copy on the battlefield instead of casting it, hence the reasoning for the new param

tool4ever: `Defined$ RememberedLKI | ZoneRegardless$ True | CopyCard$ True` works fine for me?
what steps are needed for error?

tool4ever (approved):

PR #4112, opened 2023-11-05
Addressed various errors and warnings per Eclipse and IDEA

tslever:


forge-lda/src/main/java/forge/lda/LDAModelGenerator.java
```diff
@@ -298,99 +282,4 @@ public static boolean topicContains(String card, List<Pair<String, Double>> topi
         return false;
     }
 
-    public static HashMap<String,List<Map.Entry<PaperCard,Integer>>> initializeCommanderFormat(){
```
tool4ever, line 301: some of these auto refactors look a bit aggressive
why delete useful functions and did you really use Java 8 as target?

tslever: While useful, initializeCommanderFormat was not used. If we need it back, we can retrieve it from the Git history. Java 8 and Java 11 checks seemed to pass.

tslever: I restored initializeCommanderFormat. Do you prefer the pull request consisting of the 6 commits above or the pull request consisting of all 11 commits?

tool4ever: I think we can just squash merge it later

qodana.yaml
```diff
@@ -0,0 +1,31 @@
+#-------------------------------------------------------------------------------#
```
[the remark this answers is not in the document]
tslever: What would proper integration entail? I removed qodana.yaml for now and added to our .gitignore. See below commits.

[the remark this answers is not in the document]
tool4ever: I'd say something that can be run with maven easily like already happening with Checkstyle so it's the same reproducible workflow for every developer
Otherwise having people just commit their favorite tool configs doesn't seem very useful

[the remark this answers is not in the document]
tslever: Apparently, I can use IDEA but not Eclipse (see ask-devs discord channel). IDEA has "Server-Side Analysis". Setting this up created qodana.yaml. It seems useful to enable all IDEA capabilities.

How do I use Checkstyle?

I have removed qodana.yaml from our .gitignore.

Are we good to go?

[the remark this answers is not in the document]
tool4ever: Checkstyle offers plugins for different IDE
but like I said it's already assigned a maven phase too
I don't use IntelliJ + looks like we already have a Findbugs config too (I know it's outdated)
But we can certainly consider upgrading to a better linter, though preferably something more open source...

forge-gui-desktop/src/main/java/forge/view/Main.java
```diff
@@ -77,23 +77,19 @@ public static void main(final String[] args) {
...
-    }
 
-    @SuppressWarnings("deprecation")
-	@Override
-    protected void finalize() throws Throwable {
         try {
             ExceptionHandler.unregisterErrorHandling();
-        } finally {
-            super.finalize();
+        } catch (java.io.IOException e) {
```
tool4ever, line 88: This here could potentially be a logic change in case of no command line arguments (since we return then and don't have this extra function)
It's nice if we can get rid of something deprecated, but the call to `unregisterErrorHandling()` seems to release log file access among other things
Though unfortunately I don't know this part here well enough to judge if this was even supposed to work like that during normal execution, but I'll try to investigate 🤔

tool4ever: All right, I'm currently assuming you stumbled upon something that hasn't been fully working for a while (at least on my setup)
Now it's a bit hard to tell how bad of a resource leak this could actually cause, but ideally the call can be duplicated during shutdown logic of the GUI to be safe
So while I don't really want to stall this - it would probably make sense if we try to address this here too
(as long as the task won't be too complex)

tslever: Would you lay out steps for us to follow?
P.S.: In running main with CLI arguments `sim -d Virtue_And_Valor.dck Virtue_And_Valor.dck`, two AI's play a game. What CLI arguments do I use to launch the GUI?

tool4ever: Uhh, that logic is just above the part you've been editing:
https://github.com/Card-Forge/forge/blob/4c0599bb52ec719e5b06cbde653cd849e88cba17/forge-gui-desktop/src/main/java/forge/view/Main.java#L57

tslever: Thank you, and for the run configuration without command-line arguments displayed below, I get the following crash report. Why do I need to specify decks when launching the GUI?

image

image

Hanmac: you need add this to the Params, taken from the forge.sh:
https://github.com/Card-Forge/forge/blob/master/forge-gui-desktop/src/main/config/forge.sh#L46C10-L46C1026

```
--add-opens java.desktop/java.beans=ALL-UNNAMED --add-opens java.desktop/java.awt.color=ALL-UNNAMED --add-opens java.desktop/javax.swing.border=ALL-UNNAMED --add-opens java.desktop/javax.swing.event=ALL-UNNAMED --add-opens java.desktop/sun.awt.image=ALL-UNNAMED --add-opens java.desktop/sun.swing=ALL-UNNAMED --add-opens java.desktop/javax.swing=ALL-UNNAMED --add-opens java.desktop/java.awt=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED --add-opens java.base/java.text=ALL-UNNAMED --add-opens java.desktop/java.awt.font=ALL-UNNAMED --add-opens java.desktop/java.awt.image=ALL-UNNAMED --add-opens java.base/jdk.internal.misc=ALL-UNNAMED --add-opens java.base/sun.nio.ch=ALL-UNNAMED --add-opens java.base/java.nio=ALL-UNNAMED --add-opens java.base/java.math=ALL-UNNAMED --add-opens java.base/java.util.concurrent=ALL-UNNAMED --add-opens java.base/java.net=ALL-UNNAMED -Dio.netty.tryReflectionSetAccessible=true 
```

tslever: @Hanmac, when I supply the CLI arguments you provided, I get

Unknown mode.
Known mode is 'sim', 'parse'

Hanmac: they need to be added to the VM Options

tslever: @Hanmac , that did it, thank you! I guess the main thing left on this pull request is to work on finalize with @tool4ever .

PR #4126, opened 2023-11-05
TriggerExcessDamageAll: add new trigger for Magmatic Galleon

Hanmac:
adds new Trigger

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardDamageMap.java
+++ b/forge-game/src/main/java/forge/game/card/CardDamageMap.java
@@ -144,8 +144,9 @@ public class CardDamageMap extends ForwardingTable<Card, GameEntity, Integer> {
                 runParams.put(AbilityKey.DamageAmount, excess);
                 runParams.put(AbilityKey.IsCombatDamage, isCombat);
                 game.getTriggerHandler().runTrigger(TriggerType.ExcessDamage, runParams, false);
+
+                damagedList.add(damaged.getKey());
             }
-            damagedList.add(damaged.getKey());
         }
 
         if (cause != null && cause.hasParam("ExcessSVar")) {
--- a/forge-game/src/main/java/forge/game/card/CardDamageMap.java
+++ b/forge-game/src/main/java/forge/game/card/CardDamageMap.java
@@ -126,13 +126,14 @@ public class CardDamageMap extends ForwardingTable<Card, GameEntity, Integer> {
 
             int excess = sum - (damaged.getKey().hasBeenDealtDeathtouchDamage() ? 1 : damaged.getValue());
 
+            // also update the DamageHistory, but overwrite previous excess outcomes
+            // because Rith, Liberated Primeval cares about who controlled it at this moment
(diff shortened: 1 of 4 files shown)


forge-game/src/main/java/forge/game/card/CardDamageMap.java
```diff
@@ -144,11 +145,20 @@ public void triggerExcessDamage(boolean isCombat, Map<Card, Integer> lethalDamag
                 runParams.put(AbilityKey.IsCombatDamage, isCombat);
                 game.getTriggerHandler().runTrigger(TriggerType.ExcessDamage, runParams, false);
             }
+            damagedList.add(damaged.getKey());
```
tool4ever, line 148: shouldn't this be inside `excess > 0`?

forge-game/src/main/java/forge/game/card/CardDamageMap.java
```diff
@@ -126,27 +126,29 @@ public void triggerExcessDamage(boolean isCombat, Map<Card, Integer> lethalDamag
...
             // also update the DamageHistory, but overwrite previous excess outcomes
             // because Rith, Liberated Primeval cares about who controlled it at this moment
-            lkiCache.get(damaged.getKey().getId()).setHasBeenDealtExcessDamageThisTurn(excess > 0);
-            if (excess > 0) {
-                if (cause != null && cause.hasParam("ExcessSVar")) {
-                    if ((!cause.hasParam("ExcessSVarCondition") || damaged.getKey().isValid(cause.getParam("ExcessSVarCondition"), cause.getActivatingPlayer(), cause.getHostCard(), cause))
-                            && (!cause.hasParam("ExcessSVarTargeted") || damaged.getKey().equals(cause.getTargetCard()))) {
-                        storedExcess += excess;
-                    }
+            lkiCache.get(damaged.getKey().getId()).setHasBeenDealtExcessDamageThisTurn(true);
```
tool4ever, line 135: like the comment above says the LKI needs to be always updated for _Rith, Liberated Primeval_ to work correctly

Hanmac: does inside the loop would ever make sense to call setHasBeenDealtExcessDamageThisTurn(false) ?

tool4ever: Rith looks in the past:
> Rith’s ability will trigger as long as an opponent controlled the creature as excess damage was dealt to it, even if the creature is no longer on the battlefield or you control it at the end of the turn.

that means it needs to be refreshed on LKI otherwise
- I control creature and it's dealt excess
- next opponent gains control of it and it's dealt damage again, but nonexcess

Rith would check DamageHistory LKI, see the opponent controlled instance as excess and trigger

Hanmac: imo Rith might need to be more fixed, but i will think about something

tool4ever (approved):

Agetian (approved):

PR #4137, opened 2023-11-07
Improve AI for Workshop Elders + fix AI paying Cycling PayLife costs in presence of Living End

### LINKED ISSUE #3837: Workshop Elders unable to animate treasure token
**Describe the bug**
I was playing a commander gauntlet against the AI. The AI was playing Tinker Time precon. The AI played Workshop Elders and then tried to animate a treasure token it had created. But the ability did nothing. The token did not become a creature.

**Expected behavior**
When a treasure token is targeted by Workshop Elders' triggered ability, it should become a 0/0 creature with 4 1/1 counters on it.

 - OS: Windows 11
 - Version 1.6.59-SNAPSHOT-09.29

### LINKED ISSUE #4128: AI kills itself with Street Wraith
**Describe the bug**
AI cycles Street Wraith with low health and die

**To Reproduce**
1. AI has 1-2 lifes
2. AI cycles Street Wraith

**Expected behavior**
AI shouldn't cycle Street Wraith with low health

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-11.03

**Screenshots**

street

Agetian:
Fixes #3837 
Fixes #4128

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -68,7 +68,10 @@ import forge.util.collect.FCollectionView;
 import io.sentry.Breadcrumb;
 import io.sentry.Sentry;
 
-import java.util.*;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 /**
  * <p>
@@ -1623,8 +1626,15 @@ public class AiController {
             AiPlayDecision aiPlayDecision = AiPlayDecision.CantPlaySa;
             if (useLivingEnd) {
                 if (sa.isCycling() && sa.canCastTiming(player)) {
-                    if (ComputerUtilCost.canPayCost(sa, player, sa.isTrigger()))
-                        aiPlayDecision = AiPlayDecision.WillPlay;
+                    if (ComputerUtilCost.canPayCost(sa, player, sa.isTrigger())) {
+                        if (sa.getPayCosts() != null && sa.getPayCosts().hasSpecificCostType(CostPayLife.class)
+                                && !player.cantLoseForZeroOrLessLife()
+                                && player.getLife() <= sa.getPayCosts().getCostPartByType(CostPayLife.class).getAbilityAmount(sa) * 2) {
+                            aiPlayDecision = AiPlayDecision.CantAfford;
+                        } else {
+                            aiPlayDecision = AiPlayDecision.WillPlay;
+                        }
+                    }
                 } else if (sa.getHostCard().hasKeyword(Keyword.CASCADE)) {
                     if (isLifeInDanger) { //needs more tune up for certain conditions
                         aiPlayDecision = player.getCreaturesInPlay().size() >= 4 ? AiPlayDecision.CantPlaySa : AiPlayDecision.WillPlay;
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -1622,10 +1622,11 @@ public class AiController {
                     continue;
                 }
             }
-            //living end AI decks
+            // living end AI decks
+            // TODO: generalize the implementation so that superfluous logic-specific checks for life, library size, etc. aren't needed
             AiPlayDecision aiPlayDecision = AiPlayDecision.CantPlaySa;
             if (useLivingEnd) {
-                if (sa.isCycling() && sa.canCastTiming(player)) {
+                if (sa.isCycling() && sa.canCastTiming(player) && player.getCardsIn(ZoneType.Library).size() >= 10) {
                     if (ComputerUtilCost.canPayCost(sa, player, sa.isTrigger())) {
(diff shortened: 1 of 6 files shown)


forge-ai/src/main/java/forge/ai/ability/AnimateAi.java
```diff
@@ -314,7 +314,8 @@ private boolean animateTgtAI(final SpellAbility sa) {
                 int aValue = ComputerUtilCard.evaluateCreature(animatedCopy);
 
                 // animated creature has zero toughness, don't do that
-                if (animatedCopy.getNetToughness() <= 0) {
+                if (animatedCopy.getNetToughness() <= 0
+                        && !(logic.equals("WithCounters") && animatedCopy.canReceiveCounters(CounterEnumType.P1P1))) {
```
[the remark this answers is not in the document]
Agetian: True :) Done!

forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -1623,8 +1626,15 @@ private SpellAbility chooseSpellAbilityToPlayFromList(final List<SpellAbility> a
             AiPlayDecision aiPlayDecision = AiPlayDecision.CantPlaySa;
             if (useLivingEnd) {
                 if (sa.isCycling() && sa.canCastTiming(player)) {
-                    if (ComputerUtilCost.canPayCost(sa, player, sa.isTrigger()))
-                        aiPlayDecision = AiPlayDecision.WillPlay;
+                    if (ComputerUtilCost.canPayCost(sa, player, sa.isTrigger())) {
```
tool4ever, line 1629: Honestly it makes no sense to me to create this "aggressive decision override" mode and now try to backport all kinds of sanity logic we'd normally benefit from into this single place
E.g. it's only a matter of time until we get a report where the AI cycles itself into a loss because the library was empty
Perhaps the way this mode gets enabled can be tweaked as part of the deck selection or something :/
sorry, I really don't want to sound too harsh and all...

Agetian: I have no idea tbh, though I do agree with the reasoning. Not sure if I'm the best person to do the latter though, as I'm not the author of the override logic and I'm not sure if I fully understand how exactly it can be wired better to avoid this kind of situation ^^;

forge-ai/src/main/java/forge/ai/ability/AnimateAi.java
```diff
@@ -313,10 +313,21 @@ private boolean animateTgtAI(final SpellAbility sa) {
...
                 int aValue = ComputerUtilCard.evaluateCreature(animatedCopy);
 
-                // animated creature has zero toughness, don't do that
-                if (animatedCopy.getNetToughness() <= 0
-                        && !(logic.equals("WithCounters") && animatedCopy.canReceiveCounters(CounterEnumType.P1P1))) {
-                    continue;
+                // animated creature has zero toughness, don't do that unless
+                if (animatedCopy.getNetToughness() <= 0) {
+                    boolean buffedToughness = false;
+                    SpellAbility sub = sa.getSubAbility();
```
Hanmac, line 319: you could try findSubAbilityByType

Agetian: True, good call! Done :)

forge-ai/src/main/java/forge/ai/AiController.java
Agetian: I checked the logic - as far as my possibilities go at the moment (given the available spare time), I can only either patch up the existing implementation and add a TODO for a better implementation to hopefully be coded in the future (either by me or by someone else), or I can revert my code in this branch and leave the implementation broken for both cases (life payment and library size) until someone else can patch it properly or until I can find enough time to dig deep into the question of how to rewire the logic. Please let me know what would be preferable.

tool4ever: hmn, let's just leave it alone like this now? 🤔 
It's not like there will be huge negative impact, besides the approach having this general function getting a bit messier with card specific stuff ofc

Agetian: Alright, sounds good, let's keep it as is for now then, and hopefully we can modify it later to have the relevant complication avoided

kevlahnota: > Honestly it makes no sense to me to create this "aggressive decision override" mode and now try to backport all kinds of sanity logic we'd normally benefit from into this single place E.g. it's only a matter of time until we get a report where the AI cycles itself into a loss because the library was empty Perhaps the way this mode gets enabled can be tweaked as part of the deck selection or something :/ sorry, I really don't want to sound too harsh and all...

I don't see any different play style/variation that use Living End based deck on this stats (Living End Decks): https://mtgdecks.net/prices/living-end

So the decision is forced to pile up graveyard and use cascade to cheat/play Living End (try removing the forced decision and see what the AI will do), The logic is somehow lacking in some aspect checks but I can't introduce more logic without slowing down the AI decision.

PR #4147, opened 2023-11-11
WHO, LCI, LCC achievements by Marek14

Agetian:
Needs some of the cards implemented.
Also, some ideas from Marek14 for the potential implementation that are currently unrealized:

```
If you satisfy [the Doctors achievement] the "real" way, with First to Thirteenth Doctor:
The Thirteen Doctors
The greatest reunion the Universe has ever seen!

Plus, I feel like there should be a hidden achievement for triggering that ability with exactly 2012 time counters.
```


paulsnoops:
@Agetian craft has been merged if you want to go ahead with this one :D

PR #4140, opened 2023-11-08
LCI - Craft implementation

Northmoc:
- closes #4098

PR #4167, opened 2023-11-14
LCI: dire_flail_dire_blunderbuss.txt + support

Northmoc:
needs #4140


forge-game/src/main/java/forge/game/ability/effects/ImmediateTriggerEffect.java
```diff
@@ -68,7 +68,9 @@ public void resolve(SpellAbility sa) {
             if (overridingSA instanceof AbilitySub) {
                 ((AbilitySub)overridingSA).setParent(null);
             }
-
+            if (sa.hasParam("CopyTriggeringObjects")) {
```
tool4ever, line 71: I made that obsolete by introducing "Spawner>..." syntax
But in this case it's only for itself anyway, so nothing should prevent you from using normal Count$

forge-gui/res/cardsfolder/upcoming/dire_flail_dire_blunderbuss.txt
```diff
@@ -0,0 +1,22 @@
...
+Oracle:Equipped creature gets +2/+0.\nEquip {1}\nCraft with artifact {3}{R}{R} ({3}{R}{R}, Exile this artifact, Exile another artifact you control or an artifact card from your graveyard: Return this card transformed under its owner's control. Craft only as a sorcery.)
+
+ALTERNATE
+
+Name:Dire Blunderbuss
+Colors:red
+Types:Artifact Equipment
+S:Mode$ Continuous | Affected$ Creature.EquippedBy | AddPower$ 3 | AddTrigger$ AttacksTrig | Description$ Equipped creature gets +3/+0 and has "Whenever this creature attacks, you may sacrifice an artifact other than CARDNAME. When you do, this creature deals damage equal to its power to target creature."
+SVar:AttacksTrig:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigImmediateTrig | TriggerDescription$ Whenever this creature attacks, you may sacrifice an artifact other than TRIGGERSOURCE. When you do, this creature deals damage equal to its power to target creature.
+SVar:TrigImmediateTrig:AB$ ImmediateTrigger | Cost$ Sac<1/Artifact.!TriggerSource/another artifact> | Execute$ TrigDamage | CopyTriggeringObjects$ True | TriggerDescription$ When you do, this creature deals damage equal to its power to target creature.
```
tool4ever, line 19: `NotDefinedOriginalHost` should already work

forge-game/src/main/java/forge/game/trigger/Trigger.java
```diff
@@ -129,6 +129,10 @@ public String toString(boolean active) {
                 desc = TextUtil.fastReplace(desc,"CARDNAME", CardTranslation.getTranslatedName(currentName));
                 desc = TextUtil.fastReplace(desc,"NICKNAME", Lang.getInstance().getNickName(CardTranslation.getTranslatedName(currentName)));
             }
+            if (desc.contains("TRIGGERSOURCE")) {
```
tool4ever, line 132: should use ORIGINALHOST like all the other ones
looks like _Nettlevine Blight_ was already expecting it to exist 🐣

tool4ever: maybe it can be moved from `replaceAbilityText` because that should always use this result anyway?

tool4ever (approved):

PR #4194, opened 2023-11-23
LCC: wayta_trainer_prodigy.txt and support

Northmoc:


forge-gui/res/cardsfolder/upcoming/wayta_trainer_prodigy.txt
```diff
@@ -0,0 +1,11 @@
+Name:Wayta, Trainer Prodigy
+ManaCost:R G W
+Types:Legendary Creature Human Warrior
+PT:1/5
+K:Haste
+A:AB$ Pump | Cost$ 2 G T | ValidTgts$ Creature.YouCtrl | TgtPrompt$ Select target creature you control | SubAbility$ DBFight | ReduceCost$ X | StackDescription$ None | SpellDescription$ Target creature you control fights another target creature.
+SVar:DBFight:DB$ Fight | Defined$ ParentTarget | ValidTgts$ Creature | TargetUnique$ True | TgtPrompt$ Select another target creature
+SVar:X:Count$Compare Y EQ2.2.0
+SVar:Y:AllTargeted$Valid Creature.YouCtrl
+S:Mode$ Panharmonicon | ValidMode$ DamageDoneOnce,DamageDone | ValidCard$ Creature.YouCtrl | Description$ If a creature you control being dealt damage causes a triggered ability of a permanent you control to trigger, that ability triggers an additional time.
```
tool4ever, line 10: uhh...there's no support for these modes?
(you'll need to consider the discussion of #4076)

Northmoc: I ran some tests with pinging 3/5 creatures on both sides of the battlefield with Wayta + Death Pits of Rath in play. The only time the Death Pits trigger was doubled was when the same player that controlled the creature getting pinged also controlled both Wayta and the Death Pits.

So I think this is working right now. More cases?

tool4ever (approved):

PR #4195, opened 2023-11-23
PIP: vats.txt and support

Northmoc:


forge-gui/src/main/java/forge/gamemodes/match/input/InputSelectTargets.java
```diff
@@ -256,6 +256,22 @@ protected final boolean onCardSelected(final Card card, final List<Card> otherCa
             }
         }
 
+        // If all cards must have equal toughness
+        if (tgt.isEqualToughness()) {
```
tool4ever, line 260: this is ok for GUI, but engine also needs logic in `SpellAbility.canTarget()`

tool4ever (approved):

PR #4204, opened 2023-11-24
PIP: Radiation mechanic implementation

Northmoc:


forge-gui/res/languages/de-DE.properties
```diff
@@ -2028,7 +2030,9 @@ lblChooseCardToManifest=Wähle zu manifestierende Karte
 #MeldEffect.java
 lblChooseCardToMeld=Wähle Karte zum Verschmelzen mit
 #MillEffect.java
-lblDoYouWantPutLibraryCardsTo=Lege Karte(n) von der Bibliothek nach
+lblDoYouWantPutLibraryCardsTo=Möchten Sie fräsen?
+lblMilledToZone=(zu {0})
+lblMilledCards=OVERRIDE Karten {0} gefräst
```
tool4ever, line 2035: it seems wrong to require hardcoded logic like OVERRIDE in a place like this, imo that's just potential for a mistranslation
why would it be needed anyway?

Northmoc: It was a hacky fix I put into PlayerController reveal a while ago to avoid the player Zone string suffix that was automatically added.

I think adding this boolean is better. Cards that had RevealTitle$ were all using OVERRIDE so better to just make the RevealTitle automatically skip the suffix for now

I probably could have added the OVERRIDE string earlier similar to the existing code in Clash, but at least until someone talks me out of it I think this way might be better.

tool4ever (approved):
I suppose we could merge it, rules for it will still take some time anyway?

PR #4221, opened 2023-11-26
Add puzzle PS_LCI01 (needs GameState update/fix)

Agetian:
For some reason, when the Saga goes to the graveyard, its front face trigger fires, as if it was untransformed. I tried updating its LKI info via game.updateLastStateForCard after setting the card to the Transformed card state, as well as forcing an update of the card in the setState call, but the net effect is the same - as soon as the card goes to the graveyard, its front face trigger fires away. I tried debugging this and got as far as GameAction#changeZone, in which the temporary "copied" object seems to have the front face characteristics (such as the Creature type, for instance). I couldn't figure out why this happens though. I could use some help here... What do you think @Hanmac @tool4ever ?

On a possibly related note, there seems to be an issue with other things not happening as they should in a somewhat similar-ish fashion. For instance, if a card that gives Hexproof to a player is put on the battlefield via a GameState setup, the player will visually have Hexproof but it will be possible for the opponent to target that said player with things. Not sure if it's related or a separate issue.

### DIFF
--- a/forge-gui/res/puzzle/PS_LCI01.pzl
+++ b/forge-gui/res/puzzle/PS_LCI01.pzl
@@ -1,0 +1,17 @@
+[metadata]
+Name:Possibility Storm - Lost Caverns of Ixalan #01
+URL:https://i1.wp.com/www.possibilitystorm.com/wp-content/uploads/2023/11/latest-scaled.jpg?ssl=1
+Goal:Win
+Turns:1
+Difficulty:Rare
+Description:Win this turn. Start in your first main phase (the 4th lore counter has just been put on Roar of the Fifth People and the ability is being put on the stack). Assume both graveyards start empty and that your opponent has no spells they could cast this turn. Remember that your solution must satisfy all possible blocking decisions. Good luck!
+[state]
+turn=1
+activeplayer=p0
+activephase=UPKEEP
+activephaseadvance=MAIN1
+p0life=20
+p0hand=Light 'Em Up;Sibling Rivalry;Kessig Flamebreather;Rampaging Geoderm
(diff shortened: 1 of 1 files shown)


tool4ever:
You're probably talking about _The Skullspore Nexus_ instead?

The problem is ChangesZoneAll trigger doesn't really have LKI support

Agetian:
Ahh, no, it's not The Skullspore Nexus. The issue here is, there's a transformed Huatli Poet on the battlefield (as Roar of the Fifth People) with the 4th Lore counter effect being put on stack. Once it resolves, the Saga goes to the graveyard and a trigger fires, as if it went to the graveyard front face up (as Huatli Poet).

tool4ever:
Well like I said the problem isn't really Huatli
(unless you're saying its own ETB trigger fires but I doubt that's the case)

Hanmac:
also Skullspore Nexus has a LKI problem in general

Agetian:
Hmm, interesting! Upon further inspection I figured out that in general case, it's not a problem (e.g. if this situation happens naturally in a game). It does, however, become an issue in the game state setup scenario, particularly when the dev advance phase feature is used (which I'm using here to transition to Main 1 and fire the Saga chapter trigger).

tool4ever:
I'm still not sure what you're talking about, can you show a screenshot?

The missing LKI problem isn't limited to GameState

Agetian:
Sure thing, here's what happens. The 4th chapter of the Saga triggers, the Saga goes to the graveyard, The Skullspore Nexus fires up and produces a token as if a creature died (and not a saga). If you do the same thing naturally in a game though (not via a set up state), it'll work fine and The Skullspore Nexus won't trigger.

Снимок экрана от 2023-11-27 15-47-22
Снимок экрана от 2023-11-27 15-47-15
Снимок экрана от 2023-11-27 15-47-04

Agetian:
I think I'll merge this, it's an engine issue, clearly, and has to be fixed somewhere either in GameState or elsewhere where The Skullspore Nexus misbehaves under certain conditions :/

PR #4228, opened 2023-11-27
WHO: the_five_doctors.txt and support (ChangeZoneEffect refactors)

Northmoc:
It's a big one but most is pretty simple refactoring:
 - better contextual choice buttons for choosing zone / library position for ChangeZoneEffect
 - translate zone choice prompt (remove param from cards) and choice buttons
 - translate search library prompt for fetch cards (remove param from cards)
 - support a dedicated SVar allowing or even forcing the alternate zone (refactor cards that used to use 2 different ChangeZoneEffects)
 - remove StackDesc `,,,` spacers before translating


forge-game/src/main/java/forge/game/ability/effects/ChangeZoneEffect.java
```diff
@@ -1248,6 +1227,18 @@ else if (origin.contains(ZoneType.Hand) && player.isOpponentOf(decider)) {
             if (sa.hasParam("ShuffleChangedPile")) {
                 CardLists.shuffle(chosenCards);
             }
+
+            if (sa.hasParam("DestinationAlternative")) {
+                Map<ZoneType, Integer> map = handleAltDest(sa, source, destination, libraryPos, decider);
+                if (!map.containsKey(destination) || map.get(destination) != libraryPos) {
+                    for (final Map.Entry<ZoneType, Integer> e : map.entrySet()) {
```
tool4ever, line 1234: might be better to just use something like `Pair` so you don't loop for no reason

forge-gui/res/cardsfolder/upcoming/the_five_doctors.txt
```diff
@@ -0,0 +1,8 @@
+Name:The Five Doctors
+ManaCost:5 G
+Types:Sorcery
+K:Kicker:5
+A:SP$ ChangeZone | Origin$ Library | OriginAlternative$ Graveyard | Destination$ Hand | DestinationAlternative$ Battlefield | DestAltSVar$ MANDATORY Count$TimesKicked | ChangeType$ Doctor | ChangeNum$ 5 | ConditionCheckSVar$ X | ConditionSVarCompare$ EQ0 | SubAbility$ DBChangeZone | SpellDescription$ Search your library and/or graveyard for up to five Doctor cards, reveal them, and put them into your hand. If you search your library this way, shuffle. If this spell was kicked, put those cards onto the battlefield instead of putting them into your hand.
```
tool4ever, line 5: One of the Origins should be both zones
so you can combine cards from them

tool4ever, line 5: bad Conditions

Northmoc: Hmm, in testing it already allowed me to mix and match cards from both zones as long as I indicated that I wanted to search both... Forge did a prompt for each zone, which seems correct with the "and/or" ? I'm comparing this to something like Ecological Appraisal.

tool4ever:
The other languages disappeared again?
I think you don't always have to rewrite your commits since we can still squash with merge

PR #4248, opened 2023-11-30
Hotel of Fears bugfix, planar types

Klisz:
Fixes a bug where, if a card exiled with Hotel of Fears isn't played before end of turn, the next player during their upkeep would lose life equal to that card's mana value plus the mana value of their own exiled card (and further stacking if that card isn't played, and so on). Also adds two planar types from WHO plane cards that are already merged but didn't update the TypeLists


tool4ever (approved):
thanks

PR #4250, opened 2023-12-01
Update unleash_the_flux.txt

Klisz:
Fixes a crash that would occur from a memory buffer overflow whenever the AbilityFactory tries to create Unleash the Flux's ability, due to the SubAbility loop in the previous script for that card


Hanmac:
@tool4ever check this one out

we might refactor it later with Hoarder's Greed?

tool4ever:
it would look cleaner without Repeat stuff, but might be tricky 🤔

Hanmac:
> it would look cleaner without Repeat stuff, but might be tricky 🤔

i mean still use repeat, but maybe have the repeat condition more clean without svars maybe?

tool4ever:
that's true
@Klisz 
can you rewrite it with `RememberLoser$`?
then you can check with `RepeatCheckSVar$ PlayerCountRemembered$HasPropertyYou`

Klisz:
Ah, thank you - I hadn't realized RememberLoser existed.

tool4ever (approved):

PR #4271, opened 2023-12-03
Basic Discover AI based on modified PlayAi logic.

Agetian:


forge-ai/src/main/java/forge/ai/ability/DiscoverAi.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public boolean confirmAction(Player ai, SpellAbility sa, PlayerActionConfirmMode mode, String message, Map<String, Object> params) {
+        Card c = (Card)params.get("Card");
+        for (SpellAbility s : AbilityUtils.getBasicSpellsFromPlayEffect(c, ai, CardStateName.Original)) { // TODO: other states for split cards and MDFC?
+            if (!sa.matchesValidParam("ValidSA", s)) {
+                continue;
+            }
+            if (s instanceof LandAbility) {
+                // might want to run some checks here but it's rare anyway
+                return true;
```
Hanmac, line 54: Discover doesn't let you play lands

forge-ai/src/main/java/forge/ai/ability/DiscoverAi.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    protected boolean doTriggerAINoCost(final Player ai, final SpellAbility sa, final boolean mandatory) {
+        return mandatory || checkApiLogic(ai, sa);
+    }
+
+    @Override
+    public boolean confirmAction(Player ai, SpellAbility sa, PlayerActionConfirmMode mode, String message, Map<String, Object> params) {
+        Card c = (Card)params.get("Card");
+        for (SpellAbility s : AbilityUtils.getBasicSpellsFromPlayEffect(c, ai, CardStateName.Original)) { // TODO: other states for split cards and MDFC?
+            if (!sa.matchesValidParam("ValidSA", s)) {
```
tool4ever, line 49: ValidSA doesn't exist here

forge-ai/src/main/java/forge/ai/ability/DiscoverAi.java
tool4ever: I think you need to return false or your get ClassCastException later if it encounters MDFC with land backside

Agetian: Ah, good call! Done.

PR #4284, opened 2023-12-04
Protection rework for Plural removal

Hanmac:
This is start of refactor to remove the Need for Plural->Singular for the Types


Hanmac:
i need to find where else Plural Creature Types are needed in a non Descriptive way,
like in Keywords

Descriptive like Description and similar can be updated later and they should not have an effect on the Game

forge-gui/res/cardsfolder/e/elite_inquisitor.txt
```diff
@@ -4,7 +4,7 @@ Types:Creature Human Soldier
 PT:2/2
 K:First Strike
 K:Vigilance
-K:Protection from Vampires
-K:Protection from Werewolves
-K:Protection from Zombies
+K:Protection:Vampire
+K:Protection:Werewolve
```
tool4ever, line 8: Werewolf

forge-gui/res/cardsfolder/g/greensleeves_maro_sorcerer.txt
```diff
@@ -2,7 +2,7 @@ Name:Greensleeves, Maro-Sorcerer
 ManaCost:3 G G
 Types:Legendary Creature Elemental
 PT:*/*
-K:Protection:Planeswalker,Wizards:Protection from planeswalkers and Wizards
+K:Protection:Planeswalker,Wizards:planeswalkers and Wizards
```
tool4ever, line 5: Wizard

forge-gui/res/cardsfolder/s/sword_of_dungeons_dragons.txt
```diff
@@ -2,7 +2,7 @@ Name:Sword of Dungeons & Dragons
 ManaCost:3
 Types:Artifact Equipment
 K:Equip:2
-S:Mode$ Continuous | Affected$ Creature.EquippedBy | AddPower$ 2 | AddToughness$ 2 | AddSVar$ SwordOfBodyAndMindCE | AddKeyword$ Protection from Rogues & Protection from Clerics | Description$ Equipped creature gets +2/+2 and has protection from Rogues and from Clerics.
+S:Mode$ Continuous | Affected$ Creature.EquippedBy | AddPower$ 2 | AddToughness$ 2 | AddSVar$ SwordOfBodyAndMindCE | AddKeyword$ Protection;Rogue & Protection:Cleric | Description$ Equipped creature gets +2/+2 and has protection from Rogues and from Clerics.
```
tool4ever, line 5: typo ;

forge-gui/res/cardsfolder/h/haktos_the_unscarred.txt
```diff
@@ -5,5 +5,5 @@ PT:6/1
 S:Mode$ MustAttack | ValidCreature$ Card.Self | Description$ CARDNAME attacks each combat if able.
 K:ETBReplacement:Other:ChooseNum
 SVar:ChooseNum:DB$ ChooseNumber | Min$ 2 | Max$ 4 | Defined$ You | Random$ True | SpellDescription$ As NICKNAME enters the battlefield, choose 2, 3, or 4 at random.
-S:Mode$ Continuous | Affected$ Card.Self | AddKeyword$ Protection:Card.cmcNEChosenNumber,Emblem:Protection from each mana value other than ChosenNumber | Description$ NICKNAME has protection from each mana value other than the chosen number.
+S:Mode$ Continuous | Affected$ Card.Self | AddKeyword$ Protection:Card.cmcNEChosenNumber,Emblem:each mana value other than ChosenNumber | Description$ NICKNAME has protection from each mana value other than the chosen number.
```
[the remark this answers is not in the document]
Hanmac: imo because it says: "cardname has protection" and not just "Protection", it should use Static Ability

forge-gui/res/cardsfolder/k/kitsune_riftwalker.txt
```diff
@@ -2,6 +2,6 @@ Name:Kitsune Riftwalker
 ManaCost:1 W W
 Types:Creature Fox Wizard
 PT:2/1
-K:Protection from Spirits
-K:Protection from Arcane
+K:Protection:Spirit
```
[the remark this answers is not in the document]
Hanmac: `Cephalid Snitch` is the problematic one, because it says:
> Sacrifice Cephalid Snitch: Target creature loses protection from black until end of turn.

this can cause "Protection from all colors" to split into Protection from the other 4"

forge-gui/res/cardsfolder/m/mirror_golem.txt
```diff
@@ -8,14 +8,14 @@ T:Mode$ ChangesZone | Origin$ Battlefield | Destination$ Any | Static$ True | Va
...
+S:Mode$ Continuous | Affected$ Card.Self | AddKeyword$ Protection:Creature | CheckSVar$ CheckCreature | SVarCompare$ GE1 | Description$ CARDNAME has protection from each of the exiled card's card types. (Artifact, creature, enchantment, instant, land, planeswalker, sorcery, and tribal are card types.)
+S:Mode$ Continuous | Affected$ Card.Self | AddKeyword$ Protection:Land | CheckSVar$ CheckLand | SVarCompare$ GE1
+S:Mode$ Continuous | Affected$ Card.Self | AddKeyword$ Protection:Enchantment | CheckSVar$ CheckEnchantment | SVarCompare$ GE1
+S:Mode$ Continuous | Affected$ Card.Self | AddKeyword$ Protection:Artifact | CheckSVar$ CheckArtifact | SVarCompare$ GE1
+S:Mode$ Continuous | Affected$ Card.Self | AddKeyword$ Protection:Instant | CheckSVar$ CheckInstant | SVarCompare$ GE1
+S:Mode$ Continuous | Affected$ Card.Self | AddKeyword$ Protection:Sorcery | CheckSVar$ CheckSorcery | SVarCompare$ GE1
+S:Mode$ Continuous | Affected$ Card.Self | AddKeyword$ Protection:Planeswalker | CheckSVar$ CheckPlaneswalker | SVarCompare$ GE1
+S:Mode$ Continuous | Affected$ Card.Self | AddKeyword$ Protection:Tribal | CheckSVar$ CheckTribal | SVarCompare$ GE1
```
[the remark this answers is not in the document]
tool4ever: We should probably add Battle as type too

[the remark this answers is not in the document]
Hanmac: yeah that too

Hanmac:
@tool4ever can you help me if there are other keywords using plural of Creature Types?

tool4ever:
Looks like this might have been the last one :)
Other candidates like Affinity are already done singular

Hanmac:
as next Step, we could do make this:
https://github.com/Card-Forge/forge/blob/331006e6294c637018859a64c2d234f24d343c66/forge-game/src/main/java/forge/game/ability/AbilityUtils.java#L3035
only apply for Descriptive Properties

while that might not prevent the Descriptions from looking funky, it should stop Cards from crashing

Hanmac:
in the future we could update the Description into something like this: `{plt:Vampire}` => `Vampires` to have this only done for GUI, but that's probably something @Northmoc needs to look into

forge-core/src/main/java/forge/card/CardType.java
```diff
@@ -988,6 +988,8 @@ public static boolean isAPlanarType(final String cardType) {
      * Otherwise, simply return the input.
      * @param type a String.
      * @return the corresponding type.
+     *
+     * @deprecated
      */
     public static final String getSingularType(final String type) {
```
Hanmac, line 994: imo this can be removed?

forge-gui/src/main/java/forge/gui/card/CardDetailUtil.java
```diff
@@ -345,8 +345,7 @@ public static String composeCardText(final CardStateView state, final GameView g
 
             for (final Entry<String, String> e : Sets.union(changedColorWords.entrySet(), changedTypes.entrySet())) {
                 // ignore lower case and plural form keys, to avoid duplicity
-                if (Character.isUpperCase(e.getKey().charAt(0))
-                        && !CardType.Constant.singularTypes.containsKey(e.getKey())) {
+                if (Character.isUpperCase(e.getKey().charAt(0))) {
```
Hanmac, line 348: that check can probably be removed too?

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -3032,7 +3032,9 @@ public static final String applyTextChangeEffects(final String def, final boolea
         }
```
tool4ever, line 3032: because the ChangedWords only use uppercase now we need to add lowercase version above like Any already has
for types it should be fine because they're all uppercase anyway?

but that alone isn't enough, it looks like `changedCardKeywordsByWord` isn't used anywhere

Hanmac: i fixed color word replacement above

tool4ever: it will modify the KW now, but it's still not calculating the new ones

Hanmac: it seems to have been broken for a bit longer?
changedCardKeywordsByWord needs to be added to getChangedCardKeywordsList

tool4ever:
@Hanmac 
I fixed this case:
_Artifical Evolution_ on spell _Breath Weapon_ so it can damage _Shivan Dragon_

the problem was it only updated the host stuff and not active Spell from it

tool4ever:
oh right, thanks fixed

tool4ever (approved):

forge-core/src/main/java/forge/card/CardType.java
Hanmac: normally you only want Singular to Plural for Gui stuff

I don't know of cases where you might want Plural to Singular

Hanmac:
Is this ready to merge?

tool4ever:
I think so

Hanmac:
@Agetian or @tehdiplomat  your comments? 

@Northmoc maybe you too? are the Protection Image Keys updated correctly?

Agetian (approved):

Agetian:
Looks good to me! :) :+1:

PR #4303, opened 2023-12-05
SLD: the_fourteenth_doctor.txt + support

Klisz:


forge-game/src/main/java/forge/game/ability/effects/CloneEffect.java
```diff
@@ -202,4 +203,27 @@ public void run() {
         }
     }
 
+    protected void addPumpUntil(SpellAbility sa, final Card c, long timestamp) {
```
tool4ever, line 206: Shouldn't need to duplicate that code
You could just make the one in TokenEffectBase be static or something

tool4ever (approved):
thanks

PR #4310, opened 2023-12-06
Minor typo fixes patch + Ent-Draught Basin tap fix

edg444:
I was going to do a small typo patch but found the Ent-Draught Basin error partway through, so hopefully it's okay to combine the fixes.

1 FUNCTIONAL CHANGE
Ent-Draught Basin now must tap to activate its ability, as it should be.

8 NON-FUNCTIONAL CHANGES
- Fixed six minor typos for the SpellDescription$ of various cards. These all show up fine in Deck Editor, but in the game, the SpellDescription takes over and there are typos in some of those. None of these are functional changes. See individual description for Graveyard Marshal.

- Changed two rarities (Greed C21 and Momentary Blink KHC).


tool4ever (approved):
thanks

PR #4333, opened 2023-12-09
MKM 8 cards

Simisays:


forge-gui/res/cardsfolder/upcoming/rakdos_patron_of_chaos.txt
```diff
@@ -0,0 +1,12 @@
+Name:Rakdos, Patron of Chaos
+ManaCost:4 B R
+Types:Legendary Creature Demon
+PT:6/6
+K:Flying
+K:Trample
+T:Mode$ Phase | Phase$ End of Turn | TriggerZones$ Battlefield | ValidPlayer$ You | Execute$ TrigSacrificeOpp | TriggerDescription$ At the beginning of your end step, target opponent may sacrifice two nonland, nontoken permanents. If they don't, you draw two cards.
+SVar:TrigSacrificeOpp:DB$ Sacrifice | ValidTgts$ Opponent | Amount$ 2 | Optional$ True | SacValid$ Permanent.nonLand+nonToken | SacMessage$ two nonland, nontoken permanents | RememberSacrificed$ True | SubAbility$ DBDraw
```
tool4ever, line 8: make sure only one sacrifice can't happen

Simisays: hmm any ideas how to do that. I've tried with LimitMin but haven't had much succes

tool4ever: I think the SacrificeEffect should have a param for it but maybe UnlessCost is cleaner anyway?

Simisays: I've got it to work with unlesscost but It seems like the trigger doesn't trigger anymore when the ai controls it

tool4ever: Pump as helper often needs Logic hints like Curse to work correctly
otherwise try debugging it?

Simisays: I'll remove this one for now. Forge doesn't give any errors it just skips the trigger for some reason when the ai controls the card

tool4ever: it's probably because you used wrong param: IsCurse$ should work

Simisays: you're right 
image

tool4ever (approved):

PR #4371, opened 2023-12-13
Improved AI for Sixth Sense effects

Klisz:
Improves targeting AI for Combat Research, Catlike Curiosity, Sixth Sense, and Snake Umbra, all of which are currently regarded by the AI as being Curiosity effects but have the distinction that the *enchanted creature's* controller, rather than the Aura's as in Curiosity itself, draws the card, and thus the AI should usually prioritize enchanting its own creatures.

### DIFF
--- a/forge-gui/res/cardsfolder/c/combat_research.txt
+++ b/forge-gui/res/cardsfolder/c/combat_research.txt
@@ -2,7 +2,7 @@ Name:Combat Research
 ManaCost:U
 Types:Enchantment Aura
 K:Enchant creature
-A:SP$ Attach | ValidTgts$ Creature | AILogic$ Curiosity | AITgts$ Creature.Legendary
+A:SP$ Attach | ValidTgts$ Creature | AILogic$ Curiosity | AITgts$ Creature.Legendary+YouCtrl
 S:Mode$ Continuous | Affected$ Creature.EnchantedBy | AddTrigger$ TrigDamageDone | Description$ Enchanted creature has "Whenever this creature deals combat damage to a player, draw a card."
 SVar:TrigDamageDone:Mode$ DamageDone | ValidSource$ Card.Self | ValidTarget$ Player | CombatDamage$ True | Execute$ TrigDraw | TriggerDescription$ Whenever this creature deals combat damage to a player, draw a card.
 SVar:TrigDraw:DB$ Draw
--- a/forge-gui/res/cardsfolder/m/mischievous_catgeist_catlike_curiosity.txt
+++ b/forge-gui/res/cardsfolder/m/mischievous_catgeist_catlike_curiosity.txt
@@ -16,7 +16,7 @@ ManaCost:no cost
 Colors:blue
 Types:Enchantment Aura
 K:Enchant creature
-A:SP$ Attach | ValidTgts$ Creature | TgtPrompt$ Select target creature | AILogic$ Pump | AITgts$ Creature.Legendary+YouCtrl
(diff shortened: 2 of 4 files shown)


Hanmac:
@Agetian imo this shouldn't be needed, right? normally the AI Logic `Curiosity` would already tell that it should prefer Creatures AI controls, or is there something funky?

Hanmac:
@Agetian found the culprit:

```
prefList = ComputerUtil.filterAITgts(sa, aiPlayer, prefList, true);

Card c = attachGeneralAI(aiPlayer, sa, prefList, mandatory, attachSource, sa.getParam("AILogic"));
```

`AITgts` is checked before `AILogic`

forge-gui/res/cardsfolder/m/mischievous_catgeist_catlike_curiosity.txt
```diff
@@ -16,7 +16,7 @@ ManaCost:no cost
 Colors:blue
 Types:Enchantment Aura
 K:Enchant creature
-A:SP$ Attach | ValidTgts$ Creature | TgtPrompt$ Select target creature | AILogic$ Pump
+A:SP$ Attach | ValidTgts$ Creature | TgtPrompt$ Select target creature | AILogic$ Pump | AITgts$ Creature.Legendary+YouCtrl
```
Hanmac, line 19: `AILogic$ Curiosity | AITgts$ Creature.YouCtrl`

Hanmac:
> > @Agetian found the culprit:
> > ```
> > prefList = ComputerUtil.filterAITgts(sa, aiPlayer, prefList, true);
> > 
> > Card c = attachGeneralAI(aiPlayer, sa, prefList, mandatory, attachSource, sa.getParam("AILogic"));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `AITgts` is checked before `AILogic`
> 
> Oh, a nice catch! Thanks! :) Can you take care of this (modify it so it works as needed), please? I'm currently overwhelmed with work, sadly... :/

i don't know what would the right answer ...

i think for now, it would be good enough to fix the AiTgts like this MR is doing

forge-gui/res/cardsfolder/m/mischievous_catgeist_catlike_curiosity.txt
Hanmac: change the AILogic too into Curiosity

forge-gui/res/cardsfolder/c/combat_research.txt
```diff
@@ -2,7 +2,7 @@ Name:Combat Research
 ManaCost:U
 Types:Enchantment Aura
 K:Enchant creature
-A:SP$ Attach | ValidTgts$ Creature | AILogic$ Curiosity | AITgts$ Creature.Legendary
+A:SP$ Attach | ValidTgts$ Creature | AILogic$ Curiosity | AITgts$ Creature.Legendary+YouCtrl
```
tool4ever, line 5: I don't see why these changes are needed
The AILogic already selects its own creatures

Klisz: The AITgts for Combat Research currently causes it to prefer a legendary creature an opponent controls over a nonlegendary creature the AI controls (I had a game where an effect suspended Combat Research and, once the last time counter was removed, the AI controlled only nonlegendary creatures and it targeted an opponent's legend). That being said, I have removed the added AITgts for the other three cards

tool4ever: ah right, thanks

tool4ever (approved):

PR #4376, opened 2023-12-14
CardDb: add normalized Names

### LINKED ISSUE #4365: Forge doesn't recognise cards that have the letters á, É, é, í, ó, ú, û in their names
**Describe the bug**
Forge doesn't recognise cards that have the letters á, É, é, í, ó, ú, û in their names. It writes warnings that these cards aren't supported.

**To Reproduce**
1. Choose "Constructed"
2. Choose "Net Archive Modern Decks"
3. Choose deck with Lórien Revealed
4. See warning

**Expected behavior**
Forge should recognise cards that have the letters á, É, é, í, ó, ú, û in their names

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.60-SNAPSHOT-12.11

**Screenshots**

lorien
lorien2

Hanmac:
closes #4365 

- [x] Added Normalized Names to DB
- [x] Deck Import recognize Normalized Names
- [x] Deck Import recognize Non Normalized Names
- [x] Deck Search Supports both Normalized and Non Normalized Names

The Deck import currently still can't understand the OG "û", i need to debug why


Hanmac:
@tehdiplomat what should i do with:
```
     public static final String REX_CARD_NAME = String.format("(\\[)?(?<%s>[a-zA-Z0-9&',\\.:!\\+\\\"\\/\\-\\s]+)(\\])?", REGRP_CARD);
```

?

Hanmac:
@tehdiplomat @kevlahnota the Pattern in DeckRecognizer should probably be updated for Unicode

forge-gui/res/cardsfolder/l/lim_dul_the_necromancer.txt
```diff
@@ -1,4 +1,4 @@
-Name:Lim-Dul the Necromancer
+Name:Lim-Dûl the Necromancer
```
tehdiplomat, line 1: So weird. All of the other Lim-Dul names had been normalized. Not sure how this one ended up with the accent marks.

Hanmac: This is on purpose

i want the Non Normalized Names be back on the Cards
while the Normalized Name is handled by the CardDb

tehdiplomat: No I know. I just meant why were Lim-Dul's Vault and other cards using the normalized version. I like just converting the standard name down to normalized for storage/lookups.

Hanmac: probably because our system couldn't handle the û vocals.

this MR is for changing that back

Hanmac:
The Font for Mobile doesn't need to be touched, because `û` is already in their Default Characters

Hanmac:
@tehdiplomat @kevlahnota i did some funky with the CardRulesPredicates

when searching Cards in the Deck Editor, it now searches for all the faces.

this helps when searching for cards like with Adventure

and for Card names, it even support the normalized names. so when you search for "Lim-Dul" you get "Lim-Dûl" too

forge-gui/tools/mtg-data.txt
```diff
@@ -52707,7 +52707,7 @@ White creatures you control get +0/+1.
 {R}{W}, Exile Legion's Initiative: Exile all creatures you control. At the beginning of the next combat, return those cards to the battlefield under their owner's control and those creatures gain haste until end of turn.
 DGM M
 
-Legions of Lim-Dul
+Legions of Lim-Dûl
```
tehdiplomat, line 52710: I don't think we use this file anymore. We should probably just delete it from the repo

Hanmac: yeah that should be deleted

and maybe the tools folder be cleaned too

forge-gui/res/draft/rankings.txt
```diff
@@ -38491,7 +38491,7 @@
 #103|False Summoning|C|PO2
 #104|Mind Rot|C|PO2
 #105|Dakmor Scorpion|C|PO2
-#106|Deja Vu|C|PO2
+#106|Déjà Vu|C|PO2
```
tehdiplomat, line 38494: I'm not sure this file is going to be imported this way and if its not, its going to be hard to maintain. Maybe we can keep this one with the normalized names?

tehdiplomat (approved):
As long as deck/quest loading can still handle normalized names I'm fine with the general idea of these changes. The next time they save they'll have the UTF names I guess?

forge-gui/res/draft/rankings.txt
Hanmac: i don't know where the ranking file is built from

i need to debug to see how the ranking file is parsed to see if it would support normalized names

tehdiplomat: I'm in the middle of splitting these up into individual files. I'd prefer if it wouldn't change right now.

forge-core/src/main/java/forge/card/CardDb.java
```diff
@@ -233,22 +235,34 @@ public CardDb(Map<String, CardRules> rules, CardEdition.Collection editions0, Li
...
         setCardArtPreference(cardArtPreference);
     }
 
+    private void addFaceToDbNames(ICardFace face) {
+        if (face == null) {
+            return;
+        }
+        final String name = face.getName();
+        facesByName.put(name, face);
+        final String normalName = Normalizer.normalize(name, Normalizer.Form.NFD).replaceAll("[^\\p{ASCII}]", "");
```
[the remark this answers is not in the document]
Hanmac: i just noticed that there is: `StringUtils.stripAccents`

tool4ever (approved):

PR #4385, opened 2023-12-15
ChangesZoneAll: add LKI support

### LINKED ISSUE #4335: The Skullspore Nexus trigger uses wrong power
**Describe the bug**
With the The Skullspore Nexus on the battlefield, I had a creature, with base p/t 4/4 with 3 +1/+1 counters and the power double effect of the nexus making it a 14/7, dying. 

I expected to get a 14/14 token, but got a 4/4.

It shoul use the last known information about the dying creature's power and not the one it would have in the graveyard.

Ruling:
> Use the power of the nontoken creatures that died as they last existed on the battlefield to determine the base power and toughness of the token created by the middle ability. [(2023-11-10)](https://scryfall.com/card/lci/212/the-skullspore-nexus#rulings)

This was on Android 1.6.59-SNAPSHOT-12-03

### LINKED ISSUE #4375: The War Doctor receives multiple triggers from ReplaceDying effects
**Describe the bug**
When a permanent dealt damage by an effect with ReplaceDying would die, and the resulting replacement effect exiles that creature, The War Doctor's ChangesZoneAll ability triggers twice rather than once. If several permanents are exiled simultaneously (such as via Anger of the Gods), there will be one correct "[Amount: X]" trigger (where X is the number of permanents exiled this way) and a separate "[Amount: 1]" for each permanent, as though the ability was triggering once as a ChangesZoneAll and then again as a ChangesZone.

**To Reproduce**
1. Have The War Doctor in play, along with a creature with toughness less than or equal to the number of time counters on The War Doctor
2. Attack with The War Doctor, targeting the other creature with the attack trigger
3. When the attack trigger resolves and the other creature is exiled, The War Doctor's ChangesZoneAll ability triggers twice rather than once

**Expected behavior**
The War Doctor should trigger only once when one or more permanents are exiled from the battlefield.

 - OS: Windows 10
 - Version 1.6.60-SNAPSHOT-12-13

**Additional context**
Neither Ranar the Ever-Watchful nor Hero of Bretagard (the only other two cards I'm aware of that have ChangesZoneAll triggers with Destination$ Exile and with battlefield as a possible origin zone) have this issue. After some testing, this appears to be linked with their ValidCause$ SpellAbility.YouCtrl clause (rather than the other difference, that their triggers both have battlefield as the only origin zone rather than any for The War Doctor).

### LINKED ISSUE #4386: ExiledWithSourceLKI still remembered long after the card leaves exile
**Describe the bug**
If Prowl, Stoic Strategist exiles a card, it continues to remember that card for as long as Prowl remains on the battlefield, even after that card leaves exile, and will trigger even if the card is cast later on from somewhere other than exile.

**To Reproduce**
1. Attack with Prowl, Stoic Strategist, exiling a tapped creature or Vehicle
2. Cast the exiled card
3. Convert Prowl, Pursuit Vehicle back into Prowl, Stoic Strategist
4. Return the formerly-exiled card to your hand
5. Cast that card from your hand
6. Prowl triggers again, even though this game object was not "exiled with Prowl" anymore

**Expected behavior**
Once an exiled card leaves exile, ExiledWithSourceLKI should match that card only long enough to trigger anything specifically triggering from that card leaving exile or being played; Prowl shouldn't trigger from later plays of the same card from other locations.

 - OS: Windows 10
 - Browser 1.6.60-SNAPSHOT-12-15

tool4ever:
Closes #4335
Closes #4386

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -963,7 +963,6 @@ public class GameAction {
             if (z.is(ZoneType.Battlefield)) {
                 c.runLeavesPlayCommands();
             }
-
         }
 
         // CR 603.6c other players LTB triggers should work
@@ -1041,7 +1040,6 @@ public class GameAction {
         game.getTriggerHandler().runTrigger(TriggerType.ChangesController, runParams, false);
 
         game.getTriggerHandler().clearSuppression(TriggerType.ChangesZone);
-
     }
 
     // Temporarily disable (if mode = true) actively checking static abilities.
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -916,7 +916,7 @@ public class GameAction {
     }
 
     public final CardCollection exile(final CardCollection cards, SpellAbility cause, Map<AbilityKey, Object> params) {
-        CardZoneTable table = new CardZoneTable();
+        CardZoneTable table = new CardZoneTable(game.getLastStateBattlefield(), game.getLastStateGraveyard());
         CardCollection result = new CardCollection();
         for (Card card : cards) {
             if (cause != null) {
@@ -1261,7 +1261,7 @@ public class GameAction {
             checkStaticAbilities(false, affectedCards, CardCollection.EMPTY);
             boolean checkAgain = false;
 
-            CardZoneTable table = new CardZoneTable();
+            CardZoneTable table = new CardZoneTable(game.getLastStateBattlefield(), game.getLastStateGraveyard());
 
             for (final Player p : game.getPlayers()) {
                 for (final ZoneType zt : ZoneType.values()) {
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -184,24 +184,8 @@ public class GameAction {
                 lastKnownInfo = (Card) cause.getReplacingObject(AbilityKey.CardLKI);
             }
         }
-        CardCollectionView lastBattlefield = null;
-        CardCollectionView lastGraveyard = null;
-        if (params != null) {
-            lastBattlefield = (CardCollectionView) params.get(AbilityKey.LastStateBattlefield);
-            lastGraveyard = (CardCollectionView) params.get(AbilityKey.LastStateGraveyard);
-        }
-        if (lastBattlefield == null && cause != null) {
-            lastBattlefield = cause.getLastStateBattlefield();
-        }
-        if (lastGraveyard == null && cause != null) {
-            lastGraveyard = cause.getLastStateGraveyard();
-        }
-        if (lastBattlefield == null) {
-            lastBattlefield = game.getLastStateBattlefield();
-        }
-        if (lastGraveyard == null) {
-            lastGraveyard = game.getLastStateGraveyard();
-        }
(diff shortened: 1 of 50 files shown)


forge-game/src/main/java/forge/game/ability/effects/DestroyAllEffect.java
```diff
@@ -89,10 +89,11 @@ public void resolve(SpellAbility sa) {
 
         list = GameActionUtil.orderCardsByTheirOwners(game, list, ZoneType.Graveyard, sa);
 
-        CardZoneTable table = new CardZoneTable();
         Map<AbilityKey, Object> params = AbilityKey.newMap();
         params.put(AbilityKey.LastStateBattlefield, game.copyLastStateBattlefield());
 
+        CardZoneTable table = new CardZoneTable(game.getLastStateBattlefield(), game.getLastStateGraveyard());
```
tool4ever, line 95: I need to update all other places
but in tests it already works with something like _Ruination_

forge-game/src/main/java/forge/game/trigger/TriggerChangesZone.java
```diff
@@ -102,16 +101,13 @@ public final boolean performTest(final Map<AbilityKey, Object> runParams) {
 
         if (hasParam("ValidCard")) {
             Card moved = (Card) runParams.get(AbilityKey.Card);
-            boolean leavesLKIZone = "Battlefield".equals(getParam("Origin"));
-            leavesLKIZone |= "Exile".equals(getParam("Origin")) && (moved.getZone().is(ZoneType.Graveyard) ||
-                    moved.getZone().is(ZoneType.Command) || hasParam("UseLKI"));
 
-            if (leavesLKIZone) {
+            if ("Battlefield".equals(getParam("Origin"))) {
```
Hanmac, line 105: removed from Exile also should look back aka LKI

tool4ever: hmn right, but only partially?
> 603.10a. Some zone-change triggers look back in time. These are leaves-the-battlefield abilities, abilities that trigger when a card leaves a graveyard, and **abilities that trigger when an object that all players can see is put into a hand or library**.

Hanmac: But as you said "when a card leaves a graveyard"

You did remove that

tool4ever: _Syr Konrad_ currently triggers if I return an Enchantment while _Opalescence_ is in play
I can probably fix that

tool4ever: I need to find a smart way though to differentiate if from something like _Breathless Knight_ which also has `Origin$ Graveyard` but it's still ETB trigger

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -916,7 +916,7 @@ public final Card moveToVariantDeck(Card c, ZoneType zone, int deckPosition, Spe
     }
 
     public final CardCollection exile(final CardCollection cards, SpellAbility cause, Map<AbilityKey, Object> params) {
-        CardZoneTable table = new CardZoneTable();
+        CardZoneTable table = new CardZoneTable(game.getLastStateBattlefield(), game.getLastStateGraveyard());
```
Hanmac, line 919: it seems to only affect the exile method, but shouldn't this also try to use the LastState from Cause if able?

or try to get them from params?

Hanmac (approved):

PR #4401, opened 2023-12-17
Adventure: check for decks in user plane directory too

Jorilx:
Hi everyone,
I've noticed that when using a user plane (from `USER_ADVENTURE_DIR/userplanes`) Forge looks for decks only under "common", so I've made this simple patch to make it look under the user plane directory too.


jjayers99 (changes requested):
It's a small observation, but considering how rarely user planes see use at this time, I would prefer to confirm that a custom plane is configured before making the file system call to check for deckfile.exists() and the same with any other references that check for custom / overriden data in this way.

Of more relevant note, this change should be expanded to encompass ".json" formatted files which are loaded a few lines below.

Jorilx:
Ok, I'll update the PR ASAP, thanks for your suggestion!

Jorilx:
I've studied the code better, [this line](https://github.com/Card-Forge/forge/blob/08b8d7deecc127053fef1e2006b1ff69ad3d43d0/forge-gui-mobile/src/forge/adventure/util/CardUtil.java#L680) 
`FileHandle handle = Config.instance().getFile(path);`
already handles the userplane-or-common cases, so my patch can be simplified a lot... I'll update the PR in a few minutes.

Jorilx:
I think the new patch looks nicer, but in the end the code still performs the file system checks... I'll try modifying `Config.getFile()`

jjayers99:
What you have here should be fine

jjayers99 (approved):

Agetian:
Please rebase against the current upstream master branch so that this PR can be merged.

PR #4408, opened 2023-12-19
Planechase inherent effects

### LINKED ISSUE #4262: Chaos and planeswalk abilities trigger several times if on multiple planes
**Describe the bug**
The two effects that result from nonblank planar die rolls (the planeswalking ability and chaos ensuing) are given to plane cards automatically by the CardFactory. This is in contradiction of rules 901.8 (which explicitly states that the planeswalking ability is "an inherent triggered ability" of Planechase games and "has no source")  and 901.9 (which lists them as occurring from certain die rolls, and does not list them as being abilities of planes themselves). The upshot of this is that, if multiple planes are face-up at the same time (due to Spatial Merging or Norn's Seedcore), rolling to planeswalk will cause N planeswalking triggers and rolling chaos will cause chaos to ensue N times, where N is the number of face-up planes at the time. For example, if Norn's Seedcore and Goldmeadow are both face-up, and chaos is rolled, chaos will ensue twice, and thus each plane's chaos ability will trigger twice rather than once, resulting in Norn's Seedcore adding two planes rather than one, and Goldmeadow creating two tokens rather than one. Similarly, when rolling planeswalk, two planeswalk triggers will be put on the stack; first you will planeswalk away from Norn's Seedcore and Goldmeadow, then immediately planeswalk again away from wherever you just arrived.

**To Reproduce**
1. Begin a planechase game
2. Planeswalk to Spatial Merging or Norn's Seedcore
3. Use a rigged planar die roll to have a non-blank result

**Expected behavior**
Each planeswalk result on the planar die should put only one planeswalking ability on the stack, and each chaos result should cause chaos to ensue only once, regardless of how many planes are face-up in the command zone at the time.

**Screenshots**
image

 - OS: Windows 10
 - Version: 1.6.59-SNAPSHOT-12-01

Klisz:
Closes #4262, by moving the planeswalking ability and the static chaos trigger to an effect carried with the player in a planechase game. Technically the special action to roll the planar die should be part of it too, but then you either have to make the "Planechase Effect" card visible in the UI so that human players can actually click it to roll, or rework how the UI relates to planechase entirely

### DIFF
--- a/forge-game/src/main/java/forge/game/player/Player.java
+++ b/forge-game/src/main/java/forge/game/player/Player.java
@@ -3228,12 +3228,6 @@ public class Player extends GameEntity implements Comparable<Player> {
         planesWalkTrigger.setOverridingAbility(AbilityFactory.getAbility(rolledWalk, eff));
         eff.addTrigger(planesWalkTrigger);
 
-        String chaosTrig = "Mode$ PlanarDice | ValidPlayer$ You | Result$ Chaos | TriggerZones$ Command | Static$ True";
-        String rolledChaos = "DB$ ChaosEnsues";
-        Trigger chaosTrigger = TriggerHandler.parseTrigger(chaosTrig, eff, true);
-        chaosTrigger.setOverridingAbility(AbilityFactory.getAbility(rolledChaos, eff));
-        eff.addTrigger(chaosTrigger);
-
         String specialA = "ST$ RollPlanarDice | Cost$ X | SorcerySpeed$ True | Activator$ Player | SpecialAction$ True" +
                 " | ActivationZone$ Command | SpellDescription$ Roll the planar dice. X is equal to the number of " +
                 "times you have previously taken this action this turn. | CostDesc$ {X}: ";
--- a/forge-game/src/main/java/forge/game/player/Player.java
+++ b/forge-game/src/main/java/forge/game/player/Player.java
@@ -3212,7 +3212,13 @@ public class Player extends GameEntity implements Comparable<Player> {
 
     public Card createPlanechaseEffects(Game game) {
         final String name = "Planechase Effect";
-        final Card eff = new Card(game.nextCardId(), game); //can't be DetachedCardEffect because there's no linked card
+        final Card eff = new Card(game.nextCardId(), game){
+            @Override
+            public Card getCardForUi() {
+                List<Card> currentPlanes = getOwner().getGame().getActivePlanes();
+                return (currentPlanes == null || currentPlanes.size() == 0) ? null : currentPlanes.get(0); //use current plane for the sake of UI display logic
+            }
+        };
         eff.setTimestamp(game.getNextTimestamp());
         eff.setName(name);
         eff.setOwner(this);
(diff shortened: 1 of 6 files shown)


Northmoc:
Nice solution. Definitely seems closer to the CR to me.

Seems like it would be easy enough to just put the new effect in the command zone so players could interact with it? Similar to things like the City's Blessing.

The card image should probably be some kind of image of a planar die or multiple planar dice
image

Klisz:
> Seems like it would be easy enough to just put the new effect in the command zone so players could interact with it? Similar to things like the City's Blessing.

Indeed, that's easier than what I've already done - the only issue I had was the lack of an image, so I altered it to have the effect's getCardForUi method overridden. I do like the idea of using an image of one or more planar dice, though, and alternatively the thought just occurred to me of possibly using [the cardback](https://backs.scryfall.io/large/7/8/7840c131-f96b-4700-9347-2215c43156e6.jpg?1665006192)used on physical planar cards. 

f we go either of these routes and move the planar dice special action from the planes to the effect, it should also be noted that RollPlanarDiceAi will need to be altered (since currently it uses the AIRollPlanarDiceParams SVar of the SpellAbility's host card - although, as it stands, none of the MOC or WHO planes have such an SVar anyway [I've been meaning to get around to making a MR to just add reasonable params to each of them, but as of yet the planar dice AI params are only found on the pre-2023 planes, meaning the AI will just never roll the planar die when on a 2023 plane]).

Klisz:
Moved the planar die roll to the effect, and made an attempt to update RollPlanarDiceAi, but there's an issue where the AI is still never activating the ability - I even tested changing RollPlanarDiceAi's canPlayAI to just straight-up `return true;` and it still never did, which leads me to believe that something is preventing it from even getting that far.

Klisz:
Never mind, that issue was just because I didn't realize that SpellAbility didn't have a getGame method of its own

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -3210,6 +3210,42 @@ else if (game.getRules().hasAppliedVariant(GameType.Oathbreaker)) {
...
+        String rolledWalk = "DB$ Planeswalk | Cause$ PlanarDie";
+        Trigger planesWalkTrigger = TriggerHandler.parseTrigger(trigger, eff, true);
+        planesWalkTrigger.setOverridingAbility(AbilityFactory.getAbility(rolledWalk, eff));
+        eff.addTrigger(planesWalkTrigger);
+
+        String chaosTrig = "Mode$ PlanarDice | ValidPlayer$ You | Result$ Chaos | TriggerZones$ Command | Static$ True";
+        String rolledChaos = "DB$ ChaosEnsues";
+        Trigger chaosTrigger = TriggerHandler.parseTrigger(chaosTrig, eff, true);
+        chaosTrigger.setOverridingAbility(AbilityFactory.getAbility(rolledChaos, eff));
+        eff.addTrigger(chaosTrigger);
```
Hanmac, line 3235: maybe remove this trigger and add just:
`getTriggerHandler().runTrigger(TriggerType.ChaosEnsues, runParams, false);`

in the PlanarDice class?

tool4ever (approved):
LGTM

Hanmac (approved):

PR #4409, opened 2023-12-19
Update Game.java

### LINKED ISSUE #3318: Plane card doesn't leave game when owner does
**Describe the bug**
In a Planechase game, when a player who isn't the planar controller but does own a face-up plane card leaves the game, that plane continues to exist. This causes additional issues; once a player planeswalks from such a plane, they'll planeswalk to a new plane, but the previous one will remain face-up in the command zone as well. Furthermore, the player who planeswalked will remain the planar controller of the old plane even after they pass turn, but the new plane will change planar controllers as normal.

**To Reproduce**
Begin a Planechase game with at least three players. Pass the turn such that the planar controller and the owner of the current plane are different players. Have the owner of the current plane leave the game.

**Expected behavior**
As per rule 901.10, the current plane should be removed from the game and the planar controller should turn the top card of their planar deck face up. (This already seems to behave correctly if the player who leaves the game is the planar controller as well as the owner.)

OS: Windows. Tested on build 1.6.5.7-SNAPSHOT-06.18 in both forge and forge-adventure.

Klisz:
Closes #3318

(Agetian noted in that issue that passing a null for the SpellAbility argument in Player.planeswalk would cause issues with Player.java:2701-2702, but Player.planeswalk has since been altered and no longer has those lines anyway, so I think that should be moot now?)

I've tested with all three combinations of
1. planar controller leaves the game while also the owner of an active plane (next player becomes planar controller, planeswalks)
2. planar controller leaves the game but doesn't own an active plane (next player takes control of active planes)
3. planar owner leaves the game when not the planar controller (planar controller planeswalks)

### DIFF
--- a/forge-game/src/main/java/forge/game/Game.java
+++ b/forge-game/src/main/java/forge/game/Game.java
@@ -891,7 +891,7 @@ public class Game {
                 }
             }
             runParams.put(AbilityKey.Cards, planesLeavingGame);
-            game.getTriggerHandler().runTrigger(TriggerType.PlaneswalkedFrom, runParams, false);
+            getTriggerHandler().runTrigger(TriggerType.PlaneswalkedFrom, runParams, false);
             planarController.planeswalkTo(null, new CardCollection(planarController.getZone(ZoneType.PlanarDeck).get(0)));
         }
 
--- a/forge-game/src/main/java/forge/game/Game.java
+++ b/forge-game/src/main/java/forge/game/Game.java
@@ -792,6 +792,7 @@ public class Game {
         // Rule 800.4 Losing a Multiplayer game
         CardCollectionView cards = this.getCardsInGame();
         boolean planarControllerLost = false;
+        boolean planarOwnerLost = false;
         boolean isMultiplayer = getPlayers().size() > 2;
         CardZoneTable triggerList = new CardZoneTable();
 
@@ -814,8 +815,13 @@ public class Game {
         }
 
         for (Card c : cards) {
-            if (c.getController().equals(p) && (c.isPlane() || c.isPhenomenon())) {
-                planarControllerLost = true;
+            if (c.isPlane() || c.isPhenomenon()) {
+                if (c.getController().equals(p)) {
+                    planarControllerLost = true;
+                }
+                if (c.getOwner().equals(p)) {
+                    planarOwnerLost = true;
+                }
             }
 
             if (isMultiplayer) {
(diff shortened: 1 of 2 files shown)


Agetian (approved):

forge-game/src/main/java/forge/game/Game.java
```diff
@@ -860,11 +866,23 @@ public void onPlayerLost(Player p) {
         // 901.6: If the current planar controller would leave the game, instead the next player
         // in turn order that wouldn't leave the game becomes the planar controller, then the old
         // planar controller leaves
+        if (planarControllerLost) {
+            for (Card c : getActivePlanes()) {
+                if ((c != null) && !(c.getOwner().equals(p))) {
+                    c.setController(getNextPlayerAfter(p), 0);
+                    getAction().controllerChangeZoneCorrection(c);
```
tool4ever, line 873: if you want try to also check for
> 901.10b. Abilities from phenomena owned by a player who left the game remain on the stack controlled by the new planar controller.
inside `removeInstancesControlledBy`?
Hopefully the engine can tolerate the card being gone

forge-game/src/main/java/forge/game/Game.java
```diff
@@ -860,11 +866,23 @@ public void onPlayerLost(Player p) {
...
         // from phenomena leave the game. (See rule 800.4a.) If that includes a face-up plane card
         // or phenomenon card, the planar controller turns the top card of his or her planar deck face up.the game.
-        if (planarControllerLost) {
-            getNextPlayerAfter(p).initPlane();
+        if (planarOwnerLost) {
+            Player planarController = getPhaseHandler().getPlayerTurn();
+            if (planarController.equals(p)) {
+                planarController = getNextPlayerAfter(p);
+            }
+            planarController.planeswalkTo(null, new CardCollection(planarController.getZone(ZoneType.PlanarDeck).get(0)));
```
tool4ever, line 885: from looking at the rules I'd say you should still call `planarController.leaveCurrentPlane()` before

Klisz: I've added logic to create a PlaneswalkedFrom trigger, though avoiding actually using leaveCurrentPlane for the purposes of an edge case with Norn's Seedcore (where there might be face-up planes owned by multiple players; in that case, from my understanding, only those actually owned by the player leaving the game should be planeswalked away from, while leaveCurrentPlane will put _all_ active planes on the bottom of their owners' planar decks)

tool4ever (dismissed):
nice work

tool4ever (approved):

PR #4412, opened 2023-12-19
Revert "Revert "CardDb: add normalized Names""

Hanmac:
Reverts Card-Forge/forge#4411


forge-core/src/main/java/forge/deck/DeckRecognizer.java
```diff
@@ -414,7 +414,7 @@ public static TokenKey fromString(String keyString){
     public static final String REGRP_CARD = "cardname";
     public static final String REGRP_CARDNO = "count";
 
-    public static final String REX_CARD_NAME = String.format("(\\[)?(?<%s>[a-zA-Z0-9&',\\.:!\\+\\\"\\/\\-\\s]+)(\\])?", REGRP_CARD);
+    public static final String REX_CARD_NAME = String.format("(\\[)?(?<%s>[a-zA-Z0-9âûáóéíúÉàäöüñ&',\\.:!\\+\\\"\\/\\-\\s]+)(\\])?", REGRP_CARD);
```
paulsnoops, line 417: Would something like À-ÿ work here to cover all bases?

Hanmac: i only added the ones i found, i don't know how the letter ranges works for such ones

i thought about using Unicode for the Regexp, but i thought that might be overkill

paulsnoops: As far as I know it should include everything we'd need while future proofing. I know it finds all cards on Scryfall so might be worth it if it works in Forge

https://scryfall.com/search?q=name%3A%2F%28%5B%C3%80-%C3%BF%5D%29%2F+-t%3Acard

Hanmac: i think the only ones missing are some un cards, but they should not matter

Hanmac:
@paulsnoops we probably need to talk about image keys / filenames

should we do `StringUtils.stripAccents` there too?

paulsnoops:
> should we do `StringUtils.stripAccents` there too?

I think as long as Forge will accept both versions of the name with and without accents etc. we should be good

Hanmac:
> > should we do `StringUtils.stripAccents` there too?
> 
> I think as long as Forge will accept both versions of the name with and without accents etc. we should be good

For the Card, yes.

But maybe not for the file names of the images.

That's a different logic I didn't touched yet with this
And might not be part of this MR

Hanmac:
@paulsnoops YOU need to check if the file names are still okay

otherwise we should do ANOTHER PR later

Hanmac:
@paulsnoops @tehdiplomat i need a confirmation on what to do with OLD cached files that has the wrong normalized name

otherwise this MR could be Merged

forge-core/src/main/java/forge/ImageKeys.java
```diff
@@ -367,26 +370,29 @@ public static File setLookUpFile(String filename, String fullborderFile) {
         return null;
     }
     private static File findFile(String dir, String filename) {
-        if (dir.equals(CACHE_CARD_PICS_DIR)) {
-            for (String ext : FILE_EXTENSIONS) {
-                if (ext.equals(""))
+        for (String ext : FILE_EXTENSIONS) {
```
tehdiplomat, line 373: I think I'd rather just leave this function alone and convert filename to normalized at the top, and don't try to rename. I'm not sure how writing file names in different character sets is going to react, and it feels like the type of thing thats just going to cause more problems than its solving.

Hanmac: i need to check this out, these characters should all be supported by the different file systems

if i would normalize the filename at the top, it would fail to find the already existing Filenames

@paulsnoops didn't give me the Info i would need how to work with old existing Filenames

tehdiplomat: Not sure why you think that's his responsibility. He was just pointing out sobering he thought of. He's not really the expert of that area

Hanmac: He said the old cache names will be a problem

Otherwise I would revert the changes to this function, and have it be done at later MR like I said multiple times

Hanmac:
@tehdiplomat This MR is for fixing the Card Names

fixing possible image keys, needs to be a different MR

tehdiplomat (approved):

PR #4413, opened 2023-12-20
Split rankings

tehdiplomat:


Northmoc:
Nice work!!

Hanmac (approved):

Hanmac:
depending the load, we might add "lazy" loading for the Rankings?

i mean we only need to load the Ranking for a specific edition once at a time, right?

tehdiplomat:
No I think we use relative ranking in a set to determine if you are drafting all cards.

Hanmac:
> No I think we use relative ranking in a set to determine if you are drafting all cards.

no i mean, when drafting, you only draft from one set, right? so only the info from that set should be needed?

tehdiplomat:
I understand what you mean. We have a mode where you can just draft ALL cards ever, and I think for that mode we take each cards individual spot in an edition to cross-rank it against other cards. Meaning if you draft in that mode you need all rankings loaded in.

PR #4415, opened 2023-12-20
Overload keyword

Hanmac:
finally Overload as Keyword

might need better Stack Description

Some other stuff like `CounterEffect#AllValid` could be refactored/removed too

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/PumpAllEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/PumpAllEffect.java
@@ -78,7 +78,7 @@ public class PumpAllEffect extends SpellAbilityEffect {
                 tgtC.addHiddenExtrinsicKeywords(timestamp, 0, hiddenkws);
             }
 
-            if (sa.hasParam("RememberAllPumped")) {
+            if (sa.hasParam("RememberPumped")) {
                 sa.getHostCard().addRemembered(tgtC);
             }
 
--- a/forge-game/src/main/java/forge/game/keyword/Keyword.java
+++ b/forge-game/src/main/java/forge/game/keyword/Keyword.java
@@ -134,7 +134,7 @@ public enum Keyword {
     NINJUTSU("Ninjutsu", Ninjutsu.class, false, "%s, Return an unblocked attacker you control to hand: Put this card onto the battlefield from your %s tapped and attacking."),
     OUTLAST("Outlast", KeywordWithCost.class, false, "%s, {T}: Put a +1/+1 counter on this creature. Outlast only as a sorcery."),
     OFFERING("Offering", KeywordWithType.class, false, "You may cast this card any time you could cast an instant by sacrificing a %1$s and paying the difference in mana costs between this and the sacrificed %1$s. Mana cost includes color."),
-    OVERLOAD("Overload", KeywordWithCost.class, false, "You may cast this spell for its overload cost. If you do, change its text by replacing all instances of \"target\" with \"each."),
+    OVERLOAD("Overload", KeywordWithCost.class, false, "You may cast this spell for its overload cost. If you do, change its text by replacing all instances of \"target\" with \"each.\""),
     PARTNER("Partner", Partner.class, true, "You can have two commanders if both have partner."),
     PERSIST("Persist", SimpleKeyword.class, false, "When this creature dies, if it had no -1/-1 counters on it, return it to the battlefield under its owner's control with a -1/-1 counter on it."),
     PHASING("Phasing", SimpleKeyword.class, true, "This phases in or out before you untap during each of your untap steps. While it's phased out, it's treated as though it doesn't exist."),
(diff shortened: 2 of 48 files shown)


forge-game/src/main/java/forge/game/keyword/Keyword.java
```diff
@@ -134,6 +134,7 @@ public enum Keyword {
     NINJUTSU("Ninjutsu", Ninjutsu.class, false, "%s, Return an unblocked attacker you control to hand: Put this card onto the battlefield from your %s tapped and attacking."),
     OUTLAST("Outlast", KeywordWithCost.class, false, "%s, {T}: Put a +1/+1 counter on this creature. Outlast only as a sorcery."),
     OFFERING("Offering", KeywordWithType.class, false, "You may cast this card any time you could cast an instant by sacrificing a %1$s and paying the difference in mana costs between this and the sacrificed %1$s. Mana cost includes color."),
+    OVERLOAD("Overload", KeywordWithCost.class, false, "You may cast this spell for its overload cost. If you do, change its text by replacing all instances of \"target\" with \"each."),
```
Northmoc, line 137: Doesn't the final `"` also need `\` ?

Northmoc:
Nice work!

I have an idea for StackDescription if you have other fish to fry

forge-game/src/main/java/forge/game/ability/effects/PumpAllEffect.java
```diff
@@ -78,7 +78,7 @@ private static void applyPumpAll(final SpellAbility sa,
                 tgtC.addHiddenExtrinsicKeywords(timestamp, 0, hiddenkws);
             }
 
-            if (sa.hasParam("RememberAllPumped")) {
+            if (sa.hasParam("RememberPumped")) {
```
Northmoc, line 81: Good call

tool4ever:
@Hanmac 
I tried to copy it and it had no effect, the original still worked after:
image

Hanmac:
> @Hanmac I tried to copy the overloaded Spell and it had no effect, the original still worked after: image

what exactly did you do? give me something to replicate it?
imo it still should not be able to gain another instance of the keyword?

tool4ever:
I just cast it normally from hand and followed it with _Reiterate_ 🤔

tool4ever:
The problem is somehow` CardTraitBase.copyHelper()` + `originalMapParams`?

Hanmac:
@tool4ever fixed the Spell Copy

tool4ever:
Should be good aside from the StackDescription thing

Northmoc:
I added my idea, see what you think

image

forge-gui/res/cardsfolder/c/counterflux.txt
```diff
@@ -2,6 +2,6 @@ Name:Counterflux
 ManaCost:U U R
 Types:Instant
 K:This spell can't be countered.
-A:SP$ Counter | Cost$ U U R | TargetType$ Spell | TgtPrompt$ Select target spell you don't control. | ValidTgts$ Card.YouDontCtrl | SpellDescription$ Counter target spell you don't control.
-A:SP$ Counter | Cost$ 1 U U R | AllValid$ Spell.YouDontCtrl | PrecostDesc$ Overload | CostDesc$ {1}{U}{U}{R} | NonBasicSpell$ True | SpellDescription$ (You may cast this spell for its overload cost. If you do, change its text by replacing all instances of "target" with "each.") | StackDescription$ Counter each spell you don't control.
+A:SP$ Counter | TargetType$ Spell | TgtPrompt$ Select target spell you don't control. | ValidTgts$ Spell.YouDontCtrl | SpellDescription$ Counter target spell you don't control.
```
Northmoc, line 5: Is this ValidTgts$ OK now? I thought it still needed to be Card?

Hanmac: I need to test it again, but it should still work

Hanmac: i made all Cards on the Stack be Spells

i wanted this for another MR later too

Hanmac: reverted because it messes with other stuff

Hanmac:
sometimes i get this:
> ComputerUtilMana: payManaCost() cost was not paid for Overload {1}{U}{U}{R} (You may cast this spell for its overload cost. If you do, change its text by replacing all instances of "target" with "each.") (Counterflux). Didn't find what to pay for null
> AI failed to play Counterflux (124)

i need to debug why this happened

@tool4ever @Northmoc can you try to debug this?

tool4ever:
`CounterAi` shouldn't be able to play non-targeted ones by itself currently
So I don't know how you managed to get past that (the error is only shown after it has passed testing) 😕

Hanmac:
> `CounterAi` shouldn't be able to play non-targeted ones by itself currently So I don't know how you managed to get past that (the error is only shown after it has passed testing) 😕

Surprisely the Error happens when I try to cast it with Overload cost and then the card disappears

But not always, for some reason

tool4ever:
Were you using Dev Mode maybe?
I can only get that to happen if I make AI cast something, it looks like it might mess up priority a bit 🤷‍♂️ 
In that case it should be unrelated to this PR

Hanmac:
> Were you using Dev Mode maybe? I can only get that to happen if I make AI cast something, it looks like it might mess up priority a bit 🤷‍♂️ In that case it should be unrelated to this PR

i thought it was my attempt to cast it what caused it to crash in AI ...

but yeah, that might be look into later

tool4ever (approved):

PR #4417, opened 2023-12-20
Desertion fix

tool4ever:
> The card is put onto the battlefield, but any effects that check if the original card was “cast from your hand” will not trigger or otherwise consider the card to have been cast from your hand. The card was put onto the battlefield by the effect of Desertion instead.


Northmoc (approved):
👍  Looks good to me,

PR #4424, opened 2023-12-21
 Fix Stonebinder's Familiar not triggering from Nivmagus Elemental 

tool4ever:


forge-gui/res/cardsfolder/s/stonebinders_familiar.txt
```diff
@@ -2,7 +2,7 @@ Name:Stonebinder's Familiar
 ManaCost:W
 Types:Creature Spirit Dog
 PT:1/1
-T:Mode$ ChangesZoneAll | Destination$ Exile | TriggerZones$ Battlefield | Execute$ TrigPutcounter | PlayerTurn$ True | ActivationLimit$ 1 | TriggerDescription$ Whenever one or more cards are put into exile during your turn, put a +1/+1 counter on CARDNAME. This ability triggers only once each turn.
+T:Mode$ ChangesZoneAll | ValidCard$ Card.nonToken+nonCopiedSpell | Destination$ Exile | TriggerZones$ Battlefield | Execute$ TrigPutcounter | PlayerTurn$ True | ActivationLimit$ 1 | TriggerDescription$ Whenever one or more cards are put into exile during your turn, put a +1/+1 counter on CARDNAME. This ability triggers only once each turn.
```
Northmoc, line 5: hmm, `nonToken+nonCopiedSpell` is a lot of stuff that isn't in the Oracle.

My concern is that for future similar cards it might be too easy to forget to add these things.

Would it be better long term to make `nonToken+nonCopiedSpell` default for ChangesZoneAll ?

tool4ever: it's kind of in the Oracle ("cards")
also we can't make it default because some others like Ranar don't care:
> spell or ability you control exiles one or more permanents from the battlefield

Hanmac (approved):

Hanmac:
@tool4ever you should stop adding fixes to unrelated MRs

that makes it more difficult to review them

PR #4425, opened 2023-12-22
Update kate_stewart.txt

Klisz:
Fixes issue where Kate Stewart's first ability would trigger upon suspending a permanent card

### DIFF
--- a/forge-gui/res/cardsfolder/k/kate_stewart.txt
+++ b/forge-gui/res/cardsfolder/k/kate_stewart.txt
@@ -2,7 +2,7 @@ Name:Kate Stewart
 ManaCost:1 U R W
 Types:Legendary Creature Human Scientist
 PT:3/3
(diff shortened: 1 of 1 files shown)


forge-gui/res/cardsfolder/k/kate_stewart.txt
```diff
@@ -2,7 +2,7 @@ Name:Kate Stewart
 ManaCost:1 U R W
 Types:Legendary Creature Human Scientist
 PT:3/3
-T:Mode$ CounterAddedOnce | ValidCard$ Permanent.YouCtrl | TriggerZones$ Battlefield | CounterType$ TIME | Execute$ TrigToken | TriggerDescription$ Whenever you put one or more time counters on a permanent you control, create a 1/1 white Soldier creature token.
+T:Mode$ CounterAddedOnce | ValidCard$ Permanent.YouCtrl+inZoneBattlefield | TriggerZones$ Battlefield | CounterType$ TIME | Execute$ TrigToken | TriggerDescription$ Whenever you put one or more time counters on a permanent you control, create a 1/1 white Soldier creature token.
```
tool4ever, line 5: it's important you use inRealZone because otherwise there is some LKI logic so it would still trigger for something like _Epochrasite_

tool4ever (approved):

PR #4429, opened 2023-12-23
Updates "Can't be countered" into Replacement Effects

### LINKED ISSUE #2688: CantCounter Rework
* `CardFactoryUtil.isCounterable` should probably be removed in favor of isCounterableBy
* `CardFactoryUtil.isCounterableBy` should either be moved and/or extended
if it only checking the ReplacementList, moved to AI?
* Remove this keywords:
  * "CARDNAME can't be countered."
  * "This spell can't be countered."
  * "CantBeCounteredBy"
* Currently it is done as CantHappen Replacement Effect, is this the right way or are our StaticAbility better?
Should we prefer CantHappen if an event is available?
  * For "CantBeCounteredBy", the Replacement Effect should probably get new Parameter to handle
* CounterEffect should be changed to use CantHappen result

Hanmac:
Closes #2688

This Removes the Hidden Keywords:
* `CARDNAME can't be countered.`
* `This spell can't be countered.`
* `CantBeCounteredBy`

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -2076,12 +2076,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
         final StringBuilder sbLong = new StringBuilder();
 
         for (String keyword : getHiddenExtrinsicKeywords()) {
-            if (keyword.startsWith("CantBeCounteredBy")) {
-                final String[] p = keyword.split(":");
-                sbLong.append(p[2]).append("\r\n");
-            } else {
-                sbLong.append(keyword).append("\r\n");
-            }
+            sbLong.append(keyword).append("\r\n");
         }
         if (sb.length() > 0) {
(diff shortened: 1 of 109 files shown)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -5234,7 +5227,7 @@ public final boolean isPermanent() {
     }
 
     public final boolean isSpell() {
-        return isInstant() || isSorcery() || (isAura() && !isInZone((ZoneType.Battlefield)));
+        return isInZone(ZoneType.Stack) || isInstant() || isSorcery() || (isAura() && !isInZone(ZoneType.Battlefield));
```
tool4ever, line 5230: The problem is the ability of _Lightning Storm_
when _Chimil_ is in play you can now no longer _Stifle_ it 💩

Hanmac: reworked again

tool4ever (approved):

PR #202, opened 2022-05-02
GameAction: Add ETB counters before checking statics

tool4ever:
grafik

Is this fix safe?

Tested with _Tayam, Luminous Enigma_ and it also still got vigilance counter

PR #4436, opened 2023-12-27
Saga and Read ahead refactor for WHO

### LINKED ISSUE #3868: WHO Barbara Wright
`Sagas you control have read ahead`

for this to work we need to redo saga and the Read Ahead Keyword, so that it's possible to give a card "Read Ahead" keyword without it breaking how the Saga worked

we might need to remove on how the Chapter Abilities are built, while not being connected to a Read Ahead keyword, i still would like them to be better generated somehow, especially if multiple chapter share the same ability

### LINKED ISSUE #3957: "Blink" saga not working properly; triggering first effect as Chapter II
**Describe the bug**
The Blink saga triggers its first effect as Chapter II, shuffling a creature into the library and creating a clue token under control of the same player. The effect triggers three times in a row in total. The effect that creates the Alien Angel token only occurs once, as Chapter IV.

**To Reproduce**
1. Play "Blink".
2. Allow chapters I through IV to trigger.

**Expected behavior**
The effects should alternate. I and III shuffle a creature into the library and create a clue. II and IV create an Alien Angel token.

- OS: macOS Monterey Version 12.5.1
- Browser: Chrome
- Version: 1.6.59-SNAPSHOT-10.19

**Additional context**
Perhaps the way sagas are currently coded is not equipped to handle repeated Chapters that are "out of order". I checked the script for the card, and it seems like it should be working properly, yet it does not.

Hanmac:
Closes #3868 
Closes #3957

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -363,28 +363,26 @@ public class AiController {
             }
         }
 
-        for (final Trigger tr : card.getTriggers()) {
-            if (!card.hasStartOfKeyword("Saga") && !card.hasStartOfKeyword("Read ahead")) {
+        if (card.isSaga()) {
+            for (final Trigger tr : card.getTriggers()) {
+                if (tr.getMode() != TriggerType.CounterAdded || !tr.isChapter()) {
+                    continue;
+                }
+
+                SpellAbility exSA = tr.ensureAbility().copy(activator);
+
+                if (api != null && exSA.getApi() == api) {
+                    rightapi = true;
+                }
+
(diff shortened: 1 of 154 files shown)


tool4ever:
@Hanmac 
_Barbara Wright_ + _Doubling Season_ + normal Saga ETB
previous chapters will also trigger

the problem is in GameAction `replaceCounterEffect()` is done before `checkStaticAbilities()`
which we switched as fix for #202

I wonder if we just need to calculate statics two times now, before and after replacing counters 🤔

Hanmac:
@tool4ever fixed with optional double check

tool4ever (approved):
I could not find other bugs

PR #4441, opened 2023-12-28
WHO: add can't tap check

### LINKED ISSUE #4440: Ood Sphere

Hanmac:
Closes #4440

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -51,6 +51,7 @@ import forge.game.player.PlayerCollection;
 import forge.game.replacement.ReplaceMoved;
 import forge.game.replacement.ReplacementEffect;
 import forge.game.replacement.ReplacementHandler;
+import forge.game.replacement.ReplacementLayer;
 import forge.game.replacement.ReplacementResult;
 import forge.game.replacement.ReplacementType;
 import forge.game.spellability.*;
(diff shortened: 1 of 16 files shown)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4505,14 +4506,38 @@ public final void setTapped(boolean tapped0) {
...
 
         // Run replacement effects
-        getGame().getReplacementHandler().run(ReplacementType.Tap, AbilityKey.mapFromAffected(this));
+        Map<AbilityKey, Object> repParams = AbilityKey.mapFromAffected(this);
+        repParams.put(AbilityKey.IsCombat, attacker); // right name for parameter?
+
+        switch (getGame().getReplacementHandler().run(ReplacementType.Tap, repParams)) {
+        case NotReplaced:
+            break;
+        case Updated:
```
tool4ever, line 4536: it was previously used by _Illusionary Mask_ so you probably need to that change that into Updated result to work correctly

Hanmac: That's why Updated is handled the same way as not replaced

tool4ever: you need to also update it here because it is not the default result:
https://github.com/Card-Forge/forge/blob/be2f9aa187a3b6d06f1608a66085c3173d9b0560/forge-game/src/main/java/forge/game/ability/effects/PlayEffect.java#L564

tool4ever (approved):

PR #1076, opened 2022-07-08
Make charm choices of triggers for each while it gets added to stack, not before

tool4ever:
Closes #1051

PR #4448, opened 2023-12-30
Riot: refactor into UnlessCost like Fabricate

Hanmac:
This refactors Riot into an UnlessCost similar to Fabricate

because the rules say, if you can't put counters on the creature, then it needs to gain haste

See #1076

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
@@ -1488,7 +1488,7 @@ public class ComputerUtilCombat {
                     continue;
                 }
                 if (sa.hasParam("Cost")) {
-                    if (!CostPayment.canPayAdditionalCosts(sa.getPayCosts(), sa, false)) {
+                    if (!CostPayment.canPayAdditionalCosts(sa.getPayCosts(), sa, true)) {
                         continue;
                     }
                 }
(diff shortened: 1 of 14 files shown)


forge-game/src/main/java/forge/game/cost/CostPutCounter.java
```diff
@@ -208,4 +241,27 @@ public void resetLists() {
         counterTable.clear();
     }
 
+    protected boolean isETBRelacement(final SpellAbility ability, final boolean effect) {
```
tool4ever, line 244: typo in name

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
```diff
@@ -1488,7 +1488,7 @@ public static int predictToughnessBonusOfAttacker(final Card attacker, final Car
                     continue;
                 }
                 if (sa.hasParam("Cost")) {
-                    if (!CostPayment.canPayAdditionalCosts(sa.getPayCosts(), sa)) {
+                    if (!CostPayment.canPayAdditionalCosts(sa.getPayCosts(), sa, false)) {
```
tool4ever, line 1491: this should be true too?

Hanmac: fixed

tool4ever (approved):

Agetian (approved):

PR #4449, opened 2023-12-30
GameAction: stop always trigger before etb counter

Hanmac:
fixes Dark Depths Always trigger

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -583,10 +583,15 @@ public class GameAction {
             }
         }
 
-        // Need to apply any static effects to produce correct triggers
-        checkStaticAbilities();
+        if (!table.isEmpty()) {
+            // we don't want always trigger before counters are placed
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -583,10 +583,15 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
             }
         }
 
-        // Need to apply any static effects to produce correct triggers
-        checkStaticAbilities();
+        if (!table.isEmpty()) {
+            // we don't want always trigger before counters are placed
+            game.getTriggerHandler().suppressMode(TriggerType.Always);
```
tool4ever, line 588: When you cast _Endless One_ with X=1 while your opponent controls _Vorinclex, Monstrous Raider_:
the trigger would stay supressed

Hanmac: fixed

tool4ever (approved):

PR #4483, opened 2024-01-06
Add card script for Mistakes Were Made (UNF)

Wild-W:
Card script for Mistakes Were Made (from Unfinity) The second part of its ability does nothing, but that might be fine. (See Ass Whuppin)

Oracle: "Destroy target artifact or enchantment. Create a 1/1 green Squirrel creature token for each fire extinguisher you can see from your seat."
Ass Whuppin's text reads: "Destroy target silver-bordered permanent in any game you can see from your seat."

Pretty similar scenario, and Ass Whuppin is already implemented so maybe this could be too.


paulsnoops:
Ass Whuppin' works because the only silver-bordered cards you can see are the ones in the game you're playing. This card doesn't work because we don't have fire extinguishers inside Forge.

The main problem here is "The second part of its ability does nothing, but that might be fine."
We can't really intentionally implement cards that can't do what they say (bugs aside) and the engine won't support.

PR #4484, opened 2024-01-06
Improve TriggerHandler accuracy

### LINKED ISSUE #1157: Differ Between Control Untap Haste Effects
Most Effects does "Gain Control, do Untap, Gain Haste"

very few like "Blind with Anger" does Untap before Gain Control.

This would matter under which control the creature does untap for possible trigger.

### LINKED ISSUE #4471: Triggers should only be checked once (preferably very close after the event)
When you target yourself with _Goblin Festival_ and control _Auntie Blyte_ you only get her trigger if you win the flip (but obviously it should happen in both cases).

It fails because even though the trigger is collected correctly after the damage is dealt (while you still control the Enchantment), the trigger gets tested again when the stack unfreezes for it.

But why? We have **already confirmed the event was correct** at the point between effect X and Y and that should be the final result.

So I think checking `canRunTrigger` twice is not only redundant but even wrong rules wise.

Luckily my proposed fix is minimal by changing the logic here
https://github.com/Card-Forge/forge/blob/13cf0810460741f36b9f25f4e31ad8db5f9e259d/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java#L360
into `(wt.getTriggers() != null || canRunTrigger(t, mode, runParams)`

@Hanmac 
your opinion?

(It is a bit like #1157 but not the same!)

tool4ever:
Casting _Blind with Anger_ on another players _Sphinx's Disciple_ now correctly lets the old controller draw.

Closes #1157
Closes #4471


forge-game/src/main/java/forge/game/trigger/TriggerWaiting.java
```diff
@@ -13,6 +16,7 @@ public class TriggerWaiting {
     private TriggerType mode;
     private Map<AbilityKey, Object> params;
     private List<Trigger> triggers = null;
+    private Map<Trigger, Player> controllers = null;
```
Hanmac, line 19: i don't know if we really need two extra lists, or if they can be combined into one

i mean the trigger list can also be just returned by map.keySet ?

Hanmac:
for the untap/GainControl vs GainControl/Untap i will think about it if we can't still do this as one effect or not

forge-game/src/main/java/forge/game/trigger/TriggerWaiting.java
tool4ever: Done
I think since TriggerHandler supports duplicates (maybe some trigger with LKI) we need MultiMap here too though

Hanmac: For the thing with LKI trigger, I would need to debug what the right answer should be

Hanmac: i think for LKI trigger, only one of them should match? The one with TriggerZone Battlefield?

of the LKI ones should overwrite the other ones

Hanmac (approved):

PR #4491, opened 2024-01-08
Simplify how users can open the log file when we need it

tehdiplomat:
image


forge-gui/res/languages/en-US.properties
```diff
@@ -342,6 +342,7 @@ lblTroubleshooting=Troubleshooting
 lblArticles=Articles
 lblGettingStarted=Getting Started
 lblHowtoPlay=How to Play
+lblOpenLogFile=Open Log File
```
Northmoc, line 345: `lblOpenLogFile` should probably be added to other languages too

PR #4493, opened 2024-01-08
Skip ChangesZoneAll inside Moved replacement

### LINKED ISSUE #1733: Moved replacement causing wrong ChangesZoneAll trigger
image

When you destroy the Griffin _Welcoming Vampire_ will trigger.
This is because the RiP RE only changes the zone of the copied card, so the `ChangeZoneTable` will be `{Battlefield={Battlefield=[Abbey Griffin]}}`

@Hanmac 
my current idea is to just return `game.getCardState(c)` here:
https://github.com/Card-Forge/forge/blob/e7f6b3371a2642ce625abce01c8b2bab21c2c368/forge-game/src/main/java/forge/game/GameAction.java#L382

or at the very least update the zone of c with the new one?

Edit:
probably not good enough, that would still cause `triggerChangesZoneAll` twice, once from RE ChangeZoneEffect and second from DestroyEffect

### LINKED ISSUE #4458: War Doctor triggering twice on exiling creature
**Describe the bug**
I attacked with The War Doctor and killed Merfolk Assassin with his ability, but instead of a single trigger, he triggered twice as you can see in the log.

I *think* this (triggering twice on exile) was one of the issues mentioned in something that's being worked on, but just in case it isn't I wanted to let you guys know about it.

**Screenshots**
Screenshot_20231231_191907 cleaned

tool4ever:
Closes #1733
Closes #4458


forge-game/src/main/java/forge/game/card/CardZoneTable.java
```diff
@@ -81,7 +86,8 @@ protected Table<ZoneType, ZoneType, CardCollection> delegate() {
 
     public void triggerChangesZoneAll(final Game game, final SpellAbility cause) {
         triggerTokenCreatedOnce(game);
-        if (!isEmpty()) {
+        if (!isEmpty() &&
```
Hanmac, line 89: I would probably rewrite it to pass the ChangesZone Table as Replacement Parameter

With this, the RE doesn't need to create their own Table,  but can fill the existing one

tool4ever: the problem I see there:
1. if the inner RE effect reuses the "parent" table we'd need a different sort of check in all outer places so it doesn't create a second but wrong entry?
2. and the other way around (maybe rarer) - if the RE gets replaced by another RE we'd also need to make sure only the zone changer at the bottom of that chain actually ends up in the table?

Hanmac: you pass the table via params to the deepest GameAction moveZone part

then only the final move should be put into the table

tool4ever: - `CardZoneTable.put()` will still get called by the outer "cause" after the RE is done
- so it should bail if entry origin/*/card already exists for that to work?
- then we need to be sure no weird case exists where a Table could **legally** have two such entries like some simultaneous stuff - Imo it's probably safe but what you think? 🤔 

I'd say the current solution is a bit cleaner code wise but `getCardState()` isn't that great so passing the table might be worth it.

Hanmac: > * `CardZoneTable.put()` will still get called by the outer "cause" after the RE is done
and this is what i mean should be moved into GameAction

Currently:

Effect: 
* create Table
* getOldZone
* move Card
* compare OldZone with NewZone
* add to Table
* Trigger Table

what i want:

Effect:
* create Table (if Effect is not RE, we have similar checks for Cause)
* move Card (using Table as mapParam)
* Trigger Table (if Effect is not RE)

Game Action:
* calls ReplacementEffect
* if nothing replaces it, add Card to ZoneTable via the Param **This is the Part i want to add**

tool4ever: Ok, I think I follow
will try later

tool4ever: I updated ChangeZoneEffect as PoC

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -538,8 +538,10 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
                     if (repres != ReplacementResult.NotReplaced) continue;
                 }
                 if (card == c) {
+                    storeChangesZoneAll(copied, zoneFrom, zoneTo, params);
                     zoneTo.add(copied, position, toBattlefield ? null : lastKnownInfo); // the modified state of the card is also reported here (e.g. for Morbid + Awaken)
                 } else {
+                    storeChangesZoneAll(card, zoneFrom, zoneTo, params);
```
tool4ever, line 544: The mutate handling isn't perfect - you added the components here: a653085

most likely so it works with returning all parts via UntilHostLeavesPlay?

but I think _Hero of Bretagard_ should probably only gain 1 counter when a mutated permanent gets exiled

So it might better be done inside TriggerHandler like for ChangesZone here:
https://github.com/Card-Forge/forge/blob/acb5151a8289d788afaa9653e79cf507a85ddc0e/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java#L467

but maybe something for later since it would require even more refactor

Hanmac: "or a spell or ability you control exiles one or more permanents from the battlefield" imo that need to use LKI ?

meaning it should only see one Mutated Card?

Hero of Bretagard probably need to be refactored into an Exile Trigger anyway instead of a ChangesZones Trigger

tool4ever: Yea the wording is sus, we'd need new ExiledAll but for some reason it doesn't have a ruling to ignore RE like _Soulherder_

Hanmac: yeah, but i think it is still implied

also for touching Exile, i need to add a Static/Replacment for Exile GameAction to add:
"Cause can't cause you to exile stuff" for the WHO card

that is on my TODO list

tool4ever:
I tried to update all important places

forge-game/src/main/java/forge/game/ability/AbilityKey.java
```diff
@@ -30,6 +30,7 @@ public enum AbilityKey {
     Card("Card"),
     Cards("Cards"),
     CardsFiltered("CardsFiltered"),
+    CardZoneTable("CardZoneTable"),
```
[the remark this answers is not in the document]
Hanmac: @tehdiplomat your opinion about the CardZoneTable param ?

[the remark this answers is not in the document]
tehdiplomat: I honestly don't know what "CardZoneTable" means, so I would say its probably not a good name.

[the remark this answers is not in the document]
Hanmac: the CardZoneTable there is for collecting the Card Changes Zones over multiple replacement effects

which then can trigger the *All Trigger with the right values

i don't know if there is a better name for the Parameter

Hanmac (approved):

forge-game/src/main/java/forge/game/GameAction.java
Hanmac: Hero of Bretagard  should be updated in a later MR too

PR #4502, opened 2024-01-10
Fix UntilHostLeavesPlay being dropped when replaced

### LINKED ISSUE #4452: Cards exiled with Lagrella, the Magpie don't return
**Describe the bug**
As the title says, the cards that Lagrella exiles aren't returning when she leaves the battlefield.

**Expected behavior**
Any card that is exiled with Lagrella to return once she leaves the field, no matter how she leaves the field.

tool4ever:
I'm linking the list to the new Card directly so there's no chance of a GameCommand running twice (though I'm not certain that's a real risk).

Closes #4452


tool4ever:
@Hanmac 
I think this is ready?

Hanmac (approved):
i think this is okay for now

PR #4509, opened 2024-01-11
Update snapshots-android.yml

tehdiplomat:


Hanmac:
@tehdiplomat should we try to use this for setting up the Android tools?

https://github.com/android-actions/setup-android

tehdiplomat:
@Hanmac I tried that one. It didn't work. The way I'm doing it seems to work just fine.

PR #4518, opened 2024-01-14
Support duplicate objects per effect

### LINKED ISSUE #4517: Support effects affecting the same object more than once
We know in general effects on 2+ objects should happen simultaneously but there are corner cases where those objects can be identical.

I did not check all cards for which effects might need it yet:
- LifeGain on _Angel of Destiny_
``Whenever a creature you control deals combat damage to a player, you and that player each gain that much life.``
> If a creature you control somehow deals combat damage to you (most likely because its combat damage was redirected), you gain twice that much life as a single life-gaining event.
- Draw on _Diviner Spirit_
- Counters on _Serrated Biskelion_ (this one already got reworked into a single effect by fddff49, so now if it targets itself only one Counter gets applied 🐞 )

Some considerations:
1. rewriting into a single effect should probably still be the desired solution (compared to storing all tables between etc.)
2. this means we need to sometimes* support Duplicates in getDefined...
3. To keep refactoring less messy List is probably easier than Iterable, e.g. ``CountersPutEffect`` already uses ``List<GameEntity>`` anyway
4. `CountersPutEffect` should be fine since it just fills a Table first, others might need additional logic like ``LifeGainEffect`` but not that much, e.g. the amount needs to be inlined into the Player loop to multiplicate it with `Collections.frequency(tgts, p);`

\* = I don't know yet if there are cases it could break if we just make it default
a) we add special param like "DefinedDuplicates$"
b) if the Defined contains "&" we allow it

@Hanmac
your opinion, especially on 4.?

tool4ever:
Closes #4517

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/DrawEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/DrawEffect.java
@@ -1,7 +1,10 @@
 package forge.game.ability.effects;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+
+import com.google.common.collect.Sets;
 
 import forge.game.ability.AbilityKey;
 import forge.game.ability.AbilityUtils;
@@ -53,7 +56,6 @@ public class DrawEffect extends SpellAbilityEffect {
     @Override
     public void resolve(SpellAbility sa) {
         final Card source = sa.getHostCard();
-        final int numCards = sa.hasParam("NumCards") ? AbilityUtils.calculateAmount(sa.getHostCard(), sa.getParam("NumCards"), sa) : 1;
 
         final boolean upto = sa.hasParam("Upto");
         final boolean optional = sa.hasParam("OptionalDecider") || upto;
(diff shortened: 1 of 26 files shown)


forge-game/src/main/java/forge/game/ability/effects/DrawEffect.java
```diff
@@ -53,19 +56,23 @@ protected String getStackDescription(SpellAbility sa) {
...
-        for (final Player p : getDefinedPlayersOrTargeted(sa)) {
+        final List<Player> tgts = getTargetPlayersWithDuplicates(true, "Defined", sa);
+
+        for (final Player p : Sets.newHashSet(tgts)) {
             if (!p.isInGame()) {
                 continue;
             }
 
+            int numCards = sa.hasParam("NumCards") ? AbilityUtils.calculateAmount(sa.getHostCard(), sa.getParam("NumCards"), sa) : 1;
+            numCards = Collections.frequency(tgts, p);
```
Hanmac, line 74: doesn't this needs to be "numCards *= X" ?

Agetian (approved):

tool4ever:
I could not find other effects needing this treatment, so I would leave them alone for now

forge-gui/res/cardsfolder/c/cunning_strike.txt
```diff
@@ -1,7 +1,8 @@
 Name:Cunning Strike
 ManaCost:3 U R
 Types:Instant
-A:SP$ DealDamage | Cost$ 3 U R | ValidTgts$ Creature | TgtPrompt$ Select target creature | NumDmg$ 2 | SubAbility$ DB1 | SpellDescription$ CARDNAME deals 2 damage to target creature and 2 damage to target player or planeswalker. Draw a card.
-SVar:DB1:DB$ DealDamage | NumDmg$ 2 | ValidTgts$ Player,Planeswalker | TgtPrompt$ Select target player or planeswalker | SubAbility$ DBDraw
+A:SP$ DealDamage | Cost$ 3 U R | ValidTgts$ Creature | TgtPrompt$ Select target creature | NumDmg$ 2 | DamageMap$ True | SubAbility$ DB1 | SpellDescription$ CARDNAME deals 2 damage to target creature and 2 damage to target player or planeswalker. Draw a card.
```
Hanmac, line 4: remove Cost?

Hanmac (approved):

PR #4511, opened 2024-01-12
Add Ravnica Remastered draft

pfirpfel:
Attempted to recreate the [booster collation ](https://magic.wizards.com/en/news/feature/collecting-ravnica-remastered)as good as possible:

> - 8–10 Commons
> - 3–4 Uncommons
> - 1 Rare or mythic rare
> - 1 Mana slot card
>   - 58% Guildgate (common)
>   - 33% Signet (uncommon)
>   - 9% Chromatic Lantern or a shock land (rare)
> - 1 Retro frame card of any rarity

I ignored the retro slot for now.

PR #4524, opened 2024-01-16
Improve ranking scraper

pfirpfel:
Some minor improvements:
- add proper ranking file header (was missing)
- disable debug output of json data
- fix name of altpage argument (script would not work without this fix)
- avoid duplicate card entries (was criticized here #4511 and fixed manually by paulsnoops)

Now the created ranking file looks nice and tidy.

---

Example usage (for Ravnica Remastered):
```sh
cd forge-gui/tools
python3 RankingScraper.py -c RVR    
```

I did not have all the dependencies installed and needed to do this beforehand:
```sh
 pip3 install requests 
```


pfirpfel:
@tehdiplomat Does this look ok to you?

tehdiplomat:
Yea seems fine. I think there's a more pythonic way to convert to a dict but it's fine.

PR #4541, opened 2024-01-17
Fix some rebalanced errors

paulsnoops:
Wizards' being Wizards'

The patch notes for A-Visions of Phyrexia do not reflfect the card on Arena so I've fixed it.

They also included A-Radha, Coalition Warlord from MUL when the rest were from DMU/BRO so moved it to group with the rest.


Northmoc (approved):
Looks good to me!

PR #4545, opened 2024-01-17
DestroyEffect: use Cause if able

### LINKED ISSUE #3512: Replacement effects should keep cause of the original event
https://old.reddit.com/r/mtgrules/comments/q1k1qa/ranar_and_cosmic_intervention_one_or_more_spirits/
https://magicjudge.tumblr.com/post/711336202023862272/generally-do-replacement-effects-override-the
> modifying the result doesn’t change the cause

For the above cases using `ReplacementResult.Updated` should make it easier.

The tricky ones will be stuff where we need to switch to a different effect:
- Opponents SA giving you draw
- you activate _Words of Worship_
- _Rain of Gore_ should NOT replace the life gain

Hanmac:


tool4ever:
Should be ok for Totem Armor but the issue has other cases for it to remain open imo

tool4ever:
Well multiple judges claimed it's a general problem, we should try and keep the original cause if possible

> Specifically, with Unpredictable Cyclone and Teferi's Ageless Insight on the battlefield, if I cycle a card can I replace draw with draw two and then skip cyclone because now it's not "a cycling ability" but an Ageless Insight ability?

> The short answer is no, they don’t. [...]

This does not work yet, so I don't want to close the issue?
But we can still merge this part

Hanmac:
> Well multiple judges claimed it's a general problem, we should try and keep the original cause if possible
> 
> > Specifically, with Unpredictable Cyclone and Teferi's Ageless Insight on the battlefield, if I cycle a card can I replace draw with draw two and then skip cyclone because now it's not "a cycling ability" but an Ageless Insight ability?
> 
> > The short answer is no, they don’t. [...]
> 
> This does not work yet, so I don't want to close the issue? But we can still merge this part

@tool4ever i think the Draw Cards Replacement Types needs to be able to Update the Value instead of creating a new Draw?


Found the Problem, i think there is one Jace, Mirror Mage:
> Draw a card and reveal it. Remove a number of loyalty counters equal to that card’s mana value from Jace, Mirror Mage.

Ruling:
> If the card draw from Jace’s last ability is modified by a replacement effect, you don’t reveal any card, even if the draw was replaced by one or more card draws. No loyalty counters will be removed from Jace in that case. 

don't the rulings contradict each other? 🤔

tool4ever:
Hmn nice find...

1. I would have also said Updated result would be the easiest solution. Here the same argument gets made for "drawn this way" on _Explosions of Riches_:
https://www.reddit.com/r/mtgrules/comments/mu0qio/explosion_of_riches/

> Yes, if the number of cards drawn by the instruction changes due to replacement effects the number of triggers for that spell will likewise change.

2. And I do know for certain if we don't support Updated we need to make sure all "if you do" scripts are refactored to avoid RememberDrawn$ , e.g. on _Academy Loremaster_:

➡️  Currently you could cheat away the effect creation when controlling _Teferi's Ageless Insight_

> The "If [a player] [does, doesn't, or can't]" clause checks whether the player chose to pay an optional cost or started to pay a mandatory cost, regardless of what events actually occurred.

3. It looks like this might also depend on the templating of the cause - with _Jace, Mirror Mage_ it's probably this rule which we need to respect:

> 121.6c. Some effects perform additional actions on a card after it's drawn. If the draw is replaced, the additional action is not performed on any cards that are drawn as a result of that replacement effect or any subsequent replacement effects.

Hanmac:
we could make a Special flag for Jace like "RememberIfDrawIsNotReplaced" ?

tool4ever:
Yea but I don't know how this mix of Updated and Replaced results could work because the special draw rules can turn this single action into two:
- if I first replace Amount 1->2 but then apply _Eruth, Tormented Prophet_ how would I get 2 exile effects from that?
- also it might be problematic because the _Alms Collector_ RE would no longer see the Updated value?

And this approach can't work if it's get replaced to a different type (like my own example in the issue) so perhaps trying for some generic logic that keeps the original cause (and checks that for e. g. Remember params) is better? 🤔 
(Just like you already did here for Destroy)

Hanmac:
there are two replacements ... 
DrawCards with N
and Draw with Card

but the part should be moved into another Issue ticket, so this can be merged

tool4ever (approved):
I'll try this other case next

PR #4547, opened 2024-01-18
"this combat" bug

hungriestPigeon:
Fixed a bug with cards that care about "this combat". For example warriors stand could be used offensively. Also cleaned up a little bit by moving all of the expiry for "this combat" effects to the same place.

### DIFF
--- a/forge-game/src/main/java/forge/game/combat/Combat.java
+++ b/forge-game/src/main/java/forge/game/combat/Combat.java
@@ -31,6 +31,7 @@ import forge.game.spellability.SpellAbility;
 import forge.game.spellability.SpellAbilityStackInstance;
 import forge.game.staticability.StaticAbilityAssignCombatDamageAsUnblocked;
 import forge.game.trigger.TriggerType;
+import forge.game.zone.ZoneType;
 import forge.util.CardTranslation;
 import forge.util.Localizer;
 import forge.util.collect.FCollection;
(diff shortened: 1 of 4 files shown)


forge-game/src/main/java/forge/game/combat/Combat.java
```diff
@@ -154,13 +154,20 @@ public void endCombat() {
         lkiCache.clear();
         combatantsThatDealtFirstStrikeDamage.clear();
 
-        //update view for all attackers and blockers
+        //clear tracking for cards that care about "this combat"
         for (Card c : attackers) {
```
tool4ever, line 158: - thanks, but can you explain what you mean by "could be used offensively"?
- Seems to me like `PhaseHandler.endCombat()` calls `Combat.endCombat()` anyway, so I don't understand yet how moving these parts around would fix something?
- also, before it was iterating over all battlefield cards. I think there's a risk you would miss updating some now if they get removed from combat before

hungriestPigeon: The actual bug was that clearAttackedPlayersMyCombat() wasn't being called at end of combat. This made it possible to use warrior's stand in your own declare attackers step.
In fixing the bug I had to figure out where to put the call to clearAttackedPlayersMyCombat, and noticed that tracking for "this combat" effects was being cleared in multiple places, so I moved them together. 
Good call on the issue with things being removed from combat, I will fix that.

PR #4555, opened 2024-01-19
Make Flash interact with Dryad Arbor

### LINKED ISSUE #4554: Flash needs to work partially with lands
> If a Dryad Arbor gains flash, or you have the ability to play Dryad Arbor as though it had flash (due to Teferi, Mage of Zhalfir or Scout’s Warning, for example), you can ignore the normal timing rules for when during your turn you can play a land, but not any other restrictions. You can’t play Dryad Arbor during another player’s turn, and you can’t play Dryad Arbor if you don’t have any land plays remaining.

tool4ever:
A sneaky interaction 🧙 

Closes #4554


Hanmac:
i need to touch LandAbility later for a better cleanup

Hanmac (approved):

PR #4569, opened 2024-01-20
MKM: Disguise + Cloak

### LINKED ISSUE #4532: MKM: Disguise and Cloak
* Disguise is similar to Morph, with the difference that the Face-Down card has Ward 2. But it doesn't count as Morph.
* Cloak is similar to Manifest, also with Ward 2 there too.

Qarsi Deceiver and MorphOrManifest needs an update.
`Spend this mana only to cast a face-down creature spell, pay a mana cost to turn a manifested creature face up, or pay a morph cost.`
* you can't use the Mana to turn a Disguised card face up.
* but you can use it when the card was manifested, then you can use the Disguise Ability.
* Card.java has `public boolean turnFaceUp(boolean manifestPaid, boolean runTriggers, SpellAbility cause) {` which should be removed or updated? especially the `manifestPaid` check, shouldn't that be done in the SetStateEffect instead?
* isManifestUp probably needs to be removed?
* Qarsi Deceiver probably should use isValid checks instead of Special String

Hanmac:
Closes #4532

### DIFF
--- a/forge-ai/src/main/java/forge/ai/SpellApiToAi.java
+++ b/forge-ai/src/main/java/forge/ai/SpellApiToAi.java
@@ -56,6 +56,7 @@ public enum SpellApiToAi {
             .put(ApiType.Clash, ClashAi.class)
             .put(ApiType.ClassLevelUp, AlwaysPlayAi.class)
             .put(ApiType.Cleanup, AlwaysPlayAi.class)
+            .put(ApiType.Cloak, CloakAi.class)
             .put(ApiType.Clone, CloneAi.class)
             .put(ApiType.CompanionChoose, ChooseCompanionAi.class)
             .put(ApiType.Connive, ConniveAi.class)
--- a/forge-game/src/main/java/forge/game/ability/effects/CloakEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/CloakEffect.java
@@ -25,6 +25,12 @@ public class CloakEffect extends SpellAbilityEffect {
         final int amount = sa.hasParam("Amount") ? AbilityUtils.calculateAmount(source,
                 sa.getParam("Amount"), sa) : 1;
 
+        CardZoneTable triggerList = new CardZoneTable(game.copyLastStateBattlefield(), game.copyLastStateGraveyard());
+        Map<AbilityKey, Object> moveParams = AbilityKey.newMap();
+        moveParams.put(AbilityKey.LastStateBattlefield, triggerList.getLastStateBattlefield());
+        moveParams.put(AbilityKey.LastStateGraveyard, triggerList.getLastStateGraveyard());
+        moveParams.put(AbilityKey.InternalTriggerTable, triggerList);
+
         for (final Player p : getTargetPlayers(sa, "DefinedPlayer")) {
             CardCollection tgtCards;
             if (sa.hasParam("Choices") || sa.hasParam("ChoiceZone")) {
@@ -51,19 +57,14 @@ public class CloakEffect extends SpellAbilityEffect {
                 CardLists.shuffle(tgtCards);
             }
 
-            CardZoneTable triggerList = new CardZoneTable();
-
-            Map<AbilityKey, Object> moveParams = AbilityKey.newMap();
-            moveParams.put(AbilityKey.LastStateBattlefield, triggerList.getLastStateBattlefield());
-            moveParams.put(AbilityKey.LastStateGraveyard, triggerList.getLastStateGraveyard());
-            moveParams.put(AbilityKey.InternalTriggerTable, triggerList);
             for (Card c : tgtCards) {
                 Card rem = c.cloak(p, sa, moveParams);
                 if (rem != null && sa.hasParam("RememberCloaked") && rem.isCloaked()) {
                     source.addRemembered(rem);
                 }
             }
-            triggerList.triggerChangesZoneAll(game, sa);
         }
+
+        triggerList.triggerChangesZoneAll(game, sa);
     }
 }
--- a/forge-game/src/main/java/forge/game/ability/effects/CloakEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/CloakEffect.java
@@ -1,0 +1,69 @@
+package forge.game.ability.effects;
+
+import java.util.Map;
+
+import forge.game.Game;
+import forge.game.ability.AbilityKey;
+import forge.game.ability.AbilityUtils;
+import forge.game.ability.SpellAbilityEffect;
+import forge.game.card.Card;
+import forge.game.card.CardCollection;
+import forge.game.card.CardCollectionView;
+import forge.game.card.CardLists;
+import forge.game.card.CardZoneTable;
+import forge.game.player.Player;
+import forge.game.spellability.SpellAbility;
+import forge.game.zone.ZoneType;
+import forge.util.Localizer;
+
+public class CloakEffect extends SpellAbilityEffect {
+    @Override
+    public void resolve(SpellAbility sa) {
+        final Card source = sa.getHostCard();
+        final Player activator = sa.getActivatingPlayer();
+        final Game game = source.getGame();
+        final int amount = sa.hasParam("Amount") ? AbilityUtils.calculateAmount(source,
+                sa.getParam("Amount"), sa) : 1;
+
+        for (final Player p : getTargetPlayers(sa, "DefinedPlayer")) {
+            CardCollection tgtCards;
+            if (sa.hasParam("Choices") || sa.hasParam("ChoiceZone")) {
+                ZoneType choiceZone = ZoneType.Hand;
+                if (sa.hasParam("ChoiceZone")) {
+                    choiceZone = ZoneType.smartValueOf(sa.getParam("ChoiceZone"));
+                }
+                CardCollectionView choices = game.getCardsIn(choiceZone);
+                if (sa.hasParam("Choices")) {
+                    choices = CardLists.getValidCards(choices, sa.getParam("Choices"), activator, source, sa);
+                }
+                if (choices.isEmpty()) {
+                    continue;
+                }
+
+                String title = sa.hasParam("ChoiceTitle") ? sa.getParam("ChoiceTitle") : Localizer.getInstance().getMessage("lblChooseCards") + " ";
+
+                tgtCards = new CardCollection(activator.getController().chooseCardsForEffect(choices, sa, title, amount, amount, false, null));
+            } else {
+                tgtCards = getTargetCards(sa);
+            }
+
+            if (sa.hasParam("Shuffle")) {
+                CardLists.shuffle(tgtCards);
+            }
+
+            CardZoneTable triggerList = new CardZoneTable();
+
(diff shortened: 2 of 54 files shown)


forge-gui/res/cardsfolder/upcoming/vannifar_evolved_enigma.txt
```diff
@@ -0,0 +1,10 @@
+Name:Vannifar, Evolved Enigma
+ManaCost:2 G U
+Types:Legendary Creature Elf Ooze Wizard
+PT:3/4
+T:Mode$ Phase | Phase$ BeginCombat | ValidPlayer$ You | Execute$ TrigCharm | TriggerZones$ Battlefield | TriggerDescription$ At the beginning of combat on your turn, ABILITY
+SVar:TrigCharm:DB$ Charm | Choices$ DBCloak,DBPutCounters | CharmNum$ 1
+SVar:DBCloak:DB$ Cloak | Choices$ Card.YouCtrl | SpellDescription$ Cloak a card from your hand.
+SVar:DBPutCounters:DB$ PutCounterAll | ValidCards$ Creature.Colorless+YouCtrl | CounterType$ P1P1 | CounterNum$ 1 | SpellDescription$ Put a +1/+1 counter on each colorless creature you control.
+Oracle:At the beginning of combat on your turn, choose one —\n• Cloak a card from your hand. (Put it onto the battlefield face down as a 2/2 creature with ward {2}. Turn it face up any time for its mana cost if it’s a creature card.)\n• Put a +1/+1 counter on each colorless creature you control.
```
Northmoc, line 9: Change `it’s` -> `it's`

DeckHas:Ability$Counters

forge-gui/res/cardsfolder/upcoming/vannifar_evolved_enigma.txt
```diff
@@ -0,0 +1,10 @@
...
+
```
Northmoc, line 10: Two whitespace lines?

Agetian:
Can the AI handle these similar to Morph/Manifest already or does it need additional AI support?

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -4021,9 +4060,12 @@ public static void setFaceDownState(Card c, SpellAbility sa) {
         if (sa.hasParam("FaceDownSetType")) {
             faceDown.setType(new CardType(Arrays.asList(sa.getParam("FaceDownSetType").split(" & ")), false));
         }
+        if (sa.hasParam("FaceDownKeyword")) {
+            faceDown.addIntrinsicKeywords(Arrays.asList(sa.getParam("FaceDownKeyword").split(" & ")));
```
tool4ever, line 4064: shouldn't the spell already have Ward?
I copied the disguised Phantom with _Lithoform Engine_ and the token didn't have it 🤔

Hanmac: should be updated, i think

forge-game/src/main/java/forge/game/spellability/AbilityManaPart.java
```diff
@@ -422,10 +396,6 @@ public boolean meetsManaRestrictions(final SpellAbility sa, String restrictions)
                 }
             }
 
-            if (restriction.equals("CantCastNonArtifactSpells")) {
-                return !sa.isSpell() || sa.getHostCard().isArtifact();
-            }
-
             // the payment is for a resolving SA, currently no other restrictions would allow that
             if (getSourceCard().getGame().getStack().getInstanceMatchingSpellAbilityID(sa.getRootAbility()) != null) {
```
Hanmac, line 400: @tool4ever can you check for what issue this check was for?

because i removed some restrictions above to use the lower isValid one?

tool4ever: hmn this refactor certainly breaks the logic
e. g. you can no longer pay for _Convolute_ UnlessCost with _Thran Turbine_

forge-ai/src/main/java/forge/ai/SpellApiToAi.java
```diff
@@ -56,6 +56,7 @@ public enum SpellApiToAi {
             .put(ApiType.Clash, ClashAi.class)
             .put(ApiType.ClassLevelUp, AlwaysPlayAi.class)
             .put(ApiType.Cleanup, AlwaysPlayAi.class)
+            .put(ApiType.Cloak, CloakAi.class)
```
tool4ever, line 59: I wonder if it should just be mapped to Manifest since the decisions are so closely related 🤔

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
tool4ever: the token is still without it

Hanmac: then its something else broken

the Token Spell on the Stack still has ward, but on the battlefield it has not

maybe the GameAction does something bad with the copy?

forge-game/src/main/java/forge/game/spellability/AbilityManaPart.java
Hanmac: ah UnlessCost was the bad thing i forgot, i tested all the other ones

can't we just remove the InstanceMatching check?

forge-ai/src/main/java/forge/ai/SpellApiToAi.java
Hanmac: they are slightly different so they might need extra logic later

forge-game/src/main/java/forge/game/spellability/AbilityManaPart.java
tool4ever: Unfortunately I added it on the old Gitlab
you can try and I'll see if I can break it without

Hanmac: i think this might need bigger rewrite ...
currently we don't differ correct between using mana for Cost (Spell/Ability) and using mana for Effect at this point

we probably need to pass the "forEffect" Parameter down to this function

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
Hanmac: it seems the Spell does have Ward on the Stack but the copied Card host doesn't ? or i'm seeing this wrong?

can you debug this?

tool4ever: the `abilityCastFaceDown.resolve()` needs better logic so it doesn't turn the copy face down

Hanmac: fixed i think

forge-game/src/main/java/forge/game/spellability/AbilityManaPart.java
Hanmac: especially with "Tin Street Gossip" saying: 
> Spend this Mana only to cast face-down spells or to turn creatures face up

we might need to wait for CR or oracle updates

Hanmac: i might revert this change for now

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -7061,10 +7096,13 @@ public List<SpellAbility> getAllPossibleAbilities(final Player player, final boo
             }
         }
 
-        if (isInPlay() && isFaceDown() && isManifested()) {
-            ManaCost cost = oState.getManaCost();
-            if (oState.getType().isCreature()) {
-                abilities.add(CardFactoryUtil.abilityManifestFaceUp(this, cost));
+        if (isInPlay() && isFaceDown() && oState.getType().isCreature())
```
tool4ever, line 7099: Imo it also needs a check for manacost here
e. g. currently I can unmanifest _Asmoranomardicadaistinaculdacar_ for {0}

> If the card representing that permanent isn't a creature card or it doesn't have a mana cost, it can't be turned face up this way.

forge-game/src/main/java/forge/game/ability/effects/CloakEffect.java
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            if (sa.hasParam("Shuffle")) {
+                CardLists.shuffle(tgtCards);
+            }
+
+            CardZoneTable triggerList = new CardZoneTable(game.copyLastStateBattlefield(), game.copyLastStateGraveyard());
+
+            Map<AbilityKey, Object> moveParams = AbilityKey.newMap();
+            moveParams.put(AbilityKey.LastStateBattlefield, triggerList.getLastStateBattlefield());
```
tool4ever, line 57: I kind of wonder if the lastState shouldn't be the same for all players or they really cloak after another?
e. g. with _Unexplained Absence_

Hanmac: Probably yes

forge-game/src/main/java/forge/game/spellability/AbilityManaPart.java
```diff
@@ -387,8 +387,9 @@ public boolean meetsManaRestrictions(final SpellAbility sa, String restrictions)
                 continue;
             }
 
-            if (restriction.equals("Disturb")) {
-                if (sa.isDisturb()) {
+            if (restriction.equals("FaceDownOrTurnFaceUp")) {
+                if ((sa.isSpell() && sa.getHostCard().isCreature() && sa.isCastFaceDown())
```
tool4ever, line 391: imo no creature check here?
e. g. _Lumithread Field_

Hanmac: You are wrong

It's about the Face Down Creature Spell

Hanmac: But technically, all face down spells are creature spells

tool4ever: Right, but somehow the check fails for me there
> Spend this mana only to cast face-down spells or to turn creatures face up.

Hanmac: The check can probably be removed 

Below too

Because like I said all facedown spells should be creatures.
But otherwise we might need to debug it. I think the Spell Host might not be set yet

tool4ever: fixed

tool4ever (approved):
LGTM

forge-game/src/main/java/forge/game/spellability/AbilityManaPart.java
Hanmac: i asked Dunkatog about the different wording:

* technically there is a difference between "face down spells" and "face down creature spells"
* but in practice, all face down spells are also creature spells 

but they aren't going to update the other card

Hanmac:
@Agetian @tehdiplomat review please

PR #4579, opened 2024-01-22
Merseine fixes

tool4ever:
image


Agetian (approved):

Hanmac (approved):
the commit looks bloated again, but i think that still might be okay?

PR #4599, opened 2024-01-26
MKC: serene_sleuth.txt + support

Northmoc:


forge-game/src/main/java/forge/game/ability/effects/GoadEffect.java
```diff
@@ -40,6 +41,11 @@ public void resolve(SpellAbility sa) {
                 continue;
             }
 
+            if (ungoad) {
+                tgtC.unGoad();
+                continue;
+            }
```
Hanmac, line 47: @Northmoc what do you think? is this better?

Northmoc: Yes – I think that's better! 😄

tool4ever (approved):

forge-game/src/main/java/forge/game/ability/effects/GoadEffect.java
```diff
@@ -40,6 +41,11 @@ public void resolve(SpellAbility sa) {
                 continue;
             }
 
+            if (ungoad) {
+                tgtC.unGoad();
```
tool4ever, line 45: I'm not sure if timestamps should be used here for those static ones like _Baeloth Barrityl_ 🤔

Northmoc: Of course there are no release notes on this 💢 

My feel is something with power LT Baeloth's will be no longer goaded thanks to Serene Sleuth... until the state-based check makes it goaded again

Which I think would be correct with this code

Northmoc: So I just tested this case –
Serene triggered for the Goblin, investigated, and removed the goad, but then Baeloth's static immediately re-goaded
I think this is good to ship for now until WotC clarifies, but I may be missing a rule
image

PR #4603, opened 2024-01-27
Fix Essence Vortex (unless costs not referencing targets properly)

tehdiplomat:
Repro:

Cast Essence Vortex on opponents creature
AI pays 0 life to save creature, because the Unless Ghsot ability doesn't have a reference to the Target.


tool4ever (approved):
one day we'll make a cleaner approach there

PR #4612, opened 2024-01-29
Fix ResolvedLimit not accounting for multiple controllers

tool4ever:


tool4ever:
@Hanmac 
anything needed?

Hanmac (approved):
I think that's okay or would a Map<Player,Integer> be better?

PR #4614, opened 2024-01-30
Draw: use Cause if able

tool4ever:
Another part for #3512


Hanmac (approved):
i think its okay for now until we found a card where this breaks

PR #4616, opened 2024-01-31
Amass Tokens not using the appropriate names for images

### LINKED ISSUE #4344: Question about the naming convention for the Army tokens
**Describe the bug**
Noticed recently that after the amass change with the LTR set the army tokens no longer show up with an image, nor does Forge download a new image. Just wondering what the new naming rule for the tokens is, so I can manually update them.

tehdiplomat:
Closes #4344

### DIFF
--- a/forge-gui/res/tokenscripts/b_0_0_zombie_army.txt
+++ b/forge-gui/res/tokenscripts/b_0_0_zombie_army.txt
@@ -1,0 +1,6 @@
+Name:Zombie Army Token
+ManaCost:no cost
+Types:Creature Zombie Army
+Colors:black
+PT:0/0
+Oracle:
--- a/forge-ai/src/main/java/forge/ai/ability/AmassAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/AmassAi.java
@@ -32,9 +32,11 @@ public class AmassAi extends SpellAbilityAi {
         if (!aiArmies.isEmpty()) {
             return Iterables.any(aiArmies, CardPredicates.canReceiveCounters(CounterEnumType.P1P1));
         }
-        final String tokenScript = "b_0_0_army";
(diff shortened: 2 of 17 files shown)


Hanmac:
No that doesn't work!

Card Text Changes would BREAK it

Hanmac:
like i commented there: #4344

> @kevlahnota we need to update the army token list in the token images

Hanmac (changes requested):
Card Text Changes would break it

tehdiplomat:
@Hanmac Tested with Change type and its working exactly how i expect.

Hanmac:
I think the Original Param might still be altered if a Token of the Card is created 🤔 

I need to check that out

Hanmac (dismissed):
i think this is okay for now until i find a way to break it again

forge-gui/res/tokenscripts/b_0_0_zombie_army.txt
```diff
@@ -0,0 +1,6 @@
+Name:Zombie Army Token
```
tool4ever, line 1: so we need one for each type now?
what about _Lazotep Sliver_?

tehdiplomat: Yea, but these scripts should already have been unique since they have different names.

tool4ever (approved):

Hanmac (changes requested):
Wouldn't it be better to keep the TokenScript and just alter the Image Key?

tehdiplomat:
@Hanmac Why are you blocking this now again? The token is not named just army.

Hanmac:
> @Hanmac Why are you blocking this now again? The token is not named just army.

I just say it would have been easier to just manipulate the image key instead of creating extra token scripts

PR #4619, opened 2024-02-01
Fix ChangeText not being fully applied for trigger effects + traits gained by static

tool4ever:
E. g. _Artificial Evolution_ on _Perimeter Sergeant_

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -4661,6 +4661,9 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
         SpellAbility result = storedSpellAbility.get(stAb, str);
         if (result == null) {
             result = AbilityFactory.getAbility(str, this, stAb);
+            // apply text changes from the statics host
+            result.setIntrinsic(true);
+            result.changeTextIntrinsic(stAb.getChangedTextColors(), stAb.getChangedTextTypes());
             result.setIntrinsic(false);
             result.setGrantorStatic(stAb);
             storedSpellAbility.put(stAb, str, result);
@@ -4672,6 +4675,10 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
         Trigger result = storedTrigger.get(stAb, str);
         if (result == null) {
             result = TriggerHandler.parseTrigger(str, this, false, stAb);
+            // apply text changes from the statics host
+            result.setIntrinsic(true);
+            result.changeTextIntrinsic(stAb.getChangedTextColors(), stAb.getChangedTextTypes());
+            result.setIntrinsic(false);
             storedTrigger.put(stAb, str, result);
         }
         return result;
@@ -4702,6 +4709,10 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
         ReplacementEffect result = storedReplacementEffect.get(stAb, str);
         if (result == null) {
             result = ReplacementHandler.parseReplacement(str, this, false, stAb);
+            // apply text changes from the statics host
+            result.setIntrinsic(true);
+            result.changeTextIntrinsic(stAb.getChangedTextColors(), stAb.getChangedTextTypes());
+            result.setIntrinsic(false);
             storedReplacementEffect.put(stAb, str, result);
         }
         return result;
@@ -4711,6 +4722,10 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
         StaticAbility result = storedStaticAbility.get(stAb, str);
         if (result == null) {
             result = StaticAbility.create(str, this, stAb.getCardState(), false);
+            // apply text changes from the statics host
+            result.setIntrinsic(true);
+            result.changeTextIntrinsic(stAb.getChangedTextColors(), stAb.getChangedTextTypes());
+            result.setIntrinsic(false);
             storedStaticAbility.put(stAb, str, result);
         }
         return result;
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -4658,12 +4658,20 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
     }
 
     public final SpellAbility getSpellAbilityForStaticAbility(final String str, final StaticAbility stAb) {
+        Map<String, String> changedTextColors = stAb.getChangedTextColors();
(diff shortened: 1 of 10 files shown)


Hanmac:
LockTokenScript$ isn't needed for such Keywords because they should NOT be affected by Change Text

tool4ever:
> LockTokenScript$ isn't needed for such Keywords because they should NOT be affected by Change Text
> 
> That us why it's important what Part of the Keyword can be affected by Text Changes and what Parts not

yea the KW won't be modified
but in TriggerHandler it doesn't know that the effect belongs to one
should I try to add some skip logic there?

Hanmac:
the thing is that the TriggerHandler should not need to care

the Card Traits are updated there:

```java

    public void updateChangedText() {
        final List<CardTraitBase> allAbs = ImmutableList.<CardTraitBase>builder()
            .addAll(manaAbilities)
            .addAll(nonManaAbilities)
            .addAll(triggers)
            .addAll(replacementEffects)
            .addAll(staticAbilities)
            .build();
        for (final CardTraitBase ctb : allAbs) {
            if (ctb.isIntrinsic()) {
                ctb.changeText();
            }
        }
    }
```

noticed how it doesn't touch keywords?

because it is more clean to substitute the Keyword directly like from "Protection:Red" to "Protection:Blue" than to and rebuilt the Keyword than to try to update the Keyword parts (where at this point you might not know which parts are part of the internal stuff like the TokenScripts and what parts are not)

Hanmac:
@tehdiplomat what is your opinion there?

tehdiplomat:
Yea copying back over the altered map params feels like the right move here.

tool4ever:
@Hanmac 
another case is static abilities that grant abilities aren't correctly changed, e. g. on _Regal Sliver_:
> Sliver creatures you control have "When this creature enters the battlefield, Slivers you control get +1/+1 until end of turn if you're the monarch. Otherwise, you become the monarch."

the affected creatures are correctly changed to the new type but will use the original type for their triggers

should I manually apply the staticAbility host TextChanges inside `getTriggerForStaticAbility`?

tool4ever:
I modified the trigger [here](https://github.com/Card-Forge/forge/blob/548c3d8f9462cbbe5abddf137232f11847540dcd/forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java#L862) like this and it will work:
```
actualTrigger.setIntrinsic(true);
actualTrigger.changeTextIntrinsic(stAb.getChangedTextColors(), stAb.getChangedTextTypes());
actualTrigger.setIntrinsic(false);
```

Hanmac:
> I modified the trigger [here](https://github.com/Card-Forge/forge/blob/548c3d8f9462cbbe5abddf137232f11847540dcd/forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java#L862) like this and it will work:
> 
> ```
> actualTrigger.setIntrinsic(true);
> actualTrigger.changeTextIntrinsic(stAb.getChangedTextColors(), stAb.getChangedTextTypes());
> actualTrigger.setIntrinsic(false);
> ```

setting it to Intrinsic would probably mess up some Text

like would this Mess up "Afterlife" trigger again?

tool4ever:
I don't think so?
they are keyworded and would never get added via `AddTrigger$`

It might be more problematic if I did the same below with `GainsTriggerAbsOf$` for _Idris_
but since she only checks exiled cards we can leave her alone for now?

tool4ever:
The mapParams of the new Trigger are processed via `AbilityUtils.getSVar` (which applies the changes)
but this still leaves the naked SVars for overridingAbilities:
like you quoted above `updateChangedText()` does not modify those

I'll check if I can modify `TriggerHandler.parseTrigger` accordingly but might be too much trouble 🤔

forge-game/src/main/java/forge/game/spellability/SpellAbility.java
```diff
@@ -1166,7 +1169,7 @@ public SpellAbility copy(Card host, Player activ, final boolean lki) {
 
             // don't use setHostCard to not trigger the not copied parts yet
 
-            copyHelper(clone, host);
+            copyHelper(clone, host, lki || keepTextChanges);
```
tool4ever, line 1172: If it's lki we should always keep textChanges imo -
fixes this scenario:
1. alter Human type of _Jirina, Dauntless General_
2. use _Quicksilver Elemental_ on her
3. activate new ability -> it should affect the new type of 1.

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4661,6 +4661,9 @@ public final SpellAbility getSpellAbilityForStaticAbility(final String str, fina
         SpellAbility result = storedSpellAbility.get(stAb, str);
         if (result == null) {
             result = AbilityFactory.getAbility(str, this, stAb);
+            // apply text changes from the statics host
+            result.setIntrinsic(true);
+            result.changeTextIntrinsic(stAb.getChangedTextColors(), stAb.getChangedTextTypes());
```
Hanmac, line 4666: what part does need to set it to intrinsic?

i mean can't you just pass a parameter to the changeText function to overwrite it?

tool4ever: yea I can try that, just wasn't sure how much I would need to pass that down recursively

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4669,9 +4680,20 @@ public final SpellAbility getSpellAbilityForStaticAbility(final String str, fina
     }
 
     public final Trigger getTriggerForStaticAbility(final String str, final StaticAbility stAb) {
+        Map<String, String> changedTextColors = stAb.getChangedTextColors();
+        Map<String, String> changedTextTypes = stAb.getChangedTextTypes();
         Trigger result = storedTrigger.get(stAb, str);
+        if (result != null) {
+            boolean needNewCache = !result.getChangedTextColors().equals(changedTextColors);
```
tool4ever, line 4687: if the Text of the static host gets changed *again* (after it's cached) we might be stuck with the previous version. Is there a better way compared to these checks?

Hanmac: move this check to CardTraitsBase

like "sameChangedText(CardTraitBase other)"

Hanmac: i might have moved to card traits, but this is fine too

Hanmac (approved):
okay for now i think

PR #4623, opened 2024-02-02
MKM: Forensic Gadgeteer

jumpinjackie:
Abilities copied/derived from:

~Heartstone~ Training Grounds - For the activation cost reduction ability, but to cover *your artifacts*

Riddlesmith - For the "Whenever you cast an artifact spell" trigger template, but to trigger investigation instead of looting

~Only point of doubt is I am unsure whether the `ValidActivatingPlayer$ You` constraint is correct for the cost reduction ability so that only your artifacts get the reduction bonus and not the opponent's. I couldn't figure out a good way to test that.~

All ~other~ abilities work as advertised based on the decks I've built and tested with this card.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/forensic_gadgeteer.txt
+++ b/forge-gui/res/cardsfolder/upcoming/forensic_gadgeteer.txt
@@ -2,7 +2,7 @@ Name:Forensic Gadgeteer
 ManaCost:2 U
 PT:2/3
 Types:Creature Vedalken Artificer Detective
-T:Mode$ SpellCast | ValidCard$ Artifact | ValidActivatingPlayer$ You | Execute$ TrigInvestigate | TriggerZones$ Battlefield | TriggerDescription$ Whenever you cast an artifact spell, investigate
+T:Mode$ SpellCast | ValidCard$ Artifact | ValidActivatingPlayer$ You | Execute$ TrigInvestigate | TriggerZones$ Battlefield | TriggerDescription$ Whenever you cast an artifact spell, investigate.
 SVar:TrigInvestigate:DB$ Investigate
-S:Mode$ ReduceCost | ValidCard$ Artifact | ValidActivatingPlayer$ You | Type$ Ability | Amount$ 1 | MinMana$ 1 | AffectedZone$ Battlefield | Description$ Activated abilities of artifacts you control {1} less to activate. This effect can't reduce the mana in that cost to less than one mana.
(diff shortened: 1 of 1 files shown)


forge-gui/res/cardsfolder/upcoming/forensic_gadgeteer.txt
```diff
@@ -0,0 +1,8 @@
+Name:Forensic Gadgeteer
+ManaCost:2 U
+PT:2/3
+Types:Creature Vedalken Artificer Detective
+T:Mode$ SpellCast | ValidCard$ Artifact | ValidActivatingPlayer$ You | Execute$ TrigInvestigate | TriggerZones$ Battlefield | TriggerDescription$ Whenever you cast an artifact spell, investigate
```
tool4ever, line 5: missing . (at end)

forge-gui/res/cardsfolder/upcoming/forensic_gadgeteer.txt
```diff
@@ -0,0 +1,8 @@
...
+SVar:TrigInvestigate:DB$ Investigate
+S:Mode$ ReduceCost | ValidCard$ Artifact | ValidActivatingPlayer$ You | Type$ Ability | Amount$ 1 | MinMana$ 1 | AffectedZone$ Battlefield | Description$ Activated abilities of artifacts you control {1} less to activate. This effect can't reduce the mana in that cost to less than one mana.
```
tool4ever, line 7: it needs to be scripted more like on _Training Grounds_
it matters when e. g. _Armageddon Clock_ would be in play
see the difference?

forge-gui/res/cardsfolder/upcoming/forensic_gadgeteer.txt
jumpinjackie: Done

forge-gui/res/cardsfolder/upcoming/forensic_gadgeteer.txt
jumpinjackie: Removed `ValidActivatingPlayer$ You` constraint and replaced `Artifact` with `Artifact.YouCtrl` for the `ValidCard$` constraint.

Tested this against a deck with Gadgeteer, Armageddon Clock and Well of Knowledge. Can verify I can activate their clock and wells at reduced costs if they have a Gadgeteer in play.

tool4ever (approved):
thanks!

PR #4629, opened 2024-02-04
forge-game: tests: replace TestNG with Junit 5

allentiak:
There are only a few tests with TestNG (which I originally converted from JUnit 3).

That conversion was a mistake: it's better if all tests use the same testing library.
Albeit at its latest version (JUnit 3 has been EoL'ed long time ago).

### DIFF
--- a/forge-game/pom.xml
+++ b/forge-game/pom.xml
@@ -24,9 +24,9 @@
             <version>1.2</version>
         </dependency>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.13.2</version>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <version>5.10.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/forge-game/pom.xml
+++ b/forge-game/pom.xml
@@ -24,9 +24,9 @@
             <version>1.2</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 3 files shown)


Hanmac:
Was there a benefit from using TestNG over JUnit?

forge-game/pom.xml
```diff
@@ -24,9 +28,9 @@
             <version>1.2</version>
         </dependency>
         <dependency>
-            <groupId>org.testng</groupId>
-            <artifactId>testng</artifactId>
-            <version>7.4.0</version>
+            <groupId>org.junit.jupiter</groupId>
```
tool4ever, line 31: I don't understand your reasoning, you switched us mostly to TestNG here: c7d0dde
we already have a strong dependency on it because of PowerMock
why would we need a second framework again? 👎

allentiak:
You're both right.

At this point it doesn't make much sense to move only a tiny part of the tests away from TestNG.
That being said, as I said before, I do think I made a mistake in moving to TestNG.
Simply because JUnit has a much better support for legacy JDKs. (Since August 2022, TestNG does not support code that targets JDK8 anymore, like Forge; whereas JUnit 5 still supports JDK8.)

Thanks for your feedback.
I'll try to find something more useful to work on.

forge-game/pom.xml
allentiak: (Replied below.)

allentiak:
Since 7.6.0, TestNG does not support JDK8 anymore:
https://github.com/testng-team/testng/releases/tag/7.6.0

PR #4632, opened 2024-02-04
CantExile: add checks for Effects and Costs

### LINKED ISSUE #4496: Rework for CantExile static
yeah, but i think it is still implied

also for touching Exile, i need to add a Static/Replacment for Exile GameAction to add:
"Cause can't cause you to exile stuff" for the WHO card

that is on my TODO list

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/pull/4493#discussion_r1446148766_

Hanmac:
Closes #4496

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -24,6 +24,7 @@ import forge.GameCommand;
 import forge.StaticData;
 import forge.card.CardStateName;
 import forge.card.MagicColor;
+import forge.card.CardType.Supertype;
 import forge.deck.DeckSection;
 import forge.game.ability.*;
 import forge.game.card.*;
@@ -725,7 +726,7 @@ public class GameAction {
 
     private void storeChangesZoneAll(Card c, Zone zoneFrom, Zone zoneTo, Map<AbilityKey, Object> params) {
         if (params != null && params.containsKey(AbilityKey.InternalTriggerTable)) {
-            ((CardZoneTable) params.get(AbilityKey.InternalTriggerTable)).put(zoneFrom != null ? zoneFrom.getZoneType() : null, zoneTo.getZoneType(), c);    
+            ((CardZoneTable) params.get(AbilityKey.InternalTriggerTable)).put(zoneFrom != null ? zoneFrom.getZoneType() : null, zoneTo.getZoneType(), c);
         }
     }
 
@@ -747,16 +748,11 @@ public class GameAction {
         // use FThreads.invokeInNewThread to run code in a pooled thread
         return moveTo(zoneTo, c, null, cause, params);
     }
-    public final Card moveTo(final Zone zoneTo, Card c, Integer position, SpellAbility cause) {
-        return moveTo(zoneTo, c, position, cause, AbilityKey.newMap());
-    }
 
     public final Card moveTo(final ZoneType name, final Card c, SpellAbility cause, Map<AbilityKey, Object> params) {
         return moveTo(name, c, 0, cause, params);
     }
-    public final Card moveTo(final ZoneType name, final Card c, final int libPosition, SpellAbility cause) {
-        return moveTo(name, c, libPosition, cause, AbilityKey.newMap());
-    }
+
     public final Card moveTo(final ZoneType name, final Card c, final int libPosition, SpellAbility cause, Map<AbilityKey, Object> params) {
         // Call specific functions to set PlayerZone, then move onto moveTo
         switch(name) {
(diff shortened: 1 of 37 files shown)


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -747,24 +748,23 @@ public final Card moveTo(final Zone zoneTo, Card c, SpellAbility cause, Map<Abil
...
             case Hand:          return moveToHand(c, cause, params);
             case Library:       return moveToLibrary(c, libPosition, cause, params);
             case Battlefield:   return moveToPlay(c, c.getController(), cause, params);
             case Graveyard:     return moveToGraveyard(c, cause, params);
-            case Exile:         return exile(c, cause, params);
+            case Exile:
+                if (!c.canExiledBy(cause, true)) {
+                    return null;
+                }
+                return exile(c, cause, params);
```
tool4ever, line 767: looks like only MeldEffect needs tweaking because it still calls `exile(CardCollection)` directly

Hanmac: i think i need to ask Judges what exactly happens if a meld pair can't be exiled

if it still tries to meld them or not

tool4ever: yes but you would still exile them anyway
so now that could be prevented too...

Hanmac: i added some fix, but i don't know if that handles all corner cases

forge-gui/res/cardsfolder/upcoming/the_master_multiplied.txt
```diff
@@ -0,0 +1,9 @@
+Name:The Master, Multiplied
+ManaCost:4 B R
+Types:Legendary Creature Time Lord Rogue
+PT:4/3
+K:Myriad
+S:Mode$ IgnoreLegendRule | ValidCard$ Creature.YouCtrl+token | Description$ The "legend rule" doesn't apply to creature tokens you control.
+S:Mode$ CantSacrifice | ValidCard$ Creature.YouCtrl+token | ValidCause$ Triggered.YouCtrl | ForCost$ False | Description$ Triggered abilities you control can't cause you to sacrifice or exile creature tokens you control.
+S:Mode$ CantExile | ValidCard$ Creature.YouCtrl+token | ValidCause$ Triggered.YouCtrl | ForCost$ False | Secondary$ True | Description$ Triggered abilities you control can't cause you to sacrifice or exile creature tokens you control.
```
tool4ever, line 8: If you want to keep the check in `ChangeZoneEffect.removeFromStack()` as future-proofing we might have to add `+inZoneBattlefield` here:
this is because PlayEffect CopyCard$ sets it to token early (which I think fixed some other bug)
this means right now e. g. _Ashiok's Erasure_ wouldn't work completely against a spell from _Reenact the Crime_

tool4ever (dismissed):
couldn't crash it and seems good enough for being only relevant with a single card

Hanmac:
@tehdiplomat review please

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -747,24 +748,23 @@ public final Card moveTo(final Zone zoneTo, Card c, SpellAbility cause, Map<Abil
...
     public final Card moveTo(final ZoneType name, final Card c, final int libPosition, SpellAbility cause, Map<AbilityKey, Object> params) {
         // Call specific functions to set PlayerZone, then move onto moveTo
         switch(name) {
             case Hand:          return moveToHand(c, cause, params);
             case Library:       return moveToLibrary(c, libPosition, cause, params);
             case Battlefield:   return moveToPlay(c, c.getController(), cause, params);
             case Graveyard:     return moveToGraveyard(c, cause, params);
-            case Exile:         return exile(c, cause, params);
+            case Exile:
+                if (!c.canExiledBy(cause, true)) {
```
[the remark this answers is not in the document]
Hanmac: because inside the exile function i can't check if gets exiled by cost or by effect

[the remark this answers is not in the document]
Hanmac: Otherwise i need to stop calling the exile function from moveZone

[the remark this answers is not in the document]
tool4ever: would anything even get stopped from that check there?
The way it's written each effect needs to consider the Static by itself anyway so it can adjust its logic
Same for the costs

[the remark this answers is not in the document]
Hanmac: Costs that exile doesn't call the "moveZone" function

and effects that exile are checked before too.

its mostly just a fail-safe, to prevent it from causing trouble

tool4ever (dismissed):

Agetian:
Needs conflict resolution for the merge

tool4ever (approved):

PR #4649, opened 2024-02-08
PlayEffect: support AltCosts

### LINKED ISSUE #4636: Can't use Conspiracy Unraveler's alt cost during triggers that let you cast spells
If you use Gale, Waterdeep Prodigy's trigger to try to cast something from the graveyard, it won't let you cast that spell with the collect evidence 10 alternative cost.

### LINKED ISSUE #4648: Inventive Iteration chapter 2 draws a card instead of returning an artifact
Windows, current snapshot. 

Inventive Iteration [NEO] has:
II — Return an artifact card from your graveyard to your hand. If you can't, draw a card.

I have a Commander's Sphere in my graveyard, but it's just drawing me a card.

tool4ever:
Closes #4636

There are probably some corner cases where it won't work, but that will be more of a general MayPlay problem.

This also fixes being able to cast fused spells via _Oskar, Rubbish Reclaimer_.

Closes #4648

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -2898,15 +2898,16 @@ public class AbilityUtils {
     }
 
     public static final List<SpellAbility> getBasicSpellsFromPlayEffect(final Card tgtCard, final Player controller) {
-        return getBasicSpellsFromPlayEffect(tgtCard, controller, CardStateName.Original);
+        return getSpellsFromPlayEffect(tgtCard, controller, CardStateName.Original, false);
     }
-    public static final List<SpellAbility> getBasicSpellsFromPlayEffect(final Card tgtCard, final Player controller, CardStateName state) {
+    public static final List<SpellAbility> getSpellsFromPlayEffect(final Card tgtCard, final Player controller, CardStateName state, boolean withAltCost) {
         List<SpellAbility> sas = new ArrayList<>();
-        List<SpellAbility> list = Lists.newArrayList(tgtCard.getBasicSpells());
+        List<SpellAbility> list = new ArrayList<>();
+        collectSpellsForPlayEffect(list, tgtCard.getCurrentState(), controller, withAltCost);
         CardState original = tgtCard.getState(state);
 
         if (tgtCard.isFaceDown()) {
-            Iterables.addAll(list, tgtCard.getBasicSpells(original));
+            collectSpellsForPlayEffect(list, original, controller, withAltCost);
         } else {
             if (tgtCard.isLand()) {
                 LandAbility la = new LandAbility(tgtCard, controller, null);
@@ -2921,7 +2922,7 @@ public class AbilityUtils {
             }
             if (tgtCard.isModal()) {
                 CardState modal = tgtCard.getState(CardStateName.Modal);
-                Iterables.addAll(list, tgtCard.getBasicSpells(modal));
+                collectSpellsForPlayEffect(list, modal, controller, withAltCost);
                 if (modal.getType().isLand()) {
                     LandAbility la = new LandAbility(tgtCard, controller, null);
                     la.setCardState(modal);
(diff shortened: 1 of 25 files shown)


forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -2898,15 +2898,16 @@ public static final void applyManaColorConversion(ManaConversionMatrix matrix, S
...
-        return getBasicSpellsFromPlayEffect(tgtCard, controller, CardStateName.Original);
+        return getSpellsFromPlayEffect(tgtCard, controller, CardStateName.Original, false);
     }
-    public static final List<SpellAbility> getBasicSpellsFromPlayEffect(final Card tgtCard, final Player controller, CardStateName state) {
+    public static final List<SpellAbility> getSpellsFromPlayEffect(final Card tgtCard, final Player controller, CardStateName state, boolean withAltCost) {
         List<SpellAbility> sas = new ArrayList<>();
-        List<SpellAbility> list = Lists.newArrayList(tgtCard.getBasicSpells());
+        List<SpellAbility> list = new ArrayList<>();
+        collectSpellsForPlayEffect(list, tgtCard.getCurrentState(), controller, withAltCost);
         CardState original = tgtCard.getState(state);
```
Hanmac, line 2907: better use the getState(state) instead of getCurrentState

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -2898,15 +2898,16 @@ public static final void applyManaColorConversion(ManaConversionMatrix matrix, S
...
 
         if (tgtCard.isFaceDown()) {
-            Iterables.addAll(list, tgtCard.getBasicSpells(original));
+            collectSpellsForPlayEffect(list, original, controller, withAltCost);
         } else {
             if (tgtCard.isLand()) {
```
Hanmac, line 2912: might not be much, but can the isLand check be moved to collectSpellsForPlayEffect ?

in the Future i would like the LandAbility to be generated, but that should not be that important

forge-game/src/main/java/forge/game/spellability/SpellAbilityRestriction.java
```diff
@@ -249,7 +250,7 @@ public final boolean checkZoneRestrictions(final Card c, final SpellAbility sa)
             }
             if (sa.isSpell()) {
                 final CardPlayOption o = c.mayPlay(sa.getMayPlay());
-                if (o == null) {
+                if (o == null || sa.hasSVar("IsCastFromPlayEffect")) {
```
Hanmac, line 253: from the code i think IsCastFromPlayEffect is very ugly

wouldn't be much simpler to add a boolean property to SpellAbility?

tool4ever: yea this cleans up nicely

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -2903,31 +2903,20 @@ public static final List<SpellAbility> getBasicSpellsFromPlayEffect(final Card t
     public static final List<SpellAbility> getSpellsFromPlayEffect(final Card tgtCard, final Player controller, CardStateName state, boolean withAltCost) {
         List<SpellAbility> sas = new ArrayList<>();
         List<SpellAbility> list = new ArrayList<>();
-        collectSpellsForPlayEffect(list, tgtCard.getCurrentState(), controller, withAltCost);
+        collectSpellsForPlayEffect(list, tgtCard.getState(tgtCard.getCurrentStateName()), controller, withAltCost);
```
Hanmac, line 2906: no i mean use "CardStateName state" from the get param a few lines above

tool4ever: hmn that will be a logic change? 🤔 
currently when facedown it would first add from that state and then from original

Hanmac: okay then it seems to be right

Hanmac (approved):
Looks good

Anything else needs to be refactored by MayPlay anyway

PR #4659, opened 2024-02-10
MKC: add Mentor Trigger

Hanmac:
adds Mentor Trigger

use some more logic for `isKeyword`, so the SpellAbility doesn't need extra `Keyword$ True` flags anymore

moved some parts for Special Keyword trigger that cares if the counters are "really" added to the Card or not (in case of Replacement effects)

> 702.100b A creature “evolves” when one or more +1/+1 counters are put on it as a result of its evolve ability resolving.
>
> 702.149c Some creatures with training have abilities that trigger when they train. “When this creature trains” means “When a resolving training ability puts a +1/+1 counter on this creature.”

Still needs Translation Labels ... I'm unsure about "mentors" vs "mentored" or what the best wording would be

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
@@ -553,9 +553,9 @@ public class CountersPutEffect extends SpellAbilityEffect {
                     }
 
                     if (sa.isKeyword(Keyword.MENTOR)) {
-                        final Map<AbilityKey, Object> runParams = AbilityKey.mapFromCard(sa.getHostCard());
-                        runParams.put(AbilityKey.Mentors, gameCard);
-                        game.getTriggerHandler().runTrigger(TriggerType.Mentor, runParams, false);
+                        final Map<AbilityKey, Object> runParams = AbilityKey.mapFromCard(gameCard);
+                        runParams.put(AbilityKey.Source, sa.getHostCard());
+                        game.getTriggerHandler().runTrigger(TriggerType.Mentored, runParams, false);
                     }
 
                     game.updateLastStateForCard(gameCard);
--- a/forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
@@ -18,6 +18,7 @@ import forge.game.ability.AbilityUtils;
 import forge.game.ability.SpellAbilityEffect;
 import forge.game.card.*;
 import forge.game.event.GameEventRandomLog;
+import forge.game.keyword.Keyword;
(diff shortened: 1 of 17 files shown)


Hanmac:
@tool4ever can you check what labels would be better?
or as name for the trigger?
Mentor, Mentors, Mentored ?

forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
```diff
@@ -535,17 +536,13 @@ protected void resolvePerType(SpellAbility sa, Player placer, CounterType counte
                         gameCard.addCounter(counterType, counterAmount, placer, table);
                     }
 
-                    if (sa.hasParam("Evolve")) {
-                        game.getTriggerHandler().runTrigger(TriggerType.Evolved, AbilityKey.mapFromCard(gameCard),
-                                false);
-                    }
                     if (sa.hasParam("Monstrosity")) {
```
tool4ever, line 539: Monstrosity check can be refactored too?

forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
```diff
@@ -554,9 +551,11 @@ protected void resolvePerType(SpellAbility sa, Player placer, CounterType counte
                         game.getTriggerHandler().runTrigger(TriggerType.Adapt, AbilityKey.mapFromCard(gameCard),
```
tool4ever, line 551: Adapt too?

tool4ever:
The trigger names scheme is somewhat mixed up
Looks like "Mentored" is established only as verb but not noun?
But still ok imo

Hanmac:
@tool4ever 
Evolves and Training cares about the counters put on the creature by the resolving ability
replacement effects can affect the amount, and in Trainings case, can alter the amount of trigger

* while Monstrosity and Renown trigger when the State is set
* while Adapt when the ability resolves.

(also Monstrosity and Adapt aren't keywords, they are just short-cut for the keyword action, may need extra Effect)

the new Mentors trigger when the Mentor Ability resolves. (meaning it doesn't care if the counter or how many are put on the other creature)

Hanmac:
About Trains update

> https://x.com/Dunkatog/status/1757057993626276200?s=20
> That’s a good callout. That should be “One or more counters.” We’ll correct that in the next rules update.

So it should be like Evolves

tehdiplomat:
Hanmac over here trying to get his name in the rules editor book.

tehdiplomat (dismissed):

tehdiplomat:
@Hanmac I think typically for triggers we've tried to use Past tense verbs. So "Mentored" over "Mentors" and "Trained" over "Trains"

Also Trains makes me think of the vehicle which might be confusing.

Issue #4451, opened 2023-12-30
The description of ability seems to be incorrect.


stormcat664:
**Describe the bug**
The description of the creatures targeted by the Sontaran General's ability seems to be incorrect.

**To Reproduce**
1. Trigger Sontaran General's ability.
2. Target creature.
3. See the description of the state.

**Expected behavior**
It's not "this creature can't be blocked", I think "this creature can't block this turn." is correct.

**Screenshots**
![スクリーンショット 2023-12-31 035145](https://github.com/Card-Forge/forge/assets/52299273/20ccb756-cb00-4db9-864c-0ad3f20d780f)

**Desktop (please complete the following information):**
 - OS: Windows10
 - Browser: Firefox
 - Version: forge-gui-desktop-1.6.60-SNAPSHOT-12-29

**Additional context**

stormcat664:
It doesn't update, but I believe the solution is to use SVar:DBPump instead of SVar:DBNoBlock.

PR #4663, opened 2024-02-11
Fixed Sontaran General (#4451)

Morgenmvffel:
Made that the target creatures cant block, instead of the previous unblockable attribute

### DIFF
--- a/forge-gui/res/cardsfolder/s/sontaran_general.txt
+++ b/forge-gui/res/cardsfolder/s/sontaran_general.txt
@@ -5,9 +5,8 @@ PT:5/5
 K:Trample
 K:Haste
 T:Mode$ Attacks | ValidCard$ Card.Self | TriggerZones$ Battlefield | IsPresent$ Creature.attacking+Other | NoResolvingCheck$ True | PresentCompare$ GE2 | Execute$ TrigGoad | TriggerDescription$ Battalion — Whenever CARDNAME and at least two other creatures attack, for each opponent, goad up to one target creature that player controls. Those creatures can't block this turn. (Until your next turn, they attack each combat if able and attack a player other than you if able.)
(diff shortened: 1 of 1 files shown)


forge-gui/res/cardsfolder/s/sontaran_general.txt
```diff
@@ -5,9 +5,8 @@ PT:5/5
...
-SVar:TrigGoad:DB$ Goad |ValidTgts$ Creature.OppCtrl | TgtPrompt$ Select up to one target creature that opponent controls | TargetMin$ 0 | TargetMax$ OneEach | TargetsWithDifferentControllers$ True | SubAbility$ DBNoBlock
-SVar:DBNoBlock:DB$ Effect | RememberObjects$ Targeted | ForgetOnMoved$ Battlefield | StaticAbilities$ Unblockable
-SVar:Unblockable:Mode$ CantBlockBy | ValidAttacker$ Card.IsRemembered | Description$ Those creatures can't be blocked this turn.
+SVar:TrigGoad:DB$ Goad |ValidTgts$ Creature.OppCtrl | TgtPrompt$ Select up to one target creature that opponent controls | TargetMin$ 0 | TargetMax$ OneEach | TargetsWithDifferentControllers$ True | SubAbility$ DBCantBlock
+SVar:DBCantBlock:DB$ Pump | Defined$ Targeted | KW$ HIDDEN CARDNAME can't block. | IsCurse$ True | SpellDescription$ Target creature can't block this turn.
```
tool4ever, line 9: while this would work we're slowly moving away from hidden KW
So just keep using the effect and switch the wrong param to `ValidBlocker$` instead

Morgenmvffel: can you give me an example where "Creature can't block this turn" is made without the hidden KW?

tool4ever: Ahn-Crop Crasher

tool4ever (approved):

PR #4686, opened 2024-02-13
Various card scripts and minor fixes

dracontes:
~(To be marked ready for review tomorrow, 2024/02/14 on a GMT schedule.)~

**Tested card scripts**
- [Sophia, Dogged Detective](https://scryfall.com/card/mkc/8/sophia-dogged-detective)
- [Break Out](https://scryfall.com/card/mkm/190/break-out): ~Tested insofar as the outcomes of "creature card on battlefield with haste for the turn" or "creature card in hand" go. In relative short notice, I couldn't figure out how to skip the step where the revealed card, when it's mana value 2 or less, shows up in your hand along with a prompt on whether you want to put it on the battlefield or not (unless that's how it works in Forge).~ Works better now per discussion below. 
- [Sludge Titan](https://scryfall.com/card/clu/43/sludge-titan): ~Splitting milled card selection into one optional step for each relevant type isn't the most elegant solution but I couldn't find much help otherwise in existing scripts.~ A better script is now used per discussion below.

**Minor fixes**
- Changed filename for Uncivil Unrest
- Updated deckbuilding tags for Syr Ginger, the Meal Ender and Dark Heart of the Wood
- Cleaned up formatting issues, notably vertical bar glyph that wasn't completely bordered by whitespace outside of deckbuilding tags.
- Cleaned up sundry typos.
- Completed Protean War Machine's deckbuilding tags.

**NB:** This [pull request](https://github.com/Card-Forge/forge/pull/4689) was mishap from clicking too fast. Please disregard. ([This one](https://github.com/Card-Forge/forge/pull/4701) too).


forge-gui/res/cardsfolder/upcoming/break_out.txt
```diff
@@ -0,0 +1,9 @@
+Name:Break Out
+ManaCost:R G
+Types:Sorcery
+A:SP$ Dig | Cost$ R G | DigNum$ 6 | ChangeNum$ 1 | Optional$ True | ForceRevealToController$ True | ChangeValid$ Card.Creature | RestRandomOrder$ True | RememberChanged$ True | SubAbility$ DBChangeZoneBattlefield | StackDescription$ SpellDescription | SpellDescription$ Look at the top six cards of your library. You may reveal a creature card from among them. If that card has mana value 2 or less, you may put it onto the battlefield and it gains haste until end of turn. If you didn’t put the revealed card onto the battlefield this way, put it into your hand. Put the rest on the bottom of your library in a random order.
```
tool4ever, line 4: Cost$ on spells is redundant if it's just the standard ManaCost

tool4ever, line 4: this Dig is doing too much in the wrong order
switch to PeekAndReveal (then you can also use Origin$ in ChangeZone for a cleaner prompt)

dracontes: Duly noted. Incidentally going over the notes I collated about two years ago non-permanent spell `Cost$` redundancy was part of the cleanup I unfortunately didn't get around to. I take it haring off in that direction is still welcome?

In any case, it took a while of racking my brain but I figured Break Out out satisfactorily enough, though do tell if something is amiss.

forge-gui/res/cardsfolder/upcoming/break_out.txt
```diff
@@ -0,0 +1,11 @@
+Name:Break Out
+ManaCost:R G
+Types:Sorcery
+A:SP$ PeekAndReveal | Defined$ You | PeekAmount$ 6 | NoReveal$ True | RememberPeeked$ True | SubAbility$ PickOne | StackDescription$ SpellDescription | SpellDescription$ Look at the top six cards of your library. You may reveal a creature card from among them. If that card has mana value 2 or less, you may put it onto the battlefield and it gains haste until end of turn. If you didn’t put the revealed card onto the battlefield this way, put it into your hand. Put the rest on the bottom of your library in a random order.
```
tool4ever, line 4: why don't you use RevealValid + RevealOptional + RememberRevaled = no extra ChooseCard?

dracontes: I tried a good number of iterations on those but it frankly got me nowhere as reasonable.
I reached the solution above after looking over the few examples of `PeekAndReveal` used for similar effects (I borrowed heavily from Nine-Fingers Keene's triggered ability) and noticing there's a `SubAbility` for the "put the rest on the bottom of your library in a random order" part of the effect. For `Dig` that's built-in setting. I suspected there could be a conflict for `PeekAndReveal` when using `RememberRevealed` since the revealed card and the cards that are put on the bottom of the library are called back to by `Remembered` (or `IsRemembered`). I'm certainly not disabused of that notion.

forge-gui/res/cardsfolder/upcoming/sludge_titan.txt
```diff
@@ -0,0 +1,14 @@
+Name:Sludge Titan
+ManaCost:4 BG BG
+Types:Creature Zombie Giant
+PT:6/6
+K:Trample
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigMill | TriggerDescription$ Whenever CARDNAME enters the battlefield or attacks, mill five cards. You may put a creature card and/or a land card from among them into your hand.
+T:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigMill | TriggerZones$ Battlefield | Secondary$ True | TriggerDescription$ Whenever CARDNAME enters the battlefield or attacks, mill five cards. You may put a creature card and/or a land card from among them into your hand.
+SVar:TrigMill:DB$ Mill | NumCards$ 5 | Defined$ You | RememberMilled$ True | SubAbility$ DBChangeZoneCreature
+SVar:DBChangeZoneCreature:DB$ ChangeZone | Hidden$ True | Origin$ Graveyard,Exile | Destination$ Hand | ChangeType$ Creature.YouOwn+IsRemembered | Optional$ True | SelectPrompt$ You may select a creature card. | SubAbility$ DBChangeZoneLand
```
tool4ever, line 9: do it like on _Conflux_

forge-gui/res/cardsfolder/upcoming/break_out.txt
```diff
@@ -0,0 +1,11 @@
...
+SVar:PickOne:DB$ ChooseCard | Defined$ You | Amount$ 1 | ChoiceTitle$ Choose a creature card to reveal | Choices$ Card.Creature+IsRemembered | ChoiceZone$ Library | Reveal$ True | SubAbility$ DBChangeZoneBattlefield
+SVar:DBChangeZoneBattlefield:DB$ ChangeZone | Origin$ Library | Destination$ Battlefield | Defined$ ChosenCard | ConditionDefined$ Remembered | ConditionPresent$ Card.Creature+cmcLE2+ChosenCard | ConditionCompare$ EQ1 | Optional$ True | SubAbility$ DBPump | StackDescription$ None
+SVar:DBPump:DB$ Pump | KW$ Haste | Defined$ ChosenCard | SubAbility$ DBChangeZoneHand | StackDescription$ None
```
tool4ever, line 7: you might also need to work with Imprint here so it only Pumps when you have it ETB

Hanmac: @tool4ever maybe use AnimateSubAbility for this?

tool4ever: why? it says "gains" so it shouldn't ETB with haste

forge-gui/res/cardsfolder/upcoming/sludge_titan.txt
dracontes: Much obliged for the pointer! Works like a charm :)

forge-gui/res/cardsfolder/upcoming/break_out.txt
dracontes: Figured out a way to implement `Imprint` on the script for the chosen creature gaining haste. I'm just not sure how to verify that the creature first enters the battlefield and then gains haste.

tool4ever: well you should move the Imprint to the ChangeZone line instead
(I don't want to confuse you further but currently Pump only works on battlefield by default but ChosenCard would be the old library object which will eventually need to fail due to timestamp checks)

dracontes: Took a bit to untangle but it's done :)

tool4ever: Why are you trying to Pump two things now? Imprinted will be the battlefield object

dracontes: Sorry, that was a remainder of previous digressions due to some confusion on my part. Script fixed: it functions as expected.

tool4ever (approved):
Alright, I think that should be good now

Make sure you rebase your master properly for a future PR first or create them from another branch, since I will squash this one too (so we don't end up with 100+ additional commits in the history) ⚠️ 

Thanks

PR #4687, opened 2024-02-13
Fix scripts

### LINKED ISSUE #4684: Sentinel of the Pearl Trident ETB isn't optional
Windows, current snapshot.

Sentinel of the Pearl Trident (2XM) has an ETB ability which should be optional, but the current implementation just does it without asking.

> When Sentinel of the Pearl Trident enters the battlefield, you may exile target historic permanent you control. If you do, return that card to the battlefield under its owner's control at the beginning of the next end step.

tool4ever:
Closes #4684


forge-gui/res/cardsfolder/s/sentinel_of_the_pearl_trident.txt
```diff
@@ -3,7 +3,7 @@ ManaCost:4 U
...
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigExile | OptionalDecider$ You | TriggerDescription$ When CARDNAME enters the battlefield, you may exile target historic permanent you control. If you do, return that card to the battlefield under its owner's control at the beginning of the next end step. (Artifacts, legendaries, and Sagas are historic.)
 SVar:TrigExile:DB$ ChangeZone | ValidTgts$ Permanent.Historic+YouCtrl | TgtPrompt$ Select target historic permanent you control | Origin$ Battlefield | Destination$ Exile | RememberChanged$ True | SubAbility$ DelTrigEffect
 SVar:DelTrigEffect:DB$ DelayedTrigger | Mode$ Phase | Phase$ End Of Turn | Execute$ TrigReturn | RememberObjects$ RememberedLKI | TriggerDescription$ Return the exiled card to the battlefield under its owner's control at the beginning of the next end step. | SubAbility$ DBCleanup | AILogic$ Always
```
Hanmac, line 8: should be written as Cost for "If you do" ?

tool4ever: can't be done currently because it targets

Hanmac: if it targets, then this is okay

Hanmac (approved):

PR #4696, opened 2024-02-15
Fix alternative costs from static abilities

tchataway:
Currently, when a static ability grants an alternative cost to spells (for example, Hunting Velociraptor giving Dinosaur spells "prowl", and Henzie "Toolbox" Torre giving mana value 4 and greater creature spells "blitz"), the alternative costs only apply when casting from the Hand. In the case of Toolbox, the mana value is also ignored, allowing any creature spell to be cast from the hand with blitz.

This change allows alternative costs granted from static abilities to function properly. The main change in behaviour here is that the alternative cost can be paid whenever it is supposed to be able to be paid. E.g. if you can cast creature spells from the top of your library, you can now cast them using alternative costs granted by static abilities. Basically, if you could cast it normally, and there aren't other restrictions for the alternative cost (e.g. "flashback" and "escape" only casting from the graveyard), then you are able to cast it using that cost (relevant rules here being `601.2`, `601.2b`, and `601.2f-h`).

## Overview
- Move "alternative costs from static ability" logic from inside `getAlternativeCosts` to new function `getAlternativeCostsGrantedByStaticAbilities` function which is called before alternative costs are retrieved for card
- Update "alternative costs from static ability" logic to check all alternative cost keywords, not just blitz
- Move `cmcGE4` restriction from Toolbox's static ability's `AddKeyword` parameter to its `Affected` parameter
- Change Hunting Velociraptor's static ability's `AffectedZone` parameter from `Hand` to `Stack`

## Testing
- No unit tests (I can add some if necessary)
- Local testing of both Toolbox and Hunting Velociraptor, ensuring
    - Restrictions are applied
    - Cost reductions are applied

### DIFF
--- a/forge-game/src/main/java/forge/game/GameActionUtil.java
+++ b/forge-game/src/main/java/forge/game/GameActionUtil.java
@@ -352,26 +352,28 @@ public final class GameActionUtil {
         return alternatives;
     }
     
-    /**
-     * Checks a card to see if it qualifies to receive keywords from static abilities
-     * that grant alternative casting costs to spells on the stack, for example
-     *  - Toolbox's ability that gives Blitz to creature spells with mana value 4 or greater
-     *  - Hunting Velociraptor's ability that gives Prowl to Dinosaur spells
-     * 
-     * @param source The card to check eligibility for.
-     * @return Zero or more spell abilities, each representing an alternative casting cost.
-     */
-    public static final FCollectionView<SpellAbility> getAlternativeCostsGrantedByStaticAbilities(Card source) {
-    	final FCollection<SpellAbility> alternatives = new FCollection<SpellAbility>();
-    	final Game game = source.getGame();
-    	
-        if (!game.getAction().hasStaticAbilityAffectingZone(ZoneType.Stack, StaticAbilityLayer.ABILITIES)) {
-        	return alternatives;
-        }
-        
-        // double freeze tracker, so it doesn't update view
-        game.getTracker().freeze();
-        
+	/**
+	 * Checks a card to see if it qualifies to receive keywords from static
+	 * abilities that grant alternative casting costs to spells on the stack, for
+	 * example, Toolbox's ability that gives blitz to creature spells with mana
+	 * value 4 or greater, and Hunting Velociraptor's ability that gives prowl to
+	 * Dinosaur spells.
+	 * 
+	 * @param source The card to check eligibility for.
+	 * @return Zero or more spell abilities, each representing an alternative
+	 *         casting cost.
+	 */
+	public static final FCollectionView<SpellAbility> getAlternativeCostsGrantedByStaticAbilities(Card source) {
+		final FCollection<SpellAbility> alternatives = new FCollection<SpellAbility>();
+		final Game game = source.getGame();
+
+		if (!game.getAction().hasStaticAbilityAffectingZone(ZoneType.Stack, StaticAbilityLayer.ABILITIES)) {
+			return alternatives;
+		}
+
+		// double freeze tracker, so it doesn't update view
+		game.getTracker().freeze();
+
 		Zone oldZone = source.getLastKnownZone();
 		Card creatureCandidate = source; // Candidate to receive keyword.
 		if (!source.isLKI()) {
@@ -390,7 +392,7 @@ public final class GameActionUtil {
 				// If the operation doesn't throw an IllegalArgumentException, the keyword
 				// is an alternative cost.
 				AlternativeCost.valueOf(keywordInterface.getKeyword().toString());
-			
+
 				for (SpellAbility iSa : keywordInterface.getAbilities()) {
 					// do only non intrinsic
 					if (!iSa.isIntrinsic()) {
@@ -414,7 +416,7 @@ public final class GameActionUtil {
 		game.getTracker().unfreeze();
 
 		return alternatives;
-    }
+	}
 
     public static List<OptionalCostValue> getOptionalCostValues(final SpellAbility sa) {
         final List<OptionalCostValue> costs = Lists.newArrayList();
--- a/forge-game/src/main/java/forge/game/GameActionUtil.java
+++ b/forge-game/src/main/java/forge/game/GameActionUtil.java
@@ -17,6 +17,7 @@
  */
 package forge.game;
 
+import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.List;
@@ -52,6 +53,8 @@ import forge.game.zone.Zone;
 import forge.game.zone.ZoneType;
 import forge.util.Lang;
 import forge.util.TextUtil;
+import forge.util.collect.FCollection;
+import forge.util.collect.FCollectionView;
 
 
 /**
@@ -269,35 +272,6 @@ public final class GameActionUtil {
                     foretold.putParam("AfterDescription", "(Foretold)");
                     alternatives.add(foretold);
                 }
-
-                // some needs to check after ability was put on the stack
-                // Currently this is only checked for Toolbox and that only cares about creature spells
-                if (source.isCreature() && game.getAction().hasStaticAbilityAffectingZone(ZoneType.Stack, StaticAbilityLayer.ABILITIES)) {
-                    Zone oldZone = source.getLastKnownZone();
-                    Card blitzCopy = source;
-                    if (!source.isLKI()) {
-                        blitzCopy = CardUtil.getLKICopy(source);
-                    }
-                    blitzCopy.setLastKnownZone(game.getStackZone());
-                    lkicheck = true;
-
-                    blitzCopy.clearStaticChangedCardKeywords(false);
-                    CardCollection preList = new CardCollection(blitzCopy);
-                    game.getAction().checkStaticAbilities(false, Sets.newHashSet(blitzCopy), preList);
-
-                    // currently only for Keyword BLitz, but should affect Dash probably too
-                    for (final KeywordInterface inst : blitzCopy.getKeywords(Keyword.BLITZ)) {
-                        // TODO with mana value 4 or greater has blitz.
-                        for (SpellAbility iSa : inst.getAbilities()) {
-                            // do only non intrinsic
-                            if (!iSa.isIntrinsic()) {
-                                alternatives.add(iSa);
-                            }
-                        }
-                    }
-                    // need to reset to Old Zone, or canPlay would fail
-                    blitzCopy.setLastKnownZone(oldZone);
-                }
             }
 
             // reset static abilities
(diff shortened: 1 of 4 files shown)


forge-game/src/main/java/forge/game/GameActionUtil.java
```diff
@@ -270,35 +273,6 @@ public static final List<SpellAbility> getAlternativeCosts(final SpellAbility sa
...
-                        for (SpellAbility iSa : inst.getAbilities()) {
-                            // do only non intrinsic
-                            if (!iSa.isIntrinsic()) {
-                                alternatives.add(iSa);
-                            }
-                        }
-                    }
-                    // need to reset to Old Zone, or canPlay would fail
-                    blitzCopy.setLastKnownZone(oldZone);
-                }
```
tchataway, line 301: This logic was moved, mostly unchanged, into `getAlternativeCostsGrantedByStaticAbilities`.

forge-game/src/main/java/forge/game/GameActionUtil.java
```diff
@@ -378,6 +352,72 @@ public static final List<SpellAbility> getAlternativeCosts(final SpellAbility sa
...
+     * @return Zero or more spell abilities, each representing an alternative
+     *         casting cost.
+     */
+    public static final FCollectionView<SpellAbility> getAlternativeCostsGrantedByStaticAbilities(Card source) {
+        final FCollection<SpellAbility> alternatives = new FCollection<SpellAbility>();
+        final Game game = source.getGame();
+
+        if (!game.getAction().hasStaticAbilityAffectingZone(ZoneType.Stack, StaticAbilityLayer.ABILITIES)) {
+            return alternatives;
+        }
```
tchataway, line 373: The previous logic checked if the `source` is a creature, but I removed that here, because the restriction is already checked when applying the static ability in `checkStaticAbilities`, below.

This will allow static abilities that grant alternative costs to non-creature spells to function, also.

forge-game/src/main/java/forge/game/GameActionUtil.java
```diff
@@ -378,6 +352,72 @@ public static final List<SpellAbility> getAlternativeCosts(final SpellAbility sa
...
+
+        // double freeze tracker, so it doesn't update view
+        game.getTracker().freeze();
+
+        Zone oldZone = source.getLastKnownZone();
+        Card creatureCandidate = source; // Candidate to receive keyword.
+		
+        if (!source.isLKI()) {
+            creatureCandidate = CardUtil.getLKICopy(source);
+        }
```
tchataway, line 383: I admit I don't really understand the concept of Last Known Information, so I've tried to leave it untouched, but I noticed that some of this logic was being run before the static ability checks were happening in the old location, so I included it here in case it's necessary.

There's also corresponding code that is executed afterward, which book-ends the logic.

forge-game/src/main/java/forge/game/GameActionUtil.java
```diff
@@ -378,6 +352,72 @@ public static final List<SpellAbility> getAlternativeCosts(final SpellAbility sa
...
+            try {
+                // Try to find the keyword in the list of alternative cost keywords.
+                // If the operation doesn't throw an IllegalArgumentException, the keyword
+                // is an alternative cost.
+                AlternativeCost.valueOf(keywordInterface.getKeyword().toString());
+
+                for (SpellAbility iSa : keywordInterface.getAbilities()) {
+                    // do only non intrinsic
+                    if (!iSa.isIntrinsic()) {
+                        iSa.setHostCard(source);
```
tchataway, line 400: This line is new. Without it, the alternative cost isn't copied for additional `CardPlayOption`s, which means it can only be paid when casting from the Hand.

forge-game/src/main/java/forge/game/GameActionUtil.java
```diff
@@ -378,6 +352,72 @@ public static final List<SpellAbility> getAlternativeCosts(final SpellAbility sa
...
+                        alternatives.add(iSa);
+                    }
+                }
+            } catch (IllegalArgumentException e) {
+                // The keyword interface isn't for an alternative cost keyword.
+                continue;
+            }
```
tchataway, line 407: Some places I've worked frown upon using exceptions this way.

If it violates convention here, I'll write a new function to verify the keyword represents an alternative cost rather than calling `valueOf` on the `AlternativeCost` enum.

forge-gui/res/cardsfolder/h/hunting_velociraptor.txt
```diff
@@ -3,6 +3,6 @@ ManaCost:2 R
 Types:Creature Dinosaur
 PT:3/2
 K:First Strike
-S:Mode$ Continuous | Affected$ Dinosaur.YouCtrl | AddKeyword$ Prowl:2 R | AffectedZone$ Hand | Description$ Dinosaur spells you cast have prowl {2}{R}. (You may cast a spell for its prowl cost if you dealt combat damage to a player this turn with a creature with any of its creature types.)
+S:Mode$ Continuous | Affected$ Dinosaur.YouCtrl | AddKeyword$ Prowl:2 R | AffectedZone$ Stack | Description$ Dinosaur spells you cast have prowl {2}{R}. (You may cast a spell for its prowl cost if you dealt combat damage to a player this turn with a creature with any of its creature types.)
```
tchataway, line 6: Just as an aside: I discovered that this ability didn't work because I was playing my Jurassic Park commander deck and couldn't cast Pantlaza, Sun-Favored from the Command Zone. I'm chuffed that it's working as intended, now! 😊

forge-gui/res/cardsfolder/h/henzie_toolbox_torre.txt
```diff
@@ -2,7 +2,7 @@ Name:Henzie "Toolbox" Torre
 ManaCost:B R G
 Types: Legendary Creature Devil Rogue
 PT:3/3
-S:Mode$ Continuous | Affected$ Creature.YouCtrl | AffectedZone$ Stack | AddKeyword$ Blitz:CardManaCost:Spell.Creature+cmcGE4 | Description$ Each creature spell you cast with mana value 4 or greater has blitz. The blitz cost is equal to its mana cost. (You may choose to cast that spell for its blitz cost. If you do, it gains haste and "When this creature dies, draw a card." Sacrifice it at the beginning of the next end step.)
+S:Mode$ Continuous | Affected$ Creature.YouCtrl+cmcGE4 | AffectedZone$ Stack | AddKeyword$ Blitz:CardManaCost:Spell.Creature | Description$ Each creature spell you cast with mana value 4 or greater has blitz. The blitz cost is equal to its mana cost. (You may choose to cast that spell for its blitz cost. If you do, it gains haste and "When this creature dies, draw a card." Sacrifice it at the beginning of the next end step.)
```
tchataway, line 5: The previous implementation was so close to being correct!

Hanmac: This Check is for MDFC

is the affected checked again when casting the Backside?

tchataway: I'll look into this later today (it's currently 3am local time for me).

Hanmac: For Henzie, it also needs to check the Spell Mana Value because for X Creature Spells

so you can only Blitz them when X is high enough

tchataway: Yes, that wasn't happening at all before. The exclusive restriction wasn't defined in the right place. I moved it to the `Affected` param, and now it actually does check it (i.e. `Affected Creature.YouCtrl+cmcGE4`).

tchataway: I checked and I've put together a functionality matrix of expected behaviours across the `master` branch and this one (the fork).

| Expected Behaviour                                                                 | master | fork |
|------------------------------------------------------------------------------------|:------:|:----:|
| Standard card, alternative cost printed on it, can cast from all legal locations   |    ✔️    |   ✔️  |
| Standard card, alternative cost _granted to_ it, can cast from all legal locations |        |   ✔️  |
| MDFC Front, alternative cost printed on it, can cast from all legal locations      | ✔️      | ✔️    |
| MDFC Back, alternative cost printed on it, can cast from all legal locations       | ✔️      | ✔️    |
| MDFC Front, alternative cost _granted to_ it, can cast from all legal locations    |        | ✔️    |
| MDFC Back, alternative cost _granted to_ it, can cast from all legal locations     |        |      |
| Toolbox's ability correctly applies to creatures with mana value 4 or greater      |        | ✔️    |

The main takeaway is that the fork doesn't provide "MDFC Back, alternative cost _granted to_ it, can cast from all legal locations".

I can find a solution and add it to this PR, or I can address it in a separate one. It's your call. 👍

Hanmac: The problem are X Creatures

Like Endless One with ManaCost =X

You allow to X them, but only if you choose X>=4

That's why it needs to apply to all creature Spells first, and is later checked again when you choose X and the spell is put on the Stack

In any case you choose X <=3 then you can't cast the spell and the action is canceled 


The MDFC where working with Blitz before, you need to test them different to prowl
So you should fix the MDFC in this MR too

tchataway: > The MDFC where working with Blitz before, you need to test them different to prowl
So you should fix the MDFC in this MR too

Will do!

> The problem are X Creatures
> 
> Like Endless One with ManaCost =X
> 
> You allow to X them, but only if you choose X>=4
> 
> That's why it needs to apply to all creature Spells first, and is later checked again when you choose X and the spell is put on the Stack
> 
> In any case you choose X <=3 then you can't cast the spell and the action is canceled

Was there a rules errata that lets Henzie work this way? Cursory searches online indicate that Henzie works the way you describe, but I can't find any official sources for it. I want to make sure I code this properly so having actual rules text for it would be ideal.

Thanks so much for reviewing this PR btw, I appreciate the thoroughness!

tool4ever: Hey there!

Thanks for trying to work on the engine side of Forge.

Unfortunately you picked a pretty complex area as your first project :/

We haven't been happy with the MayPlay / alternative cost handling for quite some time, but it has also proven rather difficult to fix smaller aspects of it without breaking something else at the same time...(so it has been left alone mostly for a bigger refactor)

For your X interaction it's this:
> 202.3e. When calculating the mana value of an object with an {X} in its mana cost, X is treated as 0 while the object is not on the stack, and X is treated as the number chosen for it while the object is on the stack.

tchataway: > Unfortunately you picked a pretty complex area as your first project :/
Just my luck! 😆 

> We haven't been happy with the MayPlay / alternative cost handling for quite some time, but it has also proven rather difficult to fix smaller aspects of it without breaking something else at the same time...(so it has been left alone mostly for a bigger refactor)
Right, fair enough. It is turning out to be much hairier than anticipated. Would it be better to just shelve this in lieu of a more comprehensive refactor later, then?

> For your X interaction it's this:
>
> 202.3e. When calculating the mana value of an object with an {X} in its mana cost, X is treated as 0 while the object is not on the stack, and X is treated as the number chosen for it while the object is on the stack.
Thank you! I was very confused but this makes sense. I'm obviously very new to the codebase so I'm not clear on how Forge lines things up relative to the "proposal to cast a spell". When the user clicks a card in their hand, for example, Forge appears to get all costs and alternative costs at that point, and present them as a list for the user. For X cost creatures in Toolbox's case, should it present the Blitz option the way it normally would, but with X in the cost, then prompt for X, then check the total mana value of the spell on the stack and either cancel/prompt for mana?

tool4ever: > Right, fair enough. It is turning out to be much hairier than anticipated. Would it be better to just shelve this in lieu of a more comprehensive refactor later, then?

It's your choice I guess. I'm not saying you can't try to tackle this and splitting it across consecutive PR is also sometimes easier. Just don't want you getting frustrated by this too quickly ;)

You noticed correctly that spell casting is one of the parts where we're not super accurate in the order according to the CR (sometimes it's also different for AI vs Human PlayerController). I suppose we could have less trouble in that regard if we would basically allow any card to be moved to the stack first as outlined by 601.2a. But then again this might end up too confusing GUI wise, which is one of the reasons we're doing some of these choices in sort of a "look-ahead" fashion. Another one would be this:
> 601.3a. If an effect prohibits a player from casting a spell with certain qualities, that player may consider any choices to be made during that spell's proposal that may cause those qualities to change. If any such choices could cause that effect to no longer prohibit that player from casting that spell, the player may begin to cast the spell, ignoring the effect.

So yea, that gets messy pretty quickly... 🤹‍♂️

tchataway: Ah, geez. It all makes sense now. That clears everything up, thanks @Hanmac.

I saw the comment in `getAlternativeCosts` that said `// TODO with mana value 4 or greater has blitz.` and that, coupled with the fact that the Blitz cost was appearing on the list of casting options when clicking a card, led me to believe that the behaviour hadn't been implemented yet. If I'd tried to actually click the Blitz option, I would have seen that it didn't work for mana values < 4.

It didn't help that I didn't know about rule `202.3e`, either (cheers, @tool4ever!).

Thanks for your patience, @Hanmac. I've been very obtuse.

So, to summarise:
- On `master`, static abilities that grant alternative costs work for normal cards and both faces of MDFCs in the hand, but not when able to cast from the top of the library or the command zone, etc.
- On the fork, static abilities that grant alternative costs work for normal cards and the front face of MDFCs for all zones they could normally be cast from, but not the back face of MDFCs, and it also breaks Toolbox for X costs (sorry about that 😩 )

I thought that moving the static ability checking code back to where it came from and duplicating the loop over `source.mayPlay(activator)`  would do the trick for the back face of MDFCs, but it doesn't seem to have worked. I'd prefer to find a way to enable the extra `CardPlayOption`s that minimises disruption, but I'll need to do some more investigation.

Thanks again, both of you, for taking the time to educate me on the codebase!

tchataway: > It's your choice I guess. I'm not saying you can't try to tackle this and splitting it across consecutive PR is also sometimes easier. Just don't want you getting frustrated by this too quickly ;)

Oh I'm happy to keep thrashing away at this. I've been a software engineer for over a decade now so I'm no stranger to frustration 😆

Thanks for the warm welcome. I'm looking forward to learning more about how Forge does things. I might take a look at some of the beginner friendly issues to get a bit more orientation, as well.

> So yea, that gets messy pretty quickly... 🤹‍♂️
Yes, absolutely. Those two rules, if implemented faithfully, would turn Forge into a Christopher Nolan film. 😬

Hanmac:
@tool4ever remember me to recreate the Issue about May Play again

in short summary, we need to split the MayPlay logic between multiple effects that may or may not be combined:
* Effects that alter the Cost like `Rooftop Storm`, that should be use a Static instead of a Keyword like Alternative Cost:0
* Effects that allow you to cast/play stuff from zones but without setting a cost, like `Augur of Autumn`
* Effects that combine both, but for these, the other effects can't be used too like `Bolas's Citadel`
* The Goal also was to reduce the use of Continuous for LKI stuff

i think my first attempt to split was 3+ years ago:
https://github.com/Card-Forge/forge/compare/mayPlayAlternativeCost

Hanmac:
i think in the current state,
the best thing would be if we are doing it in a three step process:

* get Base Abilities, (including Modal)
* get possible Stack Keyword from them (this needs to apply possible Modal States for LKI)
* and then apply MayPlay stuff (this should notice that it can't apply MayPlay WithoutCost stuff on Stack Keywords)

Hanmac:
@tool4ever to fix the problems addressed there, we probably still need to refactor MayPlay stuff

tool4ever:
Gonna close this for now since the logic got shuffled around and improved a decent amount since this started
But feel free to revisit this if your time allows

PR #4708, opened 2024-02-17
MKC: Duskana, the Rage Mother + 6 cards

MikeS-NZ:
MKC: Add card scripts for seven MKC cards:

- Duskana, the Rage Mother
- Experiment Twelve
- Printlifter Ooze
- Showstopping Surprise
- True Identity
- Unexplained Absence
- Veiled Ascension

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/experiment_twelve.txt
+++ b/forge-gui/res/cardsfolder/upcoming/experiment_twelve.txt
@@ -1,0 +1,11 @@
+Name:Experiment Twelve
+ManaCost:3 G
+Types:Creature Elf Lizard Warrior
+PT:4/4
+K:Trample
+K:Disguise:4 G
+T:Mode$ TurnFaceUp | ValidCard$ Card.Self | Execute$ TrigPutCounters | TriggerZones$ Battlefield | TriggerDescription$ Whenever CARDNAME or another creature you control is turned face up, put +1/+1 counters on that creature equal to its power.
+T:Mode$ TurnFaceUp | ValidCard$ Creature.Other+YouCtrl | Execute$ TrigPutCounters | TriggerZones$ Battlefield | Secondary$ True | TriggerDescription$ Whenever CARDNAME or another creature you control is turned face up, put +1/+1 counters on that creature equal to its power.
+SVar:TrigPutCounters:DB$ PutCounter | Defined$ TriggeredCard | CounterType$ P1P1 | CounterNum$ X
+SVar:X:TriggeredCard$CardPower
+Oracle:Trample\nWhenever Experiment Twelve or another creature you control is turned face up, put +1/+1 counters on that creature equal to its power.\nDisguise {4}{G} (You may cast this card face down for {3} as a 2/2 creature with ward {2}. Turn it face up any time for its disguise cost.)
--- a/forge-gui/res/cardsfolder/upcoming/printlifter_ooze.txt
+++ b/forge-gui/res/cardsfolder/upcoming/printlifter_ooze.txt
@@ -1,0 +1,10 @@
+Name:Printlifter Ooze
+ManaCost:1 G
+Types:Creature Ooze
+PT:2/2
+K:Deathtouch
+K:Disguise:3 G
+T:Mode$ TurnFaceUp | ValidCard$ Card.Self | Execute$ TrigToken | TriggerZones$ Battlefield | TriggerDescription$ Whenever CARDNAME or another creature you control is turned face up, create a 0/0 green Ooze creature token with trample. The token enters the battlefield with X +1/+1 counters on it, where X is the number of other creatures you control.
+SVar:TrigToken:DB$ Token | TokenScript$ g_0_0_ooze_trample | WithCountersType$ P1P1 | WithCountersAmount$ X
+SVar:X:Count$Valid Creature.YouCtrl
+Oracle:Deathtouch\nWhenever Printlifter Ooze or another creature you control is turned face up, create a 0/0 green Ooze creature token with trample. The token enters the battlefield with X +1/+1 counters on it, where X is the number of other creatures you control.\nDisguise {3}{G}
--- a/forge-gui/res/cardsfolder/upcoming/showstopping_surprise.txt
+++ b/forge-gui/res/cardsfolder/upcoming/showstopping_surprise.txt
@@ -1,0 +1,8 @@
+Name:Showstopping Surprise
+ManaCost:3 R R
+Types:Instant
+A:SP$ Pump | Cost$ 3 R R | ValidTgts$ Creature.YouCtrl | AILogic$ PowerDmg | TgtPrompt$ Select target creature you control | RememberObjects$ Targeted | SubAbility$ DBTurnFaceUp | StackDescription$ None | SpellDescription$ Choose target creature you control. Turn it face up if it’s face down. Then it deals damage equal to its power to each other creature.
+SVar:DBTurnFaceUp:DB$ SetState | Defined$ Remembered | Mode$ TurnFaceUp | SubAbility$ IgnitionDamage
+SVar:IgnitionDamage:DB$ DamageAll | ValidCards$ Creature.IsNotRemembered | NumDmg$ X | ConditionDefined$ Targeted | ConditionPresent$ Creature | ConditionCompare$ EQ1 | DamageSource$ ParentTarget
+SVar:X:ParentTargeted$CardPower
+Oracle:Choose target creature you control. Turn it face up if it’s face down. Then it deals damage equal to its power to each other creature.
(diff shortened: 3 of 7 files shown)


forge-gui/res/cardsfolder/upcoming/experiment_twelve.txt
```diff
@@ -0,0 +1,11 @@
+Name:Experiment Twelve
+ManaCost:3 G
+Types:Creature Elf Lizard Warrior
+PT:4/4
+K:Trample
+K:Disguise:4 G
+T:Mode$ TurnFaceUp | ValidCard$ Card.Self | Execute$ TrigPutCounters | TriggerZones$ Battlefield | TriggerDescription$ Whenever CARDNAME or another creature you control is turned face up, put +1/+1 counters on that creature equal to its power.
```
tool4ever, line 7: no reason to split into two trigger

forge-gui/res/cardsfolder/upcoming/printlifter_ooze.txt
```diff
@@ -0,0 +1,10 @@
+Name:Printlifter Ooze
+ManaCost:1 G
+Types:Creature Ooze
+PT:2/2
+K:Deathtouch
+K:Disguise:3 G
+T:Mode$ TurnFaceUp | ValidCard$ Card.Self | Execute$ TrigToken | TriggerZones$ Battlefield | TriggerDescription$ Whenever CARDNAME or another creature you control is turned face up, create a 0/0 green Ooze creature token with trample. The token enters the battlefield with X +1/+1 counters on it, where X is the number of other creatures you control.
```
tool4ever, line 7: wrong valid

forge-gui/res/cardsfolder/upcoming/showstopping_surprise.txt
```diff
@@ -0,0 +1,8 @@
+Name:Showstopping Surprise
+ManaCost:3 R R
+Types:Instant
+A:SP$ Pump | Cost$ 3 R R | ValidTgts$ Creature.YouCtrl | AILogic$ PowerDmg | TgtPrompt$ Select target creature you control | RememberObjects$ Targeted | SubAbility$ DBTurnFaceUp | StackDescription$ None | SpellDescription$ Choose target creature you control. Turn it face up if it’s face down. Then it deals damage equal to its power to each other creature.
```
tool4ever, line 4: what's the point of the extra Pump?

forge-gui/res/cardsfolder/upcoming/showstopping_surprise.txt
```diff
@@ -0,0 +1,8 @@
...
+SVar:DBTurnFaceUp:DB$ SetState | Defined$ Remembered | Mode$ TurnFaceUp | SubAbility$ IgnitionDamage
+SVar:IgnitionDamage:DB$ DamageAll | ValidCards$ Creature.IsNotRemembered | NumDmg$ X | ConditionDefined$ Targeted | ConditionPresent$ Creature | ConditionCompare$ EQ1 | DamageSource$ ParentTarget
```
tool4ever, line 6: cleanup missing

forge-gui/res/cardsfolder/upcoming/true_identity.txt
```diff
@@ -0,0 +1,8 @@
+Name:True Identity
+ManaCost:1 W
+Types:Enchantment
+K:Disguise:W
+T:Mode$ TurnFaceUp | ValidCard$ Permanent.YouCtrl | Execute$ TrigScry | TriggerZones$ Battlefield | ActivationLimit$ 1 | TriggerDescription$ Whenever CARDNAME or another permanent you control is turned face up, scry 1, then draw a card. This ability triggers only once each turn.
```
tool4ever, line 5: wrong valid

forge-gui/res/cardsfolder/upcoming/unexplained_absence.txt
```diff
@@ -0,0 +1,9 @@
+Name:Unexplained Absence
+ManaCost:3 W
+Types:Instant
+A:SP$ ChangeZone | Origin$ Battlefield | Destination$ Exile | ValidTgts$ Permanent.nonLand | TgtPrompt$ Select up to one target nonland permanent each player controls | TargetMin$ 0 | TargetMax$ OneEach | TargetsWithDifferentControllers$ True | SubAbility$ LoopTgts | StackDescription$ SpellDescription | SpellDescription$ For each player, exile up to one target nonland permanent that player controls. For each permanent exiled this way, its controller cloaks the top card of their library.
+SVar:LoopTgts:DB$ RepeatEach | RepeatSubAbility$ CloakEach | DefinedCards$ Targeted
```
tool4ever, line 5: why the extra RepeatEach?

forge-gui/res/cardsfolder/upcoming/experiment_twelve.txt
MikeS-NZ: Cleaned up!

forge-gui/res/cardsfolder/upcoming/printlifter_ooze.txt
MikeS-NZ: Fixed!

forge-gui/res/cardsfolder/upcoming/showstopping_surprise.txt
MikeS-NZ: This is the same pump method used in the script for Chandra's Ignition. There may be a more concise way to do this, but thought it might be a more secure way given the creature must also be turned face-up if face down before damage is dealt.

forge-gui/res/cardsfolder/upcoming/showstopping_surprise.txt
MikeS-NZ: Added missing cleanup

forge-gui/res/cardsfolder/upcoming/true_identity.txt
MikeS-NZ: Fixed the valid tag

forge-gui/res/cardsfolder/upcoming/unexplained_absence.txt
MikeS-NZ: This is borrowed from Luminate Primordial's script, just makes sure only the players who had a permanent exiled this way can cloak a card from library. Couldn't find a more modern example with this effect which is more concise

forge-gui/res/cardsfolder/upcoming/showstopping_surprise.txt
tool4ever: _Chandra's Ignition_ is a bit different, with the Pump acting as target helper (and it still doesn't need remembered)
Here you can just target with the SetState effect

forge-gui/res/cardsfolder/upcoming/unexplained_absence.txt
tool4ever: - _Luminate Primordial_ probably needs it because the amount for each player is variable
- in the current version you wouldn't even need the cleanup due to internal handling by RepeatEach
- However you should use RememberLKI$ since it's explicitly stating "exiled this way"
- The CloackEach line will work fine by itself then

forge-gui/res/cardsfolder/upcoming/showstopping_surprise.txt
MikeS-NZ: Okay, I am trying it with the opening SetState effect and without any Remembered$ functionality. I'm struggling with the right `Validcards$` tag for the damage part. There appears to be no such parameter that means "all creature this spell doesn't target." I've seen `Creature.IsNotRemembered' for Chandra's Ignition, but I'm looking for something like `Creature.IsNotTargeted` or `Creature.nonTargeted` here. I couldn't find any scripts with a searchable implementation of this. Any thoughts?

forge-gui/res/cardsfolder/upcoming/unexplained_absence.txt
MikeS-NZ: Removed the RepeatEach and added RememberLKI to the script, works fine! Thank you

forge-gui/res/cardsfolder/upcoming/showstopping_surprise.txt
tool4ever: the script you mentioned already uses `NotDefinedTargeted`

MikeS-NZ: I figured out why that use never appeared for me -- an old .rar card script repository used a different script for Ignition. Your suggested way works just fine in the newer snapshot. Thank you for your help and patience!

tool4ever (approved):

PR #4714, opened 2024-02-19
Card: add surveilledThisTurn

Hanmac:
for Eye of Duskmantle

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -239,6 +239,8 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
     private boolean tributed = false;
     private boolean discarded = false;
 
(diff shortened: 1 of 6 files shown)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -6893,6 +6895,7 @@ public void onCleanupPhase(final Player turn) {
         resetShieldCount();
         setBecameTargetThisTurn(false);
         setFoughtThisTurn(false);
+        setSurveilledThisTurn(false);
```
tool4ever, line 6898: currently this gets only called for battlefield cards
I guess we have to include graveyards now too? :/

Hanmac: probably, i will do that later too

Agetian (approved):

PR #4735, opened 2024-02-24
Support for The Master, Transcendent

tool4ever:


forge-game/src/main/java/forge/game/card/CardUtil.java
```diff
@@ -268,8 +268,8 @@ public static Card getLKICopy(final Card in, Map<Integer, Card> cachedMap) {
         newCopy.setMonstrous(in.isMonstrous());
         newCopy.setRenowned(in.isRenowned());
         newCopy.setSolved(in.isSolved());
-        newCopy.setSuspectedTimestamp(in.getSuspectedTimestamp());
         newCopy.setSuspected(in.isSuspected());
+        newCopy.setSuspectedTimestamp(in.getSuspectedTimestamp());
```
Hanmac, line 272: @tool4ever why exactly did you do this?

the setSuspected() function explicit was looking to the timestamp to not mess up the created properties

it might not matter because they are overwritten later anyway, but like this, there is unnecessary keywords generated

forge-gui/res/cardsfolder/e/evolving_adaptive.txt
```diff
@@ -4,10 +4,8 @@ Types:Creature Phyrexian Warrior
 PT:0/0
 K:etbCounter:OIL:1
 S:Mode$ Continuous | Affected$ Card.Self | AddPower$ X | AddToughness$ X | Description$ CARDNAME gets +1/+1 for each oil counter on it.
-T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Creature.YouCtrl+powerGTY+Other,Creature.YouCtrl+toughnessGTZ+Other | TriggerZones$ Battlefield | Execute$ TrigPutCounter | TriggerDescription$ Whenever another creature enters the battlefield under your control, if that creature has greater power or toughness than CARDNAME, put an oil counter on CARDNAME.
-SVar:TrigPutCounter:DB$ PutCounter | ConditionDefined$ TriggeredCard | ConditionPresent$ Creature.powerGTY+Other,Creature.toughnessGTZ+Other | CounterType$ OIL
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Creature.YouCtrl+Other | Condition$ Evolve | TriggerZones$ Battlefield | Execute$ TrigPutCounter | TriggerDescription$ Whenever another creature enters the battlefield under your control, if that creature has greater power or toughness than CARDNAME, put an oil counter on CARDNAME.
```
Hanmac, line 7: while it might be similar, technically its not evolve :/

tool4ever: I know, but this avoids the most extra code without affecting anything else in the engine
We've been doing this for some others too, e. g. for Revolt on _Lulu, Loyal Hollyphant_

PR #4742, opened 2024-02-25
fixed performance with OpenGL acceleration

thedevnull:
Fixed performance on Linux using OpenGL acceleration.


tehdiplomat (approved):

tehdiplomat:
Seems fine. I think thats early enough where it will get set before we care about this setting.

Hanmac (approved):

tehdiplomat:
We've been getting a lot of people reporting crashes after this merge. I'm going to roll it back for now. I think we can add this as a flag for people who want it.

PR #4744, opened 2024-02-26
Fix Crawling Sensation + related tweaks

tool4ever:
Unlike its newer cousin (_Crawling Infestation_) the trigger on _Crawling Sensation_ should see the past with "for the first time" - so perhaps might never be able to trigger on the turn it ETB.

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityKey.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityKey.java
@@ -4,6 +4,7 @@ import forge.game.GameEntity;
 import forge.game.card.Card;
 import forge.game.card.CardZoneTable;
 import forge.game.player.Player;
+import forge.game.spellability.SpellAbility;
 
 import java.util.EnumMap;
 import java.util.Map;
@@ -210,7 +211,7 @@ public enum AbilityKey {
         map.put(AbilityKey.LastStateGraveyard, table.getLastStateGraveyard());
         map.put(AbilityKey.InternalTriggerTable, table);
     }
-    public static CardZoneTable addCardZoneTableParams(Map<AbilityKey, Object> map, forge.game.spellability.SpellAbility sa) {
+    public static CardZoneTable addCardZoneTableParams(Map<AbilityKey, Object> map, SpellAbility sa) {
         CardZoneTable table = new CardZoneTable(sa.getLastStateBattlefield(), sa.getLastStateGraveyard());
         addCardZoneTableParams(map, table);
         return table;
--- a/forge-game/src/main/java/forge/game/ability/AbilityKey.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityKey.java
@@ -1,5 +1,6 @@
 package forge.game.ability;
 
+import forge.game.Game;
 import forge.game.GameEntity;
 import forge.game.card.Card;
 import forge.game.card.CardZoneTable;
@@ -213,6 +214,11 @@ public enum AbilityKey {
     }
     public static CardZoneTable addCardZoneTableParams(Map<AbilityKey, Object> map, SpellAbility sa) {
         CardZoneTable table = new CardZoneTable(sa.getLastStateBattlefield(), sa.getLastStateGraveyard());
+        addCardZoneTableParams(map, table);
+        return table;
+    }
+    public static CardZoneTable addCardZoneTableParams(Map<AbilityKey, Object> map, Game game) {
+        CardZoneTable table = new CardZoneTable(game.getLastStateBattlefield(), game.getLastStateGraveyard());
         addCardZoneTableParams(map, table);
         return table;
     }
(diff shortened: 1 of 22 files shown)


forge-game/src/main/java/forge/game/trigger/TriggerChangesZoneAll.java
```diff
@@ -31,6 +33,19 @@ public boolean performTest(Map<AbilityKey, Object> runParams) {
...
+                lastState = table.getLastStateBattlefield();
+            } else if ("Graveyard".equals(getParam("Destination"))) {
+                lastState = table.getLastStateGraveyard();
+            }
+            // currently only for Crawling Sensation
+            if (CardLists.getValidCardCount(lastState, getParam("ValidCards") + "+ThisTurnEntered", getHostCard().getController(), getHostCard(), this) > 0) {
+                return false;
+            }
+        }
+
```
Hanmac, line 48: i don't think LastStateGraveyard words there as you think

for example, what if i remove all lands that was sent there
then play Crawling Sensation
and then mill new lands?

because LastStateGraveyard is empty, this condition is successful

Hanmac: Instead, we probably need to check for:

```
list = getTurnEntered("Graveyard", "Any")
list.removeAll(table.cards)
list.isEmpty()
```

tool4ever: yea good idea
seems to work

forge-game/src/main/java/forge/game/card/CardZoneTable.java
```diff
@@ -121,9 +121,13 @@ public CardCollection filterCards(Iterable<ZoneType> origin, ZoneType destinatio
         if (origin != null) {
             for (ZoneType z : origin) {
                 CardCollectionView lkiLookup = CardCollection.EMPTY;
+                // CR 603.10a
                 if (z == ZoneType.Battlefield) {
                     lkiLookup = lastStateBattlefield;
                 }
+                if (z == ZoneType.Graveyard && destination == null) {
```
tool4ever, line 128: _Desecrated Tomb_ is now able to trigger when you _Unearth_ our boy _Grist_

Hanmac (dismissed):

forge-game/src/main/java/forge/game/ability/AbilityKey.java
```diff
@@ -210,7 +211,7 @@ public static void addCardZoneTableParams(Map<AbilityKey, Object> map, CardZoneT
         map.put(AbilityKey.LastStateGraveyard, table.getLastStateGraveyard());
         map.put(AbilityKey.InternalTriggerTable, table);
     }
-    public static CardZoneTable addCardZoneTableParams(Map<AbilityKey, Object> map, forge.game.spellability.SpellAbility sa) {
+    public static CardZoneTable addCardZoneTableParams(Map<AbilityKey, Object> map, SpellAbility sa) {
```
Hanmac, line 214: i don't know if that works

my IDE didn't wanted to import that, because the AbilityKey has a Key for SpellAbility already

forge-game/src/main/java/forge/game/ability/effects/ConniveEffect.java
```diff
@@ -68,7 +68,7 @@ public void resolve(SpellAbility sa) {
             CardCollection connivers = CardLists.filterControlledBy(toConnive, p);
             while (!connivers.isEmpty()) {
                 GameEntityCounterTable table = new GameEntityCounterTable();
-                final CardZoneTable triggerList = new CardZoneTable(sa.getLastStateBattlefield(), sa.getLastStateGraveyard());
+                final CardZoneTable triggerList = new CardZoneTable(game.getLastStateBattlefield(), game.getLastStateGraveyard());
```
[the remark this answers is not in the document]
tool4ever: The SpellAbility one is only really filled when
a) it's the castSA for stuff like "as you cast" (Presence)
b) we're inside a replacement ability (which can then be used inside AbilityUtils valid checks)

so most of the time it's unavailable for normal LKI usage inside the table and we should move away from using it in effects? 🤔 
And I suppose there can be scenarios where we need to use `game.copyLastState` because the previous one might be stale?

forge-game/src/main/java/forge/game/ability/effects/CounterEffect.java
```diff
@@ -253,6 +253,8 @@ private static boolean removeFromStack(final SpellAbility tgtSA, final SpellAbil
 
         params.put(AbilityKey.StackSa, tgtSA);
         params.put(AbilityKey.StackSi, si);
+        params.put(AbilityKey.LastStateBattlefield, game.getLastStateBattlefield());
+        params.put(AbilityKey.LastStateGraveyard, game.getLastStateGraveyard());
```
Hanmac, line 257: better use a CardZoneTable there?

The Problem lies in Line 55:
https://github.com/Card-Forge/forge/blob/a7d77028e900371769c3a3a199d0a0ff4002bc81/forge-game/src/main/java/forge/game/ability/effects/CounterEffect.java#L55-L56

at this point the SpellAbility doesn't have a LastState so it should fallback to game?

i don't know anymore if AbilityKey.addCardZoneTableParams was the right place for this or if that should have been done in SpellAbilityEffect 
but there are corner cases where you would need the Table, but don't have a SpellAbility

tool4ever: ok, I'll try to refine the general usage

tool4ever:
Since this PR is getting big we could merge these fixes now and I'll do more cleanup in a follow-up

PR #4751, opened 2024-02-27
Consolidate CardZoneTable logic

### LINKED ISSUE #2076: GameAction: investigate where LastState is still missing
Exception handling is expensive and can hide real bugs.

tool4ever:
Fixes Meld not triggering "one or more ETB"

Closes #2076

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -131,31 +131,19 @@ public class GameAction {
         // need to check before it enters
         if (c.isAura() && !c.isAttachedToEntity() && toBattlefield && (zoneFrom == null || !zoneFrom.is(ZoneType.Stack))) {
             boolean found = false;
-            try {
-                if (Iterables.any(game.getPlayers(), PlayerPredicates.canBeAttached(c, null))) {
+            if (Iterables.any(game.getPlayers(), PlayerPredicates.canBeAttached(c, null))) {
+                found = true;
+            }
+
+            if (!found) {
+                if (Iterables.any((CardCollectionView) params.get(AbilityKey.LastStateBattlefield), CardPredicates.canBeAttached(c, null))) {
                     found = true;
                 }
-            } catch (Exception e1) {
-                found = false;
             }
 
             if (!found) {
-                try {
-                    if (Iterables.any((CardCollectionView) params.get(AbilityKey.LastStateBattlefield), CardPredicates.canBeAttached(c, null))) {
-                        found = true;
-                    }
-                } catch (Exception e2) {
-                    found = false;
-                }
-            }
-
-            if (!found) {
-                try {
-                    if (Iterables.any((CardCollectionView) params.get(AbilityKey.LastStateGraveyard), CardPredicates.canBeAttached(c, null))) {
-                        found = true;
-                    }
-                } catch (Exception e3) {
-                    found = false;
+                if (Iterables.any((CardCollectionView) params.get(AbilityKey.LastStateGraveyard), CardPredicates.canBeAttached(c, null))) {
+                    found = true;
                 }
             }
             if (!found) {
@@ -184,8 +172,8 @@ public class GameAction {
                 lastKnownInfo = (Card) cause.getReplacingObject(AbilityKey.CardLKI);
             }
         }
-        CardCollectionView lastBattlefield = getLastState(AbilityKey.LastStateBattlefield, cause, params);
-        CardCollectionView lastGraveyard = getLastState(AbilityKey.LastStateGraveyard, cause, params);
+        CardCollectionView lastBattlefield = getLastState(AbilityKey.LastStateBattlefield, cause, params, false);
+        CardCollectionView lastGraveyard = getLastState(AbilityKey.LastStateGraveyard, cause, params, false);
 
         if (c.isSplitCard()) {
             boolean resetToOriginal = false;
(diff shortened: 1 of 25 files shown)


forge-game/src/main/java/forge/game/ability/effects/SacrificeAllEffect.java
```diff
@@ -84,8 +84,7 @@ public void resolve(SpellAbility sa) {
 
         Map<Integer, Card> cachedMap = Maps.newHashMap();
         Map<AbilityKey, Object> params = AbilityKey.newMap();
-        CardZoneTable table = new CardZoneTable(game.copyLastStateBattlefield(), CardCollection.EMPTY);
-        AbilityKey.addCardZoneTableParams(params, table);
+        CardZoneTable zoneMovements = AbilityKey.addCardZoneTableParams(params, sa);
 
         for (Card sac : list) {
             final Card lKICopy = CardUtil.getLKICopy(sac, cachedMap);
```
tool4ever, line 90: @Hanmac 
You added the `cachedMap` in some places, but I don't see the point here:
these are different cards so LKI wouldn't get reused anyway?

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -131,31 +131,19 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
...
             }
 
             if (!found) {
-                try {
-                    if (Iterables.any((CardCollectionView) params.get(AbilityKey.LastStateBattlefield), CardPredicates.canBeAttached(c, null))) {
-                        found = true;
-                    }
-                } catch (Exception e2) {
-                    found = false;
+                if (Iterables.any((CardCollectionView) params.get(AbilityKey.LastStateBattlefield), CardPredicates.canBeAttached(c, null))) {
```
tool4ever, line 139: maybe we can/should just reuse the `lastStateBattlefield` from above here?
what would a good test for this be 🤔

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -2608,12 +2596,13 @@ private static void unanimateOnAbortedChange(final SpellAbility cause, final Car
...
-    private CardCollectionView getLastState(final AbilityKey key, final SpellAbility cause, final Map<AbilityKey, Object> params) {
+    public CardCollectionView getLastState(final AbilityKey key, final SpellAbility cause, final Map<AbilityKey, Object> params, final boolean refreshIfEmpty) {
         CardCollectionView lastState = null;
         if (params != null) {
             lastState = (CardCollectionView) params.get(key);
         }
         if (lastState == null && cause != null) {
+            // inside RE
             if (key == AbilityKey.LastStateBattlefield) {
                 lastState = cause.getLastStateBattlefield();
```
tool4ever, line 2607: I'm still not totally convinced this should still be included here:
like if a RE does:
- move X from A to B
- and then **after** moves Y

the second part should still have a newer lastState from its perspective 🤔 

But I would like to find an example (ideally one where it matters) before I commit further

tool4ever: Hmn, I checked **Dredge** and it wasn't a problem since it's only set for the root anyway
Will probably leave it alone for now...

forge-game/src/main/java/forge/game/ability/effects/SacrificeAllEffect.java
Hanmac: just for the corner case when you sacrifice multiple cards at once
that may be attached to each other

the cachedMap is just for when getting an LKI copy, to not copy something that already got copied.

otherwise we might look into this using LKI that already got copied in LastStateBattlefield?

forge-game/src/main/java/forge/game/GameAction.java
Hanmac: yeah, better reuse the lastStateBattlefield that already get set

your current change with removing the try-catch would otherwise crash on possible null-pointer?

forge-game/src/main/java/forge/game/ability/effects/SacrificeAllEffect.java
Hanmac: i didn't mean it like that

more like:
`final Card lKICopy = zoneMovements.getLastStateBattlefield().get(sac);`

this way, we can use the LKI that was already generated for LastStateBattlefield?

forge-game/src/main/java/forge/game/GameAction.java
tool4ever: there used to be an old crash, which I failed to reproduce in newer versions
I think at this point all relevant calls already provided the lastState
And if they didn't there could be a bug hidden because we might miss out on a legal attach
Anyway with the additional change it should now be impossible

forge-game/src/main/java/forge/game/ability/effects/SacrificeAllEffect.java
tool4ever: that's how it will work?
`getLKICopy` checks the prefilled cache first
this way there's some safety if for some reason the LastState doesn't have it and just creates a new one instead

Hanmac: i mean it shouldn't even be possible

you can only Sacrifice from the Battlefield, and at that moment, LastState should already have the card

so there is no need to make an extra CacheMap for this, right?

tool4ever: In an ideal world yes?
Seems stable without, will run some longer AI tests as backup

Hanmac (approved):
looks good for me

PR #4763, opened 2024-03-01
Alchemy Incorporate

Northmoc:


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4506,7 +4511,6 @@ public final void removePerpetual(final long timestamp) {
         for (Map<String, Object> p : perpetual) {
             if (p.get("Timestamp").equals(timestamp)) {
                 toRemove = p;
-                break;
```
tool4ever, line 4509: why remove optimization?

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4496,8 +4500,9 @@ public final void executePerpetual(Map<String, Object> p) {
                 false, (Set<RemoveType>) p.get("RemoveXTypes"),
                 (long) p.get("Timestamp"), (long) 0, true, false);
         } else if (category.equals("Colors")) {
-            addColor((ColorSet) p.get("Colors"), !(boolean) p.get("Overwrite"), (long) p.get("Timestamp"),
-                (long) 0, false);
+            ColorSet colors = (ColorSet) p.get("Colors");
+            addColor(colors, !(boolean) p.get("Overwrite"), (long) p.get("Timestamp"), (long) 0, false);
+            getCurrentState().addColor(colors.getColor());
```
tool4ever, line 4505: why do we need to hard set the states now?
that seems wrong

Northmoc: I was encountering an issue whereupon being cast, cards would not display as their updated colors anymore.

So after the Animate, everything looked fine while the card was in the hand. But then upon cast, there was a misalignment between the colors displayed in Detail Pane / "Current Card Colors" (also the color coding of the spell on the stack) and the actual card item when checked in the debugger. So in the debugger the card object did have the new color identity... it just wasn't displaying properly in the GUI.

Is there a less aggressive fix?

forge-game/src/main/java/forge/game/card/Card.java
Northmoc: Some Perpetual effects have multiple entries with the same timestamp. This card for example has "Incorporate" and "Colors" categories with the same timestamp.

As I think about it more, there are further problems as Guildpact Greenwarden here has "Incorporate" and "Colors" with one timestamp, but "BoostPT" with another timestamp since that's in PumpEffect 😦

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
```diff
@@ -81,6 +82,27 @@ public static void doAnimate(final Card c, final SpellAbility sa, final Integer
...
+            c.addPerpetual(params);
+            
+            final ManaCost incMCost = new ManaCost(new ManaCostParser(incorporate));
+            incColors = ColorSet.fromMask(incMCost.getColorProfile());
+            final ManaCost newCost = ManaCost.combine(c.getManaCost(), incMCost);
+            c.addChangedManaCost(newCost, timestamp, 0);
+            c.setManaCost(c.getManaCost());
+            c.updateManaCostForView();
+
+            c.getFirstSpellAbility().getPayCosts().add(new Cost(incorporate, false));
```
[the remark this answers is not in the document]
Northmoc: Good catch

Hmm

forge-game/src/main/java/forge/game/card/Card.java
tool4ever: well I think when you implemented this function it was for a very specific card?
if the code is supposed to handle multiple entries it would need updating anyway

forge-game/src/main/java/forge/game/card/Card.java
tool4ever: if it's a GUI problem only then the logic where the view gets refreshed probably needs updating
chances are its just using the state directly because it really doesn't expect what you're doing there - since the implementation you're "abusing" was really only meant for _Volrath's Shapeshifter_

Northmoc: I think I've dodged this now

tool4ever (approved):

Hanmac:
@Northmoc it seems we need to touch this again

while it does affect the color of the card,
it does not set the mana cost directly,
but more like it that its adds additional Costs

Northmoc:
> @Northmoc it seems we need to touch this again
> 
> while it does affect the color of the card, it does not set the mana cost directly, but more like it that its adds additional Costs

Interesting – that might actually be easier

PR #4778, opened 2024-03-03
Fix River Song's Diary + Rhuk, Hexgold Nabber

tool4ever:


forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -125,7 +130,8 @@ else if (defined.equals("Enchanted")) {
                     c = sacrificed.getFirst().getEnchantingCard();
                 }
             }
-        } else if (defined.equals("TopOfGraveyard")) {
+        }
+        else if (defined.equals("TopOfGraveyard")) {
```
Hanmac, line 134: isn't the other more readable?

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -3200,6 +3206,11 @@ public static int doXMath(final int num, final String operators, final Card c, C
             return num * -1;
         } else if (s[0].contains("Times")) {
             return num * secondaryNum;
+        } else if (s[0].contains("DivideEvenlyUp")) {
+            if (secondaryNum == 0) {
+                return 0;
+            }
+            return num / secondaryNum + (num % secondaryNum == 0 ? 0 : 1);
```
Hanmac, line 3213: use Math.ceil?

`(int)Math.ceil((double)num / secondaryNum);`

tool4ever: There are claims that it can lead to bad results:
https://stackoverflow.com/questions/7139382/java-rounding-up-to-an-int-using-math-ceil

Agetian (approved):

Hanmac (approved):

PR #4794, opened 2024-03-06
Undo restore snapshot

tehdiplomat:
Stashing a snapshot when putting a SpellAbility on the stack which then we're able to restore the game state to the snapshot. Since we don't need to undo individual items, things like discarding and sacrificing are now fully undoable.


forge-ai/src/main/java/forge/ai/simulation/SpellAbilityPicker.java
```diff
@@ -90,7 +90,8 @@ public List<SpellAbility> getCandidateSpellsAndAbilities() {
     }
 
     public SpellAbility chooseSpellAbilityToPlay(SimulationController controller) {
-        printOutput = controller == null;
+        //printOutput = controller == null;
+        printOutput = true;
```
tehdiplomat, line 94: We should disable this so we don't print so much during testing

Hanmac:
@tehdiplomat can we talk about all the calls to getLKICopy now shown as deprecated?

into what form should they be updated?

forge-ai/src/main/java/forge/ai/simulation/GameSimulator.java
```diff
@@ -33,7 +33,8 @@ public class GameSimulator {
     public GameSimulator(SimulationController controller, Game origGame, Player origAiPlayer, PhaseType advanceToPhase) {
         this.controller = controller;
         copier = new GameCopier(origGame);
-        simGame = copier.makeCopy(advanceToPhase, origAiPlayer);
+        simGame = copier.makeCopy();
+        // simGame.advanceToPhase(advanceToPhase, origAiPlayer)
```
tool4ever, line 37: Don't understand this change
it's important for simulation AI to look at future combat

forge-gui/src/main/java/forge/player/HumanPlaySpellAbility.java
```diff
@@ -162,19 +162,33 @@ public final boolean playAbility(final boolean mayChooseTargets, final boolean i
...
-                && ability.checkRestrictions(human)
-                && (!mayChooseTargets || ability.setupTargets()) // if you can choose targets, then do choose them.
-                && ability.canCastTiming(human)
-                && ability.isLegalAfterStack()
-                && (isFree || payment.payCost(new HumanCostDecision(controller, human, ability, ability.isTrigger())));
+
+        // We need to split up this super conditional to know WHY somethings prequisites failed.
+        // Failing because there's no legal targets is different than failing because the player canceled paying costs
+
+        final boolean announcePrequiste = announceType() && announceValuesLikeX();
```
tool4ever, line 169: that's more expensive now and also will still let you choose targets even if you already canceled by announcing?
this should cascade in a cleaner way

forge-game/src/main/java/forge/game/phase/PhaseHandler.java
```diff
@@ -1050,12 +1051,18 @@ public void startFirstTurn(Player goesFirst, Runnable startGameHook) {
...
                         System.out.print("... " + pPlayerPriority + " plays " + chosenSa);
                     }
+
+                    boolean rollback = false;
                     for (SpellAbility sa : chosenSa) {
                         Card saHost = sa.getHostCard();
                         final Zone originZone = saHost.getZone();
 
-                        if (pPlayerPriority.getController().playChosenSpellAbility(sa)) {
+                        if (pPlayerPriority.getController().playChosenSpellAbility(sa) || sa instanceof LandAbility) {
```
tool4ever, line 1060: this seems like a weird addition
might be worth investigating if HumanPlay should really always return false instead

forge-game/src/main/java/forge/game/zone/MagicStack.java
```diff
@@ -789,18 +800,23 @@ public boolean addAllTriggeredAbilitiesToStack() {
...
-            whoAddsToStack = game.getNextPlayerAfter(whoAddsToStack);
-        } while (whoAddsToStack != null && whoAddsToStack != playerTurn);
-        // 603.3b (Strict Proctor)
-        whoAddsToStack = playerTurn;
-        do {
-            result |= chooseOrderOfSimultaneousStackEntry(whoAddsToStack, true);
-            whoAddsToStack = game.getNextPlayerAfter(whoAddsToStack);
-        } while (whoAddsToStack != null && whoAddsToStack != playerTurn);
+        // Grab players in turn order starting with the active player
+        List<Player> players = game.getPlayersInTurnOrder(playerTurn);
```
tool4ever, line 804: Looks like the logic there is wrong?
you need to shift them in the other direction instead

forge-ai/src/main/java/forge/ai/simulation/GameSimulator.java
tehdiplomat: Oh i commented this out for other reasons, but should be able to add it back in. I don't think we had any tests that covered this path so it was hard to see where we used it. 

I don't think its related to the disabled test though. That seems to be a problem with the simulation assuming cards would be ordered in the same way from the original game to the sim games. i.e. Why I had to change all of the other loops from index to findById(). The AI seems to be trying to target the opposing creature in that circumstance and I couldn't figure out where exactly that wrong decision tree was occurring.

forge-gui/src/main/java/forge/player/HumanPlaySpellAbility.java
tehdiplomat: Fair enough. I think i can flatten most of these again. At the time i was trying to figure out where the rollbacks were failing.

forge-game/src/main/java/forge/game/GameSnapshot.java
```diff
@@ -0,0 +1,511 @@
...
+            newCard.setManifested(fromCard.isManifested());
+            newCard.setSickness(fromCard.hasSickness());
+            newCard.setState(fromCard.getCurrentStateName(), false);
+        }
+
+        // This loop happens later to make sure all cards are in the correct zone first
+        for (Card newCard : toGame.getCardsIn(ZoneType.Battlefield)) {
+            Card fromCard = fromGame.findById(newCard.getId());
+
+            if (fromCard.isAttachedToEntity()) {
```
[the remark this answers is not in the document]
tehdiplomat: Got a repro scenario? I haven't seen this one.

tehdiplomat:
@Hanmac I can undo marking LKI calls as deprecated for now. The idea was to try to merge all of these concepts into a single one, but I'm not there yet. Wanted to get people using the restore for a bit first.

forge-game/src/main/java/forge/game/GameSnapshot.java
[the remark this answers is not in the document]
tool4ever: I just tried to undo _Lithoform Engine_ payment 🤷‍♂️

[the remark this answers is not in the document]
tehdiplomat: Hmm.. not crashing for me... Some other weirdness, but no crash. What were you targeting?

forge-game/src/main/java/forge/game/zone/MagicStack.java
tehdiplomat: Why do you think the logic is wrong? We start at the current player, then iterate through?

tool4ever: see Javadoc for Collections.rotate:
> For example, suppose list comprises [t, a, n, k, s].After invoking Collections.rotate(list, 1) (or Collections.rotate(list, -4)), list will comprise [s, t, a, n, k]. 

you aren't ensuring APNAP

forge-game/src/main/java/forge/game/GameSnapshot.java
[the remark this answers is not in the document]
tool4ever: Just a good old Bear Cub 🤔 
Now that I try again however it just puts the ability on the stack even if I cancel mana payment
I'll test this further after this progresses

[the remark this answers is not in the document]
tehdiplomat: Yea that was the weirdness i was seeing. Which is weird, everything else had been rolling back just fine.

forge-game/src/main/java/forge/game/zone/MagicStack.java
tehdiplomat: Ah I see.. the rotate isn't doing what i expected it to do. I'll run some tests on it to make sure its rotating as expected.

forge-game/src/main/java/forge/game/GameSnapshot.java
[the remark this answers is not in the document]
tehdiplomat: Ooohhh its not adding it anyway. Its repopulating the stack incorrectly. Shouldn't be too bad to fix.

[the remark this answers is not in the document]
tehdiplomat: Ok @tool4ever I think this should be resolved. Manapool is fixed too, but I need to do some normalizing of the data since the mana is pointing to the wrong game underneath right now. I think things should work for the most part.

forge-gui/src/main/java/forge/player/HumanPlaySpellAbility.java
tool4ever: not handled yet:
https://stackoverflow.com/questions/3152208/do-and-short-circuit-in-java

tehdiplomat: I hate Java.

forge-game/src/main/java/forge/game/Game.java
```diff
@@ -368,7 +402,7 @@ public final PlayerCollection getPlayersInTurnOrder(Player p) {
         PlayerCollection players = getPlayersInTurnOrder();
 
         int i = players.indexOf(p);
-        Collections.rotate(players, i);
+        Collections.rotate(players, -i);
```
tool4ever, line 405: your failed test:

Question:
this implies the AI is making a pretty big misplay now, but why...?

Answer:
with default direction, you are **modifying the original order of the player slots**

This took me ages to find 💩 
(Luckily I noticed on your branch when loading gamestates the boards would get swapped around depending on which turns player you loaded it)

tehdiplomat: Oh sweet. Good catch! Hopefully that's the last thing causing issues, so we can get people experimenting with it.

forge-game/src/main/java/forge/game/combat/Combat.java
```diff
@@ -72,7 +72,7 @@ public Combat(final Player attacker) {
         initConstraints();
     }
 
-    public Combat(Combat combat, GameObjectMap map) {
+    public Combat(Combat combat, IEntityMap map) {
```
tehdiplomat, line 75: Need to do some more testing about rolling back combat. Just had some weirdness occur where nothing died when canceling a spell during combat.

tehdiplomat: Ok, the map was grabbing the wrong game when restoring. This should be fixed now.

tool4ever:
Will retest this weekend

forge-game/src/main/java/forge/game/combat/Combat.java
tool4ever: Can't confirm
To me it looks like canceling (say _Doubling Cube_) any ability during combat prevents damage from every creature involved...

forge-game/src/main/java/forge/game/GameSnapshot.java
```diff
@@ -0,0 +1,458 @@
...
+        restore = false;
+        assignGameState(origGame, newGame, includeStack);
+        //System.out.println("Storing game state with timestamp of :" + origGame.getTimestamp());
+
+        return newGame;
+    }
+
+    public void restoreGameState(Game currentGame) {
+        System.out.println("Restoring game state with timestamp of :" + newGame.getTimestamp());
+        restore = true;
```
tool4ever, line 58: Hmn, I think for now there should probably be some kind of rundimentary logic that does not allow the same rollback to happen twice?
This can happen quickly during payment chains, e. g.
tapping a basic to pay for mana ability and then trying to use _Ashnod's Altar_ for the rest but in the end deciding to cancel that one
-> now you're rolling back to the wrong state and the basic gets untapped

forge-game/src/main/java/forge/game/GameSnapshot.java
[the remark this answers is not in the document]
tool4ever: Maybe it's limited to GameStates?
But this scenario related to my other concern seems to always crash for me:
tap Doubling Cube -> tap Swamp -> Activate and then cancel the Altar

[Uploading undo_test.txt…]()

forge-game/src/main/java/forge/game/GameSnapshot.java
tehdiplomat: Do you just mean that canceling any part of a cost payment is rolling back the entire thing? That's intentional for now. I think we should be a bit more precise in the future, but I'd rather just rewind the whole thing and start over, rather than leave you half paid. Ideally, it would be a stack of costs, but when i was trying to do that it was causing problems with unfreezing the stack.

forge-game/src/main/java/forge/game/GameSnapshot.java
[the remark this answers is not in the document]
tehdiplomat: Yea ok, that was it. Now I'm stashing the snapshot after loading a game state. So this should be fine now.

forge-game/src/main/java/forge/game/GameSnapshot.java
tool4ever: Hmn well currently the first cancel on top of another rolls back everything and leaves the game in an illegal state
that it applies the same state a second time after probably doesn't hurt?
So while this make even unintentional "cheating" rather easier to occur, that's not a huge concern for experimental? 🤔

tool4ever (approved):

PR #4799, opened 2024-03-09
RadiationRework: use Trigger for InternalRadiation

### LINKED ISSUE #4756: Rad Counters/Radiation Effect Rework
from the Release Notes:
https://magic.wizards.com/en/news/feature/magic-the-gathering-fallout-release-notes
> There is an inherent triggered ability associated with having rad counters. This triggered ability has no source and is controlled by the active player. The full text of this ability is "At the beginning of the precombat main phase of a player with rad counters, that player mills cards equal to the number of rad counters they have. For each nonland card milled this way, that player loses 1 life and removes one rad counter from themselves."

* I don't know if we really need an RadiationEffect class, i think that logic can mostly moved to the Player class, and with PutCounter Effect
* I don't know either if the effect from the triggered ability should be an extra effect or if the trigger can be coded without extra effect
* especially because of Strong, the Brutish Thespian and `You gain life rather than lose life from radiation.`
> Strong's last ability changes what happens when the inherent triggered ability associated with having rad counters resolves. You'll still mill cards equal to the number of rad counters you have, and you'll still remove a rad counter from yourself for each nonland card you milled, but instead of losing life, you'll gain 1 life for each nonland card you milled.

~~does Strong's effect apply as "lose Life" replacement, or happens that before?~~
~~How does this Effect applies to effects that say that you can't lose Life or that your life can't change?~~

RulesManager:
> Strong's ability isn't a replacement effect. It changes the rules of the game such that the triggered ability from rad counters causes you to gain life, not lose it. You would gain life in this scenario.

AI:
Other than Poison counters, Rad Counters might not be directly negative

CR:
> 725. Rad Counters
> 725.1. Rad counters are a kind of counter a player can have (see rule 122, “Counters”). There is an inherent triggered ability associated with rad counters. This ability has no source and is controlled by the active player. This is an exception to rule 113.8. The full text of this ability is “At the beginning of each player’s precombat main phase, if that player has one or more rad counters, that player mills a number of cards equal to the number of rad counters they have. For each nonland card milled this way, that player loses 1 life and removes one rad counter from themselves.”
> 725.1a A card that refers to life loss “from radiation” refers to life lost as a result of the triggered ability associated with rad counters.

Hanmac:
Closes #4756 

I don't really know if we still need the Radiation Effect (that puts counters) or that should just be handled by the PutCounter Effect instead

@tool4ever your opinion?

Also the Translations for the Radiation Log needs to be updated


Hanmac:
@Agetian whats your opinion? should Radiation be its own Effect, like Poison or should it be inside PutCounter?

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -863,7 +863,20 @@ public void addCounterInternal(final CounterType counterType, final int n, final
...
+        if (counterType.is(CounterEnumType.RAD)) {
+            if (newValue > 0) {
+                String setCode = null;
+                if (params.containsKey(AbilityKey.Cause)) {
+                    SpellAbility cause = (SpellAbility) params.get(AbilityKey.Cause);
+                    setCode = cause.getHostCard().getSetCode();
+                }
+                createRadiationEffect(setCode);
+            } else {
+                removeRadiationEffect();
```
tool4ever, line 877: you need to move this somewhere else
currently the effect stays around after you're cured

forge-game/src/main/java/forge/game/ability/ApiType.java
```diff
@@ -204,7 +204,9 @@ public enum ApiType {
     DamageResolve (DamageResolveEffect.class),
     ChangeZoneResolve (ChangeZoneResolveEffect.class),
     InternalLegendaryRule (CharmEffect.class),
-    InternalIgnoreEffect (CharmEffect.class);
+    InternalIgnoreEffect (CharmEffect.class),
```
tool4ever, line 207: imo we should also get rid of ChooseCompanionEffect class and map it to Charm as placeholder like those others

Agetian:
Hmm, this seems unique enough to warrant its own effect imo, in order not to complicate PutCounter effect further... But not sure :/ Any other opinions on this one, guys?

tool4ever (approved):

PR #4807, opened 2024-03-10
Keyword: Saddle

Hanmac:
Closes #4803

PR #4808, opened 2024-03-10
Move copy functions to new service

tehdiplomat:
Pre-work to simplify game rollback/snapshot code


Hanmac:
@tehdiplomat we might need to take a look at how much GameCopier shares with getLkiCopy

https://github.com/Card-Forge/forge/blob/959aa74f02f9e77e621031a708dabf184c5dc440/forge-ai/src/main/java/forge/ai/simulation/GameCopier.java#L352-L455

or when moving to a CopyService, if these functions have a common ancestor?

tehdiplomat:
@Hanmac yea thats part of the work I'm doing. I just need this move to happen first because people keep changing the underlying code and its really hard to keep track of what lines are getting added. Ideally all of these things will be flattened to a single simpler file thats testable.

Hanmac:
yeah, so should we do this PR before or after other ones like #4807 ?

tehdiplomat:
@Hanmac Either way is fine. If we have the last two things we want to get in, just note what they are on this PR, and then we can merge this. But yea ideally, I'd merge all of LKI, GameCopier, GameSnapshot into a single use case that we can use for both restoring snapshots, looking back in time, and simulating game states for the AI.

Hanmac:
simulating game states might be useful for that Layer Dependency Rule in the far future

Hanmac (approved):

PR #4815, opened 2024-03-11
Keyword: Spree

### LINKED ISSUE #4813: Keyword: Spree
maybe combine a Charm Effect with an Additional Cost Effect that adds more Cost depending on the Charm Choices?

> Spree (Choose one or more additional costs.)
+{2} —Exchange control of two target creatures
+{1} —Exchange control of two target artifacts
+{1} —Exchange control of two target enchantments

Hanmac:
Closes #4813 

i think it is already functional?

it might just need some GUI love for the generated Charm Description

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/CharmEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/CharmEffect.java
@@ -8,6 +8,7 @@ import com.google.common.collect.Lists;
 import forge.game.ability.AbilityUtils;
 import forge.game.ability.SpellAbilityEffect;
 import forge.game.card.Card;
+import forge.game.cost.Cost;
 import forge.game.player.Player;
 import forge.game.spellability.AbilitySub;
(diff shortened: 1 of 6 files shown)


Agetian (approved):

Agetian:
Looks good, yeah!

Hanmac:
@tool4ever should we try to update the Description now, or should that be later?

tool4ever:
I could check for it soon

Hanmac:
@tool4ever in `CharmEffect::makeFormatedDescription` check for `Spree$ True`
and then use this to also check for SpreeCost on the Options

tool4ever:
image

Hanmac:
the only difference might be to have Spree line rendered instead of the "Choose one or more" line (in the CharmEffect)

the difference would be that it wouldn't have an extra empty line i think

tool4ever (approved):

PR #4820, opened 2024-03-14
Improve simultaneous handling with ChangeZoneAll

### LINKED ISSUE #4819: Detention Sphere returning Wrong Exiled Cards when Bounced Before ETB.
**Describe the bug**
You may have heard of the trick where if you remove/bounce/destroy an Oblivion Ring while it's ETB is on the stack, the LTB trigger resolves before the ETB trigger and the permanent is exiled permanently.

When doing the same trick with a Detention Sphere **TWICE** the exiled cards from the first Detention Sphere return, even though it's LTB trigger has already resolved, and the new Detention Sphere should not be considered the same card? 

Maybe I'm missing something with the wording of Detention Sphere?

**To Reproduce**

Cast Detention Sphere, let it resolve.
ETB trigger: Target an opponents non-land permanent.
With the ETB on the stack, bounce the Detention Sphere (I used Kiora's Dismissal.)
LTB trigger let it resolve: Does nothing, nothing has been exiled.
Let the ETB resolve, exiling targeted permanent.

Above is the intended iteration and works correctly, however doing it twice in a row is the bug:
Cast the same Detention Sphere again, let it resolve.
ETB trigger: Target an opponents non-land permanent.
With the ETB on the stack, bounce the Detention Sphere (I used Kiora's Dismissal.)
**LTB trigger, let it resolve: This returns permanents exiled with the FIRST detention sphere!?** 
Let the ETB resolve, exiling targeted permanent.

**Expected behavior**
Each cast instance of detention sphere should be considered it's own unique card, and not retain memory of any other cards named detention sphere or any cards exiled with any other cast copy.

This bug does not happen with similar cards, like journey to nowhere or oblivion ring.

 - OS: Windows 10
 - Browser NA
 - Version forge-1.6.60

**Additional context**
Thank You!

tool4ever:
Closes #4819

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/DigUntilEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/DigUntilEffect.java
@@ -135,6 +135,10 @@ public class DigUntilEffect extends SpellAbilityEffect {
         boolean sequential = digSite == ZoneType.Library && revealedDest != null && revealedDest.equals(foundDest);
 
         CardZoneTable table = new CardZoneTable(game.copyLastStateBattlefield(), game.copyLastStateGraveyard());
+        CardZoneTable tableSeq = null;
+        if (!sequential) {
+            tableSeq = new CardZoneTable(table.getLastStateBattlefield(), table.getLastStateGraveyard());
+        }
         boolean combatChanged = false;
 
         for (final Player p : getTargetPlayers(sa)) {
@@ -148,7 +152,6 @@ public class DigUntilEffect extends SpellAbilityEffect {
             CardCollection revealed = new CardCollection();
 
             final PlayerZone library = p.getZone(digSite);
-
             final int maxToDig = maxRevealed != null ? maxRevealed : library.size();
 
             for (int i = 0; i < maxToDig; i++) {
@@ -202,7 +205,9 @@ public class DigUntilEffect extends SpellAbilityEffect {
                         foundDest = optionalNoDestination;
                     }
 
-                    CardZoneTable tableSeq = new CardZoneTable(table.getLastStateBattlefield(), table.getLastStateGraveyard());
+                    if (sequential) {
+                        tableSeq = new CardZoneTable(table.getLastStateBattlefield(), table.getLastStateGraveyard());
+                    }
                     Map<AbilityKey, Object> moveParams = AbilityKey.newMap();
                     AbilityKey.addCardZoneTableParams(moveParams, tableSeq);
 
@@ -241,7 +246,9 @@ public class DigUntilEffect extends SpellAbilityEffect {
                         game.getAction().moveTo(foundDest, c, foundLibPos, sa, moveParams);
                     }
 
-                    tableSeq.triggerChangesZoneAll(game, sa);
+                    if (sequential) {
+                        tableSeq.triggerChangesZoneAll(game, sa);
+                    }
                 }
                 revealed.removeAll(found);
             }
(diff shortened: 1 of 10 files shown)


forge-gui/res/cardsfolder/h/honor_the_fallen.txt
```diff
@@ -1,8 +1,9 @@
 Name:Honor the Fallen
 ManaCost:1 W
 Types:Instant
-A:SP$ ChangeZoneAll | Cost$ 1 W | ChangeType$ Creature | Origin$ Graveyard | Destination$ Exile | RememberChanged$ True | ForgetOtherRemembered$ True | SubAbility$ DBGainLife | SpellDescription$ Exile all creature cards from all graveyards. You gain 1 life for each card exiled this way.
-SVar:DBGainLife:DB$ GainLife | Defined$ You | LifeAmount$ X
+A:SP$ ChangeZoneAll | Cost$ 1 W | ChangeType$ Creature | Origin$ Graveyard | Destination$ Exile | RememberChanged$ True | SubAbility$ DBGainLife | SpellDescription$ Exile all creature cards from all graveyards. You gain 1 life for each card exiled this way.
```
Hanmac, line 4: Remove Cost?

forge-game/src/main/java/forge/game/ability/effects/DigUntilEffect.java
```diff
@@ -135,6 +135,10 @@ public void resolve(SpellAbility sa) {
         boolean sequential = digSite == ZoneType.Library && revealedDest != null && revealedDest.equals(foundDest);
 
         CardZoneTable table = new CardZoneTable(game.copyLastStateBattlefield(), game.copyLastStateGraveyard());
+        CardZoneTable tableSeq = null;
+        if (!sequential) {
+            tableSeq = new CardZoneTable(table.getLastStateBattlefield(), table.getLastStateGraveyard());
+        }
```
Hanmac, line 141: reuse table if not sequential?

`CardZoneTable tableSeq = sequential ? null : table;`

forge-gui/res/cardsfolder/h/honor_the_fallen.txt
tool4ever: hmn ok, but a cleanup PR for all might be easier eventually

Hanmac: i normally try to update files I'm going to touch anyways

forge-game/src/main/java/forge/game/ability/effects/DigUntilEffect.java
tool4ever: I can't do that currently

because even when revealing first the found ones get moved together before the other ones
e. g. on _Collision of Realms_:
> All creatures are shuffled into their owners' libraries at the same time. Then all players who shuffled nontoken creatures into their library this way reveal cards. **Then all creature cards revealed this way enter the battlefield at the same time. Then all other cards are put on the bottom of their owners' libraries.**

It would need more refactoring to split the Players loop into two first but that might get a bit messy with some of the checks on those cards

and I'm not sure a corner case for it exists anyway... 🤔

tool4ever: if not sequential both will be used
`table` will contain the revealed cards that should be moved after doing the found ones

Hanmac (approved):

PR #4827, opened 2024-03-17
CardType: add isOutlaw

### LINKED ISSUE #4805: CreatureType Group: Outlaw
Outlaw:

- Assassin
- Mercenary
- Pirate
- Rogue
- Warlock

i think the simplest solution would be to add a CardProperty/CardStateProperty for this.

the Creature Type List should not be part of the card because it should not be available to CardTextChanges

the isOutlaw should probably done in CardType itself, so it would reduce the calls for getCardType
isHistoric should probably be moved there too

Hanmac:
Closes #4805 

no cards for Outlaw yet

the closest effect we could make right now is "Boneyard Desecrator" but it has incomplete PT

### DIFF
--- a/forge-core/src/main/java/forge/card/CardType.java
+++ b/forge-core/src/main/java/forge/card/CardType.java
@@ -489,6 +489,33 @@ public final class CardType implements Comparable<CardType>, CardTypeView {
     }
 
     @Override
+    public final boolean isAttachment() { return isAura() || isEquipment() || isFortification(); }
+    @Override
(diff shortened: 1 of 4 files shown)


tool4ever (approved):

Hanmac:
@tool4ever 
we maybe could change some `.getType().hasSubtype("Saga")` checks into `getType().isSaga()` checks too now

Hanmac:
maybe later we could move the Party types into a Constant in CardType too

PR #117, opened 2022-04-22
Timestamp rework

Hanmac:
Draft for now, i need to update more Effects, so that the WrappedAbility Check can be removed

The Main reason for this MR, Split Timestamp into:
- GameTimestamp that is used by the Rules when an Card becomes an new Object
- LayerTimestamp that is used when the Rules talk about a Card getting a new timestamp, only used for Static Layer Effects

Also
- For Player, add a new check "isInGame" which means the Player has neither lost, nor won (in case of Limited Influence)
- Also More Effects should check for GameTimestamp in Effect so it is checked for Triggered objects
- canBeTargetedBy doesn't needs to be checked in Effect, this is handled by Magic Stack

PR #4830, opened 2024-03-17
TurnFaceUp Trigger use LKI for Card

Hanmac:
DamageSource for this now uses LKI from trigger

@tool4ever we need to update the other Cards using TriggeredCardLKICopy

we either need to update more Cards using `NoTimestampCheck` or we need to add `DealDamage` to the updated API in `WrappedAbility`

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -844,7 +844,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
             game.getReplacementHandler().run(ReplacementType.TurnFaceUp, repParams);
 
             // Run triggers
-            final Map<AbilityKey, Object> runParams = AbilityKey.mapFromCard(this);
+            final Map<AbilityKey, Object> runParams = AbilityKey.mapFromCard(CardCopyService.getLKICopy(this));
             runParams.put(AbilityKey.Cause, cause);
 
             triggerHandler.registerActiveTrigger(this, false);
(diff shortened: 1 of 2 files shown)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -844,7 +844,7 @@ public boolean turnFaceUp(boolean runTriggers, SpellAbility cause) {
             game.getReplacementHandler().run(ReplacementType.TurnFaceUp, repParams);
 
             // Run triggers
-            final Map<AbilityKey, Object> runParams = AbilityKey.mapFromCard(this);
+            final Map<AbilityKey, Object> runParams = AbilityKey.mapFromCard(CardCopyService.getLKICopy(this));
```
tool4ever, line 847: like I said this does not really fix anything
this trigger should not look back in time by default
now when casting _Giant Growth_ on it in response won't increase the damage it deals...

Hanmac: The Problem is related to https://github.com/Card-Forge/forge/pull/117

DamageDealEffect needs to be updated to handle LKI info "IF" needed

we have updated other effects to be able to use the GameObject while the Card is still in the same zone,
but fall back to LKI when it isn't

Hanmac: @tool4ever the bigger Problem:

DamageSource might be able to handle LKI and update it to GameObject
calculateAmount, the function that gets the Number with `TriggeredCard$CardPower` is not able to do that

tool4ever: DamageDealEffect should already convert its source into LKI if able (from changeZoneLKI)

but there is no logic for converting it back, so this would also use source from wrong time

Issue #1249, opened 2022-07-30
ThisTurnEntered can sometimes fail as source for Discarded


tool4ever:
Because of _Harness Infinity_ you now get cards back after with _Shadow of the Grave_.

Hanmac:
what of them using ThisTurnEntered? is it DiscardedThisTurn?

technically none of these cards should care about each other?

tool4ever:
Shadow of the Grave:
>Return to your hand all cards in your graveyard that you cycled or discarded this turn.

I think Harness Infinity might be the only card that puts card from your hand into graveyard that's not discarding.

Hanmac:
thats why i was asking what was using ThisTurnEntered, but yeah we might need extra logic for this.

same with Mill if something would look for milled cards

Hanmac:
@Agetian @tool4ever @Northmoc you guys opinion to this:

- should we make a hotfix in ChangeZoneEffect?
- or should we take the opportunity to make a better DiscardEffect? (with maybe better AI logic?)

tool4ever:
There were still other Discard bugs in the old issues I couldn't fix yet, unfortunately for the whole list I'd have to dig it up again.

Not sure how related they would be but some were replacement problems:
- _Library of Leng_ + _Tolarian Winds_ interaction
- > If you discard Obstinate Baloth and put it onto the battlefield, you’ve still discarded it. Abilities that trigger when you discard a card (such as that of Raiders’ Wake) will still trigger.

Hanmac:
it will be more clear when Discard is its own effect

Agetian:
Yes, agreed that Discard effect sounds cleaner :)

Hanmac:
we should make a list of current Discard Effects:
- Discard Defined Card, in most cases by Trigger
- Discard By Defined Player (target)
  - Optional Reveal or Look (Reveal to you)
  - Random
  - By Chooser
  - When Choose Mode, add ValidCard
  - Special Hand case

After i finished writing, there already is a discard effect ;P
We all seems to have forgotten it XD
But it only works for players currently

What we should do is to move Defined Card into the Discard Effect

Hanmac:
i think i finished the first part

but i need to read how Obstinate Baloth would interact with Madness, like if you would choose discard it for madness "If a player would discard this card, *that player discards it*" would that still cause the Replacement Effect from the Baloth to run?

tool4ever:
For Madness it might be ok because it will never be controlled by opponent?

[...]

PR #4833, opened 2024-03-18
ReplaceDiscard doesn't exist

tool4ever:
This is a second part for #1249

image

I think this cleaned up nicely...

- [x] update remaining cards

But I won't fix everything in this, more might follow later 🐌


Hanmac:
@tool4ever for a followup later, we might need a flag to check if a card got revealed while going from Hand -> (Revealed) -> Library

then we can update the Discard Trigger to see if they are in a hidden zone but revealed before

Hanmac (approved):

forge-gui/res/cardsfolder/l/library_of_leng.txt
```diff
@@ -2,7 +2,7 @@ Name:Library of Leng
 ManaCost:1
 Types:Artifact
 S:Mode$ Continuous | EffectZone$ Battlefield | Affected$ You | SetMaxHandSize$ Unlimited | Description$ You have no maximum hand size.
-R:Event$ Discard | ActiveZones$ Battlefield | ValidCard$ Card.YouCtrl | ReplaceWith$ DiscardToLibrary | Optional$ True | DiscardFromEffect$ True | Description$ If an effect causes you to discard a card, discard it, but you may put it on top of your library instead of into your graveyard.
-SVar:DiscardToLibrary:DB$ Discard | Defined$ ReplacedPlayer | Mode$ Defined | DefinedCards$ ReplacedCard | DiscardToTopOfLibrary$ True
+R:Event$ Discard | ActiveZones$ Battlefield | ValidCard$ Card.YouCtrl | Discard$ True | EffectOnly$ True | ReplaceWith$ DiscardToLibrary | Optional$ True | Description$ If an effect causes you to discard a card, discard it, but you may put it on top of your library instead of into your graveyard.
```
Hanmac, line 5: @tool4ever you forgot to update the Mode

and i didn't noticed that

tool4ever: thanks, fixed in open PR🤦‍♂️

PR #4840, opened 2024-03-19
Card: differ between Game and Layer Timestamp

Hanmac:
Smaller PR than #117 

@tool4ever can you debug why exactly Kalitas is failing? and why only one of them?

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
@@ -1835,7 +1835,7 @@ public class ComputerUtilCard {
         for (final Card c : list) {
             // remove old boost that might be copied
             for (final StaticAbility stAb : c.getStaticAbilities()) {
-                vCard.removePTBoost(c.getTimestamp(), stAb.getId());
+                vCard.removePTBoost(c.getLayerTimestamp(), stAb.getId());
                 if (!stAb.checkMode("Continuous")) {
                     continue;
                 }
@@ -1858,7 +1858,7 @@ public class ComputerUtilCard {
                     String addT = stAb.getParam("AddToughness");
                     def = AbilityUtils.calculateAmount(addT.contains("Affected") ? vCard : c, addT, stAb, true);
                 }
-                vCard.addPTBoost(att, def, c.getTimestamp(), stAb.getId());
+                vCard.addPTBoost(att, def, c.getLayerTimestamp(), stAb.getId());
             }
         }
     }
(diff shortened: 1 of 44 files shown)


tool4ever:
Hmn, I think this might come down to a logic error with the new changeZoneLKI structure
If we want to retrieve the previous LKI for Object X, then we need to use the game timestamp it had in its old zone
or in most cases nothing will be found...

tool4ever:
So there might be two use cases for changeZoneLKI:
- A - we have the new object and want to know its **previous** LKI
- B - we reference the old object but want its "final" LKI (mostly because of undone statics)

I think A might be rarer?:
1. Used by ReplacementHandler

B:
2. CardProperty: this gets used only partially with focus on controller and zone checks, I remember it fixed some things
3. DamageDeal effects: I think this fixes something like Lifelink by static missing e.g. if a creature dies before it's ability resolves
4. TriggerChangesZone DamageReceivedCondition$: I'll check that one, looks more like a incomplete LKI problem?

6. some other less common parts

in your old branch you refactored 1. to avoid the usage, should still be safe
if we don't need A without it then that could be the complete fix for the test

Hanmac:
"Used by ReplacementHandler" I need to look at this part

It probably can be changed to use LastStateBattlefield instead

Hanmac:
@tool4ever so Replacement Handler is updated

tool4ever:
For triggers we have 2 "problem cases":

- TriggerDestroyed - this uses old zone card, which means _Karmic Justice_ incorrectly triggers for enchantments animated by _Opalescence_
- TriggerSacrificed - this triggers with LKI, which means cards like _It That Betrays_ need to use `Defined$ TriggeredCard` to work (which can be broken if you exile and then return it to Graveyard)

So they either
- need both in runParams like TriggerChangesZone
- some AbilityUtils logic that falls back to changeZoneLKI

But that's something that can be decided later

forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
```diff
@@ -118,12 +118,10 @@ public List<ReplacementEffect> getReplacementList(final ReplacementType event, f
             public boolean visit(Card crd) {
                 Card c = preList.get(crd);
                 Zone cardZone = game.getZoneOf(c);
-                Zone lkiZone = game.getChangeZoneLKIInfo(c).getLastKnownZone();
 
                 // only when not prelist
-                boolean noLKIstate = c != crd || event != ReplacementType.Moved;
+                boolean noLKIstate = c != crd || event != ReplacementType.Moved || !ZoneType.Battlefield.equals(runParams.get(AbilityKey.Origin));;
```
tool4ever, line 123: turns out it's not good enough :/
the logic needs to be like before:
- if the RE host was in play, use RE host LKI
but now it's instead:
- if card LTB, use RE host LKI

this breaks the Exile part of _Bleed Dry_

Hanmac: Urgh, then it probably would have failed in my old branch too

can you try to fix it on this branch?

tool4ever: I think I healed it now

tool4ever (approved):

Issue #2193, opened 2023-01-01
Discarded trigger can fail when depending on static


tool4ever:
When you discard a non-Spirit creature with _Maskwood Nexus_ in play _Shipwreck Sifters_ does not trigger.

Easy fix would be using the new card from ``GameAction.changeZone`` to run the trigger instead of the old hand version.

Only question is if it should also trigger if you use ``Ashes of the Fallen`` instead because it only affects Graveyard.

But I think it should probably too?
> 603.10. Normally, objects that exist immediately after an event are checked to see if the event matched
any trigger conditions, and continuous effects that exist at that time are used to determine what the
trigger conditions are and what the objects involved in the event look like.
No LKI exception does apply here

As bonus we could then finally use ``Defined$ TriggeredCardLKICopy`` on cards like _Bag of Holding_.

@Hanmac
your opinion?

Hanmac:
yeah that might work

Hanmac:
for stuff like Shipwreck Sifters, i need to ask Judges if it should look at the LKI in the hand, or the new card in the graveyard. (in cases of Ashes)

imo the most important part is when the into the graveyard zone change was replaced by a zone change into the library BUT is not revealed like by sun and moon:
> If a card would be put into enchanted player’s graveyard from anywhere, instead that card is revealed and put on the bottom of that player’s library.

because if the card is not revealed then some trigger should not trigger.

Hanmac:
Some Higher Judge said:
> Discarding a Card is not listed in the exceptions 603.10, so it appears to not one of the things that "look back". so, Ashes should work with Sifters. while, something like !yix jailer should stop a non-Spirit Disturb trigger

meaning we should be able to use LKI of new Object?

Hanmac:
> Library of Leng
> If an effect causes you to discard a card, discard it, but you may put it on top of your library instead of into your graveyard.

this one doesn't reveal, so _Shipwreck Sifters_ shouldn't trigger

Hanmac:
> 701.8c If a card is discarded, but an effect causes it to be put into a hidden zone instead of into its owner’s graveyard without being revealed, all values of that card’s characteristics are considered to be undefined. If a card is discarded this way to pay a cost that specifies a characteristic about the discarded card, that cost payment is illegal; the game returns to the moment before the cost was paid (see rule 728, “Handling Illegal Actions”).

We need to respect this one

tool4ever:
[...]

PR #4845, opened 2024-03-20
Fix Pure Intentions

tool4ever:
Small part for #2193

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -1911,8 +1911,6 @@ public class GameAction {
             runParams.putAll(params);
         }
         game.getTriggerHandler().runTrigger(TriggerType.Destroyed, runParams, false);
-        // in case the destroyed card has such a trigger
-        game.getTriggerHandler().registerActiveLTBTrigger(c);
 
         final Card sacrificed = sacrificeDestroy(c, sa, params);
         return sacrificed != null;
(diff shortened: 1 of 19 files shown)


Hanmac:
are all ApiType.DealDamage updated, or is that just wishful thinking?

tool4ever:
> are all ApiType.DealDamage updated, or is that just wishful thinking?

you mean DamageEach etc.? I left those alone because of the old PR and only touched this to get rid of the `NoTimetampCheck$` bandaid

if you mean all scripts, then no, but I can check those?
(my hope was because DamageDealEffect had timestamp check for years actually it should be fairy unproblematic 😬 )

Hanmac:
> (my hope was because DamageDealEffect had timestamp check for years actually it should be fairy unproblematic 😬 )

the timestamp checks are only for the DamageSource property, but may not work for the DamageAmount (TriggeredCard) yet

See my other issue with Pyrotechnic Performer leaving the Battlefield

tool4ever:
inside DamageDealEffect they're currently done for the affected (cards) which seems right
after more accuracy from game timestamps LKI fallback for NumDmg$ can be implemented?
(ideally inside AbilityUtils to server for all effects)

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -1911,8 +1911,6 @@ public final boolean destroy(final Card c, final SpellAbility sa, final boolean
             runParams.putAll(params);
         }
         game.getTriggerHandler().runTrigger(TriggerType.Destroyed, runParams, false);
-        // in case the destroyed card has such a trigger
-        game.getTriggerHandler().registerActiveLTBTrigger(c);
```
Hanmac, line 1915: is the LTB Trigger registered somewhere deeper in the code?

Hanmac: then i think its okay?

PR #4864, opened 2024-03-22
YMKM: izzet_polarizer + support

Northmoc:


forge-game/src/main/java/forge/game/ability/effects/MakeCardEffect.java
```diff
@@ -49,7 +49,13 @@ public void resolve(SpellAbility sa) {
                     !player.getController().confirmAction(sa, null, Localizer.getInstance().getMessage(desc), null)) {
             		return;
             }
-            if (sa.hasParam("Name")) {
+            if (sa.hasParam("Choices")) {
```
tool4ever, line 52: why can't this share the code with SpellBook if you only want a different message?

tool4ever (approved):

PR #4873, opened 2024-03-23
MagicStack: fix fizzle removing too much Targets

### LINKED ISSUE #3893: Order of target legality checking with dependency restriction has unwanted influence
The way ``hasFizzled`` removes illegal targets sequentially could cause some weird results.

Example:
- Cast _Blatant Thievery_ on Opps A Creature X + Opps B Creature Y
- Before resolve have A gain control of Y
- You will now only gain control of X or Y (whichever you targeted last is kept)

The ruling on it:
> If a permanent changes controller after being targeted but before this spell resolves, you won't gain control of that permanent.

Means it should always be X that is kept?

_Chaos Mutation_ is worded differently:
> The phrase “target creatures controlled by different players” means that no two targets may be controlled by the same player.

Means it should always fizzle?

### LINKED ISSUE #4843: MagicStack needs to handle invalid Targets separate
https://github.com/Card-Forge/forge/blob/93a52a4263538910543ad222f20d056a2bcfb388/forge-game/src/main/java/forge/game/zone/MagicStack.java#L653-L684

for cards like "Target Player and Target Card they Control" they become invalid when the Player gets Hexproof because the second part about "they control" get messed up

we need to collect the invalid targets in the loop and then remove all of invalid targets together

(Effects that currently check for "canTarget" should be removed, because the MagicStack already does the stuff)

Hanmac:
Closes  #4843, #1179

Does not fix #3893 yet

@tool4ever i'm unsure about _Dead Ringers_ and if we should remove "RememberOriginalTargets" there too?


forge-game/src/main/java/forge/game/zone/MagicStack.java
```diff
@@ -637,71 +637,62 @@ public final boolean hasLegalTargeting(final SpellAbility sa) {
...
                 }
             }
         }
-        else if (sa.getTargetCard() != null) {
-            fizzle = !sa.canTarget(sa.getTargetCard());
-        } else {
-            // Set fizzle to the same as the parent if there's no target info
-            fizzle = parentFizzled;
+        if (sa.getSubAbility() != null) {
+            fizzle = hasFizzled(sa.getSubAbility(), source) && fizzle;
```
tool4ever, line 685: your refactor is wrong, I don't like that no test catch that
e. g. _Annihilate_ still resolves when target gains hexproof

Hanmac: Ugh can you try to update it?

And if possible add a test case for this?

tool4ever: - [ ] I want to check if there are some cards where it should still fizzle from ParentTarget (maybe because they got scripted weirdly)

tool4ever: I found _Breaking of the Fellowship_ which also fizzles on master if only the first target becomes illegal 💩

tool4ever: Fixed now

Hanmac:
> @tool4ever i'm unsure about _Dead Ringers_ and if we should remove "RememberOriginalTargets" there too?

can we look about Dead Ringers too if we can do this without Remember?

tool4ever:
I'm leaning towards also removing it
Imo the rules for information about illegal targets don't allow it (no longer?)

Hanmac:
@tool4ever what if we flip the script?

Have MagicStack not remove Targets, just check if there is a valid one? 🤔 

With that might might allow LKI

tool4ever:
Imo the targets still need to be removed, so no part of the effect can use them

But I have some ideas I can try after this one

Hanmac:
i was thinking in trying to write it without parent fizzle, but my idea didn't work out

PR #4876, opened 2024-03-23
BIG: Sword of Wealth and Power & Nexus of Becoming

dracontes:
Tested card scripts for:
- [Sword of Wealth and Power](https://scryfall.com/card/big/26/sword-of-wealth-and-power)
- [Nexus of Becoming](https://scryfall.com/card/big/25/nexus-of-becoming)

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/sword_of_wealth_and_power.txt
+++ b/forge-gui/res/cardsfolder/upcoming/sword_of_wealth_and_power.txt
@@ -1,0 +1,12 @@
+Name:Sword of Wealth and Power
+ManaCost:3
+Types:Artifact Equipment
+K:Equip:2
+S:Mode$ Continuous | Affected$ Creature.EquippedBy | AddPower$ 2 | AddToughness$ 2 | AddSVar$ SwordOfWealthAndPowerCE | AddKeyword$ Protection:Instant:instants & Protection:Sorcery:sorceries | Description$ Equipped creature gets +2/+2 and has protection from instants and from sorceries.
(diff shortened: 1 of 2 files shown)


forge-gui/res/cardsfolder/upcoming/sword_of_wealth_and_power.txt
```diff
@@ -0,0 +1,12 @@
+Name:Sword of Wealth and Power
+ManaCost:3
+Types:Artifact Equipment
+K:Equip:2
+S:Mode$ Continuous | Affected$ Creature.EquippedBy | AddPower$ 2 | AddToughness$ 2 | AddSVar$ SwordOfWealthAndPowerCE | AddKeyword$ Protection:Instant:instants & Protection:Sorcery:sorceries | Description$ Equipped creature gets +2/+2 and has protection from instants and from sorceries.
```
tool4ever, line 5: write it as single protection KW

dracontes: Done. I take it this is another cleanup item on the to-do list.

tool4ever: @Hanmac 
we prefer to split it for the engine or have the text closer to the print?

Hanmac: Have two Separate keywords would be easier (there is also a Rule about that)

Also haven't I updated the Plural style?
So shouldn't `Protection:Instant` be enough?

dracontes: FWIW, this is what I get with `AddKeyword$ Protection:Instant & Protection:Sorcery:sorceries`, assuming the plural style doesn't cover 'sorcery'.
image

tool4ever: I think it should, try leaving that part away

dracontes: Turns out the plural style does cover 'sorcery'. Edited the relevant file accordingly.

tool4ever (approved):

PR #4888, opened 2024-03-25
Fix Rose, Cutthroat Raider

### LINKED ISSUE #4887: Rose Cutthroat Raider doesn't do anything
After combat, if you attacked any number of players, Rose will trigger, but the trigger won't make any junk tokens.

tool4ever:
Closes #4887


Northmoc (approved):
👍

PR #4889, opened 2024-03-25
YMKM: perforator_crocodile + support

Northmoc:


tool4ever (approved):
ok for now here since the nightmare is contained in a single effect

Northmoc:
> ok for now here since the nightmare is contained in a single effect

MakeCardEffect is mostly the "support digital nonsense" effect anyway

PR #4891, opened 2024-03-25
YMKM: ominous_lockbox + support

Northmoc:


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -301,6 +301,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
     private List<String> chosenColors;
     private List<String> chosenName = new ArrayList<>();
     private Integer chosenNumber;
+    private Integer secretChosenNum;
```
tool4ever, line 304: like on #4574 no extra field should be needed

forge-gui/res/cardsfolder/upcoming/ominous_lockbox.txt
```diff
@@ -0,0 +1,10 @@
+Name:Ominous Lockbox
+ManaCost:1 U
+Types:Artifact Clue
+K:ETBReplacement:Other:ChooseNum
+SVar:ChooseNum:DB$ ChooseNumber | SecretlyChoose$ True | Min$ 1 | Max$ 20 | KeepSecret$ True | SpellDescription$ As CARDNAME enters the battlefield, secretly choose a number between 1 and 20.
```
tool4ever, line 5: can you also rename this param to Secretly?
then they should finally all be consistent...

Northmoc: I think ChooseType is already `Secretly$`

tool4ever (approved):

PR #4896, opened 2024-03-25
YMKM 6 cards

Simisays:


forge-gui/res/cardsfolder/upcoming/scour_the_scene.txt
```diff
@@ -0,0 +1,12 @@
+Name:Scour the Scene
+ManaCost:W U
+Types:Enchantment
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | TriggerZones$ Battlefield | Execute$ TrigToken | TriggerDescription$ When CARDNAME enters the battlefield, create a 2/2 white and blue Detective creature token.
+SVar:TrigToken:DB$ Token | TokenScript$ wu_2_2_detective
+T:Mode$ Sacrificed | ValidCard$ Artifact | ValidPlayer$ You | Execute$ TrigChooseCard | TriggerZones$ Battlefield | TriggerDescription$ Whenever you sacrifice an artifact, choose a creature card in your hand or a creature you control. It perpetually gets +1/+0.
+SVar:TrigChooseCard:DB$ ChooseCard | ChoiceZone$ Hand,Battlefield | ChoiceTitle$ Choose a creature card in your hand or a creature you control | Mandatory$ True | Choices$ Creature.YouOwn | SubAbility$ DBPump
+SVar:DBPump:DB$ Pump | PumpZone$ Hand,Battlefield | Defined$ ChosenCard | NumAtt$ 1 | Duration$ Perpetual | SubAbility$ DBCleanup
+SVar:DBCleanup:DB$ Cleanup | ClearChosenCard$ True
+DeckHas:Ability$Token
```
Northmoc, line 10: & Type$Detective

forge-gui/res/cardsfolder/upcoming/scour_the_scene.txt
```diff
@@ -0,0 +1,12 @@
+Name:Scour the Scene
+ManaCost:W U
+Types:Enchantment
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | TriggerZones$ Battlefield | Execute$ TrigToken | TriggerDescription$ When CARDNAME enters the battlefield, create a 2/2 white and blue Detective creature token.
+SVar:TrigToken:DB$ Token | TokenScript$ wu_2_2_detective
+T:Mode$ Sacrificed | ValidCard$ Artifact | ValidPlayer$ You | Execute$ TrigChooseCard | TriggerZones$ Battlefield | TriggerDescription$ Whenever you sacrifice an artifact, choose a creature card in your hand or a creature you control. It perpetually gets +1/+0.
+SVar:TrigChooseCard:DB$ ChooseCard | ChoiceZone$ Hand,Battlefield | ChoiceTitle$ Choose a creature card in your hand or a creature you control | Mandatory$ True | Choices$ Creature.YouOwn | SubAbility$ DBPump
```
Northmoc, line 7: Choices$ Creature.YouCtrl or stuff you had gained control of on the battlefield would not be eligible

forge-gui/res/cardsfolder/upcoming/sproutwatch_dryad.txt
```diff
@@ -0,0 +1,8 @@
+Name:Sproutwatch Dryad
+ManaCost:1 G G
+Types:Creature Dryad
+PT:2/3
+T:Mode$ Phase | Phase$ BeginCombat | TriggerZones$ Battlefield | Execute$ TrigPump | TriggerDescription$ At the beginning of each combat, Sproutwatch Dryad gains flying until end of turn if a creature you control or a card in your hand has flying. The same is true for first strike, double strike, deathtouch, haste, hexproof, indestructible, lifelink, menace, reach, trample, and vigilance.
+SVar:TrigPump:DB$ PumpAll | ValidCards$ Creature.YouCtrl | KW$ First Strike & Flying & Deathtouch & Double Strike & Haste & Hexproof & Indestructible & Lifelink & Menace & Reach & Trample & Vigilance | SharedKeywordsZone$ Hand | SharedRestrictions$ Creature.YouOwn
```
Northmoc, line 6: Shouldn't ValidCards just be Self?
I think ShardRestriction has to be YouCtrl again for the same reason as the other

forge-gui/res/cardsfolder/upcoming/sproutwatch_dryad.txt
```diff
@@ -0,0 +1,8 @@
+Name:Sproutwatch Dryad
+ManaCost:1 G G
+Types:Creature Dryad
+PT:2/3
+T:Mode$ Phase | Phase$ BeginCombat | TriggerZones$ Battlefield | Execute$ TrigPump | TriggerDescription$ At the beginning of each combat, Sproutwatch Dryad gains flying until end of turn if a creature you control or a card in your hand has flying. The same is true for first strike, double strike, deathtouch, haste, hexproof, indestructible, lifelink, menace, reach, trample, and vigilance.
```
Northmoc, line 5: CARDNAME in desc

forge-gui/res/cardsfolder/upcoming/snarlfang_vermin.txt
```diff
@@ -0,0 +1,12 @@
...
+SVar:TrigSuspect:DB$ AlterAttribute | Defined$ TriggeredTargetLKICopy | RememberAltered$ True | Attributes$ Suspected | SubAbility$ DBAnimate
+SVar:DBAnimate:DB$ Animate | Defined$ Remembered | Triggers$ CombatTrig | Duration$ Perpetual | SubAbility$ DBCleanup
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+SVar:CombatTrig:Mode$ DamageDone | CombatDamage$ True | ValidSource$ Card.Self | ValidTarget$ Creature | TriggerZones$ Battlefield | Execute$ TrigSuspect | TriggerDescription$ Whenever CARDNAME deals combat damage to a creature, if that creature is still on the battlefield, suspect that creature. It perpetually gains this ability.
+T:Mode$ ChangesZone | Origin$ Battlefield | PresentZone$ Graveyard | IsPresent$ Card.StrictlySelf | TriggerZones$ Graveyard | Destination$ Graveyard | ValidCard$ Creature.OppCtrl+IsSuspected | Execute$ TrigLoseLife | TriggerDescription$ Whenever a suspected creature an opponent controls dies while Snarlfang Vermin is in your graveyard, that opponent loses 1 life.
```
Northmoc, line 10: CARDNAME in desc

forge-gui/res/cardsfolder/upcoming/snarlfang_vermin.txt
```diff
@@ -0,0 +1,12 @@
+Name:Snarlfang Vermin
+ManaCost:B
+Types:Creature Rat
+PT:2/1
+T:Mode$ DamageDone | CombatDamage$ True | ValidSource$ Card.Self | ValidTarget$ Creature | TriggerZones$ Battlefield | Execute$ TrigSuspect | TriggerDescription$ Whenever CARDNAME deals combat damage to a creature, if that creature is still on the battlefield, suspect that creature. It perpetually gains this ability.
+SVar:TrigSuspect:DB$ AlterAttribute | Defined$ TriggeredTargetLKICopy | RememberAltered$ True | Attributes$ Suspected | SubAbility$ DBAnimate
```
Northmoc, line 6: What happens if the damaged creature dies? I feel like this line should check that the Defined is still on the battlefield.

forge-gui/res/cardsfolder/upcoming/veko_deaths_doorkeeper.txt
```diff
@@ -0,0 +1,11 @@
+Name:Veko, Death's Doorkeeper
+ManaCost:W B
+Types:Legendary Creature Spirit Cleric
+PT:1/3
+K:Extort
+A:AB$ ChangeZone | Cost$ Sac<1/Creature.nonSpirit/nonspirit creature> | RememberChanged$ True | TgtPrompt$ Choose target creature card in your graveyard | ValidTgts$ Creature.YouOwn | Origin$ Graveyard | Destination$ Hand | SorcerySpeed$ True | SubAbility$ DBAnimate | AILogic$ SacAndRetFromGrave | SpellDescription$ Return target creature card from your graveyard to your hand. It perpetually becomes a Spirit, has base power and toughness 1/1, and gains "You may pay {W/B} rather than pay this spell's mana cost." Activate only as a sorcery.
+SVar:DBAnimate:DB$ Animate | Types$ Spirit | Defined$ Remembered | Power$ 1 | staticAbilities$ PerpAltCost | Toughness$ 1 |Duration$ Perpetual | SubAbility$ DBCleanup
+SVar:PerpAltCost:Mode$ Continuous | EffectZone$ All | MayPlay$ True | MayPlayAltManaCost$ WB | MayPlayDontGrantZonePermissions$ True | Affected$ Card.Self | AffectedZone$ Hand,Graveyard,Library,Exile,Command | Description$ "You may pay {W/B} rather than pay this spell’s mana cost."
```
Northmoc, line 8: `’` -> `'`

forge-gui/res/cardsfolder/upcoming/veko_deaths_doorkeeper.txt
```diff
@@ -0,0 +1,11 @@
+Name:Veko, Death's Doorkeeper
+ManaCost:W B
+Types:Legendary Creature Spirit Cleric
+PT:1/3
+K:Extort
+A:AB$ ChangeZone | Cost$ Sac<1/Creature.nonSpirit/nonspirit creature> | RememberChanged$ True | TgtPrompt$ Choose target creature card in your graveyard | ValidTgts$ Creature.YouOwn | Origin$ Graveyard | Destination$ Hand | SorcerySpeed$ True | SubAbility$ DBAnimate | AILogic$ SacAndRetFromGrave | SpellDescription$ Return target creature card from your graveyard to your hand. It perpetually becomes a Spirit, has base power and toughness 1/1, and gains "You may pay {W/B} rather than pay this spell's mana cost." Activate only as a sorcery.
+SVar:DBAnimate:DB$ Animate | Types$ Spirit | Defined$ Remembered | Power$ 1 | staticAbilities$ PerpAltCost | Toughness$ 1 |Duration$ Perpetual | SubAbility$ DBCleanup
```
Northmoc, line 7: spacing

forge-gui/res/cardsfolder/upcoming/veko_deaths_doorkeeper.txt
```diff
@@ -0,0 +1,11 @@
+Name:Veko, Death's Doorkeeper
+ManaCost:W B
+Types:Legendary Creature Spirit Cleric
+PT:1/3
+K:Extort
+A:AB$ ChangeZone | Cost$ Sac<1/Creature.nonSpirit/nonspirit creature> | RememberChanged$ True | TgtPrompt$ Choose target creature card in your graveyard | ValidTgts$ Creature.YouOwn | Origin$ Graveyard | Destination$ Hand | SorcerySpeed$ True | SubAbility$ DBAnimate | AILogic$ SacAndRetFromGrave | SpellDescription$ Return target creature card from your graveyard to your hand. It perpetually becomes a Spirit, has base power and toughness 1/1, and gains "You may pay {W/B} rather than pay this spell's mana cost." Activate only as a sorcery.
```
Northmoc, line 6: use AnimateSubAbilty instead of remembering

forge-gui/res/cardsfolder/upcoming/veko_deaths_doorkeeper.txt
```diff
@@ -0,0 +1,11 @@
...
+SVar:DBAnimate:DB$ Animate | Types$ Spirit | Defined$ Remembered | Power$ 1 | staticAbilities$ PerpAltCost | Toughness$ 1 |Duration$ Perpetual | SubAbility$ DBCleanup
+SVar:PerpAltCost:Mode$ Continuous | EffectZone$ All | MayPlay$ True | MayPlayAltManaCost$ WB | MayPlayDontGrantZonePermissions$ True | Affected$ Card.Self | AffectedZone$ Hand,Graveyard,Library,Exile,Command | Description$ "You may pay {W/B} rather than pay this spell’s mana cost." 
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+DeckHas:Ability$Graveyard|LifeGain
```
Northmoc, line 10: `|Sacrifice`

forge-gui/res/cardsfolder/upcoming/tought_string_analyst.txt
```diff
@@ -0,0 +1,12 @@
+Name:Thought-String Analyst
```
Northmoc, line 1: file name is spelled incorrectly

forge-gui/res/cardsfolder/upcoming/tought_string_analyst.txt
```diff
@@ -0,0 +1,12 @@
...
+ManaCost:U B
+Types:Creature Human Wizard
+PT:2/2
+T:Mode$ Phase | Phase$ Upkeep | ValidPlayer$ You | Execute$ TrigDig | TriggerZones$ Battlefield | TriggerDescription$ At the beginning of your upkeep, exile the top card of target opponent's library face down. You lose life equal to its mana value. You may look at and play that card for as long as it remains exiled, and mana of any type can be spent to cast that spell.
+SVar:TrigDig:DB$ Dig | DigNum$ 1 | Defined$ Targeted | ValidTgts$ Opponent | ChangeNum$ All | DestinationZone$ Exile | ExileFaceDown$ True | RememberChanged$ True | SubAbility$ DBLoseLife
+SVar:DBLoseLife:DB$ LoseLife | LifeAmount$ X | SubAbility$ DBEffect
+SVar:X:Targeted$CardManaCost
```
Northmoc, line 8: Targeted seems wrong... should be `Remembered`?

forge-gui/res/cardsfolder/upcoming/snarlfang_vermin.txt
Simisays: Like this ?

forge-gui/res/cardsfolder/upcoming/veko_deaths_doorkeeper.txt
Simisays: Hmm I don't get this to work with AnimateSubAbility. I guess this isn't wired up outside of battlefield ?

tool4ever: without should be fine anyway

forge-gui/res/cardsfolder/upcoming/snarlfang_vermin.txt
tool4ever: condition typo

forge-gui/res/cardsfolder/upcoming/sproutwatch_dryad.txt
```diff
@@ -0,0 +1,8 @@
+Name:Sproutwatch Dryad
+ManaCost:1 G G
+Types:Creature Dryad
+PT:2/3
+T:Mode$ Phase | Phase$ BeginCombat | TriggerZones$ Battlefield | Execute$ TrigPump | TriggerDescription$ At the beginning of each combat, CARDNAME gains flying until end of turn if a creature you control or a card in your hand has flying. The same is true for first strike, double strike, deathtouch, haste, hexproof, indestructible, lifelink, menace, reach, trample, and vigilance.
+SVar:TrigPump:DB$ PumpAll | ValidCards$ Card.Self | KW$ First Strike & Flying & Deathtouch & Double Strike & Haste & Hexproof & Indestructible & Lifelink & Menace & Reach & Trample & Vigilance | SharedKeywordsZone$ Hand | SharedRestrictions$ Creature.YouCtrl
```
tool4ever, line 6: zone incomplete...

forge-gui/res/cardsfolder/upcoming/snarlfang_vermin.txt
tool4ever: also RememberAltered doesn't work like you want

Simisays: > also RememberAltered doesn't work like you want

What do you mean ? Is this not correct ?

tool4ever: not if it's already suspected

Simisays: Is the current version correct ?

Simisays: So there shoudn't be a condition on the animate. I thought the oracle meant that the suspected creature gains this ability ?

tool4ever: no, I said **different condition**
there's only one here which is the "inverening if"

Simisays: Ok thanks for explaining

forge-gui/res/cardsfolder/upcoming/snarlfang_vermin.txt
```diff
@@ -0,0 +1,11 @@
+Name:Snarlfang Vermin
+ManaCost:B
+Types:Creature Rat
+PT:2/1
+T:Mode$ DamageDone | CombatDamage$ True | ValidSource$ Card.Self | ValidTarget$ Creature | TriggerZones$ Battlefield | Execute$ TrigSuspect | TriggerDescription$ Whenever CARDNAME deals combat damage to a creature, if that creature is still on the battlefield, suspect that creature. It perpetually gains this ability.
+SVar:TrigSuspect:DB$ AlterAttribute | Defined$ TriggeredTargetLKICopy | ConditionDefined$ TriggeredTargetLKICopy | ConditionPresent$ Card.inZoneBattlefield | RememberObjects$ TriggeredTargetLKICopy | Attributes$ Suspected | SubAbility$ DBAnimate
```
tool4ever, line 6: RememberObjects makes no sense even if it doesn't exist

forge-gui/res/cardsfolder/upcoming/snarlfang_vermin.txt
```diff
@@ -0,0 +1,11 @@
...
+SVar:DBAnimate:DB$ Animate | Defined$ TriggeredTargetLKICopy | ConditionDefined$ TriggeredTargetLKICopy | ConditionPresent$ Card.inZoneBattlefield | Triggers$ CombatTrig | Duration$ Perpetual
+SVar:CombatTrig:Mode$ DamageDone | CombatDamage$ True | ValidSource$ Card.Self | ValidTarget$ Creature | TriggerZones$ Battlefield | Execute$ TrigSuspect | TriggerDescription$ Whenever CARDNAME deals combat damage to a creature, if that creature is still on the battlefield, suspect that creature. It perpetually gains this ability.
+T:Mode$ ChangesZone | Origin$ Battlefield | PresentZone$ Graveyard | IsPresent$ Card.StrictlySelf | TriggerZones$ Graveyard | Destination$ Graveyard | ValidCard$ Creature.OppCtrl+IsSuspected | Execute$ TrigLoseLife | TriggerDescription$ Whenever a suspected creature an opponent controls dies while CARDNAME is in your graveyard, that opponent loses 1 life.
```
tool4ever, line 9: you scripted intervening if here

Simisays: Do you mean that` IsPresent$ Card.StrictlySelf |` isn't necesairy here ?

tool4ever: You applied this rule:
> 603.4. A triggered ability may read "When/Whenever/At [trigger event], if [condition], [effect]." When the trigger event occurs, the ability checks whether the stated condition is true. The ability triggers only if it is; otherwise it does nothing. If the ability triggers, it checks the stated condition again as it resolves. If the condition isn't true at that time, the ability is removed from the stack and does nothing. Note that this mirrors the check for legal targets. This rule is referred to as the "intervening 'if' clause" rule.

but nothing in your trigger matches this templating, so it's only a normal trigger event instead

Simisays: I'm not sure i'm exactly following how this is scripted like an intervening if. I've used Boneyard Scourge as a template which is worded like this card

tool4ever: yea that one needs to be fixed by me
IsPresent will get checked on resolve so you need NoResolvingCheck

tool4ever: I think I'm only confusing you more at this point since you removed the condition now again :P
but it shouldn't matter too much for Alchemy and my fix there

tool4ever (approved):

PR #4900, opened 2024-03-26
YMKM: concert_kaboomist + support

Northmoc:


forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -88,6 +88,7 @@ public class Player extends GameEntity implements Comparable<Player> {
     private int spellsCastThisTurn;
     private int spellsCastThisGame;
     private int spellsCastLastTurn;
+    private List<Card> spellsCastSinceBeginningOfLastTurn = Lists.newArrayList();
```
tool4ever, line 91: This took a while since I was looking for some synergy with existing stuff ;)

Imo you could refactor it slightly to improve _Arboria_:
> Creatures can’t attack a player unless that player cast a spell or put a nontoken permanent onto the battlefield during their last turn.

Currently both parts only work as bare minimum (static trigger counting) :/

For the casting check it might be easier to consolidate everything in a MagicStack Map <int turnNr, Spells> though.
`Player` already stores a `lastTurnNr`, so with that you could just collect all the relevant entries?

I think we can still get away with periodic cleanup for memory, just the lower bound now being the minimum of each players last turn number (instead of the global last)

For the other check I was concerned we'd need to store additional `cardsAddedLastTurn` fields in `Zone` in the same way.
However there's a subtle difference:
> Arboria’s effect cares whether a player put a nontoken permanent onto the battlefield. It’s unusual for an ability to care who put a permanent onto the battlefield, as opposed to under whose control it entered.

This can matter if you look for effects like this:
> When Petradon leaves the battlefield, return the exiled cards to the battlefield under their owners’ control.

➡️ Here the trigger controller moves the cards, but they can ETB under another.

Since this seems to be also pretty unique it might be easier to have this counted simply as boolean.

While changing this PR for that obscure World would be helpful, feel free to resolve via a new issue after (or even before hehe).

WDYT?

Northmoc: It would be nice to streamline Arboria after having a look at that

I thought of generating some kind of map for all the turns but I couldn't believe we'd need to go much farther back than this... to which WotC probably says "hold my beer"

tool4ever: when you read it closely it's not "farther" imo?
just a bit different since it's not "since"

Agetian:
Is this implementation acceptable as an initial implementation or is more work required for the merge?

tool4ever:
I would have liked a cleaner implementation but it's no big deal if Northmoc is busy for now

Agetian:
Yeah, if it works except for the possible corner cases, personally I'd vote for conflict resolution and merge on this one, with the followup work brought up as an issue to finish the cleaner implementation later if viable. It becomes progressively more difficult to rebase and tweak these branches as code gets updated, especially if the original contributor is not available for the job :/

Agetian:
Attempted to resolve a conflict on this MR.

Northmoc:
Agh, sorry guys
As you have both probably noticed I have periods where I'm not as available for Forge stuff
Always feel free to @ me on Discord for a rebase
Sorry I let this get stale and for any inconvenience Agetian

Agetian:
Sure thing, no worries at all, Northmoc! I also have downtime periods like that for sure, hehe :D

PR #4907, opened 2024-03-26
Landwalk static

### LINKED ISSUE #2507: Landwalk and related ignore abilities, also CantBlockBy Rework for DefendingPlayer
* make Landwalk use CardFactoryUtil using StaticAbility CantBlockBy
* probably refactor the keyword into a more flexible: `K:Landwalk:Land.Island:Islandwalk` ?
* update the CanIgnoreLandwalk into a StaticAbility. Most of them are Static anyway (affecting all creatures?) while one is an Aura affecting Enchanted Creature

Maybe extend CantBlockBy to allow `cantBlock(attacker, defendingPlayer)` in a way to check? How many are these?
* `Can't be blocked` without Param for blocking or defending player?
* `Can't be blocked by defending Player` there i want to add a new check where Landwalk fits in and some others?
* `Can't be blocked by creature` that is the general Creature Check

if i can slot it in like this, it might cleanup some checks for the third static?

Most Can't Block is related to an Attacker, so an extra check for a generic `Can't Block` before shouldn't be needed

* move Can't Block info from Keyword Text into Hidden Keyword Text for CardView, some players where confused by it.

Edit:
* `Street Savvy` makes it that i can't just move all Landwalk related code into the second part, so even if i want to update it i still need to cheat a bit

Hanmac:
Closes #2507 

I want to update _Magnigoth Treefolk_ using one Static without the use of Basic Land Type names


forge-game/src/main/java/forge/game/keyword/Keyword.java
```diff
@@ -112,7 +112,7 @@ public enum Keyword {
     INTIMIDATE("Intimidate", SimpleKeyword.class, true, "This creature can't be blocked except by artifact creatures and/or creatures that share a color with it."),
     KICKER("Kicker", Kicker.class, false, "You may pay an additional %s as you cast this spell."),
     JUMP_START("Jump-start", SimpleKeyword.class, false, "You may cast this card from your graveyard by discarding a card in addition to paying its other costs. Then exile this card."),
-    LANDWALK("Landwalk", KeywordWithType.class, false, "This creature is unblockable as long as defending player controls a %s."),
+    LANDWALK("Landwalk", KeywordWithType.class, false, "This creature is unblockable as long as defending player controls {1:%s}."),
```
tool4ever, line 115: Can `KeywordInstance.redundant()` logic be improved to check for type too now?

might be good for GUI + engine checks

> 702.14e. Multiple instances of the same kind of landwalk on the same creature are redundant.

Hanmac: we probably might need to find a better logic for "redundant" checks

Hanmac: i refactored the Redundant Check

tool4ever (approved):
LGTM

PR #4912, opened 2024-03-27
tweak for Ward keyword display

### LINKED ISSUE #4911: Ward keyword reminder text tweaks
Ovika and Gisa (OTJ) have more complex Ward costs which create a weird reminder text in Card Detail pane
Probably the best thing to do is just drop reminder text in those situations?

Northmoc:
- closes #4911 

This should be fairly future-proof 🤞 

OLD
image

NEW!
image


Hanmac:
some of them have reminder text for complex ward costs

but I'm okay if we hide it

Northmoc:
> some of them have reminder text for complex ward costs
> 
> but I'm okay if we hide it

Good point. "Complex" should be "X cost" OR "mana + other"
Other alone often does have reminder text
This should handle it I think

tool4ever (approved):

PR #4937, opened 2024-04-01
Fix DefinedLandwalk

tool4ever:


Hanmac (approved):

Agetian (approved):

Hanmac:
@tool4ever can we look at Magnigoth Treefolk too? i want to do that without basic land types, but i forgot (the Keyword should be generated by StaticAbilityContinuous)

PR #4938, opened 2024-04-01
CI: Test on all current Java LTS versions (8, 11 ,17, and 21)

allentiak:
Tests on JDK 17 and 21 are optional (since they currently fail).

The very existence of these tests will increase the awareness on the importance of code compatibility with newer Java versions, thus enabling future transitions to be implemented gradually.

Making the code run on JDK 11 took a lot of hacking. This will reduce the effort for eventual future transitions.

### DIFF
--- a/.github/workflows/test-build.yaml
+++ b/.github/workflows/test-build.yaml
@@ -5,15 +5,23 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
+      fail-fast: true
       matrix:
-        java: [ '8', '11' ]
-    name: Test with Java ${{ matrix.Java }}
+        java: ['8', '11']
+        experimental: [false]
+        include:
+          - java: '17'
+            experimental: true
+          - java: '21'
+            experimental: true
+    name: Test with Java ${{ matrix.java }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Setup java
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
           java-version: ${{ matrix.java }}


tool4ever:
Looks like it's TestNG causing problems?
Would they even offer a version that can support 4 JDK simultaneously... 🤔 
Otherwise I think seeing all PR fail (partly) is too noisy and confusing for contributors.

tehdiplomat:
Don't you need to run your test with add-opens?

https://github.com/Card-Forge/forge/blob/152ad0b0afe33502ad4cdb5eacb20cb9b3b196f1/forge-gui-desktop/src/main/config/forge.sh#L44-L52

allentiak:
> Don't you need to run your test with add-opens?

Wouldn't this hide the fact that the current code does not compile with JDK 17+?

I think that it is important to be explicitly aware of this...
https://medium.com/continuousdelivery/if-it-hurts-do-it-more-often-f5a00cc12ffa

If we are not even aware of this problem, then no one will see an incentive to fix it...

As I said above, I think that just this awareness in itself makes merging this worth it...

Hanmac:
> Wouldn't this hide the fact that the current code does not compile with JDK 17+?

it would. if using the right ADD-OPENS parameters

allentiak:
Well, being explicit about that the code won't compile unless that hack is used is part of the "awareness raising" rationale I mentioned when I first opened this PR...

allentiak:
I mean, the first step to solve a problem is to accept its existence, right?

Hanmac:
> I mean, the first step to solve a problem is to accept its existence, right? https://jhall.io/archive/2022/06/11/if-it-hurts-do-it-more/

The "pain" you are referring too is JAVA 17+ messing up our Code we depend on it, and can't really update

tehdiplomat:
Here's my suggestion for this PR. Instead of having these failing (optional) tests run every single PR which is a lot of noise and Github Actions CPU cycles. And potentially very noisy for new people who don't understand that this is expected behavior, why don't we associate this change only during release builds that way we can keep it in mind if there's a future way to avoid add-opens, but its not just noise for the sake of noise.

PR #4942, opened 2024-04-02
Replace TestNG with Junit

allentiak:


Hanmac:
You would need to update the other Tests from `testng` too

PR #4947, opened 2024-04-03
OTJ: more saddle support and cards

Northmoc:


Northmoc:
Not sure why my rebase got tangled up with those Dracontes commits 🤷

forge-gui/res/cardsfolder/r/retrieve.txt
```diff
@@ -2,6 +2,6 @@ Name:Retrieve
 ManaCost:2 G
 Types:Sorcery
 A:SP$ ChangeZone | Origin$ Graveyard | Destination$ Hand | TgtPrompt$ Select up to one target creature card in your graveyard | TargetMin$ 0 | TargetMax$ 1 | ValidTgts$ Creature.YouOwn | SubAbility$ DBReturn | StackDescription$ SpellDescription | SpellDescription$ Return up to one target creature card
-SVar:DBReturn:DB$ ChangeZone | Origin$ Graveyard | Destination$ Hand | TargetMin$ 0 | TargetMax$ 1 | ValidTgts$ Land.YouOwn | TgtPrompt$ Select up to one target noncreature permanent card in your graveyard | SubAbility$ DBExile | StackDescription$ SpellDescription | SpellDescription$ and up to one target noncreature permanent card
+SVar:DBReturn:DB$ ChangeZone | Origin$ Graveyard | Destination$ Hand | TargetMin$ 0 | TargetMax$ 1 | ValidTgts$ Permanent.nonCreature+YouOwn | TgtPrompt$ Select up to one target noncreature permanent card in your graveyard | SubAbility$ DBExile | StackDescription$ SpellDescription | SpellDescription$ and up to one target noncreature permanent card
```
Hanmac, line 5: @tool4ever @Northmoc we probably need to reutrn the cards at the same time?

Northmoc: I do agree, but this PR is probably "big enough" without dealing with that here – there are other cards with the same problem

I'll make a new issue for it

forge-game/src/main/java/forge/game/trigger/TriggerBecomesSaddled.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public boolean performTest(Map<AbilityKey, Object> runParams) {
+        if (!matchesValidParam("ValidSaddled", runParams.get(AbilityKey.Card))) {
+            return false;
+        }
+        if (hasParam("FirstTimeSaddled")) {
+            Card v = (Card) runParams.get(AbilityKey.Card);
+            if (v.getTimesSaddledThisTurn() != 1) {
```
Hanmac, line 23: @tool4ever we had some other problems with such First Time triggers, right? Does this work that way?

tool4ever: well most likely saddling won't start happening simultaneous
so we should be relatively safe

tool4ever (approved):

forge-game/src/main/java/forge/game/card/CardProperty.java
```diff
@@ -1933,6 +1933,10 @@ else if (property.equals("blocked")) {
             if (!card.isSaddled()) {
                 return false;
             }
+        } else if (property.equals("SaddledThisTurn")) {
+            if (!source.getSaddledByThisTurn().contains(card)) {
```
Hanmac, line 1937: This should probably check for game timestamp in case the Creature got blinked?

forge-game/src/main/java/forge/game/card/CardProperty.java
```diff
@@ -1894,9 +1894,7 @@ else if (property.equals("blocked")) {
                 return false;
             }
         } else if (property.equals("CrewedThisTurn")) {
-            if (!source.getCrewedByThisTurn().contains(card)) {
-                return false;
-            }
+            if (!hasTimestampMatch(card, source.getSaddledByThisTurn())) return false;
```
Hanmac, line 1897: Shouldn't that be CrewedThisTurn ?

Hanmac (approved):

PR #4952, opened 2024-04-04
OTJ: luxurious_locomotive + support

Northmoc:


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -2387,14 +2388,22 @@ public final String keywordsToText(final Collection<KeywordInterface> keywords)
                     sbLong.append("/").append(k[3]).append("] ").append("(").append(inst.getReminderText()).append(")");
                 } else if (keyword.startsWith("Modular") || keyword.startsWith("Bloodthirst") || keyword.startsWith("Dredge")
                         || keyword.startsWith("Fabricate") || keyword.startsWith("Soulshift") || keyword.startsWith("Bushido")
-                        || keyword.startsWith("Crew") || keyword.startsWith("Tribute") || keyword.startsWith("Absorb")
+                        || keyword.startsWith("Saddle") || keyword.startsWith("Tribute") || keyword.startsWith("Absorb")
                         || keyword.startsWith("Graft") || keyword.startsWith("Fading") || keyword.startsWith("Vanishing:")
```
dracontes, line 2392: Probably doesn't affect much but I notice `"Vanishing:"` is alone among these in including a colon at the end.

tool4ever: it might be on purpose since the rules support two variants?

Northmoc: Yes, the parsing is a bit different for cards that intrinsically have vanishing (which would have the `:` in the card script) vs cards that gain vanishing as the result of an effect

Hanmac (approved):
The CrewSize might use the xCount List feature, but I think thats good enough for now

tool4ever (approved):

PR #4955, opened 2024-04-04
OTJ: felix_five_boots + support

Northmoc:


forge-game/src/main/java/forge/game/staticability/StaticAbilityPanharmonicon.java
```diff
@@ -157,6 +156,17 @@ public static boolean applyPanharmoniconAbility(final StaticAbility stAb, final
                 return false;
             }
         } else if (trigMode.equals(TriggerType.DamageDone) || trigMode.equals(TriggerType.DamageDoneOnce)) {
+            if (stAb.hasParam("ValidSource")) {
+                if (trigMode.equals(TriggerType.DamageDone) && 
+                    !stAb.matchesValidParam("ValidSource", runParams.get(AbilityKey.DamageSource))) return false;
+                else if (trigMode.equals(TriggerType.DamageDoneOnce) && 
+                    (int) runParams.get(AbilityKey.DamageAmount) <= 0) return false;
+                //DamageAmount is filtered by ValidSource in TriggerDamageDoneOnce
```
tool4ever, line 164: hmn, I have to say I don't quite understand how this comment lets us avoid filtering for the Source?

tool4ever: also isn't that the wrong mode anyway compared to Wayta?

tool4ever: @Hanmac
I believe DamageAll will also need to be included for this one?

Hanmac: yeah, probably DamageAll too

Northmoc: Should [Wayta, Trainer Prodigy](https://scryfall.com/card/lcc/7/wayta-trainer-prodigy) also have DamageAll ?

tool4ever: - I meant DamageDealtOnce instead, make sure you're not mixing up "being dealt" and "dealing"

- it looks like you're trying to use triggeringObjects? that's not how it works (they're not available)

- DamageAll needs a lot more checking, it's a bit tricky to explain but I'll try and give you an example:

Consider a combat scenario in which one of your two attacking creatures gets a warrior blocker where the damage it receives is redirected to the defending player

The final damage event will look like this:
attacker A -> 1+ dmg to Opp
blocker B -> 1+ dmg to Opp

now look at _Mindblade Render_:
> Whenever your opponents are dealt combat damage, if any of that damage was dealt by a Warrior, you draw a card and you lose 1 life.

does he trigger? yes
and does Felix double it? **no!**

because it wasn't actually **your** warrior that caused it

-> you can't check the params in isolation but a single entry in the DamageMap needs to match all of them together

you can also check the work I did for ChangesZoneAll as guideline (luckily you don't need to worry about DisableTriggers for now :P)

- Chances are you _could_ ignore it on Wayta for now because it can't match anyway

Northmoc: Thank you for all this – looks like you're on Discord right now, I might ping you

forge-game/src/main/java/forge/game/staticability/StaticAbilityPanharmonicon.java
```diff
@@ -168,4 +182,15 @@ public static boolean applyPanharmoniconAbility(final StaticAbility stAb, final
 
         return true;
     }
+
+    private static boolean checkEach(final StaticAbility stAb, final Object obj, final String param) {
```
Hanmac, line 186: Check Each shouldn't be needed, matchesValidParam already supports Iterable

forge-game/src/main/java/forge/game/staticability/StaticAbilityPanharmonicon.java
tool4ever: I believe I fixed it so now it checks based on the potential triggeringObjects too
Feel free to test
Imo we don't need to worry about ExcessDamage yet, since it seems impossible to match right now 🤔

Hanmac: are the Damage Trigger ready yet?

tool4ever: I did not have time for testing yet

tool4ever: ok, looks like it behaves the way I want

forge-game/src/main/java/forge/game/staticability/StaticAbilityPanharmonicon.java
```diff
@@ -156,10 +159,55 @@ public static boolean applyPanharmoniconAbility(final StaticAbility stAb, final
...
+                if (!stAb.matchesValidParam("ValidSource", runParams.get(AbilityKey.DamageSource))) {
+                    return false;
+                }
+                Map<Card, Integer> dmgMap = (Map<Card, Integer>) runParams.get(AbilityKey.DamageMap);
+                if (!Iterables.any(dmgMap.keySet(), Predicates.and(
+                        CardPredicates.matchesValidParam(stAb, "ValidTarget"),
+                        CardPredicates.matchesValidParam(trigger, "ValidTarget")
+                        ))) {
+                    return false;
+                }
```
Hanmac, line 201: @tool4ever DamageDealtOnce uses this Map `Map<GameEntity, Integer>`

meaning that `CardPredicates.matchesValidParam` should probably be moved to `GameObjectPredicates.matchesValidParam`

tool4ever: right, good catch

Hanmac (approved):

PR #4965, opened 2024-04-06
Create razzle-dazzler

loud1990:


forge-gui/res/cardsfolder/upcoming/razzle-dazzler
```diff
@@ -0,0 +1,10 @@
+Name:Razzle-Dazzler
```
tool4ever, line 1: filename incomplete

Northmoc: `-` needs to be `_` also
so `razzle_dazzler.txt`

tool4ever (approved):

PR #4977, opened 2024-04-07
Android manifest update

Hanmac:
this moves the Android Manifest to the more Default Location for the plugins

and enables `manifest-merger` goal.

This sets the VersionName and generates the VersionCode from it


Agetian (approved):

forge-gui-android/pom.xml
```diff
@@ -204,18 +204,17 @@
...
+                        <version>4.6.0</version>
                         <extensions>true</extensions>
                         <configuration>
                             <sign>
                                 <debug>true</debug>
                             </sign>
                             <sdk>
                                 <platform>26</platform>
-                                <buildTools>30.0.3</buildTools>
+                                <buildTools>34.0.0</buildTools>
```
tehdiplomat, line 215: This might be problematic with android build?

tehdiplomat: https://github.com/Card-Forge/forge/blob/master/.github/workflows/snapshots-android.yml#L50

Hanmac: Yeah I need to update this too


I wonder why github doesn't have a problem with d8, locally I needed to fix something, or was that done with older maven version?

Hanmac:
@Agetian maybe you know why on github we don't need to mess with the d8 file?

forge-gui-android/pom.xml
```diff
@@ -204,18 +204,17 @@
                                 <version>6.2.2</version>
                             </dependency>
                         </dependencies>
-                        <version>4.6.1</version>
+                        <version>4.6.0</version>
```
tehdiplomat, line 207: This is wrong. We need to use 4.6.1 

See this section in Android snapshot. 

`      - name: Install Android maven plugin
        run: |
          mkdir  -p ~/.m2/repository/com/simpligility/maven/plugins/android-maven-plugin/4.6.1
          cd ~/.m2/repository/com/simpligility/maven/plugins/android-maven-plugin/4.6.1
          curl -L -o android-maven-plugin-4.6.1.jar https://github.com/Card-Forge/android-maven-plugin/releases/download/4.6.1/android-maven-plugin-4.6.1.jar
          curl -L -o android-maven-plugin-4.6.1.pom https://github.com/Card-Forge/android-maven-plugin/releases/download/4.6.1/android-maven-plugin-4.6.1.pom`

tehdiplomat: I believe this library change is why we don't need to mess with d8 for the snapshot build.

Hanmac: Ahh that explains why i couldn't find the Version

forge-gui-android/pom.xml
tehdiplomat: Yea i don't remember if it was an issue or not, I know Agetian said thats the version he uses specifically and i had some problems doing a bunch of the different pieces, i don't know if this was one of them. If the snapshot works fine with it. I'm cool with updating. But maybe we also want to separate those changes to reduce variables.

Agetian:
> @Agetian maybe you know why on github we don't need to mess with the d8 file?

Hmm, not sure tbh, I haven't investigated the Github way of doing things as far as building and deploying an Android snapshot goes ^^;

Hanmac:
@Agetian can you look to replace

`Forge.CURRENT_VERSION` in there:
https://github.com/Card-Forge/forge/blob/e2f4c2cfa7a551a215192334bdfc7cfed677b01d/forge-gui-mobile/src/forge/Forge.java#L57

`BuildInfo.getVersionString` ?
https://github.com/Card-Forge/forge/blob/e2f4c2cfa7a551a215192334bdfc7cfed677b01d/forge-core/src/main/java/forge/util/BuildInfo.java#L35

that is one of the last pieces where the Version needs to be updated

Agetian:
Hmm, getVersionString simply returns "GIT" for me (on a snapshot version), with the entire BuildInfo.class.getPackage() method returning "null"... Not sure if it would return a proper version string if executed on an actual release. At any rate, kinda confusing for snapshots as such. I wonder if there's a better way to track the number, hmm...

Hanmac:
> Hmm, getVersionString simply returns "GIT" for me (on a snapshot version), with the entire BuildInfo.class.getPackage() method returning "null"... Not sure if it would return a proper version string if executed on an actual release. At any rate, kinda confusing for snapshots as such. I wonder if there's a better way to track the number, hmm...

Normally, maven should have put the version inside these Manifest files so it can be accessed by `getPackage()`
I need to check what is missing

Hanmac:
I think we need https://maven.apache.org/shared/maven-archiver/examples/manifest.html

```xml
     <configuration>
         <archive>
             <manifest>
                 <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
             </manifest>
         </archive>
     </configuration>
```

i will check this out at weekend

Agetian:
Sounds good!

Hanmac:
@Agetian the addDefaultImplementationEntries should already be enabled by default

when i run `./forge.sh`  i already got the correct Forge Version


As for the Android thing, i tried:

`JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64/ mvn -U -B -P android-debug install -Dsign.keystore=forge.keystore -Dsign.alias=Forge -Dandroid.sdk.path=/opt/android-sdk/ -T 1C  -DskipTests=true`

and i get:
```
[ERROR] Failed to execute goal com.simpligility.maven.plugins:android-maven-plugin:4.6.1:manifest-merger (update-manifest) on project forge-gui-android: Execution update-manifest of goal com.simpligility.maven.plugins:android-maven-plugin:4.6.1:manifest-merger failed: FeatureName must follow [a-zA-Z0-9][a-zA-Z0-9_]* regex, found -> [Help 1]
```

but i can't find where the problem is happening (it is difficult to debug inside the maven plugin)

Hanmac:
Linked Issue that is causing the problem:
https://github.com/simpligility/android-maven-plugin/issues/814

Hanmac:
@kevlahnota i don't understand the config there yet:
https://github.com/Card-Forge/android-maven-plugin/blob/b36d05b798f3b5cf1362795f008171309183678e/src/main/java/com/simpligility/maven/plugins/android/standalonemojos/ManifestMergerMojo.java#L120

right now it does reset the VersionCode to 1 :thinking:

Hanmac:
@kevlahnota should i try to apply this Commit into this MR too? https://github.com/Card-Forge/forge/pull/6103/commits/ad81b0e32d8366fa6a67b0b249b738824d03d997

kevlahnota:
> @kevlahnota should i try to apply this Commit into this MR too? [ad81b0e](https://github.com/Card-Forge/forge/commit/ad81b0e32d8366fa6a67b0b249b738824d03d997)

Hmm the code there will just remove all non digit from revision - 2.0.00-snapshot, to 2000 and append 0 and MMdd where MM for month and dd for the date so the versioncode if date today is october 10, it will be 200001010 as version code (also append MMdd to the finalname), I don't know if it will work on your MR

Hanmac:
> @kevlahnota i don't understand the config there yet: https://github.com/Card-Forge/android-maven-plugin/blob/b36d05b798f3b5cf1362795f008171309183678e/src/main/java/com/simpligility/maven/plugins/android/standalonemojos/ManifestMergerMojo.java#L120
> 
> right now it does reset the VersionCode to 1 🤔

@kevlahnota like i said, right now the `manifestVersionCodeUpdateFromVersion` doesn't seem to work for me, and the VersionCode in the Manifest is always set to "1".

But i don't understand yet why it does happen, and have no idea on how to debug it

Hanmac:
@kevlahnota this code right there should have updated the VersionCode

https://github.com/Card-Forge/android-maven-plugin/blob/41bbf85159145c51cb922dde9618c9cf6f138286/src/main/java/com/simpligility/maven/plugins/android/standalonemojos/ManifestMergerMojo.java#L312-L317

in theory, it should have turned `1.6.65-SNAPSHOT` into `1006065`

but for some reason it doesn't, and I don't know how to debug the maven plugin

kevlahnota:
> @kevlahnota this code right there should have updated the VersionCode
> 
> https://github.com/Card-Forge/android-maven-plugin/blob/41bbf85159145c51cb922dde9618c9cf6f138286/src/main/java/com/simpligility/maven/plugins/android/standalonemojos/ManifestMergerMojo.java#L312-L317
> 
> in theory, it should have turned `1.6.65-SNAPSHOT` into `1006065`
> 
> but for some reason it doesn't, and I don't know how to debug the maven plugin
> 
> not really for snapshots, but that should have been good enough?

did you try to pass the alpha-version?

```
<configuration>
    <manifestVersionName>${alpha-version}</manifestVersionName>
    <manifestVersionCodeUpdateFromVersion>true</manifestVersionCodeUpdateFromVersion>
</configuration>
```

Hanmac:
> > @kevlahnota this code right there should have updated the VersionCode
> > https://github.com/Card-Forge/android-maven-plugin/blob/41bbf85159145c51cb922dde9618c9cf6f138286/src/main/java/com/simpligility/maven/plugins/android/standalonemojos/ManifestMergerMojo.java#L312-L317
> > in theory, it should have turned `1.6.65-SNAPSHOT` into `1006065`
> > but for some reason it doesn't, and I don't know how to debug the maven plugin
> > not really for snapshots, but that should have been good enough?
> 
> did you try to pass the alpha-version?
> 
> ```
> <configuration>
>     <manifestVersionName>${alpha-version}</manifestVersionName>
>     <manifestVersionCodeUpdateFromVersion>true</manifestVersionCodeUpdateFromVersion>
> </configuration>
> ```

The default should be `${project.version}`, i will test `${alpha-version}` as soon as i get home

Hanmac:
@kevlahnota on my System, i had problems with the working directory.

It only works correct when calling it in `forge-gui-android`, otherwise the parameter gets messed up?

kevlahnota:
> @kevlahnota on my System, i had problems with the working directory.
> 
> It only works correct when calling it in `forge-gui-android`, otherwise the parameter gets messed up?

I can't figure it out on my system too since it always missing 4 digits that's why I resort to use the build-helper-maven-plugin (timestamp-property for appending monthdate code, regex-property to retain only numeric characters from revision, still missing 1 digit character so I append `0` before the monthdate code `${snapshot-versionCode}0${month.date}` on my PR)

Hanmac:
> missing 4 digits

That's why I added this:
`<android.manifestMerger.versionDigits>2,2,2,4</android.manifestMerger.versionDigits>`

With that, even `1.2.3` is turned into `102030000`, so you can use the last 4 digits for snapshot stuff

tehdiplomat (approved):

PR #4981, opened 2024-04-08
Update Forge.java

tehdiplomat:


Hanmac (approved):

Hanmac:
@Agetian @tehdiplomat we need to think about if we really need this CURRENT_VERSION for Mobile,
or if we can use this instead: 
https://github.com/Card-Forge/forge/blob/9eb50f441a7a5bba2613645f9d6b5d9768c24470/forge-core/src/main/java/forge/util/BuildInfo.java#L35-L41

PR #4986, opened 2024-04-09
Assist

### LINKED ISSUE #3121: Assist
> 702.132. Assist
> 702.132a Assist is a static ability that modifies the rules of paying for the spell with assist (see rules 601.2g-h). If the total cost to cast a spell with assist includes a generic mana component, before you activate mana abilities while casting it, you may choose another player. That player has a chance to activate mana abilities. Once that player chooses not to activate any more mana abilities, you have a chance to activate mana abilities. Before you begin to pay the total cost of the spell, the player you chose may pay for any amount of the generic mana in the spell’s total cost.

Some notes to think about it:

- It shouldn't reduce the amount of mana spent to cast this spell. Or otherwise, the the mana reduced for you need to be added later
- We probably need to refactor "mana is spent" cards like Pyromancer's Goggles with: `When that mana is spent to cast a red instant or sorcery spell, copy that spell and you may choose new targets for the copy.` like you are not spending it, but you still should get the trigger? Also it still should respect the Mana Restrictions
- AI should consider Assist and possible team members when choosing a value for X

tehdiplomat:
Closes #3121

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -1897,27 +1897,27 @@ public class AiController {
 
         PlayerCollection allies = player.getAllies();
 
-        if (allies.size() > 0) {
-            // AI has allies, don't help out anyone but allies.
-            if (!allies.contains(activator)) {
-                return 0;
-            }
-        } else {
+        if (allies.isEmpty()) {
             // AI only has opponents.
             // TODO: Maybe help out someone if it seems good for us, but who knows how you calculate that.
             // Probably needs some specific AI here.
             // If the spell is a creature, probably don't help.
             // If spell is a instant/sorcery, help based on the situation
             return 0;
+        } else {
+            // AI has allies, don't help out anyone but allies.
+            if (!allies.contains(activator)) {
+                return 0;
+            }
         }
 
         // AI has decided to help. Now let's figure out how much they can help
-        int mana = ComputerUtilMana.getAvailableManaEstimate(player, false);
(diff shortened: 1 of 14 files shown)


Hanmac:
@tehdiplomat @tool4ever for #3121 we need to check some SpentMana trigger. like _Pyromancer's Goggles_ 

also it needs to respect Mana Restrictions

forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -1886,6 +1887,53 @@ public boolean confirmPayment(CostPart costPart) {
...
+            // Never help your opponent in a 2 player game
+            return 0;
+        }
+
+        PlayerCollection allies = game.getPlayers().filter(new Predicate<Player>() {
+            @Override
+            public boolean apply(Player pl) {
+                return !pl.isOpponentOf(player);
+            }
+        });
```
Hanmac, line 1903: wouldn't it be easier to check `player.getAllies()` ?

tool4ever:
> @tehdiplomat @tool4ever for #3121 we need to check some SpentMana trigger. like _Pyromancer's Goggles_
> 
> also it needs to respect Mana Restrictions

Those without ValidActivatingPlayer$ like _Pyromancer's Goggles_ work already (the other player will control the copy)

Also restriction from _Ally Encampment_ was enforced in my test

paulsnoops:
Just a couple of small things in the card detail panel from my testing:
- Assist doesn't show reminder text (intentional?)
- The word Assist will only show on Creatures. It wasn't present on Instants/Sorceries.

tool4ever (approved):
looks like it could be shipped since the basics are covered
nice work

PR #4987, opened 2024-04-09
Four new OTJ cards

nthoron:
Added Resilient Roadrunner, Raucous Entertainer, Reach for the Sky, and Redrock Sentinel.


tool4ever (approved):
nice work

PR #4989, opened 2024-04-09
Fix advanced Flash logic messing with PlayEffectAi

Agetian:
E.g. Discover X or a transformed Battle - assume that an out-of-timing permanent cast without paying its mana cost is good enough to not rely on advanced Flash considerations.


tool4ever (dismissed):
not ideal but good enough as hotfix?

Hanmac:
@Agetian @tool4ever maybe look into `isCastFromPlayEffect` for when to Ignore Timing?

Hanmac (approved):

PR #4990, opened 2024-04-09
AI Pump tweaks for abilities with a sacrifice self cost + Witch's Clinic AI hint.

Agetian:


Hanmac:
@Agetian you probably need to rebase your MR ?

Agetian:
Hmm, that's what I did, and it ended up in this weird state :/

Agetian:
Ok, second time's the charm :D

Hanmac:
my main concern with "Witch's Clinic", why does AI want to give my Creatures (or even non Creatures?) Lifelink?

shouldn't that be better handled with the Pump Logic?

Agetian:
> my main concern with "Witch's Clinic", why does AI want to give my Creatures (or even non Creatures?) Lifelink?
> shouldn't that be better handled with the Pump Logic?

Most likely yes, but it'll most likely involve a custom AI logic of some sort anyway, since it doesn't go through the regular "Curse / non-Curse" path which is used for creatures. At first I thought about extending that logic for basically everything, but can't quite figure out what the repercussions would be for other potential cards (where non-creatures are pumped), since Pump has a really wide variety of use cases for non-creatures :/

tool4ever (approved):
seems ok, some people would complain if we'd limit the Clinic to creatures only anyway ;)

PR #4995, opened 2024-04-09
Fix mana doublers not working during payment

tool4ever:
Caused by the freeze changes in #4794

Maybe there's a better long term fix if you can find one?

### DIFF
--- a/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java
+++ b/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java
@@ -266,7 +266,7 @@ public class TriggerHandler {
 
         if (mode == TriggerType.Always) {
             runStateTrigger(runParams);
-        } else if (game.getStack().isFrozen() || holdTrigger) {
+        } else if ((game.getStack().isFrozen() || holdTrigger) && mode != TriggerType.TapsForMana && mode != TriggerType.ManaAdded) {
             waitingTriggers.add(new TriggerWaiting(mode, runParams));
         } else {
             runWaitingTrigger(new TriggerWaiting(mode, runParams));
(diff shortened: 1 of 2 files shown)


tehdiplomat:
Whats the repro?

tool4ever:
Uhh basically any SA that you want to also pay partly with mana from the doubled parts:
image

To be precise, it looks like after/(if) you finish casting you will get what you're owed in the pool by then

tehdiplomat:
I'm guessing the "true" resolution should be where we check if the stack is frozen and if we should ignore frozen. We should also resolve it for mana triggers.

tehdiplomat:
Yea i'm confused why those wouldn't go straight through here:

https://github.com/Card-Forge/forge/blob/master/forge-game/src/main/java/forge/game/zone/MagicStack.java#L262-L285

tool4ever:
While the stack is frozen trigger events are only collected until we reach a stable state again (that's probably simplifying things a bit but the main idea)

tehdiplomat:
I'll try to look tonight for a better long term fix

forge-game/src/main/java/forge/game/trigger/TriggerHandler.java
```diff
@@ -266,7 +266,7 @@ public final void runTrigger(final TriggerType mode, final Map<AbilityKey, Objec
 
         if (mode == TriggerType.Always) {
             runStateTrigger(runParams);
-        } else if (game.getStack().isFrozen() || holdTrigger) {
+        } else if ((game.getStack().isFrozen() || holdTrigger) && mode != TriggerType.TapsForMana && mode != TriggerType.ManaAdded) {
```
tehdiplomat, line 269: Oh i see, we're not even getting to MagicStack before they get held.  We need to test if the trigger is a mana ability here to run it right away. We can merge for now and i'll try to clean it up soon.

PR #5010, opened 2024-04-11
OTJ: more spree cards and polish

Northmoc:
Added a little UI tweak for Spree targeting since sometimes we're selecting a target for each mode
Just in case the human doing the targeting forgets what they are doing 😁

### DIFF
--- a/forge-gui/res/cardsfolder/c/collective_effort.txt
+++ b/forge-gui/res/cardsfolder/c/collective_effort.txt
@@ -5,5 +5,5 @@ K:Escalate:tapXType<1/Creature>
 A:SP$ Charm | Cost$ 1 W W | MinCharmNum$ 1 | CharmNum$ 3 | Choices$ DBDestroyCreature,DBDestroyEnchantment,DBPutCounterAll
 SVar:DBDestroyCreature:DB$ Destroy | ValidTgts$ Creature.powerGE4 | TgtPrompt$ Select target creature with power 4 or greater | SpellDescription$ Destroy target creature with power 4 or greater.
 SVar:DBDestroyEnchantment:DB$ Destroy | ValidTgts$ Enchantment | TgtPrompt$ Select target enchantment | SpellDescription$ Destroy target enchantment.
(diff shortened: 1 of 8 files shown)


forge-game/src/main/java/forge/game/ability/AbilityFactory.java
```diff
@@ -352,6 +352,11 @@ private static final TargetRestrictions readTarget(Map<String, String> mapParams
             prompt = "Select target " + tgtWhat;
         }
 
+        if (mapParams.containsKey("SpreeCost")) {
```
tool4ever, line 355: hmn I don't understand...
If they don't have SpellDescription the card would look wonky anyway:
Or are you thinking about a sub only - then a fallback might be better?

Hanmac (approved):

forge-game/src/main/java/forge/game/ability/AbilityFactory.java
tool4ever: Sorry, I still don't understand why Spree needs special treament?

It already cuts off the already targeted parts, so now we'd be showing duplicated text for no reason?

grafik

Northmoc: I'm glad you keep asking questions, don't want to change anything if it doesn't make sense.

I wired it up because unlike you I haven't selected this pref
image
Without that pref things can be a bit more confusing
I believe that pref is off by default / on a new install

I'm not sure which path is best here 🤔

Northmoc: Ah looks like it is on by default.

Easy enough, if you want that detail you should have the pref selected

Hanmac (approved):

forge-game/src/main/java/forge/game/ability/AbilityFactory.java
tool4ever: Oh I see
well since there's an option for it hardcoding kind of a partial overwrite here seems even less favorable?
(I'm not saying the feature couldn't be improved further but it feels weird to "hide" this part here - 
Also kind of takes away the control of the scripter to decide if TgtPrompt$ is needed)
We could change that option to be default in fresh install though?
Kind of makes sense to roll out with the most beginner friendly settings

tool4ever (approved):

PR #5013, opened 2024-04-11
OTJ: obeka_splitter_of_seconds + support

Northmoc:


forge-game/src/main/java/forge/game/ability/effects/AddPhaseEffect.java
```diff
@@ -39,36 +42,41 @@ public void resolve(SpellAbility sa) {
...
+            List<PhaseType> extraPhaseList = new ArrayList<>();
 
-        // Insert FollowedBy
-        if (sa.hasParam("FollowedBy")) {
-            // Currently all FollowedBy are Main2 phase, which has no step
-            PhaseType followingExtra = PhaseType.smartValueOf(sa.getParam("FollowedBy"));
-            extraPhaseList.add(followingExtra);
-        }
+            // Insert ExtraPhase
+            String extra = sa.getParam("ExtraPhase");
```
tehdiplomat, line 53: Let's pull this above the loop.

forge-game/src/main/java/forge/game/ability/effects/AddPhaseEffect.java
```diff
@@ -39,36 +42,41 @@ public void resolve(SpellAbility sa) {
...
+                extraPhaseList.addAll(PhaseType.PHASE_GROUPS.get(2));
+            } else { // Currently no effect will add End Phase
+                extraPhaseList.add(PhaseType.smartValueOf(extra));
+            }
 
-        ExtraPhase extraPhase = phaseHandler.addExtraPhase(afterPhase, extraPhaseList, nextPhase);
+            // Insert FollowedBy
+            if (sa.hasParam("FollowedBy")) {
+                // Currently all FollowedBy are Main2 phase, which has no step
+                PhaseType followingExtra = PhaseType.smartValueOf(sa.getParam("FollowedBy"));
```
tehdiplomat, line 65: We can pull the sa.getParam above the loop too. Then just check if followingExtra isn't null here instead.

forge-game/src/main/java/forge/game/ability/effects/AddPhaseEffect.java
```diff
@@ -39,36 +42,41 @@ public void resolve(SpellAbility sa) {
...
-        // Insert ExtraPhase
-        String extra = sa.getParam("ExtraPhase");
-        if (extra.equals("Beginning")) {
-            extraPhaseList.addAll(PhaseType.PHASE_GROUPS.get(0));
-        } else if (extra.equals("Combat")) {
-            extraPhaseList.addAll(PhaseType.PHASE_GROUPS.get(2));
-        } else { // Currently no effect will add End Phase
-            extraPhaseList.add(PhaseType.smartValueOf(extra));
-        }
+        while (num > 0) {
```
tehdiplomat, line 47: This is probably better as a for loop

`for(int n = num; n > 0; n--)`

Hanmac:
Need to add check that you only should be able to get extra phases on your turn (new rule)

forge-game/src/main/java/forge/game/ability/effects/AddPhaseEffect.java
```diff
@@ -20,8 +23,14 @@ public class AddPhaseEffect extends SpellAbilityEffect {
...
     public void resolve(SpellAbility sa) {
-        boolean isTopsy = sa.getActivatingPlayer().getAmountOfKeyword("The phases of your turn are reversed.") % 2 == 1;
-        PhaseHandler phaseHandler = sa.getActivatingPlayer().getGame().getPhaseHandler();
+        final Card host = sa.getHostCard();
+        final Player activator = sa.getActivatingPlayer();
+        boolean isTopsy = activator.getAmountOfKeyword("The phases of your turn are reversed.") % 2 == 1;
+        PhaseHandler phaseHandler = activator.getGame().getPhaseHandler();
+
+        //You can only add extra phases to your turn
+        if (!phaseHandler.isPlayerTurn(activator)) return;
```
tool4ever, line 32: that check is way too broad
you need to implement it by the rules:
> 500.10a If an effect that says “you get” an additional step or phase would add a step or phase to a turn other than its controller’s, no steps or phases are added.

tool4ever: nooo
this card is the only one that says that
so please don't "break" all others :P

Hanmac: technically all others also would just give you extra phases on your turns

so checking for activating player should be okay

tool4ever: @Hanmac 
no, check e. g. _Relentless Assault_
>  If you manage to cast this during a main phase of your opponent's turn, that opponent's creatures will untap and that opponent will be able to attack again. It will not allow you to attack during their turn.

tehdiplomat: It kinda sounds like we need a param to specify ownership of the phase, and return/continue if it doesn't fit?

tool4ever: or `ConditionPlayerTurn$`

Hanmac (approved):

forge-gui/res/cardsfolder/upcoming/obeka_splitter_of_seconds.txt
```diff
@@ -0,0 +1,8 @@
+Name:Obeka, Splitter of Seconds
+ManaCost:1 U B R
+Types:Legendary Creature Ogre Warlock
+PT:2/5
+K:Menace
+T:Mode$ DamageDone | ValidSource$ Card.Self | ValidTarget$ Player | Execute$ TrigAddPhase | CombatDamage$ True | TriggerDescription$ Whenever CARDNAME deals combat damage to a player, you get that many additional upkeep steps after this phase.
+SVar:TrigAddPhase:DB$ AddPhase | ConditionPlayerTurn$ True | NumPhases$ TriggerCount$DamageAmount | ExtraPhase$ Upkeep
```
tool4ever, line 7: you're still inserting it too early, maybe AfterPhase$?

> The beginning phases containing the additional upkeep steps occur after the current combat phase ends. The last step of the combat phase is the end of combat step.

tool4ever (approved):

Hanmac (approved):

PR #5016, opened 2024-04-11
Animate effect cleanup pass #2

dracontes:
Continuing from https://github.com/Card-Forge/forge/pull/5006 , here's the creature type overwriting pass for self-animators, triggered and activated. I have still to wrap my head around the static ability ones. 
As matters are before this cleanup, in the corner case of activating the second loyalty ability of [Arlinn, the Moon's Fury](https://scryfall.com/card/mid/211/arlinn-the-packs-hope-arlinn-the-moons-fury) after activating the first loyalty ability of [Sarkhan the Masterless](https://scryfall.com/card/cmm/880/sarkhan-the-masterless), it's interesting to note that, despite the card type overwrite on both, the granted subtypes remain as the effects resolve. I'm not sure if that's how it should work conceptually, even if I like the granularity of the script in practice considering how often these effects aren't exactly the same. 

animated_arlinn


forge-gui/res/cardsfolder/a/angels_tomb.txt
```diff
@@ -2,6 +2,6 @@ Name:Angel's Tomb
 ManaCost:3
 Types:Artifact
 T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Creature.YouCtrl | TriggerZones$ Battlefield | Execute$ TrigAnimateTomb | OptionalDecider$ You | TriggerDescription$ Whenever a creature enters the battlefield under your control, you may have CARDNAME become a 3/3 white Angel artifact creature with flying until end of turn.
-SVar:TrigAnimateTomb:DB$ Animate | Defined$ Self | Power$ 3 | Toughness$ 3 | Types$ Artifact,Creature,Angel | Colors$ White | OverwriteColors$ True | Keywords$ Flying
+SVar:TrigAnimateTomb:DB$ Animate | Defined$ Self | Power$ 3 | Toughness$ 3 | Types$ Artifact,Creature,Angel | RemoveCreatureTypes$ True | Colors$ White | OverwriteColors$ True | Keywords$ Flying
```
tool4ever, line 5: this change is wrong, for types it's more complicated:
> 205.1b. Some effects change an object's card type, supertype, or subtype but specify that the object retains a prior card type, supertype, or subtype. In such cases, all the object's prior card types, supertypes, and subtypes are retained. This rule applies to effects that use the phrase "in addition to its types" or that state that something is "still a [type, supertype, or subtype]." Some effects state that an object becomes an "artifact creature"; these effects also allow the object to retain all of its prior card types and subtypes. Some effects state that an object becomes a "[creature type or types] artifact creature"; these effects also allow the object to retain all of its prior card types and subtypes other than creature types, but replace any existing creature types.

dracontes: Oh well, my bad... From that I take it that animating lands, Gideon planeswalkers, Cacophony Unleashed, Halcyon Glaze and Still Life are the ones that didn't need this, right? (I should have read the Oracle text on those last two.)

tool4ever: seems so
actually the card of this comment most likely should have the change

Hanmac: yeah, becoming "artifact creature" is a tricky one

@tool4ever should we think about handling "Artifact Creature" different in code?

dracontes: I will note that in my reviewing of card scripts for this pull request I was led to think I was interpreting matters correctly because a lot of the colorless mana animating lands do have `RemoveCreatureTypes$ True`. As examples among others:
- [Cavernous Maw script](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/c/cavernous_maw.txt)
- [Dread Statuary script](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/d/dread_statuary.txt)
- [Stalking Stones script](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/s/stalking_stones.txt)

I'll then also be removing those instances in the redo for this pull request.

tool4ever: Uuh read closely
Those should be fine?

dracontes: Oops, that's the "[creature type] artifact creature" case for the last two. On the other hand both do say that the animated land is still a land. I take it the more restrictive rule takes precedence ('can't' wins over 'can', I believe).

In any case, Cavernous Maw and others that don't say anything about turning into an artifact creature are good to clean up I think.

dracontes:
Mistaken and extensive as this was, redone here, https://github.com/Card-Forge/forge/pull/5022 .

PR #5020, opened 2024-04-12
Outlast: use Keyword for check

Hanmac:
one of the cleanups to use Keyword instead of Flags


tool4ever (approved):

Hanmac:
as next clean up i might remove EternalizeOrEmbalm

PR #5021, opened 2024-04-12
Fix Orvar

tool4ever:


forge-game/src/main/java/forge/game/card/CardProperty.java
```diff
@@ -607,9 +607,12 @@ public static boolean cardHasProperty(Card card, String property, Player sourceC
...
-            final SpellAbility targetingSA = AbilityUtils.getDefinedSpellAbilities(source, def, spellAbility).get(0);
-            if (!targetingSA.canTarget(card)) {
-                return false;
+            SpellAbility targetingSA = AbilityUtils.getDefinedSpellAbilities(source, def, spellAbility).get(0);
+            while (targetingSA != null) {
+                if (targetingSA.usesTargeting() && !targetingSA.canTarget(card)) {
+                    return false;
+                }
+                targetingSA = targetingSA.getSubAbility();
             }
```
Hanmac, line 616: shouldn't that iterate all SpellAbilities first before returning false?

What if:
* Target Player
* and Target Creature that Player controls

?

tool4ever: this got implemented for _Feather, Radiant Arbiter_:
> Whenever you cast a noncreature spell that targets only CARDNAME, you may choose any number of other creatures that spell could target and pay {2} for each of those creatures.
so if it would be illegal target even partly it can't be copied imo

forge-game/src/main/java/forge/game/spellability/SpellAbility.java
```diff
@@ -2068,7 +2068,7 @@ public boolean isTargeting(GameObject o) {
         if (getTargets().contains(o)) {
             return true;
         }
-        SpellAbility p = getParent();
+        SpellAbility p = getSubAbility();
```
Hanmac, line 2071: You changed the Logic from going Up to the Parent
to go down to the Sub Abilities instead

i don't know if that would have other consequences

Especially if other functions still do the targeting in the other direction

tool4ever: I know, I checked all places it gets called and this only seems to fix corner cases
e. g. _Torchling_ wasn't able to target entwined _Abuna's Chant_

Hanmac (approved):

PR #5026, opened 2024-04-12
Embalm and Eternalize use Keyword and Refactor Copy

Hanmac:
Removed `EternalizeOrEmbalm` for Vizier of the Anointed

Refactored `Vizier of Many Faces` a bit 
Technically it could be affected by CardTextChanges, even if there isn't a card like that right now


tool4ever (approved):
maybe the CardTraitBase functions better belong in SpellAbility?
but it doesn't matter much

PR #5036, opened 2024-04-13
Typo cleanup pass #2

dracontes:
Sifted the scripts once more to deal with double whitespace, whitespace where it shouldn't be, files missing an empty line at the end, “fancier” quotation marks and apostrophes, among other perennial cleanup items.


forge-gui/res/cardsfolder/a/arm_mounted_anchor.txt
```diff
@@ -1,7 +1,7 @@
 Name:Arm-Mounted Anchor
 ManaCost:3
 Types:Artifact Equipment
-K:Equip:2:::ReduceCost$ Y: This ability costs {2} less to activate if you have one or fewer cards in hand.
+K:Equip:2:::ReduceCost$ Y:This ability costs {2} less to activate if you have one or fewer cards in hand.
```
tool4ever, line 4: it looks like these should not have the closing dot to look "perfect"

dracontes: Oh yeah, just had a closer look on Belt of Giant Strength and it turns out the period at the end of that sort of sentence is duplicated. On it then 👍 
image

dracontes: While I'm at it, I'm also fixing a number of  other typos I found in the meanwhile. 
A question though: Orvar, the All-Form has a tilde, '~', instead of the plus you'd expect here, `ConditionPresent$ Spell.IsTargeting Valid Permanent.YouCtrl~Other`. I did notice the character is mentioned in the code alongside the plus and that in the recent commit you participated in to fix Orvar this wasn't changed. Is the functionality different in some way?

tool4ever: yea that one is on purpose
I had to implement some additional escaping so it does not split "+" from the first valid part

tool4ever (approved):
you found some really bad ones, thanks

PR #5037, opened 2024-04-14
Alternative Cost For ActivatedAbilities and Spells

Hanmac:
Removed:
* CyclingForZero
* You may pay 0 rather than pay equip costs.
* CARDNAME can't crew Vehicles.
* "Alternative Cost" Keyword (sorta)

Refactored: 
* Heart of Kiran
* MayPlay moved a bit

### DIFF
--- a/forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
+++ b/forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
@@ -16,7 +16,7 @@ public class StaticAbilityAlternativeCost {
 
     static String MODE = "AlternativeCost";
 
-    public static List<SpellAbility> alternativeCosts(final SpellAbility sa, final Player pl) {
+    public static List<SpellAbility> alternativeCosts(final SpellAbility sa, final Card source, final Player pl) {
         List<SpellAbility> result = Lists.newArrayList();
         final Game game = sa.getHostCard().getGame();
         for (final Card ca : game.getCardsIn(ZoneType.STATIC_ABILITIES_SOURCE_ZONES)) {
@@ -25,7 +25,7 @@ public class StaticAbilityAlternativeCost {
                     continue;
                 }
 
-                if (!apply(stAb, sa, pl)) {
+                if (!apply(stAb, sa, source, pl)) {
                     continue;
                 }
 
@@ -33,6 +33,7 @@ public class StaticAbilityAlternativeCost {
                 // set the cost to this directly to bypass non mana cost
                 final SpellAbility newSA = sa.copyWithDefinedCost(cost);
                 newSA.setActivatingPlayer(pl);
+                newSA.setBasicSpell(false);
 
                 // CostDesc only for ManaCost?
                 newSA.putParam("CostDesc", stAb.hasParam("CostDesc") ? ManaCostParser.parse(stAb.getParam("CostDesc")) : cost.toSimpleString());
@@ -42,6 +43,9 @@ public class StaticAbilityAlternativeCost {
                 sb.append(newSA.getCostDescription());
                 // skip reminder text for now, Keywords might be too complicated
                 //sb.append("(").append(newKi.getReminderText()).append(")");
+                if (sa.isSpell()) {
+                    sb.append(" ").append(sa.getDescription());
+                }
                 newSA.setDescription(sb.toString());
 
                 result.add(newSA);
@@ -50,11 +54,11 @@ public class StaticAbilityAlternativeCost {
         return result;
     }
 
-    private static boolean apply(final StaticAbility stAb, final SpellAbility sa, final Player pl) {
+    private static boolean apply(final StaticAbility stAb, final SpellAbility sa, final Card source, final Player pl) {
         if (!stAb.matchesValidParam("ValidSA", sa)) {
             return false;
         }
-        if (!stAb.matchesValidParam("ValidCard", sa.getHostCard())) {
+        if (!stAb.matchesValidParam("ValidCard", source)) {
             return false;
         }
         if (!stAb.matchesValidParam("ValidPlayer", pl)) {
--- a/forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
+++ b/forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
@@ -32,7 +32,7 @@ public class StaticAbilityAlternativeCost {
 
                 Cost cost = new Cost(stAb.getParam("Cost"), sa.isAbility());
                 // set the cost to this directly to bypass non mana cost
-                final SpellAbility newSA = sa.copyWithManaCostReplaced(pl, cost);
+                final SpellAbility newSA = sa.isAbility() ? sa.copyWithDefinedCost(cost) : sa.copyWithManaCostReplaced(pl, cost);
                 newSA.setActivatingPlayer(pl);
                 newSA.setBasicSpell(false);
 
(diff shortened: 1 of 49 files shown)


Hanmac:
Some Question about AlternativeCost and possible X values:
How does "Kentaro, the Smiling Cat" interact with Creatures that already got X in their ManaCost?

tool4ever:
Hmn, we do need to transform that cost better so the {X} is always part of the total cost 🤔 

Hanmac:
> Hmn, we do need to transform that statics result better so the {X} is always part of the total cost 🤔 otherwise you can't use "Spend this mana only on costs that contain {X}." abilities for it... as also discussed in https://www.reddit.com/r/mtgrules/comments/n9pe1d/kentaro_the_smiling_cat_and_alternate_costs/

yeah i have seen that too ... i need to experiment on how I'm going to do that


i reordered the AlternativeCost / MayPlay and that special Stack Keyword logic
Now it checks for AlternativeCost first, then uses that ones to try to apply MayPlay on them (imo that need to refactored later too)

forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
```diff
@@ -0,0 +1,73 @@
...
+                    continue;
+                }
+
+                if (!apply(stAb, sa, source, pl)) {
+                    continue;
+                }
+
+                Cost cost = new Cost(stAb.getParam("Cost"), sa.isAbility());
+                // set the cost to this directly to bypass non mana cost
+                final SpellAbility newSA = sa.copyWithManaCostReplaced(pl, cost);
```
Hanmac, line 34: copyWithManaCostReplaced vs copyWithDefinedCost

this might be for other cards that aren't updated to have AdditionalCost out of their ManaCost

or what do you think? @tool4ever

forge-gui/res/cardsfolder/c/cho_arrim_legate.txt
```diff
@@ -2,7 +2,7 @@ Name:Cho-Arrim Legate
 ManaCost:2 W
 Types:Creature Human Soldier
 PT:1/2
-S:Mode$ Continuous | CharacteristicDefining$ True | AddKeyword$ Alternative Cost:0 | CheckSVar$ X | CheckSecondSVar$ Y | Description$ If an opponent controls a Swamp and you control a Plains, you may cast this spell without paying its mana cost.
+S:AlternativeCost | ValidSA$ Spell | ValidCard$ Card.Self | ValidPlayer$ You | Cost$ 0 | EffectZone$ All | CheckSVar$ X | CheckSecondSVar$ Y | Description$ If an opponent controls a Swamp and you control a Plains, you may cast this spell without paying its mana cost.
```
tool4ever, line 5: all these crash from missing Mode$

Hanmac: fixed

forge-gui/res/cardsfolder/o/once_upon_a_time.txt
```diff
@@ -1,7 +1,7 @@
 Name:Once Upon a Time
 ManaCost:1 G
 Types:Instant
-S:Mode$ Continuous | CharacteristicDefining$ True | AddKeyword$ Alternative Cost:0 | CheckSVar$ X | SVarCompare$ EQ0 | Description$ If this spell is the first spell you've cast this game, you may cast it without paying its mana cost.
+S:Mode$ AlternativeCost | ValidSA$ Spell | ValidCard$ Card.Self | ValidPlayer$ You | Cost$ 0 | EffectZone$ All | CheckSVar$ X | SVarCompare$ EQ0 | Description$ If this spell is the first spell you've cast this game, you may cast it without paying its mana cost.
```
tool4ever, line 4: these ones affecting itself are broken now
probably because we were able to exclude some of STATIC_ABILITIES_SOURCE_ZONES
but now Library and Hand will probably be needed too? :/

another problem is the "You" might be the wrong player, e. g. in case of cast via _Sen Triplets_

Hanmac: I added Host to StaticAbility List, we do something similar with other ones too. No need to add Library or Hand directly

as for "You", the problem is that the Alternative Cost are done before MayPlay ...
would that mean we always need to set the Controller of the card?

forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
Hanmac: This might be refactored in the Future too, Technically each Spell should have their Cost be defined by their Mana Cost, and the other parts should be defined by AdditionalCost StaticAbility

forge-gui/res/cardsfolder/o/once_upon_a_time.txt
tool4ever: Hmn, I checked and since the objects for the valids are correct another approach would be the pull the conditions into that
e. g. `ValidPlayer$ You.!castSpellThisGame` (doesn't exist)

but that involves more refactor, duplication and mixup on those parts?

Hanmac:
> Future work should probably clean up `getOptionalCostValues` parts so they're moved into their own StaticAbility class too

maybe, but these need GUI for the User to choose them.

(And more AI logic to when NOT to choose to pay OptionalCost)

forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
tool4ever: I think for now it might need some logic to support both cases?

Like this I can no longer equip _Murderer's Axe_ without discarding via _Bruenor Battlehammer_ ;/

Hanmac: ugh okay, i need to check this out

Bruenor Battlehammer should set cost to 0
New Perspectives should set the Cost of still Discard<Cardname>, but display it as 0

other AlternativeCost for Spells should still respect additional Cost (that should be rewritten as Static?)

forge-gui/res/cardsfolder/o/once_upon_a_time.txt
Hanmac: Player Property looks better imo

forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
Hanmac: @tool4ever that should fix it?

tool4ever: seems ok now

tool4ever (approved):
could merge as base for future work

Hanmac:
@tool4ever there are still 2-3 "Alternative Cost" left that uses the old keyword. or do you want to try to rework them?

Have you tested if that solves #4696 ?
I think it should. but i haven't tested it yet

tool4ever:
no
I tried to Henzie Blitz _Deepwood Legate_ from library by _Elven Chorus_ permission

looks like only _Marshland Bloodcaster_ besides _Kentaro_ remains with the old KW
I might try getting rid of it, but probably not today

Hanmac:
@tool4ever okay then you can merge this for now ... 
and i try to think of a better way to make the current MayPlay work with the Keywords

Hanmac:
@tool4ever we might also look into replacing the SVar AltCost:
 https://github.com/Card-Forge/forge/blob/master/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java#L4000
with the new static

Issue #1119, opened 2022-07-13
AI: Move UnlessCost into SpellAbility Effect logic


Hanmac:
I need to move the UnlessCost logic into SpellAbility AI part so the AI will be more able to use the UnlessCost depending on different situations

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/issues/1076#issuecomment-1183602007_

tool4ever:
See #6652

PR #5041, opened 2024-04-14
Don't transform X shards of UnlessCost before paying

tool4ever:
You can now pay for `Champions of Minas Tirith` trigger with _Rosheen Meanderer_.


Hanmac:
i updated the fixes, i don't like it when you mix fix commits with bigger changes

Hanmac (approved):
The Other stuff looks okay, i think?

Hanmac:
@tool4ever while looking at UnlessCost, we might take a look at this one: #1119 for teaching AI what the UnlessCost is about so we could implement better logic for this?

tool4ever:
thanks, I can try looking into that in follow-up

PR #4776, opened 2024-03-02
Sundry minor fixes

dracontes:
Decided to redo [this pull request](https://github.com/Card-Forge/forge/pull/4753) from scratch as well, given the unsolvable mess it stemmed from. I do hope things are more manageable now.

- ~Updated [Mythos of Snapdax](https://scryfall.com/card/iko/24/mythos-of-snapdax) to reflect the admittedly contrived possibility of a planeswalker also being a land.~ ➔ Fixed elsewhere: https://github.com/Card-Forge/forge/pull/4093 .
- ~Fixed inconsistent capitalization issues regarding a few parameters, always defaulting to the more common version.~
-- ~`sVars$` --> `SVars$`~ ➔ Looking a bit more closely, especially the relevant Java code, it turns out I had it wrong and the capitalization is consistent. The first parameter variant is uniformly associated with `Animate` effects and the second is solely used for a planeswalker emblem ([Mila, Crafty Companion // Lukka, Wayward Bonder](https://scryfall.com/card/stx/153/mila-crafty-companion-lukka-wayward-bonder)).
-- ~`staticAbilities$` --> `StaticAbilities$`~ ➔ As above, the first variant is uniformly associated with `Animate` effects and the second with generic `Effect` effects (as are planeswalker emblems). In the end, my expertise isn't such I could aim to resolve the situation in a more satisfactory manner. 
- ~Per discussion on Discord, resolving the `ValidDesc$`/`ValidDescription$` parameter duplication, with the less prevalent `ValidDesc$` as the deprecated one.~ ➔ Fixed elsewhere:  https://github.com/Card-Forge/forge/pull/5042
- ~Cleaned up typos found while sifting scripts, namely double whitespace sequences and other recurrences.~ ➔ Fixed elsewhere:  https://github.com/Card-Forge/forge/pull/5017
- ~Per discussion on Discord, added `Creature` to the `AddTypes$` parameter of the `CopyPermanent` effect of [Brenard, Ginger Sculptor](https://scryfall.com/card/woc/27/brenard-ginger-sculptor) and [Synth Infiltrator](https://scryfall.com/card/pip/40/synth-infiltrator) to follow the relevant rulings.~ ➔  Fixed elsewhere: https://github.com/Card-Forge/forge/pull/5000 
[...]

PR #5042, opened 2024-04-15
Resolving ValidDesc$ vs ValidDescription$

dracontes:
Per discussion on Discord, resolving the `ValidDesc$`/`ValidDescription$` parameter duplication, with the less prevalent `ValidDesc$` being the deprecated one (33 vs 299 hits). (Work moved from https://github.com/Card-Forge/forge/pull/4776 )

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/ChooseCardNameEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/ChooseCardNameEffect.java
@@ -45,7 +45,7 @@ public class ChooseCardNameEffect extends SpellAbilityEffect {
 
         if (sa.hasParam("ValidCards")) {
             valid = sa.getParam("ValidCards");
-            validDesc = sa.getParam("ValidDesc");
+            validDesc = sa.getParam("ValidDescription");
         }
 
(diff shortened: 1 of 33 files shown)


tehdiplomat (commented):
I think you need to update this file too https://github.com/Card-Forge/forge/blob/fb6e944b45de6ca8c812d91a88626ddc6f53bc09/forge-game/src/main/java/forge/game/ability/effects/ChooseCardNameEffect.java#L48

dracontes:
> I think you need to update this file too
> 
> https://github.com/Card-Forge/forge/blob/fb6e944b45de6ca8c812d91a88626ddc6f53bc09/forge-game/src/main/java/forge/game/ability/effects/ChooseCardNameEffect.java#L48

Working on the Java side of things feels a bit beyond my expertise, but replacing `ValidDesc` with `ValidDescription` seems to be enough here. Will do if so.

tool4ever:
since this one is GUI only it's not used anywhere else

dracontes:
> since this one is GUI only it's not used anywhere else

I hope I wasn't too bold in my editing and that this bit of code isn't redundant enough that it should be excised instead.

tool4ever (approved):

PR #5045, opened 2024-04-15
Pom: exclude tritonus-share

Hanmac:
without the exclude i get this:

```java

        try {
            audioLine = AudioSystem.getSourceDataLine(format, selectedMixer);
            audioLine.open(format);
        } catch (Exception e) {
            return;
        }

```

With this Problem:

`The method getSourceDataLine(AudioFormat, Mixer.Info) is undefined for the type AudioSystem`


Hanmac:
should be fixed by #6103

PR #5047, opened 2024-04-15
refactor LandAbility to be created by CardFactory

### LINKED ISSUE #165: make LandAbility work for ValidSA
i really need to make LandAbility work for ValidSA

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/pull/162#discussion_r860794151_

This would be great for streamlining Ziatora's Envoy

==
* LandAbility should probably be created by CardFactory when reading the CardFace in a similar way how Spell Permanent is created.
* this might cleanup other mayPlay stuff

* If Enchant Keyword is updated, Spell Permanent would needs to be updated for Auras too

Hanmac:
Closes #165
Part of my Old MayPlay rework

this changes the LandAbility to be created in CardFactory,
so it can be copied if needed

this makes it easier when we are reworking MayPlay later


Hanmac:
@tehdiplomat can you look why the test is failing?

tehdiplomat:
@Hanmac 

Evaluating... (orig score = 346)
0: [549] Simic Growth Chamber (2) -> Play land (chosen: Forest)
0: [346] Simic Growth Chamber (2) -> Play land (chosen: Simic Growth Chamber)
0: [572] Forest (3) -> Play land
BEST: Forest (3) -> Play land SCORE: 572 TIME: 2628
Current phase (MAIN1) plan with score 572:

It looks like it thinks playing the Forest is better than playing the Chamber now. Maybe something you changed also needs to update something in the calculate score logic?

tool4ever:
this is only safe because nothing can turn a non-permanent into a land right now, right? 🤔 
I'm not sure if that assumption will continue to hold though, especially with Alchemy - feels like "perpetually becomes a land" can happen anytime at this point...

Hanmac:
> "perpetually becomes a land"

The Problem there is also that if something becomes a Land, you should not be able to cast it anymore?
So as long as that doesn't happen, it should be okay?

Hanmac:
@tool4ever can we look into it? This might help us with MayPlay changes in the Future

Hanmac:
Time to fix AI again :(

Hanmac:
@tool4ever fixed AI not playing Lands via MayPlay

forge-ai/src/main/java/forge/ai/simulation/GameSimulator.java
```diff
@@ -158,9 +158,9 @@ public Score simulateSpellAbility(SpellAbility origSa, boolean resolve) {
     }
     public Score simulateSpellAbility(SpellAbility origSa, GameStateEvaluator eval, boolean resolve) {
         SpellAbility sa;
-        if (origSa instanceof LandAbility) {
+        if (origSa.isLandAbility()) {
             Card hostCard = (Card) copier.find(origSa.getHostCard());
-            if (!aiPlayer.playLand(hostCard, false)) {
+            if (!aiPlayer.playLand(hostCard, false, origSa)) {
```
tool4ever, line 163: Simulation now refuses to play lands with _Serra Paragon_ (given no other options)

Hanmac: Ugh, can you debug it what the right chante should be?

tool4ever: it wasn't failing there, but for some reason your new LandAbility had `ValidAfterStack$`

forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -1376,29 +1375,10 @@ public List<SpellAbility> chooseSpellAbilityToPlay() {
                 Card land = chooseBestLandToPlay(landsWannaPlay);
                 if ((!player.canLoseLife() || player.cantLoseForZeroOrLessLife() || ComputerUtil.getDamageFromETB(player, land) < player.getLife())
                         && (!game.getPhaseHandler().is(PhaseType.MAIN1) || !isSafeToHoldLandDropForMain2(land))) {
-                    final List<SpellAbility> abilities = Lists.newArrayList();
-
-                    // TODO extend this logic to evaluate MDFC with both sides land
```
tool4ever, line 1381: can we keep this todo? it's still valid

Hanmac: MDFC should be handled with `land.getAllPossibleAbilities(player, true);`?

tool4ever: yea but if the card has two LandAbility, e.g. _Darkbore Pathway_ AI will just always take the first one

tool4ever (approved):

PR #5050, opened 2024-04-16
Remove Alternative Cost KW

### LINKED ISSUE #5044: MayPlay AltCost defined as {X} shard are calculated too early
Hmn, we do need to transform that statics result better so the {X} is always part of the total cost 🤔 

_Originally posted by @tool4ever in https://github.com/Card-Forge/forge/issues/5037#issuecomment-2054100044_

tool4ever:
Closes #5044

### DIFF
--- a/forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
+++ b/forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
@@ -18,8 +18,9 @@ public class StaticAbilityAlternativeCost {
 
     public static List<SpellAbility> alternativeCosts(final SpellAbility sa, final Card source, final Player pl) {
         List<SpellAbility> result = Lists.newArrayList();
-        CardCollection list = new CardCollection(source.getGame().getCardsIn(ZoneType.STATIC_ABILITIES_SOURCE_ZONES));
-        list.add(source);
+        // add source first in case it's LKI (alternate host)
+        CardCollection list = new CardCollection(source);
+        list.addAll(source.getGame().getCardsIn(ZoneType.STATIC_ABILITIES_SOURCE_ZONES));
         for (final Card ca : list) {
             for (final StaticAbility stAb : ca.getStaticAbilities()) {
                 if (!stAb.checkConditions(MODE)) {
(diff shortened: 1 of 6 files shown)


forge-gui/src/main/java/forge/player/HumanPlaySpellAbility.java
```diff
@@ -264,7 +264,10 @@ private boolean announceValuesLikeX() {
         if (needX) {
             if (cost.hasXInAnyCostPart()) {
                 final String sVar = ability.getSVar("X"); //only prompt for new X value if card doesn't determine it another way
-                if ("Count$xPaid".equals(sVar) || sVar.isEmpty()) {
+                // check if X != 0 is even allowed or the X shard got removed
+                boolean replacedXshard = ability.isSpell() && ability.getHostCard().getManaCost().countX() > 0 &&
```
tool4ever, line 268: @Hanmac 
this fixed casting _Aether Tide_ via _Jodah, Archmage Eternal_ and still choosing X
that shouldn't be allowed?

Hanmac: ohm _Aether Tide_ still has X in the Additional Cost ...

but i will need to ask Judges about this

forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
```diff
@@ -36,14 +37,19 @@ public static List<SpellAbility> alternativeCosts(final SpellAbility sa, final C
...
                     newSA.putParam("CostDesc", stAb.hasParam("CostDesc") ? ManaCostParser.parse(stAb.getParam("CostDesc")) : cost.toSimpleString());
+                    sb.append(newSA.getCostDescription());
                 }
 
-                // makes new SpellDescription
-                final StringBuilder sb = new StringBuilder();
-                sb.append(newSA.getCostDescription());
                 // skip reminder text for now, Keywords might be too complicated
                 //sb.append("(").append(newKi.getReminderText()).append(")");
                 if (sa.isSpell()) {
```
Hanmac, line 55: i think you can remove the `sb.append(" ")` below too

i had it before i added the cost into the Reminder Text

forge-gui/src/main/java/forge/player/HumanPlaySpellAbility.java
Hanmac: Rules currently say that X must be 0 for _Aether Tide_

> 107.3b If a player is casting a spell that has an {X} in its mana cost, the value of X isn’t defined by the text of that spell, and an effect lets that player cast that spell while paying neither its mana cost nor an alternative cost that includes X, then the only legal choice for X is 0. This doesn’t apply to effects that only reduce a cost, even if they reduce it to zero. See rule 601, “Casting Spells.”

so while X is in the Additional Cost, it still fails because the ManaCost X says it's Zero

Hanmac (approved):
Nice

Next one is the other AltCost

PR #5053, opened 2024-04-16
150 sleeves

ChristiamMB:
hello, 
I would like to add 150 sleeves, I have already tested it on my forge and it works fine.


Hanmac:
@tehdiplomat imo is that too much for the default package?

tehdiplomat:
Yea 50 megs is probably too big. I think we should figure out how to make this downloadable

tool4ever:
hopefully not all of these are Arena assets?
Those should be excluded at least imo

ChristiamMB:
Hi, sorry, I uploaded the unoptimised one which weighs 50mb, I'm going to upload the 15mb one which is optimised.

ChristiamMB:
@tool4ever you mean remove those that are Arena assets?

tehdiplomat:
Right Arena assets should not be in our repo

ChristiamMB:
No problem, I'll modify it and upload it again.

PR #5057, opened 2024-04-16
Remove AltCost SVar

tool4ever:
This tries to keep the refactoring simple.

In a follow-up I might look into optimizing the condition checks.


Hanmac (approved):

Hanmac:
in my try i wanted to remove `OptionalCost.AltCost` but some stuff depends on it

PR #5073, opened 2024-04-19
Discord, Lord of Disharmony + Support

Jetz72:
Adds support for and an implementation of Discord, Lord of Disharmony.

Adds two properties to spellAbilityHasProperty:
* "CouldCastTiming" matches a SpellAbility that could be cast by its controller given the current turn phase and stack state. Used by Discord's "cast a copy of a spell with that name" ability to restrict the timing of the spell he's attempting to cast.
* "NamedSpell" behaves similar to the card property NamedCard, but uses only the name of the spell itself rather than all the names of the host card. Used by Discord to ensure that if he names half of a split card, an MDFC, or an adventure, that only the named half is cast.


forge-gui/res/cardsfolder/d/discord_lord_of_disharmony.txt
```diff
@@ -0,0 +1,23 @@
+Name:Discord, Lord of Disharmony
+ManaCost:2 B R
+Types:Legendary Creature Chimera
+PT:3/5
+K:Flying
+T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You | TriggerZones$ Battlefield | Execute$ ChooseName | TriggerDescription$ At the beginning of your end step, choose a random nonland Magic card name. Until your next end step, you may cast a copy of a spell with that name, and mana of any type can be spent to cast it. If you cast a spell this way, copy this ability if NICKNAME is on the battlefield.
+
```
dracontes, line 7: While I assume these helped with readability while working on the script, empty lines are only used at the end of scripts and to separate the halves/faces of split/flip/double-faced/specialize cards.

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -2824,6 +2836,26 @@ else if (o.grantsZonePermissions())
                 sAbility = sbSA.toString();
             } else if (sa.isSpell() && sa.isBasicSpell()) {
                 continue;
+            } else if (sa.hasParam("DescriptionFromChosenName") && !getNamedCard().isEmpty()) {
+                String name = getNamedCard();
+                Card namedCard = Card.fromPaperCard(StaticData.instance().getCommonCards().getUniqueByName(name), getController());
+                CardState namedState = namedCard.getStateFromCardName(name);
```
tool4ever, line 2842: this seems unneeded
`getFaceByName` should provide what you want directly?

Jetz72: Aaah, I knew something like that had to exist. Replaced it with that, and cleaned up the getStateFromCardName method I added because I kept needing it without realizing this was an option.

forge-gui/res/cardsfolder/d/discord_lord_of_disharmony.txt
Jetz72: Yeah, I saw that seemed to be standard, but wasn't sure if there was a reason for it. I've cleaned them up now.

tool4ever (approved):
nice work

PR #5089, opened 2024-04-20
Basic AI sideboarding framework

Agetian:
Currently very basic, initially will only make modifications on a loss and only with cards that are more or less equivalent when it comes to their colors and mana value. Will skip lands and cards with mana abilities in order not to screw up the mana base and curve, and will only replace creatures with creatures. Easily toggled on/off via AI profile settings.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
+++ b/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
@@ -107,6 +107,7 @@ public class PlayerControllerAi extends PlayerController {
         }
 
         boolean sbLimitedFormats = getAi().getBooleanProperty(AiProps.SIDEBOARDING_IN_LIMITED_FORMATS);
+        boolean sbSharedTypesOnly = getAi().getBooleanProperty(AiProps.SIDEBOARDING_SHARED_TYPE_ONLY);
         boolean sbPlaneswalkerException = getAi().getBooleanProperty(AiProps.SIDEBOARDING_PLANESWALKER_EQ_CREATURE);
         int sbChanceOnWin = getAi().getIntProperty(AiProps.SIDEBOARDING_CHANCE_ON_WIN);
         int sbChancePerCard = getAi().getIntProperty(AiProps.SIDEBOARDING_CHANCE_PER_CARD);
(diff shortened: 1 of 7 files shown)


forge-gui/res/ai/Cautious.ai
```diff
@@ -311,4 +311,20 @@ SACRIFICE_DEFAULT_PREF_MAX_CMC=1
 # consider the sacrifice of a matching card is a token
 SACRIFICE_DEFAULT_PREF_ALLOW_TOKENS=true
 # A creature should evaluate to no more than this much to be considered for default SacCost preference
-SACRIFICE_DEFAULT_PREF_MAX_CREATURE_EVAL=135
\ No newline at end of file
+SACRIFICE_DEFAULT_PREF_MAX_CREATURE_EVAL=135
+
+# Allow replacing a creature with a planeswalker and vice versa when sideboarding
```
tool4ever, line 316: this entry is duplicated

forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
```diff
@@ -105,8 +102,98 @@ public boolean isAI() {
...
+            } else if (cSide.getRules().getAiHints().getRemAIDecks()) {
+                continue; // don't sideboard in anything that we don't know how to play
+            } else if (cSide.getRules().getType().isLand()) {
+                continue; // don't know how to sideboard lands efficiently yet
+            }
+
+            for (PaperCard cMain : main) {
+                if (processed.contains(cMain)) {
+                    continue;
+                } else if (cMain.equals(cSide)) {
```
tool4ever, line 145: is this check so it skips identical cards?
then it probably needs to compare by name instead?

forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
```diff
@@ -105,8 +102,98 @@ public boolean isAI() {
...
+                    if ((cMain.getRules().getType().isCreature() && !cSide.getRules().getType().isCreature())
+                            || (cSide.getRules().getType().isCreature()) && !cMain.getRules().getType().isCreature()) {
+                        if (!(sbPlaneswalkerException && (cMain.getRules().getType().isPlaneswalker() || cSide.getRules().getType().isPlaneswalker()))) {
+                            continue; // Creature exception: only trade a creature for another creature unless planeswalkers are allowed as a replacement
+                        }
+                    }
+                }
+
+                if (!Card.fromPaperCard(cMain, player).getManaAbilities().isEmpty()) {
+                    continue; // Mana Ability exception: Don't sideboard out cards that produce mana, can screw up the mana base
```
tool4ever, line 165: also add it to processed then so you don't have to perform this expensive check more than once?

Agetian:
Done :) Thanks a lot, @tool4ever !

forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
```diff
@@ -105,8 +102,99 @@ public boolean isAI() {
...
+        boolean sbPlaneswalkerException = getAi().getBooleanProperty(AiProps.SIDEBOARDING_PLANESWALKER_EQ_CREATURE);
+        int sbChanceOnWin = getAi().getIntProperty(AiProps.SIDEBOARDING_CHANCE_ON_WIN);
+        int sbChancePerCard = getAi().getIntProperty(AiProps.SIDEBOARDING_CHANCE_PER_CARD);
+
+        if (!sbLimitedFormats && gameType.isCardPoolLimited()) {
+            return null;
+        }
+
+        if (MyRandom.getRandom().nextInt(100) < sbChanceOnWin) {
+            GameOutcome lastOutcome = brains.getGame().getMatch().getOutcomeById(brains.getGame().getId() - 1);
```
tool4ever, line 120: I think this calculation would risk a NPE, since the global ID might get increased by more from other matches 💣 
better grab the `lastOutcome` field directly?

Agetian: Agreed, done :) Also, fixed the related logic while I was at it.

tool4ever (approved):
LGTM

PR #5093, opened 2024-04-21
Transmuation Font

Hanmac:
The Cost Decision for AI still needs to be updated

and I'm not that happy with the Message for "1 left" when there are still 3 left (even if you choose 1 at a time)

### DIFF
--- a/forge-gui/res/cardsfolder/t/transmutation_font.txt
+++ b/forge-gui/res/cardsfolder/t/transmutation_font.txt
@@ -6,5 +6,6 @@ SVar:Blood:DB$ Token | TokenAmount$ 1 | TokenScript$ c_a_blood_draw | TokenOwner
 SVar:Clue:DB$ Token | TokenAmount$ 1 | TokenScript$ c_a_clue_draw | TokenOwner$ You | SpellDescription$ Clue
 SVar:Food:DB$ Token | TokenAmount$ 1 | TokenScript$ c_a_food_sac | TokenOwner$ You | SpellDescription$ Food
 A:AB$ ChangeZone | Cost$ 3 T Sac<3/Artifact.token+WithDifferentNames/artifact tokens with different names> | Origin$ Library | Destination$ Battlefield | ChangeType$ Artifact | ChangeNum$ 1 | SorcerySpeed$ True | SpellDescription$ Search your library for an artifact card, put it onto the battlefield, then shuffle.
+SVar:AIPreference:SacCost$Artifact.token
 Oracle:{T}: Create your choice of a Blood token, a Clue token, or a Food token.\n{3}, {T}, Sacrifice three artifact tokens with different names: Search your library for an artifact card, put it onto the battlefield, then shuffle. Activate only as a sorcery.
 
--- a/forge-gui/res/cardsfolder/t/transmutation_font.txt
+++ b/forge-gui/res/cardsfolder/t/transmutation_font.txt
@@ -1,0 +1,10 @@
+Name:Transmutation Font
+ManaCost:5
+Types:Artifact
+A:AB$ GenericChoice | Cost$ T | Defined$ You | Choices$ Blood,Clue,Food | SpellDescription$ Create your choice of a Blood token, a Clue token, or a Food token.
+SVar:Blood:DB$ Token | TokenAmount$ 1 | TokenScript$ c_a_blood_draw | TokenOwner$ You | SpellDescription$ Blood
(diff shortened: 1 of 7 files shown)


Agetian:
Yeah, this will most certainly need some AI update, I might try to check it out once I'm done with my current project unless you're faster :) Otherwise, looks good to me :) The card script has an empty line at the end.

forge-gui/src/main/java/forge/player/HumanCostDecision.java
```diff
@@ -1106,6 +1112,24 @@ public PaymentDecision visit(final CostSacrifice cost) {
         if (0 == c) {
             return PaymentDecision.number(0);
         }
+        if (differentNames) {
+            final CardCollection chosen = new CardCollection();
+            while (c > 0) {
+                final InputSelectCardsFromList inp = new InputSelectCardsFromList(controller, 1, 1, list, ability);
+                inp.setMessage(Localizer.getInstance().getMessage("lblSelectATargetToSacrifice", cost.getDescriptiveType(), "%d"));
```
tool4ever, line 1119: Try passing c instead of %d here for the right amount?

forge-gui/res/cardsfolder/t/transmutation_font.txt
```diff
@@ -0,0 +1,10 @@
+Name:Transmutation Font
+ManaCost:5
+Types:Artifact
+A:AB$ GenericChoice | Cost$ T | Defined$ You | Choices$ Blood,Clue,Food | SpellDescription$ Create your choice of a Blood token, a Clue token, or a Food token.
+SVar:Blood:DB$ Token | TokenAmount$ 1 | TokenScript$ c_a_blood_draw | TokenOwner$ You | SpellDescription$ Blood
+SVar:Clue:DB$ Token | TokenAmount$ 1 | TokenScript$ c_a_clue_draw | TokenOwner$ You | SpellDescription$ Clue
+SVar:Food:DB$ Token | TokenAmount$ 1 | TokenScript$ c_a_food_sac | TokenOwner$ You | SpellDescription$ Food
+A:AB$ ChangeZone | Cost$ 3 T Sac<3/Artifact.token+WithDifferentNames/artifact tokens with different names> | Origin$ Library | Destination$ Battlefield | ChangeType$ Artifact | ChangeNum$ 1 | SorcerySpeed$ True | SpellDescription$ Search your library for an artifact card, put it onto the battlefield, then shuffle.
+Oracle:{T}: Create your choice of a Blood token, a Clue token, or a Food token.\n{3}, {T}, Sacrifice three artifact tokens with different names: Search your library for an artifact card, put it onto the battlefield, then shuffle. Activate only as a sorcery.
```
tehdiplomat, line 9: Should we just mark this remove AI for now? Or maybe AILogic$ False for the second ability for now?

Hanmac: No need to worry, I can fix the AI Cost Decision today too

Hanmac:
@Agetian can you look into this to allow AI to sacrifice stuff with different name?

https://github.com/Card-Forge/forge/blob/d9a6acc10baeb93ee28d92907c798ff045db18ad/forge-ai/src/main/java/forge/ai/AiCostDecision.java#L449-L451

Agetian:
@Hanmac I'll try to look into that, yeah

Hanmac:
> I added some AI support that seems to work, but sadly, there's some code duplication now :/ Maybe you'd be able to somehow rewire that to avoid duplicating things, if possible?..

My suggestion is to go into AiCostDecision:

```
final AiController aic = ((PlayerControllerAi)player.getController()).getAi(); 
 CardCollectionView list = aic.chooseSacrificeType(cost.getType(), ability, isEffect(), c, null); 
 return list == null ? null : PaymentDecision.card(list); 
```

instead of directly calling `chooseSacrificeType`, maybe check for differentNames and instead of passing amount `c`, call that in a loop?


But yeah, the other check in ComputerUtilCost is for something different, so we need some code duplication ...

Just add more TODO so that i remember to change the stuff with "share names"

Agetian:
Alrighty, feel free to experiment with that a bit, I tried something similar but it didn't want to work for some reason :/ Might look into it again a bit later

Hanmac:
@Agetian @tehdiplomat @tool4ever ready for now?

Agetian:
Think so, yeah

Agetian (approved):

PR #5097, opened 2024-04-21
Cost$ cleanup: Main event: Folder 'c' part 3

dracontes:
Other fixes:
- Redundant `StackDescription$` on Creeping Chill

### DIFF
--- a/forge-gui/res/cardsfolder/c/creeping_chill.txt
+++ b/forge-gui/res/cardsfolder/c/creeping_chill.txt
@@ -1,8 +1,8 @@
 Name:Creeping Chill
 ManaCost:3 B
 Types:Sorcery
-A:SP$ DamageAll | ValidPlayers$ Player.Opponent | NumDmg$ 3 | SubAbility$ DBGainLife | SpellDescription$ CARDNAME deals 3 damage to each opponent and you gain 3 life. | StackDescription$ CARDNAME deals 3 damage to each opponent and {p:You} gains 3 life.
+A:SP$ DamageAll | ValidPlayers$ Player.Opponent | StackDescription$ CARDNAME deals 3 damage to each opponent and | NumDmg$ 3 | SubAbility$ DBGainLife | SpellDescription$ CARDNAME deals 3 damage to each opponent and you gain 3 life.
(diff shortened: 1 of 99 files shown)


forge-gui/res/cardsfolder/c/creeping_chill.txt
```diff
@@ -1,7 +1,7 @@
 Name:Creeping Chill
 ManaCost:3 B
 Types:Sorcery
-A:SP$ DamageAll | StackDescription$ CARDNAME deals 3 damage to each opponent and | Cost$ 3 B | ValidPlayers$ Player.Opponent | NumDmg$ 3 | SubAbility$ DBGainLife | SpellDescription$ CARDNAME deals 3 damage to each opponent and you gain 3 life.
+A:SP$ DamageAll | ValidPlayers$ Player.Opponent | NumDmg$ 3 | SubAbility$ DBGainLife | StackDescription$ SpellDescription | SpellDescription$ CARDNAME deals 3 damage to each opponent and you gain 3 life.
```
tool4ever, line 4: I wouldn't say it was redundant:
image

dracontes: I hope this solution is adequate. Turns out trailing into automatic text by ` and | ` is fairly rare: only two scripts do it.

tool4ever: I guess that would work
However I still don't understand why you'd change it at all since it was already fine? 🤔

dracontes: Oh, at the time I was quickly scanning the prospective 4000+ hit list for the bulk of the search and replace to check for anything amiss. Various instances of parameters out of the customary order kept showing up. I guess, in my momentary annoyance, I didn't read too closely when dealing with Creeping Chill. 
I'll do my best to do better on that count.

tool4ever (approved):

forge-gui/res/cardsfolder/c/creeping_chill.txt
tool4ever: sure, no problem

tool4ever (approved):

PR #5113, opened 2024-04-22
Deck-based AI hints metadata + predefined sideboarding plan support

Agetian:
Ok, so this is probably going to be slightly controversial/divisive, so hear me out :)
What this change does is it brings a possibility of creating a metadata line in the deck files (AIHints=...) which can be used to specify an assortment of hints for the AI for a specific deck. Currently, as a proof-of-concept demo, this brings the ability to create a predefined sideboarding plan for the AI to use (e.g. for fully transformative sideboards), but can be expanded for various AFs, for example, to specify card tutoring priorities relevant for a particular deck, etc.

The format is similar to the scripting language we use and is generally like this:
AIHints=HintName$ HintValue | HintName$ HintValue | ...

So, for the predefined AI sideboarding plan, it's possible to specify (for example):
AIHints=SideboardPlan$ Fang of Shingeki->Kami of Transience;Audacity->Michiko's Reign of Truth

The reason this is a controversial change is because this expands the deck file format to include an entry that is not specific to the deck composition itself but more to the AI's use of the deck, which may sort of make the deck format "impure". That said, our current deck format has supported non-deck data in it for the longest time (over 10 years), including a (practically unused) flag that was supposedly designed to support defining whether the deck was designed to be played by a human or by the AI (cf. PLAYER and PLAYER_TYPE definitions in DeckFileHeader.java, e.g. line 86). Some additional fields are supported as well, such as comments and tags, which expand the deck information as needed. Thus, I don't think it's completely unfeasible to add another potential field to the mix which would be AI-centric, especially considering that Forge is a game that, in the absolute majority of cases, is played against the AI opponents, as opposed to PvP network games.

Then again, I understand the concern that changes like this one would potentially "pollute" the deck format, so if there are any arguments for or against this approach, please let me know. This is the only way I'm aware of that I can support it. I tried imagining something with the extra files specifically for the purpose, but that felt too "bulky" to me and I couldn't devise a clean enough interface to make things interact as needed. So, if a different approach would be preferable, I think I wouldn't be the best man to implement it :)

### DIFF
--- a/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
+++ b/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
@@ -4,6 +4,7 @@ import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import com.google.common.collect.*;
 import forge.LobbyPlayer;
+import forge.StaticData;
 import forge.ai.ability.ProtectAi;
 import forge.card.CardStateName;
 import forge.card.ColorSet;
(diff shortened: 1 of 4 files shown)


forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
```diff
@@ -106,6 +106,38 @@ public List<PaperCard> sideboard(Deck deck, GameType gameType, String message) {
             return null;
         }
 
+        Map<PaperCard, PaperCard> sideboardPlan = Maps.newHashMap();
+        List<PaperCard> main = deck.get(DeckSection.Main).toFlatList();
+        List<PaperCard> sideboard = deck.get(DeckSection.Sideboard).toFlatList();
```
tool4ever, line 111: need check if sideboard even exists:
```
Game-0 > java.lang.NullPointerException: Cannot invoke "forge.deck.CardPool.toFlatList()" because the return value of "forge.deck.Deck.get(forge.deck.DeckSection)" is null
	at forge.ai.PlayerControllerAi.sideboard(PlayerControllerAi.java:126)
	at forge.game.Match.prepareAllZones(Match.java:275)
```

Agetian:
Added a fix for the empty sideboard NPE, thanks!
I agree about the example, I need to find a good one though :) (and not yet sure where exactly we'll add an example like that... maybe to quest decks? :/ )

Agetian:
I'm thinking about turning AI sideboarding into a GUI toggle (as opposed to the AI profile parameter), this makes it easy to enable or disable this feature (disabling it may be desired since it mostly produces random results right now unless the pre-planned sideboard strategy is defined). What do you think?

[Turn_AI_sideboarding_into_a_GUI_toggle_parameter.txt](https://github.com/Card-Forge/forge/files/15067507/Turn_AI_sideboarding_into_a_GUI_toggle_parameter.txt)

tool4ever:
I think it's fine either way
(users could just not give the deck one or work with the profiles where all parameters can be considered together?)
Otherwise a dropdown from Off->Human for AI->AI alone might be clearer

Agetian:
Agreed! I suggest that we merge this one then, and in the followup I'll try to implement the dropdown approach to selecting the sideboarding option.

PR #5154, opened 2024-04-24
Fix X shards in combat tax missing from total cost

### LINKED ISSUE #5044: MayPlay AltCost defined as {X} shard are calculated too early

tool4ever:
Closes #5044 again

Maybe only 95% clean but gets the job done considering this is somewhat corner case territory imo


Agetian (approved):

Agetian:
Looks good to me :+1:

PR #5038, opened 2024-04-14
UNF Embiggen

tehdiplomat:

PR #5155, opened 2024-04-25
Embiggen (take two)

Agetian:
After a failed merge in #5038 I tried creating a new branch and cherry-picking the relevant commits into it. Seems like it worked, but the branch shows more changes to AbilityUtils compared to the original Embiggen branch :/ Still not sure what went wrong in the first place and why things look slightly different now, so please proceed with caution. Hopefully this will work.


Agetian:
Hmm, no, apparently this doesn't work and it brings more changes than needed from wherever :/

PR #5156, opened 2024-04-25
UNF Embiggen - take four

Agetian:
The fourth attempt at cherry-picking the contents of the Embiggen branch after a failed merge. Hopefully fourth time is the charm and all that.

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -2643,20 +2643,6 @@ public class AbilityUtils {
 
             return doXMath(amount, expr, c, ctb);
         }
-        if (sq[0].contains("CardTypes")) {
-            return doXMath(countCardTypesFromList(getDefinedCards(c, sq[1], ctb), false), expr, c, ctb);
-        }
-        if (sq[0].contains("CardControllerTypes")) {
-            return doXMath(countCardTypesFromList(player.getCardsIn(ZoneType.listValueOf(sq[1])), false), expr, c, ctb);
-        }
-        if (sq[0].contains("CardControllerPermanentTypes")) {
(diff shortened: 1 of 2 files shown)


forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -2634,6 +2634,30 @@ public static int xCount(Card c, final String s, final CardTraitBase ctb) {
...
+        if (sq[0].equals("AllTypes")) {
+            List<Card> cards = getDefinedCards(c, sq[1], ctb);
+
+            int amount = countCardTypesFromList(cards, false) +
+                    countSuperTypesFromList(cards) +
+                    countSubTypesFromList(cards);
+
+            return doXMath(amount, expr, c, ctb);
+        }
+        if (sq[0].contains("CardTypes")) {
```
tool4ever, line 2646: still lots of bad changes in this file that already got refactored
please no changes except related for the new AllTypes property

Agetian: Which ones exactly are you referring to? I'm trying to figure out what is what here tbh and I got quite tangled in these changes :/

Agetian: Ok, removed most of the stuff unrelated to AllTypes, but there's also a rename for get.... -> count.... - should that go as well or should it be kept as countXXX?

tool4ever: if it compiles now it's probably fine

Agetian: Yes, tested the compile and the card itself, looks like it works fine. Will merge this one then. Thanks for looking at it! :+1:

tehdiplomat:
@Agetian Oh i would have gotten back to it. Just no time yesterday.  Thanks though.

PR #5186, opened 2024-05-01
Tribal to Kindred update

dracontes:
Looked like a simple enough case of search and replace. I left the translations alone to be done at a later date by those more knowledgeable than me. I also didn't touch `mtg-data.txt` which I presume will also be updated later. The remainder of 'tribal' instances are deck names, descriptions of decks and puzzles, etc., thus not relevant.

**Other fixes**
- Updated [Mirror Golem](https://scryfall.com/card/mrd/208/mirror-golem) to reflect the existence of battles.
- [Aminatou's Augury](https://scryfall.com/card/cmm/73/aminatous-augury) had its `Effect`'s `PlayTr` static ability referred to by both the lines for kindred and battle. Disambiguated between the two.
- ~My only conundrum is whether the reminder text on cards like [Holistic Wisdom](https://scryfall.com/card/ody/243/holistic-wisdom) should be updated to include a mention to battles.~ 
‣ Might as well then. Alphabetizing the order in [Largepox](https://scryfall.com/card/cmb2/44/largepox) will probably best wait for the Oracle update (if it happens at all).
- On `GroupDef.java`: Removed "Kindred instant" as a "group by: Type" category on the Card Catalog as it was manifestly out of place.

### DIFF
--- a/forge-gui/src/main/java/forge/itemmanager/GroupDef.java
+++ b/forge-gui/src/main/java/forge/itemmanager/GroupDef.java
@@ -106,7 +106,7 @@ public enum GroupDef {
             }),
 
     CARD_TYPE("lblType",
-            new String[] { "Planeswalker", "Creature", "Sorcery", "Instant", "Artifact", "Enchantment", "Land", "Tribal instant", "Battle" },
+            new String[] { "Planeswalker", "Creature", "Sorcery", "Instant", "Artifact", "Enchantment", "Land", "Kindred instant", "Battle" },
             new Function<Integer, ColumnDef>() {
                 @Override
                 public ColumnDef apply(final Integer groupIndex) {
@@ -127,7 +127,7 @@ public enum GroupDef {
                         if (type.isCreature()) {
                             return 1;
                         }
-                        if (type.isTribal() && type.isInstant()) {
+                        if (type.isKindred() && type.isInstant()) {
                             return 7;
                         }
                         if (type.isInstant()) {
--- a/forge-gui/src/main/java/forge/itemmanager/GroupDef.java
+++ b/forge-gui/src/main/java/forge/itemmanager/GroupDef.java
@@ -106,7 +106,7 @@ public enum GroupDef {
             }),
 
     CARD_TYPE("lblType",
-            new String[] { "Planeswalker", "Creature", "Sorcery", "Instant", "Artifact", "Enchantment", "Land", "Kindred instant", "Battle" },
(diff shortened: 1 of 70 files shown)


forge-gui/src/main/java/forge/itemmanager/GroupDef.java
```diff
@@ -106,7 +106,7 @@ public Integer apply(final InventoryItem item) {
             }),
 
     CARD_TYPE("lblType",
-            new String[] { "Planeswalker", "Creature", "Sorcery", "Instant", "Artifact", "Enchantment", "Land", "Tribal instant", "Battle" },
+            new String[] { "Planeswalker", "Creature", "Sorcery", "Instant", "Artifact", "Enchantment", "Land", "Kindred instant", "Battle" },
```
dracontes, line 109: Though I'd like to be sure that `"Kindred instant"` is right. Since, at first glance, shouldn't it be just `"Kindred"`?

dracontes: Removed the category, assuming no word to the contrary at this point is a "go ahead".

forge-gui/src/main/java/forge/itemmanager/GroupDef.java
```diff
@@ -106,7 +106,7 @@ public Integer apply(final InventoryItem item) {
             }),
 
     CARD_TYPE("lblType",
-            new String[] { "Planeswalker", "Creature", "Sorcery", "Instant", "Artifact", "Enchantment", "Land", "Tribal instant", "Battle" },
+            new String[] { "Planeswalker", "Creature", "Sorcery", "Instant", "Artifact", "Enchantment", "Land", "Battle" },
```
tool4ever, line 109: is this only used for grouping or also filtering?
wouldn't it be better to keep it as Kindred... 🤔

dracontes: Practically applied in Forge, as far as I can tell, it only shows up as a category when the "group by:" dropdown is set to "Type". 
Conceptually, I'd think it can be used for both in an "at a glance" way really. The thing is I'm not sure it's that interesting to have the cards that are also kindred set apart from the other types they are. My reckoning is, more often than not, if someone is searching for anything kindred-associated it's the creature types.

tool4ever: hmn good point :)

tool4ever (approved):
LGTM

PR #5190, opened 2024-05-02
Clean up some obsolete payment checks

tool4ever:
Fixes Autopay not working with Emerged

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -7199,6 +7199,11 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
             return false;
         }
 
+        // can't sacrifice it for mana ability if it is already marked as sacrifice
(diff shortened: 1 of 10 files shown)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -7199,6 +7199,11 @@ public final boolean canBeSacrificedBy(final SpellAbility source, final boolean
             return false;
         }
 
+        // can't sacrifice it for mana ability if it is already marked as sacrifice
+        if (source != null && source.isManaAbility() && isUsedToPay()) {
+            return false;
+        }
```
Hanmac, line 7205: that i need to look at when i refactor the CostPayment in the far future

tool4ever: yea I'm still not happy with isUsedToPay because it's tricky to reset correctly
but I will try to look into it later

PR #5202, opened 2024-05-04
Fixing card text capitalization on the Adventure Mode custom card Xira's Boss Effect

dracontes:
Missed the train on this pull request by mere seconds, https://github.com/Card-Forge/forge/pull/5201 , so once more, with feeling.


Simisays:
Great catch thanks

PR #5204, opened 2024-05-05
Fix Watcher of Hours triggering from Dust of Moments cast by opponent

tool4ever:


Hanmac (approved):

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -6222,10 +6223,10 @@ public final int addDamageAfterPrevention(final int damageIn, final Card source,
 
         DamageType damageType = DamageType.Normal;
         if (isPlaneswalker()) { // 120.3c
-            subtractCounter(CounterType.get(CounterEnumType.LOYALTY), damageIn, true);
+            subtractCounter(CounterType.get(CounterEnumType.LOYALTY), damageIn, null, true);
```
Hanmac, line 6226: i think the Source Controller is the counter remover in that case

similar to the put counter for wither damage

tool4ever: > 120.3c. Damage dealt to a planeswalker causes that many loyalty counters to be removed from that planeswalker.

vs.
> 120.3d. Damage dealt to a creature by a source with wither and/or infect causes that source's controller to put that many -1/-1 counters on that creature.

imo that means for planeswalkers it's the rules doing it?

Hanmac: hm maybe because it doesn't matter right now, we could just keep it like this?

PR #5212, opened 2024-05-06
Adventure Mode custom card cleanup

dracontes:
Mostly fixing grammar and departures from the script formatting standard for official cards. 
Not sure how welcome it would be for me to get into the weeds of templating, but I'm certainly available for relevant commentary.


Simisays:
Hey thank you for cleaning these up. 

If you find the time (and willingness) for it, a similar cleanup for the quest texts would be very welcome as well !

dracontes:
> If you find the time (and willingness) for it, a similar cleanup for the quest texts would be very welcome as well !

Those would be found in the `quests.json` files, right?

PR #5226, opened 2024-05-11
Typo Patch

edg444:
Minor typos and an oracle update.

### DIFF
--- a/forge-gui/res/cardsfolder/e/endbringer.txt
+++ b/forge-gui/res/cardsfolder/e/endbringer.txt
@@ -2,7 +2,7 @@ Name:Endbringer
 ManaCost:5 C
 Types:Creature Eldrazi
 PT:5/5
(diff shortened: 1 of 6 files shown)


forge-gui/res/cardsfolder/e/endbringer.txt
```diff
@@ -2,7 +2,7 @@ Name:Endbringer
 ManaCost:5 C
 Types:Creature Eldrazi
 PT:5/5
-K:CARDNAME untaps during each other player's untap step.
+K:Untap CARDNAME during each other player's untap step.
```
tool4ever, line 5: you can't just mess with a KW...

edg444: Oops. Didn't do my due diligence checking that one. I'm closing this and making a new patch since it's outdated now.

PR #5242, opened 2024-05-15
Umbra armor rename

Hanmac:
Upcoming rename for `Totem armor` into `Umbra armor`

should we take that change to look into AI logic for this?


Agetian:
Hmm, I don't immediately see any direct references to Totem Armor in any of the AI code... The only place where it appears to be mentioned is at EffectAi:355, but the code there doesn't reference it directly either. Do you have an idea where it might be relevant to adapt the AI code? :/

Hanmac:
> Hmm, I don't immediately see any direct references to Totem Armor in any of the AI code... The only place where it appears to be mentioned is at EffectAi:355, but the code there doesn't reference it directly either. Do you have an idea where it might be relevant to adapt the AI code? :/

My plan was for "Choose Replacement Effect" when choosing which Effect to replace the Destruction Effect
In case there are multiple ones with Totem/Umbra Armor and/or Regeneration 

But I don't know yet what's the best way would be

Hanmac:
@Agetian i added basic logic for Destory RE
checking for Shield Counter, Regeneration and Umbra Armor

Agetian (approved):

PR #5247, opened 2024-05-15
Effect cleanup and fixes

### LINKED ISSUE #5235: Cannot order multiple Captured by the Consulate triggers
**Describe the bug**
I am unable to properly order the triggers when I control multiple Captured by the Consulate. They just get added to the stack automatically.

**Ruling:** *If there is more than one Captured by the Consulate on the battlefield, the triggered ability of each one will change the target if able. The last one to resolve determines what that spell finally targets. If one player controls more than one Capture by the Consulate, they choose the order in which they resolve. If multiple players control those Auras, the first player among them in turn order (starting with the player whose turn it is now) is the one whose triggered ability resolves last.*

**Expected behavior**
When you control multiple Captured by the Consulate and they trigger, you are able to order them on the stack as you please.

**Screenshots**
Doesn't really show much beyond the triggers being on the stack, but I never got a prompt for their order.
Screenshot_20240513_164551 cleaned

 - Version: forge-gui-desktop-1.6.63-SNAPSHOT-05-12

tool4ever:
- The removed AI classes were less advanced than their normal variants anyway, so nothing of value is lost

- Closes #5235

- image
mini fix for regression from #4794


forge-game/src/main/java/forge/game/spellability/SpellAbility.java
```diff
@@ -937,36 +937,38 @@ public String getOriginalDescription() {
...
-                }
+        }
 
-                if (payCosts.isAbility() && !isEquip()) {
-                    sb.append(": ");
+        boolean equip = false;
+        StringBuilder sb = new StringBuilder();
+        // descriptors
+        if (hasParam("PrecostDesc")) {
+            equip = getParam("PrecostDesc").equals("Equip");
```
Hanmac, line 946: `getParam("PrecostDesc").equals("Equip")` is too restrictive

from this part, it might be more?
```
abilityStr.append(" | PrecostDesc$ Equip");
if (k.length > 3 && !k[3].isEmpty()) {
    abilityStr.append(" ").append(vstr);
}
```

tool4ever: good catch

forge-gui/res/cardsfolder/w/wrap_in_vigor.txt
```diff
@@ -1,5 +1,5 @@
 Name:Wrap in Vigor
 ManaCost:1 G
 Types:Instant
-A:SP$ RegenerateAll | ValidCards$ Creature.YouCtrl | SpellDescription$ Regenerate each creature you control.
+A:SP$ RegenerateAll | Defined$ Valid Creature.YouCtrl | SpellDescription$ Regenerate each creature you control.
```
dracontes, line 4: Missed this one (`RegenerateAll` --> `Regenerate`), if I'm reading the changes correctly and `RegenerateAll` was completely removed.

Agetian:
It looks like the difference with the RegenerateAllAi implementation is that it only fired it in case it could save more than one object at a time, while the default implementation will fire even if there's only one object to save. Not sure which one is more desirable tbh, I mean, sometimes you may want to use a RegenerateAll effect to save just one thing. What do you thing? Otherwise yeah, looks good to me :)

tool4ever:
Hmn good observation
I've tweaked it a bit :)

Agetian:
Yeah, this looks good :) Should RegenerateAi be hooked up to the RegenerateAllEffect then in SpellApiToAi?

tool4ever:
no, RegenerateAllEffect also stopped existing ;)

Agetian:
Gotcha, looks good! :) Thanks!

Agetian (approved):

PR #5251, opened 2024-05-16
Adding PreDH format

TheLastNarwhal:
I believe I included all legal sets, though I did omit a fair few promo sets since I couldn't tell if they would contain illegal modern cards. The legal sets are New Phyrexia and everything that came before it. Also, I couldn't figure out what the "Order" part was, so I left it blank.

I hope I did this all right... github is always intimidating when it comes to PRs.


Agetian (approved):

TheLastNarwhal:
@Agetian 
So, everything is correct with the format file? Like I said, I didn't know what the Order part was, so that's blank. All the other format files had that bit filled out with numbers.

Agetian:
For the most part, yeah. There were some sets that are not yet in Forge (namely, with the codes CED, CEI, SUM, PTC, RQS, MGB, ITP) that threw warnings on startup, and two sets had their Scryfall codes specified instead of Forge codes which made the game also throw a warning (PDRC -> DRC94, ME1 -> MED), but otherwise it looks good to me, yes. The "Order" part, as far as I understand it, only matters for the Archived formats in order to determine their sorting order in case the explicit dates aren't specified. Not sure what they were supposed to mean (or how exactly they were specified) for the other Casual formats, but things work as intended without the specification as well. I made a little followup to clean up the warnings (and I hid the non-existent sets behind a TODO label for the time being).

TheLastNarwhal:
Ah, I didn't realize that Forge didn't have some of those set codes. I figured that Forge and Scryfall would be using the same ones. Anywho, I happy that this format will be in Forge now, 'cause it seems exciting.

TheLastNarwhal:
@Agetian

Apologies if this is a bit premature and there's still format stuff that needs to be done, but I tried to build a deck using the preDH restrictions and it wasn't a selectable format. I can see the file in **/res/format/casual**, but I don't think Forge recognizes it as an actual format yet.

Agetian:
As of right now, it's only loaded along with the other custom (casual, archived) formats if the relevant option is enabled in the game options. If the intent is to make it a core format that's always available to the player, then it should be added to the list of core formats in GameFormat.java (check out the code circa line 356). I'll add it there, I believe that was the original intent, anyway :)

PR #5269, opened 2024-05-21
Basic AI logic for Tireless Provisioner

Agetian:
To be improved/generalized as needed, but works for the basic use cases.


Hanmac:
I think for a first attempt, this is okay?

Maybe it can be generalized later?

Hanmac (approved):

Agetian:
Yeah, I'll try to improve on this a bit later and see what other cards might need this kind of treatment :)

PR #5288, opened 2024-05-24
Fixing Bludgeon Brawl

dracontes:
Per ruling on [Bludgeon Brawl](https://scryfall.com/card/nph/80/bludgeon-brawl), it shouldn't overwrite existing artifact subtypes, as shown in the appended screenshot of its effect on a Fortification, so the `RemoveArtifactTypes` parameter was removed. This also brings it in line with the similar effect on [Dan Lewis](https://scryfall.com/card/who/78/dan-lewis).

image


dracontes:
Should've looked more widely for commentary on this corner case.

PR #5298, opened 2024-05-25
StaticAbilityAlternativeCost: improve X handling

tool4ever:
Fixes using _Marshland Bloodcaster_ ability for casting breaking in case the spell uses some hardcoded X SVar.

Also reworked hopefully all remaining {X} cases, so that the calculation only happens before you pay.
(This makes them work with the `CostContainsX` mana restriction - currently done in 4 cards).


Hanmac (approved):
About the To-dos, we should make followup tickets for them

PR #5313, opened 2024-05-28
Draft matters again

tehdiplomat:


forge-gui/res/editions/Conspiracy Take the Crown.txt
```diff
@@ -237,18 +237,21 @@ ScryfallCode=CN2
...
-#Emissary's Ploy
+Emissary's Ploy
 Garbage Fire
 Hired Heist
 Hold the Perimeter
 Hymn of the Wilds
 Incendiary Dissent
 Natural Unity
+Noble Banneret
+Paliano Vanguard
```
tehdiplomat, line 252: Regicide should be added to this list i think

forge-core/src/main/java/forge/card/CardFace.java
```diff
@@ -51,7 +47,7 @@ public enum FaceSelectionMethod { //
     private List<String> abilities = null;
     private List<String> staticAbilities = null;
     private List<String> triggers = null;
-    private List<String> draftActions = null;
+    private List<String> draftActions = Lists.newArrayList();
```
tool4ever, line 50: why would all instances need empty list for this 0.1% as default now?
not sure what the JVM will make of that memory wise, but feels like bad design anyway

forge-gui/res/cardsfolder/p/paliano_the_high_city.txt
```diff
@@ -0,0 +1,8 @@
+Name:Paliano, the High City
+Types:Legendary Land
+Draft:Reveal CARDNAME as you draft it.
+Draft:As you draft CARDNAME, the player to your right chooses a color, you choose another color, then the player to your left chooses a third color.
+A:AB$ Mana | Cost$ T | Produced$ Combo NotedColors | Condition$ Add one mana of any color chosen as you drafted cards named CARDNAME. | SpellDescription$ Add one mana of any color chosen as you drafted cards named Paliano, the High City.
+Oracle:Reveal CARDNAME as you draft it. As you draft CARDNAME, the player to your right chooses a color, you choose another color, then the player to your left chooses a third color.\n{T}: Add one mana of any color chosen as you drafted cards named Paliano, the High City.
```
tool4ever, line 6: bad oracle

forge-gui/res/cardsfolder/p/paliano_vanguard.txt
```diff
@@ -0,0 +1,8 @@
+Name:Paliano Vanguard
+ManaCost:1 W
+Types:Creature Human Soldier
+PT:2/2
+Draft:Draft CARDNAME face up.
+Draft:As you draft a creature card, you may reveal it, note its creature types, then turn CARDNAME face down.
+S:Mode$ Continuous | Affected$ Creature.Other+NotedTypes+YouCtrl | AddPower$ 1 | AddToughness$ 1 | Description$ Other creatures you control of a type you noted for cards named Paliano Vanguard get +1/+1.
```
tool4ever, line 7: hmn, I think you kind of will have to hardcode the "named" part into these scripts too
(so that when e.g. _Sakashima the Impostor_ copies them the abilities will still work :) )

tool4ever: or you could try to use getCardState in the code to navigate to the original trait host

forge-game/src/main/java/forge/game/spellability/AbilityManaPart.java
```diff
@@ -523,6 +521,9 @@ public final String mana(SpellAbility sa) {
             return getComboColors(sa);
         }
         String produced = this.getOrigProduced();
+        if (produced.contains("NotedColor")) {
+            produced = produced.replace("NotedColor", sa.getActivatingPlayer().getDraftNotes().get(""));
```
tool4ever, line 525: hmn, is this code reached for anything?

forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java
```diff
@@ -330,6 +312,11 @@ public String apply(String input) {
                     String[] restrictions = params.containsKey("SharedRestrictions") ? params.get("SharedRestrictions").split(",") : new String[] {"Card"};
                     addKeywords = CardFactoryUtil.sharedKeywords(addKeywords, restrictions, zones, hostCard, stAb);
                 }
+
+                if (params.containsKey("FromDraftNotes")) {
+                    addKeywords = Lists.newArrayList(hostCard.getController().getDraftNotes().getOrDefault(hostCard.getName(), "").split(","));
```
Hanmac, line 317: The Previous AddKeywords aren't even used?

And imo I would like it more if the creature(/names) were drafted there, and the Keywords are calculated

forge-gui/res/cardsfolder/p/paliano_vanguard.txt
tehdiplomat: I put them into CardProperty. I'm not sure if I love that but its more correct for now..

forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java
tehdiplomat: I'm kinda concerned about the performance consideration of not extracting the keywords out of the PaperCards. If I have to do it in the game, its going to happen every tick.

Hanmac: imo we could optimize it later

tehdiplomat: Cool. I'll add it to a futures list.

tool4ever (approved):

PR #5339, opened 2024-05-30
Added Depth Defiler, corrected Furnace Hellkite

Fulgur14:


forge-gui/res/cardsfolder/upcoming/depth_defiler.txt
```diff
@@ -0,0 +1,14 @@
+Name:Depth Defiler
+ManaCost:3 U U
+Types:Creature Eldrazi
+PT:3/5
+K:Devoid
+K:Kicker:C
+T:Mode$ SpellCast | ValidCard$ Card.Self | Execute$ TrigCharm | TriggerDescription$ When you cast this spell, choose one. If it was kicked, choose both instead.
```
tool4ever, line 7: I think you need to work with ABILITY so the full text shows up ingame

tool4ever (approved):

PR #5344, opened 2024-05-31
MH3: 3 cards

Northmoc:


forge-gui/res/cardsfolder/upcoming/breaker_of_creation.txt
```diff
@@ -0,0 +1,11 @@
+Name:Breaker of Creation
+ManaCost:6 C C
+Types:Creature Eldrazi
+PT:8/4
+T:Mode$ SpellCast | ValidCard$ Card.Self | Execute$ TrigGainLife | TriggerDescription$ When you cast this spell, you gain 1 life for each colorless permanent you control.
+SVar:TrigGainLife:DB$ GainLife | LifeAmount$ Count$Valid Permanent.Colorless+YouCtrl
+K:Hexproof:Card.nonColorless:each color
```
Hanmac, line 7: Imo, works for now, i think 

The corner case that for some reason a specific `Hexproof from red` is removed that would cause this to split up currently doesn't exist

Hanmac (approved):

PR #5352, opened 2024-06-01
MH3: detectives_phoenix + support

Northmoc:


forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -2669,8 +2671,22 @@ public static void addSpellAbility(final KeywordInterface inst, final CardState
...
+            final Cost bCost = new Cost(cost, false);
+            final boolean onlyMana = bCost.isOnlyManaCost();
+            String remTxt = inst.getReminderText();
+            if (!onlyMana) {
+                String oText = card.getCard().getOracleText();
+                String bestowStr = oText.substring(indexOf(oText, "Bestow"), indexOf(oText, "\\n"));
+                if (bestowStr.contains("(")) {
+                    String s = bestowStr.substring(indexOf(bestowStr, "(") + 1, indexOf(bestowStr, ")"));
+                    if (!s.equals("")) remTxt = s;
+                }
```
Hanmac, line 2685: The looking at the OracleText doesn't look nice to me

maybe we can change it later to something better?

Northmoc: Yeah, I thought if we had some more templating examples we would have a better place to start.

Just having the one with the non mana cost is a bit weird

tool4ever (approved):

PR #5365, opened 2024-06-03
Support MayPlay on extrinsic AltCost

### LINKED ISSUE #4725: Creatures being granted Blitz through Henzie can't be cast from the graveyard or Exile even though an ability allows you to do so
**Describe the bug**
If you have Henzie "Toolbox" Torre on the field and an ability or effect allows you to cast a creature from your grave or exile, you can't cast that spell for it's Blitz cost. (creature with cmc 4 or More) 

**To Reproduce**
1. Henzie on the field
2. An abilitiy allowing you to cast a creature from graveyard or Exile. For example Chainer, Nightmare Adept. 
3. Try casting a creature spell with cmc 4 or more from the graveyard or Exile. 
4. You are only given the option to cast the creature for its normal cost, Not its Blitz cost

**Expected behavior**
You should be able to cast the creature for its Blitz cost from the grave/Exile. 
This was possible in an earlier Version. 

 - Device: Honor View20
 - OS: Android 10
 - Forge Version forge-android-1.6.60-SNAPSHOT-02-18

**Additional context**
Tested on Android.
Tested with Chainer, Nightmare Adept and Commune with Lava.

tool4ever:
Not pretty but works:

image

Closes #4725

### DIFF
--- a/forge-game/src/main/java/forge/game/GameActionUtil.java
+++ b/forge-game/src/main/java/forge/game/GameActionUtil.java
@@ -228,6 +228,7 @@ public final class GameActionUtil {
 
                 // some needs to check after ability was put on the stack
                 if (game.getAction().hasStaticAbilityAffectingZone(ZoneType.Stack, StaticAbilityLayer.ABILITIES)) {
+                    Map<StaticAbility, CardPlayOption> oldMayPlay = source.getMayPlay();
                     Zone oldZone = source.getLastKnownZone();
                     Card stackCopy = source;
                     if (!source.isLKI()) {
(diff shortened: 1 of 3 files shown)


Hanmac:
Imo Card.getMayPlay needs to be refactored/removed anyway

Checkout my very old attempt

The Remaining MayPlay should also be implemented as Static like this:
https://github.com/Card-Forge/forge/compare/master...mayPlayAlternativeCost#diff-faf3b1b30bd35b40faaa6aa6e5a44dad244a470bb96f790524905a27c7e010abR57

tool4ever:
Fair enough
I thought you wanted a quick fix but let's just slowly pivot towards the real solution :)

Hanmac:
@tool4ever i think this still can be merged 


i just mean it would be better if we clean up mayPlay sooner than later

Hanmac (approved):

Issue #5371, opened 2024-06-04
Shell scripts aren't passing down command line args to java call


tehdiplomat:
**Describe the bug**
Trying to run command line simulations via the .sh files just launches Forge, and doesn't launch the command line mode. 

https://discord.com/channels/267367946135928833/1247522576784425010

**To Reproduce**
1. `forge.sh sim -d "The rack.dck" "The rack.dck" -n 3`

**Expected behavior**
Sim mode runs

**Screenshots**

PR #5372, opened 2024-06-04
Update forge.sh

tehdiplomat:
https://github.com/Card-Forge/forge/issues/5371

### DIFF
--- a/forge-gui-desktop/src/main/config/forge.sh
+++ b/forge-gui-desktop/src/main/config/forge.sh
@@ -38,7 +38,7 @@ jdk_version() {
 }
 v="$(jdk_version)"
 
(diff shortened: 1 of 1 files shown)


forge-gui-desktop/src/main/config/forge.sh
```diff
@@ -38,7 +38,7 @@ jdk_version() {
 }
 v="$(jdk_version)"
 
-SHAREDPARAMS='-Xmx4096m -Dfile.encoding=UTF-8 -jar $project.build.finalName$'
```
tehdiplomat, line 41: This isn't quite right. It doesn't seem to be escaping spaces right

Agetian: :+1:

Hanmac (approved):

PR #5385, opened 2024-06-06
MH3: volatile_stormdrake + support

Northmoc:


forge-gui/res/cardsfolder/upcoming/volatile_stormdrake.txt
```diff
@@ -0,0 +1,14 @@
+Name:Volatile Stormdrake
+ManaCost:1 U
+Types:Creature Drake
+PT:3/2
+K:Flying
+K:Hexproof:Activated,Triggered:activated and triggered abilities
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigExchange | TriggerDescription$ When CARDNAME enters the battlefield, exchange control of CARDNAME and target creature an opponent controls. If you do, you get {E}{E}{E}{E}, then sacrifice that creature unless you pay an amount of {E} equal to its mana value.
+SVar:TrigExchange:DB$ ExchangeControl | Defined$ Self | ValidTgts$ Creature.OppCtrl | TgtPrompt$ Select target creature an opponent controls | CantFizzle$ True | RememberExchanged$ 1 | SubAbility$ Energy
```
tool4ever, line 8: cantfizzle bad copy pasta

forge-gui/res/cardsfolder/upcoming/volatile_stormdrake.txt
```diff
@@ -0,0 +1,14 @@
...
+SVar:Energy:DB$ PutCounter | ConditionDefined$ Remembered | ConditionPresent$ Card | Defined$ You | CounterType$ ENERGY | CounterNum$ 4 | SubAbility$ Sacrifice
+SVar:Sacrifice:DB$ Sacrifice | ConditionDefined$ Remembered | ConditionPresent$ Card | Defined$ Remembered | UnlessCost$ PayEnergy<N> | UnlessPayer$ You | SubAbility$ Cleanup
```
tool4ever, line 10: why not just Defined$ Targeted?
no reason for extra logic in ControlExchange 🤔

forge-game/src/main/java/forge/game/ability/effects/SacrificeEffect.java
```diff
@@ -96,11 +96,15 @@ public void resolve(SpellAbility sa) {
...
 
         if (valid.equals("Self") && game.getZoneOf(host) != null) {
-            if (game.getZoneOf(host).is(ZoneType.Battlefield)) {
-                if (!optional || activator.getController().confirmAction(sa, null,
-                        Localizer.getInstance().getMessage("lblDoYouWantSacrificeThis", host.getName()), null)) {
-                    if (game.getAction().sacrifice(host, sa, true, params) != null && remSacrificed) {
-                        host.addRemembered(host);
+            valid = sa.hasParam("Defined") ? sa.getParam("Defined") : valid;
+            List<Card> toSac = AbilityUtils.getDefinedCards(host, sa.getParam("Defined"), sa);
+            for (Card c : toSac) {
```
tool4ever, line 101: traditionally (and somewhat weirdly) we've used SacrificeAll for Defined...
I guess you want to move away from it?

forge-gui/res/cardsfolder/upcoming/volatile_stormdrake.txt
Northmoc: Ah good point

I changed ControlExchange first
Then realized the issue was really that SacrificeEffect couldn't handle "Defined" well

forge-game/src/main/java/forge/game/ability/effects/SacrificeEffect.java
Northmoc: Yeah, every time it just seems counterintuitive

I can bring it up in the discord to see if others mind

Hanmac: I don't know if we should use DefinedCard there or not ...

In Theory, we could use both ways for `getTargetPlayers(sa)` & `getTargetCards(sa)` but that would mean a bigger rewrite

Northmoc: So would it be better to:
 - Leave this "DefinedCard" for now and refactor when we're more sure what to do
 - Just use SacrificeAllEffect and save this whole discussion for a rainy day
 - Try to refactor SacrificeEffect in this PR (I'm good to try, but don't want to do it if we aren't ready yet)

Hanmac: imo "Just use SacrificeAllEffect and save this whole discussion for a rainy day"

we could do it later

tool4ever (approved):

PR #5393, opened 2024-06-07
M3C: overclocked_electromancer + support

Northmoc:


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -269,6 +269,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
     private Map<Integer, Integer> damage = Maps.newHashMap();
     private boolean hasBeenDealtDeathtouchDamage;
     private boolean hasBeenDealtExcessDamageThisTurn;
```
Hanmac, line 271: Can this variable be removed if excessDamageThisTurnAmount > 0?

Northmoc: I wondered about that... I think so. I was just worried about the LKI part in CardDamageMap line 131

tool4ever: yea the logic for _Rith, Liberated Primeval_ probably needs to be refactored for that to work:
> if a creature or planeswalker an opponent controlled was dealt excess damage this turn

Northmoc: Maybe leave both for now and see what WotC does next? They seem to be experimenting with "excess damage" right now.

tehdiplomat (approved):

PR #5419, opened 2024-06-12
Add puzzle PS_OTJ4

Agetian:
No idea what they actually meant here:
https://i2.wp.com/www.possibilitystorm.com/wp-content/uploads/2024/05/latest-1-scaled.jpg?ssl=1

They draw 3 Restless Bivouac and 2 Shivan Reef, but say in text that there are 2 Restless Bivouac and 3 Shivan Reef :/ Any ideas, guys?...


Hanmac:
you could email them about their error?

Agetian:
True, might do that :)

Agetian:
E-mailed them regarding the error, but no answer so far, hehe :)
They weren't kidding when they said it's a puzzle - go figure what they meant, haha :D

Agetian:
Hmm, they're not responding to my requests, so I'm going to merge as is for now with the TODO entry (assuming that the graphical representation is correct), and I'll patch it up as a followup in case they fix it the other way at some point.

PR #5422, opened 2024-06-13
Oracle updates

Fulgur14:
Changed Cephalid -> Octopus
Added Cat to Freejam Regent and Skyship Stalker
Changed protection from all colors to protection from each color

If accepted, "Cephalid" can be removed from the list of creature types.


Hanmac:
for the Protection thing, i need to update the Engine first

PR #5428, opened 2024-06-15
Piper-like card update

NishacIroCode:
The objective of this PR is to address the concern associated with https://discord.com/channels/267367946135928833/1175324774067613716/1175324774067613716.  This will theoretically allow Clara Oswald, Faceless One, and The Prismatic Piper to be used as commanders and/or partners.

### DIFF
--- a/forge-core/src/main/java/forge/card/CardRules.java
+++ b/forge-core/src/main/java/forge/card/CardRules.java
@@ -52,6 +52,7 @@ public final class CardRules implements ICardCharacteristics {
     private ColorSet deckbuildingColors;
     private String meldWith;
     private String partnerWith;
+    private boolean addsWildCardColor;
     private boolean custom;
 
     public CardRules(ICardFace[] faces, CardSplitType altMode, CardAiHints cah) {
@@ -70,6 +71,7 @@ public final class CardRules implements ICardCharacteristics {
         aiHints = cah;
         meldWith = "";
         partnerWith = "";
+        addsWildCardColor = false;
 
         //calculate color identity
         byte colMask = calculateColorIdentity(mainPart);
@@ -92,6 +94,7 @@ public final class CardRules implements ICardCharacteristics {
         colorIdentity = newRules.colorIdentity;
         meldWith = newRules.meldWith;
         partnerWith = newRules.partnerWith;
+        addsWildCardColor = newRules.addsWildCardColor;
         tokens = newRules.tokens;
     }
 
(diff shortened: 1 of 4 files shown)


forge-core/src/main/java/forge/deck/DeckFormat.java
```diff
@@ -204,17 +205,19 @@ public String getDeckConformanceProblem(Deck deck) {
...
-        boolean noBasicLands = false;
+        // boolean noBasicLands = false;
 
         // Adjust minimum base on number of Advantageous Proclamation or similar cards
         CardPool conspiracies = deck.get(DeckSection.Conspiracy);
         if (conspiracies != null) {
             min -= (5 * conspiracies.countByName(ADVPROCLAMATION, false));
-            noBasicLands = conspiracies.countByName(SOVREALM, false) > 0;
+            // noBasicLands = conspiracies.countByName(SOVREALM, false) > 0;
+            conspiracies.countByName(SOVREALM, false);
```
tehdiplomat, line 215: Whats this change about?

forge-core/src/main/java/forge/item/PaperCard.java
```diff
@@ -38,7 +37,7 @@
  *
  * @author Forge
  */
-public class PaperCard implements Comparable<IPaperCard>, InventoryItemFromSet, IPaperCard, Serializable {
+public class PaperCard implements Comparable<IPaperCard>, InventoryItemFromSet, IPaperCard {
```
tehdiplomat, line 40: Let's keep the random cleanup out of this PR since you aren't even in this file.

forge-core/src/main/java/forge/deck/DeckFormat.java
NishacIroCode: Removing warnings.  I left the call to countByName because I didn't want to disrupt the code by removing it entirely, but the noBasicLands boolean is being assigned, but never used (throwing a warning).

forge-core/src/main/java/forge/item/PaperCard.java
NishacIroCode: I initially went into that file and made a modification, but then realized I didn't need it and left the cleanup.

tool4ever: to what end is making it unserializable cleanup...?

forge-core/src/main/java/forge/card/CardRules.java
```diff
@@ -485,6 +497,7 @@ public final CardRules getCard() {
             result.setNormalizedName(this.normalizedName);
             result.meldWith = this.meldWith;
             result.partnerWith = this.partnerWith;
+            result.addsWildCardColor = this.addsWildCardColor;
```
tool4ever, line 500: there's no point for changing the Reader in your current code?

forge-core/src/main/java/forge/item/PaperCard.java
NishacIroCode: IPaperCard implements it, so it was implementing it twice.

forge-core/src/main/java/forge/card/CardRules.java
NishacIroCode: From my understanding, this represents the in-game display of the card.  In this update, I did not make it so that these cards would reflect their proper color in game, but I felt it would be best to make the modification that would probably be needed for future development at the same time.

forge-core/src/main/java/forge/deck/DeckFormat.java
tehdiplomat: CountByName doesn't' really do anything except return a number. While I agree assigning to noBasicLands doesn't do anything, it at least is a place holder for someone willing to add logic for decks that don't need to have basics in them. Removing that assignment just removes all of the context of this code, and it feels worse because of that change.

NishacIroCode: With code that I commit, I refuse to leave any warnings in it, so I will comment the call and boolean initialization lines.

NishacIroCode: As this is now outdated, resolving.

tehdiplomat (approved):

PR #5439, opened 2024-06-19
Keyword: Freerunning

### LINKED ISSUE #5438: ACR: Freerunning
Alternative Cost similar to Prowl

Hanmac:
Closes #5438

### DIFF
--- a/forge-game/src/main/java/forge/game/player/Player.java
+++ b/forge-game/src/main/java/forge/game/player/Player.java
@@ -2140,6 +2140,10 @@ public class Player extends GameEntity implements Comparable<Player> {
         return !game.getDamageDoneThisTurn(true, true, sb.toString(), "Player", null, this, null).isEmpty();
     }
 
+    public final boolean hasFreerunning() {
+        return !game.getDamageDoneThisTurn(true, true, "Card.Assassin+YouCtrl,Card.IsCommander", "Player", null, this, null).isEmpty();
+    }
+
     public final void setLibrarySearched(final int l) {
         numLibrarySearchedOwn = l;
     }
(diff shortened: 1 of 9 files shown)


forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -2140,6 +2140,10 @@ public final boolean hasProwl(final Set<String> types) {
         return !game.getDamageDoneThisTurn(true, true, sb.toString(), "Player", null, this, null).isEmpty();
     }
 
+    public final boolean hasFreerunning() {
+        return !game.getDamageDoneThisTurn(true, true, "Card.Assassin+YouCtrl,Card.IsCommander", "Player", null, this, null).isEmpty();
```
tool4ever, line 2144: I think both valids need YouCtrl?

Hanmac: Yeah

forge-game/src/main/java/forge/game/spellability/SpellAbilityRestriction.java
```diff
@@ -444,6 +444,11 @@ public final boolean checkOtherRestrictions(final Card c, final SpellAbility sa,
                 return false;
             }
         }
+        if (sa.isFreerunning()) {
```
Hanmac, line 447: @tool4ever 
It might be better to make this more Generic? Same with Prowl?

tool4ever: like scripting it into SVarToCheck directly in CardFactory so it's all in one place?
I wonder if there could be a risk of overwriting from collision but most likely it's safe 🤔

Hanmac: For Freerunning we could make the Valid Hardcoded

for Prowl, we could change it to sharesCreatureTypeWith Spell Host

But yeah, we might do that later?

tool4ever (approved):

PR #5454, opened 2024-06-22
StaticAbilityBlockRestict: remove Global Rules

Hanmac:
@Northmoc @tool4ever just one GlobalRule remaining

### DIFF
--- a/forge-game/src/main/java/forge/game/combat/CombatUtil.java
+++ b/forge-game/src/main/java/forge/game/combat/CombatUtil.java
@@ -24,7 +24,6 @@ import com.google.common.collect.Lists;
 import forge.card.mana.ManaCost;
 import forge.game.Game;
 import forge.game.GameEntity;
-import forge.game.GlobalRuleChange;
 import forge.game.ability.AbilityKey;
 import forge.game.card.*;
 import forge.game.cost.Cost;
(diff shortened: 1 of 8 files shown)


forge-game/src/main/java/forge/game/combat/CombatUtil.java
```diff
@@ -460,18 +460,14 @@ public static boolean canBlock(final Card blocker, final Combat combat) {
         final Game game = blocker.getGame();
         final int blockers = combat.getAllBlockers().size();
 
-        if (blockers > 1 && game.getStaticEffects().getGlobalRuleChange(GlobalRuleChange.onlyTwoBlockers)) {
-            return false;
-        }
-
-        if (blockers > 0 && game.getStaticEffects().getGlobalRuleChange(GlobalRuleChange.onlyOneBlocker)) {
+        if (blockers >= StaticAbilityBlockRestrict.globalBlockRestrict(game)) {
```
tool4ever, line 463: I'm currently thinking we might not need two / it was wrong:
> 802.4b. When determining whether a defending player's blocks are legal, ignore any creatures attacking other players and any blocking creatures controlled by other players.

This matters e. g. with Myriad and _Dueling Grounds_
otherwise only the first defender can make a block?

so it should always only check per player...

Hanmac: https://magicjudge.tumblr.com/post/137740905557/multiplayer-game-i-have-a-silent-arbiter-and
> Attacking and blocking restrictions like Silent Arbiter’s apply separately to each player, so each defending player can block with one creature

Seems like it

So I try to remove the global ones

tool4ever (approved):

Northmoc (approved):
Excellent!

Issue #3582, opened 2023-08-01
DeclareCombatantsEffect needs to use timestamps because of different durations


tool4ever:
Example:

Cast _Master Warcraft_ followed by _Melee_.

If you get a second combat phase the effect from Master Warcraft (which lasts whole turn) is lost, because the `removeOverrides` gets executed by Melee.

PR #5458, opened 2024-06-23
Invasion Plans using Timestamp instead

Hanmac:
removed the last of the GlobalRules


forge-gui/res/cardsfolder/i/invasion_plans.txt
```diff
@@ -2,7 +2,8 @@ Name:Invasion Plans
 ManaCost:2 R
 Types:Enchantment
 S:Mode$ MustBlock | ValidCreature$ Creature | Description$ All creatures block each combat if able.
-S:Mode$ Continuous | GlobalRule$ The attacking player chooses how each creature blocks each combat. | Description$ The attacking player chooses how each creature blocks each combat.
+R:Event$ DeclareBlocker | Layer$ Control | ReplaceWith$ DBUpdateChooser | Description$ The attacking player chooses how each creature blocks each combat.
```
tool4ever, line 5: this shouldn't be RE that let's you choose
this matters especially if multiple _Master Warcraft_ would resolve

Hanmac: Then Master Warcraft would need to be reworked too

The thing probably needs to use Card Timestamp logic?

tool4ever: yea #3582

Hanmac:
@tool4ever we could also refactor the remaining cards and remove DeclareCombatantsEffect

Hanmac:
the cards updated

there is one Problem, Brutal Hordechief has this:
> Creatures your opponents control block this turn if able, and you choose how those creatures block.

this has some corner cases, if you have a teammate
* if on your opponents turn, they attack other opponents, but also your teammate
* you should be able to choose how those block for opponents creatures, but not for your teammate

tool4ever:
Hmn, I guess instead of `Declare*$ True` we can use Defined to store that as PlayerCollection

If the data fields stay in Game it will need more logic to find the right declarer

Or we move them to Player instead, we might even provide the old declarers as the Affected$ - that would make it easier to grab the new one for the turn based action? 🤔 
_Opposition Agent_ already does it like that

Hanmac:
@tool4ever i think my new idea using `| Affected$ Player.Opponent | DeclaresBlockers$ You` should work

tool4ever (approved):
nice work

PR #5465, opened 2024-06-25
Cipher: do it like Epic

tool4ever:


Hanmac (approved):
As long as no other card would grant Cipher, then it shouldn't be a problem.

PR #5474, opened 2024-06-27
Genku trigger description

tehdiplomat:


forge-gui/res/cardsfolder/upcoming/genku_future_shaper.txt
```diff
@@ -2,7 +2,7 @@ Name:Genku, Future Shaper
 ManaCost:2 W U
 Types:Legendary Creature Moonfolk Wizard
 PT:2/5
-T:Mode$ ChangesZone | Origin$ Battlefield | Destination$ Any | ValidCard$ Permanent.YouCtrl+nonToken | TriggerZones$ Battlefield | Execute$ TrigCharm | TriggerDescription$ Whenever another nontoken permanent you control leaves the battlefield, choose one that hasn't been chosen this turn. Create a creature token with those characteristics.
+T:Mode$ ChangesZone | Origin$ Battlefield | Destination$ Any | ValidCard$ Permanent.YouCtrl+nonToken | TriggerZones$ Battlefield | Execute$ TrigCharm | TriggerDescription$ Whenever another nontoken permanent you control leaves the battlefield, create a creature token with one of the following characteristics - ABILITY.
```
Hanmac, line 5: Where does the "choose one that hasn’t been chosen this turn." go? shouldn't that also be part of the Trigger Description?

tehdiplomat: It's already being added by the ABILITY macro 
image

Hanmac: Hm, while the trigger might not be the same, I think that's good enough

PR #5486, opened 2024-06-29
UNF Nearby Planet

NCat39:
Attempted to add UNF's Nearby Planet using the Omo effect from M3C (and to a lesser extent Planar Nexus from MH3) as a reference.

Tested card in Forge, seems to work as intended. The displayed typeline is messier than with the Mistform Ultimus/changeling effect for creatures (Mistform displays "(All)" whereas this spells out the types), but it is consistent with the referenced Modern Horizons cards.

### DIFF
--- a/forge-gui/res/cardsfolder/n/nearby_planet.txt
+++ b/forge-gui/res/cardsfolder/n/nearby_planet.txt
@@ -1,0 +1,8 @@
+Name:Nearby Planet
+ManaCost:no cost
+Types:Land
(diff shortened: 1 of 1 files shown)


Simisays:
@paulsnoops I think this is the first nonlegal UNF card ? Might this mess up the legality for vintage and legacy ?

NCat39:
> @paulsnoops I think this is the first nonlegal UNF card ? Might this mess up the legality for vintage and legacy ?

I made sure to confirm that UNF had nonlegal cards already. Killer Cosplay is implemented already

paulsnoops:
> @paulsnoops I think this is the first nonlegal UNF card ? Might this mess up the legality for vintage and legacy ?

It will show up as legal yeah but so do other acorn cards for now. The unset cards are already marked as "funny" in the edition we just need to filter them out of formats by default without adding them to the ban list etc. We can go ahead with this and figure it out later.

PR #5495, opened 2024-06-30
Backup Plan with hot swappable Input

tehdiplomat:


forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -1341,6 +1352,15 @@ public void updateAllZonesForView() {
         }
     }
 
+    public final List<PlayerZone> getExtraZones() {
+        return extraZones;
+    }
+
+    public void resetExtraZones(ZoneType type) {
+        extraZones.removeIf(z -> z.getZoneType().equals(type));
```
tehdiplomat, line 1360: After we might want to set to null if extraZones is empty

PR #5504, opened 2024-07-02
If Commander would goto hand, don't move it to command zone

tehdiplomat:


forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -1718,6 +1718,16 @@ public final boolean aiShouldRun(final ReplacementEffect effect, final SpellAbil
             hostCard = game.getCardState(hostCard);
         }
 
+        if (effect.hasParam("CommanderMoveReplacement")) {
```
tool4ever, line 1721: what is this for - 
it should already reach the logic in `ChangeZoneAi.doReturnCommanderLogic`?

tehdiplomat: Ah i didn't know about that function. But what I'm adding is still needed. Bouncing to hand shouldn't send to Command Zone.

tool4ever: yea most likely only in corner cases
might still be good to move it to the same place? 🤔

tehdiplomat: Yep, I'll move it to the function you pointed out. THanks for the eyes.

tool4ever (approved):

PR #5512, opened 2024-07-04
Fix Avatar Creature type being treated as Vanguard avatar

Jetz72:
I looked around and couldn't find a reason that this deck section validator also includes the creature type. If anyone knows why then I can try tackling this a different way.


tehdiplomat:
Yea looks like the person who refactored this just didn't know that Vanguards used to be called Avatars.

tehdiplomat (approved):

PR #5515, opened 2024-07-05
Draft changes

Simisays:
- Decreased the minimum amount of cards needed for a draft to be valid 
- Increased the chance of nonstandard sets showing up
- decreased the chance of jumpstart


forge-gui-mobile/src/forge/adventure/data/AdventureEventData.java
```diff
@@ -303,11 +303,11 @@ public BoosterDraft getDraft() {
     public static Predicate<CardEdition> selectSetPool() {
         final int rollD100 = MyRandom.getRandom().nextInt(100);
         Predicate<CardEdition> rolledFilter;
-        if (rollD100 < 40) {
+        if (rollD100 < 30) {
```
ianelletson, line 306: Love this; I think it would be great to expose these constants in the configuration file, but that change can and should be separate from this one.

tehdiplomat (approved):

PR #5544, opened 2024-07-07
Keyword: Impending

### LINKED ISSUE #5481: DSK: Impending Keyword
> Impending N—Cost (If you cast this spell for its impending cost, it enters with N time counters and isn’t a creature until the last is removed. At the beginning of your end step, remove a time counter from it.)

Similar to Fading ...
* ETB with Time Counters
* At end step, remove time counter
* StaticAbility(?) that stops it from being a creature while it has time counters? (This also could be a pump Effect with Duration)

Hanmac:
Closes #5481

First try of Impending Keyword and Everywhere Land Token. These might be updated later when we know more about the rules.

The Current Implementation of Impending uses a Static Ability that can be enabled by any Time counter.

The Everywhere Land currently has the Land types hard coded, so it can be affected by Card Text Changes.
While The Creature should not, that's why the "LockTokenScript"

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -2494,7 +2494,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
                         || keyword.startsWith("Bestow") || keyword.startsWith("Surge")
                         || keyword.startsWith("Transmute") || keyword.startsWith("Suspend")
                         || keyword.startsWith("Dash") || keyword.startsWith("Disturb")
-                        || keyword.equals("Undaunted") || keyword.startsWith("Monstrosity")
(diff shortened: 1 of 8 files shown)


Agetian (approved):

tool4ever:
debut stream only one day away - we could wait if that confirms some things already but then it seems fine

Hanmac:
> debut stream only one day away - we could wait if that confirms some things already but then it seems fine

No, this is for the Next Set in September 😜

tool4ever (approved):

Hanmac:
Info for the Land Token:
https://x.com/WotC_Matt/status/1810351857644326935
> I would imagine it has no printed land types. Where would you put them?

so maybe it is a static? maybe not? We could change that later?

PR #5558, opened 2024-07-10
Keyword: Offspring

Hanmac:
* I don't know if linkedCastSA is enough or if I need some extra Valid Check for the ETB Trigger
* I don't know if The "Offspring {2}" should be in the Trigger Description? Currently, it isn't Parsed right by the Stack Description?
* I need to Update the Keyword Description for the Card. It is still "Offspring:2"


forge-game/src/main/java/forge/game/spellability/StackItemView.java
```diff
@@ -66,6 +67,8 @@ void updateOptionalCost(SpellAbilityStackInstance si) {
             if (cost == OptionalCost.Kicker1 || cost == OptionalCost.Kicker2)
                 kicked = true;
             if (cost == OptionalCost.Entwine)
+                offspring = true;
+            if (cost == OptionalCost.Offspring)
```
tool4ever, line 71: switch booleans

Hanmac: imo we should look to rewrite updateOptionalCost in a later day

tool4ever:
Hopefully they didn't create extrinsic Offspring
Otherwise we can't use `linkedCastSA` like on Squad

Hanmac:
> Hopefully they didn't create extrinsic Offspring Otherwise we can't use `linkedCastSA` like on Squad

yeah ... but i think, i still want to change it with "is cast with Offering" instead

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -1594,6 +1594,24 @@ public static void addTriggerAbility(final KeywordInterface inst, final Card car
...
+            String costDesc = cost.toSimpleString();
+            if (!cost.isOnlyManaCost()) {
+                costDesc = "—" + costDesc;
+            }
+
+            final String trigStr = "Mode$ ChangesZone | Destination$ Battlefield " +
+                    "| ValidCard$ Card.Self+linkedCastSA | Secondary$ True " +
+                    "| TriggerDescription$ Offspring " + costDesc + " (" + inst.getReminderText() + ")";
+
+            final String effect = "DB$ CopyPermanent | Defined$ TriggeredCard | NumCopies$ 1 | SetPower$ 1 | SetToughness$ 1";
```
tool4ever, line 1609: should be LKICopy?

Hanmac: i copied from Squad ... maybe that needs LKICopy too

Hanmac:
> Hopefully they didn't create extrinsic Offspring Otherwise we can't use `linkedCastSA` like on Squad

There is one, with Zinnia, Valley's Voice
> Creature spells you cast have offspring {2}

Hanmac:
@tool4ever the Cost and the Tigger are Probably linked, so i need to do something similar to this like with Conspire:
https://github.com/Card-Forge/forge/blob/7e200eae8bec038c50a2f6aa26877b5c08c3ff40/forge-game/src/main/java/forge/game/GameActionUtil.java#L597-L616

tool4ever:
this is also a problem for _Adipose Offering_ cast via _Herigast_

Hanmac:
> this is also a problem for _Adipose Offering_ cast via _Herigast_

Emerge is different, because it's an Alternative Cost, not an Optional Cost.

Squad for example has `If a spell has multiple instances of squad, each is paid separately.`

Hanmac:
> this is also a problem for _Adipose Offering_ cast via _Herigast_

I asked Rules Manager:
https://x.com/Hanmac10/status/1811001507006873839

Hanmac:
@tool4ever i updated it with `Zinnia, Valley's Voice` that grants Offspring to all your Creature Spells

Hanmac:
I asked Rules Manager if the trigger would be an reflexive/delayed trigger 

https://x.com/Hanmac10/status/1812432938203525347

Because that might complicate the code (and makes it not work with Delney)

tool4ever:
There is still the problem that when you cast a _Clone_ and let it copy a creature which had Offspring paid it will trigger for the copy too :/

Hanmac:
> There is still the problem that when you cast a _Clone_ and let it copy a creature which had Offspring paid it will trigger for the copy too :/

yeah that we need to check out too :(

sadly i haven't had an Answer for my "Delayed Trigger" Question

Fulgur14:
@Hanmac Probably, yes.

tool4ever:
@Hanmac
I think I fixed it by checking that the stack LKI already had the same trigger ID

tool4ever (approved):

Hanmac:
> @Hanmac I think I fixed it by checking that the stack LKI already had the same trigger ID

i was thinking about a better way last night, but that might be done when we get CR

PR #5594, opened 2024-07-13
Support another case of reflexive triggers getting duplicated

tool4ever:
> In a case where you create more than one Cat Warrior creature token with Ajani, Nacatl Avenger's second ability (for example, because of the effect of Doubling Season), the reflexive ability will trigger once for each Cat Warrior you made this way. If a replacement effect causes you to create different tokens instead (for example, the effect of Divine Visitation), the creation of those tokens won't cause the reflexive ability to trigger.


forge-game/src/main/java/forge/game/ability/effects/TokenEffectBase.java
```diff
@@ -191,6 +191,10 @@ protected TokenCreateTable makeTokenTable(TokenCreateTable tokenTable, final boo
                 if (sa.hasParam("RememberTokens")) {
                     host.addRemembered(moved);
                 }
+                // used for some reflexive trigger
+                if (sa.hasParam("RememberOriginalTokens") && moved.getRules().getNormalizedName().equals(sa.getParam("TokenScript"))) {
```
tool4ever, line 195: @Hanmac
can we do a better check? though it does seem to work 🤔

Hanmac: Compare with original TokenTable from before the Replacement Effect in line 85

Hanmac (approved):
i think that should be okay for now?

I think maybe better would be to copy the old Table

then you can do `originalTokens.contains(p, prototype)` to check if the prototype is still with the player in case a prototype is done for multiple players?

tool4ever:
it looks like if there are multiple owner they get their own prototype anyway

PR #5598, opened 2024-07-13
Fix for QuestCommander deck editor

Jetz72:
Resolves an issue where quest commander was functioning as quest mode in some ways, like commander mode in other ways, and like quest commander in no way whatsoever.

In doing this, I deleted this condition from the deck editor's constructor:
```
if (editorType0 == EditorType.QuestCommander) //fix saving quest commander
    editorType = EditorType.Quest;
else
    editorType = editorType0;
```
Removing something commented with the words "fix saving" leaves me very nervous because I'm not sure exactly what the issue there was. I did run into these lines in `QuestMenu.java` that seem like they could cause a quest commander deck saving issue since they don't include `EditorType.QuestCommander`:
```
((DeckController<Deck>)EditorType.Quest.getController()).setRootFolder(FModel.getQuest().getMyDecks());
((DeckController<DeckGroup>)EditorType.QuestDraft.getController()).setRootFolder(FModel.getQuest().getDraftDecks());
```

I added such a line in, and haven't encountered any issues while testing it. Still, if that's all it needed, I'm not sure why the workaround solution was used. 

All this to say, if anybody knows any more details about this, please chime in.

### DIFF
--- a/forge-core/src/main/java/forge/deck/Deck.java
+++ b/forge-core/src/main/java/forge/deck/Deck.java
@@ -539,7 +539,12 @@ public class Deck extends DeckBase implements Iterable<Entry<DeckSection, CardPo
     public CardPool getAllCardsInASinglePool() {
         return getAllCardsInASinglePool(true);
     }
+
     public CardPool getAllCardsInASinglePool(final boolean includeCommander) {
+        return getAllCardsInASinglePool(includeCommander, false);
+    }
+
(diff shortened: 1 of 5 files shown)


tehdiplomat:
@Jetz72 the person who added commander quest was very unfamiliar with the codebase and slapped it together without knowing how things should be done

tehdiplomat (approved):

PR #5619, opened 2024-07-16
Added Blacksmith's Talent + Sword token

Fulgur14:


forge-gui/res/cardsfolder/upcoming/dour_port_mage.txt
```diff
@@ -0,0 +1,8 @@
+Name:Dour Port-Mage
+ManaCost:1 U
+Types:Creature Frog Wizard
+PT:1/3
+T:Mode$ ChangesZoneAll | ValidCards$ Creature.Other+YouCtrl | Origin$ Battlefield | Destination$ Ante,Command,Exile,Hand,Library | TriggerZones$ Battlefield | Execute$ TrigDraw | TriggerDescription$ Whenever one or more other creatures you control leave the battlefield without dying, draw a card.
```
Hanmac, line 5: @tool4ever ChangesZoneAll currently doesn't have `ExcludedDestinations` yet, would that be better?

Fulgur14: I think it would. I'm bothered by the Ante zone here, but it's necessary because of Jeweled Bird.

Fulgur14: Is there any remaining problem with this commit?

Hanmac: @tool4ever we could move this to an extra Issue about adding ExcludedDestinations to ChangesZoneAll

tool4ever: Sure, I'm trying to stick to reviewing only on weekdays anyway ;)

tool4ever (approved):

Hanmac:
@tool4ever it seems `Destination$ Ante,Command,Exile,Hand,Library` didn't work at all and is currently breaking

PR #5663, opened 2024-07-20
Add Attraction Light Icons

Jetz72:
Adds Attraction Light icons to the deck builders and card details.

Refactored functional variants to be merged in the card face rather than the card factory. This was done because CardRules has no awareness of the PaperCard's chosen variant. New methods on IPaperCard - getMainFace and getOtherFace, now allow easy access to faces with variant properties applied when the full card object hasn't yet been created (e.g. in the deck builder card list).

"[card] has neither ManaCost nor Color" warning now actually displays when these characteristics are missing. I corrected a couple of these.


forge-core/src/main/java/forge/card/CardFace.java
```diff
@@ -28,11 +28,12 @@ public enum FaceSelectionMethod { //
     
     private final static List<String> emptyList = Collections.unmodifiableList(new ArrayList<>());
     private final static Map<String, String> emptyMap = Collections.unmodifiableMap(new TreeMap<>());
+    private final static Set<Integer> emptySet = Collections.unmodifiableSet(new HashSet<>());
 
     private String name;
     private String altName = null;
     private CardType type = null;
-    private ManaCost manaCost = ManaCost.NO_COST;
+    private ManaCost manaCost = null;
```
Hanmac, line 36: I don't know about this case, I know it wouldn't trigger `if ( manaCost == null && color == null )`
but I would prefer if we still have this "No_Cost" as Default

@tool4ever your opinion?

Hanmac:
@Jetz72 @tool4ever I don't know if it would be better, but what about using `BitSet` for the Attraction Lights?

forge-core/src/main/java/forge/card/CardFace.java
Jetz72: The line added to `assignMissingFields` - `if ( manaCost == null ) manaCost = ManaCost.NO_COST;` - should ensure that no CardFace ever makes it out of the card reader with this null mana cost. I mainly changed it so that a variant face can know when to fall back to the cost of the parent card: `if(variant.manaCost == null) variant.manaCost = this.manaCost;`. 

I think this change would only matter if someone creates a CardFace without calling `assignMissingFields`*, which would present similar issues when the triggers and abilities lists were also all left null.

*Functional variants previously did this, but this PR cleans up that behavior too.

Hanmac: my question was more if we should have kept the default "No_Cost/No_Color" thing, and we could remove the ManaCost:no_cost from the lands

@tool4ever your opinion?

Jetz72:
> I don't know if it would be better, but what about using `BitSet` for the Attraction Lights?

Could rework it if it would read or run better. If anything though I'd probably just make it into its own custom class with a bitmask int and a handful of helper methods.

forge-core/src/main/java/forge/card/CardFace.java
tool4ever: cleaning up the lands like that would make sense too

Hanmac: I think we can close that issue for now (making ManaCost Required)

and then we can make a new MR to make ManaCost optional again?

Hanmac (approved):

PR #5666, opened 2024-07-21
Text change support for Class + Saga

### LINKED ISSUE #5649: Keywords with Referencing SVars should be affected by text changes
Keywords like:
* K:Chapters for Sagas
* K:Class for Class

(and maybe others) do reference SVars and these may be affected by Card Text Changes.

For example, this Class thing:
```
K:Class:2:G U:AddReplacementEffects$ SharkCatch
SVar:SharkCatch:Event$ CreateToken | ActiveZones$ Battlefield | ValidPlayer$ You | ValidToken$ Fish | ReplaceWith$ TokenReplace1 | Description$ If you would create a Fish token, create a 3/3 blue Shark creature token instead.
SVar:TokenReplace1:DB$ ReplaceToken | Type$ ReplaceToken | ValidCard$ Fish | TokenScript$ u_3_3_shark
```

First: the SVar proablaly should not reference a CreatureType name, so when Shark gets changed to Human,  `K:Class:2:G U:AddReplacementEffects$ SharkCatch` should not be changed `K:Class:2:G U:AddReplacementEffects$ HumanCatch` because it could break the Reference to the SVar.

But inside the Replacement Effect, The Token needs to be updated into an `u_3_3_human` one (without breaking the Tokenscript name)

Otherwise, we need to think of a way to refactor them without Keywords if possible

tool4ever:
image

Closes #5649

### DIFF
--- a/forge-game/src/main/java/forge/game/TriggerReplacementBase.java
+++ b/forge-game/src/main/java/forge/game/TriggerReplacementBase.java
@@ -1,6 +1,7 @@
 package forge.game;
 
 import java.util.EnumSet;
+import java.util.Map;
 import java.util.Set;
 
 import forge.game.card.Card;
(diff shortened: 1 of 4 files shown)


Hanmac (approved):
Looks good I think?

Of not, we might update it later more

PR #5673, opened 2024-07-22
Improve anchor word handling

### LINKED ISSUE #5672: Text change doesn't work on cards with replacing contrary choice ETB
The "Siege" cards Animate themselves for their abilities.

So far it only seems to matter for creature type in _Mirrodin Besieged_ ?

tool4ever:
While not truly linked should still be in a better state like this.

Closes #5672


forge-game/src/main/java/forge/game/card/CardProperty.java
```diff
@@ -138,6 +138,10 @@ public static boolean cardHasProperty(Card card, String property, Player sourceC
             if (source.hasChosenCard(card)) {
                 return false;
             }
+        } else if (property.startsWith("ChosenMode")) {
+            if (!card.getChosenMode().equals(property.substring(10))) {
```
Hanmac, line 142: Probably need an extra check because ChosenModes are linked?

Or are we doing that way later with my LinkedAbility MR?

tool4ever: yea, just transforming these scripts with statics will help later

Hanmac (approved):
Good enough for now, I think


one thing we need to take care of, the Anchor Word "Dragon" should not be affected by CardTextChanges

PR #143, opened 2022-04-26
LinkedAbilities V2

Hanmac:
This is my current Draft for Linked Abilities

Current Problem:
* Cards that refer to Original Host because they cause problems which card is meant by this (examples follow later)

PR #5676, opened 2024-07-22
Update cardstates after creating token from prototype

### LINKED ISSUE #5677: Bug with Eriette, the Beguiler
**Describe the bug**

**Eriette, the Beguiler:** *Whenever an Aura you control becomes attached to a nonland permanent an opponent controls with mana value less than or equal to that Aura's mana value, gain control of that permanent for as long as that Aura is attached to it.*

When the aura attached to the creature that Eriette took control of is removed the control of that creature doesn't revert to its owner.

**Expected behavior**
For control of the creature to return to its owner.

 - Forge Version: forge-gui-desktop-1.6.64-SNAPSHOT-07-22

tool4ever:
Reverts the previous cleanup.

@Hanmac
The problem is when the final token gets created from its prototype the traits that get copied over will have that original CardState instead.

Since this seems to be a bigger flaw in the engine related to #143 I probably can't do anything else here 🤔

Closes #5677

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardState.java
+++ b/forge-game/src/main/java/forge/game/card/CardState.java
@@ -748,6 +748,21 @@ public class CardState extends GameObject implements IHasSVars {
         }
     }
 
+    public void resetOriginalHost() {
+        final List<CardTraitBase> allAbs = ImmutableList.<CardTraitBase>builder()
+            .addAll(manaAbilities)
+            .addAll(nonManaAbilities)
+            .addAll(triggers)
+            .addAll(replacementEffects)
+            .addAll(staticAbilities)
+            .build();
+        for (final CardTraitBase ctb : allAbs) {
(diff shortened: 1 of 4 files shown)


Hanmac:
i think it should be easier to fix the Token Creation about that "OriginalHost" for such case means the Created Token, and not the Token Prototype

forge-game/src/main/java/forge/game/card/CardState.java
```diff
@@ -748,6 +748,21 @@ public void addIntrinsicKeywords(Collection<KeywordInterface> intrinsicKeywords2
         }
     }
 
+    public void resetOriginalHost() {
+        final List<CardTraitBase> allAbs = ImmutableList.<CardTraitBase>builder()
```
Hanmac, line 752: When such immutable list is used for multiple functions, we better separate it into its own function

Hanmac (approved):

PR #5691, opened 2024-07-26
Initial Gift development

### LINKED ISSUE #5463: BLB: Gift Keyword
Two Version, one for Permanent Spells and one for instant or sorcery spell
> As you cast a permanent spell with gift, you can promise the gift to an opponent if you want. This is optional, and if you have more than one opponent—ooh, scary!—you choose one and only one of them at most. They don't receive the gift yet. You're just promising it. When the permanent enters, the chosen opponent receives the gift. In Kitnap's case, they draw a card.
> Each card with gift will have additional or alternative effects if the gift was or wasn't promised. In Kitnap's case, promising the gift will change how its other enters-the-battlefield triggered ability functions. If you agree to let an opponent draw a card, you can avoid putting three stun counters on the creature you just … um, adopted.
> But not every gift is a card, and not every card with gift is a permanent. Let's look at an instant and a Fish.
> Promising the gift for an instant or sorcery spell works just like it does for a permanent spell. You do this as you cast the spell. It's an additional cost. The opponent receiving the gift is a little different. As the instant or sorcery spell starts resolving, first the opponent receives their gift. In Parting Gust's case, they create a tapped 1/1 blue Fish creature token. Yay! A free snack and/or pet and/or good luck charm. Then the spell's other effects happen.

The "Gift a tapped fish" Token uses a Named one, so LockTokenScript and no ChangeCardText

Other Gift Effects:

- Gift a Treasure
- Gift a card
- Gift an extra Turn

Also, "Whenever you gift" Trigger

tehdiplomat:
Closes #5463


forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -1255,6 +1225,22 @@ public static void addTriggerAbility(final KeywordInterface inst, final Card car
 
             saRebel.setIntrinsic(intrinsic);
             inst.addTrigger(etbTrigger);
+        } else if (keyword.equals("Gift")) {
```
Hanmac, line 1228: The ETB trigger only makes sense for Permanents, so i think we could skip them for others?

Agetian (approved):

Agetian:
Probably worth updating GameState to support |Promised for the save state and puzzle purposes, but I can take care of it later once this PR is in.

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -1352,6 +1354,28 @@ public static void resolve(final SpellAbility sa) {
...
+        Player controller = sa.getActivatingPlayer();
+        Card source = sa.getHostCard();
+        // do blessing there before condition checks
+        if (sa.isSpell() && !source.isPermanent() && source.hasKeyword(Keyword.ASCEND)) {
+            if (controller.getZone(ZoneType.Battlefield).size() >= 10) {
+                controller.setBlessing(true);
+            }
+        }
+
+        if (sa.isSpell() && !(sa instanceof SpellPermanent) && sa.getHostCard().hasKeyword(Keyword.GIFT) && sa.isOptionalCostPaid(OptionalCost.PromiseGift)) {
```
Hanmac, line 1367: I don't know if `!source.isPermanent()` wouldn't be better like above with Ascend

Hanmac:
I probably would have written it like `K:Gift:GiftEffect`, but that might be changed later

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -1888,6 +1889,18 @@ public final void setChosenPlayer(final Player p) {
         view.updateChosenPlayer(this);
     }
 
+    public final boolean hasPromisedGift() {
+        return promisedGift != null;
+    }
+    public final Player getPromisedGift() {
+        return promisedGift;
+    }
+    public final void setPromisedGift(final Player p) {
```
tool4ever, line 1898: this needs to be called in `copySpellHost`

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -1255,6 +1225,22 @@ public static void addTriggerAbility(final KeywordInterface inst, final Card car
...
+            // Gift is a special keyword that is used to create a trigger for permanents when they enter the battlefield
+            // On casting a Gift needs to be promised to an opponent
+            final SpellAbility saGift = AbilityFactory.getAbility(card.getSVar("GiftEffect"), card);
```
Hanmac, line 1231: What happens when you do this for non-Permanents?

`card.getFirstSpellAbility().setAdditionalAbility("GiftAbility", saGift)` ?

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
tehdiplomat: Is source.isPermanent right? Does that calculate properly for things like Adventure?

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
tehdiplomat: Yea i know. I wasn't done with this part yet.

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
Hanmac: At this place, shortly before the Spell is going to resolve, 
the Spell Host has changed into the Adventure Spell Half (so it is not a Permanent, and it shouldn't have had Gift anyway)

forge-game/src/main/java/forge/game/card/Card.java
tehdiplomat: Lemme know if thats what you meant

Hanmac (approved):

Hanmac (approved):

tool4ever (approved):

PR #5693, opened 2024-07-26
Pawprint: first try

### LINKED ISSUE #5555: Pawprint Modal Spells
> Choose up to five {J} worth of modes. You may choose the same mode more than once.
> {J} — Each player sacrifices a creature.
> {J}{J} — Draw a card for each creature you controlled that died this turn.
> {J}{J}{J} — Each opponent loses X life, where X is the number of creature cards in your graveyard.

Modes have "Paw Cost", you get N Symbols. so CharmMax should be N too.
Then you can select Modes up to the amount of Symbols for this Spell.

AI might need some extra logic for this

Hanmac:
Closes #5555 

- [x] Human Player
- [x] AI Player
- [x] New Icon, the old Phyrexian needs to be renamed to H first #5698

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/CharmAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/CharmAi.java
@@ -26,6 +26,7 @@ public class CharmAi extends SpellAbilityAi {
     protected boolean checkApiLogic(Player ai, SpellAbility sa) {
         final Card source = sa.getHostCard();
         List<AbilitySub> choices = CharmEffect.makePossibleOptions(sa);
+        final int pawprintLimit = sa.hasParam("Pawprint") ? AbilityUtils.calculateAmount(source, sa.getParam("Pawprint"), sa) : 0;
 
         final int num;
         final int min;
@@ -66,13 +67,13 @@ public class CharmAi extends SpellAbilityAi {
              * minimum choice requirements with canPlayAi() alone.
              */
             chosenList = min > 1 ? chooseMultipleOptionsAi(choices, ai, min)
-                    : chooseOptionsAi(choices, ai, timingRight, num, min, sa.hasParam("CanRepeatModes"));
+                    : chooseOptionsAi(choices, ai, timingRight, num, min, pawprintLimit, sa.hasParam("CanRepeatModes"));
         }
 
         if (chosenList.isEmpty()) {
             if (timingRight) {
                 // Set minimum choices for triggers where chooseMultipleOptionsAi() returns null
-                chosenList = chooseOptionsAi(choices, ai, true, num, min, sa.hasParam("CanRepeatModes"));
+                chosenList = chooseOptionsAi(choices, ai, true, num, min, pawprintLimit, sa.hasParam("CanRepeatModes"));
                 if (chosenList.isEmpty() && min != 0) {
                     return false;
                 }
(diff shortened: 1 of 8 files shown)


kevlahnota:
@Hanmac  OK, I'm replacing all instances of old phyrexian code **P** to designate the **H** code, gonna make a PR later

Hanmac:
@Agetian can you help me looking into the AI code? it should be in CharmAI, but it seems AI is also ignoring "allowRepeat" :/

Agetian:
> @Agetian can you help me looking into the AI code? it should be in CharmAI, but it seems AI is also ignoring "allowRepeat" :/

Yeah, I checked out the code and it looks like allowRepeat isn't really used in the code (not sure if it was always that way or it got lost over the years), it looks like a simple solution would be to only remove the chosen choice if allowRepeat is set to false, e.g.:

```
            if (!allowRepeat) {
                choices.removeAll(chosenList);
            }
```

Circa lines 111 and 122 or CharmAi. What do you think? Would that do the trick for the card you're implementing? Can you try it in action, please?

tool4ever:
The problem with allowing AI to repeat modes freely is that it wouldn't take into account any previous choices.

Imagine if your opponent had a huge threat - then without additional logic it'd end up going for "Each player sacrifices a creature" 5 times, which in worst case could lead to only itself sacrificing on the remaining 4 rounds haha :/

Agetian:
Ah, yes, that's true. Hmm, not sure what the optimal logic here would be, as this can get pretty complicated pretty quickly :) I need to think about a good enough starting implementation that would be safe for the AI to use and yet allow it to sanely repeat modes :)

Hanmac:
> The problem with allowing AI to repeat modes freely is that it wouldn't take into account any previous choices.
> 
> Imagine if your opponent had a huge threat - then without additional logic it'd end up going for "Each player sacrifices a creature" 5 times, which in worst case could lead to only itself sacrificing on the remaining 4 rounds haha :/

yeah that need to be optimized later too :(

@Agetian one thing we need to try first is for AI to not use more Pawprint Points as allowed

Agetian:
> @Agetian one thing we need to try first is for AI to not use more Pawprint Points as allowed

That's easier, the basic version can probably just count Pawprints as the AI makes choices (per the defined parameter) and short circuit (stop choosing) if it maxes out.

Agetian:
Added some basic AI logic to limit according to the number of Pawprints, the exact choices will depend on the subability logic. This raises a question, however. In particular, should the if branches below that code (which guarantee the min amount of choice) also be limited by Pawprint? This could, in theory, lead to an infinite loop (if there's ever a situation, for whatever reason, where the AI is forced to choose a certain min number of options but is limited at the same time by the Pawprint limit). The currently implemented card doesn't allow for situations like that, so I dunno :/

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
```diff
@@ -93,13 +94,22 @@ protected boolean checkApiLogic(Player ai, SpellAbility sa) {
     }
 
     private List<AbilitySub> chooseOptionsAi(List<AbilitySub> choices, final Player ai, boolean isTrigger, int num,
-            int min, boolean allowRepeat) {
+            int min, int pawprintLimit, boolean allowRepeat) {
```
Hanmac, line 97: i would just pass the SpellAbility sa to this function, so you don't need to have extra pawprintLimit

like i accidently commited there:
https://github.com/Card-Forge/forge/pull/5713/files#diff-fd1617932269fd1af8b5b0c76a9d2597fd611b6c478b39f62bc7baa035d614a6

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
```diff
@@ -93,13 +94,22 @@ protected boolean checkApiLogic(Player ai, SpellAbility sa) {
...
             sub.setActivatingPlayer(ai, true);
             if (AiPlayDecision.WillPlay == aic.canPlaySa(sub)) {
+                if (pawprintLimit > 0) {
+                    int curPawprintAmount = AbilityUtils.calculateAmount(sub.getHostCard(), sub.getParamOrDefault("Pawprint", "0"), sub);
+                    if (pawprintAmount + curPawprintAmount > pawprintLimit) {
+                        continue;
+                    } else {
+                        pawprintAmount += curPawprintAmount;
+                    }
+                }
```
Hanmac, line 112: @Agetian because the later Choices are more expensive for Pawprint (and Spree?) I think it might be a good idea to flip the choices list, so AI tries to pay for the most expensive choices first

Agetian: Agreed :) Did it for Pawprint at the moment, should we always do it instead, in case it matters for more stuff? (can you also tell me what Spree card you're referring to, please?)

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
Agetian: Ah, indeed, done :) :+1:

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
tool4ever: looks like some extra test changes sneaked in though

Hanmac: fixed the sneaky changes

Hanmac:
@Agetian Spree Cards like this:
https://scryfall.com/search?q=keyword%3Aspree&unique=cards&as=grid&order=name

But i was wrong, it isn't always the case that the most expensive SubAbility is last ... so we should probably address this in a later MR

PR #5697, opened 2024-07-27
update Phyrexian designation code

kevlahnota:
- #5693 to support Bloomburrow Pawprint {P} designation


tehdiplomat:
Looks like you have a failing test. It seems like you updated that file, so maybe a toString needs to updated that wasn;t?

kevlahnota:
> Looks like you have a failing test. It seems like you updated that file, so maybe a toString needs to updated that wasn;t?

fixed

Hanmac:
@kevlahnota commando back

All the other still uses WP or something

It's only about exactly "{P}" used for Rage Extractor

kevlahnota:
> @kevlahnota commando back
> 
> All the other still uses WP or something
> 
> It's only about exactly "{P}" used for Rage Extractor

I can rollback and only replace the single P to use H, what's your final take on this @tehdiplomat

kevlahnota:
need to wait for final ruling update...

Hanmac:
> need to wait for final ruling update...

No need to close it, my sources say it's only one icon
(And even if not, we can do the other ones later)

Agetian:
Will reopen for now, it's definitely workable and can be updated promptly once we have information.

kevlahnota:
> Will reopen for now, it's definitely workable and can be updated promptly once we have information.

gonna close this, since pawprint are not really related to manashards, its only GUI display, gonna close this and push a new PR

PR #5702, opened 2024-07-27
Fixes for Mistbreath Elder, Long River Lurker, Waterspout Warden

Glorax:
Most of Mistbreath Elder's effects didn't work and it was a 4|4 for G instead of a 2|2.

Waterspout Warden had Haste for some reason.

Long River Lurker's effect straight-up crashed Forge due to a naming mismatch.

### DIFF
--- a/forge-gui/res/cardsfolder/c/cutthroat_centurion.txt
+++ b/forge-gui/res/cardsfolder/c/cutthroat_centurion.txt
@@ -4,5 +4,5 @@ Types:Artifact Creature Phyrexian Warrior
 PT:2/2
 A:AB$ Pump | Cost$ Sac<1/Creature.Other;Artifact.Other/another creature or artifact> | Defined$ Self | NumDef$ 2 | NumAtt$ +2 | AILogic$ Aristocrat | ActivationLimit$ 1 | SpellDescription$ CARDNAME gets +2/+2 until end of turn. Activate only once each turn.
 DeckHas:Ability$Sacrifice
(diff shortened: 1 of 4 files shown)


forge-gui/res/cardsfolder/upcoming/mistbreath_elder.txt
Glorax, line 1: In my attempts to fix Mistbreath Elder, I ended up majorly revising the structure of its effects, getting rid of the DBBranch. I don't know exactly why it didn't work, but either the Branch effect was formatted incorrectly, or Branch is straight-up broken.

forge-gui/res/cardsfolder/c/cutthroat_centurion.txt
Glorax, line 1: Does this AIPreference section actually serve any purpose anymore? I notice there are a many cards, like Deadly Dispute, that lack it entirely.

forge-gui/res/cardsfolder/upcoming/mistbreath_elder.txt
tool4ever: Weird, I can only assume you tested with an older version since the wrong PT was already fixed
but your cleanup is fine too

Glorax: I think it was the version from the 25th; you must have caught the wrong PT very quickly.

forge-gui/res/cardsfolder/c/cutthroat_centurion.txt
tool4ever: the AI can't really know how valuable the effect it's paying for would be in relation to its sacrifice
so this helps with establishing what's still acceptable in the average situation

tool4ever (approved):

PR #5703, opened 2024-07-27
Miscallaneous Cleanup

Jetz72:
A collection of IDE-suggested changes to utilize language and utility features, available since Java 8 or earlier. Biggest chunk is the cleanup of manual boxing and unboxing of reference types, as well as removing reference types where they were no longer needed.

No changes here should affect code semantics, and performance should be roughly equal or better. Changes were organized into commits based on type of change; it'll probably be easier to review them one commit at a time.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -1173,8 +1173,8 @@ public class AiAttackController {
         while (!attritionalAttackers.isEmpty() && humanLife > 0 && attackRounds < 99) {
             // sum attacker damage
             int damageThisRound = 0;
-            for (int y = 0; y < attritionalAttackers.size(); y++) {
-                damageThisRound += attritionalAttackers.get(y).getNetCombatDamage();
+            for (Card attritionalAttacker : attritionalAttackers) {
+                damageThisRound += attritionalAttacker.getNetCombatDamage();
             }
             // remove from player life
             humanLife -= damageThisRound;
--- a/forge-gui-mobile/src/forge/itemmanager/views/ItemListView.java
+++ b/forge-gui-mobile/src/forge/itemmanager/views/ItemListView.java
@@ -22,6 +22,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.stream.IntStream;
 
 import com.badlogic.gdx.math.Rectangle;
 
@@ -154,9 +155,7 @@ public final class ItemListView<T extends InventoryItem> extends ItemView<T> {
     @Override
     public void selectAll() {
         selectedIndices.clear();
-        for (Integer i = 0; i < getCount(); i++) {
-            selectedIndices.add(i);
-        }
+        IntStream.range(0, getCount()).forEach(i -> selectedIndices.add(i));
         onSelectionChange();
     }
 
--- a/forge-gui-mobile/src/forge/itemmanager/views/ItemListView.java
+++ b/forge-gui-mobile/src/forge/itemmanager/views/ItemListView.java
@@ -61,7 +61,7 @@ public final class ItemListView<T extends InventoryItem> extends ItemView<T> {
 
     private final ItemList list = new ItemList();
     private final ItemListModel listModel;
-    private List<Integer> selectedIndices = new ArrayList<>();
+    private final List<Integer> selectedIndices = new ArrayList<>();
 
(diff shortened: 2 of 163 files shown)


forge-gui-mobile/src/forge/itemmanager/views/ItemListView.java
```diff
@@ -154,9 +155,7 @@ public void zoomSelected() {
     @Override
     public void selectAll() {
         selectedIndices.clear();
-        for (Integer i = 0; i < getCount(); i++) {
-            selectedIndices.add(i);
-        }
+        IntStream.range(0, getCount()).forEach(i -> selectedIndices.add(i));
```
Hanmac, line 158: doesn't `selectedIndices::add` would work too like in the file before?

Jetz72: Took me a moment to figure out why the IDE did one one way and one the other. `selectedIndicies` is declared as final in ImageView but not here in ItemListView. The IDE was apparently accounting for the possibility that the ArrayList `selectedIndicies` points to could change in the middle of a loop. The lambda would account for that, whereas the method reference would resolve the reference once, then call `add` on the same object until the loop's conclusion even if the field changed.

Regardless, there isn't actually any code that would ever replace `selectedIndicies`, let alone in the middle of the select all loop, so I've changed it to final and updated it here.

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -1173,8 +1173,8 @@ public int compare(Pair<GameEntity, Integer> r1, Pair<GameEntity, Integer> r2) {
         while (!attritionalAttackers.isEmpty() && humanLife > 0 && attackRounds < 99) {
             // sum attacker damage
             int damageThisRound = 0;
-            for (int y = 0; y < attritionalAttackers.size(); y++) {
-                damageThisRound += attritionalAttackers.get(y).getNetCombatDamage();
+            for (Card attritionalAttacker : attritionalAttackers) {
+                damageThisRound += attritionalAttacker.getNetCombatDamage();
             }
```
Hanmac, line 1178: @tehdiplomat @Agetian is it worth it to change this into a sum function with Stream lib?

tool4ever: does this work?
`damageThisRound = Aggregates.sum(attritionalAttackers, Card::getNetCombatDamage)`

Hanmac: Probably, Aggregates is one of ours from before Java8 times (I think we could update that later?)

but I was more thinking about one of these:
https://www.baeldung.com/java-stream-sum

I think `Collectors` might be the most useful for us?

I tried this example
```java
/* Online Java Compiler and Editor */

import java.util.*;

public class HelloWorld{

static public class Damage
{
    protected int damage;
    public Damage(int i) {damage = i;}
    public int getDamage() { return damage;}
}
     public static void main(String []args){
         List<Damage> list = new ArrayList<>();
         list.add(new Damage(1));
         list.add(new Damage(2));
         list.add(new Damage(3));
        System.out.format("Damage: %d", list.stream().collect(
            Collectors.summingInt(Damage::getDamage)));
     }
}
```

Jetz72: I was thinking the same thing too when I saw that one.

The IDE does have an inspection that can convert for loops into streams. It's off by default, though, and I can kinda see why - it's kinda subjective where you draw the line between the stream form being shorter and simpler vs dense and difficult.

This block comes out looking okay:
image

This one though...
image
More concise, but arguably trickier to read, and very likely more awkward to edit if you wanted to do something more complicated within that for loop structure.

And this one goes from "hard to process" to "complete nightmare":
image

"for loop -> stream" was something I was considering for a second wave of this cleanup process, but evidently there'd need to be a rule of thumb for which ones to apply and which ones to leave alone. Number of terms in the stream? Length of the resulting line? Only do it for single statement loops? Or could just leave that one alone...

Hanmac: yeah i think that is for later MR, right now, `Aggregates.sum` might be good enough

but i think from the examples of that java sum page, `Collectors.summingInt` is imo my favorite.

tehdiplomat: Yea lets do that as a different PR. These changes look like a bunch of good cleanups for now.

tehdiplomat (approved):

Issue #5675, opened 2024-07-22
Heirloom Epic


Hanmac:
> {4}, {T}: Draw a card. For each mana in this ability’s activation cost, you may tap an untapped creature you control rather than pay that mana.

@tool4ever wouldn't this work with Announce with an TapXType + ReduceCost/ReduceAmount?

While thinking more about it, it probably needs to be inside CostAdjustment with an optional SwitchCost Logic

tool4ever:
The problem is the cost needs to be adjusted before:
> You can't tap more creatures than the amount of mana in the cost of Heirloom Epic's ability. Effects that increase or reduce the activation cost of Heirloom Epic's ability apply before costs are paid and will increase or decrease the total number of creatures you can tap.

Hanmac:
> The problem is the cost needs to be adjusted before:
> 
> > You can't tap more creatures than the amount of mana in the cost of Heirloom Epic's ability. Effects that increase or reduce the activation cost of Heirloom Epic's ability apply before costs are paid and will increase or decrease the total number of creatures you can tap.

Isn't that what "AnnounceMax" is for?

Hanmac:
While thinking more about it, it probably needs to be inside CostAdjustment with an optional SwitchCost Logic

Hanmac:
@tool4ever for this to work, the RaiseCost needs to be After all the Other Raise Cost, so it can use the maximum of ManaCost can be used:

> 601.2f The player determines the total cost of the spell. Usually this is just the mana cost. Some spells have additional or alternative costs. Some effects may increase or reduce the cost to pay, or may provide other alternative costs. Costs may include paying mana, tapping permanents, sacrificing permanents, discarding cards, and so on. The total cost is the mana cost or alternative cost (as determined in rule 601.2b), plus all additional costs and cost increases, and minus all cost reductions. **If multiple cost reductions apply, the player may apply them in any order.** If the mana component of the total cost is reduced to nothing by cost reduction effects, it is considered to be {0}. It can’t be reduced to less than {0}. Once the total cost is determined, any effects that directly affect the total cost are applied. Then the resulting total cost becomes “locked in.” If effects would change the total cost after this time, they have no effect.

the current rule only says about that you can choose the order they apply for cost reductions

> 118.7. What a player actually needs to do to pay a cost may be changed or reduced by effects.

I think effects like "Heirloom Epic" are more like Changing Cost than Raising/Reducing it 🤔

tool4ever:
[...]

PR #5704, opened 2024-07-27
Rottenmouth Viper

Glorax:


forge-gui/res/cardsfolder/upcoming/rottenmouth_viper.txt
```diff
@@ -0,0 +1,19 @@
+Name:Rottenmouth Viper
+ManaCost:5 B
+Types:Creature Elemental Snake
+PT:6/6
+A:SP$ PermanentNonland | Cost$ 5 B Sac<X/Permanent.nonLand/nonland permanent> | AILogic$ SacToReduceCost
```
Hanmac, line 5: @tool4ever can we try to make this `Sac<X/Permanent.nonLand/nonland permanent>` as Additional Cost?

tool4ever: hmn I guess it could be scripted more like the March cycle?
though this would still require the extra spell to use Announce$ instead
but in either case the API should be `PermanentCreature`

Hanmac: I need to check if we could add an Announce on the RaiseCost/ReduceCost so we don't need that extra Spell


Also, this might be more interesting, such Combined Reduce Cost (add other cost) seems to apply AFTER the normal RaiseCost effects (so you could reduce them too)

tool4ever: > Also, this might be more interesting, such Combined Reduce Cost (add other cost) seems to apply AFTER the normal RaiseCost effects (so you could reduce them too)

What you mean? This should already work?

Hanmac: If the ReplaceCost  depends on the Maximum of the Raise Mana Cost like #5675 , it would matter in which order you apply them

Glorax: I used Torgaar as a basis.

tool4ever: Why is PermanentCreature inappropriate for a creature spell? :p

Hanmac: @Glorax can the Type be changed into PermanentCreature?

Glorax: Yes.

Hanmac:
@tool4ever if that is ready, I would squash it?

tool4ever (approved):

PR #5713, opened 2024-07-28
SacrificeOnce Trigger

### LINKED ISSUE #5674: Camellia, the Seedmiser
Needs "SacrificedOnce" trigger

Because the trigger cares about "You" sacrificing stuff, it should collect the LKI and trigger per Player.

Most sacrifice stuff already applies to a List, or is already in a for each player loop

Hanmac:
Some things i need to consider:

- [x] I really don't like the Destroy Effect that does Sacrifice instead, we should refactor them #5719
- ~~[ ] For Cost, i don't know, but i feel we need to handle them different for stuff that does "Sacrifice itself AND other", for this special case, maybe it shouldn't trigger there :(~~

Closes #5674

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -1442,10 +1442,7 @@ public class GameAction {
                 sacrificeList = (CardCollection) GameActionUtil.orderCardsByTheirOwners(game, sacrificeList, ZoneType.Graveyard, null);
                 orderedSacrificeList = true;
             }
-            for (Card c : sacrificeList) {
-                c.updateWasDestroyed(true);
-                sacrifice(c, null, true, mapParams);
-            }
+            sacrifice(sacrificeList, null, true, mapParams);
             setHoldCheckingStaticAbilities(false);
 
             table.triggerChangesZoneAll(game, null);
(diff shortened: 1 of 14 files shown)


tool4ever:
you need to add it to `registerActiveLTBTrigger`

Hanmac:
> you need to add it to `registerActiveLTBTrigger`

@tool4ever feel free to add it

forge-game/src/main/java/forge/game/ability/effects/DestroyEffect.java
```diff
@@ -71,12 +73,37 @@ public void resolve(SpellAbility sa) {
...
+
+        final boolean remDestroyed = sa.hasParam("RememberDestroyed");
+
+        SpellAbility cause = sa;
+        if (sa.isReplacementAbility()) {
+            cause = (SpellAbility) sa.getReplacingObject(AbilityKey.Cause);
+        }
+
+        // TODO remove Sacrifice inside Destroy
+        for (Card c : game.getAction().sacrifice(sacList, cause, true, params)) {
```
Hanmac, line 101: @tool4ever if we would remove the Destroy/Sacrifice thing, I could drop this extra piece of code

Hanmac: Waiting for #5744 to be merged first

Hanmac:
@tool4ever i think this MR is ready?

tool4ever (approved):

PR #5715, opened 2024-07-28
Update anduril_flame_of_the_west.txt

Fulgur14:
Equipments that trigger on the equipped creature attacking should check the same creature on resolution, but they often check the equipped creature (which may have incorrect results if the equipment is moved to another creature with the ability on stack). If this looks alright, I can change them to work with TriggeredAttacker instead.


forge-gui/res/cardsfolder/a/anduril_flame_of_the_west.txt
```diff
@@ -6,7 +6,7 @@ T:Mode$ Attacks | ValidCard$ Card.AttachedBy | Execute$ TrigBranch | TriggerDesc
 SVar:TrigBranch:DB$ Branch | BranchConditionSVar$ X | BranchConditionSVarCompare$ EQ1 | TrueSubAbility$ Legendary | FalseSubAbility$ NonLegendary
 SVar:Legendary:DB$ Token | TokenAmount$ 2 | TokenScript$ w_1_1_spirit_flying | TokenOwner$ You | TokenTapped$ True | TokenAttacking$ True
 SVar:NonLegendary:DB$ Token | TokenAmount$ 2 | TokenScript$ w_1_1_spirit_flying | TokenOwner$ You | TokenTapped$ True
-SVar:X:Count$Valid Creature.EquippedBy+Legendary
+SVar:X:Count$Valid Creature.TriggedAttacker+Legendary
```
tool4ever, line 9: yea good call
but actually you should use `TriggeredAttacker$Valid Card.Legendary`
because even when the creature gets removed in response LKI should take over instead

tool4ever (approved):

PR #5716, opened 2024-07-29
TriggeredAttacker corrections

Fulgur14:


Hanmac:
@tool4ever shouldn't this be `TriggeredAttackerLKICopy`?

tool4ever:
yea

Hanmac:
@tool4ever about these `DB$ Destroy | Defined$ Self | Sacrifice$ True` cards like Ordeal, can we try to rewrite them using Sacrifice Effect instead of Destroy Effect? 

imo it feels wrong if the Destroy Effect doesn't Destroy (See also #5713 )

tool4ever:
ugh, I'd have to check but I'm not sure if that requires switching some engine parts around too?

Hanmac:
> ugh, I'd have to check but I'm not sure if that requires switching some engine parts around too?

i don't think that should be needed, `Sacrifice | Defined$ Self` should already work?

tool4ever (approved):

Issue #210, opened 2022-05-02
etbCounter needs to be Removed in the Future


Hanmac:
etbCounter currently makes problems with CardTextChanges because the value for X is not affected by it but it should.
Example: `Ulasht, the Hate Seed`

In the work of changing ETB Keywords into a Form that used "ReplacementResult.Updated" instead of using ETBInternal,
it will be much cleaner to use normal replacement Effects instead of using a Keyword

This is part of the Cleanup of CardFactoryUtil to remove everything what isn't a "real Keyword"

Northmoc:
Just some cards that I think might need special attention when doing this refactor:
 - Agent's Toolkit
 - the IKO "this card enters with your choice of counter A or counter B" stuff

Hanmac:
yeah, extracting from a keyword into an replacement effect isn't that big of a problem.

it's just to not use the keyword anymore

Hanmac:
@tool4ever if we do not remove this keyword, then maybe you can do your thing with #5666 with these Keywords too so it would fix `Ulasht, the Hate Seed`?

tool4ever:
I don't know, there's no benefit in keeping it? (except saving some editing work?)

Hanmac:
> I don't know, there's no benefit in keeping it? (except saving some editing work?)

I don't think so ... 

Having fake keywords (like this) or hidden keywords only causes Problems Rules-wise ... 

So i'm for replacing them with Replacement Effects when we need to touch them anyway

Hanmac:
@dracontes should we do another round with etbCounter removal? Or do we wait with this after a Bloomborrow release?

dracontes:
> should we do another round with etbCounter removal? Or do we wait with this after a Bloomborrow release?

My apologies for the late reply. I'm fine with the `etbCounter` removal pass being after a Bloomburrow release as I'm currently organizing my notes and going over minor items in them.

PR #5720, opened 2024-07-29
ETB update: Java code, adventure custom cards, token scripts

dracontes:
Changing "enters the battlefield" to "enters" (and the respective variants of the first to the variants of the second) as detailed on Gatherer where it's functionally relevant. Items that were ignored:
- Code comments
- Language/translation files
- Puzzle and deck descriptions

I hope my tinkering with the code is at least useful to start off. I do feel as some strings for look-up get rather short, maybe some internal keywording would be more useful in the long term.

**Edit 30-07-2024:** As requested by Hanmac, replaced keyworded ETB tapped replacement effects with the code that would be generated by the game engine. Split the difference on `As it enters` collapsing onto the aforementioned replacement effects mostly for clarity.

**Edit 04-08-2024:** Setting to draft and splitting this PR off into more focused ones.


tool4ever:
Unfortunately we haven't gotten around to refactoring that tapped RE without keyword
this means all of these PR will have to be merged really close together, otherwise only updated scripts would work :/

Hanmac:
one thing to notice, for cards like "Uncharted Haven" with:
> Uncharted Haven enters tapped. As it enters, choose a color.

these still needs to be two replacement effects

Hanmac:
> Unfortunately we haven't gotten around to refactoring that tapped RE without keyword this means all of these PR will have to be merged really close together, otherwise only updated scripts would work :/

yeah, these "fake" ETB keywords are on my hit list to remove if able

especially #210

Jetz72:
Could the RE temporarily be reworked to use either wording? You might even leave it that way for a while, since it'd also be a breaking change for custom sets.

Hanmac:
@dracontes @tool4ever @tehdiplomat your opinion of using the change and replace `K:CARDNAME enters tapped.` with a Replacement Effect?

dracontes:
> these still needs to be two replacement effects

In regard to the `SpellDescription$` for the choose-a-color replacement effect: do I follow the letter of the cards ("`As it enters`") or do I keep it "`As CARDNAME enters`"? Unfortunately, I didn't get around to addressing this conundrum. Fortunately there are no relevant cards in the 'a' folder so I'm well in time to do so.

dracontes:
> @dracontes your opinion of using the change and replace `K:CARDNAME enters tapped.` with a Replacement Effect?

I'd certainly go for something like `K:ETBReplacement:Tapped`. Looking over scripts it seems like we have a good starting point for paring down in, say, Razorverge Thicket's version (<https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/r/razorverge_thicket.txt>).

Hanmac:
> > @dracontes your opinion of using the change and replace `K:CARDNAME enters tapped.` with a Replacement Effect?
> 
> I'd certainly go for something like `K:ETBReplacement:Tapped`. Looking over scripts it seems like we have a good starting point for paring down in, say, Razorverge Thicket's version (https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/r/razorverge_thicket.txt).

Something like ETBReplacement is just a short fix, this one is on my hit-list too


So I think `K:CARDNAME enters tapped.` should be replaced with this:
```
Event$ Moved | ValidCard$ Card.Self | Destination$ Battlefield | ReplacementResult$ Updated | ReplaceWith$ ETBTapped | Description$ CARDNAME enters tapped.
```

@tool4ever your opinion?

tool4ever:
Ideally and eventually yes
There are probably a few that can then be swapped to NICKNAME too

dracontes:
> In regard to the `SpellDescription$` for the choose-a-color replacement effect

Had a look and it turns out cards and card faces like [Apex Observatory](https://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=642482), [Sunken Citadel](https://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=637010), [Mirage Mesa](https://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=655203) and [Tarnation Vista](https://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=660326), already collapsed an ETB choice replacement effect into the same paragraph as the ETB tapped one. Among these only Mirage Mesa keeps the replacement effects separate, apparently the correct interpretation.

dracontes:
Well, `\cardsfolder` is fully uploaded. Regarding this PR should I split off what seems to me to be less questionable: `\adventure\common\custom_cards` and `\tokenscripts`? Also, is it worthwhile to bring `AiController.java` up to speed?

Hanmac:
> Well, `\cardsfolder` is fully uploaded. Regarding this PR should I split off what seems to me to be less questionable: `\adventure\common\custom_cards` and `\tokenscripts`? Also, is it worthwhile to bring `AiController.java` up to speed?

first i would do the custom_cards and the tokenscripts

for the AIController thing, should be separate MR

Issue #5717, opened 2024-07-29
Migrate guava Function/Predicate to java


Hanmac:
These functions are Legacy:
```java
import com.google.common.base.Function;
import com.google.common.base.Predicate;
```

use these instead

```java
import java.util.function.Function;
import java.util.function.Predicate;
```

If possible, try to remove the usage of anonymous classes with lambda expressions and/or method references instead of classes.

Jetz72:
I think we could also get guava's `Optional`s and `Supplier`s. Was already planning to do "anonymous -> lambda -> method reference" wherever possible in the follow up to #5703. I can include this as well.

Hanmac:
@Jetz72 I think an easy thing to refactor could be to remove the `CardPredicates.Accessors` and replace `fnGetNetPower` with `Card::getNetPower`

and because of these Functional Interfaces, it shouldn't make a difference

Jetz72:
So I tried inlining the items in `CardPredicates.Accessors` as method references, and that worked great. Then the next thing on my agenda was the anonymous to lambda to method reference cleanup. But it turns out a lot of the static ones are in the same boat as the accessors - it just made more sense to have `PaperCard::getName` in the code than `PaperCard.FN_GET_NAME`, since the latter is just an extra step in getting to `PaperCard::getName`. So I tried inlining all the static Function fields, and that went well for all but a couple ones that were already a bit messy to begin with (e.g. [DeckProxy](https://github.com/Card-Forge/forge/blob/0a474c955a023a9de85b0eb48ae06224b71da57d/forge-gui/src/main/java/forge/deck/DeckProxy.java#L678)). I've since started on some of the other functional interface types and I'm starting to see the same pattern again.

So now it's a question of how far to go with this. A huge chunk of methods in CardPredicates reduce to method references and could be cleanly inlined. With the addition of some simple helper methods on Card, I think most of the remaining ones could be simplified the same way. 

This would also simplify the process of migrating from the old Guava libraries to Java ones. I tried having the IDE do it automatically but it had very little success. Guava's `Predicate` and `Function` types are used explicitly in many places, and to change one you often have to change a large web of them. There are also a couple points where they're fed back into other Guava classes that ask for Guava predicates specifically. Lambdas and method references are much more flexible and aren't specific to a single functional interface. In many cases already I've been able to cut out the import statement entirely.

Does this seem like a good direction to proceed in?

Hanmac:
[...]

PR #5737, opened 2024-07-31
Miscallaneous Cleanup Part 2 - Lambdas and Method References

Jetz72:
Follow up to #5703. This time the biggest focus was converting anonymous classes to lambdas, and lambdas to method references wherever possible. This also helps a bit with shaking the dependencies on guava libraries (#5717).

Static `Function` fields have also been inlined as method references. There's potential to do the same with other functional interface types, namely predicates. But those may take a few extra steps, and this is probably already enough of a leap for one PR.


forge-core/src/main/java/forge/card/CardRulesPredicates.java
```diff
@@ -160,19 +145,16 @@ public static Predicate<CardRules> joinedType(final PredicateString.StringOp op,
...
-                final Set<String> set = card.getType().getCreatureTypes();
-                for (final String creatureType : creatureTypes) {
-                    if (set.contains(creatureType)) {
-                        return true;
-                    }
+        return card -> {
+            if (!card.getType().isCreature()) { return false; }
+
+            final Set<String> set = card.getType().getCreatureTypes();
+            for (final String creatureType : creatureTypes) {
```
Hanmac, line 152: might use `!Collections.disjoint` to check for overlapping

forge-core/src/main/java/forge/deck/Deck.java
```diff
@@ -125,12 +123,7 @@ public List<PaperCard> getCommanders() {
             result.add(c.getKey());
         }
         if (result.size() > 1) { //sort by type so signature spell comes after oathbreaker
-            Collections.sort(result, new Comparator<PaperCard>() {
-                @Override
-                public int compare(final PaperCard c1, final PaperCard c2) {
-                    return Boolean.compare(c1.getRules().canBeSignatureSpell(), c2.getRules().canBeSignatureSpell());
-                }
-            });
+            Collections.sort(result, (c1, c2) -> Boolean.compare(c1.getRules().canBeSignatureSpell(), c2.getRules().canBeSignatureSpell()));
```
Hanmac, line 126: would `Comperator.comparing(c -> c.getRules().canBeSignatureSpell())` work?

forge-core/src/main/java/forge/CardStorageReader.java
```diff
@@ -227,14 +227,11 @@ public final Iterable<CardRules> loadCards() {
...
-            public int compare(final CardRules o1, final CardRules o2) {
-                if (loadingTokens) {
-                    return String.CASE_INSENSITIVE_ORDER.compare(o1.getNormalizedName(), o2.getNormalizedName());
-                }
-                return String.CASE_INSENSITIVE_ORDER.compare(o1.getName(), o2.getName());
+        final Set<CardRules> result = new TreeSet<>((o1, o2) -> {
+            if (loadingTokens) {
+                return String.CASE_INSENSITIVE_ORDER.compare(o1.getNormalizedName(), o2.getNormalizedName());
             }
+            return String.CASE_INSENSITIVE_ORDER.compare(o1.getName(), o2.getName());
```
Hanmac, line 234: What about `Comperator.comparing(CardRules::getName, String.CASE_INSENSITIVE_ORDER)` ?

forge-core/src/main/java/forge/deck/Deck.java
Jetz72: Yeah, that looks better.

forge-core/src/main/java/forge/CardStorageReader.java
Jetz72: Looking at it, I think the `loadingTokens` check can also be moved out of the comparator, since that field is final.

forge-core/src/main/java/forge/card/CardRulesPredicates.java
Hanmac: see there: https://www.baeldung.com/java-check-elements-between-lists

Collections.disjoint returns true if there is no element from set A matching elements from set B,
so we just negate them.

it should stop and return false if first match is found (and then negated)

Jetz72: Ah, I see what you mean. Done.

tool4ever (approved):

PR #5744, opened 2024-07-31
Destroy: refactor fake into SacrificeAll

### LINKED ISSUE #5719: Destroy Sacrifice$ Cards needs to be reworked
@tool4ever about these `DB$ Destroy | Defined$ Self | Sacrifice$ True` cards like Ordeal, can we try to rewrite them using Sacrifice Effect instead of Destroy Effect? 

imo it feels wrong if the Destroy Effect doesn't Destroy (See also #5713 )

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/issues/5716#issuecomment-2255706281_

I think the Engine should support them already?

Currently 18 Cards left: https://github.com/search?q=repo%3ACard-Forge%2Fforge%20%22Sacrifice%24%20True%22&type=code

Also, some effects uses the opposite: `Sacrifice | Destroy$ True` these should be changed too

Hanmac:
Closes #5719 

This reworks "Sacrifice" Effects that were hiding inside Destroy Effects via SacrificeAll

i'm not yet happy with them, i think i would like them more if they were Sacrifice Effects instead of All.
or that Defined is required then using Targets?

- [x] Remove Engine Support for `Sacrifice$`
- [x] move AI for MadSarkhanDragon

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/DamageEachAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/DamageEachAi.java
@@ -26,6 +26,7 @@ public class DamageEachAi extends DamageAiBase {
             }
             sa.resetTargets();
             sa.getTargets().add(weakestOpp);
+            return weakestOpp.canLoseLife() && !weakestOpp.cantLoseForZeroOrLessLife();
         }
         
         final String damage = sa.getParam("NumDmg");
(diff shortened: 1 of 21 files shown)


Hanmac:
@tool4ever please check out this MR too

tool4ever (approved):
more cleanup would indeed be nice but good enough for now

PR #5746, opened 2024-07-31
TokenEffect no longer has state

tool4ever:
No more unique effect classes for each SA lol...

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/ApiType.java
+++ b/forge-game/src/main/java/forge/game/ability/ApiType.java
@@ -189,7 +189,7 @@ public enum ApiType {
     TapOrUntap (TapOrUntapEffect.class),
     TapOrUntapAll (TapOrUntapAllEffect.class),
     TimeTravel (TimeTravelEffect.class),
-    Token (TokenEffect.class, false),
+    Token (TokenEffect.class),
     TwoPiles (TwoPilesEffect.class),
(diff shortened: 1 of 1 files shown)


Hanmac:
What is the difference?

forge-game/src/main/java/forge/game/ability/ApiType.java
```diff
@@ -189,7 +189,7 @@ public enum ApiType {
     TapOrUntap (TapOrUntapEffect.class),
     TapOrUntapAll (TapOrUntapAllEffect.class),
     TimeTravel (TimeTravelEffect.class),
-    Token (TokenEffect.class, false),
+    Token (TokenEffect.class),
```
Hanmac, line 192: @tool4ever if there are no more "non-stateless" Types, can the extra logic be removed?

tool4ever: imo it doesn't hurt to keep it since it's a clean implementation
(should the need for it ever arise again)

tool4ever:
> What is the difference?

I checked with sampling:
image

You see all other Effects can share the same instance because of this logic:
https://github.com/Card-Forge/forge/blob/a58a197945f4daa6b5198b4a7be8df6d84a2b91a/forge-game/src/main/java/forge/game/ability/ApiType.java#L236

PR #5748, opened 2024-07-31
Generate individual booster slots with individual replacement percent…

tehdiplomat:
…ages

Convert MH3 to booster slot percentages


Agetian:
Looks good to me, nice job! 👍 There are some merge conflicts that need resolving. Also, would be nice if someone else also looked at the code 😊

PR #5798, opened 2024-08-03
Fix ChangeZoneEffect ordering to library; Aetherspouts prompt

### LINKED ISSUE #5466: Ordering the top of player's library  sometimes works incorrectly
**Describe the bug**
When using effects which allow you to order the top X cards of your library, after the screen where you sort them starting with "closest to top", the cards will sometimes be in incorrect order. This seems to happen consistently with Brainstorm and Brainsurge

**To Reproduce**
Here is my most recent example

1. Cast Brainsurge from hand
2. Choose Echo of Eons and Heliod's Intervention to put back
3. On the order screen, order so that Echo of Eons is "Closest to top"
4. Activate Elminster's -3 ability, exiling Heliod's Intervention 

**Expected behavior**
The expected behaviour is to have the card I selected as "Closest to top" be the one actually on top of the library

**Screenshots**

 - Device: Xiaomi Redmi Note 12 pro 
 - OS: Android 11
 - Version: 1.6.63-SNAPSHOT-06-15

**Additional context**
This might be intended behavior, but in that case I think it's unintuitive for the card that you selected as "Closest to top" to not actually end up being closest to top

Jetz72:
Resolves #5466.

While looking for cards to test that issue, I ran into [Aetherspouts](https://scryfall.com/card/mic/97/aetherspouts), which I think badly needs a rework for UX purposes. Currently, it prompts you for the ordering of the cards, then prompts you one card at a time whether to send the card to the top or bottom. This causes the order on top to be reversed, much like Brainsurge was doing here, since it puts cards on top of the deck one at a time. 

It also didn't specify which card you were moving for each individual prompt, meaning you had to memorize the order you put them in if you want to decide case by case. I fixed this last issue by adding support for `ShowCurrentCard` to generic choices, but I think the implementation probably needs to be redone at some point. Possibly to one where you split into piles for the top and bottom first, then order them.


forge-game/src/main/java/forge/game/ability/effects/ChangeZoneEffect.java
```diff
@@ -516,10 +516,15 @@ private void changeKnownOriginResolve(final SpellAbility sa) {
...
             } else {
-                tgtCards = GameActionUtil.orderCardsByTheirOwners(game, tgtCards, destination, sa);
+                if (sa.hasParam("Chooser"))
+                    tgtCards = chooser.getController().orderMoveToZoneList(tgtCards, destination, sa);
+                else
+                    tgtCards = GameActionUtil.orderCardsByTheirOwners(game, tgtCards, destination, sa);
+
+                final CardCollection reversed = new CardCollection(tgtCards);
+                Collections.reverse(reversed);
+                tgtCards = reversed;
```
Hanmac, line 527: @tool4ever i think the extra Reverse seems a bit ugly ... can we do something about this?

Jetz72: Wasn't too happy with that part either, but I couldn't find an elegant way to reverse a CardCollectionView.

Hanmac: @tool4ever probably needs to look into the `orderMoveToZoneList` and `orderCardsByTheirOwners` functions, that for destination=Library already reverses the CardCollection

tool4ever: Since this seems more of a convenience thing to select the order opposite from the zone changes it should probably belong to GUI code (PlayerControllerHuman)
or just switch the "bottom" + "top" labels around? :p
either way then all places in effect classes could be cleaned up

Jetz72: It's a question of whether you want the output of `orderMoveToZoneList` to mean "the player wants the cards in this order" versus "this is the order you should move cards one-by-one into this zone and position". Whatever is settled on, it would probably be good idea to add a Javadoc comment explaining the function's behavior when ordering cards for the top of a deck.

tool4ever: Second one makes more sense?
So all loops can be for each

Hanmac: i don't know, we probably need more info like @tehdiplomat

tehdiplomat: Ideally, we would have a different UI element thats a Scry type. Where you can reorder a library AND send to a different zone (including bottom of the library). I know people have had similar issues with Sylvan Library, where the ordering and the place on library didn't up matching their expectations. It seems weird to reverse the array when we already have labels to say which order things will be in.

Jetz72:
Just waiting on a resolution to the a resolution to the above. I could go back and apply `reverse` inside `orderMoveToZoneList` and remove it elsewhere, or just leave it as is. Just need to know which to do.

tool4ever:
Imo the contract between engine and GUI/AI should be that it just always gets returned the order to start moving from the first element
then GUI is free to do whatever is more user friendly...

tool4ever (approved):
thanks for seeing this through

dusan-madzarevic:
As the original reporter, thanks from me as well :)

PR #5845, opened 2024-08-04
UNF last 3 attractions

Simisays:
Tunnel of Love needs some more work, it doesn't seem to exile the card chosen by

### DIFF
--- a/forge-gui/res/cardsfolder/s/swinging_ship.txt
+++ b/forge-gui/res/cardsfolder/s/swinging_ship.txt
@@ -1,0 +1,12 @@
+Name:Swinging Ship
+ManaCost:no cost
+Types:Artifact Attraction
+Variant:A:Lights:2 6
+Variant:B:Lights:4 6
+K:Visit:TrigEffect
+SVar:TrigEffect:DB$ Effect | Triggers$ STShip | SpellDescription$ Visit — After the first combat phase this turn, there's an additional combat phase. At the beginning of that combat, untap all creatures that attacked this turn.
+SVar:STShip:Mode$ Phase | Phase$ EndCombat | ValidPlayer$ You | FirstCombat$ True | Execute$ TrigAdditionalCombat | TriggerDescription$ After the first combat phase this turn, there's an additional combat phase. At the beginning of that combat, untap all creatures that attacked this turn.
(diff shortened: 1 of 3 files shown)


forge-gui/res/cardsfolder/t/tunnel_of_love.txt
```diff
@@ -0,0 +1,15 @@
+Name:Tunnel of Love
+ManaCost:no cost
+Types:Artifact Attraction
+Variant:A:Lights:2 6
+Variant:B:Lights:3 6
+K:Visit:TrigChoose
+SVar:TrigChoose:DB$ ChooseCard | ValidTgts$ Opponent | Choices$ Creature.TargetedPlayerCtrl | AILogic$ WorstCard | Mandatory$ True | Amount$ 1 | SubAbility$ DBChooseYou | SpellDescription$ Choose an opponent. They choose a creature they control, then you choose a creature you control. You may exile the chosen creatures. If you do, return them to the battlefield under their owners’ control at the beginning of the next end step. Otherwise, the chosen creatures fight each other.
```
tool4ever, line 7: - this shouldn't target
- you need to remember each choice

forge-gui/res/cardsfolder/s/swinging_ship.txt
```diff
@@ -0,0 +1,12 @@
+Name:Swinging Ship
+ManaCost:no cost
+Types:Artifact Attraction
+Variant:A:Lights:2 6
+Variant:B:Lights:4 6
+K:Visit:TrigEffect
+SVar:TrigEffect:DB$ Effect | Triggers$ STShip | SpellDescription$ Visit — After the first combat phase this turn, there's an additional combat phase. At the beginning of that combat, untap all creatures that attacked this turn.
```
tool4ever, line 7: why the extra effect?

Simisays: I've used Zariel, Archduke of Avernus as a template since that cards effect sounds most similar to this card. Would you impement it diffently ?

tool4ever: it should probably be coded more like _World at War_
however the "first combat" part might be difficult then without more support
I'm just gonna approve it since taking Unfinity too seriously is against the spirit lol?

tool4ever (approved):

PR #5853, opened 2024-08-05
GameActionUtil: store OptionalKeywordAmount in CastSA

Hanmac:
@tool4ever i moved the amount from the trigger object, into the CastSA info

might need a better name for "CastSAOptKeyword"

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
@@ -907,13 +907,16 @@ public class CardFactoryUtil {
             String abString = "DB$ CopySpellAbility | Defined$ TriggeredSpellAbility | MayChooseTarget$ True";
             String[] k = keyword.split(":");
             if (k.length > 2) {
-                abString = abString + " | " + k[2];
+                abString += " | " + k[2];
             }
 
             final Trigger casualtyTrigger = TriggerHandler.parseTrigger(trigScript, card, intrinsic);
-            casualtyTrigger.setOverridingAbility(AbilityFactory.getAbility(abString, card));
-            casualtyTrigger.setSVar("Casualty", "0");
-            casualtyTrigger.setSVar("CasualtyPaid", "0");
+            SpellAbility sa = AbilityFactory.getAbility(abString, card);
+            sa.setSVar("CasualtyPaid", "Count$CastSAHasOptKeyword");
+            sa.setSVar("Casualty", "Count$CastSAOptKeyword");
+            casualtyTrigger.setOverridingAbility(sa);
+            casualtyTrigger.setSVar("CasualtyPaid", "Count$CastSAHasOptKeyword");
+            casualtyTrigger.setSVar("Casualty", "Count$CastSAOptKeyword");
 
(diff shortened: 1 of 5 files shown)


Hanmac:
@tehdiplomat your opinion on the Name of the Functions?

forge-gui/res/cardsfolder/o/ob_nixilis_the_adversary.txt
```diff
@@ -2,7 +2,7 @@ Name:Ob Nixilis, the Adversary
 ManaCost:1 B R
 Types:Legendary Planeswalker Nixilis
 Loyalty:3
-K:Casualty:X:NonLegendary$ True | SetLoyalty$ Casualty:The copy isn't legendary and has starting loyalty X.
+K:Casualty:X:NonLegendary$ True | SetLoyalty$ CasualtyPaid:The copy isn't legendary and has starting loyalty X.
```
tool4ever, line 5: broken now for X=0

Hanmac: fixed

Hanmac:
@tool4ever can you help me find a better name for the attribute and functions?

tool4ever:
well `OptionalGenericCostPaid` already exists
so maybe `OptionalKeywordCostPaid[Times]`?

Hanmac:
@tool4ever I got some better idea for this MR

if I extend the Map into a Table using Keyword as additional Key like this:
`Map<Pair<Long, Long>, Integer>` → `Table<Keyword, Pair<Long, Long>, Integer>`

then I can use this for other cases like Multikicker too

Hanmac:
@tool4ever I renamed my stuff and used a Table for future uses (like Multikicker)

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -907,13 +907,16 @@ public static void addTriggerAbility(final KeywordInterface inst, final Card car
...
+                abString += " | " + k[2];
             }
 
             final Trigger casualtyTrigger = TriggerHandler.parseTrigger(trigScript, card, intrinsic);
-            casualtyTrigger.setOverridingAbility(AbilityFactory.getAbility(abString, card));
-            casualtyTrigger.setSVar("Casualty", "0");
-            casualtyTrigger.setSVar("CasualtyPaid", "0");
+            SpellAbility sa = AbilityFactory.getAbility(abString, card);
+            sa.setSVar("CasualtyPaid", "Count$hasOptionalKeywordAmount");
+            sa.setSVar("Casualty", "Count$OptionalKeywordAmount");
```
tool4ever, line 916: the trigger should only need the first SVar and the sa only the second when inside that if part?

Hanmac: probably, should i change that, or should i keep it? i mean the extra doesn't hurt?

Hanmac:
also on my local branch, i testing out Multikicker for Human, it works

AI might need more love

tool4ever (approved):

PR #5854, opened 2024-08-06
File end cleanup for August 2024: every relevant script folder except for 'upcoming'

dracontes:
Making sure script files end in an empty new line and that their line breaks are all set to Unix (LF) rather than Windows (CR LF),


Hanmac (approved):

Hanmac:
@dracontes @tool4ever we could set ".gitattributes" for the [forge-gui/res/cardsfolder] so files should always be in the wanted result?

dracontes:
> we could set ".gitattributes" for the [forge-gui/res/cardsfolder] so files should always be in the wanted result?

I'm all for either or both of these formatting guidelines being automated.

PR #5863, opened 2024-08-06
Update AiController.java

Hanmac:
Add RepParams for AddCounter


tool4ever:
there's also a check in `ComputerUtil.isETBprevented()` that needs updating, then it should be ok?

tool4ever (approved):

PR #5865, opened 2024-08-06
Fix Uphill Battle

tool4ever:


dracontes:
The [[Double Major]] case now works as expected with the token copy entering untapped. 
I did however also want to consider [[Garth One-Eye]] which creates a copy of a card that is then cast. In that case, choosing Shivan Dragon, the resulting token entered untapped, which, as the result of a cast spell, probably shouldn't be the case (see appended screenshot with game log).
image

forge-game/src/main/java/forge/game/ability/effects/PlayEffect.java
```diff
@@ -169,6 +169,7 @@ public void resolve(final SpellAbility sa) {
             // so it gets added to stack
             card.setCopiedPermanent(card);
             card.setToken(true);
+            card.setZone(controller.getZone(ZoneType.None));
```
Jetz72, line 172: Would it work to just cut out `card.setCopiedPermanent` and `card.setToken` and replace them with `card.setCopiedSpell(true)`? I think PlayEffect is the only place where a card is made into a token while on the stack.

tool4ever: yea I'm not super happy with how these params are overlapping, investigating a good path for cleanup might be a worthy project
but just as a note if you didn't notice, I did have to fix this area for that last time:
https://github.com/Card-Forge/forge/pull/5230/commits/b05f9f1

dracontes (approved):
The Garth!Shivan Dragon now enters tapped. Ran a few more tests with cast from exile effects, like foretell, plot, and plain static command zone effects like Dream-Thief's Bandana to be on the safe side. Also ran Panglacial Wurm through its paces. All of it works as expected 👍

PR #5881, opened 2024-08-08
GameActionUtil: use Multikicker as KeywordExtraCost

Hanmac:
- [x] Multikicker uses new Table
- [x] AI need some more love, i disabled some code, but AI should use new chooseNumberForKeywordCost instead
- [x] some more cleanup the remaining `isAnnouncing("Multikicker")`

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtil.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtil.java
@@ -124,8 +124,6 @@ public class ComputerUtil {
         // Spell Permanents inherit their cost from Mana Cost
         final Cost cost = sa.getPayCosts();
 
-        // Remember the now-forgotten kicker cost? Why is this needed?
-        //sa.getHostCard().setKickerMagnitude(source.getKickerMagnitude());
         game.getStack().freezeStack(sa);
 
         // TODO: update mana color conversion for Daxos of Meletis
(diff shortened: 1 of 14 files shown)


Agetian (changes requested):
Please make sure the AI code is properly updated to support the new format before merging the code. Sadly, I won't be able to help out before August 20-21 or so (out of town, can only look at stuff and maybe meege in some smaller modifications). Overall, nice idea!

Hanmac:
One goal that may be for this MR is to move chooseNumberForKeywordCost more into the SpellAbilityEffectAI parts

so the AI might decide if it really wants to sacrifice something for Casualty, if the Player doesn't have the Mana to pay for the first unless cost

Hanmac:
@tool4ever can you test if the AI is good enough for it to play it with Multikicker if able?

tool4ever:
doesn't seem like it, never casts _Everflowing Chalice_ now :/

Hanmac:
> doesn't seem like it, never casts _Everflowing Chalice_ now :/

fixed AI

tool4ever (approved):
seems ok, it should really be hooked up earlier in `calculateManaCost` somehow already to avoid potential misplay but this is cleaner for now

Hanmac:
@Agetian can you look more into the AI parts?

or can this be merged?

Agetian:
I'll be back in town in 2-3 days and I'll take a deeper look at it ;) sadly, can't test or assess things before then, I only have my phone with me ... ^^;

Agetian (approved):

Agetian:
Looks good to me :) I think the commented out parts in the AI can be safely removed

Agetian (approved):

PR #5889, opened 2024-08-10
GamePieceType and isCollectible

Jetz72:
This consolidates the following card boolean properties into an enum called `GamePieceType`:
* `token`
* `copiedSpell`
* `isImmutable`
* `isAttractionCard`

Since these states are mutually exclusive and all indicate the kind of physical card or object that a Card behaves as, this seemed more appropriate than adding more flags as other nontraditional card types are introduced. Other GamePieceTypes for existing nontraditional card types - namely Planar, Scheme, Avatar, and Dungeon - are also included. See javadoc in `GamePieceType.java` for more details.

Also added is a Card flag called `collectible`. This distinguishes between cards in a player's collection that they brought to the game, and cards that were conjured or otherwise created during the game. Currently the only use of this is to prevent the edge case of a conjured card or token card (e.g. Time Sidewalk) being anted and lost when the player doesn't truly own a copy. See javadoc of `Card::isCollectible` for more details.


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -248,6 +248,8 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
             // CR 707.12 casting of a card copy
             if (zoneTo.is(ZoneType.Stack) && c.isRealToken()) {
                 copied.setCopiedPermanent(c.getCopiedPermanent());
+                //TODO: Feels like this should fit here and seems to work but it'll take a fair bit more testing to be sure.
+                //copied.setGamePieceType(GamePieceType.COPIED_SPELL);
```
tool4ever, line 252: copying a spell is still different from casting, so really shouldn't get switched here

Jetz72: I don't think identifying it as a copy should take away from it being "cast" since there's still the distinction between moving to the stack versus being created in that zone already. After the point where the copy is cast and put on the stack, the CR doesn't seem to distinguish it from a copied spell.

I've been looking into options to consolidate these because I've seen a number of places where a copied card that gets cast needs to be distinguished from a token in the wrong zone by checking copiedPermanent, e.g. 
https://github.com/Card-Forge/forge/blob/05b72654739c8641443203e92c315fc38e68e4da/forge-game/src/main/java/forge/game/GameAction.java#L1675-L1677

https://github.com/Card-Forge/forge/blob/05b72654739c8641443203e92c315fc38e68e4da/forge-game/src/main/java/forge/game/zone/Zone.java#L132-L138

There was also the recent change to the Adventure ability, where checking if the spell on the stack is a card necessitated both nonToken and nonCopiedSpell.

It's something I have an interest in revisiting, maybe with a few unit tests covering the function and related game logic to make sure nothing is getting broken. But if you'd prefer, I can clean up the comment.

tool4ever: gotcha, it looks like most places check `SpellAbility.isCopied()` for their logic so some consolidation seems well spent
thanks, more engine work is always appreciated :)

Hanmac (approved):

PR #5895, opened 2024-08-10
Miscallaneous Cleanup Part 3

Jetz72:
Continuing from #5737, this set of adjustments is a bit more varied in scope. A few more lambdas and method references have been applied that were passed over in the previous batch. Some helper methods have been applied and some wrapper methods have been unwrapped.

A significant change is the switch from `new FileInputStream` to `Files.newInputStream` in a few places, particularly when reading cards and other resources. As I understand it, the underlying implementation it uses on modern JVMs is much less prone to stalling for garbage collection. In practice, I've noticed the reported time it takes to load cards has gone from ~1700ms to ~1300ms, and I've yet to experience any prolonged hangs that I've occasionally encountered on startup previously. Fingers crossed.

### DIFF
--- a/forge-core/src/main/java/forge/CardStorageReader.java
+++ b/forge-core/src/main/java/forge/CardStorageReader.java
@@ -382,25 +382,14 @@ public class CardStorageReader {
      * @return a new Card instance
      */
     protected final CardRules loadCard(final CardRules.Reader reader, final File file) {
-        FileInputStream fileInputStream = null;
-        try {
-            fileInputStream = new FileInputStream(file);
+        try (InputStream fileInputStream = java.nio.file.Files.newInputStream(file.toPath())) {
             reader.reset();
             final List<String> lines = readScript(fileInputStream);
             return reader.readCard(lines, Files.getNameWithoutExtension(file.getName()));
         } catch (final FileNotFoundException ex) {
             throw new RuntimeException("CardReader : run error -- file not found: " + file.getPath(), ex);
         } catch (final Exception ex) {
-            System.out.println("Error loading cardscript " + file.getName() + ". Please close Forge and resolve this.");
-            throw ex;
-        } finally {
-            try {
-                assert fileInputStream != null;
-                fileInputStream.close();
-            } catch (final IOException ignored) {
-                // 11:08
-                // PM
-            }
+            throw new RuntimeException("Error loading cardscript " + file.getName() + ". Please close Forge and resolve this.", ex);
         }
     }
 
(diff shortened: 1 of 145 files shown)


forge-core/src/main/java/forge/CardStorageReader.java
```diff
@@ -382,25 +382,14 @@ private List<String> readScript(final InputStream inputStream) {
      * @return a new Card instance
      */
     protected final CardRules loadCard(final CardRules.Reader reader, final File file) {
-        FileInputStream fileInputStream = null;
-        try {
-            fileInputStream = new FileInputStream(file);
+        try (InputStream fileInputStream = java.nio.file.Files.newInputStream(file.toPath())) {
```
Hanmac, line 385: use `import java.nio.file.Files` ?

Jetz72: Clashes with `import com.google.common.io.Files`, which is used for `Files.getNameWithoutExtension`.

Hanmac: Sadly Java doesn't gave import aliases

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -7950,26 +7950,14 @@ public void addChosenModes(SpellAbility ability, String mode, boolean yourCombat
                 }
             }
         } else {
-            List<String> result = chosenModesTurn.get(original);
-            if (result == null) {
-                result = Lists.newArrayList();
-                chosenModesTurn.put(original, result);
-            }
+            List<String> result = chosenModesTurn.computeIfAbsent(original, k -> Lists.newArrayList());
```
Hanmac, line 7953: k -> Lists.newArrayList() can't be replaced with `Lists::newArrayList` or something, right?

Jetz72:
```
java: incompatible types: bad return type in lambda expression
    inference variable E has incompatible bounds
      equality constraints: java.lang.String
      lower bounds: forge.game.spellability.SpellAbility
```
Apparently not. It's trying to plug the key into the parameters of newArrayList, but that would create an `ArrayList<SpellAbility>` when the map wants an `ArrayList<String>`.

Hanmac: ah yeah because it is a Function and not a Supplier

Hanmac (approved):

tool4ever (approved):

PR #5898, opened 2024-08-10
Fix token CMC when copying

tool4ever:
This decade old shortcut would mean turning ordinary tokens into copies with e.g. _Mirrorweave_ would still keep their CMC at 0.

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -7230,9 +7230,6 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars {
         return getCMC(SplitCMCMode.CurrentSideCMC);
     }
     public int getCMC(SplitCMCMode mode) {
-        if (isToken() && getCopiedPermanent() == null) {
-            return 0;
-        }
         if (lkiCMC >= 0) {
             return lkiCMC; // a workaround used by getLKICopy
         }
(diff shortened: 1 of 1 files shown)


Hanmac:
Dies this still respect:
> 202.3b The mana value of a transforming double-faced permanent or spell’s back face is calculated as though it had the mana cost of its front face. If a permanent or spell is a copy of the back face of a transforming double-faced card (even if the card representing that copy is itself a double-faced card), the mana value of the copy is 0.

tool4ever:
@Hanmac 
if I understand correctly:
- token from _Helm of the Host_ equipped onto _Blitzwing, Cruel Tormentor_
- token then transforms into _Blitzwing, Adaptive Assailant_
- token should have CMC=0

?

seems this was also broken

Hanmac:
That fix should be for melded too

tool4ever:
Hmn during testing this didn‘t seem necessary?
Because from the rules change you can only make copies that can transform
And if you‘d clone a meld it would put it as Original state which has no mana cost anyway

Jetz72:
Maybe "even if that copy is represented by two other meld cards" refers to a scenario where you have two different melded permanents in play, and one becomes a copy of the other via something like Nanogene Conversion? That rule could really use an example though.

tool4ever:
Yes, but I meant the way the engine handles it iternally when getting a Cloned state it wouldn't run into this path any longer
though I can just future-proof it here for good measure :)

Hanmac (approved):

PR #5913, opened 2024-08-12
73 cards from Unknown Events 2023 (some being in mb2)

Lykrast:
Several cards from [gavin's unknown events](https://mtg.fandom.com/wiki/Unknown_Event) in 2023. There's an edition file and what I could implement for Philadelphia, Minneapolis, Barcelona, Indiana (which was the same as Barcelona but with different prizes), and Las Vegas ~~(where I short circuited on motivation so there's only like 3)~~.

That's currently 73 cards and 10 tokens.

Some notes:

* Some of these cards are in mystery booster 2 so they could be salvaged if needed (at least Mox Poison, Anax and Cymede and Kynaios and Tiro, and Questing Cosplayer)
* They don't have set codes and are not on scryfall, so I put ~~UNK~~ UE + first 3 letters of the town + 23 (for 2023, would have put 24 for the 2024 ones), could be changed
* Phila, Unsealed only has the text from the mirran victory (cause mirrans won that event), don't know if it should have the phyrexian victory text too like the actual printed card
* ~~Prize cards and the "you make the cards" are separated in the set files~~ Prize cards are marked as promos, and "you make the cards" got a bonus sheet like kaldheim snow lands cause they are not given the same way in the draft
* ~~Phila doesn't seem to show up as part of the set and I don't know how to fix it~~
* Draft is not made I dunno how the sealed formats work and if you can make them (mtg wiki has the booster composition for most of them)
* Haven't done files for the 2024 events ~~and most of the simple Las Vegas ones cause as I said I short circuited on motivation~~ and figured might as well show what I have now.
* **I kept the printed wording for these cards** (because they don't have oracle text) **but I can do the bloomburrow etb errata if needed.**


forge-gui/res/tokenscripts/role_questing.txt
```diff
@@ -0,0 +1,12 @@
+Name:Questing
+ManaCost:no cost
+Types:Enchantment Aura Role
+K:Enchant creature
+A:SP$ Attach | Cost$ 0 | ValidTgts$ Creature | AILogic$ Pump
+S:Mode$ Continuous | Affected$ Creature.EnchantedBy | AddKeyword$ Vigilance & Deathtouch & Haste | Description$ Enchanted creature has vigilance, deathtouch, and haste.
+S:Mode$ CantBlockBy | ValidAttacker$ Creature.EnchantedBy | ValidBlocker$ Creature.powerLE2 | Description$ Enchanted creature can't be blocked by creatures with power 2 or less.
+S:Mode$ CantPreventDamage | IsCombat$ True | ValidSource$ Creature.YouCtrl | Description$ Combat damage that would be dealt by creatures you control can't be prevented.
+T:Mode$ DamageDone | ValidSource$ Creature.EnchantedBy | ValidTarget$ Opponent | CombatDamage$ True | TriggerZones$ Battlefield | Execute$ MoreDamage | TriggerDescription$ Whenever enchanted creature deals combat damage to an opponent, it deals that much damage to target planeswalker that player controls.
+SVar:MoreDamage:DB$ DealDamage | ValidTgts$ Planeswalker.ControlledBy TriggeredTarget | TgtPrompt$ Select target planeswalker that player controls | NumDmg$ X
```
Hanmac, line 10: Don't know if the Trigger should also be gained to the Enchanted Creature?

Or if the DealDamage needs extra Damage Source

Hanmac:
@tool4ever we need to find a way to make such Test Cards not legal in Commander?

forge-gui/res/tokenscripts/role_questing.txt
Lykrast: Oh yeah you're right, it should be giving the abilities to the creature it'd make more sense given the wording.
Was kinda short-circuiting on those las vegas cards.

tool4ever:
Disabling non-legal cards in preferences should work thanks to the edition file

Lykrast:
Good news I have found what was wrong with Phila.
It looks like the tests still don't like it though, I think because of the wonky set codes I made up, though I really don't know what would be a fix for that (since those cards aren't on scryfall anyway).

Lykrast:
Hey guess what I got a bit of motivation and did the las vegas cards I could. Also fixed the etb tapped with the bloomburrow update.

For now I kept the printed text on these cards (since they don't have oracle text and likely will never get any update), but I'm open to doing a bloomburrow etb errata pass on them.

forge-gui/res/editions/Unknown Event Barcelona 2023 Playtest Cards.txt
```diff
@@ -0,0 +1,71 @@
+[metadata]
+Code=UNKBAR23
```
tool4ever, line 2: try shorten to less than 8
e. g. UEBAR23 etc.

Lykrast: Done

tool4ever (dismissed):

tool4ever (approved):

PR #5920, opened 2024-08-13
Update Sentry

kevlahnota:
- fix Java System info/properties


Hanmac:
@kevlahnota do we really need the Hints anywere?

I mean, isn't the data already stored in `bread.setData("SA", sa.toString());` ?

Also, `hint.set(sa.toString(), sa);` could cause problems if we really want to use the Hint data?

Because if we don't use them, we also could just skip the Hint Data?

kevlahnota:
> @kevlahnota do we really need the Hints anywere?
> 
> I mean, isn't the data already stored in `bread.setData("SA", sa.toString());` ?
> 
> Also, `hint.set(sa.toString(), sa);` could cause problems if we really want to use the Hint data?
> 
> Because if we don't use them, we also could just skip the Hint Data? The Hint Objects might be useful when using Events to filter them. But then they should have keys similar to the Breadcrumb Data, like: `hint.set("SA", sa);`
 
we could pass it as null or don't use it, Sentry change the passing of object to Hint
image

Hanmac (approved):

PR #5923, opened 2024-08-14
Edition updates: MB2, SLD, YBLB

paulsnoops:


Hanmac:
@paulsnoops @tehdiplomat shouldn't MLB be marked as "Fun" because of the Playtest cards?

paulsnoops:
> @paulsnoops @tehdiplomat shouldn't MLB be marked as "Fun" because of the Playtest cards?

It's supposed to be a draft set but I forgot to mark those cards with F in the collector number, will fix now

Hanmac:
@paulsnoops also isn't the Collector Number wrong? Doesn't they start with 999-Text?

https://scryfall.com/card/mb2/999-AC/avacyns-collar-the-symbol-of-her-church

paulsnoops:
Yeah but for now we don't support -XXX in collector numbers

Issue #5392, opened 2024-06-07
M3C Desert Warfare Return to Battlefield Not Happening


DaggerV:
Desert Warfare (M3C) triggers both effects as intended. However, the deserts never actually return to the battlefield at the end of turn, but remain in the graveyard.

tool4ever:
the problem is the combination of two trigger events where one uses LKI and the other doesn't

PR #5926, opened 2024-08-14
Update desert_warfare.txt

dracontes:
After a bit of methodical testing, this PR closes https://github.com/Card-Forge/forge/issues/5392 : sacrificed, milled and discarded Deserts are now returned to the battlefield as expected.

**Edit:** Even the case where you have an [Yedora, Grave Gardener](https://scryfall.com/card/mkc/197/yedora-grave-gardener), a [Dune Chanter](https://scryfall.com/card/otc/31/dune-chanter) (to give all lands the Desert subtype), and sacrifice outlets for both creatures (such as [Ashnod's Altar](https://scryfall.com/card/cmm/368/ashnods-altar)) and Forests (like [Fortitude](https://scryfall.com/card/usg/253/fortitude)): the once face-down Forest Desert is returned to the battlefield as the front-face creature, which I believe is the expected behavior.


tool4ever (approved):
the problem is it should still be LKICopy but that can be fixed later

Hanmac (approved):

PR #5933, opened 2024-08-15
You're In Command + Support

Jetz72:
Seems to work from my testing so far. Not sure if there needs to be some kind of View or network update if the game suddenly becomes a commander variant in the middle of it. I'm also tempted to try moving all the remaining "createCommanderEffect" calls into the `Player.setCommander` method so they can be more easily tracked and managed.


forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -2766,12 +2767,34 @@ public void onMulliganned() {
     public List<Card> getCommanders() {
         return commanders;
     }
-    public void setCommanders(List<Card> commanders0) {
-        if (commanders0 == commanders) { return; }
-        commanders = commanders0;
+    public void setCommanders(List<Card> commanders) {
+        if (new HashSet<>(this.commanders).containsAll(commanders)) { return; }
```
Hanmac, line 2771: imo should that be more optimized?

Should this.commanders be a CardCollection to stop duplicates?

Jetz72: That would make sense, yes. And I also realized it should probably be a test if the sets match exactly; if you have commanders A and B, setCommanders([A]) should remove B. I'll have another shot at this bit.

Hanmac:
What is this `IncludeAllComponentCards` used for?

Jetz72:
> What is this `IncludeAllComponentCards` used for?

Per the ruling on You're In Command: 
>If You’re in Command targets a creature that’s represented by multiple cards, each of those cards become your commander.

Currently, Defined or Targeted selectors will only include the primary card of a merged or melded permanent. I didn't see any way to have `getDefinedCardsOrTargeted` return all the other cards that make up a combined permanent, so that's what that parameter is for.

forge-gui/res/cardsfolder/y/youre_in_command.txt
```diff
@@ -0,0 +1,6 @@
+Name:You're in Command
+ManaCost:1 W
+Types:Sorcery
+A:SP$ AlterAttribute | ValidTgts$ Creature.YouOwn+YouCtrl | IncludeAllComponentCards$ True | Attributes$ Commander | RememberAltered$ True | SubAbility$ DBDemote | SpellDescription$ Choose target creature you own and control. That creature becomes your commander. Any other commanders you have are no longer your commander. (That creature starts with a commander tax of {0}.)
+SVar:DBDemote:DB$ AlterAttribute | Defined$ ValidAll Card.YouOwn+IsNotRemembered+IsCommander | Attributes$ Commander | Activate$ False
```
tool4ever, line 5: needs DBCleanup

forge-game/src/main/java/forge/game/player/Player.java
Jetz72: Alright, new version should be safe from duplicates, and should skip over any commanders on the input list that the player already has as their commander. It also manages the commander effect internally rather than leaving that to the call site. I also added a `copyCommandersToSnapshot` method so the map from commanders to commander effects could be preserved when copying the game state, and moved the logic for the other commander maps in there too.

forge-gui/res/cardsfolder/y/youre_in_command.txt
Jetz72: Added.

dracontes:
FWIW, I can report that Champions of Archery // Join the Group is still functioning as expected after the recent update.

forge-ai/src/main/java/forge/ai/GameState.java
```diff
@@ -1327,10 +1324,7 @@ else if (info.startsWith("OnAdventure")) {
                     c.setExiledWith(c); // This seems to be the way it's set up internally. Potentially not needed here?
                     c.setExiledBy(c.getController());
                 } else if (info.startsWith("IsCommander")) {
-                    c.setCommander(true);
-                    List<Card> cmd = Lists.newArrayList(player.getCommanders());
-                    cmd.add(c);
-                    player.setCommanders(cmd);
+                    player.addCommander(c);
```
tool4ever, line 1327: I can no longer cast my commander when loading a state

Jetz72:
Alright, take 3. Commander effects are now one-per-player and serve all commanders that player owns. 

I noticed that Signature Spells in Oathbreaker seem to be considered commanders which I don't think is correct. I hacked around that for the commander effect in the Oathbreaker format by looking for `Permanent.IsCommander+YouOwn` vs `Spell.IsCommander+YouOwn` but a proper `IsSignatureSpell` should be implemented down the line. As it stands, signature spells will revive Skyfire Pheonix, can be paid for by Jeweled Lotus, and be trapped forever in the command zone after playing You're In Command.

forge-ai/src/main/java/forge/ai/GameState.java
Jetz72: Should be fixed in the latest revision.

dracontes:
Tested Champions of Archery // Join the Group again. 

This time, while the commander replacement effect and the commander tax work fine, it turns out that the commander damage tracking isn't working properly, perhaps as far as the case for multiple commanders is concerned. 
Each combat the tracker logs each hit for each commander but fails to add the damage up, instead replacing an older logged hit amount with either the most recent one for that commander or just blanking the amount each combat damage step there's combat damage dealt by any commander. Take Champions of Archery (121) which, according to the game log, should have 26 commander damage to its name and Silvos, Rogue Elemental (129) should have 38.

incommand_battlefield
_Screenshot of board state for this test run._
 
incommand_partiallog
_Game log showing the game continuing despite enough commander damage being dealt._

incommand_logging
_Card detail panes for two of the commanders that dealt damage for the game log above showing that the damage amounts aren't being logged properly._

Also, mildest of complaints, but the automatic Commander Effect text snippet could be a bit more context aware.

commander_adventure_prompt

dracontes:
>  [I]t turns out that the commander damage tracking isn't working properly, perhaps as far as the case for multiple commanders is concerned.

Well, felt my curiosity piqued enough to check: commander damage tracking does work as expected for the single commander case, the partner commander case, the case where a mutate pile that was made into a commander with You're in Command left the battlefield and then was recast from the command zone as its component parts.
I'm looking at the script for Champions of Archery // Join the Group, which I realize I should have supplied right away,`*` and contrasting it with You're in Command. I can only think, for the snippet reproduced below, that the `IsNotCommander`  bit in the argument for `Defined$` is somehow messing things up. I added it to head off issues with duplicating the attribute in case it's for some reason already there. Maybe that's unnecessary?
```
SVar:YouAreCommander:DB$ AlterAttribute | Defined$ ValidCommand Card.IsRemembered+IsNotCommander | Attributes$ Commander | SubAbility$ DBCleanup
```
,`*` Here it is: [champions_of_archery.txt](https://github.com/user-attachments/files/16737490/champions_of_archery.txt)

Jetz72:
Got it. It was an issue with snapshot restoring. Also cleaned up the "by Commander Effect", and updated CardCopyService to properly restore DetachedCardEffects.

`IsNotCommander` should be unnecessary for your card script. Making a card that's already a commander into your commander will have no effect.

tool4ever (changes requested):
```
Game-0 > java.lang.ClassCastException: class forge.game.card.Card cannot be cast to class forge.game.ability.effects.DetachedCardEffect (forge.game.card.Card and forge.game.ability.effects.DetachedCardEffect are in unnamed module of loader 'app')
	at forge.game.player.Player.copyCommandersToSnapshot(Player.java:2815)
	at forge.ai.simulation.GameCopier.makeCopy(GameCopier.java:133)
	at forge.ai.simulation.GameSimulator.<init>(GameSimulator.java:36)
	at forge.ai.simulation.SpellAbilityPicker.evaluateSa(SpellAbilityPicker.java:381)
	at forge.ai.simulation.SpellAbilityPicker.chooseSpellAbilityToPlayImpl(SpellAbilityPicker.java:195)
	at forge.ai.simulation.SpellAbilityPicker.formulatePlanWithPhase(SpellAbilityPicker.java:119)
	at forge.ai.simulation.SpellAbilityPicker.createNewPlan(SpellAbilityPicker.java:154)
	at forge.ai.simulation.SpellAbilityPicker.chooseSpellAbilityToPlay(SpellAbilityPicker.java:113)
	at forge.ai.AiController.chooseSpellAbilityToPlay(AiController.java:1354)
	at forge.ai.PlayerControllerAi.chooseSpellAbilityToPlay(PlayerControllerAi.java:808)
```

Jetz72:
Game snapshots are going to be the death of me.

tool4ever (approved):
potential for future cleanup, but seems stable now, thanks

PR #5951, opened 2024-08-19
Break out main loop step for easier testing and AI dev: now with a feature flag!

marthinwurer:
- Broke out the main game loop into mainLoopStep() and setupFirstTurn()
- Added a feature flag in settings to enable it for games
- Implemented some basic combat tests
- Ported some SpellAbilityPickerSimulationTests over to simply SpellAbilityPickerTests
- Added a (currently ignored) game fuzzing test
- Added some basic combat tests
- Probably more, I wrote most of this two weeks ago


Hanmac:
I don't see why `public void startFirstTurn(Player goesFirst, Runnable startGameHook)` can't call the new `mainGameLoop()` there too

no need to have the code duplicated

tool4ever:
Well, I guess this attempt looks less risky than your previous one :p
Imo a preference is overkill though and might actually make things harder since users switching that around would lead to inconsistent reports (even with the log print)

marthinwurer:
Closing to have #6727 be the main effort for this

PR #5961, opened 2024-08-21
Adding 2022 Commander decks to be purchased in the Quest shop

Sunnovah:


forge-gui/res/quest/precons/Bedecked Brokers.dck
```diff
@@ -0,0 +1,103 @@
+[shop]
+WinsToUnlock=0
+Credits=5000
+MinDifficulty=0
+MaxDifficulty=5
+[metadata]
```
tool4ever, line 6: @tehdiplomat
isn't there a better way yet than to duplicate these files? 😢
https://github.com/Card-Forge/forge/blob/master/forge-gui/res/quest/commanderprecons/Bedecked%20Brokers%20[NCC]%20[2022].dck

tehdiplomat: Yea if we already have these files, then we should just have a list of directories of where to pull them from.

Sunnovah: They are in commanderprecons. I can just add the cost block to everything in there as long as the line in forge/forge-gui/src/main/java/forge/localinstance/properties/ForgeConstants.java):

public static final String QUEST_PRECON_DIR             = QUEST_DIR + "precons" + PATH_SEPARATOR;

Can be coded to read from both directories.

tool4ever: I think we might
but wouldn't it be better to only use them when quest is commander mode too? 🤔

Sunnovah: Maybe, depends if there are any useful cards in the precons. Isn't there a commander plane in the regular quest mode as well?

tehdiplomat: @Sunnovah were you going to make the update? Just add prices to original spots, and have the files be pulled from both locations?

PR #5963, opened 2024-08-22
Adventure enemy freeze fix

misha-colbourne:
Fixed dungeon enemies not waiting for you to get out of their unfreeze range before engaging again, they now idle correctly until out of range.


Simisays:
Looks good, thanks !

Agetian (approved):

PR #5970, opened 2024-08-22
Fixed missing elements from HUD in towns/dungeons/etc

misha-colbourne:
Fixed missing elements from HUD when not on world map (now translucent instead of invisible)


kevlahnota:
the hidden elements are intended to be invisible since some tiledmap objects views are blocked by those HUD Elements and players may miss important objects. Worst view would be portrait mode on mobile. I guess a gesture on the upper right corner touch area to show temporarily the playar avatar and stats would be better.

misha-colbourne:
For now I'll add an entry to the game options for this, so at least players on pc and such can see them, leaving it toggled off by default so it's an opt-in to have them visible

misha-colbourne:
Locally have it looking like so
World map:
image
Dungeon map:
image

Would that not be translucent enough for players to see the objects/enemies/etc through? Without the backing box I'd imagine that's ok?

kevlahnota (approved):

PR #5975, opened 2024-08-23
Fix AI for Fireball and Shatterskull Smashing

### LINKED ISSUE #5899: AI doesn't cast Shatterskull Smashing
**Describe the bug**
AI doesn't cast Shatterskull Smashing even if it have enough lands and player has 2+ creatures

**To Reproduce**
1. AI has untapped 8+ lands on the battlefield and 2 Shatterskull Smashing in hands
2. Player has 2 dorks and Baneslayer Angel
3. AI doesn't cast Shatterskull Smashing

**Expected behavior**
AI should cast Shatterskull Smashing

**Additional context**
AI casts Spikefield Hazard and Bala Ged Recovery

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.64-SNAPSHOT-08.04

**Screenshots**

shatt
shatt2

Agetian:
Might need a refinement, since I'm not sure if the other if branches are also broken now and need similar treatment (e.g. the InYourHand if branch), and also I'm not sure if we need to keep the previous checks, which are no longer relevant at least for Fireball and Shatterskull Smashing.
Fixes #5899

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/DamageDealAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/DamageDealAi.java
@@ -95,8 +95,8 @@ public class DamageDealAi extends DamageAiBase {
         final String damage = sa.getParam("NumDmg");
         int dmg = AbilityUtils.calculateAmount(source, damage, sa);
 
-        if (damage.equals("X") || sourceName.equals("Crater's Claws")) {
-            if (sa.getSVar(damage).equals("Count$xPaid") || sourceName.equals("Crater's Claws")) {
+        if (damage.equals("X") || source.getSVar("X").equals("Count$xPaid") || sourceName.equals("Crater's Claws")) {
+            if (sa.getSVar("X").equals("Count$xPaid") || sa.getSVar(damage).equals("Count$xPaid") || sourceName.equals("Crater's Claws")) {
                 dmg = ComputerUtilCost.getMaxXValue(sa, ai, sa.isTrigger());
 
                 // Try not to waste spells like Blaze or Fireball on early targets, try to do more damage with them if possible


Hanmac (approved):

forge-ai/src/main/java/forge/ai/ability/DamageDealAi.java
```diff
@@ -95,8 +95,8 @@ protected boolean canPlayAI(Player ai, SpellAbility sa) {
         final String damage = sa.getParam("NumDmg");
         int dmg = AbilityUtils.calculateAmount(source, damage, sa);
 
-        if (damage.equals("X") || sourceName.equals("Crater's Claws")) {
-            if (sa.getSVar(damage).equals("Count$xPaid") || sourceName.equals("Crater's Claws")) {
+        if (damage.equals("X") || source.getSVar("X").equals("Count$xPaid") || sourceName.equals("Crater's Claws")) {
```
tool4ever, line 98: I don't understand, Fireball should have already passed the old checks?

Agetian: It looks like the card (as well as other similar cards) were updated, so the "damage" parameter no longer contained what was expected (e.g. NumDmg was pointing to Y, not to X, and X still had Count$xPaid). I'm not sure when it was rewritten and what else may have been rewritten (and to what extent), but at least for these two cards, the checks are now different.

tool4ever: I see, I was only looking at Fireball
for Shatterskull Smashing, it has always been that way

PR #5986, opened 2024-08-24
Quick-pickup for gold-only/shard-only rewards

misha-colbourne:
This is a QoL change that makes it so picking up a gold-only/shard-only reward doesn't take you to the reward scene any more, but rather grants it immediately to avoid interrupting play as much.

It also utilizes accounting labels in the HUD (which are visible albeit translucent in dungeons/towns/etc by this change https://github.com/Card-Forge/forge/pull/5970) to display the increased values accordingly. 

See the video in this discord post for an example: 
https://discord.com/channels/267367946135928833/1084900654616682597/1276372484773511208


kevlahnota:
What happens if the user has a lot of gold or shards? Does the animated text will be out of bounds? I think maybe the animated label should appear above the pickup item for a few seconds...

misha-colbourne:
> What happens if the user has a lot of gold or shards? Does the animated text will be out of bounds? I think maybe the animated label should appear above the pickup item for a few seconds...

The current implementation utilises the accounting labels, which essentially replace text in a singular label to show the increase for a time before showing the new total. Not seeing the number after the +/- in full in the case of large gold/shard totals isn't too big a deal, as the total will update shortly after anyhow.

As such I think this current implementation should still go in as a nice improvement for the majority of cases, but allow to also propose addition idea which I may try in the coming days: As placing the +/- text above/below as you suggest probably would overlap other elements too much, I think a label that can appear above the player's worldmap character when they pick up the gold/shards would make for a nice way to see how much they picked up - whilst also being away from screen edges so should be fully readable regardless of the device being played on.

kevlahnota:
> > What happens if the user has a lot of gold or shards? Does the animated text will be out of bounds? I think maybe the animated label should appear above the pickup item for a few seconds...
> 
> The current implementation utilises the existing "accounting labels", which essentially replace text in a singular label to show the increase for a time before showing the new total. Not seeing the number after the +/- in full in the case of large gold/shard totals isn't too big a deal, as the total will update shortly after anyhow.
> 
> As such I think this current implementation should still go in as a nice improvement for the majority of cases, but allow to also propose addition idea which I may try in the coming days: As placing the +/- text above/below as you suggest probably would overlap other elements too much, I think a label that can appear above the player's worldmap character when they pick up the gold/shards would make for a nice way to see how much they picked up - whilst also being away from screen edges so should be fully readable regardless of the device being played on.

Just don't animate the shard and coin on the game hud, I have a better implementation in mind to display the loot pickups.

misha-colbourne:
> Just don't animate the shard and coin on the game hud, I have a better implementation in mind to display the loot pickups.

Why not have both? The accounting label may as well stay in addition to your idea, surely? Can make it toggleable in options if you like

PR #5988, opened 2024-08-25
CopyPermanentEffect: use DefinedName for special copies

### LINKED ISSUE #5416: Creating Named Token by Card Name
> 111.11. If an effect instructs a player to create a token by name, **doesn’t define any other characteristics for that token**, and the name is not one of the types in the list of predefined tokens above, that player uses the card with that name in the Oracle card reference to determine the characteristics of that token.
> Example: Disa the Restless has the ability “Whenever one or more creatures you control deal combat damage to a player, create a Tarmogoyf token.” As that ability resolves, its controller creates a token with the same characteristics as the card named Tarmogoyf, as determined by the Oracle card reference.

For now, we could keep our Token Scripts, but that is the difference between:
* Create a Tarmagoyf token. (has mana cost)
* Create a 1/1 white Kithkin Soldier creature token named Goldmeadow Harrier. It has "{W}, Tap: Tap target creature." (has no mana cost)

Also, Tarmagoyf token should not be affected by Card Text Changes. So they should be marked with `LockTokenScript`

Hanmac:
Closes #5416

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/CopyPermanentEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/CopyPermanentEffect.java
@@ -13,7 +13,9 @@ import com.google.common.base.Predicates;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 
+import forge.ImageKeys;
 import forge.StaticData;
+import forge.card.CardRarity;
 import forge.card.CardRulesPredicates;
 import forge.card.CardStateName;
 import forge.game.Game;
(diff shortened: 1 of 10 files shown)


forge-game/src/main/java/forge/game/ability/effects/CopyPermanentEffect.java
```diff
@@ -297,6 +306,13 @@ public static Card getProtoType(final SpellAbility sa, final Card original, fina
         } else {
             copy.setState(copy.getCurrentStateName(), true, true);
         }
+        if (sa.hasParam("DefinedName")) {
```
tool4ever, line 309: in this case you can just do `copy = original` above instead of another CardFactory?

Hanmac: like this?

tool4ever (dismissed):

tool4ever (approved):

PR #5991, opened 2024-08-26
Translation support for functional variants

### LINKED ISSUE #5967: card data not translation
image

image
Regarding "Garbage Elemental", the card names in res/languages/cardnames*.txt cannot distinguish the following situations
image

Jetz72:
Fixes #5967.

Seems to work as best I can tell, though it'd be nice to have eyes on it from someone experienced in using Forge under a translation. Little tricky to playtest it myself when I can't read most of the text.

Syntax for these is: 
```
Garbage Elemental $A|Garbage Elemental|Creature — Elemental|Frenzy 2...
Garbage Elemental $B|Garbage Elemental|Creature — Elemental|When Garbage Elemental...
```


forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
```diff
@@ -82,8 +82,7 @@ public final String getStackDescriptionWithSubs(final Map<String, String> params
                 if (params.containsKey("SpellDescription")) {
                     if (rawSDesc.contains(",,,,,,")) rawSDesc = rawSDesc.replaceAll(",,,,,,", " ");
                     if (rawSDesc.contains(",,,")) rawSDesc = rawSDesc.replaceAll(",,,", " ");
-                    String spellDesc = CardTranslation.translateSingleDescriptionText(rawSDesc,
-                            sa.getHostCard().getName());
+                    String spellDesc = CardTranslation.translateSingleDescriptionText(rawSDesc, sa.getHostCard());
```
tool4ever, line 85: this doesn't feel right:
`Card.getName()` has additional logic which you now seem to bypass for the plain CardState value? :dizzy_face:

tool4ever: oh nevermind, I think I misread that part

Jetz72: I saw that `Card.getName()` has logic for applying changed names, but I ended up getting the name to use for translations - the "translation key" as I've been calling it - from the card state, which I believe does step around the `ChangedCardName` logic. I'm not sure if that's the right call. The thing that seemed weird about using `Card.getName()` is that changing the name would change the oracle text that the translator supplies. Is that intended? If not I could rework it.

tool4ever (approved):

Issue #3307, opened 2023-06-21
Remove Hidden Keywords Meta Issue


Hanmac:
This is a Meta Issue for all the others refering Hidden Keywords that needs to be refactored, either as ReplacementEffect=CantHappen or as Static Ability that doesn't use Layer

* IgnoreLandfall #2507 
* `CARDNAME can't be regenerated` #3304 
* Block as If Shadow #2550
* Can't Untap #984
* `CARDNAME's power and toughness are switched` , this probably can be done as flags instead of keyword? should also use Layer?
*  `All creatures able to block CARDNAME do so.` & `CARDNAME must be blocked if able.`
* `CARDNAME can't be countered.`  #2688
* `CARDNAME assigns no combat damage`
* `Colorless Damage Source` 
* `CARDNAME can't attack or block.`
* `CARDNAME's activated abilities can't be activated.`

Both `CARDNAME can't attack or block.` and  `CARDNAME's activated abilities can't be activated.` are tricky because of
`Volrath's Curse` and `Lost in Thought` having `ignore this effect` on them, so they either needs to be one static ability, or some trickery

Hanmac:
currently `Heart of Kiran` isn't directly a hidden keyword, but its ability still should be reworked into a better way to replace the cost

see the other cost replacer:

* "CyclingForZero" (Activation)
* "You may pay 0 rather than pay equip costs." (Activation)
* "You may pay 0 rather than pay the echo cost for permanents you control." (Effect)

tool4ever:
After all are finished the logic for `isStackingKeyword` can also be changed to use Keyword class information instead of ugly `NonStackingKWList.txt`

Hanmac:
@tool4ever i already started this with `KeywordInterface.redundant`

also related to CardTextChanges, i want the keyword teach about what part of the Keyword should be affected by Text Changes, so it would simplify this issue: #2512


If you want you can try to update the CantUntap stuff again (using CantHappen Replacement?)

tool4ever:
about IgnoreEffectCost$:
the other 2 cards that use it to affect Players wouldn't really need to be done as Continuous either
- so the logic for it could be moved to work on generic StaticAbility instead
- to avoid some redundant code the CantSearchLibrary stuff should also be Static instead of RE
- this would allow cleaning up adding statics on Players (only used by _Damping Engine_ it seems)
  - to make this work we might consider supporting more than one mode together, e.g. `Mode$ CantPlayLand, CantBeCast`

[...]

PR #5998, opened 2024-08-27
StaticAbilityAssignNoCombatDamage: Static and first batch

### LINKED ISSUE #5992: Assigns no combat damage as StaticAbility
Change `CARDNAME assigns no combat damage` as StaticAbility to be used with Effect, in most cases, it doesn't even need Remember but can use EffectSource

- [ ] Create new StaticAbility
- [ ] Update existing cards
- [ ] maybe tweak AI to not use Pump Effects on cards that are affected by `assigns no combat damage`

Hanmac:
Closes #5992 

Static Ability and first batch of cards, others will be added later to this MR

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
@@ -898,6 +898,9 @@ public abstract class SpellAbilityEffect {
         } else if ("UntilHostLeavesPlayOrEOT".equals(duration)) {
             host.addLeavesPlayCommand(until);
             game.getEndOfTurn().addUntil(until);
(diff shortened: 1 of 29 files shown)


tool4ever:
It might be interesting to check that it's not worse for performance on average case?
(now needs to iterate all statics for matching one each time the combat power gets used
vs.
refreshing KW but only if one of the few relevant cards become involved)

Hanmac:
@tool4ever the difference is the rules ... technically hidden keywords shouldn't exist, and they cause more problems that it is worth

The main issue is StaticAbilityContinuous + HiddenKeywords because these are updated each time the Layers are updated.
And Normal Pump Abilities should probably use Layers too

See #3307 for more issues

Hanmac:
@tool4ever i updated the remaining cards

tool4ever (approved):

Issue #5548, opened 2024-07-09
Update forge.command according to forge.sh


Hanmac:
The macOS specific `forge.command` file needs to be updated, like the forge.sh one

Hanmac:
@tehdiplomat can you look at this if you can update `forge.command` in a similar way than `forge.sh` ?

tehdiplomat:
I dont have a mac, so not really sure how i could even test it.

PR #6008, opened 2024-08-29
respect JAVA_HOME env in linux/mac launch scripts

janAkali:
Forge should use java version set by JAVA_HOME first and if it's not set - only then use plain "java" invocation.

Also fixed warnings reported by `shellcheck` utility:
- [SC2164](https://www.shellcheck.net/wiki/SC2164) "`cd` command can fail.." => "cd ... || exit"
- [SC2155](https://www.shellcheck.net/wiki/SC2155) "`local foo=...`" => "local foo; foo="
- [SC2086](https://www.shellcheck.net/wiki/SC2086) added double quotes where word-splitting is not needed


Hanmac:
do you have a mac to test the scripts?

can you update the forge.command script too? #5548
it still misses the version checks from the forge.sh

Archargelod:
> do you have a mac to test the scripts?

No, I don't have a mac to test scripts, but the changes I made are POSIX-compliant. It should work the same with every default shell on any OS (dash, bash, zsh, etc..).

> can you update the forge.command
AFAIK there's no difference between ".sh" and ".command" file extensions in MacOS. (see https://apple.stackexchange.com/questions/24262/what-is-the-difference-between-command-tool-and-sh-file-extensions)
Having to support two identical launch scripts seems unnecessary. Maybe it'd be better to remove them? (question is addressed to maintainers)

Hanmac:
> > do you have a mac to test the scripts?
> 
> No, I don't have a mac to test scripts, but the changes I made are POSIX-compliant. It should work the same with every default shell on any OS (dash, bash, zsh, etc..).
> 
> > can you update the forge.command
> 
> AFAIK there's no difference between ".sh" and ".command" file extensions in MacOS. (see https://apple.stackexchange.com/questions/24262/what-is-the-difference-between-command-tool-and-sh-file-extensions) Having to support two identical launch scripts seems unnecessary. Maybe it'd be better to remove them? (question is addressed to maintainers)

I think MacOS reacts differently with the ".command" file extension.

so either copy the .sh to the .command file extension too, or make the .command call the .sh if able

Archargelod:
Oh I missed that you was talking about `forge.command` that launches the desktop client. In my previous message I meant that we can remove two ".command" scripts: [forge-gui-mobile-dev/src/main/config/forge-adventure.command](https://github.com/Card-Forge/forge/blob/239e327f0ba8587b1666fefba0a5a40fa78e47f4/forge-gui-mobile-dev/src/main/config/forge-adventure.command#L2) and [forge-adventure/src/main/config/forge-adventure-editor.command](https://github.com/Card-Forge/forge/blob/239e327f0ba8587b1666fefba0a5a40fa78e47f4/forge-adventure/src/main/config/forge-adventure-editor.command#L2) because they have "*_mac.sh" files beside them that are more up to date.

But with "forge.command" it seems like it's possible to copy "forge.sh" script with minimal changes. First, the she-bang:
```
#!/usr/bin/zsh
```
I've tested this script both in bash and zsh (on Linux machine) and there's only one line that would work differently in zsh:
```
elif [[ -n $(type -p java) ]]
```

`type` in zsh returns a string "foo not found" even if it can't find the path, but we can fix it to check the exit code instead:
```
elif (type -p java)
```
or replace it with `whence` that's also builtin and doesn't return anything on fail:
```
elif [[ -n $(whence java) ]]
```

Hanmac:
@tehdiplomat @Agetian @kevlahnota can someone check this MR?

kevlahnota:
@Hanmac I think it's fine however once my PR is merged along with Java 17 update, the -illegal-access permit shouldn't work and should be removed and we need to use add opens paramater to cleanup the code.

Agetian:
Agreed with @kevlahnota :+1:

jamincollins:
## Findings

### Logic correctness

#### 1. Unquoted java command invocation breaks when JAVA_HOME contains spaces
The sh launch scripts construct `java="${JAVA_HOME:+$JAVA_HOME/bin/}java"` and invoke `$java` without quotes. If JAVA_HOME contains spaces, word splitting occurs and the command fails with command not found.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("/bin/bash", "-c", "JAVA_HOME=\"/opt/my java\"; java=\"${JAVA_HOME:+$JAVA_HOME/bin/}java\"; which $java || echo \"FAIL\"");
pb.inheritIO().start().waitFor();
```
Impact: Users installing JDK under paths with spaces cannot launch Forge; script fails silently.

#### 2. jdk_version arithmetic comparison crashes when no java is available
`forge.sh` sets `v="$(jdk_version)"` where `jdk_version` returns `no_java` if `java_cmd` is empty. Later `[[ $v -ge 17 ]]` performs integer comparison on a non-numeric string, causing bash to error `integer expression expected` and abort the script.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("/bin/bash", "-c", "unset JAVA_HOME; PATH=\"\"; v=\"no_java\"; [[ $v -ge 17 ]] && echo ok || echo \"FAIL\"");
pb.inheritIO().start().waitFor();
```
Impact: Systems without java on PATH and without JAVA_HOME cause hard failure instead of graceful error.

#### 3. sh scripts use JAVA_HOME without existence check
`forge-adventure-editor.sh`, `forge-adventure-editor-mac.sh`, `forge-adventure.sh`, `forge-adventure-mac.sh` expand to `java="${JAVA_HOME:+$JAVA_HOME/bin/}java"` unconditionally. No `-x` test is performed. If JAVA_HOME is set to a non-existent directory, the script attempts to execute a non-existent binary and fails.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("/bin/bash", "-c", "JAVA_HOME=\"/nonexistent\"; java=\"${JAVA_HOME:+$JAVA_HOME/bin/}java\"; $java -version 2>&1 || echo \"FAIL\"");
pb.inheritIO().start().waitFor();
```
Impact: Setting JAVA_HOME to an invalid location prevents fallback to PATH java, breaking launches.

#### 4. cd $(dirname "${0}") is unquoted in sh scripts
The sh scripts use `cd $(dirname "${0}")` without quotes. If the script resides in a path containing spaces, dirname output is word-split and cd fails, leaving the working directory unchanged and causing jar path resolution to fail.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("/bin/bash", "-c", "mkdir -p '/tmp/a b'; touch '/tmp/a b/forge.sh'; cd $(dirname '/tmp/a b/forge.sh'); pwd");
pb.inheritIO().start().waitFor();
```
Impact: Installation under paths with spaces results in wrong cwd and launch failure.

#### 5. JAVA_HOME precedence change is a breaking behavioural change
`forge.sh` previously preferred `type -p java` over JAVA_HOME. The PR reverses the order to check JAVA_HOME first. Users with JAVA_HOME pointing to an older JDK while PATH contains a newer JDK will now run the older JDK, potentially breaking features requiring newer Java.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("/bin/bash", "-c", "export JAVA_HOME=/opt/jdk11; export PATH=/opt/jdk17/bin:$PATH; java_cmd=\"\"; [[ -n \"$JAVA_HOME\" && -x \"$JAVA_HOME/bin/java\" ]] && java_cmd=\"$JAVA_HOME/bin/java\" || java_cmd=\"java\"; $java_cmd -version");
pb.inheritIO().start().waitFor();
```
Impact: Silent downgrade of java version for existing installations; backwards incompatible.

#### 6. cd || exit added only to forge.sh, not sh scripts
`forge.sh` now uses `cd "$(dirname "${0}")" || exit`, improving failure handling. The sh scripts retain `cd $(dirname "${0}")` without error handling, so failures are silent and subsequent commands run in the wrong directory.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("/bin/bash", "-c", "cd /nonexistent/path; echo $?");
pb.inheritIO().start().waitFor();
```
Impact: Inconsistent error handling across launch scripts; partial failures harder to diagnose.

### Performance and security

#### 7. jdk_version spawns java process on every launch
`jdk_version` invokes `$java_cmd -Xms32M -Xmx32M -version` each time the script runs, adding ~100-300ms startup latency. This is acceptable for desktop launch but is unnecessary for scripts that never inspect version.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("java", "-version");
pb.start().waitFor();
long dur = System.nanoTime() - start;
System.out.println(dur / 1_000_000);
```
Impact: Increased cold start time; no security issue but performance regression vs direct invocation.

## Verdict
Request Changes

The PR correctly implements JAVA_HOME preference and fixes shellcheck warnings, but the implementation introduces word-splitting vulnerabilities, missing existence checks in sh scripts, a breaking precedence change, and an unhandled no-java arithmetic error. These issues are reachable in real environments with spaces in paths or misconfigured JAVA_HOME.

## Recommended Fixes
1. Quote all command invocations: use `"$java"` and `"$java_cmd"` everywhere, and quote `cd "$(dirname "${0}")"`.
2. Add existence check in sh scripts before using JAVA_HOME, falling back to PATH java if `$JAVA_HOME/bin/java` is not executable.
3. Guard jdk_version result before integer comparison: if `v` is not numeric, exit with a clear error message instead of bash arithmetic failure.
4. Document the precedence change as breaking and consider a compatibility flag.
5. Apply `cd ... || exit` and quoted dirname to all launch scripts for consistency.
---

jamincollins:
## Findings

### Logic correctness

#### 1. Unquoted java command invocation breaks when JAVA_HOME contains spaces
The sh launch scripts construct `java="${JAVA_HOME:+$JAVA_HOME/bin/}java"` and invoke `$java` without quotes. If JAVA_HOME contains spaces, word splitting occurs and the command fails with command not found.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("/bin/bash", "-c", "JAVA_HOME=\"/opt/my java\"; java=\"${JAVA_HOME:+$JAVA_HOME/bin/}java\"; which $java || echo \"FAIL\"");
pb.inheritIO().start().waitFor();
```
Impact: Users installing JDK under paths with spaces cannot launch Forge; script fails silently.

#### 2. jdk_version arithmetic comparison crashes when no java is available
`forge.sh` sets `v="$(jdk_version)"` where `jdk_version` returns `no_java` if `java_cmd` is empty. Later `[[ $v -ge 17 ]]` performs integer comparison on a non-numeric string, causing bash to error `integer expression expected` and abort the script.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("/bin/bash", "-c", "unset JAVA_HOME; PATH=\"\"; v=\"no_java\"; [[ $v -ge 17 ]] && echo ok || echo \"FAIL\"");
pb.inheritIO().start().waitFor();
```
Impact: Systems without java on PATH and without JAVA_HOME cause hard failure instead of graceful error.

#### 3. sh scripts use JAVA_HOME without existence check
`forge-adventure-editor.sh`, `forge-adventure-editor-mac.sh`, `forge-adventure.sh`, `forge-adventure-mac.sh` expand to `java="${JAVA_HOME:+$JAVA_HOME/bin/}java"` unconditionally. No `-x` test is performed. If JAVA_HOME is set to a non-existent directory, the script attempts to execute a non-existent binary and fails.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("/bin/bash", "-c", "JAVA_HOME=\"/nonexistent\"; java=\"${JAVA_HOME:+$JAVA_HOME/bin/}java\"; $java -version 2>&1 || echo \"FAIL\"");
pb.inheritIO().start().waitFor();
```
Impact: Setting JAVA_HOME to an invalid location prevents fallback to PATH java, breaking launches.

#### 4. cd $(dirname "${0}") is unquoted in sh scripts
The sh scripts use `cd $(dirname "${0}")` without quotes. If the script resides in a path containing spaces, dirname output is word-split and cd fails, leaving the working directory unchanged and causing jar path resolution to fail.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("/bin/bash", "-c", "mkdir -p '/tmp/a b'; touch '/tmp/a b/forge.sh'; cd $(dirname '/tmp/a b/forge.sh'); pwd");
pb.inheritIO().start().waitFor();
```
Impact: Installation under paths with spaces results in wrong cwd and launch failure.

#### 5. JAVA_HOME precedence change is a breaking behavioural change
`forge.sh` previously preferred `type -p java` over JAVA_HOME. The PR reverses the order to check JAVA_HOME first. Users with JAVA_HOME pointing to an older JDK while PATH contains a newer JDK will now run the older JDK, potentially breaking features requiring newer Java.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("/bin/bash", "-c", "export JAVA_HOME=/opt/jdk11; export PATH=/opt/jdk17/bin:$PATH; java_cmd=\"\"; [[ -n \"$JAVA_HOME\" && -x \"$JAVA_HOME/bin/java\" ]] && java_cmd=\"$JAVA_HOME/bin/java\" || java_cmd=\"java\"; $java_cmd -version");
pb.inheritIO().start().waitFor();
```
Impact: Silent downgrade of java version for existing installations; backwards incompatible.

#### 6. cd || exit added only to forge.sh, not sh scripts
`forge.sh` now uses `cd "$(dirname "${0}")" || exit`, improving failure handling. The sh scripts retain `cd $(dirname "${0}")` without error handling, so failures are silent and subsequent commands run in the wrong directory.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("/bin/bash", "-c", "cd /nonexistent/path; echo $?");
pb.inheritIO().start().waitFor();
```
Impact: Inconsistent error handling across launch scripts; partial failures harder to diagnose.

### Performance and security

#### 7. jdk_version spawns java process on every launch
`jdk_version` invokes `$java_cmd -Xms32M -Xmx32M -version` each time the script runs, adding ~100-300ms startup latency. This is acceptable for desktop launch but is unnecessary for scripts that never inspect version.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("java", "-version");
pb.start().waitFor();
long dur = System.nanoTime() - start;
System.out.println(dur / 1_000_000);
```
Impact: Increased cold start time; no security issue but performance regression vs direct invocation.

## Verdict
Request Changes

The PR correctly implements JAVA_HOME preference and fixes shellcheck warnings, but the implementation introduces word-splitting vulnerabilities, missing existence checks in sh scripts, a breaking precedence change, and an unhandled no-java arithmetic error. These issues are reachable in real environments with spaces in paths or misconfigured JAVA_HOME.

## Recommended Fixes
1. Quote all command invocations: use `"$java"` and `"$java_cmd"` everywhere, and quote `cd "$(dirname "${0}")"`.
2. Add existence check in sh scripts before using JAVA_HOME, falling back to PATH java if `$JAVA_HOME/bin/java` is not executable.
3. Guard jdk_version result before integer comparison: if `v` is not numeric, exit with a clear error message instead of bash arithmetic failure.
4. Document the precedence change as breaking and consider a compatibility flag.
5. Apply `cd ... || exit` and quoted dirname to all launch scripts for consistency.
---

PR #6017, opened 2024-08-30
Fix Conspiracy removing types without choice for new one

### LINKED ISSUE #5420: Kindred appears in the wrong order
Since the change from "Tribal" to "Kindred", the ordering of types on kindred cards has been incorrect. Hence, a card that should say "Kindred X" will say "Instant Kindred", "Enchantment Kindred", "Artifact Kindred", etc.

tool4ever:
I also changed it so it avoids adding empty traits in this case.

Closes #5420

### DIFF
--- a/forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java
+++ b/forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java
@@ -422,26 +422,30 @@ public final class StaticAbilityContinuous {
             if (params.containsKey("AddAllCreatureTypes")) {
                 addAllCreatureTypes = true;
             }
-            if (params.containsKey("RemoveSuperTypes")) {
-                remove.add(RemoveType.SuperTypes);
-            }
-            if (params.containsKey("RemoveCardTypes")) {
-                remove.add(RemoveType.CardTypes);
-            }
-            if (params.containsKey("RemoveSubTypes")) {
-                remove.add(RemoveType.SubTypes);
-            }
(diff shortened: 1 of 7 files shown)


forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java
```diff
@@ -422,26 +422,30 @@ public static CardCollectionView applyContinuousAbility(final StaticAbility stAb
...
-                remove.add(RemoveType.CreatureTypes);
-            }
-            if (params.containsKey("RemoveArtifactTypes")) {
-                remove.add(RemoveType.ArtifactTypes);
-            }
-            if (params.containsKey("RemoveEnchantmentTypes")) {
-                remove.add(RemoveType.EnchantmentTypes);
+
+            // overwrite doesn't work without new value (e.g. Conspiracy missing choice)
+            if (addTypes == null || !addTypes.isEmpty()) {
```
Hanmac, line 427: Isn't this wrong? What if something would just remove types without adding them? Wouldn't that break now?

tool4ever: if it would only remove and not overwrite like _Alpine Moon_ addTypes will be null

Hanmac: Hm, I somehow don't like the difference between `addTypes` not used, and `addTypes` being empty 🤔

Hanmac (approved):

PR #6018, opened 2024-08-31
27 cards from 2024 Unknown Events (1 is in mystery booster 2)

Lykrast:
More cards from [Gavin's unknown event](https://mtg.fandom.com/wiki/Unknown_Event), following from #5913 to catch up to 2024. Although gavin hasn't made a video about the gencon event, so haven't included an edition for it yet and I haven't tried any of the 2 cards that are documented.

2 editions (chicago and amsterdam), 27 cards, 5 tokens.

Notes:

* Halving season is in mystery booster 2 too
* The Bear Force Pilot's token has no name on the card, and in the video Gavin says the token is supposed to be a 0/0
* The Duke of Midrange's actual card has 2 printing errors according to Gavin's video, which have been fixed here (the toughness is 1+\* to match the text, and the card us cast for free)
* The Keeper of Dark Pacts uses arena's intensity mechanic to keep track of its effect cause I did not know how to do it otherwise, and it works
* The Magic Bandit says to copy a played land, which nothing in the game currently does, so I went with making a token copy of the played card, which seems to work (even though it'll have inconsistent interactions with token doublers compared to token spells, but I don't think the rules actually cover "copying a land drop")
* The Rebellious Intelligence's reminder text tells you to get any card for its effect outside of draft, but I stuck to sideboard to match wish cards

EDIT: oh just saw that #5416 got done! I haven't managed to get dev forge running locally (these scripts were tested through the custom folder), but I'll probably do another pass of the unknown cards that use that after the next update.

### DIFF
--- a/forge-gui/res/cardsfolder/a/ateoclock.txt
+++ b/forge-gui/res/cardsfolder/a/ateoclock.txt
@@ -2,7 +2,7 @@ Name:Ate-o'-Clock
 ManaCost:2 G
 Types:Instant
 K:Replicate:2 G
-A:SP$ CopyPermanent | Choices$ Creature.token+YouCtrl | NumCopies$ 1 | Populate$ True | SubAbility$ DBProlif | StackDescription$ SpellDescription | SpellDescription$ Populate.
+A:SP$ CopyPermanent | Choices$ Creature.token+YouCtrl | NumCopies$ 1 | Populate$ True | SubAbility$ DBProlif | SpellDescription$ Populate. Proliferate. Investigate. Regenerate target creature.
 SVar:DBProlif:DB$ Proliferate | SubAbility$ DBInvestigate
 SVar:DBInvestigate:DB$ Investigate | SubAbility$ DBRegen
 SVar:DBRegen:DB$ Regenerate | ValidTgts$ Creature | TgtPrompt$ Select target creature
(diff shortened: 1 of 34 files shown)


tool4ever:
Development should better be done with snapshots anyway:
https://downloads.cardforge.org/dailysnapshots/

Lykrast:
> Development should better be done with snapshots anyway: https://downloads.cardforge.org/dailysnapshots/

Oooh right! Thanks forgot about those. Will do that if I get motivated to backfill the affected unknown cards before next release (and sending love to people who will do the duskmourn, don't know if I'll have time to contribute).

tool4ever (dismissed):

forge-gui/res/cardsfolder/a/ateoclock.txt
```diff
@@ -0,0 +1,11 @@
+Name:Ate-o'-Clock
+ManaCost:2 G
+Types:Instant
+K:Replicate:2 G
+A:SP$ CopyPermanent | Choices$ Creature.token+YouCtrl | NumCopies$ 1 | Populate$ True | SubAbility$ DBProlif | StackDescription$ SpellDescription | SpellDescription$ Populate.
+SVar:DBProlif:DB$ Proliferate | SubAbility$ DBInvestigate
```
tool4ever, line 6: SpellDescription incomplete

Lykrast: Should be fixed

tool4ever (approved):

PR #6024, opened 2024-09-01
GameLoss: use CantHappen for cantLose

### LINKED ISSUE #4757: CantLose and CantWin as StaticAbilities/CantHappen
Cases:

- [ ]  "You can't win the game." → StaticAbility
- [ ]  "You can't lose the game." → GameLoss Replacement + `GameLossReason` parameter
  - [ ]  "You don't lose the game for having 0 or less life."
  - [ ]  "You don't lose the game for drawing from an empty library."

Similar to other #3307, "CantLose" and "CantWin" should be via StaticAbilities instead of keywords
CantHappen Replacement would be fine too

Also @tool4ever "Your opponents can't win the game." keyword probably doesn't exist

https://github.com/Card-Forge/forge/blob/511ccd697c7b1d7dea8933715a5b7973be8ea141/forge-gui/res/cardsfolder/c/courageous_resolve.txt#L8

Hanmac:
Closes #4757


Hanmac:
@tool4ever you need to comment on this one

tool4ever:
you need to give me at least 24 hours for bigger stuff lol :p

tool4ever (approved):

forge-game/src/main/java/forge/game/CardTraitBase.java
```diff
@@ -235,6 +236,13 @@ public boolean matchesValid(final Object o, final String[] valids, final Card sr
                     return true;
                 }
             }
+        } else if (o instanceof GameLossReason) {
+            for (String s : valids) {
+                GameLossReason valid = GameLossReason.smartValueOf(s);
+                if (((GameLossReason) o).name().equals(valid.name())) {
+                    return true;
+                }
+            }
```
Hanmac, line 245: @tool4ever I really need to find a better idea for this, to match an Enum against some valid strings

One main problem is, that can't use Abstract code inside enum classes, so i might need to move that to a helper?

tool4ever: `((GameLossReason) o).name().equals(s)` should also work already?

Hanmac: yeah, maybe equals IgnoreCase

my idea was more to find a way to not have all EnumTypes I want to check there

like some kind of:
```
enum GameLossReason implements IValidEnum
{
   public bool isValid(String valids[])
}
```

then in the MatchesValid, i could do:
```
} else if (o instanceof IValidEnum) {
  return ((IValidEnum)o).isValid(valids);
}
```

tool4ever: using Enum as class name should make it work for others?

Issue #3543, opened 2023-07-27
Clean up redundancy of `StaticAbility.checkConditions` into CardTraitBase


tool4ever:
fixes the card for now because `StaticAbility.checkConditions` does not support IsPresent2
we could probably try and merge it into CardTraitBase checks but that is a bigger refactor

_Originally posted by @tool4ever in https://github.com/Card-Forge/forge/pull/3540#discussion_r1276188781_

PR #6037, opened 2024-09-02
Intervening if triggers with tapped/untapped state

Fulgur14:


Hanmac:
@tool4ever i don't like the use of CheckSvar there 

do we have a better way for this?

Hanmac:
@Fulgur14 I don't know if you can debug that right now, but can you check what the difference between `IsPresent` vs `CheckSVar` is from debugging?

Fulgur14:
Probably not at the moment.

tool4ever:
IsPresent$ means "if this card is in the specified zone..."

That said, most like _Bottomless Vault_ can actually be left alone anyway, since putting a storage counter on them wouldn't make sense after they LTB...

Fulgur14:
I know, but I thought that it would be better to make it consistent so the code could be re-used safely.

tool4ever:
Well it is faster from a programming view, since otherwise it checks all the cards in the zone 👍

Hanmac:
> IsPresent$ means "if this card is in the specified zone..."

can you debug why the tap check didn't work for IsPresent while `Count$ValidSelf Card.untapped` is better?

Found the reason, is uses special Host for `ValidSelf`:
https://github.com/Card-Forge/forge/blob/19d8c7bac6f63a111f4844c007d1a527ecfc59e8/forge-game/src/main/java/forge/game/ability/AbilityUtils.java#L1950-L1952

@tool4ever the problem is that `IsPresent` there doesn't have `PresentDefined` like done for Restrictions or `ConditionDefined` for Conditions

If it had, then imo better would be this?
`IsPresent$ Card.untapped | PresentDefined$ Self`

tool4ever:
Somewhat related to #3543
might be done later imo

tool4ever:
@Fulgur14 
these can now be updated, thanks

forge-gui/res/cardsfolder/v/voodoo_doll.txt
```diff
@@ -5,8 +5,8 @@ T:Mode$ Phase | Phase$ Upkeep | ValidPlayer$ You | Execute$ TrigPutCounter | Tri
 T:Mode$ Phase | Phase$ End of Turn | IsPresent$ Card.Self+untapped | ValidPlayer$ You | Execute$ TrigDestroy | TriggerZones$ Battlefield | TriggerDescription$ At the beginning of your end step, if CARDNAME is untapped, destroy CARDNAME and it deals damage to you equal to the number of pin counters on it.
 A:AB$ DealDamage | Cost$ X X T | ValidTgts$ Any | NumDmg$ X | SpellDescription$ CARDNAME deals damage equal to the number of pin counters on it to any target. X is the number of pin counters on CARDNAME.
 SVar:TrigPutCounter:DB$ PutCounter | CounterType$ PIN | CounterNum$ 1 | Defined$ Self
-SVar:TrigDestroy:DB$ Destroy | Defined$ Self | SubAbility$ DBDealDamageYou | ConditionDefined$ Self | ConditionPresent$ Card.untapped
-SVar:DBDealDamageYou:DB$ DealDamage | Defined$ You | NumDmg$ X | ConditionDefined$ Self | ConditionPresent$ Card.untapped
+SVar:TrigDestroy:DB$ Destroy | Defined$ Self | SubAbility$ DBDealDamageYou | PresentDefined$ Self | IsPresent$ Card.untapped
```
tool4ever, line 8: this script should not have been changed

forge-gui/res/cardsfolder/n/nim_abomination.txt
```diff
@@ -2,6 +2,6 @@ Name:Nim Abomination
 ManaCost:2 B
 Types:Creature Zombie
 PT:3/4
-T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You | TriggerZones$ Battlefield | IsPresent$ Card.Self+untapped | Execute$ TrigLoseLife | TriggerDescription$ At the beginning of your end step, if CARDNAME is untapped, you lose 3 life.
+T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You | TriggerZones$ Battlefield | PresentDefined$ Self | IsPresent$ Card.untapped | Execute$ TrigLoseLife | TriggerDescription$ At the beginning of your end step, if CARDNAME is untapped, you lose 3 life.
 SVar:TrigLoseLife:DB$ LoseLife | Defined$ You | LifeAmount$ 3 | ConditionDefined$ Self | ConditionPresent$ Card.untapped
```
tool4ever, line 6: condition check here no longer needed

forge-gui/res/cardsfolder/m/mana_vault.txt
```diff
@@ -4,7 +4,7 @@ Types:Artifact
 A:AB$ Mana | Cost$ T | Produced$ C | Amount$ 3 | SpellDescription$ Add {C}{C}{C}.
 K:CARDNAME doesn't untap during your untap step.
 T:Mode$ Phase | Phase$ Upkeep | ValidPlayer$ You | TriggerZones$ Battlefield | OptionalDecider$ You | Execute$ TrigUntap | TriggerDescription$ At the beginning of your upkeep, you may pay {4}. If you do, untap CARDNAME.
-T:Mode$ Phase | Phase$ Draw | ValidPlayer$ You | IsPresent$ Card.Self+tapped | Execute$ TrigDamage | TriggerDescription$ At the beginning of your draw step, if CARDNAME is tapped, it deals 1 damage to you.
+T:Mode$ Phase | Phase$ Draw | ValidPlayer$ You | PresentDefined$ Self | IsPresent$ Card.tapped | Execute$ TrigDamage | TriggerDescription$ At the beginning of your draw step, if CARDNAME is tapped, it deals 1 damage to you.
 SVar:TrigUntap:AB$ Untap | Cost$ 4 | Defined$ Self
 SVar:TrigDamage:DB$ DealDamage | Defined$ You | NumDmg$ 1 | ConditionDefined$ Self | ConditionPresent$ Card.tapped
```
tool4ever, line 9: condition check here no longer needed

tool4ever (approved):

PR #6044, opened 2024-09-03
Room: First Spell Part

### LINKED ISSUE #6033: DSK: Room Mechanic
* Create a new SplitMode like Split (CombinedCMC)
* Have the Default Face be combined
* When it is cast by their SplitFace they enter as that face. (entering would cause them to unlock this door)
* Special Action to unlock the other face (Setting State to Default?)
  > Spent this Mana only to cast Room Spells and Unlock Doors 
* when it isn't cast, it enters with all rooms locked.
  > The permanent will be an object on the battlefield but with no name, mana value, or abilities. It's still a Room enchantment.
* The Rooms should not be copyable. LeftRoomUnlocked/RightRoomUnlocked should be Flags on Card. (See Setting Correct CardState for Flipped)
  This might not be enough. The Copy would need to copy the other states too.
* When a Door is unlocked, it triggers "When you unlock this Door"
  * This also needs to trigger when a Room is enters when cast
* need CalculateAmount for "unlocked Doors among Rooms you control" also for "Different Door Names"
* When all Doors are unlocked (via Set State to Default?) it triggers "Fully unlocked a Room" (see Eerie Ability Word)
  * we could add such a Trigger first to make Eerie cards scriptable even without Room cards

I don't know yet if it should continue to use SplitMode or if we should create a New Mode.

Hanmac:
Closes #6033

- [x] Split Parts can be cast, and they enter the battlefield
- [x] Visual they are off right now, but the ability work (static and trigger)
- [x] Special all-locked State Missing
- [x] Unlock Door Special Action Missing
- [x] Unlock Door and Fully Unlock Room Trigger Missing
   - [x] `Unlock this Door` should look for the CardState property of the Special Action

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/UnlockDoorEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/UnlockDoorEffect.java
@@ -2,8 +2,8 @@ package forge.game.ability.effects;
 
 import java.util.Map;
 
-import org.testng.collections.Lists;
-import org.testng.collections.Maps;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 
 import forge.StaticData;
 import forge.card.CardStateName;
--- a/forge-game/src/main/java/forge/game/ability/effects/UnlockDoorEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/UnlockDoorEffect.java
@@ -1,0 +1,76 @@
+package forge.game.ability.effects;
+
+import java.util.Map;
+
+import org.testng.collections.Lists;
+import org.testng.collections.Maps;
+
+import forge.StaticData;
+import forge.card.CardStateName;
+import forge.card.ICardFace;
+import forge.game.Game;
+import forge.game.ability.SpellAbilityEffect;
+import forge.game.card.Card;
+import forge.game.card.CardCollection;
+import forge.game.card.CardLists;
+import forge.game.player.Player;
+import forge.game.spellability.SpellAbility;
+import forge.game.zone.ZoneType;
+import forge.util.Localizer;
+
+public class UnlockDoorEffect extends SpellAbilityEffect {
+
+    @Override
+    public void resolve(SpellAbility sa) {
+        final Card source = sa.getHostCard();
+        final Game game = source.getGame();
+        final Player activator = sa.getActivatingPlayer();
+        
+        CardCollection list;
+        
+        if (sa.hasParam("Choices")) {
+            Player chooser = activator;
+            String title = sa.hasParam("ChoiceTitle") ? sa.getParam("ChoiceTitle") : Localizer.getInstance().getMessage("lblChoose") + " ";
+
+            CardCollection choices = CardLists.getValidCards(game.getCardsIn(ZoneType.Battlefield), sa.getParam("Choices"), activator, source, sa);
+
+            Card c = chooser.getController().chooseSingleEntityForEffect(choices, sa, title, Maps.newHashMap());
+            if (c == null) {
+                return;
+            }
+            list = new CardCollection(c);
+        } else {
+            list = getTargetCards(sa);
+        }
+        
+        for (Card c : list) {
+            switch (sa.getParamOrDefault("Mode", "ThisDoor")) {          
+            case "ThisDoor":
+                c.unlockRoom(activator, sa.getCardStateName());
+                break;
+            case "Unlock":
+                //List<ICardFace> faces = c.getLockedRoomNames().stream().map(face -> StaticData.instance().getCommonCards().getFaceByName(face)).collect(Collectors.toList());
+
+                Map<ICardFace, CardStateName> map = Maps.newHashMap();
+                
+                for (CardStateName faceStateName : c.getLockedRooms()) {
+                    if (!c.hasState(faceStateName)) {
+                        continue;
+                    }
+                    String faceName = c.getState(faceStateName).getName();
+                    ICardFace face = StaticData.instance().getCommonCards().getFaceByName(faceName);
+                    map.put(face, faceStateName);
+                }
+
+                // need to choose Room Name
+                ICardFace chosen = activator.getController().chooseSingleCardFace(sa, Lists.newArrayList(map.keySet()), "Choose Room to unlock");
+                if (chosen == null) {
+                    continue;
+                }
+                c.unlockRoom(activator, map.get(chosen));
+                break;
+            }
+        }
+    }
+
+}
--- a/forge-game/src/main/java/forge/game/ability/effects/UnlockDoorEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/UnlockDoorEffect.java
@@ -1,18 +1,16 @@
 package forge.game.ability.effects;
 
-import java.util.Map;
+import java.util.List;
+import java.util.stream.Collectors;
 
-import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 
-import forge.StaticData;
-import forge.card.CardStateName;
-import forge.card.ICardFace;
 import forge.game.Game;
 import forge.game.ability.SpellAbilityEffect;
 import forge.game.card.Card;
 import forge.game.card.CardCollection;
 import forge.game.card.CardLists;
+import forge.game.card.CardState;
 import forge.game.player.Player;
 import forge.game.spellability.SpellAbility;
 import forge.game.zone.ZoneType;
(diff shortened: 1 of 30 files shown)


Hanmac:
I will add Unlock Door Param/Action for SetState.

With this, I can better control into which state it should change:

- EmptyRoom → LeftSide → DefaultSide (both)
- EmptyRoom → RightSide → DefaultSide (both)

tool4ever:
One thing that might be different is that old Split cards could have different types.
But now it seems to be shared?
So even in empty state it still needs to be a Room Enchantment? 🤔

Hanmac:
> One thing that might be different is that old Split cards could have different types. But now it seems to be shared? So even in empty state it still needs to be a Room Enchantment? 🤔

The empty state will be room specific, so it will not collide with old split cards

Hanmac:
@tool4ever added Unlock Action and Unlock Trigger (might add more info like GameLog Message?)

I haven't tested it yet with clone Effects.


@kevlahnota really need to help me to fix the visual on the battlefield

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -7468,6 +7479,15 @@ public List<SpellAbility> getAllPossibleAbilities(final Player player, final boo
             }
         }
 
+        if (isInPlay() && isRoom()) {
```
tool4ever, line 7482: SorcerySpeed + PhasedOut checks missing

Hanmac:
@tool4ever can you help me to wire it what should happen if a something becomes a copy of a Room, or a Room becomes a copy of another Room?

tool4ever:
Need to fix this crash when copying permanent:
```
EDT > java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at forge.card.CardDb$CardRequest.fromString(CardDb.java:203)
	at forge.card.CardDb.getCard(CardDb.java:541)
	at forge.view.arcane.CardPanel.displayIconOverlay(CardPanel.java:487)
	at forge.view.arcane.CardPanel.paintChildren(CardPanel.java:368)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1128)
	at forge.view.arcane.CardPanel.paint(CardPanel.java:277)
```

forge-game/src/main/java/forge/game/trigger/TriggerUnlockDoor.java
```diff
@@ -0,0 +1,50 @@
...
+            return false;
+        }
+
+        if (!matchesValidParam("ValidPlayer", runParams.get(AbilityKey.Player))) {
+            return false;
+        }
+
+        if (hasParam("ThisDoor")) {
+            CardState state = (CardState) runParams.get(AbilityKey.CardState);
+            if (!getCardState().equals(state)) {
```
tool4ever, line 29: @Hanmac
this logic needs to be fixed somehow
right now it fails for copied spells because then the trigger uses the original CardState

Hanmac: What's your opinion? Just compare the CardStateName? (after ValidCard compared against `Card.Self` ?)

tool4ever: seems safer 🤔

Hanmac: is this good?

tool4ever: works now

Hanmac:
@tool4ever @Agetian @tehdiplomat  i need some better logic for "Choose Door to unlock"

my function needs the CardStateName, but the GUI function might want the ICardFace name?

forge-game/src/main/java/forge/game/ability/effects/UnlockDoorEffect.java
```diff
@@ -0,0 +1,76 @@
...
+
+                Map<ICardFace, CardStateName> map = Maps.newHashMap();
+                
+                for (CardStateName faceStateName : c.getLockedRooms()) {
+                    if (!c.hasState(faceStateName)) {
+                        continue;
+                    }
+                    String faceName = c.getState(faceStateName).getName();
+                    ICardFace face = StaticData.instance().getCommonCards().getFaceByName(faceName);
+                    map.put(face, faceStateName);
```
tool4ever, line 62: It looks like at least on Desktop `GuiChoose` should accept a list of CardStateView just fine...
might be worth a shot to see how it'd display, then doing it without the Map seems possible 🤔

tool4ever:
I checked out the card rendering and the way split states get handled makes drawing something meant for both a bit tricky (this could also work with Fuse)
I think the best I could do for now is skip right for the "shared type line" if you think that helps?

forge-game/src/main/java/forge/game/ability/effects/UnlockDoorEffect.java
Hanmac: we need to be careful with that, because for Network, the object needs to be Serializeable

if you can get CardStateView working, then you could try this too

tool4ever: it does extend TrackableObject so like all View classes I except it can be serialized

forge-gui/res/cardsfolder/upcoming/ghostly_keybearer.txt
```diff
@@ -0,0 +1,8 @@
+Name:Ghostly Keybearer
+ManaCost:3 U
+Types:Creature Spirit
+PT:3/3
+K:Flying
+T:Mode$ DamageDone | ValidSource$ Card.Self | ValidTarget$ Player | Execute$ TrigUnlock | CombatDamage$ True | TriggerDescription$ Whenever CARDNAME deals combat damage to a player, unlock a locked door of up to one target Room you control.
```
tool4ever, line 6: @Hanmac
I tried to unlock _Copy Enchantment_ Room but it fails from `getLockedRooms` check 🤔

forge-game/src/main/java/forge/game/card/CardView.java
```diff
@@ -40,6 +41,14 @@ public static CardStateView getState(Card c, CardStateName state) {
         return s == null ? null : s.getView();
     }
 
+    public static Map<CardStateView, CardState> getStateMap(Iterable<CardState> states) {
+        Map<CardStateView, CardState> stateViewCache = Maps.newLinkedHashMap();
+        for (CardState state : states) {
+            stateViewCache.put(state.getView(), state);
```
tool4ever, line 47: this doesn't work yet because (not sure yet if bug or not) both cardstates from room have same id
so map never gets both entries :/

Hanmac: should be fixed?

forge-gui/res/cardsfolder/upcoming/ghostly_keybearer.txt
Hanmac: what exactly did you do so i can reproduce it?

tool4ever: - Cast _Copy Enchantment_ and copy one of the rooms in this PR
- try to use _Ghostly Keybearer_ trigger to unlock
- it seems to fail because it's not `CardSplitType.Split`

Hanmac: check removed

tool4ever (approved):
seems no room got printed with RE / or activated ability 🤷‍♂️
so I feel like it's pretty ready

Hanmac:
I think the LookOrUnlock part might need some polish 

Also the AI part might need to be tweaked

tool4ever (changes requested):
@Hanmac
we need to fix `sharesNameWith(final String name)` with locked checks somehow

If I cast the same room once with left and once with right _Echoing Calm_ can destroy both

Hanmac:
> @Hanmac we need to fix `sharesNameWith(final String name)` with locked checks somehow
> 
> If I cast the same room once with left and once with right _Echoing Calm_ can destroy both

that should be fixed?

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -5987,9 +6000,21 @@ public final boolean sharesNameWith(final String name) {
         boolean shares = getName(true).equals(name);
 
         // Split cards has extra logic to check if it does share a name with
-        if (isSplitCard()) {
-            shares |= name.equals(getState(CardStateName.LeftSplit).getName());
-            shares |= name.equals(getState(CardStateName.RightSplit).getName());
+        if (!shares) {
+            if (isInPlay()) {
```
tool4ever, line 6004: ugh, this two names stuff is really confusing me...
How do they get overwritten?
Here's another case:
image

- both Rooms must be unlocked to only the same side
- then animate one to enchant it
- if you target the creature with the Boardwipe the room gets destroyed too
- but if you target the room the creature lives

something can't be right there 🤔

Hanmac: Ugh, I probably need to debug that more

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -7476,6 +7501,15 @@ public List<SpellAbility> getAllPossibleAbilities(final Player player, final boo
             }
         }
 
+        if (isInPlay() && isRoom() && !isPhasedOut() && player.canCastSorcery()) {
```
tool4ever, line 7504: I'm starting to think checking `isRoom` isn't correct:
> 116.2m. A player who controls a permanent that has one or more locked halves (see rule 709.5) may pay the mana cost of a locked half of that permanent to give that permanent the appropriate unlocked designation.

So it should still work when enchanted by _Minimus Containment_?
Or only if it doesn't use `RemoveAllAbilities$`? 🤔

Hanmac: That's probably for rules manager to asked

Hanmac: Also Special Action might not be removed by that

Hanmac: fixed, it doesn't need to check for Room anymore

forge-game/src/main/java/forge/game/card/Card.java
Hanmac: ugh, can you help me to debug this?

tool4ever: if the Room is enchanted with _Witness Protection_ it will only have one name instead of two?

tool4ever: should be ok for now I think

tool4ever (approved):

Issue #3397, opened 2023-07-03
FCollection View vs unmodifiable vs Immutable Refactor


Hanmac:
Currently the FCollection View sits in the FCollection Class and doesn't implement either java Collection or java List
so each time we get a View and want to iterate over it in a different way, or want to get the size/empty check we need to cast to List or make a new Array List?

small steps:

- Make View implement List and Collection class, but make them fail on change, maybe implement them as Immutable List?
- Also use Immutable for EmptyFCollection ?
- Maybe add an Interface for FCollection that has all the methods (including the Modifying ones) and have View and Empty Collection use that Interface
- Implement FCollection as [ForwardingList](https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/ForwardingList.html), + some Set Functions then it should be more easy to put a Immutable List inside for Empty?

Hanmac:
My ideas so far to combine the points above:

* new FCollectionInterface that extends from Collection Interface
* new FCollectionAbstract that extends from ForwardingList, (so we could inject Immutable/Unmodifiable Views?)
* Then FCollection will extend FCollectionAbstract using [SetUniqueList](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/list/SetUniqueList.html)


That should be the best of all worlds 🤔

Hanmac:
[SetUniqueList](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/list/SetUniqueList.html) isn't sortable

might need some trickery?

Jetz72:
What does forwarding to a SetUniqueList add that our current FCollection implementation isn't already capable of?

Hanmac:
> What does forwarding to a SetUniqueList add that our current FCollection implementation isn't already capable of?

I wanted to try to reduce our own code for such collection to a community maintained one, to reduce the sources for possible errors

Jetz72:
Would something like this work then?
* `IFCollection` - interface that implements Collection, possibly List, `stream()`, plus any other helper methods.
* `FCollection` - Implements `IFCollection`, extends `SetUniqueList` (rather than forwarding to it), and holds the list that the parent SetUniqueList operates on.
* `FCollectionView` - Implements `IFCollection`, wraps an `FCollection`, except all write methods throw unsupported operation exceptions.
* `EmptyFCollection` - Extends `FCollectionView` with an empty collection, all methods have trivial implementations.

[...]

PR #6047, opened 2024-09-03
Migrate Guava Predicates and Functions to Java implementations

Jetz72:
Per #5717.

Not complete yet but wouldn't hurt to get some eyes on it while it's easier to revise things.

The switch-over of the implementation is done, but still need to finish cleaning up the helper methods. Many can be simplified through method references (e.g. `::equals` and `::contains`), but others don't have a good equivalent.

The dependency on Guava's `Iterables` is one point of uncertainty at the moment. It's too convenient to get rid of, but a handful of the filter methods still want Guava predicates. Could reimplement these manually in an `IterableUtils` class, or could just use the `predicate::test` hack to keep using the original library.


Jetz72:
Finally got the nasty part of this migration done, primarily [ #76aa196 ](https://github.com/Card-Forge/forge/pull/6047/commits/76aa19622ec65db5a1e71fed1792c183a97818a5). Some cases were clear improvements when migrated from Iterables to Streams. Some cases were skipped because they would have needed substantial rewrites or just looked much worse when I tried to replace them. And some cases could be argued either way, and if anyone wants to weigh in on the matter, I'm all ears.

Couple things I'd like to do before finishing up:
* Add some stream utilities like collectors and reducers for common use cases. Biggest example of this is `Aggregates.random` - Rather than collecting a stream into a list only to pass it over to this method, it'd be nice to be able to just cap off the stream with `.collect(Aggreagtes.collectToRandom)` or something and spit out the chosen item directly.
* Clean up the CollectionSuppliers class since it's literally just method references now.
* Once we've dropped support for Java 8, I'd like to switch some uses of `predicate.negate()` back to `Predicate.not(predicate)`. For elaborate predicates with and's and or's , I believe it's easier to read when you have the "not" up front and the inverted logic within the parentheses. There are also a couple cases where `toUnmodifiableList` would come in handy.
* Run the "optimize imports" IDE task since a lot of those have been swapped around and switched to and from wildcards during this process.

Once that's done, final steps will be to split out the predicate/function methods from migration aid classes `Iterables` and `Predicates` into `IterableUtils` or something. The remaining Iterables methods that don't need predicate support can be pointed back to Guava's Iterables.

Hanmac:
I think one issue for example is that FCollectionView isn't a Collection type, while most of the filters/stream functions would like to have a Collection instead of an Iterable.

i made an extra Issue for this: #3397 

I wonder is we really need FCollectionView as a Type and just return FCollection instead and hope for the best. (i think FCollection was just something to make the Collection not modifyable)

I will try to make an extra MR for this

Jetz72:
FCollectionView could use some fleshing out. Not sure whether that means adding a stream method, implementing a more generally useful interface in it, or dropping it entirely. But I didn't actually run into it all that often while replacing uses of `Iterables`.

The one that gave me the most trouble was actually just variables defined as `Iterable<...>`. Many functions take or return a value typed as Iterable since it's more efficient to pass around a view created by `Iterables.filter` than it is to recreate an entire collection. Unfortunately those iterables aren't streamable, so I couldn't do much with them without either changing the types at their roots or wrapping them in a lambda. With that in mind, there probably is merit to keeping FCollectionView around if we actually apply it for things like filters.

forge-ai/src/main/java/forge/ai/ability/DamageDealAi.java
```diff
@@ -893,7 +892,11 @@ private boolean damageChooseRequiredTargets(final Player ai, final SpellAbility
 
             // See if there's an indestructible target that can be used
             CardCollection indestructible = CardLists.filter(ai.getCardsIn(ZoneType.Battlefield),
-                    Predicates.and(CardPredicates.Presets.CREATURES, CardPredicates.Presets.PLANESWALKERS, CardPredicates.hasKeyword(Keyword.INDESTRUCTIBLE), CardPredicates.isTargetableBy(sa)));
+                    CardPredicates.CREATURES
+                            .and(CardPredicates.PLANESWALKERS) //TODO: Should this be "or" Planeswalkers?
+                            .and(CardPredicates.hasKeyword(Keyword.INDESTRUCTIBLE))
+                            .and(CardPredicates.isTargetableBy(sa))
+            );
```
Jetz72, line 899: Per the TODO, should this be "creatures or planeswalkers"? The way it was written previously would only find cards that have both types.

forge-core/src/main/java/forge/card/CardRulesPredicates.java
```diff
@@ -546,108 +535,55 @@ public PredicateSplitType(final CardSplitType type) {
...
+    public static final Predicate<CardRules> IS_BATTLE = CardRulesPredicates.coreType(true, CardType.CoreType.Battle);
+    public static final Predicate<CardRules> IS_INSTANT = CardRulesPredicates.coreType(true, CardType.CoreType.Instant);
+    public static final Predicate<CardRules> IS_SORCERY = CardRulesPredicates.coreType(true, CardType.CoreType.Sorcery);
+    public static final Predicate<CardRules> IS_ENCHANTMENT = CardRulesPredicates.coreType(true, CardType.CoreType.Enchantment);
+    public static final Predicate<CardRules> IS_NON_CREATURE_SPELL =
+            CardRulesPredicates.IS_SORCERY
+                    .or(CardRulesPredicates.IS_INSTANT)
+                    .or(CardRulesPredicates.IS_PLANESWALKER)
+                    .or(CardRulesPredicates.IS_ENCHANTMENT) //TODO: Battles? Is testing these one by one really the best way to check "non-creature"?
+                    .or(CardRulesPredicates.IS_ARTIFACT.and(CardRulesPredicates.IS_CREATURE.negate()));
```
Jetz72, line 572: Per the TODO, I believe Battles should be added here? Though it may be better to just rewrite this as something derived from the type rather than maintaining a list of them here.

Hanmac: Why not just use `CardRulesPredicates.IS_CREATURE.negate()` for non-creature-spell?

forge-ai/src/main/java/forge/ai/ability/DamageDealAi.java
tool4ever: yea no doubt OR is the better choice

forge-core/src/main/java/forge/card/CardRulesPredicates.java
tool4ever: this is a good catch as the current form has become outdated, leading to some chance that card type ratios from deck generation aren't reaching the desired values (or depending on the higher level logic at least cause more looping)

Imo we could go with NonCreature + NonLand + `!isVariant()` just to be safe

Jetz72:
Added `StreamUtils.random` and converted a handful of uses of `Aggregates.random` to it instead. I didn't bother changing the ones that just passed the source collection right into the method, just the ones that had other operations that could also be part of a stream pipeline. I was gonna write up `StreamUtils.max` and `min` methods that took accessor functions like their Aggregate counterparts, but turns out you can just use `stream.max(Comparator.comparing(whatever))`. 

Added some filter predicates that can be tacked on to `CardDb.get[All/Unique]Cards` to replace the non-promo and non-reprint variants of those methods.

When replacing Aggregates.random calls in the quest rewards, I realized I was cleaning up 5 copies of the same method with a different filter, so I ended up merging those into one plus some helper methods.

There are some lingering items that have been accumulated on my to-do list - e.g. individual methods that could be rewritten to take advantage of streams but aren't easy replacements, parameters or return types that could be shifted from Iterable to a more concrete collection, simplifying things elsewhere... But I'll probably just leave those alone because I have to draw the line somewhere. 

The biggest thing I'll probably leave for another PR is a cleanup of `CardLists.filter` and its derivatives. Those methods create a new collection every time they're called, but often enough there isn't actually a need for one. There are many places where it's only used for counting items, picking a single item, or populating another collection (possibly with another call to `CardLists.filter`), and I wonder if the similar name to `Iterables.filter` has created the expectation that it behaves similarly.

That and whatever else I come across at the last minute aside, should be done soon. Need to remove the migration aid classes, resolve merge conflicts, cleanup imports, and spend some time testing it.

forge-core/src/main/java/forge/util/StreamUtil.java
```diff
@@ -0,0 +1,109 @@
...
+        int sampleCount = 0;
+
+        public RandomReservoir(int size) {
+            this.maxSize = size;
+            this.samples = new ArrayList<>(size);
+        }
+
+        public void accumulate(T next) {
+            sampleCount++;
+            if(sampleCount < maxSize) {
```
tool4ever, line 98: I believe you introduced a logic error here

Jetz72: Good catch. I tried a few different implementations of that collector to try and get parallel support in before giving up and going back to something close to the original. Must have missed that `<=` in the process.

tool4ever:
@Hanmac
do you still want to check anything? otherwise since release happened we might ship it
maybe keep it unsquashed bc of the size? 🤔

kevlahnota:
@Hanmac is this PR good, I think java is more performant than guava, also do we need guava (I don't know if java covers all other collections  guava have)

Hanmac:
@kevlahnota i will look over this MR, but i think it should be good enough?

we can't really drop guava fully, i would miss my Table support

Jetz72:
I looked at cutting out more of Guava and recreating the functionality within other util classes, but the scope of that started expanding too quickly for this PR at least. They have a huge number of helper methods that we use, and many of those are implemented through internal helper classes which would either need to be reproduced as well or replaced with inline logic. 

If we want to further lessen the dependency on Guava, I think a good direction would be expanding the functionality of `FCollection`, `FCollectionView`, and their derivatives. Passing data around through those instead of `Iterable`s would reduce our reliance on Guava's `Iterables` methods.

For performance improvements, as mentioned above, I think `CardLists` could do with some scrutiny. I believe many methods there create entirely new collections each time they're called, and they're often used even when the call site only needs a filtered view.

kevlahnota (changes requested):
the toList method directly from stream are missing on older Android (12 and below tested, not sure on 13 and above), so this will break mobile builds. 
[#4bbc165460cbe87a32eb4227f5b4fda26959d9f9](https://github.com/Card-Forge/forge/pull/6047/commits/4bbc165460cbe87a32eb4227f5b4fda26959d9f9)

Jetz72:
Gah. I was excited when I saw we'd moved up to Java 17. Should I just treat the language level as though it were 11, then?

kevlahnota:
> Gah. I was excited when I saw we'd moved up to Java 17. Should I just treat the language level as though it were 11, then?

Hmm I just discovered it when I update the auditer. Don't know which Java api are excluded on android.

Jetz72:
Android 12 and 13 reportedly support up to Java 11, and Android 14 partially supports Java 17. 
https://developer.android.com/build/jdks#compileSdk

From what I can tell, the D8 compiler has support for backporting newer Java API members via desugaring, but it sounds like that's only being used to enable stuff from the Java 11 API on Android <= 11. 
https://developer.android.com/studio/write/java11-default-support-table
https://issuetracker.google.com/issues/304335616

If the project is gonna be set to Java 17, though, It would probably be a good idea to add a github workflow to test the android build, or otherwise report unsupported API members. It seems like an easy thing to overlook.

Worth keeping an eye on this in case desugaring support for 17 gets added someday - https://github.com/google/desugar_jdk_libs/blob/master/CHANGELOG.md

Hanmac (approved):

tehdiplomat:
Looks like this broke entering the Inn/Drafting in Adventure mode.

PR #6052, opened 2024-09-04
TriggerFullyUnlock: Trigger used for Eerie

Hanmac:
Trigger that will be used For Room.

Early commit so Eerie cards can be scripted.

### DIFF
--- a/forge-game/src/main/java/forge/game/trigger/TriggerFullyUnlock.java
+++ b/forge-game/src/main/java/forge/game/trigger/TriggerFullyUnlock.java
@@ -1,0 +1,41 @@
+package forge.game.trigger;
+
+import java.util.Map;
+
+import forge.game.ability.AbilityKey;
+import forge.game.card.Card;
+import forge.game.spellability.SpellAbility;
+import forge.util.Localizer;
+
+public class TriggerFullyUnlock extends Trigger {
+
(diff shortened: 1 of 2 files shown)


forge-game/src/main/java/forge/game/trigger/TriggerFullyUnlock.java
```diff
@@ -0,0 +1,41 @@
+package forge.game.trigger;
+
+import java.util.Map;
+
+import forge.game.ability.AbilityKey;
+import forge.game.card.Card;
+import forge.game.spellability.SpellAbility;
+import forge.util.Localizer;
+
+public class TriggerFullyUnlock extends Trigger {
```
tool4ever, line 10: imo it should be just TriggerUnlock so it can be used for "When you unlock this door..." too
like a Side$ param with Left/Right/Both

Hanmac: Imo I think extra Trigger would make more sense

Because for example when you later unlock the second door, it triggers both:
* Unlock (this) Door "Left/Right"
* Fully Unlock Room

tool4ever: Hmn, I'm not sure I see the advantage there but it's no big deal anyway :)

tool4ever (approved):

PR #6055, opened 2024-09-04
Manifest Dread

### LINKED ISSUE #6032: DSK: Manifest Dread
Update of Manifest Effect

- use `Dread$ True` param.
- Then look at Top 2 cards, and select one.
- Manifest this one.
- And move the other into Graveyard.
- ChangeZoneTable should trigger only once.

tool4ever:
Closes #6032


Hanmac (approved):
I'm not 100% sure if the changes Zone Trigger happens at the same time, but that can be cleanup later

PR #6063, opened 2024-09-04
Eerie DSK cards

Fulgur14:


forge-gui/res/cardsfolder/upcoming/stalked_researcher.txt
```diff
@@ -0,0 +1,10 @@
+Name:Stalked Researcher
+ManaCost:1 U
+Types:Creature Human Wizard
+PT:3/3
+K:Defender
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Enchantment.YouCtrl | TriggerZones$ Battlefield | Execute$ TrigCanAttack | TriggerDescription$ Eerie — Whenever an enchantment you control enters and whenever you fully unlock a Room, CARDNAME can attack this turn as though it didn't have defender.
+T:Mode$ FullyUnlock | ValidCard$ Card.Room | ValidPlayer$ You | Secondary$ True | Execute$ TrigCanAttack | TriggerDescription$ Eerie — Whenever an enchantment you control enters and whenever you fully unlock a Room, CARDNAME can attack this turn as though it didn't have defender.
+SVar:TrigCanAttack:Mode$ CanAttackDefender | ValidCard$ Card.EffectSource | Description$ EFFECTSOURCE can attack this turn as though it didn't have defender.
```
tool4ever, line 8: this isn't the way to script a static

forge-gui/res/cardsfolder/upcoming/victor_valgavoths_seneschal.txt
```diff
@@ -0,0 +1,13 @@
+Name:Victor, Valgavoth's Seneschal
+ManaCost:1 W B
+Types:Legendary Creature Human Warlock
+PT:3/3
+K:Vigilance
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Enchantment.YouCtrl | TriggerZones$ Battlefield | Execute$ TrigSurveil | TriggerDescription$ Eerie — Whenever an enchantment you control enters and whenever you fully unlock a Room, surveil 2 if this is the first time this ability has resolved this turn. If it's the second time, each opponent discards a card. If it's the third time, put a creature card from a graveyard onto the battlefield under your control.
```
tool4ever, line 6: this is a very problematic card for us because we have to script it as two triggers
but the ResolvedLimit should obviously be shared between both

most likely you have no other option than switching to ugly StoreSVar counting :/

tool4ever: hmn, actually can you try if it would still work?
just use some functioning trigger instead

tool4ever: tough I'm pretty sure it doesn't, otherwise we could have cleaned up _Corruption of Towashi_ too...

Fulgur14: I don't think I can test it right now, I'm going to bed soon.

tool4ever (approved):
thanks

PR #6079, opened 2024-09-05
Keyword: link StaticAbility into it

### LINKED ISSUE #6075: Find a cleaner way to fix Fumiko
> X is variable. The bushido bonus is calculated each time Fumiko’s bushido trigger resolves, based on the number of attackers at that time.

as for the Keywords stuff, i will try some idea when i get home to have `KeywordInterface` implement `getSVarFallback` via StaticAbility and CardState source

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/pull/6041#discussion_r1745601605_

Hanmac:
Closes #6075

i couldn't get Soulshift working yet, because that should use LKI

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -5823,9 +5823,15 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
             if (StringUtils.isNumeric(s)) {
                count += Integer.parseInt(s);
             } else {
-                String svar = StringUtils.join(parse);
-                if (state.hasSVar(svar)) {
-                    count += AbilityUtils.calculateAmount(this, state.getSVar(svar), null);
+                StaticAbility st = inst.getStatic();
+                // TODO make keywordinterface inherit from CardTrait somehow, or invent new interface
+                if (st != null && st.hasSVar(s)) {
+                    count += AbilityUtils.calculateAmount(this, st.getSVar(s), null);
+                } else {
+                    String svar = StringUtils.join(parse);
+                    if (state.hasSVar(svar)) {
+                        count += AbilityUtils.calculateAmount(this, state.getSVar(svar), null);
+                    }
                 }
             }
         }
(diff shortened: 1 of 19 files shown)


tool4ever:
Doesn't seem to work fully yet:
_Takeno, Samurai General_ should also buff _Fumiko_ (while there are attacking creatures)

Hanmac:
Fumiko + Takeno fixed

Hanmac:
this fallback might be able to be removed, i didn't find any example for it

```java
String svar = StringUtils.join(parse);
if (state.hasSVar(svar)) {
    count += AbilityUtils.calculateAmount(this, state.getSVar(svar), null);
}
```

tool4ever (approved):
should work for now 🤔

Hanmac:
I want to refactor Soulshift X but that doesn't work yet

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -5821,9 +5823,15 @@ public final int getKeywordMagnitude(final Keyword k, CardState state) {
...
-                    count += AbilityUtils.calculateAmount(this, state.getSVar(svar), null);
+                StaticAbility st = inst.getStatic();
+                // TODO make keywordinterface inherit from CardTrait somehow, or invent new interface
+                if (st != null && st.hasSVar(s)) {
+                    count += AbilityUtils.calculateAmount(this, st.getSVar(s), null);
+                } else {
+                    String svar = StringUtils.join(parse);
+                    if (state.hasSVar(svar)) {
+                        count += AbilityUtils.calculateAmount(this, state.getSVar(svar), null);
+                    }
```
Hanmac, line 5834: @tool4ever can we safely remove this part?

tool4ever: fallback might still be needed in some cases when "fully" intrinsic e. g. _Petrified Wood-Kin_ with "Bloodthirst X" (though this one might need to be unhardcoded first)

Hanmac: The difference there is that Bushido Xis gained via Static Ability, while Bloodthirst X is a direct keyword.


Bloodthirst X is hardcoded because it is special in the rules:

> 702.54b “Bloodthirst X” is a special form of bloodthirst. “Bloodthirst X” means “This permanent enters with X +1/+1 counters on it, where X is the total damage your opponents have been dealt this turn.”

and also doesn't use getKeywordMagnitude
* This one is only used for Toxic and Bushido
* And only Bushido is gained via special Static

PR #6081, opened 2024-09-06
Update Adventure BG files

kevlahnota:


kevlahnota:
> @kevlahnota why do all fallback_skin files need to exist 3 times? can't they all point to the same location? :/ I guess the current pom files defaults to their own filters and assets and packaged into their own jars / apk container. Maybe it can be done to update their pom to get those fallback skins at single source when building

PR #6086, opened 2024-09-06
Overgrown Zealot (DSK)

Fulgur14:
Note that wording is a bit different than on Tin Street Gossip, "permanents" vs. "creatures". This would matter for Yedora, Grave Gardener.


tool4ever:
ugh, you're right, I need to touch it again :/

tool4ever (approved):

PR #6096, opened 2024-09-07
Impending: use Effect for Last Time counter removed

### LINKED ISSUE #5481: DSK: Impending Keyword

Hanmac:
Closes #5481 

This makes Impending use Effect for the "last Time counter removed" trigger to work

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
@@ -463,6 +463,11 @@ public abstract class SpellAbilityEffect {
         card.addChangedSVars(Collections.singletonMap("EndOfTurnLeavePlay", "AtEOT"), card.getGame().getNextTimestamp(), 0);
     }
 
+    protected static SpellAbility getExileSpellAbility(final Card card) {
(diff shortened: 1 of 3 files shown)


tool4ever:
This is so:
a) it can't lose creature again if a time counter gets added later gain
b) "loses all abilities" doesn't interact with it
?

Hanmac:
> This is so: a) it can't lose creature again if a time counter gets added later gain b) "loses all abilities" doesn't interact with it ?

That's my current understanding of how it should work, yes.


For example if you cast it while Solemnity us on the field, it's just a non Creature forever

tool4ever (approved):
I wonder if the duration will be "for as long as" (so phasing out should also break it), but that can be checked when CR is updated

Hanmac:
> I wonder if the duration will be "for as long as" (so phasing out should also break it), but that can be checked when CR is updated

I asked if it works like "for as long as" and the answer was no.

Especially because Solemnity would have stopped the Effect from even starting

PR #6103, opened 2024-09-08
Update Android build (SDK 35), Java 17, LibGDX 1.13.0

kevlahnota:
- Java 17
- targetSDK 35 (min 26)
- add modded uber-apk-signer.jar for signing cert
- add standalone proguard 7.6.0
- update some dependency
- update new default skin
- update LibGDX (1.13.0)
- add standalone gdx particle editor
- create jar installer with izpack


tehdiplomat:
Awesome!

Some thoughts:
- We should bump the versions up to 2.0.0 and announce that we're no longer going to be supporting Java 8 at the same time we merge this PR.
- It might be nice to do one more full release that supports 1.8 before this gets merged.
- Test with java 8 workflow `.github/workflows/test-build.yaml` won't work for this PR, but we can just bypass protections so we don't need to merge it in a different PR

Hanmac:
@kevlahnota would be nice if we could implement https://github.com/Card-Forge/forge/pull/4977 in this update

kevlahnota:
> @kevlahnota would be nice if we could implement #4977 in this update

Just push a new release / update the plugin  https://github.com/Card-Forge/android-maven-plugin
it should work since the PR is merged.

Hanmac:
@kevlahnota for Signing, checkout https://github.com/MuntashirAkon/apksig-android

kevlahnota:
> @kevlahnota for Signing, checkout ~~ https://github.com/MuntashirAkon/apksig-android ~~
> 
> Better: https://github.com/jenkinsci/android-signing-plugin

Reverted to sdk 29, seems target sdk 33 and above somehow don't allow installation without proper signature/signing on latest android version (tested with android 14)

Jetz72:
> Reverted to sdk 29, seems target sdk 33 and above somehow don't allow installation without proper signature/signing on latest android version (tested with android 14)

Isn't it just that it needs to be signed with signature scheme v2?
https://developer.android.com/about/versions/11/behavior-changes-11#minimum-signature-scheme
Should be possible but it would need an upgrade from jarsigner.

I think the requirement that the app target the latest Android version is only if you're releasing an app on the Google Play store.

kevlahnota:
> > Reverted to sdk 29, seems target sdk 33 and above somehow don't allow installation without proper signature/signing on latest android version (tested with android 14)
> 
> Isn't it just that it needs to be signed with signature scheme v2? https://developer.android.com/about/versions/11/behavior-changes-11#minimum-signature-scheme Should be possible but it would need an upgrade from jarsigner. Which I guess would also mean an upgrade from simpligility, since that hasn't been updated since 2019? I'm not too experienced with Maven workflows.
> 
> I think the requirement that the app target the latest Android version is only if you're releasing an app on the Google Play store.

From android sdk 30 and above, the resources.arsc needs to be uncompressed inside the apk, the rest have no problem being compressed. Then you need to zipalign and sign with v2 and it will install fine. I have a working build on my local computer but It's a hassle to manually fix those things. Yes it's requirement for playstore but using latest sdk stops the nagging popup that the apk is outdated and latest sdk supports more java api (sdk 34 uses java 17)

.github/workflows/snapshots-android.yml
```diff
@@ -89,7 +89,7 @@ jobs:
           d=$(date +%m-%d)
           # Replace date in forge-gui-mobile/src/forge/Forge.java
           sed -i -e "s/-SNAPSHOT/-SNAPSHOT-${d}/g" forge-gui-mobile/src/forge/Forge.java
-          mvn -U -B -P android-release-build,android-release-sign install -e -Dsign.keystore=forge.keystore -Dsign.alias=Forge -Dsign.storepass=${{ secrets.SIGN_STORE_PASS }} -Dsign.keypass=${{ secrets.SIGN_STORE_PASS }}  -Dcardforge-repo.username=${{ secrets.FTP_USERNAME }} -Dcardforge-repo.password=${{ secrets.FTP_PASSWORD }} -Dandroid.sdk.path=/usr/local/lib/android/sdk -Dandroid.buildToolsVersion=30.0.3 -Dmaven.test.skip=true
+          mvn -U -B -P android-release-build install -e -Dcardforge-repo.username=${{ secrets.FTP_USERNAME }} -Dcardforge-repo.password=${{ secrets.FTP_PASSWORD }} -Dandroid.sdk.path=/usr/local/lib/android/sdk -Dandroid.buildToolsVersion=33.0.2 -Dmaven.test.skip=true
```
tehdiplomat, line 92: Signing isn't needed anymore?

kevlahnota: It's included on the pom file, it will sign with v2, v3 on install phase

Jetz72: Are those keystore passwords just placeholders for testing?

kevlahnota: check publish.bat in android module.

Jetz72: Knowing that the keystore credentials have already been in a publicly visible file for the last 9 years isn't doing much to mitigate my concern... Someone would still need the keystore itself to do anything malicious but if those are the actual passwords then that's a layer of security that isn't doing anything.

kevlahnota: If the github actions template works on the pom file, feel free to update it since I don't know how those variable interact from github to maven via pom file

Jetz72: I'll admit I don't know much about that either, and the credentials would need to be changed by someone with more access than me anyway. I could open an issue for it but I'm hesitant to give security problems more visibility than necessary. It's probably not a huge problem in the short term though.

.github/workflows/publish-android.yml
```diff
@@ -71,19 +71,19 @@ jobs:
...
+          #mvn install:install-file -Dfile=android-maven-plugin-4.6.2.jar -DgroupId=com.simpligility.maven.plugins -DartifactId=android-maven-plugin -Dversion=4.6.2 -Dpackaging=jar
           cd -
           mvn install -Dmaven.test.skip=true
           mvn dependency:tree
           
       - name: Build/Install/Publish to GitHub Packages Apache Maven
         run: |
           export _JAVA_OPTIONS="-Xmx2g"
-          mvn -U -B -P android-release-build,android-release-sign,android-release-upload install -e -Dsign.keystore=forge.keystore -Dsign.alias=Forge -Dsign.storepass=${{ secrets.SIGN_STORE_PASS }} -Dsign.keypass=${{ secrets.SIGN_STORE_PASS }}  -Dcardforge-repo.username=${{ secrets.FTP_USERNAME }} -Dcardforge-repo.password=${{ secrets.FTP_PASSWORD }} -Dandroid.sdk.path=/usr/local/lib/android/sdk -Dandroid.buildToolsVersion=30.0.3 -Dmaven.test.skip=true
+          mvn -U -B -P android-release-build,android-release-upload install -e -Dcardforge-repo.username=${{ secrets.FTP_USERNAME }} -Dcardforge-repo.password=${{ secrets.FTP_PASSWORD }} -Dandroid.sdk.path=/usr/local/lib/android/sdk -Dandroid.buildToolsVersion=35.0.0 -Dmaven.test.skip=true
```
tehdiplomat, line 86: Is Storepass not needed?

kevlahnota: > Is Storepass not needed?

yes, since it's already defined on the pom

tehdiplomat:
Ah damn, sorry for all the merge conflicts.

kevlahnota:
> Ah damn, sorry for all the merge conflicts.

done

Hanmac:
@kevlahnota while checking pom.xml stuff, can you look if we really need `<alpha-version>1.6.65-SNAPSHOT</alpha-version>` ? Shouldn't that always be the same as `version` after maven does add -SNAPSHOT again?

kevlahnota:
> @kevlahnota while checking pom.xml stuff, can you look if we really need `<alpha-version>1.6.65-SNAPSHOT</alpha-version>` ? Shouldn't that always be the same as `version` after maven does add -SNAPSHOT again?

on android pom I remove it, and use the ${revision}, I append the monthdate code after it to the generated property ${snapshot-version}

.github/workflows/snapshots-android.yml
```diff
@@ -29,30 +29,32 @@ jobs:
...
+          export PATH=$PWD/apache-maven-3.8.1/bin:$PATH
+          export MAVEN_HOME=$PWD/apache-maven-3.8.1
           mvn --version
           
       - name: Set Up Android tools
         run: |
-          JAVA_HOME=${JAVA_HOME_11_X64} ${ANDROID_SDK_ROOT}/cmdline-tools/latest/bin/sdkmanager --sdk_root=$ANDROID_SDK_ROOT "platform-tools"
-          JAVA_HOME=${JAVA_HOME_11_X64} ${ANDROID_SDK_ROOT}/cmdline-tools/latest/bin/sdkmanager --sdk_root=$ANDROID_SDK_ROOT --install "platforms;android-26"
-          JAVA_HOME=${JAVA_HOME_11_X64} ${ANDROID_SDK_ROOT}/cmdline-tools/latest/bin/sdkmanager --sdk_root=$ANDROID_SDK_ROOT --install "build-tools;30.0.3"
+          JAVA_HOME=${JAVA_HOME_17_X64} ${ANDROID_SDK_ROOT}/cmdline-tools/latest/bin/sdkmanager --sdk_root=$ANDROID_SDK_ROOT --install "build-tools;26.0.0"
```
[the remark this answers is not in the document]
kevlahnota: I'm surprised too, i found the answer here: https://github.com/simpligility/android-maven-plugin/issues/811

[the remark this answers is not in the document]
kevlahnota: I tried to install both "build-tools;30.0.3" and "build-tools;35.0.0" without the platform-26, and it compiled fine(I'm really confused). I can only guess that earlier build tools extract to different folder whereas the modified android maven plugin is finding the old build tools path by default on generate sources step...

tehdiplomat (approved):
2️⃣

.github/workflows/snapshots-android.yml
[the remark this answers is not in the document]
kevlahnota: just patched the android maven plugin, and it doesn't need the old build tools.

Hanmac:
Do we need the ParticleEditor? (or the replacement?)

kevlahnota:
> Do we need the ParticleEditor? (or the replacement?)

Its for modifying the effects or create a new particle effects, the new one is better and is stand alone. looks good and the user can download if they plan to add new effects..  though it's Scene2D (like the adventure mode and not Swing). It's really optional.
image
image
image

Hanmac (dismissed):
We might update the forge.sh files too

forge-gui-desktop/src/main/config/forge.sh
```diff
@@ -1,52 +1,3 @@
...
-elif [[ $v -ge 11 ]]
-then
-    java --illegal-access=permit $SHAREDPARAMS
-else
-    java $SHAREDPARAMS
-fi
+java -Xmx4096m --add-opens java.desktop/java.beans=ALL-UNNAMED --add-opens java.desktop/javax.swing.border=ALL-UNNAMED --add-opens java.desktop/javax.swing.event=ALL-UNNAMED --add-opens java.desktop/sun.swing=ALL-UNNAMED --add-opens java.desktop/java.awt.image=ALL-UNNAMED --add-opens java.desktop/java.awt.color=ALL-UNNAMED --add-opens java.desktop/sun.awt.image=ALL-UNNAMED --add-opens java.desktop/javax.swing=ALL-UNNAMED --add-opens java.desktop/java.awt=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED --add-opens java.base/java.text=ALL-UNNAMED --add-opens java.desktop/java.awt.font=ALL-UNNAMED --add-opens java.base/jdk.internal.misc=ALL-UNNAMED --add-opens java.base/sun.nio.ch=ALL-UNNAMED --add-opens java.base/java.nio=ALL-UNNAMED --add-opens java.base/java.math=ALL-UNNAMED --add-opens java.base/java.util.concurrent=ALL-UNNAMED --add-opens java.base/java.net=ALL-UNNAMED -Dio.netty.tryReflectionSetAccessible=true -Dfile.encoding=UTF-8 -jar $project.build.finalName$
```
Hanmac, line 3: @kevlahnota can you add the changes for this, checking for JAVA_HOME? #6008 

Also are the add-opens in the forge.sh needed, when the pom.xml already adds them?

kevlahnota: The java home doesn't work when i test here though I'm on windows and using liberica jdk,  double clicking the sh file here with my changes works. Also the add opens in the pom are for launch4j when producing the exe file for windows.

kevlahnota: @Hanmac I update the replacer plugin so it adds needed java 17 args to run. the contributor in #6008 should just update his PR so the JAVA_HOME environment vars replacement should work on all OS (since it fails here on windows upon testing)

Hanmac: The JAVA_HOME thing is only done for sh files (under linux/mac)

And wasn't done for cmd files yet
Which probably should be different 🤔

kevlahnota:
@tehdiplomat @Hanmac seems the new LibGDX is good, no crashes so far on testing here.

kevlahnota:
@tehdiplomat this will create the forge.zip (archive) and forge-installer-2.0.00-SNAPSHOT.jar (installer) in forge-installer/target module. either of those two can be distributed.. though the github actions needs update to reflect this change..

tehdiplomat (approved):

PR #6114, opened 2024-09-09
Support for Valgavoth's Onslaught

tool4ever:


Hanmac (approved):
works for now i think, even if the confusion of Amount vs Times

Hanmac:
I might clean that up later with its own effect, but it is good for now

PR #6116, opened 2024-09-10
Player: use LifeReduced CantHappen for Archon of Coronation

Hanmac:
use LifeReduced CantHappen instead of Keyword

Also fixes FatefulHour flag on trigger/replacement


forge-game/src/main/java/forge/game/CardTraitBase.java
```diff
@@ -311,7 +311,10 @@ protected boolean meetsCommonRequirements(Map<String, String> params) {
             if ("True".equalsIgnoreCase(params.get("Bloodthirst")) != hostController.hasBloodthirst()) return false;
         }
         if (params.containsKey("FatefulHour")) {
-            if ("True".equalsIgnoreCase(params.get("FatefulHour")) != (hostController.getLife() > 5)) return false;
+            if ("True".equalsIgnoreCase(params.get("FatefulHour")) != (hostController.getLife() < 5)) return false;
```
tool4ever, line 314: should actually be <= 5

tool4ever (approved):

PR #6122, opened 2024-09-10
2 DSK cards, some DSC cards, 5 Schemes

Fulgur14:
I noticed an issue with Because I Have Willed It, but not sure how to correct: it should only count spells cast by the triggered player, not all spells you don't control.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/a_premonition_of_your_demise.txt
+++ b/forge-gui/res/cardsfolder/upcoming/a_premonition_of_your_demise.txt
@@ -6,5 +6,5 @@ SVar:DBDig:DB$ Dig | DigNum$ 2 | Reveal$ True | ChangeNum$ All | ChangeValid$ Ca
 SVar:DBImmediateTrigger:DB$ ImmediateTrigger | ConditionDefined$ Remembered | ConditionPresent$ Card.nonLand | Execute$ TrigDamage | SubAbility$ DBCleanup | SpellDescription$ When you reveal one or more nonland cards this way, this scheme deals damage equal to their total mana value to any target.
 SVar:TrigDamage:DB$ DealDamage | ValidTgts$ Any | NumDmg$ X
 SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
-SVar:X:Remembered.nonLand$SumCMC
+SVar:X:Count$Valid Remembered.nonLand$SumCMC
 Oracle:When you set this scheme in motion, reveal the top two cards of your library and put them into your hand. When you reveal one or more nonland cards this way, this scheme deals damage equal to their total mana value to any target.
--- a/forge-gui/res/cardsfolder/upcoming/a_premonition_of_your_demise.txt
+++ b/forge-gui/res/cardsfolder/upcoming/a_premonition_of_your_demise.txt
@@ -1,0 +1,10 @@
+Name:A Premonition of Your Demise
+ManaCost:no cost
+Types:Scheme
+T:Mode$ SetInMotion | ValidCard$ Card.Self | Execute$ TrigRepeatEach | TriggerZones$ Command | TriggerDescription$ When you set this scheme in motion, for each card type among permanents your opponents control, you may put a permanent card of that type from your hand onto the battlefield.
+SVar:DBDig:DB$ Dig | DigNum$ 2 | Reveal$ True | ChangeNum$ All | ChangeValid$ Card | DestinationZone$ Hand | RememberChanged$ True | SubAbility$ DBImmediateTrigger
+SVar:DBImmediateTrigger:DB$ ImmediateTrigger | ConditionDefined$ Remembered | ConditionPresent$ Card.nonLand | Execute$ TrigDamage | SpellDescription$ When you reveal one or more nonland cards this way, this scheme deals damage equal to their total mana value to any target.
+SVar:TrigDamage:DB$ DealDamage | ValidTgts$ Any | NumDmg$ X | SubAbility$ DBCleanup
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+SVar:X:Remembered.nonLand$SumCMC
+Oracle:When you set this scheme in motion, reveal the top two cards of your library and put them into your hand. When you reveal one or more nonland cards this way, this scheme deals damage equal to their total mana value to any target.
--- a/forge-gui/res/cardsfolder/upcoming/a_premonition_of_your_demise.txt
+++ b/forge-gui/res/cardsfolder/upcoming/a_premonition_of_your_demise.txt
@@ -3,8 +3,8 @@ ManaCost:no cost
 Types:Scheme
 T:Mode$ SetInMotion | ValidCard$ Card.Self | Execute$ TrigRepeatEach | TriggerZones$ Command | TriggerDescription$ When you set this scheme in motion, reveal the top two cards of your library and put them into your hand. When you reveal one or more nonland cards this way, this scheme deals damage equal to their total mana value to any target.
 SVar:DBDig:DB$ Dig | DigNum$ 2 | Reveal$ True | ChangeNum$ All | ChangeValid$ Card | DestinationZone$ Hand | RememberChanged$ True | SubAbility$ DBImmediateTrigger
-SVar:DBImmediateTrigger:DB$ ImmediateTrigger | ConditionDefined$ Remembered | ConditionPresent$ Card.nonLand | Execute$ TrigDamage | SpellDescription$ When you reveal one or more nonland cards this way, this scheme deals damage equal to their total mana value to any target.
-SVar:TrigDamage:DB$ DealDamage | ValidTgts$ Any | NumDmg$ X | SubAbility$ DBCleanup
+SVar:DBImmediateTrigger:DB$ ImmediateTrigger | ConditionDefined$ Remembered | ConditionPresent$ Card.nonLand | Execute$ TrigDamage | SubAbility$ DBCleanup | SpellDescription$ When you reveal one or more nonland cards this way, this scheme deals damage equal to their total mana value to any target.
+SVar:TrigDamage:DB$ DealDamage | ValidTgts$ Any | NumDmg$ X
 SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
 SVar:X:Remembered.nonLand$SumCMC
 Oracle:When you set this scheme in motion, reveal the top two cards of your library and put them into your hand. When you reveal one or more nonland cards this way, this scheme deals damage equal to their total mana value to any target.
--- a/forge-gui/res/cardsfolder/upcoming/a_premonition_of_your_demise.txt
+++ b/forge-gui/res/cardsfolder/upcoming/a_premonition_of_your_demise.txt
@@ -1,7 +1,7 @@
 Name:A Premonition of Your Demise
 ManaCost:no cost
 Types:Scheme
-T:Mode$ SetInMotion | ValidCard$ Card.Self | Execute$ TrigRepeatEach | TriggerZones$ Command | TriggerDescription$ When you set this scheme in motion, for each card type among permanents your opponents control, you may put a permanent card of that type from your hand onto the battlefield.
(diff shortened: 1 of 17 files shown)


forge-gui/res/cardsfolder/upcoming/a_premonition_of_your_demise.txt
```diff
@@ -0,0 +1,10 @@
+Name:A Premonition of Your Demise
+ManaCost:no cost
+Types:Scheme
+T:Mode$ SetInMotion | ValidCard$ Card.Self | Execute$ TrigRepeatEach | TriggerZones$ Command | TriggerDescription$ When you set this scheme in motion, for each card type among permanents your opponents control, you may put a permanent card of that type from your hand onto the battlefield.
```
tool4ever, line 4: wrong description

forge-gui/res/cardsfolder/upcoming/a_premonition_of_your_demise.txt
```diff
@@ -0,0 +1,10 @@
+Name:A Premonition of Your Demise
+ManaCost:no cost
+Types:Scheme
+T:Mode$ SetInMotion | ValidCard$ Card.Self | Execute$ TrigRepeatEach | TriggerZones$ Command | TriggerDescription$ When you set this scheme in motion, for each card type among permanents your opponents control, you may put a permanent card of that type from your hand onto the battlefield.
+SVar:DBDig:DB$ Dig | DigNum$ 2 | Reveal$ True | ChangeNum$ All | ChangeValid$ Card | DestinationZone$ Hand | RememberChanged$ True | SubAbility$ DBImmediateTrigger
+SVar:DBImmediateTrigger:DB$ ImmediateTrigger | ConditionDefined$ Remembered | ConditionPresent$ Card.nonLand | Execute$ TrigDamage | SpellDescription$ When you reveal one or more nonland cards this way, this scheme deals damage equal to their total mana value to any target.
+SVar:TrigDamage:DB$ DealDamage | ValidTgts$ Any | NumDmg$ X | SubAbility$ DBCleanup
```
tool4ever, line 7: cleanup should happen after creating

forge-gui/res/cardsfolder/upcoming/a_premonition_of_your_demise.txt
```diff
@@ -0,0 +1,10 @@
+Name:A Premonition of Your Demise
+ManaCost:no cost
+Types:Scheme
+T:Mode$ SetInMotion | ValidCard$ Card.Self | Execute$ TrigRepeatEach | TriggerZones$ Command | TriggerDescription$ When you set this scheme in motion, for each card type among permanents your opponents control, you may put a permanent card of that type from your hand onto the battlefield.
+SVar:DBDig:DB$ Dig | DigNum$ 2 | Reveal$ True | ChangeNum$ All | ChangeValid$ Card | DestinationZone$ Hand | RememberChanged$ True | SubAbility$ DBImmediateTrigger
+SVar:DBImmediateTrigger:DB$ ImmediateTrigger | ConditionDefined$ Remembered | ConditionPresent$ Card.nonLand | Execute$ TrigDamage | SpellDescription$ When you reveal one or more nonland cards this way, this scheme deals damage equal to their total mana value to any target.
+SVar:TrigDamage:DB$ DealDamage | ValidTgts$ Any | NumDmg$ X | SubAbility$ DBCleanup
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+SVar:X:Remembered.nonLand$SumCMC
```
tool4ever, line 9: was this line tested?

Fulgur14: No, I mean, technically nonLand is not even needed there since lands have mana value 0, it was just futureproofing.

tool4ever: well still seems like invalid syntax...?

Fulgur14: ok, so how should it look?

forge-gui/res/cardsfolder/upcoming/my_followers_ascend.txt
```diff
@@ -0,0 +1,10 @@
+Name:My Followers Ascend
+ManaCost:no cost
+Types:Scheme
+T:Mode$ SetInMotion | ValidCard$ Card.Self | Execute$ TrigBranch| TriggerZones$ Command | TriggerDescription$ When you set this scheme in motion, put five +1/+1 counters on a creature you control and it gains flying and vigilance until end of turn. If you control no creatures, instead create a 4/4 colorless Scarecrow artifact creature token with vigilance.
+SVar:TrigBranch:DB$ Branch | BranchConditionSVar$ X | BranchConditionSVarCompare$ EQ0 | TrueSubAbility$ DBToken | FalseSubAbility$ DBPutCounter
+SVar:DBToken:DB$ Token | TokenAmount$ 1 | TokenScript$ a_4_4_scarecrow_vigilance | TokenOwner$ You
+SVar:DBPutCounter:DB$ PutCounter | Choices$ Creature.YouCtrl | CounterType$ P1P1 | CounterNum$ 5 | SubAbility$ DBPump
+SVar:DBPump:DB$ Pump | Defined$ ChosenCard | KW$ Flying & Vigilance
```
tool4ever, line 8: invalid defined

forge-gui/res/cardsfolder/upcoming/only_i_know_what_awaits.txt
```diff
@@ -0,0 +1,9 @@
+Name:Only I Know What Awaits
+ManaCost:no cost
+Types:Scheme
+T:Mode$ SetInMotion | ValidCard$ Card.Self | Execute$ TrigRepeatEach | TriggerZones$ Command | TriggerDescription$ When you set this scheme in motion, for each card type among permanents your opponents control, you may put a permanent card of that type from your hand onto the battlefield.
+SVar:DBRepeatEach:DB$ RepeatEach | RepeatTypesFrom$ Permanent.OppCtrl | RepeatSubAbility$ ChooseCard | SubAbility$ DBChangeZoneAll
+SVar:ChooseCard:DB$ ChooseCard | Choices$ Card.YouOwn+ChosenType | RememberChosen$ True | ChoiceTitle$ You may choose a card of this type to put onto the battlefield | ChoiceTitleAppend$ ChosenType | ChoiceZone$ Hand
```
tool4ever, line 6: no need to use extra remember

tool4ever: can be cleaned from L8 too

forge-gui/res/cardsfolder/upcoming/a_premonition_of_your_demise.txt
tool4ever: Hmn I don't think you can filter a value with Defined via Valid and then apply another function
It would have to be Count$Valid I guess?

Fulgur14: I think I'm a bit lost... is this right?

tool4ever: nope, anyway now most cases should work 🤷‍♂️

Fulgur14: Thanks :)

forge-gui/res/cardsfolder/upcoming/my_followers_ascend.txt
```diff
@@ -0,0 +1,11 @@
+Name:My Followers Ascend
+ManaCost:no cost
+Types:Scheme
+T:Mode$ SetInMotion | ValidCard$ Card.Self | Execute$ TrigBranch| TriggerZones$ Command | TriggerDescription$ When you set this scheme in motion, put five +1/+1 counters on a creature you control and it gains flying and vigilance until end of turn. If you control no creatures, instead create a 4/4 colorless Scarecrow artifact creature token with vigilance.
+SVar:TrigBranch:DB$ Branch | BranchConditionSVar$ X | BranchConditionSVarCompare$ EQ0 | TrueSubAbility$ DBToken | FalseSubAbility$ DBPutCounter
+SVar:DBToken:DB$ Token | TokenAmount$ 1 | TokenScript$ a_4_4_scarecrow_vigilance | TokenOwner$ You
```
paulsnoops, line 6: Please use: c_4_4_a_scarecrow_vigilance

forge-gui/res/cardsfolder/upcoming/cathartic_parting.txt
```diff
@@ -0,0 +1,6 @@
+Name:Cathartic Parting
+ManaCost:1 G
+Types:Sorcery
+A:SP$ ChangeZone | Origin$ Battlefield | Destination$ Library | ValidTgts$ Artifact.OppCtrl,Enchantment.OppCtrl | TgtPrompt$ Select target artifact or enchantment an opponent controls | SubAbility$ DBChangeZone | Shuffle$ True | SpellDescription$ The owner of target artifact or enchantment an opponent controls shuffles it into their library. You may shuffle up to four target cards from your graveyard into your library.
+SVar:DBChangeZone:DB$ ChangeZone | Origin$ Graveyard | Destination$ Library | ValidTgts$ Card.YouOwn | TgtPrompt$ Select target card from your graveyard | TargetMin$ 0 | TargetMax$ 4 | Shuffle$ True
```
tool4ever, line 5: Seems like the "may" still needs to be addressed:
But should it allow skipping of only some of the targets?
Or do you think rather just offer to not shuffle at all?

Fulgur14: I've noticed that problem, Release notes just say
"You may target zero cards in your graveyard and still choose to shuffle your library."

The card I used as template (Devious Cover-Up) had the same problem.

tool4ever: I'll fix it in a follow-up

forge-gui/res/cardsfolder/upcoming/you_cannot_hide_from_me.txt
```diff
@@ -0,0 +1,12 @@
+Name:You Cannot Hide from Me
+ManaCost:no cost
+Types:Ongoing Scheme
+T:Mode$ Phase | Phase$ BeginCombat | ValidPlayer$ You | TriggerZones$ Command | Execute$ TrigPump | TriggerDescription$ At the beginning of combat on your turn, up to one target creature gets +2/+2 and gains vigilance until end of turn. It can't be blocked this turn.
+SVar:TrigPump:DB$ Pump | ValidTgts$ Creature | TargetMin$ 0 | TargetMax$ 1 | TgtPrompt$ Select up to one target creature | NumAtt$ 2 | NumDef$ 2 | KW$ Vigilance | SubAbility$ DBEffect
+SVar:DBEffect:DB$ Effect | ExileOnMoved$ Battlefield | RememberObjects$ ParentTarget | StaticAbilities$ Unblockable
+SVar:Unblockable:Mode$ CantBlockBy | ValidAttacker$ Card.IsRemembered | Description$ This creature can't be blocked this turn.
+T:Mode$ Phase | Phase$ End Of Turn | ValidPlayer$ Opponent | Execute$ Abandon | TriggerZones$ Command | CheckSVar$ X | SVarCompare$ LTY | TriggerDescription$ At the beginning of each opponent's end step, if that player has less than half their starting life total, abandon this scheme.
+SVar:Abandon:DB$ Abandon
+SVar:X:TriggeredPlayer$LifeTotal
```
tool4ever, line 10: imo this seems untested because you're trying to use triggeringsobjects while checking the condition

Fulgur14: The schemes were generally so straining that I had no energy left for testing :) Their effects are usually novel and complicated.

tool4ever: uhh...do you mean it's better if I fix it?

Fulgur14: Probably, I'm not sure how :(

tool4ever: I think you could just use PlayerCountDefinedActivePlayer$ so it's not reliant on the trigger

Fulgur14: OK, and in this case, even theoretical later implementation of shared turns wouldn't break it.

tool4ever (approved):

PR #6127, opened 2024-09-11
Add XstartOnFirstThread to Mac runner

kholbekj:
I know very little about java development, but this seems to at least be a workaround for the currently not launchable adventure file on mac.


kevlahnota:
the script is already added last year, and on my PR #6103 , the script will be automatically added...
image

PR #6140, opened 2024-09-13
Update ogre_battlecaster.txt

dracontes:
Following up on a report that [Ogre Battlecaster](https://scryfall.com/card/j22/36/ogre-battlecaster) wasn't getting the power buff after targeted card was cast. Turns out a `Cleanup` happened a bit too early. Tested the fix to satisfaction, including X spells.

### DIFF
--- a/forge-gui/res/cardsfolder/o/ogre_battlecaster.txt
+++ b/forge-gui/res/cardsfolder/o/ogre_battlecaster.txt
@@ -3,12 +3,12 @@ ManaCost:2 R
 Types:Creature Ogre Shaman
 PT:3/3
 K:First Strike
-T:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigDT | TriggerDescription$ Whenever CARDNAME attacks, you may cast target instant or sorcery card from your graveyard by paying {R}{R} in addition to its other costs. If that spell would be put into a graveyard, exile it instead. When you cast that spell, CARDNAME gets +X/+0 until end of turn, where X is that spell's mana value.
-SVar:TrigDT:DB$ DelayedTrigger | ValidTgts$ Instant.YouOwn,Sorcery.YouOwn | TgtZone$ Graveyard | TgtPrompt$ Select target instant or sorcery card | Mode$ SpellCast | ValidCard$ Card.IsTriggerRemembered | Execute$ TrigPump | RememberObjects$ Targeted | SubAbility$ DBPlay | TriggerDescription$ When you cast that spell, CARDNAME gets +X/+0 until end of turn, where X is that spell's mana value.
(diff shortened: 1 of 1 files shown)


forge-gui/res/cardsfolder/o/ogre_battlecaster.txt
```diff
@@ -5,9 +5,9 @@ PT:3/3
 K:First Strike
 T:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigDT | TriggerDescription$ Whenever CARDNAME attacks, you may cast target instant or sorcery card from your graveyard by paying {R}{R} in addition to its other costs. If that spell would be put into a graveyard, exile it instead. When you cast that spell, CARDNAME gets +X/+0 until end of turn, where X is that spell's mana value.
 SVar:TrigDT:DB$ DelayedTrigger | ValidTgts$ Instant.YouOwn,Sorcery.YouOwn | TgtZone$ Graveyard | TgtPrompt$ Select target instant or sorcery card | Mode$ SpellCast | ValidCard$ Card.IsTriggerRemembered | Execute$ TrigPump | RememberObjects$ Targeted | SubAbility$ DBPlay | TriggerDescription$ When you cast that spell, CARDNAME gets +X/+0 until end of turn, where X is that spell's mana value.
-SVar:DBPlay:DB$ Play | Defined$ Targeted | ValidSA$ Spell | ReplaceGraveyard$ Exile | PlayRaiseCost$ R R | Optional$ True | SubAbility$ DBCleanup
+SVar:DBPlay:DB$ Play | Defined$ Targeted | ValidSA$ Spell | ReplaceGraveyard$ Exile | PlayRaiseCost$ R R | Optional$ True
+SVar:TrigPump:DB$ Pump | Defined$ Self | NumAtt$ +X | SubAbility$ DBCleanup
```
tool4ever, line 9: the point of the previous order was to get rid of the trigger if you didn't cast it
otherwise it would still be around and potentially apply for e.g. a flashback cast

we need to think/debug what broke there

tool4ever: Caused by #4794
but that can't be reverted easily, I think a different workaround is needed here now

tool4ever: new approach should work :p

tool4ever (approved):

forge-gui/res/cardsfolder/o/ogre_battlecaster.txt
dracontes: Indeed: passes the testing I threw at the previous iteration of this fix. Much obliged for catching the issues with it👍 
I did think a rewrite could be what this one needed but opted to first test what seemed to be a simpler fix.

PR #6175, opened 2024-09-18
Allows for additional cards to be added to worlds.txt in quest mode.

repeperilka:
By writing "Extra:cardname@set;cardname@set..." you can get extra cards not found in the previously given sets into worlds in quest mode. When there are no allowed sets or additional cards, the world shop and pool will behave as it did before (It'll give cards from all sets), and if there are no allowed sets but there are additional cards, all of the cards shown on the store and initial card pool will only be composed of the additional cards specified.

To accommodate for this, the constructors for GameFormat and GameFormatQuest have been changed together with its filters and the return value of getAllCards() and when checking if a set is allowed in the format, the result will depend on if the format has additional cards.
Colored booster pack generation accounts for additional cards when building its filters
Added random colored boosters as win rewards called "Chaos" plus the color.
The function for generating singles in the shop now uses the same system as the initial card pool for starting a quest because it's faster.


forge-gui/res/quest/world/worlds.txt
```diff
@@ -37,4 +37,5 @@ Name:Kaladesh|Sets:AER, KLD, MPS, KLR
...
...ade;Tortured Existence;Goblin Motivator;Harried Spearguard;Reckless Lackey;Barbed Batterfist;Hobblefiend;Rally at the Hornburg;Ratcatcher Trainee;Red Herring;Volatile Wanderglyph;Yavimaya Steelcrusher;Spelleater Wolverine;Spellgorger Weird;Tuskeri Firewalker;Chimney Rabble;Wildfire Elemental;Improvised Club;Ambitious Assault;Gnawing Crescendo;Sulfurous Blast;Wrangle;Wrenn's Resolve;Raid Bombardment;Bayou Groff;Gnarlid Colony;Gnoll Hunter;Jade Avenger;Thornweald Archer;Voracious Varmint;Wary Thespian;Werebear;Contagious Vorrac;Deepwood Denizen;Phantom Tiger;Tuskguard Captain;Bloom Hulk;Jade Guardian;Urban Daggertooth;Voracious Typhon;Silverback Shaman;Greater Tanuki;Dread Linnorm;Massive Might;Tamiyo's Safekeeping;Grapple with the Past;Master's Rebuke;Prizefight;Clear Shot;Edge of Autumn;You Meet in a Tavern;Wild Growth;Silver Drake;Body Dropper;Goblin Anarchomancer;Goblin Electromancer;Infuse with Vitality;Dog Walker;War Flare;Iron Apprentice;Filigree Familiar;Skyscanner;Juggernaut;Candy Trail;Leonin Bola;Prophetic Prism;Star Compass;Jagged Barrens;Bristling Backwoods;Eroded Canyon;Abraded Bluffs;Basilisk Gate;Conduit Pylons;Crystal Grotto;Quicksand;Shire Terrace;Tocasia's Dig Site
```
tehdiplomat, line 41: This.. seriously explodes this file. Can we look into doing this a different way? Maybe having worlds be individual files, and have them have sections kinda like an Edition? Sets, Cards, etc.

repeperilka (commented):
I don't know what I did here.
Hope I didn't mess anything up.

repeperilka:
Just found out the changes I made break the spell shop for the Random Commander world

PR #6183, opened 2024-09-21
Reorder card entries in 111 edition files.

Ayora29:
No card add, no card remove. Only reorder, numerically then alphabetically.


paulsnoops:
Moving collector numbers starting with a letter doesn't always make sense. Some that start with F*** should stay in the order as is. For example: SLD F798 officially is collector number 798; the F is only used for internal purposes to let Forge know it's a "funny" card.

Have you tested for breaking any image file name numbering? What I mean is for example:
(I'm making this up not using a real edition as I haven't looked through them all)
1 Fireball (Fireball1.fullborder)
35 Fireball (Fireball2.fullborder)
1★ Fireball (Fireball3.fullborder)

If you were to reorder 1★ to be next to 1 such as (which to be fair does make sense):
1 Fireball
1★ Fireball
35 Fireball

Then Forge would show Fireball2.fullborder as 1★ and Fireball3.fullborder as 35, which would show the images the wrong way around. This would really apply to any card that has the same CardName where the collector number order has changed.

Ayora29:
Hello @paulsnoops ,

Ok I see, thank you for your feedback. I 'm going to do further testing and come back.

Ayora29:
Hello @paulsnoops 
I left the "F collector numbers" as they were.

I ran a script to detect multiple copies in the same set of card I have re-order.
I did find some arts that are not correctly displayed (according to scryfall) : 
Mishra's Factory (ATQ)
Strip Mine (ATQ)
Urza's Power Plant (ATQ)
Urza's Tower (ATQ)
But it seems the arts were already not the good ones before my re-order (still according to scryfall). If you can confirm?

paulsnoops:
> Hello @paulsnoops I left the "F collector numbers" as they were.
> 
> I ran a script to detect multiple copies in the same set of card I have re-ordered. I did find some arts that are not correctly displayed (according to scryfall) : Mishra's Factory (ATQ) Strip Mine (ATQ) Urza's Power Plant (ATQ) Urza's Tower (ATQ) But it seems the arts were already not the good ones before my re-order (still according to scryfall). If you can confirm?

Yeah go ahead and keep this change, it's not a big deal to have one outlier

Ayora29:
@paulsnoops Ok great. Do you need anything else for approving the PL?

paulsnoops:
> @paulsnoops Ok great. Do you need anything else for approving the PL?

I will try and review it properly at some point later this week, unless someone else wanted to before I get a chance, thanks for your work

paulsnoops (approved):

paulsnoops:
@Ayora29 looks good thanks, any problems we can fix at a later date. I'll fix those ATQ images on the ftp this weekend if they aren't lining up

PR #6196, opened 2024-09-25
Fix valiant cards only triggering once per game rather than once per turn

Ampersandnz:
Copies the cleanup logic for `becameTargetThisTurn`

FLUP to #6178


tool4ever (approved):
oops. thanks for the catch

Issue #6200, opened 2024-09-26
Alphabetically sort MTGO Cubes so they are easier to find


myyk:
**Is your feature request related to a problem? Please describe.**
I often like to draft the latest MTGO Cube and I have trouble finding it on the list because the format like "August 2024" and it leave the drafts kind of scrambled.

**Describe the solution you'd like**
I would like to look at the list and easily see order so I can find the draft I want to do. Since they are largely just "Month YYYY" it would sort better as "YYYY-MM"

**Describe alternatives you've considered**
I guess it's possible someone else wants it sorted like it is by everything grouped by the month it came out in whatever year. It's bizarre though as even MTGO has shifted their update schedule over the years.

myyk:
PR is still ready to merge, pending approval.

myyk:
Please don't auto-close this, it's been ready for months.

PR #6201, opened 2024-09-26
 Alphabetically sort MTGO Cubes so they are easier to find

myyk:
Implements #6200.

I have one commit as moves and one commit as edits for easier reviewing. It's too bad that it didn't show them as moves for all the `.draft` files in github, but that's out of my control.


tehdiplomat:
Is this going to cause issues for people who are in between gauntlets, or have their preferences pointed to the old names?

myyk:
@tehdiplomat, I'm unfamiliar with Gauntlet, but as far as I can tell from playing with the UI it doesn't seem related to draft or cube so I think those are totally separate and safe if in progress.

As far as drafted decks from a cube go, the decks seem unrelated to the cube they were drafted from. And there's no in progress saving of a draft, so nothing to worry about from that front.

tehdiplomat (dismissed):

forge-gui/res/cube/MTGO Legacy Cube 2017-01.dck
```diff
@@ -1,5 +1,5 @@
 [metadata]
-Name=MTGO Legacy Cube January 2017
+Name=MTGO Vintage Cube 2017-01
```
myyk, line 2: Oops, I'll fix this

myyk:
@myyk I triple checked all changes after fixing the 1 typo I found.

Agetian:
Is this MR ready?

myyk:
It is

Agetian (approved):

PR #6219, opened 2024-09-30
Edit pile cleanup: part 2, folders 'h' to 'q'

dracontes:


forge-gui/res/cardsfolder/h/hierophant_bio_titan.txt
```diff
@@ -5,8 +5,8 @@ PT:12/12
 K:Vigilance
 K:Reach
 K:Ward:2
-A:SP$ PermanentCreature | Cost$ 10 G G RemoveAnyCounter<X/P1P1/Creature.YouCtrl/among creatures you control>
-S:Mode$ ReduceCost | ValidCard$ Card.Self | Type$ Spell | Amount$ Y | EffectZone$ All | Relative$ True | Description$ Frenzied Metabolism — As an additional cost to cast this spell, you may remove any number of +1/+1 counters from among creatures you control. This spell costs {2} less to cast for each counter removed this way.
+A:SP$ PermanentCreature | Cost$ 10 G G RemoveAnyCounter<X/P1P1/Creature.YouCtrl/any number of creatures> | AdditionalDesc$ This spell costs {2} less to cast for each counter removed this way.
```
tool4ever, line 8: the type description seems wrong?
also PrecostDesc missing

dracontes: The issue is the automatic text for the Card Detail and the top paragraph on the Prompt add "you control" by default. On the other hand the generated text for the second Prompt paragraph doesn't do this. So fixing the latter by way of type description would break the other two.

image

As for `PrecostDesc`, it currently doesn't work for these spell abilities. Still I can add it so you can fix it later. For convenience, here's my attempt at adding it to A-Cobbled Lancer.

image

[a-cobbled_lancer_b.txt](https://github.com/user-attachments/files/17270094/a-cobbled_lancer_b.txt)

tool4ever: then just "among creatures" would be closer? 🤔
I'll try and check the PreCost as follow-up

dracontes: Given the behavior, my preference for the second paragraph would be "Choose a creature you control to remove a [counter kind] counter from it. ([N] left. You can choose the same permanent more than once.)" Though, working with what we have, fair enough.

Issue #5916, opened 2024-08-12
Inconsistent plurals in costs


dracontes:
**Describe the bug**
Automatic text generation for display of costs in-game has trouble with _some_ complex costs. A few examples:
- tapping more than one untapped permanent for an activated ability ([Adaptive Gemguard](https://scryfall.com/card/lci/3/adaptive-gemguard))
![image](https://github.com/user-attachments/assets/bdf55709-8713-4771-8d85-c747111d7347)
→ removing the baked-in description as it worked int the case of Ulamog the Defiler, only results in revealing the argument for the cost (`Artifact;Creature`).
- discarding more than one card as a mandatory additional cost ([Abandon Hope](https://scryfall.com/card/tmp/107/abandon-hope))
![image](https://github.com/user-attachments/assets/aa704285-8e44-4f6e-8e7a-f4f4ee99fffb)
→ changing the baked-in description doesn't do anything to the displayed text, though comparing with [Aether Tide](https://scryfall.com/card/exo/27/aether-tide), I take it the additional cost should be structured differently.
- discarding more than one card for an activated ability ([Advanced Stitchwing](https://scryfall.com/card/2x2/36/advanced-stitchwing))
![image](https://github.com/user-attachments/assets/a95b165d-9526-41f9-9adb-fb5738af0837)
→ As for Abandon Hope, changing the baked-in description (or adding it) doesn't do anything to the displayed text.
- sacrificing more than one permanent for an activated ability ([Akul the Unrepentant](https://scryfall.com/card/otj/189/akul-the-unrepentant))
![image](https://github.com/user-attachments/assets/904f27b5-3448-4238-9bab-537a2f2d16c2)

I could go on but you get the gist.

**To Reproduce**
1. Start a regular game and go quickly through the initial checks.
2. Then use Dev Mode to put the relevant cards in hand. Hovering the mouse cursor over them should reveal any issues with the automatic text on the Card Detail pane.

**Expected behavior**
That the plurals would be more consistent. At the very least this should mean that baked-in descriptions when provided shouldn't be wholly ignored or entrained into the automatic generation process that's used on simpler arguments for costs, which admittedly has humorous outcomes. On the other hand, the automatic generation process could be improved. I do reckon that the `Card` case for Abandon Hope and the Stitchwing should be straightforward.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version: `1.6.64-SNAPSHOT-08-05`

kevlahnota:
[...]

PR #6220, opened 2024-09-30
Edit pile cleanup: part 3, folders 'r' to 'u' & 'rebalanced'

dracontes:


forge-gui/res/cardsfolder/r/relentless_skaabs.txt
```diff
@@ -3,5 +3,5 @@ ManaCost:3 U U
 Types:Creature Zombie
 PT:4/4
 K:Undying
-A:SP$ PermanentCreature | Cost$ 3 U U ExileFromGrave<1/Creature>
+A:SP$ PermanentCreature | Cost$ 3 U U ExileFromGrave<1/Creature> | SpellDescription$ As an additional cost to cast this spell, exile a creature card from your graveyard.
```
tool4ever, line 6: why are we adding it if the autogeneration works fine?

dracontes: The automatic text doesn't show up in full on the Prompt, and I'd rather err on the side of clarity, especially in those admittedly rarer cases where mana costs aren't paid and casting is mandatory.
image

tool4ever: Hmn, can't say I like reproducing this (especially in the script)
I'll try and check if it can be included from the code instead

dracontes: Fair. I'll be checking the gnarlier variants for issues and address what I can.

tool4ever: check #6233
this should let you avoid these changes?

dracontes: So far, I feel like the issues I'm coming across are an extension of what I've noted here, https://github.com/Card-Forge/forge/issues/5916 , that time for non-permanent spells and activated abilities, 
image.

dracontes: > check #6233 this should let you avoid these changes?

To be clear, the above was done without reference to your fix, though I'll be sure to test it as soon as possible.

dracontes: Preliminary fix testing done: works fine indeed. While reverting the simpler additional cost variants should be easy enough, the more complex variants do need a closer look.
image

tool4ever: ok, I've changed it so can you can still use the override with SpellDescription if there's a really exotic case
however usually the changes should rather be made/fixed inside the part description handling imo
it's also worth mentioning a newer approach without manual spell definition (see #2472) could fix some more cleanly (though your desired full text in prompt might not happen then)

dracontes: Fair. I'll do my best to do as you recommend 👍

dracontes: As a confounding contrast to [Redemptor Dreadnought](https://scryfall.com/card/40k/164/redemptor-dreadnought) ( [card script](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/r/redemptor_dreadnought.txt) ), I came across [Graven Archfiend](https://scryfall.com/card/ysnc/7/graven-archfiend) ( [card script](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/g/graven_archfiend.txt) ), which uses `S:Mode$ OptionalCost` for its optional additional cost instead. The issue with the Dreadnought is the text on the Prompt is fairly off base, while the Archfiend works like kicker does. 

image

So, should the optional additional costs be changed to static abilities or is there a reason why `A:SP$ PermanentCreature` is the better choice the Dreadnought?

tool4ever: _Redemptor Dreadnought_ was written at a time OptionalCost static didn't exist yet
it could be updated now to improve the text (while I would prefer if it could still be autogenerated in that case that's something for later)
I believe some cards like _Gorex_ are still impossible in that fashion though

PS:
It looked to me like the remaining parts of this run all had at least 1 script that was changed in this way, so I was going to wait until we have an update approach that can be applied on all

dracontes: Duly noted 👍

forge-gui/res/cardsfolder/u/unexpected_allies.txt
```diff
@@ -5,5 +5,5 @@ A:SP$ Pump | ValidTgts$ Creature.YouCtrl+nonToken | TgtPrompt$ Select target non
 SVar:TrigEffect:DB$ Pump | ConditionCheckSVar$ X | Defined$ Targeted | KW$ First Strike
 SVar:X:Count$Valid Creature.NotDefinedTargeted+YouCtrl+sharesNameWith Targeted/Plus.Y
 SVar:Y:Count$ValidGraveyard Creature.YouOwn+sharesNameWith Targeted
-DeckHas:Keyword$Double Team|First Strike
+DeckHas:Keyword$DoubleTeam|FirstStrike
```
tool4ever, line 8: why are the spaces getting removed?

forge-gui/res/cardsfolder/r/relentless_skaabs.txt
tool4ever: Oh, I see the Dreadnought would also display its flavor words if it gets changed :)

tool4ever: Uhm anyway, I'm just gonna try and rephrase since I feel I might not have been clear? 😅
The cards don't need to be updated identical as of yet. Just try and get rid of any "duplicated" text that my PR would display now instead - any bad phrasing that still occurs should really be fixed in code since it would still display wrong in card detail imo (but I leave that to your discretion if it doesn't collide with anything else).

Then I can merge all together...

forge-gui/res/cardsfolder/u/unexpected_allies.txt
dracontes: At least for first strike, I started my edit from a situation where about half of the mentions in the context of `Deck*:` don't have spaces. I had a quick look over the code and it seemed like that was accounted for first strike, double strike and double team. I take it the look I gave it was too quick?

forge-gui/res/cardsfolder/r/relentless_skaabs.txt
dracontes: I'm certainly leaving the second paragraph on the Prompt for later.

forge-gui/res/cardsfolder/u/unexpected_allies.txt
tool4ever: well it's good you spotted it
but in fact it seems like this was a misunderstanding or typo that has spread
it won't work if it's written wrong / without spaces

dracontes: Duly noted. I'll revert and propagate the newer fix where relevant.

dracontes: For Unexpected Allies, should it be "Double team" instead?

tool4ever: yea

tool4ever (approved):

PR #6223, opened 2024-09-30
Edit pile cleanup: part 6, folder 'upcoming' part 3

dracontes:


tool4ever (approved):
just line endings I suppose

dracontes:
> just line endings I suppose

Indeed. While the bulk of the other two remaining is also that, they do need a closer look. I'll check them again FWIW,

PR #6226, opened 2024-10-01
InTurnOrder parameter check for specific cards that force turn order of opponents

TwentyToedToad:
* Added `InTurnOrder` parameter check for case when getting opponents in turn order to [AbilityUtils.java](forge-game/src/main/java/forge/game/ability/AbilityUtils.java)
* Added `getOpponentsInTurnOrder()` for same case to [Player.java](forge-game/src/main/java/forge/game/player/Player.java)

Note: Not many cards require forced turn order of opponents. Honestly, I could only find three. Turn order breaks when AI activates the ability due to normally just getting the opponents of the player. With AI the human is always at the top of the players list from [Game.java](forge-game/src/main/java/forge/game/Game.java)'s `getPlayers()` method. I have not tested this in net play at all so I am not sure how players are ordered in Game.java in an online game. ~~If there was an issue before this should solve the problem if Manifold Insights and Protection Racket are updated with `InTurnOrder$ True`. I will be committing chard changes for both and the new card Sadistic Shell Game.~~

Added updated cards to PR.

[Manifold Insights](https://scryfall.com/card/cm2/44/manifold-insights)
[Protection Racket](https://scryfall.com/card/ncc/39/protection-racket)
[Sadistic Shell Game](https://scryfall.com/card/dsc/24/sadistic-shell-game)

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -1193,11 +1193,7 @@ public class AbilityUtils {
         } else if (defined.equals("You")) {
             players.add(player);
         } else if (defined.equals("Opponent")) {
-            if (sa.hasParam("InTurnOrder") && sa.getParam("InTurnOrder").equals("True")) {
-                players.addAll(player.getOpponentsInTurnOrder());
-            } else {
-                players.addAll(player.getOpponents());
-            }
+            players.addAll(player.getOpponents());
         } else if (defined.startsWith("NextPlayerToYour")) {
             Direction dir = defined.substring(16).equals("Left") ? Direction.Left : Direction.Right;
             players.add(game.getNextPlayerAfter(player, dir));
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -1193,7 +1193,11 @@ public class AbilityUtils {
         } else if (defined.equals("You")) {
             players.add(player);
         } else if (defined.equals("Opponent")) {
-            players.addAll(player.getOpponents());
+            if (sa.hasParam("InTurnOrder") && sa.getParam("InTurnOrder").equals("True")) {
+                players.addAll(player.getOpponentsInTurnOrder());
+            } else {
+                players.addAll(player.getOpponents());
+            }
         } else if (defined.startsWith("NextPlayerToYour")) {
             Direction dir = defined.substring(16).equals("Left") ? Direction.Left : Direction.Right;
             players.add(game.getNextPlayerAfter(player, dir));
--- a/forge-game/src/main/java/forge/game/player/Player.java
+++ b/forge-game/src/main/java/forge/game/player/Player.java
@@ -303,6 +303,10 @@ public class Player extends GameEntity implements Comparable<Player> {
         return game.getPlayers().filter(PlayerPredicates.isOpponentOf(this));
     }
 
+    public final PlayerCollection getOpponentsInTurnOrder() {
+        return game.getPlayersInTurnOrder(this).filter(PlayerPredicates.isOpponentOf(this));
+    }
+
     public final PlayerCollection getRegisteredOpponents() {
         return game.getRegisteredPlayers().filter(PlayerPredicates.isOpponentOf(this));
     }
--- a/forge-game/src/main/java/forge/game/player/Player.java
+++ b/forge-game/src/main/java/forge/game/player/Player.java
@@ -303,10 +303,6 @@ public class Player extends GameEntity implements Comparable<Player> {
         return game.getPlayers().filter(PlayerPredicates.isOpponentOf(this));
     }
 
-    public final PlayerCollection getOpponentsInTurnOrder() {
-        return game.getPlayersInTurnOrder(this).filter(PlayerPredicates.isOpponentOf(this));
-    }
-
     public final PlayerCollection getRegisteredOpponents() {
         return game.getRegisteredPlayers().filter(PlayerPredicates.isOpponentOf(this));
     }
--- a/forge-game/src/main/java/forge/game/player/Player.java
+++ b/forge-game/src/main/java/forge/game/player/Player.java
@@ -300,7 +300,7 @@ public class Player extends GameEntity implements Comparable<Player> {
      * Should keep player relations somewhere in the match structure
      */
     public final PlayerCollection getOpponents() {
-        return game.getPlayers().filter(PlayerPredicates.isOpponentOf(this));
+        return game.getPlayersInTurnOrder(this).filter(PlayerPredicates.isOpponentOf(this));
     }
 
     public final PlayerCollection getRegisteredOpponents() {
--- a/forge-gui/res/cardsfolder/m/manifold_insights.txt
+++ b/forge-gui/res/cardsfolder/m/manifold_insights.txt
@@ -2,7 +2,7 @@ Name:Manifold Insights
 ManaCost:2 U
 Types:Sorcery
 A:SP$ PeekAndReveal | PeekAmount$ 10 | RevealValid$ Card | RememberRevealed$ True | SubAbility$ ChooseNonLand | SpellDescription$ Reveal the top ten cards of your library. Starting with the next opponent in turn order, each opponent chooses a different nonland card from among them. Put the chosen cards into your hand and the rest on the bottom of your library in a random order.
-SVar:ChooseNonLand:DB$ RepeatEach | RepeatPlayers$ Player.Opponent | InTurnOrder$ True | RepeatSubAbility$ OpponentsChoose | SubAbility$ ShipToBottom
+SVar:ChooseNonLand:DB$ RepeatEach | RepeatPlayers$ Player.Opponent | StartingWith$ Opponent | RepeatSubAbility$ OpponentsChoose | SubAbility$ ShipToBottom
 SVar:OpponentsChoose:DB$ ChooseCard | ChoiceZone$ Library | Choices$ Card.nonLand+IsRemembered | Defined$ Player.IsRemembered | ForgetChosen$ True | SubAbility$ ChosenToHand
 SVar:ChosenToHand:DB$ ChangeZone | Origin$ Library | Destination$ Hand | Defined$ ChosenCard
 SVar:ShipToBottom:DB$ ChangeZoneAll | ChangeType$ Card.IsRemembered | Origin$ Library | Destination$ Library | LibraryPosition$ -1 | RandomOrder$ True | SubAbility$ DBCleanup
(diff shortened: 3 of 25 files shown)


forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -1193,7 +1193,11 @@ public static PlayerCollection getDefinedPlayers(final Card card, final String d
         } else if (defined.equals("You")) {
             players.add(player);
         } else if (defined.equals("Opponent")) {
-            players.addAll(player.getOpponents());
+            if (sa.hasParam("InTurnOrder") && sa.getParam("InTurnOrder").equals("True")) {
```
tool4ever, line 1196: this is a good catch but will require a wider approach imo as this wording seems to only emphasize the default anyway, e.g.:
> As Dredge the Mire resolves, first the next opponent in turn order (or, if it’s an opponent’s turn, the opponent whose turn it is) chooses a creature card in their graveyard, then each other opponent in turn order does the same knowing the choices made before them. Then all the chosen cards are returned at the same time.

forge-gui/res/cardsfolder/m/manifold_insights.txt
```diff
@@ -2,7 +2,7 @@ Name:Manifold Insights
 ManaCost:2 U
 Types:Sorcery
 A:SP$ PeekAndReveal | PeekAmount$ 10 | RevealValid$ Card | RememberRevealed$ True | SubAbility$ ChooseNonLand | SpellDescription$ Reveal the top ten cards of your library. Starting with the next opponent in turn order, each opponent chooses a different nonland card from among them. Put the chosen cards into your hand and the rest on the bottom of your library in a random order.
-SVar:ChooseNonLand:DB$ RepeatEach | RepeatPlayers$ Player.Opponent | RepeatSubAbility$ OpponentsChoose | SubAbility$ ShipToBottom
+SVar:ChooseNonLand:DB$ RepeatEach | RepeatPlayers$ Player.Opponent | InTurnOrder$ True | RepeatSubAbility$ OpponentsChoose | SubAbility$ ShipToBottom
```
tool4ever, line 5: here the extra param wouldn't work yet since the RepeatPlayers one uses a different syntax

I think what needs to happen instead is RepeatEachEffect calling `getDefinedPlayersOrTargeted` because there is already some logic to sort the players:

https://github.com/Card-Forge/forge/blob/4fcd6f577acffa7d64e0e99437bfe36896953b7b/forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java#L303-L306

so this just needs some fixing to make it work even if the AP isn't included

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -303,6 +303,10 @@ public final PlayerCollection getOpponents() {
         return game.getPlayers().filter(PlayerPredicates.isOpponentOf(this));
     }
 
+    public final PlayerCollection getOpponentsInTurnOrder() {
+        return game.getPlayersInTurnOrder(this).filter(PlayerPredicates.isOpponentOf(this));
```
tool4ever, line 307: did you test it?
looking at the code I'd expect it now respects ReverseTurnOrderEffect...
but that would still not return them based on where this player is seated 🤔

forge-gui/res/cardsfolder/upcoming/sadistic_shell_game.txt
```diff
@@ -0,0 +1,8 @@
+Name:Sadistic Shell Game
+ManaCost:4 B
+Types:Sorcery
+A:SP$ RepeatEach | RepeatPlayers$ Opponent | RepeatSubAbility$ DBChooseCard | SubAbility$ DBDestroy | InTurnOrder$ True | StackDescription$ SpellDescription | SpellDescription$ Starting with the next opponent in turn order, each player chooses a creature you don’t control. Destroy the chosen creatures.
+SVar:DBChooseCard:DB$ ChooseCard | Choices$ Creature | Reveal$ True | Defined$ Player.IsRemembered | ControlledByPlayer$ Player.Opponent | Mandatory$ True | ChoiceTitle$ Choose a creature to destroy | RememberChosen$ True | AILogic$ OppPreferred
```
tool4ever, line 5: ControlledByPlayer incorrect

forge-gui/res/cardsfolder/upcoming/sadistic_shell_game.txt
```diff
@@ -0,0 +1,8 @@
...
+SVar:DBDestroy:DB$ DestroyAll | ValidCards$ Card.Creature+IsRemembered | SubAbility$ DBCleanup | StackDescription$ Destroy the chosen creatures.
+SVar:DBCleanup:DB$ Cleanup | ClearChosenCard$ True | ClearRemembered$ True
+Oracle: Starting with the next opponent in turn order, each player chooses a creature you don’t control. Destroy the chosen creatures.
```
tool4ever, line 8: spacing

forge-gui/res/cardsfolder/upcoming/sadistic_shell_game.txt
```diff
@@ -0,0 +1,8 @@
+Name:Sadistic Shell Game
+ManaCost:4 B
+Types:Sorcery
+A:SP$ RepeatEach | RepeatPlayers$ Opponent | RepeatSubAbility$ DBChooseCard | SubAbility$ DBDestroy | InTurnOrder$ True | StackDescription$ SpellDescription | SpellDescription$ Starting with the next opponent in turn order, each player chooses a creature you don’t control. Destroy the chosen creatures.
```
tool4ever, line 4: that is actually incorrect since the choices should still be made by all players

looks like this card will be a bit complicated to get right, let me know if you'd rather I take over at any point

forge-game/src/main/java/forge/game/player/Player.java
TwentyToedToad: Yes. It returns them in appropriate turn order. That said, I will do some more testing just to be sure. I'll amend this comment with results.

forge-gui/res/cardsfolder/upcoming/sadistic_shell_game.txt
TwentyToedToad: I'm not sure what you mean. Each opponent makes a choice and then the cards are destroyed. If you could clarify I can look into it a bit more.

forge-gui/res/cardsfolder/upcoming/sadistic_shell_game.txt
TwentyToedToad: I'll look at some other examples of `ControlledByPlayer$`

forge-gui/res/cardsfolder/m/manifold_insights.txt
TwentyToedToad: Understood. I really should have looked more into this individual card before committing a change for it.

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
TwentyToedToad: That is a very good catch. I did not like adding this InTurnOrder check to only one place (Opponent). I'll look into it more.

forge-gui/res/cardsfolder/upcoming/sadistic_shell_game.txt
TwentyToedToad: Addressed in recent [commit ](https://github.com/Card-Forge/forge/pull/6226/commits/261308bf219e657dff36908625ce7760906fa91a)

forge-gui/res/cardsfolder/upcoming/sadistic_shell_game.txt
TwentyToedToad: Addressed in recent [commit ](https://github.com/Card-Forge/forge/pull/6226/commits/261308bf219e657dff36908625ce7760906fa91a)

forge-game/src/main/java/forge/game/player/Player.java
TwentyToedToad: See [this google doc](https://docs.google.com/document/d/15iNuRxOa5xcXwVdCdrEt2RZqIjsF3CTDyZfjUPOy0d0/edit?usp=sharing)

All seems to be good. I can test more if you have any suggestions of cases to test.

forge-gui/res/cardsfolder/upcoming/sadistic_shell_game.txt
tool4ever: nope, that filtering is not what the card says
you should only use `Choices$ Creature.YouDontCtrl`

forge-gui/res/cardsfolder/upcoming/sadistic_shell_game.txt
tool4ever: it's not identical to _Manifold Insights_:
"each **player** chooses"
only who starts is defined differently than the rules default

forge-gui/res/cardsfolder/upcoming/sadistic_shell_game.txt
TwentyToedToad: Ah! Sorry. That makes much more sense. I will get better at this as I look into more cards and the code itself!

forge-gui/res/cardsfolder/upcoming/sadistic_shell_game.txt
TwentyToedToad: Oh man. I even played this card at my table the other night and we definitely did not let me choose. That is just a misread on mine and my pods part. I'll fix this right away.

forge-game/src/main/java/forge/game/player/Player.java
tool4ever: yea seems fine, thanks

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
tool4ever: - mb something like this to align the order instead?
```
// try sort in APNAP order
PlayerCollection sortedUnique = sa.getHostCard().getGame().getPlayersInTurnOrder(sa.getHostCard().getGame().getPhaseHandler().getPlayerTurn());
sortedUnique.retainAll(resultUnique);
return sortedUnique;
```
- the Opponent ordering in AbilityUtils might still be updated with the new default since quite a few places only call out to `getDefinedPlayers()` so a few corner cases could probably continue to hide for future cleanup

forge-gui/res/cardsfolder/m/manifold_insights.txt
tool4ever: no problem, all easy cards for this set have been finished for a while, so now this helps improving the engine :)

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
TwentyToedToad: @tool4ever I created a new commit that reverted all of the `InTurnOrder `stuff. I went with changing `StartingWithActivator ` to `StartingWith` and that new parameter can take `You `or `Oppponent`. If you would like me to create a new PR for these changes and remove this PR I can do so. Please see [commit](https://github.com/Card-Forge/forge/pull/6226/commits/b7e4640ba598d11c8762b3f76c34d69df74050b8).

tool4ever: that's fine, we usually squash merge if needed

tool4ever:
I refined it:
- can't always assume that ``getNextPlayerAfter`` returns an opponent on first call (instead of a teammate)
- other effects will benefit if the improved logic is made default
- this still leaves the problem of a huge number of places not passing through that part. I'm considering the "brute-force" approach and simply updating the main call in Player. We better test this a bit though...

tool4ever (approved):

PR #6230, opened 2024-10-01
Support second main

### LINKED ISSUE #6215: Second Main Phase cards interaction with cards that generate extra Main Phases
I'm not really sure if I should put this to Bug Report or Feature Request, so I'll just choose the latter.

There was a functionality change a while ago that the Second Main Phase will only occur once. Any further main phases beyond it will be 3rd, 4th, etc.

In the case that you have a card that triggers on the Second Main Phase and you can get additional Main Phases beyond the second one (like Aggravated Assault), those Main Phases will not trigger said cards.

It looks like the cards that trigger on the Second Main Phase can still trigger from extra main phases generated by Aggravated Assault. This should not be the case.

I tested this on Cautious Survivor, but I think the other cards with similar effects also do the same thing.

Also, World at War received a change from **"First Postcombat Main Phase"** to **"Second Main Phase"**. So I think it should also receive a similar functionality change.

https://www.reddit.com/r/magicTCG/comments/1e9wb5w/fyi_the_change_in_templating_from_postcombat_main/?share_id=s9s2uIA-oR4I54b6f5_RO&utm_content=1&utm_medium=android_app&utm_name=androidcss&utm_source=share&utm_term=1
image

### LINKED ISSUE #6217: Engine does not activate Illusionary Mask correctly
**Describe the bug**
Illusionary Mask does not bring creature to the battlefield as expected

**To Reproduce**
1. Illusionary Mask in play with creature card in hand with Mana Cost 3, during main phase
2. Click on Illusionary Mask, choose X=3, tap 3 mana matching the mana cost of the creature in hand.
3. Click on creature card in hand 
4. Nothing happens.

**Expected behavior**
The creature should be placed in play face down as a 2/2 creature.

 - OS: Windows 10
 - Version 1.6.64

tool4ever:
Closes #6215
Closes #6217


Agetian (approved):

Agetian:
:+1:

PR #6232, opened 2024-10-02
Update @Artist names in edition txt files

Ayora29:
Update @Artist names with the ones from Scryfall.


forge-gui/res/editions/Modern Horizons 3.txt
```diff
@@ -591,6 +589,14 @@ A29 R A-Guide of Souls @Ryan Valle
 A38 M A-Ocelot Pride @Chris Seaman
 A122 C A-Galvanic Discharge @Zoltan Boros
 
+[full art]
```
paulsnoops, line 592: How come you've moved these to [cards] and duplicated them here too?

Ayora29: Oh yes, that' a bug off my script. It was expecting a [full art] section like the one from Bloomburrow.
Thanks for double checking. I have fixed it.

paulsnoops (approved):

PR #6233, opened 2024-10-02
Display additional cost text during payment

tool4ever:


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -7715,12 +7719,6 @@ public void cleanupCopiedChangesFrom(Card c) {
         }
     }
 
-    public void forceTurnFaceUp() {
-        getGame().getTriggerHandler().suppressMode(TriggerType.TurnFaceUp);
```
Hanmac, line 7719: wasn't the suppressMode for when you force Turn up a card that gets removed from the battlefield or something?

and that's why shouldn't trigger?

tool4ever: yea but it gets called with runTriggers=false anyway

PR #6234, opened 2024-10-02
gui-desktop : add keyboard shortcut to press OK/Yes button

Ayora29:
gui-desktop : Match : add a configurable keyboard shortcut to press OK/yes button from prompt and pop-ups.
It allows to quickly pass phases and accept stack events.
Default to 'space bar'.  
JButtons already have a default shortcut to space bar by default but for whatever reason it does NOT always works during match.

Should not mess with the rest of the UI because like I said JButtons already have a default shortcut to space bar.

Don't know if I'm allowed to offer this kind of modification.

Don't know If I miss some Panel/pop-ups. In all the match I've done so far, it worked.


forge-gui/res/languages/es-ES.properties
```diff
@@ -418,6 +418,7 @@ lblSHORTCUT_SHOWTARGETING=Partida: alternar la orientación visual de superposic
 lblSHORTCUT_AUTOYIELD_ALWAYS_YES=Partida: ceder automáticamente en cada habilidad de la pila (Siempre Sí)
 lblSHORTCUT_AUTOYIELD_ALWAYS_NO=Partida: ceder automáticamente en cada habilidad de la pila (Siempre No)
 lblSHORTCUT_MACRO_RECORD=Partida: Grabar una macro de secuencia de acciones
+lblSHORTCUT_PRESS_BUTTON=Partida: Presione el bot�n Aceptar/S�
```
kevlahnota, line 421: Presione el botón Aceptar/Sí

forge-gui/res/languages/it-IT.properties
```diff
@@ -415,6 +415,7 @@ lblSHORTCUT_SHOWTARGETING=Incontro: attiva / disattiva gli indicatori dei bersag
 lblSHORTCUT_AUTOYIELD_ALWAYS_YES=Incontro: consenso automatico in pila: sempre Sì
 lblSHORTCUT_AUTOYIELD_ALWAYS_NO=Incontro: consenso automatico in pila: sempre No
 lblSHORTCUT_MACRO_RECORD=Incontro: registra una macro (sequenza di azioni)
+lblSHORTCUT_PRESS_BUTTON=Incontro: premere il pulsante OK/S�
```
kevlahnota, line 418: Premere il pulsante OK/Sì

forge-gui/res/languages/de-DE.properties
```diff
@@ -418,6 +418,7 @@ lblSHORTCUT_AUTOYIELD_ALWAYS_NO=Duel: Auto-Bestätigen von Fähigkeiten auf dem
 lblSHORTCUT_MACRO_RECORD=Duell: Aktion-Abfolge-Makro aufnehmen
 lblSHORTCUT_MACRO_NEXT_ACTION=Duel: führe nächste Aktion im gespeicherten Makro aus
 lblSHORTCUT_CARD_ZOOM=Duell: Zoome ausgewählte Karte
+lblSHORTCUT_PRESS_BUTTON=Duell: Dr�cken Sie die OK/Ja-Taste
```
kevlahnota, line 421: Drücken Sie die OK/Ja-Taste

forge-gui/res/languages/es-ES.properties
Ayora29: Fixed. Thank you

forge-gui/res/languages/it-IT.properties
Ayora29: Fixed. Thank you

forge-gui/res/languages/de-DE.properties
Ayora29: Fixed. Thank you

tehdiplomat (approved):

Agetian:
Hmm, the default here seems not to be the best choice - when searching for cards by name, pressing Space confusingly confirms the search instead of letting the search continue. Case in point: easily replicated in dev mode using the "Add Card To Hand" option. Or anything that says "Name a card" - for instance, you can try naming anything with a space in it (e.g. Ugin's Labyrinth).

Ayora29:
Hum I see, I can precent the shortcut to trigger for panel with "lblNameTheCard" (Name the card) title.
Don't know if there is a more elegant way to do this.

Agetian:
That'll probably not catch the all desired cases... In particular, for instance, the dev mode options use other prompts for the window title.

Ayora29:
Indeed. :( Well, I guess I can revert ListChooser.java.
The shortcut will still trigger to pass phases and accept stack events as this iw the main usage for this shortcut.
As Chooser often allows for keyboard input, the shortcut is not a good idea here.

Agetian:
Yeah, I agree. Maybe @kevlahnota or @tehdiplomat can suggest a better solution, but for now, I think reverting ListChooser would be optimal...

tehdiplomat:
I'm just gonna revert the whole thing and we can try again a different way.

PR #6256, opened 2024-10-08
Show chosen mana for mana abilities as symbols

Jetz72:
Before:
Screenshot from 2024-08-23 17-57-30

After:
Screenshot from 2024-08-24 12-20-44

This also wraps `showOptionDialog` and `showInputDialog` in callables to ensure they're always run on the event dispatch thread. At some point it'd be a good idea to sprinkle some assertions throughout the desktop UI logic and make sure the rest of it always is on that thread too. I ran into some tricky to debug deadlock issues without it.


forge-game/src/main/java/forge/game/ability/effects/ManaEffect.java
```diff
@@ -106,25 +111,26 @@ public void resolve(SpellAbility sa) {
                             choiceString.append(" ");
                         }
                         choiceString.append(choice);
+                        choiceSymbols.append(MagicColor.toSymbol(choice));
```
tehdiplomat, line 114: Is there a way to choose Colorless from a Snow source?

Jetz72: Are you asking if there's a card/combo that lets you make that choice? The notification only shows up when generating mana with a nonmana ability, which really narrows the scope on top of being Snow, and most mana choosing effects are "of any color" which narrows it further. Maybe put Elemental Resonance on Ulalek, then Rimefeather Owl puts an ice counter on Elemental Resonance? Doesn't look like there's a notification in that branch though...

tehdiplomat: Cool. If someone finds a weird corner case we can display Snow if applicable. Looks good.

tehdiplomat (approved):

PR #6257, opened 2024-10-08
Add Demonic Covenant and supporting changes

TwentyToedToad:
- Added a method to CardTypeView that supports getting all types of a card (super, sub and core).
- Added a method to CardType that checks if a another card shares all types with the card.
- Added a parameter to CardProperty for "sharesAllCardTypesWithOther".
- Added a parameter to GameAction.sacrifice() for "ShowSacrificedCards".
  - Without showing that Demonic Covenant was sacrificed due to the milling the end of turn seems extremely abrupt. In a multiple player game it gets confusing quickly when an Enchantment just disappears without notification of some sort. I did not notice any sacrifices going to the Game Log and felt this was an pretty good solution.

### DIFF
--- a/forge-core/src/main/java/forge/card/CardType.java
+++ b/forge-core/src/main/java/forge/card/CardType.java
@@ -312,21 +312,6 @@ public final class CardType implements Comparable<CardType>, CardTypeView {
     }
 
     @Override
-    public Iterable<String> getAllTypes() {
-        final Set<String> allTypes = Sets.newHashSet();
-        for (final Supertype st : supertypes) {
-            allTypes.add(st.name());
-        }
-        for (final CoreType ct : coreTypes) {
-            allTypes.add(ct.name());
-        }
-        for (final String st : subtypes) {
-            allTypes.add(st);
-        }
-        return allTypes;
-    }
-
-    @Override
     public Set<String> getLandTypes() {
         final Set<String> landTypes = Sets.newHashSet();
         if (isLand()) {
@@ -806,10 +791,8 @@ public final class CardType implements Comparable<CardType>, CardTypeView {
         if (ctOther == null) {
             return false;
         }
-        List<String> ctOtherCoreTypes = new ArrayList<>();
-        ctOther.getAllTypes().forEach(ctOtherCoreTypes::add);
-        for (final String type : getAllTypes()) {
-            if (!ctOtherCoreTypes.contains(type)) {
+        for (final CoreType type : getCoreTypes()) {
+            if (!ctOther.hasType(type)) {
                 return false;
             }
         }
--- a/forge-core/src/main/java/forge/card/CardType.java
+++ b/forge-core/src/main/java/forge/card/CardType.java
@@ -312,6 +312,21 @@ public final class CardType implements Comparable<CardType>, CardTypeView {
     }
 
     @Override
+    public Iterable<String> getAllTypes() {
+        final Set<String> allTypes = Sets.newHashSet();
+        for (final Supertype st : supertypes) {
+            allTypes.add(st.name());
+        }
+        for (final CoreType ct : coreTypes) {
+            allTypes.add(ct.name());
+        }
+        for (final String st : subtypes) {
+            allTypes.add(st);
+        }
+        return allTypes;
+    }
+
+    @Override
     public Set<String> getLandTypes() {
         final Set<String> landTypes = Sets.newHashSet();
         if (isLand()) {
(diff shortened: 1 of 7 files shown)


forge-gui/res/cardsfolder/upcoming/demonic_covenant.txt
```diff
@@ -0,0 +1,14 @@
+Name:Demonic Covenant
+ManaCost:4 B B
+Types:Kindred Enchantment Demonic
```
tool4ever, line 3: wrong type

forge-gui/res/cardsfolder/upcoming/demonic_covenant.txt
```diff
@@ -0,0 +1,14 @@
...
+T:Mode$ AttackersDeclared | ValidAttackers$ Creature.Demon+YouCtrl | TriggerZones$ Battlefield | Execute$ TrigDrawAndDamage | TriggerDescription$ Whenever one or more Demons you control attack a player, you draw a card and lose 1 life.
```
tool4ever, line 4: wrong trigger, it's `AttackersDeclaredOneTarget`

forge-core/src/main/java/forge/card/CardType.java
```diff
@@ -788,6 +802,20 @@ public boolean sharesCardTypeWith(final CardTypeView ctOther) {
         return false;
     }
 
+    public boolean sharesAllCardTypesWith(final CardTypeView ctOther) {
+        if (ctOther == null) {
+            return false;
+        }
+        List<String> ctOtherTypes = new ArrayList<>();
+        ctOther.getAllTypes().forEach(ctOtherTypes::add);
+        for (final String type : getAllTypes()) {
```
tool4ever, line 811: The naming might have confused you but if you check above you'll see this only refers to the coreTypes
so you should just compare both cards amounts for that (so avoid extra list)

TwentyToedToad: I just looked at the rules and you are correct. I updated everything and removed what wasn't needed.

[Changes](https://github.com/Card-Forge/forge/pull/6257/commits/1f7a9197a1b09468ccfd32faf8352372d2757394)

(Sorry about the accidental force push!)

forge-gui/res/cardsfolder/upcoming/demonic_covenant.txt
```diff
@@ -0,0 +1,14 @@
+Name:Demonic Covenant
+ManaCost:4 B B
+Types:Kindred Enchantment Demon
+T:Mode$ AttackersDeclaredOneTarget | ValidAttackers$ Creature.Demon+YouCtrl | TriggerZones$ Battlefield | Execute$ TrigDrawAndDamage | TriggerDescription$ Whenever one or more Demons you control attack a player, you draw a card and lose 1 life.
```
tool4ever, line 4: `AttackedTarget$ Player` missing

forge-core/src/main/java/forge/card/CardType.java
```diff
@@ -788,6 +787,18 @@ public boolean sharesCardTypeWith(final CardTypeView ctOther) {
         return false;
     }
 
+    public boolean sharesAllCardTypesWith(final CardTypeView ctOther) {
+        if (ctOther == null) {
+            return false;
+        }
+        for (final CoreType type : getCoreTypes()) {
+            if (!ctOther.hasType(type)) {
```
tool4ever, line 795: the script probably works already since you're checking both cards anyway

but the logic here should still be fixed, right now your approach is not symmetrical (imo comparing sizes will be cleanest)

forge-gui/res/cardsfolder/upcoming/demonic_covenant.txt
```diff
@@ -0,0 +1,14 @@
...
+SVar:TrigDrawAndDamage:DB$ Draw | SubAbility$ DBLoseLife
+SVar:DBLoseLife:DB$ LoseLife | LifeAmount$ 1
+T:Mode$ Phase | Phase$ End of Turn | TriggerZones$ Battlefield | Execute$ TrigCreateAndMill | TriggerDescription$ At the beginning of your end step, create a 5/5 black Demon creature token with flying, then mill two cards.
+SVar:TrigCreateAndMill:DB$ Token | TokenAmount$ 1 | TokenScript$ b_5_5_demon_flying | TokenOwner$ You | SubAbility$ DBMill
+SVar:DBMill:DB$ Mill | NumCards$ 2 | RememberMilled$ True | ShowMilledCards$ True | SubAbility$ DBSacrifice | SpellDescription$ If two cards that share all their card types were milled this way, sacrifice CARDNAME.
```
tool4ever, line 9: description splitting doesn't work on triggers, move it to TriggerDescription

forge-core/src/main/java/forge/card/CardType.java
TwentyToedToad: Would checking the sizes of each actually ensure they both have the same core types? I'm a bit confused on that.

tool4ever: Hmn why not?
It's a Set so no worrying about any doubling

TwentyToedToad: The Core Types of one of the cards could be "Instant" and the other "Creature". Both Sets would have a size of 1 but the cards would not share types. Unless I am misunderstanding the rules of the card.

tool4ever: haha no, you're absolutely right and I'm too sleepy already :p

anyway the problem with your current implementation is you might not catch if `ctOther` has any additional types as long as it has at least the ones from this instance

TwentyToedToad: I'll get that fixed up! Thank you :)

tool4ever (approved):

PR #6259, opened 2024-10-08
Add Mine Is the Only Truth

TwentyToedToad:
Added supporting member variables and accessors To Player.java
Added supporting check to AbilityUtils.xCount()


tool4ever (approved):
nice work

PR #6265, opened 2024-10-09
Add Rendmaw Creaking Nest and supporting changes

TwentyToedToad:
Added a Black 2/2 Bird Creature with Flying token
Added a property to check for number of types a card has

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardProperty.java
+++ b/forge-game/src/main/java/forge/game/card/CardProperty.java
@@ -35,10 +35,7 @@ import forge.util.collect.FCollectionView;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 
-import java.util.Collections;
-import java.util.List;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
 
 public class CardProperty {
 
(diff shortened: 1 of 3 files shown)


forge-gui/res/cardsfolder/upcoming/rendmaw_creaking_nest.txt
```diff
@@ -0,0 +1,12 @@
+Name:Rendmaw, Creaking Nest
+ManaCost:3 B G
+Types:Legendary Artifact Creature Scarecrow
+PT:5/5
+K:Menace
+K:Reach
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ DBTokens | TriggerDescription$ When CARDNAME enters each player creates a tapped 2/2 black Bird creature token with flying. The tokens are goaded for the rest of the game.
```
tool4ever, line 7: when a trigger is caused by 2+ different events we still use the printed description
but we add Secondary$ param after the first one so it doesn't show up ingame

forge-game/src/main/java/forge/game/card/CardProperty.java
```diff
@@ -1529,6 +1526,9 @@ else if (prop.isEmpty() && dmgSource.equalsWithGameTimestamp(source)) {
             } else if (property.startsWith("numColors")) {
                 rhs = property.substring(11);
                 y = card.getColor().countColors();
+            } else if (property.startsWith("numTypes")) {
+                rhs = property.substring(10);
+                y = ((Collection<?>)card.getType().getCoreTypes()).size();
```
tool4ever, line 1531: this is bad for type safety
use `Iterables.size`

TwentyToedToad: Big mistake, thanks for the catch!

forge-gui/res/cardsfolder/upcoming/rendmaw_creaking_nest.txt
tool4ever: both need the same TriggerDescription
(relevant if e.g. you're asked to order simultaneous stack entry)

tool4ever (approved):

PR #6266, opened 2024-10-09
Twenty toed toad add fear of sleep paralysis and supporting rules changes

TwentyToedToad:
Added a Static Ability for Can't Remove Counters.
Added Fear of Sleep Paralysis


Hanmac:
@TwentyToedToad instead of a new Static, you could use a RemoveCounter ReplacementEffect with `Layer$ CantHappen`

so we don't need a new Static

PR #6267, opened 2024-10-09
Add Experimental Lab // Staff Room

TwentyToedToad:


forge-gui/res/cardsfolder/upcoming/experimental_lab_staff_room.txt
```diff
@@ -0,0 +1,21 @@
+Name:Experimental Lab
+ManaCost:3 G
+Types:Enchantment Room
+T:Mode$ UnlockDoor | ValidPlayer$ You | ValidCard$ Card.Self | ThisDoor$ True | Execute$ TrigManifest | TriggerDescription$ When you unlock this door, manifest dread, then put two +1/+1 counters and a trample counter on that creature.
+SVar:TrigManifest:DB$ ManifestDread | Amount$ 1 | RememberManifested$ True | SubAbility$ DBPutP1P1Counter
+SVar:DBPutP1P1Counter:DB$ PutCounter | Defined$ RememberedCard | CounterType$ P1P1 | CounterNum$ 2 | SubAbility$ DBPutTrampleCounter
+SVar:DBPutTrampleCounter:DB$ PutCounter | Defined$ RememberedCard | CounterNum$ 1 | CounterType$ Trample | SubAbility$ DBCleanup
```
tool4ever, line 7: use PutCounterAll instead like on _Guide of Souls_  because this "put" should be a single event

forge-gui/res/cardsfolder/upcoming/experimental_lab_staff_room.txt
```diff
@@ -0,0 +1,21 @@
...
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+AlternateMode:Split
+Oracle:(You may cast either half. That door unlocks on the battlefield. As a sorcery, you may pay the mana cost of a locked door to unlock it.)\nWhen you unlock this door, manifest dread, then put two +1/+1 counters and a trample counter on that creature.
+
+ALTERNATE
+
+Name:Staff Room
+ManaCost:2 G
+Types:Enchantment Room
+T:Mode$ DamageDone | ValidSource$ Creature.YouCtrl | ValidTargets$ Opponent | CombatDamage$ True | Execute$ TrigChoose | TriggerDescription$ Whenever a creature you control deals combat damage to a player, turn that creature face up or put a +1/+1 counter on it.
```
tool4ever, line 17: ValidTarget wrong (and also with typo)

forge-gui/res/cardsfolder/upcoming/experimental_lab_staff_room.txt
```diff
@@ -0,0 +1,21 @@
...
+SVar:TrigChoose:DB$ GenericChoice | Choices$ DBPutCounter,DBTurnFaceUp | Defined$ You | AILogic$ Random | ShowChoice$ ExceptSelf
```
tool4ever, line 18: this card is somewhat problematic for us because this rule will probably apply here:
> 608.2d. If an effect of a spell or ability offers any choices other than choices already made as part of casting the spell, activating the ability, or otherwise putting the spell or ability on the stack, the player announces these while applying the effect. The player can't choose an option that's illegal or impossible [...]

this means e. g. if the creature is already face up usually you will need to place the counter

I'll try to think about some solution that can work here

TwentyToedToad: Yeah I was looking at the same ruling when writing it. Wasn't sure if I should add supporting changes to the code or wait to see if you had a better idea haha.

forge-gui/res/cardsfolder/upcoming/experimental_lab_staff_room.txt
tool4ever: I don't think that effect supports Defined$ though

TwentyToedToad: You're right. It's ValidCards$

tool4ever: `ValidCards$ Card.IsRemembered`
Defined (AbilityUtils) vs Valid (CardProperty) syntax

PR #6280, opened 2024-10-11
Add My Will Is Irresistible

TwentyToedToad:


forge-gui/res/cardsfolder/upcoming/my_will_is_irresistible.txt
```diff
@@ -0,0 +1,10 @@
+Name:My Will Is Irresistible
+ManaCost:no cost
+Types:Scheme
+T:Mode$ SetInMotion | ValidCard$ Card.Self | TriggerZones$ Command | Execute$ DBChooseOpp | TriggerDescription$ When you set this scheme in motion, choose up to three nonland permanents you don't control. Target opponent chooses one of those permanents. You gain control of the rest.
+SVar:DBChooseOpp:DB$ ChoosePlayer | Defined$ You | Choices$ Player.Opponent | RememberChosen$ True | SubAbility$ DBChoosePerms | ChoiceTitle$ Select target opponent.
```
tool4ever, line 5: you need to target the Opponent

forge-gui/res/cardsfolder/upcoming/my_will_is_irresistible.txt
```diff
@@ -0,0 +1,10 @@
+Name:My Will Is Irresistible
+ManaCost:no cost
+Types:Scheme
+T:Mode$ SetInMotion | ValidCard$ Card.Self | TriggerZones$ Command | Execute$ DBChooseOpp | TriggerDescription$ When you set this scheme in motion, choose up to three nonland permanents you don't control. Target opponent chooses one of those permanents. You gain control of the rest.
+SVar:DBChooseOpp:DB$ ChoosePlayer | Defined$ You | ValidTgts$ Opponent | SubAbility$ DBChoosePerms | ChoiceTitle$ Select target opponent.
```
tool4ever, line 5: there's no reason for this extra effect now
just do the targeting on L7

TwentyToedToad: Ah, sorry. Before I change it I just want to verify rules.

Its been a few days since I wrote this but I believe I was trying to replicate the paper rules with this card (or at least how I interpreted them):

1. Card goes on stack with "Target opponent"
2. Activator selects three cards
3. "Target opponent" selects one card
4. Change of ownership happens

If this is incorrect let me know and I'll fix things to be accurate.

tehdiplomat: Right, but it doesn't matter which line in the script does the targeting. You don't need to make sure it happens "first" Forge knows to do it before it goes on the stack.

TwentyToedToad: Oh, okay. Thank you <3

tool4ever (approved):

PR #6287, opened 2024-10-12
Fear of Sleep Paralysis and support

tool4ever:
- the logic for this was tricky, _Dramatist's Puppet_ should now correctly force you to put (or remove) another Stun counter on the target if the other choice is impossible

- Untap costs checks extended

- some AI parts updated to avoid cheating

- Also fixed _Afiya Grove_ trigger moving counters if it gets phased out in response

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/CountersMoveEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/CountersMoveEffect.java
@@ -101,8 +101,7 @@ public class CountersMoveEffect extends SpellAbilityEffect {
         // uses for multi sources -> one defined/target
         // this needs given counter type
         if (sa.hasParam("ValidSource")) {
-            CardCollectionView srcCards = game.getCardsIn(ZoneType.Battlefield);
-            srcCards = CardLists.getValidCards(srcCards, sa.getParam("ValidSource"), activator, host, sa);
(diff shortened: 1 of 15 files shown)


forge-game/src/main/java/forge/game/ability/effects/CountersMoveEffect.java
```diff
@@ -147,11 +146,6 @@ public void resolve(SpellAbility sa) {
             Map<CounterType, Integer> countersToAdd = Maps.newHashMap();
 
             for (Card src : srcCards) {
-                // rule 121.5: If the first and second objects are the same object, nothing happens
```
Hanmac, line 150: Why remove this?

tool4ever: code cleanup because removeCounter helper checks for it anyway

Hanmac:
I'm going to need to tinker about the other case with draw -> replaced by lifegain case too

tool4ever:
did you get an answer on that? I can't see on Twitter 💩

Hanmac:
> did you get an answer on that? I can't see on Twitter 💩

Don't see a answer yet
That's probably the case

But that can be done for later MR

Hanmac (approved):
I think that's ok?

PR #6289, opened 2024-10-12
Added Smoky Lounge//Misty Salon DSK

loud1990:
Sorry I thought I updated my other commit you commented on but I guess not. Still learning how github works. This one is fixed with updated triggerdescription and restrictvalid.


tool4ever (approved):
thanks

PR #6325, opened 2024-10-14
Add Winter Cynical Opportunist

tehdiplomat:


forge-game/src/main/java/forge/game/ability/effects/ChangeZoneEffect.java
```diff
@@ -1159,6 +1162,11 @@ else if (origin.contains(ZoneType.Hand) && player.isOpponentOf(decider)) {
                             fetchList = CardLists.getValidCards(fetchList, "Card.powerLE" + totpower, source.getController(), source, sa);
                         }
                     }
+                    if (totalCardTypes != null) {
+                        if (totCardTypes >= 0) {
+                            fetchList = CardLists.getValidCards(fetchList, "Card.numTypesLE" + totCardTypes, source.getController(), source, sa);
```
tool4ever, line 1167: this doesn't seem right
imo you need at least "four or more card types among them" to exile anything
but your logic mirrors the other params which are meant for "or less"

tehdiplomat: Yea, maybe this filter just doesn't matter. Lemme fiddle with it some more.

tool4ever: I guess it depends on how the GUI support should behave
in theory just ChooseCard any and then a Condition for ChangeZone would already work 🤔

tehdiplomat: Right, ideal case is counting the types like we do for Mount. But functionality, We don't need as much. Maybe i don't even need engine changes.

forge-gui/res/cardsfolder/w/winter_cynical_opportunist.txt
```diff
@@ -0,0 +1,14 @@
+Name:Winter, Cynical Opportunist
+ManaCost:2 B G
+Types:Legendary Creature Human Warlock
+PT:2/5
+K:Deathtouch
+T:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigMill | TriggerDescription$ Whenever NICKNAME attacks, mill three cards.
+SVar:TrigMill:DB$ Mill | NumCards$ 3
+T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You | Delirium$ True | TriggerZones$ Battlefield | Execute$ TrigExileDelirium | TriggerDescription$ Delirium — At the beginning of your end step, you may exile any number of cards from your graveyard with four or more card types among them. If you do, put a permanent card from among them onto the battlefield with a finality counter on it.
+SVar:TrigExileDelirium:DB$ ChangeZone | Origin$ Graveyard | Destination$ Exile | ChangeType$ Card.YouOwn | ChangeNum$ X | RememberChanged$ True | Hidden$ True | WithTotalCardTypes$ 4 | SubAbility$ DBReturn
+SVar:DBReturn:DB$ ChangeZone | Origin$ Exile | Destination$ Battlefield | WithCountersType$ FINALITY | Hidden$ True | ChangeType$ Permanent.YouOwn+IsRemembered | ChangeNum$ 1 | SubAbility$ DBCleanup
```
tool4ever, line 10: this line probably needs Mandatory$?

tool4ever (approved):

PR #6337, opened 2024-10-15
Adventure: Reward Scene Sort by Rarity

nwg5817:
Minor change to sort by rarity. This improves booster pack QoL as well as general shop browsing. It also improves monster loot viewing.
It does check to make sure it's a card, it doesn't affect the capital stores.

Mostly done for booster packs, because it's closer to opening boosters IRL then a pile.

I feel capital stores should organize the Heart and Teleport first, we could sort by cost as a fall back and that would go here.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/scene/RewardScene.java
+++ b/forge-gui-mobile/src/forge/adventure/scene/RewardScene.java
@@ -26,6 +26,7 @@ import forge.item.PaperCard;
 import forge.sound.SoundEffectType;
 import forge.sound.SoundSystem;
 import forge.util.ItemPool;
+import java.util.Comparator;
 
 /**
  * Displays the rewards of a fight or a treasure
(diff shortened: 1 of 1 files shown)


forge-gui-mobile/src/forge/adventure/scene/RewardScene.java
```diff
@@ -318,6 +319,14 @@ void updateCollectionPool() {
     }
 
     public void loadRewards(Array<Reward> newRewards, Type type, ShopActor shopActor) {
+        // Sort the rewards based on the rarity of the card inside the reward/ lets give items rarity
+        newRewards.sort(Comparator.comparing(reward -> {
+            if (reward.getCard() != null && reward.getCard().getRarity() != null) {
```
tehdiplomat, line 324: If we're going to be sorting them, should we also consider color or name as secondary sorts? Doesn't necessarily have to be in this PR but just a thought..

nwg5817: I considered it. I thought it might be overdoing it, but in the large booster box dumps it would be best. We can accept and I'll add later.

tehdiplomat (approved):

Hanmac (approved):

PR #6358, opened 2024-10-16
Adventure: Constructed as default mode

nwg5817:
Aligns the default menu options with the default new player experience.

Players will expect that constructed is the default/normal mode for first time play.


nwg5817:
To be more clear, it sets the default deck mode on new game to Constructed instead of Jump Start. Basic logic to be able to switch out defaults later as it had no sorting before.

kevlahnota (approved):

Hanmac (approved):

PR #6367, opened 2024-10-16
Add Unnatural Summons (YDSK)

TwentyToedToad:


forge-gui/res/cardsfolder/upcoming/unnatural_summons.txt
```diff
@@ -0,0 +1,8 @@
+Name:Unnatural Summons
+ManaCost:1 G U
+Types:Sorcery
+K:Rebound
+S:Mode$ ReduceCost | ValidCard$ Card.Self | Type$ Spell | Amount$ X | EffectZone$ All | Description$ This spell costs {1} less to cast if you weren't the starting player.
```
tool4ever, line 5: description doesn't match oracle

PR #6369, opened 2024-10-16
Adventure (Gamepad) : Reward Scene Zoom QoL/Bugfix

nwg5817:
Currently when zooming cards on controller, if you do not zoom out, zoomed cards will begin to stack on top of each other. You need to either leave the shop or zoom out in reverse.
This fixes that bug. Detail now acts as a toggle and you can now move through the items zoomed in.

We don't have low level controller access at the scene level, this seemed like the best approach.


nwg5817:
Bugged not opening will test more

PR #6371, opened 2024-10-16
Add Harrowing Swarm (YDSK)

TwentyToedToad:


forge-gui/res/cardsfolder/upcoming/harrowing_swarm.txt
```diff
@@ -0,0 +1,9 @@
+Name:Harrowing Swarm
+ManaCost:1 G
+Types:Sorcery
+A:SP$ ManifestDread | SubAbility$ DBAnimate | SpellDescription$ Manifest dread. Then each face-down creature you control gains "This permanent costs {2} less to turn face up" and "When this permanent is turned face up, if it's a creature, put a +1/+1 counter on it."
+SVar:DBAnimate:DB$ AnimateAll | ValidCards$ Creature.faceDown+YouCtrl | Zone$ Battlefield | staticAbilities$ ReduceCost | Triggers$ TurnedFaceUp | Duration$ Permanent
+SVar:ReduceCost:Mode$ ReduceCost | Amount$ 2 | Type$ TurnFaceUp | ValidCard$ Card.Self | Activator$ You | EffectZone$ All | Description$ This permanent costs {2} less to turn face up
```
tool4ever, line 6: please test your scripts, I can tell this part won't work

TwentyToedToad: Ah. It works great when Harrowing Swarm is cast only once. After casting it 3 times in a row the first manifested card gets 3 +1/+1 counters. I'm not sure if that is supposed to stack, but I would guess not?

tool4ever: no that's fine if I understand you correctly
it just gains another instance of those abilities each time
but the Type$ param has wrong syntax

TwentyToedToad: It does reduce the cost of turning it face up, but that is probably just a side effect of whatever I messed up here. I'll look into it more closely and make sure I'm not reducing the cost of other things.

PR #6386, opened 2024-10-19
SLD: Marvel x MTG Superdrop

dracontes:
As seen on [Secret Lair | Marvel Superdrop](https://magic.wizards.com/en/news/announcements/secret-lair-marvel-superdrop).

Card scripts tested and fully working for:
- Iron Man, Titan of Innovation
- Wolverine, Best There Is
- Storm, Force of Nature

Card scripts tested to a point, and in need of support, for:
- Black Panther, Wakandan King
→ Contrasting [CountersMoveEffect.java](https://github.com/Card-Forge/forge/blob/master/forge-game/src/main/java/forge/game/ability/effects/CountersMoveEffect.java) with [CountersRemoveEffect.java](https://github.com/Card-Forge/forge/blob/master/forge-game/src/main/java/forge/game/ability/effects/CountersRemoveEffect.java) the former doesn't have a `RememberAmount$` (or `RememberRemoved$`) parameter yet so the gaining life and drawing a card part isn't functional. I set the script up so, in principle, just adding one of the missing parameters should have the card working. My assumption for how the effect does the check is by counting the counters removed from the land and checking whether any end up on the creature.
- Captain America, First Avenger
→ Informed as I was while working on [Three Dog, Galaxy News DJ](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/t/three_dog_galaxy_news_dj.txt), that the game engine remembers objects used for paying costs, I tried to find if unattaching as a payment was considered for the activated ability. Turns out it isn't, but it does seem like something that should be implemented, so I set up the script to take advantage of that possible future implementation, lifting the pattern from `Sacrifice`.

Other updates:
- Added the Hero creature type to `TypeLists.txt`.

@Fulgur14 , I do believe we should compare notes on these.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/captain_america_first_avenger.txt
+++ b/forge-gui/res/cardsfolder/upcoming/captain_america_first_avenger.txt
@@ -2,8 +2,8 @@ Name:Captain America, First Avenger
 ManaCost:R W U
 Types:Legendary Creature Human Soldier Hero
 PT:4/4
-A:AB$ DealDamage | Cost$ 3 Unattach<Equipment.Attached/Equipment attached to NICKNAME> | PrecostDesc$ Throw . . . - | ValidTgts$ Any | TgtPrompt$ Select any target to distribute damage to | NumDmg$ X | TargetMin$ 1 | TargetMax$ 3 | DividedAsYouChoose$ X | SpellDescription$ He deals damage equal to that Equipment's mana value divided as you choose among one, two, or three targets.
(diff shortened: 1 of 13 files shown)


forge-gui/res/cardsfolder/upcoming/captain_america_first_avenger.txt
```diff
@@ -0,0 +1,9 @@
+Name:Captain America, First Avenger
+ManaCost:R W U
+Types:Legendary Creature Human Soldier Hero
+PT:4/4
+A:AB$ DealDamage | Cost$ 3 Unattach<Equipment.Attached/Equipment attached to NICKNAME> | PrecostDesc$ Throw . . . — | ValidTgts$ Any | TgtPrompt$ Select any target to distribute damage to | NumDmg$ X | TargetMin$ 1 | TargetMax$ 3 | DividedAsYouChoose$ X | SpellDescription$ He deals damage equal to that Equipment's mana value divided as you choose among one, two, or three targets.
```
tool4ever, line 5: @Hanmac
any idea how we can support this?

Hanmac: What exactly is the problem?

tool4ever: I needed partial payment decision early for the division
but should work now

tool4ever (approved):

Issue #5289, opened 2024-05-24
Anje Falkenrath / Falkenrath Gorger interaction


dwetuski:
**Describe the bug**
Anje Falkenrath's triggered ability is disregarding madness gained from Falkenrath Gorger.

**To Reproduce**
1. Cast Anje Falkenrath.
2. Cast Falkenrath Gorger.
3. Activate Anje's ability to discard a Vampire creature card.
4. The card is discarded.
5. A card draw goes onto the stack.
6. The madness ability goes onto the stack.

**Expected behavior**
6. The madness ability and Anje's triggered untap ability go onto the stack in the order the player chooses.

**Desktop (please complete the following information):**
 - Forge v.1.6.62 (forge.GuiDesktop@6ddf90b0)
 - Linux (6.1.52-valve16-1-neptune-61 amd64)
 - Java Version - 1.8.0_401

**Additional context**
"If you discard a card with madness to pay the cost of Anje's activated ability, you'll draw a card last. You may untap Anje before or after you choose whether to cast the discarded card, but not after you've chosen to cast it but before that spell resolves." (https://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=470583)

This choice of order is very important, because Anje will only untap if her triggered ability resolves before the madness triggered ability does.  The discarded Vampire loses madness when it enters the battlefield, and Anje's triggered ability rechecks for madness when it's ready to resolve (rule 603.4).

tool4ever:
@Hanmac 
this is closely related to #2193

I guess we could change Discarded trigger to use the new object for now but it won't fix all corner cases?

Hanmac:
> @Hanmac this is closely related to #2193
> 
> I guess we could change Discarded trigger to use the new object for now but it won't fix all corner cases?

@tool4ever maybe

you can try to change Discarded Trigger

but we need to be careful if the Discard is getting replaced with a hidden Zone (without reveal)
like _Library of Leng_

so for _wheel of sun and moon_ we need to recognize that:
Discard > **Reveal** > Library

tool4ever:
I know
but I've no idea how to fix that part currently

tool4ever:
It's worth noting that there is currently only "legal" nonland card (_Asmoranomardicadaistinaculdacar_) without mana cost.

Because of how it's currently coded _Falkenrath Gorger_ would fail to give it Madness (if it was also a Vampire).

But being unpayable shouldn't lead to that.

tool4ever:
[...]

PR #6388, opened 2024-10-19
Grant abilities based on mana cost even if a card doesn't have one not just for Flashback

tool4ever:
> If a card with no mana cost gains flashback, it has no flashback cost. It can't be cast this way.

Related to #5289

### DIFF
--- a/forge-core/src/main/java/forge/card/mana/ManaCost.java
+++ b/forge-core/src/main/java/forge/card/mana/ManaCost.java
@@ -42,7 +42,7 @@ public final class ManaCost implements Comparable<ManaCost>, Iterable<ManaCostSh
 
     private List<ManaCostShard> shards;
     private final int genericCost;
-    private final boolean hasNoCost; // lands cost
+    private boolean hasNoCost = true; // lands cost
     private String stringValue; // precalculated for toString;
 
     private Float compareWeight = null;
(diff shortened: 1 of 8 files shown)


forge-core/src/main/java/forge/card/mana/ManaCost.java
```diff
@@ -281,6 +281,9 @@ public String toString() {
      * @return
      */
     public String getShortString() {
+        if (isNoCost()) {
+            return "-1";
```
Hanmac, line 285: I don't like that one

I would have preferred the "no cost" stuff

tool4ever: Hmn, but this is basically the definition for it:
https://github.com/Card-Forge/forge/blob/42893234b36f5d889702d4c55ddf4afb20ccb98f/forge-core/src/main/java/forge/card/mana/ManaCost.java#L50-L50

If I'd write it out it would be treated as two different parts and fail the parsing 🤷‍♂️

Hanmac:
I still think cards without Mana Cost should not gain Keywords based on that Mana Cost 🤔 

See the reason why all the Chosen Color/Type effects got excluded

tool4ever:
I think it's allowed:
> 118.6. Some objects have no mana cost. This represents an unpayable cost. An ability can also have an unpayable cost if its cost is based on the mana cost of an object with no mana cost.

For chosen it's only because of a special rule:
> 607.5a. If an object gains an ability that refers to a choice, but either (a) doesn't copy that ability's linked ability or (b) does copy the linked ability but no choice is made for it, then the choice is considered to be "undefined." If an ability refers to an undefined choice, that part of the ability won't do anything.

forge-core/src/main/java/forge/card/mana/ManaCost.java
Hanmac: ah there is the no cost i was looking for:
```java
face.setManaCost("no cost".equals(value) ? ManaCost.NO_COST
                                : new ManaCost(new ManaCostParser(value)));
```

it is only used for reading cardfaces

Hanmac: while i'm not happy about that "-1", that can be fixed later

Hanmac (approved):

PR #6392, opened 2024-10-20
Fix the SpellSmith missing some cards.

CollinJ:
The main issues is that getAllCards returns a card list that is unique by name. That means each card only has a single rarity. In this case, cards like "Forbidden Orchard" and "Choke" de-deuplicate to a version of the card with 'S' rarity. And when filtering by rarity, the 'S' rarity will be filtered out.
Another similar issue is that when a card exists in multiple sets with different rarities, you might filter to a set in which the card is 'R', but the de-duped version in allCards has rarity 'U', so using the cardDB to check the rarity to filter on doesn't work.

When filtering by a specific set, the solution is to use the rarity of the card in that set.
When not filtering by a set but filtering by rarity, we have to address the root cause of the de-duping.

WHen de-duping, I changed the logic to ignore 'S' rarities. This fixes the rarity filter when not filtering by an edition (although if there are multiple rarities you still have to kind of guess). This has the side-effect of making the card image used something more legible. Before 'Choke' de-duped to the MPS_AHK set which has illegible text. By ignoring 'S' rarities, it not de-dups to 8ED.

More work could probably be done to improve de-duping, but perfect correctness would probably involve de-duping only after filtering, which is likely too slow.


Agetian:
Is this tested to work correctly in the "Use Card Variants from All Sets" mode where all the cards from all sets (including special sets) are allowed? Please test this to ensure that e.g. the cards from MPS_AKH set (with the "S" rarity) are still available in the said mode and don't cause any issues.

CollinJ:
That config option is only referenced in `pullCard()` which I didn't touch. Also it just uses the card name from the filtered card pool, so it doesn't matter what the rarity/edition is in the filtered card pool.
```
        if (Config.instance().getSettingData().useAllCardVariants) {
            if (!edition.isEmpty()) {
                currentReward = new Reward(CardUtil.getCardByNameAndEdition(P.getCardName(), edition));
            } else {
                currentReward = new Reward(CardUtil.getCardByName(P.getCardName())); // grab any random variant if no set preference is specified
            }
        }
```
And `getCardByName` will use the full card database. I edited pullCard to always choose choke:
```
        for (PaperCard card : cardPool)
        	if (card.getName().equals("Choke")) P = card;
```
And it generated all card images for Choke when filtering by nothing or by rarity: "U".

Agetian (approved):

PR #6412, opened 2024-10-22
display locked room, fix nonrotated card panels

kevlahnota:


Hanmac:
In Case of `CardStateName.EmptyRoom`, both rooms are locked and both 🔒 should be displayed

Jetz72:
Is there a reason to use separate overlays for the left and right locks, rather than one sprite drawn across either half of the card?

kevlahnota:
> Is there a reason to use separate overlays for the left and right locks, rather than one sprite drawn across either half of the card?

just lazy to compute exact coordinates when needed. you can try to modify it and use the extra space to something

Issue #5604, opened 2024-07-14
AI activates some activated abilities as soon as it can with no reason


MorihiroS:
**Describe the bug**
AI activates activated abilities of some cards as soon as it can with no reason. It's ineffectively.
Some of this cards:
- [ ] Ugin's Labyrinth (second ability)
- [ ] Canopy lands
- [ ] Zuran Orb
- [ ] Skirk Prospector

**To Reproduce**
1. AI plays Ugin's Labyrinth
2. AI uses second ability of Ugin's Labyrinth immediately

or

1. AI play Fiery Islet and casts Ragavan, Nimble Pilferer
2. AI successfully attacks with Ragavan, Nimble Pilferer and takes Treasure token
3. AI sacrifices Fiery Islet and now it hasn't lands

or

1. AI casts Zuran Orb
2. AI sacrifices all of its lands immediately

or

1. AI has one untapped land
2. AI casts Skirk Prospector for land's mana
3. AI sacrifices Skirk Prospector and casts Aether Vial

**Expected behavior**
AI should activate activated abilities of this and similar cards in accordance with the situation.
For example:
- AI activates second ability of the Ugin's Labyrinth only after it has enough mana to cast exiled card or in response to an attempt to destroy the Ugin's Labyrinth;
- AI sacrifices canopy lands only after it already has X lands and/or no cards in its hand
- I don't know, may be AI should activate the Zuran Orb ability when it has less than X life or when Titania, Protector of Argoth is on the battlefield
- AI doesn't sacrifice Skirk Prospector before it has 2+ lands and only for cast a spell with cmc = number of its lands + 1

**Additional context**
Similar with #4214

**Desktop**
 - OS: Windows 10
 - Java Version: 1.8.0_202-b08
 - Forge Version: 1.6.64-SNAPSHOT-07-13

**Screenshots**

![labir](https://github.com/user-attachments/assets/711b8311-586b-4bb0-ae08-a76120d3bf7c)
![canopy](https://github.com/Card-Forge/forge/assets/115870280/f68d5c6c-5441-42ff-9880-061c1cdbd1bf)
![canopy2](https://github.com/Card-Forge/forge/assets/115870280/f114855b-ad36-472c-92cc-f8b39dcb8769)
![zuran](https://github.com/Card-Forge/forge/assets/115870280/ae692c2f-4850-47c9-8f71-e37ab9e71c6e)
[...]

PR #6418, opened 2024-10-23
Various improvements to AI-related logic (Canopy lands, Ugin's Labyrinth, Zuran Orb, Pyrite Spellbomb, Eladamri)

Agetian:
Some work towards #5604.


forge-ai/src/main/java/forge/ai/ability/ChooseCardAi.java
```diff
@@ -58,11 +59,27 @@ protected boolean checkApiLogic(final Player ai, final SpellAbility sa) {
     protected boolean checkAiLogic(final Player ai, final SpellAbility sa, final String aiLogic) {
         final Card host = sa.getHostCard();
         final Game game = ai.getGame();
+
         ZoneType choiceZone = ZoneType.Battlefield;
+        CardCollectionView choices = new CardCollection();
         if (sa.hasParam("ChoiceZone")) {
-            choiceZone = ZoneType.smartValueOf(sa.getParam("ChoiceZone"));
+            String definedChoiceZone = sa.getParam("ChoiceZone");
```
tool4ever, line 66: try to use classes own helpers instead of manually coding like `listValueOf` here

Agetian: Done :) Thanks for the heads up, didn't realize listValueOf was a thing :)

tool4ever: Still seems way overcomplicated :p
why not just have the code look like this?
```
        List<ZoneType> choiceZone = Lists.newArrayList(ZoneType.Battlefield);
        if (sa.hasParam("ChoiceZone")) {
            choiceZone = ZoneType.listValueOf(sa.getParam("ChoiceZone"));
        }
        CardCollectionView choices = game.getCardsIn(choiceZone);
```

Agetian: Ah, indeed :) Thanks a lot for the suggestion! Implemented.

Hanmac:
I wonder if AI shouldn't use Draw Effects (without Discard before or After),
if the resulting Draw would get them over their hand discard on their turn?

Like, check for: (pseudocode)
```
if (!handSizeUnlimited && numDraw + cardsInHand > handSize) {
  return false;
}
```

Agetian:
Prolly a good idea, yeah, though it's a bit tricky since the AI may be able to use the said card on its turn and get some advantage from having the extra card :/ I'll need to think how to best do this. Prolly a point for the next MR :)

Hanmac (approved):

Issue #5137, opened 2024-04-24
Arboria rework


tool4ever:
This took a while since I was looking for some synergy with existing stuff ;)

Imo you could refactor it slightly to improve _Arboria_:
> Creatures can’t attack a player unless that player cast a spell or put a nontoken permanent onto the battlefield during their last turn.

Currently both parts only work as bare minimum (static trigger counting) :/

For the casting check it might be easier to consolidate everything in a MagicStack Map <int turnNr, Spells> though.
`Player` already stores a `lastTurnNr`, so with that you could just collect all the relevant entries?

I think we can still get away with periodic cleanup for memory, just the lower bound now being the minimum of each players last turn number (instead of the global last)

For the other check I was concerned we'd need to store additional `cardsAddedLastTurn` fields in `Zone` in the same way.
However there's a subtle difference:
> Arboria’s effect cares whether a player put a nontoken permanent onto the battlefield. It’s unusual for an ability to care who put a permanent onto the battlefield, as opposed to under whose control it entered.

This can matter if you look for effects like this:
> When Petradon leaves the battlefield, return the exiled cards to the battlefield under their owners’ control.

➡️ Here the trigger controller moves the cards, but they can ETB under another.

Since this seems to be also pretty unique it might be easier to have this counted simply as boolean.

While changing this PR for that obscure World would be helpful, feel free to resolve via a new issue after (or even before hehe).

WDYT?

_Originally posted by @tool4ever in https://github.com/Card-Forge/forge/pull/4900#discussion_r1559687924_

tool4ever:
It might be a bit more problematic than I thought since we can't just always check for the Cause controller instead.
Some cards will make different players do the moving, e. g. _Second Sunrise_:
> Each player returns to the battlefield all...

So looks like we'll need an additional param.
But I'm not sure yet if one of the cases is more common, so that making it the default one would hopefully avoid lots of script adjustments 🤔

tool4ever:
I stumbled upon _Land Equilibrium_:
> This effect applies no matter how the land would enter the battlefield: because an opponent plays it, or because a spell or ability allows that opponent to put it onto the battlefield. Note that it doesn’t matter whose control the land enters the battlefield under. If the opponent would put the land onto the battlefield under someone else’s control (as a result of Yavimaya Dryad’s ability, for example), that opponent will still have to sacrifice a land.

@Fulgur14
do you know any other cards that need "Card Putter" info?

Fulgur14:
[...]

PR #6420, opened 2024-10-23
Support engine tracking card mover vs. controller

tool4ever:
Adds support that will be needed for #5137


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -209,11 +209,27 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
...
+        Player putter = c.getController();
+
         // Don't copy Tokens, copy only cards leaving the battlefield
         // and returning to hand (to recreate their spell ability information)
         if (toBattlefield || (suppress && zoneTo.getZoneType().isHidden())) {
             copied = c;
 
+            // in some cases it's always affected that puts them in play (initially)
+            String defPutter;
+            if (cause == null || (!cause.hasParam("Putter") && (c.isToken() ||
```
Hanmac, line 221: I don't know if I like it there

I probably would move the Putter logic into `AbilityKey.addCardZoneTableParams`

tool4ever: hmn, the problem is the Putter can be different players during the same effect

and most effects currently don't need to modify moveParams between cards and reuse them instead

so I'd have to refactor those further, especially when they pass down to helpers in SpellAbilityEffect like with discard?

Hanmac: imo it would be more clean if the stuff is done in the specific SpellAbilityEffect classes (or a helper function)

Hanmac: @tool4ever can you move the Putter Logic into Effects instead?

tool4ever: I tried but like I said above it got messy fast ;/
Not sure yet if I can find a cleaner approach

Hanmac: i still think it would be cleaner in `SpellAbilityEffect` like classes

tool4ever: maybe I'll be able to finish this one day 🤷‍♂️
but having one open compared to all of yours doesn't seem too bad :P

PR #6460, opened 2024-10-29
Fix some scripts

tool4ever:
I think this matters since _Idris, Soul of the TARDIS_ exiling _Chlorophant_ should only provide its first trigger to her :p


Hanmac:
Can you look into:
> RuntimeException: AbilityFactory : getAbility -- Eidolon of Astral Winds has no SVar: TrigAnimate

Hanmac:
What is the issue this trying to fix?

tool4ever:
This just improves rules accuracy on relevant scripts so they're all triggers granted by statics.

I can't fix _ Eidolon of Astral Winds _ because right now it only exists for the scripter who reported his own typo 💩

Hanmac (approved):

PR #6461, opened 2024-10-29
Another 12 FDN/J25 cards

Fulgur14:


forge-gui/res/cardsfolder/upcoming/electroduplicate.txt
```diff
@@ -0,0 +1,8 @@
+Name:Electroduplicate
+ManaCost:2 R
+Types:Sorcery
+K:Flashback:2 R R
+A:SP$ CopyPermanent | ValidTgts$ Creature.YouCtrl | TgtPrompt$ Select target creature you control | AddKeywords$ Haste | AddSVars$ TrigEOT | SpellDescription$ Create a token that's a copy of target creature you control, except it has haste and "At the beginning of the end step, sacrifice this token."
```
tool4ever, line 5: trigger missing

Fulgur14: Isn't it just after this line?

tool4ever: it's not an activated ability, so you should use AddTriggers

tool4ever (approved):

PR #6489, opened 2024-11-01
Edit pile cleanup: Deck tags, first pass: folder 'a', part 1

dracontes:
A fairly extensive pass: 4566 changed files on my side, though overlap with previous PRs resulted in some edit redundancy. 

**The gist:**
- More reordering of lines around `Deck*:` tags
- Propagation and clean up of `Deck*:` tags, notably tagging token scripts and various extra deck cards.
- Minor edits


forge-gui/res/cardsfolder/a/adaptive_sporesinger.txt
```diff
@@ -7,5 +7,6 @@ T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.S
 SVar:TrigCharm:DB$ Charm | Choices$ DBPump,DBProliferate
 SVar:DBPump:DB$ Pump | NumAtt$ 2 | NumDef$ 2 | ValidTgts$ Creature | KW$ Vigilance | SpellDescription$ Target creature gets +2/+2 and gains vigilance until end of turn.
 SVar:DBProliferate:DB$ Proliferate | SpellDescription$ Proliferate. (Choose any number of permanents and/or players, then give each another counter of each kind already there.)
-DeckHas:Ability$Proliferate
+DeckHas:Ability$Proliferate|Counters
```
tool4ever, line 10: this card doesn't have a "Counters" ability by itself

forge-gui/res/cardsfolder/a/accomplished_automaton.txt
```diff
@@ -3,5 +3,5 @@ ManaCost:7
 Types:Artifact Creature Construct
 PT:5/7
 K:Fabricate:1
-DeckHas:Ability$Counters|Token
+DeckHas:Ability$Counters|Token & Type$Servo
```
tool4ever, line 6: I wanna say for consistency this should rather be done via engine support as well
At least it seems not too difficult, just not sure it's in my schedule for the rest of the year :p
So probably better to skip the double work (sorry)

dracontes: The same would be valid for the likes of embalm and eternalize, then?

forge-gui/res/cardsfolder/a/aethersquall_ancient.txt
```diff
@@ -6,4 +6,6 @@ K:Flying
 T:Mode$ Phase | Phase$ Upkeep | ValidPlayer$ You | TriggerZones$ Battlefield | Execute$ TrigEnergy | TriggerDescription$ At the beginning of your upkeep, you get {E}{E}{E} (three energy counters).
 SVar:TrigEnergy:DB$ PutCounter | Defined$ You | CounterType$ ENERGY | CounterNum$ 3
 A:AB$ ChangeZoneAll | Cost$ PayEnergy<8> | ChangeType$ Creature.StrictlyOther | SorcerySpeed$ True | Origin$ Battlefield | Destination$ Hand | SpellDescription$ Return all other creatures to their owners' hands. Activate only as a sorcery.
+DeckHas:Ability$Counters
+DeckHints:Ability$Counters
```
tool4ever, line 10: especially the hints seem too generic for energy imo 🤔 

I guess `Ability$Proliferate` is an alternative but honestly just because it's possible we really shouldn't flood every counter producer with that (especially in cases like this card which already works not too bad on its own)
Not sure what could be the right boundary - maybe only the really strong synergy ones like _Simic Ascendancy_?

FWIW `Ability$Energy` is also used on 3 cards so that's useless and should probably be removed instead (or is Energy enough of an Archetype to bother with completing it?)

Perhaps we can try and write down some rough decision guidance while working through these PR...

dracontes: Fair on being more judicious about it. I have been more inclined to hint proliferate on cards that distribute counters.

forge-gui/res/cardsfolder/a/accomplished_automaton.txt
tool4ever: well those aren't just tokenscripts keyworded which my approach could handle
while still tokens I'd need to support loose params in general, I'm not even sure that would be worth it yet

dracontes: Oh my mistake then: it's living weapon and for Mirrodin that are relevant here. Will clean up.

forge-gui/res/cardsfolder/a/adaptive_sporesinger.txt
tool4ever: imo yes, I can see why it could be interpreted that way but remember it's mostly for deckbuilding

e.g. _Bloom Hulk_ gains nothing if you'd encourage a combination with this one

forge-gui/res/cardsfolder/a/accomplished_automaton.txt
dracontes: And just so it's clear: is `Investigate` something that your approach can handle? I ask because I notice that, in the code, it strays a bit from the pattern for fabricate, living weapon and the like.

tool4ever: not sure what you mean :)

dracontes: Well, looking over [InvestigateEffect.java](https://github.com/Card-Forge/forge/blob/master/forge-game/src/main/java/forge/game/ability/effects/InvestigateEffect.java), `c_a_clue_draw` shows up but it isn't associated with the `StringBuilder` element in a comparable way to the keyword abilities we've discussed (as seen on [CardFactoryUtil.java](https://github.com/Card-Forge/forge/blob/master/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java)). At least not so clearly associated that I feel confident reverting the relevant work I've done. So I'd like confirmation that, for instance, the` DeckHas:` line on [Alquist Proft, Master Sleuth](https://github.com/dracontes/forge/blob/rb-editpile-decktag-a-1-patch-2/forge-gui/res/cardsfolder/a/alquist_proft_master_sleuth.txt) is either to remove or to keep.

forge-gui/res/cardsfolder/a/aethersquall_ancient.txt
dracontes: While reviewing keywords with issues to address, I tagged:
- scripts with awaken with the creature type and keyword ability
- scripts with a level up cost of 3 mana or more with proliferate

As for energy, I feel something like `Ability$Counters.Energy` to go alongside `Ability$Counters.P1P1` could be more useful. Of course, some character other than a period can be used as a separator for this layer of classification.

forge-gui/res/cardsfolder/a/accomplished_automaton.txt
tool4ever: Right, that would actually be a third variant to support 😅

I might consider a more generic approach for this one where we can hardcode some implication logic for Ability$ values, e.g. `Investitigate -> Color$Colorless & Type$Artifact|Clue` and potentially store them in an extra file

but that's also something I'd only attempt after implementing what I originally commented on first 🤹‍♂️

forge-gui/res/cardsfolder/a/aethersquall_ancient.txt
tool4ever: hmn working towards some Ability groups like that would also let you use logic where a rather generic tag can be fulfilled with more specific ones based on it 🥇

`DeckHints:Ability$Counters` ➡️ `DeckHas:Ability$Counters.Energy` + `DeckHas:Ability$Counters`
but
`DeckHints:Ability$Counters.Energy` ➡️ `DeckHas:Ability$Counters.Energy`

I already added the basics for that in #564

dracontes: Oh okay then, I'll propagate that with (I hope) the required judiciousness.

Also, while we're on Ability groups, I've had the vague notion that `Mana.Prismatic`/`Mana.Colors`/`Mana.AnyColor` for the likes of Birds of Paradise could be useful to go along with `Mana.Colorless`.

dracontes: Just to confirm: from looking over the code, when we're dealing with multiple kinds of counters (for instance, [Phyrexian Hydra](https://scryfall.com/card/mbs/85/phyrexian-hydra) ) the correct script snippet is `Ability$Counters.M1M1|Counters.Poison`, right?

tool4ever: that should fit

dracontes: A fair bit. I think I've got the `M1M1` and `P1P1` counters adequately covered. Now it's charge counters, maybe time counters and a smattering of other aesthetic ones afterward. And then it's making the pass to recheck `Proliferate` for relevance.

tool4ever: woah alright, you didn't have to take care of all types at once 💪

tool4ever:
I'll probably close this after the other parts time out and wait for the revised edits

PR #6503, opened 2024-11-02
Added script text for Squad Rallier and Stab cards.

earno12:
These belong to set Foundations (FDN).

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/squad_rallier.txt
+++ b/forge-gui/res/cardsfolder/upcoming/squad_rallier.txt
@@ -1,0 +1,7 @@
+Name:Squad Rallier
+ManaCost:3 W
+Types:Creature Human Scout
+PT:3/4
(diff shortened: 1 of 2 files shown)


forge-gui/res/cardsfolder/upcoming/squad_rallier.txt
```diff
@@ -0,0 +1,7 @@
+Name:Squad Rallier
+ManaCost:3 W
+Types:Creature Human Scout
+PT:3/4
+A:AB$ Pump | Cost$ 2 W | Execute$ TrigDig | SpellDescription$ Look at the top four cards of your library. You may reveal a creature card with power 2 or less from among them and put it into your hand. Put the rest on the bottom of your library in a random order.
```
tool4ever, line 5: no reason for extra Pump

tool4ever (approved):
thanks

PR #6511, opened 2024-11-02
Combine Unknown Event editions into DA1

### LINKED ISSUE #6182: Scryfall card images for Unknown Events
* Scryfall is adding card images for the unknown events, under the set code DA1 https://scryfall.com/sets/da1
* Several cards from those events are in forge #5913 and #6018 *(I plan to make more with the rule for card tokens when I get time)*
* The implemented ones use set codes like UELAS23 to distinguish eachother
* No clue how the forge image database thingy work to connect them
* Thought it was at least good to inform of that in case relevant people didn't know that scryfall added those cards

paulsnoops:
Closes #6182

I have deleted all of the Unknown Event editions (which don't exist) and combined them into DA1 to align with Scryfall. All cards but 6 from the deleted editions are present in Scryfall's DA1 but I have added these 6 to the end of the list and will reach out to Scryfall to get them added there.

@Lykrast tagging you here so you can see the changes.


Lykrast:
Ooh thanks! And for combining the cards that came out since then!
A bit of a shame it's separating how they were in different events (I was kinda hoping to one day make the separate draft formats but I got no clue how to do that + lots of cards aren't implementable + given I did not have motivation to do more cards lately that was probably not gonna be a thing I'd do for a long time) but like that doesn't change much.

PR #6516, opened 2024-11-02
Add events for plotting cards into the game log

Ampersandnz:
(copied from the existing Foretold and Phased events)

image


Hanmac (approved):

Hanmac:
@tool4ever 

```java
String sb = TextUtil.concatWithSpace(getActivatingPlayer().toString(), "has plotted", c.toString());
game.getGameLog().add(GameLogEntryType.STACK_RESOLVE, sb);
game.fireEvent(new GameEventCardPlotted(c, getActivatingPlayer()));
```

for later, how about trying to move the messages into an extra Handler that looks for the Events?

tool4ever:
I think `GameLogFormatter` could already catch them?

Hanmac:
> I think `GameLogFormatter` could already catch them?

you are right, maybe we need to move more stuff to GameLogFormatter instead

PR #6520, opened 2024-11-03
update VCardDisplayArea, VZoneDisplay

kevlahnota:
prevent NPE and update some translations


forge-gui-desktop/src/main/java/forge/screens/match/controllers/CCombat.java
```diff
@@ -78,13 +78,19 @@ private static String getCombatDescription(final CombatView localCombat, final G
         }
 
         display.append("\n");
-
+        PlayerView controller = null;
         if (defender instanceof CardView) {
-            final PlayerView controller = ((CardView) defender).getController();
-            display.append(Lang.getInstance().getPossesive(controller.getName())).append(" ");
+            controller = ((CardView) defender).getController();
+            if (controller == null)
```
tool4ever, line 84: I don't like such null checks that just hide the bug instead of really fixing the cause 🤔

kevlahnota: So I just leave it crashing why it didn't have controller? I can just revert it if you want.

tool4ever: Nah that's not what I meant
since in this case the View is supposed to represent the ingame object it shouldn't be possible to have no controller during combat, so that's what needs debugging

kevlahnota: play a network game and most object will have null, I mean It's a hard to reproduce on single player but It shouldn't crash the GUI or you can't finish any game. If its really null then it will crash somewhere else.

Issue #6528, opened 2024-11-03
Bug Report: "Mr. Foxglove" does not work as intended with Draw Replacement Effects.


Skuzee:
**Describe the bug**
_Mr. Foxglove_ says:  
Whenever Mr. Foxglove attacks, draw cards equal to the number of cards in defending player's hand minus the number of cards in your hand. **If you didn't draw cards this way**, you may put a creature card from your hand onto the battlefield.  

With a replacement a effect like _Teferi's Ageless Insight_ that replaces/doubles the number of cards drawn, Forge prompts you to put a creature into play despite having drawn cards with Mr. Foxgloves ability.

**To Reproduce**
1. You have Mr. Foxglove and Teferi's Ageless Insight on the battlefield under your control.  
2. Have less cards than an opponent.  
3. Have another creature in hand.  
4. Attack with Mr. Foxglove.  
5. You will draw double cards and then forge will prompt you to put a creature into play from your hand.

**Expected behavior**
The attack ability will trigger, Teferi's Ageless Insight replacement effect will change the number of cards you draw, Mr. Foxglove should still be the source of card draw. You draw twice as many cards and do not put a creature into play.

**Desktop (please complete the following information):**
 - OS: Windows 10 Enterprise
 - Browser: N/A
 - Version [e.g. 22] forge-installer-2.0.00-SNAPSHOT-10.28.tar

**Additional context**
Thanks!

IceBen4444:
I checked the code and the bug luckily ends at Foxglove.
[The card's definition](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/m/mr_foxglove.txt) has a bit about `RememberDrawn$ True` while it should be `RememberDrawn$ AllReplaced` like on other cards that care about "this way".
[A proper example is Ancient Excavation](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/a/ancient_excavation.txt)

PR #6530, opened 2024-11-04
Fix Windfall

### LINKED ISSUE #6115: Windfall effects have incorrect count for NPC's with "instead" effects
In a 'wheel' deck with Library of Leng and Jace's Archivist in play; I noticed the NPC's only drew 7, even though I had 24 cards in hand.
I suspect that discarding to an alternate zone doesn't add to the NPC's card count.  It's probably the same with Rest in Peace or cards that never hit the graveyard due to 'instead' effects.  Interestingly, it did not affect me, I drew the proper number. 

This sorely nerfs card-draw tax effects like Orcish Bowmasters or Smothering Tithe.

1.6.65 snapshot

tool4ever:
Closes #6115


Hanmac (approved):

Hanmac:
@tool4ever does this close #6528 too?

tool4ever:
no, I think that might need a script patch but no more time right now

PR #6547, opened 2024-11-08
AI hint for Evolved Sleeper

Agetian:
Prevents perpetual reuse of the second ability


Hanmac:
Wouldn't it be better if using a Generic Check and check if the Counter is Useless? `ComputerUtil.isUselessCounter`

Or more speficlic, if Counter is a KeywordCounter and CardName already has that Keyword? `type.isKeywordCounter() && c.hasKeyword(type.toString())`

Agetian:
Sounds like a plan! I'll give it a shot ;) Thanks for the tip!

Agetian:
This seems to do the job :)

PR #6560, opened 2024-11-10
Attempt to fix Warlord's Elite AI payment

### LINKED ISSUE #2177: AI Cast Warlord's Elite with 3 basic lands and an artifact
The AI had 2 forests, a plains and a Goblin Firebomb (artifact) out, but managed to cast Warlord's Elite. It failed when I tried to do it myself, but the AI could!

Agetian:
Aims to fix #2177 , using a card memory algorithm similar to the one employed in checkTapTypeCost. Seems to work, but I'm not sure. Thoughts? :)


tool4ever:
This looks like you're wanting to fix this old problem?
`SVar:AIPaymentPreference:DontPayTapCostWithManaSources`
Might need some debugging if this handles all scenarios instead 🤔

Agetian:
Tested the DontPayTapCostWithManaSources thing, sadly, that's a different thing and isn't covered with this fix. For example, Zahid still will be paid in a cheaty way if we remove DontPayTapCostWithManaSources. I think I'll apply this fix for now (for the other issue) and I'll take a deeper dive into the other issue later :)

Agetian:
Hmm, can you send me the test case where it fails? I haven't seen this happen in my tests, so the conditions or involved cards must be different.

tool4ever:
Hmn, I'm only trying with minimal setup, e.g. AI has 3 Plains in play + Warlord's Elite in hand

Agetian:
Ok, reproduced it :) Will take a look as to why this is happening

Agetian:
I guess it's too late to bail on the payment decision when in AiCostDecision, so I'll need to find a way to determine this earlier. I'll think of a better approach if possible.

Agetian:
For now, I added the DontPayTapCostWithManaSources as a temporary stop-gap measure. Can't think of a proper fix for all these cases (help is welcome if anyone is willing to fix this one for good), but this imo is better than no fix at all - yes, the AI won't pay for this with land (or land creatures or whatever), but at least it'll play it at all and it'll play it without cheating :/

Issue #3652, opened 2023-08-18
Cryptic Spires does not work properly


Goddfrey:
**Describe the bug**
Cryptic Spires says "As you create your deck, circle two of the colors below" (out of 5 basic mana colors). In game you can tap it to add one mana of either of the circled colors.

How it's implemented here in the quest mode is you choose the two colors before the game starts. Whatever I'm fine with it.

But the real bug is it produces two colorless when tapped.

**To Reproduce**
Play Cryptic Spires
Tap it for mana

**Expected behavior**
Should add 1 mana of one of the chosen colors. Not two colorless. 

**Screenshots**

![cryptic-spires-bug](https://github.com/Card-Forge/forge/assets/49716936/b3b202ca-1010-4fb2-ac86-b0e2aa8c2977)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Forge 1.6.57, Java 8

Hanmac:
@tool4ever can you look into this Problem?

pfirpfel:
Can confirm, still bugged. There is a color selection at the start of the game, but it still produces 2 colorless mana instead.

Hanmac:
@pfirpfel that's because @tool4ever or others hasn't looked at the problem yet

Agetian:
I tried debugging this, it looks like the information about the chosen colors is lost between the time it's actually chosen and the standard game flow begins. My guess is that this information is reset somewhere when the library is prepared or something along those lines, but I haven't been able to establish the exact cause yet. This should probably be done somewhere later in the chain of game events then?

Agetian:
Actually, I'm not sure why this is happening, since the next thing after doing the Spires color choice routine is drawing the hand, and that doesn't seem to run anything that would reset the chosen color info :/

Agetian:
By the time GameAction calls changeZone on the Spires when you play it as a land, the game tries to grab the last known info via getLKICopy, and the source in there already has no information on the chosen colors (the relevant array is empty).

Northmoc:
I had the thought that there might actually be 10 different card scripts... each instance gets locked in when drafting / constructing decks... but got a little stuck on bringing the theory to completion

This could lead to implementing Attractions in Unfinity, too

Perpetual stuff might be lost between matches, though that is an interesting tactic too

Agetian:
Ah, makes sense! :+1:

hmorr5:
[...]

PR #6570, opened 2024-11-12
add support for Cryptic Spire

kevlahnota:


tool4ever:
Imo it should be implemented by the rules instead like I mentioned in #3652

> 207.5. One card (Cryptic Spires) has a set of symbols below the text box that represent each color and an ability that instructs a player to circle two of those colors as they create their deck. To circle a color, the player circles (or otherwise clearly indicates) that color's symbol. **The mana symbol of each circled color is considered part of that card's printed rules text (see rule 613.1) and affects that card's color identity** (see rule 903.4).

kevlahnota:
> Imo it should be implemented by the rules instead like I mentioned in #3652
> 
> > 207.5. One card (Cryptic Spires) has a set of symbols below the text box that represent each color and an ability that instructs a player to circle two of those colors as they create their deck. To circle a color, the player circles (or otherwise clearly indicates) that color's symbol. **The mana symbol of each circled color is considered part of that card's printed rules text (see rule 613.1) and affects that card's color identity** (see rule 903.4).

so what's not working to support this card?

tool4ever:
- the placement in `runPreOpeningHandActions` is obviously wrong, since deck legality check was already passed before. But yes, making it properly work while deck building seems like a bigger task that could also be attempted separately later :)

- > The circled colors are part of the copiable characteristics of the card. If another permanent becomes a copy of Cryptic Spires, its mana ability will produce the same colors as the permanent that was copied.

This means you need to put it as CardState field instead and transfer it in `copyFrom` so CloneStates apply it correctly. 
Then `getChosenColorSpire` can obtain the right colors from `currentState`.

- Isn't _The Prismatic Piper_ choice also broken? I wonder if this solution can be generalized a bit to handle both at least (though it doesn't need new copiable value)

kevlahnota:
ok I think this needs just to save to deck file and load the definition from it upon restart.. and make the clone effect copy the chosenspire values
image

kevlahnota:
@Jetz72  what do you think is the better method to save the papercard info onto the deck and when it's reloaded, composed again from text string from the deck file? With this once you set it, it will stay the same until the user change it again.

kevlahnota:
Ok, seems this is ready, to set the colorid,  the player will need to go to deck editor and set the color id of the card, then save. It will persist until changed again. Though the card script needs to add a new vars so we can remove the "Cryptic Spires" name comparison for future use if somehow wotc add a similar card.

Jetz72:
I'm not sure if any parts of the code base will break now that they have to handle marked instances of PaperCards. But I suspect this is one of those situations where the easiest way to find out is to just try it and watch for bug reports.

tool4ever:
thanks, I should have time to check it out tomorrow :)

forge-gui-desktop/src/main/java/forge/screens/deckeditor/controllers/ACEditorBase.java
```diff
@@ -576,5 +580,24 @@ public void addMoveAlternateItems(final String verb, final String dest) {
                     InputEvent.SHIFT_DOWN_MASK | Toolkit.getDefaultToolkit().getMenuShortcutKeyMask(),
                     InputEvent.ALT_DOWN_MASK | Toolkit.getDefaultToolkit().getMenuShortcutKeyMask());
         }
+        public void addSetColorID() {
+            String label = localizer.getMessage("lblColorIdentity");
+            CardManager cardManager = (CardManager) CDeckEditorUI.SINGLETON_INSTANCE.getCurrentEditorController().getDeckManager();
+            PaperCard existingCard = cardManager.getSelectedItem();
+            int val;
+            if ((val = existingCard.getRules().getSetColorID()) > 0)
```
tool4ever, line 588: wrong logic since it's supposed to be 2?

forge-game/src/main/java/forge/game/ability/effects/CloneEffect.java
```diff
@@ -199,6 +199,8 @@ public void run() {
             if (sa.hasParam("RememberCloneOrigin")) {
                 tgtCard.addRemembered(cardToCopy);
             }
+            // spire
+            tgtCard.setChosenColorID(cardToCopy.getChosenColorID());
```
tool4ever, line 203: @Hanmac
wouldn't it be better if the field was in CardState?
then it doesn't need to be copied around manually (which could also break source since it keeps old values overwritten even if clone effect ends)

forge-gui-desktop/src/main/java/forge/screens/deckeditor/controllers/ACEditorBase.java
kevlahnota: The value is set by the card script setcolorid:val, and the editor wasnt for cryptic spires only. The setting for colorid can be reused.

tool4ever: that's great and all but I'm telling you it just doesn't work right now...

kevlahnota: Seems to work here, are you on the catalog menu? You ahould be in the main or side board

forge-game/src/main/java/forge/game/ability/effects/CloneEffect.java
kevlahnota: @tool4ever  @Hanmac  is there anything left needed except for the change if this will be on the cardstate or not? It's working on the player side now (except the AI part since it doesn't generete it's deck on the fly and put the code there what's the best mana combo to use according to deck land stats to be compliant on rules if that's really needed),

tool4ever: I don't know about anything else right now

kevlahnota: Ok I'll just merge this and I'll just update it if needed more changes.

PR #6575, opened 2024-11-13
Some fixes

### LINKED ISSUE #6551: (FDN) Infernal Vessel's death trigger is not working.

### LINKED ISSUE #6553: Needletooth Pack's Morbid shouldn't trigger at opponent end step.
**Screenshots**
image

### LINKED ISSUE #6573: Strange interaction when cloning XX cost card
A rare occurrence as there is only 1 x cost clone spell, but if you clone any XX creature using Altered Ego and you add X to the cost, the creature enters with double the +1/+1 counters.  I have done this twice with Sporocyst and Altered Ego.

-Have Sporocyst (XXG) in play
-Cast Altered Ego (X2UG) FOR 4
-Result eight +1/+1 counters

### LINKED ISSUE #6574: Potential issue with how Forge handles The Mindskinner replacement effect
**Describe the bug**
I'm not fully sure if how Forge handles Mindskinner is wrong, but at first glance it seems like it would be. It appears that Forge is "bundling" all the sources of damage into a single mill, which doesn't seem right since Mindskinner should replace _each_ source of damage with a mill. At least that's how it reads to me.

**Example:** I have The Mindskinner and Zellix, Sanity Flayer on the field and I attacked one player with 15 creatures. I expected each player to mill 15 times, but there were only two mills, one for each opponent left. As a result I only got two horror tokens even though there were a lot of creatures milled. I think I probably would've gotten a few more tokens if the mills were not bundled.

Also, when displaying the mills in the log, it looks like Forge is carrying over the first player's mills into the second player's mill log. This isn't an issue with Mindskinner, per se, but probably with how Forge records the milled cards and shows it in the log. Minor issue, but would be nice if it didn't do this.

**Expected behavior**
For each source of damage in combat to be a separate mill when The Mindskinner is on the field.

**Screenshots**
This shows that for all those sources of damage, I only got two mills. It also shows the display log issue I mentioned.
Screenshot_20241113_140711

(I need to grab the newest snapshot, but I think it would still have this issue)
 - Forge Version: forge-installer-2.0.00-SNAPSHOT-11.09

tool4ever:
Closes #6573
Closes #6551
Closes #6574
Closes #6553

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -1640,9 +1640,15 @@ public class AbilityUtils {
                 return doXMath(calculateAmount(c, sq[v ? 1 : 2], ctb), expr, c, ctb);
             }
 
+            SpellAbility sa = null;
             if (ctb instanceof SpellAbility) {
-                final SpellAbility sa = (SpellAbility) ctb;
+                sa = (SpellAbility) ctb;
+            } else if (sq[0].contains("xPaid") && ctb instanceof TriggerReplacementBase) {
+                // try avoid fallback
+                sa = ((TriggerReplacementBase) ctb).getOverridingAbility();
+            }
 
+            if (sa != null) {
                 // special logic for xPaid in SpellAbility
                 if (sq[0].contains("xPaid")) {
                     SpellAbility root = sa.getRootAbility();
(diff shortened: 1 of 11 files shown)


forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -1640,9 +1640,15 @@ public static int xCount(Card c, final String s, final CardTraitBase ctb) {
                 return doXMath(calculateAmount(c, sq[v ? 1 : 2], ctb), expr, c, ctb);
             }
 
+            SpellAbility sa = null;
             if (ctb instanceof SpellAbility) {
-                final SpellAbility sa = (SpellAbility) ctb;
+                sa = (SpellAbility) ctb;
+            } else if (sq[0].contains("xPaid") && ctb instanceof TriggerReplacementBase) {
```
Hanmac, line 1646: What exactly is the reason for this?

tool4ever: I need the better logic for the "if X is 5 or more" ravenous check or it still draws

Hanmac: i will need to think of a better way, but i think this is good enough for now

Hanmac (approved):

PR #6577, opened 2024-11-14
AI Attack & getSpellAbilityToPlay Timeout

kevlahnota:


kevlahnota:
> @kevlahnota about the AI timeout part, how about separating it into extra Threads per creature?
> 
> threadSafeIterable would that change make problems for this ConcurrentModificationException?
> 
> When Touching FCollection, how about looking at #3397 ? Especially looking at `SetUniqueList`

I changed the FCollection LinkedList to Lists.newCopyOnWriteArrayList() so it's already thread safe, threadSafeIterable returns the list directly from FCollection, also it uses Sets.newConcurrentHashSet(), (I think this is a shortcut to ConcurrentHashmap.newkeyset).

forge-gui/src/main/java/forge/gamemodes/quest/QuestUtilUnlockSets.java
```diff
@@ -17,12 +17,6 @@
  */
 package forge.gamemodes.quest;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.List;
-import java.util.Map;
```
tool4ever, line 24: I wonder why tests didn't fail from checkstyle before 🤔

forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -1565,79 +1596,94 @@ private SpellAbility chooseSpellAbilityToPlayFromList(final List<SpellAbility> a
...
+                if (opinion != AiPlayDecision.WillPlay)
+                    return 0;
 
-            return sa;
+                spells.add(sa);
+                return 0;
+            }));
+        }
+        //timeout 5 seconds? even the AI don't acquire all, there should be SA to cast if valid
+        CompletableFuture<?>[] futuresArray = futures.toArray(new CompletableFuture<?>[0]);
```
tool4ever, line 1669: please make it an opt-in option or users will just complain about passive AI in some cases and we can't determine easily if API logic was bad or just boardstate too messy

personally I prefer to wait instead of taking a suboptimal play anyway

kevlahnota: I can add that option 5 sec, 10 sec, maybe for off 1 hour is good enough for you to timeout?

tool4ever: 1 hour probably means deadlock haha
5sec, 1min, off or something?

kevlahnota: I'll just put 1 hour for off (most likely it will be finish unless an error occurred)

Agetian:
I agree with @tool4ever's observations, and after these tweaks it looks like a very useful and nice option! :)

Hanmac:
> I changed the FCollection LinkedList to Lists.newCopyOnWriteArrayList() so it's already thread safe, threadSafeIterable returns the list directly from FCollection, also it uses Sets.newConcurrentHashSet(), (I think this is a shortcut to ConcurrentHashmap.newkeyset).

I just thought we could clean up some code if we make:
`FCollection = new SetUniqueList(Lists.newCopyOnWriteArrayList(), Sets.newConcurrentHashSet())`

less own code means less errors ;P

forge-core/src/main/java/forge/util/collect/FCollection.java
```diff
@@ -43,12 +42,38 @@ public static <T> FCollection<T> getEmpty() {
...
      */
-    private final LinkedList<T> list = Lists.newLinkedList();
+    private List<T> LIST;
+    private List<T> list() {
+        List<T> result = LIST;
+        if (result == null) {
+            synchronized (this) {
+                result = LIST;
+                if (result == null) {
+                    result = Lists.newCopyOnWriteArrayList();
```
tool4ever, line 70: by switching the default iterator you will produce crashes imo

because some code uses add() or remove()
and LinkedList did support it...

but read JavaDoc of `CopyOnWriteArrayList`

```
 * The iterator will not reflect additions, removals, or changes to
 * the list since the iterator was created.  Element-changing
 * operations on iterators themselves ({@code remove}, {@code set}, and
 * {@code add}) are not supported. These methods throw
 * {@code UnsupportedOperationException}.
```

Hanmac: In most of our use cases, we need to make a copy via threadSafeIterable anyway, 
otherwise, we get `ConcurrentModificationException`

And the Test seems to pass so far?

tool4ever: I trust my brain/IDE more here ;)

Here's one example from PlayAi:
image

this will call remove() on a CardCollection, which is a FCollection
I don't see a way this doesn't crash now

Jetz72:
> I changed the FCollection LinkedList to Lists.newCopyOnWriteArrayList() so it's already thread safe

Isn't that going to be a massive hit to performance for a lot of the single-threaded code? `CopyOnWriteArrayList` really does copy the whole list every time you add or remove list items. Adding a single item goes from O(1) to O(N), and adding iteratively goes from O(M) to O(N * M). Using it for every single FCollection might be excessive.

forge-core/src/main/java/forge/util/collect/FCollection.java
Hanmac: imo a better way would be to get the ones into a `toRemove` collection, and then remove them after the loop?

forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -1547,8 +1574,12 @@ private SpellAbility getSpellAbilityToPlay() {
 
         Iterables.removeIf(saList, spellAbility -> { //don't include removedAI cards if somehow the AI can play the ability or gain control of unsupported card
             // TODO allow when experimental profile?
-            return spellAbility.isLandAbility() || (spellAbility.getHostCard() != null && ComputerUtilCard.isCardRemAIDeck(spellAbility.getHostCard()));
+            return spellAbility.isLandAbility() || (spellAbility.getHostCard() != null && ComputerUtilCard.isCardRemAIDeck(spellAbility.getHostCard()))  || !spellAbility.canCastTiming(player);
```
tool4ever, line 1577: does this fix anything? it's already checked by canPlaySa so redundant check just adds extra load

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -144,8 +147,15 @@ public static boolean handlePlayingSpellAbility(final Player ai, SpellAbility sa
...
         }
-        //Should not arrive here
-        System.out.println("AI failed to play " + sa.getHostCard());
+        // FIXME: Should not arrive here, though the card seems to be stucked on stack zone and invalidated and nowhere to be found, try to put back to original zone and maybe try to cast again if possible at later time?
+        System.out.println("[" + sa.getActivatingPlayer() + "] AI failed to play " + sa.getHostCard() + " [" + sa.getHostCard().getZone() + "]");
+        sa.setSkip(true);
+        if (host != null && hz != null) {
+            Card c = game.getAction().moveTo(hz.getZoneType(), host, null, null);
+            for (SpellAbility csa : c.getSpellAbilities()) {
+                csa.setSkip(true);
```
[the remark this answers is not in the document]
tool4ever: oh nm, I see you do it more selectively in ComputerUtilMana anyway
hmn I guess then it's easiest for now

forge-game/src/main/java/forge/game/cost/CostPayment.java
```diff
@@ -229,8 +229,8 @@ public final boolean payComputerCosts(final CostDecisionMakerBase decisionMaker)
      * @return a {@link forge.game.mana.Mana} object.
      */
     public static Mana getMana(final Player player, final ManaCostShard shard, final SpellAbility saBeingPaidFor,
-            final byte colorsPaid, Map<String, Integer> xManaCostPaidByColor) {
-        final List<Pair<Mana, Integer>> weightedOptions = selectManaToPayFor(player.getManaPool(), shard,
+            final byte colorsPaid, Map<String, Integer> xManaCostPaidByColor) { // player.getManaPool() is not threadsafe if this is called somwewhere concurrently
+        final List<Pair<Mana, Integer>> weightedOptions = selectManaToPayFor(new ManaPool(player), shard,
```
tool4ever, line 233: you're just breaking AI ability to pay from pool completely because you're now calling this with new empty one??

forge-core/src/main/java/forge/util/collect/FCollection.java
kevlahnota: > I trust my brain/IDE more here ;)
> 
> Here's one example from PlayAi: image
> 
> this will call remove() on a CardCollection, which is a FCollection I don't see a way this doesn't crash now

Can you write a simple test so it will crash if the iterator add or remove is not supported?

kevlahnota:
> > I changed the FCollection LinkedList to Lists.newCopyOnWriteArrayList() so it's already thread safe
> 
> Isn't that going to be a massive hit to performance for a lot of the single-threaded code? `CopyOnWriteArrayList` really does copy the whole list every time you add or remove list items. Adding a single item goes from O(1) to O(N), and adding iteratively goes from O(M) to O(N * M). Using it for every single FCollection might be excessive.

I did profile it and seems ok so far. I don't know any concurrent list that is available. How does the change react to your machine when using  it with lits of token/card generation?

forge-ai/src/main/java/forge/ai/AiController.java
kevlahnota: on my test, the amount of sa is reduced though the canplaysa check is used for counter, the saList = ComputerUtilAbility.getSpellAbilities(cards, player); don't have any filters for cancasttiming.

forge-core/src/main/java/forge/util/collect/FCollection.java
tool4ever:
```
    @Test
    void testBadIteratorLogic() {
        CardCollection cc = new CardCollection(new Card(1, null));
        Iterator<Card> it = cc.iterator();
        it.next();
        it.remove();
    }
```

forge-game/src/main/java/forge/game/cost/CostPayment.java
kevlahnota: how come? they should be tally or it will not cast anything. Is there anything that can be reproducible with this change?

tool4ever: Please read the code...
if you call `selectManaToPayFor` with an always empty mana pool `weightedOptions` will always be empty
**this function can no longer return anything except null**

forge-core/src/main/java/forge/util/collect/FCollection.java
kevlahnota: ah yes, i see it so does card collection use this? the docs says use list.remove() but I think i need to update those unless I can change to different concurrentlist which isn't available on default libs

forge-ai/src/main/java/forge/ai/AiController.java
tool4ever: I don't understand what you're saying but pasting checks left and right like that is just asking for trouble
The SA isn't fully prepared for checking at this point, you will produce some misplay

forge-game/src/main/java/forge/game/cost/CostPayment.java
kevlahnota: I rolled back it a while ago

kevlahnota:
@Hanmac I encountered concurrent modification on Card.class for the Iterable getchangedcardtypes. Is it safe to return ImmutableList.copyOf instead of Iterables.unmodifiableIterable?

Hanmac:
> @Hanmac I encountered concurrent modification on Card.class for the Iterable getchangedcardtypes. Is it safe to return ImmutableList.copyOf instead of Iterables.unmodifiableIterable?

I saw the error too on sentry but I can't explain it yet

Do you got a case to reproduce that?

If yes I could make a short hot fix

forge-core/src/main/java/forge/util/collect/FCollection.java
tool4ever: I suppose the direct calls have been avoided, I probably don't know enough about concurrency to consider any other implications from this 🤷‍♂️

kevlahnota:
> > @Hanmac I encountered concurrent modification on Card.class for the Iterable getchangedcardtypes. Is it safe to return ImmutableList.copyOf instead of Iterables.unmodifiableIterable?
> 
> I saw the error too on sentry but I can't explain it yet
> 
> Do you got a case to reproduce that?
> 
> If yes I could make a short hot fix

no but if it occurs before this change then the bug already exists before. No idea why it happens though

kevlahnota:
@tool4ever I changed the ManaPool floatingmana -> arrayListMultimap to a custom ConcurrentMultimap implementation, seems to work fine now but the AI seems to pick or tap a single land at end of turn? I wonder why it does that and on the log it says AI picked {W}.

forge-core/src/main/java/forge/util/collect/ConcurrentMultiMap.java
```diff
@@ -0,0 +1,135 @@
...
+import com.google.common.collect.Multiset;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Queue;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+public class ConcurrentMultiMap<K, V> {
```
Hanmac, line 13: Why exactly using your on class instead of using the collection4 package?
It wouldn't be much of a problem, I was probably going to use something of that package anyway

And should this implement https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Multimap.html ?

kevlahnota: I tried implementing the guava multimap but you need to fulfill all those overrides (seems overkill for the manapool sake version), though the earlier version from collection4 package works too.

Hanmac:
> > > @Hanmac I encountered concurrent modification on Card.class for the Iterable getchangedcardtypes. Is it safe to return ImmutableList.copyOf instead of Iterables.unmodifiableIterable?
> > 
> > 
> > I saw the error too on sentry but I can't explain it yet
> > Do you got a case to reproduce that?
> > If yes I could make a short hot fix
> 
> no but if it occurs before this change then the bug already exists before. No idea why it happens though

Yeah, i thought i could find a case to make the error reproduceable.

@tool4ever for a smaller HOTFIX before this MR, what about using getchangedcardtypes a ImmutableList.copy?
And maybe we could cache the result of getTypeWithChanges ? and or maybe cache the List?
Because it is also used for `hasRemoveIntrinsic`

forge-ai/src/main/java/forge/ai/AiCardMemory.java
```diff
@@ -79,21 +79,21 @@ public enum MemorySet {
     private final Set<Card> memRevealedCards;
 
     public AiCardMemory() {
```
Hanmac, line 81: @tool4ever @kevlahnota might it be cleaner if we change the structure into this?

`Map<MemorySet, Set<Card>>`, and have the whole logic inside?
Then we don't need extra properties for each MemorySet-Type

forge-core/src/main/java/forge/util/collect/ConcurrentMultiMap.java
tool4ever: To me the question is why does ManaPool need to be concurrent now?

If you're wanting to check multiple SA for p(l)ayability in some non-deterministic way won't some checks possibly fail if the pool is already being drained from another Future/Thread at the same time?!

forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -1574,79 +1605,99 @@ private SpellAbility chooseSpellAbilityToPlayFromList(final List<SpellAbility> a
...
-                        continue;
-                    }
                 }
-            }
 
-            sa.setActivatingPlayer(player, true);
-            SpellAbility root = sa.getRootAbility();
+                // dumb check needed? don't really know why the AI tapped the land for nothing at EOT of his opponent.
+                // without this, you can see the weird tapping of single land -> mana ability. with this, it doesn't confuse human player
+                if (sa.isManaAbility() && game.getPhaseHandler() != null && game.getPhaseHandler().is(PhaseType.END_OF_TURN) && !game.getPhaseHandler().isPlayerTurn(player))
```
tool4ever, line 1657: are you saying it only happens on this branch?
then some logic got messed up somehow and should be fixed instead...
otherwise I can try and fix it on master

forge-core/src/main/java/forge/util/collect/ConcurrentMultiMap.java
kevlahnota: I thought we only get a single spellability to play, the only difference is they're computed concurrently but ai will play at a time, by the time the ai cast or activate the top sa, wouldn't the next sa checked again if feasible to play? If you have better idea to access the manapool without concurrent modification then please enlighten.

Jetz72:
> I did profile it and seems ok so far. I don't know any concurrent list that is available. How does the change react to your machine when using it with lots of token/card generation?

Nothing significant in gameplay, though it's hard to pin down what's causing sluggishness just by watching, and I haven't had much luck finding a reasonably priced profiler on Linux. I ran some basic benchmarks - creating and destroying a ton of tokens, shuffling a deck a bunch of times. Destroying things seems unaffected funnily enough, but other operations do experience exponential slowdown with higher numbers and the CopyOnWriteArrayList.


> ### Creating a deck and shuffling it 10000 times
> 
> #### ArrayList
> ##### 100 card deck
> Create: 0.15320649
> Shuffle: 0.413653924
> 
> ##### 1000 card deck
> Create: 0.42519795
> Shuffle: 2.900546915
> 
> ##### 3000 card deck
> Create: 0.937716733
> Shuffle x 10000: 9.418635845
> 
> #### CopyOnWriteArrayList
> ##### 100 card deck
> Create: 0.155789148
> Shuffle: 0.755845171
> 
> ##### 1000 card deck
> Create: 0.466518411
> Shuffle: 15.762016363
> 
> ##### 3000 card deck
> Create: 3.240357319
> Shuffle: 2:04.585301495
> 
> 
> ### Creating and destroying tokens
> 
> #### ArrayList
> ##### 100 tokens
> Create: 0.464475841
> Destroy: 0.11633868
> 
> ##### 1000 tokens
> Create: 7.275628844
> Destroy: 6.799610756
> 
> ##### 3000 tokens
> Create: 1:03.685734948
> Destroy: 3:49.847811285
> 
> 
> #### CopyOnWriteArrayList
> 
> ##### 100 tokens
> Create: 0.45074327
> Destroy: 0.112953013
> 
> ##### 1000 tokens
> Create: 12.786269518
> Destroy: 7.194827712
> 
> ##### 3000 tokens
> Create: 3:39.469679407
> Destroy: 4:03.47748994


Obviously these numbers are far above what you could expect to have in an actual game, but Forge already has a number of issues with scaling to larger board states and this would compound that somewhat. A custom implementation may be better than the out-of-the-box list. You could also just make separate thread-safe collections for concurrent operations to minimize the impact, perhaps either a subclass of FCollection or a method that instructs it to switch its underlying list to a thread-safe one when it's needed.

forge-ai/src/main/java/forge/ai/AiController.java
kevlahnota: I was thinking on the sorting of SA. On this branch I moved the sort after the SA is gathered with timeout since the order will not be the same and needs to be sorted again. Gonna try tomorrow again.

Hanmac:
> You could also just make separate thread-safe collections for concurrent operations to minimize the impact, perhaps either a subclass of FCollection or a method that instructs it to switch its underlying list to a thread-safe one when it's needed.

That's what the `threadSafeIterable`` function was for, while not changing the underlying structure, it still returns a new list that then can be safely iterated.

kevlahnota:
> > I did profile it and seems ok so far. I don't know any concurrent list that is available. How does the change react to your machine when using it with lots of token/card generation?
> 
> Nothing significant in gameplay, though it's hard to pin down what's causing sluggishness just by watching, and I haven't had much luck finding a reasonably priced profiler on Linux. I ran some basic benchmarks - creating and destroying a ton of tokens, shuffling a deck a bunch of times. Destroying things seems unaffected funnily enough, but other operations do experience exponential slowdown with higher numbers and the CopyOnWriteArrayList.
> 
> > ### Creating a deck and shuffling it 10000 times
> > #### ArrayList
> > ##### 100 card deck
> > Create: 0.15320649
> > Shuffle: 0.413653924
> > ##### 1000 card deck
> > Create: 0.42519795
> > Shuffle: 2.900546915
> > ##### 3000 card deck
> > Create: 0.937716733
> > Shuffle x 10000: 9.418635845
> > #### CopyOnWriteArrayList
> > ##### 100 card deck
> > Create: 0.155789148
> > Shuffle: 0.755845171
> > ##### 1000 card deck
> > Create: 0.466518411
> > Shuffle: 15.762016363
> > ##### 3000 card deck
> > Create: 3.240357319
> > Shuffle: 2:04.585301495
> > ### Creating and destroying tokens
> > #### ArrayList
> > ##### 100 tokens
> > Create: 0.464475841
> > Destroy: 0.11633868
> > ##### 1000 tokens
> > Create: 7.275628844
> > Destroy: 6.799610756
> > ##### 3000 tokens
> > Create: 1:03.685734948
> > Destroy: 3:49.847811285
> > #### CopyOnWriteArrayList
> > ##### 100 tokens
> > Create: 0.45074327
> > Destroy: 0.112953013
> > ##### 1000 tokens
> > Create: 12.786269518
> > Destroy: 7.194827712
> > ##### 3000 tokens
> > Create: 3:39.469679407
> > Destroy: 4:03.47748994
> 
> Obviously these numbers are far above what you could expect to have in an actual game, but Forge already has a number of issues with scaling to larger board states and this would compound that somewhat. A custom implementation may be better than the out-of-the-box list. You could also just make separate thread-safe collections for concurrent operations to minimize the impact, perhaps either a subclass of FCollection or a method that instructs it to switch its underlying list to a thread-safe one when it's needed.

Hmm FCollection needs List without duplicates (the cards  has its own id, so they're all unique I guess), @Hanmac suggested the SetUniqueList but it's another addition that needs to be tested. Eitherway, the copyonwriteareaylist works fine on less than 1000 objects but will suffer when used on generating ai genetic decks (but that generation is slow even before the change of fcollection)

kevlahnota:
> > You could also just make separate thread-safe collections for concurrent operations to minimize the impact, perhaps either a subclass of FCollection or a method that instructs it to switch its underlying list to a thread-safe one when it's needed.
> 
> That's what the `threadSafeIterable`` function was for, while not changing the underlying structure, it still returns a new list that then can be safely iterated.

I was misled by the name, since why it needs to return another list if the list structure is already threadsafe/concurrent.

forge-core/src/main/java/forge/util/collect/ConcurrentMultiMap.java
tool4ever: - checking SA includes mana test
- if more than one is available and the pool isn't empty then how are you keeping it synchronized when AI checks A + B at roughly the same time? **maybe you didn't consider it but the pool is actively being modified/drained even when only testing**

the whole AI logic isn't set up to jump around between threads even if the data structures might not crash from it now lol

Jetz72: Would it be cleaner + efficient enough to implement threading through something more like the GameSimulator AI - copy the whole game state to separate threads and evaluate them without any risk of tampering with shared resources?

forge-ai/src/main/java/forge/ai/AiController.java
kevlahnota: I don't want to argue on what you believe but the point of this change is to have a timeout regardless if the boardstate is complex or not. The current problem is where the number to process those objects using sequential approach will eventually be slower than processing them concurrently and anything that's within the timeout will have results than waiting forever to wait for the game to proceed. Also the threaded computation are only specifically selected on the most expensive computation so it will still have results of it timeout or not.  This was intended to make the game move on those state where it's hard to process  all. I don't know how long you play the game but it's tiresome to wait for the AI if it's not producing results without timeout. All this change was to have a better play experience than having a perfect AI logic that will stall over time if the boardstate becomes to much to process.

tool4ever: yes I understand all that and I'm not against it

frankly often when I debugged some slowdown it was at least partially caused by very few cards/API which run a badly optimized path (often custom written AILogic$)

but anyway my point was why another structural change on top of this?
why force check all API first when there's a good chance you could find a playable candidate way before the last element?
is it because you have no control over the order the Futures get calculated so you can't just return with one of the first matches?

kevlahnota: Because you don't know what will be finish first specially if the spellabilities is big. Example you have 10 creatures that have deals damage to target creature. And the opponent have a lot of possible target. If the first one didn't find good target, it will take the next one again and will have the same result as the first one leaving the computation to process all of them sequentially to just not find anything feasible. If you process them concurrently, you will have faster results. The cons of this approach is if those abilities are unique, you need to sort them after. unlike the previous implementation that is already sorted (if verify transitivity didn't occur) and if the logic passes, it will be returned, the cons of that approach is if the top sa will fail the logic, it will process the next one and wait if it passes or not and so on.

forge-gui/src/main/java/forge/gamemodes/quest/QuestUtilUnlockSets.java
tool4ever: I guess the redundant logic just isn't smart enough?

forge-ai/src/main/java/forge/ai/AiCardMemory.java
kevlahnota: > @tool4ever @kevlahnota might it be cleaner if we change the structure into this?
> 
> `Map<MemorySet, Set<Card>>`, and have the whole logic inside? Then we don't need extra properties for each MemorySet-Type

already updated the logic

kevlahnota:
@Hanmac @Agetian @tool4ever is there any changes needed?  FCollection and ManaPool shouldn't throw ConcurrentModification anymore and seems to work fine.

kevlahnota:
> Well my main concern isn't really addressed: while you might prevent plain out crashing you're still creating race conditions if you allow multiple threads to access the manapool concurrently. Either the floating mana or the arrays for conversion will have a chance of getting messed up. I don't see how you're working around this unless you mark `canPlayAndPayFor` or something as `synchronized`. (And no, I can't give you a fully working example because that's just the nature of nondeterminism)
> 
> Also the "dumb check" you had to add clearly hints that something with your custom data alternative doesn't fully work and will break some existing AI behaviour.

Ok I'll revert the that Manapool and implement a lock similar to what I did to FCollection and remove the dumb check and restore the sort above it (I still think that's is the cause the of the issue) to satisfy your concern.

forge-gui/src/main/java/forge/gamemodes/match/HostedMatch.java
```diff
@@ -155,6 +155,7 @@ public void startGame() {
 
         game = match.createGame();
         game.EXPERIMENTAL_RESTORE_SNAPSHOT = FModel.getPreferences().getPrefBoolean(FPref.MATCH_EXPERIMENTAL_RESTORE);
+        game.AI_TIMEOUT = FModel.getPreferences().getPrefInt(FPref.MATCH_AI_TIMEOUT);
```
tool4ever, line 158: this logic doesn't work because the players get created with the default before
(It's probably a good idea to have more than one value though in case we want to combine it with AI difficulty or something later)

tool4ever:
👍
(I think some parts still look a bit like duct tape but thanks for all the work. And maybe the concurrent SA approach can be reused later for simulation together with normal one, we could talk on Discord in case more AI stuff is on your roadmap)

forge-gui/src/main/java/forge/gamemodes/match/HostedMatch.java
kevlahnota: remove the definition on player and just take the value from the game

forge-core/src/main/java/forge/util/collect/FCollection.java
```diff
@@ -43,12 +44,40 @@ public static <T> FCollection<T> getEmpty() {
     /**
      * The {@link Set} representation of this collection.
      */
-    private final Set<T> set = Sets.newHashSet();
+    private Set<T> SET;
```
Hanmac, line 47: Just a question, why don't you use the default value anymore? I mean there shouldn't be a time where the set/list is null?

kevlahnota: Everytime a new cardcollection is created and isn't used because of invalid condition saves a bit of memory. On my local copy when i profile it (Card, player, spellability, etc), unused vars and not initialized if not needed produce good results.

tool4ever (approved):

forge-core/src/main/java/forge/util/collect/FCollection.java
Jetz72: Is that the result of a bunch of fields using them as the default value? Like these in `Card.java`?

```java
    private List<GameCommand> leavePlayCommandList = Lists.newArrayList();
    private final List<GameCommand> untapCommandList = Lists.newArrayList();
    private final List<GameCommand> changeControllerCommandList = Lists.newArrayList();
    private final List<GameCommand> unattachCommandList = Lists.newArrayList();
    private final List<GameCommand> faceupCommandList = Lists.newArrayList();
    private final List<GameCommand> facedownCommandList = Lists.newArrayList();
    private final List<GameCommand> phaseOutCommandList = Lists.newArrayList();
    private final List<Object[]> staticCommandList = Lists.newArrayList();
```

Might be worth giving the same lazy-init treatment to other commonly used classes.

kevlahnota: I refactor the initial lazy init and added a lazy initialization from the article here: https://4comprehension.com/leveraging-lambda-expressions-for-lazy-evaluation-in-java/

kevlahnota: > @kevlahnota checkout this:
> 
> https://guava.dev/releases/21.0/api/docs/com/google/common/base/Suppliers.html#memoize-com.google.common.base.Supplier- `public static <T> Supplier<T> memoize(Supplier<T> delegate)`

done

kevlahnota: @Hanmac @Agetian Anything else needed?

Agetian: Can't think of anything at the moment :)

PR #6606, opened 2024-11-22
Miracle&Madness: use ManaCost instead of CardManaCost

Hanmac:
This fixes the Problem with Miracle and Madness not using the right value for Spells to cast.

I'm not 100% happy with PlayReduceCost yet, because i think this should reduce the cost BEFORE you choose values for X or how 2-hybrid costs are chosen.

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
@@ -1460,12 +1460,17 @@ public class CardFactoryUtil {
             final String[] k = keyword.split(":");
             final String manacost = k[1];
 
-            final String trigStr = "Mode$ Exiled | ValidCard$ Card.Self | Madness$ True | Secondary$ True"
-                    + " | TriggerDescription$ Play Madness " + ManaCostParser.parse(manacost) + " - " + card.getName();
+            StringBuilder desc = new StringBuilder("Play Madness ");
+            if (!"ManaCost".equals(manacost)) {
+                desc.append(ManaCostParser.parse(manacost)).append(" ");
+            }
+            desc.append(" - " + card.getName());
+
+            final String trigStr = "Mode$ Exiled | ValidCard$ Card.Self | Secondary$ True | TriggerDescription$ " + desc.toString();
 
             final String playMadnessStr = "DB$ Play | Defined$ Self | ValidSA$ Spell | PlayCost$ " + manacost +
                     " | ConditionDefined$ Self | ConditionPresent$ Card.StrictlySelf+inZoneExile" +
-                    " | Optional$ True | RememberPlayed$ True | Madness$ True";
+                    " | Optional$ True | RememberPlayed$ True";
 
             final String moveToYardStr = "DB$ ChangeZone | Defined$ Self.StrictlySelf | Origin$ Exile" +
                     " | Destination$ Graveyard | TrackDiscarded$ True | ConditionDefined$ Remembered | ConditionPresent$ Card" +
@@ -1518,7 +1523,10 @@ public class CardFactoryUtil {
             final String manacost = k[1];
             final String abStrReveal = "DB$ Reveal | Defined$ You | RevealDefined$ Self"
                     + " | MiracleCost$ " + manacost;
-            final String abStrPlay = "DB$ Play | Defined$ Self | Optional$ True | PlayCost$ " + manacost;
+            String abStrPlay = "DB$ Play | Defined$ Self | Optional$ True | PlayCost$ " + manacost;
+            if (k.length >= 2) {
+                abStrPlay += " | PlayReduceCost$ " + k[2];
+            }
 
             String revealed = "DB$ ImmediateTrigger | TriggerDescription$ CARDNAME - Miracle";
 
--- a/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
@@ -2395,8 +2395,17 @@ public class CardFactoryUtil {
 
             inst.addReplacement(re);
         } else if (keyword.startsWith("Madness")) {
+            final String[] k = keyword.split(":");
+            final String manacost = k[1];
+
+            StringBuilder desc = new StringBuilder("Madness");
+            if (!"ManaCost".equals(manacost)) {
+                desc.append(" ").append(ManaCostParser.parse(manacost));
+            }
+            desc.append(": If you discard this card, discard it into exile.");
+
             String repeffstr = "Event$ Moved | ActiveZones$ Hand | ValidCard$ Card.Self | Discard$ True | Secondary$ True "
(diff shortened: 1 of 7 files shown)


forge-game/src/main/java/forge/game/trigger/TriggerExiled.java
```diff
@@ -84,7 +84,7 @@ public final boolean performTest(final Map<AbilityKey, Object> runParams) {
             return false;
         }
 
-        if (hasParam("Madness")) {
+        if (isKeyword(Keyword.MADNESS)) {
             if (cause == null || !cause.isKeyword(Keyword.MADNESS)) {
```
tool4ever, line 88: the check here could probably use some more accuracy:
currently multiple _Falkenrath Gorger_ cause more than one trigger, but only the one doing the exiling should apply

tool4ever:
it's actually starting to look like they want the cost reduction to work on X shards too 🤔

So maybe consistency with external foretell is something to consider later

tool4ever:
But the stupid thing is some judges say it still needs to be treated a bit differently:
464637515_582731397426704_445586349640712613_n

forge-game/src/main/java/forge/game/trigger/TriggerExiled.java
Hanmac: They are not the same "ctb", but part of the same keyword. So checking if they are the same keyword should work

Hanmac:
> But the stupid thing is some judges say it still needs to be treated a bit differently: 464637515_582731397426704_445586349640712613_n

I don't think that is the case, I need to ask Dunk again if he answers

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -1460,12 +1460,17 @@ public static void addTriggerAbility(final KeywordInterface inst, final Card car
...
-                    + " | TriggerDescription$ Play Madness " + ManaCostParser.parse(manacost) + " - " + card.getName();
+            StringBuilder desc = new StringBuilder("Play Madness ");
+            if (!"ManaCost".equals(manacost)) {
+                desc.append(ManaCostParser.parse(manacost)).append(" ");
+            }
+            desc.append(" - " + card.getName());
+
+            final String trigStr = "Mode$ Exiled | ValidCard$ Card.Self | Secondary$ True | TriggerDescription$ " + desc.toString();
 
             final String playMadnessStr = "DB$ Play | Defined$ Self | ValidSA$ Spell | PlayCost$ " + manacost +
```
tool4ever, line 1471: needs ValidSA so you can't play enchantment lands 🐛

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -1518,7 +1523,10 @@ public static void addTriggerAbility(final KeywordInterface inst, final Card car
             final String manacost = k[1];
             final String abStrReveal = "DB$ Reveal | Defined$ You | RevealDefined$ Self"
                     + " | MiracleCost$ " + manacost;
-            final String abStrPlay = "DB$ Play | Defined$ Self | Optional$ True | PlayCost$ " + manacost;
+            String abStrPlay = "DB$ Play | Defined$ Self | Optional$ True | PlayCost$ " + manacost;
```
tool4ever, line 1526: needs ValidSA$ so you can't play enchantment land 🐛

Hanmac:
I found this:
https://github.com/Estellarium/Aminatous-Enchantment-Discount/blob/main/README.md

forge-game/src/main/java/forge/game/trigger/TriggerExiled.java
tool4ever: Self should check same host, but your way might be even safer

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
tool4ever: the branch for `tgtSA.isLandAbility()` happens before that though

tool4ever:
A random Git repository just for this? I trust it 0% :p

The bigger question for me is if Instant/Sorcery on e.g. _Virtue of Strength_ should really be allowed...

But I think we can just merge this after my findings above get eliminated, then we'd still have 3 cases fixed good 👍

forge-game/src/main/java/forge/game/trigger/TriggerExiled.java
Hanmac: doesn't work, because TriggerExiled is on the new Card object, so the Keyword Instance is different (and the Id of the Trigger is different too)

Maybe TriggerExiled should use LKI?

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
Hanmac: hm imo the isLandAbility Check there should check if `"ManaCost".equals(sa.getParam("PlayCost"))`

Hanmac: i added a check to isLandAbility part

i wanted to move the parts even more up, like when spells and cards are checked, but i failed with the java syntax a bit

forge-game/src/main/java/forge/game/trigger/TriggerExiled.java
tool4ever: that seems wrong / not worth the trouble
can `kw.getStatic()` be a match?

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
tool4ever: Well I don't see how that's better since for something like _Strength of the Harvest_ you shouldn't even be allowed to choose land backside to confuse things
but it's your PR ;)

Hanmac: i'm trying to move the check up

right now, i added the check to the whole SpellAbility List

forge-game/src/main/java/forge/game/trigger/TriggerExiled.java
Hanmac: i added getStatic check

what i find curious is that the two Madness didn't asked me to choose the ReplacementEffect 
(can you help debug this?)

Hanmac: hm i need to test it with intrinsic Madness too

tool4ever: well looks like that doesn't work either, the description for the RE is too hardcoded

tool4ever: I think for intrinsic the logic might still need a small change
either both are null or equal?

tool4ever (dismissed):
LGTM

tool4ever (approved):

PR #6609, opened 2024-11-23
YDSK: improvising_aerialist.txt + support

Northmoc:


forge-gui/res/cardsfolder/i/improvising_aerialist.txt
```diff
@@ -0,0 +1,7 @@
+Name:Improvising Aerialist
+ManaCost:1 W
+Types:Creature Human Survivor
+PT:3/2
+T:Mode$ Phase | Phase$ Main | PhaseCount$ 2 | ValidPlayer$ You | PresentDefined$ Self | IsPresent$ Card.tapped | Execute$ TrigPerpetual | TriggerDescription$ Survival — At the beginning of your second main phase, if CARDNAME is tapped, CARDNAME and the top creature card in your library without flying perpetually gain flying.
+SVar:TrigPerpetual:DB$ Pump | PumpZone$ Battlefield,Library | Defined$ Self & ValidLibrary Creature.TopLibraryCreature | KW$ Flying | Duration$ Perpetual
```
squee1968, line 6: Shouldn't the defined parameter be Defined$ Self & ValidLibrary Creature.TopLibraryCreature.withoutFlying?

Northmoc: The syntax doesn't work that way but your overall point is correct!

paulsnoops (dismissed):

forge-game/src/main/java/forge/game/card/CardProperty.java
```diff
@@ -650,6 +650,10 @@ public static boolean cardHasProperty(Card card, String property, Player sourceC
             if (cards.isEmpty() || !card.equals(cards.get(0))) {
                 return false;
             }
+        } else if (property.startsWith("TopLibrary_")) {
+            CardCollection cards = new CardCollection(card.getOwner().getCardsIn(ZoneType.Library));
```
tool4ever, line 654: new CardCollection() shouldn't be needed?

tool4ever: Ideally all these properties for Top + Bottom would be cleaned up together (with only the Collections.reverse as difference when needed) 🤔
but that could be done later, feel free to resolve

tool4ever (approved):
should work ok like this

PR #6342, opened 2024-10-15
Add Lurker in the Deep (YDSK)

TwentyToedToad:

PR #6615, opened 2024-11-23
YDSK: lurker_in_the_deep.txt + support

Northmoc:
Replaces #6342


tool4ever (approved):

Hanmac:
@Northmoc the isEmpty checks break Android

PR #6616, opened 2024-11-23
UNF: eelectrocute.txt + support

Northmoc:


forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -2320,6 +2320,12 @@ public static int xCount(Card c, final String s, final CardTraitBase ctb) {
             return doXMath(player.getNumRollsThisTurn(), expr, c, ctb);
         }
 
+        if (sq[0].startsWith("YouRolledThisTurn")) {
+            int n = Integer.parseInt(sq[0].substring(17));
+            boolean didRoll = player.getDiceRollsThisTurn().contains(n);
+            return doXMath(calculateAmount(c, didRoll ? "1" : "0", ctb), expr, c, ctb);
```
Jetz72, line 2326: Since you're using a list to track rolls this turn, you could maybe broaden the functionality by outputting the number of times the given number was rolled. Alternatively, could store it as a set instead and ignore duplicate rolls right from the start.

Northmoc: Whichever we choose WotC will pivot the other way... but something that cares about "how many ones you've rolled this turn" or similar does sound believable

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -2320,6 +2320,11 @@ public static int xCount(Card c, final String s, final CardTraitBase ctb) {
             return doXMath(player.getNumRollsThisTurn(), expr, c, ctb);
         }
 
+        if (sq[0].startsWith("YouRolledThisTurn")) {
+            int n = Integer.parseInt(sq[0].substring(17));
```
Hanmac, line 2324: how about using calculate Amount for the Parameter too instead of parseInt?

Hanmac (approved):

PR #6618, opened 2024-11-23
magar_of_the_magic_strings.txt bug fix

### LINKED ISSUE #6422: NPE when casting spells copied with Magar of the Magic Strings
**Describe the bug**
When the 3/3 creature created by Magar of the Magic Strings deals damage and triggers Forge crashes with an NPE.

**To Reproduce**
1. Use Magar of the Magic Strings ability
2. Attack with the creature to trigger the ability
3. Forge will softcrash that game with an NPE

 - Forge Version: forge-2.0.00-SNAPSHOT-10.22

Northmoc:
closes #6422

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -3004,7 +3004,7 @@ public class AbilityUtils {
                 sp.setCardState(original);
                 list.add(sp);
             }
-            if (tgtCard.isModal()) {
+            if (tgtCard.isModal() && tgtCard.getState(CardStateName.Modal) != null) {
                 collectSpellsForPlayEffect(list, tgtCard.getState(CardStateName.Modal), controller, withAltCost);
             }
         }
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -3004,7 +3004,7 @@ public class AbilityUtils {
                 sp.setCardState(original);
                 list.add(sp);
             }
-            if (tgtCard.isModal() && tgtCard.getState(CardStateName.Modal) != null) {
+            if (tgtCard.isModal() && tgtCard.hasState(CardStateName.Modal)) {
                 collectSpellsForPlayEffect(list, tgtCard.getState(CardStateName.Modal), controller, withAltCost);
             }
         }


forge-gui/res/cardsfolder/m/magar_of_the_magic_strings.txt
```diff
@@ -2,10 +2,13 @@ Name:Magar of the Magic Strings
...
+SVar:DBNoteName:DB$ NameCard | ChooseFromDefinedCards$ Targeted | SubAbility$ DBChangeZone | StackDescription$ None
+SVar:DBChangeZone:DB$ ChangeZone | Defined$ Targeted | Origin$ Graveyard | Destination$ Battlefield | FaceDown$ True | FaceDownPower$ 3 | FaceDownToughness$ 3 | FaceDownSetType$ Creature | AnimateSubAbility$ DBAnimate | SubAbility$ DBCleanup | StackDescription$ None
+SVar:DBAnimate:DB$ Animate | NoteNamedCard$ True | Defined$ Remembered | Triggers$ DealsCDTrig | Replacements$ ReplaceDies | Duration$ Permanent
 SVar:DealsCDTrig:Mode$ DamageDone | ValidSource$ Card.Self | ValidTarget$ Player | CombatDamage$ True | Execute$ TrigPlay | TriggerZones$ Battlefield | TriggerDescription$ Whenever this creature deals combat damage to a player, you may create a copy of the card with the noted name. You may cast the copy without paying its mana cost.
-SVar:TrigPlay:DB$ Play | Defined$ RememberedLKI | ZoneRegardless$ True | CopyCard$ True | ValidSA$ Spell | WithoutManaCost$ True | Optional$ True
+SVar:TrigPlay:DB$ Play | CopyFromChosenName$ True | ValidSA$ Spell | WithoutManaCost$ True | Optional$ True
```
tool4ever, line 10: uuhh, you may want to explain what you determined as the cause of the crash so I can understand :)
this wouldn't really be a correct way for fix, I think I pointed out it's supposed to work like this back when you implemented it:
> 707.14. One card (Magar of the Magic Strings) instructs a player to note the name of a particular card in a graveyard and create a copy of the card with the noted name. To do so, use the characteristics of that card as it last existed in the graveyard to determine the copiable values of the copy.

Northmoc: Ah I forgot about that part of the CR and just re read the card
Sorry it's been so long since I looked at this that I forgot some of the prior conversation 😅 

I think the cause of the crash is in the second collectSpellsForPlayEffect call in getSpellsFromPlayEffect is sending state as null

I'll debug it some more

Northmoc: OK on further digging it seems that the copied card made by PlayEffect isModal (for AbilityUtils.getSpellsFromPlayEffect) but it doesn't have the Modal state

Northmoc: Hmm this also seems wrong... and comes a bit earlier
image

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -3010,6 +3010,7 @@ public static final List<SpellAbility> getSpellsFromPlayEffect(final Card tgtCar
         }
 
         for (SpellAbility s : list) {
+            if (s == null) continue;
```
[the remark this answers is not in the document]
Hanmac: In what case is the first SpellAbility of a Land = null?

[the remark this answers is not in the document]
tool4ever: If LKI also has Modal state it shouldn't be needed? 🤔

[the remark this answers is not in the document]
tool4ever: but I'm still not sure if we can really allow it in all cases:
> 712.8a. While a double-faced card is outside the game or in a zone other than the battlefield or stack, it has only the characteristics of its front face.

Hanmac (approved):
good for now?

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -3004,7 +3004,7 @@ public static final List<SpellAbility> getSpellsFromPlayEffect(final Card tgtCar
                 sp.setCardState(original);
                 list.add(sp);
             }
-            if (tgtCard.isModal()) {
+            if (tgtCard.isModal() && tgtCard.getState(CardStateName.Modal) != null) {
```
tool4ever, line 3007: hasState() seems cleaner?

PR #6631, opened 2024-11-25
Update draftable edition sections

Ayora29:
- FileSection.java : Use a LinkedHashMap to parse edition section, preserving section order and improve performance.
- CardEdition.java : Increment art index when multiple cards with the same name are in the same sheet. Add new section names, remove irrelevant ones.
- Edition txt files of draftable sets : update "sections" and "Booster=" parts to better reflect reallity (cf scryfall and mtg.fandom.com) and reduce amount of unecessary additionnal sections/sheets.
- printsheets.txt : remove unecessary or dead sections
- No card addition, no card remove, no cards in different order compared to before (exept some basic lands)
- I have tested drafts for all editions that I have modified


forge-gui/res/editions/Modern Horizons 3.txt
```diff
@@ -591,7 +588,6 @@ Replace=.042F fromSheet("MH3 commanders")
 A29 R A-Guide of Souls @Ryan Valle
 A38 M A-Ocelot Pride @Chris Seaman
 A122 C A-Galvanic Discharge @Zoltan Boros
-A193 R A-Nadu, Winged Wisdom @Daren Bader
```
tehdiplomat, line 594: Was this removed on purpose?

forge-gui/res/editions/War of the Spark.txt
```diff
@@ -347,3 +353,5 @@ voja_friend_to_elves
 w_0_3_wall_defender
 w_2_2_soldier_vigilance
 w_4_4_angel_flying_vigilance
+
+
```
paulsnoops, line 357: Not a big deal here but 2 empty lines should be 1

Agetian:
Is this branch now ready? Can someone please test and confirm/approve? Leaving it open for a long time may result in a similar effect as before (the need to merge + some difficult to track issues)

paulsnoops:
I didn't think those changes to MH3 edition were causing the crash so I've just tested again and it's still crashing for me:

main > java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
	at java.base/java.util.Objects.checkIndex(Objects.java:365)
	at java.base/java.util.ArrayList.get(ArrayList.java:428)
	at forge.util.FileSection.parseSections(FileSection.java:109)
	at forge.game.GameFormat$Reader.read(GameFormat.java:375)
	at forge.game.GameFormat$Reader.read(GameFormat.java:345)
	at forge.util.storage.StorageReaderRecursiveFolderWithUserFolder.readAll(StorageReaderRecursiveFolderWithUserFolder.java:105)
	at forge.util.storage.StorageBase.<init>(StorageBase.java:44)
	at forge.game.GameFormat$Collection.<init>(GameFormat.java:465)
	at forge.model.FModel.initialize(FModel.java:208)
	at forge.Singletons.initializeOnce(Singletons.java:56)
	at forge.view.Main.main(Main.java:64)

Agetian:
Hmm, probably worth checking out in the debugger which file it's unhappy about in particular... I'm only going to be able to do this on the weekend, so hopefully someone is faster, hehe :)

Ayora29:
@paulsnoops  @Agetian  Ok, you should have a blank file that I don't have and it is causing the Index out of bounds Execption.
May be a custom set file or something like that?
I have added a check for empty text in the parseSections method.
The error should be gone now hopefully.

forge-core/src/main/java/forge/util/FileSection.java
```diff
@@ -105,25 +105,25 @@ public static Map<String, List<String>> parseSections(final List<String> lines)
...
         String section = null;
 
-        do{
-            String line = lines.get(lineNumber++);
-            if (line.startsWith("[") && line.endsWith("]")) {
-                section = line.substring(1, line.length() - 1);
-                if(!result.containsKey(section)) {
-                    result.put(section, new ArrayList<>());
+        if(null != lines && !lines.isEmpty()){
+            do{
```
tehdiplomat, line 109: Why is this a do/while instead of a for loop? 

for(int lineNumber = 0; lineNumber < lines.size(); lineNumber++) {}

Ayora29: it is code optimization. it allows to do the same thing with less line of code and less if statetemens inside the loop.

paulsnoops:
> @paulsnoops @Agetian Ok, you should have a blank file that I don't have and it is causing the Index out of bounds Execption. May be a custom set file or something like that? I have added a check for empty text in the parseSections method. The error should be gone now hopefully.

It loads fine now, thanks for supporting custom editions

forge-core/src/main/java/forge/util/FileSection.java
tehdiplomat: I worry that you embed the increment here. `lineNumber++` which might confuse people of what is actually going on, and if someone decides to add code that interacts with lineNumber after this increment, it won't be the same value as what gets used on the line itself. I wouldn't call that more optimal, as I'm sure the compiler handles the scenarios pretty similarly.

    ```java
    if(null == lines || lines.isEmpty()) {
        return result;
    }
    for(int lineNumber = 0; lineNumber<lines.size(); lineNumber++) {
                String line = lines.get(lineNumber);
                if (line.startsWith("[") && line.endsWith("]")) {
                    section = line.substring(1, line.length() - 1);
                    if(!result.containsKey(section)) {
                        result.put(section, new ArrayList<>());
                    }
                }
                else if(null != section && !line.isEmpty() && !line.startsWith("#")){
                    result.get(section).add(line);
                }
            }
``` 

In fact, we don't even care about the line number, so we could just do...

```java
if (lines == null) {
   return result;
}

for(String line : lines) {
   if (line.startsWith("#") continue;

   if (line.startsWith("[") && line.endsWith("]")) {
          section = line.substring(1, line.length() - 1);
          if(!result.containsKey(section)) {
                 result.put(section, new ArrayList<>());
          }
    } else if(null != section && !line.isEmpty()){
        result.get(section).add(line);
    }
}
```

tehdiplomat: This also seems like we're potentially adding lines to sections that can be just spaces now? since we aren't trimming or string comparison equivilants of that.

Ayora29: Hello @tehdiplomat , Indeed, I agree the for loop is more friendly. The algorithm do not require a do/while at all, didn't realize that. I have updated the code with your sugestion and have added a trim to the line as well.

paulsnoops (approved):

PR #6636, opened 2024-11-26
Chosen letter support (Unsets)

Northmoc:
A decent draft, I think

### DIFF
--- a/forge-game/src/main/java/forge/game/ForgeScript.java
+++ b/forge-game/src/main/java/forge/game/ForgeScript.java
@@ -26,6 +26,7 @@ import org.apache.commons.lang3.StringUtils;
 
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 public class ForgeScript {
@@ -91,6 +92,29 @@ public class ForgeScript {
                 default:
                     return false;
             }
+        } else if (property.startsWith("ChosenLetter")) {
+            final List<String> letters = source.getChosenLetters();
+            if (letters == null) return false; // some things check before letters have been chosen
+            String name = cardState.getName();
+            if (name.startsWith("A-")) name = name.substring(2); // remove Alchemy tag
+            if (property.contains("FirstWord")) name = name.split(" ")[0];
+            boolean found = false;
+            if (property.endsWith("Starts")) {
+                for (String l : letters) {
+                    if (name.startsWith(l)) {
+                        found = true;
+                        break;
+                    }
+                }
+            } else if (property.endsWith("Contains")) {
+                for (String l : letters) {
+                    if (name.contains(l) || name.contains(l.toLowerCase())) {
+                        found = true;
+                        break;
+                    }
+                }
+            }
+            return found;
         } else if (property.equals("Outlaw")) {
             return type.isOutlaw();
         } else if (property.startsWith("non")) {
--- a/forge-game/src/main/java/forge/game/ForgeScript.java
+++ b/forge-game/src/main/java/forge/game/ForgeScript.java
@@ -24,6 +24,7 @@ import forge.game.zone.ZoneType;
 import forge.util.Expressions;
 import org.apache.commons.lang3.StringUtils;
 
+import java.text.Normalizer;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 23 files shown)


Hanmac:
> In case you’re wondering, we’ve chosen to allow Y to be both a consonant and a vowel, depending on what the card cares about. Y? Because we’re nice, and because Y has a great legal team. 

xD

forge-game/src/main/java/forge/game/ability/effects/ChooseLetterEffect.java
```diff
@@ -0,0 +1,27 @@
...
+public class ChooseLetterEffect extends SpellAbilityEffect {
+
+    @Override
+    public void resolve(SpellAbility sa) {
+        final Card card = sa.getHostCard();
+        List<String> letters = Lists.newArrayList("A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N",
+                "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z");
+        List<String> consonants = Lists.newArrayList("B", "C", "D", "F", "G", "H", "J", "K", "L", "M", "N", "P", "Q",
+                "R", "S", "T", "V", "W", "X", "Y", "Z");
+        List<String> choices = sa.hasParam("Consonant") ? consonants : letters;
```
Hanmac, line 20: Imo, I would have used Constants for this? And only make a new List if "Exclude" is used?

Also need extra list for Vowels

Northmoc: Sounds good – I am not sure how to set that up so it would be good to learn

Hanmac: Like this?

```java

List<String> LETTERS = Lists.newArrayList();
//... inside function:
List<String> choices = LETTERS;
if (sa.hasParam("Exclude"))  {
  choices = Lists.newArrayList(choices); // makes copy
  choices.removeAll(Arrays.asList(sa.getParam("Exclude").split(",")));
}


```

hm the `chooseLetter` implementations does remove the letter anyway from the list, so you can't give them the constant ? :(

forge-game/src/main/java/forge/game/ForgeScript.java
```diff
@@ -91,6 +92,29 @@ public static boolean cardStateHasProperty(CardState cardState, String property,
                 default:
                     return false;
             }
+        } else if (property.startsWith("ChosenLetter")) {
+            final List<String> letters = source.getChosenLetters();
+            if (letters == null) return false; // some things check before letters have been chosen
+            String name = cardState.getName();
```
Jetz72, line 98: Maybe apply `.toUpperCase` to `name` here so you don't have to check it twice below? Also in case we ever have any cards that start with a lowercase letter. Can't think of any right now besides "capital offense".

Hanmac:
Apropos Letter in Card Name, we should normalize them for names like `Bartolomé` so `é` counts as `e`

see:
https://github.com/Card-Forge/forge/blob/1bee1a498f1966a0e8df5201b09be08a147ca8f6/forge-core/src/main/java/forge/util/TextUtil.java#L49-L54

forge-game/src/main/java/forge/game/ability/effects/ChooseLetterEffect.java
```diff
@@ -13,15 +13,14 @@ public class ChooseLetterEffect extends SpellAbilityEffect {
     @Override
     public void resolve(SpellAbility sa) {
         final Card card = sa.getHostCard();
-        List<String> letters = Lists.newArrayList("A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N",
+        List<String> letters = Lists.newArrayList("A", "B", "C", "D", "E", "É", "F", "G", "H", "I", "J", "K", "L", "M", "N",
```
Hanmac, line 16: shouldn't that `É` be normalized by Normalizer.Form.NFD ?

Hanmac:
@Northmoc you still need an extra List if you want to support vowels (because Y can be both)
(or is there currently not something that uses them?)

Also `Gray Merchant of Alphabet` might say:
>  The chosen letter needn’t be in the English alphabet. 

But i think for us, it is more simple if we normalize them?

Northmoc:
> @Northmoc you still need an extra List if you want to support vowels (because Y can be both) (or is there currently not something that uses them?)

Correct - nothing uses vowels yet

> Also `Gray Merchant of Alphabet` might say:
> 
> > The chosen letter needn’t be in the English alphabet.
> 
> But i think for us, it is more simple if we normalize them?

I think so too

PR #6643, opened 2024-11-28
Fix cheating class level 3 by copying level 2 activation

### LINKED ISSUE #6634: When activating Carionette (Tempest) it remains in the graveyard
**Describe the bug**
The card Carionette from the Tempest set, reads "2BB: exile Carionette and target creature unless that creature's controller pays 2. Activate only if Carionette is in your graveyard".

When I activate Carionette in the graveyard it and AI opponents does not pay the 2 mana cost, it exiles the opponent card but remains in my graveyard and can be used again.

**To Reproduce**
1. Start Quest mode (haven't tried in other modes)
2. Have Carionette in the graveyard
3. Click on Carionette, to activate ability. 
4. Make sure opponents does not pay the 2 mana cost
5. Carionette remains in the graveyard

**Expected behavior**
Carionette should be exiled.

 - OS: Windows
 - Version Forge 1.6.65 Quest mode

tool4ever:
Probably doesn't make sense if the trigger would still fire either.

Closes #6634

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/ClassLevelUpEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/ClassLevelUpEffect.java
@@ -19,7 +19,7 @@ public class ClassLevelUpEffect extends SpellAbilityEffect {
     public void resolve(SpellAbility sa) {
         final Card host = sa.getHostCard();
         final Game game = host.getGame();
-        final int level = host.getClassLevel();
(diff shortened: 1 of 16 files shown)


Hanmac (dismissed):
i would try to move the ClassLevelUp check to conditions, but i think it is good enough for now

Hanmac (approved):

PR #6652, opened 2024-12-02
SpellAbilityAi: move willPayUnlessCost to EffectSpecific Code

### LINKED ISSUE #6421: Volatile Stormdrake bug - opponent doesn't pay energy counters but doesn't sacrifice creature.
Volatile Stormdrake enters, targets 5 MV creature (Marneus Calgar), player gets 4 energy counters, player doesn't pay any, creature isn't sacrificed. No object on board visible that would prevent the sacrifice effect.

Log
image

Boardstate
image

Forge updated latest version.

Hanmac:
This is for moving "willPayUnlessCost" to Effect Specific AI code

with that we will clean up when AI should not Pay for Unless Cost

It is still work in Progress

Closes #6421


Hanmac:
Probably next one to refactor is LifeLE cost ones.

While not UnlessCost, AI need help with Effects like `Braids, Arisen Nightmare` to not kill themselves

Hanmac:
@tool4ever can you take a peak on my refactoring to see if I'm on the right track?

Especially the LoseLife/DealDamage one
currently, AI is just taking the lifeloss without reaction

Hanmac:
or @Agetian , what do you think about my WIP?

tool4ever:
Hmn, while I think this was rather low priority logic seems ok?

Hanmac:
> Hmn, while I think this was rather low priority logic seems ok?

That is one of the parts I'm tying to port into Ai Effects,
another part would the logic for Additional Costs, so AI has more control over when to Pay Costs like Kicker, and "Gift a Fish"

Agetian:
Yeah, definitely a good effort, and the implementation looks good to me so far! :+1:

Hanmac:
@Agetian especially these two:
* may draw. If you do, discard
* may discard. If you do, draw

might need more love

Hanmac:
@tool4ever can you use this MR to add the PayEnergy Logic Problem we had this week?

i try to move the last piece of "WillAttack" logic, but i didn't had time yet

tool4ever:
I don't really know how to fix that currently either :/

forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
```diff
@@ -1222,19 +1222,19 @@ public String chooseProtectionType(String string, SpellAbility sa, List<String>
...
+            emptyAbility.setTriggeringObjects(sa.getTriggeringObjects());
+            emptyAbility.setReplacingObjects(sa.getReplacingObjects());
+            emptyAbility.setTrigger(sa.getTrigger());
+            emptyAbility.setReplacementEffect(sa.getReplacementEffect());
+            emptyAbility.setSVars(sa.getSVars());
+            emptyAbility.setCardState(sa.getCardState());
+            emptyAbility.setXManaCostPaid(sa.getRootAbility().getXManaCostPaid());
+            emptyAbility.setTargets(sa.getTargets().clone());
+
             boolean result = ComputerUtil.playNoStack(player, emptyAbility, getGame(), true); // AI needs something to resolve to pay that cost
```
Hanmac, line 1238: @tool4ever can we find a way for AI to pay the Cost without the Need of an extra empty AbilityStatic ?

Using EmptySa also wouldn't fix the problem.

Hanmac:
@tool4ever @Agetian i changed how AI pays for UnlessCost, no need for extra emptyAbility.

probably need some intense testing?

forge-gui/src/main/java/forge/player/HumanPlaySpellAbility.java
```diff
@@ -109,7 +109,9 @@ public final boolean playAbility(final boolean mayChooseTargets, final boolean i
             ability.setPaidLife(0);
         }
 
-        ability = GameActionUtil.addExtraKeywordCost(ability);
+        if (ability.isSpell() && !c.isCopiedSpell()) {
```
tool4ever, line 112: this is redundant as addExtraKeywordCost already checks it at the start

Hanmac: Yeah, just to be sure in case I might touch that in the future

tool4ever:
We could change Echo + Cumulative upkeep in SacrificeEffect to use `payCostToPreventEffect`?
technically they are UnlessCost so later logic when not to pay can be added 🤔

Hanmac:
> We could change Echo + Cumulative upkeep in SacrificeEffect to use `payCostToPreventEffect`? technically they are UnlessCost so later logic when not to pay can be added 🤔

that is a bit more tricky, because Echo has an extra trigger, but you could try it.
And like with: 
> You may pay {0} rather than pay the echo cost for permanents you control.

And for Cumulative upkeep, there is also this: "Spend this mana only to pay cumulative upkeep costs."

forge-ai/src/main/java/forge/ai/ability/DrawAi.java
```diff
@@ -547,4 +549,53 @@ public boolean confirmAction(Player player, SpellAbility sa, PlayerActionConfirm
...
+            return false;
+        }
+
+        // Risk Factor Effects
+        for (Player p : AbilityUtils.getDefinedPlayers(host, sa.getParam("Defined"), sa)) {
+            if (p.isOpponentOf(payer)) {
+                if (!p.canDraw()) {
+                    return false;
+                }
+                if (cost.hasSpecificCostType(CostDamage.class)) {
```
tool4ever, line 575: all these extra parts checks don't make sense to me

they should already be covered by the `super` anyway

Hanmac: The generic super doesn't check for Damage in comparison to the amount of cards being drawn

forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
```diff
@@ -1220,26 +1221,13 @@ public String chooseProtectionType(String string, SpellAbility sa, List<String>
...
-        emptyAbility.setXManaCostPaid(sa.getRootAbility().getXManaCostPaid());
-        emptyAbility.setTargets(sa.getTargets().clone());
-
-        if (ComputerUtilCost.willPayUnlessCost(sa, player, cost, alreadyPaid, allPayers)) {
-            boolean result = ComputerUtil.playNoStack(player, emptyAbility, getGame(), true); // AI needs something to resolve to pay that cost
-            if (!emptyAbility.getPaidHash().isEmpty()) {
-                // report info to original sa (Argentum Masticore)
-                sa.setPaidHash(emptyAbility.getPaidHash());
+        if (SpellApiToAi.Converter.get(sa.getApi()).willPayUnlessCost(sa, player, cost, alreadyPaid, allPayers)) {
+            if (!ComputerUtilCost.canPayCost(cost, sa, player, true)) {
```
tool4ever, line 1225: you missed this place:
image

AI now cheats on the mana

Hanmac: like this?

Hanmac:
@tool4ever can we look about this?

```
        if (cost == null) {
            // Is this fork even used for anything anymore?
            if (ComputerUtilMana.payManaCost(ai, sa, false)) {
                game.getStack().addAndUnfreeze(sa);
                return true;
            }
        } else {
```

that doesn't feel right imo, can this be removed?

forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
tool4ever: extra object shouldn't be needed but seems to work for now

tool4ever:
I can't think of any case for it.
The Daxos TODO also seems outdated.

tool4ever (approved):
good enough for now?

PR #6657, opened 2024-12-03
fix FCollection

kevlahnota:
should fix slowness on access by using concurrent collection and avoid concurrentmodificatonexception without using locks


Hanmac:
im still thinking about 
[SetUniqueList](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/list/SetUniqueList.html)

but in my last attempt it wasn't sortable, I probably need to try again

Jetz72:
Still wondering if cloning the game state via `GameCopier` and running AI logic on separate threads with their own copies might be a safer - or at least more maintainable - approach to AI multi-threading than exposing a single game state to multiple threads at once. 

Does this bring us back to losing `iterable.remove()`?

Hanmac:
> Does this bring us back to losing `iterable.remove()`?

imo removing items at the same time as iterating them is too problematic anyway

Jetz72:
> imo removing items at the same time as iterating them is too problematic anyway

It's a useful tool in the right context, and in the places where it is appropriate, the alternatives are rarely elegant. Collecting a list of items to remove and then removing them all in a separate operation is a bit of a messy workaround when more typical collections can support doing it in a single step.

But the broader concern I have is the increased complexity of working with `FCollection`s the more they differ from regular `Collection`s. Having to remember to use certain `CollectionUtil` methods instead of the ones in Java's `Collections` library is another example. It's a very easy mistake to make because methods like `Collections.shuffle` will still accept an `FCollection` without warning or error.

tool4ever:
I'm seeing somewhat better performance with this approach:
only takes around 90 seconds now (so 50 faster / "only" 40 slower than before the previous PR 😵)

kevlahnota:
> > imo removing items at the same time as iterating them is too problematic anyway
> 
> It's a useful tool in the right context, and in the places where it is appropriate, the alternatives are rarely elegant. Collecting a list of items to remove and then removing them all in a separate operation is a bit of a messy workaround when more typical collections can support doing it in a single step.
> 
> But the broader concern I have is the increased complexity of working with `FCollection`s the more they differ from regular `Collection`s. Having to remember to use certain `CollectionUtil` methods instead of the ones in Java's `Collections` library is another example. It's a very easy mistake to make because methods like `Collections.shuffle` will still accept an `FCollection` without any warning or error before runtime.

Hmm I think the only way to use Collections.shuffle like before is to use a different concurrent collection for FCollection but I don't have idea to recreate a custom collection that's faster without using locks or synchronizedbor it would end up throwing again concurrentmodification.
On Sentry, some methods are failing with concurrentmodification like the staticeffects since  hostedmatch is run on a thread and it's refactored wayback 2015 to run multiple matches. It could use locks or concurrent collection. Though concurrent is not way faster than the regular collection that's not thread safe.

kevlahnota:
@Hanmac  @tehdiplomat I think this should be refactored later on if needed be. I think those iterator.remove() can be refactored to get similar results if it's really needed. This fix should be enough to fasten up the collection.

Hanmac:
i really don't like that CopyOnWriteArrayList makes everything more complicated :(

kevlahnota:
> i really don't like that CopyOnWriteArrayList makes everything more complicated :(

me too but I have no choice since there's no other available concurrent list on Java, and FCollection implements a List, I can't find a faster way without breaking concurrent modification

Hanmac:
> > i really don't like that CopyOnWriteArrayList makes everything more complicated :(
> 
> me too but I have no choice since there's no other available concurrent list on Java, and FCollection implements a List, I can't find a faster way without breaking concurrent modification

i think instead we could go a middle way? Have the concurrent modification exceptions better show where the problem was happening?

Right now, concurrent modification happens when you do this:

```
for (O : list) {
  list.change();
}
```

the problem with the Exceptions is more that they trigger in the iterator, but it would be more useful if they would trigger where the change happend.

kevlahnota:
> > > i really don't like that CopyOnWriteArrayList makes everything more complicated :(
> > 
> > 
> > me too but I have no choice since there's no other available concurrent list on Java, and FCollection implements a List, I can't find a faster way without breaking concurrent modification
> 
> i think instead we could go a middle way? Have the concurrent modification exceptions better show where the problem was happening?
> 
> Right now, concurrent modification happens when you do this:
> 
> ```
> for (O : list) {
>   list.change();
> }
> ```
> 
> the problem with the Exceptions is more that they trigger in the iterator, but it would be more useful if they would trigger where the change happend.

hmmm the problem also is some threads access the collection while the other thread is still modifying it will produce concurrentmodification hence I put a lock so no other threads can access it unless the thread that modifies the collection is finished. however it will slow down as soon as the collection gets bigger (like the all swamp land deck test).

kevlahnota:
@Hanmac just changed to use Collections.synchronized so it has protection on add and remove though on iteration, it will throw ConcurrentModification if it's not allowed to do so.

Hanmac (approved):
Collections.synchronized might be better for now, I think?

If the problems happen, we can fix them later?

Hanmac:
As for ConcurrentModification, we still need to investigate when they happen

Hanmac (approved):

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -784,9 +785,10 @@ private Pair<Integer, Integer> getDamageFromBlockingTramplers(final List<Card> b
             for (Card blocker : remainingBlockers) {
                 if (CombatUtil.canBlock(attacker, blocker) && damage > 0) {
                     damage -= ComputerUtilCombat.shieldDamage(attacker, blocker);
-                    remainingBlockers.remove(blocker);
+                    removedBlockers.add(blocker);
```
kevlahnota, line 788: @Hanmac does this also work? just found this before merging the fix.

Hanmac: Yeah, that should fix it, i think

kevlahnota:
gonna merge this and lets see if there still some cases needed.

PR #6664, opened 2024-12-05
Fix ConcurrentModificationException

tool4ever:
Imo this method is so rarely called that removing via Iterator seems pointless (vs. readability).

### DIFF
--- a/forge-ai/src/main/java/forge/ai/GameState.java
+++ b/forge-ai/src/main/java/forge/ai/GameState.java
@@ -1133,7 +1133,7 @@ public abstract class GameState {
             p.getZone(zt).removeAllCards(true);
         }
 
-        p.setCommanders(Lists.newArrayList());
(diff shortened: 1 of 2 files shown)


Hanmac (approved):
i wish we could use functions like [retainAll](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ArrayList.html#retainAll(java.util.Collection)) to remove old Commanders

Hanmac (changes requested):
@tool4ever setCommanders is only ever called with empty ArrayList ... so how about removing the function and use a new `clearCommanders` instead?

Hanmac (approved):
I would have defined a `clearCommanders`, but that is fine too

Jetz72:
That was mostly written as one of those "in case anybody needs it later" methods. Seems like a reasonable cut.

PR #6665, opened 2024-12-05
Fix for noted types for Volo's journal

### LINKED ISSUE #4693: AI makes illegal choices when triggering Volo's Journal
Windows, current snapshot.

The AI had `Volo, Itinerant Scholar` [CLB], and his journal. It cast `Shimmering Efreet` [VIS] which has the creature type Efreet, triggering journal, and chose Sliver. There's no effect making that a valid choice. I tested it myself and was correctly limited to choosing types that the spell actually had, so I guess it's just that the AI choice isn't made from the restricted set.

neoFuzz:
Aims to stop "Sliver" being the default option.

Closes #4693

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtil.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtil.java
@@ -2567,9 +2567,6 @@ public class ComputerUtil {
                 }
             }
 
-            if (sa.hasParam("Note") && sa.getHostCard().hasAnyNotedType())
-                validTypes.removeAll((Collection<String>) sa.getHostCard().getNotedTypes());
-
             if (!CardType.isACreatureType(chosen) || invalidTypes.contains(chosen)) {
                 chosen = validTypes.size() == 1 ? (String) validTypes.toArray()[0] :
                         ComputerUtilCard.getMostProminentType(ai.getAllCards(), validTypes, false);
--- a/forge-ai/src/main/java/forge/ai/ComputerUtil.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtil.java
@@ -2570,7 +2570,8 @@ public class ComputerUtil {
             if (sa.getHostCard().hasAnyNotedType())
                 validTypes.removeAll((Collection<String>) sa.getHostCard().getNotedTypes());
             if (!CardType.isACreatureType(chosen) || invalidTypes.contains(chosen)) {
-                chosen = ComputerUtilCard.getMostProminentType(ai.getAllCards(), validTypes, false);
+                chosen = validTypes.size() == 1 ? (String) validTypes.toArray()[0] :
+                        ComputerUtilCard.getMostProminentType(ai.getAllCards(), validTypes, false);
                 //chosen = "Sliver";
             }
 
--- a/forge-ai/src/main/java/forge/ai/ComputerUtil.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtil.java
@@ -2567,8 +2567,9 @@ public class ComputerUtil {
                 }
             }
 
-            if (sa.getHostCard().hasAnyNotedType())
+            if (sa.hasParam("Note") && sa.getHostCard().hasAnyNotedType())
                 validTypes.removeAll((Collection<String>) sa.getHostCard().getNotedTypes());
+
             if (!CardType.isACreatureType(chosen) || invalidTypes.contains(chosen)) {
                 chosen = validTypes.size() == 1 ? (String) validTypes.toArray()[0] :
                         ComputerUtilCard.getMostProminentType(ai.getAllCards(), validTypes, false);
--- a/forge-ai/src/main/java/forge/ai/ComputerUtil.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtil.java
@@ -2544,39 +2544,32 @@ public class ComputerUtil {
 
                 if (logic.equals("MostProminentOnBattlefield")) {
                     chosen = ComputerUtilCard.getMostProminentType(game.getCardsIn(ZoneType.Battlefield), valid);
-                }
-                else if (logic.equals("MostProminentComputerControls")) {
+                } else if (logic.equals("MostProminentComputerControls")) {
                     chosen = ComputerUtilCard.getMostProminentType(ai.getCardsIn(ZoneType.Battlefield), valid);
-                }
-                else if (logic.equals("MostProminentComputerControlsOrOwns")) {
+                } else if (logic.equals("MostProminentComputerControlsOrOwns")) {
                     CardCollectionView list = ai.getCardsIn(Arrays.asList(ZoneType.Battlefield, ZoneType.Hand));
                     if (list.isEmpty()) {
                         list = ai.getCardsIn(Arrays.asList(ZoneType.Library));
                     }
                     chosen = ComputerUtilCard.getMostProminentType(list, valid);
-                }
-                else if (logic.equals("MostProminentOppControls")) {
+                } else if (logic.equals("MostProminentOppControls")) {
                     CardCollection list = ai.getOpponents().getCardsIn(ZoneType.Battlefield);
                     chosen = ComputerUtilCard.getMostProminentType(list, valid);
                     if (!CardType.isACreatureType(chosen) || invalidTypes.contains(chosen)) {
                         list = CardLists.filterControlledBy(game.getCardsInGame(), ai.getOpponents());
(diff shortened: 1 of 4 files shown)


forge-game/src/main/java/forge/game/ability/effects/ChooseTypeEffect.java
```diff
@@ -154,4 +154,12 @@ public void resolve(SpellAbility sa) {
             throw new InvalidParameterException(sa.getHostCard() + "'s ability resulted in no types to choose from");
         }
     }
+
+    private static List<String> getInvalidTypes(SpellAbility sa) {
+        if(sa.getHostCard().hasAnyNotedType())
```
Hanmac, line 159: this check is already done in in line 119

it is also important to check for hasParam "Note"

forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
```diff
@@ -875,12 +875,29 @@ public static String getMostProminentType(final CardCollectionView list, final C
...
-        for (final Entry<String, Integer> entry : typesInDeck.entrySet()) {
-            final String type = entry.getKey();
+        // Check if the valid list is not empty
+        if (!valid.isEmpty()) {
+            // Iterate through typesInDeck and consider only valid types
+            for (final Entry<String, Integer> entry : typesInDeck.entrySet()) {
+                final String type = entry.getKey();
+
+                // Only consider the types that are in the valid list
+                if (valid.contains(type)) {
```
tool4ever, line 885: you can save a lot of lines if you just check `valid.isEmpty() ||` here first instead

forge-game/src/main/java/forge/game/ability/effects/ChooseTypeEffect.java
neoFuzz: I was getting `empty`, which didn't sit well with me, but It was meant to be the first time selection.

Pushing an update now

forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
neoFuzz: I'll work it out 😅

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -2572,7 +2572,13 @@ else if (logic.equals("MostProminentInComputerGraveyard")) {
                 }
             }
             if (!CardType.isACreatureType(chosen) || invalidTypes.contains(chosen)) {
-                chosen = "Sliver";
+                // assume invalid types is all non-valid types
+                List <String> invalid = Lists.newArrayList(CardType.getAllCreatureTypes());
```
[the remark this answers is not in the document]
neoFuzz: I don't understand the question 100%?
tag it in the SpellAbility AILogic?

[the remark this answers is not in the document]
tool4ever: well moving away from the hardcoded default is probably a good idea
but if you just move the lines after `logic != null` branch out of it you can reuse them here so only L2580 should be needed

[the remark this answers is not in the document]
neoFuzz: So the whole `if` block?
I used `getMostProminentType` as a decision making tool, since you might as well note the types with the most first.

Making a script tag wouldn't be bad since only 2 cards currently use this singleton noted creature types abilitiy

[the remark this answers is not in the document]
neoFuzz: I completely reduced the code there. `invalid` wasn't used. Also added check for `validType.size() == 1` skipping get most prominent type

forge-game/src/main/java/forge/game/ability/effects/ChooseTypeEffect.java
```diff
@@ -154,4 +154,10 @@ public void resolve(SpellAbility sa) {
             throw new InvalidParameterException(sa.getHostCard() + "'s ability resulted in no types to choose from");
         }
     }
+
+    private static List<String> getInvalidTypes(SpellAbility sa) {
+        return sa.hasParam("InvalidTypes") ?
+                Arrays.asList(sa.getParam("InvalidTypes").split(",")) :
+                new ArrayList<>();
+    }
```
kevlahnota, line 162: Would it be better to use Collections.emptyList() if it isn't mutable?

neoFuzz: This snippet was already like that, I just moved it to a function. If you think it's worth the change, I'll investigate

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -2544,35 +2544,35 @@ public static String chooseSomeType(Player ai, String kindOfType, SpellAbility s
...
                     boolean includeTokens = !logic.endsWith("NonToken");
                     chosen = ComputerUtilCard.getMostProminentType(ai.getAllCards(), valid, includeTokens);
-                }
-                else if (logic.equals("MostProminentInComputerGraveyard")) {
+                } else if (logic.equals("MostProminentInComputerGraveyard")) {
                     chosen = ComputerUtilCard.getMostProminentType(ai.getCardsIn(ZoneType.Graveyard), valid);
                 }
             }
+
+            if (sa.getHostCard().hasAnyNotedType())
```
Hanmac, line 2570: add `sa.hasParam("Note")` inside the If before the other check, so it only does it when the ability uses `Note`

neoFuzz: `if (sa.hasParam("Note") && sa.getHostCard().hasAnyNotedType())` or individual `if` statements?

neoFuzz:
I have written a test to make sure it the choices are unique. Where would I add this test too?

forge-ai/src/main/java/forge/ai/ComputerUtil.java
Hanmac: yeah like this

forge-gui-desktop/src/test/java/forge/ai/simulation/GameSimulationTest.java
```diff
@@ -2548,4 +2551,70 @@ public void testBasicSpellFizzling() {
...
+        for(Card card : cards) {
+            if (card.getName().equals("Volo's Journal")) {
+                // All words are present in the iterable
+                AssertJUnit.assertTrue(areWordsInIterable(words,card.getNotedTypes()));
+            }
+        }
+
+    }
+
+    protected boolean areWordsInIterable(List<String> words, Iterable<String> iterable) {
```
Hanmac, line 2605: Why not use `list.containsAll` ?

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -2544,35 +2544,36 @@ public static String chooseSomeType(Player ai, String kindOfType, SpellAbility s
...
                     boolean includeTokens = !logic.endsWith("NonToken");
                     chosen = ComputerUtilCard.getMostProminentType(ai.getAllCards(), valid, includeTokens);
-                }
-                else if (logic.equals("MostProminentInComputerGraveyard")) {
+                } else if (logic.equals("MostProminentInComputerGraveyard")) {
                     chosen = ComputerUtilCard.getMostProminentType(ai.getCardsIn(ZoneType.Graveyard), valid);
                 }
             }
+
+            if (sa.hasParam("Note") && sa.getHostCard().hasAnyNotedType())
```
tool4ever, line 2570: this part shouldn't be needed:
the normal assumption is the api effect of the engine will already provide the legal options with its logic, so AI code can focus on logic around good choice etc.

neoFuzz: I've checked it. The types are the remaining. eg: human wizard and human already noted leaves wizard left.
It'll be removed.

forge-gui-desktop/src/test/java/forge/ai/simulation/GameSimulationTest.java
neoFuzz: Expanded to check for unique entries.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
neoFuzz: removed

tool4ever (approved):
seems good, thanks :)

PR #6668, opened 2024-12-06
Improve AI for First Responder, improve evaluation for self-sac abilities e.g. Sakura-Tribe Elder.

### LINKED ISSUE #4025: AI would rather kill a Sakura-Tribe Elder than a more valid target
**Describe the bug**
AI would rather kill Sakura-Tribe Elder than Noble Hierarch, for example, which makes no sense.

**To Reproduce**
1. Player casts Sakura-Tribe Elder and Noble Hierarch.
2. AI casts Lightning Bolt and targets Sakura-Tribe Elder.

**Expected behavior**
Sakura-Tribe Elder and similar should be at the bottom of AI's priority for destruction or exile.

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-10.28

**Screenshots**

sakura

### LINKED ISSUE #6667: AI prioritizes returning lowest mana value creature with First Responder over all other options
First Responder is a card that may return another creature to hand at end step and puts a number of +1/+1 counters on First Responder equal to that creature's power.  The AI will always choose the lowest MV creature, regardless of that creature's power, ETB effects, status as a token or whether the creature is about to be sacrificed.  The AI will choose 0 mana cards like Ornithopter over tokens, but will choose tokens over anything else, even if they are 0 power Eldrazi Spawn.  The AI will always return a creature if able, even if it is strictly detrimental (such as returning an Eldrazi Spawn)

**To Reproduce**
1. Create a board state where the AI has First Responder, an Ornithopter, and any other creature that would logically be a better choice.
2. Go to End Step.
3. The AI will return the Ornithopter.

**Expected behavior**
First Responder controlled by a human has a variety of play patterns, whether you want to maximize ETB effects or push damage quickly. The safest behavior for the AI is to treat it as a 'rescue bounce' effect, and the AI should rescue cards based on the value of doing so. A simple list of priorities would be:
1. Creatures that are sacrificed at EoT such as blitzed cards or Groundbreaker.
2. Creatures that are enchanted by an opponent's Aura or affected by opponent's abilities that make them unable untap (Stun Counters eg.).
3. Dashed Creatures or creatures that will return to hand anyway at end of turn.
4. Creatures with positive ETB Abilities.
5. Non-token Creatures by highest power.
6. Return nothing.

For safety, the AI should probably never return token creatures or creatures owned by another player if given a choice.  There are certainly instances where it is worth doing, but there are many, many more where it is incorrect to do so.

**Screenshots**

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 2.0.01-SNAPSHOT-12.03

Agetian:
Slightly better decision making for First Responder. Fixes #6667.
Fixes #4025 .

### DIFF
--- a/forge-ai/src/main/java/forge/ai/CreatureEvaluator.java
+++ b/forge-ai/src/main/java/forge/ai/CreatureEvaluator.java
@@ -280,8 +280,9 @@ public class CreatureEvaluator implements Function<Card, Integer> {
                     }
                 }
             }
+        } else if (ComputerUtilCost.isSacrificeSelfCost(sa.getPayCosts())) {
+            return -20; // can be sacrificed in response to ability or spell, thus, less prioritable
         }
-
         // default value
         return 10;
     }
--- a/forge-ai/src/main/java/forge/ai/CreatureEvaluator.java
+++ b/forge-ai/src/main/java/forge/ai/CreatureEvaluator.java
@@ -281,7 +281,7 @@ public class CreatureEvaluator implements Function<Card, Integer> {
                 }
             }
         } else if (ComputerUtilCost.isSacrificeSelfCost(sa.getPayCosts())) {
-            return -20; // can be sacrificed in response to ability or spell, thus, less prioritable
+            return -10; // can be sacrificed in response to ability or spell, thus, less prioritable
         }
         // default value
         return 10;
--- a/forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
@@ -1570,6 +1570,8 @@ public class ChangeZoneAi extends SpellAbilityAi {
                 }
             } else if (logic.startsWith("ExilePreference")) {
                 return doExilePreferenceLogic(decider, sa, fetchList);
+            } else if (logic.equals("BounceOwnTrigger")) {
+                return doBounceOwnTriggerLogic(decider, fetchList);
             }
         }
         if (fetchList.isEmpty()) {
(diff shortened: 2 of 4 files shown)


tool4ever:
I suspect the user wanting his favorite interactions fixed will just report the AI will now prefer bolting _Birds of Paradise_ over something like _Spore Frog_ at EOT next.
Does your change still make sense from a heuristics POV? ugh probably...

Agetian:
True, though I feel like when the AI prioritizes between two similar-ish cards, one of which can be sacrificed in response to an ability or a spell, and the other one can't, a slight de-prioritization of the card that can be sac'ed might be warranted. I'm not sure if any side effects might come out of this though :/  -20 is basically enough to cover a +1 power or toughness difference, or a couple keyworded or non-keyworded abilities, but not much else. Think it's too much of a difference and we should make it -10 instead? Or is this minimal enough to not worry about it at the moment and tweak if anything specific pops up?

Agetian:
I tested a bit more and decided to go down to -10, which still works for similarly tiered creatures, but doesn't affect greater differences between the two, which may be trickier to differentiate. Overall, the question here is:
1. Do we need this implemented at all, as in, is it relevant enough to be noticeable? Probably yes, if you consider simple cases where the two cards are nearly identical when it comes to the 
2. Is the criterion of having a self-sacrifice cost an adequate one? I'm still sorta split on this one, since many different possibilities exist where you'd still want to throw e.g. a removal spell on a card that can be removed in response, just to get this specific effect (e.g. get the opponent to take it off the board via a sacrifice). Which is why I'm going down to the minimal possible difference (currently of -10) where this would not affect any bigger difference than maybe a +1/+0 or a single ability when it comes to evaluating two cards side by side.
I don't think this solution will please any specific person or should even aim to do so :) just considering something that seems like a good idea, at least for now :)
If we decide that it's not a good idea to de-prioritize like that, I'll prolly just remove this portion (revert it).

tool4ever (approved):
I think it's ok for now

Imo a more general overhaul of CreatureEvaluator could be something worth doing in the future but that should probably involve somewhat more analysis of all its usages and discussion (always happy to talk with you :D)

E. g. it might need at least a flag for what cause it's used (targeted removal vs. casting?)

Most importantly it needs to be fast so ideally should only check for direct card attributes
(which is one of the things that worries me about static "Cant Block" stuff)

Hanmac:
@tool4ever yeah we might look into better Evaluation for different causes like "for discard" or something

Agetian:
Indeed, that would be the optimal approach, though it's a bit of a bigger overhaul that needs some planning and quite a bit of time investment, so prolly a project for the next time I get an opportunity to do a deep dive into coding :)

PR #6678, opened 2024-12-11
Mark gainlands as L rarity in Foundations file

tehdiplomat:


Hanmac:
yeah Scryfall still doesn't has the L Rarity

Hanmac (approved):

PR #6690, opened 2024-12-15
fix: Add Set Codes to Commander Pre Cons

remggo:
This commit adds the set cons to the Commander Pre Cons so that they are correctly sorted when grouping for Set.


Agetian (approved):
Looks good to me!

Hanmac (approved):

PR #6700, opened 2024-12-17
Full Control unleashed

### LINKED ISSUE #112: Abilities with some unpayable costs must still be allowed to be put on the stack
> Activating mana abilities is done before paying costs when activating an ability. This means that if you control a creature which can be tapped for mana, such as Avacyn's Pilgrim, you may both tap it for mana and untap it to pay costs for Halo Fountain's activated abilities.

Currently Forge will not let you activate the ability without a tapped creature though, so you're forced to tap for mana first.

This matters because there might be triggers related to it that will then be on the stack before your ability, e.g.:

grafik

But you should be able to reach the opposite order as well if you so desire.

Another related example is _Bond of Agony_:
if you're at 1 life Forge will not let you choose X > 1. But maybe you want to cast it for X=2 because you can pay that part with _The Great Henge_.

tool4ever:
Closes #112

Also fixed wrong value for _Momentous Fall_ when spending green mana before sacrificing _Omnath, Locus of Mana_ to pay.


forge-gui/src/main/java/forge/gamemodes/match/input/InputPassPriority.java
```diff
@@ -124,8 +124,8 @@ private void passPriority(final Runnable runnable) {
     @Override
     protected final void onPlayerSelected(Player selected, final ITriggerEvent triggerEvent) {
         PlayerController pc = selected.getController();
-        if (pc.isGuiPlayer()) {
-           pc.setFullControl(!pc.isFullControl());
+        if (!pc.isAI()) {
```
Hanmac, line 127: Isn't that check broken for network player?

forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
```diff
@@ -1672,4 +1672,9 @@ public CardCollection chooseCardsForEffectMultiple(Map<String, CardCollection> v
 
         return choices;
     }
+
+    @Override
+    public List<CostPart> orderCosts(List<CostPart> costs) {
+        return null;
```
Hanmac, line 1678: Wouldn't it be more easy, to have them return costs List instead of null?

forge-game/src/main/java/forge/game/cost/CostPayment.java
```diff
@@ -136,7 +131,12 @@ public final void refundPayment() {
 
     public boolean payCost(final CostDecisionMakerBase decisionMaker) {
         adjustedCost = CostAdjustment.adjust(cost, ability);
-        final List<CostPart> costParts = adjustedCost.getCostPartsWithZeroMana();
+        List<CostPart> costParts = adjustedCost.getCostPartsWithZeroMana();
+
+        if (adjustedCost.getCostParts().size() > 1 && decisionMaker.getPlayer().getController().isFullControl(FullControlFlag.ChooseCostOrder)) {
+            // if mana part is shown here it wouldn't include reductions, but that's just a minor inconvenience
+            costParts = decisionMaker.getPlayer().getController().orderCosts(costParts);
+        }
```
Hanmac, line 139: I would make the `.isFullControl(FullControlFlag.ChooseCostOrder)` check inside the `orderCosts` function for this controller

Hanmac:
@tool4ever you are doing fixes in the wrong branches again 

doing the Token or Rani fixes in this PR doesn't make this more likely to be merged

forge-gui/src/main/java/forge/gamemodes/match/input/InputPassPriority.java
Jetz72: What's the reason for hooking the opening and closing of this menu into the input system? Seems like something that should be entirely on the UI side.

tool4ever: Hmn but this is still inside forge-gui?
Anywhere higher up the chain seems problematic as we wouldn't know we're in safe state for this (no other input active)

Jetz72: What I was getting at is that opening a menu seems like something lower down the chain - handled entirely within the desktop and mobile UI implementations. Right now it appears to be hooked into the input function for choosing a player, which is responding with an instruction to open a menu if a chosen player isn't currently needed.

I guess this works as a way to open the menu when clicking on a player portrait when it wouldn't otherwise do anything. But I would think the input state logic shouldn't be concerned with how the UI makes the Full Control settings available to the user - whether it's a menu opened by clicking a portrait, right clicking in the details panel, a dropdown at the top of the screen, etc. 

You might consider either detecting the input state entirely within the front end (the mana pool icons in [CField.java](https://github.com/Card-Forge/forge/blob/c2015bf0201b3b4f793def0f894cd3c3b03f0f31/forge-gui-desktop/src/main/java/forge/screens/match/controllers/CField.java#L76-L78) do this), or opening the menu through other gestures (right click, long press, swipe, etc.) so as not to share with the "choose player" input.

tool4ever: Well I didn't really consider changing the options in the middle of targeting worthwhile.
I could switch to right click and long press instead, though those ``instanceof`` checks in CField aren't looking clean either...
Guess I'll sleep on it :)

forge-gui-desktop/src/test/java/forge/gamesimulationtests/util/PlayerControllerForTests.java
```diff
@@ -786,4 +786,10 @@ public List<SpellAbility> chooseSpellAbilitiesForEffect(List<SpellAbility> spell
 		// TODO Auto-generated method stub
 		return null;
 	}
+
+    @Override
+    public List<CostPart> orderCosts(List<CostPart> costs) {
+        // TODO Auto-generated method stub
+        return null;
```
Hanmac, line 793: Probably need to return costs too,
or otherwise tests might crash?

forge-gui/src/main/java/forge/gamemodes/match/input/InputPassPriority.java
tool4ever: should be fine now

PR #6717, opened 2024-12-23
Contraptions

Jetz72:
Adds full support for Contraption cards and implementations for every card that uses them (minus a couple that use other un-mechanics).

Features:
- Contraption extra deck, similar to the Attraction deck.
- "Whenever you crank..." trigger, used by all contraptions.
- "Assemble a Contraption" effect. Puts a contraption into the battlefield and prompts you to assign it to a sprocket. Pulls double duty for "Reassemble" effects, where you move a contraption onto another sprocket.
- Crank counter, displayed as a command zone effect, advanced by a turn based action or by card effects. When it advances to the next sprocket, you may crank any contraptions attached to that sprocket.
- Contraptions sorted into piles based on which sprocket they're attached to.
- State based action that destroys unassembled contraptions and reassembles ones that changed controller.
- Full support for Steamflogger Boss's heretofore stubbed out replacement effect.
- 45 contraptions and 21 contraption support cards. Merry Christmas.

Also includes:
- New icons for deck sections and extra zones. Currently only available as HD icons; can add some low-res fallbacks if needed. I also put these on a separate sprite sheet since they were getting kinda scattered around the all-purpose "buttons" one. Can reevaluate this if it causes problems.
- Extra deck sections are now accessible in-game. The ante tab on desktop and the sideboard tab on mobile have been replaced with a dropdown that can hold any extra zones that are in use.
- Fix for AnimateAI not using its preferred target for mandatory targeted abilities.
- Fix for target prompts being blank when a targeted DB is more than 2 SubAbilities out from the StackDescription


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -1585,6 +1590,18 @@ private boolean stateBasedAction704_attach(Card c, CardCollection unAttachList)
         return checkAgain;
     }
 
+    private boolean stateBasedAction_Contraption(Card c) {
+        //There isn't a rule-based source for this SBA, but it seems like something that should exist given the one for Space Sculptor.
```
tool4ever, line 1594: They said this on the Unstable mechanics article:
> In some very unusual cases (and in this set, that's saying something), a Contraption will find its way onto the battlefield without being assembled. If that ever happens, the card either goes to the scrapyard if it's an actual Contraption card or it goes to the graveyard if it's not (for example, a creature card that became a copy of a Contraption).

forge-gui/src/main/java/forge/gui/card/CardDetailUtil.java
```diff
@@ -296,9 +296,7 @@ public static String composeCardText(final CardStateView state, final GameView g
             area.append("Effect");
         }
         // card text
-        if (area.length() != 0) {
-            area.append("\n");
-        }
+        area.append("\n");
```
tool4ever, line 299: image

tool4ever (commented):
Looks like a pretty clean implementation, I see no major showstoppers :)

forge-game/src/main/java/forge/game/GameAction.java
Jetz72: Ahhhh, I'd been reading the wiki article which doesn't mention that bit. Good catch.

forge-gui/src/main/java/forge/gui/card/CardDetailUtil.java
Jetz72: I think the idea was to clean that up by removing the conditionals and just trim off the blank line at the end. Then I guess I forgot to do the second part of that. Should look right now.

tool4ever (approved):

PR #6723, opened 2024-12-26
[[Adventure]] Jace quickfix

Simisays:
I'll rework this boss more profoundly later, this is only a hotfix.

### DIFF
--- a/forge-gui/res/adventure/common/custom_cards/jace_boss_effect.txt
+++ b/forge-gui/res/adventure/common/custom_cards/jace_boss_effect.txt
@@ -2,11 +2,11 @@ Name:Jace Boss Effect
 ManaCost:no cost
 Colors:blue
 Types:Enchantment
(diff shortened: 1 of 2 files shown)


forge-gui/res/adventure/common/custom_cards/jace_boss_effect.txt
```diff
@@ -3,7 +3,7 @@ ManaCost:no cost
 Colors:blue
 Types:Enchantment
 T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You | TriggerZones$ Command | Execute$ TrigChoose | TriggerDescription$ At the beginning of your end step, conjure a duplicate of a random creature card from your opponent's library onto the battlefield under your control. The duplicate perpetually becomes a blue Illusion creature with base power and toughness 2/2 and has "When this creature becomes the target of a spell or ability, sacrifice it."
-SVar:TrigChoose:DB$ ChooseCard | Choices$ Creature.OppOwn | ChoiceZone$ Library | AtRandom$ True | SubAbility$ DBConjure
+SVar:TrigChoose:DB$ ChooseCard | Choices$ Creature.OppOwn+cmcLE5 | ChoiceZone$ Library | AtRandom$ True | SubAbility$ DBConjure
```
Agetian, line 6: This probably needs to be reflected in the trigger description and in the Oracle text below (since neither currently says that it only affects creatures with mana value 5 or less)

PR #6729, opened 2024-12-29
Order hand option

tool4ever:
Revisited this idea from some older attempt, feels pretty helpful when drawing lots of cards


Agetian (approved):

Agetian:
A nice QoL effort! :) Looks good to me for the initial implementation, might be a good idea to eventually add other sorting options too :+1:

PR #6733, opened 2024-12-31
Extra cost gift rework

Hanmac:
- [x] Used PromisedGift in xCount
- [x] moved chooseOptionalCosts into SpellAbilityAi (so it can be used later)
- [x] Check for Invalid Targets if cost would be added
- [x] Update SpellAbilityAi cases for when it should use OptionalCost or not (this is probably for PromisedGift specific)

### DIFF
--- a/forge-ai/src/main/java/forge/ai/SpellAbilityAi.java
+++ b/forge-ai/src/main/java/forge/ai/SpellAbilityAi.java
@@ -13,6 +13,7 @@ import forge.card.mana.ManaCost;
 import forge.card.mana.ManaCostParser;
 import forge.game.GameEntity;
 import forge.game.card.Card;
+import forge.game.card.CardCopyService;
 import forge.game.card.CardState;
 import forge.game.card.CounterType;
 import forge.game.cost.Cost;
@@ -23,6 +24,8 @@ import forge.game.player.Player;
 import forge.game.player.PlayerActionConfirmMode;
 import forge.game.player.PlayerController.BinaryChoiceType;
 import forge.game.spellability.AbilitySub;
+import forge.game.spellability.OptionalCost;
+import forge.game.spellability.OptionalCostValue;
 import forge.game.spellability.SpellAbility;
 import forge.game.spellability.SpellAbilityCondition;
 import forge.game.zone.ZoneType;
@@ -305,7 +308,7 @@ public abstract class SpellAbilityAi {
      */
     public boolean chkDrawbackWithSubs(Player aiPlayer, AbilitySub ab) {
         final AbilitySub subAb = ab.getSubAbility();
-        return SpellApiToAi.Converter.get(ab.getApi()).chkAIDrawback(ab, aiPlayer) && (subAb == null || chkDrawbackWithSubs(aiPlayer, subAb));
+        return SpellApiToAi.Converter.get(ab).chkAIDrawback(ab, aiPlayer) && (subAb == null || chkDrawbackWithSubs(aiPlayer, subAb));
     }
 
     public boolean confirmAction(Player player, SpellAbility sa, PlayerActionConfirmMode mode, String message, Map<String, Object> params) {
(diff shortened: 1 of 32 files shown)


forge-ai/src/main/java/forge/ai/SpellAbilityAi.java
```diff
@@ -410,4 +413,61 @@ public CounterType chooseCounterType(List<CounterType> options, SpellAbility sa,
...
+                if (ComputerUtilCard.checkNeedsToPlayReqs(copy, kickedSaCopy) != AiPlayDecision.WillPlay) {
+                    continue; // don't choose kickers we don't want to play
+                }
+            }
+
+            if (ComputerUtilCost.canPayCost(fullCostSa, player, false)) {
+                // check for additional Cost breaking targets AI had already chosen
+                boolean targeting = false;
+                SpellAbility sub = fullCostSa;
+                while (sub != null) {
```
tool4ever, line 442: doesn't make sense imo
AI would choose optional cost before API sets targets anyway so this might just stop it from playing them at all

Hanmac: normally yes, targets would be set after optional Cost, but not for AI
AI would set targets before the spell is added to the Stack (and optional Costs are chosen)

tested with `Into the Flood Maw`:
- without gift, bounce creature
- with gift, bounce nonland

at this point in the code, right now, ChangeZoneAi didn't know about Gift
which means, it would set the Target on the "bounce creature" SA

if now AI would choose to add the Optional Cost, both SA become invalid

my next work needs to be teach AI in ChangeZoneAi about Gift and what that means for targets.
So that AI can use it to bounce nonlands (that are better than a fish token)

tool4ever: that's not correct:
**`chooseOptionalCosts` is called before `canPlayAI`**
the reason why you think it works is because on the second try it might reuse the old target from the previous iteration - but that can obviously lead to other problems

You almost fixed it by cleaning up the ValidStack checks, but I still had to patch in the castSA so it knows which sub is targetless (should also help with other cards in general)

Hanmac: and now you broke it by forcing the AI to pay for optional cost

tool4ever: ?
that's just the standard behaviour until more logic gets added

Hanmac: This MR was planned to fix the Logic when to use Optional Costs
Like with Into the Flood Maw

Hanmac: @tool4ever the problem is this part:

https://github.com/Card-Forge/forge/blob/386ed8a08290c6201c8a6207e500d03ba5d4bf16/forge-ai/src/main/java/forge/ai/ComputerUtilAbility.java#L117-L138

If AI decides to add Optional Cost, it does Ignore the Base Ability, even if there is no benefit

tool4ever: How about:
```
if list.contains (PromisedGift cost)
also add original SA before it
```
then it will be checked first

Hanmac: @tool4ever found the main problem: "optionalCosts" in SpellAbility isn't deep copied when SpellAbility.copy is called

meaning when adding optionalCosts to the second SA, it does add it also to the first one
which then messes up the AI

i will try to fix it tomorrow, if you are not faster

tool4ever: I can't confirm that observation, in fact this logic seems to prevent that from happening:
https://github.com/Card-Forge/forge/blob/33e26c1ad2c4a915f2c883002df576772a786790/forge-game/src/main/java/forge/game/spellability/SpellAbility.java#L534-L538

Hanmac: Maybe you can help me debug why AI always used the Optional Cost SA

The OptionalCost = copy should have been worked?

tool4ever: Certainly not always? It's just gonna depend on API specifics but I had no problem getting it to play the basic _Blooming Blast_ on master with the approach I mentioned above...
But imo there's no good case for custom written logic so I wonder why you don't just offer support for scripters like with `NeedsToPlayKicked` instead 🤔

Hanmac: @tool4ever what about this part?

https://github.com/Card-Forge/forge/blob/33e26c1ad2c4a915f2c883002df576772a786790/forge-game/src/main/java/forge/game/spellability/SpellAbility.java#L440-L457

doesn't canPlayWithOptionalCost messes with the Original SA instead of making a copy?
for some reason it did go into that line, i need to debug more

tool4ever: no, that also copies it first

Hanmac: i think the main problem i run into was that AI is slightly different for SubAbilties

so for non-gift:
* Spell Change Zone with Target
* **SubAbility** ChangeZone without Target

and for gift:
* Spell Change Zone without Target
* **SubAbility** ChangeZone with Target

For AI, the second one looks more preferable

but that's probably something @Agetian need to look into later

Hanmac:
@tool4ever i used AITgts for these 2 cards which have `0.1` and `1.0` targets to stop it from using it on Targets where Additional Cost is not needed

Agetian:
Looks like a nice effort, thanks! :+1:

Agetian (approved):

PR #6736, opened 2025-01-01
Remove obsolete GameEvent

tool4ever:
This avoids extra handling and redundant View updates for tokens which got introduced to improve their card stack refreshes.
Tested on mobile and actually makes it smoother on Desktop now too.


forge-game/src/main/java/forge/game/ability/effects/ProtectAllEffect.java
```diff
@@ -87,6 +88,7 @@ public void resolve(SpellAbility sa) {
 
             for (final Card tgtC : list) {
                 tgtC.addChangedCardKeywords(gainsKWList, null, false, timestamp, null, true);
+                game.fireEvent(new GameEventCardStatsChanged(tgtC));
```
Hanmac, line 91: Why not do this event in the addChangedCardKeywords function?

tool4ever: because most other effects like Pump can do more than just add KW, so while the View properties should be updated it's better to only call the event once at the end

Hanmac (approved):

forge-game/src/main/java/forge/game/ability/effects/ProtectAllEffect.java
Hanmac: I still think the event should be part of a function

tool4ever: Hmn, not sure it helps to override it everywhere when usually the effect will have to do the work instead e.g.:

image

I might do more work on the events later, maybe then it becomes easier 🤔

PR #6743, opened 2025-01-04
remove add-opens from mandatory.java.args, use MANIFEST.mf

Hanmac:
fixes the --add-opens problem with the run-scripts using the Manifest.mf

### DIFF
--- a/forge-gui-desktop/pom.xml
+++ b/forge-gui-desktop/pom.xml
@@ -18,28 +18,7 @@
         <parsedVersion.incrementalVersion>0</parsedVersion.incrementalVersion>
         <timestamp>${maven.build.timestamp}</timestamp>
         <maven.build.timestamp.format>yyyy-MM-dd HH:mm:ss</maven.build.timestamp.format>
-        <mandatory.java.args>--add-opens java.desktop/java.beans=ALL-UNNAMED
-            --add-opens java.desktop/javax.swing.border=ALL-UNNAMED
-            --add-opens java.desktop/javax.swing.event=ALL-UNNAMED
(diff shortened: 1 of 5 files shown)


forge-gui-desktop/pom.xml
```diff
@@ -18,28 +18,7 @@
...
-            --add-opens java.desktop/java.awt.font=ALL-UNNAMED
-            --add-opens java.base/jdk.internal.misc=ALL-UNNAMED
-            --add-opens java.base/sun.nio.ch=ALL-UNNAMED 
-            --add-opens java.base/java.nio=ALL-UNNAMED
-            --add-opens java.base/java.math=ALL-UNNAMED
-            --add-opens java.base/java.util.concurrent=ALL-UNNAMED
-            --add-opens java.base/java.net=ALL-UNNAMED 
-            -Dio.netty.tryReflectionSetAccessible=true -Dfile.encoding=UTF-8
-        </mandatory.java.args>
+        <mandatory.java.args>-Xmx4096m -Dio.netty.tryReflectionSetAccessible=true -Dfile.encoding=UTF-8</mandatory.java.args>
```
tool4ever, line 21: where will they get picked up for the MANIFEST if you remove them here?

Hanmac: @tool4ever there:

https://github.com/Card-Forge/forge/blob/9ee2cfbf1846dc0cb8f423b029bd6d2d3d5cadb9/forge-gui-desktop/pom.xml#L170-L177

The Manifest.mf stuff doesn't have problems with such NewLines, it adds LineBreaks anyway

PR #6746, opened 2025-01-04
Fix some NPEs from Sentry

tehdiplomat:


Hanmac:
we need to add some CodeStype to check if java imports should be before or after the forge imports

tehdiplomat:
@Hanmac Removed teh import optimzes.

Hanmac (approved):

PR #6751, opened 2025-01-07
Improved AI tapland playing

marthinwurer:
Improved AI's handling of playing tapped lands. 

AI now considers what colors it has available and if an additional untapped land would help them cast a spell this turn before choosing to play an untapped land over a tapped land. When picking the land to play, it also takes into account the colors it can tap for, the amount of mana it can tap for, and the number of basic land types it has.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -30,6 +30,7 @@ import forge.card.CardType;
 import forge.card.MagicColor;
 import forge.card.mana.ManaAtom;
 import forge.card.mana.ManaCost;
+import forge.card.mana.ManaCostShard;
 import forge.deck.Deck;
 import forge.deck.DeckSection;
 import forge.game.*;
@@ -73,6 +74,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
+import static forge.ai.ComputerUtilMana.getAvailableManaEstimate;
 import static java.lang.Math.max;
 
 /**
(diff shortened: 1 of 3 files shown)


forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -570,11 +576,41 @@ private Card chooseBestLandToPlay(CardCollection landList) {
...
+                boolean found = false;
+                for (Card c : nonLandsInHand) {
+                    // TODO make this work better with split cards and Monocolored Hybrid
+                    if (c.getManaCost().getCMC() == max_inc + mana_available) {
+                        found = true;
+                        break;
+                    }
+                }
+
+                if (found) {
```
tool4ever, line 603: Looks like you're attempting to improve the early game sequencing so AI has a better chance of dropping a tapland when it's not needed? 🤔
I'd argue this currently comes at the cost of worse late game decisions, since at that time you'll usually have a much higher chance of extra drawing into something castable + you'd rather cast multiple (or big X) spells anyway which your logic above would miss and risk AI to drop all its taplands first

Considering you're also increasing the calculations done on all that tradeoff doesn't seem worth it imo...

Maybe turn count could be a rough estimate of when to stop doing this new behaviour

forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -570,11 +576,41 @@ private Card chooseBestLandToPlay(CardCollection landList) {
...
+
+                int mana_available = 0;
+                for (Card c : player.getCardsIn(ZoneType.Battlefield)) {
+                    mana_available += c.getMaxManaProduced();
+                }
+
+                boolean found = false;
+                for (Card c : nonLandsInHand) {
+                    // TODO make this work better with split cards and Monocolored Hybrid
+                    if (c.getManaCost().getCMC() == max_inc + mana_available) {
```
tool4ever, line 597: doesn't this need kind of a sanity check like `max_inc` > 0 so you're not thinking something like _Maze of Ith_ gets you in casting range?

forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -570,11 +576,41 @@ private Card chooseBestLandToPlay(CardCollection landList) {
...
-                landList = nonTappedLands;
+                // get the costs of the nonland cards in hand and the mana we have available.
+                // If adding one won't make something new castable, then pick a tapland.
+                int max_inc = 0;
+                for (Card c : nonTappedLands) {
+                    max_inc = max(max_inc, c.getMaxManaProduced());
+                }
+
+                int mana_available = 0;
+                for (Card c : player.getCardsIn(ZoneType.Battlefield)) {
```
tool4ever, line 590: should probably use `getAvailableManaEstimate` instead

marthinwurer: Thanks, I didn't know that existed.

forge-ai/src/main/java/forge/ai/AiController.java
marthinwurer: Good catch!

forge-ai/src/main/java/forge/ai/AiController.java
marthinwurer: I agree that it's not great for late game performance, especially with decks that draw a lot. It slipped my mind because I was focused on the early game where it's most evident. Instead of a turn cutoff, we might be able to use a mana available cutoff. I'd prefer that to be dynamic, based on the maximum mana value of any card in the deck, rather than a pure turn thing. Of course, that might cause problems with reanimator decks or other decks that cheat things in, but it's a reasonable start. Another possible improvement would be to check the combinations of different cards in hand and choose an untapped land if any combination could be cast with an additional mana instead of just any one card. X spells would then always want an untapped land, if the AI would want to play it that turn. We could also skip the tapped lands if there's enough lands in hand too. There are lots of options for improvement. 

Would adding a mana available threshold (say, 5 or 6) for the tapped land logic get the logic to a mergeable point?

tool4ever: Hmn yea good ideas :)
maybe just basing it on `getAvailableManaEstimate` would be fine if it now gets called anyway

I think this function still needs to be reasonably fast, so not too much O(n^2) stuff
We're still doing heuristics here after all, there's always simulation for the "perfect play" approach

marthinwurer: Updated it with calling the other function as well as checking for playable x spells and for the current amount of mana available.

tool4ever (approved):

PR #6754, opened 2025-01-07
Replace Cling with jupnp

neoFuzz:
```
Update forthline cling dependency to the more supported jupnp, removes http source.
FServerManager.java file formatted as a teduly
```
I believe this complies but Android builds fail to complete. The proguard step ends in error which fails the build

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
+++ b/forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
@@ -19,10 +19,10 @@ import io.netty.channel.socket.nio.NioServerSocketChannel;
 import io.netty.handler.codec.serialization.ClassResolvers;
 import io.netty.handler.logging.LogLevel;
 import io.netty.handler.logging.LoggingHandler;
-import org.fourthline.cling.UpnpService;
-import org.fourthline.cling.UpnpServiceImpl;
-import org.fourthline.cling.support.igd.PortMappingListener;
-import org.fourthline.cling.support.model.PortMapping;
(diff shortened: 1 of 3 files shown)


Hanmac:
@Agetian @kevlahnota can you check the Android built with proguard?

forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
```diff
@@ -73,6 +60,39 @@ public static FServerManager getInstance() {
...
+            if (in != null) {
+                try {
+                    in.close();
+                } catch (IOException e) {
+                    e.printStackTrace();
+                }
+            }
+        }
+        return null;
+    }    private final Thread shutdownHook = new Thread(() -> {
```
Hanmac, line 82: move this property back again?

neoFuzz:
```
public static FServerManager getInstance() {
        if (instance == null) {
            instance = new FServerManager();
            instance.shutdownHook = new Thread(() -> {
                if (instance.isHosting()) {
                    instance.stopServer(false);
                }
            });
        }
        return instance;
    } 
```
Would this work instead? Seems more logical

Hanmac:
@tehdiplomat I think this is clean now? I reverted some of the code changes back to keep the MR as minimum as possible

I needed to manually started the Android build in the Actions, but they did run

Hanmac (approved):

Agetian:
Agreed, I'd say let's give it a shot and see how it goes :)

PR #6758, opened 2025-01-09
CardView: added extra check for Ward and Annihilator icons

### LINKED ISSUE #6755: Very minor issue with Idris, Soul of the TARDIS and Ward/Annihilator icons
**Describe the bug**
After the recent fix Idris is now gaining Ward/Annihilator, but the icons for those abilities aren't appearing on the card overlay. It's not a big deal, like the title of this report says, but it would be nice if they appeared.

 - Forge Version: forge-installer-2.0.02-SNAPSHOT-01.07

Hanmac:
Closes #6755

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardView.java
+++ b/forge-game/src/main/java/forge/game/card/CardView.java
@@ -1603,7 +1603,7 @@ public class CardView extends GameEntityView {
         }
         void updateKeywords(Card c, CardState state) {
             c.updateKeywordsCache(state);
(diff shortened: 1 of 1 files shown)


tool4ever:
@Hanmac
I'm having a hard time deciding if this is correct behaviour or a bug (so we should manually remove KW for Idris instead)

E.g. would those triggers still be disabled by _Nowhere to Run_ ? 🤔

Another problem related to undesired interactions can occur with "fake" trigger, e.g. when exiling a card that normally has Daybound (but it was a copy) Idris could still make it Day.

tool4ever (approved):

PR #6760, opened 2025-01-11
Move Altar of the Pantheon to its own static

Hanmac:
> 700.5a A player’s devotion to each color and combination of colors, taking into account any effects that modify devotion, is calculated after considering any copy, control, or text-changing effects but before any other effects that modify the characteristics of permanents. This is an exception to 613.10. See also rule 613, “Interaction of Continuous Effects.”


tool4ever (approved):
working now
> If Altar of the Pantheon loses its abilities, you'll still get a bonus to your devotion. This is because its effect is applied before any effect that would make it lose its abilities.

PR #6767, opened 2025-01-13
Misc cleanup

### LINKED ISSUE #6765: Spirit cleric tokens from Hallowed Haunting do not tap for white mana with convoke
The white spirit cleric tokens produced by Hallowed Haunting tap for colorless but not white mana for a convoke spell.

 - OS: Windows 10
 - Browser: Chrome
 - Version: Forge 2.0.00

tool4ever:
Closes #6765


.mvn/local-settings.xml
```diff
@@ -1,9 +1,8 @@
-<!--
-    Derived from: https://stackoverflow.com/a/67002852
--->
 <settings xmlns="http://maven.apache.org/SETTINGS/1.2.0"
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
           xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.2.0 http://maven.apache.org/xsd/settings-1.2.0.xsd">
+<!--
+    Derived from: https://stackoverflow.com/a/67002852
```
Hanmac, line 5: This section can be removed

Hanmac:
You are doing way too many unrelated changes in one MR again

.mvn/local-settings.xml
tool4ever: I thought it might be a good idea to keep it commented out in case we need it for something else again but I guess that's not very likely :)

tool4ever:
Sorry but time is short lately and not splitting simple stuff across too many branches is one way I try to still bring some fixes :/
Though I might have one holi-day next week in which I could see about reorganizing this a bit

forge-gui/res/cardsfolder/p/parnesse_the_subtle_brush.txt
```diff
@@ -4,8 +4,8 @@ Types:Legendary Creature Vampire Wizard
...
 SVar:TrigCounter:DB$ Counter | Defined$ TriggeredSourceSA | UnlessCost$ PayLife<4> | UnlessPayer$ TriggeredSourceSAController
-T:Mode$ SpellCopy | ValidActivatingPlayer$ You | NoResolvingCheck$ True | TriggerZones$ Battlefield | Execute$ TrigTarget | TriggerDescription$ Whenever you copy a spell, up to one target opponent may also copy that spell. They may choose new targets for that copy.
-SVar:TrigTarget:DB$ Pump | ValidTgts$ Opponent | SubAbility$ DBCopy
-SVar:DBCopy:DB$ CopySpellAbility | TargetMin$ 0 | TargetMax$ 1 | Defined$ TriggeredSpellAbility | AILogic$ Always | Controller$ Targeted | Optional$ True | MayChooseTarget$ True
+T:Mode$ SpellCopy | ValidActivatingPlayer$ You | TriggerZones$ Battlefield | Execute$ TrigTarget | TriggerDescription$ Whenever you copy a spell, up to one target opponent may also copy that spell. They may choose new targets for that copy.
+SVar:TrigTarget:DB$ Pump | ValidTgts$ Opponent | TargetMin$ 0 | TargetMax$ 1 | SubAbility$ DBCopy
+SVar:DBCopy:DB$ CopySpellAbility | Defined$ TriggeredSpellAbility | AILogic$ Always | Controller$ Targeted | Optional$ True | MayChooseTarget$ True
```
Hanmac, line 9: might still move the Target in the CopySpellAbility line and skip the extra Pump one?

Hanmac:
why the removal of "forge_checks.xml" ?

tool4ever:
identical file already exists in the same location in forge-gui

Hanmac:
> identical file already exists in the same location in forge-gui

That file might be per-project basis, for example, I have each sub forge directory as different Project inside my eclipse

tool4ever:
it's just checkstyle configuration which should eventually be handled identically across subprojects and therefore only require a single source file that needs to be updated

Hanmac:
> it's just checkstyle configuration which should eventually be handled identically across subprojects and therefore only require a single source file that needs to be updated

The problem for this: if the IDE only sees the SubProjects like in my case, it can't see the files outside, so it wouldn't see that configuration

and I already added `.checkstyle` to gitignore so it wouldn't pollute the "untracked" files for me

tool4ever:
ugh I'll just restore it then 🤷‍♂️

Hanmac:
Yeah, it might need to be clean up, but I don't think we should do that for now with this MR

forge-gui/res/cardsfolder/p/parnesse_the_subtle_brush.txt
tool4ever: Hmn, looks like that would require some more refactoring first as that gets prevented pretty early here:

https://github.com/Card-Forge/forge/blob/f92c570bff5ba7624181729f264053b2d3a1a6c9/forge-game/src/main/java/forge/game/ability/AbilityFactory.java#L211-L216

most likely not worth it / too unrelated? 🤔

Hanmac: ugh, ok so i will look into this another time

PR #6770, opened 2025-01-15
Add ${androidBuildTools} to pom.xml

neoFuzz:
Straightforward, add ${androidBuildTools} so the Android Build Tools version can updated via the variable


Hanmac:
Should the `platform` a line above also be replaced by variable?

neoFuzz:
> Should the `platform` a line above also be replaced by variable?

Good idea. makes that central too

tehdiplomat:
I'll merge this and test it on todays Android build. If it breaks, i'll roll it back.

PR #6778, opened 2025-01-16
Improved AI Shockland play

marthinwurer:
- AI now knows that taplands can enter untapped
- Added a hacky fix to the shockland payment logic so that it will pay if there's just a card of that mana cost. 
- Added tests for AI land playing logic

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -23,6 +23,7 @@ import com.google.common.collect.Lists;
 import forge.ai.AiCardMemory.MemorySet;
 import forge.ai.ability.ChangeZoneAi;
 import forge.ai.ability.LearnAi;
+import forge.ai.ability.TapAi;
 import forge.ai.simulation.GameStateEvaluator;
 import forge.ai.simulation.SpellAbilityPicker;
 import forge.card.CardStateName;
@@ -74,6 +75,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 import static forge.ai.ComputerUtilMana.getAvailableManaEstimate;
+import static forge.game.ability.AbilityUtils.calculateUnlessCost;
 import static java.lang.Math.max;
 
 /**
@@ -542,6 +544,8 @@ public class AiController {
         }
 
         // try to skip lands that enter the battlefield tapped if we might want to play something this turn
+        // TODO figure out a better life level for shocking in lands
+        boolean check_shocks = player.getLife() > 10;
         if (!nonLandsInHand.isEmpty()) {
             CardCollection nonTappedLands = new CardCollection();
             for (Card land : landList) {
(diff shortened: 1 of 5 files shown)


forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -562,6 +566,14 @@ private Card chooseBestLandToPlay(CardCollection landList) {
                     if (reSA == null || !ApiType.Tap.equals(reSA.getApi())) {
                         continue;
                     }
+                    // check unlesscost parameters. At this point, it should be a tapapi ability
+                    if (reSA.hasParam("UnlessCost")) {
+                        String unlessCost = reSA.getParam("UnlessCost").trim();
+                        Cost cost = calculateUnlessCost(reSA, unlessCost, true);
+                        if (SpellApiToAi.Converter.get(ApiType.Tap).willPayUnlessCost(reSA, player, cost, false, new PlayerCollection(player))) {
```
Hanmac, line 573: `SpellApiToAi.Converter.get(reSA)` 
no need for extra Type import

Also, i don't like how `calculateUnlessCost` gets extra imported

forge-ai/src/main/java/forge/ai/ability/TapAi.java
```diff
@@ -105,12 +105,15 @@ public boolean willPayUnlessCost(SpellAbility sa, Player payer, Cost cost, boole
...
                                 continue;
                             }
-                            // Check if the AI intends to play the card and if it can pay for it with the mana it has
-                            boolean willPlay = ComputerUtil.hasReasonToPlayCardThisTurn(payer, c);
-                            boolean canPay = c.getManaCost().canBePaidWithAvailable(ColorSet.fromNames(ComputerUtilCost.getAvailableManaColors(payer, source)).getColor());
-                            if (canPay && willPlay) {
+                            if (payer.getLife() > 10 || landsize < 3) {
                                 return true;
                             }
+//                            // Check if the AI intends to play the card and if it can pay for it with the mana it has
```
tool4ever, line 111: way too hacky

forge-ai/src/main/java/forge/ai/AiController.java
marthinwurer: Thanks for catching and handling this!

Hanmac:
Why the SimultionTest -> AiTest changes?

forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -542,6 +542,8 @@ private Card chooseBestLandToPlay(CardCollection landList) {
         }
 
         // try to skip lands that enter the battlefield tapped if we might want to play something this turn
+        // TODO figure out a better life level for shocking in lands
+        boolean check_shocks = player.getLife() > 10;
```
Hanmac, line 546: unused variable?

forge-ai/src/main/java/forge/ai/ability/TapAi.java
marthinwurer: After sleeping on it, I agree that it's a bit too hacky. I think the API that's being called in the original code represents the correct line of thought, but the actual implementation of that API doesn't match up with what I'd think it would. I don't think that the internal canPlaySA stuff should be caring about whether or not the AI wants to play it, just whether it has the ability to. Basically 'can' vs 'should'. I don't know what all is relying on the current can implying should logic there, though. I can try setting up some unit tests to make sure I don't break anything too much in the process. I probably won't be able to work on this much until the weekend.

marthinwurer:
> Why the SimultionTest -> AiTest changes?

I wanted to reuse the setup of those tests for my land drop tests and figured that I should move them up and into a non-simulation-specific and more generic parent class.

forge-ai/src/main/java/forge/ai/AiController.java
marthinwurer: Yep, good catch! A side effect of coding past midnight 😅

forge-ai/src/main/java/forge/ai/ability/TapAi.java
Jetz72: In my experience, those AI ability methods tend to play a bit fast and loose with the responsibilities of "can I play it", "should I play it", and "actually play it" (they handle assigning targets as well for instance). Took me a while to puzzle out which ones were called and when on the couple occasions I've dabbled in them. Maybe I'm not seeing the bigger picture but it'd be nice to have more documentation someday for how they're meant to be used.

PR #6780, opened 2025-01-16
Update ConniveAi.java

tehdiplomat:


forge-ai/src/main/java/forge/ai/ability/ConniveAi.java
```diff
@@ -18,6 +18,11 @@ protected boolean canPlayAI(Player ai, SpellAbility sa) {
             return false; // can't draw anything
         }
 
+        final int num = AbilityUtils.calculateAmount(host, sa.getParamOrDefault("ConniveNum", "1"), sa);
+        if (num == 0) {
```
tehdiplomat, line 22: I didn't test this yet. It might break if you need to payX for Connive.

Hanmac: Good thing that there isn't a PayX right now

But the trigger ones still need to target

Hanmac: There is also 
> Whenever you attack, target attacking creature connives X, where X is the number of attacking creatures.

But that can't be 0 either, while targeting

Agetian:
It looks like there's a compile-time error of some sort during the tests, can you please check it out?

tehdiplomat:
Yea i know i didnt import. I'll get back to this later in the week

tehdiplomat:
@Hanmac @Agetian alright this is ready to go now.

Hanmac (approved):

PR #6792, opened 2025-01-18
Card: add RenderForUI to disable Render for Effects

Hanmac:


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -6452,16 +6453,20 @@ public final String getMostRecentSet() {
...
+        }
         Card uiCard = getCardForUi();
         if(uiCard == null)
             return "";
         return uiCard.currentState.getImageKey();
     }
     public final void setImageKey(final String iFN) {
+        if (!getRenderForUI()) {
+            return;
+        }
```
Jetz72, line 6467: Will this affect the images shown in dialogs when the user is prompted to make a choice related to a card? I can't remember if DetachedCardEffects show up in those.

Hanmac: Probably not? 🤷‍♂️

tool4ever: I can't test right now but from looking at things I'd expect this breaks at least Companion Effect?

tool4ever (approved):
it should probably be moved out of effects package at some point

Issue #5642, opened 2024-07-18
Dependency Layers Meta Issue


Hanmac:
Collection of issues related to "Layer Dependency"
- #1850
- #4108
- #4680
- #4841
- #5636 

Which Layers should be affected?

* 613.1d Layer 4: Type-changing effects are applied.
Cards like Blood Moon
or Graaz, Unstoppable Juggernaut applying to previous non-creatures that got animated
* 613.1f Layer 6: Ability-adding effects, keyword counters, ability-removing effects, and effects that say an object can’t have an ability are applied.
like Cavalry Master that refers to other creatures with abilities
or stuff that makes the card loose all abilities
* 613.4b Layer 7b: Effects that set power and/or toughness to a specific number or value are applied. Effects that refer to the base power and/or toughness of a creature apply in this layer.
Andrios, Roaming Explorer for example, the pump ability otherwise wouldn't make the static ability apply for these cards

for which Layer 4 and unlikely Layer 6 could have ways to cause a dependency Loop

Related Issues:
- Pump Effects might need to use Static Layers too:
#144

Hanmac:
This is important for _Lifecraft Engine_
> Vehicle creatures you control are the chosen creature type in addition to their other types.

For this to work correct, Animate Effects would need to use StaticAbility Layers

@tool4ever @tehdiplomat @Northmoc  I might need some input there
Like if i need to do Animate → StaticAbility, then I don't want all the Effects in the Command Zone :(

Hanmac:
Places that use addChangedCardTypes to be changed for Layer 4 using this as LinkedHashMap:

- StaticAbilityContinuous
- MANABOND CounterType (playtest)
- Bestow (Add Aura Type + Remove CreatureType)
- Reconfigure (Remove Creature Type)
- AnimateEffect (base)
- AmassEffect

tool4ever:
Imo one of the better ressources (not too outdated and lots of examples) is:
https://outsidetheasylum.blog/dependency/

Like the link suggests we might consider implementing it as directed graph, which when visualized might help players and our debugging too.
This one looks promising: https://jgrapht.org/javadoc/org.jgrapht.core/org/jgrapht/alg/cycle/package-summary.html

Here's my rough pseudocode:

```
For Layer X
[...assume all CDA are already applied the same way if really needed]
	while unapplied Statics left
		For each possible pair of Statics {A,B}, {B,A} etc.
		(it's noteworthy that this should be a flat loop instead of recursion after each static)
[...]

PR #6797, opened 2025-01-19
AmassEffect: use Effect in Command Zone for Type Change

Hanmac:
Part of #5642 to reduce the amount of addChangedCardTypes


tool4ever:
I'm not sure there exists a dependency where not applying this one first (or rather always like before) would be required.
So for a short while until we support them this might behave a bit worse?
I guess ideally it would just call `AnimateEffectBase.doAnimate` but that's a pretty big refactor 😅

Hanmac:
@tool4ever effects that removes all creature types except chosen ones for example should apply first, and then the add Creature Type after

I might look into it refactoring AnimateBase later

But first I want to remove the use of addChangedCardTypes

tool4ever:
If you cast something like _Mercurial Transformation_ after Amass it should just be applied by timestamp order and overwrite everything.

tool4ever (dismissed):

Hanmac:
> If you cast something like _Mercurial Transformation_ after Amass it should just be applied by timestamp order and overwrite everything.

No, i mean using `Conspiracy` to transform some non Army into an Army. And then use the Amass Effect to turn it onto a Zombie/Orc/Sliver.


That's why this doesn't work:
> I'm not sure there exists a dependency where not applying **this one first** (or rather always like before) would be required.

tool4ever:
But that works already, this doesn't lead to a dependency by itself and you will have an Army Zombie

Hanmac:
> But that works already, this doesn't lead to a dependency by itself and you will have an Army Zombie

Yes, but i don't want it to be broken later by `applying this one first`

tool4ever (approved):

PR #6800, opened 2025-01-20
Minor cleanup

tool4ever:


forge-game/src/main/java/forge/game/ability/effects/ChangeZoneEffect.java
```diff
@@ -908,14 +908,6 @@ private void changeZonePlayerInvariant(Player chooser, SpellAbility sa, List<Pla
                 decider = player;
             }
 
-            if (sa.usesTargeting()) {
```
Hanmac, line 911: @tool4ever is that really not used anymore?

tool4ever: it's the sole responsibility of MagicStack to check target legality and remove them
the check is either redundant or wrong in the rare case that a player becomes illegal during resolving

Hanmac: I think the last piece that needed this was something like #4843, but that got fixed

Hanmac (approved):

PR #6819, opened 2025-01-21
AnimateSubAbility rework

### LINKED ISSUE #6816: AnimateSubAbility rework
Judges say the effect from it doesn't exist while doing ETB lookahead:
https://old.reddit.com/r/askajudge/comments/1ggxkoo/containment_priest_enduring_creatures/

My current idea in GameAction.changeZone:

1. Moved RE
2. check Cause param and create/reuse DetachedCardEffect
3. Remember the moving card onto effect
4. Apply statics, do ChangesZone trigger etc.

So this might actually be good for us to also clean up the multiple code for it in different effects.

tool4ever:
- [x] add zone ordering
- [x] update remaining scripts
- [x] update remaining effects

Closes #6816


tool4ever:
Already works like I want:
image
⬇️ only one will get creature type removed from timestamp order
image

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -725,6 +726,50 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
         return copied;
     }
 
+    private void handleStaticEffect(Card copied, SpellAbility cause) {
+        if (cause != null && cause.hasParam("StaticEffect") && copied.isPermanent()) {
+            final Card source = cause.getHostCard();
+            if (cause.hasParam("StaticEffectCheckSVar")) {
+                String cmp = cause.getParamOrDefault("StaticEffectSVarCompare$", "GE1");
```
Hanmac, line 733: the $ in the Param?

forge-game/src/main/java/forge/game/GameActionUtil.java
```diff
@@ -853,12 +857,28 @@ public static CardCollectionView orderCardsByTheirOwners(Game game, CardCollecti
                     subList.add(c);
                 }
             }
+            if (sa != null && sa.getActivatingPlayer() == p && sa.hasParam("StaticEffect")) {
+                // create helper card for ordering
+                eff = new DetachedCardEffect(sa.getHostCard(), "Static Effect of " + sa.getHostCard());
+                subList.add(eff);
```
Hanmac, line 863: This Detached Card Effect is not the same Effect used for GameAction, right?

This one is just for filtering?

forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
```diff
@@ -585,14 +585,16 @@ protected static void addLeaveBattlefieldReplacement(final Card eff, final Strin
...
+        return createEffect(sa, controller, name, image, controller.getGame().getNextTimestamp());
+    }
+    public static Card createEffect(final SpellAbility sa, final Player controller, final String name, final String image, final long timestamp) {
         final Card hostCard = sa.getHostCard();
         final Game game = hostCard.getGame();
         final Card eff = new Card(game.nextCardId(), game);
 
-        eff.setGameTimestamp(game.getNextTimestamp());
+        eff.setGameTimestamp(timestamp);
         eff.setName(name);
```
Hanmac, line 598: @tool4ever should we more setSetCode and setRarity be moved to this place too?

from there:
https://github.com/Card-Forge/forge/blob/6d0bf1924248a8188dcec04c2f5382cf7fa45700/forge-game/src/main/java/forge/game/ability/effects/EffectEffect.java#L165-L170

forge-gui/res/cardsfolder/r/recommission.txt
```diff
@@ -1,11 +1,9 @@
 Name:Recommission
 ManaCost:1 W
 Types:Sorcery
-A:SP$ ChangeZone | Origin$ Graveyard | Destination$ Battlefield | AnimateSubAbility$ DBConditionEffect | ValidTgts$ Artifact.YouCtrl+cmcLE3,Creature.YouCtrl+cmcLE3 | TgtPrompt$ Select target artifact or creature card with mana value 3 from your graveyard | SpellDescription$ Return target artifact or creature card with mana value 3 or less from your graveyard to the battlefield. If a creature enters this way, it enters with an additional +1/+1 counter on it.
-SVar:DBConditionEffect:DB$ Effect | RememberObjects$ Remembered | ReplacementEffects$ ETBCreat | SubAbility$ DBCleanup
+A:SP$ Effect | RememberObjects$ Targeted | ReplacementEffects$ ETBCreat | ExileOnMoved$ Graveyard | SubAbility$ DBReturn | SpellDescription$ Return target artifact or creature card with mana value 3 or less from your graveyard to the battlefield. If a creature enters this way, it enters with an additional +1/+1 counter on it.
```
Hanmac, line 4: The problem, it doesn't respect `If a creature enters this way`

If for some reason it can't enter right now, the Effect does linger until EOT.

So if they enter somehow later, they still get the additional counters, but shouldn't

tool4ever: Hmn true (though this would also be a problem on the old version)

Any idea?

Hanmac: Maybe we could add a Trigger/GameCommand that does remove the Effect when the SpellAbility is removed from the Stack?

Also, i don't like it for the AI that the Effect happens before the ChangeZone Effect ... but that's something @Agetian can look into later

forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
tool4ever: yea I think that logic can be default

forge-gui/res/cardsfolder/r/recommission.txt
tool4ever: found a way

I did consider leaving parts of the old AnimateSubAbility$ behind for these, but that felt more like keeping a hacky workaround in the end with too little gain

Hanmac: hm it is a bit hacky ... maybe i will think about refactoring it later

forge-game/src/main/java/forge/game/GameActionUtil.java
tool4ever: indeed
in theory it might be reused but that would require more logic than worth it

tool4ever:
Well I maybe should have started with a more complicated one first...
Had to rearrange some parts so when _Bronzehide Lion_ returns as Aura it still gets a chance to become attached.
Now I'm a bit less happy with it and need to check if this causes no bad side effects ;(

>  If a nontoken permanent is a copy of Bronzehide Lion, it returns from its owner’s graveyard as an Aura with the two abilities granted by Bronzehide Lion and none of its normal abilities. **If it has any enters-the-battlefield replacement effects, those won’t apply.** The Aura keeps its name, colors, and any supertypes it may have.

Also the bold part confuses me:
- Is that a contradiction to the ruling that _Containment Priest_ can exile a returning _Enduring Innocence_?
- Unless I find a flaw in my thought process might have to ask Judge next 🤔

Hanmac:
> Well I maybe should have started with a more complicated one first... Had to rearrange some parts so when _Bronzehide Lion_ returns as Aura it still gets a chance to become attached. Now I'm a bit less happy with it and need to check if this causes no bad side effects ;(
> 
> > If a nontoken permanent is a copy of Bronzehide Lion, it returns from its owner’s graveyard as an Aura with the two abilities granted by Bronzehide Lion and none of its normal abilities. **If it has any enters-the-battlefield replacement effects, those won’t apply.** The Aura keeps its name, colors, and any supertypes it may have.
> 
> Also the bold part confuses me:
> 
>     * Is that a contradiction to the ruling that _Containment Priest_ can exile a returning _Enduring Innocence_?
> 
>     * Unless I find a flaw in my thought process might have to ask Judge next 🤔

ugh that sounds extra ugly ... i think the difference is that Bronzehide Lion would remove its own abilities, so it would also remove any affected replacement effects

tool4ever:
Based on the Judge answer it looks like this rules area is in pretty bad shape
Accordingly I will restore/align with the old behaviour which will respect the ruling at least until something new and official gets posted

Hanmac (approved):
the only things i don't like are the ones that create the Effect first ... and i wish the ChangesZone Effect would know about this :/

Hanmac (changes requested):
Need to fix the test cases

tool4ever:
@Hanmac
was the EffectZone changes

Hanmac (approved):

PR #6824, opened 2025-01-22
January 2025 cleanup: pass 6, part A

dracontes:
Dealing with typos and other lapses. Also chipping away at the work for `TgtPrompt$` while I was at it. Notable fixes:
- [Fear of Change](https://scryfall.com/card/ydsk/21/fear-of-change) was conjuring a card when it entered an empty battlefield, so I added a relevant condition to the `DB$ NameCard` line. Also the script was adjusted since neither `ChoiceTitle$` or `TgtPrompt$` displayed. All tested to satisfaction.


tool4ever (approved):
thanks

PR #6836, opened 2025-01-22
[[Adventure]] Changes to the restricted cards list.

Simisays:


Agetian:
I'm not sure if we should really restrict these cards in Forge, Adventure mode or not. I mean, they are a part of the history of Magic, even if there are indeed reasons to believe that they're not up to the community standards anymore due to their possible subtext interpretation and references. Still, I'm not a particular fan of rewriting history or pretending that things didn't happen or that these cards don't exist or never existed. Probably this could be an option, maybe even set to "on" by default - as in, the restricted list could be editable and the player could opt to either choose not to play with these cards or keep them in, depending on what makes them feel comfortable. This would then be similar to Scryfall, which, even though it restricts viewing the relevant graphics by default, still allows to both see and download the relevant art for someone who would like to see those cards. I do understand that the content of the cards may feel offensive/insulting to some people, and if anything, I'd recommend defaulting to keeping these cards restricted, while still offering players the option to play with all the cards from the early sets if they would like to experience them, except the ones that may be overpowered mechanically.

Granted, this is just my $0.02. I don't see any other good reason (I mean rule-wise or mechanics-wise) to restrict those cards, compared to the other cards on the list restricted for balance reasons.

Agetian:
By the way, one of the cards suggested for restriction is called "Stone-Throwing Devils", not "Devil".

Agetian (changes requested):
Voting to make this change optional, decided on Adventure start or in game options similar to "use all card variants" option, with default set to "on" due to negative historical connotations in card names, rule text or art, but with the ability for the player to opt out of any restrictions not based on game balance and mechanical considerations.

PR #6841, opened 2025-01-23
Cleanup: EffectZone$ Command, part 2

dracontes:


tool4ever:
@Hanmac
Don't really understand what makes the test fail from this 🐛

Hanmac:
@tool4ever we might need to debug it, but i can't right now

i restarted the test, to see if it was a hickup

Hanmac:
@tool4ever found it:

https://github.com/Card-Forge/forge/blob/2f390114b5f3d2db4e154da124ea70fbb52f2a7e/forge-game/src/main/java/forge/game/ability/effects/EffectEffect.java#L197
adds it to mapParams

but when the StaticAbility is copied, the map params are discarded:
https://github.com/Card-Forge/forge/blob/2f390114b5f3d2db4e154da124ea70fbb52f2a7e/forge-game/src/main/java/forge/game/CardTraitBase.java#L740-L741

imo it might be better to add a different property to StaticAbility, like this for Trigger and Replacement Effects:
https://github.com/Card-Forge/forge/blob/2f390114b5f3d2db4e154da124ea70fbb52f2a7e/forge-game/src/main/java/forge/game/TriggerReplacementBase.java#L47-L52

tool4ever (approved):

PR #6853, opened 2025-01-24
7 DFT + 1 DRC card

Fulgur14:


forge-gui/res/cardsfolder/upcoming/boom_scholar.txt
```diff
@@ -0,0 +1,8 @@
+Name:Boom Scholar
+ManaCost:1 R G
+Types:Creature Goblin Advisor
+PT:3/3
+S:Mode$ ReduceCost | ValidCard$ Permanent.YouCtrl+Other | Type$ Ability.Exhaust | Amount$ 2 | AffectedZone$ Battlefield | Description$ Exhaust abilities of other permanents you control cost {2} less to activate.
```
Hanmac, line 5: Could do `Activated.Exhaust` ?

Fulgur14: Yeah, I re-used code from Agatha of the Soul Cauldron. I guess Activated is not considered essential because no other abilities can have their cost reduced? Should Agatha be changed?

Hanmac: Agatha of the Vile Cauldron?, probably yeah
Trigger and Replacement effects are excluded right now anyway, but we might never know when that changes

tool4ever (approved):

PR #6861, opened 2025-01-24
StaticAbility: use getActiveZone functions

### LINKED ISSUE #6856: Static Abilties need better Zone Param like Trigger and Replacements
Currently, the Effect Zone for Static Abilities is checked there: (yeah i know, if forgot to touch CharacteristicDefining there)
https://github.com/Card-Forge/forge/blob/dd9bca2bd787fbff1afcbb7654f05bd9072a0621/forge-game/src/main/java/forge/game/staticability/StaticAbility.java#L330-L341
And also there?
https://github.com/Card-Forge/forge/blob/dd9bca2bd787fbff1afcbb7654f05bd9072a0621/forge-game/src/main/java/forge/game/staticability/StaticAbilityCantTarget.java#L107
And there?
https://github.com/Card-Forge/forge/blob/dd9bca2bd787fbff1afcbb7654f05bd9072a0621/forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java#L58-L60

Some Effects might manually set it:
https://github.com/Card-Forge/forge/blob/2f390114b5f3d2db4e154da124ea70fbb52f2a7e/forge-game/src/main/java/forge/game/ability/effects/EffectEffect.java#L197
adds it to mapParams

but when the StaticAbility is copied, the map params are discarded:
https://github.com/Card-Forge/forge/blob/2f390114b5f3d2db4e154da124ea70fbb52f2a7e/forge-game/src/main/java/forge/game/CardTraitBase.java#L740-L741

imo it might be better to add a different property to StaticAbility, like this for Trigger and Replacement Effects:
https://github.com/Card-Forge/forge/blob/2f390114b5f3d2db4e154da124ea70fbb52f2a7e/forge-game/src/main/java/forge/game/TriggerReplacementBase.java#L47-L52

I propose this:
- set = null, use Default (non-CDA = Battlefield)
- set is empty = ALL zones (so we don't need an extra full set?, or have ZoneType.valueOf("All") → full set?)
- otherwise, check if current Zone is in Set

Hanmac:
Closes #6856


forge-game/src/main/java/forge/game/zone/ZoneType.java
```diff
@@ -60,6 +60,9 @@ public static ZoneType smartValueOf(final String value) {
     }
 
     public static List<ZoneType> listValueOf(final String values) {
+        if ("All".equals(values)) {
+            return List.of(Battlefield, Hand, Graveyard, Exile, Stack, Library, Command);
```
dracontes, line 64: I took the liberty of starting work on replacing the arguments in the relevant `AffectedZone$` parameters: `Ante` is missing on this list.

Hanmac: @dracontes i don't know if Ante should be there?

dracontes: Well, ante showed up while I was doing this clean up, on Leyline of Transformation ([on repository](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/l/leyline_of_transformation.txt)), and I thought I should mention it.
Looking into it a bit more closely, it's admittedly the corner case of corner cases: a deprecated game zone with a single card, [Jeweled Bird](https://scryfall.com/card/chr/102/jeweled-bird), that moves cards outside of it.
I'm not sure how not considering it would interact with the sort of Conspiracy-like effects `AffectedZone$ All` covers. I am noticing a few perpetual effects, like Wilson, Bear Comrade ([on repository](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/w/wilson_bear_comrade.txt)), are still explicit, instead of being flagged as such. I take the latter is something to be addressed then.

Hanmac: @tool4ever your opinion? something we can touch later?

forge-game/src/main/java/forge/game/staticability/StaticAbility.java
```diff
@@ -327,16 +338,16 @@ public final boolean checkConditions() {
             return false;
         }
 
-        if (hasParam("EffectZone")) {
-            if (!getParam("EffectZone").equals("All")) {
+        if (!isCharacteristicDefining()) {
```
tool4ever, line 341: it seems this would change behaviour of _Lost Order of Jarkeld_ because it has both params
most likely it its script should be changed to nonCDA?

tool4ever: Ok, here's a PR that should remove all wrong CDA: #6872

afterwards @dracontes should be safe to clean up EffectZone$ from all remaining ones

forge-game/src/main/java/forge/game/zone/ZoneType.java
tool4ever: I think without a good case we can continue ignoring it just like those other Un-zones

Also that Wilson is scripted with the older perpetual approach
I wouldn't worry too much about it

forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
```diff
@@ -53,10 +54,11 @@ public static List<SpellAbility> alternativeCosts(final SpellAbility sa, final C
                     newSA.putParam("ManaRestriction", stAb.getParam("ManaRestriction"));
                 }
 
-                if (stAb.hasParam("AffectedZone")) {
-                    newSA.getRestrictions().setZone(ZoneType.smartValueOf(stAb.getParam("AffectedZone")));
-                } else if (!stAb.getHostCard().isImmutable() && stAb.hasParam("EffectZone") && !"All".equals(stAb.getParam("EffectZone"))) {
-                    newSA.getRestrictions().setZone(ZoneType.smartValueOf(stAb.getParam("EffectZone")));
+                if (!stAb.getHostCard().isImmutable()) {
```
tool4ever, line 57: Hmn, this doesn't look like equivalent logic too me
why do you no longer want to transfer AffectedZone to the AltCostSA?

Hanmac: I checked, there is no AlternativeCost effect with AffectedZone

forge-game/src/main/java/forge/game/staticability/StaticAbility.java
dracontes: To be clear, `EffectZone$ All` parameters in the context of static ability `S:` lines, right?

tool4ever: I was talking about CharacteristicDefining$ because it will make them redundant/ignored

dracontes: So, to be sure we're on the same page, [this line on Cognivore](https://github.com/Card-Forge/forge/blob/5385678076ffa3a2481da87fa2ac670bcb4d9d9c/forge-gui/res/cardsfolder/c/cognivore.txt#L6C1-L6C68) (and similar ones elsewhere) should have `EffectZone$ All` removed.

tool4ever: yea

forge-game/src/main/java/forge/game/staticability/StaticAbilityAlternativeCost.java
tool4ever: gotcha
can you just do `zones.get(0)` instead of `zones.stream().findFirst().get()` ?

tool4ever: hmn I guess then it doesn't matter

tool4ever (approved):

PR #6866, opened 2025-01-25
ReplaceToken: use Param from ReplacementEffect

### LINKED ISSUE #6855: Cleanup ReplaceTokenEffect
ReplaceTokenEffect uses inside AddToken Replacement uses ValidCard to match against the tokens.
But that mirrors the same as ValidToken in CreateToken.

The SpellAbility at resolve knows what it's ReplacementEffect is, so wouldn't it be better to just let the ReplacementEffect do the filter gain?

Hanmac:
Closes #6855


tool4ever (approved):
👍

PR #6868, opened 2025-01-25
Lost Monarch of Ifnir (DRC)

Fulgur14:


forge-gui/res/cardsfolder/upcoming/lost_monarch_of_ifnir.txt
```diff
@@ -0,0 +1,13 @@
+Name:Lost Monarch of Ifnir
+ManaCost:3 B
+Types:Creature Zombie Noble
+PT:4/4
+K:Afflict:3
+S:Mode$ Continuous | Affected$ Zombie.Other+YouCtrl | AddKeyword$ Afflict:3 | Description$ Other Zombies you control have afflict 3.
+T:Mode$ Phase | Phase$ Main | PhaseCount$ 2 | ValidPlayer$ You | TriggerZones$ Battlefield | ConditionCheckSVar$ X | Execute$ TrigMill | TriggerDescription$ At the beginning of your second main phase, if a player was dealt combat damage by a Zombie this turn, mill three cards, then you may return a creature card from your graveyard to your hand.
```
tool4ever, line 7: you still need to test it

Fulgur14: Hm, yeah, the condition seems to be always true, even if nothing attacked, not to mention dealt combat damage. So what is the problem...?

tool4ever: ConditionCheckSVar does not exist on triggers...

Fulgur14: OK, should work now, but... is there a reason for all these differences in syntax? Like, say, why does Pump use NumAtt$ and NumDef$ but Continuous uses AddPower$ and AddToughness$? It's like English, but without the excuse of being left unmanaged for centuries :(

tool4ever: well it depends :D
sometimes like this case for SpellAbility you'd need both (Condition*+ CheckSVar as restriction)
technically triggers only have Condition so I suppose the param name got shortened :p

as for Pump, yea that's just caused by legacy debt
maybe that one will sort itself if we refactor it with Layers logic

tool4ever (approved):

PR #6869, opened 2025-01-25
Dependency Graphs

### LINKED ISSUE #1850: Ability removal layer dependency not applying
I cast Lignify on an Archetype of Imagination. Apart from the Archetype itself, all the other creatures still had or didn't have flying as if the effect were there.

Since they're both an effect that applies in the ability layer, I believe due to dependency rules that the Lignify effect should apply first, ignoring timestamps, and totally removing the Archetype's effect.

I haven't had a chance to investigate whether this applies to other similar cards.

### LINKED ISSUE #4108: Imprisoned in the Moon does not suppress Pia Nalaar, Consul of Revival haste ability
**Describe the bug**
I put Imprisoned in the Moon on my opponent's Pia Nalaar, Consul of Revival on turn 3. Several turns later, my opponent was able to generate a couple of Thopters. They had haste even though nothing should give them haste, and he immediately attacked with them. 

**Expected behavior**
Imprisoned in the Moon should suppress Pia Nalaar, Consul of Revival's abilities such that Thopters do not get haste.

**Screenshots**
2023-11-03 19_07_29-Forge_ 1 6 59-SNAPSHOT-11 04

 - OS: Windows 11
 - Version: 1.6.59-SNAPSHOT-11.04

### LINKED ISSUE #4680: Layers Issue- Dryad of the Ilysian Grove and Life and Limb
If I have a Dryad of the Ilysian Grove and Life and Limb on the field, with non-Forest lands and no Saprolings and Dryad entering second, then Life and Limb should depend on Dryad (and not vice versa) which would cause all lands I control to become 1/1 Saproling creatures; but the engine only has cards that are already Forests become Saprolings. It seems like it's defaulting to timestamps when it should use dependency.

### LINKED ISSUE #4841: Kormus bell + Urbor, Tomb of Yawgmoth doesn't function right
**Describe the bug**
Kormus Bell does not make opponents' lands into creatures after having played Urborg, Tomb of Yawgmoth

**To Reproduce**
1. Play Urborg, Tomb of Yawgmoth... All lands are now Swamps.
2. Play Kormus Bell, all Swamps are now Black 1/1 creatures.

**Expected behavior**
regardless of order played, Kormus bell should always make swamps into creatures, and Urborg should always make all lands into swamps... this is a known combo... follow that up with a anti-anthem like Elesh norn and everything dies.

**Screenshots**
image

**Additional context**

### LINKED ISSUE #6669: Issue with Ygra, Eater of All and Impending
**Describe the bug**
I'm assuming this is a layers issue, but when Ygra is on the field and you cast Overlord of the Hauntwoods via Impending, it's treated as a "creature" and gains the artifact/food types. If you cast Overlord of the Hauntwoods via Impending first, then cast Ygra the typing is correct and Overlord of the Hauntwoods doesn't gain the creature only typing of artifact/food.

**Expected behavior**
For the Impending creatures to not gain artifact/food typing from Ygra when they are enchantments.

**Screenshots**
Screenshot_20241206_064252

 - Forge Version: forge-installer-2.0.01-SNAPSHOT-12.03

tool4ever:
This implements the algorithm I've outlined in #5642 :)

- [x] tweaked `StaticEffect` so it can undo single layers only
- [x] add test that would fail on master

To handle more cases `generateStaticAbilityResult` will need additional logic for the other layers.
But just comparing the affected objects should already support quite a lot.

Closes #4680
Closes #1850
Closes #4108
Closes #4841
Closes #6669


forge-game/pom.xml
```diff
@@ -34,5 +35,10 @@
             <artifactId>sentry-logback</artifactId>
             <version>7.15.0</version>
         </dependency>
+        <dependency>
+            <groupId>org.jgrapht</groupId>
```
Hanmac, line 39: guava has its own graph lib you might want to check out?

Hanmac:
```
final Comparator<StaticAbility> comp = (a, b) -> ComparisonChain.start()
                .compareTrueFirst(a.isCharacteristicDefining(), b.isCharacteristicDefining())
                .compare(a.getHostCard().getLayerTimestamp(), b.getHostCard().getLayerTimestamp())
                .result();
        staticAbilities.sort(comp);
```

maybe remove this, and separate them inside the layers loop?

forge-game/pom.xml
tool4ever: I compared them but it's rather basic...

Most importantly this already gives us cycle finding methods which if you check out those classes isn't all that trivial
I really do not want to reimplement one of these by hand 😅
E.g. here's the paper for the one I've chosen: https://link.springer.com/article/10.1007/BF01931370

Second bonus is that it also has support for visualization which I may try to add at some point too :)

tool4ever:
I think right now I do need them pre-sorted once, so I can always start checking from the earliest

tool4ever:
@Hanmac
we could define the `comp` as Static class field though
maybe better for performance instead of always creating it each time?

Hanmac:
> @Hanmac we could define the `comp` as Static class field though maybe better for performance instead of always creating it each time?

Look what I meant:
https://github.com/Card-Forge/forge/issues/5642#issuecomment-2601982519

I think filtering the list for CDA first would help

tool4ever:
Feel free and try to tweak it 🤔

I tested the sub-issues and easier ones work now, those have been linked for closing

Hanmac:
@tool4ever should we add more test cases?

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -1232,6 +1255,115 @@ public boolean visit(final Card c) {
...
+        // now the earlist one left is the correct choice
+        List<StaticAbility> statics = Lists.newArrayList(dependencyGraph.vertexSet());
+        statics.sort(Comparator.comparing(s -> s.getHostCard().getLayerTimestamp()));
+
+        return statics.get(0);
+    }
+
+    private List<Object> generateStaticAbilityResult(StaticAbilityLayer layer, StaticAbility stAb) {
+        List<Object> results = Lists.newArrayList();
+        if (layer == StaticAbilityLayer.CONTROL) {
```
Hanmac, line 1361: why is there only a case for Control? shouldn't the other layer also have something there?

tool4ever: see my description:
> To handle more cases generateStaticAbilityResult will need additional logic for the other layers.
But just comparing the affected objects should already support quite a lot.

For that we should investigate which params from the other layers are worth comparing.
But with my implementation this can be done in steps with more specific PR building on it.

e. g. for
> Layer 6: Yixlid Jailer and Necrotic Ooze
the part for GainsAbilitiesOf$ would need some refactoring so it can be reused

tool4ever:
@Hanmac
last test updated to prove loop detection works based on
https://apps.magicjudges.org/forum/topic/11405/

tool4ever:
Performance seems ok for average AI games:
image

It should be O(n*(n-1)) for the amount of statics. The logic shortcut I added probably helps...

Hanmac:
> Performance seems ok for average AI games: image
> 
> It should be O(n*(n-1)) for the amount of statics. The logic shortcut I added probably helps...

What does the rest of show? I mean should part is the most used one and should be optimized?

tool4ever:
Here's one from a BO5 4 player game with random decks:
image

We can still expect a lesser improvement if I refactor the players do be handled in their own loop. Just need to do more research if they even need dependency checking, so that's another follow-up which can be handled soon after this survives a few snapshots. 🤞

tool4ever:
That's a bit harder to tell recently because the AI timeout stuff creates so many threads.
Aggregating their calls in VisualVM requires some trickery I'm still trying to figure out.

But here's a Top10 of Hotspots:
image

If there's another quick-win I haven't found it this way yet.

Hanmac:
@tool4ever i think you can move the Comperator into a Const if you want

but i still think removing the CDA first from the list and then sorting by timestamp / dependency later separate might be better?

tool4ever:
There's no benefit from multiple sort calls since CDA need timestamp order too and this will also make sure they get applied first anyway.

I'm already skipping them for dependency order right now until a case for it gets supported.

➡️ Comparator extracted and converted to native Java

Hanmac:
i haven't found an example for CDA dependency yet, because they can only affect itself, it needs to be something on the same card, or gained something via copy layers

The Full dependency is only in effect when the TreeTable are replaced with LinkedHashMap
and for that to work correct, we need to update the Pump and Animate effects

tool4ever:
Indeed more work / PR will still remain...

Luckily with such effects:
- they're immutable so removal can't be caused in this way
- affect ``Card.IsRemembered`` or ``EffectSource`` so that's fixed too
- which only leaves the somewhat rarer third case

But is anything else needed here?
Let's please merge to fix what's already possible and worry about the rest step by step :)

Hanmac (approved):
i think this is good for now?

We just need to update the other effects soonish

PR #6888, opened 2025-01-26
Speed / Start your engines!

### LINKED ISSUE #6859: Speed and Start your Engines
- Player Property: Speed, starts at 0 max is 4
- Keyword `Start your Engines` uses a Static Always Trigger to update the Speed from 0 → 1
- Then the Player uses a DetachedCardEffect similar to the Ring Effect with a Trigger that triggers when an Opponent loses Life (only once per your turn)
- For both the Keyword and the Trigger we need a new SpellAbilityEffect that can increase your Speed.
- Player needs a getter to return `getSpeed`, and also a bool `isMaxSpeed`
- the MaxSpeed is used as Condition for CardTraits (it might be that the abilities aren't intrinsic and MaxSpeed adds them, but we aren't sure)

Northmoc:
- Closes #6859


forge-gui/res/cardsfolder/upcoming/embalmed_ascendant.txt
```diff
@@ -0,0 +1,12 @@
+Name:Embalmed Ascendant
+ManaCost:1 W B
+Types:Creature Zombie
+PT:1/2
+K:Start your engines
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigToken | TriggerDescription$ When this creature enters, create a 2/2 black Zombie creature token.
+SVar:TrigToken:DB$ Token | TokenScript$ b_2_2_zombie
+T:Mode$ ChangesZone | Origin$ Battlefield | Destination$ Graveyard | ValidCard$ Creature.YouCtrl | TriggerZones$ Battlefield | Execute$ TrigLoseLife | TriggerDescription$ Max speed — Whenever a creature you control dies, each opponent loses 1 life and you gain 1 life.
```
Hanmac, line 8: Forgot Max Speed Param on Trigger?

Hanmac (approved):
i think this is good to go?
The Sound is already icing on the Cake

and for possible "Start your Engine!" Keyword Icon from Arena, we can add that later

PR #6897, opened 2025-01-26
TriggerCrewedSaddled: unify

### LINKED ISSUE #6857: TriggerSaddled and TriggerCrewed Rework
For Canyon Vaulter, we need `Whenever this creature saddles a Mount`
and it seems we would need something like this for Saddle too:
https://github.com/Card-Forge/forge/blob/dd9bca2bd787fbff1afcbb7654f05bd9072a0621/forge-game/src/main/java/forge/game/zone/MagicStack.java#L424-L429

But if we could access the PaidList as CardProperty inside the Trigger for ValidSa, then we could just do this in `AbilityCast` Trigger above, and remove this Crewed Trigger.

Like this: `ValidSA$ Activated.Crew+Vehicle,Activated.Saddle+Mount` + in PayList check

----

The only Problem is `Tiana, Angelic Mechanic`, the Arena only card that says:
> When Tiana, Angelic Mechanic or another legendary creature you control crews a Vehicle, that Vehicle perpetually gets +1/+0.

Tiana triggers twice, so we need a per crew-member variant anyway

Hanmac:
Closes #6857


forge-game/src/main/java/forge/game/zone/MagicStack.java
```diff
@@ -421,11 +421,25 @@ public final void add(SpellAbility sp, SpellAbilityStackInstance si, int id) {
...
-                // Trigger crews!
-                runParams.put(AbilityKey.Vehicle, sp.getHostCard());
-                runParams.put(AbilityKey.Crew, sp.getPaidList("TappedCards", true));
-                game.getTriggerHandler().runTrigger(TriggerType.Crewed, runParams, false);
+            if (sp.isCrew() && sp.getHostCard().getType().hasSubtype("Vehicle")) {
+                Iterable<Card> crews = sp.getPaidList("TappedCards", true);
+                if (crews != null) {
+                    Map<AbilityKey, Object> crewParams = AbilityKey.mapFromCard(sp.getHostCard());
+                    for (Card c : crews) {
+                        crewParams.put(AbilityKey.Crew, c);
```
tool4ever, line 429: each trigger needs own runParams
or you overwrite it before it can be used

forge-gui/res/cardsfolder/a/adrestia.txt
```diff
@@ -6,11 +6,7 @@ K:Landwalk:Island
 T:Mode$ Attacks | ValidCard$ Card.Self | CheckSVar$ X | TriggerZones$ Battlefield | Execute$ TrigDraw | TriggerDescription$ Whenever CARDNAME attacks, if an Assassin crewed it this turn, draw a card. CARDNAME becomes an Assassin in addition to its other types until end of turn.
 SVar:TrigDraw:DB$ Draw | NumCards$ 1 | SubAbility$ DBAnimate
 SVar:DBAnimate:DB$ Animate | Defined$ Self | Types$ Assassin
-T:Mode$ BecomesCrewed | ValidVehicle$ Card.Self | Execute$ RememberCrew | Static$ True
-SVar:RememberCrew:DB$ PumpAll | ValidCards$ Assassin.TriggeredCrew | RememberPumped$ True
-T:Mode$ Phase | Phase$ End of Turn | Static$ True | Execute$ DBCleanup
-SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
-SVar:X:Remembered$Amount
+SVar:X:Crewed$Valid Assassin
```
tool4ever, line 9: this isn't fully accurate because the condition should look back in time
But I don't think crewedByThisTurn can be fully changed to the LKI paidList because some others use it for ingame stuff?

forge-gui/res/cardsfolder/upcoming/canyon_vaulter.txt
```diff
@@ -0,0 +1,8 @@
+Name:Canyon Vaulter
+ManaCost:1 W
+Types:Creature Kor Pilot
+PT:3/1
+T:Mode$ Crewed | ValidCrew$ Card.Self | Execute$ TrigPump | TriggerZones$ Battlefield | TriggerDescription$ Whenever this creature saddles a Mount or crews a Vehicle during your main phase, that Mount or Vehicle gains flying until end of turn.
+T:Mode$ Saddled | ValidCrew$ Card.Self | Execute$ TrigPump | TriggerZones$ Battlefield | Secondary$ True | TriggerDescription$ Whenever this creature saddles a Mount or crews a Vehicle during your main phase, that Mount or Vehicle gains flying until end of turn.
+SVar:TrigPump:DB$ Pump | Defined$ TriggeredCard | KW$ Flying
```
tool4ever, line 7: don't these need *LKICopy for now? :/
or should we try and fix that soon and leave new cards this way?

Hanmac: The TriggeredCard there isn't an LKI copy, so i don't think that would cause a problem?
but just in case, i added more LKICopy

forge-gui/res/cardsfolder/a/adrestia.txt
Hanmac: i changed the crewedThisTurn into an LKI list

feel free to test that

forge-gui/res/cardsfolder/upcoming/canyon_vaulter.txt
tool4ever: it's needed currently because without you could flicker it, AbilityUtils would lookup the new object and wrongly affect it

forge-gui/res/cardsfolder/a/adrestia.txt
tool4ever: ok works fine

tool4ever (approved):

PR #6908, opened 2025-01-27
[Adventure] further cleanup of the restricted cards

Simisays:
By adding MB2 to the restricted list all the non-convention cards become banned as well. This way fixes that. 

The restricted list also contained non existent sets which I have removed.

### DIFF
--- a/forge-gui/res/adventure/common/config.json
+++ b/forge-gui/res/adventure/common/config.json
@@ -84,7 +84,8 @@
     "PPC1",
     "UND",
     "PUST",
-    "DA1"
+    "DA1",
+    "UST"
   ],
   "difficulties": [
     {
--- a/forge-gui/res/adventure/common/config.json
+++ b/forge-gui/res/adventure/common/config.json
@@ -59,7 +59,15 @@
     "The Heron Moon",
     "Gobland",
     "Fetching Garden",
-    "Mox Poison"
+    "Mox Poison",
+    "Wisedrafter's Will",
+    "New Master of Arms",
+    "Halving Season",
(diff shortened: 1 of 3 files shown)


Agetian (approved):

forge-gui/res/adventure/common/config.json
```diff
@@ -67,14 +84,8 @@
     "PPC1",
     "UND",
     "PUST",
-    "UEPHI23",
-    "UEMIN23",
-    "UELAS23",
-    "UEIND23",
-    "UEBAR23",
-    "MB2",
-    "UNF",
```
Jetz72, line 76: Is Unfinity ready to come off the restricted list? Still a handful of funny cards mixed with the legal ones, and Attraction deck support for the adventure deck editor is still a work in progress.

Simisays: I woudn't mind it being off, since these are actually commander/eternal legal cards I was mainly waiting for the attraction/contraption support

Jetz72: Still working on the updates to the deck editor to support those, and hope to have it ready relatively soon. Just thought it might cause some confusion if the support cards start appearing before the Attractions/Contraptions themselves are available.

Simisays: Oh but I readded unstable already in a later edit, didn't want to remove it.  It's the last set on the list

tehdiplomat (approved):

Agetian:
Please resolve the merge conflicts so that we can merge the MR :)

PR #6919, opened 2025-01-29
Additional Discard Cost to RaiseCost

Hanmac:
part of #2472

- [ ] Check AI if it still understands that these have a discard cost. Probably need an update
- [ ] Need to update the ones that does `As an additional cost to cast this spell, discard X cards.`
- [ ] Maybe remove `AlternateAdditionalCost` if possible

### DIFF
--- a/forge-gui/res/cardsfolder/b/big_score.txt
+++ b/forge-gui/res/cardsfolder/b/big_score.txt
@@ -1,7 +1,8 @@
 Name:Big Score
 ManaCost:3 R
 Types:Instant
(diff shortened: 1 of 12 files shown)


Hanmac:
@tool4ever your opinion?

tool4ever:
Yea I think this will require some more refactor because unfortunately API related AI checks (e.g. `willPayCosts`) don't consider the adjusted cost this way :/

Agetian:
This will definitely need an AI-side update, I might be able to check it out this weekend, but not sure yet, since it's the report week at work :/

PR #6922, opened 2025-01-29
19 DFT cards (Mimeoplasm and its imprints)

Fulgur14:


forge-gui/res/cardsfolder/upcoming/mimeoplasm_revered_one.txt
```diff
@@ -0,0 +1,14 @@
+Name:Mimeoplasm, Revered One
+ManaCost:X B G U
+Types:Legendary Creature Ooze
+PT:0/0
+R:Event$ Moved | ValidCard$ Card.Self | Destination$ Battlefield | ReplaceWith$ DBChangeZone | ReplacementResult$ Updated | Description$ As NICKNAME enters, exile up to X creature cards from your graveyard. It enters with three +1/+1 counters on it for each creature card exiled this way.
+SVar:DBChangeZone:DB$ ChangeZone | Origin$ Graveyard | Destination$ Exile | ChangeType$ Creature.YouOwn | SelectPrompt$ Select up to X creature cards from your graveyard to exile | ChangeNum$ X | Hidden$ True | RememberChanged$ True | SubAbility$ DBPutCounter
```
tool4ever, line 6: try put ETB$ on this line too

forge-gui/res/cardsfolder/upcoming/radiant_lotus.txt
```diff
@@ -0,0 +1,8 @@
+Name:Radiant Lotus
+ManaCost:6
+Types:Artifact
+A:AB$ Mana | Cost$ T Sac<X/Artifact> | XMin$ 1 | ValidTgts$ Player | Produced$ Any | Amount$ X | RestrictValid$ CantCastNonArtifactSpells | SubAbility$ DBConvert | SpellDescription$ Choose a color. Target player adds three mana of the chosen color for each artifact sacrificed this way.
```
tool4ever, line 4: bad sub

tool4ever, line 4: color choice still needs to be done by activator (who might be different than target), this wouldn't currently?

tool4ever, line 4: no restriction

Fulgur14: I noticed that Spectral Searchlight specifically mentions who does choosing, so I assumed it's not needed here.

tool4ever: both are different
I don't see how it's supposed to work right now since your Defined$ would just overwrite the target

Fulgur14: Hm, is there a way to do it, or does it need new functionality? I don't think there are any cards where one player chooses the color and another one gets the mana...

tool4ever: just use ChooseColor effect too

Fulgur14: Like this?

tool4ever (approved):

PR #6957, opened 2025-02-01
refactor Max Speed round 1

### LINKED ISSUE #6942: MaxSpeed Static
> After the rules got released, we might look at Max Speed again.
> For example, this one:
> Max speed — {T}: Add {R} for each Goblin you control.
> 
> Is this just an Activated Ability with a restriction? And can it be used with Agatha's Soul Cauldron?
> or is it a Static Ability that grants an Activated Ability?
> 
> 🤔  

 _Originally posted by @Hanmac in [#6859](https://github.com/Card-Forge/forge/issues/6859#issuecomment-2624371717)_

Matt_Tabak on BlueSky said, while it is a static that grants this Ability, it seems more of a Characteristic Defining one?
Means it works in all zones. (Links will be added later)

@tool4ever that also means, for your Dependency Layer thing, it should apply to itself in Exile before Agatha's Soul Cauldron grants the Ability to your creatures.

Northmoc:
closes #6942 (INCLUDING tooltip problem referred to later in that issue)

### DIFF
--- a/forge-game/src/main/java/forge/game/event/GameEventSpeedUp.java
+++ b/forge-game/src/main/java/forge/game/event/GameEventSpeedUp.java
@@ -1,6 +1,14 @@
 package forge.game.event;
 
+import forge.game.player.Player;
+
 public class GameEventSpeedUp extends GameEvent {
+
+    public final Player player;
+
+    public GameEventSpeedUp(Player affected) {
+        player = affected;
+    }
 
     @Override
     public <T> T visit(IGameEventVisitor<T> visitor) {
--- a/forge-gui/res/cardsfolder/upcoming/aether_syphon.txt
+++ b/forge-gui/res/cardsfolder/upcoming/aether_syphon.txt
@@ -3,6 +3,7 @@ ManaCost:1 U U
 Types:Artifact
 K:Start your engines
 A:AB$ Draw | Cost$ 2 T | SpellDescription$ Draw a card.
-T:Mode$ Drawn | ValidCard$ Card.YouCtrl | TriggerZones$ Battlefield | MaxSpeed$ True | Execute$ TrigMill | TriggerDescription$ Max speed — Whenever you draw a card, each opponent mills two cards.
+S:Mode$ Continuous | CharacteristicDefining$ True | Affected$ Card.Self | Condition$ MaxSpeed | AddTrigger$ DrawTrig | Description$ Max speed — Whenever you draw a card, each opponent mills two cards. (Each opponent puts the top two cards of their library into their graveyard.)
+SVar:DrawTrig:Mode$ Drawn | ValidCard$ Card.YouCtrl | TriggerZones$ Battlefield | Execute$ TrigMill | Secondary$ True | TriggerDescription$ Max speed — Whenever you draw a card, each opponent mills two cards.
 SVar:TrigMill:DB$ Mill | Defined$ Opponent | NumCards$ 2
 Oracle:Start your engines! (If you have no speed, it starts at 1. It increases once on each of your turns when an opponent loses life. Max speed is 4.)\n{2}, {T}: Draw a card.\nMax speed — Whenever you draw a card, each opponent mills two cards. (Each opponent puts the top two cards of their library into their graveyard.)
--- a/forge-gui/res/cardsfolder/upcoming/amonkhet_raceway.txt
+++ b/forge-gui/res/cardsfolder/upcoming/amonkhet_raceway.txt
@@ -3,5 +3,6 @@ ManaCost:no cost
 Types:Land
 K:Start your engines
 A:AB$ Mana | Cost$ T | Produced$ C | SpellDescription$ Add {C}.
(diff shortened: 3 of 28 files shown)


forge-game/src/main/java/forge/game/event/GameEventSpeedUp.java
```diff
@@ -1,7 +1,15 @@
...
+import forge.game.player.Player;
+
 public class GameEventSpeedUp extends GameEvent {
 
+    public final Player player;
+
+    public GameEventSpeedUp(Player affected) {
+        player = affected;
+    }
+
```
Jetz72, line 12: You could generalize this to a ChangeSpeed event so that speed going up and down are handled in the same places. Would just need an extra delta value and a check in the sound handler to determine if the sound plays. Wouldn't alter the functionality but it'd be a little bit cleaner.

Northmoc: Sounds good – I'll do that as I refactor the whole thing into SBA

forge-gui/res/cardsfolder/upcoming/aether_syphon.txt
```diff
@@ -3,6 +3,7 @@ ManaCost:1 U U
 Types:Artifact
 K:Start your engines
 A:AB$ Draw | Cost$ 2 T | SpellDescription$ Draw a card.
-T:Mode$ Drawn | ValidCard$ Card.YouCtrl | TriggerZones$ Battlefield | MaxSpeed$ True | Execute$ TrigMill | TriggerDescription$ Max speed — Whenever you draw a card, each opponent mills two cards.
+S:Mode$ Continuous | CharacteristicDefining$ True | Affected$ Card.Self | Condition$ MaxSpeed | AddTrigger$ DrawTrig | Description$ Max speed — Whenever you draw a card, each opponent mills two cards. (Each opponent puts the top two cards of their library into their graveyard.)
```
tool4ever, line 6: @Hanmac
I think it needs `EffectZone$ + AffectedZone$ All` instead of ``CharacteristicDefining$`` :/
but could be hardcoded in the constructor when MaxSpeed is used?

forge-gui/res/cardsfolder/upcoming/amonkhet_raceway.txt
```diff
@@ -3,5 +3,6 @@ ManaCost:no cost
 Types:Land
 K:Start your engines
 A:AB$ Mana | Cost$ T | Produced$ C | SpellDescription$ Add {C}.
-A:AB$ Pump | PrecostDesc$ Max speed — | Cost$ T | Activation$ MaxSpeed | ValidTgts$ Creature | KW$ Haste | SpellDescription$ Target creature gains haste until end of turn.
+S:Mode$ Continuous | CharacteristicDefining$ True | Affected$ Card.Self | Condition$ MaxSpeed | AddAbility$ ABPump | Description$ Max speed — {T}: Target creature gains haste until end of turn.
+SVar:ABPump:AB$ Pump | Secondary$ True | PrecostDesc$ Max speed — | Cost$ T | ValidTgts$ Creature | KW$ Haste | SpellDescription$ Target creature gains haste until end of turn.
```
tool4ever, line 7: Secondary missing?

forge-gui/res/cardsfolder/upcoming/aether_syphon.txt
Jetz72: Could possibly implement it like Visit and Chapter abilities, where a pseudo-keyword points to the ability in question and then expands into the actual static ability.

Hanmac: MaxSpeed can only affect itself, so i have no problem to define it as CDA?

i don't really think i can have `Condition$ MaxSpeed` alter the other two Attributes ... it isn't like something i did with Exhaust :/

Hanmac: > Could possibly implement it like Visit and Chapter abilities, where a pseudo-keyword points to the ability in question and then expands into the actual static ability.

i 'm a bit sour with such pseudo-keywords ... i already fall on my nose about some of these hidden keywords i'm trying to remove
stuff like `etbCounter` causes way too many problems

tool4ever: I was mostly asking because of some extra rules for CDA
like the copy exception thing and dependencies...

tool4ever: being CDA makes it work for now until I can support that case properly I guess

tool4ever (approved):

forge-game/src/main/java/forge/game/event/GameEventSpeedUp.java
Hanmac: @tool4ever @Northmoc should we still change the Event into a SpeedChangedEvent? for SpeedDown too?

Hanmac (approved):
we go the GameEventSpeed change later

PR #6971, opened 2025-02-04
Fix Primal Wellspring trigger not working with Chun-Li

tool4ever:
Turns out in scenarios with frozen stack `registerOneTrigger` is pretty problematic and easily leads to the trigger getting lost.

Switching to a real delayed trigger avoids such "race condition".


Hanmac (approved):
i don't really like the hard coded `getSpawningAbility().hasParam("TriggersWhenSpent")`
but i think that is good enough for now

PR #6972, opened 2025-02-05
`instanceof` Pattern variable changes

neoFuzz:
Changes to files to use `instanceof` pattern variable. Since it was used in one place already, it sounds like it should work everywhere.

After testing the Android build and running it, I found that it works as intended on my device. I suggest that further tests be conducted by others to confirm its reliability.


tool4ever (approved):
seems ok
Android is weird like that, best to check if any problems with this get reported in Discord in the next few days imo - revert is always an option

PR #6982, opened 2025-02-07
Add Speed Tracker to Command Zone

Jetz72:
Makes the player's speed more prominent by tracking it with a Command zone effect. ~Similar to the Contraptions' "CRANK!" counter, this is done by adjusting the number of counters on the effect to match the player's speed.~ (Edit: Added a setter that can supply text for this directly.) It also has a flipped state when at maximum speed.

Still need the [tracker's images](https://scryfall.com/card/tdft/14/start-your-engines!-max-speed) to be added to the tokens directory so it can display properly. I've used "t:speed" and "t:max_speed" as image keys, but can change them as needed.


Hanmac:
I don't really like the adding of counter types which aren't real counter types :/

Jetz72:
Is the suggestion to just cut it from the CounterEnumType and leave it as a more open `CounterType`, or to move away from using counters for this sort of thing entirely?

I would someday like to expand the counter overlay text to also display other things in an easily visible way, but not yet sure what the scope of something like that would be. Arbitrary text? Controllable by card script at all? Context sensitive, something like the detail text hints idea that was being bounced around the other day?

Jetz72:
Did the thing discussed above. `card.setOverlayText` will now let you show any string the way a counter would. Updated the Crank counters from contraptions to use this feature too. The fake counters have been removed.

tool4ever:
Should `TrackableProperty.Speed` be removed too to avoid redundancy? 🤔

Jetz72:
Removed the parts that were already done by #6987.

> Should `TrackableProperty.Speed` be removed too to avoid redundancy? 🤔

Done.

tool4ever (approved):
LGTM

PR #6986, opened 2025-02-08
StaticAbilityContinous now have AffectedDefined for card list

### LINKED ISSUE #6931: StaticAbilityContinuous: add way to use DefinedCards/DefinedPlayer
For `StaticAbilityContinuous.getAffectedCards`, I would like a way to use getDefinedCards instead of a Valid Param.

This is for Effects that are created in CommandZone using `isRemembered`
If the Effect already knows what cards it is going to look for, it would be faster instead of iterating over all cards in the given Zone.

This might help us in the future when we are trying to refactor Pump and Animate Effects

Hanmac:
closes #6931 

Defined List should be faster than using a Valid over all cards

### DIFF
--- a/forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java
+++ b/forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java
@@ -1032,7 +1032,7 @@ public final class StaticAbilityContinuous {
 
         CardCollection definedCards = null;
         if (stAb.hasParam("AffectedDefined")) {
-            definedCards = AbilityUtils.getDefinedCards(hostCard, stAb.getParam("AffectedDefined"), stAb);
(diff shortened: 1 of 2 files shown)


forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java
```diff
@@ -1032,18 +1030,28 @@ public static CardCollectionView getAffectedCards(final StaticAbility stAb, fina
         // non - CharacteristicDefining
         CardCollection affectedCards = new CardCollection();
 
+        CardCollection definedCards = null;
+        if (stAb.hasParam("AffectedDefined")) {
+            definedCards = AbilityUtils.getDefinedCards(hostCard, stAb.getParam("AffectedDefined"), stAb);
```
tool4ever, line 1035: needs phased out check imo

tool4ever (approved):

PR #6989, opened 2025-02-09
Remove outdated logic

### LINKED ISSUE #4842: WrappedAbility Timestamp issue
Problem with Trigger and Timestamps:

https://github.com/Card-Forge/forge/blob/93a52a4263538910543ad222f20d056a2bcfb388/forge-game/src/main/java/forge/game/trigger/WrappedAbility.java#L503-L522

WrappedAbility does mess with the Triggered Objects,
but the better way would be to have the Effects handle it

currently the way is to update the Effects with `equalsGameTimestamp` to check if the timestamp is still OK or use LKI instead
for this, it might be better if the Triggered Objects where LKI to begin with so they can't be messed up again?

it would be better if all effects are updated, and the Part could be removed from WrappedAbility

tool4ever:
Closes #4842


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -625,31 +619,18 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
...
-        }
-
         if (zoneFrom == null) {
             return copied;
         }
 
-        if (!c.isRealToken() && !toBattlefield) {
-            copied.clearDevoured();
-            copied.clearDelved();
-            copied.clearExploited();
```
Hanmac, line 639: would it make sense to remove this from the Replacement Prevented Part too?

Hanmac: clearDevoured and clearExploited aren't used anymore and can be removed?

Hanmac (approved):

PR #6992, opened 2025-02-09
Bestow as Static Effect

### LINKED ISSUE #6811: animate Bestow as Static Ability
Part of ##5642 

Bestow Effect should use StaticAbility Layers if able

For Layer4, instead of addChangedCardTypes, we need:
* Add Aura Type + Remove CreatureType + Remove Other Enchantment Types

it needs to effect the Spell before it is put on the Stack, and then Later on the Battlefield

Hanmac:
Closes #6811 

currently it causes flickering

and sometimes, the Bestow Effect isn't removed as it should, but i can't find the remaining one
and sometimes, the effect might be removed, but the static effect of the effect isn't cleared

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -250,7 +250,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
     private int exertThisTurn = 0;
     private PlayerCollection exertedByPlayer = new PlayerCollection();
 
-    private long bestowTimestamp = -1;
+    private Card bestowEffect = null;
     private long transformedTimestamp = 0;
     private long prototypeTimestamp = -1;
     private long mutatedTimestamp = -1;
@@ -3741,7 +3741,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
     public final void addLeavesPlayCommand(final GameCommand c) {
         leavePlayCommandList.add(c);
     }
- 
+
     public void addStaticCommandList(Object[] objects) {
         staticCommandList.add(objects);
     }
@@ -4757,7 +4757,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
     public void addDraftAction(String s) {
         draftActions.add(s);
     }
- 
+
     private int intensity = 0;
     public final void addIntensity(final int n) {
         intensity += n;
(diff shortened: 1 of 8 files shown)


forge-gui/src/main/java/forge/player/HumanPlay.java
```diff
@@ -104,6 +104,8 @@ public final static boolean playSpellAbility(final PlayerControllerHuman control
                     if (rollback.isInZone(ZoneType.Exile)) {
                         rollback.addMayLookTemp(p);
                     }
+                } else {
```
tool4ever, line 107: else is wrong
just move it into the block above

Hanmac: in theory `undoAlternateHost` could handle reverting `castFaceDown` too

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -6827,43 +6827,40 @@ public boolean isClassCard() {
...
+        bestowEffect = SpellAbilityEffect.createEffect(null, this, this.getController(), "Bestow Effect", getImageKey(), getGame().getNextTimestamp());
+        bestowEffect.setRenderForUI(false);
+        bestowEffect.addRemembered(this);
+
+        String s = "Mode$ Continuous | AffectedDefined$ RememberedCard | EffectZone$ Command "
+                + " | AddType$ Aura | RemoveType$ Creature | RemoveEnchantmentTypes$ True | AddKeyword$ Enchant creature";
+        bestowEffect.addStaticAbility(s);
+
+        GameCommand until = SpellAbilityEffect.exileEffectCommand(getGame(), bestowEffect);
+        addLeavesPlayCommand(until);
```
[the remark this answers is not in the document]
Hanmac: Leaving the stack without resolving should already call `unanimateBestow` directly

Hanmac:
@tool4ever i couldn't pinpoint the cause of the issues yet (both visual and functional)

but my gut is telling me that the Remembered List of the LKI could be a cause.
Right now when the Effect are copied, the remembered still uses the original objects?

https://github.com/Card-Forge/forge/blob/a4b27321ac4c4a4d8377bbf5ee7bdd88581a3056/forge-game/src/main/java/forge/game/card/CardCopyService.java#L342-L344

forge-game/src/main/java/forge/game/card/Card.java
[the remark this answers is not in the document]
tool4ever: it seems CounterEffect would call it
but ChangeZoneEffect not currently

Hanmac:
> Probably need to manually add the effect to Command instead so you have greater control over when statics are applied 🤔

I think the main problem are that Bestow uses LKI -> StaticEffect to check.

And then the LKI object isn't used anymore, but the StaticEffect might still linger ...
I tried debugging it, but i haven't found all cases where the BestowEffect should be removed again after checks.

Especially the ones where you Cancel the Spell cast

Hanmac:
@tool4ever i think this one is the next to revive, thanks to hidden Static.

It should work without causing too many problems with LKI

Hanmac:
i need some more AI tests to check if i need more `undoAlternateHost`

tool4ever:
I'm having trouble testing AI, need to find out if it's broken on master too :/

tool4ever:
This is still done twice here:
https://github.com/Card-Forge/forge/blob/cdb75c48edeeae435868c3ecdef31f4399f546cf/forge-game/src/main/java/forge/game/cost/CostAdjustment.java#L51-L54

I was hoping to remove such special case because it would be better to trust the altHost

but it seems it would still be reached at least from here:
https://github.com/Card-Forge/forge/blob/cdb75c48edeeae435868c3ecdef31f4399f546cf/forge-game/src/main/java/forge/game/spellability/Spell.java#L110

ideally the connection SA <> altHost is so strong it only gets created once 🤔

(and then later `setSplitStateToPlayAbility` for the real card)

forge-game/src/main/java/forge/game/spellability/Spell.java
```diff
@@ -170,6 +174,7 @@ public Card getAlternateHost(Card source) {
             }
 
             source.animateBestow();
+            source.updateKeywords();
```
tool4ever, line 177: this doesn't really do anything right now

forge-game/src/main/java/forge/game/spellability/Spell.java
```diff
@@ -100,15 +100,19 @@ public Card canPlayFromHost() {
...
-        if (!activator.getController().isFullControl(FullControlFlag.AllowPaymentStartWithMissingResources) &&
-                !CostPayment.canPayAdditionalCosts(this.getPayCosts(), this, false)) {
-            return null;
+            if (!activator.getController().isFullControl(FullControlFlag.AllowPaymentStartWithMissingResources) &&
+                    !CostPayment.canPayAdditionalCosts(this.getPayCosts(), this, false)) {
+                return null;
+            }
+        } finally {
+            undoAlternateHost(card);
         }
```
Hanmac, line 116: @tool4ever i think the problem right now for the AI is this part, that tries to "undo" the changes that Alt Host did

now i'm not 100 % sure if that is the right idea

PR #7005, opened 2025-02-12
Oracle update: During your turn

dracontes:
A while ago I noticed the text on [At Knifepoint](https://scryfall.com/card/otj/193/at-knifepoint) had changed with "During your turn" replacing the original "As long as it's your turn". More recent work had me come across a similar change for [Angry Mob](https://scryfall.com/card/me1/1/angry-mob). Thus, while I was at it regarding recent cleanups, I did as well for this item, with reference to Scryfall and Gatherer.


tool4ever (approved):
thanks

Hanmac (approved):

PR #7009, opened 2025-02-13
Misc cleanup

tool4ever:


forge-gui/src/main/java/forge/gui/control/FControlGameEventHandler.java
```diff
@@ -451,12 +451,6 @@ public Void visit(final GameEventPlayerStatsChanged event) {
         return processCards(cards, cardsRefreshDetails);
     }
 
-    public Void visit(final GameEventSpeedChanged event) {
-        Player p = event.player;
-        processPlayer(p, livesUpdate);
-        return processEvent();
-    }
```
Hanmac, line 458: @tool4ever wouldn't that stop the GUI from updating?

forge-game/src/main/java/forge/game/ability/effects/ChangeZoneAllEffect.java
```diff
@@ -107,10 +107,8 @@ public void resolve(SpellAbility sa) {
             final Zone originZone = game.getZoneOf(c);
 
             // Fizzle spells so that they are removed from stack (e.g. Summary Dismissal)
-            if (sa.hasParam("Fizzle")) {
```
Hanmac, line 110: @tool4ever can we look to remove other Fizzle Params from the ChangeZone Effects?

forge-gui/src/main/java/forge/gui/control/FControlGameEventHandler.java
tool4ever: Because #6982 changed it to Command Zone display updating Player stats no longer serves a purpose

forge-game/src/main/java/forge/game/ability/effects/ChangeZoneAllEffect.java
tool4ever: should be fixed

Hanmac (approved):

Hanmac:
@tool4ever another thing that could be cleanup is the functions that handle "removed from stack"
There are two:
* `ChangeZoneEffect.removeFromStack`
* `CounterEffect.removeFromStack`

Maybe some of the Common Code could be moved to Generic SpellAbilityEffect, or to MagicStack?

PR #7013, opened 2025-02-13
Dependency tab

tool4ever:
I also evaluated graph drawing options but it seemed pretty complex to obtain nice looking results so copyable text felt like the better approach in the end.

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -707,7 +707,7 @@ public class GameAction {
             // needed for ETB lookahead like Bronzehide Lion
             stAb.putParam("AffectedZone", "All");
             SpellAbilityEffect.addForgetOnMovedTrigger(eff, "Battlefield");
-            game.getAction().moveToCommand(eff, cause);
+            eff.getOwner().getZone(ZoneType.Command).add(eff);
         }
 
         eff.addRemembered(copied);
@@ -723,7 +723,6 @@ public class GameAction {
             }
         }
     }
-
 
     private void storeChangesZoneAll(Card c, Zone zoneFrom, Zone zoneTo, Map<AbilityKey, Object> params) {
         if (params != null && params.containsKey(AbilityKey.InternalTriggerTable)) {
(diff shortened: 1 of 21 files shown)


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -707,7 +707,7 @@ private Card setupStaticEffect(Card copied, SpellAbility cause) {
             // needed for ETB lookahead like Bronzehide Lion
             stAb.putParam("AffectedZone", "All");
             SpellAbilityEffect.addForgetOnMovedTrigger(eff, "Battlefield");
-            game.getAction().moveToCommand(eff, cause);
+            eff.getOwner().getZone(ZoneType.Command).add(eff);
```
Hanmac, line 710: why not use the moveToCommand one?

tool4ever: performance
it would lead to here:
https://github.com/Card-Forge/forge/blob/11913085efe87ad7970a2de305fe288e8e61130c/forge-game/src/main/java/forge/game/GameAction.java#L106-L112

and I didn't like that during resolve

forge-game/src/main/java/forge/game/GameView.java
```diff
@@ -200,15 +205,29 @@ void updateGameLog(GameLog gameLog) {
...
+        return get(TrackableProperty.Dependencies);
+    }
+    public void setDependencies(Table<StaticAbility, StaticAbility, StaticAbilityLayer> dependencies) {
+        StringBuilder sb = new StringBuilder();
+        StaticAbilityLayer layer = null;
+        for (Cell<StaticAbility, StaticAbility, StaticAbilityLayer> dep : dependencies.cellSet()) {
+            if (layer != dep.getValue()) {
+                layer = dep.getValue();
+                sb.append("Layer " + (StaticAbilityLayer.CONTINUOUS_LAYERS.indexOf(layer) + 1)).append(": ");
+            }
```
Hanmac, line 222: I have questions about this table:
- is it that only two StaticAbility have only a dependency in 1 layer? Or could they be in multiple?
- about the layer variable, you want to use it as prefix, in the description, assuming the table is somewhat ordered by their value, instead of their keys?
- the number doesn't work for the sub layers, like 7c. Would it make sense to add the number to `StaticAbilityLayer` ?

tool4ever: 1. I think in theory it might be possible to produce, but would probably involve some pretty specific copy steps?
But if you mean I should rather store them as Set to be safe I agree!
2. yea because that's the order layers get processed it made sense to display it the same way
3. ah good point, guess I'll add them there

Hanmac (approved):

PR #7015, opened 2025-02-14
Bundle of Assorted Fixes

### LINKED ISSUE #3485: Loose control of keyboard
'm running the Linux version, I'm using Pop!_OS (Ubuntu/Debian distro). My Java Version is openjdk 11.0.19 2023-04-18. Forge version is 1.5.56.

Whenever I use Forge, I can't use my backspace key, enter key, delete key, most of the letters on my keyboard. I have to log out and log back in to get control.

### LINKED ISSUE #4678: Issue with Forge causing undesirable/destabilizing effects on linux distros
**Describe the bug**
Just saw this being posted (again) on the Forge discord and one of the devs saying they've never seen anyone complain about this issue before, which is probably true, but this issue is definitely a real thing.

Screenshot_20240212_133221

Here's one of the bug reports that I remember: https://github.com/Card-Forge/forge/issues/3485

And for ease of access here's the original report I made long ago: https://www.slightlymagic.net/forum/viewtopic.php?f=26&t=31106&p=251894&hilit=manjaro#p251894

From what I can tell it only seems to happen if you use the content downloader in Forge, 'cause I haven't touched that thing since I figured that bit out and I have had zero issues like this crop up again.

### LINKED ISSUE #5941: Import Parses Tyrannax Rex Incorrectly
**Describe the bug**
When using the Import Feature, the importer interprets the "Rex" in "Tyrannax Rex" as a set, instead of part of its name. Adding the card via Card Catalog works normally.

**To Reproduce**
1. Go to Deck Editor
2. Click on Import
3. Input "1 Tyrannax Rex" as a card to import
4. The importer displays "Tyrannax [Rex]" and of course doesn't fetch the card.

**Expected behavior**
The importer should recognize Tyrannax Rex as the _name_ of the card, not interpret "Rex" as a set.

**Screenshots**

 - OS: Windows
 - Version 1.6.64

**Additional context**
I'm only a amateur programmer myself and a bit of a novice, but I suspect the fact that "Rex" has three letters might be what's causing the parser to trip and think it's a set code?

Jetz72:
Small handful of minor fixes and changes.

* Fix importing when the last few letters of the card name are also a set code. Closes #5941.
* Add text formatting support for "[Keyword]cycling" and "Affinity for [Keyword]". Not really perfect since the reminder text for the former still comes out as "for a [keyword] card" instead of "for a card with [keyword]", but the text formatting there hooks into several other keywords so this seems acceptable for now. Unblocks #7003.
* "Number of attractions you visited this turn" now counts attractions that have left the battlefield or have stopped being Attractions, per rule 608.2i. Doesn't close any issues because it's a niche rule in an edge case scenario involving multiple cards from a pool that are almost never used... but it's been bugging me so here it is. 
* Remove the wake lock code for desktop. Java does not offer any proper API for this functionality, so it worked by simulating a key press of "F15" every 30 seconds or so. It was written with the assumption that the F15 key would typically never have any effect, but that assumption has not held up. We could replace it with a different solution that simulates a mouse jitter, but I'd like to try removing it completely first. Desktop OSes like Windows are a bit less strict about sleep conditions than Android, and it's possible the network traffic will be enough to keep the CPU awake. If anyone objects, or if we encounter any issues caused by this in the future, I don't mind implementing an alternative. Closes #4678 and properly closes #3485.

### DIFF
--- a/forge-core/src/main/java/forge/deck/DeckRecognizer.java
+++ b/forge-core/src/main/java/forge/deck/DeckRecognizer.java
@@ -472,7 +472,8 @@ public class DeckRecognizer {
             "side", "sideboard", "sb",
             "main", "card", "mainboard",
             "avatar", "commander", "schemes",
-            "conspiracy", "planes", "deck", "dungeon"};
+            "conspiracy", "planes", "deck", "dungeon",
+            "attractions", "contraptions"};
 
     private static CharSequence[] allCardTypes(){
         List<String> cardTypesList = new ArrayList<>();
@@ -670,6 +671,9 @@ public class DeckRecognizer {
                     // ok so the card has been found - let's see if there's any restriction on the set
                     return checkAndSetCardToken(pc, edition, cardCount, deckSecFromCardLine,
                                                 currentDeckSection, true);
+                // On the off chance we accidentally interpreted part of the card's name as a set code, e.g. "Tyrranax Rex"
+                if (data.isMTGCard(cardName + " " + setCode))
+                    continue;
                 // UNKNOWN card as in the Counterspell|FEM case
                 return Token.UnknownCard(cardName, setCode, cardCount);
             }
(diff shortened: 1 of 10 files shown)


tool4ever (approved):
I agree with your reasoning :)

PR #7018, opened 2025-02-15
checkStaticAbilities: skip wrong zone earlier for less looping

tool4ever:


Hanmac (approved):

Hanmac:
@tool4ever we might look into isSuppressed a bit more, i think after some changes, nothing outside of Commander Replacement for Meld/Mutated does call that anymore?

PR #7022, opened 2025-02-16
Effects don't need Exiled trigger

tool4ever:


Hanmac (approved):

Hanmac:
@tool4ever i did an early return instead

PR #7061, opened 2025-02-22
Replace net deck links

paulsnoops:


tehdiplomat (approved):

tool4ever:
doesn't seem ideal we hardcode the full URL there but oh well

tehdiplomat:
Ya it was already like that. I think it would be better to just be relative URLs in the long term and in the code we just pointed it to the base directory and add the relative path after.

PR #7070, opened 2025-02-23
Don't update view of LKI instead of real Card

### LINKED ISSUE #7067: Manifest dread in combination with adventure card and/or */* has incorrect power/thoughness
**Describe the bug**
I used manifest dread ability to put Beanstalk Giant card into play. It should be a manifest token with 2/2, but instead, it was 11/11 (because I had 11 lands, and Beanstalk Giant is \*/\*, where * is number of my lands)

I'm not sure if this is some interaction between adventure or any \*/\* would cause this when interacting with manifest dread.

**To Reproduce**
1. Select Miracle Workers Duskmourn commander deck
2. Cast your commander Zimone, 
3. somehow set up that the top card in your deck is Beanstalk Giant
4. Drop a land to cause manifest dread, and select Beanstalk Giant from the top of your deck to manifest him
5. He will now be a number of your lands / number of your lands manifest token instead of being 2/2 manifest token

**Expected behavior**
All manifest tokens should be fixed 2/2 (unless some special condition powers them up), and they shouldn't have the ability of their bottom down card.

Here's a picture of my board, the second manifest token should be 2/2, but it is 11/11. The bottom right card (Breanstalk giant) is the face down card from that manifest token.

**Screenshots**

Image

 - Device: PC
 - OS: Win10
 - Version 2.0.2

tool4ever:
Closes #7067

### DIFF
--- a/forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
+++ b/forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
@@ -66,7 +66,6 @@ public class ReplacementHandler {
 
     public List<ReplacementEffect> getReplacementList(final ReplacementType event, final Map<AbilityKey, Object> runParams, final ReplacementLayer layer) {
         final CardCollection preList = new CardCollection();
-        boolean checkAgain = false;
         Card affectedLKI = null;
         Card affectedCard = null;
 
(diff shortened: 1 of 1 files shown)


Agetian:
Looks good to me, not sure if I may be missing something :)

Hanmac (approved):

PR #7074, opened 2025-02-24
Delegate rendering of cards in deck viewer to piles all the time

marthinwurer:
See the thread "Mobile deck editor doesn't show cards unless you scroll" on the discord server

### DIFF
--- a/forge-gui-mobile/src/forge/itemmanager/views/ImageView.java
+++ b/forge-gui-mobile/src/forge/itemmanager/views/ImageView.java
@@ -869,33 +869,18 @@ public class ImageView<T extends InventoryItem> extends ItemView<T> {
                 if (isCollapsed) {
                     return;
                 }
-
-                float visibleLeft = getScrollLeft();
-                float visibleRight = visibleLeft + getWidth();
-                for (Pile pile : piles) {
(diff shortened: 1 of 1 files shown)


forge-gui-mobile/src/forge/itemmanager/views/ImageView.java
```diff
@@ -869,33 +869,18 @@ public void draw(Graphics g) {
...
 
-            final float visibleTop = getScrollValue();
-            final float visibleBottom = visibleTop + getScroller().getHeight();
-            for (ItemInfo itemInfo : items) {
-                if (itemInfo == null)
-                    continue;
-                if (itemInfo.getBottom() < visibleTop) {
+            float visibleLeft = getScrollLeft();
+            float visibleRight = visibleLeft + getWidth();
+            for (Pile pile : piles) {
```
tehdiplomat, line 876: Do these piles exist if groupBy is null?

marthinwurer: Yes, they do. I tested it with combinations of groups and piles and they all work in manual testing.

tehdiplomat (approved):

Issue #2773, opened 2023-03-29
MOM: Battle Mechanic


paulsnoops:

tool4ever:
> Any Target: A spell or ability may require "any target." "Any target" is the same as "target creature, player, or planeswalker."

We could just hardcode that in `AbilityFactory.readTarget` so `ValidTgts$ Any` works?

Hanmac:
> > Any Target: A spell or ability may require "any target." "Any target" is the same as "target creature, player, or planeswalker."
> 
> We could just hardcode that in `AbilityFactory.readTarget` so `ValidTgts$ Any` works?

imo i would prefer to AnyDamageTarget ? or something like that?

i don't know if we should do it in readTarget or in the isValid part

tehdiplomat:
That was just an example. Damageable would be the Property. So it would be `GameObject.Damageable` in this case; then Card and Player defining Damageable on their own.

Hanmac:
we should probably move this discussion to a Sub Ticket what we should use to simplify the damage for "any target" stuff

tehdiplomat:
https://github.com/Card-Forge/forge/issues/2833

Hanmac:
@tehdiplomat can this Issue be closed or do we need it for more AI stuff?

Agetian:
Needs additional AI support + the ability to actually defend the battles for both the human and the AI (#3107 attempts to address this).

tool4ever:
There are also still the two missing SBA?

Hanmac:
> There are also still the two missing SBA?

Yes, but i need to think on how to code them:
* one of them needs to be able to choose new player as protector, but I don't know how to get it working for AI
* one of them is for Battle loosing their Siege Subtype, bit I don't know that can happen with current cards

PR #7075, opened 2025-02-24
Fix battles attacking/blocking

tool4ever:
```
Game-0 > java.lang.NullPointerException: Cannot invoke "forge.game.player.Player.isOpponentOf(forge.game.player.Player)" because the return value of "forge.game.card.Card.getProtectingPlayer()" is null
	at forge.game.combat.CombatUtil.getAllPossibleDefenders(CombatUtil.java:73)
	at forge.game.combat.Combat.initConstraints(Combat.java:124)
	at forge.game.combat.Combat.<init>(Combat.java:79)
	at forge.game.phase.PhaseHandler.onPhaseBegin(PhaseHandler.java:295)
```


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -1659,11 +1659,23 @@ private boolean stateBasedAction_Saga(Card c, CardCollection sacrificeList) {
...
+        if (((c.getProtectingPlayer() == null || !c.getProtectingPlayer().isInGame()) &&
+                (game.getCombat() == null || game.getCombat().getAttackersOf(c).isEmpty())) ||
+                (c.getType().hasStringType("Siege") && c.getController().equals(c.getProtectingPlayer()))) {
+            Player newProtector = c.getController().getController().chooseSingleEntityForEffect(c.getController().getOpponents(), null, "Choose an opponent to protect this battle", null);
+            // seems unlikely unless range of influence gets implemented
+            if (newProtector == null) {
+                removeList.add(c);
+            } else {
+                c.setProtectingPlayer(newProtector);
+            }
```
Hanmac, line 1674: Yeah, i wanted to implement that, but i always forgot

Hanmac:
https://github.com/Card-Forge/forge/issues/2773#issuecomment-1547257297
the SBA are referenced there

> 704.5x If a Siege’s controller is also its designated protector, that player chooses an opponent to become its protector. If no player can be chosen this way, the battle is put into its owner’s graveyard. See rule 310, “Battles.”


There is also this part:
> 310.8a As a battle enters the battlefield, its controller chooses a player to be its protector. Which players may be chosen as its protector are determined by its battle type (see rule 310.11). **If it has no battle types, its controller becomes its protector.**

but I don't think that case is possible right now

Hanmac (approved):

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -1659,11 +1659,23 @@ private boolean stateBasedAction_Saga(Card c, CardCollection sacrificeList) {
 
     private boolean stateBasedAction_Battle(Card c, CardCollection removeList) {
         boolean checkAgain = false;
-        if (!c.getType().isBattle()) {
-            return false;
+        if (!c.isBattle()) {
+            return checkAgain;
+        }
+        if (((c.getProtectingPlayer() == null || !c.getProtectingPlayer().isInGame()) &&
+                (game.getCombat() == null || game.getCombat().getAttackersOf(c).isEmpty())) ||
```
Hanmac, line 1666: @tool4ever why the extra Combat check there?

tool4ever: > 704.5w. If a battle has no player in the game designated as its protector and **no attacking creatures are currently attacking that battle**, that battle's controller chooses an appropriate player to be its protector

PR #7093, opened 2025-02-27
Hidden and Double Agenda better as Keyword

Hanmac:
this is for reducing the use of `hasKeyword("string")`

i couldn't test it really, because my draft isn't working as i want


tool4ever:
looks ok I guess?

tool4ever:
@Hanmac
I think for testing you can also manually add them in deck editor?

Hanmac:
@tool4ever yeah i tried with draft but that didn't work
i didn't know if could add them via deck editor. (or somehow cheat them?)

can you test them until can check on them later?

Jetz72:
I believe you can add them to a constructed deck in the mobile editor at least, but no variant besides draft will actually utilize them in game. You could probably test it by opening up `RegisteredPlayer.java`, go down to the `restoreDeck()` method and add `assignConspiracies();` to the bottom of it.

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -239,14 +239,14 @@ public static boolean handleHiddenAgenda(Player player, Card card) {
 
         card.turnFaceDown();
         card.addMayLookAt(player.getGame().getNextTimestamp(), ImmutableList.of(player));
-        card.addSpellAbility(abilityRevealHiddenAgenda(card));
+        ki.addSpellAbility(abilityRevealHiddenAgenda(card));
```
tool4ever, line 242: this somehow breaks it

Hanmac: shouldn't the Keyword still add the SpellAbility to the Card?

Or do I need Extra check for TurnUp somewhere else?

Hanmac: should be fixed, i used Restriction instead of Condition because i somehow could turned it back face-down when i shouldn't

PR #7098, opened 2025-02-28
Make canBlock check if the blocker is a creature.

Agetian:
TokenAi also checks if the spawned token is a creature before running checks.
Fixes e.g. the AI using Fountainport to produce a Treasure token as a potential blocker.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/TokenAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/TokenAi.java
@@ -202,11 +202,11 @@ public class TokenAi extends SpellAbilityAi {
 
         if (sa.isPwAbility() && alwaysFromPW) {
             return true;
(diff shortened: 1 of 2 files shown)


forge-ai/src/main/java/forge/ai/ability/TokenAi.java
```diff
@@ -202,11 +202,11 @@ protected boolean checkApiLogic(final Player ai, final SpellAbility sa) {
 
         if (sa.isPwAbility() && alwaysFromPW) {
             return true;
-        } else if (game.getPhaseHandler().is(PhaseType.COMBAT_DECLARE_ATTACKERS)
+        } else if (game.getCombat() != null
```
tool4ever, line 205: why this change?

Agetian: Ah, that was a remainder from an experiment, good catch, thanks! :)

Hanmac (approved):

PR #7108, opened 2025-03-02
StaticAbility InfectDamage for Phyrexian Unlife

Hanmac:
Changes _Phyrexian Unlife_ into a Static Ability

or should this new class be merged with StaticAbilityWitherDamage? (WitherDamage uses Source, this uses Target)


tool4ever (approved):
Even when combined into one class technically it should still be two modes?

Also I wonder if some AI checks should be updated to consider it or if the overhead without KW isn't worth it...

Hanmac:
Yes still two modes

I don't know if AI needs more update, but we can do that later

PR #7111, opened 2025-03-02
Update MagicFest 2025.txt

tehdiplomat:


tehdiplomat:
@paulsnoops It looks like this edition has a bunch more cards on Scryfall now too? Not sure if you have this in your plan.

paulsnoops:
> @paulsnoops It looks like this edition has a bunch more cards on Scryfall now too? Not sure if you have this in your plan.

Yeah I was going to tomorrow, but just pushed an update here :)

paulsnoops (approved):

PR #7024, opened 2025-02-16
DA1: Updating Phila, Introducing Delphia

dracontes:
Couldn't help feel that our implementation of [Phila, Unsealed](https://scryfall.com/card/da1/MA30/phila-unsealed) could hew a bit closer to the card as written, while allowing access to the possible versions of the card (like the Arena buffs/debuffs). So I iterated the script structure until arriving at a hopefully reasonable way of doing so, with support from this [fix to functional variants](https://github.com/Card-Forge/forge/pull/7019). While I was at it, I did [Delphia, Undecided](https://scryfall.com/card/da1/MZ02/delphia-undecided) as well, since it's a conceptually similar card. Except for some issues with card display outside of games, shared with other functional variant scripts, these work as expected.

PR #7114, opened 2025-03-03
Sanguine Brushstroke: disambiguating between Arena and Mystery Booster versions

dracontes:
- [**Arena**](https://scryfall.com/card/ymid/32/sanguine-brushstroke): Script as currently exists in Forge moved to the rebalanced folder and given the requisite name prefix as that's what happened with it according to the MTG Wiki [Rebalanced card](https://mtg.wiki/page/Rebalanced_card) article.
- [**Mystery Booster**](https://scryfall.com/card/mb2/260/sanguine-brushstroke): Adapted the existing script, reinstating the life gain on the sacrifice triggered ability, and per the stated intent of the card's rulings, the enters triggered ability now conjures the paper version of Blood Artist, as the Arena one isn't listed on Gatherer.

### DIFF
--- a/forge-gui/res/cardsfolder/rebalanced/a-sanguine_brushstroke.txt
+++ b/forge-gui/res/cardsfolder/rebalanced/a-sanguine_brushstroke.txt
@@ -1,0 +1,10 @@
+Name:A-Sanguine Brushstroke
+ManaCost:1 B B
+Types:Enchantment
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigToken | TriggerDescription$ When CARDNAME enters, create a Blood token and conjure a card named Blood Artist onto the battlefield.
+SVar:TrigToken:DB$ Token | TokenScript$ c_a_blood_draw | SubAbility$ DBMakeCard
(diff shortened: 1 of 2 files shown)


tehdiplomat:
@dracontes Was this ready to go?

dracontes:
> @dracontes Was this ready to go?

Had a discussion with @paulsnoops and it turns out this would break archival decks as is. 
An option would be to have the Mystery Booster version as the one with the prefixed name. However, my work on [Phila and Delphia](https://github.com/Card-Forge/forge/pull/7024) came up, I recalled a discussion had months ago about [Earthquake Dragon](https://scryfall.com/card/clb/228/earthquake-dragon) and its [rebalanced version for Arena](https://scryfall.com/card/hbg/A-209/a-earthquake-dragon), and we agreed it's better to expand @Jetz72 's variant system to have more flexible, comprehensive way to address name overlaps and variations resulting from rebalancing, etc.
Probably should have mentioned that sooner but it seemed like there was enough work to contend with already.

paulsnoops:
It's also important to note that A-Sanguine Brushstroke doesn't exist so I'd rather we didn't name it this way. This card being Arena/Alchemy only doesn't have a separate rebalanced version. Could we close this for now until a better solution is found?

tehdiplomat:
I'm gonna mark this as closed for now. We can reopen if an official decision is made.

PR #7135, opened 2025-03-09
gui-desktop: sort preferences

Takzzg:
Sort user preferences for gui-desktop.

before: https://postimg.cc/gallery/GDC0JXr
after: https://postimg.cc/gallery/XqgXQRS

Some preferences remain uncategorized, suggestions appreciated

Note-worthy changes:
- VSubmenuPreferences.java: 
  - divide cctor into different functions
  - new SectionDivider class (line 645)
 - new translation keys (en-US & es-ES)


tehdiplomat:
I would say Deck Conformance is Rules category. 

Maybe make some of the categories more generic "Draft" vs "Draft Ranking Options"

Also, I would prefer if the troubleshooting options are at the top rather than the bottom.

Issue #4760, opened 2024-02-29
Player Hidden Keywords Meta Issue


Hanmac:
Player Hidden Keywords that needs to be rewritten into StaticAbility or ReplacementEffects CantHappen

- "You may pay 0 rather than pay the echo cost for permanents you control."
 
- "You draw a card during each opponent's draw step."
 
- "Creatures your opponents control attack a player other than you if able."
 
- "You assign combat damage of each creature attacking you."
 
- "Foretell on any player's turn"

 
- "All damage is dealt to you as though its source had infect."
- "Damage doesn't cause you to lose life."
 
- Search Library
  - "LimitSearchLibrary"
  - "CantSearchLibrary"
  - "Spells and abilities you control can't cause you to search your library."

- Cant Win or Lose
  - "You can't lose the game."
  - "You can't win the game."
  - "You don't lose the game for having 0 or less life."
  - "You don't lose the game for drawing from an empty library."
 
- "You draw cards from the bottom of your library instead of the top of your library."
 
- "Creatures you control can boast twice during each of your turns rather than once."
 
- "Piracy"
- "CyclingForZero"
- "You may pay 0 rather than pay equip costs."
 
- "Shaman's Trance"

- "PayLifeInsteadOf:B"

tool4ever:
Currently I'm liking CantHappen more because it gives us easier support for GameLog already 🤔

Hanmac:
> Currently I'm liking CantHappen more because it gives us easier support for GameLog already 🤔

I prefer to use CantHappen when there is already an event to Replace 
and where it makes sense


i mean,  "Foretell on any player's turn", can't really be a "CantHappen" event, right? xD

tool4ever:
no, but for those extra GameLog feedback isn't really useful anyway

Hanmac:
@tool4ever for example, the Search Library stuff might use Replacement Effect and CantHappen for 2 of them

Hanmac:
"As you cascade, you may put a land card from among the exiled cards onto the battlefield tapped."
should probably be part of the DigUntil Effect? Should that be done as Replacement Effect?

"Your devotion to each color and each combination of colors is increased by one."
This one is a bit tricky because it needs to be done in between Layer 3 and 4 (so at the end of layer 3?)

Hanmac:
- "Your devotion to each color and each combination of colors is increased by one."

This one is a bit tricky because it needs to be done in between Layer 3 and 4 (so at the end of layer 3?)

PR #7143, opened 2025-03-11
Use StaticAbility for Topsy Turvy

Hanmac:
part of #4760 

also reduce the amount of `getAmountOfKeyword(string)`


tool4ever (approved):
I still wish UN cards wouldn't lead to more engine checks as long as they don't exist in a game

Hanmac:
> I still wish UN cards wouldn't lead to more engine checks as long as they don't exist in a game

It's probably better to use small checks than to update them with HiddenKeywords

PR #7146, opened 2025-03-12
Filter out targets for each player based on previous targets

tehdiplomat:


forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
```diff
@@ -914,6 +914,12 @@ private static boolean isPreferredTarget(final Player ai, final SpellAbility sa,
         if (sa.isSpell()) {
             list.remove(source); // spells can't target their own source, because it's actually in the stack zone
         }
+
+        // Filter out cards TargetsForEachPlayer
+        if (sa.hasParam("TargetsForEachPlayer")) {
```
tool4ever, line 919: I'd probably just filter by `sa.canTarget` again
then you'll fix it for all relevant dependency-like restrictions at once?
e. g. TargetsWithDifferentCMC$

tehdiplomat: Meh.. I think I'd rather have one function that does these relative comparisons than needing to waste time looping around canTarget for things that are already filtered out.

tehdiplomat: Alright I did the thing.

PR #7150, opened 2025-03-13
Fix Griselbrand overdrawing in one turn

tehdiplomat:


forge-ai/src/main/java/forge/ai/ability/DrawAi.java
```diff
@@ -78,7 +78,10 @@ protected boolean checkApiLogic(Player ai, SpellAbility sa) {
                     || (sa.getHostCard().isLand() && ai.getLandsInPlay().size() >= 5); // TODO: make this configurable in the AI profile
         }
 
-        return true;
+        // Is the AI in danger of decking out?
+
+        // Allow drawing cards if the AI has less cards than the max hand size
+        return ai.getCardsIn(ZoneType.Hand).size() <= ai.getMaxHandSize();
```
tool4ever, line 84: Superior logic already exists in `targetAI` (which does more than its name might imply):
https://github.com/Card-Forge/forge/blob/137d87c3dfe36773cd7e05726809386dc65745ea/forge-ai/src/main/java/forge/ai/ability/DrawAi.java#L502-L528

so that should be tweaked instead so game win logic etc. stays functional

tehdiplomat: Good spot. I moved it as a sub-check here.

tool4ever (approved):

Issue #6011, opened 2024-08-29
Can't Untap effects needs to be rewritten


Hanmac:
Can't Untap effects needs to be rewritten, some could be Statics, some could be SpellAbilityEffects.

Different Cases:

- [ ] Keyword `CARDNAME doesn't untap during your untap step.` → Static Ability using "CantUntap(Phase)", don't need for `Continuous` ?
   - [ ] Also use it for effects like "doesn't untap during its controller's untap step for as long as EFFECTSOURCE remains on the battlefield."
  - [ ] Some Effects target a Player: `Creatures target player controls don't untap during that player's next untap step.`
  - [ ] Some Effects target a Player and be affected cards: `Tap all creatures target player controls. Those creatures don't untap during that player's next untap step.` the Tricky part is "**that player**"
  - [ ] Drawback of Spells and Activated Effects that say: `doesn't untap during your next untap step` need to remember You as affected player.
- [ ] Hidden Keyword `This card doesn't untap during your next untap step.` → CantUntapTurnEffect using N for counting down on the Card. This is only when it means Controller

Better issue for #984

See also:

- `You may choose not to untap CARDNAME during your untap step.` should be a static?
- Also `untaps during each <Player> untap step` which should be affected by some of the stuff above
  - `CARDNAME untaps during each other player's untap step`
  - `CARDNAME untaps during each opponent's untap step.`

Old Branch that is getting recycled: https://github.com/Card-Forge/forge/tree/untapRework

i probably try to do this in smaller sections

Hanmac:
@tool4ever what do you think? Should these be Static or CantHappen Replacement Effects?

tool4ever:
I don't like that "Can't" wording isn't used which might imply Untap via Effect or Cost would still work?
But maybe I'm overthinking it...

Hanmac:
Now that stun counters also use Replacement Effects for Untap, 
I might try to update some of the effects into a CantHappen Replacement

@tool4ever can you look if `untaps during each <Player> untap step` can be a static trigger?

Hanmac:
@tool4ever now only the "next untap step" are missing

tool4ever:
After #7392 some AI logic might need to be moved

PR #7162, opened 2025-03-15
Cant untap first step

Hanmac:
part of #6011 

This is the first batch of creatures and other cards that does have intrinsic `K:CARDNAME doesn't untap during your untap step.`

i might extend this MR to static abilities adding it.
But for this, i need to update the ReplacmentEffect for `its controllers untap step`

### DIFF
--- a/forge-game/src/main/java/forge/game/phase/Untap.java
+++ b/forge-game/src/main/java/forge/game/phase/Untap.java
@@ -168,9 +168,6 @@ public class Untap extends Phase {
 
         cardsWithKW.addAll(cardsWithKW2);
         for (final Card cardWithKW : cardsWithKW) {
-            if (!cardWithKW.canUntap(player)) {
-                continue;
-            }
             if (cardWithKW.untap(player)) {
                 untapMap.computeIfAbsent(cardWithKW.getController(),
                         i -> new CardCollection()).add(cardWithKW);
--- a/forge-game/src/main/java/forge/game/phase/Untap.java
+++ b/forge-game/src/main/java/forge/game/phase/Untap.java
@@ -98,7 +98,7 @@ public class Untap extends Phase {
         //exerted need current player turn
         final Player playerTurn = c.getGame().getPhaseHandler().getPlayerTurn();
 
-        return !c.isExertedBy(playerTurn);
+        return c.canUntap(playerTurn);
     }
 
     public static final Predicate<Card> CANUNTAP = Untap::canUntap;
@@ -152,7 +152,7 @@ public class Untap extends Phase {
         });
 
         for (final Card c : list) {
-            if (optionalUntap(c)) {
+            if (optionalUntap(c, player)) {
                 untapMap.computeIfAbsent(player, i -> new CardCollection()).add(c);
             }
         }
@@ -168,10 +168,10 @@ public class Untap extends Phase {
 
         cardsWithKW.addAll(cardsWithKW2);
         for (final Card cardWithKW : cardsWithKW) {
-            if (cardWithKW.isExertedBy(player)) {
+            if (!cardWithKW.canUntap(player)) {
                 continue;
             }
-            if (cardWithKW.untap(true)) {
+            if (cardWithKW.untap(player)) {
                 untapMap.computeIfAbsent(cardWithKW.getController(),
                         i -> new CardCollection()).add(cardWithKW);
             }
@@ -203,7 +203,7 @@ public class Untap extends Phase {
             }
         }
         for (Card c : restrictUntapped) {
-            if (optionalUntap(c)) {
+            if (optionalUntap(c, player)) {
                 untapMap.computeIfAbsent(player, i -> new CardCollection()).add(c);
             }
         }
@@ -228,7 +228,7 @@ public class Untap extends Phase {
         game.getTriggerHandler().runTrigger(TriggerType.UntapAll, runParams, false);
     }
 
-    private static boolean optionalUntap(final Card c) {
+    private static boolean optionalUntap(final Card c, Player phase) {
         boolean untap = true;
 
         if (c.hasKeyword("You may choose not to untap CARDNAME during your untap step.") && c.isTapped()) {
@@ -247,7 +247,7 @@ public class Untap extends Phase {
             untap = c.getController().getController().chooseBinary(new SpellAbility.EmptySa(c, c.getController()), prompt.toString(), BinaryChoiceType.UntapOrLeaveTapped, defaultChoice);
         }
         if (untap) {
-            if (!c.untap(true)) untap = false;
+            if (!c.untap(phase)) untap = false;
         }
         return untap;
     }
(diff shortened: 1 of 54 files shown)


forge-gui/res/cardsfolder/b/bewitching_leechcraft.txt
Fulgur14, line 1: "When this Aura" enters

forge-gui/res/cardsfolder/d/deep_slumber_titan.txt
```diff
@@ -2,10 +2,10 @@ Name:Deep-Slumber Titan
...
+R:Event$ Moved | ValidCard$ Card.Self | Destination$ Battlefield | ReplacementResult$ Updated | ReplaceWith$ ETBTapped | Description$ This creature enters tapped.
 SVar:ETBTapped:DB$ Tap | Defined$ Self | ETB$ True
-K:CARDNAME doesn't untap during your untap step.
-T:Mode$ DamageDoneOnce | ValidTarget$ Card.Self | Execute$ TrigUntap | TriggerDescription$ Whenever CARDNAME is dealt damage, untap it.
+R:Event$ Untap | ValidCard$ Card.Self | UntapStep$ You | Layer$ CantHappen | Description$ This creature doesn't untap during your untap step.
+T:Mode$ DamageDoneOnce | ValidTarget$ Card.Self | Execute$ TrigUntap | TriggerDescription$ Whenever this creature is dealt damage, untap it.
 SVar:TrigUntap:DB$ Untap | Defined$ Self
 AI:RemoveDeck:Random
-Oracle:Deep-Slumber Titan enters tapped.\nDeep-Slumber Titan doesn't untap during your untap step.\nWhenever Deep-Slumber Titan is dealt damage, untap it.
+Oracle:This creature enters tapped.\nDeep-Slumber Titan doesn't untap during your untap step.\nWhenever this creature is dealt damage, untap it.
```
Fulgur14, line 11: Second ability should be "This creature" as well.

forge-gui/res/cardsfolder/j/jokulmorder.txt
```diff
@@ -3,12 +3,12 @@ ManaCost:4 U U U
...
-T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigSacUnless | TriggerDescription$ When CARDNAME enters, sacrifice it unless you sacrifice five lands.
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigSacUnless | TriggerDescription$ When this creature enters, sacrifice it unless you sacrifice five lands.
 SVar:TrigSacUnless:DB$ Sacrifice | UnlessCost$ Sac<5/Land> | UnlessPayer$ You
-K:CARDNAME doesn't untap during your untap step.
-T:Mode$ LandPlayed | ValidCard$ Island.YouCtrl | Execute$ TrigUntap | TriggerZones$ Battlefield | OptionalDecider$ You | TriggerDescription$ Whenever you play an Island, you may untap CARDNAME.
+R:Event$ Untap | ValidCard$ Card.Self | UntapStep$ You | Layer$ CantHappen | Description$ This creature doesn't untap during your untap step.
+T:Mode$ LandPlayed | ValidCard$ Island.YouCtrl | Execute$ TrigUntap | TriggerZones$ Battlefield | OptionalDecider$ You | TriggerDescription$ Whenever you play an Island, you may untap this creature.
 SVar:TrigUntap:DB$ Untap | Defined$ Self
 SVar:NeedsToPlayVar:Y GE5
 SVar:Y:Count$Valid Land.YouCtrl
```
Fulgur14, line 14: This creature

forge-gui/res/cardsfolder/f/freyalises_winds.txt
```diff
@@ -3,7 +3,7 @@ ManaCost:2 G G
 Types:Enchantment
 T:Mode$ Taps | TriggerZones$ Battlefield | Execute$ TrigPutCounter | TriggerDescription$ Whenever a permanent becomes tapped, put a wind counter on it.
 SVar:TrigPutCounter:DB$ PutCounter | Defined$ TriggeredCardLKICopy | CounterType$ WIND | CounterNum$ 1
-R:Event$ Untap | ActiveZones$ Battlefield | ValidCard$ Permanent.counters_GE1_WIND | ReplaceWith$ RepRemoveCounter | UntapStep$ True | Description$ If a permanent with a wind counter on it would untap during its controller's untap step, remove all wind counters from it instead.
+R:Event$ Untap | ActiveZones$ Battlefield | ValidCard$ Permanent.counters_GE1_WIND | ReplaceWith$ RepRemoveCounter | UntapStep$ Player | Description$ If a permanent with a wind counter on it would untap during its controller's untap step, remove all wind counters from it instead.
```
tool4ever, line 6: this logic seems wrong because you're now only checking that it's any untap step (instead of controller)

tool4ever: could add `+ActivePlayerCtrl`?

forge-game/src/main/java/forge/game/phase/Untap.java
```diff
@@ -168,10 +168,10 @@ private void doUntap() {
 
         cardsWithKW.addAll(cardsWithKW2);
         for (final Card cardWithKW : cardsWithKW) {
-            if (cardWithKW.isExertedBy(player)) {
+            if (!cardWithKW.canUntap(player)) {
```
tool4ever, line 171: this line is redundant now because untap call below would just return false anyway

forge-gui/res/cardsfolder/f/freyalises_winds.txt
tool4ever, line 6: and perhaps instead of UntapStep just use `ActivePhases$ Untap`

Hanmac: This probably would collide with `Bewitching Leechcraft` one?

The ReplacementEffect that does still untap for a cost?

tool4ever: hmn good point, the result would change (though I'm not sure which one the rules might actually consider valid)

tool4ever:
We need to think about how checks in `isUselessCreature` + `CreatureEvaluator` could be updated for AI

forge-gui/res/cardsfolder/f/freyalises_winds.txt
Hanmac: my gut is telling me, it should be affected by it

Hanmac: i added ActivePlayerCtrl

tool4ever (approved):
good first step imo
maybe I have time for AI later

Hanmac:
@tool4ever should we add the other ones with this MR too?
the other does add it via static? "Enchanted doesn't untap"?

tool4ever:
I wouldn't mind a follow-up either so I can try to spin off some parts

PR #7174, opened 2025-03-17
Refactor Telekinesis

tool4ever:


Hanmac:
Looks okay i think, but i would have liked if you would have made the fixes separate, and the CantUntap stuff in other MR. (or directly into my MR)

PR #7200, opened 2025-03-19
Fix Valiant tracking incorrectly with controller change

tool4ever:


Hanmac:
@tool4ever 

Can we try to unify the code parts from SpellAbilityStackInstance.updateTarget and MagicStack?
(maybe move the unified code to SpellAbility?)

the distinctObjects loop looks nearly the same

Agetian (approved):
Looks good to me :+1:

Hanmac (approved):
unification can be done later

PR #7206, opened 2025-03-19
WASD movement for Adventure mode

Drecon84:
I messed up earlier, but this time I tested it enough to feel like it is actually all in order now.


forge-gui-mobile/src/forge/adventure/util/KeyBinding.java
```diff
@@ -29,16 +29,26 @@ public enum KeyBinding {
...
+        this.name=name;
+        this.defaultBinding=binding=defaultBinding;
+        this.defaultBindingController=bindingController=defaultBindingController;
+        this.alternativeBinding=alternativeBinding;
+    }
+
     public boolean isPressed(int key)
     {
-        return key==binding||key==bindingController;
+        return key==binding||key==bindingController||key==alternativeBinding;
```
tehdiplomat, line 51: This is starting to feel cumbersome. Should we just have an array/set of integers that get passed into each binding and we just check if the key that's pressed is included in the array?

Drecon84: Ooh, I like that. I was trying to do it while changing as little to the code as possible, but the whole construction with the double constructor and sets of different amounts of variables is a little bit messy. 

I'll change the script up to use the array, it seems much cleaner in practice.

Drecon84: Thinking about it a second longer, I think I'll look at the option to make the whole construction a dictionary. Seems a bit cleaner.

tehdiplomat (approved):

Drecon84:
@tehdiplomat The fix for the text box was merged, but the WASD keybind stuff is still reverted. Was that something I should have put back in or is that something that someone from the regulars should still un-revert?

PR #7209, opened 2025-03-19
Add DetainEffect

Hanmac:
Added Detain Effect to move some into extra logic

I will transport the AI logic tomorrow


Hanmac:
@tool4ever should i add a property to the CardView something like "detained by <Players>" ? or is that overkill for just 10 cards?

Agetian:
Looks good to me, though it needs the AI update (move from Pump) :)

tool4ever:
Imo View update isn't needed since effect will be available to check + the duration is fixed anyway

But what could be done later is a more general approach for these kind of statics to display
Or ideally give clearer feedback, e.g. which CantBeActivated caused fail when human attempts to activate

Hanmac:
I couldn't get to write the AI stuff this weekend

@Agetian can you use this as start for more code? (i took part of TapAi)

Agetian:
> I couldn't get to write the AI stuff this weekend
> @Agetian can you use this as start for more code? (i took part of TapAi)

Yeah, I can take a look at how to improve it :) I'll try to figure out my plans for the week and schedule this ;)

forge-gui/res/cardsfolder/l/linvala_shield_of_sea_gate.txt
```diff
@@ -4,7 +4,10 @@ Types:Legendary Creature Angel Wizard
 PT:3/3
 K:Flying
 T:Mode$ Phase | Phase$ BeginCombat | ValidPlayer$ You | TriggerZones$ Battlefield | CheckSVar$ X | SVarCompare$ EQ4 | Execute$ TrigPump | TriggerDescription$ At the beginning of combat on your turn, if you have a full party, choose target nonland permanent an opponent controls. Until your next turn, it can't attack or block, and its activated abilities can't be activated.
-SVar:TrigPump:DB$ Pump | ValidTgts$ Permanent.nonLand+OppCtrl | TgtPrompt$ Choose target nonland permanent an opponent controls | KW$ HIDDEN CARDNAME can't attack or block. & HIDDEN CARDNAME's activated abilities can't be activated. | IsCurse$ True | Duration$ UntilYourNextTurn | AILogic$ DetainNonLand
+SVar:TrigPump:DB$ Effect | ValidTgts$ Permanent.nonLand+OppCtrl | TgtPrompt$ Choose target nonland permanent an opponent controls | RememberObjects$ Targeted | StaticAbilities$ DBCantAttack,DBCantBlockBy,DBCantBeActivated | ForgetOnMoved$ Battlefield | IsCurse$ True | Duration$ UntilYourNextTurn | AILogic$ DetainNonLand
```
tool4ever, line 7: the logic still needs to be moved
or maybe just use DetainEffect anyway?

Hanmac: @tool4ever i tried more generic but it is still WIP

Hanmac:
@Agetian @tool4ever now the others are cleaned up, this just needs more logic for DetainAi (and maybe for the generic EffectAi)

Hanmac:
@tool4ever especially check out the EffectAi section, we might need that

forge-ai/src/main/java/forge/ai/ability/EffectAi.java
```diff
@@ -383,6 +385,118 @@ protected AiAbilityDecision checkApiLogic(final Player ai, final SpellAbility sa
...
+        } else if (sa.hasParam("RememberObjects")) { //generic
+            boolean cantAttack = false;
+            boolean cantBlock = false;
+            boolean cantActivate = false;
+
+            String duraction = sa.getParam("Duration");
+
+            String matchStr = "Card.IsRemembered";
+
+            for (String st : sa.getParam("StaticAbilities").split(",")) {
```
tool4ever, line 393: seems good
but it really needs it's own helper function that can also be used by DetainAI and especially AttachAI because a lot of Auras are also doing that

Hanmac: This is not for DetainEffect anymore, but might be useful for #8172

Hanmac:
@Agetian we need more AI logic for this one too

Agetian:
Yeah, I still need to look into this one :+1: 
Do you have any ideas for what can be added to the AI?

Hanmac:
> Yeah, I still need to look into this one 👍 Do you have any ideas for what can be added to the AI?

Most of the stuff is related to `CARDNAME can't attack or block.` & `CARDNAME's activated abilities can't be activated.`

so stuff from PumpAi that would make opponents creatures not be able to attack/block

Hanmac:
@tool4ever now with HiddenStatics, what do you think about moving the Static into a hidden effect via a Property?
(we could add a Detained Icon)

Or would the `UntilYourNextTurn` mess with that too much? 🤔 

@kevlahnota might add this icon:
Duels_Detain

Hanmac:
@tool4ever 
If i want to add a detainedBy PlayerCollection,
then the hiddenStatic might not even be needed?
i can just return true/false for canAttack/canBlock checks, right?

See `exertedByPlayer`

tool4ever:
hmn I guess that could be slightly better for performance

Hanmac:
I removed `CARDNAME's activated abilities can't be activated.`

we just need to update the icon, but @kevlahnota can do that later

Hanmac:
i might need to add more hints to the AI into `isUseless` checks

tool4ever:
yea this looks nice now
I'll do some clean up tomorrow

PR #7240, opened 2025-03-23
Store cards by collectorNumber instead of artIndex; PaperCard flag support

Jetz72:
This revises the way cards are written to files, with two primary goals. The first is to switch from storing cards by artIndex to the collector number. Ever the thorn in our side, the Secret Lair set showed the problems with using art index for storage, when new card variants were being added with lower collector numbers. This should resolve that.

The other purpose is to generalize the logic Cryptic Spires was using to store its marked color identity. Going forward this will make it easier to implement more cards that do this (if Wizards prints new ones or if we want to add in certain playtest cards). But this also supports other properties we might want to add to a specific copy of a card. The `noSell` flag has been rolled into this, allowing for several parts of the Adventure collection logic to be simplified. We no longer have to keep track of how many copies of each card are unsellable - now an unsellable copy registers as a separate item in the collection.

At some point we could use this system for the `isFoil` flag as well, and streamline a few other parts of the code. I didn't go that far this time because the scope was already getting a bit broad.

I've checked that older decks and adventure saves are imported correctly when loading them, but it wouldn't hurt to give it a spin to make sure I didn't miss any edge cases.

Also renames "chosenColorID" to "markedColors" for clarity. Since "chosen____" properties are defined within the game and aren't copyable, think it makes sense to distinguish them from markings on a card made outside the game.


forge-core/src/main/java/forge/card/ColorSet.java
```diff
@@ -292,13 +294,10 @@ public byte getColor() {
     @Override
     public String toString() {
         if (this.orderWeight == -1) {
-            return "n/a";
+            return "C";
```
tehdiplomat, line 297: Is this right? Colorless is used to map to 0 not -1?

Jetz72: I think we can just remove that entirely. It shouldn't even be possible for `orderWeight` to become -1. Looks like the check was added back in 2011, when  `ColorSet.getNullColor()` returned a distinct constant with an `orderWeight` of -1. If the color is 0 it'll  fall into `shardOrderLookup[0]` and print 'C'.

tehdiplomat:
I haven't forgotten about this. Once we get the next release out we'll get this out the door.

forge-core/src/main/java/forge/deck/io/DeckSerializer.java
```diff
@@ -61,6 +61,8 @@ private static List<String> serializeDeck(Deck d) {
         }
     
         for(Entry<DeckSection, CardPool> s : d) {
+            if(s.getValue().isEmpty())
```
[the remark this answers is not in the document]
Jetz72: Considered always including Main and Sideboard since most decks use them. But the pattern is broken by dedicated Planar and Scheme decks which only have those specific sections and card types.

tehdiplomat (approved):

PR #7245, opened 2025-03-24
Fix Danny Pink triggering once for each type

tool4ever:


Hanmac:
@tool4ever i wonder if part of `CounterTypeAddedAll` should be added to `replaceCounterEffect` ?
After the `Apply counter after replacement effect` loop?

dracontes:
Tested it for both situations (Danny Pink + animated Agent's Toolkit and Stalwart Successor + Tayam) plus an overlap of those: Works as expected 👍

Hanmac (approved):

Hanmac (approved):

PR #7250, opened 2025-03-24
fix: copy to clipboard now collates copies of a card together.

HeitorBittenc:


forge-gui-mobile/src/forge/deck/FDeckViewer.java
```diff
@@ -57,8 +58,15 @@ public static void copyDeckToClipboard(Deck deck) {
...
             deckList.append(s.toString()).append(": ");
             deckList.append(nl);
+            int count = 0;
             for (final Entry<PaperCard, Integer> ev : cp) {
-                deckList.append(ev.getValue()).append(" ").append(ev.getKey().getCardName()).append(nl);
+                count = cp.countByName(ev.getKey());
+                deckList.append(count).append(" ").append(ev.getKey().getCardName()).append(nl);
+                if(count != ev.getValue())
+                {
+                    cp.removeAll(cp.getFilteredPool(PaperCardPredicates.name(ev.getKey().getCardName())));
```
Hanmac, line 67: isn't that a `ConcurrentModificationException`?

you can't alter cp while iterating over it

HeitorBittenc: You are right, i will create another list and see if it contains there instead of altering cp..

forge-gui-mobile/src/forge/deck/FDeckViewer.java
```diff
@@ -50,15 +51,19 @@ public static void copyDeckToClipboard(Deck deck) {
...
             CardPool cp = deck.get(s);
             if (cp == null || cp.isEmpty()) {
                 continue;
             }
             deckList.append(s.toString()).append(": ");
             deckList.append(nl);
+            Set<String> accounted = new HashSet<>();
             for (final Entry<PaperCard, Integer> ev : cp) {
-                deckList.append(ev.getValue()).append(" ").append(ev.getKey().getCardName()).append(nl);
+                if (!accounted.contains(ev.getKey().getCardName())) {
```
tehdiplomat, line 63: I don't think this is going to do what you think it should do. The way I'm reading this only the first entry of a name will be outputted to the copy with however many of that specific paper card. That means if I have 20 Plains from 10 different sets, only the first version of that Plains will show up.

tehdiplomat: What you probably want to do is make a new map thats String -> Integer, and if the string isn't a key added it as a key and populate with the current value. If it is already a key just add ev.getValue() to that.

HeitorBittenc: cp.countByName(ev.getKey()) already will count all cards where the key equals that name. Now if we want to show from what set a card is, then yeah it will have to change.

tehdiplomat: Ah I see. So it looks like you end up nesting a loop here. (We loop through the entire deck, and count for each key but if the key already showed up we skip the nested loop), as opposed to just tallying each element based off the name of the keyed item. Probably not the biggest deal, but if I was confused reading the PR, its likely someone else will be confused when they stumble on this code later. (The important piece is buried in a concatenation line)

If you never care about the value _in this loop_ why are we looping over the Entry and not just the keys?

HeitorBittenc: Fair enough, I'll add some comments for clarification and make it loop only on the Key.

tehdiplomat (approved):

PR #7254, opened 2025-03-25
Endure effect

Hanmac:
- [x] Better Action Message
- [x] Need cards
- [x] Needs AI logic (similar to Explore?)

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/dusyut_earthcarver.txt
+++ b/forge-gui/res/cardsfolder/upcoming/dusyut_earthcarver.txt
@@ -2,6 +2,7 @@ Name:Dusyut Earthcarver
 ManaCost:5 G
 Types:Creature Elephant Druid
 PT:4/4
+K:Reach
 T:Mode$ ChangesZone | ValidCard$ Card.Self | Origin$ Any | Destination$ Battlefield | Execute$ TrigEndure | TriggerZones$ Battlefield | TriggerDescription$ When this creature enters, it endures 3. (Put three +1/+1 counters on it or create a 3/3 white Spirit creature token.)
 SVar:TrigEndure:DB$ Endure | Num$ 3
 Oracle:Reach\nWhen this creature enters, it endures 3. (Put three +1/+1 counters on it or create a 3/3 white Spirit creature token.)
--- a/forge-gui/res/cardsfolder/upcoming/dusyut_earthcarver.txt
+++ b/forge-gui/res/cardsfolder/upcoming/dusyut_earthcarver.txt
@@ -1,0 +1,7 @@
+Name:Dusyut Earthcarver
+ManaCost:5 G
+Types:Creature Elephant Druid
+PT:4/4
+T:Mode$ ChangesZone | ValidCard$ Card.Self | Origin$ Any | Destination$ Battlefield | Execute$ TrigEndure | TriggerZones$ Battlefield | TriggerDescription$ When this creature enters, it endures 3. (Put three +1/+1 counters on it or create a 3/3 white Spirit creature token.)
+SVar:TrigEndure:DB$ Endure | Num$ 3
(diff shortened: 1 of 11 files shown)


dracontes:
Tested it on my side with one of the unclaimed endure cards. Works as expected, including the interaction with Solemnity. Here's the Descendant of Stoms script for convenience.
[descendant_of_storms.txt](https://github.com/user-attachments/files/19443700/descendant_of_storms.txt)

For the Prompt message I'd suggest "Do you want to put N +1/+1 counters on this creature instead of creating a N/N white Spirit creature token?" with N replaced by the correct values for the specific situation. On the other hand, maybe having the "Yes" and "No" buttons replaced by "Counters" and "Token" respectively might be worthwhile, either by itself or in conjunction with the message.

dracontes:
> the cards are not really claimed, more like blocked because of Effect so feel free to add them to this MR

Since it's the first time I'm contributing to a branch other than the master, I'd rather have confirmation that this is the correct setup to contribute, before going ahead with it heedlessly and mess it up.

image

> About the Prompt, technically the Effect is similar to Fabricate, which uses PutCounter as Unless Cost, but I don't know 🤷‍♂️

Oh okay: like a "you may discard a card. If you do, draw a card" process then.
> `You may put put N +1/+1 counters on this creature. If you don't, create a N/N white Spirit creature token.`

Hanmac:
> > the cards are not really claimed, more like blocked because of Effect so feel free to add them to this MR
> 
> Since it's the first time I'm contributing to a branch other than the master, I'd rather have confirmation that this is the correct setup to contribute, before going ahead with it heedlessly and mess it up.
> 
> image

Yeah that's one way, other way would be to commit them directly into this branch

> 
> > About the Prompt, technically the Effect is similar to Fabricate, which uses PutCounter as Unless Cost, but I don't know 🤷‍♂️
> 
> Oh okay: like a "you may discard a card. If you do, draw a card" process then.
> 
> > `You may put put N +1/+1 counters on this creature. If you don't, create a N/N white Spirit creature token.`

yeah something like that

forge-gui/res/cardsfolder/upcoming/dusyut_earthcarver.txt
```diff
@@ -0,0 +1,7 @@
+Name:Dusyut Earthcarver
+ManaCost:5 G
+Types:Creature Elephant Druid
+PT:4/4
```
tool4ever, line 4: KW missing

Hanmac:
@tool4ever can you look at the MessageFormat ? i tried to use the same key multiple times, but i somehow failed :/

Hanmac:
@Agetian just the AI is missing

Agetian:
> @Agetian just the AI is missing

Will do this one today. Do you have any ideas when the AI should prefer one over the other? It could be a pretty difficult choice :)

tool4ever:
The logic for Fabricate should be generalized imo, the only difference is X 1/1 vs. X/X

Hanmac:
I will add the confirmAction for AI later today so it can decide if it wants +1/+1 or Spirit tokens

Agetian:
Ok, I added a rudimentary EndureAi, which, for now, adapts the code from Fabricate in order to make the decision, and also uses the basic checks based on ExploreAi (I decided to keep the targeting checks in, just in case there's ever a "...target creature endures X" card in the future). Unfortunately, couldn't figure out a clean enough way to refactor TokenAi Fabricate code in order to generalize it for use through EndureAi, they use different token templates and different ways to count the amount :/ Need a bit more time to do that than I currently have with my workload, hopefully someone can help out with that if/when needed.

Hanmac:
@Agetian i updated the logic a bit, the token was created as 0/0, i needed to set the base power and toughness and apply static checks for testing.

Agetian:
> @Agetian i updated the logic a bit, the token was created as 0/0, i needed to set the base power and toughness and apply static checks for testing.

Nice, thanks a lot for help! :)

tool4ever (approved):
AI looks a bit rough but should be ok

PR #7262, opened 2025-03-25
3 TDC + 1 TDM cards

Fulgur14:
Should cards like Conspicuous Snoop be changed similar to this Thundermane Dragon? I'm not sure they would work correctly with things like Adventures or MDFC.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/sunpearl_kirin.txt
+++ b/forge-gui/res/cardsfolder/upcoming/sunpearl_kirin.txt
@@ -5,7 +5,7 @@ PT:2/1
 K:Flash
 K:Flying
 T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigChange | TriggerDescription$ When this creature enters, return up to one other target nonland permanent you control to its owner's hand. If it was a token, draw a card.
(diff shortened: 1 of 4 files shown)


forge-gui/res/cardsfolder/upcoming/sunpearl_kirin.txt
```diff
@@ -0,0 +1,11 @@
+Name:Sunpearl Kirin
+ManaCost:1 W
+Types:Creature Kirin
+PT:2/1
+K:Flash
+K:Flying
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigChange | TriggerDescription$ When this creature enters, return up to one other target nonland permanent you control to its owner's hand. If it was a token, draw a card.
+SVar:TrigChange:DB$ ChangeZone | Origin$ Battlefield | Destination$ Hand | TargetMin$ 0 | TargetMax$ 1 | ValidTgts$ Permanent.YouCtrl+Other | RememberChanged$ True | SubAbility$ DBDraw | TgtPrompt$ Select up to one other target permanent you control
```
tool4ever, line 8: wrong tgts

tool4ever:
most likely the Creature restriction could be merged into the Affected since I'm not aware of any corner case (library or power shouldn't be able to make it illegal in this kind of way)

but this is ok too

tool4ever (approved):

PR #7268, opened 2025-03-26
Adventure: Booster shops now remove only restricted cards, keeping…

HeitorBittenc:
… their sets + Removed Booster Shop Image Fetch

### DIFF
--- a/forge-core/src/main/java/forge/card/PrintSheet.java
+++ b/forge-core/src/main/java/forge/card/PrintSheet.java
@@ -68,7 +68,7 @@ public class PrintSheet {
         for(PaperCard card : cards)
             cardsWithWeights.remove(card);
     }
-    public boolean removeByName(String name)
+    public PaperCard removeByName(String name)
     {
         PaperCard toBeRemoved = null;
         for(PaperCard card : cardsWithWeights.toFlatList())
@@ -79,7 +79,7 @@ public class PrintSheet {
             }
         }
 
-        return toBeRemoved != null && cardsWithWeights.remove(toBeRemoved);
+        return toBeRemoved;
     }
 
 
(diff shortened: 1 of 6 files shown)


forge-core/src/main/java/forge/card/PrintSheet.java
```diff
@@ -68,6 +68,20 @@ public void removeAll(Iterable<PaperCard> cards) {
...
+        for(PaperCard card : cardsWithWeights.toFlatList())
+        {
+            if(card.getCardName().equals(name))
+            {
+                toBeRemoved = card;
+            }
+        }
+
+        return toBeRemoved != null && cardsWithWeights.remove(toBeRemoved);
+    }
```
Jetz72, line 83: Doesn't deleting the restricted cards like this alter the pack data outside adventure mode too?

forge-core/src/main/java/forge/card/PrintSheet.java
```diff
@@ -79,7 +79,7 @@ public boolean removeByName(String name)
             }
         }
 
-        return toBeRemoved != null && cardsWithWeights.remove(toBeRemoved);
+        return toBeRemoved;
```
Jetz72, line 82: I don't see that either this function or the place where it's called actually remove the card anymore. Also it'd only find a single copy with the given name, when a print sheet might have more than one printing.

But I'm generally not a fan of modifying Forge's static data at all for a case like this. I think it'd make more sense to generate filtered versions of any print sheets containing restricted cards, and sample cards from those instead. You might make a function that tests if a print sheet has any restricted cards on it. If not, return the print sheet. If so, create a copy with the restricted cards filtered out and return that. Could optionally cache the output of that function, like in an adventure-specific version of `StaticData.printSheets`.

HeitorBittenc: Got it, i tried making a copy before but ran into some troubles, then i tried to do this, but yeah modifying static data might not be the best approach. Will see what i can do.

PR #7278, opened 2025-03-27
Removed unnecessary request for image fetching

HeitorBittenc:


tehdiplomat:
@HeitorBittenc I think you missed a spot here too? https://github.com/Card-Forge/forge/blob/master/forge-gui-mobile/src/forge/util/LibGDXImageFetcher.java#L107-L111

tehdiplomat (approved):

PR #7292, opened 2025-03-28
TDM final spoiler batch - part 3

Fulgur14:


tool4ever (approved):
thanks

forge-gui/res/cardsfolder/upcoming/rainveil_rejuvenator.txt
```diff
@@ -0,0 +1,8 @@
+Name:Rainveil Rejuvenator
+Types:Creature Elephant Druid
```
squee1968, line 2: missing mana cost

PR #7293, opened 2025-03-28
Fix for main quest

Drecon84:
The main quest had an objective that would not complete. This fixes that and removes a comma that was breaking things elsewhere.
(Again there are a bunch of trailing files that aren't mine but that I suspect are just part of the main branch at this point? I hope it won't be a problem).


tehdiplomat:
@Drecon84 Can you try rebasing with your change? I think its picking up some of the line ending changes from cards uploaded directly in the UI.

tehdiplomat:
Thanks for the fix. It looks like i don't have access to resolve your conflicts, so I created the same PR here: https://github.com/Card-Forge/forge/compare/fix-lastquest?expand=1

Drecon84:
At least the fix for the trailing comma worked out. 

I will figure out how to do this correctly. I have no idea how somehow all of those other files got caught up. I still have my fix for the main quest. I'll figure out how to do this stuff cleanly. Shouldn't be that hard.. 
Thanks for the patience. I just don't have enough experience with Git when working together with others. I'll get there though.

PR #7297, opened 2025-03-29
Omen: first attempt

Hanmac:
This adds logic for Omen SubType

@tehdiplomat @tool4ever should we combine the CardSplitType ?


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -3029,14 +3044,23 @@ else if (o.grantsZonePermissions())
...
-                CardState advState = getState(CardStateName.Adventure);
+                CardState advState = getState(CardStateName.Secondary);
                 StringBuilder sbSA = new StringBuilder();
                 sbSA.append(Localizer.getInstance().getMessage("lblAdventure"));
                 sbSA.append(" — ").append(CardTranslation.getTranslatedName(advState.getName()));
                 sbSA.append(" ").append(sa.getPayCosts().toSimpleString());
                 sbSA.append(": ");
                 sbSA.append(sAbility);
                 sAbility = sbSA.toString();
+            } else if (sa.isOmen() && state.getStateName().equals(CardStateName.Original)) {
```
tehdiplomat, line 3055: Is it worth combining these sections since the only difference is the translation used?

tehdiplomat (approved):

Hanmac:
@tehdiplomat i think we could cleanup the AlternateMode or should we keep that for now?

Hanmac:
@tool4ever rules updates:


> 720.2a If an effect refers to a card, spell, or permanent that “has an Omen,” it refers to an object that has the alternative characteristics of an Omen spell, even if the object currently doesn’t use them.

Currently, there is no card that cares for "has an Omen"

> 720.3d As an Omen spell resolves, its controller shuffles it into its owner’s library instead of putting it into its owner’s graveyard as it resolves.

That should be okay for now? Or should that be changed in the future?
Like, should that be an ReplacementEffect instead? (this is our way of doing `self-replacement effect`, right?)
Or even be part of the MagicStack code?

tool4ever:
Hmn, I think you should probably change it because that wording makes it like Adventure - replacing **rules** and not **effect** parts.
It matters so e.g. Feather is allowed to replace Omen too: https://www.reddit.com/r/magicTCG/comments/d1wlec/

On a side note I suspect TriggerShuffled doesn't fully check between shuffler vs. library owner as being done here. 🤕

Hanmac:
> Hmn, I think you should probably change it because that wording makes it like Adventure - replacing **rules** and not **effect** parts. It matters so e.g. Feather is allowed to replace Omen too: https://www.reddit.com/r/magicTCG/comments/d1wlec/

i will do that later today or at the week end.

> 
> On a side note I suspect TriggerShuffled doesn't fully check between shuffler vs. library owner as being done here. 🤕

for this maybe an extra MR is needed

Hanmac:
@tool4ever the corner case might be very fabricated, but both Adventure and Omen Replacement effects depending on the Type

So wouldn't it be better if I would move them into CardState as intrinsic Replacements like the Planeswalker ones?

PR #7310, opened 2025-03-30
Mobilize X: enable the keyword to pass the SVars to the static

Hanmac:
the other Mobilize X card should be easy without this MR

### DIFF
--- a/forge-game/src/main/java/forge/game/CardTraitBase.java
+++ b/forge-game/src/main/java/forge/game/CardTraitBase.java
@@ -565,8 +565,8 @@ public abstract class CardTraitBase extends GameObject implements IHasCardView, 
     }
 
     protected IHasSVars getSVarFallback() {
-        if (this.getKeyword() != null) {
-            return this.getKeyword();
+        if (this.getKeyword().getStatic() != null) {
+            return this.getKeyword().getStatic();
         }
         if (getCardState() != null)
             return getCardState();
--- a/forge-game/src/main/java/forge/game/CardTraitBase.java
+++ b/forge-game/src/main/java/forge/game/CardTraitBase.java
@@ -565,8 +565,8 @@ public abstract class CardTraitBase extends GameObject implements IHasCardView, 
     }
 
     protected IHasSVars getSVarFallback() {
-        if (this.getKeyword() != null && this.getKeyword().getStatic() != null) {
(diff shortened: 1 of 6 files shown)


forge-game/src/main/java/forge/game/CardTraitBase.java
```diff
@@ -565,8 +565,8 @@ public CardView getCardView() {
     }
 
     protected IHasSVars getSVarFallback() {
-        if (this.getKeyword() != null && this.getKeyword().getStatic() != null) {
-            return this.getKeyword().getStatic();
+        if (this.getKeyword() != null) {
+            return this.getKeyword();
```
tool4ever, line 569: I've some concern here because this altering of the order could cause problems in cases like this:
```
K:Monstrosity:5:7 G G:X:creature card in your graveyard
SVar:X:Count$ValidGraveyard Creature.YouCtrl
```

Now if for some reason the host gets a different cardState (e.g. Clone) it would fail to find the X?

Hanmac: I don't think so, the it's either on the same card, or use Static Ability to gain the Keyword or the Activated Ability?

tool4ever: - Sacrifice _Death Kiss_ for _The Ever-Changing 'Dane_
- Activate Monstrosity
- Sacrifice some other vanilla creature
- Monstrosity resolve gives no counters 💣

Hanmac: Yeah, that one should have used Static instead
Another reason why I want to change
(But I think I can revert that bit for now?)

tool4ever: If no static available falling back to CardState seems safer right now

Hanmac: until all effects are changed to static layers

Hanmac:
@tool4ever i don't know if Keywords should hold Svars themselves, but that might be useful in the future?

tool4ever:
Well it seems the script would work fine alone 🤷‍♂️
Architecture wise this just seems to complicate things so I'd prefer a good use case before carrying empty TreeMap on all KW? :/

tool4ever (approved):

PR #7118, opened 2025-03-04
Highway Reaver fix

dracontes:
As requested [here](https://github.com/Card-Forge/forge/pull/7090#discussion_r1979239154), replacing `Animate` with `Pump`,

Issue #6372, opened 2024-10-16
Aminatou, Veil Piercer Does Not Correctly Reduce Room Costs


Quantumdit:
**Describe the bug**
Aminatou does not seem to correctly reduce the cost of Rooms with her Miracle ability. It seems that the she uses the costs of both rooms combined, then reduces that by {4}. When the Room is played this way, only the room selected when casting the spell is unlocked, while the other room remains locked.

**Instance 1**
On the board I had Aminatou, Veil Piercer, along with two cards (Starfield Mystic and Inquisitive Glimmer) that reduced the mana costs of enchantments by {1} (each). When I drew Bottomless Pool // Locker Room at the beginning of my turn and chose the cast Locker Room ({4}{U}) for the Miracle cost (using Aminatou, Veil Piercer), I expected to pay only {U}, as I should have had a 6 colorless mana discount (2 from the enchantment discount sources, 4 from Aminatou's ability). Instead, I was forced to pay {U}{U}. See screenshots 1, 2 and 3 for the immediate aftermath.

**Instance 2**
On the board I had Aminatou, Veil Piercer, along with Secret Arcade // Dusty Parlor, with only Dusty Parlor unlocked. When I drew Cramped Vents // Access Maze, I chose to cast Access Maze ({5}{B}{B}) for the Miracle cost (using Aminatou, Veil Piercer). I expected to pay {1}{B}{B}, but was forced to pay {4}{B}{B}{B}. Additionally, Dusty Parlor allowed me to put 7 +1/+1 counters on a creature, implying that the mana value of the spell was 7, which is correct.

**Instance 3**
On the board I had Aminatou, Veil Piercer, along with Secret Arcade // Dusty Parlor, with only Dusty Parlor unlocked. When I drew Dazzling Theatre // Prop Room, I chose to play Dazzling Threatre ({3}{W}). I expected to pay only {W}, but instead had to pay {1}{W}{W}. Additionally, Dusty Parlor allowed me to put 4 +1/+1 counters on a creature implying that the mana value of the spell was 4, which is correct.

**To Reproduce**
1. Play Aminatou, Veil Piercer
2. Play Secret Arcade // Dusty Parlor, and have Dusty Parlor unlocked
3. Use Aminatou's ability to play a Room using the Miracle cost
4. Select which room to unlock
5. Observe that the Room spell's cost is the cost of both rooms combined
6. Pay the cost and play the Room
7. Resolve Dusty Parlor's trigger
[...]

PR #7319, opened 2025-04-01
Cleanup: Targeted effects that grant flashback

dracontes:
Prompted by the upcoming [Songcrafter Mage](https://scryfall.com/card/tdm/225/songcrafter-mage) and [earlier discussion](https://github.com/Card-Forge/forge/pull/7118) on [Highway Reaver](https://scryfall.com/card/ydft/22/highway-reaver) regarding the display of costs associated with granted abilities, I experimented a bit with the effects that grant flashback. 
I found that in the case of `Pump` effects, the flashback cost can be displayed, and this is the quality of life update herein. As for the relevant `PumpAll` effects (eg. [Past in Flames](https://scryfall.com/card/mm3/105/past-in-flames)) a similar edit results in the flashback cost somehow being a singular colorless mana. These I left alone.

While I was at it, I standartized relevant `SVar` nomenclature and added a few missing AI flags.

### DIFF
--- a/forge-gui/res/cardsfolder/a/archmages_newt.txt
+++ b/forge-gui/res/cardsfolder/a/archmages_newt.txt
@@ -4,8 +4,8 @@ Types:Creature Salamander Mount
 PT:2/2
 K:Saddle:3
 T:Mode$ DamageDone | ValidSource$ Card.Self | ValidTarget$ Player | CombatDamage$ True | Execute$ TrigBranch | TriggerDescription$ Whenever CARDNAME deals combat damage to a player, target instant or sorcery card in your graveyard gains flashback until end of turn. The flashback cost is equal to its mana cost. That card gains flashback {0} until end of turn instead if CARDNAME is saddled. (You may cast that card from your graveyard for its flashback cost. Then exile it.)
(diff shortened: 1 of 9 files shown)


Hanmac:
The problem there are Split cards (and others like that)
Does it still work correct with such cards?

What about spells with no mana cost?
This check doesn't work anymore
https://github.com/Card-Forge/forge/blob/7f9260f54c509eb081d3ff57a176b624e92274cf/forge-game/src/main/java/forge/game/GameActionUtil.java#L162-L164

tool4ever:
Yea unfortunately not possible since it happens to be a functional change (in Forge) right now:
> 702.1b. An effect that grants an object a keyword ability may define a variable in that ability based on characteristics of that object or other information about the game state. For these abilities, the value of that variable is constantly reevaluated.

> Example: Fire//Ice is a split card whose halves have the mana costs {1}{R} and {1}{U}. Past in Flames reads "Each instant and sorcery card in your graveyard gains flashback until end of turn. The flashback cost is equal to its mana cost." Fire//Ice has "Flashback {2}{U}{R}" while it is in your graveyard, but if you choose to cast Fire, the resulting spell has "Flashback {1}{R}."

Partly related to #6372

dracontes:
> The problem there are Split cards (and others like that) Does it still work correct with such cards?

No it doesn't, Also for the screenshot below the card detail only updated to show the flashback cost after I had attempted to cast the card once, 
Oh well... Good to know for Songcrafter Mage at least.

image
 
> What about spells with no mana cost? 

Seems to work as expected (excusing "Flashback no cost"): I can't cast Evermind from my graveyard.

image

> Yea unfortunately not possible since it happens to be a functional change (in Forge) right now:

Fair. Reverted that bit of the changes made in this PR. I hope the rest is reasonable enough.

Hanmac (approved):

Issue #1828, opened 2022-11-08
Defiler + Card Text Change Effect


Hanmac:
Defiler of Instinct

> As an additional cost to cast red permanent spells, you may pay 2 life. **Those spells cost {R} less to cast if you paid life this way. This effect reduces only the amount of red mana you pay.**

if an Effect would replace red with green for example, then the second part would be *Those spells cost {R} less to cast if you paid life this way. This effect reduces only the amount of green mana you pay.*

but because this part should happen after you choose how to play hybrid mana, the R symbol can't reduce the amount of green mana you pay. so the Ability should fail

Hanmac:
This might need #309 because of the nature of Hybrid Mana Cost
(Rules wise, you need to choose the Mana Cost first, before the Defiler can reduce it)

PR #7321, opened 2025-04-01
Harmonize first try

Hanmac:
The Raise Cost doesn't work yet

but that is the easiest way right now to code Harmonize without extra java code

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
@@ -2314,6 +2314,40 @@ public class CardFactoryUtil {
             re.setOverridingAbility(saExile);
 
             inst.addReplacement(re);
+        } else if (keyword.startsWith("Harmonize")) {
+            StringBuilder sb = new StringBuilder();
+            sb.append("Event$ Moved | ValidCard$ Card.Self | Origin$ Stack | ExcludeDestination$ Exile ");
+            sb.append("| ValidStackSa$ Spell.Harmonize+castKeyword | Description$ Harmonize");
+
+            if (keyword.contains(":")) {
+                final String[] k = keyword.split(":");
+                final Cost cost = new Cost(k[1], false);
+                sb.append(cost.isOnlyManaCost() ? " " : "—").append(cost.toSimpleString());
+                sb.append(cost.isOnlyManaCost() ? "" : ".");
+
+                String extraDesc =  k.length > 3 ? k[3] : "";
+                if (!extraDesc.isEmpty()) { // extra params added in GameActionUtil, desc added here
+                    sb.append(cost.isOnlyManaCost() ? ". " : " ").append(extraDesc);
+                }
+            }
+
+            sb.append(" (").append(inst.getReminderText()).append(")");
+
+            String repeffstr = sb.toString();
+
+            String abExile = "DB$ ChangeZone | Defined$ Self | Origin$ Stack | Destination$ Exile";
+
+            SpellAbility saExile = AbilityFactory.getAbility(abExile, card);
+
+            if (!intrinsic) {
+                saExile.setIntrinsic(false);
+            }
+
+            ReplacementEffect re = ReplacementHandler.parseReplacement(repeffstr, host, intrinsic, card);
+
+            re.setOverridingAbility(saExile);
+
+            inst.addReplacement(re);
         } else if (keyword.startsWith("Graft")) {
             final String[] k = keyword.split(":");
             final String m = k[1];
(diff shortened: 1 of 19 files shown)


tool4ever:
Made it work
currently it shouldn't collide imo? 🤔

Hanmac:
Yeah, we need to test the extrinsic variant

to fix it with Undergrowth, maybe have OptionalCostValue know about if the extra cost is intrinsic or extrinsic?

dracontes:
> Made it work 

Testing it on my side with Songcrafter Mage ([script here](https://github.com/dracontes/forge/blob/rb-harmonizecards/forge-gui/res/cardsfolder/upcoming/songcrafter_mage.txt)), the issue is that even announcing the power of the creature I want to tap to pay for the harmonize cost I'm still forced to pay the full mana cost of, in this case, [Lightning Blast](https://scryfall.com/card/tpr/138/lightning-blast).

image

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -3964,6 +3996,15 @@ else if ("Artifact".equals(t))
             String effect = "Mode$ CantBlockBy | ValidAttacker$ Creature.Self | ValidBlocker$ Creature.withoutFlying+withoutReach | Secondary$ True" +
                     " | Description$ Flying (" + inst.getReminderText() + ")";
             inst.addStaticAbility(StaticAbility.create(effect, state.getCard(), state, intrinsic));
+        } else if (keyword.startsWith("Harmonize")) {
+            String raiseEffect = "Mode$ OptionalCost | ValidCard$ Card.Self | ValidSA$ Spell.Harmonize | Secondary$ True"
+                    + " | Cost$ tapXType<1/Creature.powerEQHarmonize/creature for Harmonize> | EffectZone$ All "
+                    + " | Description$ Harmonize (" + inst.getReminderText() + ")";
+            inst.addStaticAbility(StaticAbility.create(raiseEffect, state.getCard(), state, intrinsic));
+            String reduceEffect = "Mode$ ReduceCost | ValidCard$ Card.Self | Type$ Spell.Harmonize | Secondary$ True"
+                    + " | Amount$ Harmonize | EffectZone$ All | CheckSVar$ Harmonize"
```
tool4ever, line 4005: the problem here with extrinsic:
`addChangedCardKeywordsInternal` doesn't/can't update the host when moving to stack, so it fails to see the Announce on the old one

tool4ever: yea that seems to be the place
but I'm not sure we can safely do it without breaking rollback

tool4ever:
@dracontes
can you test again?
thanks

tool4ever:
Hmn, I'm not fully happy with some of that logics robustness...
Maybe it can be improved, I'll try sleeping on it 🐑

dracontes:
> Hmn, I'm not fully happy with some of that logics robustness...

FWIW, tested again: all the relevant tests are passed with flying colors. 
image

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -3964,6 +3996,15 @@ else if ("Artifact".equals(t))
             String effect = "Mode$ CantBlockBy | ValidAttacker$ Creature.Self | ValidBlocker$ Creature.withoutFlying+withoutReach | Secondary$ True" +
                     " | Description$ Flying (" + inst.getReminderText() + ")";
             inst.addStaticAbility(StaticAbility.create(effect, state.getCard(), state, intrinsic));
+        } else if (keyword.startsWith("Harmonize")) {
+            String raiseEffect = "Mode$ OptionalCost | ValidCard$ Card.Self | ValidSA$ Spell.Harmonize | Secondary$ True"
+                    + " | Cost$ tapXType<1/Creature.powerEQHarmonize/creature for Harmonize> | EffectZone$ All "
+                    + " | Description$ Harmonize (" + inst.getReminderText() + ")";
+            inst.addStaticAbility(StaticAbility.create(raiseEffect, state.getCard(), state, intrinsic));
+            String reduceEffect = "Mode$ ReduceCost | ValidCard$ Card.Self | Type$ Spell.Harmonize | Secondary$ True"
```
tool4ever, line 4004: @Hanmac
We still need to find a cleaner way to connect this
I assume right now if a card has two instances of Harmonize it would apply twice?

Hanmac: I have a plan for this, but it takes some bigger changes.

For the Card/SpellAbility remembering which optional cost was paid, I want to add an extra list especially for the generic cost to hold the Static Ability and/or keyword that did it.

Then in the ReduceCost, you could do something like "is linked" or something.

The problem is that OptionalCostValue can't be the object for this, because that one still needs to be serialize-able.

OR maybe better, what if we change Optional Cost Static? Add a param like "ReduceCost"
So while CostAdjustment, add the Static Abilities that Increased or just only the Optional Cost ones, and then for reducing, it already has the needed parameter 
(That can help clean up the Defiler cards #1828 )

Hanmac: Is this ready for now? Or should I try to do something with the Defilers?

Hanmac: @tool4ever what instead of the Announce, we move the Cost increase to addExtraKeywordCost like with Casualty?

tool4ever: it's still experimental and I forgot split card as final test: they fail the KW check 😢

Hanmac: Should we still move it to `addExtraKeywordCost`?

tool4ever: it might be cleaner, will try after dinner

tool4ever (approved):
This PR gave me nightmares

PR #7327, opened 2025-04-01
Call the Spirit Dragons

squee1968:
This should be pretty close, however, I've lost the ability to playtest.


dracontes:
Okay, I did some playtesting on my side. Managed to fix a number of issues, namely the enchantment triggering on all upkeeps and the choice for putting counters being among creatures and not restricted to Dragons alone. But, try as I may, I can't get your explicit sequence to work beyond putting one counter on one single Dragon.

[call_the_spirit_dragons.txt](https://github.com/user-attachments/files/19559797/call_the_spirit_dragons.txt)

image

I should add, that I tried a similar tack a while ago to no avail (among a few), and I wasn't the only person to try this card and fail. Y'know, sometimes the "Wait for rules or engine changes" flag is there for a reason.

squee1968:
Ahhhh, I see where I messed up. I wonder why your script doesn't work either? Seems like yours is scripted perfectly. Thanks for the assistance.

PR #7329, opened 2025-04-02
Gladiolus Amicitia (FIN)

Fulgur14:


forge-gui/res/cardsfolder/upcoming/gladiolus_amicitia.txt
```diff
@@ -0,0 +1,9 @@
+Name:Gladiolus Amicitia
+ManaCost:4 R G
+Types:Legendary Creature Human Warrior
+PT:6/6
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigChange | TriggerDescription$ When NICKNAME enters, search your library for a land card, put it onto the battlefield tapped, then shuffle.
```
dracontes, line 5: The name is quoted in full on the [card's rules text](https://www.reddit.com/r/magicTCG/comments/1jpikis/fin_gladiolus_amicitia/) so it should be CARDNAME. Oracle also needs to be corrected similarly.

Fulgur14: Sorry, I got too used to legends that never use their full name :)

dracontes (approved):

PR #7333, opened 2025-04-03
Remove Quest items on NG+

Drecon84:
This removes quest items from a NG+ run in adventure mode. It might not remove all when a file has multiple NG+'s already, but multiple resets will remove all of the items. I decided to remove the quest item tag from the teleport runes, since they are not added by a quest but can be bought in the store instead. Keeping them seems a good move for NG+.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
+++ b/forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
@@ -1027,6 +1027,17 @@ public class AdventurePlayer implements Serializable, SaveFileContent {
         return true;
     }
 
+    public void removeAllQuestItems(){
+        for (String s : inventoryItems) {
+            ItemData data = ItemData.getItem(s);
+            if(data != null){
+                if(data.questItem){
+                    removeItem(data.name);
+                }
+            }
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
```diff
@@ -1027,6 +1027,17 @@ public boolean addItem(String name) {
         return true;
     }
 
+    public void removeAllQuestItems(){
+        for (String s : inventoryItems) {
+            ItemData data = ItemData.getItem(s);
+            if(data != null){
+                if(data.questItem){
```
tehdiplomat, line 1034: Small nit: Can we combine these conditionals there's not really any reason for these to be nested?

`if(data != null && data.questItem){`

Drecon84: Oh yeah, At first I was doing something else where there was a reason to split them up but when I cleaned it up I forgot to check if I could combine them again. Fix coming right up :)

Drecon84: Done.

tehdiplomat (approved):

PR #7347, opened 2025-04-05
Adventure and Omen as ReplacementEffect in CardState

Hanmac:
This Updates Omen into an ReplacementEffect and moves both of them into Card State depending on the Type

(assuming something could mess with Spell Subtypes)

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardFactory.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactory.java
@@ -281,13 +281,6 @@ public class CardFactory {
             } else if (state != CardStateName.Original) {
                 CardFactoryUtil.setupKeywordedAbilities(card);
             }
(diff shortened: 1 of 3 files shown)


Agetian:
Does this still work for the AI ?

Hanmac:
> Does this still work for the AI ?

which part? the Adventure or the Omen?

for  Adventure i just moved the Location
for Omen, we didn't had much of AI yet?

Agetian:
Yeah, I mean, would the AI still be able to cast the Adventure part?

Hanmac:
> Yeah, I mean, would the AI still be able to cast the Adventure part?

i haven't changed that, the one that allows to be cast from Adventure is still the Effect that gets created by the replacement effect

Agetian:
Kk nice 🙂

tool4ever (approved):

PR #7348, opened 2025-04-05
Put ForColor  for Call of the Spirit Dragons

Hanmac:
it was easier to add it to CounterPut instead of using RepeatEach

@Agetian AI might need some logic for ForColor so it doesn't crash?

The card itself might need some Needs Dragon in Play param

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
@@ -7,6 +7,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 
+import forge.card.MagicColor;
 import forge.game.Game;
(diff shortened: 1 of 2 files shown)


Agetian:
Yes, at the very least, I think a NeedsToPlay:Dragon.YouCtrl should be added so that there's at least one dragon (possibly we can make it 2+ or 3+, but not sure if it'll be too restrictive for the AI that way?), and for ForColor, add a branch into the AI where it chooses appropriately. Might be able to take a look at the latter, but not yet sure when exactly, I'll see if anything works this weekend (eh, busy days at work... :/ )

dracontes:
FWIW, tested it and it's working as expected on my side.

Hanmac:
@Agetian seems the AI doesn't need extra logic for the trigger to run

Agetian:
Nice! Sounds good :)

Agetian (approved):

Issue #2550, opened 2023-02-24
Shadow use CantBlockBy and static abilities


Hanmac:
Shadow can be these two abilities:
* `CantBlockBy | ValidAttacker$ Card.Self | ValidBlocker$ Card.withoutShadow`
* `CantBlockBy | ValidAttacker$ Card.withoutShadow | ValidBlocker$ Card.Self`

Then there is Aetherflame Wall and Aether Web with
`can block creatures with shadow as though they didn’t have shadow`
This can be done in the first ability in checking withoutShadow similar to IfReach.
But i Probably need to check the other ability too because of this rule:
>  If Aetherflame Wall gains shadow, it won’t be able to block any creatures (not even those with shadow). 

Problem there is Wall of Diffusion and Heartwood Dryad with:
`CARDNAME can block creatures with shadow as though CARDNAME had shadow`
The wording is slightly different so i assume that if gains Shadow, it still would be able to block Creatures with Shadow.

for the second one i need to ask judges.

or if combined. how Heartwood Dryad attached with Aether Web can block creatures with Shadow?

Some Matrix Data Table:

  | Attacker | Attacker
-- | -- | --
Blocker | without Shadow | with Shadow
without Shadow | x |  
with Shadow |   | x
Aetherflame Wall without Shadow | x | x
Aetherflame Wall with Shadow |   |  
Heartwood Dryad without Shadow | x | x
Heartwood Dryad with Shadow |   | x
Heartwood Dryad with Aether Web and without Shadow | x | ?
Heartwood Dryad with Aether Web and with Shadow | ? | ?

`x` = "can block"
`?` = are the rules question below

tool4ever:
> 609.4a. If two effects state that a player may (or a creature can) do the same thing "as though" different conditions were true, both conditions could apply. If one "as though" effect satisfies the requirements for another "as though" effect, then both effects will apply.

Hanmac:
> 609.4a. If two effects state that a player may (or a creature can) do the same thing "as though" different conditions were true, both conditions could apply. If one "as though" effect satisfies the requirements for another "as though" effect, then both effects will apply.

@tool4ever what would that mean for the `Heartwood Dryad` with `Aether Web` cases?

tool4ever:
Unfortunately I have no real idea :/
In the rules it actually has an example but it's for combining _Vedalken Orrery_ with _Shaman's Trance_

Hanmac:
[...]

PR #7356, opened 2025-04-06
Card: extraEffects for intrinsic Effects

Hanmac:
Added way to get the internal Effects in case of LKI

in the future, some Statics like CantBlockBy might need to use LKI like for #2550 
(so i can alter the LKI of both attacker and blocker, without adding LKI Effects to the CommandZone)

I still don't know yet if i need the `Map<Effect, Command>` if most of them should be added to LeaveTheBattlefield anyway? @tool4ever

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardCopyService.java
+++ b/forge-game/src/main/java/forge/game/card/CardCopyService.java
@@ -2,6 +2,8 @@ package forge.game.card;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+
+import forge.GameCommand;
 import forge.card.CardStateName;
 import forge.card.CardType;
 import forge.game.Game;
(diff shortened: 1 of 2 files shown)


forge-game/src/main/java/forge/game/card/CardCopyService.java
```diff
@@ -339,9 +346,15 @@ public Card getLKICopy(Map<Integer, Card> cachedMap) {
         newCopy.setIntensity(copyFrom.getIntensity(false));
         newCopy.setPerpetual(copyFrom);
 
-        newCopy.addRemembered(copyFrom.getRemembered());
-        newCopy.addImprintedCards(copyFrom.getImprintedCards());
-        newCopy.setChosenCards(copyFrom.getChosenCards());
+        Map<Card, GameCommand> extraEffects = Maps.newHashMap();
+        for (Card e : copyFrom.getExtraEffects()) { // should not need to run the game commands for LKI
```
tool4ever, line 350: - I don't like all that extra for a playtest card
- not quite sure how this is supposed to work, but the concept of LKI effects seems a bit exotic rules wise
- switching Remembered etc. to extra LKI without good reason seems wrong, nothing says LKI isn't allowed to reference ingame objects

Hanmac: I need it for other effects too, like the #6811 right now doesn't work because I need to create an effect for the spell you would cast
(And then it did flicker with the type)

Hanmac: The problem with Remembered is this:
When Card A with internal Effect B makes a LKI copy of A, and A->B is linked, I need to copy B too. But the copy of B still points to original A, not the copy of A

Hanmac:
Replaced with #9166

PR #7361, opened 2025-04-08
AttachAi: make KeepTapped logic generic

Hanmac:
with that we should be able to remove the "KeepTapped" logic from the Auras


tool4ever:
LGTM

Hanmac:
@tool4ever do you want to remove the KeepTrapped Flag on the Cards with this MR or another one?

Hanmac:
@tool4ever can you update the cards for me in this MR, or should i do that later?

tool4ever:
maybe on the weekend but not near my IDE right now

Hanmac:
@tool4ever 
Hm apropos attach, we might need to separate the logic between such Auras that keep stuff tapped,
and Equipment like `Cement Shoes` that does have it as drawback.

i will think about the second ones if they need some extra logic

Hanmac:
@tool4ever @Agetian i think this is ready?
i will do the Equipments on a later day

Agetian (approved):

PR #7365, opened 2025-04-09
Feature/network improvements

TimothyWright95:
Seperated server properties into their own file, this will eventually help facilitate a headless server.
Fixed the localhost ip mapping to give the correct IP Address instead of failing and defaulting to "localhost"
Fixed UPnP as well as added some additional options and choices regarding UPnP
Added localization strings to all language files. (Translators will need to translate these, but the current English string is there for easy reference so they dont have to search the en-US file)


forge-gui/pom.xml
```diff
@@ -49,20 +49,49 @@
...
-            <groupId>org.jupnp</groupId>
-            <artifactId>org.jupnp</artifactId>
-            <version>3.0.2</version>
-        </dependency>
-        <dependency>
-            <groupId>org.jupnp</groupId>
-            <artifactId>org.jupnp.support</artifactId>
-            <version>3.0.2</version>
+            <groupId>com.simtechdata</groupId>
+            <artifactId>WaifUPnP</artifactId>
```
tehdiplomat, line 53: Does this need to be in proguard? The old library was. Either way we should probably remove that reference if we don't need the old library.

forge-gui/src/main/java/forge/gamemodes/net/ProtocolMethod.java
```diff
@@ -108,7 +110,7 @@ private enum Mode {
             this.toInvoke = toInvoke;
         }
     }
-
+    private final static Logger logger = LogManager.getLogger(ProtocolMethod.class);
```
tehdiplomat, line 113: Forge already has a logger, why are we creating a new one?

forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
```diff
@@ -75,6 +83,35 @@ public static FServerManager getInstance() {
...
+
+                case 3:
+                    FModel.getNetPreferences().setPref(ForgeNetPreferences.FNetPref.UPnP, "NEVER");
+                    FModel.getNetPreferences().save();
+                default:
+                    //case 1 defaults to here
+                    startUPnP = false;
+                    break;
+            }
+        } else startUPnP = UPnPOption.equalsIgnoreCase("ALWAYS");
```
tehdiplomat, line 113: Can we wrap this in braces? This same line else is vary jarring within this code style.

forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
```diff
@@ -75,6 +83,35 @@ public static FServerManager getInstance() {
     }
 
     public void startServer(final int port) {
+        this.port = port;
+        String UPnPOption = FModel.getNetPreferences().getPref(ForgeNetPreferences.FNetPref.UPnP);
+        boolean startUPnP;
+        if(UPnPOption.equalsIgnoreCase("ASK")) {
+            switch (SOptionPane.showOptionDialog(localizer.getMessageorUseDefault("lblUPnPQuestion", String.format("Attempt to open port %d automatically?", port), port),
```
tehdiplomat, line 90: Not sure I'm a huge fan of nesting this dialog inside the switch. Makes it hard to follow whats actually happening.

forge-gui/src/main/java/forge/gamemodes/net/NetConnectUtil.java
```diff
@@ -122,7 +122,22 @@ public static void copyHostedServerUrl() {
 
     public static ChatMessage join(final String url, final IOnlineLobby onlineLobby, final IOnlineChatInterface chatInterface) {
         final IGuiGame gui = GuiBase.getInterface().getNewGuiGame();
-        final FGameClient client = new FGameClient(FModel.getPreferences().getPref(FPref.PLAYER_NAME), "0", gui);
+        String hostname = url;
+        int port = FModel.getNetPreferences().getPrefInt(ForgeNetPreferences.FNetPref.NET_PORT);
+
+        //see if port specified in URL
+        int urlSeparatorIndex = url.indexOf(':');
```
tehdiplomat, line 129: I know this isn't your code, but I think this can be written cleaner.

forge-gui/pom.xml
TimothyWright95: Apologies but I am not sure what proguard is but based on a quick Google search it appears to be some sort of optimizer for android apks. If that is correct than the new library should be added as the UPnP stuff won't work without it.

As for jupnp it's no longer needed and only supports up to java 11 and as far as i can tell wasnt actually working, which is why I replaced it with a different library. So i don't see an issue removing it from proguard. 
The new library is only ever used in the FServerManager to open a port if the router supports UPnP

forge-gui/src/main/java/forge/gamemodes/net/ProtocolMethod.java
TimothyWright95: When looking through the project i saw instances of an AppllicationLogger but it was all contained in an android portion of the code. I saw some includes for slf4j but no instances of where any implementation of an slf4j logger was being used (logback was in the dependencies but i didnt see it used anywhere). I also saw some log4j.xml configs but they didnt actually create any appenders. 

In the networking code there were never any log statements just system.out.println.

I added the log4j2 dependencies so that i can start outputting logs to a file as well as the console, which will be necessary  for a headless server. 
It also will allow me to add in thread name to the logs, which will be a necessity once i implement multiple game lobbies on a single server (each lobby will get its own thread).

If i missed an implementation of a logger somewhere i'm fine swapping over to that instead. I just couldn't find one.

forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
TimothyWright95: I can shove the switch dialog off in a function (callUPnPDialog()) if that would meke it easier to follow, or do you want me to remove the dialog from the switch completely, set a boolean, and call the function later?

forge-gui/src/main/java/forge/gamemodes/net/NetConnectUtil.java
TimothyWright95: int port = FModel.getNetPreferences().getPrefInt(ForgeNetPreferences.FNetPref.NET_PORT);

If this is the line you are refering too, i can shove it off into a function getServerConfigPort(), that should make following the code easier. If your refering to the url serperator code beneath i can look into making that better

forge-gui/pom.xml
tool4ever: Java UPnP support isn't in a great state but JUPnP seems like the better maintained project to me (even if we don't need all features)
More importantly it should also support Android which our tests often won't detect: https://github.com/adolfintel/WaifUPnP/pull/14

Imo the Java 11 limitation you mention is for building the library from source, not using it...

The migration to this fork of Cling was only done by another contributor a couple of months ago, maybe they forgot some required calls 🤔
Ideally we can avoid switching again - though I do like the control dialog you added and thanks for diving in!

forge-gui/src/main/java/forge/gamemodes/net/ProtocolMethod.java
tool4ever: we use a pretty basic/old library and perhaps switching would have its benefits, but we should stick to a single framework (and handling it with this PR seems a bit too much scope creep imo)

https://github.com/Card-Forge/forge/blob/master/forge-gui/src/main/java/forge/error/ExceptionHandler.java

forge-gui/src/main/java/forge/gamemodes/net/NetConnectUtil.java
tehdiplomat: I just meant this whole block of code. It's basically trying to split by colon and do somethings based off the length, but its doing it awkwardly and swallowing an exception in the meantime.

forge-gui/pom.xml
neoFuzz: Probably did miss some calls. Used this fork of Cling/jupnp since it had been further developed over the old version we were using

forge-gui/src/main/java/forge/gamemodes/net/NetConnectUtil.java
TimothyWright95: ok, i replaced it with a URI parcer, looks much nicer and should hopefully be much easier to follow along with

forge-gui/pom.xml
TimothyWright95: i kept getting missing dependency exceptions when trying to initially fix the jupnp code, tried fixing that for a few hours before i decided to switch it up. I did quit trying to fix it after seeing it was java 11 compatible, not realizing that was just for building from source. 
 
If the new library doesn't support android, i don't think that's permissible so ill look into switching back to JuPNP

forge-gui/res/languages/en-US.properties
```diff
@@ -2985,6 +2985,14 @@ lblForgeUnableDetermineYourExternalIP=Forge was unable to determine your externa
...
 lblYourConnectionToHostWasInterrupted=Your connection to the host ({0}) was interrupted.
 lblConnectedIPPort=Connected to {0}:{1}
+#FServerManager.java
+lblUPnPTitle=UPnP option
+lblUPnPQuestion=Attempt to open port {0} automatically?
+lblUPnPFailed=UPnP failed to open port {0}. You will need to manually create a port forward for external clients to connect.
+lblUPnPAlways=Always
+lblUPnPNotNow=Not Now
+lblUPnPNever=Never
+lblUPnPJustOnce=Just Once
```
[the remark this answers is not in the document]
TimothyWright95: Sure, the localization files just seems to be broken down into which files they were used in, i can make them generic easily, is there a better place to put them in the localization file, or jst leave them where they are and make the key generic (aka remove the UPnP)

forge-gui/src/main/java/forge/gamemodes/net/NetConnectUtil.java
```diff
@@ -122,7 +124,27 @@ public static void copyHostedServerUrl() {
 
     public static ChatMessage join(final String url, final IOnlineLobby onlineLobby, final IOnlineChatInterface chatInterface) {
         final IGuiGame gui = GuiBase.getInterface().getNewGuiGame();
-        final FGameClient client = new FGameClient(FModel.getPreferences().getPref(FPref.PLAYER_NAME), "0", gui);
+        String hostname;
+        //todo Should forge have a default port defined in Forge.Constants?
+        // Current implementation would mean the users preferred hosting port is their
+        //  default connection port
```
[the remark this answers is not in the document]
TimothyWright95: Kindove what i was thinking as well, ill make that change

forge-gui/src/main/java/forge/localinstance/properties/ForgeConstants.java
```diff
@@ -266,6 +264,7 @@ public final class ForgeConstants {
     public static final String DECK_TINY_LEADERS_DIR= DECK_BASE_DIR + "tiny_leaders" + PATH_SEPARATOR;
     public static final String DECK_BRAWL_DIR       = DECK_BASE_DIR + "brawl" + PATH_SEPARATOR;
     public static final String MAIN_PREFS_FILE      = USER_PREFS_DIR + "forge.preferences";
+    public static final String SERVER_PREFS_FILE    = USER_PREFS_DIR + "server.preferences";
```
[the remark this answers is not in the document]
TimothyWright95: The end goal would be to have a server preferencies block in the forge preferences GUI, but im not so good with GUI stuff, at work we pretty much just use javaFX for all our GUI stuff and it looks like a coder made it instead of a designer, which is fine since the applications are all internally facing but probably wont translate very well to doing GUI work in forge. 

I can look into it to see if theres an easy copy paste i can do to create a new server preferences block though and wire it all up to the new server preferences file.

forge-gui/pom.xml
tool4ever: #7367

Doesn't look like I can test with my router but this should work in theory now

forge-gui/src/main/java/forge/gamemodes/net/ProtocolMethod.java
tool4ever: well that class does combine System.out with a Logfile

forge-gui/pom.xml
TimothyWright95: ill try testing it with my router when I get home, had to set up UPnP to test the other library so afaik it should still work with the jupnp library as well

forge-gui/src/main/java/forge/gamemodes/net/ProtocolMethod.java
TimothyWright95: yep just noticed the redirect

        System.setOut(new PrintStream(new MultiplexOutputStream(System.out, logFileStream), true));

forge-gui/src/main/java/forge/localinstance/properties/ForgeConstants.java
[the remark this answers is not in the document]
TimothyWright95: Ok i was able to add the new server preferences portion to the preferences GUI
I did not see a localized combo box so i created one, that way non english speaking users can understand what the dropdown options represent. The backing preferences file will still use english constants so as to not cause issues if someone were to change the translation.

Since the server preferences are now separated into their own server.preferences file, it was necessary to make all the preference Enums implement a IPref interface so that they can be passed into the same function and later separated using instanceof

Jetz72 (approved):
[Approving not because I've tested this but because I personally have no further concerns; Github doesn't just let me mark my change requests resolved]

>TODO: (In scope)
The new server preferences section needs to be added to Android and IOS perhaps?

The mobile UI is available on desktop by launching Forge Adventure then selecting "Classic Mode".

tehdiplomat:
Gonna merge this. Will keep an eye on network reports in the next few days...

tehdiplomat (approved):

Issue #7012, opened 2025-02-13
Dependency Layer 6


Hanmac:
addChangedCardTraits
- [x] StaticAbilityContinous
- [ ] MANABOND CounterType (playtest) #7010
- [ ] AnimateEffectBase
- [x] Suspected (Can't Block) #6990
- [ ] PlayerKeywords

addChangedCardKeywords
- [x] StaticAbilityContinous
- [x] Suspected (Menace) #6990
- [ ] Mana Keywords (ManaSpent to cast Spells, give them Keywords)
- [ ] Ability CounterType
- [ ] PumpAllEffect
- [ ] PumpEffect
- [ ] DebuffEffect (when Protection from all Colors is split)
- [ ] ProtectAllEffect
- [ ] ProtectEffect
- [ ] ControlGainEffect
- [ ] ChangeZoneEffect (Uneath)
- [ ] CloneEffect and TokenEffectBase with PumpKeyword
- [ ] Bestow #6811

PR #7369, opened 2025-04-10
Unearth: add gain Haste as StaticLayer

Hanmac:
Part of #7012 

reduces the use of `addChangedCardKeywords` and uses Layer instead

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/ChangeZoneEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/ChangeZoneEffect.java
@@ -675,8 +675,6 @@ public class ChangeZoneEffect extends SpellAbilityEffect {
 
                     eff.addRemembered(movedCard);
 
-                    // Add forgot trigger
-                    addExileOnMovedTrigger(eff, "Battlefield");
                     movedCard.addLeavesPlayCommand(exileEffectCommand(game, eff));
 
                     game.getAction().moveToCommand(eff, sa);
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/ability/effects/ChangeZoneEffect.java
```diff
@@ -663,9 +663,26 @@ private void changeKnownOriginResolve(final SpellAbility sa) {
...
+                    String s = "Mode$ Continuous | Affected$ Card.IsRemembered | EffectZone$ Command | AddKeyword$ Haste";
+                    eff.addStaticAbility(s);
+
+                    // If it would leave the battlefield, exile it instead of putting it anywhere else.
+                    addLeaveBattlefieldReplacement(eff, "Exile");
+
+                    eff.addRemembered(movedCard);
+
+                    // Add forgot trigger
+                    addExileOnMovedTrigger(eff, "Battlefield");
```
tool4ever, line 679: adding both is redundant

Hanmac: What would you prefer? With only one defined Creature, the Command might be better?

tool4ever: yea then no trigger needs to be parsed and tested

tool4ever (approved):

PR #7370, opened 2025-04-10
Fix AI removing creature from combat when checking Endure AI during an attack

### LINKED ISSUE #7368: AI triggering Endure while attacking causes the Enduring creature to be removed from combat
**Describe the bug**
There are three cards in TDM that can Endure while attacking -- Descendant of Storms, Sinkhole Surveyor, and Inspirited Vanguard. When the AI attacks with these cards and they Endure, they are immediately removed from combat before moving to declare blockers, though the Endure trigger resolves. Other attacking creatures not Enduring are not removed from combat. This bug does not occur when the user triggers Endure on an attacking creature.

**To Reproduce**
1. Build the AI a deck with full playsets of the creatures listed above, full playsets of different relevant tutors, and lands.
2. Play a game against the AI and allow it to play the creatures and attack with them to get the Endure triggers.

**Expected behavior**
The AI should choose to put +1/+1 counters on the attacking creature, and you should take damage equal to that creature's new power after you declare no blockers.

 - OS: Windows 10
 - Version: 2.0.04-SNAPSHOT-04.07

**Additional context**
I'm still familiarizing myself with the code base, but I was able to determine that the attacking creature is being removed from combat when the AI calls canBeBlockedProfitably in the shouldPutCounters method of the EndureAi class. As far as I can tell, the issue is with the copy of the attacker the AI is making to simulate blocks. When it gets added to the simulated combat, something is causing the original attacking creature to leave the actual combat.

Agetian:
Fixes #7368.
The exact reason for this behavior is unknown, but copying the card with getLKICopy and then passing it to canBeBlockedProfitably removes the original from combat. Could it be that they are linked by ID? Using copyStats and assigning a new ID seems to do the job (the bug doesn't happen anymore), if a LKI copy would be preferred here, we need to test if getting a LKI copy and assigning a new ID would do the trick (not sure if there's a method for that, getLKICopy doesn't support it inherently).

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/EndureAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/EndureAi.java
@@ -51,7 +51,7 @@ public class EndureAi extends SpellAbilityAi {
 
         // need a copy for one with extra +1/+1 counter boost,
         // without causing triggers to run
-        final Card copy = CardCopyService.getLKICopy(source);
+        final Card copy = CardCopyService.copyStats(source, ai, true);
         copy.setCounters(CounterEnumType.P1P1, copy.getCounters(CounterEnumType.P1P1) + amount);
         copy.setZone(source.getZone());
 
--- a/forge-gui/res/lists/altwin-achievements.txt
+++ b/forge-gui/res/lists/altwin-achievements.txt
@@ -5,6 +5,7 @@ Azor's Elocutors|The Filibuster|Talk might be cheap, but it can buy you victory!
 Barren Glory|The Clean Slate|When you have nothing, you can lose nothing... so you can win everything!
 Battle of Wits|The Great Library|So many answers, so little time to look through them...
 Biovisionary|The Clique|And now my... I mean our plan is complete!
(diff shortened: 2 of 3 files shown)


Hanmac:
Lki need to keep the ID of the original for other effects

For this problem, I and @tool4ever need to investigate deeper

Agetian:
The same approach is used in TokenAi.willPayUnlessCost at line 423, should that be converted to copyStats as well in order to avoid this same issue? (Or another mechanism which would get a LKI copy and then assign a new ID, perhaps)

Agetian:
Yeah, this works as a stop-gap measure but there might be some kind of a deeper underlying issue with how it works, feel free to investigate further when able :+1:

Agetian:
Most likely, yeah, but I couldn't figure out a way to do it :D
Feel free to have at it, though ;)

Agetian:
It should, and it does, yes, but it looks like those methods called from within somehow check all the possible Combat objects during the given phase. I tried tracing through it, but couldn't determine where exactly this is happening (that other Combat objects get affected)

PR #7208, opened 2025-03-19
TDC: 4 cards

dracontes:
Tested to satisfaction:
- Transcendent Dragon
- Transforming Flourish
- Will of the Jeskai

Needing support:
- Voracious Bibliophile: The trigger works as far as the condition goes. For the resulting draw effect, I got as far as lifting the `SVar:X:TriggerObjectsSpellAbilityTargets$` line from [Arcee, Acrobatic Coupe](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/a/arcee_sharpshooter_arcee_acrobatic_coupe.txt) but the results are as follows on casting a [Donate](https://scryfall.com/card/uds/31/donate) (2 distinct targets) for the variations that seemed relevant:
• `TriggerObjectsSpellAbilityTargets$Amount`: draw 1 card
• `TriggerObjectsSpellAbilityTargets$Valid Any`: no card draw
Either option would be fine to implement: unfortunately my aptitude for Java doesn't stretch that far yet.

PR #7374, opened 2025-04-10
Cleanup: Optional wheel update

dracontes:
As discussed [here](https://github.com/Card-Forge/forge/pull/7208#discussion_r2005290115), updating existing optional wheel effects formatted as an `UnlessCost$` by following [Ruin Grinder](https://scryfall.com/card/c21/57/ruin-grinder) instead, to comply with relevant rulings. 
- [Imposing Grandeur](https://scryfall.com/card/voc/24/imposing-grandeur)
- [Sail into the West](https://scryfall.com/card/ltc/68/sail-into-the-west)
- [Will of the Jeskai](https://scryfall.com/card/tdc/40/will-of-the-jeskai)

Also updating: 
- Ruin Grinder so the selectable options more closely reflect its dies effect.
- Adding a `StackDescription$` to Imposing Grandeur and Sail into the West so their Stack items read a bit better.

### DIFF
--- a/forge-gui/res/cardsfolder/i/imposing_grandeur.txt
+++ b/forge-gui/res/cardsfolder/i/imposing_grandeur.txt
@@ -1,8 +1,13 @@
 Name:Imposing Grandeur
 ManaCost:4 R
 Types:Sorcery
-A:SP$ RepeatEach | RepeatPlayers$ Player | RepeatSubAbility$ DBDraw | SpellDescription$ Each player may discard their hand and draw cards equal to the greatest mana value of a commander they own on the battlefield or in the command zone.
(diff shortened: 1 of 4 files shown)


Hanmac:
i don't really like it, especially because AI doesn't really understand GenericChoice

but maybe that can be cleaned up in the future later by me and @tool4ever ?

tool4ever (approved):
ok for now since it's really shouldn't be a cost

Hanmac:
I still don't like it, and @Agetian might need to help with AI

Agetian:
Most likely needs an extension to GenericChoice AI, yeah. Might be able to look at it a little later.

PR #7378, opened 2025-04-11
Card: add CounterKeywordEffects

Hanmac:
not ready yet
just commited to be looked at by @tool4ever

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -64,6 +64,7 @@ import org.apache.commons.lang3.tuple.Triple;
 
 import java.util.*;
 import java.util.Map.Entry;
+import java.util.stream.Collectors;
 
 import static java.lang.Math.max;
 
@@ -169,6 +170,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
     private final Multimap<Long, Keyword> cantHaveKeywords = MultimapBuilder.hashKeys().hashSetValues().build();
 
     private final Map<CounterType, Long> counterTypeTimestamps = Maps.newHashMap();
+    private final Map<CounterType, Card> counterTypeKeywordEffects = Maps.newHashMap();
 
     private final Map<Long, Integer> canBlockAdditional = Maps.newTreeMap();
     private final Set<Long> canBlockAny = Sets.newHashSet();
@@ -1793,7 +1795,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
             view.updateCounters(this);
         }
         if (newValue <= 0) {
-            removeCounterTimestamp(counterType);
+            removeCounterTimestamp(counterType, false);
         } else if (addCounterTimestamp(counterType, false)) {
             updateKeywords();
         }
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -1802,12 +1804,52 @@ public void addCounterInternal(final CounterType counterType, final int n, final
...
+            getGame().getAction().moveToCommand(result, null);
+
+            counterTypeKeywordEffects.put(counterType, result);
+        }
+
+        // update timestamp
+        result.setGameTimestamp(timestamp);
+
+        // update amount
+        for (StaticAbility stAb : result.getStaticAbilities()) {
```
tool4ever, line 1843: this can be else?

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -1827,28 +1869,25 @@ public boolean addCounterTimestamp(CounterType counterType, boolean updateView)
...
-        return removeCounterTimestamp(counterType, true);
-    }
     public boolean removeCounterTimestamp(CounterType counterType, boolean updateView) {
         Long old = counterTypeTimestamps.remove(counterType);
         if (old != null) {
             removeChangedCardTypes(old, 0, updateView);
             removeChangedCardTraits(old, 0);
-            removeChangedCardKeywords(old, 0, updateView);
+        }
+        if (!counterType.isKeywordCounter()) {
```
tool4ever, line 1886: logic error?

forge-game/src/main/java/forge/game/card/Card.java
Hanmac: My goal was to add something to StaticAbilityContinous that would Multiply the AddKeywords,
so i just would need to update the SVar on the Effect.

But i didn't found the right way yet

Hanmac:
@tool4ever right now, LKI copy causes StackOverflowError

i need something like this #7356 to prevent LKI from making new Cards in the CommandZone

tool4ever:
we use LKI for
**a.** look back in time (and strictly only this case is backed by rules)
**b.** look into the future

maybe there needs to be more unique logic for the desired direction?

Hanmac:
> we use LKI for **a.** look back in time (and strictly only this case is backed by rules) **b.** look into the future
> 
> maybe there needs to be more unique logic for the desired direction?

The Problem is more like with my other changes, Card objects point to other Cards in the Command Zone
and for their effects to apply correct, i would need to make LKI copies of these ones too

Hanmac:
@tool4ever i had an idea to fix a big chunk of our Dependency Static Ability problem
(i will use it for this ones first)

First i was thinking about turning StaticAbility into an Interface, so a fake-one can be used for the Dependency System without defining a real StaticAbility.
But this has too many compications (like turning CardTraits into an Interface too)

So instead of creating Effect Cards that pollute the command zone, 
how about adding the StaticAbilities that causes such changes (like the AnimateEffect part of Amass) or Keyword Counters into a Static Ability that is hidden from the other ones?
Like they are not in `getStaticAbilities` so they can't be removed
It doesn't cause Card Effects in the CommandZone, so it doesn't cause problems with LKI
And all other places that look for StaticAbilities (and their Effects), could add these too?

The only difference is to extend `getLayerTimestamp` and move it to StaticAbility where i can overwrite it for special cases

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -1817,67 +1817,39 @@ public void addCounterInternal(final CounterType counterType, final int n, final
...
+            result = counterTypeKeywordStatic.computeIfAbsent(counterType, ct -> {
+                String s = "Mode$ Continuous | AffectedDefined$ Self | AddType$ Land | RemoveCardTypes$ True | RemoveSubTypes$ True | RemoveAllAbilities$ True | AddAbility$ ManaReflected";
+                StaticAbility stAb = StaticAbility.create(s, this, currentState, true);
+                String abStr = "AB$ ManaReflected | Cost$ T | Valid$ Defined.Self | ColorOrType$ Color | ReflectProperty$ Is | SpellDescription$ Add one mana of any of this card's colors.";
+                stAb.setSVar("ManaReflected", abStr);
+                return stAb;
+            });
+        } else if (counterType.isKeywordCounter()) {
+            result = counterTypeKeywordStatic.computeIfAbsent(counterType, ct -> {
+                return StaticAbility.create("Mode$ Continuous | AffectedDefined$ Self | AddKeyword$ " + ct.toString(), this, currentState, true);
```
tool4ever: it's supposed to work in all zones

PR #7379, opened 2025-04-11
JUPnP android fix

tool4ever:


Hanmac:
Imo might be better if that part gets overwritten by the -android lib?
I don't like it when the non-android poms depend on android libs

tool4ever:
hmn good point, that means I need to extract the logic from forge-gui right?

Hanmac:
Maybe add a `GuiBase.getUpnpServiceConfiguration` ?

PR #7385, opened 2025-04-11
Svar fallback changes

### LINKED ISSUE #7383: Lier + X spells broken
Attempting to cast e.g. _Fireball_ checks the wrong CardState for the SVars.

The problem is the change in `CardTraitBase.getSVarFallback()` done by #6079.

Hanmac:
@tool4ever can you check this out?

And if possible, add a TestCase for Flashback + Adventure?

Closes #7383

### DIFF
--- a/forge-game/src/main/java/forge/game/CardTraitBase.java
+++ b/forge-game/src/main/java/forge/game/CardTraitBase.java
@@ -2,12 +2,14 @@ package forge.game;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import org.apache.commons.lang3.ArrayUtils;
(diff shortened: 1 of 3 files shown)


Hanmac:
@tool4ever it works for Fireball, what do you think about the code? can it be done better?

tool4ever (approved):
I tried to find some SVar collision but it seems to work
Not sure we can do it cleaner currently

Hanmac:
@tool4ever i don't really like `CardTraitBase.getSVars()` change, maybe i try to remove this function (it is only really used by CreateEffect)

same with `getDirectSVars()` it is only used by CharmOrder

PR #7389, opened 2025-04-12
removed getDirectSVars

Hanmac:
I removed `getDirectSVars` and did some cleanup

one other thing i would like to remove is in createEffect, the call for `eff.setSVars(sa.getSVars());`

This is the reason why this still exist:
```java

    @Override
    public Map<String, String> getSVars() {
        Map<String, String> res = Maps.newHashMap();
        // TODO reverse the order
        for (IHasSVars s : getSVarFallback()) {
            res.putAll(s.getSVars());
        }
        res.putAll(sVars);
        return res;
    }
```

so what if Effect just uses `effectSourceAbility` instead? Or is the problem in case the Svars get cleaned up?

### DIFF
--- a/forge-game/src/main/java/forge/game/cost/CostAdjustment.java
+++ b/forge-game/src/main/java/forge/game/cost/CostAdjustment.java
@@ -128,14 +128,10 @@ public class CostAdjustment {
         } else if (st.hasParam("Amount")) {
             String amount = st.getParam("Amount");
             if ("Escalate".equals(amount)) {
-                SpellAbility sub = sa;
-                while (sub != null) {
-                    if (sub.getDirectSVars().containsKey("CharmOrder")) {
-                        count++;
-                    }
-                    sub = sub.getSubAbility();
+                SpellAbility tail = sa.getTailAbility();
+                if (tail.hasSVar("CharmOrder")) {
+                    count = tail.getSVarInt("CharmOrder") - 1;
                 }
-                --count;
             } else if ("Strive".equals(amount)) {
                 for (TargetChoices tc : sa.getAllTargetChoices()) {
                     count += tc.size();
--- a/forge-game/src/main/java/forge/game/CardTraitBase.java
+++ b/forge-game/src/main/java/forge/game/CardTraitBase.java
@@ -607,7 +607,7 @@ public abstract class CardTraitBase extends GameObject implements IHasCardView, 
     }
 
     @Override
-    public final void setSVar(final String name, final String value) {
+    public void setSVar(final String name, final String value) {
         sVars.put(name, value);
     }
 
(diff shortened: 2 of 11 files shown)


forge-game/src/main/java/forge/game/cost/CostAdjustment.java
```diff
@@ -128,14 +128,10 @@ private static void applyRaise(final Cost cost, final SpellAbility sa, final Sta
         } else if (st.hasParam("Amount")) {
             String amount = st.getParam("Amount");
             if ("Escalate".equals(amount)) {
-                SpellAbility sub = sa;
-                while (sub != null) {
-                    if (sub.getDirectSVars().containsKey("CharmOrder")) {
-                        count++;
-                    }
-                    sub = sub.getSubAbility();
+                SpellAbility tail = sa.getTailAbility();
```
tool4ever, line 131: I don't think this works:
if one mode is effects A -> B -> C your logic will not find it?

Hanmac: CharmOrder is set for sorting, so the Bottom of the SpellAbility Chain should have the highest CharmOrder

using getTailAbility to get to them Bottom, and using getSVar, sould look up for the next SpellAbility with CharmOrder param and return it

tool4ever: you will only check the direct parent B and miss the SVar on A

tool4ever: I really don't see how calling `C.hasSVar` would reach A:

```
    @Override
    public boolean hasSVar(final String name) {
        return sVars.containsKey(name) || findSVar(name).isPresent();
    }
```

```
    protected Optional<IHasSVars> findSVar(final String name) {
        return getSVarFallback().stream().filter(f -> f.hasSVar(name)).findFirst();
    }
```

```
    protected List<IHasSVars> getSVarFallback() {
        List<IHasSVars> result = Lists.newArrayList();
        if (getParent() != null) {
            result.add(getParent());
        }
        result.addAll(super.getSVarFallback());
        return result;
    }
```

Hanmac: C.hasSVar calls B.hasSVar which calls A.hasSVar

C.hasSVar fails to find it in sVars
then calls C.findSVar which calls C.getSVarFallback
which calls C.getParent to get B

then it calls B.hasSVar fails to find it in sVars
then calls B.findSVar which calls B.getSVarFallback
which calls B.getParent to get A

then it calls A.hasSVar which finds it in sVars

tool4ever (approved):

Hanmac:
@tool4ever can we do something about `eff.setSVars(sa.getSVars());` ?
or for later MR?

PR #7393, opened 2025-04-13
CantUntap Third Step

tool4ever:
Patches `your next`, so remaining usage of the hidden KW should be reduced to only the types `player's|controller's/controllers' next`, making the final refactor a bit easier later.


Hanmac:
I don't know about `ExileEff` why can't it be part of the duration?

tool4ever:
guess my brain is a bit weird lately, looks cleaner now

Hanmac (approved):

PR #7400, opened 2025-04-14
AttachAI: remove getFirstAttachSpell

Hanmac:
removes the last use of getFirstAttachSpell


Agetian (dismissed):

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -2777,19 +2777,15 @@ private boolean attachAuraOnIndirectETB(final Card source, Map<AbilityKey, Objec
             return false;
         }
 
-        SpellAbility aura = source.getFirstAttachSpell();
```
tool4ever, line 2780: Why do you want to build extra SA when we could use the real one in most cases? 🤔

Hanmac: Because I want to get rid of getFirstAttachSpell function

tool4ever: Hmn, I guess I'm not seeing what's wrong with it...so should the logic be moved into this function instead then?

Hanmac: i try to remove unnecessary functions instead, and it is only used for AI, and it should be happy with EmptySA too

PR #7408, opened 2025-04-16
Fix always targeting Graveyard with Aura

### LINKED ISSUE #7406: Casting an aura opens the graveyard selection
**Describe the bug**
I don't know if this started to happen before the 4/15 snapshot, but the 4/10 snapshot didn't have this issue. When you cast an aura the graveyard selection opens.

**To Reproduce**
1. Cast an aura (one that doesn't enchant something in the graveyard)
2. See the graveyard selection pop up.

**Expected behavior**
For the graveyard selection to only open if the aura in question can enchant something there.

 - Forge Version: forge-installer-2.0.04-SNAPSHOT-04.15

tool4ever:
Closes #7406

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardFactory.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactory.java
@@ -443,7 +443,6 @@ public class CardFactory {
             } else if (c.isAura()) {
                 String desc = "";
                 String extra = "";
-                boolean grave = false;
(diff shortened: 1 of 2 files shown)


Hanmac:
CanBeEnchantedBy should have handled the Graveyard from lighting up

Also doesn't really work that way if cards gain Enchant keywords

Hanmac (approved):

PR #7417, opened 2025-04-17
Bug fix for previous PR that introduced a GUI thread block

TimothyWright95:
Fixed mobile server host UPnP dialog from accidentally blocking the GUI thread indefinitely


TimothyWright95:
Previous PR is #7365

tool4ever (approved):

PR #7418, opened 2025-04-17
Feature/network improvement android

TimothyWright95:
Added the new server hosting options to the android settings screen.
Added new IntegerSelectSetting which is used by the server port setting.
Added new LocalizedSelectSetting which is used by the UPnP setting.


tool4ever (approved):
a bit tougher to review with all the autoformatting...
also not convinced these extra spaces really help with readability, though it seems less of a factor with such GUI class 🤷‍♂️

TimothyWright95:
> a bit tougher to review with all the autoformatting... also not convinced these extra spaces really help with readability, though it seems less of a factor with such GUI class 🤷‍♂️

Ya sorry about that I must have hit auto format in intelij or something

PR #7433, opened 2025-04-20
[Adventure] Quest order fix

Drecon84:
I changed this number to test some stuff but apparently my attempt to change it back before committing was unsucessful earlier. 
Currently the main quest is out of order, this will put things right again.


tehdiplomat (approved):

tehdiplomat:
Part of me feels like quests.json needs to be broken into different files so its easier to see how the quests change.

Drecon84:
> Part of me feels like quests.json needs to be broken into different files so its easier to see how the quests change.

I had been thinking the same thing. It's getting a place on my ever-growing list of things I would like to do something about.

PR #7435, opened 2025-04-20
Update cards with Static CantAttack or CantBlock

Hanmac:
Part of #3307

Removed `K:CARDNAME can't block.`

and updated "S:Mode"  Statics that did add `CARDNAME can't attack or block.` & `CARDNAME's activated abilities can't be activated.`

i didn't updated Cards yet that does this via Pump/Animate/Effect

`Volrath's Curse` and `Lost in Thought` are still tricky

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/AttachAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/AttachAi.java
@@ -834,7 +834,27 @@ public class AttachAi extends SpellAbilityAi {
         int totPower = 0;
         final List<String> keywords = new ArrayList<>();
 
+        boolean cantAttack = false;
+        boolean cantBlock = false;
+
         for (final StaticAbility stAbility : attachSource.getStaticAbilities()) {
+            if (stAbility.checkMode(StaticAbilityMode.CantAttack)) {
+                String valid = stAbility.getParam("ValidCard");
+                if (valid.contains(stCheck) || valid.contains("AttachedBy")) {
+                    cantAttack = true;
+                }
+            } else if (stAbility.checkMode(StaticAbilityMode.CantBlock)) {
+                String valid = stAbility.getParam("ValidCard");
+                if (valid.contains(stCheck) || valid.contains("AttachedBy")) {
+                    cantBlock = true;
+                }
+            } else if (stAbility.checkMode(StaticAbilityMode.CantBlockBy)) {
+                String valid = stAbility.getParam("ValidBlocker");
+                if (valid.contains(stCheck) || valid.contains("AttachedBy")) {
+                    cantBlock = true;
+                }
+            }
+
             if (!stAbility.checkMode(StaticAbilityMode.Continuous)) {
                 continue;
             }
@@ -886,6 +906,12 @@ public class AttachAi extends SpellAbilityAi {
             prefList = CardLists.filter(prefList, c -> containsUsefulCurseKeyword(keywords, c, sa));
         } else if (totPower < 0) {
             prefList = CardLists.filter(prefList, c -> c.getNetPower() > 0 && ComputerUtilCombat.canAttackNextTurn(c));
+        }
+
+        if (cantAttack) {
+            prefList = CardLists.filter(prefList, c -> c.isCreature() && ComputerUtilCombat.canAttackNextTurn(c));
+        } else if (cantBlock) { // TODO better can block filter?
+            prefList = CardLists.filter(prefList, c -> c.isCreature() && !ComputerUtilCard.isUselessCreature(ai, c));
         }
 
         //some auras aren't useful in multiples
--- a/forge-gui/res/cardsfolder/a/akron_legionnaire.txt
+++ b/forge-gui/res/cardsfolder/a/akron_legionnaire.txt
@@ -2,7 +2,7 @@ Name:Akron Legionnaire
 ManaCost:6 W W
 Types:Creature Giant Soldier
 PT:8/4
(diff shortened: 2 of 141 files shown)


tool4ever (dismissed):

Hanmac:
@tool4ever @tehdiplomat 
`Volrath's Curse` and `Lost in Thought` made me unsure if this is the right way

should we do "CantAttack$ True" & "CantBlock$" and "CantActivate$ Activated" instead? :thinking: 
(i know the main goal was to reduce the amount of StaticAbilityContinous Layer runs?)

otherwise we might need to reconfig how "IgnoreEffect" works for these two :/

tool4ever:
> should we do "CantAttack$ True" & "CantBlock$" and "CantActivate$ Activated" instead?

that seems wrong, could as well just leave it as hidden KW in that case

> otherwise we might need to reconfig how "IgnoreEffect" works for these two :/

none of the IgnoreEffectCost$ should need to use Continuous

Hanmac:
> > should we do "CantAttack$ True" & "CantBlock$" and "CantActivate$ Activated" instead?
> 
> that seems wrong, could as well just leave it as hidden KW in that case

I still want to remove hidden keywords and replace them with something more dynamic


> 
> > otherwise we might need to reconfig how "IgnoreEffect" works for these two :/
> 
> none of the IgnoreEffectCost$ should need to use Continuous

you got any better idea in changing this?
```
S:Mode$ Continuous | Affected$ Creature.EnchantedBy | AddHiddenKeyword$ CARDNAME can't attack or block. & CARDNAME's activated abilities can't be activated. | IgnoreEffectCost$ Sac<1/Permanent> | Description$ Enchanted creature can't attack or block, and its activated abilities can't be activated. That creature's controller may sacrifice a permanent for that player to ignore this effect until end of turn.
```

my current attempt would split them into three static abilities, which then `IgnoreEffectCost` doesn't work correctly anymore

if i would use "CantAttack$ True" & "CantBlock$" and "CantActivate$ Activated", then it could stay one static

tool4ever:
refactoring (or even duplicating) statics back into rules layer is a bad trade-off

why not make  `checkMode` use `contains` instead of `equals`:
➡️ `Mode$ CantAttack,CantBlock,CantBeActivated`

way more dynamic, since it'll also allow other modes to be combined freely (as long as the params don't clash)

Hanmac:
> refactoring (or even duplicating) statics back into rules layer is a bad trade-off
> 
> why not make `checkMode` use `contains` instead of `equals`: ➡️ `Mode$ CantAttack,CantBlock,CantBeActivated`
> 
> way more dynamic, since it'll also allow other modes to be combined freely (as long as the params don't clash)

Because CantBlock**By** right now uses different params
And CantBeActivated often checks for nonManaAbilities

tool4ever:
so what? you would just add the params needed (for each mode)

Hanmac:
> so what? you would just add the params needed (for each mode)

I will try something with extra CantBlock Static Mode


but you might need to help me with looking into these, I will make extra MR for them later

* "CARDNAME can't attack or block alone"
* "CARDNAME can't block unless at least two other creatures block."
* "CARDNAME can't block unless a creature with greater power also blocks."

tehdiplomat:
Have we considered something like this:

`S:Mode$ MultiMode | Modes$ CantAttack,CantBlock,CantBeActivated |  | IgnoreEffectCost$ Sac<1/Permanent> | Description$ Enchanted creature can't attack or block, and its activated abilities can't be activated. That creature's controller may sacrifice a permanent for that player to ignore this effect until end of turn.`

And then chain the Modes that way through SVars?

Hanmac:
@tool4ever @tehdiplomat this is an update to remove the Hidden Keywords

or should i try to do MultiMode with `CantAttack,CantBlock` with this PR?

forge-ai/src/main/java/forge/ai/ability/AttachAi.java
```diff
@@ -833,7 +833,27 @@ private static Card attachAICursePreference(final SpellAbility sa, final List<Ca
         int totPower = 0;
         final List<String> keywords = new ArrayList<>();
 
+        boolean cantAttack = false;
+        boolean cantBlock = false;
+
         for (final StaticAbility stAbility : attachSource.getStaticAbilities()) {
+            if (stAbility.getParam("Mode").equals(StaticAbilityCantAttackBlock.CantAttackMode)) {
+                String valid = stAbility.getParam("ValidCard");
+                if (valid.contains(stCheck) || valid.contains("AttachedBy")) {
```
tool4ever, line 842: I wonder if we should just remove Enchanted vs. EquippedBy
technically it shouldn't matter for these cases:
> 301.5f. An ability of a permanent that refers to the "equipped creature" refers to whatever creature that permanent is attached to, even if the permanent with the ability isn't an Equipment.

tool4ever:
is there another case where it matters that it's kept as single ability?

forge-ai/src/main/java/forge/ai/ability/AttachAi.java
Hanmac: maybe yes, but i still think it helps the AI for now

we can clean that up later

Hanmac:
> is there another case where it matters that it's kept as single ability?

mostly for `Volrath's Curse` and `Lost in Thought`

i will test them soon

forge-ai/src/main/java/forge/ai/ability/AttachAi.java
Hanmac: @tool4ever what we could do is to cleanup the Card Properties there, and their negations:

https://github.com/Card-Forge/forge/blob/194662c9c7155a52513936bd9b277087b621440b/forge-game/src/main/java/forge/game/card/CardProperty.java#L1373-L1396

tool4ever (approved):

Issue #7288, opened 2025-03-28
Update Token Images List into a Token to Scryfall Mapping


Hanmac:
https://github.com/Card-Forge/forge/blob/878da9b06f3c544150af751b2b52ffd5174d4d65/forge-gui/res/lists/token-images.txt#L932-L937

Update the List into something like this?
```
g_2_2_ooze https://scryfall.com/card/tdsc/13/ooze
g_2_2_ooze_mitotic https://scryfall.com/card/tncc/26/ooze
g_2_2_ooze_mitotic_m11 https://scryfall.com/card/tm11/5/ooze
g_2_2_ooze_mitotic_ncc https://scryfall.com/card/tncc/26/ooze
g_2_2_ooze_mitotic_pca https://scryfall.com/card/tpca/17/ooze
g_2_2_ooze_rna https://scryfall.com/card/trna/7/ooze
g_2_2_ooze_dsc https://scryfall.com/card/tdsc/13/ooze
```
Or maybe even shorter? We could guess from the set to the token set? `dsc` → `tdsc`, we just need to define the CollectorNumber?
(or can we get the collector number from the Set?)

```
g_2_2_ooze_mitotic_m11 Ooze#5
g_2_2_ooze_mitotic_ncc  Ooze#26
g_2_2_ooze_mitotic_pca  Ooze#17
g_2_2_ooze_rna  Ooze#7
g_2_2_ooze_dsc  Ooze#13
```

`https://scryfall.com/card/tdsc/13` does redirect to `https://scryfall.com/card/tdsc/13/ooze`

or do we use Scryfall UUID?
`https://scryfall.com/card/786e11b0-1b63-4243-a16d-e75caebe20d7` does redirect to `https://scryfall.com/card/tdsc/13/ooze`

Or I the Future, if the Set files are updated with CollectorNumber for Tokens, then we could get rid of  the Token Images file

Commander Sets might have Regular Set Tokens
So we might need some kind of fallback

Jetz72:
Dropping `token-images.txt` and defining the collector numbers in the edition file seems like a good direction. Linking our sets to Scryfall's token sets is a complication, and I don't like blindly assuming that the token set code is just gonna be the same set code prefixed by a 'T'. 

A simple approach would be to just include the set code in the line for the token script:
```
c_1_1_a_myr TTDC 30
c_1_1_a_servo TTDC 31
c_1_1_a_soldier TTDC 32
```

[...]

PR #7439, opened 2025-04-21
CardEdition: add TokenInSet

Hanmac:
First part for #7288 

This allows the Token Scripts to be added with CollectorNumber and Artist in the CardEdition

next steps are TokenDb and PaperToken

### DIFF
--- a/forge-core/src/main/java/forge/card/CardEdition.java
+++ b/forge-core/src/main/java/forge/card/CardEdition.java
@@ -165,18 +165,60 @@ public final class CardEdition implements Comparable<CardEdition> {
         }
     }
 
+    private static final Map<String, String> sortableCollNumberLookup = new HashMap<>();
+    /**
+     * This method implements the main strategy to allow for natural ordering of collectorNumber
+     * (i.e. "1" < "10"), overloading the default lexicographic order (i.e. "10" < "1").
+     * Any non-numerical parts in the input collectorNumber will be also accounted for, and attached to the
+     * resulting sorting key, accordingly.
+     *
+     * @param collectorNumber: Input collectorNumber tro transform in a Sorting Key
+     * @return A 5-digits zero-padded collector number + any non-numerical parts attached.
+     */
+    public static String getSortableCollectorNumber(final String collectorNumber){
+        String inputCollNumber = collectorNumber;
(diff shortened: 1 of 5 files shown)


forge-core/src/main/java/forge/card/CardEdition.java
```diff
@@ -165,41 +165,15 @@ public static List<String> getNames() {
         }
     }
 
-    public static class CardInSet implements Comparable<CardInSet> {
-        public final CardRarity rarity;
+    public static class InSet {
```
tehdiplomat, line 168: I wonder if we should be pulling these out of this file now that there's 3 subclasses? Also, I was looking at this briefly last month, and IntelliJ suggests converting these to `Record` types. Which removes a lot of this boilerplate.

tehdiplomat (commented):
I might push a commit to this branch in relation to TokenCode if I can find the stash I had when i was fiddling around with that.

Hanmac:
> I might push a commit to this branch in relation to TokenCode if I can find the stash I had when i was fiddling around with that.

Feel free to update more of the stuff

Like I couldn't got the TokenInSet into PaperToken yet

forge-core/src/main/java/forge/card/CardEdition.java
Hanmac: Ah interesting 🤔 

If it works for Android, I can try to change them into records (with or without base class?)

It's just that I need to find a new place for the CollectorNumber function

Hanmac: changed into records

Hanmac: Records might not work with Android 11 (my phone, for example)

@Agetian can you check the Android build for this?

Hanmac: @tehdiplomat looks like the android build tools can backport it? Or is that something you can only test with runtime?

tehdiplomat: Lets just pull it out of this PR for now. We can run a different PR with the upgrade later and fool around with how functional it is in android.

tehdiplomat (approved):

PR #7452, opened 2025-04-25
Add CantBlock Mode

Hanmac:
Rework Part for #7435

so in the future when we have Multiple Modes Combined we can use ValidCard

AI part will be added to a later date after some more rework

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -6765,7 +6765,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
 
             String s = "Mode$ Continuous | AffectedDefined$ RememberedCard | EffectZone$ Command | AddKeyword$ Menace | AddStaticAbility$ SuspectedCantBlockBy";
             StaticAbility suspectedStatic = suspectedEffect.addStaticAbility(s);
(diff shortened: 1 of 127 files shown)


tool4ever:
imo adding another static for this just complicates things
- either CantBlockBy should just check ValidCard$
  - directly
  - or as fallback
- or when multiple modes get used it wouldn't be a big deal to just combine them as needed e.g. both ValidCard$ + ValidAttacker$

Hanmac:
YOU wanted an extra Mode

> either CantBlockBy should just check ValidCard$

But that complicates things. if CantBlockBy only has one ValidCard, does it check for the Attacker or the Blocker?

What about the one that has "Cardname can't be blocked" or better "Cardname can't block and can't be blocked"?

tool4ever:
> YOU wanted an extra Mode

that was a misunderstanding

> But that complicates things. if CantBlockBy only has one ValidCard, does it check for the Attacker or the Blocker?
 
fair enough, imo combining the params would still be easier then but maybe this one needs a POC that it will work first though

> What about the one that has "Cardname can't be blocked" or better "Cardname can't block and can't be blocked"?

hmn, the second case is indeed tricky - `ValidAttackerOrBlocker$`? 🤔

Hanmac:
@tehdiplomat can you review this?

It's kinda blocking for my #7452 → #7435 → #7209 pipeline

tehdiplomat:
Seems fine to me.

tool4ever (approved):

PR #7459, opened 2025-04-27
Card: add FacedownImageKey for better logic and set codes

Hanmac:
This adds set codes to different face-down images

in the future, this can be used for set specific token images

### DIFF
--- a/forge-ai/src/main/java/forge/ai/GameState.java
+++ b/forge-ai/src/main/java/forge/ai/GameState.java
@@ -13,6 +13,7 @@ import forge.card.mana.ManaAtom;
 import forge.game.Game;
 import forge.game.GameEntity;
 import forge.game.ability.AbilityFactory;
+import forge.game.ability.ApiType;
 import forge.game.ability.effects.DetachedCardEffect;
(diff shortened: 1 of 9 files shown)


Jetz72:
Toward the end of the function you might even add a switch on a card's GamePieceType, and add the unique card backs for Planechase, Scheme, Attraction, and Contraption cards.

Hanmac:
> Toward the end of the function you might even add a switch on a card's GamePieceType, and add the unique card backs for Planechase, Scheme, Attraction, and Contraption cards.

That's maybe something for later MR
Also i don't know if scryfall has images for this

Jetz72:
> That's maybe something for later MR
> Also, I don't know if scryfall has images for this

It does - there's a button under cards of each of those types to view the back, but yeah I suppose that can wait.

tehdiplomat:
Yea seems fine. I'd agree that other card backs can be managed in a different way. I think the biggest question for those is where do we load the cardbacks from. How does that interact with card sleeves?

tehdiplomat (approved):

PR #7475, opened 2025-05-02
Support for non-Siege Battles

Jetz72:
Not currently used by any official cards but the rules define their behavior to an extent and it was requested to support some custom cards. This doesn't fully support custom types - the engine won't even acknowledge unknown subtypes, much less handle their intrinsic abilities or state based actions. But this does enable various designs that are protected by their controllers.

### DIFF
--- a/forge-core/src/main/java/forge/card/CardType.java
+++ b/forge-core/src/main/java/forge/card/CardType.java
@@ -26,6 +26,7 @@ import org.apache.commons.lang3.StringUtils;
 
 import java.util.*;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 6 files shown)


tehdiplomat:
Looks like your test build is failing

> Error:  /home/runner/work/forge/forge/forge-game/src/main/java/forge/game/card/CardFactory.java:21:8: Unused import - com.google.common.collect.Iterables. [UnusedImports]

Jetz72:
Gah. Deleted a fallback branch at the last second because I realized I'd made a wrong assumption writing it. Forgot to re-check for imports. Should be clean now.

tehdiplomat (approved):

PR #7479, opened 2025-05-02
PaperToken: add CollectorNumber and update Downloader

Hanmac:
- Adds CollectorNumber to PaperToken
- Updates Token ImageKey to something more parseable, like with PaperCard
- Updates Downloader to use correct CollectorNumber if possible

This Updates the resulting the filename in the folder be be changed:
from `c_a_clue_draw2_mkm.jpg` with idx = 2
to `c_a_clue_draw_mkm_17.jpg` with nr = 17

Or should the filename to be changed to `mkn/17_c_a_clue_draw.jpg` with set based subfolder and Nr first?

### DIFF
--- a/forge-core/src/main/java/forge/token/TokenDb.java
+++ b/forge-core/src/main/java/forge/token/TokenDb.java
@@ -75,6 +75,17 @@ public class TokenDb implements ITokenDatabase {
         return new PaperToken(rulesByName.get(name), edition, name, t.collectorNumber, t.artistName);
     }
 
+    // try all editions to find token
+    protected PaperToken fallbackToken(String name) {
+        for (CardEdition edition : this.editions) {
+            String fullName = String.format("%s_%s", name, edition.getCode().toLowerCase());
+            if (loadTokenFromSet(edition, name)) {
+                return Aggregates.random(allTokenByName.get(fullName));
+            }
+        }
+        return null;
+    }
+
     @Override
     public PaperToken getToken(String tokenName) {
         return getToken(tokenName, CardEdition.UNKNOWN.getCode());
@@ -88,6 +99,10 @@ public class TokenDb implements ITokenDatabase {
         // token exist in Set, return one at random
         if (loadTokenFromSet(realEdition, tokenName)) {
             return Aggregates.random(allTokenByName.get(fullName));
+        }
+        PaperToken fallback = this.fallbackToken(tokenName);
+        if (fallback != null) {
+            return fallback;
         }
 
         if (!extraTokensByName.containsKey(fullName)) {
--- a/forge-core/src/main/java/forge/token/TokenDb.java
+++ b/forge-core/src/main/java/forge/token/TokenDb.java
@@ -1,10 +1,15 @@
 package forge.token;
 
+import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Multimap;
+
 import forge.card.CardDb;
 import forge.card.CardEdition;
 import forge.card.CardRules;
+import forge.item.IPaperCard;
 import forge.item.PaperToken;
+import forge.util.Aggregates;
 
 import java.util.*;
 import java.util.function.Predicate;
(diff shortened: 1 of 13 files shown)


tehdiplomat (dismissed):

Hanmac:
@tehdiplomat your opinion on totally new token name by set folder?
> Or should the filename to be changed to mkn/17_c_a_clue_draw.jpg with set based subfolder and Nr first?

Hanmac:
There is also currently a Problem with Sets that doesn't define the Token,
but also doesn't have a fallback set.

This happens for some sets like Commander, but also for some missing token.
Like banana from "Kibo, Uktabi Prince" still doesn't have a Token Image

forge-core/src/main/java/forge/token/TokenDb.java
```diff
@@ -38,38 +43,79 @@ public boolean containsRule(String rule) {
...
+        String fullName = String.format("%s_%s", tokenName, realEdition.getCode().toLowerCase());
+
+        // token exist in Set, return one at random
+        if (loadTokenFromSet(realEdition, tokenName)) {
+            return Aggregates.random(allTokenByName.get(fullName));
+        }
+        PaperToken fallback = this.fallbackToken(tokenName);
+        if (fallback != null) {
+            return fallback;
+        }
```
Hanmac, line 106: @tehdiplomat @tool4ever i added a fallback in case the Token isn't found in the right set so it tries any other set

i don't know yet if that makes the last part with "extraTokenByName" obsolete

Hanmac:
Something funky with Specialize i found:

PaperCard does define extra functions:
https://github.com/Card-Forge/forge/blob/71256972c4685cbfd73c055aa6d55fa6e3952f45/forge-core/src/main/java/forge/item/PaperCard.java#L398-L402

But CardFactory ignores them and uses something different?
https://github.com/Card-Forge/forge/blob/71256972c4685cbfd73c055aa6d55fa6e3952f45/forge-game/src/main/java/forge/game/card/CardFactory.java#L217-L219

But then Card does reset the Imagekey in the CardState anyway and resets them with the ones from PaperCard:
https://github.com/Card-Forge/forge/blob/71256972c4685cbfd73c055aa6d55fa6e3952f45/forge-game/src/main/java/forge/game/card/Card.java#L6528

while Desktop ImageCache somehow messes with both:
https://github.com/Card-Forge/forge/blob/71256972c4685cbfd73c055aa6d55fa6e3952f45/forge-gui-desktop/src/main/java/forge/ImageCache.java#L175-L200


i need to debug what is going on there

Hanmac:
Some funk with DMU and DMC tokens i found:

> Dominaria United series of 26 tokens
> Dominaria United comes with two sets of tokens: a series of 26 bearing either DMU or DMC set codes, and a separate series of 12 tokens with exclusively the DMC set code. We don't know why they did things this way. We've opted to file the entire series of 26 here regardless of set code to avoid archival conflicts.

so https://scryfall.com/card/tdmu/7/insect is under TDMU while still having DMC, to not collide with https://scryfall.com/card/tdmc/7/bear which also has DMC

tehdiplomat:
Yea i think replicating how we treat set images is good. So having the set code folder. What's the advantage of putting the collector number?

Hanmac:
> Yea i think replicating how we treat set images is good. So having the set code folder. What's the advantage of putting the collector number?

For example are treasure token
Imo `Set/CN_treasure` is more clean than `Set/treasureIdx`

Hanmac:
@tehdiplomat it seems with FIN, there are separate CollectorNumber for Foil Token

https://scryfall.com/card/tfin/5/hero
https://scryfall.com/card/tfin/29/hero (notice the little ★ at the bottom)

(I saw some different Token with a Star as Part of the Collector Number, but I need to search it again)

might be useful for later MR if we want to have `PaperToken#isFoil` return something else

Jetz72:
> Scryfall says that cn5 has foil and nonfoil versions, while cn29 only has foil versions?

The latter is a surge foil. Maybe it's in the fancier version of the commander decks?

Hanmac:
@tehdiplomat @tool4ever is this MR ready?

i will try to use this later to update the Card Image Keys too

tehdiplomat (approved):

PR #7489, opened 2025-05-03
UNF: Added the 4 eternal-format-legal dice modification/reroll cards

autumnmyst:
This adds the 2 dice roll modification cards:
- [Xenosquirrels](https://scryfall.com/card/unf/96/xenosquirrels)
- [Night Shift of the Living Dead](https://scryfall.com/card/unf/81/night-shift-of-the-living-dead)

The dice roll exchange card:
- [Vedalken Squirrel-Whacker](https://scryfall.com/card/unf/63/vedalken-squirrel-whacker)

And the reroll card:
- [Monitor Monitor](https://scryfall.com/card/unf/52/monitor-monitor)

## Important things to note:
Due to a lack of clarity in the comprehensive rules and card rulings regarding rerolling, modifying, and exchanging dice roll results, I have had to make some judgement calls for how specific interactions should operate. Many forum posts have alternate interpretations for how these cards interact with other cards like [Barbarian Class](https://scryfall.com/card/afr/131/barbarian-class), [Bamboozling Beeble](https://scryfall.com/card/unf/37/bamboozling-beeble), and [Iron Mastiff](https://scryfall.com/card/clb/317/iron-mastiff)

Additionally, I didn't add support for planar dice rerolling with Monitor Monitor. That is one incomplete aspect here, and I added a comment about it in the card script.

I will clarify the functionality I have decided upon after scouring many sources and the overall purpose of the cards effect.

#### 1. An initial **roll action** occurs, returning a list of natural results:
- Replacement effects are evaluated; this checks things like rolling more dice and ignoring the lowest, and VSW exchange effect.
- The number of dice to roll is calculated, and the dice are rolled and added natural 
- The dice that should be ignored are removed from the natural results
- The dice that players get to choose to ignore are chosen and removed from the natural results

#### 2. If the roll only cares about the highest result, all other results are removed from the natural results
- This is here once and once only

#### 3. Reroll effects are checked
- Currently, this is just Monitor Monitor
- For each card that can reroll, you choose a number of natural results to reroll. Those results are removed, and a new **roll action** occurs with dice equal to the number of results removed (with the same replacement effect triggers as the initial **roll action**)
- VSW's exchange effect can activate here for a second time technically, however, the ruling on the card emphasizes it should only happen once. While technically this replacement can stack, rerolled dice aren't supposed to trigger effects, as they are to be treated as if they never happened.

#### 4. Dice roll modifications are made
- This includes Xenosquirrels and Night Shift

#### 5. Vedalken Squirrel-Whacker exchanges are decided
- This is the last step before the the results are finalized 
- This is placed last for 2 reasons:
- 1. it's a replacement effect that references the "result" of the die roll, and a ruling on the card specified that a "result" is after all modifications
- 2. Exchange effects for P/T are static, and modifying one will actual lead to changing the other's value. Having VSW last makes this functionally identical to this exchange behavior

#### 7. Any roll-unrelated modifications are made, such as cards that ask a player to roll die and add or subtract from the result for a results table
- Shouldn't squirrel whacker swap with this result? I treat this differently, as VSW is a replacement effect specifically modifying the dice roll itself. I believe it makes the most sense to come up with a set of final results for the dice roll before moving on to what the card says to do with that roll. The word "result" is used in both cases here, but these modification are completely independent from the roll. Cards that modify in this way are modifying a number returned to the card from the roll, they no long longer care about the details of the roll itself.

- Additionally, there are no reroll-unrelated (on the card that rolls them) modifications for d6 rolls, only d20. So the order of this and VSW is never questioned in practice

## References
[CR 706 Dice Rolling](https://mtg.fandom.com/wiki/Dice_rolling)

If anyone has more official rulings for how these interactions should function, let me know and I can make those changes. Otherwise this is likely good enough for now.


forge-gui/res/cardsfolder/e/elvish_impersonators.txt
```diff
@@ -4,7 +4,11 @@ Types:Creature Elf
...
 PT:*/*
 K:ETBReplacement:Other:TrigRoll
-SVar:TrigRoll:DB$ RollDice | ResultSVar$ SetPwr | SubAbility$ RollTough | SpellDescription$ As CARDNAME enters, roll a six-sided die twice. Its base power becomes the first result and its base toughness becomes the second result.
-SVar:RollTough:DB$ RollDice | ResultSVar$ SetTgn | SubAbility$ DBAnimate
-SVar:DBAnimate:DB$ Animate | Defined$ Self | Power$ SetPwr | Toughness$ SetTgn | Duration$ Permanent
+SVar:TrigRoll:DB$ RollDice | ResultSVar$ diePwr | SubAbility$ RollTough | SpellDescription$ As CARDNAME enters, roll a six-sided die twice. Its base power becomes the first result and its base toughness becomes the second result.
+SVar:RollTough:DB$ RollDice | ResultSVar$ dieTgn | SubAbility$ StorePower
+SVar:StorePower:DB$ StoreSVar | SVar$ CurrentPower | Type$ CountSVar | Expression$ CurrentPower/Plus.diePwr | SubAbility$ StoreToughness
+SVar:StoreToughness:DB$ StoreSVar | SVar$ CurrentToughness | Type$ CountSVar | Expression$ CurrentToughness/Plus.dieTgn
+S:Mode$ Continuous | CharacteristicDefining$ True | SetPower$ CurrentPower | SetToughness$ CurrentToughness
```
tool4ever, line 11: these changes seem wrong
this is no CharacteristicDefining ability either

autumnmyst: I made this change when I noticed `getBasePower()` and `getBaseToughness()` returned 0 for cards with stats set through animate, but returned the actual value for cards with stats set with static effects.

Are you sure this isn't character defining?
**604.3a** "A static ability is a characteristic-defining ability if it meets the following criteria: (1) It defines an object’s colors, subtypes, power, or toughness; (2) it is printed on the card it affects, it was granted to the token it affects by the effect that created the token, or it was acquired by the object it affects as the result of a copy effect or text-changing effect; (3) it does not directly affect the characteristics of any other objects; (4) it is not an ability that an object grants to itself; and (5) it does not set the values of such characteristics only if certain conditions are met."

I do see your point that the card text isn't worded like other CD abilities, in that it says "becomes" instead of "is" but I think it checks all the CD boxes no?

tool4ever: I think that's a case where the wording got changed and we still refer to it for the printed value only, instead of Layer 7c
anyway the relevant rule in this case is:
> 208.2b. The card may have a static ability that creates a replacement effect that sets the creature's power and toughness to one of a number of specific values as it enters the battlefield or is turned face up. (See rule 614, "Replacement Effects.") Such an ability is worded "As [this creature] enters . . . ," "As [this creature] is turned face up . . . ," or "[This creature] enters as . . ." and lists two or more specific power and toughness values (and may also list additional characteristics). The characteristics chosen or determined with these effects affect the creature's copiable values. (See rule 707.2.) While the card isn't on the battlefield, its power and toughness are each considered to be 0.

forge-game/src/main/java/forge/game/ability/effects/RollDiceEffect.java
```diff
@@ -80,78 +136,149 @@ private static int rollDiceForPlayer(SpellAbility sa, Player player, int amount,
...
-        if (sa != null && sa.hasParam("UseHighestRoll")) {
-            naturalRolls.subList(0, naturalRolls.size() - 1).clear();
+        // Vedalken Exchange
+        List<Card> vedalkenSwaps = new ArrayList<>(dicePTExchanges);
+        if (!vedalkenSwaps.isEmpty()) {
+            DieRollResult rollToSwap;
+            do {
+                rollToSwap = player.getController().chooseRollToSwap(resultsList);
+                if (rollToSwap != null) {
+                    Card c = player.getController().chooseCardToDiceSwap(vedalkenSwaps, rollToSwap.getModifiedValue());
```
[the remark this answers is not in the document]
autumnmyst: My thinking was it would be easier to add AI logic with separate functions for each dice action, but that's probably not something to be concerned with at the moment. I removed a couple functions and used the one you suggested. I also replaced the increment/decrement function with `chooseBinary()`

forge-gui/res/cardsfolder/e/elvish_impersonators.txt
autumnmyst: Ah, okay that makes sense. I'll revert the elves, though for VSW I'll need a way to get the correct base power.

forge-game/src/main/java/forge/game/ability/effects/RollDiceEffect.java
```diff
@@ -80,78 +136,163 @@ private static int rollDiceForPlayer(SpellAbility sa, Player player, int amount,
...
-
-        for (int i = 0; i < amount; i++) {
-            int roll = MyRandom.getRandom().nextInt(sides) + 1;
-            // Play the die roll sound
-            player.getGame().fireEvent(new GameEventRollDie());
-            player.roll();
-            naturalRolls.add(roll);
+        // Reroll Phase:
+        String monitorKeyword = "Once each turn, you may pay {1} to reroll one or more dice you rolled.";
+        CardCollection canRerollDice = new CardCollection(getRerollCards(player, monitorKeyword));
```
tool4ever, line 151: should be refactored without extra CardCollection

forge-game/src/main/java/forge/game/ability/effects/RollDiceEffect.java
```diff
@@ -80,78 +136,149 @@ private static int rollDiceForPlayer(SpellAbility sa, Player player, int amount,
...
+                    parts = c.getSVar("CurrentToughness").split("\\$");
+                    int cToughness = Integer.parseInt(parts[1]);
+                    List<String> choices = Arrays.asList("Power", "Toughness");
+                    String powerOrToughness = player.getController().chooseRollSwapValue(choices, rollToSwap.getModifiedValue(), cPower, cToughness);
+                    if (powerOrToughness != null) {
+                        int tempRollValue = rollToSwap.getModifiedValue();
+                        switch (powerOrToughness) {
+                            case "Power":
+                                rollToSwap.setModifiedValue(cPower);
+                                c.setSVar("CurrentPower", "Number$" + tempRollValue);
```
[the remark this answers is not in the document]
tool4ever: use `getCurrentPower()` and `addNewPT()`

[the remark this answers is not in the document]
autumnmyst: Okay will do, didn't realize that `getCurrentPower()` returned current base

forge-game/src/main/java/forge/game/ability/effects/RollDiceEffect.java
autumnmyst: done, made getRerollCards return the filtered CardCollection

forge-game/src/main/java/forge/game/ability/effects/RollDiceEffect.java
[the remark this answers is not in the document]
autumnmyst: done, it's animated now. Also localized the power / toughness choice labels

forge-game/src/main/java/forge/game/ability/effects/RollDiceEffect.java
```diff
@@ -179,23 +316,140 @@ private static int rollDiceForPlayer(SpellAbility sa, Player player, int amount,
...
+        switch (player.getGame().getReplacementHandler().run(ReplacementType.RollDice, repParams)) {
+            case NotReplaced:
+                break;
+            case Updated: {
+                amount = (int) repParams.get(AbilityKey.Number);
+                ignore = (int) repParams.get(AbilityKey.Ignore);
+                //noinspection unchecked
+                ignoreChosenMap = (Map<Player, Integer>) repParams.get(AbilityKey.IgnoreChosen);
+                //noinspection unchecked
+                dicePTExchanges.addAll((Set<Card>) repParams.get(AbilityKey.DicePTExchanges));
```
tool4ever, line 419: you shouldn't need to add them here since you already passed the same Set to ReplacementHandler
it does not make copies by itself

tool4ever:
nice work for your first PR, especially appreciate the detailed description :)
while we've been avoiding some of your parts like hardcoded keywords for a while I think it should be fine on a first iteration in case of Unfinity
Since you did a good job of keeping your changes grouped together I'm not seeing anything too concerning
I'll give this a test drive during the week after you pass the build

forge-game/src/main/java/forge/game/ability/effects/ReplaceEffect.java
```diff
@@ -48,6 +49,13 @@ public void resolve(SpellAbility sa) {
             for (Player key : AbilityUtils.getDefinedPlayers(card, sa.getParam("VarKey"), sa)) {
                 m.put(key, m.getOrDefault(key, 0) + AbilityUtils.calculateAmount(card, varValue, sa));
             }
+        } else if ("CardSet".equals(type)) {
+            Set<Card> cards = (Set<Card>) params.get(varName);
+            List<Card> list = AbilityUtils.getDefinedCards(card, varValue, sa);
+            if (!list.isEmpty()) {
+                cards.add(list.get(0));
+                params.put(varName, cards);
```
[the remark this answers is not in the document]
autumnmyst: Yeah that's true, for a while I didn't realize I needed to add [DicePTExchanges ](https://github.com/Card-Forge/forge/pull/7489/files/a0a6b5e4ef93dcb7d38678e07d5e9317cab6bca2#diff-75b34c46abddbed8878188436b88539c4c68935748e8fa6da6f614323c4b4526R45) in `ReplaceRollDice.java` and this was a workaround I had but forgot to remove. Just removed it

forge-game/src/main/java/forge/game/ability/effects/RollDiceEffect.java
autumnmyst: Yeah, I forgot to remove that, same reason as above comment

autumnmyst:
> nice work for your first PR, especially appreciate the detailed description :) while we've been avoiding some of your parts like hardcoded keywords for a while I think it should be fine on a first iteration in case of Unfinity Since you did a good job of keeping your changes grouped together I'm not seeing anything too concerning I'll give this a test drive during the week after you pass the build

Thanks! I tried to make the GUI as intuitive as possible.

Also, do you have any idea why the virtual framebuffer might be erroring?

Jetz72:
> Also, do you have any idea why the virtual framebuffer might be erroring?

Looks like  you added methods to PlayerController without updating PlayerControllerForTest.

autumnmyst:
> > Also, do you have any idea why the virtual framebuffer might be erroring?
> 
> Looks like you added methods to PlayerController without updating PlayerControllerForTest.

That was it, just checked the file and added those functions. Not sure why I was able to debug while those weren't implemented, but they are now. Thanks for the help.

tool4ever (approved):

PR #7496, opened 2025-05-04
feature/Refactored-server-url-parsing-logic

TimothyWright95:
New URL parsing logic fixes prior issue with port being null.
New parsing logic allows for localhost and similar network computer names, via InetAdress.getByName()
This now allows any valid URL, IP, or hostname that resolves via DNS to a valid IP


tool4ever (approved):
thanks

PR #7501, opened 2025-05-07
Fixing an issue in which touching the space the map occupies outside …

Arguas:
…of the world map does not allow the player to move (very relevant on maps with content in the top left corner)


Arguas:
Tested under the following conditions:
- Within town/dungeon where the hud is disabled
- Within town/dungeon with debugMap enabled
- On world map without debug mode enabled
- On world map *with* debug mode enabled

This also fixes a bug in which clicking on the location where the map zoom button would be pressed where it is alpha'd out still would cause the zoom map to appear

tehdiplomat (approved):

PR #7504, opened 2025-05-08
CardEdition: add collector number for other

Hanmac:
Added Other Collector Number
(need to update the other sets)

so other references to Helper cards (facedown/manifest) and Emblem can be updated to check if they are part of the set

### DIFF
--- a/forge-core/src/main/java/forge/card/CardEdition.java
+++ b/forge-core/src/main/java/forge/card/CardEdition.java
@@ -560,15 +560,6 @@ public final class CardEdition implements Comparable<CardEdition> {
         return sheets;
     }
 
-    public EditionEntry findOther(String name) {
-        for (EditionEntry o : this.otherMap.values()) {
-            if (name.equals(o.name)) {
-                return o;
-            }
-        }
-        return null;
-    }
-
     public static class Reader extends StorageReaderFolder<CardEdition> {
         private final boolean isCustomEditions;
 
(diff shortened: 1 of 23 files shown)


forge-core/src/main/java/forge/card/CardEdition.java
```diff
@@ -296,6 +297,42 @@ public int compareTo(TokenInSet o) {
             return thisCollNr.compareTo(othrCollNr);
         }
     }
+    public static class OtherInSet implements Comparable<OtherInSet> {
```
tehdiplomat, line 300: I think I'm even less happy now that there are 3 of these.

Hanmac: i'm open for suggestions

should i use records instead?

tehdiplomat: Can we just reduce all of these down to ObjectInSet with an enum of { card, token, other }? And for non-cards             rarity and functionalVariantName be set to null for cards?

Jetz72: Maybe `EditionEntry`, for a more clear name?

tehdiplomat: I tested the one that was added for Network play. Have we added it anywhere else? `EditionEntry` sounds great to me. I don't think I need this to be an entry. I was just getting annoyed by all of the repetition.

tehdiplomat (approved):

tehdiplomat (approved):

PR #7505, opened 2025-05-08
Some support for Desert Cenote

### LINKED ISSUE #7498: On-damage trigger still occuring even when source of trigger left the battlefield prior to damage being dealt
**Describe the bug**
An opponent's creature, which was enchanted with [sigil of sleep](https://scryfall.com/card/ddi/25/sigil-of-sleep), put a triggered ability onto the stack that was going to damage me.
In response, I destroyed the creature. After this, when sigil of sleep was no longer in play, priority passed around and the damage was dealt.
Sigil of sleep's effect still triggered even though it was no longer on the battlefield. Unless I'm misunderstanding the rules this is incorrect?

I'd guess that the root cause is related to whatever code is looking back at the last known state of the creature when determinining how much damage to deal, and that it's incorrectly also picking up some other side-effects that should no longer apply.

**Logs**
[sigil_of_sleep.log](https://github.com/user-attachments/files/20050746/sigil_of_sleep.log)

**Screenshots**
Image

**Repro game state**
1. Pass to your own turn
2. Two nekusar triggers will be put onto the stack
3. [optionally] allow the extra draw to resolve]
4. Before the damage ping(s) resolve, cast swords to plowshares on nekusar
5. Resolve the abilities and see that sigil of sleep's effect triggers from the damage

[sigil_of_sleep_game_state.txt](https://github.com/user-attachments/files/20050748/sigil_of_sleep_game_state.txt)

**Desktop:**
 - OS: Windows
 - Version: 2.0.04-SNAPSHOT-04.28

tool4ever:
Closes #7498

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityKey.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityKey.java
@@ -93,7 +93,6 @@ public enum AbilityKey {
     Mana("Mana"),
     MergedCards("MergedCards"),
     Mode("Mode"),
-    Modifier("Modifier"),
     MonstrosityAmount("MonstrosityAmount"),
     NaturalResult("NaturalResult"),
     NewCard("NewCard"),
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -2220,7 +2220,7 @@ public class AbilityUtils {
         // Count$IfCastInOwnMainPhase.<numMain>.<numNotMain> // 7/10
         if (sq[0].contains("IfCastInOwnMainPhase")) {
             final PhaseHandler cPhase = game.getPhaseHandler();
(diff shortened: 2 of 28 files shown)


dracontes:
Unfortunately, `ColorsFrom$ Card.inZoneHand+YouOwn`, `ColorsFrom$ Card.inZoneHand+YouCtrl` and `ColorsFrom$ Hand` are all getting me this not-particularly-useful color selection prompt:

image

tool4ever:
You need to use proper Defined Syntax with ValidHand

dracontes:
The color choosing replacement effect is working now, being able to select from zero to two colors. 
However, the linked mana ability displays the same issue as before: when selecting two colors, only mana of the first selected color is produced. Also when I select no color that mana ability now produces two colorless mana.

image

tool4ever:
Sigh looks like ManaEffect also needs some work then

tool4ever:
@dracontes
should work now?

dracontes:
> @dracontes should work now?

Indeed it does: passes all relevant tests.

PR #7513, opened 2025-05-09
Adventure - Fix for 'Fetch' objectives

entrailmix:
If a user Receives an item specified in a 'Fetch' objective before the objective has started, it was impossible for them to finish the quest. Now, as long as they are in the PointofInterest where the item was found, they can enter their inventory and Use the item instead. This will prevent main quest softlocks.

I tested this on the only 'Fetch' objective in the game so far, stage 2 of The Dig Site.
I acquired the item before beginning the quest. I completed the quests up until that objective, entered the PoI and used enter.
image

I also deleted an empty line.

This is my first time using Github, IDEs, and contributing! Let me know your comments.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/data/AdventureQuestStage.java
+++ b/forge-gui-mobile/src/forge/adventure/data/AdventureQuestStage.java
@@ -330,7 +330,7 @@ public class AdventureQuestStage implements Serializable {
                         status = ++progress1 >= count1 ? COMPLETE : status;
                 }
                 else if (event.type == AdventureQuestEventType.USEITEM) {
-                    if ((itemNames.isEmpty()) || itemNames.contains(event.item.name))
+                    if ((itemNames.isEmpty()) || (event.item != null && itemNames.contains(event.item.name)))
                         status = ++progress3 >= count3 ? COMPLETE : status;
                 }
                 break;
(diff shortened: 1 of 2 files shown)


forge-gui-mobile/src/forge/adventure/data/AdventureQuestStage.java
```diff
@@ -325,9 +325,14 @@ public AdventureQuestController.QuestStatus handleEvent(AdventureQuestEvent even
...
-                status = event.type == AdventureQuestEventType.RECEIVEITEM
-                        && (itemNames.isEmpty()) || (event.item != null && itemNames.contains(event.item.name))
-                        && ++progress1 >= count1 ? COMPLETE : status;
+                if (event.type == AdventureQuestEventType.RECEIVEITEM) {
+                    if ((itemNames.isEmpty()) || (event.item != null && itemNames.contains(event.item.name)))
+                        status = ++progress1 >= count1 ? COMPLETE : status;
+                }
+                else if (event.type == AdventureQuestEventType.USEITEM) {
+                    if ((itemNames.isEmpty()) || (event.item != null && itemNames.contains(event.item.name)))
+                        status = ++progress3 >= count3 ? COMPLETE : status;
```
tehdiplomat, line 334: I'm a little bit worried about how this whole class interacts with its different progress/count variables. But let's try it and keep an eye on reports.

tehdiplomat (approved):

PR #7514, opened 2025-05-09
FIN: Y'shtola Rhul

dracontes:
[As revealed recently.](https://www.reddit.com/r/magicTCG/comments/1kirvi7/fin_yshtola_rhul_borderless_ver/) While I imagine there's probably a way to do this card that doesn't involve a static trigger to count end steps, said way is currently escaping me. In any case, it's working as expected even so.


tool4ever:
check what happens if this is the only creature you control :p

tool4ever (approved):

PR #7517, opened 2025-05-10
Tiered: first example

Hanmac:
OR should Spree/SpreeCost and Tiered/TieredCost be simplified?

Also should they check for the Keyword instead of Param?


tool4ever:
I guess some lines could be saved if both would check for something like ModeCost$

and I can't think of a case where having the param but not the keyword could happen 🤷‍♂️

Hanmac:
@tool4ever Spree and Tiered only work with Intrinsic, so there is no way to add or remove this keyword

I updated the code to use ModeCost for both

tool4ever (approved):

PR #7518, opened 2025-05-10
5 FIN cards

Fulgur14:
Not much time today, so I only implemented a few.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/absolute_virtue.txt
+++ b/forge-gui/res/cardsfolder/upcoming/absolute_virtue.txt
@@ -1,0 +1,8 @@
+Name:Absolute Virtue
+ManaCost:6 W U
+Types:Legendary Creature Avatar Warrior
+PT:8/8
+R:Event$ Counter | ValidCard$ Card.Self | ValidSA$ Spell | Layer$ CantHappen | Description$ This spell can't be countered.
+K:Flying
+S:Mode$ Continuous | Affected$ You | AddKeyword$ Protection:Player.OpponentOf PlayerUID_ChosenPlayerUID:all opponents of ChosenPlayerName | DefinedKW$ You | Description$ You have protection from each of your opponents. (You can't be dealt damage, enchanted, or targeted by anything controlled by your opponents.)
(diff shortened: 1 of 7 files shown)


forge-gui/res/cardsfolder/upcoming/absolute_virtue.txt
```diff
@@ -0,0 +1,8 @@
+Name:Absolute Virtue
+ManaCost:6 W U
+Types:Legendary Creature Avatar Warrior
+PT:8/8
+R:Event$ Counter | ValidCard$ Card.Self | ValidSA$ Spell | Layer$ CantHappen | Description$ This spell can't be countered.
+K:Flying
+S:Mode$ Continuous | Affected$ You | AddKeyword$ Protection:Player.OpponentOf PlayerUID_ChosenPlayerUID:all opponents of ChosenPlayerName | DefinedKW$ You | Description$ You have protection from each of your opponents. (You can't be dealt damage, enchanted, or targeted by anything controlled by your opponents.)
```
tool4ever, line 7: DefinedKW$ doesn't exist here...

Fulgur14: Alright, so what should I do? I couldn't find any good example.

tool4ever: try less complicated: `Protection:Player.Opponent:each of your opponents`

Fulgur14: OK -- technically, it's not exactly the same (should be separate protection ability from each opponent, like with "protection from all colors"), but I don't think it matters anywhere.

tool4ever: hmn yea maybe @Hanmac needs to change that when relevant?

Hanmac: It only matters for "Cephalid Snitch" which does remove `protection from black`
And other color related ones

Fulgur14: Yes, it's a potential problem, but not a current one.

tool4ever (approved):

PR #7519, opened 2025-05-11
Mobile Deck Editor Reorganization and Feature Update

### LINKED ISSUE #3652: Cryptic Spires does not work properly

### LINKED ISSUE #6993: I want to function to change version of cards.
Currently, to change the version of a card, it is necessary to search by entering the card name in the search field.
With the recent increase in card variations, I would like to have an option to easily change the version of a card.

In the deck editor, add a right-click menu for cards, such as “Search for this card name in the catalog.

Seems less time consuming to implement.

**Additional context**
I may have requested something similar before, but I really want it.

### LINKED ISSUE #7861: Can exit draft deck editor without adding lands
**Describe the bug**
After drafting, if you exit the draft deck editor before adding lands, you end up with wastes in your deck and cannot edit the deck before playing. It is not possible to recover from this once you get this far.

**To Reproduce**
1. Complete a booster draft
2. Build a deck, but don't add basic lands
3. Click out of the deck editor

**Expected behavior**
There should be a warning about deck size if there are less than 40 cards in the deck, preventing inadvertent exit from the deck editor.
It should be possible to go back into the deck editor before starting round one.
It should also be possible to add basic lands to the deck while sideboarding, without having to add them to your sideboard before round one.

**Screenshots**
None

 - Windows 10

### LINKED ISSUE #8141: Drafts crash in image view when category containing a final card is collapsed
**Describe the bug**
When drafting in Classic mode, if cards are grouped by type and the "basic land" category is minimized, maximizing that category to select a basic land as the last card in a pack will result in an error. This may apply to other card types as well if they are left as a final card, but this is most easy to replicate with basic lands. The crash can be avoided if the user switches to another sort method.

**To Reproduce**
1. Open Forge-Adventure and select "Classic."
2. Select New Game, then Block/Set, then select any format that includes basic lands as draftable cards. I've replicated this with both Theros block (choosing BNG/THS/THS and JOU/BNG/THS) and Core Set 2021.
3. Choose to group by any grouping that isolates basic lands. (Any grouping except "set" should work.") Use the arrow to minimize the section that contains basic lands.
4. Make any choices in the draft until a final card in the pack remains.
5. Maximizing the category containing the land and then attempting to select the final card will result in an error like:

" java.lang.NullPointerException: Cannot invoke "forge.item.PaperCard.getRules()" because "card" is null
	at forge.deck.DeckSection$Validators.lambda$static$5(DeckSection.java:116)
	at forge.deck.DeckSection.validate(DeckSection.java:52)
	at forge.deck.DeckSection.matchingSection(DeckSection.java:57)
	at forge.deck.FDeckEditor$CatalogPage.onCardActivated(FDeckEditor.java:1555)
	at forge.deck.FDeckEditor$DraftPackPage.onCardActivated(FDeckEditor.java:2013)
	at forge.deck.FDeckEditor$CardManagerPage.lambda$new$0(FDeckEditor.java:958)
	at forge.itemmanager.ItemManager.activateSelectedItems(ItemManager.java:943)
	at forge.itemmanager.views.ImageView.tap(ImageView.java:566)
	at forge.itemmanager.views.ItemView$Scroller.tap(ItemView.java:68)
	at forge.Forge$MainInputProcessor.tap(Forge.java:1363)
	at forge.toolbox.FGestureAdapter.touchUp(FGestureAdapter.java:245)
	at forge.toolbox.FGestureAdapter.touchUp(FGestureAdapter.java:177)
	at com.badlogic.gdx.InputEventQueue.drain(InputEventQueue.java:70)
	at com.badlogic.gdx.backends.lwjgl3.DefaultLwjgl3Input.update(DefaultLwjgl3Input.java:190)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Window.update(Lwjgl3Window.java:414)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:181)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:153)
	at forge.app.GameLauncher.<init>(GameLauncher.java:51)
	at forge.app.Main.main(Main.java:30)"

6. However, the card can be selected and the draft can be completed if the grouping of the cards is changed after the land category is maximized. Switching to List View also allows the card to be chosen.

**Expected behavior**
Maximizing the category containing a final card should allow that card to be chosen without changing card groupings.

 - OS: MacOS Sequoia 15.5
 - Version: 2.0.05-SNAPSHOT-07.20

Jetz72:
Ongoing overhaul to the mobile deck editor. Large parts of it have been revised, along with parts of the systems it ties into.

Major goals of this update:

- Make the AdventureDeckEditor inherit from the main one, sharing code where appropriate.
- Add support for variant cards in adventure mode (Though there's no inherent way to get them yet.)
- Adds placeholder Attraction and Contraption if you have less than the minimum number in your Adventure deck.
- Add Draft-Matters support to mobile and adventure (Draft-matters cards seem to work, but the sets that use them are designed for 4 player drafts, which is out of scope for now.)
- Give the adventure editor support for favorites, replacing prints, importing+exporting, (Favorites & Replacing prints done, Importing to come later.)
- Remove the adventure-specific filter dropdown, move auto-sell into its own tab.
- Allow sorting the adventure catalog by price, showing the price outside the shop, and make prices show in list view. (Done)
- Remove the `EditorType` enum used across a bunch of switch statements in the editor, replacing it with `DeckEditorConfig`, a class that supplies these behavioral differences through overrides of its methods. Vastly reduces hard-coding of game-type specific things into the deck editor itself.
- Lots of logic that used separate "addCards" and "removeCards" calls when moving cards around have been refactored into a single "moveCards" operation that ensures cards aren't either spawned from nothing or lost into the void.
- Adds the ability to re-open the deck editor freely in Adventure Events. (Closes #7861.)

Closes #6993; Closes #3652; Closes #8141.


forge-core/src/main/java/forge/util/ItemPool.java
```diff
@@ -269,11 +269,12 @@ public void removeAllFlat(final Iterable<T> flat) {
         // need not set out-of-sync: either remove did set, or nothing was removed
     }
 
-    public void removeIf(Predicate<T> test) {
-        for (final T item : items.keySet()) {
-            if (test.test(item))
-                remove(item);
-        }
+    public void removeIf(Predicate<T> filter) {
+        items.keySet().removeIf(filter.negate());
```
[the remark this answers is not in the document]
Jetz72: Dunno how I missed that. Bad copy-paste during a merge? I think I eliminated the place it was used so it didn't come up, but good catch by the AI.

Jetz72:
Think this is just about ready for review. I'm sure I didn't catch everything but I've at least tested each component a few times for outstanding issues. 

Stuff to do in future PRs or projects:
* Importing decks in adventure mode. Logic's mostly written but was getting a bit out of scope. I aim to add this soon.
* Full conspiracy and commander draft support. Conspiracy cards seem to work but 4 player draft formats are a separate animal. And 4 player anything in Adventure mode is one more animal removed from that.
* Reorganize the move labels so they're assembled as ("Add to" + "zone") rather than ("Add" + "to zone"). Might be complications in other languages for that but it's currently doubling the number of localized strings needed here.
* Move the commander filter logic over to be with the regular filters.
* Partners in Oathbreaker (Probably with #8066).
* Redo `getColOverrides` in the adventure editor to not need so much near-duplicate code. Probably needs a bunch of overhauls to the underlying logic though.
* Change Adventure save logic to just serialize the whole deck instead of splitting each deck into sections and storing them individually. (May just do this the next time we have to add a section to save; #7799 would help a bit with the migration logic here.)

kevlahnota:
Hmmm changing art for the card has a advanced search button if there's a lot of variants (like basic land), and if isn't disabled or hidden (Adventure Mode), then the UI will be stucked after trying to exit from the advanced search screen unless this is also covered though, there's a pr that allows you to advance search like scryfall, I think that's more preferable option I guess in the future.

forge-gui/src/main/java/forge/gamemodes/planarconquest/ConquestUtil.java
```diff
@@ -137,25 +137,6 @@ public static CardPool getAvailablePool(Deck deck) {
             availableCards.remove(pc);
         }
 
-        // Remove all cards in the main deck
```
tool4ever, line 140: what's the reasoning for this removal?

Jetz72: The old `refresh()` method in `CatalogPage` passed all the responsibility of building the Planar Conquest card pool over to `ConquestUtil::getAvailablePool` - 
https://github.com/Card-Forge/forge/blob/4db0b6a67fdeae964e698e53bd50461bf383c841/forge-gui-mobile/src/forge/deck/FDeckEditor.java#L1483-L1485

But  aside from providing inventory, this extra color identity and singleton filtering was essentially doing what the Commander and QuestCommander branches were already doing. In the revision, `ConquestUtil::getAvailablePool` is only used as a player inventory supplier (Adventure and Quest inventories are supplied by their EditorConfigs in similar ways), which is routed to the card pool for the refresh method where the commander filtering is applied:

https://github.com/Card-Forge/forge/blob/76e821ccc9e61e1bd62773ee43d54962144e950b/forge-gui-mobile/src/forge/deck/FDeckEditor.java#L358-L361

https://github.com/Card-Forge/forge/blob/76e821ccc9e61e1bd62773ee43d54962144e950b/forge-gui-mobile/src/forge/deck/FDeckEditor.java#L1750-L1775

Still not the perfect spot for it per the TODO, but card inventories are now treated similarly across our three progression-based modes.

Jetz72:
> Hmmm changing art for the card has a advanced search button if there's a lot of variants (like basic land), and if isn't disabled or hidden (Adventure Mode), then the UI will be stucked after trying to exit from the advanced search screen unless this is also covered though, there's a pr that allows you to advance search like scryfall, I think that's more preferable option I guess in the future.

I was unaware of this search feature, but I was unable to reproduce any hang with it. I selected it, filtered for set UNF, hit the back arrow, and got a list of UNF cards that I was able to select and use. Do I need to use it in a particular way? Or in a particular context?
image

kevlahnota:
> > Hmmm changing art for the card has a advanced search button if there's a lot of variants (like basic land), and if isn't disabled or hidden (Adventure Mode), then the UI will be stucked after trying to exit from the advanced search screen unless this is also covered though, there's a pr that allows you to advance search like scryfall, I think that's more preferable option I guess in the future.
> 
> I was unaware of this search feature, but I was unable to reproduce any hang with it. I selected it, filtered for set UNF, hit the back arrow, and got a list of UNF cards that I was able to select and use. Do I need to use it in a particular way? Or in a particular context? image

try running adventure mode then change the art in the adventure deck editor, your screenshot is classic mode not adventure mode.

Jetz72:
Alright, I think I've patched up the back button functionality enough for it to not break in adventure mode. Still has problems, but it should be more resistant against soft-locking or navigating anywhere weird now.

Also took out the home menu bar from the advanced filter screen, and restored the ability to use it to filter the card manager while in adventure mode. 

Finally, ported over the foil toggle. The new foil effect looks really good by the way, @kevlahnota.

forge-core/src/main/java/forge/deck/DeckFormat.java
```diff
@@ -191,12 +195,57 @@ public Range<Integer> getMainRange() {
...
+                return 2; //These are two-of.
+            return getExtraSectionMaxCopies(section);
+        }
+        else
+            return this.getMaxCardCopies();
+    }
+
+    public int getExtraSectionMaxCopies(DeckSection section) {
+        return switch (section) {
+            case Avatar, Commander, Planes, Dungeon, Attractions, Contraptions -> 1;
```
tehdiplomat, line 224: Can't wait til WotC prints a Brothers Yamazaki type card that can be partner commander with itself.

forge-gui-desktop/src/test/java/forge/BoosterDraftTest.java
```diff
@@ -66,11 +67,16 @@ public CardPool nextChoice() {
...
     @Override
-    public boolean setChoice(final PaperCard c) {
+    public boolean setChoice(final PaperCard c, DeckSection section) {
         System.out.println(c.getName());
         return false;
     }
 
+    @Override
+    public void skipChoice() {
+        System.out.println("Skip.");
```
tehdiplomat, line 77: How often is this called? I could see it being annoying if its printed a lot.

forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
```diff
@@ -238,6 +243,8 @@ public static BoosterDraft createDraft(final LimitedPoolType draftType, final Ca
             }
         }
 
+        draft.shouldShowDraftLog = block.getName().contains("Conspiracy"); //TODO: Make this an actual property on the block and/or edition data.
```
tehdiplomat, line 249: Agree this should just look for all available packs and see if that edition has a "NeedsDraftLog" config.

forge-core/src/main/java/forge/deck/DeckFormat.java
Jetz72: Yeah. I figure we'll have to go through and revise this logic to hook into some more dynamic deck construction rules someday. I drafted up a proposal for that in #8066. For now this is just the existing logic that I've moved out of the deck editor itself because this seems like a better place for it.

forge-gui-desktop/src/test/java/forge/BoosterDraftTest.java
Jetz72: It's in `BoosterDraftTest` rather than `BoosterDraft`, and it's used for a few obscure draft-matters cards, so basically never. I only had it print because that's what the above `setChoice` method does and if anyone ever does run into it it'd make sense to print something rather than have a conspicuous absence in the logs.

forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
Jetz72: Only thing I'd want to remember is for drafts with the complete card pool, where it'd be needed for the one-in-several-thousand chance a draft-matters card comes up. Might tack on a system message like "(The draft log shows actions that happen during the draft due to cards such as Agent of Acquisitions)", so that players aren't asking themselves "what's this button here that opens a blank panel that tells me almost nothing?".

forge-core/src/main/java/forge/deck/DeckFormat.java
Hanmac: > Can't wait til WotC prints a Brothers Yamazaki type card that can be partner commander with itself.

You mean like `Mothers Yamazaki`
> Partner with itself (When this enters, target player may put Mothers Yamazaki into their hand from their library, then shuffle. A Commander deck can include two of this card, and they can be your commanders.)

forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
tehdiplomat: Thats a good point. I was just thinking for a mixed booster draft like..MH2-Time Spiral-Conspiracy That we would show the draft log even though there isn't really a "block" associated with it.

forge-core/src/main/java/forge/deck/DeckFormat.java
tehdiplomat: Because of course they already printed that.

PR #7520, opened 2025-05-11
FIN: 4 Job select cards and support

dracontes:
As revealed recently; tested to satisfaction:
- [Black Mage's Rod](https://scryfall.com/card/fin/90/black-mages-rod)
- [Dragoon's Lance](https://scryfall.com/card/fin/17/dragoons-lance)
- [Summoner's Grimoire](https://scryfall.com/card/fin/205/summoners-grimoire)
- [White Mage's Staff](https://scryfall.com/card/fin/42/white-mages-staff)

The only issue is that I'm not sure how to implement the flavor words for the equip abilities of Dragoon's Lance and Summoner's Grimoire so they display on the in-game Card Detail. However, as that's fundamentally a cosmetic issue, I'll leave it to be resolved at a latter date.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/summoners_grimoire.txt
+++ b/forge-gui/res/cardsfolder/upcoming/summoners_grimoire.txt
@@ -1,0 +1,12 @@
+Name:Summoner's Grimoire
+ManaCost:3 G
+Types:Artifact Equipment
+K:Job select
+S:Mode$ Continuous | Affected$ Creature.EquippedBy | AddType$ Shaman | AddTrigger$ TrigAttack | Description$ Equipped creature is a Shaman in addition to its other types and has "Whenever this creature attacks, you may put a creature card from your hand onto the battlefield. If that card is an enchantment card, it enters tapped and attacking."
+SVar:TrigAttack:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigChoose | TriggerZones$ Battlefield | OptionalDecider$ You | TriggerDescription$ Whenever this creature attacks, you may put a creature card from your hand onto the battlefield. If that card is an enchantment card, it enters tapped and attacking.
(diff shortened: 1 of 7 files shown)


Hanmac:
Yeah, we can skip the flavor words for now

Hanmac (dismissed):

forge-gui/res/cardsfolder/upcoming/summoners_grimoire.txt
```diff
@@ -0,0 +1,13 @@
+Name:Summoner's Grimoire
+ManaCost:3 G
+Types:Artifact Equipment
+K:Job select
+S:Mode$ Continuous | Affected$ Creature.EquippedBy | AddType$ Shaman | AddTrigger$ TrigAttack | Description$ Equipped creature is a Shaman in addition to its other types and has "Whenever this creature attacks, you may put a creature card from your hand onto the battlefield. If that card is an enchantment card, it enters tapped and attacking."
+SVar:TrigAttack:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigChoose | TriggerZones$ Battlefield | OptionalDecider$ You | TriggerDescription$ Whenever this creature attacks, you may put a creature card from your hand onto the battlefield. If that card is an enchantment card, it enters tapped and attacking.
+SVar:TrigChoose:DB$ ChooseCard | Mandatory$ True | ChoiceZone$ Hand | Choices$ Creature.YouOwn | ChoiceTitle$ Choose a creature card in your hand | SubAbility$ DBChangeOne
```
tool4ever, line 7: why is it Mandatory?

Hanmac, line 7: @tool4ever I wish that would work without extra Choose Ability 😒

tool4ever: doesn't look like other cards with such condition exist yet, otherwise some extra code could be worth it

Hanmac: Took a small peak, 
Might change that in the future 

Good for now I think 🤔 

Question for later what happens if the creature would only be an Enchantment on the battlefield, but I can ask the Rules Manager later

PR #7521, opened 2025-05-11
Fix mobile deck editor selecting wrong cards under certain conditions

Jetz72:
I'm not sure I fully grasp the scope of this issue, but apparently if the mobile deck editor is in ImageView mode, `groupBy` is set to something non-null, `pileBy` is null, and you change your search filters in a way that causes a group that had cards in it to become empty, the group will be emptied of items but it won't clear the piles within that group. They won't be visible, but `getItemAtPoint` can still see them, which can cause it to return an item that was previously filtered out. That item that should no longer exist later gets referred to by its outdated index, then that wrong index gets used to look up a real item that actually is visible in the list. 

End result: you click one card and the menu opens for a different card, or for a null card and you get problems. The recent addition of `card.hasNoSellValue` in the Adventure editor's `buildMenu` method turned this bug into an immediate NPE, but I suspect the underlying issue has been there much longer. 

This PR changes `getItemAtPoint` to ignore empty groups and their ghost piles, and blocks off the NPE by preventing `buildMenu` from being called without a valid card selected.

### DIFF
--- a/forge-gui-mobile/src/forge/itemmanager/ItemManager.java
+++ b/forge-gui-mobile/src/forge/itemmanager/ItemManager.java
@@ -1003,6 +1003,8 @@ public abstract class ItemManager<T extends InventoryItem> extends FContainer im
     private class ContextMenu extends FDropDownMenu {
         @Override
         protected void buildMenu() {
+            if (getSelectedItem() == null)
+                return;
             contextMenuBuilder.buildMenu(this, getSelectedItem());
         }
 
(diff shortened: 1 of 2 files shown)


tool4ever:
related to #7074?

Jetz72:
Doesn't look that way. The root of the problem here is in `updateLayout`, not `draw`.

https://github.com/Card-Forge/forge/blob/0b5f14cd9caae3b2b9eb337b9d08dd82f7a908d0/forge-gui-mobile/src/forge/itemmanager/views/ImageView.java#L437-L476

When it's going through updating groups, if `pileBy` is null, it'll skip the first block 440 where piles get cleared and remade. Then if there are no items left in the group, it'll short circuit in that block on 459 before the `if (pileBy == null)` block can clear them. It might be better to clear the piles somewhere in this function under these circumstances, but I haven't checked if having zero piles would break assumptions elsewhere.

tehdiplomat:
This might be related to the adventure deck editor sometimes grabbing the first card of the wrong pile when adding/removing cards. I'm not sure if we have an active issue for this. But I've seen this before, and I've seen people complain about it, but I had no idea how to dig into it.

tehdiplomat (approved):

PR #7536, opened 2025-05-12
Alchemy Rebalancing for May 13, 2025

paulsnoops:
Source: https://magic.wizards.com/en/news/mtg-arena/announcements-may-12-2025#Alchemy

Alchemy only - no new script needed:
- [x] Ribald Shanty
- [x] Enduring Friendship
- [x] Naktamun Shines Again
- [x] Awestruck Cygnet


tehdiplomat (approved):

tehdiplomat:
I wonder if we need a mechanism for Alchemy Rebalancing cards to redownload the image.. 

Maybe something in the cardscript with a date code and if the image file is before the date code then we redownload?

PR #7537, opened 2025-05-12
Adding monochrome icon to the android app

rpg2014:
Added monochrome icon pngs for each mipmap size and added them to the xml configs for both icons (standard + rounded).  The monochrome icons were made by editing the respective icon size in paint.net.

I'm not super familiar with android development, but this is a big pet peeve of mine, so I'd like to contribute this fix!

Thanks for maintaining this project!


tehdiplomat (approved):
Seems fine. If it is problematic with the next build we can roll it back and re-release

rpg2014:
The icon looks good to me! Thanks

PR #7546, opened 2025-05-13
Support for Lord Jyscal Guado

tool4ever:
Use "Any"


dracontes:
Tested a few cases that came to mind including proliferating, moving counters, [tribute](https://scryfall.com/card/bng/147/fanatic-of-xenagos): working as expected 👍

PR #7550, opened 2025-05-13
10 FIN/FIC cards, batch 5

Fulgur14:


forge-gui/res/cardsfolder/upcoming/locke_treasure_hunter.txt
```diff
@@ -0,0 +1,14 @@
+Name:Locke, Treasure Hunter
+ManaCost:1 B R
+Types:Legendary Creature Human Rogue
+PT:2/3
+S:Mode$ CantBlockBy | ValidAttacker$ Creature.Self | ValidBlocker$ Creature.powerGEX | Description$ NICKNAME can't be blocked by creatures with greater power.
+T:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigMill | TriggerZones$ Battlefield | TriggerDescription$ Mug — Whenever NICKNAME attacks, each player mills a card. If a land card was milled this way, create a Treasure token. Until end of turn, you may cast a spell from among those cards.
+SVar:TrigMill:DB$ Mill | Defined$ Player | NumCards$ 1 | RememberMilled$ True | SubAbility$ DBToken
+SVar:DBToken:DB$ Token | TokenAmount$ 1 | TokenScript$ c_a_treasure_sac | TokenOwner$ You | ConditionDefined$ Remembered | ConditionPresent$ Land | SubAbility$ DBEffect
+SVar:DBEffect:DB$ Effect | RememberObjects$ RememberedCard | StaticAbilities$ EffSModeContinuous | SubAbility$ DBCleanup | ForgetOnMoved$ Exile
```
Fulgur14, line 9: Should it be "ForgetOnMoved$ Graveyard, Exile" or something like that?

tool4ever: unless ruled otherwise I'd assume this follows the other allowed cases 👍

Fulgur14: It's a bit unclear, since I presume it means "remove from Remembered if the card leaves this zone", but shouldn't it be standard to forget a card when it moves zones in any way? Effects that track cards across the zones are the exception, not the rule.

tool4ever: well there's also a good amount of effects that combine something that affects cards + another on rules

tool4ever (approved):

PR #7578, opened 2025-05-16
How to suplex a train

Fulgur14:
There's something strange with Phantom Train's ability where its P/T display doesn't update properly after getting the counter -- maybe because the counter is gained before animating?


Hanmac:
yeah, @tool4ever might look into the PT-View refresh

PR #7604, opened 2025-05-18
Fix Henzie again with SvarFallback

Hanmac:
@tool4ever can you check if that doesn't break anything else?

otherwise the Henzie Svar rename might be enough for now?


Hanmac:
> We might try and check if this very old code is also obsolete:
> 
> https://github.com/Card-Forge/forge/blob/fd727a909b32eccae30c6d5e6aa11681f4ba6b85/forge-game/src/main/java/forge/game/spellability/SpellPermanent.java#L64-L66

Probably

I also would like to refactor the setSvars(getSVars) in Effect

This is the only real place where SpellAbility getSVars is used

tool4ever (approved):
👍

Hanmac:
@tool4ever we can cleanup more later

PR #7607, opened 2025-05-18
~ fix IterableUtil.and and Iterable.or usage

Hanmac:


Hanmac:
@Jetz72 i still got the same problems with Java21

Jetz72:
These changes shouldn't be necessary, but we can implement them to ensure Eclipse's compiler behaves consistently. IntelliJ is quick to warn that these type parameters can be inferred, and it's possible they get caught up in some future code cleanup. Hopefully whatever bug is causing this is resolved by then.

Hanmac:
@Jetz72 or @tehdiplomat needs to approve this MR

Jetz72 (approved):

PR #7636, opened 2025-05-21
Update Libgdx to 1.13.5 (fixes Steam Deck mobile backport crash)

Agetian:
Updates Libgdx dependency from 1.13.0 to the point release 1.13.5, which fixes mobile backport crashing on Steam Deck during any match (when trying to visualize cards).

This works on my Linux desktop without any issue as well.
Could use some testing on Android, perhaps, since I don't have a working Android build setup anymore.


Hanmac (approved):

tehdiplomat:
Happy to just merge once tests pass. The easiest way to test Android is to just let a snapshot run honestly. if it breaks we can rollback and fire a new android build.

PR #7637, opened 2025-05-21
Add TextBoxExchangeEffect Ability and 'Deadpool, Trading Card'

kvn1338:
- Added `ExchangeTextBox` Effect
- Added `Deadpool, Trading Card`

Currently the text exchange is not represented graphically, but the tool tip will show the exchanged text and functionally it works.

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/TextBoxExchangeEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/TextBoxExchangeEffect.java
@@ -1,0 +1,124 @@
+package forge.game.ability.effects;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Table.Cell;
+
+import forge.game.Game;
+import forge.game.ability.SpellAbilityEffect;
+import forge.game.card.Card;
+import forge.game.card.CardState;
+import forge.game.card.CardCopyService;
+import forge.game.event.GameEventCardStatsChanged;
+import forge.game.keyword.KeywordInterface;
+import forge.game.keyword.KeywordsChange;
+import forge.game.replacement.ReplacementEffect;
+import forge.game.spellability.SpellAbility;
+import forge.game.staticability.StaticAbility;
+import forge.game.trigger.Trigger;
+
+import java.util.List;
+
+/**
+ * Exchanges text boxes between two creatures.
+ */
+public class TextBoxExchangeEffect extends SpellAbilityEffect {
+    @Override
+    protected String getStackDescription(final SpellAbility sa) {
+        final List<Card> tgtCards = getTargetCards(sa);
+        Card c1;
+        Card c2;
+        if (tgtCards.size() == 1) {
+            c1 = sa.getHostCard();
+            c2 = tgtCards.get(0);
+        } else {
+            c1 = tgtCards.get(0);
+            c2 = tgtCards.get(1);
+        }
+        return c1 + " exchanges text box with " + c2 + ".";
+    }
+
+    @Override
+    public void resolve(final SpellAbility sa) {
+        final List<Card> tgtCards = getTargetCards(sa);
+        final Card host = sa.getHostCard();
+        if (tgtCards.isEmpty()) {
+            return;
+        }
+
+        final Card c1;
+        final Card c2;
+        if (tgtCards.size() == 1) {
+            c1 = host;
+            c2 = tgtCards.get(0);
+        } else {
+            c1 = tgtCards.get(0);
+            c2 = tgtCards.get(1);
+        }
+
+        // snapshot the original text boxes before modifying
+        final Card copy1 = CardCopyService.getLKICopy(c1);
+        final Card copy2 = CardCopyService.getLKICopy(c2);
+
+        final Game game = host.getGame();
+        final long ts = game.getNextTimestamp();
+
+        swapTextBox(c1, copy2, ts);
+        swapTextBox(c2, copy1, ts);
+
+        game.fireEvent(new GameEventCardStatsChanged(c1));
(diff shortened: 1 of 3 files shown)


Agetian:
Should this card be RemAIDeck for now? I might actually think of some AI for this, not sure what the logic should be though off the top of my head :/

forge-game/src/main/java/forge/game/ability/effects/TextBoxExchangeEffect.java
```diff
@@ -0,0 +1,124 @@
...
+
+        swapTextBox(c1, copy2, ts);
+        swapTextBox(c2, copy1, ts);
+
+        game.fireEvent(new GameEventCardStatsChanged(c1));
+        game.fireEvent(new GameEventCardStatsChanged(c2));
+    }
+
+    private static void swapTextBox(final Card to, final Card from, final long ts) {
+        to.getChangedCardTraitsByText().clear();
```
Hanmac, line 73: imo clearing the CardTraitsByText might be dangerous

and I wish there was a way to code this via Static Layers @tool4ever

forge-gui/res/cardsfolder/d/deadpool_trading_card.txt
```diff
@@ -0,0 +1,10 @@
+Name:Deadpool, Trading Card
+ManaCost:2 B R
+Types:Legendary Creature Mutant Mercenary Hero
+PT:5/3
+K:ETBReplacement:Other:DBExchangeText:Optional
+SVar:DBExchangeText:DB$ ExchangeTextBox | ValidTgts$ Creature.Other | TgtPrompt$ Choose a creature to exchange text boxes with | SpellDescription$ You may exchange CARDNAME's text box and another creature's.
```
tool4ever, line 6: this doesn't target
do ChooseCardEffect first

forge-game/src/main/java/forge/game/ability/effects/TextBoxExchangeEffect.java
```diff
@@ -0,0 +1,124 @@
...
+        if (tgtCards.size() == 1) {
+            c1 = host;
+            c2 = tgtCards.get(0);
+        } else {
+            c1 = tgtCards.get(0);
+            c2 = tgtCards.get(1);
+        }
+
+        // snapshot the original text boxes before modifying
+        final Card copy1 = CardCopyService.getLKICopy(c1);
```
tool4ever, line 59: I suppose extra LKI gets you around some additional code but it would probably still be cleaner/safer if you'd just prepare the Arrays for each card and only after modify both

forge-game/src/main/java/forge/game/ability/effects/TextBoxExchangeEffect.java
```diff
@@ -0,0 +1,124 @@
...
+        for (StaticAbility st : src.getStaticAbilities()) {
+            if (st.isIntrinsic()) {
+                statics.add(st.copy(to, false));
+            }
+        }
+        to.addChangedCardTraitsByText(spells, triggers, reps, statics, ts, 0);
+
+        List<KeywordInterface> kws = Lists.newArrayList();
+        for (KeywordInterface kw : src.getIntrinsicKeywords()) {
+            kws.add(kw.copy(to, false));
```
tool4ever, line 114: not quite sure why you switched you logic here
should just also call `getKeywords()` then you don't need to mess with `getChangedCardKeywordsByText()` above

forge-gui/res/cardsfolder/d/deadpool_trading_card.txt
tool4ever, line 6: SpellDescription doesn't quite match

kvn1338:
Hey guys, thank you for your feedback I made some changes along those lines.

This is my first contribution to Forge so I'm so familiar with how to do stuff. From my testing everything seems to work fine.

forge-game/src/main/java/forge/game/ability/effects/TextBoxExchangeEffect.java
```diff
@@ -0,0 +1,163 @@
...
+        swapTextBox(c2, data1, ts);
+
+        game.fireEvent(new GameEventCardStatsChanged(c1));
+        game.fireEvent(new GameEventCardStatsChanged(c2));
+    }
+
+    private static void swapTextBox(final Card to, final TextBoxData from, final long ts) {
+        to.getChangedCardTraitsByText().clear();
+        to.clearChangedCardKeywords(false);
+        to.copyChangedTextFrom(from.textHolder);
```
tool4ever, line 69: Ok, so the reason why you might have thought you needed to copy the previous text changes extra is that normally `CardTraitBase.copyHelper` should not apply them again

However we really shouldn't mess with it manually considering there's also a duration-based exchange already printed (_Exchange of Words_)

`SpellAbility.copy()` is already overloaded to support that, probably that should be done for the other traits too 🤔

forge-game/src/main/java/forge/game/ability/effects/TextBoxExchangeEffect.java
```diff
@@ -0,0 +1,124 @@
...
+
+        final Card c1;
+        final Card c2;
+        if (tgtCards.size() == 1) {
+            c1 = host;
+            c2 = tgtCards.get(0);
+        } else {
+            c1 = tgtCards.get(0);
+            c2 = tgtCards.get(1);
+        }
```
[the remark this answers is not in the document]
tool4ever: I've considered it, but it's not really relevant right now

We might add it later if this effect survives on its own?

tool4ever (approved):

PR #7643, opened 2025-05-22
~fix Token ImageKeys without Set

Hanmac:


tool4ever (approved):

Hanmac:
@tehdiplomat we might need to find a good image source for stuff like "hidden" which doesn't have a set

(and also I couldn't find a way to get a "back" image from Scryfall)

PR #7645, opened 2025-05-22
 Edgar, King of Figaro and support

tool4ever:
Effect needed cleanup.

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/FlipCoinEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/FlipCoinEffect.java
@@ -63,7 +63,7 @@ public class FlipCoinEffect extends SpellAbilityEffect {
         for (final Player flipper : playersToFlip) {
             if (noCall) {
                 int countHeads = flipCoins(flipper, sa, amount);
-                int countTails = countHeads - amount;
+                int countTails = Math.abs(countHeads - amount);
                 if (countHeads > 0) {
                     if (sa.hasParam("RememberResult")) {
                         host.addFlipResult(flipper, "Heads");
@@ -132,7 +132,7 @@ public class FlipCoinEffect extends SpellAbilityEffect {
                 }
             } else {
                 int countWins = flipCoins(flipper, sa, amount);
-                int countLosses = countWins - amount;
+                int countLosses = Math.abs(countWins - amount);
                 if (countWins > 0) {
                     if (sa.hasParam("RememberWinner")) {
                         host.addRemembered(flipper);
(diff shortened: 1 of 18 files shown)


forge-game/src/main/java/forge/game/ability/effects/FlipCoinEffect.java
```diff
@@ -196,57 +130,59 @@ public void resolve(SpellAbility sa) {
...
-     */
-    public boolean flipCoinNoCall(final SpellAbility sa, final Player flipper, final int multiplier, final String varName) {
-        boolean result = false;
-        int numSuccesses = 0;
-
+    public static int flipCoins(final Player flipper, final SpellAbility sa, final int amount) {
+        return flipCoins(flipper, sa, amount, "");
+    }
+    public static int flipCoins(final Player flipper, final SpellAbility sa, final int amount, final String info) {
+        int multiplier = getFlipMultiplier(flipper);
```
Hanmac, line 172: @tool4ever i wish we could do something about this hidden Keyword: 
`"If you would flip a coin, instead flip two coins and ignore one."`

Hanmac:
@tool4ever can this be cleaned up too?

https://github.com/Card-Forge/forge/blob/6dc508b631ccc6c6b612e068e696edc040b2aeab/forge-gui/res/cardsfolder/g/goblin_assassin.txt#L6-L7

Would This Work?
`FlipACoin | Flipper$ Player | NoCall$ True | RememberResult$ True`

tool4ever:
Only two scripts even use RememberResult 🤔
and logic-wise one can freely treat Heads=Win
so maybe more cleanup later?

forge-game/src/main/java/forge/game/ability/effects/FlipCoinEffect.java
```diff
@@ -196,57 +131,59 @@ public void resolve(SpellAbility sa) {
...
-            for (int i = 0; i < multiplier; i++) {
-                flipResults.add(MyRandom.getRandom().nextBoolean());
-            }
-            flipper.getGame().fireEvent(new GameEventFlipCoin());
-            result = flipResults.size() == 1 ? flipResults.iterator().next() : flipper.getController().chooseFlipResult(sa, flipper, BOTH_CHOICES, false);
-            if (result) {
-                numSuccesses++;
+            Boolean fixedResult = StaticAbilityFlipCoinMod.fixedResult(flipper);
+            for (int i = 0; i < amount; i++) {
+                won = flipCoin(flipper, sa, multiplier, fixedResult,"");
```
Hanmac, line 179: @tool4ever shouldn't you pass info there?

Hanmac:
> and logic-wise one can freely treat Heads=Win
> so maybe more cleanup later?

There are triggers happen for "whenever you win a coin flip", these should not trigger for cards like `Two-Headed Giant` and `Goblin Assassin`

That's why it can't be used that easy


But from what i have seen, `Goblin Assassin` is the only `NoCall` one without `HeadsSubAbility` or `TailsSubAbility`

Hanmac (approved):

PR #7646, opened 2025-05-22
FIN: 3 cards

dracontes:
Tested to satisfaction:
- [Ancient Adamantoise](https://scryfall.com/card/fin/172/ancient-adamantoise)
- [Sidequest: Hunt the Mark // Yiazmat, Ultimate Mark](https://www.reddit.com/r/magicTCG/comments/1ksgjmg/fin_sidequest_hunt_the_mark_yiazmat_ultimate_mark/)
- [Stolen Uniform](https://www.reddit.com/r/magicTCG/comments/1ksern5/fin_stolen_uniform_ff6/)

Cleanup:
- The Regalia


forge-gui/res/cardsfolder/upcoming/stolen_uniform.txt
```diff
@@ -0,0 +1,11 @@
...
+A:SP$ Pump | ValidTgts$ Creature.YouCtrl | TgtPrompt$ Select target creature you control | RememberObjects$ ThisTargetedCard | SubAbility$ DBPumpTwo | SpellDescription$ Choose target creature you control and target Equipment. Gain control of that Equipment until end of turn. Attach it to the chosen creature. When you lose control of that Equipment this turn, if it's attached to a creature you control, unattach it.
+SVar:DBPumpTwo:DB$ Pump | ValidTgts$ Equipment | ImprintCards$ ThisTargetedCard | SubAbility$ DBGainControl
+SVar:DBGainControl:DB$ GainControl | Defined$ Imprinted | LoseControl$ EOT | SubAbility$ DBAttach
+SVar:DBAttach:DB$ Attach | Object$ Imprinted | Defined$ Remembered | ForgetRemembered$ True | SubAbility$ DBDelayTrig
+SVar:DBDelayTrig:DB$ DelayedTrigger | Mode$ ChangesController | ValidCard$ Card.IsTriggerRemembered | ValidOriginalController$ You | RememberObjects$ Imprinted | ConditionDefined$ Imprinted | ConditionPresent$ Card.Equipment+AttachedTo Creature.YouCtrl | ThisTurn$ True | Execute$ TrigUnattach | SubAbility$ DBCleanup | TriggerDescription$ When you lose control of that Equipment this turn, if it's attached to a creature you control, unattach it.
```
tool4ever, line 8: the Condition you used is for resolving abilities, you need to script it like a normal trigger condition

forge-gui/res/cardsfolder/upcoming/stolen_uniform.txt
```diff
@@ -0,0 +1,11 @@
+Name:Stolen Uniform
+ManaCost:U
+Types:Instant
+A:SP$ Pump | ValidTgts$ Creature.YouCtrl | TgtPrompt$ Select target creature you control | RememberObjects$ ThisTargetedCard | SubAbility$ DBPumpTwo | SpellDescription$ Choose target creature you control and target Equipment. Gain control of that Equipment until end of turn. Attach it to the chosen creature. When you lose control of that Equipment this turn, if it's attached to a creature you control, unattach it.
+SVar:DBPumpTwo:DB$ Pump | ValidTgts$ Equipment | ImprintCards$ ThisTargetedCard | SubAbility$ DBGainControl
+SVar:DBGainControl:DB$ GainControl | Defined$ Imprinted | LoseControl$ EOT | SubAbility$ DBAttach
```
tool4ever, line 6: I see no reason why you can't target the Equipment with the effect for it
(unfortunately not for the Attach too because they decided on a different targeting order :/)

forge-gui/res/cardsfolder/upcoming/stolen_uniform.txt
dracontes: Well, at least figured out only `ValidCard$` was needed here. New script version working as expected.

forge-gui/res/cardsfolder/upcoming/stolen_uniform.txt
dracontes: I got nowhere trying to be terse, even inverting the targeting order, because as best I can tell, `GainControl` lines don't support either `Imprint$ True` or `ImprintCards$` and imprinting the Equipment seems necessary for the delayed trigger to work.

dracontes: Okay, managed to figure it out, while taking into account the NEO's creature Equipment cards.

tool4ever: I'm not confident equipment creatures wouldn't mess it up now
imo just remember the first target and use ParentTargeted for the other

forge-gui/res/cardsfolder/upcoming/stolen_uniform.txt
```diff
@@ -0,0 +1,9 @@
+Name:Stolen Uniform
+ManaCost:U
+Types:Instant
+A:SP$ Pump | ValidTgts$ Creature.YouCtrl | TgtPrompt$ Select target creature you control | SubAbility$ DBGainControl | SpellDescription$ Choose target creature you control and target Equipment. Gain control of that Equipment until end of turn. Attach it to the chosen creature. When you lose control of that Equipment this turn, if it's attached to a creature you control, unattach it.
+SVar:DBGainControl:DB$ GainControl | ValidTgts$ Equipment | Defined$ ThisTargetedCard | LoseControl$ EOT | SubAbility$ DBAttach
```
tool4ever, line 5: why add Defined when it's the same as Tgts?

forge-gui/res/cardsfolder/upcoming/stolen_uniform.txt
tool4ever: wrong change, scripting it like that isn't "intervening if"

I'm not aware of any scripts with wrong usage

forge-gui/res/cardsfolder/upcoming/stolen_uniform.txt
dracontes: Refreshed my memory of reconfigure cards namely the quote below, which meant `nonCreature` was an unnecessary restriction for Stolen Uniform's effect.
> "An Equipment creature with reconfigure can be attached to creatures by effects other than its reconfigure ability, such as the activated ability of Brass Squire." - [Reconfigure on MTG Wiki](https://mtg.wiki/page/Reconfigure)

I did however come across some odd behavior.
- Loading a game state file with a reconfigure Equipment attached to a creature results in a bug report, from which you can continue playing (log, bug report and game state file appended),
[forge.log](https://github.com/user-attachments/files/20399849/forge.log)
[2025-05-22-00_reconf_game.txt](https://github.com/user-attachments/files/20399859/2025-05-22-00_reconf_game.txt)
[state_stolenu_recon.txt](https://github.com/user-attachments/files/20399920/state_stolenu_recon.txt)
- When following Stolen Uniform's resolution process for a targeted reconfigure Equipment that's a creature, there's this step along the way that seems a bit odd, considering it didn't happen for the version of the script with `nonCreature` (gane state file appended). The more recent script does otherwise behave as expected.
[state_stolenu_recon_3.txt](https://github.com/user-attachments/files/20399961/state_stolenu_recon_3.txt)
image

tool4ever: you scripted it so there can be two options for only one attachment, so I'm a bit confused why this confuses you :)
`Defined$ Valid Creature.targetedBy`

forge-gui/res/cardsfolder/upcoming/stolen_uniform.txt
dracontes: I hope that fixes all the issues I can fix on my side,

Well, regarding delayed triggers with "intervening if" and the `Condition` pair there's [Infinite Authority](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/i/infinite_authority.txt). Now that you've mentioned it, I'm also not too sure about other delayed triggers like those of the versions of [Alora, Rogue Companion](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/a/alora_rogue_companion.txt).

forge-gui/res/cardsfolder/upcoming/stolen_uniform.txt
tool4ever: getting better
but you should still just use:
```
Object$ ParentTarget
Defined$ Remembered
RememberObjects$ ParentTarget
```

instead of doing this awkward manual filtering of the battlefield

also ForgetRemembered$ doesn't exist there...

forge-gui/res/cardsfolder/upcoming/stolen_uniform.txt
tool4ever: the effect of _Infinite Authority_ (Destroy) can't be reverted (unlike Attach)
so logic-wise it's easier to just check directly if it even makes sense to create the trigger

while I don't recommend looking at Alchemy scripts for proper rules implementation there are no "intervening if" at all on any Alora

forge-gui/res/cardsfolder/upcoming/stolen_uniform.txt
dracontes: Well, I abandoned `ParentTarget` as an anchor(?) as I got nowhere with it on the initial terse version. Just tested the last version of it with reasonable alterations (intervening if) and still a no-go.

I take it `ParentTarget` refers to the last previous line that has a `ValidTgts$` parameter (and its argument) and not the first in the effect sequence.

In any case, substituting the arguments you listed where relevant do have the current version of the script working in both regular and reconfigure equipment cases.

tool4ever (approved):

PR #7650, opened 2025-05-23
Update carrionette.txt

dracontes:
It came to my attention that [Carrionette](https://scryfall.com/card/tmp/111/carrionette)'s activated ability was able to target creature cards in graveyards. something it shouldn't do per rules text. The targeting is herein restricted to correct this behavior.

carrionette


Hanmac:
@dracontes probably `TgtZone$ Battlefield` better?

tool4ever:
> @dracontes probably `TgtZone$ Battlefield` better?

I suspect that might still lead to conflict with `adjustChangeZoneTarget`

Hanmac (approved):

PR #7657, opened 2025-05-23
Flavor for Equip

### LINKED ISSUE #7533: Support flavor words in Keywords
Yeah, we can skip the flavor words for now

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/issues/7520#issuecomment-2869526931_

Northmoc:
- closes #7533

image

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
@@ -3004,6 +3004,14 @@ public class CardFactoryUtil {
                 System.err.println("Malformed Equip entry! - Card: " + card.toString());
                 return;
             }
+            boolean hasFlav = false;
+            String flavor = "";
+            // Flavor keyword titles should be last in the card script K: line
+            if (keyword.contains(":Flavor ")) {
+                flavor = (keyword.split(":Flavor ", 2)[1]) + (" — ");
+                keyword = keyword.substring(0, keyword.indexOf(":Flavor "));
+                hasFlav = true;
+            }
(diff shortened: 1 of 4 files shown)


tool4ever:
sorry for the long delay
I was kind of expecting them to flood more keywords with this

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -3013,7 +3021,9 @@ else if (part instanceof CostCollectEvidence) {
             if (card.hasSVar("AttachAi")) {
                 abilityStr.append("| ").append(card.getSVar("AttachAi"));
             }
-            abilityStr.append(" | PrecostDesc$ Equip");
+            abilityStr.append(" | PrecostDesc$ ");
```
tool4ever, line 3035: instead of this code just store it in the keyword?
and add it here for all:
https://github.com/Card-Forge/forge/blob/d4a484522e67a0be59dcdb97e1f5054600575ed2/forge-game/src/main/java/forge/game/spellability/SpellAbility.java#L1059

tool4ever: `formatSpellAbility` could then add italic tags via `replaceAll` on `sa.getKeyword().getFlavor()`

Hanmac: @tool4ever formatSpellAbility would need to work with Ability words like "Landfall" too
so making it only work for Keywords is not the right way

Hanmac (approved):

PR #7665, opened 2025-05-24
(Adventure Mode): Fix abandoned town

Arguas:
Adventure Mode: Fix Quest_APortalToNowhere, in which the file path got moved but the point of interest json was not updated through the editor to accommodate it.

I spent some time to go through this dungeon and everything seems to work, so I figure we should just turn this back on!


tehdiplomat:
THere seems to be a TON of changes to this file. I can't even tell what changes are real? Can we push something that just fixes the abandoned town without all of the other file formatting changes?

Arguas:
One thing that helps is if you go to the gear in the diff at the very top and select the checkbox to hide whitespace changes, but it still is a lot more of indenting changes in that case. The only real change is the abandoned town. 

I generated this change by using the adventure editor tool, changed the file path, and saved the result. I can directly in-line make the same change to the file, and put that up for review. Will do so shortly.

Arguas:
OK, direct change within the point of interest json applied. We probably figure out why using the tool changed between tabs/spaces.

PR #7666, opened 2025-05-24
(Adventure) Fix draft scene transition

Arguas:
Fixing a bug in which the transition screen's non-blocking of the start match button can be clicked multiple times, which results in a crash when the match ends, regardless of outcome.

Quick comment, this also removes two useless `continue;` calls that the editor flags as such because they are the last lines of their respective blocks before the end of their for loop iteration anyway, so just doing very safe code cleanup while in the same file. If you want me to restore it, I'm perfectly OK with that, just say so.


Arguas:
Bumping

tehdiplomat (approved):

PR #7669, opened 2025-05-24
Adventure: added support for dynamic deck count

SprinkleMeTimbers:
When playing Adventure for a while, it's possible to run out of deck slots when trying different builds. Currently the deck count is statically set to 10 due to the underlying structure of the data (array) and storage solution (trying to see if the user has deck of an index stored, for each index up till 10).

This MR replaces the underlying structure (with an ArrayList) and the save load mechanism to allow dynamic deck counts. I've extended the adventure deck editor scene/UI to support adding and deleting decks. The player is always bound to have at least 10 decks (trying to delete below the minimum limit just clears that deck, as was the case before). The save/load checks whether the deck count is saved and if so, loads decks up till the count, otherwise proceeds with the legacy load.

I've also added a utility method for an integer-selecting combo box - didn't really need it in the end but might come in handy for someone else later on.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
+++ b/forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
@@ -459,6 +459,8 @@ public class AdventurePlayer implements Serializable, SaveFileContent {
         boolean hasDynamicDeckCount = data.containsKey("deckCount");
         if (hasDynamicDeckCount) {
             int dynamicDeckCount = data.readInt("deckCount");
+            // in case the save had previously saved more decks than the current version allows (in case of the max being lowered)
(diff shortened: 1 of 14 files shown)


tehdiplomat:
We had a conversation recently about this feature request, and there was concerns about adding this capability leading to issues with adventure mode itself. I don't think we should be adding this without really understanding how it will impact loading and saving adventure saves. I'd much rather solve this problem by export/importing decks.

tehdiplomat:
THe conversation was before your PR existed. I wouldn't necessarily say anything about your PR is the problem, more about what would the extra data lead to potential save corruptions. That's an issue we already have in our quest mode, and I worry 40 more decks would cause a serious burden in that area.

tehdiplomat (approved):
THanks for bearing with us.

PR #7676, opened 2025-05-25
(Adventure) Add booster pack grant debug

Arguas:
This is a new debug command to be able to instantly grant a booster pack of the type specified within its command in order to test opening packs. The pack granter checks to see if a booster template exists for the corresponding pack to exist before granting.


Arguas:
Bumping

tool4ever:
Thank you, we will merge this functionality from above PR and all should be good

PR #7702, opened 2025-05-27
Diamond Weapon (FIN)

Fulgur14:


forge-gui/res/cardsfolder/upcoming/diamond_weapon.txt
```diff
@@ -0,0 +1,9 @@
+Name:Diamond Weapon
+ManaCost:7 G G
+Types:Legendary Artifact Creature Elemental
+PT:8/8
+S:Mode$ ReduceCost | ValidCard$ Card.Self | Type$ Spell | Amount$ X | EffectZone$ All | Description$ This spell costs {1} less to cast for each permanent card in your graveyard.
+K:Reach
+K:Prevent all combat damage that would be dealt to CARDNAME.
```
tool4ever, line 7: just write as RE directly so you can include the flavor word

Hanmac: Also i don't like that Keyword anyway :/

tool4ever: will be tricky to remove completely because of _Glittering Lynx_ 🤔

Hanmac: Yeah, but any other instances of the keyword can be removed
And for _Glittering Lynx_, I will think of something else to make it work

PR #7728, opened 2025-05-29
Details Labels and Zooming on Adventure World Map

MD200210:
Multiple Buttons which cycle Visibility when being pressed:
Details -> Events: Display Drafts/Jumpstarts at Inns
Events -> Reputation: Remove Labels. Display Town Reputation if != 0
Reputation -> Names: Remove Labels. Display Names of visited Dungeons (visited is saved in PointOfInterestChanges)
Names -> Details: Remove Labels

Zooming by creating a new ScrollPane without a ScrollListener and then attaching a new Input Listener which activates the Zoom methods
WorldMap image gets scaled and moved to maintain viewing position
Labels and MiniMapPlayer get moved as well


MD200210:
I'll probably remake this Pull Request to fix the extra Commits and the mising Line Break in fin.rnk

PR #7729, opened 2025-05-29
Preventing negative power to cause power drop with +X effects.

Fulgur14:


Fulgur14:
> Would it be easier to do this check in the X-math logic or the pump effect logic instead of every card that uses it? What cards shouldn't output 0 when computing negative power values?

Basically, apart from doubling, things that set something that can be negative (like power, toughness, or life total) to a value.

tool4ever:
not relevant

PR #7736, opened 2025-05-30
The FINAL Fantasy cards

Fulgur14:


forge-gui/res/cardsfolder/upcoming/blitzball.txt
```diff
@@ -0,0 +1,8 @@
+Name:Blitzball
+ManaCost:3
+Types:Artifact
+A:AB$ Mana | Cost$ T | Produced$ Any | SpellDescription$ Add one mana of any color.
+A:AB$ Draw | PrecostDesc$ GOOOOAAAALLL! — | Cost$ T Sac<1/CARDNAME> | NumCards$ 2 | CheckSVar$ X | SpellDescription$ Draw two cards. Activate only if an opponent was dealt combat damage by a legendary creature this turn.
+SVar:X:PlayerCountRegisteredOpponents$HasPropertywasDealtCombatDamageThisTurnBy Card.Legendary GE 1
```
tool4ever, line 6: bad type & syntax

tool4ever: currently the check stops at any creature which I'll have to fix
but even after it would then fail to parse the amount

tool4ever (approved):

PR #7757, opened 2025-06-01
ImageKeys: unify PaperCard imageKeys

### LINKED ISSUE #7532: Specialize ImageKey refactor
Something funky with Specialize i found:

PaperCard does define extra functions:
https://github.com/Card-Forge/forge/blob/71256972c4685cbfd73c055aa6d55fa6e3952f45/forge-core/src/main/java/forge/item/PaperCard.java#L398-L402

But CardFactory ignores them and uses something different?
https://github.com/Card-Forge/forge/blob/71256972c4685cbfd73c055aa6d55fa6e3952f45/forge-game/src/main/java/forge/game/card/CardFactory.java#L217-L219

But then Card does reset the Imagekey in the CardState anyway and resets them with the ones from PaperCard:
https://github.com/Card-Forge/forge/blob/71256972c4685cbfd73c055aa6d55fa6e3952f45/forge-game/src/main/java/forge/game/card/Card.java#L6528

while Desktop ImageCache somehow messes with both:
https://github.com/Card-Forge/forge/blob/71256972c4685cbfd73c055aa6d55fa6e3952f45/forge-gui-desktop/src/main/java/forge/ImageCache.java#L175-L200

I need to debug what is going on there

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/issues/7479#issuecomment-2853565037_

Hanmac:
Closes #7532

The Goal is to simplify the ImageKeys into a more readable format `c:Name|Set|CN`
and when it is finally used as filename, into `Set/CN_Name.fullborder.jpg`
Or would this be better `Set/Name.CN.fullborder.jpg` ala XMage?

Also updated the scryfall downloader to support artCrop on Desktop
* Split Cards right now have problems with artCrop, but for this i want to talk to Scryfall about it.
* I was experimenting with artCrop for tokens, but it caused more problems
* Front and Back Side might have different Artist, but we can't handle that right now

I don't know if this really was needed?
`CardView.getCardForUi(ImageUtil.getPaperCardFromImageKey(cardView.getCurrentState().getTrackableImageKey()));` 

The Artist Info on Mobile, for this it seems the good way was to add a new Attribute, but it causes crashs on Player Effects like Speed
* For these i think i will make an extra PR, that adds PaperOther so i can better access the Artist from there too
right now, the Artist Info for Tokens falls back to WOTC?
* I need to refactor it anyway because with Artist Attribute, it would leak info from FaceDown Cards, and we cant have that

Specialize does have different CollectorNumber needed for Scryfall:
* `c:Jaheira, Ruthless Harper|HBG|16b` is turned into `HBG/16b_Jaheira, Ruthless Harper.fullborder.jpg`
* but that makes finding EditionEntry or PaperCard by CollectorNumber but that's only needed for ArtCrop?

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardView.java
+++ b/forge-game/src/main/java/forge/game/card/CardView.java
@@ -1335,6 +1335,10 @@ public class CardView extends GameEntityView {
         }
         void updateImageKey(Card c) {
             set(TrackableProperty.ImageKey, c.getImageKey());
+            // currently only works for Cards
+            if (!c.getGamePieceType().equals(GamePieceType.CARD)) {
+                return;
+            }
             IPaperCard pc = c.getPaperCard();
             if (pc != null) {
                 set(TrackableProperty.Artist, pc.getArtist());
@@ -1342,6 +1346,10 @@ public class CardView extends GameEntityView {
         }
         void updateImageKey(CardState c) {
             set(TrackableProperty.ImageKey, c.getImageKey());
+            // currently only works for Cards
+            if (!c.getCard().getGamePieceType().equals(GamePieceType.CARD)) {
+                return;
+            }
             IPaperCard pc = c.getCard().getPaperCard();
             if (pc != null) { // currently Artist is per Card
                 set(TrackableProperty.Artist, pc.getArtist());


Agetian:
Will this be also compatible with the file name scheme we're using at the moment? For the users of full card art collections like myself, it'll be a total nightmare trying to rename all the cards to the new format. If the file naming scheme changes completely, I'd say we need some kind of a converter to convert old collections to the new format. I wish it were unnecessary and we had a Scryfall downloader that would grab all the proper art (like in XMage), but we don't have that, unfortunately, so that's a moot point :( What do you think?

Hanmac:
> Will this be also compatible with the file name scheme we're using at the moment? For the users of full card art collections like myself, it'll be a total nightmare trying to rename all the cards to the new format. If the file naming scheme changes completely, I'd say we need some kind of a converter to convert old collections to the new format. I wish it were unnecessary and we had a Scryfall downloader that would grab all the proper art (like in XMage), but we don't have that, unfortunately, so that's a moot point :( What do you think?

i try to add fallbacks, but the new Filename will have CollectorNumber instead of Artindex so SecretLair doesn't mess up our files anymore

Hanmac:
The different situations of ImageKeys: (With/Without Set, With/Without CN, With/Without Artist, maybe back)
lets me think it might be better to have them as Key-Value:
`c:Jaheira, Ruthless Harper|HBG|16b` into `c:Name$ Jaheira, Ruthless Harper | Set$ HBG | CN$ 16b | Artist$ Mila Pesic`

Then I don't need to ask the Edition/EditionEntry when I want to display the Artist

otherwise, displaying the Artist from the Card might show hidden info

Jetz72:
> The different situations of ImageKeys: (With/Without Set, With/Without CN, With/Without Artist, maybe back) lets me think it might be better to have them as Key-Value: `c:Jaheira, Ruthless Harper|HBG|16b` into `c:Name$ Jaheira, Ruthless Harper | Set$ HBG | CN$ 16b | Artist$ Mila Pesic`
> 
> Then I don't need to ask the Edition/EditionEntry when I want to display the Artist

Sounds like the amount of data here is growing beyond what a string is good at storing. Maybe the image key needs to be expanded into a full on class/record for containing image metadata? Then we could pile on whatever extra metadata we want  - specialize mode, front or back, artist, flipped flag, whether we even expect it exists on Scryfall or if it's something local like an Adventure card or placeholder. It could just have its own methods for translating itself into a filename, a legacy filename, a Scryfall URL, a hashcode, etc.

Hanmac:
Blocked by #7846

PR #7782, opened 2025-06-03
Update fin.rnk

dracontes:
Correction per [Scryfall](https://scryfall.com/card/fin/162/summon:-gf-cerberus).


tehdiplomat (approved):
Oh i thought i had caught this one

PR #7787, opened 2025-06-04
Adventure:  Made tooltips instant and reduced size

EfourC:
I felt that the card tooltips in rewards/shops were too slow and too large, so I propose some changes here.

This PR:
1. Makes tooltips in a RewardScene instant (like in matches).  It does this by extending the ToolTipManager class to override the built-in libgdx methods that control animations for tooltips.
2. Reduces the size of the Reward tooltips to 70% of the screen instead of 95% on Desktop. The reduced size is similar to matches (and 95% felt too big on a large monitor). If isAndroid() is true, I left the size as-is since there are so many sizes and form-factors of Android devices (I don't know if a tooltip size change there would be a problem for some users).

I tested in various rewards screens (pickups, quest reward, card packs, shops) on Desktop. Feels snappy and so much nicer with instant tooltips.


tehdiplomat:
@EfourC Looks like you have a merge conflict

EfourC:
@tehdiplomat Thanks! resolved now

tehdiplomat (approved):

PR #7799, opened 2025-06-05
Split Adventure save compatibility handling into separate methods

Jetz72:
Renamed this issue because A) I think the original issue may have been fixed incidentally by the deck editor rewrite, and B) because the save compat stuff has grown to be the main feature of this PR.

This adds version numbers for adventure saves, and some methods that are applied when loading a save from an earlier version than Forge expects. This lets us split a bunch of special case backwards compatibility logic out of the main `AdventurePlayer.load()` method. In the future, any time we make a breaking change to the format of an Adventure Save, we can increment `WorldSave.ADVENTURE_SAVE_VERSION` and add the migration logic to the appropriate methods.

Original issue text below:

---

Attempts to resolve the issue discussed [here](https://discord.com/channels/267367946135928833/1379495497542140034). A user posted a file where A) They owned a deck containing over a thousand cards that weren't actually in their collection, and B) several cards in their auto-sell list also were not in their collection. It's not certain how this happened, but my working theory is that the cards were in both a deck and the auto-sell list, got auto-sold, and weren't removed from the deck as part of this process (because there's no logic to do so).

This PR attempts to solve the issue a bit upstream from the actual issue by preventing cards in auto-sell from being moved directly into decks. I'll see about doing a more direct fix to the sell logic in #7519.

This PR also adds some logic to clean up this issue from existing saves. Any cards that aren't in the collection will be removed from the auto-sell list (factoring in quantities appropriately). Any cards that aren't in the collection but are in decks will be added to the collection. I was going to just remove these from the decks as well, but that would likely cause substantial confusion for other users in this situation. We can change this in the future if this turns into the source of a card duplication bug.

Partially to that end, I've added a value `ADVENTURE_SAVE_VERSION` for adventure saves to track. Doesn't do much for now (And I haven't looked into whether the new dynamic deck count logic's legacy path can merge into it), but in the future we can increment it any time we need to modify the save system, and then check it to determine when migration logic should be applied. I've also added a field that stores the Forge version the file was saved with, for reference when debugging saves in the future.


forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
```diff
@@ -38,6 +40,11 @@ public class AdventurePlayer implements Serializable, SaveFileContent {
     public static final int MIN_DECK_COUNT = 10;
     // this is a purely arbitrary limit, could be higher or lower; just meant as some sort of reasonable limit for the user
     public static final int MAX_DECK_COUNT = 20;
+    /**
+     * Increment this any time a breaking change to a save is made that will require conversion.
+     * Conversion logic can be applied in onSaveVersionBump.
+     */
+    public static final int ADVENTURE_SAVE_VERSION = 1;
```
Jetz72: Just realized I should probably shift this outward to the world save data or header, then pass it in to the load method instead.

Jetz72: Alright, moved it to the world data. It'd fit even better in the header, but since that class is serialized rather than stored as a map, and it currently has no versioning whatsoever, and breaking it would break basically all saves, it'd take someone more confident in their understanding of serialization than me to add it there.

Jetz72:
@kevlahnota I've had this PR for a while to implement versioning and migration handling for Adventure Saves. Do you think it makes sense to roll your inventory item migration logic into something like this?

PR #7818, opened 2025-06-08
Add "Copy Collection to Clipboard" button to adventure mode deck editor

rawbeans:
Adds new button to the adventure mode deck editor that enables users to copy full collection to clipboard.

This is useful when using external tools and websites for deck building.

PR includes translations for all supported languages.

Screenshot 2025-06-07 at 5 51 12 PM

### DIFF
--- a/forge-gui-mobile/src/forge/deck/FDeckViewer.java
+++ b/forge-gui-mobile/src/forge/deck/FDeckViewer.java
@@ -16,7 +16,6 @@ import forge.menu.FPopupMenu;
 import forge.screens.FScreen;
 import forge.screens.match.MatchController;
 import forge.toolbox.FOptionPane;
-import forge.util.ItemPool;
 
 public class FDeckViewer extends FScreen {
     private static FDeckViewer deckViewer;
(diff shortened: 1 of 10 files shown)


tehdiplomat (approved):

forge-gui-mobile/src/forge/deck/FDeckViewer.java
```diff
@@ -72,6 +73,21 @@ public static void copyDeckToClipboard(Deck deck) {
...
+        final String nl = System.lineSeparator();
+        final StringBuilder collectionList = new StringBuilder();
+        java.util.Set<String> accounted = new java.util.HashSet<>();
+        for (final java.util.Map.Entry<PaperCard, Integer> entry : pool) {
+            String cardName = entry.getKey().getCardName();
+            if (!accounted.contains(cardName)) {
+                collectionList.append(entry.getValue()).append(" ").append(cardName).append(nl);
+                accounted.add(cardName);
+            }
+        }
```
Jetz72: I think this would only include the quantity of the first print in the list. Like if you had 3 Forests from M15 and 4 Forests from MH3, it'd only say you had 3 Forests and ignore all subsequent printings. The `copyDeckToClipboard` method uses `countByName` to get the full total across all prints.

Also, why the fully qualified identifiers to `Set`, `HashSet`, and `Map.Entry`?

rawbeans: Good catch. I've updated to use countByName and removed unnecessary fully qualified names.

Jetz72 (approved):

tehdiplomat (approved):

PR #7822, opened 2025-06-08
Cleanup: DBCleanup

dracontes:
Tidying `DBCleanup` lines, testing when appropriate to confirm functionality:
- some standardization
- removing where irrelevant
- establishing links with upstream lines where missing and relevant

Notable incidental fixes:
- Ancient Vendetta: Updated so Prompt and Stack read a bit better for its effect.
- King Narfi's Betrayal: Thought it best to have the exiled cards be remembered by the command zone effect so that information is accessible after the last chapter resolves. Also the effect wasn't exiling itself  after the last of the exiled cards was cast.


forge-gui/res/cardsfolder/b/blue_dragon.txt
```diff
@@ -7,6 +7,5 @@ T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.S
 SVar:TrigPump:DB$ Pump | ValidTgts$ Creature.OppCtrl | TgtPrompt$ Choose target creature an opponent controls (-3/-0) | NumAtt$ -3 | Duration$ UntilYourNextTurn | SubAbility$ DBPump
```
tool4ever, line 7: wrong tgts

tool4ever (approved):

PR #7826, opened 2025-06-09
Improve ai tests

tehdiplomat:


forge-ai/src/main/java/forge/ai/ManaPaymentService.java
```diff
@@ -0,0 +1,1437 @@
...
+                            if (part instanceof CostSacrifice && !part.payCostFromSource()) {
+                                unpreferredCost = true;
+                            }
+                        }
+                        needsLimitedResources = !unpreferredCost;
+                    }
+                }
+
+                // Check for abilities that can produce multiple mana (like Serra's Sanctum)
+                if (m.hasParam("Amount")) {
```
tool4ever, line 1054: that's...a lot of additional code
is this mostly refactoring and only new logic here?

maybe the old can be removed in that case so it's less messy?

tool4ever, line 1054: code block seems overcomplicated
just calling `calculateAmount` should handle all that already

forge-ai/src/main/java/forge/ai/ManaPaymentService.java
```diff
@@ -0,0 +1,1437 @@
...
+        }
+
+        // Sort high output mana sources by potential output (highest first)
+        highOutputManaSources.sort((a, b) -> Integer.compare(
+            manaOutputPotential.getOrDefault(b, 0),
+            manaOutputPotential.getOrDefault(a, 0)
+        ));
+
+        // Add high output sources first - they're most efficient
+        sortedManaSources.addAll(sortedManaSources.size(), highOutputManaSources);
```
tool4ever, line 1130: does this lead to negative impacts?
I guess you added it to nudge it into passing your test...?

but does this mean instead AI will now waste _Serra's Sanctum_ in Main1 for a white one drop and then fail to pay for something bigger in M2? 👎

Hanmac:
@tehdiplomat we need to look at AutoPay and cards like `{T}, Tap an untapped creature you control: Add one mana of any color.` with also another mana ability like `Scene of the Crime`

for some reason i had it that it fails to make me the right mana, it aborts, untaps my stuff, and somehow i can keep the mana

forge-ai/src/main/java/forge/ai/ManaPaymentService.java
tehdiplomat: Yea most of this is refactoring ComputerUtilMana into an object oriented service we can call. Then there's a bit of Agentic to fix a few specific system tests. I left the old stuff around for now just in case the new things were causing any problems and we can turn it off. At some point soon, I plan to remove the old file.

tehdiplomat: Yea the AI got crazy with this one. I'll rewrite it.

forge-ai/src/main/java/forge/ai/ManaPaymentService.java
tehdiplomat: I'll add another test case to verify that we're not overpaying here.

tehdiplomat:
@Hanmac I'll add it to the list when I'm looking into Signets.

Hanmac:
> @Hanmac I'll add it to the list when I'm looking into Signets.

Signets aren't that big of a problem, I think.
I think the main with `Scene of the Crime` is that it has two mana abilities,
that somehow conflict with each other?

PR #7833, opened 2025-06-10
Fix missing card art for all cards from MB1 and PLIST

rawbeans:
**Issue**
About a year ago, Scryfall moved all cards from MB1, MB2 into The List, and updated its set code from PLIST to PLST. This caused an issue in Forge where all card arts from these sets failed to load.

**Solution**
* Update The List edition file to include all cards returned by the Scryfall API
* Update tryGetCard to search in PLST instead of MB1, MB2 and PLIST

This will handle deck lists and save files that use the deprecated sets.

I've included the script I used to update the edition file from Scryfall.

**Possible Next Steps**
* Update all decklists to use PLST with the correct collector number.

I'll leave it up to the maintainers of this repo to decide if this optional step should be carried out. These code changes will work with or without migrating the deck lists. I have included a script in this PR to update all deck lists to use the new PLST set in place of MB1, MB2 and PLIST.

### DIFF
--- a/forge-core/src/main/java/forge/card/CardDb.java
+++ b/forge-core/src/main/java/forge/card/CardDb.java
@@ -655,9 +655,9 @@ public final class CardDb implements ICardDatabase, IDeckGenPool {
             // ALSO: Set Code are always UpperCase
             CardEdition edition = editions.get(reqEditionCode.toUpperCase());
 
-            // Check if this is a MB1, MB2, or PLIST card and try PLST first
+            // Check if this is a MB1 or PLIST card and use PLST instead
             // These sets have been deprecated by Scryfall and replaced with PLST
-            if (Set.of("MB1", "MB2", "PLIST").contains(reqEditionCode)) {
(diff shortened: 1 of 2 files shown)


paulsnoops:
Not had time to review this yet but just a quick note. MB2 has not been deprecated and is not part of PLST. There was also some discussion on Discord about this and the MB1 edition file whether it should be deleted or how the draft for it will work. I will try and get around to looking at this in the next week or so. Thanks for your contribution!

rawbeans:
@paulsnoops Thanks, you are absolutely right. MB2 is still a valid set on Scryfall. I thought I had read somewhere that it was also deprecated and am guilty of only testing API calls with MB1 and PLIST.

I have updated this PR to exclude MB2.

paulsnoops:
@tehdiplomat @Hanmac We good to close this in favour of #7972 ?

kevlahnota:
- closed in favor of #7972

Issue #7771, opened 2025-06-02
New ImageKeys Design Hash/Record


Hanmac:
> The different situations of ImageKeys: (With/Without Set, With/Without CN, With/Without Artist, maybe back) lets me think it might be better to have them as Key-Value: `c:Jaheira, Ruthless Harper|HBG|16b` into `c:Name$ Jaheira, Ruthless Harper | Set$ HBG | CN$ 16b | Artist$ Mila Pesic`
> 
> Then I don't need to ask the Edition/EditionEntry when I want to display the Artist

Sounds like the amount of data here is growing beyond what a string is good at storing. Maybe the image key needs to be expanded into a full on class/record for containing image metadata? Then we could pile on whatever extra metadata we want  - specialize mode, front or back, artist, flipped flag, whether we even expect it exists on Scryfall or if it's something local like an Adventure card or placeholder. It could just have its own methods for translating itself into a filename, a legacy filename, a Scryfall URL, a hashcode, etc.

_Originally posted by @Jetz72 in https://github.com/Card-Forge/forge/issues/7757#issuecomment-2932023655_

Hanmac:
I moved this to extra issue so we can Brainstorm about what such remade Image Key should be able to do.

* become FilePath
* become Scryfall Link
* have CollectorNumber and Artist available (in addition to Name and Set)

Jetz72:
I think it'd make most sense to have a block of metadata that describes everything we'd need to know about a card/token/effect's image. 
I think it'd need:
* Card's set code and collector number.
* Card State Name - Whether it's the front or back, specialize color, if it's a meld face, or other card state info.
* Card Name - Needed at least for generating old-format file paths as fallbacks.
* Card Source - Such as whether the card is an official card on ScryFall or a custom one. See below.
* ~~If the image should be rendered upside down (i.e. flipped state)~~
* Artist

Possibly also:
* [any other data needed to make a Scryfall URL or file name? If I'm forgetting something...]
* Language? Gets a bit complicated.
* If the card has failed to download before? Might be better tracked elsewhere, especially if this class is immutable.
* Functional Variant ID? Might just be part of the collector number, but I think the old-format file paths use it.
* A resource file path if the card is from Forge. Orrrrr, we could move all of Forge's original cards into one folder? Or use the set code to point it to either the Adventure or Quest directory?
* Whether the image is a full-border or art-crop.

[...]

PR #7838, opened 2025-06-10
feature: adds option to prefetch all images when loading a card pool

SethMilliken:
Images are currently downloaded on-demand on mouseover, which can feel sluggish. This change adds a new preference that, when enabled, will prefetch all images from a (non-infinite) card pool when opening it. Primarily for all Sanction Format modes.

Disabled by default.

Added only to desktop. Although the mobile version also loads images on-demand, it does so as soon as the cards are visible, which is less of an issue.

Note: will require adjustment when `imageKey` refactor (#7757) lands.

### DIFF
--- a/forge-gui-mobile/src/forge/screens/settings/SettingsPage.java
+++ b/forge-gui-mobile/src/forge/screens/settings/SettingsPage.java
@@ -495,10 +495,6 @@ public class SettingsPage extends TabPage<SettingsScreen> {
                                 }
                             },
                 4);
-        lstSettings.addItem(new BooleanSetting(FPref.UI_PREFETCH_CARD_POOL_IMAGES,
-                                    Forge.getLocalizer().getMessage("lblPrefetchDeckImages"),
-                                    Forge.getLocalizer().getMessage("nlPrefetchDeckImages")),
-                4);
         lstSettings.addItem(new BooleanSetting(FPref.UI_ENABLE_ONLINE_IMAGE_FETCHER,
                         Forge.getLocalizer().getMessage("cbImageFetcher"),
                         Forge.getLocalizer().getMessage("nlImageFetcher")),


forge-gui-desktop/src/main/java/forge/itemmanager/CardManager.java
```diff
@@ -92,6 +93,17 @@ protected Iterable<Entry<PaperCard, Integer>> getUnique(Iterable<Entry<PaperCard
...
 
+    @Override
+    public void updateView(final boolean forceFilter, final Iterable<PaperCard> itemsToSelect) {
+        prefetchImages();
+        super.updateView(forceFilter, itemsToSelect);
+    }
+
+    private void prefetchImages() {
+        if (isInfinite() || getPool().isEmpty()) return;
+        GuiBase.getInterface().prefetchAllImages(getPool());
```
tehdiplomat: I'm a little concerned about how this is fetching. Thats potentially a lot of downloads immediately, and woud be going against Scryfalls 10 requests per minute limitations. https://scryfall.com/docs/api

SethMilliken: Heard. `ImageFetcher` uses the event dispatch thread, which seems to effectively rate-limit the requests by executing them serially, and each request takes enough time that we stay under the 10 requests/second Scryfall API limit. I can look into making that rate-limiting explicit rather than happenstance.

tehdiplomat: Yea i mostly worry with all of the new contributors that this would get changed without anyone noticing if there's not explicit protections around it. That's part of the reason we wholesale removed the bulk downloaders when switching over to Scryfall download.

forge-gui/src/main/java/forge/util/ImageFetcher.java
```diff
@@ -81,6 +81,15 @@ private String getScryfallDownloadURL(PaperCard c, String face, boolean useArtCr
...
 
+    @FunctionalInterface
+    public interface FetchFunction {
+        void fetch(String imageKey);
+    }
+
+    public FetchFunction fetchWithoutCallbackFunc() {
+        return (String key) -> fetchImage(key, () -> {});
+    }
+
```
Jetz72: Would it work to just have a `fetchImageWithoutCallback(String imageKey)` that forwards to `fetchImage`, and use it in the stream as `getImageFetcher()::fetchImageWithoutCallback`?

Jetz72 (commented):
Maybe there should be an sanity cap on how many fetches can be queued? If someone opened a thousand-plus card quest inventory would this queue all of those?

forge-gui/src/main/java/forge/util/ImageFetcher.java
SethMilliken: Haha. Yes. This is a vestige of a much earlier approach I had taken. Fixed.

SethMilliken:
> Maybe there should be an sanity cap on how many fetches can be queued? If someone opened a thousand-plus card quest inventory would this queue all of those?

This would be easy enough with a `.limit(MAX_FETCH_SIZE)` in the `prefetchAllImages` stream (would 100 be a reasonable number for that constant?). Before I resort to that, though, I'm going to see if I can find a more general solution that can live in `ImageFetcher`, so that we address the rate-limit problem regardless of where the requests originate.

Jetz72:
If you're looking to expand the scope of this and the logic for queuing image downloads, you might consider a priority queue or similar structure. If you have a hundred images queued and the user mouses over a card with no image, it'd be good for that card to be able to skip to the front of the line. 

Might also be worth considering a way to cancel pending downloads if they're no longer of interest - namely if user leaves the screen where they're requested. This may be more trouble than it's worth, though - it'd have to be hooked into all the screens that use it. 

You should also be aware that we're exploring the possibility of changing up image keys from a string to an object. See #7771 and #7757.

SethMilliken:
> You should also be aware that we're exploring the possibility of changing up image keys from a string to an object. See #7771 and #7757.

Yup. I made note of that in my PR description.

SethMilliken:
> If you're looking to expand the scope of this and the logic for queuing image downloads, you might consider a priority queue or similar structure. If you have a hundred images queued and the user mouses over a card with no image, it'd be good for that card to be able to skip to the front of the line.
> 
> Might also be worth considering a way to cancel pending downloads if they're no longer of interest - namely if user leaves the screen where they're requested. This may be more trouble than it's worth, though - it'd have to be hooked into all the screens that use it.

I've been tinkering with some ideas for refactoring `ImageFetcher` that would make doing these easier. Seemed best to wait for those `imageKey` changes to settle first.

forge-gui-desktop/src/main/java/forge/itemmanager/CardManager.java
SethMilliken: Now rate-limiting to 10 req/sec using a `ScheduledExecutorService`. https://github.com/Card-Forge/forge/pull/7838/commits/1cab968e9404724c528d1fc33e5915a781284e1c

SethMilliken:
Moved prefetch method to desktop `CardManager`, the only place it is actually used. Build was broken since `GuiMobile` did not implement `IGuiBase::prefetchAllImages`.

SethMilliken:
Clearly I ought to have made this a "Draft" to begin with.

Latest changes:
- Now naming thread pools
- Switched relevant `System.*` logging to use slf4j `Logger`s instead
- ensure path is always removed from queue on `notifyObservers`

Jetz72:
> * Switched relevant `System.*` logging to use slf4j `Logger`s instead

I think if we want to start utilizing another logger, it's the kind of thing that should be done in a big commit that hits the whole project, not just one area. Not opposed to exploring it, but it'd be good to discuss options and benefits first.

SethMilliken:
> > * Switched relevant `System.*` logging to use slf4j `Logger`s instead
> 
> I think if we want to start utilizing another logger, it's the kind of thing that should be done in a big commit that hits the whole project, not just one area. Not opposed to exploring it, but it'd be good to discuss options and benefits first.

Ah, looks like I misread the repo tea leaves. All the plumbing was already there (slf4j in the `pom`s and  `logback.xml` in resources directory with an originating comment "replace log4j with logback"); I thought I was just turning on the tap. But, yeah, now that I look I don't see us actually using it anywhere else.

I'll get the ball rolling on the independent conversation about logging facilities, and I'll drop that commit before this lands.

Jetz72:
Its come up before ([e.g.](https://github.com/Card-Forge/forge/pull/7365#discussion_r2034338701)). But I don't think the matter has gotten its own dedicated discussion, issue, or PR yet.

PR #7840, opened 2025-06-11
Details Labels and Zooming on Adventure World Map

MD200210:
Multiple Buttons which cycle Visibility when being pressed:
Details -> Events: Display Drafts/Jumpstarts at Inns
Events -> Reputation: Remove Labels. Display Town Reputation if != 0
Reputation -> Names: Remove Labels. Display Names of visited Dungeons (visited is saved in PointOfInterestChanges)
Names -> Details: Remove Labels

Zooming by creating a new ScrollPane without a ScrollListener and then attaching a new Input Listener which activates the Zoom methods
WorldMap image gets scaled and moved to maintain viewing position
Labels and MiniMapPlayer get moved as well


tehdiplomat:
@MD200210 Can you post some screenshots on what it looks like?

MD200210:
image
image
image

MD200210:
Top one is the closest Zoom and very close to the original Zoom of the Map
Bottom one is the farthest Zoom
This is the default screen and Zoom when opening the map
image

PR #7846, opened 2025-06-12
add ImageKey class

### LINKED ISSUE #7771: New ImageKeys Design Hash/Record

Hanmac:
Closes #7771

still in work, but just commited this state

### DIFF
--- a/forge-core/src/main/java/forge/item/ArtStyle.java
+++ b/forge-core/src/main/java/forge/item/ArtStyle.java
@@ -1,0 +1,13 @@
+package forge.item;
+
+public enum ArtStyle {
+    Normal("fullborder", "normal"),
+    Crop("artcrop", "art_crop");
+
+    public final String filename;
+    public final String scryfall;
+    ArtStyle(String filename, String scryfall) {
+        this.filename = filename;
+        this.scryfall = scryfall;
+    }
+}
--- a/forge-core/src/main/java/forge/item/ImageKey.java
+++ b/forge-core/src/main/java/forge/item/ImageKey.java
@@ -3,7 +3,8 @@ package forge.item;
 import java.util.List;
 
 import org.apache.commons.lang3.StringUtils;
-import org.testng.collections.Lists;
+
+import com.google.common.collect.Lists;
 
 import forge.StaticData;
 import forge.card.CardEdition;
--- a/forge-core/src/main/java/forge/item/ImageKey.java
+++ b/forge-core/src/main/java/forge/item/ImageKey.java
@@ -1,0 +1,97 @@
+package forge.item;
+
+import java.util.List;
+
+import org.apache.commons.lang3.StringUtils;
+import org.testng.collections.Lists;
+
+import forge.StaticData;
+import forge.card.CardEdition;
+import forge.card.CardStateName;
+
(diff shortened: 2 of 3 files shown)


Jetz72:
What's the idea behind the `List<String>` of variations of the filepath?

The `ImageType` and `ArtStyle` enums might be simple enough and narrow enough in purpose to be nested within the `ImageKey` class, unless you plan to add more functionality to them later.

Hanmac:
> What's the idea behind the `List<String>` of variations of the filepath?

https://github.com/Card-Forge/forge/blob/7e6f77234544b10d28371f5a33062aa4e4308f85/forge-core/src/main/java/forge/ImageKeys.java#L157-L180

The plan is to use the variations for `ImageKeys::getImageFile`, and then call it in a loop:

```java
for (String s : paths) {
    file = findFile(dir, s);
    if (file != null) {
        cachedCards.put(filename, file);
        return file;
    }
}
```

Also, the first path returned from the List should be used for the Downloader too

Hanmac:
> The `ImageType` and `ArtStyle` enums might be simple enough and narrow enough in purpose to be nested within the `ImageKey` class, unless you plan to add more functionality to them later.

i want to add these too:
https://github.com/Card-Forge/forge/blob/7e6f77234544b10d28371f5a33062aa4e4308f85/forge-core/src/main/java/forge/ImageKeys.java#L16-L22

but i don't know the FileStructure for them yet, nor can these be downloaded from Scryfall (yet)
i want to mail Scryfall to add Booster Images

Jetz72:
> The plan is to use the variations for `ImageKeys::getImageFile`, and then call it in a loop:
> `...`
> Also, the first path returned from the List should be used for the Downloader too

What's all that support for no-set and no-collector-number tokens used for? The Scryfall URL method won't work without those fields, so can anything new be downloaded to those paths?

Hanmac:
> > The plan is to use the variations for `ImageKeys::getImageFile`, and then call it in a loop:
> > `...`
> > Also, the first path returned from the List should be used for the Downloader too
> 
> What's all that support for no-set and no-collector-number tokens used for? The Scryfall URL method won't work without those fields, so can anything new be downloaded to those paths?

It's not just for download, it's for finding images too, like Custom Sets

Hanmac:
@Eradev I need to add your new ⚡ Logic to the Download abilities of this class

Issue #6823, opened 2025-01-22
Adventure Mode: Quality of life UI/input requests


trunksbomb:
**Is your feature request related to a problem? Please describe.**
Adventure mode is fun, but there are some common quality of life features and UI improvements that would make the experience better.

**Describe the solution you'd like**

1. Escape should always bring you either completely out of a screen or one step back. Some screens (the shops, for example) use Enter to leave the shop and Escape does nothing on that screen
2. Dialog boxes should be closable with Escape or Enter (or space?). On multiple choice dialog boxes, this is possible using the arrow keys to navigate first to a button, then hitting Enter. On single button dialogs, though, the arrow keys do not highlight a button and the button is not highlighted by default
3. A preview of the battlefield card granted by a piece of equipment would be nice when hovering that equipment in the shop, so I can make an informed decision about what items I want to buy
5. Better UI clarity in the deck management screen. Some specifics:   
   1. The color and type filter button backgrounds are a black gradient background on a dark blue background - a lighter color or even a light color border would help emphasize these buttons
   2. Likewise, the text highlight on the search filter is hard to see, so it wasn't obvious to me at first that re-focusing the search box was automatically highlighting the text I put in there
   3. Search button appears to always be grayed out and clicking does nothing - only Enter while focused on search box or unfocusing the search box (by clicking somewhere else) will apply your typed search term
   4. After typing a search term and hitting Enter, search box is no longer focused and requires clicking it again to type
   5. Enter on the Number Pad does not submit your search term - only the "real" Enter key does
   6. I'm not intimately familiar with the card type icons in MtG, A simple text label popup when hovering the filter icons would be neat here
   7. Escape does not exit this screen

**Describe alternatives you've considered**
I currently don't have those QOL features, which seems to be my alternative now. 

**Additional context**

trunksbomb:
Added

> There's no "held button prevention" when pressing Escape to bring up the main menu. If you press Escape to bring up the menu but don't let go immediately, then letting go of Escape will exit the menu. This can cause you to unexpectedly close the menu if you held the button for too long.

rawbeans:
I just discovered this game yesterday. I was annoyed by many of these menu UX issues as well. I will start working on fixes and comment here when I have opened PRs. I don't have permission to re-open issues yet.

PR #7848, opened 2025-06-12
Fix several UX issues with menu navigation and dialog boxes

rawbeans:
This PR contains a number of UX improvements and bug fixes for menu navigation in Adventure Mode

The majority of these fixes are related to keybindings that only work when using game pads or Android devices, and do not have keyboard support.

* Change keybinding to exit the shop and other rewards screens from Use to Back. This is consistent with all other menus that let you press ESC to go back. It also prevents accidental purchase of cards when attempting to exit shop.
* Fix enter key not working to save game in save menu
* Fix enter key not working to load game in load menu
* Add support for ESC key to go back from deck editor screen
* Fix main menu flickering issue when opened
* Fix issue where main menu opens and closes with a single key press
* Enable ESC (Back KeyBinding) to dismiss console
* Fix autofocus in dialogs on desktop (with keyboard). This allows you to press enter to dismiss single option dialogs.

This resolves most of the issues listed here: https://github.com/Card-Forge/forge/issues/6823

### DIFF
--- a/forge-gui-mobile/src/forge/Forge.java
+++ b/forge-gui-mobile/src/forge/Forge.java
@@ -1235,7 +1235,15 @@ public class Forge implements ApplicationListener {
             if (keyInputAdapter != null) {
                 return keyInputAdapter.keyUp(keyCode);
             }
-            return false;
+            // if no active key input adapter, give current screen or overlay a chance to handle key
+            FContainer container = FOverlay.getTopOverlay();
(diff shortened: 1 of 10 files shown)


Agetian:
Looks good to me, but I haven't tested this on Android and I'm definitely not the best at figuring out the UI code details, so some assistance in reviewing would be much appreciated :)

PR #7850, opened 2025-06-13
FIN Fix Uncommons

MD200210:
Not splitting the Uncommons caused all 3 to be replaced instead of individually randomizing them

### DIFF
--- a/forge-gui/res/editions/Final Fantasy.txt
+++ b/forge-gui/res/editions/Final Fantasy.txt
@@ -8,7 +8,7 @@ Prerelease=6 Boosters, 1 RareMythic+
 BoosterBox=30
 # https://mtgscribe.com/2025/05/11/final-fantasy-play-booster-fact-sheet/
 BoosterSlots=Common,Common-Guest,Uncommon,RareMythic,AnyLand,Wildcard,WildcardFoil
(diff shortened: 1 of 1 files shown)


Hanmac:
@tehdiplomat can you check this out? IMO, I would like `3 Uncommon` more, then the Replace Rules needs to be investigated

forge-gui/res/editions/Final Fantasy.txt
```diff
@@ -8,7 +8,7 @@ Prerelease=6 Boosters, 1 RareMythic+
 BoosterBox=30
 # https://mtgscribe.com/2025/05/11/final-fantasy-play-booster-fact-sheet/
 BoosterSlots=Common,Common-Guest,Uncommon,RareMythic,AnyLand,Wildcard,WildcardFoil
-Booster=6 Common,1 Common-Guest,3 Uncommon, 1 RareMythic, 1 Wildcard, 1 WildcardFoil+, 1 AnyLand
+Booster=6 Common, 1 Common-Guest, 1 Uncommon, 1 Uncommon, 1 Uncommon, 1 RareMythic, 1 Wildcard, 1 WildcardFoil+, 1 AnyLand
```
tehdiplomat, line 11: Yea this isn't how this should be fixed. If there's an issue with the replacement code, we should fix the replacement code

tehdiplomat:
I'm going to close this PR in favor of fixing the real problem: https://github.com/Card-Forge/forge/pull/7854

PR #7852, opened 2025-06-14
proposal: adopt a logging framework

SethMilliken:
I'd like to open up the conversation on adopting a logging framework. Relying solely on `System.*.print` statements seems brittle and does not provide the flexibility for each person to to regulate their own logging volume and specificity.

I've seen some vestiges of prior attempts at introducing logging frameworks, but none of these appears to have stuck. Is there a strong argument against using a logging framework? Are there issues with cross-platform support, for instance? Is anyone wary of using or unhappy with using a logging framework? Or is this just something that never got sufficient momentum behind it to make it happen?

In case anyone needs convincing, here are some potential advantages of adopting a logging framework:
- Timestamped entries help assess timing and performance issues.
- Thread names in log messages help us better understand concurrency in the app.
- Including class names in log messages helps us more quickly navigate to the callsites of originating messages.
- Assigning a distinct log level to each message enables easy filtering.
- Being able to assign log messages to `DEBUG` and `TRACE` provides additional implicit code documentation without cluttering up the default log experience.
- Adjustable logging thresholds could give us another lever to solicit more detailed bug reports from users.
- Being able to easily log to a file makes log analysis easier, and opens up opportunities to compare logging profiles over time.
- Mapped Diagnostic Context (MDC) support could facilitate more sophisticated visibility, e.g. tracing of specific events as they propagate through the system.

This PR:

- Establishes SLF4J API implemented by logback as our log framework.
- Provides some initial documentation with guidance for using and configuring logging.
- Provides a well-documented, flexible, and easily overridable default logback configuration.
- Adds necessary dependencies to the `pom.xml` (n.b. `logback-core` is a transitive dependency of `logback-classic`)
- Removes vestiges of `log4j`.
- Adds an initial `Logger` with a single log statement in the `Main` class of `forge-gui-desktop`; the minimal amount needed to initialize the on-demand log subsystem.

My hope is that this provides enough for folks to easily kick the tires without having to jump through all of the hoops required by starting from scratch, as well as some concrete examples to ground the conversation.


SethMilliken:
Added [another section](https://github.com/Card-Forge/forge/blob/303f13eec23ec8b4e380c6873a40fbc237d059c4/LOGGING.md#other-logging-in-forge) to try to document other preexisting varieties of logging in the codebase.

SethMilliken:
Removed threadname from default pattern, per Jetz. Parameterized `loggingPattern` variable to faciliate temporarily substituting a different pattern when working on code that may benefit from different details (and for consistency with the other variables).

Edit: And updated documentation to reflect the change.

SethMilliken:
Rebased. Cleaned up some of the documentation based on git spelunking and Discord conversations.

SethMilliken:
Added `slf4j-api` dependency to all submodules to facilitate adding new loggers.

SethMilliken:
Rebased.

SethMilliken:
Rebased.

tool4ever:
@SethMilliken
hey, seems no Dev found the time to look at this closer
unfortunately GitHub doesn't allow reopening this
but it still helped me with testing how we can approach this area (see linked PR) 🏆

Issue #7865, opened 2025-06-19
ConcurrentModificationException


CTimmerman:
**Describe the bug**
The program is responsive but the game had been stuck for hours. Checking forge.log it ends with an exception:

[forge.log](https://github.com/user-attachments/files/20822417/forge.log)

**To Reproduce**
1. Play AI-only games of Prossh vs Rainbow. 

[decks.zip](https://github.com/user-attachments/files/20822455/decks.zip)

2. Wait until error.

**Expected behavior**
No error.

**Screenshots**

![Image](https://github.com/user-attachments/assets/23608757-7930-4128-abd2-09f3b8f2075e)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version: 2.0.4 stable

**Additional context**
I've probably entered these decks in the same session, given the amount of art downloads.

PR #7867, opened 2025-06-19
Fix ConcurrentModificationException (#7865)

CTimmerman:
Fix ConcurrentModificationException (#7865)

I haven't seen 50 tokens with Prossh vs Omnath, Locus of All again yet, but my 20+ test games did not produce an error or get stuck like before.


Agetian (approved):

tool4ever:
it don't really see what could cause it there but it's only a small hotfix anyway 🤷‍♂️

Hanmac:
@tool4ever you might never know where a ConcurrentModificationException is coming from.

There is one Exception on Sentry about `canRegenerate`

Jetz72:
> it don't really see what could cause it there but it's only a small hotfix anyway 🤷‍♂️

Doesn't the AI do combat simulation on multiple threads since late last year? Maybe one of those touched the ManaPool.

Issue #7858, opened 2025-06-15
The Mightstone and Weakstone art won't load


CTimmerman:
**Describe the bug**
The Mightstone and Weakstone art won't load.

**To Reproduce**
1. Go to Deck Editor.
2. Search The Mightstone and Weakstone.
3. See art won't load.

**Expected behavior**
Load the card art.

**Screenshots**

![Image](https://github.com/user-attachments/assets/bcb4414a-ab27-4144-92a6-d1df71ab93bf)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version: 2.0.4 stable

**More context**

forge.log says:
Attempting to fetch: https://api.scryfall.com/cards/bro/238/en?format=image&version=normal
Failed to download card [C:\Users\C\AppData\Local\Forge\Cache\pics\cards\BRO\The Mightstone and Weakstone.full.jpg] image: Can't get input stream from URL!
Attempting to fetch: https://api.scryfall.com/cards/bro/238/?format=image&version=normal
Failed to download card [C:\Users\C\AppData\Local\Forge\Cache\pics\cards\BRO\The Mightstone and Weakstone.full.jpg] image: Can't get input stream from URL!
Attempting to fetch: https://api.scryfall.com/cards/bro/238/en?format=image&version=normal
Failed to download card [C:\Users\C\AppData\Local\Forge\Cache\pics\cards\BRO\Urza, Planeswalker.full.jpg] image: Can't get input stream from URL!
Attempting to fetch: https://api.scryfall.com/cards/bro/238/?format=image&version=normal
Failed to download card [C:\Users\C\AppData\Local\Forge\Cache\pics\cards\BRO\Urza, Planeswalker.full.jpg] image: Can't get input stream from URL!

https://api.scryfall.com/cards/bro/238/en?format=image&version=normal says:

```json
{
  "object": "error",
  "code": "not_found",
  "status": 404,
  "details": "No card found with the given ID or set code and collector number."
}
```

237 and 239 work, so it's a Scryfall problem: https://github.com/scryfall/gatherer-bugs/issues/196

Gatherer, Companion, and Cockatrice all load the card art fine though.

CTimmerman:
Reported to Scryfall via https://scryfall.com/contact because their regular bug form only applies to their site and not API I guess:

> BRO238 does not redirect to BRO238a image like BRO225 does
> 
[...]

PR #7872, opened 2025-06-20
Fix Meld art download (#7858)

CTimmerman:
Fix Meld art download (#7858)


Hanmac (approved):
I will keep that in mind for my later ImageKey fixes

CTimmerman:
Unfortunately, this broke the other half of the meld card, which doesn't share a collectors number: https://github.com/scryfall/gatherer-bugs/issues/196#issuecomment-2993242003

Issue #7401, opened 2025-04-15
New THISTYPE for Description


Hanmac:
I would like to add `THISTYPE` similar to `CARDNAME` for Descriptions

it mostly matters for Cost Parts that does check for "CARDNAME" should check for "THISTYPE" too

Variations:
* "this land"
* "this creature"
* "this artifact"
* "this enchantment"
* "this permanent" (maybe difficult? Could check for Non-Intrinsic?)
* "this token"

uppercase ones
* "this Aura"
* "this Equipment"
* "this Case"
* "this Attraction"

I think that it shouldn't use changed Type, so it could just use the Original Type of CardState
That's why I want to add `getThisType` to CardState

Jetz72:
How would this work for granted abilities, such as from equipment, auras, and animate effects? Do we try and discern the type of object the ability's being applied to?

Hanmac:
@Jetz72 
Smash equipment like `Deconstruction Hammer` still use Original Host Name

`Mortarpod` with `Sacrifice this creature` will fall back to Permanent (until I cheat something)

Hanmac:
I need this for making better Reminder Text of the Station Keyword

Hanmac:
Example function for adding to `CardTraitBase``:

```java
public String getThisType()
{
   CardTypeView type = getCardState().getType();
   if (type.isAura())
       return "Aura";
   if (type.isEquipment())
       return "Equipment";
   if (type.isContraption())
       return "Contraption";
   if (type.isAttraction())
       return "Attraction";

   if (type.hasSubtype("Saga"))
       return "Saga";
   if (type.hasSubtype("Case"))
       return "Case";
   if (type.hasSubtype("Spacecraft"))
       return "Spacecraft";
   if (type.hasSubtype("Planet"))
       return "Planet";

   // with 2 or more core types, its just a permanent
   if (Iterables.size(type.getCoreTypes()) >= (type.isKindred() ? 3 : 2)) {
       return "permanent";
   }
   for (CardType.CoreType ct : type.getCoreTypes()) {
     if (ct == CardType.CoreType.Kindred) {
       continue;
     }
     // TODO add translation
     return ct.name().toLowerCase(Locale.ENGLISH);

}
```

also need Token and nonIntrinsic checks

Hanmac:
[...]

PR #7885, opened 2025-06-22
Add Spacecraft

Hanmac:
Adds new Spacecraft Type

Adds Station Keyword (might need AI logic?)

Allows Legendary Vehicle and Spacecraft as Commander


tool4ever (approved):
LGTM

Hanmac:
> Starting with Edge of Eternities, legendary Vehicles and legendary Spacecrafts **with a printed power and toughness** are eligible to be your deck's commander.

So, I need to do some more changes there :/, no Commander "The Eternity Elevator".

Also I need to do #7401 to fix the Station Description

tool4ever:
I was just about to do the commander part so you can finish Warp

PR #7897, opened 2025-06-27
Creating Multiple Planes in Adventure

shenshinoman:
Updated the code to allow plane swapping via a dropdown in Adventure's settings menu, aswell as building in an alpha and pre-alpha of two new planes, (Innistrad and Crystal_Kingdoms) along with new enemies, updated quests, maps, etc. (Mostly in Innistrad so far.) There is some excess bloat in the two planes folders right now (enemies etc. sometimes bug during the updates right now since everything is so early. So for stability sake, large portions of the common folders are currently duplicated. This will be cleaned back out once they are more stable.)

I am also looking for aid in setting up the change of planes to access updated, unique, jumpstart packs. The requisite boosters-special.txt and printsheets.txt files are in the added subfolders under blockdata (if preferred, these can happily be moved fully back into the individual plane folders. They are just there for original ease of comparison.) As-is, if the data in those sheets is appended to the data inside the current primary sheets (or overwritten with them) everything in Adventure works. However I am trying to incorporate keeping such changes separate from the main core-experience.


shenshinoman:
gonna clear the baggage. closing this pull, will make another when things are cleaner.

PR #7910, opened 2025-07-01
Adventure Config File -  add restrictedEvents, allowedJumpstart, and support per-plane configs

Jetz72:
Adds some support for #7899. In particular, a config file in a plane's directory should now be able to override the one in common. Ideally the config fields should be merged, but that has some logistics that'd need to be worked out (like if lists are additive or not, if it's possible to unset a field...), so this seems like a decent way to add the needed functionality for now.

Also adds `restrictedEvents`, replacing the hardcoded list in `AdventureEventData`. Sets that are designed for Commander Draft have been added to this list.

Adds basic support for `allowedJumpstart`, which if present will replace the default Jumpstart event sets. It's not defined in the default config but was requested for #7899.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/data/AdventureEventData.java
+++ b/forge-gui-mobile/src/forge/adventure/data/AdventureEventData.java
@@ -404,6 +404,11 @@ public class AdventureEventData implements Serializable {
                     legalBlocks.add(b);
                 }
             }
+            for (CardBlock b : FModel.getFantasyBlocks()) {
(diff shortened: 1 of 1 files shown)


shenshinoman:
while not a reviewer. I can confirm, via testing, that the config file update works for plane changes. A la #7899  and does more cleanly on the code side than the config file change I made there. This should take full precedence over that config file, as enabling the plane swaps was the only change made in that config file.

PR #7914, opened 2025-07-02
Keyword: Warp

Hanmac:
from my current understanding (until next week),
it counts as warped when exiled by the Warp ability.

### DIFF
--- a/forge-game/src/main/java/forge/game/GameActionUtil.java
+++ b/forge-game/src/main/java/forge/game/GameActionUtil.java
@@ -248,6 +248,14 @@ public final class GameActionUtil {
                     alternatives.add(foretold);
                 }
 
+                if (source.isWarped() && activator.equals(source.getOwner()) && !source.enteredThisTurn() && !source.getManaCost().isNoCost()) {
+                    final SpellAbility normalWarp = sa.copy(activator);
+                    normalWarp.getRestrictions().setZone(ZoneType.Exile);
+                    // no alternative cost for this
+                    normalWarp.putParam("AfterDescription", "(From Warp)");
+                    alternatives.add(normalWarp);
+                }
+
                 if (activator.canCastSorcery() && source.isPlotted() && source.isInZone(ZoneType.Exile) && activator.equals(source.getOwner()) && !source.enteredThisTurn()) {
                     final SpellAbility plotted = sa.copyWithNoManaCost(activator);
                     plotted.setAlternativeCost(AlternativeCost.Plotted);
--- a/forge-game/src/main/java/forge/game/keyword/Keyword.java
+++ b/forge-game/src/main/java/forge/game/keyword/Keyword.java
@@ -200,6 +200,7 @@ public enum Keyword {
     VANISHING("Vanishing", KeywordWithAmount.class, false, "This permanent enters with {%d:time counter} on it. At the beginning of your upkeep, remove a time counter from it. When the last is removed, sacrifice it."),
     VIGILANCE("Vigilance", SimpleKeyword.class, true, "Attacking doesn't cause this creature to tap."),
     WARD("Ward", KeywordWithCost.class, false, "Whenever this permanent becomes the target of a spell or ability an opponent controls, counter it unless that player pays %s."),
+    WARP("Warp", KeywordWithCost.class, false, "You may cast this card from your hand for its warp cost. Exile this creature at the beginning of the next end step, then you may cast it from ecile on a later turn."),
     WITHER("Wither", SimpleKeyword.class, true, "This deals damage to creatures in the form of -1/-1 counters."),
 
     // mayflash additional cast
--- a/forge-gui/res/cardsfolder/upcoming/anticausal_vestige.txt
+++ b/forge-gui/res/cardsfolder/upcoming/anticausal_vestige.txt
@@ -1,0 +1,10 @@
+Name:Anticausal Vestige
+ManaCost:6
+Types:Creature Eldrazi
+PT:7/5
+T:Mode$ ChangesZone | ValidCard$ Card.Self | Origin$ Battlefield | Execute$ TrigDraw | TriggerDescription$ When this creature leaves the battlefield, draw a card, then you may put a permanent card with mana value less than or equal to the number of lands you control from your hand onto the battlefield tapped.
+SVar:TrigDraw:DB$ Draw | SubAbility$ DBChangeZone
+SVar:DBChangeZone:DB$ ChangeZone | ChangeType$ Permanent.YouCtrl+cmcLEX | Origin$ Hand | Tapped$ True | Destination$ Battlefield | TgtPrompt$ Select permanent card with mana value less than or equal to the number of lands you control
+SVar:X:Count$Valid Land.YouCtrl
+K:Warp:4
+Oracle:When this creature leaves the battlefield, draw a card, then you may put a permanent card with mana value less than or equal to the number of lands you control from your hand onto the battlefield tapped.\nWarp {4} (You may cast this card from your hand for its warp cost. Exile this creature at the beginning of the next end step, then you may cast it from ecile on a later turn.)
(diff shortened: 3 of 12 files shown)


forge-game/src/main/java/forge/game/keyword/Keyword.java
```diff
@@ -200,6 +200,7 @@ public enum Keyword {
     VANISHING("Vanishing", KeywordWithAmount.class, false, "This permanent enters with {%d:time counter} on it. At the beginning of your upkeep, remove a time counter from it. When the last is removed, sacrifice it."),
     VIGILANCE("Vigilance", SimpleKeyword.class, true, "Attacking doesn't cause this creature to tap."),
     WARD("Ward", KeywordWithCost.class, false, "Whenever this permanent becomes the target of a spell or ability an opponent controls, counter it unless that player pays %s."),
+    WARP("Warp", KeywordWithCost.class, false, "You may cast this card from your hand for its warp cost. Exile this creature at the beginning of the next end step, then you may cast it from ecile on a later turn."),
```
squee1968: typo

forge-gui/res/cardsfolder/upcoming/anticausal_vestige.txt
```diff
@@ -0,0 +1,10 @@
+Name:Anticausal Vestige
+ManaCost:6
+Types:Creature Eldrazi
+PT:7/5
+T:Mode$ ChangesZone | ValidCard$ Card.Self | Origin$ Battlefield | Execute$ TrigDraw | TriggerDescription$ When this creature leaves the battlefield, draw a card, then you may put a permanent card with mana value less than or equal to the number of lands you control from your hand onto the battlefield tapped.
+SVar:TrigDraw:DB$ Draw | SubAbility$ DBChangeZone
+SVar:DBChangeZone:DB$ ChangeZone | ChangeType$ Permanent.YouCtrl+cmcLEX | Origin$ Hand | Tapped$ True | Destination$ Battlefield | TgtPrompt$ Select permanent card with mana value less than or equal to the number of lands you control
+SVar:X:Count$Valid Land.YouCtrl
+K:Warp:4
+Oracle:When this creature leaves the battlefield, draw a card, then you may put a permanent card with mana value less than or equal to the number of lands you control from your hand onto the battlefield tapped.\nWarp {4} (You may cast this card from your hand for its warp cost. Exile this creature at the beginning of the next end step, then you may cast it from ecile on a later turn.)
```
squee1968: typo

Hanmac:
there is some leak ala "Warp Equal to their Mana Value"
which means that it probably should be moved to the Place where Flashback:ManaValue is built

forge-gui/res/cardsfolder/upcoming/timeline_culler.txt
```diff
@@ -0,0 +1,8 @@
+Name:Timeline Culler
+ManaCost:B B
+Types:Creature Drix Warlock
+PT:2/2
+K:Haste
+S:Mode$ Continuous | Affected$ Card.Self | MayPlay$ True | ValidSA$ Spell.Warp | AffectedZone$ Graveyard | EffectZone$ Graveyard | Description$ You may cast CARDNAME from your graveyard using its wrap ability.
+K:Warp:B PayLife<2>
+Oracle:Haste\nYou may cast this card from your graveyard using its wrap ability.\nWarp—{B}, Pay 2 life. (You may cast this card from your hand for its warp cost. Exile this creature at the beginning of the next end step, then you may cast it from exile on a later turn.)
```
tool4ever: wrong Oracle

Hanmac: stupid oracle text, one of them said: "if you do" , the other doesn't

forge-gui/res/cardsfolder/upcoming/timeline_culler.txt
```diff
@@ -0,0 +1,8 @@
+Name:Timeline Culler
+ManaCost:B B
+Types:Creature Drix Warlock
+PT:2/2
+K:Haste
+S:Mode$ Continuous | Affected$ Card.Self | MayPlay$ True | ValidSA$ Spell.Warp | AffectedZone$ Graveyard | EffectZone$ Graveyard | Description$ You may cast CARDNAME from your graveyard using its wrap ability.
+K:Warp:B PayLife<2>
+Oracle:Haste\nYou may cast this card from your graveyard using its wrap ability.\nWarp—{B}, Pay 2 life. (You may cast this card from your hand of graveyard for its warp cost. If you do, exile this creature at the beginning of the next end step, then you may cast it from exile on a later turn.)
```
tool4ever: "wrap"

Hanmac:
@tool4ever should we merge this for now?

Or should we do this different?
- "isWarped" might be better as property ala plotted if there is something that can make it "warped" (then i don't need ExiledSA)?
- The Keyword might need to be generated somewhere else if something gives "Warp" Keyword to other Cards/Spells (see Blitz)

forge-game/src/main/java/forge/game/GameActionUtil.java
```diff
@@ -248,6 +248,14 @@ public static List<SpellAbility> getAlternativeCosts(final SpellAbility sa, fina
                     alternatives.add(foretold);
                 }
 
+                if (source.isWarped() && activator.equals(source.getOwner()) && !source.enteredThisTurn() && !source.getManaCost().isNoCost()) {
+                    final SpellAbility normalWarp = sa.copy(activator);
```
tool4ever: you need to build it earlier so real AltCost see it here
otherwise how can friend Henzie blitz it?

Hanmac: I don't know
casting from Warp isn't an Alternative Cost (but Alternative Zone)

so my Gut says, you can Blitz warped Cards from Exile

tool4ever: yes, you should allow it just like for Adventure

Hanmac: I think it depends on if there is something that make a card in exile become "warped"

if not, then i might do a Mayplay Effect similar to the Adventure Part

tool4ever:
> "isWarped" might be better as property ala plotted if there is something that can make it "warped" (then i don't need ExiledSA)?

I think the other fields could be cleaned up with it...?
`exiledSA.isKeyword(Keyword.PLOT) || exiledSA.hasParam("Plotted")`

Hanmac:
> > "isWarped" might be better as property ala plotted if there is something that can make it "warped" (then i don't need ExiledSA)?
> 
> I think the other fields could be cleaned up with it...? `exiledSA.isKeyword(Keyword.PLOT) || exiledSA.hasParam("Plotted")`

Doesn't work, because "BecomesPlotted" is a Trigger

forge-game/src/main/java/forge/game/GameActionUtil.java
tool4ever: why resolve, it doesn't work ;/

Hanmac: I will do Blitz stuff later

tool4ever: Hmn, it's not only for Blitz
All AltCost like _As Foretold_ can't work like this
I'll try a fix imo

Hanmac:
I still don't know if I should like `ExiledSA` or not 🤔

tool4ever:
I don't care too much because all of that probably needs refactor when linked ability support gets a better implementation.

FWIW a dumb corner case with _Skullbriar_ + _Evelyn_ exists because she says: "exiled by an ability you controlled" 💩

Hanmac:
> I don't care too much because all of that probably needs refactor when linked ability support gets a better implementation.

I meant the check that is behind the `isWarped` function.
I don't know if a simple `bool warped = false` would have been better

Hanmac:
Next card I'm going to add:
> choose a creature you control, or a warped creature card you own in exile

this should work with my warped property

=
what I'm scared, if there is a card that says: "exile card X, it becomes warped" :/

PR #7828, opened 2025-06-09
Fix #7814: Counters OOM crashes the game

CTimmerman:
Closes #7814

PR #7920, opened 2025-07-02
Don't freeze on ridiculous number of counters.

CTimmerman:


tool4ever:
Please don't add your own duplicates to the queue like that: #7828

PR #7921, opened 2025-07-03
Add unique suffix for reversible cards

### LINKED ISSUE #8602: Missing keys: t:speed and t:max_speed
`t:speed` and `t:max_speed` show up as default cards in the command zone, but it would be nice if they had the official image and didn't cause output spam. Patch coming up.

Eradev:
* Fix the fetching of reversible versions of cards in the REX, TDM and SLD sets by adding an unique suffix (☇), because a single entry in SLD (1012b M Icingdeath, Frost Tyrant) is not a reversible card, but named using b. Also prevents clashes with sets that have multiples arts using a,b,c,[...] suffixes. Also added it for tokens.
* Fix to some SLD ids, added missing ones.
* Removed the ARN hack as it was no longer present in the editions txt file.
* Fix speed and max_speed images. ( Fixes #8602  )

### DIFF
--- a/forge-core/src/main/java/forge/ImageKeys.java
+++ b/forge-core/src/main/java/forge/ImageKeys.java
@@ -31,6 +31,8 @@ public final class ImageKeys {
     public static final String MONARCH_IMAGE         = "monarch";
     public static final String THE_RING_IMAGE        = "the_ring";
     public static final String RADIATION_IMAGE       = "radiation";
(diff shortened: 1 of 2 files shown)


Hanmac:
> Fixed dungeons images fetch

I need to fix that in a different MR
Because the ImageFetch is the wrong place to get the Dungeon type

paulsnoops (approved):

PR #7922, opened 2025-07-03
Fix MB2 funny card numbers + fetch

Eradev:
* Allow fetching funny cards;
* Fix MB2 collector numbers for the funny cards.


Jetz72:
Not certain about this, but from a quick check of `StaticData`, `CardDb`, `CardEdition` and a couple other places, I think the only place we're actually checking for the "F" prefix is in the audit. Everything else seems to use the sections in the edition file. I wonder if we could change the audit function and then just remove the "F" prefix when loading the edition. Then we wouldn't need to watch for it downstream.

PR #7923, opened 2025-07-03
Encode UTF-8 symbols in collector number

Eradev:
Encode UTF-8 symbols in collector number for maximum compatibility.


Hanmac:
I need to keep that in mind when changing image keys

Probably should use better URL encode methods 🤔 @tool4ever

tool4ever:
well if there are more symbols not handled like why does the cross not need encoding too?

Eradev:
I just realized that it might just be in IntelliJ. I tried with ARN and the cross did not work after I cleared my image cache.

Still, UTF-8 symbols should be encoded for maximum compatibility. I'll change the PR to reflect that.

tehdiplomat:
This broke Android. Reverting

PR #7932, opened 2025-07-04
Adding extra Planes to Adventure mode and providing an easy method to swap between them.

shenshinoman:
There are a ton of files here, as it adds a mostly completed "Old School Shandalar" which is the base game, but limited to pre-scourge sets. Aswell as adding in an Innistrad-only plane that is in an alpha state with unique enemies, quests, etc. And finally also brings in a third plane, Crystal Kingdoms, which is Final Fantasy only, but in a very pre-alpha state.

This also modifies a couple base game files in order to allow the plane switching menu in the Settings scene. As well as adding new Jumpstarts for Innistrad, and the code to properly limit inn events based on allowedEditions and restrictedEvents in the config json.


forge-gui/res/adventure/Shandalar Old School/config.json
```diff
@@ -0,0 +1,747 @@
...
+    "ZNC",
+    "ZNE",
+    "ZNR"
+  ],
+  "difficulties": [
+    {
+      "name": "Easy",
+      "startingLife": 16,
+      "startingShards": 5,
+      "staringMoney": 500,
```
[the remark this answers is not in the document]
shenshinoman: this typo exists in the base files. changing this in this file would simply break this file.

forge-gui/res/adventure/Shandalar Old School/config.json
```diff
@@ -0,0 +1,747 @@
...
+      },
+      "startItems": [
+        "Manasight Amulet",
+        "Leather Boots"
+      ]
+    },{
+      "name": "Normal",
+      "startingLife": 12,
+      "startingShards": 2,
+      "staringMoney": 250,
```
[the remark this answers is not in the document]
shenshinoman: this typo exists in the base files. changing this in this file would simply break this file.

forge-gui-mobile/pom.xml
```diff
@@ -21,6 +21,15 @@
                     <target>17</target>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>22</source>
+                    <target>22</target>
+                    <compilerArgs>--enable-preview</compilerArgs>
```
tehdiplomat: Whats this plugin used for?

tehdiplomat:
Hey I haven't lost track of this, just lots of different balls in the air. Gonna try to pull this fork down in the next few days and give it a spin. Didn't see anything too notable in the code changes. 

How do you feel about all of the config files? I've been thinking for a while that they might be too nested, definitely not something i want you to change here, but now that we're opening this gate, I want to make sure that we keep other planes in mind for potentially splitting out these files for easier maintainability/modification.

shenshinoman:
No problem. I understand completely and wasn't too worried. 

As for the config files, they've worked out fine for me so far. However I wouldn't be against greater granularity along them. Either way, I also plan to update the wiki with the process for building planes once everything is tested and confirmed its working as intended and then gets joined to live. To try and help any other interested individuals out there.

forge-gui/res/adventure/Innistrad/printsheets.txt
```diff
@@ -0,0 +1,763 @@
+
+#----------------------SHEN CUSTOM JUMPSTART PACKS START HERE---------------------------------#
+
+[ISD Dearly Departed 1]
```
tehdiplomat, line 4: Is this duplicated? Maybe we should try to load printsheets from its default location as well as res/adventure/*/printsheets.txt? As well?

forge-gui-mobile/pom.xml
shenshinoman: Sorry, that plug-in was in accident and should be removed. (it serves no purpose for what is in this and doesn't effect any of it. i thought i removed that before the commit XD)

I initially added it, to see if some of the JDK 22 preview functionality from it would allow a smooth appending of the plane specific sheets to the main sheets... But it didn't work out.

forge-gui/res/adventure/Innistrad/printsheets.txt
shenshinoman: you'll find similar in boosters-special.txt, and the data in both is appended to the equivalent files in the blockdata folder.

I did, in fact, try to find a way for the system to initialize them separately, working with Friorsol and Jetz via the discord. Ultimately, we all came to the realization, as quoted from Jetz directly. "The problem with world-local print sheets and booster templates is that to generate a booster, we generally pass the booster template to new UnOpenedProduct, which uses BoosterGenerator.makeSheet to get the needed print sheets. That ultimately uses StaticData's print sheets, and doesn't leave much room for us to supply our own unless we reconstruct the whole pipeline."

In short, those sheets get initialized in the StaticData.java long before anything in Adventure initializes. So slipping it into the main blockdata sheets "boosters-special.txt" and "printsheets.txt" was the current work-around, and those two files are sitting in the adventure Innistrad folder as a self-reminder to, eventually, build the proper pipeline through the entire game to actually be able to use them by the plane, and thus remove the current dirty implementation of their data being appended to the core games sheets.

tehdiplomat:
Looks like something else caused a merge conflict. I'll pull this down tonight and check it out

tehdiplomat (approved):
Let

PR #7938, opened 2025-07-04
Fix #7937: offspring token key malformed

### LINKED ISSUE #7937: Offspring token image key is malformed
**Describe the bug**
Offspring token image key is malformed.

**To Reproduce**
1. Play Coruscation Mage with Offspring.
2. See "Offspring token image key is malformed" in program output.

**Expected behavior**
Use a proper key.

Easy fix coming up.

CTimmerman:
Closes #7937


Hanmac (approved):

Hanmac:
i probably need to get the collectorNumber from the Token, but i will change that later

PR #7940, opened 2025-07-04
Fix useless game log entries (#7939)

### LINKED ISSUE #7939: Useless entries in GameLog
**Describe the bug**
Game log is full of useless entries like "A controlled by A" and "A didn't attack" instead of damage dealt.

**To Reproduce**
1. Try to tell what's going on by only looking at the game log.
2. See mostly useless data described above.

**Expected behavior**
Be more like Cockatrice and show damage before noops.

**Additional context**
Patch coming up.

CTimmerman:
Closes #7939

### DIFF
--- a/forge-game/src/main/java/forge/game/GameLogFormatter.java
+++ b/forge-game/src/main/java/forge/game/GameLogFormatter.java
@@ -159,6 +159,7 @@ public class GameLogFormatter extends IGameEventVisitor.Base<GameLogEntry> {
         if (newLobbyPlayer == null) {
             message = localizer.getMessage("lblLogPlayerHasRestoredControlThemself", p.getName());
         } else {
+            if (p.getName() == newLobbyPlayer.getName()) return null;
             message = localizer.getMessage("lblLogPlayerControlledTargetPlayer", p.getName(), newLobbyPlayer.getName());
         }
         return new GameLogEntry(GameLogEntryType.PLAYER_CONTROL, message);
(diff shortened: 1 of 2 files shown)


forge-game/src/main/java/forge/game/GameLogFormatter.java
```diff
@@ -247,9 +248,7 @@ public GameLogEntry visit(final GameEventAttackersDeclared ev) {
             if (sb.length() > 0) sb.append("\n");
             sb.append(localizer.getMessage("lblLogPlayerAssignedAttackerToAttackTarget", ev.player, Lang.joinHomogenous(attackers), k));
         }
-        if (sb.length() == 0) {
-            sb.append(localizer.getMessage("lblPlayerDidntAttackThisTurn").replace("%s", ev.player.toString()));
```
tehdiplomat: Are you suggesting that someone not declaring attackers when they could have is not a useful log?

CTimmerman: Yes. Not playing any spells in a turn is also not noteworthy.

CTimmerman: It's also a lie in this case:
image

CTimmerman:
Still relevant, i'm afraid.

tehdiplomat (approved):

forge-game/src/main/java/forge/game/GameLogFormatter.java
tehdiplomat: I don't really agree, and if people complain about missing it, we'll put it back.

tehdiplomat:
BTW this PR created issues for other users, myself included where logs just seemingly stopped showing up.

Jetz72:
Presumably because the default log level (and by extension, the one that most users currently use) is set to "Damage" - i.e. "only show damage events and things that are more significant". And this PR moved damage events above the events for adding things to the stack and resolving them, plus a few other types. 

Could revert the entry type ordering, though I'm not sure the original order makes much sense either - I'd think that adding things to the stack and resolving them would be one of the highest priority things to keep record of when logging game events.

tehdiplomat:
Yea i think this is more about being aware of the types of changes we're making and how its going to affect otehr people.

CTimmerman:
I don't need to know that my **Torbran, Thane of Red Fell** etc trigger every time I deal damage, but would like to know why i died, which this patch fixes. See the issue.

Also, a recent patch appears to have undone the final log order so exports are to be read bottom up again.

PR #7948, opened 2025-07-05
Add missing radiation token

Eradev:
* Add missing radiation token
* Add missing [other] entries
* Minor code fix (deprecated + superfluous code)


Hanmac:
i would revert the DeckFormat Changes and do in a separate MR

Hanmac:
the Changes on Player will probably break the build

Hanmac (approved):

PR #7952, opened 2025-07-06
Huge Alchemy Cleanup

Eradev:
* Removed Alchemy cards that were removed from the game, and from Scryfall.
* Removed the need to hyphenate via code by changing their Collector No according to Scryfall.
* Added missing Alchemy entries into sets.


paulsnoops:
I get why you'd want to remove those Alchemy cards that no longer exist but this will break the Archived formats that used those rebalances

paulsnoops:
The Alchemy cards from MH2 were intentionally put into J21 so they can be used in Archived Arena formats (and show up correctly) where MH2 isn't present. I don't know if it's better to change those two lines in the MH2 edition to comments with a new comment explaining this or remove them from J21 and I can add them as 'Additional' in the relevant format files.

Eradev:
> I get why you'd want to remove those Alchemy cards that no longer exist but this will break the Archived formats that used those rebalances

This is tricky because it wouldn't keeping them in the set lists also unbalance those same formats after they were effectively removed? Seriously, Alchemy cards are such a headache. Maybe keep their scripting files and try to find a way to add those extra cards into these format specific settings?

New users will also be unable to get pictures of them anymore via Scryfall.

paulsnoops:
> This is tricky because it wouldn't keeping them in the set lists also unbalance those same formats after they were effectively removed? Seriously, Alchemy cards are such a headache. Maybe keep their scripting files and try to find a way to add those extra cards into these format specific settings?
> 
> New users will also be unable to get pictures of them anymore via Scryfall.

For me, the ideal situation is to have "versions" of the same card inside the original script. So instead of having any Alchemy scripts at all we can just have for example dated versions inside the original script which changes only the lines required. That way if the version is dated we could try and look at the date inside the format file and Forge will just use the correct version of the script (which will also help with different rebalances of the same card showing up in the correct archived format by date). But that's a way off right now so I'd rather not delete anything from scripts or editions until we can figure out another way of doing it.

Edit: added some words

paulsnoops:
In regards to users being unable to fetch images, Sol and I have spoken before about hosting a small selection of images that Scryfall doesn't have to help with this, but only for those kind of cards that have since been removed from Scryfall rather than delete them from Forge because we'd rather keep historical changes.

tool4ever:
> > This is tricky because it wouldn't keeping them in the set lists also unbalance those same formats after they were effectively removed? Seriously, Alchemy cards are such a headache. Maybe keep their scripting files and try to find a way to add those extra cards into these format specific settings?
> > New users will also be unable to get pictures of them anymore via Scryfall.
> 
> For me, the ideal situation is to have "versions" of the same card inside the original script. So instead of having any Alchemy scripts at all we can just have for example dated versions inside the original script which changes only the lines required. That way if the version is dated we could try and look at the date inside the format file and Forge will just use the correct version of the script (which will also help with different rebalances of the same card showing up in the correct archived format by date). But that's a way off right now so I'd rather not delete anything from scripts or editions until we can figure out another way of doing it.
> 
> Edit: added some words

I think there was already some discussion a while ago that functional variants, e.g. _Garbage Elemental_ could help solve this case

Hanmac:
> Archived formats that used those rebalances

That doesn't work either if Wizards does rebalance the cards multiple times

Eradev:
> > Archived formats that used those rebalances
> 
> That doesn't work either if Wizards does rebalance the cards multiple times

Indeed. I think only Omnath had that treatment (so far), but Forge only has the 2nd reworked version.

Eradev:
Also take into account the errata'd cards as well. It's near impossible to account for all the changes to perfectly emulate the archived formats over the years.

Hanmac (approved):

Hanmac:
@paulsnoops for "archived formats" we need to think about something else

paulsnoops:
I understand the conversation about different versions of rebalances/errata etc. but I don't agree with removing cards that once existed. Forge has always kept a historical archive of Magic and removing these will also break any decks/quests that contain them.

Hanmac:
> historical archive

The problem is that we can't keep every different version, otherwise we would need to every different oracle version too

what we could do, but that is for a different MR, is to make a fallback, that if it tries to load a Rebalanced Card that doesn't exist, it does load the normal version instead

Jetz72:
> I think there was already some discussion a while ago that functional variants, e.g. _Garbage Elemental_ could help solve this case

Yes, we could remove the alchemy scripts and merge their contents into their paper counterparts, by prefixing the modified lines with `Variant:Alchemy:`. Then we could just modify forge to automatically default to that variant if it's found while playing in an Alchemy format.

Perhaps scripts for retired alchemy cards could be supported by using a particular naming scheme, e.g. `Variant:Alchemy20221006:...` would mean to use that variant in archived alchemy formats dated prior to October of 2022.

paulsnoops:
Sounds like good solutions all around :)

Jetz72:
The point about breaking existing decks, quests, and adventure saves is also worth considering. We should probably wait on this until we've settled on a course of action and have logic in place for migrating older card lists. Otherwise if someone loads a save before we have a migration solution in place, that data would just be lost.

Eradev:
I've added the fallback to get the non-Alchemy version of a missing card.

I believe what @Jetz72 had proposed will require another PR and much more work..

paulsnoops:
For now, seeing as though we have #8210 if we just leave those rebalanced cards out of the editions but keep the card scripts they should fall into 'unknown' edition. That way we still have the scripts until we can go the variants route from @Jetz72's suggestion? Then we can merge this?

Eradev:
I have restored the files and see that they're added correctly.

image

Hanmac (approved):

PR #7954, opened 2025-07-06
Non-alchemy sets cleanup

Eradev:
* Added missing cards.
* Changed Master Edition from MED to ME1 to reflect Scryfall's ID.
* Changed Masterpiece Series - Amonkhet from MPS_AKH to MP2 to reflect Scryfall's ID, as well as the set's name.
* Replaced 'unsupported' cards by their non-promo equivalent so they don't popup at every launch.

### DIFF
--- a/forge-gui/res/editions/Year of the Rabbit 2023.txt
+++ b/forge-gui/res/editions/Year of the Rabbit 2023.txt
@@ -4,6 +4,7 @@ Date=2023-02-10
 Name=Year of the Rabbit 2023
 Type=Promo
 ScryfallCode=PL23
+TokensCode=PL23
 
 [cards]
 1 R Rabbit Battery @TSWCK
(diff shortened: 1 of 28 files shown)


forge-gui/res/editions/Year of the Rabbit 2023.txt
```diff
@@ -4,6 +4,7 @@ Date=2023-02-10
 Name=Year of the Rabbit 2023
 Type=Promo
 ScryfallCode=PL23
+TokensCode=PL23
```
tool4ever, line 7: not needed, should just fallback

Eradev: From my tests it defaults to "T[Edition Code]" and never tries "[EditionCode]".

Hanmac (approved):

PR #7955, opened 2025-07-06
Fix typo

Eradev:


Hanmac:
@kevlahnota can this change break existing Adventure?

kevlahnota:
> @kevlahnota can this change break existing Adventure?

I think so since the serialized data includes the typo variable. There should be a migration of some sort to make the old file work and update to newer version when the old file is loaded and hooked to the new variable. To be safe maybe just leave the typo alone until the user really needs to upgrade the old save file but we don't know if we come to that point.

Eradev:
Isn't that value only used at the start of the game anyway?

kevlahnota:
> Isn't that value only used at the start of the game anyway?

Did you try new game plus?

Eradev:
Yes, it isn't even used in NewGame+. I'm not sure why we're storing it in the save file?

EDIT: Still added support for old variable in case it is needed in the future.

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
```diff
@@ -300,10 +300,15 @@ public void setColorIdentity(ColorSet set) {
 
     @Override
     public void load(SaveFileData data) {
-        clear(); //Reset player data.
+        clear(); // Reset player data.
         this.statistic.load(data.readSubData("statistic"));
         this.difficultyData.startingLife = data.readInt("startingLife");
-        this.difficultyData.startingMoney = data.readInt("startingMoney");
+        // Support for old typo
+        if (data.containsKey("staringMoney")) {
```
kevlahnota, line 307: does the old save work without this check? or just use a starting money from the config file? have you seen the updateDifficulty method used by SaveLoadScene.java for NewGamePlus? if it works without migration then I think it's good to remove this I guess.

Eradev: Yes the old save work without using this. updateDifficulty does assign the value to the variable but it isn't used anywhere. NewGame+ doesn't grant new money.

PR #7961, opened 2025-07-07
Prefer random token from set

Eradev:


forge-core/src/main/java/forge/token/TokenDb.java
```diff
@@ -101,12 +102,23 @@ public PaperToken getToken(String tokenName) {
...
+    @Override
+    public PaperToken getToken(String tokenName, String edition, int artIndex) {
         CardEdition realEdition = editions.getEditionByCodeOrThrow(edition);
         String fullName = String.format("%s_%s", tokenName, realEdition.getCode().toLowerCase());
 
-        // token exist in Set, return one at random
+        // Token exists in edition, return token at artIndex or a random one.
         if (loadTokenFromSet(realEdition, tokenName)) {
-            return Aggregates.random(allTokenByName.get(fullName));
+            List<PaperToken> list = new ArrayList<>(allTokenByName.get(fullName));
```
tool4ever: use something like `Iterables.getLast` to avoid extra array cost in busy functions

Hanmac:
i don't really see the reason behind this

And `artIndex` is going to be replaced by #7846 anyway soon?

Eradev:
I'm not sure what you're working on, but I just wanted to use the set's tokens when possible so it isn't always the default one. One of Magic's strength IMO is the wide different arts and being locked into one when there would be more flavorful ones is depressing.

tool4ever (approved):
LGTM

Hanmac (approved):
i don't really like the changes with the art index
but i'm going to remove them later anyway

PR #7968, opened 2025-07-08
Fix Crash for Rug of Smothering

kevlahnota:
TriggeredActivator not defined on AbilityUtils calculateAmount, hence produces ClassCastException (found on Sentry)

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -658,7 +658,7 @@ public class AbilityUtils {
                 val = playerXCount(getDefinedPlayers(card, key, sa), calcX[1], card, ability);
             }
             else if (calcX[0].startsWith("TriggeredPlayer") || calcX[0].startsWith("TriggeredTarget")
-                    || calcX[0].startsWith("TriggeredDefendingPlayer")) {
+                    || calcX[0].startsWith("TriggeredDefendingPlayer") || calcX[0].startsWith("TriggeredActivator")) {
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -733,7 +733,14 @@ else if (calcX[0].startsWith("TriggerObjects")) {
                 else if (calcX[0].startsWith("Triggered") || (calcX[0].startsWith("CardTriggered"))) {
                     final SpellAbility root = sa.getRootAbility();
                     final int s = calcX[0].startsWith("Triggered") ? 9 : 13;
-                    list = new CardCollection((Card) root.getTriggeringObject(AbilityKey.fromString(calcX[0].substring(s))));
+                    final Object triggeringObject = root.getTriggeringObject(AbilityKey.fromString(calcX[0].substring(s)));
```
tool4ever: thanks but not needed, just add the param here:
https://github.com/Card-Forge/forge/blob/6986b5a53d167f3f8e41aab610df6d9ac8cb0fe9/forge-game/src/main/java/forge/game/ability/AbilityUtils.java#L660

Hanmac: yeah, i was just searching for that part too, to pass through `playerXProperty`

Hanmac (approved):

Hanmac:
@kevlahnota next time, use "Squash Merge", so all the changes gets compressed into one Commit

That's especially useful if the MR has some changes that get reverted by later commit

PR #7972, opened 2025-07-09
Migration to PLST

Eradev:
* Fixed PLST Coll.No. and updated the list, fixing the missing pictures;
* Removed FMB1 (Merged into PLST);
* Modified MB1 to use PLST cards, keeping the set for drafts as discussed;
* Will attempt to transfer the cards from the deprecated sets into PLST on load;
* Fixed the audit to only account for Coll.No. `^F\d+` as Funny cards;


kevlahnota:
@Hanmac @paulsnoops @Agetian I think this is good, any opinion about this PR?

PR #7973, opened 2025-07-09
Fix Meld cards in FIN

Eradev:
* FIN doesn't use a prefix for the front of the meld cards.
* Fix FIN edition artists


Hanmac:
@Eradev I reported the inconsistency to Scryfall (including some Error with the CN in SIR)

> 
> From Innistrad Remastered (INR):
> Hanweir Garrison is 157a
> Midnight Scavengers is 123 instead of 123a?
> Bruna, the Fading Light is 14a
> 
> Shadows over Innistrad Remastered (SIR)
> For this pair, I think the Melded Card has the wrong Collector Number?
> I added new Gatherer Links: 
> Graf Rats 115a should be 123? https://gatherer.wizards.com/SIR/en-us/123/graf-rats
> Chittering Host 115b is correct https://gatherer.wizards.com/SIR/en-us/115/chittering-host
> Midnight Scavengers 123 should be 115a? https://gatherer.wizards.com/SIR/en-us/115/midnight-scavengers
> 
> The Brothers' War (BRO) 
> The Mightstone and Weakstone is 238 instead of 238a?

Also, i didn't report this, but these one does mess with other "a" and "b" parts too:
> The Brothers' War Promos (PBRO) This one has "s" and "p" marks after "a" and "b"
> Phyrexian Dragon Engine 163as
> Mishra, Lost to Phyrexia 163bs

PR #7976, opened 2025-07-09
Update PRM, TD0, TD1, EOE, EOC

Eradev:


forge-gui/res/editions/Edge of Eternities.txt
```diff
@@ -6,30 +6,155 @@ Type=Expansion
...
+# 384 M Exalted Sunborn @Makoron
+# 385 R Starfield Vocalist @nina
 386 M Sothera, the Supervoid @Mateusz Urbanowicz
+389 M Mutinous Massacre @Aogachou
+391 M The Endstone @Hidetaka Tenjin
+398 R Singularity Rupture @Néstor Ossandón Leal
+399 R Emissary Escort @Lius Lasahido
+
+[tokens]
+# 1 c_1_1_sliver @Igor Grechanyi
```
Hanmac: we know that the Nummer is 1, so no need for # ?

Eradev: True. I t wasn't on Scryfall yet so I had commented it, but it'll just fail the fetch for now.

forge-gui/res/editions/Edge of Eternities.txt
```diff
@@ -6,30 +6,155 @@ Type=Expansion
...
 [cards]
+# 1 R Anticausal Vestige @Chase Stone
 2 M Tezzeret, Cruel Captain @Chris Rahn
+# 4 R Astelli Reclaimer @Carly Milligan
+6 C Banishing Light @Rovina Cai
+14 U Emergency Eject @Leon Tukker
+# 15 M Exalted Sunborn @Scott M. Fischer
+# 19 R Haliya, Guided by Light @Kieran Yanner
 35 R The Seriema @Sergey Glushakov
+# 37 U Starfield Shepherd @Marta Nael
```
tehdiplomat, line 17: You shouldn't need to comment out spoiled cards that aren't scripted yet.

Eradev: I'll know for next time. :)

PR #7989, opened 2025-07-10
Attempt at Eumidian Wastewaker -- could it work like this?

Fulgur14:


.github/eumidian_wastewaker.txt
```diff
@@ -0,0 +1,15 @@
+Name:Eumidian Wastewaker
+ManaCost:2 B B
+Types:Creature Insect Cleric
+PT:4/4
+T:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigChooseCard | TriggerZones$ Battlefield | TriggerDescription$ Whenever this creature attacks, you and defending player each discard a card or sacrifice a permanent. You draw a card for each land card put into a graveyard this way.
+SVar:DBChoose:DB$ ChooseCard | Defined$ You | Mandatory$ True | Choices$ Card.YouOwn+inZoneHand,Permanent.CanBeSacrificedBy You | ChoiceTitle$ Choose a card or permanent | AILogic$ WorstCard | SubAbility$ DBChooseOpp
```
tool4ever: it's an interesting idea
can probably use ControlledByPlayer$ to get rid of the second ChooseCard

Fulgur14: Maybe like this? Is there generic "CanBeSacrificed"? There are few unsacrificeable things now and they must be excluded.

.github/eumidian_wastewaker.txt
```diff
@@ -0,0 +1,15 @@
+Name:Eumidian Wastewaker
+ManaCost:2 B B
+Types:Creature Insect Cleric
+PT:4/4
+T:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigChooseCard | TriggerZones$ Battlefield | TriggerDescription$ Whenever this creature attacks, you and defending player each discard a card or sacrifice a permanent. You draw a card for each land card put into a graveyard this way.
+SVar:DBChoose:DB$ ChooseCard | Defined$ You | Mandatory$ True | Choices$ Card.YouOwn+inZoneHand,Permanent.CanBeSacrificedBy You | ChoiceTitle$ Choose a card or permanent | AILogic$ WorstCard | SubAbility$ DBChooseOpp
+SVar:DBChooseOpp:DB$ ChooseCard | Defined$ TriggeredDefendingPlayer | Mandatory$ True | Choices$ Card.inZoneHand+OwnedBy TriggeredDefendingPlayer,Permanent.CanBeSacrificedBy TriggeredDefendingPlayer | ChoiceTitle$ Choose a card or permanent | AILogic$ WorstCard | SubAbility$ DBDiscard
```
tool4ever: CanBeSacrificedBy doesn't have that additional value to be scripted
strictly, CanBeDiscardedBy would also need to exist - but its only printed once in combination with CantSacrifice so can't cheat out of it anyway? 🤔

.github/eumidian_wastewaker.txt
tool4ever, line 1: fix path

.github/eumidian_wastewaker.txt
Fulgur14: Actually, I could do this -- it accepts permanents that can be sacrificed by either player, but the Chooser will limit it to their own permanents anyway.

.github/eumidian_wastewaker.txt
tool4ever: fixed the general errors, please test

Fulgur14: This version doesn't seem to allow discarding,

tool4ever (approved):

PR #7992, opened 2025-07-10
J21 Update

Eradev:
Sync data with Scryfall

### DIFF
--- a/forge-gui/res/editions/Jumpstart Historic Horizons.txt
+++ b/forge-gui/res/editions/Jumpstart Historic Horizons.txt
@@ -37,751 +37,348 @@ ScryfallCode=J21
 29 C Skyshroud Ambush @Andreas Zafiratos
 30 U Skyshroud Lookout @Olena Richards
 31 C Veteran Charger @Leanna Crossan
-32 U Abiding Grace
-33 U Abzan Battle Priest
-34 U Abzan Falconer
-35 U Aethershield Artificer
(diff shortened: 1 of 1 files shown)


forge-gui/res/editions/Jumpstart Historic Horizons.txt
```diff
@@ -37,755 +37,380 @@ ScryfallCode=J21
...
+762 R Zabaz, the Glimmerwasp @Jason Felix
+763 C Ichor Wellspring @Steven Belledin
+770 C Khalni Garden @Ryan Pancoast
+776 R Sliver Hive @Igor Kieryluk
 
 [rebalanced]
 A438 U A-Dragon's Rage Channeler @Martina Fackova
 A529 C A-Unholy Heat @Kari Christensen
 
 [tokens]
```
Hanmac: J21 doesn't have token images, so these shouldn't be defined there?

Eradev: Indeed I shouldn't. I'll fix it.

tehdiplomat:
Typically we've been averse to just widespread changes like this because of how Forge might interact when referencing defined cards are removed. Have you tested saves in Quest/Adventure/etc that have these cards that now no longer exist after this change is made? What happens? What do we want to happen?

Eradev:
It's already handled. It replaces the card with another version of it if it exists.

image

image

PR #7994, opened 2025-07-10
runeblade_raiser.txt and support

Northmoc:


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -7272,6 +7272,18 @@ public void updateTriggers(List<Trigger> list, CardState state) {
                 list.clear();
             }
             list.addAll(ck.getTriggers());
+            if (!ck.getRemovedTriggers().isEmpty()) {
+                List<Trigger> toRemove = Lists.newArrayList();
+                for (final Trigger t : ck.getRemovedTriggers()) {
+                    for (final Trigger l : list) {
+                        if (t.getMode() == l.getMode() && t.toString().equals(l.toString())) {
```
Hanmac, line 7279: i don't think `t.toString().equals(l.toString())` is good enough for comparison :/

Northmoc: I'm not crazy about it either... maybe "good enough for this one Alchemy card" though?

I'll poke around a bit and see if it can be improved without major surgery to the engine

Let me know if you have an idea how to make it better

forge-gui/res/cardsfolder/r/runeblade_raiser.txt
```diff
@@ -0,0 +1,10 @@
+Name:Runeblade Raiser
+ManaCost:B
+Types:Creature Human Warrior
+PT:2/1
+R:Event$ Moved | ValidCard$ Card.Self | Destination$ Battlefield | ReplacementResult$ Updated | ReplaceWith$ ETBTapped | Description$ This creature enters tapped.
+SVar:ETBTapped:DB$ Tap | Defined$ Self | ETB$ True
+T:Mode$ ChangesZone | Origin$ Battlefield | Destination$ Graveyard | ValidCard$ Card.Self | Execute$ TrigReturn | TriggerDescription$ When this creature dies, return it to the battlefield under its owner's control. It perpetually loses this ability.
+SVar:TrigReturn:DB$ ChangeZone | Defined$ TriggeredNewCardLKICopy | Origin$ Graveyard | Destination$ Battlefield | RememberChanged$ True | SubAbility$ DBAnimate
+SVar:DBAnimate:DB$ Animate | Defined$ Remembered | RemoveThisTrigger$ True | Duration$ Perpetual
```
tool4ever, line 9: cleanup missing

forge-game/src/main/java/forge/game/card/Card.java
tool4ever: probably at least switch to `getOriginalParam("TriggerDescription")`
though multiple copies could still quickly make that fall apart...

but if we obviously can't use the normal ID maybe we could just store the intrinsic index of the trigger and use that for removal on the new object? 🤔

jamincollins:
## Findings

#### 1. RemoveThisTrigger parameter not propagated to AnimateEffectBase copy
Description:
The AnimateEffectBase adds support for RemoveThisTrigger parameter by collecting triggers to remove, but the check `sa.hasParam("RemoveThisTrigger")` only checks the original ability, not considering that the ability may be copied via AbilityFactory. If the SpellAbility is cloned with different params, the trigger reference may be stale.
Test case:
```java
// code that proves/illustrates the finding
SpellAbility sa = AbilityFactory.getAbility(card, "T:Mode$ ChangesZone | Origin$ Battlefield | Destination$ Graveyard | ValidCard$ Card.Self | Execute$ TrigReturn | TriggerDescription$ When this creature dies, return it to the battlefield under its owner's control.", null);
sa.setParam("RemoveThisTrigger", "True");
// sa.getTrigger() may return null if trigger not yet parsed
if (sa.hasParam("RemoveThisTrigger") && sa.getTrigger() == null) {
    // removedTriggers will be empty, trigger not removed
}
```
Impact: The RemoveThisTrigger may silently fail if trigger is not available at animation time, leading to trigger not being removed per MTG rules.

#### 2. Trigger removal uses toString equality for matching
Description:
In Card.updateTriggers, removed triggers are matched via `t.getMode() == l.getMode() && t.toString().equals(l.toString())`. This is fragile: toString may change with internal state, and mode equality is reference equality not logical equality.
Test case:
```java
Trigger t1 = TriggerHandler.parseTrigger("...", card, false, sa);
Trigger t2 = TriggerHandler.parseTrigger("...", card, false, sa);
t1.setId(1);
t2.setId(2);
// t1.toString() == t2.toString() but mode objects differ? 
// getMode() returns same enum instance typically, but toString equality may be insufficient.
```
Impact: Trigger removal may fail to match, leaving unwanted triggers active, violating perpetual loss rule.

#### 1. RemovedTriggers list initialized but never cleared on copy
Description:
CardTraitChanges.copy creates new list for removedTriggers but does not handle null. If removedTriggers is null, copy will NPE.
Test case:
```java
CardTraitChanges ctc = new CardTraitChanges(null, null, null, null, null, null, false, false);
CardTraitChanges copy = ctc.copy(card, true);
// copy.removedTriggers initialized via Lists.newArrayList() even if original null, OK
// but constructor accepts null removedTriggers, stored as null? 
// In constructor, if removedTriggers != null, addAll. So field stays null. copy will NPE on for loop.
```
Impact: NullPointerException on copy when removedTriggers not supplied.

#### 2. AnimateEffectBase does not guard against null sa.getTrigger()
Description:
removedTriggers.add(sa.getTrigger()) without null check. If RemoveThisTrigger param present but getTrigger() returns null, null added to list, later iteration NPE.
Test case:
```java
if (sa.hasParam("RemoveThisTrigger")) {
    removedTriggers.add(sa.getTrigger()); // may add null
}
```
Impact: NullPointerException during trait application.

#### 1. Card.addChangedCardTraits signature change breaks callers
Description:
The method signature changed to include removedTriggers, but several callers were updated with null. However, `addChangedCardTraitsByText` still creates CardTraitChanges with null removedTriggers, which is fine, but `getChangedCardTraitsList` creates CardTraitChanges with 6 args (landManaAbilities, null, null, null, null,null). The constructor expects 8 args: spells, removedAbilities, trigger, removedTriggers, replacements, statics, removeAll, removeNonMana. The call passes null for removedTriggers correctly, but older code may compile against old signature.
Test case:
```java
// Existing code compiled with old signature may still reference old method with 5 args
Card card = ...
card.addChangedCardTraits(spells, removedAbilities, trigger, replacements, statics, removeAll, removeNonMana, ts, id);
// This now resolves to new signature with removedTriggers inserted, causing arg shift.
```
Impact: Potential compile/runtime mismatch, breaking backwards compatibility.

#### 2. Player.addChangedCardTraits call passes null for removedTriggers but signature expects 9 args
Description:
Player.java line 1012 now passes null for removedTriggers, but the call originally had 8 args. The change is correct, but if any other caller missed update, it will fail at runtime.
Impact: Silent failure or mis-ordered arguments.

#### 1. RemoveThisTrigger interacts with perpetual animation
Description:
Runeblade Raiser uses RemoveThisTrigger$ True on DBAnimate subability, which is perpetual. The trigger removal occurs only during animate effect setup, but perpetual animation may reapply traits multiple times, possibly removing the trigger multiple times or failing to remove after first application.
Test case:
```java
// First animate: removedTriggers contains trigger reference
// Unanimate: trait removed, trigger returns?
// Perpetual animation may cause duplicate removal entries.
```
Impact: Trigger may be removed prematurely or not at all, breaking "perpetually loses this ability" rule.

#### 2. Trigger removal based on toString may collide with distinct triggers
Description:
Different triggers with same toString but different internal state will be incorrectly removed.
Impact: Unintended trigger removal affecting other cards.

### Performance and security
#### 1. O(n²) trigger removal loop
Description:
Card.updateTriggers iterates over removedTriggers and for each iterates over entire list to find match, resulting in O(n²) complexity.
Test case:
```java
for (final Trigger t : ck.getRemovedTriggers()) {
    for (final Trigger l : list) {
        if (t.getMode() == l.getMode() && t.toString().equals(l.toString())) { ... }
    }
}
```
Impact: Performance degradation with many triggers.

#### 2. No validation of RemoveThisTrigger param value
Description:
AnimateEffectBase checks hasParam("RemoveThisTrigger") but does not validate value is "True". Param may exist with false value.
Impact: Incorrect trigger removal.

## Verdict
Request Changes
The PR introduces trigger removal support for Runeblade Raiser but has several correctness and safety issues: fragile trigger matching via toString, potential NPEs from null triggers, O(n²) removal, and missing validation of param values. The changes need hardening before merge.

## Recommended Fixes
1. Validate RemoveThisTrigger param is "True" and sa.getTrigger() is non-null before adding to removedTriggers.
2. Use stable identifier for trigger matching instead of toString equality (e.g., compare Trigger IDs or canonical string).
3. Guard CardTraitChanges.copy against null removedTriggers and initialize safely.
4. Add unit tests for perpetual trigger removal and edge cases with null triggers.
5. Refactor trigger removal to use Set for O(n) matching.
---

jamincollins:
## Findings

#### 1. RemoveThisTrigger parameter not propagated to AnimateEffectBase copy
Description:
The AnimateEffectBase adds support for RemoveThisTrigger parameter by collecting triggers to remove, but the check `sa.hasParam("RemoveThisTrigger")` only checks the original ability, not considering that the ability may be copied via AbilityFactory. If the SpellAbility is cloned with different params, the trigger reference may be stale.
Test case:
```java
// code that proves/illustrates the finding
SpellAbility sa = AbilityFactory.getAbility(card, "T:Mode$ ChangesZone | Origin$ Battlefield | Destination$ Graveyard | ValidCard$ Card.Self | Execute$ TrigReturn | TriggerDescription$ When this creature dies, return it to the battlefield under its owner's control.", null);
sa.setParam("RemoveThisTrigger", "True");
// sa.getTrigger() may return null if trigger not yet parsed
if (sa.hasParam("RemoveThisTrigger") && sa.getTrigger() == null) {
    // removedTriggers will be empty, trigger not removed
}
```
Impact: The RemoveThisTrigger may silently fail if trigger is not available at animation time, leading to trigger not being removed per MTG rules.

#### 2. Trigger removal uses toString equality for matching
Description:
In Card.updateTriggers, removed triggers are matched via `t.getMode() == l.getMode() && t.toString().equals(l.toString())`. This is fragile: toString may change with internal state, and mode equality is reference equality not logical equality.
Test case:
```java
Trigger t1 = TriggerHandler.parseTrigger("...", card, false, sa);
Trigger t2 = TriggerHandler.parseTrigger("...", card, false, sa);
t1.setId(1);
t2.setId(2);
// t1.toString() == t2.toString() but mode objects differ? 
// getMode() returns same enum instance typically, but toString equality may be insufficient.
```
Impact: Trigger removal may fail to match, leaving unwanted triggers active, violating perpetual loss rule.

#### 1. RemovedTriggers list initialized but never cleared on copy
Description:
CardTraitChanges.copy creates new list for removedTriggers but does not handle null. If removedTriggers is null, copy will NPE.
Test case:
```java
CardTraitChanges ctc = new CardTraitChanges(null, null, null, null, null, null, false, false);
CardTraitChanges copy = ctc.copy(card, true);
// copy.removedTriggers initialized via Lists.newArrayList() even if original null, OK
// but constructor accepts null removedTriggers, stored as null? 
// In constructor, if removedTriggers != null, addAll. So field stays null. copy will NPE on for loop.
```
Impact: NullPointerException on copy when removedTriggers not supplied.

#### 2. AnimateEffectBase does not guard against null sa.getTrigger()
Description:
removedTriggers.add(sa.getTrigger()) without null check. If RemoveThisTrigger param present but getTrigger() returns null, null added to list, later iteration NPE.
Test case:
```java
if (sa.hasParam("RemoveThisTrigger")) {
    removedTriggers.add(sa.getTrigger()); // may add null
}
```
Impact: NullPointerException during trait application.

#### 1. Card.addChangedCardTraits signature change breaks callers
Description:
The method signature changed to include removedTriggers, but several callers were updated with null. However, `addChangedCardTraitsByText` still creates CardTraitChanges with null removedTriggers, which is fine, but `getChangedCardTraitsList` creates CardTraitChanges with 6 args (landManaAbilities, null, null, null, null,null). The constructor expects 8 args: spells, removedAbilities, trigger, removedTriggers, replacements, statics, removeAll, removeNonMana. The call passes null for removedTriggers correctly, but older code may compile against old signature.
Test case:
```java
// Existing code compiled with old signature may still reference old method with 5 args
Card card = ...
card.addChangedCardTraits(spells, removedAbilities, trigger, replacements, statics, removeAll, removeNonMana, ts, id);
// This now resolves to new signature with removedTriggers inserted, causing arg shift.
```
Impact: Potential compile/runtime mismatch, breaking backwards compatibility.

#### 2. Player.addChangedCardTraits call passes null for removedTriggers but signature expects 9 args
Description:
Player.java line 1012 now passes null for removedTriggers, but the call originally had 8 args. The change is correct, but if any other caller missed update, it will fail at runtime.
Impact: Silent failure or mis-ordered arguments.

#### 1. RemoveThisTrigger interacts with perpetual animation
Description:
Runeblade Raiser uses RemoveThisTrigger$ True on DBAnimate subability, which is perpetual. The trigger removal occurs only during animate effect setup, but perpetual animation may reapply traits multiple times, possibly removing the trigger multiple times or failing to remove after first application.
Test case:
```java
// First animate: removedTriggers contains trigger reference
// Unanimate: trait removed, trigger returns?
// Perpetual animation may cause duplicate removal entries.
```
Impact: Trigger may be removed prematurely or not at all, breaking "perpetually loses this ability" rule.

#### 2. Trigger removal based on toString may collide with distinct triggers
Description:
Different triggers with same toString but different internal state will be incorrectly removed.
Impact: Unintended trigger removal affecting other cards.

### Performance and security
#### 1. O(n²) trigger removal loop
Description:
Card.updateTriggers iterates over removedTriggers and for each iterates over entire list to find match, resulting in O(n²) complexity.
Test case:
```java
for (final Trigger t : ck.getRemovedTriggers()) {
    for (final Trigger l : list) {
        if (t.getMode() == l.getMode() && t.toString().equals(l.toString())) { ... }
    }
}
```
Impact: Performance degradation with many triggers.

#### 2. No validation of RemoveThisTrigger param value
Description:
AnimateEffectBase checks hasParam("RemoveThisTrigger") but does not validate value is "True". Param may exist with false value.
Impact: Incorrect trigger removal.

## Verdict
Request Changes
The PR introduces trigger removal support for Runeblade Raiser but has several correctness and safety issues: fragile trigger matching via toString, potential NPEs from null triggers, O(n²) removal, and missing validation of param values. The changes need hardening before merge.

## Recommended Fixes
1. Validate RemoveThisTrigger param is "True" and sa.getTrigger() is non-null before adding to removedTriggers.
2. Use stable identifier for trigger matching instead of toString equality (e.g., compare Trigger IDs or canonical string).
3. Guard CardTraitChanges.copy against null removedTriggers and initialize safely.
4. Add unit tests for perpetual trigger removal and edge cases with null triggers.
5. Refactor trigger removal to use Set for O(n) matching.
---

PR #7996, opened 2025-07-11
Fix Bands with other for humans

tehdiplomat:


Eradev:
Can't you just make the check non case-sensitive so that doesn't happen again?

tehdiplomat:
Its defined in the Keyword. It broke because it was refactored and not tested.

Eradev:
Then couldn't it use the keyword instead of hard coding it again?

card.hasKeyword(Keyword.BANDSWITH), or even make a card.hasBanding() with a TrackableProperty?

Hanmac (approved):

tehdiplomat:
TrackableProperty doesn't work because its not a CardView. But yea, I can just compare to the keyword itself.

PR #7998, opened 2025-07-11
Hemosymbic Mite (EOE)

Fulgur14:


forge-gui/res/cardsfolder/upcoming/hemosymbic_mite.txt
```diff
@@ -0,0 +1,7 @@
+Name:Hemosymbic Mite
+Types:Creature Mite
```
squee1968, line 3: missing mana cost?

tool4ever (approved):

PR #8008, opened 2025-07-12
Allow the display of the placeholder in shops and reward windows.

Eradev:
#7859 only for the desktop version. I can't manage to build the Android version so I haven't touched it.

Allows to display the placeholder in the tooltip by dragging down on the card.


forge-gui-mobile/src/forge/adventure/util/RewardActor.java
```diff
@@ -548,30 +562,66 @@ private void generateBackFace(Reward r, Texture t) {
...
+        };
+    }
+
     private void switchTooltip() {
         if (!Reward.Type.Card.equals(reward.type))
             return;
-        if (!reward.getCard().hasBackFace())
-            return;
         if (GuiBase.isAndroid() || Forge.hasGamepad()) {
+            if (!reward.getCard().hasBackFace())
```
kevlahnota, line 592: is this check still necessary?

forge-gui-mobile/src/forge/card/CardImageRenderer.java
```diff
@@ -146,7 +158,7 @@ public static void drawCardImage(Graphics g, CardView card, boolean altState, fl
         y += headerHeight;
 
         float artWidth = w - 2 * artInset;
-        float artHeight = artWidth / CardRenderer.CARD_ART_RATIO;
+        float artHeight = !showArtBox ? 0f : artWidth / CardRenderer.CARD_ART_RATIO;
```
kevlahnota, line 161: it will hide the art image if the selected border options in the settings is "Art", what happens on classic mode view and the deck editor views?

forge-gui-mobile/src/forge/adventure/util/RewardActor.java
Eradev: Yes, it is. I just moved the check from above to inside this function to retain the same functionality as before on Android.

forge-gui-mobile/src/forge/card/CardImageRenderer.java
Eradev: It doesn't affect Deck Editor. Only the Reward/Shop screen.

kevlahnota (approved):

PR #8009, opened 2025-07-12
Support code2 and alias in sheets

Eradev:
Fixes a crash when creating AKH block boosters.


forge-core/src/main/java/forge/card/CardEdition.java
```diff
@@ -555,6 +555,20 @@ public List<PrintSheet> getPrintSheetsBySection() {
             }
 
             sheets.add(sheet);
+
+            String code2 = this.getCode2();
```
tool4ever: Hmn, I don't think duplicating static data like that is a smart choice memory wise
shouldn't we have a global function that translates these back to the main code instead?

forge-core/src/main/java/forge/card/CardEdition.java
```diff
@@ -852,7 +852,10 @@ public CardEdition get(final String code) {
             }
 
             CardEdition baseResult = super.get(code);
-            return baseResult == null ? aliasToEdition.get(code) : baseResult;
+
+            return baseResult == null
+                ? aliasToEdition.get(code)
+                : baseResult;
```
Hanmac: why this extra lines?

Eradev: Removed. Force of habit.

Hanmac (approved):

forge-core/src/main/java/forge/item/generation/BoosterGenerator.java
```diff
@@ -64,6 +64,28 @@ private static PaperCard generateFoilCard(List<PaperCard> cardList) {
...
 
+    public static PrintSheet tryGetStaticSheet(String sheetName) {
+        if (staticSheetsCorrespondance.containsKey(sheetName)) {
+            return StaticData.instance().getPrintSheets().get(staticSheetsCorrespondance.get(sheetName));
+        };
+
+        String passedEdition = sheetName.split(" ")[0];
+        CardEdition realEdition = StaticData.instance().getEditions().get(passedEdition);
+
+        if (realEdition == null) {
```
tool4ever: is there even a use case where this can happen?
if we know this will just crash no point in doing it like this

Eradev: Don't know what I was doing there. Must've been drunk. Fixed.

PR #8014, opened 2025-07-12
Update scouring_swarm.txt

Fulgur14:


forge-gui/res/cardfolder/upcoming/eumidian_wastewaker.txt
```diff
@@ -9,6 +9,6 @@ SVar:DBSac:DB$ SacrificeAll | ValidCards$ Card.ChosenCardStrict | SubAbility$ DB
 SVar:DBDraw:DB$ Draw | NumCards$ X | SubAbility$ DBCleanup
 SVar:DBCleanup:DB$ Cleanup | ClearChosenCard$ True
 K:Encore:6 B B
-SVar:X:Count$ValidGraveyard Land.ChosenCard
+SVar:X:Count$ValidGraveyard Land.ChosenCard+!token
```
tool4ever: why do you think this is needed?

Fulgur14: A token land could be sacrificed, but it wouldn't count as a land card put into graveyard.

tool4ever: we are never truly adding a token to its final zone
I'd be interested if you can find a case where it matters that we don't wait for the responsible SBA instead

Fulgur14: OK then.

tool4ever (approved):

PR #8024, opened 2025-07-13
Fix Spacecraft type collision

tool4ever:


forge-core/src/main/java/forge/card/CardType.java
```diff
@@ -659,33 +659,36 @@ public void sanisfySubtypes() {
...
+        if (isBattle()) {
+            allowedTypes.add(Predicates.IS_BATTLE_TYPE);
         }
-        if (!isPlane()) {
-            subtypes.removeIf(Predicates.IS_PLANAR_TYPE);
+        if (isPlane()) {
+            allowedTypes.add(Predicates.IS_PLANAR_TYPE);
         }
+
+        subtypes.removeIf(t -> allowedTypes.stream().noneMatch(p -> p.test(t)));
```
Hanmac: `subtypes.retainAll(allowedTypes)`

?

tool4ever: doesn't exist like that ;(

I assume building it as Set\<String> would be more expensive

Hanmac: How about:

```java
        Predicate<String> remove = s -> true;
        if (isCreature() || isKindred()) {
            remove.and(Predicates.IS_CREATURE_TYPE.negate());
        }

        subtypes.removeIf(remove);
```

then you don't need to steam the allowedTypes when the Predicate can handle this?

Hanmac (approved):

PR #8031, opened 2025-07-14
Fix token generation in puzzles

### LINKED ISSUE #8026: Puzzle Mode is broken for any puzzle containing token generation
**Describe the bug**
As of the current build, any puzzle containing token generation (via the T: tag) will fail to load completely, with at least one half of the battlefield being empty due to the token creation error.

**To Reproduce**
1. Try playing e.g. Possibility Storm - Foundations 02 in Puzzle Mode.
2. See the puzzle fail to load.

**Expected behavior**
The puzzle should load and generate tokens correctly.

Eradev:
Fixes #8026


Hanmac:
i used `CardEdition.UNKNOWN_CODE` which uses the constant directly

Hanmac (approved):

Issue #8025, opened 2025-07-13
[Planar Conquest] Starting commanders and planeswalkers are chosen from promo sets


Agetian:
**Describe the bug**
Something must have broken over time in the Planar Conquest mode, since the starting commanders and planeswalkers are once again selected primarily from the promo sets, resulting in various alternative art and alternatively named planeswalkers being chosen instead of the actual core ones. This doesn't respect the art choice option as well, so, no matter whether you choose "original art" or "latest art", the choice will still be the same.

**To Reproduce**
1. On mobile Forge, start a new Planar Conquest.
2. Choose any plane.
3. Look at the choices for the commanders and planeswalkers. Many will come from SLD or from various promo sets.

**Expected behavior**
Ideally, this should respect the art preference option. Alternatively, it should at least provide the original art in English language so that it's easy enough to figure out what is what.

Agetian:
Any opinions, @kevlahnota @Hanmac @tool4ever ? I tried debugging this, but couldn't determine the origin of the issue. Trying to limit the choice of sets (by filtering out promo types etc.) in NewConquestScreen (e.g. EnsureRegionsLoaded) doesn't seem to do the trick...

Eradev:
This was fixed already?

Agetian:
Yep, it's fixed, all good :) Thanks!

PR #8033, opened 2025-07-14
Planar Conquest fixes

Eradev:
Should alleviate problems in #8025

* Only add commanders once.
* Fix REPRINT_SET_TYPES check for Planeswalkers.
* Allow `cards.txt` to specify a set.

### DIFF
--- a/forge-gui/res/conquest/planes/Alara/cards.txt
+++ b/forge-gui/res/conquest/planes/Alara/cards.txt
@@ -18,14 +18,14 @@ Duskmantle Prowler
 Duty-Bound Dead
 Knight of Glory
 Knight of Infamy
-Nefarox, Overlord of Grixis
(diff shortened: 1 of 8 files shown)


forge-gui/res/conquest/planes/Alara/cards.txt
```diff
@@ -18,14 +18,14 @@ Duskmantle Prowler
...
 Servant of Nefarox
 Cathedral of War
-Meren of Clan Nel Toth
-Jazal Goldmane
-Jhessian Thief
-Rhox Faithmender
+Meren of Clan Nel Toth|C15
+Jazal Goldmane|C14
+Jhessian Thief|ORI
+Rhox Faithmender|J25
```
Agetian: The Alara plane probably shouldn't use the new sets (e.g. J25) as it was designed with the original Alara block in mind

Issue #8034, opened 2025-07-14
[AI] AI casts Olivia's Wrath when it doesn't have any vampire in play.


Eradev:
**Describe the bug**
AI will cast Olivia's Wrath even when it doesn't have any vampire in play.

**Expected behavior**
Should only cast if it can kill one or more creatures.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version 2.0.0.5-SNAPSHOT-07.14

Agetian:
Should be fixed with #8035, please reopen in case this still happens.

PR #8035, opened 2025-07-14
AI hint for Olivia's Wrath

Agetian:
Helps #8034


Hanmac:
@Agetian I think this is a General problem, that AI should not use -X/-X abilities where X would resolve to Zero 🤔

Agetian:
Yeah, true, I'll look into why that is happening too

PR #8041, opened 2025-07-14
Warp/Void EOE scripts, part 3

Fulgur14:


forge-gui/res/cardsfolder/upcoming/starwinder.txt
```diff
@@ -0,0 +1,9 @@
+Name:Starwinder
+ManaCost:5 U U
+Types:Creature Leviathan
+PT:7/7
+T:Mode$ DamageDone | ValidSource$ Creature.YouCtrl | ValidTarget$ Player | CombatDamage$ True | TriggerZones$ Battlefield | Execute$ TrigDraw | OptionalDecider$ You | TriggerDescription$ Whenever a creature you control deals combat damage to a player, you may draw that many cards.
```
tool4ever: not optional

Fulgur14: No, this is, in fact, optional.

tool4ever: I was still on my first coffee!

tool4ever: However it's still important you move the Optional to the effect line because of 121.2b

tool4ever: for some reason it's also called OptionalDecider there :p

tool4ever (approved):

PR #8042, opened 2025-07-14
Ai ability updates

tehdiplomat:


forge-ai/src/main/java/forge/ai/AiAbilityDecision.java
```diff
@@ -0,0 +1,9 @@
+package forge.ai;
+
+public record AiAbilityDecision(int rating, AiPlayDecision decision) {
```
tool4ever, line 3: image

tehdiplomat: No I tested this. The syntactic sugar is in the Build Tools we use:

https://github.com/Card-Forge/forge/blob/11815a4cf2e395c5d41fcea40465379b1285aaa2/forge-core/src/main/java/forge/card/CardEdition.java#L210

tool4ever: oh good news

forge-ai/src/main/java/forge/ai/SpellAbilityAi.java
```diff
@@ -220,30 +246,30 @@ protected boolean doTriggerAINoCost(final Player aiPlayer, final SpellAbility sa
...
-                    return true;
+                    return new AiAbilityDecision(100, AiPlayDecision.WillPlay);
                 }
             }
 
-            return false;
+            return new AiAbilityDecision(0, AiPlayDecision.TargetingFailed);
         }
-        return true;
+        return new AiAbilityDecision(0, AiPlayDecision.CantPlayAi);
```
tool4ever: logic inverted
the problem is canPlayWithoutRestrict leads to checkApiLogic which when not defined for api only causes random success
this breaks something like a Static Cleanup trigger in an almost impossible to debug way

forge-ai/src/main/java/forge/ai/SpellAbilityAi.java
```diff
@@ -81,23 +86,23 @@ protected boolean canPlayWithoutRestrict(final Player ai, final SpellAbility sa)
...
             if (!alwaysOnDiscard && !checkPhaseRestrictions(ai, sa, ai.getGame().getPhaseHandler(), logic)) {
-                return false;
+                return new AiAbilityDecision(0, AiPlayDecision.MissingPhaseRestrictions);
             }
         } else if (!checkPhaseRestrictions(ai, sa, ai.getGame().getPhaseHandler())) {
-            return false;
+            return new AiAbilityDecision(0, AiPlayDecision.MissingPhaseRestrictions);
         }
 
         if (!checkApiLogic(ai, sa)) {
```
tool4ever: can you also refactor this one?
it's actually used by more effects

I'll gladly update the ones still overriding `canPlayAI` in a follow-up because they really should just use one of the more tailored entrypoints

forge-ai/src/main/java/forge/ai/SpecialCardAi.java
```diff
@@ -78,16 +78,17 @@ public class SpecialCardAi {
 
     // Arena and Magus of the Arena
     public static class Arena {
-        public static boolean consider(final Player ai, final SpellAbility sa) {
+        public static AiAbilityDecision consider(final Player ai, final SpellAbility sa) {
```
tool4ever, line 81: it seems like some of the `consider` functions here have been arbitrarily skipped

maybe there's some kind of reasoning but then it should probably be documented there at least

forge-ai/src/main/java/forge/ai/SpellAbilityAi.java
tehdiplomat: Sure i'll fiddle with it tonight.

forge-ai/src/main/java/forge/ai/SpecialCardAi.java
tehdiplomat: Ahh yea, i didn't realize how these were setup at first, i'll take another pass through and set them all to have the same interface.

tool4ever (approved):
LGTM

PR #8046, opened 2025-07-15
Tokens cleanup (D-E sets)

Eradev:


forge-gui/res/editions/Dominaria.txt
```diff
@@ -308,6 +308,10 @@ ScryfallCode=DOM
...
 11 g_1_1_saproling @Mike Burns
-11 g_1_1_saproling @Mike Burns
-11 g_1_1_saproling @Mike Burns
+12 g_1_1_saproling @Mike Burns
+13 g_1_1_saproling @Mike Burns
 14 c_0_0_a_construct_total_artifacts @Victor Adame Minguez
+
+[other]
+15 emblem_jaya_ballard @Magali Villeneuve
+16 emblem_teferi-_ero_of_dominaria @Chris Rallis
```
Hanmac: Is this a Typo? Shouldn't it be `_hero`?

forge-gui/res/editions/Double Masters.txt
```diff
@@ -745,10 +745,10 @@ ScryfallCode=2XM
 4 w_2_2_cat @Jesper Ejsing
 5 w_1_1_human_soldier @Deruchenko Alexander
 6 w_1_1_soldier @Jakub Kasper
-7 u_2_1_a_phyrexian_myr @Daarken
+7 u_2_1_a_myr @Daarken
```
Hanmac, line 748: @Eradev the problem there is that there is no token script for u_2_1_a_myr anymore because it got updated to the phyrexian one

Eradev: Hmm... Maybe we should keep it as tokens don't get errata'd, even if it is not used anymore at the moment? M3C already has the updated token.

forge-gui/res/editions/Dominaria.txt
Eradev: Fixed.

forge-gui/res/editions/Double Masters.txt
Hanmac: Then old LivingWeapon cards wouldn't find the old token anymore, for example
because then it would look for germ_phyrexian in old set

Eradev: I'll do some testing but it should fallback to searching for that specific token if it isn't in the set.

Eradev: Yeah there is no problem on that front. You can try with Batterskull in NPH. It doesn't have a Germ token and it'll fetch the Phyrexian one without any problems.

Hanmac: > it'll fetch the Phyrexian one without any problems.

but not from the right set, right?

Eradev: Not every set has the right tokens. It should PREFER the current set if it exists, but fallback to an existing one if it doesn't.

forge-gui/res/editions/New Phyrexia.txt
```diff
@@ -191,6 +190,6 @@ TokenFallbackCode=MBS
 
 [tokens]
 1 g_3_3_beast @Dave Allsop
-2 r_1_1_phyrexian_goblin_haste @Jaime Jones
```
Hanmac: The problem with this token is:
`Chancellor of the Forge` creates "Phyrexian Goblin creature token with haste",
but there is no image for Haste Token

Eradev: There is already a phyrexian goblin token. It is a closer match. I'll add it to the sets that has it temporarily until they make one with haste (if ever)

https://scryfall.com/search?as=grid&order=released&q=oracleid%3Ab43340e3-bb76-4501-8db2-040b7cae33b7+include%3Aextras&unique=prints

Eradev: Duplicated in #8049

forge-gui/res/editions/Double Masters.txt
Hanmac: @Eradev how about "duplicate" the TokenScripts for this, so that both are defined like in the other MRs?

Eradev: There is already a Phyrexian Myr token printed in another set.

Hanmac: I still would prefer to keep it consistent with existing tokens

Eradev: I only duplicated them when existing tokens don't exist yet though (as in no printed version exist)? Otherwise I think it is better to have their correct printing to reflect their type and name correctly without having to double check.

Hanmac:
Found the note again I have been looking for 
> Dominaria United series of 26 tokens
> 
> Dominaria United comes with two sets of tokens: a series of 26 bearing either DMU or DMC set codes, and a separate series of 12 tokens with exclusively the DMC set code. We don't know why they did things this way. We've opted to file the entire series of 26 here regardless of set code to avoid archival conflicts.

Eradev:
We should keep them in DMU anyway as there would be conflicts if we try to move them to DMC. They share some numbers. :/

forge-gui/res/editions/Double Masters.txt
Hanmac: @tehdiplomat what do you think? should the old Images be duplicated for the new TokenScripts?
or should they just fallback to newer Token images?

Hanmac:
@Eradev need to rebase/resolve conflict

Hanmac (approved):

PR #8048, opened 2025-07-15
New 16x16 sprites and avatars for planeswalkers &c.

lemtom:
The chibi sprites of the various planeswalkers and several iconic creatures look really nice, but are static and don't really fit the established scale and style. Largely based on existing sprites, I've created 16x16 alternatives (plus matching avatars).

I've based myself on the card art, but I can only do so much in 16x16 pixels.


Simisays:
This looks awesome, a huge improvement to be sure !

The downscale in the enemies.json for these enemies will have to be removed but that can be done later.

lemtom:
Aha, sorry, I was not aware of that. I was already proud I noticed the atlas. 😂

Agetian (approved):

PR #8049, opened 2025-07-15
Tokens cleanup (F-G sets)

Eradev:


forge-gui/res/editions/Guilds of Ravnica Guild Kit.txt
```diff
@@ -137,15 +136,20 @@ TokenFallbackCode=GRN
...
-6 rw_1_1_soldier_haste @David Palumbo
-7 g_3_3_centaur @Slawomir Maniak
+# 5☇ bg_1_1_insect @Ben Wootten
+6 rw_1_1_soldier @David Palumbo
 8 g_1_1_saproling @Brad Rigney
+# 8☇ gw_2_2_elf_knight_vigilance @Tomasz Jedruszek
 9 g_5_5_wurm_trample @Anthony Palumbo
 10 gw_8_8_elemental_vigilance @Yeong-Hao Han
+# 10☇ g_3_3_centaur @Slawomir Maniak
 11 voja @Richard Sardinha
```
Hanmac: The Token for GK1 are wrong on Scryfall, i already wrote them a message, but no response

check the Numbers of the Tokens compared with Scryfall

https://scryfall.com/card/tgk1/4/goblin-soldier has wrong Collector Number, that's GRN 004, not GK1 004

the correct GK1 004 Goblin token is this image:
https://scryfall.com/card/tgk1/3/weird-goblin?back

Eradev: And it's missing GTK7 I believe. That thing is so confusing. XD

EDIT: Nvm. That's the centaur...

Hanmac: i think you should keep the old numbers for now, until i get some answers from scryfall

forge-gui/res/editions/Fallen Empires.txt
```diff
@@ -198,10 +198,3 @@ ScryfallCode=FEM
 100 U Ruins of Trokair @Mark Poole
 101 R Sand Silos @Pat Lewis
 102 U Svyelunite Temple @Mark Poole
-
-[tokens]
-w_1_1_citizen
```
tehdiplomat, line 203: Why are we removing these tokens? They are definitely generated by cards in this set.

Eradev: There are no copies of these tokens in the set. They'll fallback to using an existing one from another set.

tehdiplomat: Thats not exactly true. Scryfall may not have them, but tokens exist.

Eradev: The tokens list is mostly there for download purposes. Nothing stops users from putting using their own tokens in their folder.

Hanmac:
Fallout Set probably needs to be renamed?
its just `Fallout (PIP)` on Scryfall

Hanmac:
Keep

PR #8058, opened 2025-07-15
Cleaning up Innistrad structure

shenshinoman:
Cleaning up the file structure to move all Innistrad maps under the Innistrad folder, as well as some minor logic change to Shandalar Old Border's config.json to make maintenance easier. (By setting all the desired sets to be viewed in allowedEditions, it will only shows those, and we no longer have to restrict every new edition as it comes in.)

Finally, added block info for Alchemy: Innistrad, so it should start being available in events as well.


shenshinoman:
as an aside: I really have no idea why it says I've modified all the "upcoming" cards. I had merely merged the newest commits into my branch before making the pull, and every "change" to those is showing removing all the text then re-adding all the text exactly as-is.

Hanmac:
> as an aside: I really have no idea why it says I've modified all the "upcoming" cards. I had merely merged the newest commits into my branch before making the pull, and every "change" to those is showing removing all the text then re-adding all the text exactly as-is.

That's a bug in github, we forced the cards to be LF, but when @Fulgur14 is uploading them, they are CRLF (and github doesn't fix that)

so each PR will show these files as changed

i will make a Commit fixing these soon, and then you can rebase your MR

tehdiplomat (approved):

tehdiplomat:
@Hanmac is this PR waiting on your proposed changes?

Hanmac:
> @Hanmac is this PR waiting on your proposed changes?

There is no changes waiting for me, feel free to merge this

Hanmac:
@shenshinoman my only question, why the remove of the one "Chocobo Racetrack"?

shenshinoman:
> @shenshinoman my only question, why the remove of the one "Chocobo Racetrack"?

Technically, that wasn't me. Someone else removed it in another push. (I think it was throwing errors that it wasnt ready/found?). That I had merged into mine before pushing the main commit, to make sure no cobflicts.

It got flagged as edited on my build then. The difference showing it being removed (and my version re-adding it) was what drew my attention. As doing so was now only 39 cards, I edited my original deck with an extra Sabotender to bring it back to 40, rather than get a random waste. 

As the deck, and plane it is part of, is my creation but INCREDIBLY early. (Like, I wouldn't be insulted if we wanted to just remove that plane from live until I have more. I just know some folks had shown interest in it.) I took the quick fix as it's a starter deck for that plane, and planned to later actually find out why it was removed once I am dedicating more time to expanding that specific plane.

Hanmac:
@tehdiplomat if you merge this, please do a Squash Merge if able

PR #8061, opened 2025-07-15
Force LF in cardsfolder

Eradev:
Force LF in text files inside cardsfolder so we don't always have to commit line changes


tool4ever:
cardsfolder already has it so this probably fixes nothing:
https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/.gitattributes

Eradev:
> cardsfolder already has it so this probably fixes nothing: https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/.gitattributes

I didn't see that it was lf in cardsfolder.... Why is that one folder not crlf like everything else?

Hanmac:
The problem is that @Fulgur14 upload the files directly into github, and that somehow ignores gitattributes

Fulgur14:
I don't really understand it myself, I'm not sure if I should be doing something differently. Lately, when I just do small number of new cards, I create the files in github, and that seems to not have problems, but it's a bit troublesome doing it with large number of cards.

Eradev:
Oh I see. The Web UI only allows LF so it makes sense that they end up conflicting then.

Hanmac:
other way around, the The Web UI doesn't check when you upload the files directly
(it would enforce them in the MRs, but it doesn't)

Eradev:
Would the workflow route work?

Hanmac:
Different direction, we want to normalize them to LF instead

PR #8063, opened 2025-07-15
Update famished_worldsire.txt

Valensior:
ability of putting cards from library on battlefield could affect all cards. added "changevalid$ land"


tool4ever (approved):
thanks!

PR #8065, opened 2025-07-15
Better image fetcher error

Eradev:
* Display file location if user wants to put their own missing art when no download URL was found.
* Early exit if file already exists.
* Stop spamming "key is malformed" message.


kevlahnota (changes requested):
needs to resolve conflicts

kevlahnota (approved):

PR #8077, opened 2025-07-16
Advanced filtering using the search box

Eradev:
Proof of concept and beginning of the implementation of filtering using the search box à la Scryfall.

image


Hanmac:
A way bigger change for later would be to allow `()`
And stuff like `t:creature or t:artifact`

Eradev:
~~It's just a starting point. Hopefully people will add onto it. :)~~

Now it does!

forge-gui/src/main/java/forge/itemmanager/advancedsearchparsers/RarityParser.java
```diff
@@ -0,0 +1,226 @@
...
+    /**
+     * Handles exact rarity check of a card
+     * @param tokenValue Token value
+     * @return Predicate or null
+     */
+    public static Predicate<PaperCard> handleExact(String tokenValue) {
+        Predicate<PaperCard> predicate = null;
+        switch(tokenValue) {
+            case "c":
+            case "common":
```
Hanmac: Add case for "l" "land"

also, the following methods have duplicated code,
probably extract the "string" to CardRarity into an extra function

Eradev: Done

Hanmac:
More love for the RarityParser:
* Use `CardRarity.smartValueOf` to get the Enum Value
* Use `CardRarity.compareTo`  that compares the ordinal instead of extra RARITY_RANK

Eradev:
`CardRarity.smartValueOf` doesn't work for "land", it'd need to be "basic land". Same for "Mythic Rare"

Eradev:
And the RARITY_RANK is there to stop processing on rarity not supported like special or unknown.

Hanmac:
Hm, I still want to clean up the CardRarity stuff

I will do a smaller MR to support "land" and "Mythic" too

tool4ever (approved):

Hanmac (approved):
i will do some CardRarity cleanup later

PR #8080, opened 2025-07-16
Try to use current set when creating a card from an effect

### LINKED ISSUE #7445: Select images to be used on cards conjured by Alchemy's card effects
For example, in Mtg Arena, the effect of [[Thayan Evokers]] conjures a [[Lightning Bolt]] illustrated in Commander Legends: Battle for Baldur's Gate (CLB).

In the current Forge, the above effect conjures a [[Lightning Bolt]] illustrated in Ravnica: Cluedo Edition (CLU) (this is the same illustration as M10). 

This is probably because the CLU illustration is the most recent [[Lightning Bolt]] in existence, and the conjuring in Forge references the newest illustration.
(It seems to show the newest downloaded in the cache, not the newest implemented.)

Of course, this does not affect the operation of the play at all, but I was concerned about the conjuring of cards from different sets and different dimensions.

Personally, I want to can [[Oracle of the Alpha]] conjures a Power 9 for the Vintage Masters illustration.

Image
Image

Eradev:
Fixes #7445 

* Try to use current set when creating a card from an effect
* Removed conjured-only cards from shops/rewards (Might need more testing)
* Fix CardDb fallback

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/MakeCardEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/MakeCardEffect.java
@@ -152,11 +152,8 @@ public class MakeCardEffect extends SpellAbilityEffect {
                             pc = Iterables.getLast(IterableUtil.filter(pack, PaperCardPredicates.name(name)));
                         } else {
                             // Try to get the card in the sa host's current edition
-                            pc = StaticData.instance().getCommonCards().getCard(name, sa.getHostCard() != null ? sa.getHostCard().getSetCode() : CardEdition.UNKNOWN_CODE);
-
-                            if (pc == null) {
-                                pc = StaticData.instance().getCommonCards().getCard(name);
-                            }
+                            String editionCode = sa.getHostCard() != null ? sa.getHostCard().getSetCode() : CardEdition.UNKNOWN_CODE;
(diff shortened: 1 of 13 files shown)


forge-game/src/main/java/forge/game/ability/effects/MakeCardEffect.java
```diff
@@ -150,7 +151,12 @@ public void resolve(SpellAbility sa) {
                         if (pack != null) {
                             pc = Iterables.getLast(IterableUtil.filter(pack, PaperCardPredicates.name(name)));
                         } else {
-                            pc = StaticData.instance().getCommonCards().getUniqueByName(name);
+                            // Try to get the card in the sa host's current edition
+                            pc = StaticData.instance().getCommonCards().getCard(name, sa.getHostCard() != null ? sa.getHostCard().getSetCode() : CardEdition.UNKNOWN_CODE);
+
+                            if (pc == null) {
```
tool4ever: imo this part shouldn't be needed
because CardDB is already smart enough to give you fallback

Eradev: I crashed with Oracle of the Alpha (MB2) without it. I can fix the CardDb instead tho.

PR #8095, opened 2025-07-17
More console commands

Eradev:
More console commands for easier testing.


Eradev:
Oops. Based on wrong branch

PR #8096, opened 2025-07-17
More console commands

Eradev:
More console commands for easier testing.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/stage/ConsoleCommandInterpreter.java
+++ b/forge-gui-mobile/src/forge/adventure/stage/ConsoleCommandInterpreter.java
@@ -11,11 +11,13 @@ import forge.adventure.data.EnemyData;
 import forge.adventure.data.PointOfInterestData;
 import forge.adventure.data.WorldData;
 import forge.adventure.pointofintrest.PointOfInterest;
+import forge.adventure.util.AdventureEventController;
 import forge.adventure.util.Current;
 import forge.adventure.util.Paths;
 import forge.adventure.world.WorldSave;
 import forge.card.CardEdition;
 import forge.card.ColorSet;
+import forge.deck.CardPool;
 import forge.deck.Deck;
 import forge.deck.DeckProxy;
 import forge.game.GameType;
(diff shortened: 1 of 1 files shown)


forge-gui-mobile/src/forge/adventure/stage/ConsoleCommandInterpreter.java
```diff
@@ -260,7 +262,49 @@ private ConsoleCommandInterpreter() {
...
+                    amount = Integer.parseInt(s[1]);
+                } catch (NumberFormatException ignored) {}
+            }
+
+            for(int i=0; i<amount; i++) {
+                Current.player().addBooster(AdventureEventController.instance().generateBooster(edition.getCode()));
+            }
+
+            return "Added " + amount + " " + edition.getCode() + " booster(s)";
+        });
```
Jetz72: Think we've had a PR pending for a while to do this - #7676, Think yours is a bit better since it does quantity. Though I noticed that one has a `hasBoosterTemplate` check on the edition. Is that needed?

forge-gui-mobile/src/forge/adventure/stage/ConsoleCommandInterpreter.java
```diff
@@ -260,7 +262,49 @@ private ConsoleCommandInterpreter() {
...
+        registerCommand(new String[]{"clearNoSell"}, s -> {
+            CardPool cards = Current.player().getCards();
+            for(PaperCard c : cards.getFilteredPool(c -> c.getMarkedFlags().noSellValue).toFlatList()) {
+                cards.remove(c);
+            }
+            return "Removed all no sell flagged cards.";
```
Jetz72, line 311: Haven't these been changes so you can just sell them for 0 gold? Also while I do prefer camelCase to the multi-word commands, this does kinda break from the established style for commands.

Eradev: It was mostly to clear my inventory quickly after testing. I'll remove the caps.

forge-gui-mobile/src/forge/adventure/stage/ConsoleCommandInterpreter.java
Eradev: Oh I didn't see the PR. I'll add it to mine.

PR #8098, opened 2025-07-18
Better Games In Match Selection

krafczyk:
The number of games in a match is handled by the `FPref.UI_MATCHES_PER_GAME` preference value. This is only modifiable with a combobox on the new constructed game pages.

This preference value also affects other game modes such as booster draft and sealed draft. The player then sees no obvious way to change the number of games in the match on these pages.

This PR does the following:
* Adds a ModelBinder interface for linking GUI elements to model values more directly
* Adds a series of Binder specializations for existing GUI frameworks (desktop and mobile)
* Adds a 'Best Of' combo box to the Booster Draft and Sealed Draft pages for both desktop and mobile
* Adds binders so changes to the combo box are propagated to the saved preferences.

UI changes have been tested with desktop builds, and the changes work.

ModelBinder and its helpers also provide a path for expansion to other UI elements in the future.

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/gui/SwingPrefBinders.java
+++ b/forge-gui-desktop/src/main/java/forge/gui/SwingPrefBinders.java
@@ -1,0 +1,26 @@
+package forge.gui;
+
+import javax.swing.*;
+import java.awt.event.ItemEvent;
+import forge.localinstance.properties.ForgePreferences;
+import forge.model.FPrefsBinder;
+
+public class SwingPrefBinders {
+  public static final class ComboBox extends FPrefsBinder<JComboBox<String>, String> {
+    public ComboBox(ForgePreferences.FPref key, JComboBox<String> box) {
+      super(
+        key,
+        box,
+        b -> (String) b.getSelectedItem(),
+        (b, s) -> b.setSelectedItem(s),
+        s -> s,
+        s -> s);
+
+      box.addItemListener(e -> {
+        if (e.getStateChange() == ItemEvent.SELECTED) {
+          this.save();
+        }
+      });
+    }
+  }
+}
--- a/forge-gui-desktop/src/main/java/forge/screens/home/CLobby.java
+++ b/forge-gui-desktop/src/main/java/forge/screens/home/CLobby.java
@@ -1,8 +1,11 @@
 package forge.screens.home;
 
+import java.awt.event.ItemEvent;
+
 import java.util.Arrays;
 import java.util.Vector;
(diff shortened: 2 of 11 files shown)


krafczyk:
@paulsnoops , who is in charge of GUI changes? Is a small QOL change like this something that can be looked at soon?

paulsnoops:
> @paulsnoops , who is in charge of GUI changes? Is a small QOL change like this something that can be looked at soon?

I don't think we have a dedicated GUI person but please can you ask in #contribution-questions in the Discord rather than tagging devs not involved in the PR, thank you

Jetz72:
Do you have any ideas of where else we might apply these `ModelBinder` and `FPrefsBinder` classes? Could stuff like input validation and feedback for specific options or errors be worked into them in an elegant way?

Would it be reasonable to make the UI element hold its own binder object, instead of the screen containing it? Feels a little bloat-y for the containing screen to have two fields - one for the UI control, one for the an object that manages the UI control.

krafczyk:
This is my first contribution touching GUI components, so I don't have specific examples right this moment, however for this change, I wanted the 'Games in Match' combobox to reflect the current value stored in prefs no matter which 'screen' we were looking at. This would require a lot of boiler plate since I want this to appear on several screens (constructed lobby, load draft, load sealed, load winston for example), and I looked for a solution.

I felt the binder was small enough to represent this notion of synchronization, and could be reused for various GUI components, which I attempted to do by specializing it for FPrefs and the mobile/desktop.

**Could the UI element hold the binder:** Yes of course! I even thought of this, but didn't want to touch more stuff. I felt the scope was getting larger than I anticipated, so this was a compromise. I agree it would even be better, but I'm not sure how to do that since the desktop UI uses swing elements directly like JComboBox, so I'm not sure how to add an attribute to it unless we're willing to create a new class for it. the mobile element is easier because I believe most of that is under our control.

**Could the Binder handle input validation and feedback?**: Yes I bet it could! I wonder if it would make more sense to build out the FModel a bit with setters/getters that perform validation, then the binder would interact with those setters and getters instead. Specializations for each GUI framework could handle presentation of feedback for the user (for example launching a dialog), but I don't know enough yet about forge's GUI structure to know if this is the right angle for that.

krafczyk:
These binders can be used anywhere we want UI elements to reflect the current state of the forge model. I could imagine lots of settings could be handled with this setup.

kevlahnota (approved):

tool4ever (approved):
thanks

PR #8111, opened 2025-07-18
Adventure: Add give booster command

lilymnky-F:
Add a command to the debug console to give the player booster packs.

The syntax is "give booster SET" with SET being a valid set code. There is error checking both for if the set code is invalid, and if the set code cannot be generated as a booster pack.


Jetz72:
Someone beat you to it by a day (and someone else beat them to it by a month and a half but their PR fell through the cracks) - #8096, #7676.

lilymnky-F:
Got it, my bad

PR #8112, opened 2025-07-18
Final EOE batch (part 4)

Fulgur14:


forge-gui/res/cardsfolder/r/robber_of_the_rich.txt
```diff
@@ -11,5 +11,5 @@ SVar:STPlay:Mode$ Continuous | MayPlay$ True | MayPlayIgnoreColor$ True | Affect
 SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
 SVar:X:Count$ValidHand Card.DefenderCtrl
 SVar:Y:Count$ValidHand Card.YouOwn
-SVar:Z:Count$CreaturesAttackedThisTurn Creature.Rogue
+SVar:Z:Count$CreaturesAttackedThisTurn Creature.Rogue+YouCtrl
```
tool4ever: change not needed for this property

Fulgur14: Really? I thought that it would apply on opponent's turn, if they attacked with a Rogue.

tool4ever: in this case no card needed it yet, should probably be renamed at some point

tool4ever (approved):

PR #8127, opened 2025-07-19
Devour: better keyword

### LINKED ISSUE #4503: Devour needs better Wording/Handling
The Special Cases need to be coded better, i think:

* `Devour Food 3`  in Description, Food needs to be upper case
* `Devour artifact 1` artifact needs to be lower case, can be done via checking for CardType?
* `Devour land 3`
* `Devour N` should still use `Creature` internal, but not in Description
* `Devour X` the X is special there, can be done via `Devour::X:desc` ?

i don't like that we have `K:Devour Food:3` i would not like them have space separated

the Keyword might better use `KeywordWithAmountAndType` or new `KeywordWithTypeAndAmount` for the reminder text?

Hanmac:
Closes #4503


Agetian:
Looks good :)

PR #8133, opened 2025-07-20
Consolidate some PlayEffect logic

tool4ever:
Credits to:

- image

- image

### DIFF
--- a/forge-game/src/main/java/forge/game/player/Player.java
+++ b/forge-game/src/main/java/forge/game/player/Player.java
@@ -1705,22 +1705,16 @@ public class Player extends GameEntity implements Comparable<Player> {
     }
 
     public final boolean canPlayLand(final Card land, final boolean ignoreZoneAndTiming, SpellAbility landSa) {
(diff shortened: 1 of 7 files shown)


forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -1705,22 +1705,16 @@ public final Card playLandNoCheck(final Card land, SpellAbility cause) {
...
                 return false;
             }
-        }
-
-        // CantBeCast static abilities
-        if (StaticAbilityCantBeCast.cantPlayLandAbility(landSa, land, this)) {
-            return false;
-        }
 
-        if (land != null && !ignoreZoneAndTiming) {
```
Hanmac, line 1723: What's with the remove of `ignoreZoneAndTiming`?

tool4ever: it's not removed
just merged with the existing block above
(because I move the more expensive static check lower)

Hanmac (approved):

PR #8138, opened 2025-07-20
Tokens cleanup (R-S sets)

Eradev:


Hanmac (approved):

forge-gui/res/tokenscripts/g_1_1_insect_infect.txt
```diff
@@ -0,0 +1,7 @@
+Name:Insect Token
```
tool4ever, line 1: why are we adding tokenscripts for nothing?

Hanmac: @tool4ever see other MRs about the old non Phyrexian token

Hanmac: I didn't wanted to delete this one :/

I thought it was another g_1_1_insect

tool4ever: still doesn't make sense to me:
we shouldn't have to backport our internal naming scheme after errata
we get the image from CN anyway?
https://api.scryfall.com/cards/tsom/4

Hanmac: that why i suggested in #8046 to do it like this with duplicated entries:
https://github.com/Card-Forge/forge/blob/3e0d1a036cd9b0fcaa8bb3ed4c64eadf08876b8e/forge-gui/res/editions/Foundations.txt#L839-L841

but we didn't came to an conclusion yet

@tehdiplomat your input for this too please

tehdiplomat: I'm presonally not a fan of changing all of these tokens away from what we had them. So i don't think i really have further input.

Eradev: Tokens aren't cards and therefore aren't errata'd so it still officially exists as a 1/1 green Insect token, even if it isn't used at the moment.

forge-gui/res/editions/Scars of Mirrodin.txt
```diff
@@ -266,7 +265,7 @@ ScryfallCode=SOM
 1 w_2_2_cat @Scott Chou
 2 w_1_1_soldier @Goran Josic
 3 r_1_1_goblin @Goran Josic
-4 g_1_1_phyrexian_insect_infect @Adrian Smith
+4 g_1_1_insect_infect @Adrian Smith
```
Hanmac, line 268: @Eradev if we do fallback to newer TokenName too, then i would say it could be approved

PR #8146, opened 2025-07-21
Fix for red castle

Drecon84:
Hard mode didn't have the defeat dialogue because it spawned a different version of Lathliss without the defeat code.

(I had to duplicate the defeat code, which means that it is a little more error-prone to maintain, but hopefully that won't actually come up too much)


Simisays (approved):
👍

kevlahnota (approved):

PR #8147, opened 2025-07-21
Diplomatic Relations errata

Fulgur14:


tool4ever:
it's weird that it still showed a Diff

Hanmac:
@tool4ever probably because it wasn't rebased

PR #8148, opened 2025-07-21
Cast after warping only for owner

tool4ever:


tehdiplomat (approved):

Hanmac:
@tool4ever probably need to set the Effect Owner too

Right now, if the player that cast the Warp leaves the Game, the effect is gone too

Hanmac (approved):

PR #8153, opened 2025-07-22
Draft definitions for EOE

krafczyk:
Draft definitions for EOE guided using scryfall API data and https://mtgscribe.com/2025/07/16/edge-of-eternities-play-booster-fact-sheet/

### DIFF
--- a/forge-gui/res/blockdata/blocks.txt
+++ b/forge-gui/res/blockdata/blocks.txt
@@ -143,4 +143,4 @@ Aetherdrift, 3/6/DFT, DFT
 Tarkir Dragonstorm, 3/6/TDM, TDM
 Final Fantasy, 3/6/FIN, FIN
 Alchemy: Innistrad, 3/6/ISD, YMID
-Edge of Eternity, 3/6/EOE, EOE
+Edge of Eternities, 3/6/EOE, EOE
--- a/forge-gui/res/blockdata/blocks.txt
+++ b/forge-gui/res/blockdata/blocks.txt
@@ -143,3 +143,4 @@ Aetherdrift, 3/6/DFT, DFT
 Tarkir Dragonstorm, 3/6/TDM, TDM
 Final Fantasy, 3/6/FIN, FIN
 Alchemy: Innistrad, 3/6/ISD, YMID
+Edge of Eternity, 3/6/EOE, EOE
--- a/forge-gui/res/draft/rankings/eoe.rnk
+++ b/forge-gui/res/draft/rankings/eoe.rnk
@@ -1,17 +1,17 @@
 //Rank|Name|Rarity|Set
 #1|Singularity Rupture|R|EOE
 #2|Zero Point Ballad|R|EOE
-#3|Celestial Colonnade|R|EOE
+#3|Celestial Colonnade|R|EOS
 #4|Astelli Reclaimer|R|EOE
 #5|Beyond the Quiet|R|EOE
 #6|Cosmogrand Zenith|M|EOE
 #7|Exalted Sunborn|M|EOE
 #8|Quantum Riddler|M|EOE
 #9|Ouroboroid|M|EOE
-#10|Creeping Tar Pit|R|EOE
-#11|Lavaclaw Reaches|R|EOE
-#12|Lumbering Falls|R|EOE
-#13|Raging Ravine|R|EOE
+#10|Creeping Tar Pit|R|EOS
+#11|Lavaclaw Reaches|R|EOS
+#12|Lumbering Falls|R|EOS
+#13|Raging Ravine|R|EOS
 #14|Elegy Acolyte|R|EOE
 #15|Mightform Harmonizer|R|EOE
 #16|Alpharael, Stonechosen|M|EOE
@@ -22,7 +22,7 @@
 #21|Xu-Ifit, Osteoharmonist|R|EOE
 #22|Syr Vondam, Sunstar Exemplar|R|EOE
 #23|Starwinder|R|EOE
-#24|Stirring Wildwood|R|EOE
+#24|Stirring Wildwood|R|EOS
 #25|Sunstar Chaplain|R|EOE
 #26|Warmaker Gunship|R|EOE
 #27|Frenzied Baloth|R|EOE
@@ -55,12 +55,12 @@
 #54|Station Monitor|U|EOE
 #55|Memorial Vault|R|EOE
 #56|Tannuk, Steadfast Second|M|EOE
-#57|Green Sun's Zenith|M|EOE
-#58|Nexus of Fate|M|EOE
-#59|Ancient Tomb|M|EOE
-#60|Blinkmoth Nexus|R|EOE
-#61|Hissing Quagmire|R|EOE
-#62|Shambling Vent|R|EOE
+#57|Green Sun's Zenith|M|SPG
+#58|Nexus of Fate|M|SPG
+#59|Ancient Tomb|M|EOS
+#60|Blinkmoth Nexus|R|EOS
+#61|Hissing Quagmire|R|EOS
+#62|Shambling Vent|R|EOS
 #63|Pulsar Squadron Ace|U|EOE
 #64|Rescue Skiff|U|EOE
 #65|Sunstar Expansionist|U|EOE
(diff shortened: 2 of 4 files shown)


forge-gui/res/editions/Edge of Eternities Stellar Sights.txt
```diff
@@ -1,12 +1,188 @@
 [metadata]
 Code=EOS
 Date=2025-08-01
-Name=Edge of Eternities: Stellar Sights
+Name=Edge of Eternities: Steller Sights
```
tehdiplomat: Isn't it Stellar?

forge-gui/res/draft/rankings/eoe.rnk
```diff
@@ -0,0 +1,322 @@
+//Rank|Name|Rarity|Set
+#1|Singularity Rupture|R|EOE
+#2|Zero Point Ballad|R|EOE
+#3|Celestial Colonnade|R|EOE
```
paulsnoops: This should have the set EOS instead of EOE; there are quite a few cards in this draft from the sets EOS and SPG

forge-gui/res/editions/Edge of Eternities Stellar Sights.txt
```diff
@@ -1,12 +1,188 @@
...
 Type=Collector_Edition
 ScryfallCode=EOS
 
 [cards]
+1 M Ancient Tomb @Bruce Brenneise
```
paulsnoops: Looks like you're not working with the latest files here? The editions already have all cards in the latest updates

Edit: Your branch looks like it 200+ commits behind master

forge-gui/res/editions/Edge of Eternities.txt
```diff
@@ -4,32 +4,519 @@ Date=2025-08-01
...
+265 L Mountain @Adam Paquette
+266 L Forest @Adam Paquette
 367 L Plains @Adam Paquette
 368 L Island @Adam Paquette
 369 L Swamp @Adam Paquette
 370 L Mountain @Adam Paquette
 371 L Forest @Adam Paquette
+
+
+[viewport]
```
paulsnoops: These section names need to match what we can support otherwise they don't work. If you use the latest updates from the edition files they have already been grouped to what Forge knows

forge-gui/res/editions/Edge of Eternities.txt
```diff
@@ -4,32 +4,519 @@ Date=2025-08-01
...
+393 U Starfield Shepherd @Marta Nael
+394 U Annul @Carlos Palma Cruchaga
+395 U Umbral Collar Zealot @Dmitry Burmak
+396 U Kavaron Harrier @Hardy Fowler
+397 U Pull Through the Weft @Andrew Mar
+398 R Singularity Rupture @Néstor Ossandón Leal
+399 R Emissary Escort @Lius Lasahido
+
+[special guests]
+1 Warping Wall|EOE
```
paulsnoops: These should have |SPG instead of |EOE

forge-gui/res/editions/Edge of Eternities Stellar Sights.txt
krafczyk: Yep, I'll fix it.

forge-gui/res/editions/Edge of Eternities.txt
krafczyk: So, I see that these sections are actually defined here: `forge-core/src/main/java/forge/card/CardEdition.java`, So fine, I can take what's in master, however the issue is the distribution of rarities in the borderless section is different from the full draft set. This means It'll be impossible to achieve the stated probabilities here: https://mtgscribe.com/2025/07/16/edge-of-eternities-play-booster-fact-sheet/

I can change all the surreal, triumphant, and viewport sections to references to borderless, but the card frequencies will be incorrect.

forge-gui/res/editions/Edge of Eternities.txt
krafczyk: Adressing.

krafczyk:
@paulsnoops , I've updated my PR adressing your concerns.

forge-gui/res/blockdata/blocks.txt
```diff
@@ -143,3 +143,4 @@ Aetherdrift, 3/6/DFT, DFT
 Tarkir Dragonstorm, 3/6/TDM, TDM
 Final Fantasy, 3/6/FIN, FIN
 Alchemy: Innistrad, 3/6/ISD, YMID
+Edge of Eternity, 3/6/EOE, EOE
```
paulsnoops: Edge of Eternity > Edge of Eternities

forge-gui/res/editions/Edge of Eternities.txt
```diff
@@ -4,6 +4,54 @@ Date=2025-08-01
 Name=Edge of Eternities
 Type=Expansion
 ScryfallCode=EOE
+TokensCode=TEOE
```
paulsnoops: I don't think this is needed? I think we look for T by default @Hanmac

forge-gui/res/editions/Edge of Eternities.txt
```diff
@@ -4,6 +4,54 @@ Date=2025-08-01
...
+https://mtgscribe.com/2025/07/16/edge-of-eternities-play-booster-fact-sheet/
```
paulsnoops: Needs to be commented with #

forge-gui/res/editions/Edge of Eternities.txt
Hanmac: Yeah TokenCode isn't needed in this case

forge-gui/res/blockdata/blocks.txt
krafczyk: done.

forge-gui/res/editions/Edge of Eternities.txt
krafczyk: removed.

forge-gui/res/editions/Edge of Eternities.txt
krafczyk: fixed.

krafczyk:
@paulsnoops latest issues addressed.

paulsnoops:
Thanks for your contribution!

@tehdiplomat I've looked it over and all issues I saw have been fixed but I'm not so familiar with the slots, could you take a look?

forge-gui/res/editions/Edge of Eternities.txt
```diff
@@ -4,6 +4,53 @@ Date=2025-08-01
 Name=Edge of Eternities
 Type=Expansion
 ScryfallCode=EOE
+#https://mtgscribe.com/2025/07/16/edge-of-eternities-play-booster-fact-sheet/
+BoosterSlots=Common,Common-Guest,Uncommon,RareMythic,AnyLand,Wildcard,WildcardFoil
+BoosterSlots=Common,Common-Guest,Uncommon,RareMythic,AnyLand,Wildcard,WildcardFoil
```
tehdiplomat: So you listed this twice. We can remove one of these lines. 

Also you are defining the Booster Slot "Wildcard" and "WildcardFoil" but you actually refer to them as "WildcardA" and "WildcardB" down below. The first two names make more sense to me. Similarly, you are defining an "AnyLand" slot, but the slot you are using below you call "BasicLand". Doesn't matter which you call it, but the name needs to be consistent in the BooserSlot line and the sets created below.

Did you try opening Forge up with these files and running a draft? I would expect things to break without the suggested changes?

krafczyk: I fixed the double list.

I just forgot to update the names. I guess the slot names 'Wildcard' and 'WildcardFoil' always kind of bothered me. At least here they have different card distributions. And the WildcardFoil has to have the + after it since the name 'Foil' doesn't do anything. It's just a preference to make clear they're different distributions and not just a foil version of the same thing.

In any case, I changed it back to the usual style because it isn't that important.

You're absolutely right, I forgot to run the draft. I've done that and corrected the issues which came up. It ran without crashing and seemed right to me. Currently some card arts aren't getting the EOE versions (for example virus beetle), and an SPG card came up, but it fetched a very old card art of that. I don't know if this is due to anything in these sheets.

tehdiplomat: Alright. we can merge and figure those problems out next.

tehdiplomat (approved):

krafczyk:
Thanks @tehdiplomat and @paulsnoops for your help!

PR #8165, opened 2025-07-22
Add g_1_1_insect_flying token (EOC)

Eradev:


tehdiplomat (approved):

Hanmac:
Isn't that a bug? The EOC Insect token should be black 
Not Green

Eradev:
Oh indeed. I'll fix it.

Hanmac:
I fixed the Edition, the TokenScript can be removed again

PR #8167, opened 2025-07-23
CopyPermanentEffect: fix DefinedName TokenKey

Hanmac:
This should fix the broken ImageKey for DefinedName (Tarmagoyf)


tool4ever (approved):
all this imagekey logic seems weird

Hanmac:
> all this imagekey logic seems weird

And that is before my upcoming changes with Record ImageKey

PR #8189, opened 2025-07-24
add nyx effect to desktop card renders

### LINKED ISSUE #7222: Nyx Frame for all Enchantments
With Foundations, all enchantments have the Nyx Frame now

So this can either be simplified or removed:
https://github.com/Card-Forge/forge/blob/8580108d1e4dc5a68819a31351529792c4a5bfa1/forge-game/src/main/java/forge/game/card/CardView.java#L1811-L1815

And it seems the desktop doesn't use Nyx Background yet? Or Background at all?

kevlahnota:
- closes #7222


Agetian:
Looks good to me! :+1:

PR #8190, opened 2025-07-24
Fix Offering being an AltCost

tool4ever:
While also linking the cost would be nice it seems rather messy since the decision which card is the offering needs to happen earlier than normal.

But this still moves us one step in the right direction :)


Hanmac:
While i would like to get rid of `getSacrificedAsOffering` in some way or another, it looks good enough for now

PR #8202, opened 2025-07-25
Added the -c flag to determine maximum simulation time

humra:


forge-gui-desktop/src/main/java/forge/view/SimulateMatch.java
```diff
@@ -130,6 +130,10 @@ public static void simulate(String[] args) {
             }
         }
 
+        if (params.containsKey("c")) {
+            rules.setSimTimeout(Integer.parseInt(params.get("c").get(0)));
```
tehdiplomat, line 134: Why did you decide to attach simTimeout to gameRules instead of inside Simulatematch?

humra: I figured that the duration would fall under gameRules parameters as they hold the type of information I found similar in function, and it also allowed the simulateSingleMatch() to remain the same apart from the hard-coded 120 second value. 
Also, this way if somebody else would want to use the match duration elsewhere it makes it easier for them to access it.

tool4ever (approved):
thanks

PR #8206, opened 2025-07-25
16x16 avatars for every enemy

lemtom:
To keep scaling more consistent, I went over every enemy that did not have a proper 16x16 avatar, and created a new one.

For enemies with big sprites, this is sometimes just a carefully selected snippet of their sprite. Sometimes I started from downscaling a too large avatar, or upscaling the sprite (and then manually cleaning things up). Others are based on existing avatars, but often I just worked from scratch. I generally tried to use the overworld sprites as the main inspiration.

There is definitely still room for improvement, but I hope that this can be a good first step towards a more consistent presentation.


Simisays (approved):
Awesome 👍 !

PR #8228, opened 2025-07-28
Updating Shandalar Old Border to fix the land shops. Also adding addi…

shenshinoman:
Updating Shandalar Old Border to fix the land shops. Also adding additional scene data, and language labels, to make it more clear that the decks used in Old Border are not jumpstart decks but instead official WotC starter decks from the different editions. Both these things have been reported as bug threads in the Discord, so this pull request is meant to fix the actual bug with land shops, and resolve future confusion on the decks. The "help" button when the new decks are selected in old Border explains the new "classic starter" deck type when in Old Border.


shenshinoman:
Oh, hey look at that. A bunch of new text was added to the various properties as I was adding as well... let me go back and merge in the original files and then add my one line.

forge-gui/res/adventure/Shandalar Old Border/config.json
```diff
@@ -135,7 +135,7 @@
       "rewardMaxFactor" : 1.5,
       "sellFactor": 0.6,  
       "shardSellRatio": 0.95,
-      "starterDecks": {
+      "oldStarterDecks": {
```
tehdiplomat, line 138: So what does `Starter decks` mean for this world?

shenshinoman: so, most planes `starterDecks` are decks built using Jumpstart packs. When on the new game screen, there is an info button even, that if you click on it even pops up and says as much. "Old Border" does not utilize jumpstarts and instead has a series of deck files that are faithful recreations of WotC's old "starter decks" from the timeframes covered. This has lead to folks being confused and even dropping in bug reports on the discord, and RemusLord's original repo over this fact.

This adds a different category that when clicking the information button explicitly states they are the WotC starter decks, and what sets they are from based on the chosen difficulty. (As this changes between easy, medium, hard, and insane.) It also makes a different label name in the new game scene. You also will _not_ see the usual "starter" decks option in the decks while on Old Border.

shenshinoman: an example of what I mean with starter decks.

https://mtg.fandom.com/wiki/Seventh_Edition/2-Player_Starter_Set

they are, literally, called starter decks for most of the sets. Some were called "theme decks". They were built by RemusLord in his original version. This update is more to make it clear to the user what they are getting.

tehdiplomat: Maybe i just don't like the name "OldStarterDecks" it seems to just be talking about Theme decks, but also the decks under Constructed also seem like Theme decks? What is the difference there?

tehdiplomat: Maybe we can just call it "alternateConstructedDecks". It seems weird that we have two different constructed deck types, and they aren't just lumped together though. (And randomly chosen between)

Jetz72: Is there any practical difference between starter decks and old border starter decks besides the descriptions and labels? Could we just make the descriptions for the starter decks customizable from the config file instead of adding a whole new category that has different descriptions hard-coded in?

shenshinoman: Having had a greater chance to chat with Remus. it does in fact look like all the decks are theme decks. The limitation came in with how adventure mode currently does one deck, per color, per difficulty level. Meanwhile the original jumpstart decks weren't using jumpstart packs so he put the remaining twenty there as well. So, functionally, it'd be as effective as saying "Standard 2" as in a second filtering of them.

As for the deck categories, and description for the categories. Those are all hard-coded in. It hasn't been a problem before, since we just had Shandalar, and the descriptions added only a couple years ago. (I was around when it happened.) The pull here was mostly working within the framework that was already in use. We can look to trying to define them in the config file, (or if we want to split things up a bit, a new separate file.)

shenshinoman: Addendum: being able to randomly select between multiple decks per color choice as described above would, probably, be very useful for future customization of planes as a whole. However any attempt I have tried to do so has failed. That is likely just be due to the limits in my own abilities.

shenshinoman: As the broken land shops are an actual bug, and are entirely contained within the files under Shandalar Old Border as a directory that doesn't directly impact any of the discussion here. I'm going to go ahead and split that out to it's own pull request.

Issue #5374, opened 2024-06-05
HybridMana Cost choose before Cost Adjustment


Hanmac:
From 601.2b:

> If a cost that will be paid as the spell is being cast includes hybrid mana symbols, the player announces the nonhybrid equivalent cost they intend to pay. If a cost that will be paid as the spell is being cast includes Phyrexian mana symbols, the player announces whether they intend to pay 2 life or a corresponding colored mana cost for each of those symbols.

That would help with other Cost Adjustment if the ManaCost is already "normalized"

tool4ever:
This might be difficult for AI again. To start working on this I'd need some good ideas how it could make the decisions at that point without messing up its payment chances 🤔 
Or maybe it would still be acceptable to let it pay the old way at first?

Hanmac:
I think we should first think on how it could be done Game Mechanics wise
(And Gui dialog)

Connecting that to AI later will probably be the easier part

tool4ever:
Ideally the dialog would only show if the user wants to override which half gets picked
that means it doesn't need to be that pretty :p

Hanmac:
IF we could get the AI ready, we could try to Automate it with AI trying to pick the Most favorite outcome

Issue #4437, opened 2023-12-27
Cost payment issue with Lion's Eye Diamond


TheLastNarwhal:
**Describe the bug**
I say this is a cost issue with LED, but really it's probably a cost issue with any effect that has multiple costs that can't be rolled back.

When you start to pay the costs for activating LED it first asks you if you want to discard all the cards in your hand, if you say yes your hand is immediately discarded. It then asks you if you want to sacrifice LED, now normally this isn't an issue, because you will say yes, but if you say no then the whole process stops except that your hand is still gone.

**To Reproduce**
1. LED on the field.
2. Activate LED.
3. Say yes to the discard, but no to the sacrifice.
4. You've now discarded your hand as a cost and can do this whenever you want at instant speed.

**Expected behavior**
To be forced to continue with the cost payment if you've said yes to the discard, because at that point you're commited to the activation.

Hanmac:
General Problem with Mana Rollback

Technically, you also should be able to pay the costs in any order.

TheLastNarwhal:
This sounds like a deeper problem than I had assumed. I'm guessing that there isn't really an easy fix to this other than to just not say no?

PR #8229, opened 2025-07-28
Bugfix/phyrexian mana

darkorbit17493:
This pull request fixes multiple bugs related to phyrexian mana in Forge.

1. An NPE caused by having no sources for mana shards and attempting to pay a cost that has multiple phyrexian mana colors
2. Incomplete phyrexian mana payment: This bug occured whenever the player attempted to pay a cost which has both mana part and another part such as discard, exile from grave etc. When the player chose to pay life for phyrexian mana the life was instantly paid which allowed for the player to then cancel the payment for the second non-mana part of the cost. This allowed for the player to illegally pay their whole life by constantly paying phyrexian life and cancelling the other cost parts. This can be observed and recreated using the acctivated ability of the card "Drivnod, Carnage Dominus" and rejecting to exile creatures from the graveyard after paying life for the {B/P} cost.
3. Similarly to bugfix 2 there was the same bug caused for the autopayment option as well since the autopay iterates through each mana shard from the cost and tries to pay it. For phyrexian mana shard it immediately pays the 2 life regardless of whether the full cost could be paid or if the next cost part was declined.
4. The autopayment did not validate the phyrexian mana at all and it immediately tried to pay the cost resulting in payment of life with mana shards still remaining to be paid. The method responsible for both validation and payment is called "payManaCost()" and has a boolean parameter "test" which if set to true validates and if set to false pays the mana cost. Code for when it is true has been added so now the autopay rejects costs with phyrexian mana if the player does not have enough life to pay.
5. The priority for phyrexian mana was set to the lowest so that the game chooses to pay it last for autopayment. This is because before the game tried to pay phyrexian mana first using mana instead of life which resulted in the autopayment not working for situations in which it should have. Example: if you had 4 or 5 mountains on the battlefield and a "Moltensteel Dragon" ({4}{R/P}{R/P}) in your hand. Previously when attempting to cast it the autopay button would be greyed out since the autopayer would first attempt to pay the {R/P} costs resulting in having generic mana costs remaining which are impossible to pay with life. Now the other shards are paid first and only if there aren't enough mana sources to pay for the phyrexian mana then life is paid.

In Summary: Phyrexian mana is now paid after all costs have been legally declared so no life is illegaly paid. Validation is fixed. Autopayment is fixed.


forge-game/src/main/java/forge/game/cost/CostPayment.java
```diff
@@ -168,6 +182,7 @@ public boolean payCost(final CostDecisionMakerBase decisionMaker) {
             }
         }
 
+        this.payPhyrexianLife(decisionMaker);
```
tool4ever, line 185: thanks for looking into this
unfortunately costs are allowed to be paid in any order, so always moving this to the very end only trades one problem for another :/
see #309 for our goal to fixing it cleanly
see #4437 for the bigger problem of reverting complex cost parts - for this the "undo restore" option is still being tested

forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
```diff
@@ -752,8 +754,9 @@ private static boolean payManaCost(final ManaCostBeingPaid cost, final SpellAbil
             }
 
             if (saPayment == null) {
-                if ((!toPay.isPhyrexian() && !lifeInsteadOfBlack) || !ai.canPayLife(2, false, sa)
-                        || (ai.getLife() <= 2 && !ai.cantLoseForZeroOrLessLife())) {
+                phyLifeToPay += 2;
+                if ((!toPay.isPhyrexian() && !lifeInsteadOfBlack) || !ai.canPayLife(phyLifeToPay, false, sa)
```
tool4ever, line 758: this looks like a nice find
the AI was indeed in risk of attempting to pay multiple shards with more life than available

forge-game/src/main/java/forge/game/cost/CostPayment.java
darkorbit17493: I see. I haven't really considered other complex costs such as for "Lion's Eye Diamond". I thought the problem was way narrower than it is. Would you be interested in a modified solution of what I've done to have the class Cost (or potentially a helper class) to memorize the player's payment choices and to then apply them after everything has been chosen/validated in an order the player has also chosen ? This way nothing even has to be undone/restored because nothing will be done in the first place. Only the intentions of the player will be remembered. It also would not need extra UI for those choices apart for the UI for payment order.

tool4ever: Unfortunately that would also break some scenarios that have decisions that could depend on each other, e.g.

activating the _Coin of Fate_ ability on an copy from _Clever Impersonator_
he should be able to sacrifice and then exile himself
(already works if you use `FullControlFlag.ChooseCostOrder`)

Hanmac: We need some gui help,
But #5374 for Hybrid and Phyrexian Mana could be interesting for replacing Phyrexian Mana with PayLife early 

@kevlahnota we need to think about some GUI for this

darkorbit17493: @tool4ever I think I bit more than I can chew since the problem is bigger than I thought and there probably isn't really a simple solution as costs can interact with one another as you've shown with copying a card like "Coin of Fate". If I think of something maybe I could open another pull request to propose another solution. But now since this is my first pull request what is usually done in this situation ? Should I close the pull request since I read that it can be referenced/reopened in the future while closed if needed. Or is this decided/done by maintainers ?

tool4ever: hehe no worries :)
yea feel free to close it

PR #8231, opened 2025-07-28
Resolving the land shop issue in Shandalar Old World.

shenshinoman:
The land shops were desired to be a bit more mixed and random for Shandalar Old Border while RemusLord was working on it. When they were brought into the main system, the new shop logic he built was reliant on java files not in core forge. Instead of adding entirely new java files for something that was only being used by the one world. I built up a series of shop decks and accessed via it's own shops.json to use them, and Remus approved the final result. This will have zero impact outside Shandalar Old Border.


Agetian:
Looks like a good idea :) What do you think @Simisays ?

PR #8234, opened 2025-07-29
Streamline logic so canPlay always before canPay

tool4ever:
I wouldn't do it if it only simplified the code again since this was changed in 726976f

but performance-wise I'm seeing nothing negative, especially since the extra guards for targeted SA or Ward don't have to be considered now :)

Also cherry-picked fix from #8229


tool4ever:
it fixes all `ReduceCost.*ValidTarget$` (around 40 cards)

e.g. the AI would never have been able to recognize it can activate _Ghostfire Blade_ if only the cost reduction made it payable

PR #8235, opened 2025-07-29
Android dev build

krafczyk:
Enables a development build of the android apk. Developers can build the dev version using the maven target `android-test-build`. Here's my build command:

```
mvn \
  -U -B -P android-test-build install -e \
  -Dmaven.repo.local="${FORGE_PROJECT_ROOT}/maven_cache" \
  -Dandroid.sdk.path="$ANDROID_SDK_ROOT" \
  -Dandroid.buildToolsVersion=35.0.0 \
  -Dmaven.test.skip=true
```

The dev apk is installable along side the release apk and doesn't collide with it's resources enabling developers to test changes they're making without uninstalling the release version or interfering with their save data. Developers can also uninstall the dev apk version without concern when trying a new version.

* Layouts/asset references now use R to deal with changes in locations
  between the release and dev build
* getContext() is removed in favor of calling methods directly with this
  or implicit this.
* Manifest is adjusted to give a new name and application ID to this new
  dev version.


Hanmac:
Have you checked these Exception I got with Sentry?

NullPointerException
> RuntimeException: Unable to resume activity {forge.app.dev/forge.app.Main}: java.lang.NullPointerException: Attempt to invoke interface method 'void com.badlogic.gdx.backends.android.AndroidInput.onResume()' on a null object reference at android.app.ActivityThread.performResumeActivity(ActivityThread.java:5899) at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:5972) at android.app.servertransaction.ResumeActivityItem.execute(ResumeActivityItem.java:57) at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:60) at android.app.servertransaction.TransactionExecutor.executeLifecycleItem(TransactionExecutor.java:225)

SuperNotCalledException
> SuperNotCalledException: Activity {forge.app.dev/forge.app.Main} did not call through to super.onResume() at android.app.Activity.performResume(Activity.java:9539) at android.app.ActivityThread.performResumeActivity(ActivityThread.java:5869) at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:5972) at android.app.servertransaction.ResumeActivityItem.execute(ResumeActivityItem.java:57) at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:60) ... (11 additional frame(s) were not displayed)
> RuntimeException: Unable to resume activity {forge.app.dev/forge.app.Main}: android.util.SuperNotCalledException: Activity {forge.app.dev/forge.app.Main} did not call through to super.onResume() at android.app.ActivityThread.performResumeActivity(ActivityThread.java:5899) at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:5972) at android.app.servertransaction.ResumeActivityItem.execute(ResumeActivityItem.java:57) at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:60) at android.app.servertransaction.TransactionExecutor.executeLifecycleItem(TransactionExecutor.java:225) ... (10 additional frame(s) were not displayed)

krafczyk:
I did hit these issues and got around them by using the 'R' variable to find the various assets. What build did you make/test? What command did you run for the build so I can replicate it?

Hanmac:
@krafczyk I got this messages via Sentry
Means probably the ones that you got while testing

krafczyk:
@Hanmac , please bear with me. I'm not an android dev so there are many things I don't know. I did get these errors while developing this commit. They came after I installed to my phone with `adb` then tried to launch by touching the icon. The app would crash instantly, and I used logcat to get the stacktrace which looked similarly to what you have.

I fixed this issue with some changes in `Main.java`. Now I can open the dev app and run it normally without it crashing and stacktraces appearing in logcat.

I'm not sure what you mean by 'messages via Sentry'. I'm developing using a text editor, and I build/deploy to my phone using maven/adb directly in the command line.

Where would I get these Sentry messages?

Hanmac:
Check your options

"Report to Sentry" was enabled 
So while you were testing, and getting these exceptions,
I got emails for them

If you fixed these problems, that's fine by me

krafczyk:
Oh are you the only one that can see these messages? I'd there any way I could see them? I could bell now confident the issue is resolved that way.

krafczyk:
I'm not finding a 'Report to Sentry' option in the game settings menu. I'm also not finding it looking for the string "Report to Sentry" with `grep -ri "Report to Sentry" *` in the root of the repo.

Is there somewhere else I need to look?

krafczyk:
Additionally, I'd like to test that the release build of this branch launches properly, but I don't have access to the forge keystore to test it.

krafczyk:
Oh is it the 'Automatic Bug Reports' option on mobile? because it's unchecked for me.

Hanmac:
i think it is fine

@kevlahnota and @Agetian are the ones for the Android builds, so they should check out the other changes

forge-gui-android/src/forge/app/Main.java
```diff
@@ -140,18 +140,18 @@ public AndroidAudio createAudio(Context context, AndroidApplicationConfiguration
...
         try {
-            PackageInfo pInfo = getContext().getPackageManager().getPackageInfo(getContext().getPackageName(), 0);
+            PackageInfo pInfo = getPackageManager().getPackageInfo(getPackageName(), 0);
             versionString = pInfo.versionName;
-        } catch (Exception e) {
+        } catch (PackageManager.NameNotFoundException e) {
             versionString = "0.0";
         }
-        setContentView(getResources().getIdentifier("main", "layout", getPackageName()));
+        setContentView(R.layout.main);
```
kevlahnota, line 148: Why you need to change how the resources are called?

kevlahnota:
@krafczyk 
Changing the resources are called without generating it first will have problems at compiling...

Screenshot_2025-07-30-15-32-04-86_40deb401b9ffe8e1df2f1cc5ba480b12

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project forge-gui-android: Compilation failure: Compilation failure: 
Error:  /home/runner/work/forge/forge/forge-gui-android/src/forge/app/Main.java:[148,25] package R does not exist
Error:  /home/runner/work/forge/forge/forge-gui-android/src/forge/app/Main.java:[151,37] package R does not exist
Error:  /home/runner/work/forge/forge/forge-gui-android/src/forge/app/Main.java:[154,38] package R does not exist
Error:  /home/runner/work/forge/forge/forge-gui-android/src/forge/app/Main.java:[294,39] package R does not exist
Error:  /home/runner/work/forge/forge/forge-gui-android/src/forge/app/Main.java:[295,40] package R does not exist
Error:  -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
Error:  
Error:  After correcting the problems, you can resume the build with the command
Error:    mvn <args> -rf :forge-gui-android
Error: Process completed with exit code 1.
```

The workflow used in running the test build is here: https://github.com/Card-Forge/forge/blob/master/.github/workflows/test-android-build.yml

forge-gui-android/src/forge/app/Main.java
krafczyk: I want these two applications to live side-by-side on the same phone without interfering with eachother. I found the dev application (forge.app.dev) crashed being unable to find its resources. Checking around indicated that using 'R' directly was the best strategy to support multiple build targets like this. Moving to 'R' solved the null pointer crashes I was encountering for this dev build.

I don't know how to build the release version as I don't think I have access to the forge keystore for signing so I dont' think I can build it? I would really appreciate if someone who does can test it.

krafczyk:
@kevlahnota , I see the error now in the github workflow.

I believe this can be fixed by adding a `import forge.app.R;` line, but I removed it because jdtls was complaining that it was redundant to me. But I can add it back to see if it fixes this issue. I did not get this error when I built the `android-test-build` target locally though which is weird to me.

Additionally, I see the `android-test-build` target is being used in the workflow for testing specifically. Do folks think it's reasonable to adjust this target slightly for my purpose like I'm doing here, or would it be better to create a new target like `android-dev-build` or something?

krafczyk:
Strange, I'm not seeing this error at all locally.

kevlahnota:
> @Hanmac , please bear with me. I'm not an android dev so there are many things I don't know. I did get these errors while developing this commit. They came after I installed to my phone with `adb` then tried to launch by touching the icon. The app would crash instantly, and I used logcat to get the stacktrace which looked similarly to what you have.
> 
> I fixed this issue with some changes in `Main.java`. Now I can open the dev app and run it normally without it crashing and stacktraces appearing in logcat.
> 
> I'm not sure what you mean by 'messages via Sentry'. I'm developing using a text editor, and I build/deploy to my phone using maven/adb directly in the command line.
> 
> Where would I get these Sentry messages?

anyway to answer your question, Sentry is a plugin for java that reports bug to the sentry server which devs have access. Now on android, the Lifecycle Activity is also reported if errors are detected. Upon checking, the app you created reported those errors mentioned. 
image

krafczyk:
@kevlahnota ,  Okay, I figured out the issue.

It seems the two always-run workflows are invoking maven without the android tools and profiles enabled, and hitting targets like `test` and `install` which by default attempt to build all modules automatically. These are failing when trying to build the `forge-gui-android` module as the android maven plugin isn't properly initialized and so the `R.java` file isn't being generated.

The latest commit attempts to remedy this by excluding the `forge-gui-android` module from these commands.

There may be a way to exclude this automatically unless an android profile (like `android-test-build` or similar) is active without the user having to specify it on the command line like I've done here.

krafczyk:
@kevlahnota  I've rebased and re-organized these commits.This commits allows a 'multi-forge' installation on the same android phone. Here's a review of the problem, and why the 'R' class is needed. Can you trigger the tests?

Because multi-forge installation requires the app to use the generated
'R' class, if the android maven plugin isn't set up properly, building
the forge-gui-android module will fail with an import error looking for
the 'R' class.

This currently happens if the user runs any default maven target like
'test', 'verify', or 'install' without an android profile enabled. The
android profiles set up the android-maven-plugin, so running `mvn test`
for example without an android profile enabled results in the import
error.

This commit aims to correct this situation by forcing the developer to
specify that they want to build the android components by explicitly
activating the `android` profile. Then we also introduce a variable
called `android.variant` which specifies which type of apk to build. The
default is the new 'dev' build which is built from the
`android-test-build` profile. This way if the user activates the
`android` profile, at least one android apk profile will be active and
if the user doesn't have the android sdk set up properly it will yield a
meaningful error.

We also adjust application id and name of the android-test-build profile
to 'forge.app.dev' and 'Forge (dev)' so it's clear to the user which
version it is.

* Add 'android' profile to the root pom. Activating this profile adds
  the forge-gui-android module to the build and enables the
  `android-test-build` profile by default. the developer can select an
  alternate variant by specifying `-Dandroid.variant=release-build` for
  example to get the release build instead.
* Add `android.variant` variable to control which android profile is
  active.
* Adjust applicationid, application name, and sentry sources to enable
  parallel installation of the 'Forge (dev)' version on the same phone
  as the release version.
* To enable the installation of multiple versions of Forge on the same
  phone, application resource location needs to be robust against
  application id changes among other things. The 'R' class is the
  officially supported method to achieve this.
* Change how some class methods are called to reflect modern usage

krafczyk:
Superseded by https://github.com/Card-Forge/forge/pull/8352.

PR #8237, opened 2025-07-29
Refactor Foil Effect for Mobile

kevlahnota:
Should fix foil rendering for Full, Crop and Art renders
Rotation of zoomed cards and magnifier will reflect the UI_ROTATE preference

### DIFF
--- a/forge-gui-mobile/src/forge/Graphics.java
+++ b/forge-gui-mobile/src/forge/Graphics.java
@@ -18,6 +18,7 @@ import com.badlogic.gdx.math.Rectangle;
 import com.badlogic.gdx.math.Vector2;
 import com.badlogic.gdx.math.Vector3;
 import com.badlogic.gdx.scenes.scene2d.utils.ScissorStack;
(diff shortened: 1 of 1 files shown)


forge-gui-mobile/src/forge/Graphics.java
```diff
@@ -106,13 +107,28 @@ public void end() {
...
-        batch.dispose();
-        shapeRenderer.dispose();
-        shaderOutline.dispose();
-        shaderGrayscale.dispose();
-        shaderUnderwater.dispose();
-        shaderWarp.dispose();
-        if (dummyTexture != null) dummyTexture.dispose();
+        try {
+            batch.dispose();
+        } catch (Exception ignored) {}
```
Hanmac: Is there a better way than this many amount of try-catch?

Like this?
```
public void dispose () {
		for (Disposable disposable : disposables) {
			disposable.dispose();
		}
	}
```

kevlahnota: done

PR #8238, opened 2025-07-29
Cleanup API handling

tool4ever:


tehdiplomat:
I wonder if we should add some rudimentary tests for casting spells into a Ward permanent?

Hanmac:
More Tests would be great (if they don't get into a timeout)

tehdiplomat:
https://github.com/Card-Forge/forge/issues/8239 We can tackle it later unless @tool4ever feels like doing it now.

PR #8243, opened 2025-07-29
 Clean up redundant "AI" from function names

tool4ever:


Jetz72:
Could we explore some more descriptive names for these methods? Is `chkDrawback` a holdover from when sub-abilities were referred to as "drawbacks" everywhere? `doTrigger` is also a little unclear but that kinda stems from the fact that the method kinda shares the responsibilities of "do I want to do this?" and "how should I do this if I have to?".

tool4ever:
@Jetz72
I'd love to roll out some more consistent name scheme but so far I couldn't come up with any that kept things precise enough without misleading (my desire for perfection might not help here 😅)

right now we have:
- "can" vs. "will" but their meaning isn't strictly followed
- "check" somewhat generic
- "do" really not making things clearer

That said, I kind of like the different start of these names since it helps me when quickly comparing different API to see which logic area I'm looking at. I guess we could move the respective "Entity" to the front, though that would also break from the standard "verb first". But maybe staring at these for too long has just caused me some overthinking 🤯

Maybe we should try to eventually write some more Javadoc for this class on how these functions are *ideally* meant to be used? That might help with making sure the names fit... 🤔

PR #8264, opened 2025-07-31
Align invalid Player property result with Card (for Lozhan)

tool4ever:


forge-game/src/main/java/forge/game/player/PlayerProperty.java
```diff
@@ -490,7 +490,9 @@ else if (kind.startsWith("Type")) {
                 }
             }
             return false;
-        }  
+        } else {
+            return false;
```
tehdiplomat, line 494: Does this mean we're always returning false instead of returning true?

Hanmac: Its also true for Card->CardState that does return false for Properties that doesn't exist

Hanmac (approved):

forge-game/src/main/java/forge/game/player/PlayerProperty.java
tool4ever: I tried to think of a scenario where we'd want that but came up short

PR #8278, opened 2025-08-01
Fixing Ultima, Origin of Oblivion

### LINKED ISSUE #8257: AI Uses Ultima on Own Lands
Can you make it so the AI reads the Blight Counters from Ultima as a downside so it doesn't nuke its own utility lands please? I'm having this issue on the latest stable build. There is rarely a scenario where an Ultima deck wants to make its own lands into Wastes, so I would really appreciate it if this was fixed so I can test against this deck more easily.

dracontes:
Closes #8257 .

Confirmed AI behavior in bug report (2025-07-19 build):
- AI focuses targeting of attack trigger on own lands probably because the ramping self aspect is likely weighed well above denying other player resources.
- AI does this targeting non-systematically:
-> targeting lands they control that already have blight counters on them.
-> targeting nonbasic lands they control, especially ones with nonmana activated abilities as was reported as the main issue.

Iterating through fixes, some more or less strict, I came upon the argument for `AITgts$` herein which guides the AI into somewhat more strategic behavior. It, admittedly, doesn't completely avoid the nullification of nonbasic lands the AI controls but it does delay that eventuality substantially.

As for attempted fixes, using `IsCurse$ True` flipped the switch, as it were: the AI focused exclusively on the opponent's resources to the detriment of developing their own.
In either case, with or without `IsCurse$`, there was no effect trying to supplement `AITgts$` with an argument extension that would presumably expand behavior to include both targeting the AI's lands and their opponent's.
Thus I decided to keep this edit simple, and address the salient annoying behavior that belies the strategy the AI prefers. I'm however open to input otherwise.

### DIFF
--- a/forge-gui/res/cardsfolder/u/ultima_origin_of_oblivion.txt
+++ b/forge-gui/res/cardsfolder/u/ultima_origin_of_oblivion.txt
@@ -4,7 +4,7 @@ Types:Legendary Creature God
 PT:4/4
 K:Flying
 T:Mode$ Attacks | ValidCard$ Creature.Self | Execute$ TrigPutCounter | TriggerDescription$ Whenever NICKNAME attacks, put a blight counter on target land. For as long as that land has a blight counter on it, it loses all land types and abilities and has "{T}: Add {C}."
-SVar:TrigPutCounter:DB$ PutCounter | ValidTgts$ Land | CounterType$ BLIGHT | CounterNum$ 1 | AITgts$ Land.YouCtrl+Basic+notnamedWastes+counters_LT1_BLIGHT | SubAbility$ DBEffect
+SVar:TrigPutCounter:DB$ PutCounter | ValidTgts$ Land | CounterType$ BLIGHT | CounterNum$ 1 | AITgts$ Land.YouCtrl+Basic+notnamedWastes+notnamedSnow-Covered Wastes+counters_LT1_BLIGHT | SubAbility$ DBEffect
 SVar:DBEffect:DB$ Effect | RememberObjects$ Targeted | StaticAbilities$ BlightStatic | ForgetOnMoved$ Battlefield | ForgetCounter$ BLIGHT | Duration$ Permanent
 SVar:BlightStatic:Mode$ Continuous | Affected$ Card.IsRemembered | RemoveLandTypes$ True | RemoveAllAbilities$ True | AddAbility$ ColorlessMana | Description$ For as long as that land has a blight counter on it, it loses all land types and abilities and has "{T}: Add {C}."
 SVar:ColorlessMana:AB$ Mana | Cost$ T | Produced$ C | SpellDescription$ Add {C}.
--- a/forge-gui/res/cardsfolder/u/ultima_origin_of_oblivion.txt
+++ b/forge-gui/res/cardsfolder/u/ultima_origin_of_oblivion.txt
@@ -4,7 +4,7 @@ Types:Legendary Creature God
 PT:4/4
 K:Flying
 T:Mode$ Attacks | ValidCard$ Creature.Self | Execute$ TrigPutCounter | TriggerDescription$ Whenever NICKNAME attacks, put a blight counter on target land. For as long as that land has a blight counter on it, it loses all land types and abilities and has "{T}: Add {C}."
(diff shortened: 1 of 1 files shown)


Hanmac:
What if you add `,Land.OppCtrl+!Basic+counters_LT1_BLIGHT` ?

So it targets Basic you control & NonBasic Opponent Controls?

(until we fix it further?)

dracontes:
> What if you add `,Land.OppCtrl+!Basic+counters_LT1_BLIGHT` ?

I already attempted a variant of that (with `nonBasic` instead of `!Basic`; those were the argument extensions mentioned), but I did test your variant for comprehensiveness. Unfortunately, the AI keeps targeting their own lands alone, even favoring their nonbasics over the opponent's nonbasics.

image

thejunecooperative:
I posted something about this in the original comment since I posted the original issue, but considering I typically play Commander, when the AI pilots my Ultima deck it still targets its own lands that are colorless, this is why it was problematic for it to target its own nonbasics, because in Commander you'd only have colorless lands in a colorless Commander deck. You COULD put him in the 99 of a Commander deck or in a noncolorless constructed deck, but the issue I'm having is it will target its own colorless lands, even nonbasics with good abilities that are colorless like Ancient Tomb (which yes has a downside but it's just an example, it won't care about hitting any regardless of whether it's all upside). I can see how targeting your own lands to ramp you is sensible behavior in a non-colorless deck, but in the deck I was facing it results in it ruining its own manabase without ever burdening its own opponents of which there are many. If it were to prioritize colored lands this would be much closer to working as intended for my use case, because it's a colorless deck. It would prevent it from hitting things like Field of the Dead on opponents' board, but at least it would be targeting lands that are harmed by this before it hits its own in Commander if it doesn't need to ramp. Focusing away from nonbasics that are colored for example is actually not much better in this instance, so I suspect if you made it prioritize colored lands it would behave more rationally in Commander without sacrificing its ramp in a 1v1 context! You could even make it prioritize highest multicolor lands if you wanted which would make it focus on removing color fixing which is a viable strategy in my deck that I often pursue, and it would ideally not lead to too much weird behavior, and it wouldn't needlessly target monocolor deck's lands first since they are rarely going to get color screwed by this even though its possible. (Or you could make it also prioritize colored nonbasics to still hit stuff like Valakut, Molten Pinnacle, the ball is entirely in your court.)

dracontes:
_Quoting from the [issue comment](https://github.com/Card-Forge/forge/issues/8257#issuecomment-3145343260) which covers about the same ground as the comment above._

> (I forget whether it would affect supertypes)

No, it wouldn't: the triggers' effect is only removing land (sub)types.

> So unless Ultima is played in the 99 of a Commander deck, steals lands from other players (such as with Nazgul Battlemace- which I have.), or in another format then it's only negative to hit their own lands.

I understand it's frustrating but the other formats and other lines of play do exist and the expectation is those should be considered. On the other hand, Ultima might in the end be determined to not be something the AI can handle as it is and the card'll be flagged accordingly, until eventually someone gets around to do a deeper dive on it.

> thus if you have a Wastes and target it that

Fair catch on Wastes, regarding the current focus on basic lands for the AI targeting.

> Could you possibly check to see if the lands tap for colorless before targeting them? 

It's a possibility, though I think the focus on basic lands for Ultima's controller's side (excluding Wastes) directs behavior as well as it can be directed. I'll give that option a test though.
ETA: Turns out this is the terser approach.

Hanmac (approved):
We should later fix the opponent's lands too

thejunecooperative:
I was just hoping by checking for colorless you'd have the best of both worlds where it ramps you when necessary and color screws the opponents. I brought up Commander because I use Forge for goldfishing and the context is important, but I did try to recommend a solution that doesn't eliminate the play lines you recommended, since it wouldn't be used that way outside of limited. Granted, it's very handy in draft or adventure mode so this is important I agree. I just hope this leads to a better play experience, but once again I'm just trying to find a balance where Ultima doesn't harm its own nonbasics without sacrificing the obvious use case in limited, and I figured targeting colored lands is probably the most consistently logical use case? I hope this works but if it still seems like it doesn't I'll likely let you all know? Anyway thank you for getting to this!

PR #8288, opened 2025-08-02
Fix scripts

### LINKED ISSUE #8285: Shinryu's extra win effect is not triggered.
**Describe the bug**
Shinryu, Transcendent Rival's extra win effect that worked in previous version did not trigger when I tried it today.

**To Reproduce**
1. Transform to Shinryu, then defeat the opponent.

**Expected behavior**
Win the game.

 - OS: Windows 10
 - Browser Firefox
 - Version forge-installer-2.0.06-SNAPSHOT-08.01

tool4ever:
Closes #8285

### DIFF
--- a/forge-gui/res/cardsfolder/c/cavalcade_of_calamity.txt
+++ b/forge-gui/res/cardsfolder/c/cavalcade_of_calamity.txt
@@ -2,6 +2,6 @@ Name:Cavalcade of Calamity
 ManaCost:1 R
 Types:Enchantment
 T:Mode$ Attacks | ValidCard$ Creature.powerLE1+YouCtrl | TriggerZones$ Battlefield | Execute$ TrigDamage | TriggerDescription$ Whenever a creature you control with power 1 or less attacks, CARDNAME deals 1 damage to the player or planeswalker that creature is attacking.
-SVar:TrigDamage:DB$ DealDamage | Defined$ Player.TriggeredDefender & Valid Planeswalker.TriggeredDefender | NumDmg$ 1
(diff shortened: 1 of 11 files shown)


forge-gui/res/cardsfolder/c/cavalcade_of_calamity.txt
```diff
@@ -2,6 +2,6 @@ Name:Cavalcade of Calamity
 ManaCost:1 R
 Types:Enchantment
 T:Mode$ Attacks | ValidCard$ Creature.powerLE1+YouCtrl | TriggerZones$ Battlefield | Execute$ TrigDamage | TriggerDescription$ Whenever a creature you control with power 1 or less attacks, CARDNAME deals 1 damage to the player or planeswalker that creature is attacking.
-SVar:TrigDamage:DB$ DealDamage | Defined$ TriggeredDefender.Player & Valid Planeswalker.TriggeredDefender | NumDmg$ 1
+SVar:TrigDamage:DB$ DealDamage | Defined$ Player.TriggeredDefender & Valid Planeswalker.TriggeredDefender | NumDmg$ 1
```
Hanmac: what is the reason there?

* `TriggeredDefender.Player` should be getDefined("TriggeredDefender") + Property Player
* `Player.TriggeredDefender` while this is all Player + TriggeredDefender check

imo the first one should be faster?
(i think we miss the `Player` property?)

tool4ever: yea neither `Card.Card` nor `Player.Player` are valid syntax

feels stupid to change that imo

you'd prefer `TriggeredDefender.Opponent` instead?

Hanmac: yeah, `TriggeredDefender.Opponent` is fine too

PR #8292, opened 2025-08-03
Draft options

tehdiplomat:


tehdiplomat:
I think we'll let https://github.com/Card-Forge/forge/pull/7519/files merge before this one.

PR #8302, opened 2025-08-04
Fix AI timeout still keeping Future in calculation even when infinite loop

tool4ever:
`Thread.stop()` is somewhat brutal but still safer than having e.g. AI api calling ``checkStaticAbilities()`` after we're already back in engine code 💣


Hanmac:
@tool4ever 
> UnsupportedOperationException: None
    at java.lang.Thread.stop(Thread.java:1079)
    at forge.ai.AiController.chooseSpellAbilityToPlayFromList(AiController.java:1751)

PR #8305, opened 2025-08-04
CardStateName: combine Modal with Transformed

Hanmac:
make Modal and Transform use the same Backside CardStateName

we can still differ between Modal and Transform cards via the SplitState

but now we can extend the function to include some Modal cards too

### DIFF
--- a/forge-core/src/main/java/forge/card/CardRules.java
+++ b/forge-core/src/main/java/forge/card/CardRules.java
@@ -175,7 +175,7 @@ public final class CardRules implements ICardCharacteristics {
         }
         for (ICardFace face : getAllFaces()) {
             for (String spell : face.getAbilities()) {
-                if (spell.contains("AB$ Transform")) {
+                if (spell.contains("AB$ SetState") && spell.contains("Mode$ Transform")) {
                     return true;
                 }
             }


Hanmac:
> maybe can get rid of `setBackSide`?

I will need to think about that

It's also for cases when a TDFC becomes a copy of a non-TDFC

then it is still on the backside (transformed) but the CardState should be from the Original of the other one
(but i need to debug this first)

Edit: there is also Meld which also uses setBackSide

Hanmac:
Peter Parker can transform now

forge-core/src/main/java/forge/card/CardRules.java
```diff
@@ -166,6 +166,24 @@ public Iterable<ICardFace> getAllFaces() {
...
+    public boolean isTransformable() {
+        if (CardSplitType.Transform == getSplitType()) {
+            return true;
+        }
+        if (CardSplitType.Modal != getSplitType()) {
+            return false;
+        }
+        for (ICardFace face : getAllFaces()) {
+            for (String spell : face.getAbilities()) {
+                if (spell.contains("AB$ Transform")) {
```
tool4ever: What about other stuff like _Moonmist_?
I doubt it should be limited to intrinsic...

Hanmac: Ugh, I need to debug!
"AB$ Transform" can't be true for Peter Parker, I wonder what happened?
Probably should have checked for
`spell.contains("AB$ SetState") && spell.contains("Mode$ Transform")`

`isTransformable` is the check if something else is transformable
Moonmist itself is not transformable

Hanmac:
I need to debug why the faulty code let me Transform Peter Parker, but i think that is okay for now

Issue #7887, opened 2025-06-22
[Adventure Mode] Crash when drafting


tuliorcc:
When picking the last card in a draft, the game freezes and crashes. Happens more often in FIN drafts.

Crash Log:

```
Attempting to fetch: https://api.scryfall.com/cards/fin/330/en?format=image&version=normal
C:\Users\user\AppData\Local\Forge\Cache\pics\cards\FIN\The Wind Crystal2.fullborder.jpg
Saved image to C:\Users\user\AppData\Local\Forge\Cache\pics\cards\FIN\The Wind Crystal2.fullborder.jpg
EDT > java.lang.NullPointerException: Cannot invoke "forge.gamemodes.limited.DraftPack.contains(Object)" because "thisBooster" is null
    at forge.gamemodes.limited.BoosterDraft.setChoice(BoosterDraft.java:527)
    at forge.adventure.scene.AdventureDeckEditor$DraftPackPage.afterCardPicked(AdventureDeckEditor.java:105)
    at forge.adventure.scene.AdventureDeckEditor$DraftPackPage$2.run(AdventureDeckEditor.java:129)
    at forge.adventure.scene.AdventureDeckEditor$DraftPackPage$2.run(AdventureDeckEditor.java:125)
    at forge.adventure.scene.AdventureDeckEditor$CardManagerPage.lambda$addItem$6(AdventureDeckEditor.java:937)
    at forge.menu.FMenuItem$1.run(FMenuItem.java:117)
    at com.badlogic.gdx.utils.Timer$TimerThread.runPostedTasks(Timer.java:322)
    at com.badlogic.gdx.utils.Timer$TimerThread$1.run(Timer.java:272)
    at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:197)
    at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:153)
    at forge.app.GameLauncher.<init>(GameLauncher.java:51)
    at forge.app.Main.main(Main.java:30)

```
**Desktop:**
 - OS: Windows 11
 - Version 2.0.05 - Snapshot Jun 20, 2025

Jtgroos:
Same here.

Excactly with FIN Drafts.  crashes always when i pick the last card.

i was so hyped to finaly find a FIN Draft...

Jetz72:
Reportedly still happening. Might be addressed by #7519; will check soon.

[...]

PR #8313, opened 2025-08-06
Collect all of slots paper cards before adding it to the final result

tehdiplomat:


Jetz72 (approved):
Seems to fix the root cause behind #7887.

PR #8320, opened 2025-08-07
Sync wiki from main repo

tool4ever:
- we can update code and docs in the same PR
- new contributors can use the normal way from their fork for reviewing and without us having to manually manage permissions
- it's easier to include them as part of our builds for offline access

- [x] update links to work locally & remote
- [ ] update action settings before merging

### DIFF
--- a/docs/Adventure-Items.md
+++ b/docs/Adventure-Items.md
@@ -1,2 +1,0 @@
-# Adventure Item Data
-(Placeholder, actual content coming soon)
--- a/docs/Adventure-Mode.md
+++ b/docs/Adventure-Mode.md
@@ -5,7 +5,7 @@ Adventure mode is a work-in-progress game mode where you explore the ever-changi
 # How to run?
 
 1. Step 1: Download the latest version of Forge https://downloads.cardforge.org/dailysnapshots/ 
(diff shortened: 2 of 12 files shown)


.github/workflows/sync-wiki.yml
```diff
@@ -0,0 +1,27 @@
...
+      - docs/**
+      - .github/workflows/sync-wiki.yml
+concurrency:
+  group: publish-wiki
+  cancel-in-progress: true
+permissions:
+  contents: write
+jobs:
+  publish-wiki:
+    if: github.repository_owner == 'tool4ever'
```
Hanmac, line 15: shouldn't this be Card-Forge?

tool4ever: yes, I'm deploying it to my fork for now until it's approved so we don't make the transition before it's merged

Hanmac:
Can't wait until we need to update the Card Images Page XD

tool4ever:
I plan to do some more restructuring after it's live

Agetian (dismissed):

Agetian:
Good idea, looks good too! :)

PR #8325, opened 2025-08-07
Fix voting crash

### LINKED ISSUE #8324: Vote breaks NetPlay
```
io.netty.handler.codec.EncoderException: java.io.NotSerializableException: forge.game.ability.effects.InvestigateEffect
	at io.netty.handler.codec.MessageToByteEncoder.write(MessageToByteEncoder.java:125)
```

Caused by #8286

tool4ever:
Closes #8324

### DIFF
--- a/forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
+++ b/forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
@@ -1463,10 +1463,10 @@ public class PlayerControllerHuman extends PlayerController implements IGameCont
     @Override
     public Object vote(final SpellAbility sa, final String prompt, final List<Object> options,
                        final ListMultimap<Object, Player> votes, Player forPlayer, boolean optional) {
-        if (options.get(0) instanceof SpellAbility) {
+        if (sa.hasParam("Choices")) {
             return chooseSpellAbilitiesForEffect(Lists.newArrayList(IterableUtil.filter(options, SpellAbility.class)), sa, prompt, 1, null).get(0);
         }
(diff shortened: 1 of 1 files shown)


Hanmac:
so it did crash for Player/Card too? i thought i didn't changed that?

and you forgot to tag me, thats why i did this: #8331

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1463,10 +1463,10 @@ public String chooseRollSwapValue(List<String> swapChoices, Integer currentResul
     @Override
     public Object vote(final SpellAbility sa, final String prompt, final List<Object> options,
                        final ListMultimap<Object, Player> votes, Player forPlayer, boolean optional) {
-        if (optional) {
-            return getGui().oneOrNone(prompt, options);
+        if (options.get(0) instanceof SpellAbility) {
+            return chooseSpellAbilitiesForEffect((List<SpellAbility>)(Object) options, sa, prompt, 1, null).get(0);
```
Hanmac: this doesn't respect the optional param

tool4ever: it's only used once for Cards so I didn't bother there

tool4ever:
> so it did crash for Player/Card too? i thought i didn't changed that?
> 
> and you forgot to tag me, thats why i did this: #8331

it's likely the other types were already affected

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
Hanmac: I don't like this `List<SpellAbility>)(Object)` cast

Imo a `IterableUtil.filter(options, SpellAbility.class)` would be cleaner

PR #8332, opened 2025-08-08
Fix Phantom Train not updating PT

Hanmac:
This should fix that the PT isn't updated when the Train becomes a creature

might remove the `updatePTforView` from there:
https://github.com/Card-Forge/forge/blob/000fa14c177c2ff6dd15beb743a6665d8bb3db53/forge-game/src/main/java/forge/game/ability/effects/AnimateEffect.java#L203-L206


tool4ever:
yea the extra crew update is redundant now

tool4ever (approved):

PR #8333, opened 2025-08-08
update progressbar, update boosterdraft readfile

kevlahnota:
also removed unneccesary settings since forge uses shaders for round corners for rendering borders


Hanmac:
Maybe in the future to deprecate the `borderlessCardList` file?

PR #8335, opened 2025-08-08
Phase label fix for multiplayer

tovernaar123:
Me and my friends noticed that when playing forge multiplayer only the host has the correct phase labels enabled (others seem to skip main phase 1 and 2), this pull request should fix that. 
This is my first pull request for this project so if you preferer something to be different feel free to ask.


forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
```diff
@@ -1165,83 +1172,55 @@ public boolean isUiSetToSkipPhase(final PlayerView playerTurn, final PhaseType p
...
-        prefs.setPref(FPref.PHASE_HUMAN_FIRSTSTRIKE,      fvHuman.getLblFirstStrike().getEnabled());
-        prefs.setPref(FPref.PHASE_HUMAN_COMBATDAMAGE,     fvHuman.getLblCombatDamage().getEnabled());
-        prefs.setPref(FPref.PHASE_HUMAN_ENDCOMBAT,        fvHuman.getLblEndCombat().getEnabled());
-        prefs.setPref(FPref.PHASE_HUMAN_MAIN2,            fvHuman.getLblMain2().getEnabled());
-        prefs.setPref(FPref.PHASE_HUMAN_EOT,              fvHuman.getLblEndTurn().getEnabled());
-        prefs.setPref(FPref.PHASE_HUMAN_CLEANUP,          fvHuman.getLblCleanup().getEnabled());
+
+        //loops over all fields
+        for (int playerIndex = 0; playerIndex < fieldViews.size(); playerIndex++) {
+            // This is needed for multiplayer games and if there is more than one human in the game (the index is not just 0)
```
tool4ever, line 1193: the comment might be a bit misleading:
this still only supports 1 AI and 1 human before it just overwrites the states right?

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
```diff
@@ -822,6 +827,8 @@ public void finishGame() {
         FloatingZone.closeAll(); //ensure floating card areas cleared and closed after the game
         final GameView gameView = getGameView();
         if (hasLocalPlayers() || gameView.isMatchOver()) {
+            //Otherwise only the host will save in a multiplayer match.
+            writeMatchPreferences();
```
tool4ever: this means in local game preferences will get written an additional time?
should probably have some condition to avoid that...
also is this really only relevant for Desktop?

thanks for looking into this

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
tovernaar123: It should work with more player, it will treat all non controlled players as ai in terms of the active labels. Should I use another default?

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
tovernaar123: Yeah I assume the same problem is also on android but I would not really know how to fix that, is their some shared logic file I could change that would effect both gui's? And yes I should add an extra check to make sure no 2 writes happen

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
tool4ever: I think I understand what you meant now, so let me rephrase:
iterating over all players might imply you can store the phase stops for multiple AI or human. but when it finds a second of either type the preferences are just going to be written in same keys as before

not a big deal you refactored it like that, maybe we'll eventually figure out a way to extend support in this direction

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
tool4ever: After some debugging it seems to me that either host or client gets to save
but it depends on who acts in `ViewWinLose` first

doesn't look like we have a more generic solution available yet that could catch it when that class just stops its displaying

so just a check for netplay would be enough imo, so the double saving is limited to that mode at least

and for android @kevlahnota might later adapt some parts

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
tovernaar123: I added an extra comment to make it more clear that it can still overwrite the data.

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
tovernaar123: I could not really find a nice way to check for a netgame so I have added a simple boolean check I hope that that is fine too.

tool4ever: ugh are you sure a new `CMatchUi` object is created after each game? otherwise this breaks saving after the first one...
what about `GuiBase.isNetworkplay()`?

tool4ever:
@MostCromulent
want to check if this still relevant + make a fresh PR if fixing it is viable? 🙏

MostCromulent:
> want to check if this still relevant + make a fresh PR if fixing it is viable? 🙏

TBH I haven't been able to replicate the problem, but I think that's due to local network testing conditions using two instances of forge reading the same user preferences. The AI assures me this is still an issue with the default preferences on remote network play so I'll look at submitting a separate PR for it.

PR #8336, opened 2025-08-08
Expanding my Innistrad plane, further. 

shenshinoman:
Expanding my Innistrad plane further for public enjoyment. 

- New enemies. 
- New maps. 
- Second biome partially populated. 

Everything is contained to the Innistrad plane and it's folders, so should have no impact on the main game outside that plane.


shenshinoman:
made a few tweaks update to this in the second commit. in order to fix some bugs I've found in my own testing.

PR #8337, opened 2025-08-09
Perpetual: use Record for different types

Hanmac:
This uses Records with applyEffect instead of ambiguous Map

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
@@ -88,7 +88,7 @@ public abstract class AnimateEffectBase extends SpellAbilityEffect {
             final ManaCost incMCost = new ManaCost(new ManaCostParser(sa.getParam("Incorporate")));
             PerpetualIncorporate p = new PerpetualIncorporate(timestamp, incMCost);
             c.addPerpetual(p);
-            p.applyCard(c);
+            p.applyEffect(c);
         }
         
         if (!addType.isEmpty() || !removeType.isEmpty() || addAllCreatureTypes || !remove.isEmpty()) {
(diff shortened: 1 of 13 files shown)


forge-game/src/main/java/forge/game/card/perpetual/PerpetualTypes.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.Set;
+
+import forge.card.CardType;
+import forge.card.RemoveType;
+import forge.game.card.Card;
+
+public record PerpetualTypes(long timestamp, CardType addTypes, CardType removeTypes, Set<RemoveType> removeXTypes) implements PerpetualInterface {
+
+    @Override
+    public long getTimestamp() {
```
tool4ever, line 12: just make base abstract instead to move these?

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
```diff
@@ -294,12 +256,7 @@ static void doUnanimate(final Card c, final long timestamp) {
     static void handleColors(final Card c, final ColorSet colors, final long timestamp, final boolean overwrite, 
                                 final boolean perpetual) {
         if (perpetual) {
-            Map <String, Object> params = new HashMap<>();
-            params.put("Colors", colors);
-            params.put("Overwrite", overwrite);
-            params.put("Timestamp", timestamp);
-            params.put("Category", "Colors");
-            c.addPerpetual(params);
+            c.addPerpetual(new PerpetualColors(timestamp, colors, overwrite));
```
tool4ever: this never gets called now...?

forge-game/src/main/java/forge/game/card/perpetual/PerpetualTypes.java
Hanmac: Records doesn't support that:
https://stackoverflow.com/questions/71139939/why-java-records-do-not-support-inheritance

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
Hanmac: it was still called once for `Legion Reconsecrator`

tool4ever (approved):

PR #8347, opened 2025-08-10
Don't search library while opponent controls Ob Nixilis, Unshackled (#8346)

### LINKED ISSUE #8346: AI searches library while opponent controls Ob Nixilis, Unshackled
AI searches library while opponent controls Ob Nixilis, Unshackled.

I fixed the TODO. Patch coming up.

CTimmerman:
Fixes #8346.

Test result from game log (why is that most recent on top?):

```
Resolve stack: Imprint — Whenever a nontoken creature dies, you may exile that card. If you do, return each other card exiled with Mimic Vat to its owner's graveyard. [Zone Changer: Ob Nixilis, Unshackled (404)]
Resolve stack: Evolving Wilds (14) - Sedar searches their library for a Land.Basic card, puts it onto the battlefield tapped, then shuffles.
Resolve stack: Evolving Wilds (221) - Veldahar searches their library for a Land.Basic card, puts it onto the battlefield tapped, then shuffles.
Add to stack: Veldahar activated Evolving Wilds (221)
Add to stack: Sedar activated Evolving Wilds (14)
Add to stack: Veldahar triggered Mimic Vat (227)
Resolve stack: Murder (43) - Destroy Ob Nixilis, Unshackled (404).
Add to stack: Sedar cast Murder (43) targeting [Ob Nixilis, Unshackled (404)]
```

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
@@ -22,7 +22,10 @@ import forge.game.player.PlayerActionConfirmMode;
 import forge.game.spellability.AbilitySub;
 import forge.game.spellability.SpellAbility;
 import forge.game.spellability.TargetRestrictions;
+import forge.game.staticability.StaticAbility;
 import forge.game.staticability.StaticAbilityMustTarget;
+import forge.game.trigger.Trigger;
+import forge.game.trigger.TriggerType;
 import forge.game.zone.ZoneType;
 import forge.util.Aggregates;
 import forge.util.MyRandom;
@@ -346,7 +349,28 @@ public class ChangeZoneAi extends SpellAbilityAi {
                     return true;
                 });
             }
-            // TODO: prevent ai searching its own library when Ob Nixilis, Unshackled is in play
+
+            // Searching own library
+            if (p == ai && origin != null && origin.size() == 1 && origin.get(0).equals(ZoneType.Library)) {
+                for (final Player op : ai.getOpponents()) {
+                    for (Card c : op.getCardsIn(ZoneType.Battlefield)) {
+                        // Ob Nixilis, Unshackled
+                        for (Trigger trigger : c.getTriggers()) {
+                            if (TriggerType.SearchedLibrary.equals(trigger.getMode())
+                                && "TrigSac".equals(trigger.getParam("Execute"))) {
+                                    return new AiAbilityDecision(0, AiPlayDecision.CantPlayAi);
+                            }
+                        }
+                        // Opposition Agent
+                        for (StaticAbility stAb : c.getStaticAbilities()) {
+                            if ("You".equals(stAb.getParam("ControlOpponentsSearchingLibrary"))) {
+                                return new AiAbilityDecision(0, AiPlayDecision.CantPlayAi);
+                            }
+                        }
+                    }
+                }
+            }
+
             if (origin != null && origin.size() == 1 && origin.get(0).isKnown()) {
                 // FIXME: make this properly interact with several origin zones
                 list = CardLists.getValidCards(list, type, source.getController(), source, sa);


forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
```diff
@@ -380,7 +380,12 @@ private static AiAbilityDecision hiddenOriginCanPlayAI(final Player ai, final Sp
                     return true;
                 });
             }
-            // TODO: prevent ai searching its own library when Ob Nixilis, Unshackled is in play
+            // prevent ai searching its own library when Ob Nixilis, Unshackled is in play
+            for (final Player op : ai.getOpponents()) {
+                if (op.isCardInPlay("Ob Nixilis, Unshackled")) {
```
Hanmac: imo it would be better to check for the Trigger instead
I will try to write a code piece later

tool4ever (changes requested):
checks are once again just too vague, e.g. this would also stop other kind of hidden mode zone changing when the cards are in play

CTimmerman:
> checks are once again just too vague, e.g. this would also stop other kind of hidden mode zone changing when the cards are in play

I did the explicit name first, then rewrote it to be generic to address https://github.com/Card-Forge/forge/pull/8347#discussion_r2265905705 and that matches only the one card last i checked, so I don't see how it's too vague now. Is it my addition of Opposition Agent effects? "ControlOpponentsSearchingLibrary" sounds like something you also don't want to trigger with a library search.

tool4ever:
you're not considering the bigger picture which would cause negative impact: the method is in fact already named "hiddenOriginCanPlayAI" to express it handles way more variants

CTimmerman:
So this patch isn't broad enough? What else should it cover?

tool4ever:
It's possible to search other libraries than your own...

jamincollins:
## Findings
#### 1. Trigger detection relies on Execute parameter string match
The code identifies Ob Nixilis, Unshackled by checking `TriggerType.SearchedLibrary.equals(trigger.getMode()) && "TrigSac".equals(trigger.getParam("Execute"))`. This couples AI logic to internal Forge parameter naming rather than card identity or trigger effect. If Ob Nixilis is re-implemented with a different Execute name, or another SearchedLibrary trigger uses TrigSac for a different purpose, detection will be incorrect.
Test case:
```java
// Simulate a different SearchedLibrary trigger that also uses Execute="TrigSac" but for another effect
Trigger t = new Trigger(Map.of("Mode","SearchedLibrary","Execute","TrigSac"), card, false);
assert trigger.getMode() == TriggerType.SearchedLibrary;
assert "TrigSac".equals(trigger.getParam("Execute"));
// Current code would incorrectly block library search for non-Ob Nixilis trigger
```
Impact: False positives/negatives if parameter names change or are reused.

#### 2. Opposition Agent check uses static ability param equality to "You"
The code checks `stAb.getParam("ControlOpponentsSearchingLibrary")` equals `"You"`. This assumes the param value is exactly "You" and that the static ability is present on battlefield cards. It does not verify the static ability is actually active for the current player or that the ability belongs to Opposition Agent. Other cards could set same param.
Test case:
```java
StaticAbility stAb = card.getStaticAbilities().get(0);
String val = stAb.getParam("ControlOpponentsSearchingLibrary");
assert "You".equals(val); // passes for Opposition Agent, also passes for any card with same param
```
Impact: Potential false blocking.

#### 3. Origin null check and size check may miss multi-zone origins
The guard `origin != null && origin.size() == 1 && origin.get(0).equals(ZoneType.Library)` only blocks when origin is exactly one Library. Abilities that search from multiple zones, or use list with extra zones, will bypass check. Also `origin` could be empty list.
Test case:
```java
List<ZoneType> origin = Arrays.asList(ZoneType.Library, ZoneType.Graveyard);
assert origin.size() == 2;
// Current check fails, AI may search library despite Ob Nixilis
```
Impact: Incomplete protection.

#### 4. Opponent iteration assumes battlefield cards have triggers/static abilities
Loop iterates `op.getCardsIn(ZoneType.Battlefield)` without null check on `op.getCardsIn`. Also `c.getTriggers()` and `c.getStaticAbilities()` may be empty; code is safe but iterates over all cards each time, no early exit per opponent.
Test case:
```java
for (Player op : ai.getOpponents()) {
    for (Card c : op.getCardsIn(ZoneType.Battlefield)) { // what if getCardsIn returns null?
}
```
Impact: Minor performance, potential NPE if API changes.

#### 5. Early return CantPlayAi on first opponent match prevents evaluating other players
The code returns `CantPlayAi` immediately when any opponent has matching trigger/static ability. If `pDefined` contains multiple players, returning early may skip evaluation for other players. Also `p == ai` check is inside loop over pDefined, but return exits entire method.
Test case:
```java
List<Player> pDefined = Arrays.asList(ai, opponent);
for (Player p : pDefined) {
    if (p == ai && ...) { return CantPlayAi; }
    // never reaches opponent player check
}
```
Impact: Could incorrectly block abilities that target opponent library.

#### 6. Check runs after canSearchLibraryWith filter
The library search block runs after `!ai.canSearchLibraryWith(sa, p)` filtering. If `canSearchLibraryWith` already prevents searching, the extra check is redundant. Order may cause double filtering or missed cases where `canSearchLibraryWith` allows but Ob Nixilis should block.
Test case:
```java
boolean allowed = ai.canSearchLibraryWith(sa, ai); // true
// then block check runs, returns CantPlayAi
```
Impact: Interaction unclear; may be correct but order warrants review.

#### 7. No check for SearchOwnLibrary param on trigger
TriggerSearchedLibrary.performTest checks `SearchOwnLibrary` param to ensure target equals player. Code does not verify trigger has `SearchOwnLibrary` true, so may block for triggers that fire when opponent searches opponent's library.
Test case:
```java
// Trigger with Mode=SearchedLibrary, Execute=TrigSac, SearchOwnLibrary=false
// Current code would still block AI searching own library
```
Impact: Over-blocking.

### Performance and security
#### 8. Nested loops over opponents, battlefield cards, triggers, static abilities
Complexity O(opponents * battlefield cards * (triggers + static)). No caching, runs on every ChangeZone AI evaluation. Could be expensive in multiplayer games with large board.
Test case:
```java
// 4 opponents, 60 cards each, 2 triggers per card
// ~ 480 trigger checks per evaluation
```
Impact: Potential performance degradation.

## Verdict
Request Changes
The PR addresses Ob Nixilis and Opposition Agent avoidance but uses brittle parameter string matching, lacks verification of SearchOwnLibrary, and may over-block or miss edge cases. Needs more robust detection and broader coverage.

## Recommended Fixes
1. Use card identity or trigger effect semantics rather than Execute string "TrigSac".
2. Verify trigger has SearchOwnLibrary true and ValidPlayer matches opponent.
3. Check static ability's effective control via game state rather than param string equality.
4. Handle multi-zone origins and ensure check applies only when searching own library.
5. Move check earlier and ensure it doesn't short-circuit evaluation for other players.

---

PR #8350, opened 2025-08-11
reduce adventure music bitrate

kevlahnota:


Agetian:
Why would this be needed, @kevlahnota ? The difference in file size seems rather small, is it really worth reducing the quality of the music for this gain? Any other opinions, @tool4ever @Simisays @Hanmac ?

kevlahnota:
@Agetian I just halved the bitrate since I thought it was needed as @Hanmac sugested on discord. I tried ogg but it was larger than this. Anyway gonna close this since it's not really priority.

PR #8352, opened 2025-08-11
android-dev-build

krafczyk:
Added the `android-dev-build` profile to trigger building a developer apk.

Developer apk is at application id forge.app.dev, and has name 'Forge (dev)'. It is installable in parallel with the official release app allowing developers to test the android build on their phones without interfering with their existing installation.

Updated how android resources are found to accommodate changes in application id at runtime. This method doesn't rely on the 'R' package.

### DIFF
--- a/forge-core/src/main/java/forge/item/PaperCardPredicates.java
+++ b/forge-core/src/main/java/forge/item/PaperCardPredicates.java
@@ -50,6 +50,13 @@ public abstract class PaperCardPredicates {
         return new PredicateNames(what);
     }
 
+    /**
+     * Filters on a card foil status
+     */
+    public static Predicate<PaperCard> isFoil(final boolean isFoil) {
+        return new PredicateFoil(isFoil);
+    }
+
     private static final class PredicatePrintedWithRarity implements Predicate<PaperCard> {
         private final CardRarity matchingRarity;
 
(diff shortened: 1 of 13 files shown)


forge-gui-android/src/forge/app/Main.java
```diff
@@ -90,6 +90,26 @@ public class Main extends AndroidApplication {
     private TextView progressText;
     private String versionString;
 
+    // The package name the resources are compiled under (stable across dev/prod).
+    // If you ever change the base app package, update this constant once.
+    private static final String RES_PKG_FALLBACK = "forge.app";
+
+    private int resId(String type, String name) {
+        // 1) Try fully-qualified with *runtime* package
+        int id = getResources().getIdentifier(getPackageName() + ":" + type + "/" + name, null, null);
```
kevlahnota: why the deftype and packagename is null? 
image

some examples: https://stackoverflow.com/questions/4427608/android-getting-resource-id-from-string

krafczyk: I'm using a fully qualified resource name so we don't need the type or package. But I think it could also work by calling like `getIdentifier(name, type, package)`. Do you want me to change it to that?

krafczyk:
@kevlahnota  I updated `resId` to explicitly use all arguments instead of the fully qualified name. It seems to work, so I'm fine with this.

kevlahnota (approved):

krafczyk:
Rebased for merge, can someone approve the workflows?

PR #8356, opened 2025-08-12
CardState: Fix Card copies itself though ReplacementEffect

### LINKED ISSUE #1715: cards like Primal Plasma needs to be rewritten because of copyable values
cards like Primal Plasma needs to be rewritten because of copyable values, but i first need to think how i should do that

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/issues/1713#issuecomment-1282322063_

Hanmac:
Closes #1715 

that should fix the Potential of a StackOverflow to mark the new ReplacementEffect as hasRun


Hanmac:
@tool4ever i tested it and it didn't crash

but there is also this ruling for Primal Clay:
>  If an object enters the battlefield as a copy of Primal Clay, it copies the values determined by its enters-the-battlefield replacement effect, but its power and toughness are determined by the copy's own enters-the-battlefield replacement effect. This can cause you to have a 3/3 creature with flying, or a 1/6 creature with flying and defender, for example. 

Or there, for Corrupted Shapeshifter:
>  If a permanent enters the battlefield as a copy of Corrupted Shapeshifter, there will be two copy effects to apply to it. One will copy the values chosen for the Corrupted Shapeshifter being copied. And you'll choose the values for the copy's own enters-the-battlefield replacement effect. The resulting permanent will have the abilities from both effects, and its power and toughness will be determined by the last effect to apply. For example, if a creature entering the battlefield is copying an 0/12 Corrupted Shapeshifter with defender, you could make it a 3/3 or an 0/12, either way with flying and defender.

forge-gui/res/cardsfolder/a/aquamorph_entity.txt
```diff
@@ -3,9 +3,9 @@ ManaCost:2 U U
...
-K:ETBReplacement:Other:MorphChoice
-SVar:MorphChoice:DB$ GenericChoice | Defined$ You | Choices$ BigTough,BigPower | SpellDescription$ As CARDNAME enters or is turned face up, it becomes your choice of 5/1 or 1/5.
-SVar:BigPower:DB$ Animate | Defined$ Self | Duration$ Permanent | Power$ 5 | Toughness$ 1 | SpellDescription$ CARDNAME is 5/1
-SVar:BigTough:DB$ Animate | Defined$ Self | Duration$ Permanent | Power$ 1 | Toughness$ 5 | SpellDescription$ CARDNAME is 1/5
+R:Event$ Moved | ValidCard$ Card.Self | Destination$ Battlefield | Layer$ Copy | ReplacementResult$ Updated | ReplaceWith$ MorphChoice | Description$ As CARDNAME enters or is turned face up, it becomes your choice of 5/1 or 1/5.
+R:Event$ TurnFaceUp | ValidCard$ Card.Self | ActiveZones$ Battlefield | Layer$ Copy | Secondary$ True | ReplaceWith$ MorphChoice | Description$ As CARDNAME enters or is turned face up, it becomes your choice of 5/1 or 1/5.
+SVar:MorphChoice:DB$ GenericChoice | Defined$ You | Choices$ BigTough,BigPower
+SVar:BigPower:DB$ Clone | Defined$ Self | Duration$ Permanent | SetPower$ 5 | SetToughness$ 1 | SpellDescription$ CARDNAME is 5/1
```
tool4ever: effect doesn't support this Duration$

forge-game/src/main/java/forge/game/card/CardState.java
```diff
@@ -750,11 +750,23 @@ public final void copyFrom(final CardState source, final boolean lki, final Card
                 triggers.add(tr.copy(card, lki));
             }
         }
+        ReplacementEffect runRE = null;
+        if (ctb != null && source.getCard().equals(ctb.getHostCard())) {
```
tool4ever: I guess it's a style question but I'd prefer this over nesting 3 simple if?
```
        if (ctb instanceof SpellAbility sp && sp.isReplacementAbility()
                && source.getCard().equals(ctb.getHostCard()))
```

tool4ever (approved):
in my tests I couldn't get it to fail

PR #8375, opened 2025-08-14
Import from CubeCobra for Draft and Sealed

antoniomartinelli:
Add a feature to import cubes from Cube Cobra.
Works for both Draft and Sealed.
1. Select Booster Draft or Sealed
2. "Import from CubeCobra" option
3. Paste the cube id (or the cube cobra url)


antoniomartinelli:
@Hanmac rebased on a clean branch from master

forge-gui/src/main/java/forge/gamemodes/limited/CubeImporter.java
```diff
@@ -0,0 +1,41 @@
+package forge.gamemodes.limited;
+
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
+
+public class CubeImporter {
+
+    public CustomLimited importFromCubeCobra(String urlInputString) {
```
Hanmac: I would have named the Class CubeCobraImporter, because it only works for CubeCobra

The function below `public static CustomLimited parseFromURL(final URL url)` should be part of this class too,
not in CustomLimited

forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
```diff
@@ -223,6 +223,34 @@ protected boolean generateProduct() {
...
+                        Localizer.getInstance().getMessage("lblEnterCubeCobraURL") + ":",
+                        Localizer.getInstance().getMessage("lblImportCube"),
+                        null);
+
+                if (cubeCobraId == null) {
+                    return false;
+                }
+
+                try {
+                    CubeImporter importer = new CubeImporter();
```
tehdiplomat: I'm fine with making this sound generic, but then everything else needs to be generic. Send the URL into the constructor, and parse it (either in the constructor or when we try to import).

The call below this one should be `importer.import()` and leave the concern of deciding who it is and what the ID is to the importer class. To start, if you provide a host that's not CubeCobra we can warn the user that we don't support cubes from X host at this time.

forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
```diff
@@ -223,6 +223,34 @@ protected boolean generateProduct() {
...
+                /*
+                 * Import a cube from CubeCobra.
+                 * Default settings are 3 boosters with a size of 15 cards.
+                 */
+                String cubeCobraId = SOptionPane.showInputDialog(
+                        Localizer.getInstance().getMessage("lblEnterCubeCobraURL") + ":",
+                        Localizer.getInstance().getMessage("lblImportCube"),
+                        null);
+
+                if (cubeCobraId == null) {
```
tehdiplomat: Is this an ID or a URL? Based off the other functions it sounds like a URL. So lets just call it cubeURL. (We don't know its coming from CubeCobra yet)

forge-gui/src/main/java/forge/gamemodes/limited/CubeImporter.java
```diff
@@ -0,0 +1,41 @@
...
+        //                           or    https://cubecobra.com/cube/list/cubeid
+        String cubeId = extractCubeId(urlInputString);
+        if (cubeId == null) {
+            return null;
+        }
+
+        try {
+            URL cubecobraUrl;
+            try {
+                cubecobraUrl = new URI("https://cubecobra.com/cube/download/forge/" + cubeId).toURL();
```
tehdiplomat: These constant lookup URLs probably should be defined at the top of the file.

forge-gui/src/main/java/forge/gamemodes/limited/CubeImporter.java
```diff
@@ -0,0 +1,41 @@
...
+
+        } catch (Exception e) {
+            System.err.println("Error importing cube: " + e.getMessage());
+            return null;
+        }
+    }
+
+    // Extracts the cube ID from a CubeCobra URL.
+    private String extractCubeId(String url) {
+        if (url.contains("cubecobra.com")) {
```
tehdiplomat: FYI, this doesn't actually validate that the url is comig from cubecobra. It could be `cubecobra.commandzone.net`

forge-gui/src/main/java/forge/gamemodes/limited/CustomLimited.java
```diff
@@ -122,6 +126,24 @@ public static CustomLimited parse(final List<String> dfData, final IStorage<Deck
         return cd;
     }
 
+    public static CustomLimited parseFromURL(final URL url) {
+        List<Pair<String, Integer>> slots = SealedTemplate.genericNoSlotBooster.getSlots();
+
+        final CustomLimited cd = new CustomLimited("CubeCobra Imported", slots);
```
tehdiplomat: Do we not hve access to the real name? Imported date might be helpful especially if we plan to store these things in a future iteration so people don't have to download each time.

forge-gui/src/main/java/forge/gamemodes/limited/CubeImporter.java
tehdiplomat: I'm fine with having this one be generic if the plan is to add other Cube hosting services. But as I mentioned above, I think it should just be `import`

Agree abou parseFromURL being pulled into this class.

forge-gui/src/main/java/forge/gamemodes/limited/CubeImporter.java
```diff
@@ -0,0 +1,41 @@
...
+            String[] parts = url.split("/");
+            return parts[parts.length - 1];
+        }
+        return url;
```
tehdiplomat: I think we want to return null if we failed to extract the cube id?

forge-gui/src/main/java/forge/gamemodes/limited/CubeImporter.java
Jetz72: I think if we got another importer down the line, we'd want to roll it into the same class anyway. So I think the name makes sense, even if the method names are a bit targeted.

forge-gui/src/main/java/forge/gamemodes/limited/CubeImporter.java
antoniomartinelli: The thing is about QoL. The idea is to make the method work both with the cube cobra url or the cube id. In the first case we just extract the id from the url and pass it. maybe I can rename the `String url` parameter to something more generic like `String inputStr` and explain the method in a comment.

forge-gui/src/main/java/forge/gamemodes/limited/CubeImporter.java
antoniomartinelli: I will move it to the CubeImporter class

forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
antoniomartinelli: it should be an Id. But for better QoL we shoud implicitly support URL too. So I'm renaming the variable to a more generic `inputCubeId`

forge-gui/src/main/java/forge/gamemodes/limited/CustomLimited.java
antoniomartinelli: Done, appended also a date to the name

forge-gui/src/main/java/forge/gamemodes/limited/CubeImporter.java
Jetz72: Maybe include some sanity checks to ensure it could plausibly be a valid Cube ID? Dunno what characters Cube Cobra allows allows in its IDs but I imagine slashes are off the table at least.

antoniomartinelli: Sanity checks added.

forge-gui/src/main/java/forge/gamemodes/limited/CubeImporter.java
antoniomartinelli: added a switch in case of multiple cube hosting platforms. A longer and preciser URL is analyzed now.

forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
antoniomartinelli: I agree, btw `import` is already taken by java as a name. `importCube` may be a choice here. Added a potential support to Cube Artisan which is the only alternative to Cube Cobra afaik

antoniomartinelli:
Thanks for your suggestions. I went through some heavy refactoring and generalisation for the CubeImporter class. Let me know what do you think 😃

krafczyk:
@antoniomartinelli very nice functionality, I'll defer to the others for implementation specifics, but they're making great suggestions esp. regarding the option to expand to more services in the future.

I wonder if it's difficult to add 'memory' to this importer? Like 'recently imported ids'? So the user doesn't have to type in LSVCube every time for example?

antoniomartinelli:
> @antoniomartinelli very nice functionality, I'll defer to the others for implementation specifics, but they're making great suggestions esp. regarding the option to expand to more services in the future.
> 
> I wonder if it's difficult to add 'memory' to this importer? Like 'recently imported ids'? So the user doesn't have to type in LSVCube every time for example?

I could use the `ForgeProfileProperties` to remember and propose the last inserted cube id. As you mentioned on discord, in future updates I would like to add the possibility to actually download the cube and save it locally and add it to the list of custom cubes available.

antoniomartinelli:
regarding #8375 I'm not sure of which can be a correct way to implement this without using `ForgePreferences`. If this implementation can create problems we can safely rollback it. 

For the rest the feature should be complete for its first iteration. In the future I plan to support to properly download the cubes to local user .dck files.

forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
tehdiplomat: I don't see this as qol factor if passing the id only works for a single provider (since we assume cube cobra without any guidance). I'll merge for now but if purple start asking for support for this feature when multiple providers are available and doesn't do what we they expect we should change it.

Jetz72:
Ideally the stuff for saving the last imported URL should be rendered obsolete at some point by adding support for actually storing these cubes in the cache directory.

tehdiplomat (approved):

PR #8376, opened 2025-08-14
AlphaStrike - Don't try to attack itself.

### LINKED ISSUE #8377: Creatures that must attack each turn if able now always lock the game with "Attack declaration invalid"
**Describe the bug**

If you have a creature on the board that must attack each turn if able (eg. Juggernaut, Iron Golem, etc) it will now always show "Attack declaration invalid", even if you declared such a creature as an attacker.

**To Reproduce**
1. Resolve a creature that must attack every turn
2. Enter combat, declare it as an attacker
3. Game will always alert "Attack declaration invalid"

**Expected behavior**

My creatures should be tapped and attacking.

**Screenshots**

Example board state where I tried to attack with a Juggernaut

Image

 - OS: Windows 10, Steam Deck
 - Browser N/A
 - Version: 2.0.06-SNAPSHOT-08.13

**Additional context**

This did not happen in the 2.0.06-SNAPSHOT-08.11 build. This must be some regression that was introduced recently.

Eradev:
Closes #8377

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/match/input/InputAttack.java
+++ b/forge-gui/src/main/java/forge/gamemodes/match/input/InputAttack.java
@@ -132,7 +132,7 @@ public class InputAttack extends InputSyncronizedBase {
             final List<GameEntity> mustAttack = StaticAbilityMustAttack.entitiesMustAttack(c);
             if (!mustAttack.isEmpty()) {
                 for (final GameEntity defender : mustAttack) {
-                    if (CombatUtil.canAttack(c, defender)) {
+                    if (!c.equals(defender) && CombatUtil.canAttack(c, defender)) {
                         combat.addAttacker(c, defender);
                         refreshCards.add(CardView.get(c));
                         break;


Eradev:
Should I move the check into canAttack(), even if it looks silly as a check? @Hanmac

Hanmac:
> Should I move the check into canAttack(), even if it looks silly as a check? @Hanmac

I was thinking that too, but i think that is such a corner case that `StaticAbilityMustAttack.entitiesMustAttack` returns the same Card object

Maybe even further down the line? After this:
https://github.com/Card-Forge/forge/blob/ca20d98a2cab7c0c9ce08fea1e5551f4a440ed56/forge-game/src/main/java/forge/game/staticability/StaticAbilityCantAttackBlock.java#L38-L43

check:
> if defender instanceof Card, check if attacker and defender have same controller
> if defender instanceof Player, check if attacker's controller is the same as defender

tool4ever:
@Hanmac
imo basic rules logic should not be part of a StaticAbility class

@Eradev
function is used heavily also by AI, need to check it's not assuming there is no controller check 🤔
e.g. if it looks for something to steal it can attack with next turn

Hanmac:
@Hanmac where else would you put the check "creature can't attack its controller"?

Hanmac:
@tool4ever where else would you put the check "creature can't attack its controller"?

i think we should merge the fix now, and refactor it later

Eradev:
> @Eradev function is used heavily also by AI, need to check it's not assuming there is no controller check 🤔 e.g. if it looks for something to steal it can attack with next turn

I mean for the self-attack part (which was the bug), it still can't attack itself even if it changes controller.

The rest might be outside of the scope?

PR #8378, opened 2025-08-14
build: add macOS app bundle support

rappazzo:
- Add native macOS .app bundles for Forge and Forge Adventure
- Create macOS Maven profile with Info.plist and launcher scripts
- Generate optional DMG for distribution (macOS only)
- Remove redundant .command scripts in favor of native app bundles
- Update README with comprehensive build instructions

----

It looks like there was some previous work or attempt to build MacOS app files.  I am not clear on why those are not working.  I was able to build these and run both on my Mac OS system, so maybe that is good enough to switch to.  

Please let me know if there is anything that I need to do to adjust them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,41 @@ Join the **Forge community** on [Discord](https://discord.gg/HcPJNyD66a)!
 3. **User Data Management:** Previous players’ data is preserved during upgrades.
 4. **Java Requirement:** Ensure you have **Java 17 or later** installed.
 
+**Platform-specific launchers:**
+- **macOS:** `Forge.app`, `Forge Adventure.app`, and `Adventure Editor.app`
+- **Windows:** `forge.exe` and `forge-adventure.exe`
+- **Linux:** `forge.sh` and `forge-adventure.sh`
+- **Cross-platform:** JAR files
+
 ### 📱 Android Installation
 - _(Note: **Android 11** is the minimum requirements with at least **6GB RAM** to run smoothly. You need to enable **"Install unknown apps"** for Forge to initialize and update itself)_
 - Download the **APK** from the [Snapshot Build](https://github.com/Card-Forge/forge/releases/tag/daily-snapshots). On the first launch, Forge will automatically download all necessary assets.
+
+## 🔨 Building from Source
+
+To build Forge for all platforms:
+
+```bash
+git clone https://github.com/Card-Forge/forge.git
+cd forge
+mvn clean package -DskipTests
+```
+
+**Platform-specific builds:**
+- **All platforms:** `mvn clean package -DskipTests` (default)
+- **Windows/Linux:** `mvn clean package -Pwindows-linux -DskipTests`
+- **MacOS app bundles:** `mvn clean package -Pmacos -DskipTests`
+- **Force DMG creation:** `mvn clean package -Pmacos-dmg -DskipTests` (only available on MacOS)
+
+**Requirements:**
+- Java 17 or later
+- Maven 3.6+
+- Git
+
+**Output locations:**
+- Desktop apps: `forge-gui-desktop/target/`
+- Adventure mode: `forge-gui-mobile-dev/target/`
+- Installers: `forge-installer/target/`
 
 ---
 
--- a/adventure-editor/src/main/config/adventure-editor.command
+++ b/adventure-editor/src/main/config/adventure-editor.command
@@ -1,3 +1,0 @@
-#!/bin/sh
-cd $(dirname "${0}")
-java -Xmx4096m -Dfile.encoding=UTF-8 -jar $project.build.finalName$
--- a/forge-gui-desktop/src/main/config/forge.command
+++ b/forge-gui-desktop/src/main/config/forge.command
@@ -1,3 +1,0 @@
-#!/bin/sh
-cd $(dirname "${0}")
-java $mandatory.java.args$ -jar $project.build.finalName$
--- a/forge-gui-desktop/src/main/java/forge/view/FView.java
+++ b/forge-gui-desktop/src/main/java/forge/view/FView.java
@@ -103,6 +103,12 @@ public enum FView {
 		frmDocument.setMinimumSize(new Dimension(800, 600));
 		frmDocument.setLocationRelativeTo(null);
 		frmDocument.setDefaultCloseOperation(WindowConstants.DISPOSE_ON_CLOSE);
+		frmDocument.addWindowListener(new java.awt.event.WindowAdapter() {
+			@Override
+			public void windowClosing(java.awt.event.WindowEvent e) {
+				System.exit(0);
+			}
+		});
 		frmDocument.setIconImage(FSkin.getIcon(FSkinProp.ICO_FAVICON));
 
 		// Frame components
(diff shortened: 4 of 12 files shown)


Hanmac:
The biggest problem probably is the exec to `hdiutil` inside the maven

I don't know if the workflow can automate that
and might/will crash on non MacOS machines

rappazzo:
I included `failonerror="false"`, so I would hope that doesn't crash.
```
+                                        <exec executable="hdiutil" failonerror="false">
```
I'll try to find a windows VM to build on

Eradev:
There are functions not implemented in IOSAdapter IIRC. I don't know if it'll impact this. (I don't have anything Apple)

rappazzo:
This is just for MacOS, not iOS.  Is that what you mean?  The app bundle is essentially a script wrapper.  It shouldn't be doing anything that the old `.command` files weren't doing.

Hanmac:
@rappazzo 
There are places in the POM files that replaces `$project.build.finalName$` with `${project.build.finalName}-jar-with-dependencies.jar` for the command files

https://github.com/Card-Forge/forge/blob/113dc4d5e84f505ceda021cb5cf36a9046ada343/forge-gui-desktop/pom.xml#L141-L151
https://github.com/Card-Forge/forge/blob/113dc4d5e84f505ceda021cb5cf36a9046ada343/forge-gui-mobile-dev/pom.xml#L126-L136

that would be better than `java -jar forge-gui-desktop-*.jar`

rappazzo:
After running in windows, I do see that it is failing on the missing `hdiutil`.  I will adjust and update this PR

Hanmac:
i found this maven plugin:
https://github.com/federkasten/appbundle-maven-plugin
but it is probably not continued anymore

better/newer plugin:
https://github.com/perdian/macosappbundler-maven-plugin

(ugh the newer plugin uses `hdiutil` too internally)

Hanmac:
I found the old code:
https://github.com/Card-Forge/forge/blob/feb6062f2ff5969fc404096de023d9bc3ed1d35a/forge-gui-desktop/pom.xml#L484
calls this:
https://github.com/Card-Forge/forge/blob/feb6062f2ff5969fc404096de023d9bc3ed1d35a/forge-gui-desktop/pom.xml#L517

and then this file does the calls to `hdiutil`
https://github.com/Card-Forge/forge/blob/master/forge-gui-desktop/src/main/config/create-dmg

@tehdiplomat maybe we can reactivate the pom workflow using this plugin? 
https://github.com/perdian/macosappbundler-maven-plugin

rappazzo:
I've updated the build script (and this PR) to conditionally call hdiutil on MacOS and not other OSs.  I'll look at your other suggestions in a bit and see if I can do something with that.

rappazzo:
@Hanmac I think I have adjusted as you suggested.  Please confirm that the latest is correct.   

I'll look at the old version now

Hanmac:
the `--add-opens` for the macos files should not be needed, these should be baked into the jar files already

rappazzo:
I've updated this to remove `--add-opens` in the app launcher.  I ran a build on my windows VM and it succeeded (it just didn't create the dmg file).  Let me know if there is anything else to adjust.

rappazzo:
While using the app on Mac OS, I noticed that the application didn't exit when I hit the close button.  I added a new commit to this branch which attempts to fix that (see the changes in FView.java).  If you prefer this as a separate PR, I can manage that, please let me know.

Hanmac:
like i commented multiple times, i would prefer is the whole dmg stuff could be done via plugin like:
 https://github.com/perdian/macosappbundler-maven-plugin

rappazzo:
I've reimplemented the MacOS app build to use a maven plugin.  I ended up using [javapackager](https://github.com/javapackager/JavaPackager), as I was unable to get a good build using the plugin that you previously suggested.

Let me know if there is anything else to work on here.  Thanks for the look.

rappazzo:
@Hanmac Have you had a chance to review my updated version of this?

jamincollins:
## Findings
### Build System
#### 1. Launcher script uses unreplaced Maven placeholder
Description: The macOS launcher scripts `forge-installer/src/main/resources/macos/forge` and `forge-installer/src/main/resources/macos/forge-adventure` contain `$project.build.finalName$-jar-with-dependencies.jar`. The `replacer` plugin is configured only for `libs/*.txt, install.xml, eng.xml`. These macOS resource files are not processed, so the placeholder remains literal at runtime.

Test case:
```java
// Simulate launcher execution
String scriptContent = "#!/bin/bash\ncd \"$(dirname \"$0\")/../Resources\"\njava -Xdock:icon=\"forge.icns\" -jar $project.build.finalName$-jar-with-dependencies.jar";
boolean containsPlaceholder = scriptContent.contains("$project.build.finalName$");
// containsPlaceholder == true -> will fail to find jar
```
Impact: macOS app bundles will fail to launch because the jar name contains literal `$project.build.finalName$`.

### Build Configuration
#### 2. macOS profile activated by default on all platforms
Description: `forge-installer/pom.xml` defines `<profile><id>macos</id><activation><activeByDefault>true</activeByDefault></activation>` . The `javapackager` plugin executions run in `package` phase on all builds, not just macOS.

Test case:
```java
// pom.xml activation
boolean activeByDefault = true;
boolean runningOnLinux = true;
// javapackager execution will still be invoked -> potential build failure
```
Impact: CI builds on Linux/Windows may attempt macOS app packaging, causing unnecessary failures or slow builds. The profile should be activated by OS property, not by default.

#### 3. Missing JVM memory configuration in macOS launchers
Description: Legacy `.command` files used `-Xmx4096m` or `$mandatory.java.args$`. New macOS launcher scripts use plain `java -Xdock:icon=...` with no heap settings.

Test case:
```java
ProcessBuilder pb = new ProcessBuilder("bash", "forge");
pb.inheritIO().start();
// Process will use default heap (~256MB) and likely OOM on first load
```
Impact: Users launching via app bundle may encounter OutOfMemory errors, breaking the experience.

### Backwards Compatibility
#### 4. Removal of .command launchers without deprecation period
Description: `adventure-editor/src/main/config/adventure-editor.command`, `forge-gui-desktop/src/main/config/forge.command`, `forge-gui-mobile-dev/src/main/config/forge-adventure.command` are deleted. `INSTALLATION.txt` and `README.md` are updated, but existing users with scripts referencing these files will break.

Test case:
```java
File commandFile = new File("adventure-editor/src/main/config/adventure-editor.command");
boolean exists = commandFile.exists(); // false after PR
// Existing user scripts calling this path will fail
```
Impact: Breaking change for users automating launches via the .command files. No migration path provided.

### Data Integrity
#### 5. Info.plist copyright year stale
Description: Both `Info.plist` and `Info-adventure.plist` contain `Copyright © 2024 Forge Community`. The project is 2025.

Test case:
```java
String plist = "<?xml...><string>Copyright © 2024 Forge Community...";
assert plist.contains("2024"); // true, should be 2025
```
Impact: Minor, but indicates templates not maintained.

## Verdict
Request Changes
The PR introduces macOS app bundle support but contains critical launcher defects (unreplaced Maven placeholder, missing memory settings) and a build profile that activates on all platforms. These issues will prevent successful launches and break CI.

## Recommended Fixes
1. Process macOS launcher scripts through the replacer plugin or use Maven resource filtering to replace `$project.build.finalName$` with `${project.build.finalName}`.
2. Add JVM memory arguments to macOS launcher scripts, e.g. `-Xmx4096m -Dfile.encoding=UTF-8`.
3. Change macOS profile activation from `activeByDefault` to OS-specific activation: `<os><family>mac</family></os>`.
4. Consider keeping `.command` files as deprecated wrappers for one release cycle, or document migration clearly.
5. Update copyright year in Info.plist files to current year.

---

PR #8382, opened 2025-08-14
TLA/TLE cards (14th August), Batch 1

Fulgur14:


forge-gui/res/cardsfolder/upcoming/beetle_headed_merchants.txt
```diff
@@ -0,0 +1,10 @@
+Name:Beetle-Headed Merchants
+ManaCost:4 B
+Types:Creature Human Citizen
+PT:5/4
+T:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigDraw | TriggerDescription$ Whenever this creature attacks, you may sacrifice another creature or artifact. If you do, draw a card and put a +1/+1 counter on this creature.
+SVar:TrigDraw:DB$ Draw | UnlessCost$ Sac<1/Artifact.Other;Creature.Other/another creature or artifact> | UnlessPayer$ You | UnlessSwitched$ True | SubAbility$ DBPutCounter
```
tool4ever: why UnlessCost?

Fulgur14: Hm, right if I make it as AB$, and the cost is not paid, SubAbility$ won't execute either, right?

tool4ever (approved):

PR #8386, opened 2025-08-14
TLA: Earthbend

### LINKED ISSUE #8365: TLA: Earthbend
Needs to be done as extra Effect:

* Takes Num as Param for the Amount
* Does "Target Land" Internal. (for this, I will do some changes with AbilityFactory to make it depend on the EffectApi)
* Then does the Animate and the PutCounter
* And two Delayed Trigger for dies & exiled
* And finishes with extra "Earthbend" trigger

The AbilityFactory changes might be done later, i can live without them

Hanmac:
Closes #8365


Hanmac:
@tool4ever @Fulgur14 might add DeckHints and different apostrophe later

Hanmac:
@Agetian might need to look into the AI for this

Agetian:
This looks pretty similar to the concept of a manland, think we can reuse parts of the relevant logic here?

PR #8391, opened 2025-08-15
Update the forge.sh script to be more usable

krafczyk:
As a developer when I build forge with a `mvn install` command, I don't want to go through a whole rigamarole to package the jars before running. I try to run directly out of the `forge-gui-desktop/target` directory for instance. These changes make that easier as the developer can now directly run `bash forge-gui-desktop/target/forge.sh` after running the `install` target and run the built forge. It sets up a link to `forge-gui/res` so it can find it's resources and it checks if `JAVA_HOME` is set, and if so uses the appropriate `java` executable.

It's possible people have a different use pattern in mind. Let me know if I've got it wrong. Otherwise, I'm just trying to share how I do it now.

### DIFF
--- a/forge-gui-desktop/src/main/config/forge.sh
+++ b/forge-gui-desktop/src/main/config/forge.sh
@@ -1,3 +1,27 @@
 #!/bin/sh
+set -euo pipefail
+
 cd $(dirname "${0}")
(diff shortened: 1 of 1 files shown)


krafczyk:
Additionally, this `forge.sh` allows an option so the user can specify the `forge-gui-mobile-dev` jar instead if they want to test the mobile gui.

PR #8395, opened 2025-08-15
Display flip cards correctly

Eradev:
In Forge and Adventure.

Flip cards are cards like Nezumi Shortfang.

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/screens/match/controllers/CPicture.java
+++ b/forge-gui-desktop/src/main/java/forge/screens/match/controllers/CPicture.java
@@ -69,12 +69,12 @@ public class CPicture implements ICDoc {
      */
     void showCard(final CardView c, final boolean isInAltState, final boolean mayView, final boolean mayFlip) {
         final CardStateView toShow = c != null && mayView ? c.getState(isInAltState) : null;
-        boolean showFlipped = c != null && c.isFlipped();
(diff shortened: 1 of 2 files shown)


forge-gui-desktop/src/main/java/forge/gui/CardPicturePanel.java
```diff
@@ -157,4 +163,25 @@ public void showAsDisabled(){
     }
 
     public void showAsEnabled(){ this.panel.setAlpha(0.0f); }
+
+    private BufferedImage rotateImage180(BufferedImage image) {
```
tool4ever, line 167: I don't understand what you're fixing, Forge has supported showing them flipped for a long time?

grafik

kevlahnota (approved):

forge-gui-desktop/src/main/java/forge/gui/CardPicturePanel.java
kevlahnota: try using forge with card images with flipped cards and try to flip them to see the issue. Scryfall don't store the flipped cards with alternate back faces since the flip cards don't really have back image but rotated front image.

tool4ever: hmn ok I think I see

PR #8404, opened 2025-08-16
Update vorpal_sword.txt

dracontes:
As is, the AI flag enforcing chump blocking (as done for Summon: Primal Odin), is granted through an Animate effect to the equipped creature when the Equipment's ability is activated. This will result in non-optimal AI behavior with lines of play involving multiple combat phases ([Aggravated Assault](https://scryfall.com/card/ons/185/aggravated-assault), etc.), equip cost discounts or infinite mana to re-equip the Sword onto further creatures as happenstance may require. That issue is addressed herein following precedent in cards of the [Sword of X and Y cycle](https://github.com/Card-Forge/forge/blob/a31c0358a48bf34df52429709ec2374faf5828b8/forge-gui/res/cardsfolder/s/sword_of_fire_and_ice.txt).

However since the activated ability works at instant speed, and thus a human player can activate it after blockers are declared, a finer control regarding AI behavior is desirable, namely available mana being accounted for. I take it `MustBeBlocked:AttackingPlayerConservative` isn't too relevant?

### DIFF
--- a/forge-gui/res/cardsfolder/v/vorpal_sword.txt
+++ b/forge-gui/res/cardsfolder/v/vorpal_sword.txt
@@ -3,9 +3,9 @@ ManaCost:B
 Types:Artifact Equipment
 K:Equip:B B
 S:Mode$ Continuous | Affected$ Creature.EquippedBy | AddPower$ 2 | AddKeyword$ Deathtouch | Description$ Equipped creature gets +2/+0 and has deathtouch.
-A:AB$ Animate | Cost$ 5 B B B | Defined$ Self | Triggers$ TrigDamageDone | SubAbility$ DBAddSVar | SpellDescription$ Until end of turn, CARDNAME gains "Whenever equipped creature deals combat damage to a player, that player loses the game."
-SVar:DBAddSVar:DB$ Animate | Defined$ Equipped | sVars$ MustBeBlocked
-SVar:MustBeBlocked:True
+A:AB$ Animate | Cost$ 5 B B B | Defined$ Self | Triggers$ TrigDamageDone | staticAbilities$ EquippedMustBeBlocked | SpellDescription$ Until end of turn, CARDNAME gains "Whenever equipped creature deals combat damage to a player, that player loses the game."
+SVar:EquippedMustBeBlocked:Mode$ Continuous | Affected$ Creature.EquippedBy | AddSVar$ VorpalSwordCE
+SVar:VorpalSwordCE:SVar:MustBeBlocked:True
 SVar:TrigDamageDone:Mode$ DamageDone | ValidSource$ Creature.EquippedBy | ValidTarget$ Player | CombatDamage$ True | Execute$ DBLose | TriggerDescription$ Whenever equipped creature deals combat damage to a player, that player loses the game.
 SVar:DBLose:DB$ LosesGame | Defined$ TriggeredTarget
 Oracle:Equipped creature gets +2/+0 and has deathtouch.\n{5}{B}{B}{B}: Until end of turn, Vorpal Sword gains "Whenever equipped creature deals combat damage to a player, that player loses the game."\nEquip {B}{B}


jamincollins:
## Findings

#### 1. MustBeBlocked application timing changed from SubAbility to staticAbilities
The change replaces `SubAbility$ DBAddSVar` with `staticAbilities$ EquippedMustBeBlocked`, shifting from Animate effect sub-ability sVars addition to a continuous static ability that adds an SVar via AddSVar.

The old implementation used `DBAddSVar:DB$ Animate | Defined$ Equipped | sVars$ MustBeBlocked`. Animate effect processes `sVars` param to add SVars to the animated card at animation time.

The new implementation uses `SVar:EquippedMustBeBlocked:Mode$ Continuous | Affected$ Creature.EquippedBy | AddSVar$ VorpalSwordCE` with `SVar:VorpalSwordCE:SVar:MustBeBlocked:True`. This creates a continuous static ability that adds SVar `MustBeBlocked` via AddSVar to equipped creature while the animate effect is active.

The Animate ability is triggered by the spell's Animate effect with cost 5 B B B and lasts until end of turn. The original design used SubAbility to add MustBeBlocked SVar during animation. The new design uses staticAbilities parameter on the Animate ability to add a continuous static ability that applies while the Animate effect is active.

Test case:
```java
// Verify static ability is added on animate and removed on end of turn
SpellAbility animateSA = createVorpalSwordAnimateAbility();
Card equippedCreature = new Card();
animateSA.resolve();
assertTrue(equippedCreature.hasSVar("MustBeBlocked"));
endTurn();
assertFalse(equippedCreature.hasSVar("MustBeBlocked"));
```
Impact: If staticAbilities are not properly cleaned up when Animate effect expires, MustBeBlocked persists beyond intended duration, violating MTG rules for temporary effects.

#### 2. AddSVar chain uses nested SVar definition
The new SVar chain `AddSVar$ VorpalSwordCE` then `SVar:VorpalSwordCE:SVar:MustBeBlocked:True` uses SVar reference to SVar. If `VorpalSwordCE` SVar definition is missing or malformed, AddSVar may silently fail.

In the old implementation, `sVars$ MustBeBlocked` directly referenced the MustBeBlocked SVar definition. The new indirection adds a layer.

Test case:
```java
// SVar reference chain breaks if intermediate SVar missing
Card c = new Card();
c.addChangedSVars(Map.of("VorpalSwordCE", "SVar:MustBeBlocked:True"), timestamp, 0);
// If SVar parser fails on SVar:MustBeBlocked:True format, MustBeBlocked never applied
assertTrue(c.hasSVar("MustBeBlocked"));
```
Impact: Null pointer or parsing error in SVar chain could cause MustBeBlocked to not apply, allowing creature to attack without being blocked when it should be forced.

#### 3. Continuous static ability affects equipped creature even if equipment removed mid-turn
The static ability `Mode$ Continuous | Affected$ Creature.EquippedBy` applies while the Animate effect is active. If equipment is unequipped during the turn, the static ability may still be attached to the previously equipped creature.

Test case:
```java
// Equipment removed while animate active
Card creature = new Card();
Card sword = new Card();
sword.equip(creature);
sword.animate();
creature.unequip(sword);
// Static ability may still be on creature if Affected$ Creature.EquippedBy not re-evaluated
```
Impact: MustBeBlocked could persist on creature after equipment removed, causing incorrect combat rules.

#### 4. SVar definition format change from direct to nested
Old: `SVar:MustBeBlocked:True` directly defined. New: `SVar:VorpalSwordCE:SVar:MustBeBlocked:True` nests SVar reference.

If other cards or scripts reference `SVar:MustBeBlocked` directly, they may break. The SVar name remains the same but definition format changed.

Test case:
```java
// Check existing cards referencing MustBeBlocked SVar definition
String sVarDef = card.getSVarDefinition("MustBeBlocked");
assertEquals("True", sVarDef); // Old format
// New format requires parsing SVar:MustBeBlocked:True from VorpalSwordCE
```
Impact: Backwards compatibility risk if parser expects simple value but gets nested SVar reference.

#### 5. staticAbilities parameter on AB$ Animate may interact differently with AnimateEffect
AnimateEffect processes `staticAbilities` param to add static abilities to animated card. The old SubAbility approach added sVars directly via Animate effect's sVars handling. The new approach adds a continuous static ability that adds sVars.

Potential difference in timing: SubAbility runs as part of Animate effect resolution. staticAbilities may be added during ability initialization, possibly with different timestamp handling.

Test case:
```java
// Verify timestamp ordering
SpellAbility animateSA = createVorpalSwordAnimateAbility();
long before = System.currentTimeMillis();
animateSA.resolve();
long after = System.currentTimeMillis();
// Check if static ability timestamp falls within animate duration
```
Impact: Timestamp ordering may cause MustBeBlocked to be removed prematurely or too late relative to other effects.

#### 6. Interaction with combat blocking AI
`MustBeBlocked` SVar is checked in `AiBlockController` and `ComputerUtilCombat`. Changing how it's applied may affect AI blocking decisions.

Test case:
```java
// AI blocking logic with MustBeBlocked
Card attacker = new Card();
attacker.addSVar("MustBeBlocked", "True");
AiBlockController controller = new AiBlockController();
boolean shouldBlock = controller.willBlock(attacker);
// Verify AI correctly forces block
```
Impact: If MustBeBlocked applied via static ability with different timestamp than before, AI may not recognize it in time, leading to suboptimal play.

#### 7. Interaction with multiple equip/unequip cycles in same turn
If Vorpal Sword is unequipped and re-equipped during same turn with Animate active, static ability may accumulate or conflict.

Test case:
```java
// Multiple equip/unequip
sword.equip(creature1);
sword.animate();
sword.unequip(creature1);
sword.equip(creature2);
// Verify MustBeBlocked only on creature2
```
Impact: SVar may leak to previous creature or double apply.

### Performance and security

#### 8. Additional static ability layer adds processing overhead
Creating continuous static ability with AddSVar adds extra layer processing per combat phase. For each equipped creature, Forge evaluates static abilities.

Test case:
```java
// Performance measurement
for (int i = 0; i < 1000; i++) {
    evaluateStaticAbilities(equippedCreature);
}
```
Impact: Minor performance regression in games with many equipment effects; unlikely to be significant but measurable.

#### 9. SVar name collision risk
Using `VorpalSwordCE` as intermediate SVar name may collide with other cards if name reused.

Test case:
```java
// Check for duplicate SVar names across cards
Set<String> sVarNames = getAllSVarNames();
assertFalse(sVarNames.contains("VorpalSwordCE") && count > 1);
```
Impact: Low risk but could cause unexpected interactions if another card defines same SVar name.

## Verdict
Request Changes

The PR changes MustBeBlocked application from direct sVars addition via SubAbility to staticAbilities with AddSVar chain. While functionally similar, the timing, cleanup, and interaction semantics differ. Need verification that static ability is properly cleaned up at end of turn and that MustBeBlocked applies correctly when equipment changes mid-turn.

## Recommended Fixes
1. Verify AnimateEffect correctly cleans up staticAbilities added via staticAbilities param, ensuring MustBeBlocked is removed at end of turn.
2. Add test to confirm MustBeBlocked SVar is applied immediately upon Animate resolution and removed after turn ends.
3. Verify behavior when equipment is unequipped while Animate active - MustBeBlocked should not persist on former equipped creature.
4. Document SVar chain format change for future maintainers.

---

PR #8410, opened 2025-08-16
fix: move CHANGES.txt generation to target directory

rappazzo:
- Generate CHANGES.txt in forge-gui-desktop/target/ instead of source tree
- Update installer to copy from target directory for all build profiles
- Add CHANGES.txt to .gitignore since it's generated
- Remove hardcoded fromRef to use latest tag automatically
- Remove maven-release-plugin exclusion for untracked file

Fixes issue where CHANGES.txt appeared as modified after every build.

--

While working on #8378, I keep seeing changes to `forge-gui/release-files/CHANGES.txt`, and within that file it looked like it included a lot more changes than I expected (all the way back to 2011).  It likely shouldn't be in the codebase at all, so I removed it, and set the generated location to be in the target directory.  Further, the release related scripts that use this file were adjusted to refer to the new location.

As far as reducing the changes in this file to the latest tag, I did try to include the case where the changes are being generated from a new tag (where the current revision is tagged).  In this case, the changes will be generated from the previous tag.

### DIFF
--- a/forge-gui-desktop/pom.xml
+++ b/forge-gui-desktop/pom.xml
@@ -35,7 +35,6 @@
                             <goal>timestamp-property</goal>
                         </goals>
                         <configuration>
-                            <!-- generate MonthDate code to month.date property -->
                             <name>month.date</name>
                             <pattern>MM.dd</pattern>
                         </configuration>
@@ -47,7 +46,6 @@
                             <goal>regex-property</goal>
                         </goals>
                         <configuration>
-                            <!-- generate versionName from revision property to snapshot-version property  -->
                             <name>snapshot-version</name>
                             <value>${revision}</value>
                             <regex>-SNAPSHOT</regex>
@@ -189,7 +187,7 @@
   java.base/java.nio
   java.base/java.math
   java.base/java.util.concurrent
-  java.base/java.net                                     
+  java.base/java.net
 ]]>
                             </Add-Opens>
                             <Main-Class>forge.view.Main</Main-Class>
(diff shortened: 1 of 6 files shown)


forge-gui-desktop/pom.xml
```diff
@@ -208,6 +209,31 @@
...
+                <executions>
+                    <execution>
+                        <id>get-changelog-from-tag</id>
+                        <phase>initialize</phase>
+                        <goals>
+                            <goal>execute</goal>
+                        </goals>
+                        <configuration>
+                            <source>
+                                def proc = ['bash', '-c', 'if git describe --exact-match --tags HEAD 2>/dev/null | grep -q "^forge-"; then git tag --sort=-version:refname | grep "^forge-" | sed -n "2p" || echo "HEAD"; else git tag --sort=-version:refname | grep "^forge-" | head -1 || echo "HEAD"; fi'].execute()
```
tool4ever: Please keep build process platform independent

rappazzo:
I've updated the changelog generation to use the groovy plugin to invoke the git commands.

kevlahnota (changes requested):
the android pom needs update too so the assets.zip have the changes.txt https://github.com/Card-Forge/forge/blob/55e20e96f93ec2772b2432b0058f5a49e9f548f5/forge-gui-android/pom.xml#L489C45-L489C124
the

rappazzo:
I've updated this for android and one other location.  Please take a look.

kevlahnota:
~~Anyway does the changes.txt still generated even if you didn't target desktop? Example build only android?~~
Seems to work fine gonna merge this

tool4ever:
not sure if these git commands work when building from Eclipse (with EGit only) but not a big deal?

rappazzo:
I was using `mvn` in a terminal to test and build.  I don't have an eclipse setup, so I don't have a way to have known that.  It may just be something to inject into the environment when eclipse runs the build command.

tool4ever:
Well I checked inside last Assets.zip (snapshot) and it looks like it now always only contains the very last commit...?
so your code seems a bit too strict

rappazzo:
It should be using the previous tag to generate the changes.txt file.  When I just ran it now, it produced a full change log

PR #8427, opened 2025-08-18
Break out main loop step: third time's the charm

marthinwurer:
- Broke out the main game loop into mainLoopStep() and setupFirstTurn()
- Showed how the broken out main loop step can be used in integration testing of AI logic and game functionality
- The broken out game step will allow for future simulation-style AI to use the main game engine's code instead of duplicating functionality

I recommend using the jetbrains diff viewer for this, github's doesn't show the changes as well. Most changes are simply indentation.

I broke each step of the changes out into their own commit. Hopefully that makes them easier to follow.


forge-gui-desktop/src/test/java/forge/ai/AITest.java
```diff
@@ -0,0 +1,190 @@
...
+import forge.gui.GuiBase;
+import forge.item.IPaperCard;
+import forge.item.PaperToken;
+import forge.localinstance.properties.ForgePreferences.FPref;
+import forge.model.FModel;
+
+public class AITest {
+    private static boolean initialized = false;
+
+    public Game resetGame() {
```
tool4ever, line 33: can't believe it's been over 2+ years again already since your first attempt 😱

but I recall we've had this code duplication before :p
https://github.com/Card-Forge/forge/pull/3669/commits/3a88b7ef147b9b534e27e573caf229cd236dad32

marthinwurer: Oh yeah, I wanted to keep things simple so I didn't mess with the old tests. I'll go ahead and fix that up.

marthinwurer: Even gave it the same commit message :)

forge-game/src/main/java/forge/game/phase/PhaseHandler.java
```diff
@@ -1024,132 +1019,147 @@ public void startFirstTurn(Player goesFirst, Runnable startGameHook) {
...
+    }
 
-                    if (chosenSa == null) {
-                        break; // that means 'I pass'
-                    }
-                    if (DEBUG_PHASES) {
-                        System.out.print("... " + pPlayerPriority + " plays " + chosenSa);
-                    }
+    public void mainLoopStep() {
+        StopWatch sw = new StopWatch();
```
tool4ever: It doesn't make sense to create a new object each step
maybe turn it into a class field again?

tool4ever (commented):
this seems to be working on my end
but I would really like another Dev to also look at it

tehdiplomat:
Seems to be playing ok. I'll merge after tomorrows snap so give us a day of playing with it in case it needs to be reverted

PR #8428, opened 2025-08-18
Handful of Deck Editor updates

Jetz72:
* Enables "Replace Card Variant" option in Quest and Planar Conquest
* Fixes the filter buttons and the group/pile-by dropdowns being crushed on narrower layouts. I see #8426 beat me to this but this one should be a bit more flexible - depending on available space it'll either try to fit all the buttons on one line or split them onto two.
* Disabled the conformity test for custom cards in Adventure Mode, regardless of the user settings.
* Added some command line parameters for overriding the width and height of the window. You can use these to make an IDE launch config specifically for portrait dimensions.
* Added a dev cheat command to add cards to player inventories and draft packs. It's found in the top right menu in the editor, and it takes input in the form of CardRequest strings prefixed by quantity e.g. `4 Sol Ring|SLD|[2093]`. I really wish I did this back when I was working on patching up the draft-matters stuff.


kevlahnota:
hmm I've refactor the portrait mode item editor but not yet pushed my final changes, anyway gonna check this a bit later..

kevlahnota:
Ok tested this and looks good, gonna merge this and modify the portrait layout later.

PR #8435, opened 2025-08-18
Player: extra Flag for Element Bend

Hanmac:
Extra Trigger type for generic Element Bending (for Aang the Avatar)
also extra Condition for having all 4 bend types in a turn

might need better property names


Hanmac:
@tool4ever this should clean up the trigger needed for Aang the Avatar

Jetz72 (approved):

PR #8445, opened 2025-08-19
Firebending cards (TLA)

Fulgur14:


Hanmac:
after #8435 , you might be able to script the Avatar too

Fulgur14:
I'll try, but I tend to have troubles deciphering the proper syntax from the code, so it will definitely need a separate commit :) But I already have most of it...

forge-gui/res/cardsfolder/upcoming/the_rise_of_sozin_fire_lord_sozin.txt
```diff
@@ -0,0 +1,30 @@
...
+SVar:ExileYard:DB$ ChangeZone | ValidTgts$ Opponent | ChangeType$ Card.NamedCard | Origin$ Graveyard | DefinedPlayer$ Targeted | Chooser$ You | Destination$ Exile | ChangeNum$ Y | Hidden$ True | RememberChanged$ True | SubAbility$ ExileHand | StackDescription$ None
+SVar:ExileHand:DB$ ChangeZone | Origin$ Hand | Destination$ Exile | DefinedPlayer$ Targeted | ChangeType$ Card.NamedCard | ChangeNum$ Y | Chooser$ You | SubAbility$ ExileLib | RememberChanged$ True | StackDescription$ None
+SVar:ExileLib:DB$ ChangeZone | Origin$ Library | Destination$ Exile | DefinedPlayer$ Targeted | ChangeType$ Card.NamedCard | ChangeNum$ Y | Chooser$ You | Shuffle$ True | SubAbility$ DBCleanup | StackDescription$ None
+SVar:DBTransform:DB$ ChangeZone | Origin$ Battlefield | Destination$ Exile | RememberChanged$ True | SubAbility$ DBReturn | SpellDescription$ Exile this Saga, then return it to the battlefield transformed under your control.
+SVar:DBReturn:DB$ ChangeZone | Defined$ Remembered | Origin$ Exile | Destination$ Battlefield | Transformed$ True | GainControl$ True | SubAbility$ DBCleanup
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+SVar:X:Remembered$Amount
+SVar:Y:SVar$X/NMinus.4
```
tool4ever: Oracle?

Fulgur14: How did I forget THAT?

forge-gui/res/cardsfolder/upcoming/the_rise_of_sozin_fire_lord_sozin.txt
```diff
@@ -0,0 +1,31 @@
+Name:The Rise of Sozin
+ManaCost:4 B B
+Types:Enchantment Saga
+K:Chapter:3:DBDestroyAll,DBNameCard,DBTransform
+SVar:DBDestroyAll:DB$ DestroyAll | ValidCards$ Creature | SpellDescription$ Destroy all creatures.
+SVar:DBNameCard:DB$ NameCard | Defined$ You | SubAbility$ ExileYard | StackDescription$ REP Choose a_{p:You} chooses a & Search target opponent's_{p:You} searches {p:Targeted}'s & and exile_and exiles & That player_{p:Targeted} | SpellDescription$ Choose a card name. Search target opponent's graveyard, hand, and library for up to four cards with that name and exile them. That player shuffles.
+SVar:ExileYard:DB$ ChangeZone | ValidTgts$ Opponent | ChangeType$ Card.NamedCard | Origin$ Graveyard | DefinedPlayer$ Targeted | Chooser$ You | Destination$ Exile | ChangeNum$ Y | Hidden$ True | RememberChanged$ True | SubAbility$ ExileHand | StackDescription$ None
```
tool4ever: Why not single ChangeZone?

Fulgur14: Not sure, I took this code from Ancient Vendetta.

Hanmac:
@Fulgur14 Aang should work with Trigger `ElementalBend` (the generic of these four), and `AllFourBend` as Condition for all four of them this turn

Fulgur14:
> @Fulgur14 Aang should work with Trigger `ElementalBend` (the generic of these four), and `AllFourBend` as Condition for all four of them this turn

OK, added as a new commit.

forge-gui/res/cardsfolder/upcoming/the_rise_of_sozin_fire_lord_sozin.txt
Fulgur14: > Why not single ChangeZone?

I checked various cards with Lobotomy effect, and it seems all of them are written like this, with multiple ChangeZone effects. So if a change is warranted, it should probably be its own commit.

tool4ever: _Lobotomy_ is different because it doesn't use "up to"
these two can be cleaned up

Fulgur14: It's not just those, it's also The Stone Brain and Unmoored Ego. If this can be done in a single change, Lobotomy might not work (because of its mandate to exile all cards, which can be enforced in graveyard but not in hidden zones), but Memoricide might?

Fulgur14: OK, so like this?

Hanmac: Have you tested if it does shuffle the Library at the end?
Or do you need extra Shuffle SubAbility?

Fulgur14: Hm, I thought that including Library will shuffle automatically? Cards like Agency Outfitter have specific parameter that says that shuffling is not mandatory.

Hanmac: i can't test it right now, so i'm not sure

Fulgur14: Yes, it shuffles.

Hanmac (approved):

PR #8450, opened 2025-08-19
AI logic for Krumar Initiate

### LINKED ISSUE #8449: AI needs logic for endures X with X cost
[Krumar Initiate](https://scryfall.com/card/tdm/84/krumar-initiate)
> {X}{B}, {T}, Pay X life: This creature endures X. Activate only as a sorcery. (Put X +1/+1 counters on it or create an X/X white Spirit creature token.)

AI doesn't have logic for "Choose X", so it does endure for 0.

Because it does have both X for Mana and for PayLife, it might need some checks to not choose X too much either.

Agetian:
Fixes #8449.

Not sure how to properly limit the upper value for X, this is pretty conservative when it comes to timing and the AI won't go overboard (past the danger limit), but otherwise it'll choose quite freely.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/EndureAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/EndureAi.java
@@ -2,16 +2,14 @@ package forge.ai.ability;
 
 import com.google.common.collect.Sets;
 
-import forge.ai.AiAbilityDecision;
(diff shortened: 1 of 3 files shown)


Agetian:
Hmm yeah we can prolly generalize this a little. I'll merge this in and then think some more, I'll submit a followup MR if I do that ;)

PR #8455, opened 2025-08-19
Fix Boros Signet etc. (#8454)

### LINKED ISSUE #8454: Most Signets don't work
Boros Signet etc. don't work to work around an unspecified other bug. I've fixed the signets and could not find said mystery bug.

Proof of fix:
Image

CTimmerman:
Fixes #8454

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
@@ -1744,7 +1744,7 @@ public class ComputerUtilMana {
             // if a mana ability has a mana cost the AI will miscalculate
             // if there is a parent ability the AI can't use it
             final Cost cost = a.getPayCosts();
-            if (cost.hasManaCost() || (a.getApi() != ApiType.Mana && a.getApi() != ApiType.ManaReflected)) {
+            if (/* cost.hasManaCost() || */ (a.getApi() != ApiType.Mana && a.getApi() != ApiType.ManaReflected)) {
                 continue;
             }
 


Hanmac:
that ignores the comment above:
> if a mana ability has a mana cost the AI will miscalculate

CTimmerman:
The AI seems to work fine in my screenshot, so how does it miscalculate?

Hanmac:
For example, there is currently the problem with cards like ` Scene of the Crime `, with two mana abilties like this:
> {T}: Add {C}.
> {T}, Tap an untapped creature you control: Add one mana of any color.

effects like this could cause the Auto Tapper to make more mana than it should

=
But the check for ManaCost on ManaAbility are most likely for stuff like: `Manaforge Cinder`, that has:
> {1}: Add {B} or {R}. Activate no more than three times each turn.

CTimmerman:
> effects like this could cause the Auto Tapper to make more mana than it should

Because one card has two mana abilities?

> > {1}: Add {B} or {R}. Activate no more than three times each turn.

So the limit of three times is not enforced yet? Or will the AI waste all mana on making B or R?

If other mana producers are still problematic, a check for "AI:RemoveDeck:All" should suffice.

Hanmac:
> > effects like this could cause the Auto Tapper to make more mana than it should
> 
> Because one card has two mana abilities?

Not really. It can fail in this example:
 `Llanowar Elves`
 `Scene of the Crime`

Now if you want to cast something that cost {G}{G}, the auto tapper might try to get the second G from the Scene of the Crime, while tapping the Elves again. It fails, of course, but in reverting, it doesn't clean up all the right mana abilities.
Leaving you with extra Mana.

Hanmac:
> So the limit of three times is not enforced yet? Or will the AI waste all mana on making B or R?

It isn't about the limit. `Bog Initiate` can cause this too:
> {1}: Add {B}.

the problem there is that if AI wants to get B Mana via this Ability it might miscalculate that it needs extra {1} again

tool4ever (changes requested):
This isn‘t a real fix
Avoiding cheating is a higher priority

tehdiplomat:
Signets need to have tests written around them to validate their changes. But I agree this doesn't seem to actually fix anything

CTimmerman:
`getAvailableManaEstimate` thinks _Arcane Signet_ produces 2 mana. I'm not even testing in Commander.

image

Decks:

```
[metadata]
Name=test_signets
[Main]
1 Azorius Signet|KHC|[97]
1 Boros Signet|GK1|[97]
1 Dimir Signet|GK1|[22]
1 Golgari Signet|ONC|[131]
1 Gruul Signet|GK2|[103]
1 Izzet Signet|GK1|[46]
1 Nirkana Revenant|ROE|[120]
1 Orzhov Signet|AA1|[9]
1 Rakdos Signet|C20|[249]
1 Selesnya Signet|RVR|[267]
1 Simic Signet|MKC|[236]
9 Swamp|PO2|[157]
```
```
[metadata]
Name=test_signets2
[Main]
1 Arcane Signet|PLST|[OTC-252]
1 Azorius Signet|RVR|[250]
1 Black Lotus|2ED|[233]
1 Blacker Lotus|SLD|[F869]
1 Bog Initiate|INV|[95]
1 Boros Signet|PRM|[62393]
1 Dimir Signet|PRM|[62435]
1 Dirtwater Wraith|PLST|[MIR-117]
1 Golgari Signet|TD0|[A112]
1 Gruul Signet|PLST|[C16-256]
1 Izzet Signet|PLST|[MM3-223]
1 Llanowar Elves|FDN|[439]
1 Orzhov Signet|PRM|[62399]
1 Rakdos Signet|TD0|[A116]
1 Scene of the Crime|MKM|[267]
1 Selesnya Signet|TD0|[A118]
1 Simic Signet|TD0|[A119]
1 Sol Ring|3ED|[274]
```

PR #8459, opened 2025-08-19
Avatar Aang (TLA)

Fulgur14:


forge-gui/res/cardsfolder/upcoming/avatar_aang_aang_master_of_elements.txt
```diff
@@ -0,0 +1,28 @@
+Name:Avatar Aang
+ManaCost:R G W U
+Types:Legendary Creature Human Avatar Ally
+PT:4/4
+K:Flying
+K:Firebending:2
+T:Mode$ ElementalBend | TriggerZones$ Battlefield | Execute$ TrigDraw | TriggerDescription$ Whenever you waterbend, earthbend, firebend, or airbend, draw a card. Then if you've done all four this turn, transform CARDNAME.
+SVar:TrigDraw:DB$ Draw | SubAbility$ DBTransform
+SVar:DBTransform:DB$ SetState | Defined$ Self | Mode$ Transform | Condition$ AllFourBend
```
Hanmac: `Condition$ AllFourBend` doesn't work like that (yet)

It's done like Morbid:

`X:Count$AllFourBend.1.0`

tool4ever: Condition not existing now

Fulgur14: > Condition not existing now

I changed that already?

tool4ever: you keep forgetting that CheckSVar is still a restriction and not a condition ;)

Fulgur14: This might be because Forge syntax makes less sense -- albeit only marginally -- than the syntax of English.

tool4ever (approved):

PR #8471, opened 2025-08-20
Basic AI improvement for Ambrosia Whiteheart

Agetian:
To be followed up later with better logic, but better than nothing so far.


Agetian:
Hmm, looks like this was updated recently, I'll re-check and commit a different MR

PR #8472, opened 2025-08-20
Basic AI improvement for Ambrosia Whiteheart

Agetian:
Needs better logic eventually (probably with BounceMe), but works alright for now, it seems.


Agetian:
Grr, Github shenanigans again :/

PR #8473, opened 2025-08-20
Ambrosia Whiteheart AI slight improvement

Agetian:
Just so that it doesn't bounce everything at all :)


tool4ever:
Hmn...
I don't understand why you add extra logic again when I've already included a check for subability so it wouldn't happen anyway? 🤔

Agetian:
Oh, I didn't realize that was already covered by the new check, I tested it yesterday and the AI bounced everything, but I guess the Git update didn't compkete on my side, it was really uncooperative yesterday :/ Anyhow, I'll revert this merge for now and I'll test some more later just in case 👍

PR #8476, opened 2025-08-20
YEOE: 7 cards

dracontes:
Tested card scripts for:
- [Brood Astronomer](https://scryfall.com/card/yeoe/18/brood-astronomer) and support for the `Tapped` flag in `DraftEffect.java`
- [Eumidian Lifeseed](https://scryfall.com/card/yeoe/19/eumidian-lifeseed)
- [Mutable Pupa](https://scryfall.com/card/yeoe/20/mutable-pupa)
=> Works for the most part except for the case where the creature that triggered the second triggered ability leaves the battlefield before the trigger resolves: the Pupa gets none of that creature's abilities. Tried a variety of LKI options to no avail.
- [Network Marauder](https://scryfall.com/card/yeoe/8/network-marauder)
- [Prototype X-8](https://scryfall.com/card/yeoe/24/prototype-x-8)
- [Sliver Weftwinder](https://scryfall.com/card/yeoe/25/sliver-weftwinder)
- [Volatile Orbit](https://scryfall.com/card/yeoe/27/volatile-orbit)

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/mutable_pupa.txt
+++ b/forge-gui/res/cardsfolder/upcoming/mutable_pupa.txt
@@ -1,0 +1,8 @@
+Name:Mutable Pupa
+ManaCost:G
+Types:Creature Insect
+PT:1/1
+K:Evolve
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Creature.Other+YouCtrl | TriggerZones$ Battlefield | Execute$ TrigPump | TriggerDescription$ Whenever another creature you control enters, this creature perpetually gains flying if that creature has flying. The same is true for first strike, double strike, deathtouch, haste, hexproof, indestructible, lifelink, menace, reach, trample, and vigilance.
+SVar:TrigPump:DB$ Pump | Defined$ Self | KW$ First Strike & Flying & Deathtouch & Double Strike & Haste & Hexproof & Indestructible & Lifelink & Menace & Reach & Trample & Vigilance | SharedKeywordsZone$ Battlefield | SharedRestrictions$ Card.TriggeredCard | Duration$ Perpetual
+Oracle:Evolve\nWhenever another creature you control enters, this creature perpetually gains flying if that creature has flying. The same is true for first strike, double strike, deathtouch, haste, hexproof, indestructible, lifelink, menace, reach, trample, and vigilance.
--- a/forge-gui/res/cardsfolder/upcoming/sliver_weftwinder.txt
+++ b/forge-gui/res/cardsfolder/upcoming/sliver_weftwinder.txt
@@ -1,0 +1,11 @@
+Name:Sliver Weftwinder
+ManaCost:W U B R G
+Types:Legendary Creature Sliver
+PT:6/6
+K:Warp:3
+S:Mode$ Continuous | Affected$ Sliver.YouOwn | AffectedZone$ Hand | AddKeyword$ Warp:3 | Description$ Sliver cards in your hand have warp {3}.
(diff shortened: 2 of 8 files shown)


forge-gui/res/cardsfolder/upcoming/sliver_weftwinder.txt
```diff
@@ -0,0 +1,11 @@
+Name:Sliver Weftwinder
+ManaCost:W U B R G
+Types:Legendary Creature Sliver
+PT:6/6
+K:Warp:3
+S:Mode$ Continuous | Affected$ Sliver.YouOwn | AffectedZone$ Hand | AddKeyword$ Warp:3 | Description$ Sliver cards in your hand have warp {3}.
+S:Mode$ Continuous | Affected$ Creature.Sliver+YouCtrl | AddTrigger$ WeftwinderTrig | Description$ Sliver creatures you control have “When this creature enters, conjure a random card from the Slivers spellbook into the top ten cards of your library at random, then draw a card.”
+SVar:WeftwinderTrig:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ WeftwinderConjure | TriggerDescription$ When this creature enters, conjure a random card from the Slivers spellbook into the top ten cards of your library at random, then draw a card.
```
squee1968, line 8: Needs TriggerZones$ Battlefield

dracontes: Does it? The continuous ability that grants the ETB trigger is only active on the battlefield and generally the enters triggers that only have `Card.Self` as `ValidCard$` don't seem like they'd need a `TriggerZones$` parameter as the card is only tracking itself. 
A quick survey shows that, for these trigger conditions, of 3948 `Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self` combos only 44 are explicit about `TriggerZones$ Battlefield`.

@Hanmac , @tool4ever : something to clean up either way?

squee1968: I know Thendar, the Overminer was triggering from the library. I fixed it locally, and it works fine now. I figured this needed them, too.

forge-gui/res/cardsfolder/upcoming/mutable_pupa.txt
```diff
@@ -0,0 +1,8 @@
+Name:Mutable Pupa
+ManaCost:G
+Types:Creature Insect
+PT:1/1
+K:Evolve
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Creature.Other+YouCtrl | TriggerZones$ Battlefield | Execute$ TrigPump | TriggerDescription$ Whenever another creature you control enters, this creature perpetually gains flying if that creature has flying. The same is true for first strike, double strike, deathtouch, haste, hexproof, indestructible, lifelink, menace, reach, trample, and vigilance.
+SVar:TrigPump:DB$ Pump | Defined$ Self | KW$ First Strike & Flying & Deathtouch & Double Strike & Haste & Hexproof & Indestructible & Lifelink & Menace & Reach & Trample & Vigilance | SharedKeywordsZone$ Battlefield | SharedRestrictions$ Card.TriggeredCard | Duration$ Perpetual
```
tool4ever: not sure if I understand correctly, but extending the Zone$ there probably makes sense

dracontes: Tried that while using the LKI argument variants on `SharedRestrictions$`. Didn't think to do so with the plain argument, which does work as expected.

tool4ever (approved):

Issue #7102, opened 2025-03-01
Gleemox available in Cube Draft


billyhewlett:
**Is your feature request related to a problem? Please describe.**
When you draft Gleemox in a cubedraft, it disappears when you draft it.  The AI can draft and play Gleemox just fine. 

**Describe the solution you'd like**
Gleemox should be draftable by humans just like in the MTGO PowerMax Cube, or LSV's cube, etc.

**Describe alternatives you've considered**
An alternative would be for Gleemox to not be draftable by AI either, but at that point, you might as well just not put it in your cube.

antoniomartinelli:
I went through this issue and it is about how [Gleemox](https://scryfall.com/card/prm/26584/gleemox) is coded. It takes in consideration the "This card is banned" memish rule. [Here ](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/g/gleemox.txt) you can see that it has `K:DeckLimit:0:This card is banned.` , so this basically make it fail every deck legality check because you can't include more then 0 copies of the card. 
We can:
- Remove this limitation of the card (Preferred solution)
- Keep this limitation and the card as coded right now. If you want to play with it, disable deck legality checks in the options.

billyhewlett:
Thank you very much for replying!

I recommend removing the limitation, but I appreciate that there is
something that can be done in the meantime locally.

If you do change the limitation, does that go in a patch notes?  I don't
update often because it's difficult to upload my cube afterwards.

Billy

antoniomartinelli:
Yes, the fix  is going to be released in the next version. If you are interested, I recently released a function to import cubes directly from cubecobra. It is available in the latest daily snapshots.

billyhewlett:
That is very interesting.  Normally updating Forge wipes out the
/Android/obb/forge.app/Forge/... directory, so each time I update I use a
program (Zarciver, one of the few that can copy into restricted /Android
folder) to copy my cube.dck, cube.draft, and cubeRankings.txt file into the
/cube, /draft, and /draft folders.  It's necessary to create a
cubeRankings.txt file because rankings_cubecobra.txt isn't kept current so
that new cards are rated "0" which messes up the drafting.

Updating the cube is more difficult.  To get the rankings I sort
cubecobra by ELO, export, then I wrote a python script to extract the
ordering of the cards from a .dck to match the rankings_cubecobra.txt
ranking schema.  One problem is getting the files onto the android device:
due to updates in android, files downloaded from email have chmod
[...]

PR #8486, opened 2025-08-21
Remove banned card restriction from Gleemox

antoniomartinelli:
This addresses issue #7102. 
The 0 cards maximum restriction that models the "This card is banned" is removed. This is making Gleemox fail any deck legality check.


Hanmac:
it would be better to debug why `DeckLimit` isn't working for your cases

antoniomartinelli:
`DeckLimit` is working as intended because it is actually rejecting any drafted deck that includes Gleemox.

antoniomartinelli:
Reverted the Gleemox change and made the `canHaveAnyNumberOf` skip in case of Limited format. But I'm not totally sure. Maybe in some edge cases a check on the number of cards in limited should be needed?

forge-core/src/main/java/forge/deck/DeckFormat.java
```diff
@@ -418,7 +418,7 @@ public String getDeckConformanceProblem(Deck deck) {
                 }
             }
 
-            if (canHaveAnyNumberOf(simpleCard)) {
+            if (this == DeckFormat.Limited || canHaveAnyNumberOf(simpleCard)) {
```
tool4ever, line 421: this is against the rules:
> If you're fortunate enough to open or draft ten or more Nazgûl during a Limited event, you can include only nine of them in your deck. The rest remain in your sideboard.

antoniomartinelli: I expected these scenarios. The simplest solution would be to remove the "this card is banned" restriction, which does not make sense as a statement.

tool4ever: I tend to disagree, that's more like the main point of this card
it'd just be better if it wasn't available while the preference for non-legal ones isn't enabled

antoniomartinelli: Personally I would consider the "this card is banned" as a sort of flavor text. The card is already banned in every known format. The only format it can be considered "legal" is custom cubes where it is actually played. Making it even more illegal with a restriction of 0 cards maximum does not make sense for me 😂

tool4ever: but that's not how it works in Forge right now because the PRM set is also full of legal cards

anyway users can just turn off the setting to check for deck legality if they want to play with it, so both ways of playing are covered just fine 🤷‍♂️

antoniomartinelli:
For me we can close this PR and consider the related issue a non-issue.

PR #8489, opened 2025-08-21
Fix Bloomburrow lands

Eradev:


forge-gui/res/editions/Bloomburrow.txt
```diff
@@ -24,8 +24,8 @@ Base=Uncommon:fromSheet("BLB cards")
 Base=RareMythic:fromSheet("BLB cards")
 
 [FullArtLand]
-Base=Land:fromSheet("BLB full art")
-Replace=.20F Land:fromSheet("BLB full art")+
+Base=Land:fromSheet("BLB fullart")
```
tool4ever, line 27: what's the problem here?
this isn't the only time a title with a space gets used

Eradev: Haven't found out exactly why yet, but leaving the space gives me lands from different sets. Still investigating the root cause.

image

image

tool4ever: imo a space shouldn't cause such bug, way too easy to get wrong again later
or did this never occur before?
if it can't be fixed need to also check the other places and probably be documented somewhere

Eradev: Made a proper fix that should correct this.

forge-core/src/main/java/forge/card/CardEdition.java
```diff
@@ -561,7 +563,37 @@ public List<PrintSheet> getPrintSheetsBySection() {
...
+                    var setCode = split[1].trim();
+
+                    if (split.length > 2) {
+                        // Search within a set with ArtIndex
+                        try {
+                            var artIndex = Integer.parseInt(split[2]);
+                            cardFound = cardDb.getCard(cardName, setCode, artIndex);
+                        } catch (NumberFormatException  e) {
+                            // Invalid ArtIndex so ignore
+                            cardFound = cardDb.getCard(cardName, setCode);
```
Hanmac: I don't like the ArtIndex there,
can't we use `Sword of Fire and Ice|SPG|62` or something to have the right Collector Number?
(or might need to change how EditionEntry is parsed?)

Jetz72: We've discussed it in [a discord thread](https://discord.com/channels/267367946135928833/1405972912237707387) - the print sheets use Card Request syntax so it's currently possible to use `Sword of Fire and Ice|SPG|[62]` on a custom sheet. 

But isn't `cardMap` used to store the cards from sections defined in `EditionSectionWithCollectorNumbers`? Those sections don't even use this syntax. I think this part is just for defining the basic `[cards]` print sheet and such.  Custom booster sheets are handled in the next for-loop, since they're kept in `customPrintSheetsToParse`.

forge-gui/res/editions/Bloomburrow.txt
```diff
@@ -446,38 +446,38 @@ Replace=.012F RareMythic:fromSheet("BLB cards")
...
+Swamp|BLB|3
+Swamp|BLB|4
+Island|BLB|1
+Island|BLB|2
+Island|BLB|3
+Island|BLB|4
+Forest|BLB|1
+Forest|BLB|2
+Forest|BLB|3
+Forest|BLB|4
```
Jetz72: Those leading numbers represent the frequency of a card on a print sheet; I don't think we can get rid of them.

Eradev:
I've reverted my changes for now. "full art" was a reserved section name.

image

Jetz72 (approved):
Ah, so that was the root cause behind it.

PR #8498, opened 2025-08-22
HumanPlay: Offering & Emerge cleanup

Hanmac:


tool4ever:
doesn't work, I can no longer cast any offering

forge-gui/src/main/java/forge/player/HumanPlay.java
```diff
@@ -517,28 +517,7 @@ private static boolean handleOfferingConvokeAndDelve(final SpellAbility ability,
...
-            emerge.setUsedToPay(false);
-            if (!manaInputCancelled) {
-                game.getAction().sacrifice(new CardCollection(emerge), ability, false, params);
-                ability.setSacrificedAsEmerge(game.getChangeZoneLKIInfo(emerge));
-            } else {
-                ability.resetSacrificedAsEmerge();
-            }
-        }
-        if (!table.isEmpty() && !manaInputCancelled) {
-            table.triggerChangesZoneAll(game, ability);
```
tool4ever, line 539: you can't just remove this trigger...

Hanmac: The trigger is in CostPayment

tool4ever: another one is there but your change would still fail to include the delved cards

PR #8500, opened 2025-08-22
Refactor event generation, add `set event` command.

Jetz72:
Shifts some logic around when creating events so that the individual parts are a bit more self contained.

Added support for creating an `AdventureEventData` with a predetermined card block.

Added a field to MapStage to track the local inn of the last map the player entered.

Added the `set event` command to the developer console, which replaces the event at the local Inn. It has one mandatory parameter and one optional one. The first can be either a name of a card block, or the name or set code of a card edition. If a card edition is specified, it will search for a valid card block containing only that edition. The second parameter is the event format - either Draft or Jumpstart. It usually defaults to Draft if not specified.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/data/AdventureEventData.java
+++ b/forge-gui-mobile/src/forge/adventure/data/AdventureEventData.java
@@ -129,16 +129,6 @@ public class AdventureEventData implements Serializable {
                 style = AdventureEventController.EventStyle.Bracket;
         }
         this.style = style;
-
-        switch (style) {
-            case Swiss:
-            case Bracket:
-                this.rounds = (participants.length / 2) - 1;
-                break;
-            case RoundRobin:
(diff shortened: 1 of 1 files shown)


Jetz72:
Integrated the changes for the variable pod sizes and round robin format. The `set event` command seems to work correctly with SPM events.

Moved some logic for setting up the AdventureEventData closer to the rest of that class's initialization, though I think it'd make sense down the line to take most of the card block, reward, and/or participation generation stuff and move it out to the controller class.

Jetz72:
> Is this leading a path to regenerating drafts for spending resources?

That hadn't occurred to me, but it's an option. Would need a few tweaks to do it while in the InnScene; currently the command is made to work while out of the inn but in a town. Main goal here was just to make our lives easier when we're testing specific event types and card blocks with unique draft rules.

PR #8499, opened 2025-08-22
YEOE: 4 cards, batch 3

dracontes:
Tested card scripts for:
- [Monoist Gravliner](https://scryfall.com/card/yeoe/11/monoist-gravliner) with functional trigger support
- [Spirited Simulacrum](https://scryfall.com/card/yeoe/30/spirited-simulacrum)
- [Squadron Carrier](https://scryfall.com/card/yeoe/2/squadron-carrier)

Needing engine update:
- [Thought Partition](https://scryfall.com/card/yeoe/4/thought-partition)
 => ~~The effect sequence works as far as changing the chosen card's color to white. Couldn't figure out how to change the mana cost on my own.~~ Functional engine support provided by https://github.com/Card-Forge/forge/pull/8504

PR #8504, opened 2025-08-22
AnimateBase: PerpetualManaCost

Hanmac:
PerpetualManaCost for #8499


dracontes:
Tested it on my side, Works for what's expected of Thought Partition's function. 

The only issue I came across is that on games against an AI opponent the centered mana cost overlay vanishes at some point from the cards in your hand (and may return at another point). In my testing, it seems to happen consistently when playing multiple lands with "Play Unlimited Lands" turned on in Dev Mode: sometimes as early as the  second land, sometimes as late as the fourth. Confirmed that this is something that also happens for a month old build (2025-07-19).

Hanmac:
@tool4ever check out this MR

tool4ever (approved):

PR #8508, opened 2025-08-23
AbilityUtils: add Precalc Option

### LINKED ISSUE #261: Draw X and Lose X life where X is set
Cards like Corpse Augur with:
> When Corpse Augur dies, you draw X cards and you lose X life, where X is the number of creature cards in target player’s graveyard.

or cards like Champion of Dusk:
> When Champion of Dusk enters the battlefield, you draw X cards and you lose X life, where X is the number of Vampires you control.

are a bit problematic because the X should only be calculated once when the Trigger/Ability starts resolving.
meaning Draw Replacement Effects shouldn't affect the value X later on.

maybe we could use LastState might not work for the Die Trigger? @Northmoc @tool4ever can you test that?

a change might be problematic for AI to understand that.

I'm still waiting for the confirmation from Jess. https://twitter.com/Hanmac10/status/1522124991277551618

Hanmac:
Closes #261

its mostly for "Draw X and ..." cards because Dredge Replacement could affect the board while drawing X

i don't know which other cards should need to be updated by this

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -1328,6 +1328,12 @@ public class AbilityUtils {
             game.getTriggerHandler().resetActiveTriggers();
         }
 
+        if (sa.hasParam("Precalc")) {
+            for (String s : sa.getParam("Precalc").split(",")) {
+                sa.setSVar(s, String.valueOf(calculateAmount(sa.getHostCard(), s, sa)));
+            }
+        }
+
         resolvePreAbilities(sa, game);
 
         // count times ability resolves this turn
--- a/forge-gui/res/cardsfolder/a/arabella_abandoned_doll.txt
+++ b/forge-gui/res/cardsfolder/a/arabella_abandoned_doll.txt
@@ -3,7 +3,7 @@ ManaCost:R W
 Types:Legendary Artifact Creature Toy
 PT:1/3
 T:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigDamageAll | TriggerDescription$ Whenever CARDNAME attacks, it deals X damage to each opponent and you gain X life, where X is the number of creatures you control with power 2 or less.
-SVar:TrigDamageAll:DB$ DamageAll | ValidPlayers$ Player.Opponent | NumDmg$ X | SubAbility$ DBGainLife
+SVar:TrigDamageAll:DB$ DamageAll | ValidPlayers$ Player.Opponent | NumDmg$ X | Precalc$ X | SubAbility$ DBGainLife
 SVar:DBGainLife:DB$ GainLife | Defined$ You | LifeAmount$ X
 SVar:X:Count$Valid Creature.YouCtrl+powerLE2
 Oracle:Whenever Arabella, Abandoned Doll attacks, it deals X damage to each opponent and you gain X life, where X is the number of creatures you control with power 2 or less.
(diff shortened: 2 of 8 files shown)


tool4ever:
imo we still kind of need a ruling to confirm 🤔

Hanmac:
Good luck trying to get an answer from the rules managers

tool4ever:
Well I asked here for something at least and the results are still somewhat mixed:
https://www.reddit.com/r/askajudge/comments/1mybu5t/can_effects_with_more_than_one_x_reference_use/

tool4ever:
We might try to check if some cards are on Arena to confirm there?

Hanmac:
@tool4ever that's an Idea.
Assuming that Arena isn't faulty there ;p

That is also why I added it as a flag, so we can do it for specific cards that need that.

tool4ever:
it seems Arena has no Dredge, need to find another scenario if possible :/

dracontes:
Confirmed the curious behavior of the X being continuously evaluated for Corpse Augur. Also confirmed the fix herein works as expected.

As for the rules quandary, I poked around a bit as my assumption, when the rules managers don't reply, is that, beyond their being otherwise busy, they reckon there's enough information in the Comprehensive Rules already to come to a good enough interpretation. Though I can't say the quoted comment in **CR 608.2c** is particularly helpful considering the international audience for the game: 
> "Don’t just apply effects step by step without thinking in these cases—read the whole text and apply the rules of English to the text."
> - https://mtg.wiki/page/Resolving_spells_and_abilities

Still **CR 610.1.** does state the following:
> "A one-shot effect does something just once and doesn’t have a duration. Examples include dealing damage, destroying a permanent, creating a token, and moving an object from one zone to another."
> - https://mtg.wiki/page/One-shot_effect

So I don't think the actions involved in resolving Corpse Augur's triggered ability (drawing cards, losing life) happen at different points in time, at least as far as the game is concerned, as the assumption is that the whole effect is a one-shot, and thus without duration. Also, the X needs to be determined at the start of the effect's action sequence to know how many cards will be drawn and to which of those draws a replacement effect (like dredge) will be applied to (and how much life will be lost). Then, with that in mind, **CR 608.2h.** and **CR 107.3i.** require the value of X to be locked in for the whole action sequence.

tool4ever:
they can't really happen simultaneously otherwise "Draw X, Lose X" vs. "Lose X, Draw X" would have to be treated equally by Forge
meaning any zone changes happening during either part due to replacements need to be grouped together among some other funky stuff

dracontes:
Well, my main point is that as far as making sense of what **CR 608.2h.** and **CR 107.3i** mean for calculating values while carrying out the actions pertaining to a one-shot effect, the rules do state that kind of an effect doesn't have a duration, and thus, the relevant actions are to be considered (in some aspects) as though they are simultaneous. That does agree with **608.2h.** and **107.3i**'s statements about the calculation only happening once per effect and X having the same value throughout, even if in the practical terms of humans playing the game, as **CR 608.2c** states, you take the actions in the order instructed by the rules text.
Then again, there's the warning at the end of that latter item that you're not to stolidly follow the instructions in sequence but consider the effect's text in its entirety (replacements included) to plan the resulting actions out. I'd say that argues for a one-shot effect to be more taken as a whole rather than taken piecemeal via its parts.

Hanmac:
Matt Tabak says: the value of X will change
https://bsky.app/profile/wotcmatt.bsky.social/post/3lxxnwoj76k2w

PR #8511, opened 2025-08-23
Fix Achievement timestamp display on MacOS

### LINKED ISSUE #8510: Bad character render on Achievement timestamp on MacOS
**Describe the bug**
There is a boxed X character that shows up in the timestamp when hovering on the Achievements screen on MacOS.

**To Reproduce**
0. On MacOS
1. Go to Game Settings -> Achievements
2. Hover over Match Win Streak Achievement
3. Notice the timestamp has a boxed X character printed between the time and AM/PM

**Expected behavior**
There should be no boxed X character printed.

**Screenshots**

 - OS: MacOS 15.6.1
 - Java 21
 - Forge 2.0.05

Image

dennisfriedrichsen:
fixes #8510


tool4ever (commented):
well, the real problem seems to be unicode guys trying to be clever:
https://bugs.openjdk.org/browse/JDK-8304925
and maybe our default skin font doesn't have NBSP? :/
but shouldn't MacOS also do some fallback or is there really no alternative in any font on your system... 🤔

dennisfriedrichsen:
> well, the real problem seems to be unicode guys trying to be clever: https://bugs.openjdk.org/browse/JDK-8304925 and maybe our default skin font doesn't have NBSP? :/ but shouldn't MacOS also do some fallback or is there really no alternative in any font on your system... 🤔

I'm just running the stock MacOS.  Didn't change any relevant settings like fonts or anything that I know of.  The problem occurs on a relatively recent Mac (within the last 18 months) and on a 2020 M1 Mac.

The line that ACTUALLY fixed the issue is: narrow nbsp
The other 2 I left in to possibly prevent problems.

I'm open to however you'd like this fixed.  My goal was that it should Just Work on a default MacOS install with the most recent Java LTS.  And it does not sound like Java will be changing this.

Issue #8495, opened 2025-08-22
Disabling non-legal cards removes it from any legal sets it might have been printed


autumnmyst:
**Describe the bug**
Disabling non-legal cards prevents the loading of cards like [Steamflogger Boss](https://scryfall.com/card/fut/121/steamflogger-boss) despite the fact that they are in some legal sets.

This also has an odd side effect of disabling the card [Fast // Furious](https://scryfall.com/card/j21/730/fast-furious) since there is a [test card with the same name](https://scryfall.com/card/unk/CR15a/fast-furious), despite it being a completely different effect.

The fact that the fast furious in UNK is not the same fast furious in J21 and MH2 is a separate issue (and honestly should never have happened, as two different magic cards shouldn't have the same name, but here we are). The card should be remade for UNK with the correct effect (and perhaps with a slightly different name for functionality's sake), and the fact that the J21 and MH2 "fast furious" is disabled is a symptom of the above issue.

I can even understand the removal of steamflogger boss from the normally loaded card due to it's joke nature, but there is no such argument for "fast furious" and it's very clearly a bug that it's missing.

**To Reproduce**
1. Go to Advanced Settings
2. Disable "Enable Non-Legal Cards"
3. Restart forge to reload cards
4. Go to "Audit Card and Image Data"
5. Both "fast furious" and "steamflogger boss" are listed as unimplemented
6. Look up "fast furious" and "steamflogger boss" in card search
7. They will not be present

**Expected behavior**
Both "fast furious" and "steamflogger boss" show up in legal sets (the former is a completely different card that happens to have the same name as a test card) but is seems that whenever a card is part of a non-legal set, it is not loaded, even if it has been printed in a legal set

**Screenshots**

**OS:**
 - Tested in on Windows, MacOS, and Android
 - Version Snapshot 2.0.06 (but this has been around for a while before this)

**Notes**

This can probably just be fixed with a small change to loading non-legal cards in https://github.com/autumnmyst/forge/blob/master/forge-core/src/main/java/forge/StaticData.java#L95

autumnmyst:
Looking into this, it seems there is no situation in which the same card is legal in some printings, but not in others. Steamflogger Boss should be added to the UST eternal cards list as per its legality/boarder color, and Fast // Furious should be made a separate card for the UNK set. I've done both these things in #8520

PR #8520, opened 2025-08-24
Add new duplicate cards directory to deal with duplicate test cards and implement correct Fast // Furious card in UNK

autumnmyst:
The purpose of this PR is to introduce a new duplicate cards folder to the cardsfolder directory for managing cards with duplicate names (none of which are tournament legal, but are still real cards).

The folder will include mostly test cards as they are implemented, such as:

**Red Herring** in Mystery Booster [CMB1](https://scryfall.com/card/cmb1/62/red-herring) (homonymous with [MKM](https://scryfall.com/card/mkm/142/red-herring)).
**Pick your Poison** in Mystery Booster [CMB1](https://scryfall.com/card/cmb1/97/pick-your-poison) (homonymous with [MKM](https://scryfall.com/card/mkm/170/pick-your-poison))
**Unquenchable Fury** in [Battle the Horde](https://scryfall.com/card/tbth/10/unquenchable-fury) (homonymous with [Neon Dynasty](https://scryfall.com/card/nec/23/unquenchable-fury))
**Fast // Furious** in [Unknown Event](https://scryfall.com/card/unk/CR15a/fast-furious) (homonymous with [Modern Horizons 2 - MH2](https://scryfall.com/card/mh2/123/fast-furious))

The last card, Fast // Furious, is the only one present in forge currently, however it was actually the MH2 card in functionality, both making it incorrect, and also causing the MH2 card to not load if one disabled non-legal cards in the settings (see issue link below).

My proposal is to adopt a similar naming convention to that of rebalanced alchemy cards by prepending T- to the card name. I have added the correct "Fast // Furious" playtest card under the name "T-Fast // T-Furious", since cards are loaded once no matter how many sets they appear in. This way, disabling non-legal cards doesn't prevent the loading of the tournament legal Fast // Furious.

Sidenote: also moved Steamflogger Boss to the eternal section of UST setlist so as to fix the issue linked below.

Fixes: #8495

### DIFF
--- a/forge-gui/res/editions/Unknown Event.txt
+++ b/forge-gui/res/editions/Unknown Event.txt
@@ -396,7 +396,7 @@ PLA014 C The Food Court @
 RB14 R Another Night in Vegas @
 UU14a U Tax Draw @
 UU14b U Tax Sweeper @
-CR15a C Fast // Furious @
+CR15a C T-Fast // T-Furious @
 PLA015 C High Noon At Thunder Junction @
 RR15 R Force of Rowan @
 RW15 R Crux of Mirrodin @
--- a/forge-gui/res/cardsfolder/duplicate/t-fast_t-furious.txt
+++ b/forge-gui/res/cardsfolder/duplicate/t-fast_t-furious.txt
@@ -1,0 +1,17 @@
+Name:T-Fast
+ManaCost:1 R
+Types:Instant
+K:Fuse
+A:SP$ Pump | ValidTgts$ Creature | TgtPrompt$ Select target creature | KW$ Haste | SubAbility$ DBUnblockable | SpellDescription$ Target creature gains haste until end of turn. It can’t be blocked this turn except by Vehicles or by creatures with haste.
+SVar:DBUnblockable:DB$ Effect | RememberObjects$ Targeted | ExileOnMoved$ Battlefield | StaticAbilities$ Unblockable
+SVar:Unblockable:Mode$ CantBlockBy | ValidAttacker$ Card.IsRemembered | ValidBlocker$ Creature.withoutHaste+nonVehicle | Description$ This creature can’t be blocked this turn except by Vehicles or by creatures with haste.
+AlternateMode:Split
+Oracle:Target creature gains haste until end of turn. It can’t be blocked this turn except by Vehicles or by creatures with haste.\nFuse (You may cast one or both halves of this card from your hand.)
+
+ALTERNATE
+
+Name:T-Furious
+ManaCost:2 R
+Types:Instant
+A:SP$ Pump | ValidTgts$ Creature | TgtPrompt$ Select target creature to get +3/+0 | NumAtt$ +3 | SpellDescription$ Target creature gets +3/+0 until end of turn.
+Oracle:Target creature gets +3/+0 until end of turn.\nFuse (You may cast one or both halves of this card from your hand.)
--- a/forge-gui/res/editions/Unstable.txt
+++ b/forge-gui/res/editions/Unstable.txt
@@ -128,7 +128,6 @@ ScryfallCode=UST
 90 C Just Desserts @Zoltan Boros
 91 C Painiac @McLean Kendree
 92 U Party Crasher @Mike Burns
-93 R Steamflogger Boss @Warren Mahy
 94 R Steamflogger of the Month @Warren Mahy
 95 U Steamflogger Temp @Jeff Miracola
 96 U Steamfloggery @Emrah Elmasli
@@ -275,6 +274,9 @@ ScryfallCode=UST
 214 L Swamp @John Avon
 215 L Mountain @John Avon
 216 L Forest @John Avon
+
+[eternal]
+93 R Steamflogger Boss @Warren Mahy
 
 [tokens]
 1 w_4_4_angel_flying @Magali Villeneuve


dracontes:
I believe the current consensus is to use @Jetz72 's variant system (as seen for [Garbage Elemental](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/g/garbage_elemental.txt)) to address these overlaps. The idea in that process is, in part, to fold all the cards in `\rebalanced` into the scripts of their paper variants.

Jetz72:
> I believe the current consensus is to use Jetz72 's variant system (as seen for [Garbage Elemental](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/g/garbage_elemental.txt)) to address these overlaps. The idea in that process is, in part, to fold all the cards in `\rebalanced` into the scripts of their paper variants.

Has this been attempted before? I remember thinking these name collisions seemed a bit different from regular functional variants since the two cards with the same name were completely disparate from one another, unlike other cases where the variant represents a tweak to the original. They're also meant to have separate legalities; I'm hoping that proves resolvable when working on Alchemy stuff.

dracontes:
> Has this been attempted before?

Not that I know of. However in discussion with @paulsnoops about name overlaps, such as [No Way Out (MID)](https://scryfall.com/card/mid/116/no-way-out) and [No Way Out (UNK)](https://scryfall.com/card/unk/PLA031/no-way-out-(playtest)) , I was given the impression the situation would be more straightforward to manage if card names weren't pre- or suffixed.

But while I was at it, might as well try and "glue" two cards together. 
Prompted as I was by No Way Out, probably not the best choice of types to mush together, an artifact and a plane. As a result, the game engine considers the plane backside a spell (on the Deck Editor, shows up for main deck section but not for the plane section.) Still the characteristics seem to be replaced adequately otherwise.
image
[hedron_archive.txt](https://github.com/user-attachments/files/21986921/hedron_archive.txt)
[Test Cards.txt](https://github.com/user-attachments/files/21986923/Test.Cards.txt)
I do think the variant system should work adequately in practical terms for Fast // Furious (unless it clashes with the alternate system).

I do have a notion that the variant system can hew a bit closer in scripting practice to the alternate system: in having all-caps separator that doesn't need to be commented out and a flag that tells the game what to do. Something like
- VARIANT (for pure variants in the style of [Everythingamajig](https://scryfall.com/search?q=Everythingamajig&unique=cards&as=grid&order=name))
- REBALANCED (for all the tweaks)
- NAMEOVERLAP (for the Fast // Furious cases)
Though in the end I'm not sure how feasible that is.

Eradev:
Maybe we should replace the name as a primary key and change it to a proper ID field? That way we can have multiple cards with the same name and not have a prefix or suffix in the way for CARDNAME and what not.

autumnmyst:
For now, could we just remove Fast // Furious from Unknown Event as it's not actually the correct card in master? (and also moving steamflogger boss to eternal)

autumnmyst:
Following up, I was able to implement both cards in one file using the variant system, however they don't seem to be separable in terms of legality; since it's in UKN, the only way to play with any variant is to enable non-legal cards.

To be clear, my main goal with this PR was to fix the fact that the legal card fails to load when disabling non-legal cards. Implementing the correct test-card was just a nice addition on the side. If the preferred solution is a refactor of the card loading system (e.g. based on card ID instead of just name) the I will leave the addition of the test card to a later PR and just fix the legal Fast // Furious issue with this PR.

To address this, I have the logic for UNK Fast // Furious under Variant $B however I'm leaving it as $UNSUPPORTED the UNK set to avoid loading an illegal card when non-legal cards are disabled, but I am moving said unsupported variant to the "eternal" category of UNK so as to not prevent the loading of the legal cards.

This means UNK Fast // Furious will be unavailable in forge for the time being (no different from current state, as it's not the correct card), but can easily be added in when variant specific legality is implemented.

autumnmyst:
@Jetz72 @dracontes Any stopping this from being merged? I think this implementation is fine until there's a better way to differentiate variants of the same card.

forge-gui/res/cardsfolder/f/fast_furious.txt
```diff
@@ -1,16 +1,46 @@
...
 ManaCost:2 R
 Types:Instant
-A:SP$ Discard | NumCards$ 1 | Mode$ TgtChoose | SubAbility$ DBDraw | SpellDescription$ Discard a card, then draw two cards.
-SVar:DBDraw:DB$ Draw | NumCards$ 2
-DeckHas:Ability$Discard
-AlternateMode:Split
-Oracle:Discard a card, then draw two cards.
+Oracle:<Unsupported Variant>
+
+# --- VARIANT A ---
```
tehdiplomat: I really don't like the concept here of blowing up the card script like this. It makes it really hard to understand whats going on.

Jetz72: It should ideally be a different card script, but the support for registering an entirely separate script under the same name and then differentiating them in the edition file isn't there yet. I think it's fine to let them share a script until something more suitable is built. That'd take us from "legal card missing when un-cards turned off" to "illegal card present when un-cards turned off", which feels like a less severe issue.

Eradev: I think we should just add some more robust checks where we load the cards before touching the scripts.

tool4ever: It does feel tempting to reuse this system 🤔
but considering the use case also needs more support to manage different set / legality combinations on top is probably too much to happen in the same file

Though I agree that those older cards need to be treated as legal again at least

Jetz72: I think I mentioned it in some other issue or in a discord conversation but the solution I'd go with long term is to add an optional parameter to an edition entry that lets us manually specify the name of the script file to apply to the resulting card. Then we add a folder containing card scripts that don't get included in the card pool (e.g. as an upcoming card) unless that parameter specifically calls for them.

Ideally we should also do away with the behavior where un-cards are prevented from loading entirely when they're disabled. I think they should still be loaded, but not allowed in any format or random generation by default.

I'd rather not leave a legal card inaccessible while waiting for a much larger change. We could just drop support for the playtest card entirely until we have better handling for name-collisions of completely distinct cards (as opposed to partially distinct, like the variant system was built for), but I think doing it this way is an alright compromise short-term.

autumnmyst: I also don't like how messy it ended up using variants (it doesn't mesh well with alternate cards). I agree with @Jetz72 here that we should probably drop support for the test card in the mean time.

I will just remove it for the time being until we have a better way to deal with this system.

forge-gui/res/editions/Unknown Event.txt
```diff
@@ -396,7 +396,6 @@ PLA014 C The Food Court @
 RB14 R Another Night in Vegas @
 UU14a U Tax Draw @
 UU14b U Tax Sweeper @
-CR15a C Fast // Furious @
```
tool4ever, line 399: might as well just comment it out with a reason so it doesn't get readded by some future sync PR

autumnmyst: What's the standard for commenting in these edition files? (edit: I think I see an example using #)

Eradev: Just put a # in front.

autumnmyst: cool, done

tool4ever (dismissed):

autumnmyst:
sorry, just adding a reason beside the comment, quick re-approve?

PR #8531, opened 2025-08-25
Game: only reverse NonactivePlayers if more than one

Hanmac:
Reversing the Turn Order should have no Effect there if only 2 players

### DIFF
--- a/forge-game/src/main/java/forge/game/Game.java
+++ b/forge-game/src/main/java/forge/game/Game.java
@@ -421,7 +421,7 @@ public class Game {
         // Don't use getPlayersInTurnOrder to prevent copying the player collection twice
         final PlayerCollection players = new PlayerCollection(ingamePlayers);
         players.remove(phaseHandler.getPlayerTurn());
-        if (!getTurnOrder().isDefaultDirection()) {
+        if (players.size() > 1 && !getTurnOrder().isDefaultDirection()) {
             Collections.reverse(players);
         }
         return players;


tool4ever:
shouldn't hurt I guess though not good enough for multiplayer so the user might still complain:
so right now I'm thinking about why `runWaitingTrigger` even needs APNAP order there 🤔

Hanmac:
> I'm thinking about why `runWaitingTrigger` even needs APNAP order there

That is part of the rules
Trigger are put on the stack with Active Player First

tool4ever:
Sure but that‘s not happening there but later in `MagicStack`

Hanmac:
@tool4ever 

You mean `addAllTriggeredAbilitiesToStack` right?
That calls `getPlayersInTurnOrder` which would be affecting from the ReverseTurnOrder problem too
So i still need to look into that later

But, we can try to remove `getNonactivePlayers`, because it is only used in a few spaces where it might not be needed anymore

Edit1:
`addAllTriggeredAbilitiesToStack` should check first if `simultaneousStackEntryList` is empty before getting player list

~~Also, `activePlayerSAs` should not be needed to be a Property on MagicStack, if it only gets used and cleared while inside `chooseOrderOfSimultaneousStackEntry`
i mean, shouldn't it always be empty when checking `hasStateTrigger`?~~
this doesn't work because state trigger are messing with it

Hanmac:
@tool4ever the last two places where `getNonactivePlayers` is used:

https://github.com/Card-Forge/forge/blob/61a2c7cadb0ce096fa2239d0738feaf40ddb40db/forge-game/src/main/java/forge/game/zone/MagicStack.java#L960-L970
for MagicStack, maybe find a better way to set?
resetSpellCastSinceBegOfYourLastTurn/addSpellCastSinceBegOfYourLastTurn/setSpellCastSinceBegOfYourLastTurn is only ever called there

https://github.com/Card-Forge/forge/blob/7cc2cf530dc955540f607f2d6afc0d597d72a35b/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java#L345-L351
for Trigger handler, you said they are added later to the stack, so checking for APNAP isn't needed?

PR #8537, opened 2025-08-25
migrate Callback to Consumer Interface

### LINKED ISSUE #5717: Migrate guava Function/Predicate to java

kevlahnota:
- closes #5717

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/GuiDesktop.java
+++ b/forge-gui-desktop/src/main/java/forge/GuiDesktop.java
@@ -15,6 +15,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Collection;
 import java.util.List;
+import java.util.function.Consumer;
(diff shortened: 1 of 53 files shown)


kevlahnota:
@Hanmac @Jetz72 this is ok for android since Consumer was added on API 24 (Android 7) which is below the minimum Android API we support. https://developer.android.com/reference/java/util/function/Consumer

Hanmac:
Doesn't the WaitCallback places needs to be updated too? Or how does it work?

kevlahnota:
> WaitCallback

I updated it too, implements Consumer<T>, Runnable, it works the same way as before

PR #8544, opened 2025-08-26
Option to remove unsupported card from collection

Eradev:
Add an option to remove unsupported cards from collection. This could happen if an user remove or change a custom set mid-run.

image


kevlahnota:
This is ok for now, I'll refactor and maybe we should announce the unsupported and removed them all at once I guess if multiple unsupported cards are available due to some modding done to user configs.

kevlahnota (approved):

Issue #1125, opened 2022-07-14
State trigger from Olivia, Crimson Bride might trigger more than once


tool4ever:
It happens with more than one creature:
`chooseOrderOfSimultaneousStackEntry` clears them from its list before playing, so when `TriggerType.Always` gets fired next time `hasStateTrigger` doesn't find the other one.

I tried fixing it by moving the `removeAll` below instead, but that also seems to cause an infinite loop in a test somehow.

PR #8549, opened 2025-08-26
Saga ETB counters aren't intrinsic

tool4ever:
> 714.3a. As a Saga without the read ahead ability enters the battlefield, its controller puts a lore counter on it.

It matters with _Xu-Ifit_ so they still enter with a lore counter

### DIFF
--- a/forge-game/src/main/java/forge/game/GameView.java
+++ b/forge-game/src/main/java/forge/game/GameView.java
@@ -214,18 +214,20 @@ public class GameView extends TrackableObject {
         return get(TrackableProperty.Dependencies);
     }
     public void setDependencies(Table<StaticAbility, StaticAbility, Set<StaticAbilityLayer>> dependencies) {
+        if (dependencies.isEmpty()) {
+            set(TrackableProperty.Dependencies, "");
+            return;
+        }
         StringBuilder sb = new StringBuilder();
-        if (!dependencies.isEmpty()) {
-            StaticAbilityLayer layer = null;
-            for (StaticAbilityLayer sal : StaticAbilityLayer.CONTINUOUS_LAYERS_WITH_DEPENDENCY) {
-                for (Cell<StaticAbility, StaticAbility, Set<StaticAbilityLayer>> dep : dependencies.cellSet()) {
-                    if (dep.getValue().contains(sal)) {
-                        if (layer != sal) {
-                            layer = sal;
-                            sb.append("Layer " + layer.num).append(": ");
-                        }
-                        sb.append(dep.getColumnKey().getHostCard().toString()).append(" <- ").append(dep.getRowKey().getHostCard().toString()).append("\n");
+        StaticAbilityLayer layer = null;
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/zone/MagicStack.java
```diff
@@ -847,46 +847,39 @@ public final void addSimultaneousStackEntry(final SpellAbility sa) {
...
 
         return result;
     }
 
     private boolean chooseOrderOfSimultaneousStackEntry(final Player activePlayer, boolean isAbilityTriggered) {
-        if (simultaneousStackEntryList.isEmpty()) {
+        if (!activePlayer.isInGame()) {
+            return false;
+        }
+        if (!hasSimultaneousStackEntries()) {
```
Hanmac, line 880: chooseOrderOfSimultaneousStackEntry is only ever called from addAllTriggeredAbilitiesToStack,
which already checks for hasSimultaneousStackEntries

forge-game/src/main/java/forge/game/zone/MagicStack.java
```diff
@@ -847,46 +847,39 @@ public final void addSimultaneousStackEntry(final SpellAbility sa) {
...
             return false;
         }
-
         activePlayerSAs.clear();
```
Hanmac, line 883: Can you check if activePlayerSAs can be moved from a property into a local variable there?
it is only ever set in this function? so it should not be set otherwise?

forge-game/src/main/java/forge/game/GameView.java
```diff
@@ -214,19 +214,18 @@ public String getDependencies() {
         return get(TrackableProperty.Dependencies);
     }
     public void setDependencies(Table<StaticAbility, StaticAbility, Set<StaticAbilityLayer>> dependencies) {
-        if (dependencies.isEmpty()) {
-            return;
-        }
```
Hanmac: why not just `set(TrackableProperty.Dependencies, "");` before the return?

forge-game/src/main/java/forge/game/zone/MagicStack.java
tool4ever: yea but if if loops many players it could get empty between so shortcut might still get taken

forge-game/src/main/java/forge/game/zone/MagicStack.java
tool4ever: I'm 99% sure it can't be moved because in some corner cases the `hasStateTrigger` check would fail

tool4ever: oh yea, see #1126

Hanmac: Ugh its this #1125 
i might try to debug it later
i wonder where else it does trigger from

Hanmac (approved):

PR #8551, opened 2025-08-26
Fix Muraganda Petroglyphs vs. Saga with no abilities

tool4ever:
@Hanmac
doesn't feel fully clean, do you have a better idea? :/

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -3562,7 +3562,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
             return false;
         }
         if (!getReplacementEffects().isEmpty()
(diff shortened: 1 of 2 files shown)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -3561,7 +3561,8 @@ public boolean hasNoAbilities() {
         if (!getStaticAbilities().isEmpty()) {
             return false;
         }
-        if (!getReplacementEffects().isEmpty()) {
+        if (!getReplacementEffects().isEmpty()
+                && (getReplacementEffects().size() > 1 || getReplacementEffects().get(0) != getCurrentState().sagaRep)) {
```
Hanmac: You could just check for "isSaga" + not Read Ahead?

Hanmac (approved):

PR #8557, opened 2025-08-26
Add new card filters

Eradev:
Added `is:vanilla` to return vanilla creatures and lands, and `is:custom` for custom cards.


tool4ever (approved):

Hanmac:
I might have moved the IsVanilla function to CardRules itself

PR #8564, opened 2025-08-27
Event: use Record

Hanmac:
This turns all the Event classes into records implementing GameEvent interfaces

### DIFF
--- a/forge-game/src/main/java/forge/game/GameLogFormatter.java
+++ b/forge-game/src/main/java/forge/game/GameLogFormatter.java
@@ -29,25 +29,25 @@ public class GameLogFormatter extends IGameEventVisitor.Base<GameLogEntry> {
     @Override
     public GameLogEntry visit(GameEventGameOutcome ev) {
         // Turn number counted from the starting player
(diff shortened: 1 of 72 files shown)


Hanmac:
@CTimmerman you touched the GameEvents recently, what do you think of the records?

CTimmerman:
> @CTimmerman you touched the GameEvents recently, what do you think of the records?

It appears that events are now JRE14 objects that store constructor arguments and make them available via getters. I wonder what the net code size and more importantly, performance change is. It looks like both have [improved](https://minborgsjavapot.blogspot.com/2023/09/java-records-are-trusted-and.html). Excellent!

kevlahnota:
Looks fine anyway does record can be used as Table or Multimaps as replacement to guava?

Hanmac:
> Looks fine anyway does record can be used as Table or Multimaps as replacement to guava?

Not really. Records are per definition "immutable" 
and are more like a single Entry.

So they can't replace Map/Table structure.

Examples:
- CardTraitChanges might not be a record so easily because the copy function is too complicated (that could be reworked)
- CounterType could be turned into a Record because it has no changed variables (i might do that later)

tool4ever (approved):

PR #8567, opened 2025-08-28
TriggerHandler: no need for APNAP

Hanmac:
The order of the Trigger will be handled later in MagicStack

### DIFF
--- a/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java
+++ b/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java
@@ -445,7 +445,7 @@ public class TriggerHandler {
             controller = regtrig.getHostCard().getController();
         }
         // If the runParams contains MergedCards, it is called from GameAction.changeZone()
-        if (runParams.containsKey(AbilityKey.MergedCards)) {
+        if (runParams.get(AbilityKey.MergedCards) != null) {
             // Check if the trigger cares the origin is from battlefield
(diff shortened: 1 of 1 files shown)


Hanmac:
@tool4ever if this gets merged, we can remove `getNonactivePlayers`

tool4ever:
I'm checking it "soon"

forge-game/src/main/java/forge/game/trigger/TriggerHandler.java
```diff
@@ -342,12 +342,9 @@ private boolean runWaitingTrigger(final TriggerWaiting wt) {
...
         }
 
-        // AP
-        checkStatics |= runNonStaticTriggersForPlayer(playerAP, wt, delayedTriggersWorkingCopy);
-
-        // NAPs
-        for (final Player nap : game.getNonactivePlayers()) {
-            checkStatics |= runNonStaticTriggersForPlayer(nap, wt, delayedTriggersWorkingCopy);
+        // the trigger will be ordered later in MagicStack
+        for (final Player player : game.getPlayers()) {
```
tool4ever: if we agree that the order here is irrelevant, then should just get rid of looping too

Hanmac: you mean just ignore the "ForPlayer" part?
you can try if you want

tool4ever (approved):

PR #8570, opened 2025-08-29
Support Optional for Destroy effect.

Eradev:
Support for `Destroy | Optional$ True`

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/DestroyEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/DestroyEffect.java
@@ -3,17 +3,24 @@ package forge.game.ability.effects;
 import java.util.List;
 import java.util.Map;
 
+import com.google.common.collect.Iterables;
+
 import forge.game.Game;
 import forge.game.GameActionUtil;
 import forge.game.ability.AbilityKey;
+import forge.game.ability.AbilityUtils;
 import forge.game.ability.SpellAbilityEffect;
 import forge.game.card.Card;
 import forge.game.card.CardCollectionView;
 import forge.game.card.CardUtil;
 import forge.game.card.CardZoneTable;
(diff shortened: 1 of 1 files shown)


Hanmac:
got an example why the SpellAbility gets the Optional instead of the Trigger?

Eradev:
I tried coding this:

image

```text
Name:Avatar of Amosis
ManaCost:2 B B
Types:Creature Avatar
PT:2/2

K:Deathtouch

K:Tribute:3
SVar:TrigNotTribute:DB$ Destroy | Optional$ True | ValidTgts$ Creature | TgtPrompt$ Select target creature. | SpellDescription$ When this creature enters, if tribute wasn't paid, you may destroy target creature.

Oracle:Deathtouch\nTribute 3 (As this creature enters, an opponent of your choice may put three +1/+1 counters on it.)\nWhen this creature enters, if tribute wasn't paid, you may destroy target creature.

SVar:TributeAILogic:DontControlCreatures
DeckHas:Ability$Counters
```

Unless there is an optional trigger for `K:Tribute`?

Eradev:
I based it off Dig which also has an optional component to it.

Hanmac:
So it's a totally custom one, or an Arena one? (i see no copyright, so it's probably more custom)

for how the Ability is worded, i'm not 100 % sure if it can be an OptionalTrigger
I could enable Optional for the Tribute Keyword if that is what you prefer

(OptionTrigger are somewhat tricky, but they are for the GUI to support "always yes/always no")

tool4ever:
I wouldn't add extra engine support since there's a good chance it'll just get cleaned up again during some later refactor if no script in Forge needs it
but ``DestroyAll`` already supports it anyway so that should be the better option imo

Eradev:
Managed to make it work with DestroyAll. Thanks for the tip.

PR #8572, opened 2025-08-29
CounterType: turn wrapper into an interface

Hanmac:
This turns this wrapper:

CounterType:
* Enum
* String

into an Interface:
* Enum implements Interface
* Record implements Interface

one plus: for Enum, it doesn't need to create a wrapper anymore because the Enum is already the object

`(CounterType.get(CounterEnumType.P1P1)` could also be simplified to `CounterEnumType.P1P1`

i might add a different Record/Type for Custom to support different colors


Hanmac:
@CTimmerman i would be interested in how that looks for profiling

Agetian:
Looks good to me :)

Agetian (dismissed):

CTimmerman:
> @CTimmerman i would be interested in how that looks for profiling

Me too, but i'm currently working on making more cards play better. JProfiler has a 10-day evaluation period (counted per day used) but is just a GUI around the opensource debugging protocol so any (Java?) debugger should do for profiling, like this:

```python
"""JVM profiler by Cees Timmerman 2025-08-30."""
import re
import subprocess
import time

RE_THREAD_INFO = re.compile(r'''"([^"]*)".*?cpu=(\d+)''')
jcmd = r"C:\Program Files\Java\jdk-24\bin\jcmd.exe"
pid = ""

while not pid.strip():
    result = subprocess.run([jcmd], capture_output=True, text=True)
    print(result.stdout)
    pid = input("Pid? ")

traces: dict = {}
line_counts: dict = {}
for i in range(100):
    time.sleep(1)
    result = subprocess.run([jcmd, pid, "Thread.print"], capture_output=True, text=True)
    capture_trace = False
    prev_line = ""
    thread_name = ""
    trace: list = []
    for line in result.stdout.splitlines():
        if line.endswith('RUNNABLE'):  # Active thread can also be java.lang.Thread.State: TIMED_WAITING (sleeping)
            # print(f"Runnable {prev_line}")
            m = RE_THREAD_INFO.match(prev_line)
            if not m:
                print(f"WACK: {line}")
                continue

            cpu_time = int(m.group(2))
            thread_name = m.group(1)
            # print(f"CPU: {cpu_time}ms for {thread_name}")
            capture_trace = True
            continue

        if not line.strip() and len(trace) > 1:
            key = "\n".join([v for v in trace])
            # if not "forge" in key:
            #     trace = []
            #     capture_trace = False
            #     continue
            if key in traces:
                traces[key] += 1
            else:
                traces[key] = 1

            trace = []
            capture_trace = False

        if capture_trace and line.startswith("\t"):
            trace.append(line.strip())
            if line in line_counts:
                line_counts[line] += 1
            else:
                line_counts[line] = 1

        prev_line = line

    if len(traces) > 0:
        print("\nSampled stacks in increasing order of frequency:")
        traces_by_frequency = dict(sorted(traces.items(), key=lambda item: item[1]))
        for k, v in traces_by_frequency.items():
            print(f"{v}: {k}\n")

# TODO: Make flame graph.
lines_by_frequency = dict(sorted(line_counts.items(), key=lambda item: item[1]))
print("\nLine counts / flame chart:\n" + "\n".join(f"{v}: {k}" for k, v in lines_by_frequency.items()))
```

But for small changes I prefer ye olde million times loop with 
```java
for (int i = 1_000_000; i > 0; --i) {
  // old code
}
long end = System.nanoTime();
long elapsedNanos = end - start;
double elapsedMillis = elapsedNanos / 1_000_000.0;
System.out.println("Elapsed: " + elapsedMillis + " ms");

start = System.nanoTime();
for (int i = 1_000_000; i > 0; --i) {
  // new code
}
end = System.nanoTime();
elapsedNanos = end - start;
elapsedMillis = elapsedNanos / 1_000_000.0;
System.out.println("Elapsed 2: " + elapsedMillis + " ms");
```

Which is how i learned that `card.getStaticAbilities()` is 4% slower than `cardState.getStaticAbilities()`

Overall, i'd just time the test suite of a new build and compare it with the old one. In an enterprise that'd make a nice dashboard while waiting for MR approvals.

PR #8578, opened 2025-08-30
Behold the Sinister Six! and support

tool4ever:


forge-game/src/main/java/forge/game/spellability/SpellAbility.java
```diff
@@ -1500,6 +1500,22 @@ else if(prop.equals("powerLECardsInGraveyard")
                 }
             }
 
+            if (tr.isDifferentCMC() && entity instanceof Card) {
```
Hanmac, line 1503: @tool4ever what do you think? should we group the `entity instanceof Card` ones into a bigger if/extra function?

so we could do `entity instanceof Card cEntity`

tool4ever: imo more smelly than the multiple casts is that always having to add such logic in two places, especially because the other is in GUI module...

feel free to try and unify that somehow here or later 🤔

Hanmac (approved):

PR #8579, opened 2025-08-30
The Soul Stone and support

tool4ever:


forge-gui/res/cardsfolder/upcoming/the_soul_stone.txt
```diff
@@ -0,0 +1,9 @@
+Name:The Soul Stone
+ManaCost:1 B
+Types:Legendary Artifact Infinity Stone
```
tool4ever, line 3: is it multiword type or not? :/

PR #8580, opened 2025-08-30
GameActionUtil: fix getFirstSpellAbility using castSA

Hanmac:
This should fix the Emerge + Prototype thing

i also did a fix for Prototype Stack Description


tool4ever:
yea that does improve things
seems only one KW cleanup missing somewhere:
image

Hanmac:
Yeah, but that can be cleaned up later

tool4ever:
ok, worth nothing that it also happens if cast without prototype

PR #8585, opened 2025-08-30
Update urzas_miter.txt

Ellios77:
Corrects detection of artifact being sacrificed when put into graveyard vs other means

### DIFF
--- a/forge-gui/res/cardsfolder/u/urzas_miter.txt
+++ b/forge-gui/res/cardsfolder/u/urzas_miter.txt
@@ -1,10 +1,13 @@
 Name:Urza's Miter
 ManaCost:3
 Types:Artifact
-T:Mode$ Sacrificed | Static$ True | ValidCard$ Artifact.YouCtrl | TriggerZones$ Battlefield | ValidPlayer$ You | RememberTriggeringCard$ True | Execute$ DelTrigChangeZone
-SVar:DelTrigChangeZone:DB$ DelayedTrigger | Mode$ ChangesZone | Static$ True | ValidCard$ Artifact.YouCtrl+IsRemembered | Origin$ Battlefield | Destination$ Any | TriggerZones$ Battlefield | Execute$ DBCleanup
-SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True | ClearTriggered$ True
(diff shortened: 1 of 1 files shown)


Eradev:
I think we could add a condition in the code inside `Trigger.java -> meetsRequirementsOnTriggeredObjects`

```java
        else if ("NotBeenSacrificed".equals(condition)) {
            final SpellAbility trigSA = (SpellAbility) runParams.get(AbilityKey.Cause);
            if (trigSA != null && !Iterables.isEmpty(trigSA.getPaidList("Sacrificed"))) {
                return false;
            }
        } 
```

Then we could change Urza's Miter with this:

```text
Name:Urza's Miter
ManaCost:3
Types:Artifact
T:Mode$ ChangesZone | ValidCard$ Artifact.YouCtrl | Origin$ Battlefield | Destination$ Graveyard | TriggerZones$ Battlefield | Condition$ NotBeenSacrificed | OptionalDecider$ You | Execute$ TrigDraw | TriggerDescription$ Whenever an artifact you control is put into a graveyard from the battlefield, if it wasn't sacrificed, you may pay {3}. If you do, draw a card.
SVar:TrigDraw:AB$ Draw | Cost$ 3 | NumCards$ 1 | Defined$ You
DeckHints:Type$Artifact
Oracle:Whenever an artifact you control is put into a graveyard from the battlefield, if it wasn't sacrificed, you may pay {3}. If you do, draw a card.
```

tool4ever:
hmn, I feel like this script worked at some point so it looks like maybe the engine got changed
I don't think your new version handles cleanup in all cases though

@Eradev
that doesn't work
paidList is only for costs

Eradev:
> hmn, I feel like this script worked at some point so it looks like maybe the engine got changed I don't think your new version handles cleanup in all cases though
> 
> @Eradev that doesn't work paidList is only for costs

@tool4ever 

It did in my tests. ¯\\_(ツ)_/¯

If a SA had to Sac in order to activate, then it'll show up as a cost.

Eradev:
I guess would check for cases where it wasn't a cost?

```java
            if (trigSA != null &&
                    (!Iterables.isEmpty(trigSA.getPaidList("Sacrificed")) ||
                     trigSA.getApi().equals(ApiType.Sacrifice) ||
                     trigSA.getApi().equals(ApiType.SacrificeAll))) {
                return false;
            }
```

Ellios77:
Okay, should now work in all cases with cleanup.
If the artifact isn't sacrificed but it didn't go to graveyard it cleans up without activating the optional triggered ability.

tool4ever:
@Eradev
First, checking for non-empty paid list only isn't good enough:
you could match on something like _Annihilating Glare_

Then that approach would still need to check for more causes:
- BalanceEffect
- Offering/Emerge
- Sacrifice as SBA

@Ellios77
there's a rule that static abilities like on _Fractured Realm_ ignore reflexive triggers, so your workaround would make it fail to trigger an additional time

but don't you guys worry, I'll demonstrate an easier fix soon :)

PR #8587, opened 2025-08-31
GameAction: cleanup Static giving Abilities to the Stack

Hanmac:
still WIP

This fixes the problem with Herigast adding doubled Emerge to Spells on the Stack

(castFrom wasn't set, so wasCast was false when Line567 was checked)

Normal cases already works,
but what now got broken is if Herigast got sacrificed, especially for Emerge Cost.
Because it is still on the field when Line565 is checked.
So the Keyword i still there for Line567
then Herigast gets sacrificed later by some Human code (i want to clean up for later MR)
at this place i need to pinpoint, the castSa Keyword needs to be added again

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -205,6 +205,7 @@ public class GameAction {
             copied.setGameTimestamp(c.getGameTimestamp());
 
             if (zoneTo.is(ZoneType.Stack)) {
+                copied.setCastFrom(zoneFrom);
                 // try not to copy changed stats when moving to stack
 
                 // copy exiled properties when adding to stack
@@ -452,7 +453,7 @@ public class GameAction {
                 if (c.getCastSA() != null && !c.getCastSA().isIntrinsic() && c.getKeywords().contains(c.getCastSA().getKeyword())) {
                     KeywordInterface ki = c.getCastSA().getKeyword();
                     ki.setHostCard(copied);
-                    copied.addChangedCardKeywordsInternal(ImmutableList.of(ki), null, false, copied.getGameTimestamp(), null, true);
+                    copied.addChangedCardKeywordsInternal(ImmutableList.of(ki), null, false, copied.getGameTimestamp(), ki.getStatic(), true);
                 }
                 // TODO hot fix for non-intrinsic offspring
                 Multimap<StaticAbility, KeywordInterface> addKw = MultimapBuilder.hashKeys().arrayListValues().build();
@@ -566,7 +567,8 @@ public class GameAction {
         // 400.7g try adding keyword back into card if it doesn't already have it
         if (zoneTo.is(ZoneType.Stack) && cause != null && cause.isSpell() && !cause.isIntrinsic() && c.equals(cause.getHostCard())) {
             if (cause.getKeyword() != null && !copied.getKeywords().contains(cause.getKeyword())) {
-                copied.addChangedCardKeywordsInternal(ImmutableList.of(cause.getKeyword()), null, false, game.getNextTimestamp(), null, true);
+                KeywordInterface kw = cause.getKeyword();
+                copied.addChangedCardKeywordsInternal(ImmutableList.of(cause.getKeyword()), null, false, copied.getGameTimestamp(), kw.getStatic(), true);
             }
         }
 


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -452,7 +453,7 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
                 if (c.getCastSA() != null && !c.getCastSA().isIntrinsic() && c.getKeywords().contains(c.getCastSA().getKeyword())) {
                     KeywordInterface ki = c.getCastSA().getKeyword();
                     ki.setHostCard(copied);
-                    copied.addChangedCardKeywordsInternal(ImmutableList.of(ki), null, false, copied.getGameTimestamp(), null, true);
+                    copied.addChangedCardKeywordsInternal(ImmutableList.of(ki), null, false, copied.getGameTimestamp(), ki.getStatic(), true);
```
tool4ever: reuse `addKw`?

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -566,7 +567,8 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
         // 400.7g try adding keyword back into card if it doesn't already have it
         if (zoneTo.is(ZoneType.Stack) && cause != null && cause.isSpell() && !cause.isIntrinsic() && c.equals(cause.getHostCard())) {
             if (cause.getKeyword() != null && !copied.getKeywords().contains(cause.getKeyword())) {
-                copied.addChangedCardKeywordsInternal(ImmutableList.of(cause.getKeyword()), null, false, game.getNextTimestamp(), null, true);
+                KeywordInterface kw = cause.getKeyword();
```
tool4ever, line 570: Ideally we avoid doing this a second time 🤔
or does it get dropped from the `checkStaticAbilities` above...?

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -205,6 +205,7 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
             copied.setGameTimestamp(c.getGameTimestamp());
 
             if (zoneTo.is(ZoneType.Stack)) {
+                copied.setCastFrom(zoneFrom);
```
tool4ever: should be together with L228 imo

tool4ever: if you need it earlier this should become obsolete:
https://github.com/Card-Forge/forge/blob/48a5046d9371b5eda6e9f709e2fe645e634dff55/forge-game/src/main/java/forge/game/GameAction.java#L829-840

Hanmac: setCastForm in later part is set regardless of cause:
```
if (zoneTo.is(ZoneType.Stack)) {
            c.setCastFrom(zoneFrom);
```
might be cleaned up later :thinking:

Hanmac: yeah that might be cleaned up, i think
needs to be tested

Hanmac:
> Ideally we avoid doing this a second time 🤔
or does it get dropped from the checkStaticAbilities above...?

i think there are two cases we need to look after:
- First if the source of the Static Ability Keyword (i'm looking at you Henzie) gets sacrificed as Additional Cost. Then it should be gone when checkStaticAbilities is run. (Need to test that?) so the place should add the Keyword again
- Second is the Special case with Emerge. At that moment, the Herigast isn't sacrificed yet.

Also that are two different places:
- one is to Stack, 400.7g
- the other is from Stack to Battlefield 400.7a

Hanmac:
@tool4ever you problably need to help me to debug the corner cases for this

and i should write some testCases so it doesn't break again

But the Human Emerge is slightly different from the AI Emerge (see #8498)

tool4ever:
I'm currently thinking it's nearly impossible to reinject the KW if the static ability granting it disappears in all possible scenarios

what if we just always add the cast KW a second time when moving to stack but change `KeywordCollection` into `linkedHashSetValues()`?
- then it is always available as fallback
- because the KW gets cached it should always be the same object
- normally no card should have the same KeywordInterface twice anyway?

Hanmac:
> I'm currently thinking it's nearly impossible to reinject the KW if the static ability granting it disappears in all possible scenarios
> 
> what if we just always add the cast KW a second time when moving to stack but change `KeywordCollection` into `linkedHashSetValues()`?
> 
>     * then it is always available as fallback
> 
>     * because the KW gets cached it should always be the same object
> 
>     * normally no card should have the same KeywordInterface twice anyway?

I would need to think about it.
Each KeywordInterface/Instance should be unique.
(I was planing to have them extend CardTraitBase stuff or add a CardTraitInterface, but that is too big of a change yet)
But i might turn GameObject into an interface as first step

But for this problem, i need to check out these issues:
> First if the source of the Static Ability Keyword (i'm looking at you Henzie) gets sacrificed as Additional Cost. Then it should be gone when checkStaticAbilities is run.
> Second is the Special case with Emerge. At that moment, the Herigast isn't sacrificed yet.

but i didn't had time to debug them yet

tool4ever:
I still need to test a bit
but right now everything seems to work

Hanmac:
we should probably write test-cases so other changes doesn't break it.

(also the thing with Human Emerge vs AI Emerge)

tool4ever:
done, also looked fine when AI used it

tool4ever (approved):

PR #8590, opened 2025-09-01
Consider more cards playable in main1 (#8589)

### LINKED ISSUE #8589: Consider more cards playable in main1
**Describe the bug**
`Gahiji, Honored One` and many more cards lack an `SVar:PlayMain1`.

**Expected behavior**
Consider cards playable in the first main phase based on their abilities, which could change during the game.

**Additional context**
Patch coming up.

CTimmerman:
Fixes #8589. My Naya deck win rate improved by at least 50% when AI played it against its unpatched self.

I moved 0-mana, Exalted, and Extort checks before the RaidTest check as i don't think they reduce the number of blockers. The checks after RaidTest probably apply to cards that share that property with e.g. Blitz. I'm too tired to check atm.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtil.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtil.java
@@ -1141,56 +1141,274 @@ public class ComputerUtil {
     public static boolean castPermanentInMain1(final Player ai, final SpellAbility sa) {
         final Card card = sa.getHostCard();
         final CardState cardState = card.isFaceDown() ? card.getState(CardStateName.Original) : card.getCurrentState();
+        final CardCollection myCreatures = ai.getCreaturesInPlay();
 
+        // 2909+ cards match `ManaCost:no cost`
+        // trigger/boost boardstate while you can
+        if (card.getManaCost().isZero()) return true;
+
+        // 1432+ cards match `PlayMain1`
+        // manually chosen, e.g. Bello, Bard of the Brambles
         if (card.hasSVar("PlayMain1")) {
-            if (card.getSVar("PlayMain1").equals("ALWAYS") || sa.getPayCosts().hasNoManaCost()) {
+            if (Set.of("ALWAYS", "TRUE", "True").contains(card.getSVar("PlayMain1")) || sa.getPayCosts().hasNoManaCost()) {
                 return true;
             } else if (card.getSVar("PlayMain1").equals("OPPONENTCREATURES")) {
                 // Only play these main1 when the opponent has creatures (stealing and giving them haste)
                 if (!ai.getOpponents().getCreaturesInPlay().isEmpty()) {
                     return true;
                 }
-            } else if (!card.getController().getCreaturesInPlay().isEmpty()) {
+            } else if (!myCreatures.isEmpty()) {
                 return true;
             }
         }
 
-        // cast Backup creatures in main 1 to pump attackers
-        if (cardState.hasKeyword(Keyword.BACKUP)) {
-            for (Card potentialAtkr: ai.getCreaturesInPlay()) {
-                if (ComputerUtilCard.doesCreatureAttackAI(ai, potentialAtkr)) {
-                    return true;
-                }
-            }
-        }
-
-        // if AI has no speed, play start your engines on Main1
-        if (ai.noSpeed() && cardState.hasKeyword(Keyword.START_YOUR_ENGINES)) {
-            return true;
-        }
-
-        // cast Blitz in main 1 if the creature attacks
-        if (sa.isBlitz() && ComputerUtilCard.doesSpecifiedCreatureAttackAI(ai, card)) {
-            return true;
-        }
-
+        // What not to cast in main1
+        // 51+ cards match `RaidTest`
+        // Raid is better after combat
         // try not to cast Raid creatures in main 1 if an attack is likely
         if ("Count$AttackersDeclared".equals(card.getSVar("RaidTest")) && !cardState.hasKeyword(Keyword.HASTE)) {
-            for (Card potentialAtkr: ai.getCreaturesInPlay()) {
+            for (Card potentialAtkr: myCreatures) {
                 if (ComputerUtilCard.doesCreatureAttackAI(ai, potentialAtkr)) {
                     return false;
                 }
             }
         }
 
-        if (card.getManaCost().isZero()) {
+        // Activated abilities
+        for (SpellAbility nma : cardState.getNonManaAbilities()) {
+            final Cost cost = nma.getPayCosts();
+            if (cost.hasSpecificCostType(CostTap.class) && card.isAbilitySick()) continue;
+
+            // 1761+ cards match `A:AB\$.*Cost\$[^|]*Sac`
+            // sac as combat trick
+            // A Killer Among Us etc
+            // if (cost.hasSpecificCostType(CostSacrifice.class)) {
+            //     return true;
+            // }
+
+            // 244+ cards match `A:AB\$[^|]*(PumpAll|PutCounterAll)`
+            // Jazal Goldmane etc
+            if (myCreatures.size() >= 1
+                && nma.getApi() != null && Set.of(
+                    ApiType.PumpAll,
+                    ApiType.PutCounterAll
+                    ).contains(nma.getApi())) {
+                if (card.isCreature()) {
+                    final String nmaCost = nma.getParam("Cost");
+                    if (nmaCost == null || !nmaCost.contains("T")) {
+                        return true;
+                    }
+                    continue;
+                }
+            }
+
+            // 8+ cards match `A:AB.*?BeforeCombat`
+            // Apprentice Necromancer only gets here if it has haste.
+            // Arms Race already has SVar:PlayMain1:ALWAYS
+            // Cogwork Assembler cost is steep.
+            // Kiki-Jiki, Mirror Breaker etc.
+            if ("BeforeCombat".equals(nma.getParam("AILogic"))) {
+                return true;
+            }
+        }  // activated abilities
+
+        if (myCreatures.size() >= 1) {
+            // Static abilities
+            for (StaticAbility sta : cardState.getStaticAbilities()) {
+                final Set<StaticAbilityMode> mode = sta.getMode();
+
+                // 1138+ cards match `S:Mode\$ Continuous.*?Affected\$[^|]*?YouCtrl`
+                // Saryth, the Viper's Fang etc
+                if (mode.contains(StaticAbilityMode.Continuous)) {
+                    final String affected = sta.getParam("Affected");
+                    if (affected != null && affected.contains("YouCtrl")) {
+                        return true;
+                    }
+                }
+
+                // 32+ cards match `S:Mode\$ Panharmonicon`
+                // double triggers
+                if (mode.contains(StaticAbilityMode.Panharmonicon)) {
+                    return true;
+                }
+
+                String validCard = sta.getParam("ValidCard");
+                if (validCard == null) validCard = "";
+                // 26+ cards match `S:Mode\$ CantBlock.*ValidCard\$(?![^|]*(You|Self))`
+                // Bedlam; Heat Wave etc
+                if (mode.contains(StaticAbilityMode.CantBlock) || mode.contains(StaticAbilityMode.CantBlockUnless)) {
+                    if (!validCard.contains("You") && !validCard.contains("Self")) return true;
+                }
+
+                // 21+ card match `S:Mode\$ CombatDamageToughness(?!.*Self)`
+                // Ancient Lumberknot; Belligerent Brontodon etc
+                if (mode.contains(StaticAbilityMode.CombatDamageToughness)) {
+                    if (validCard.contains("Self")) continue;
+                    int moreT = 0;
+                    for (Card c : myCreatures) {
+                        if (!c.hasSickness()) moreT += c.getNetToughness() - c.getNetPower();
+                    }
+                    if (moreT > 0) return true;
+                }
+
+                // 18+ cards match `S:Mode\$ CantGainLife.*ValidPlayer\$[^|]*(?!You)`
+                // Archfiend of Despair etc
+                if (mode.contains(StaticAbilityMode.CantGainLife)) {
+                    final String validPlayer = sta.getParam("ValidPlayer");
+                    if (validPlayer != null && validPlayer.contains("You")) continue;
+                    for (Player p : ai.getOpponents()) {
+                        for (Card c : p.getCreaturesInPlay()) {
+                            if (c.hasKeyword(Keyword.LIFELINK) && !c.isTapped()) return true;
+                        }
+                    }
+                }
+
+                // 17+ cards match `S:Mode\$ CantBlockBy.*ValidAttacker\$[^|]*You`
+                // Bower Passage etc
+                if (mode.contains(StaticAbilityMode.CantBlockBy)) {
+                    final String validAttacker = sta.getParam("ValidAttacker");
+                    if (validAttacker != null && validAttacker.contains("You")) return true;
+                }
+
+                // 11+ card match `S:Mode\$ CantPreventDamage.*(?!Card.Self)`
+                // Everlasting Torment etc
+                if (mode.contains(StaticAbilityMode.CantPreventDamage)) {
+                    final String validSource = sta.getParam("ValidSource");
+                    if (!"Card.Self".equals(validSource)) return true;
+                }
+
+                // 8+ cards match `S:Mode\$ CantBeActivated.*Activator\$[^|]*(?!You)`
+                // Drana and Linvala; City of Solitude etc
+                if (mode.contains(StaticAbilityMode.CantBeActivated)) {
+                    final String activator = sta.getParam("Activator");
+                    if ("Opponent".equals(activator) || "Player.NonActive".equals(activator)) return true;
+                }
+
+                // 5+ cards match `S:Mode\$ NoCleanupDamage`
+                // Patient Zero etc
+                if (mode.contains(StaticAbilityMode.NoCleanupDamage)) {
+                    if (!"Card.Self".equals(validCard)) return true;
+                }
+
+                // 4+ cards match `S:Mode\$ CanAttackDefender.*ValidCard\$[^|]*You`
+                // Guardians of Oboro etc
+                if (mode.contains(StaticAbilityMode.CanAttackDefender)) {
+                    if (!validCard.contains("You")) continue;
+                    for (Card c : myCreatures) {
+                        if (c.hasKeyword(Keyword.DEFENDER) && !c.hasSickness()) return true;
+                    }
+                }
+
+                // 3+ cards match `S:Mode\$ IgnoreHexproof`
+                // Glaring Spotlight etc
+                if (mode.contains(StaticAbilityMode.IgnoreHexproof)) {
+                    return true;
+                }
+
+                // Everlasting Torment
+                if (mode.contains(StaticAbilityMode.WitherDamage)) {
+                    return true;
+                }
+            }  // static abilities
+
+            // 34+ cards match `K:Exalted`
+            if (cardState.hasKeyword(Keyword.EXALTED)) {
+                return true;
+            }
+
+            // 26+ cards match `K:Backup`
+            // cast Backup creatures in main 1 to pump attackers
+            if (cardState.hasKeyword(Keyword.BACKUP)) {
+                for (Card potentialAtkr: myCreatures) {
+                    if (ComputerUtilCard.doesCreatureAttackAI(ai, potentialAtkr)) {
+                        return true;
+                    }
+                }
+            }
+        } // boosts other creature
+
+        // Triggered abilities
+        for (Trigger trigger : cardState.getTriggers()) {
+            if (!"Battlefield".equals(trigger.getParam("TriggerZones"))) continue;
+
+            // 303+ cards match `Phase\$ BeginCombat.*?ValidPlayer\$(?![^|]*Opponent)`
+            // Odric, Lunarch Marshal etc. Ouroboroid is useful alone
+            final String phase = trigger.getParam("Phase");
+            if ("BeginCombat".equals(phase)) {
+                String validPlayer = trigger.getParam("ValidPlayer");
+                if (validPlayer == null || !validPlayer.contains("Opponent")) {
+                    return true;
+                }
+            }
+
+            if (myCreatures.size() < 1) continue;
+
+            final TriggerType mode = trigger.getMode();
+            if (mode != null) {
+                // 658+ cards match `Mode\$ ChangesZone.*?ValidCards?\$[^|]*Creature.*?TriggerZones\$ Battlefield`
+                // death triggers
+                // Al Bhed Salvagers; Dennick, Pious Apparition etc
+                if ("Graveyard".equals(trigger.getParam("Destination"))
+                    && Set.of("Any", "Battlefield").contains(trigger.getParamOrDefault("Origin", ""))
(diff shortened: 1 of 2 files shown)


forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1165,42 +1177,244 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
...
+            if (mode.contains(StaticAbilityMode.WitherDamage)) {
+                return true;
+            }
+        }  // static abilities
+
+        // Activated abilities
+        for (SpellAbility nma : cardState.getNonManaAbilities()) {
+            // 1710+ cards match `A:AB\$.*Cost\$[^|]*Sac`
+            // sac as combat trick
+            final String cost = nma.getParam("Cost");
```
Hanmac: Wouldn't it be better to check the real Cost Object?

You might want to check if the Sac Cost also Contains the Tap Cost (unless the creature would have haste)

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1165,42 +1177,244 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
...
         }
 
-        // cast Backup creatures in main 1 to pump attackers
-        if (cardState.hasKeyword(Keyword.BACKUP)) {
-            for (Card potentialAtkr: ai.getCreaturesInPlay()) {
-                if (ComputerUtilCard.doesCreatureAttackAI(ai, potentialAtkr)) {
+        // Triggered abilities
+        for (Trigger trigger : cardState.getTriggers()) {
+            final String triggerZones = trigger.getParam("TriggerZones");
+            if (!"Battlefield".equals(triggerZones)) continue;
```
Hanmac: not all Trigger has TriggerZones set (unless you want to exclude ETB and LTB on purpose)

like for ChangedZone Trigger, it is set there, to the `getActiveZone` set:
https://github.com/Card-Forge/forge/blob/0abc224e1bfc7507c3f1e9d97872e9600629566b/forge-game/src/main/java/forge/game/trigger/TriggerChangesZone.java#L216

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1165,42 +1177,244 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
...
+
+            final TriggerType mode = trigger.getMode();
+            if (mode != null) {
+                // 1312+ cards match `Mode\$ ChangesZone.*?TriggerZones\$ Battlefield`
+                // enter/leave triggers
+                // Admonition Angel etc
+                if (mode == TriggerType.ChangesZone || mode == TriggerType.ChangesZoneAll) {
```
Hanmac: Not all ETB trigger would be positive

Like `Hunted Bonebrute` would create tokens under opponents control, and would be better to be played after Combat?

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1152,6 +1152,18 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
         final Card card = sa.getHostCard();
         final CardState cardState = card.isFaceDown() ? card.getState(CardStateName.Original) : card.getCurrentState();
 
+        // 545+ cards match `K:Flash\n`
+        // save for later
+        if (cardState.hasKeyword(Keyword.FLASH)) {
```
tool4ever: too generic imo:
if you could improve your boardstate for a better combat then flash shouldn't always stop you

forge-ai/src/main/java/forge/ai/ComputerUtil.java
tool4ever: too generic, besides `PermanentCreatureAi` already checks `hasETBTrigger()`
if you're worried about Landfall try improving the logic around `PlayBeforeLandDrop`

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1165,42 +1177,244 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
...
+                // 1312+ cards match `Mode\$ ChangesZone.*?TriggerZones\$ Battlefield`
+                // enter/leave triggers
+                // Admonition Angel etc
+                if (mode == TriggerType.ChangesZone || mode == TriggerType.ChangesZoneAll) {
+                    return true;
+                }
+
+                // 1009+ cards match `Mode\$ SpellCast.*?ValidActivatingPlayer\$ You`
+                // Aetherflux Reservoir etc
+                if (mode == TriggerType.SpellCast && "You".equals(trigger.getParam("ValidActivatingPlayer"))) {
```
tool4ever: sequence wise this function should decide if there is something important enough to cast before combat (vs. leaving mana open for combat tricks)
_Aetherflux Reservoir_ does not qualify, it should just be cast before other spells ideally so that's more of a use case for `AIPriorityModifier`

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1165,42 +1177,244 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
...
+            if (validCard == null) validCard = "";
+            // 26+ cards match `S:Mode\$ CantBlock.*ValidCard\$(?![^|]*(You|Self))`
+            // Bedlam; Heat Wave etc
+            if (mode.contains(StaticAbilityMode.CantBlock) || mode.contains(StaticAbilityMode.CantBlockUnless)) {
+                if (!validCard.contains("You") && !validCard.contains("Self")) return true;
+            }
+
+            // 20+ card match `S:Mode\$ CombatDamageToughness(?!.*Self)`
+            // Ancient Lumberknot etc
+            if (mode.contains(StaticAbilityMode.CombatDamageToughness)) {
```
tool4ever: too situational imo

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1165,42 +1177,244 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
...
+                if (!validCard.contains("Self")) return true;
+            }
+
+            // 18+ cards match `S:Mode\$ CantGainLife.*ValidPlayer\$[^|]*(?!You)`
+            // Archfiend of Despair etc
+            if (mode.contains(StaticAbilityMode.CantGainLife)) {
```
tool4ever: too inaccurate:
at least check if opp also has untapped Lifelinkers

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1165,42 +1177,244 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
...
+            // 16+ cards match `S:Mode\$ CantBlockBy.*ValidAttacker\$[^|]*You`
+            // Bower Passage etc
+            if (mode.contains(StaticAbilityMode.CantBlockBy)) {
+                final String validAttacker = sta.getParam("ValidAttacker");
+                if (validAttacker != null && validAttacker.contains("You")) return true;
+            }
+
+            // 9+ cards match `S:Mode\$ IgnoreLegendRule`
+            // allow cloning your commander with Mirror Gallery; Myriad etc
+            if (mode.contains(StaticAbilityMode.IgnoreLegendRule)) {
```
tool4ever: too situational as default:
would at least need to check you have no non-legendary

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1165,42 +1177,244 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
...
+            // Everlasting Torment etc
+            if (mode.contains(StaticAbilityMode.CantPreventDamage)) {
+                final String validSource = sta.getParam("ValidSource");
+                if (!"Card.Self".equals(validSource)) return true;
+            }
+
+            // 4+ cards match `S:Mode\$ CanAttackDefender.*ValidCard\$[^|]*You`
+            // Guardians of Oboro etc
+            if (mode.contains(StaticAbilityMode.CanAttackDefender)) {
+                if (validCard.contains("You")) return true;
```
tool4ever: too inaccurate

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1165,42 +1177,244 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
...
+            // 4+ cards match `S:Mode\$ NoCleanupDamage`
+            // Patient Zero etc
+            if (mode.contains(StaticAbilityMode.NoCleanupDamage)) {
+                if (!"Card.Self".equals(validCard)) return true;
+            }
+
+            // 3+ cards match `S:Mode\$ ActivateAbilityAsIfHaste.*ValidCard\$[^|]*You`
+            // Thousand-Year Elixir etc
+            if (mode.contains(StaticAbilityMode.ActivateAbilityAsIfHaste)) {
+                if (validCard.contains("You")) return true;
```
tool4ever: too inaccurate

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: I think a human deck creator would have anticipated that, but the AI is probably less discerning. I'll see if there are more negative than positive effects or the negative ones can be filtered out.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: BeginCombat triggers can be responded to with Flash. Are instants not considered before declaring attacks?

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: Combat damage modifiers should not be played before combat? I see below checking for a creature that can attack and has greater T than P could suffice.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1165,42 +1177,244 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
...
+                    if (nmaCost == null || !nmaCost.contains("T")) {
+                        return true;
+                    }
+                    continue;
+                }
+            }
+            
+            // 43+ cards match `K:Bestow`
+            final String bestow = nma.getParam("Bestow");
+            if (bestow != null && bestow.equals("True")) {
```
tool4ever: what's the point of casting in main1 just because it has Bestow?

forge-ai/src/main/java/forge/ai/ComputerUtil.java
tool4ever: _Ancient Lumberknot_ is always a potential buff but _Belligerent Brontodon_ isn't

forge-ai/src/main/java/forge/ai/ComputerUtil.java
tool4ever: you can test, but I don't think AI would currently flash them in there

forge-ai/src/main/java/forge/ai/ComputerUtil.java
Hanmac: Check if the AI would play the Triggered Ability if it wasn't mandatory
this should filter out the bad ones

(see `getDamageFromETB` and `checkETBEffects`)

Also see `AiController.doTrigger`

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1165,42 +1177,244 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
...
+            if (mode.contains(StaticAbilityMode.Continuous)) {
+                final String affected = sta.getParam("Affected");
+                if (affected != null && affected.contains("YouCtrl")) {
+                    return true;
+                }
+            }
+
+            // 201+ cards match `S:Mode\$ ReduceCost.*?Activator\$ You`
+            // reduce cost to enable more plays
+            if (mode.contains(StaticAbilityMode.ReduceCost) && "You".equals(sta.getParam("Activator"))) {
```
tool4ever: again not worth the risk:
ai might just waste mana before combat

maybe they should get a slight boost in `getSpellAbilityPriority()` though (at least during early game)

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: `forge\forge-ai\src\main\java\forge\ai\ability\PermanentCreatureAi.java` `checkPhaseRestrictions` already appears to save Flash cards when needed.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: Afaik a permanent with TriggerZones "Battlefield" catches any _other_ card entering or leaving the battlefield, and cards with ETBs are handled elsewhere.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: `Hunted Bonebrute` matches `Mode\$ ChangesZone.*?(?!TriggerZones\$ Battlefield)` and such ETBs are handled elsewhere according to https://github.com/Card-Forge/forge/pull/8590#discussion_r2313352737

CTimmerman: https://github.com/Card-Forge/forge/pull/8590#discussion_r2314541603

CTimmerman: `PlayBeforeLandDrop` still makes sure to drop land later in main1.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: Spellslinger decks run such and prowess, right? So why not play _Aetherflux Reservoir_ in main1?

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: _Ancient Lumberknot_ says "Each creature you control with toughness greater than its power assigns combat damage equal to its toughness rather than its power."

_Belligerent Brontodon_ says "Each creature you control assigns combat damage equal to its toughness rather than its power."

I'll add the P/T check to make those cards play better in decks not designed around their ability.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: Only 3 cards match. Why put any in the deck if they're not useful? `card.isAbilitySick()` checks for this very ability.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: Because Bestow is like an aura, and auras are already played in main1 because they boost creatures.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: Mana rocks are typically played first, and cost reduction should be even better.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: _Etherium Sculptor_ is not legendary. Your suggested addition breaks this combo: https://www.youtube.com/shorts/KlYmlTzA9kw?feature=share

forge-ai/src/main/java/forge/ai/ComputerUtil.java
tool4ever: then you should test it
your logic does nothing because it's gonna be treated as Attach spell

forge-ai/src/main/java/forge/ai/ComputerUtil.java
tool4ever: you aren't understanding what I'm saying:
this function is for "main1 before main2"
hijacking it to add your "permanent A before B" logic just isn't supposed to happen here

forge-ai/src/main/java/forge/ai/ComputerUtil.java
tool4ever: that's nice but we're not adding very broad checks in general AI logic in the hopes of maybe playing a deck with some relevant combo...
FWIW allow me to demonstrate how you're managing to break very basic AI behaviour:
grafik
in case it's not clear enough - thanks to your change AI would prefer to lose from attack next turn instead of grabbing the win so _Mirror Gallery_ can look threatening 💩

it's up to you to ensure you're not making too many such assumptions, otherwise your changes aren't mergeable

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: Attach spells are also permanents, but I see Bestow is already checked in `attachAIPumpPreference`.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: What do you mean? Don't you think `Alisaie Leveilleur`, and `Jet Medallion` are worth playing in M1?

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: Thanks for the example. Can't that be solved by `getSpellAbilityPriority`?

forge-ai/src/main/java/forge/ai/ComputerUtil.java
tool4ever: no, _Jet Medallion_ does not help your combat directly
you'd need to at least be able to afford two additional playMain1 spells just to reach break even on mana spent vs. the risk of tapping out too early
this means on average it's a bad play

CTimmerman: Ok, but _Auriok Steelshaper_ would help equip and is a creature, and _Ballyrush Banneret_ is a creature as well.

Is spending 1W in M1 worth a body and a 1 discount on Kithkin and Soldier spells? Deck-dependent.

_Blood Funnel_ can pay for itself in monoblack, but requires creatures, which is probably checked when deciding to play it. (as well as it drawbacks). _Fluctuator_ and _Catalyst Stone_ also pay for themselves in the right decks. But all these have `AI:RemoveDeck:Random`

I'll leave ReduceCost to the card's `SVar:PlayMain1`.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
CTimmerman: Is _Ajani's Chosen_ worth playing in M1? _Ajani's Welcome_ is marked as such, maybe because it's only 1 mana.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1165,42 +1171,240 @@ public static boolean castPermanentInMain1(final Player ai, final SpellAbility s
...
-        // cast Backup creatures in main 1 to pump attackers
-        if (cardState.hasKeyword(Keyword.BACKUP)) {
+        // Activated abilities
+        for (SpellAbility nma : cardState.getNonManaAbilities()) {
+            final Cost cost = nma.getPayCosts();
+            if (cost.hasSpecificCostType(CostTap.class) && card.isAbilitySick()) continue;
+
+            // 1715+ cards match `A:AB\$.*Cost\$[^|]*Sac`
+            // sac as combat trick
+            if (cost.hasSpecificCostType(CostSacrifice.class)) {
```
tool4ever: this one is still too generic

but the PR is moving in the right direction 👍

CTimmerman: Thanks. I noticed it playing Sakura-Tribe Elder in M1 now, and immediately sacrificing it for land (which it also did in M2 and i couldn't find why) but that could boost landfall etc.

forge-ai/src/main/java/forge/ai/ComputerUtilAbility.java
```diff
@@ -349,6 +354,30 @@ private static int getSpellAbilityPriority(SpellAbility sa) {
                 if (source.isInPlay() && source.hasSVar("EndOfTurnLeavePlay")) {
                     p += 1;
                 }
+
+                // Triggered abilities
+                for (Trigger trigger : source.getTriggers()) {
+                    if (!"Battlefield".equals(trigger.getParam("TriggerZones"))) continue;
```
tool4ever: use `zonesCheck`
also mode can't be null anyway

forge-ai/src/main/java/forge/ai/ComputerUtilAbility.java
```diff
@@ -349,6 +354,30 @@ private static int getSpellAbilityPriority(SpellAbility sa) {
...
+                    final TriggerType mode = trigger.getMode();
+                    if (mode != null) {
+                        // 1024+ cards match `Mode\$ SpellCast.*?ValidActivatingPlayer\$ You`
+                        // Aetherflux Reservoir etc
+                        if (mode == TriggerType.SpellCast && "You".equals(sa.getParam("ValidActivatingPlayer"))) {
```
tool4ever: should also include ``SpellCastOrCopy`` because of Magecraft

forge-ai/src/main/java/forge/ai/ComputerUtilAbility.java
CTimmerman: `if (!trigger.zonesCheck(Zone.Battlefield)) continue;` doesn't work. Feel free to take over.

tool4ever: I've cherry picked this file via https://github.com/Card-Forge/forge/pull/9210/commits/51e10e8f2b1d86056d42725f72f4cb52ddaaafa9

CTimmerman:
For some reason the implementation of contains that's used here does not check for null.
image

CTimmerman:
GitLab for one leaves out main/master commits in its compare feature for rebased versions. Finally a thing it's better at than GitHub.

Anyway, this version appears to work fine for me.

CTimmerman:
What's keeping this from getting merged?

tool4ever:
It took almost 50 comments to even reach this state and yet it still looks messy in some parts to me:
sorry but I know further reviewing will almost certainly be time I'd rather spent on another PR, so personally I'm saving this for a (very) rainy day 🤷‍♂️

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -1141,56 +1141,274 @@ public static boolean waitForBlocking(final SpellAbility sa) {
...
+        } // boosts other creature
+
+        // Triggered abilities
+        for (Trigger trigger : cardState.getTriggers()) {
+            if (!"Battlefield".equals(trigger.getParam("TriggerZones"))) continue;
+
+            // 303+ cards match `Phase\$ BeginCombat.*?ValidPlayer\$(?![^|]*Opponent)`
+            // Odric, Lunarch Marshal etc. Ouroboroid is useful alone
+            final String phase = trigger.getParam("Phase");
+            if ("BeginCombat".equals(phase)) {
```
tool4ever, line 1337: ok, I briefly checked if this could be the candidate for the next cherry picking but I estimate there'd be at least 50 cards where this might produce a false positive (invalid conditions etc.) 👎

if it's possible to replace it you'll probably have to evaluate the trigger and its API in a much deeper fashion for the AI opinion

then you can also cleanup the SVar from the covered script cases so it doesn't get copied over and over onto new cards again...

Issue #8513, opened 2025-08-23
AI spends 15+ minutes on doubling counters


CTimmerman:
The accepted patch for #7814 fixed the program freeze and OOM crash, but it's still taking 15 minutes to double 16 million counters on Heroes' Bane despite this gaming laptop having 12x 2.6 GHz cores running at 3.5 Ghz.

<img width="1680" height="1050" alt="Image" src="https://github.com/user-attachments/assets/0a7f40c1-3043-4db1-bda9-878161499e2a" />

<img width="1680" height="1050" alt="Image" src="https://github.com/user-attachments/assets/1b6e50f3-980e-407d-9622-938ac837db1f" />

I'll leave it running over night, but my last patch would limit it to 999_999 counters which is enough for me.

Hanmac:
At this point, I would say "deal with it"

I would suggest running it through a Profiler to check which part of the AI is hanging the most

I assume it is still somewhere in the Trigger handling that causes this 


What happens next? Are you complaining that the Limit for counters is **Int** instead of **Long**?

CTimmerman:
One night later and it's still using 50% of 12x 3.7 Ghz. Games should not last this long. My patch #7828 fixes this by limiting counters to 999999 as noted multiple times already.

<img width="1680" height="1050" alt="Image" src="https://github.com/user-attachments/assets/00bf42ba-4486-402a-b2b1-33ed5d8f442c" />

Hanmac:
Like I said, the information is useless if you didn't run a Profiler to check which part takes the most cpu time

CTimmerman:
Please point me to a VS Code profiler for Java then. Also note that without a limit, even a 100x perf boot is not enough for this one simple card with the current braindead AI (that my patch would fix). Also, the GUI can't even display the current P/T. Do you want me to open an issue for that even though I don't care about such high numbers?

<img width="1680" height="1050" alt="Image" src="https://github.com/user-attachments/assets/3eb25d0d-89c2-4998-a847-fa0d74ded0d0" />

Hanmac:
> current braindead AI

you assume the problem is the AI, but you can't prove it.
I assume its mostly the Engine itself that can't keep up.

i hope you notice that 1073741824 is nearly half way through 2147483647, which is `Integer.MAX_VALUE`

> Please point me to a VS Code profiler for Java then. 

why do i need to help you find a tool when you could google it yourself?

for Eclipse i found this Memory Tool
https://eclipse.dev/mat/

[...]

PR #8601, opened 2025-09-01
CountersPutAi: prevent AI from going overflow with Hero's Bane

Hanmac:
related to #8513

first small fix to prevent AI from using Hero's Bane to overflow on Counters and on PT
(might be expanded later)

i will also later add limits in Card and GameEntity to prevent it from overflowing,
even if AI would activate it.

@CTimmerman can you check if that fixes the Hero's Bane Problem?
(the final one that caused the overflow?)


CTimmerman:
It should, but this patch doesn't cover all overflows caused by counters as listed in `forge\forge-game\src\main\java\forge\game\card\CounterEnumType.java` (unlike my patch which also keeps the P/T in visual rage for a 4 player GUI at 1080p, but a 999_999 counter limit might be too low to max out with a hand full of doubling effects).

Hanmac:
> It should, but this patch doesn't cover all overflows caused by counters as listed in `forge\forge-game\src\main\java\forge\game\card\CounterEnumType.java` (unlike my patch which also keeps the P/T in visual rage for a 4 player GUI at 1080p, but a 999_999 counter limit might be too low to max out with a hand full of doubling effects).

It's first small piece 
The real overflow fix will be later in the Engine itself

This part is only for fixing Hero's Bane (that AI would still activate)

CTimmerman:
Adding 536+e6 counters was slow but this game ended in less than 10 minutes:

image

image

test_HeroesBane.dck:
```
[metadata]
Name=test_HeroesBane
[Main]
4 Creepy Doll|ISD|[220]
18 Forest|ELD|[268]
4 Heroes' Bane|PJOU|[126★]
4 Pariah's Shield|RAV|[267]
```

PR #8603, opened 2025-09-02
Fix missing keys: t:speed and t:max_speed (#8602)

### LINKED ISSUE #8602: Missing keys: t:speed and t:max_speed

CTimmerman:
Closes #8602.

### DIFF
--- a/forge-game/src/main/java/forge/game/player/Player.java
+++ b/forge-game/src/main/java/forge/game/player/Player.java
@@ -1966,10 +1966,10 @@ public class Player extends GameEntity implements Comparable<Player> {
         CardState speedFront = speedEffect.getState(CardStateName.Original);
         CardState speedBack = speedEffect.getState(CardStateName.Backside);
 
-        speedFront.setImageKey("t:speed");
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -1966,10 +1966,10 @@ public final void createSpeedEffect() {
         CardState speedFront = speedEffect.getState(CardStateName.Original);
         CardState speedBack = speedEffect.getState(CardStateName.Backside);
 
-        speedFront.setImageKey("t:speed");
+        speedFront.setImageKey("t:speed|DFT|14");
```
tool4ever, line 1969: why can't we just define this in the edition file - like all other markers?

CTimmerman: Maybe because this is the only place `t:speed` appears in the entire repo?

PR #8615, opened 2025-09-02
Mayhem for Lands

tool4ever:


Hanmac:
Probably need to check it with 'no cost'

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -2776,10 +2776,9 @@ public static void addSpellAbility(final KeywordInterface inst, final CardState
             final String cost = params[1];
 
             final StringBuilder sbAttach = new StringBuilder();
-            sbAttach.append("SP$ Attach | Cost$ ");
+            sbAttach.append("SP$ Attach | ValidTgts$ Creature | Cost$ ");
             sbAttach.append(cost);
             sbAttach.append(" | AILogic$ ").append(params.length > 2 ? params[2] : "Pump");
-            sbAttach.append(" | Bestow$ True | ValidTgts$ Creature");
```
Hanmac, line 2782: Is the "Bestow$" flag not needed anymore?

tool4ever: if it ever was checked somewhere instead of better just using KW/AltCost it seems to be refactored already 🤷‍♂️

Hanmac:
For the land, I still need to fix the Reminder Text of Mayhem without a Cost:

`Mayhem (You may play this card from your graveyard if you discarded it this turn. Timing rules still apply.)`

it will then crash there:
https://github.com/Card-Forge/forge/blob/5234365cc1c285f4d8ce9b74df993097812f34d2/forge-game/src/main/java/forge/game/card/Card.java#L3335-L3337

tool4ever:
Unless the land gains instant/sorcery type I don't see how that's possible

Hanmac:
> Unless the land gains instant/sorcery type I don't see how that's possible

ah ok i found the other section

https://github.com/Card-Forge/forge/blob/5234365cc1c285f4d8ce9b74df993097812f34d2/forge-game/src/main/java/forge/game/card/Card.java#L2475-L2478

but it doesn't have Reminder text there 🤔 

i might fix that later

Hanmac:
@tool4ever this can be merged too?
i will tweak the Mayhem reminder later

PR #8619, opened 2025-09-03
Fix Arachnus Spinner not choosing a target for the Web

Jetz72:
Fix contributed by Atticus Von Schwoll on Discord ([thread](https://discord.com/channels/267367946135928833/1412578856807825591)).


Hanmac:
@Jetz72 @tool4ever we should try if it can be done in one line

Just adding the `AttachedTo$ ThisTargetedCard` might work?

Hanmac:
@Jetz72 i will try to make it better later

but the SpellDescription probably needs to be moved up a bit

PR #8622, opened 2025-09-03
Optional fixes

tool4ever:


forge-gui/res/cardsfolder/t/talus_paladin.txt
```diff
@@ -4,7 +4,7 @@ Types:Creature Human Knight Ally
 PT:2/3
 T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self,Ally.Other+YouCtrl | OptionalDecider$ You | Execute$ TrigPumpAll | TriggerDescription$ Whenever CARDNAME or another Ally you control enters, you may have Allies you control gain lifelink until end of turn, and you may put a +1/+1 counter on CARDNAME.
 SVar:TrigPumpAll:DB$ PumpAll | ValidCards$ Ally.YouCtrl | KW$ Lifelink | SubAbility$ DBCounter
-SVar:DBCounter:DB$ PutCounter | Defined$ Self | CounterType$ P1P1 | CounterNum$ 1 | OptionalDecider$ You
+SVar:DBCounter:DB$ PutCounter | Defined$ Self | CounterType$ P1P1 | CounterNum$ 1 | Optional$ True
```
Hanmac: >  The choices you make as Talus Paladin’s ability resolves are made individually. First you decide whether to have Allies you control gain lifelink until end of turn. Then, regardless of that decision, you decide whether to put a +1/+1 counter on Talus Paladin. 

You should be able to choose just the +1/+1 counter

so the Pump needs to be Optional too (and maybe the `OptionalDecider$ You` from the trigger removed)

PR #8626, opened 2025-09-03
Black Cat and gang (SPM)

Fulgur14:


forge-gui/res/cardsfolder/upcoming/spider_verse.txt
```diff
@@ -0,0 +1,9 @@
+Name:Spider-Verse
+ManaCost:3 R R
+Types:Enchantment
+S:Mode$ IgnoreLegendRule | ValidCard$ Spider.YouCtrl | Description$ The "legend rule" doesn't apply to Spiders you control.
+T:Mode$ SpellCast | ValidCard$ Card.!wasCastFromYourHand | ValidActivatingPlayer$ You | TriggerZones$ Battlefield | Execute$ TrigCopy | OptionalDecider$ You | TriggerDescription$ Whenever you cast a spell from anywhere other than your hand, you may copy it. If you do, you may choose new targets for the copy. If the copy is a permanent spell, it gains haste. Do this only once each turn.
```
tool4ever: ResolvedLimit missing

forge-gui/res/cardsfolder/upcoming/shadow_of_the_goblin.txt
```diff
@@ -0,0 +1,12 @@
+Name:Shadow of the Goblin
+ManaCost:1 R
+Types:Enchantment
+T:Mode$ Phase | Phase$ Main1 | ValidPlayer$ You | TriggerZones$ Battlefield | Execute$ TrigDiscard | TriggerDescription$ Unreliable Visions — At the beginning of your first main phase, discard a card. If you do, draw a card.
+SVar:TrigDiscard:DB$ Discard | Defined$ You | NumCards$ 1 | Mode$ TgtChoose | RememberDiscarded$ True | SubAbility$ DBDraw
```
tool4ever, line 5: Cost is cleaner

Fulgur14: But this is mandatory discard.

forge-gui/res/cardsfolder/upcoming/mob_lookout.txt
```diff
@@ -0,0 +1,7 @@
+Name:Mob Lookout
+ManaCost:1 UB
+Types:Creature Human Rogue Villain
+PT:0/3
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigConnive | TriggerDescription$ When CARDNAME enters, draw a card.
```
tool4ever: wrong description

forge-gui/res/cardsfolder/upcoming/mob_lookout.txt
```diff
@@ -0,0 +1,7 @@
...
+SVar:TrigConnive:DB$ Connive | Defined$ Targeted | ValidTgts$ Creature.YouCtrl | TgtPrompt$ Select target creature you control
```
tool4ever: why Defined?

Fulgur14: That's how it's written in Spymaster's Vault.

tool4ever: it's just bloat, you know right?

Fulgur14: Well, I have problem recognizing that in general :) 95% of card scripting is just breaking down what was there before and reusing it, and I think I am not bad with that. But if the template I choose is wrong, I might not recognize that. The way I work is sometimes closer to AI algorithm than to what a normal human would do.

tool4ever: ugh I would really like that to be a much lower number, though that doesn't mean you haven't recognized some good corner cases already in the past :)

in short:
all common effects should support either Defined$ or ValidTgts$ depending on what the card needs (barring the \*All variants which are often limited to Valid*$ selection)
it rarely makes sense to use both together unless the effect is part of some rather advanced chain

I'm kind of waiting on getting my wiki PR merged since it'd allow me an easier workflow when improving the API docs for such stuff

Fulgur14: There are docs?

tool4ever: I think if you read them now you'd just confuse yourself lol :P

Fulgur14: Yeah, I learned my actual job the same way, just by assimilation of examples :D

forge-gui/res/cardsfolder/upcoming/shadow_of_the_goblin.txt
tool4ever: ok, no big deal right now

PR #8633, opened 2025-09-04
Fix CollNo pattern after #7921

Eradev:
Found it after crashing with "give print" command.


Hanmac (approved):

Hanmac:
@Eradev maybe, later these days, we could change the patterns to use group names?

PR #8637, opened 2025-09-04
Sandman's Quicksand (SPM) -- needs support

Fulgur14:


Hanmac:
I would have added: "Mayhem" there

https://github.com/Card-Forge/forge/blob/11d45e0fa318a315468aa73a92906946bbb5c24e/forge-game/src/main/java/forge/game/ability/AbilityUtils.java#L1723-L1726

and imo it would be enough to rename the checks like `isMayhem` similar to `isMadness` in Card in SpellAbility

Hanmac:
@tool4ever your opinion on the Card property name?

Or should we skip it and use `CastSA Spell.Mayhem` instead?
(imo fewer properties might be better?)

tool4ever:
yea since we really track it via the `SpellAbility` field it doesn't make sense to have the calls chained more than needed

Hanmac:
hm, the AbilityUtils one can be added later if needed

tool4ever (approved):

PR #8648, opened 2025-09-05
Final SPM batch

Fulgur14:


forge-gui/res/cardsfolder/upcoming/mister_negative.txt
```diff
@@ -0,0 +1,12 @@
+Name:Mister Negative
+ManaCost:5 W B
+Types:Legendary Creature Human Villain
+PT:5/5
+K:Vigilance
+K:Lifelink
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ StoreLife | OptionalDecider$ You | TriggerDescription$ Darkforce Inversion — When CARDNAME enters, you may exchange life totals with target opponent. If you lost life this way, draw that many cards.
+SVar:StoreLife:DB$ StoreSVar | SVar$ PreviousLife | Type$ Calculate | Expression$ Count$YourLifeTotal | SubAbility$ DBExchangeLife
+SVar:ExchangeLife:DB$ TrigExchangeLife | Optional$ True | ValidTgts$ Opponent | TgtPrompt$ Select target opponent to exchange life totals with | SubAbility$ DBDraw
```
tool4ever: bad API

tool4ever: Optional$ doesn't exist

Fulgur14: It's in Cliffside Market.

Fulgur14: I... don't get what that means.

forge-gui/res/cardsfolder/upcoming/sinister_hideout.txt
```diff
@@ -0,0 +1,8 @@
+Name:Sinister Hideout
+ManaCost:no cost
+Types:Land
+R:Event$ Moved | ValidCard$ Card.Self | Destination$ Battlefield | ReplacementResult$ Updated | ReplaceWith$ ETBTapped | Description$ This land enters tapped.
+SVar:ETBTapped:DB$ Tap | Defined$ Self | ETB$ True
+A:AB$ Mana | Cost$ T | Produced$ Combo U B R | SpellDescription$ Add {U} or {B}.
```
Hanmac: wrong Combo Mana

forge-gui/res/cardsfolder/upcoming/mister_negative.txt
tool4ever: I guess sometimes your copy paste helps to clean up old cards :p

tool4ever: you should have seen an error when testing the card because of TrigExchangeLife

Fulgur14: BTW, is this solution with storing the best option?

tool4ever: #8650

Fulgur14: Hm, RememberDifference doesn't seem enough. If I understand it correctly, it computes difference between life totals of players after they are exchanged, but a simple lifegain doubling effect is enough to make opponent's life after exchange not equal to your life before exchange.

tool4ever: that's why I added RememberOwnLoss$

Fulgur14: Ah. I missed that.

PR #8650, opened 2025-09-05
Fix Profane Transfusion

tool4ever:
>  Use the difference between the life totals after the exchange to determine the size of the Horror token. Because replacement effects can modify the life gain and life loss, this number may be different from what it was before the exchange.


Hanmac:
@tool4ever looks good for that case

Can we do something with MR Negative too?

Hanmac (approved):

PR #8659, opened 2025-09-06
Fix AI sometimes tapping mana sources for 0

### LINKED ISSUE #5616: Alena Kessig Trapper is tapped for 0 mana when using "auto"
When I use "Auto" to pay for spells, the system always taps Alena Kessig trapper for mana, even when she only produces 0...

Should be easy to reproduce, it happens for me any time.
Just make sure she produces 0 mana and you cast a spell.

Additonal Note. now that I think about it: it seems it only happens when exactly all mana is spent on the spell. Coincidentially this was the case all the time this happened.

tool4ever:
Closes #5616

### DIFF
--- a/forge-game/src/main/java/forge/game/spellability/SpellAbility.java
+++ b/forge-game/src/main/java/forge/game/spellability/SpellAbility.java
@@ -354,7 +354,7 @@ public abstract class SpellAbility extends CardTraitBase implements ISpellAbilit
     public int totalAmountOfManaGenerated(SpellAbility saPaidFor, boolean multiply) {
         int result = 0;
         AbilityManaPart mp = getManaPart();
-        if (mp != null && metConditions() && mp.meetsManaRestrictions(saPaidFor)) {
+        if (mp != null && mp.meetsManaRestrictions(saPaidFor)) {
             result += amountOfManaGenerated(multiply);
         }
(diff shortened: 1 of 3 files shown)


forge-game/src/main/java/forge/game/spellability/SpellAbility.java
```diff
@@ -354,7 +354,7 @@ public int amountOfManaGenerated(boolean multiply) {
     public int totalAmountOfManaGenerated(SpellAbility saPaidFor, boolean multiply) {
         int result = 0;
         AbilityManaPart mp = getManaPart();
-        if (mp != null && metConditions() && mp.meetsManaRestrictions(saPaidFor)) {
```
Hanmac, line 357: this is for chained Mana Effects like:
* Effect A: condition=false
* Effect B: condition=true

tool4ever: the check is already inside `amountOfManaGenerated`
so it's redundant and only costs runtime

Hanmac: but not in `InputMana::activateManaAbility`

tool4ever: `activateManaAbility -> totalAmountOfManaGenerated -> amountOfManaGenerated -> metConditions`

nothing is skipped

Hanmac (approved):

PR #8676, opened 2025-09-08
Fix crash when equipping The Reaver Cleaver

Eradev:


xanxer6rB:
Thank you for catching that. How did the crash happen? 
Curious because I didn't get any game lock ups or java errors popping up when I was testing the changes.

Eradev:
> Thank you for catching that. How did the crash happen? Curious because I didn't get any game lock ups or java errors popping up when I was testing the changes.

Immediately when trying to equip.

Hanmac (approved):

PR #8681, opened 2025-09-09
Deck Importer support for Adventure, Quest, and Planar Conquest

Jetz72:
Adds support for importing decks from the clipboard in these modes, similar to constructed formats. The resulting deck will use only cards from the player's inventory. If a requested print is unavailable, the import will first try to coerce it to a different print of the same card. Otherwise, if it's a basic land, it'll be added freely similar to the Add Basic Lands dialog. If none of those fallbacks happen, the print will just be left missing and the user will be notified.

Options for preferred prints are not currently supported when using cards from the inventory. 

Adds support for showing specific per-card error messages when they fail to import, as the desktop importer does.

Fixes some issues with the editions supplied to the Add Basic Lands dialog, and adds support for them in Planar Conquest.

### DIFF
--- a/forge-core/src/main/java/forge/deck/DeckRecognizer.java
+++ b/forge-core/src/main/java/forge/deck/DeckRecognizer.java
@@ -536,7 +536,7 @@ public class DeckRecognizer {
             PaperCard tokenCard = token.getCard();
 
             if (isAllowed(tokenSection)) {
-                if (!tokenSection.equals(referenceDeckSectionInParsing)) {
+                if (tokenSection != referenceDeckSectionInParsing) {
                     Token sectionToken = Token.DeckSection(tokenSection.name(), this.allowedDeckSections);
                     // just check that last token is stack is a card placeholder.
                     // In that case, add the new section token before the placeholder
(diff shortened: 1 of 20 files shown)


Hanmac:
When touching the hasBasicLands function, can you get rid of this one? (or update it calling `ed.hasBasicLands()`?)
https://github.com/Card-Forge/forge/blob/ac9b73935f29804ba568cbad61ec011ebf95a634/forge-core/src/main/java/forge/card/CardEdition.java#L1021-L1032

Jetz72:
> When touching the hasBasicLands function, can you get rid of this one? (or update it calling `ed.hasBasicLands()`?)

I'll forward it to the method and deprecate it, but fully removing it will require verifying that that null check isn't needed.

forge-core/src/main/java/forge/deck/DeckRecognizer.java
```diff
@@ -1087,28 +1062,22 @@ private static Pair<String, String> getManaNameAndSymbol(String matchedMana) {
...
-            case MagicColor.Constant.GREEN:
-            case "g":
-                return Pair.of(localizer.getMessage("lblGreen"), MagicColor.Color.GREEN.getSymbol());
-            case MagicColor.Constant.COLORLESS:
-            case "c":
-                return Pair.of(localizer.getMessage("lblColorless"), MagicColor.Color.COLORLESS.getSymbol());
-            default: // Multicolour
-                return Pair.of(localizer.getMessage("lblMulticolor"), "");
-        }
+        return switch (matchedMana.toLowerCase()) {
```
Hanmac: getManaNameAndSymbol is a private method that isn't used
should we remove it?

Jetz72: Sure. Returning a pair is usually a sign some responsibility needs to be moved elsewhere anyway.

Hanmac:
Good by me, think?

@tehdiplomat @kevlahnota your opinions?

As for `getLocalisedMagicColorName`, I'm going to remove that in the future

PR #8687, opened 2025-09-10
Spm new set

tehdiplomat:


Hanmac:
@tehdiplomat probably need to rebase after conflicts are resolved

tehdiplomat:
@Hanmac rebased. I'll put upcoming migration in a different PR, and just push it through

Issue #8682, opened 2025-09-09
Similfy MagicColor stuff


Hanmac:
Right now, there are multiple ways of Colors being stored:

* as Byte (mutable)
* as String (short/long)
* as Enum (`MagicColor.Color`, not really used?)
* as ColorSet(immutable, Iterable Byte, but EnumSet too?)

CardFace in core uses ColorSet

CardState stores it as byte (to be changed by CloneState)
Same with CardColor (with is byte + flag)
but returns it as ColorSet after CardColor is applied to CardState one.
In Card, chosenColor it is stored as Set<String>

addColor is used for:
* CardState in CardFactory in CloneState (changes Byte)
* Card in Animate and StaticAbility (adds CardColor)

setColor is used for:
* CardState in CardFactory in CloneState (changes Byte)
* Creating Effects and Emblems (Emblems are colorless)

`PlayerController.chooseColors` uses List<String> (can't use ColorSet as parameter, because Colorless might be allowed too)

Functions like `getMostProminentColors` or `getColorByProminence` should return ColorSet or EnumColor, not byte or string
`getMostProminentColorsFromList` should not use `List<String>` or `Iterable<String>`

ChangeTextEffect for ChangeColorWord only uses "Choose Choose"
Might better to simplify the code?
Also would be better if it would create an Effect that uses StaticAbility Layers
Would be cool if i could change changedTextColors `Map<String,String>` to something using Enum

Produced Color should not be a String ala "B B B" anymore, but more like a `EnumMap<Color, Integer>` to prevent overflows

Jetz72:
Without digging into it to verify how this would affect the various systems, my vote would be for `MagicColor` being an `enum`, and `ColorSet` being an extension of `EnumSet`. Colorless can be null or an empty set, respectively. 

We could possibly add some helpers to bulk-process them using bytes and bitmask logic if it would substantially improve performance. But doing that in isolated cases can make code harder to read and work with. If someone unfamiliar with `MagicColor` and its conventions is working on code that uses color stored as a byte, it's not immediately clear where to find the functions for operating on and inspecting that color. Whereas dedicated classes can have helper methods, meaning they'd only have to type "`this.color.`" for the IDE to give them a list of relevant methods.

Hanmac:
> ColorSet being an extension of EnumSet

I don't know right now. ColorSet in the current state is immutable

> Colorless can be null or an empty set

[...]

PR #8690, opened 2025-09-10
MagicColor: use Enum for Human chooseColors

Hanmac:
Part of #8682

the smallest change for now to start

next part will be ChosenColors to be a ColorSet

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/ManaReflectedEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/ManaReflectedEffect.java
@@ -104,9 +104,9 @@ public class ManaReflectedEffect extends SpellAbilityEffect {
                 } else if (colors.size() == 1) {
                     baseMana = MagicColor.toShortString(colors.iterator().next());
                 } else if (colors.contains("colorless")) {
-                    baseMana = player.getController().chooseColorAllowColorless(Localizer.getInstance().getMessage("lblSelectManaProduce"), sa.getHostCard(), ColorSet.fromNames(colors))).getShortName();
+                    baseMana = player.getController().chooseColorAllowColorless(Localizer.getInstance().getMessage("lblSelectManaProduce"), sa.getHostCard(), ColorSet.fromNames(colors)).getShortName();
                 } else {
-                    baseMana = player.getController().chooseColor(Localizer.getInstance().getMessage("lblSelectManaProduce"), sa, ColorSet.fromNames(colors))).getShortName();
+                    baseMana = player.getController().chooseColor(Localizer.getInstance().getMessage("lblSelectManaProduce"), sa, ColorSet.fromNames(colors)).getShortName();
                 }
             } else {
                 colorMenu = ColorSet.fromMask(mask);


Hanmac:
Maybe for this MR too, i will change how 
`chooseColor` and `chooseColorAllowColorless` works using `ColorSet` to `EnumSet<Color>`

Hanmac:
@tool4ever your opinion on changes?

tool4ever:
Well currently to me it doesn't look like this makes the code actually easier to read in most of the affected parts and sometimes it's rather worse, e.g.: 
image

so I'm not sure how much benefit this actually brings 🤷‍♂️

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1833,53 +1832,51 @@ public List<AbilitySub> chooseModeForAbility(final SpellAbility sa, List<Ability
     @Override
     public List<String> chooseColors(final String message, final SpellAbility sa, final int min, final int max,
                                      List<String> options) {
-        options = options.stream().map(DeckRecognizer::getLocalisedMagicColorName).collect(Collectors.toList());
-        List<String> choices = getGui().getChoices(message, min, max, options);
-        return choices.stream().map(DeckRecognizer::getColorNameByLocalisedName).collect(Collectors.toList());
+        List<MagicColor.Color> enumOptions = options.stream().map(s -> MagicColor.Color.fromByte(MagicColor.fromName(s))).collect(Collectors.toList());
+        List<MagicColor.Color> enumChoices = getGui().getChoices(message, min, max, enumOptions);
```
tool4ever, line 1839: this almost certainly breaks netplay

Hanmac:
> Well currently to me it doesn't look like this makes the code actually easier to read in most of the affected parts
> so I'm not sure how much benefit this actually brings 🤷‍♂️

baby steps, I'm going to refactor them step by step

as for ChangeText, i might convert it into an Effect to make it cleaner

forge-core/src/main/java/forge/item/PaperCardPredicates.java
```diff
@@ -76,24 +76,20 @@ private PredicatePrintedWithRarity(final CardRarity rarity) {
...
-                }
+            if (card.getRules().getColor().toEnumSet().contains(operand)) {
+                return true;
             }
             if (card.getRules().getType().hasType(CardType.CoreType.Land)) {
-                for (final byte color : card.getRules().getColorIdentity()) {
-                    if (color == operand) {
-                        return true;
-                    }
+                if (card.getRules().getColorIdentity().toEnumSet().contains(operand)) {
```
tool4ever: what's wrong with iterating?
creating new Set from List objects each time seems like a pretty bad refactor performance wise when filtering larger PaperCard pools 👎

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
Hanmac: MagicColor.Color is an Enum with only String keys

i'm pretty sure that works with GUI, but @kevlahnota might test it

forge-core/src/main/java/forge/item/PaperCardPredicates.java
Hanmac: i wanted to remove `Iterable<byte>` in favor for `iterable<Enum>` in later future

tool4ever:
can this be closed or are there any party here left not already merged?

Hanmac:
> can this be closed or are there any party here left not already merged?

There are some changes, like `MagicColor.Color chooseColorAllowColorless` but I should do it with another MR instead of this one

PR #8694, opened 2025-09-11
fix Keyword Advanced Search

kevlahnota:


Jetz72:
We could probably improve performance substantially if that `getKeywordSet` method used EnumSets instead of HashSets. Though ideally we could find a way to store them that reuses the same instance for identical sets of keywords.

kevlahnota:
> We could probably improve performance substantially if that `getKeywordSet` method used EnumSets instead of HashSets. Though ideally we could find a way to store them that reuses the same instance for identical sets of keywords.

hmmm well the deck editor cares for the printed papercard values so I guess it wouldn't need to call the getCardForUi since we only need to compare printed values

PR #8714, opened 2025-09-15
Fix crash after removing target when copying divided spell

tool4ever:


Hanmac (approved):

forge-game/src/main/java/forge/game/ability/effects/ChangeZoneEffect.java
```diff
@@ -928,7 +928,7 @@ private void changeZonePlayerInvariant(Player chooser, SpellAbility sa, List<Pla
 
             List<ZoneType> origin = Lists.newArrayList();
             if (sa.hasParam("Origin")) {
-                origin = ZoneType.listValueOf(sa.getParam("Origin"));
+                origin.addAll(ZoneType.listValueOf(sa.getParam("Origin")));
```
Hanmac, line 931: Or:
`origin = sa.hasParam("Origin") ? ZoneType.listValueOf(sa.getParam("Origin")) : Lists.newArrayList()`

tool4ever: this actually also prevents corner case crash because `listValueOf` returns immutable (sideboard removal when controlled)

Hanmac: Ah, I didn't notice the remove later

PR #8715, opened 2025-09-15
MagicColor.Color adds ShortName and Symbol

Hanmac:
Adds ShortName and Symbol

Small Piece of #8690 that can be implemented early


Jetz72:
Does changing language currently require a restart? Writing a localized string to a constant would prevent it from being updated later.

Hanmac:
> Does changing language currently require a restart? Writing a localized string to a constant would prevent it from being updated later.

Language used for GUI (that includes this enums) requires a restart

PR #8719, opened 2025-09-16
update AltZoneLayout VCardDisplayArea for Landscape

kevlahnota:
Screenshots also enabled Alt Player Layout
image

- Off (Default Layout)
image

- Vertical (Alt Layout with old Vertical Scroll Card Display Area)
image

- Horizontal (Alt Layout with Horizontal Scroll Card Display Area)
image

- fix mouse scroll to scroll horizontall while ShiftKey is pressed

### DIFF
--- a/forge-gui-mobile/src/forge/screens/match/views/VManaPool.java
+++ b/forge-gui-mobile/src/forge/screens/match/views/VManaPool.java
@@ -69,7 +69,7 @@ public class VManaPool extends VDisplayArea {
         float x = 0;
         float y = 0;
 
-        if (Forge.isLandscapeMode()) {
+        if (Forge.isLandscapeMode() && !Forge.altZoneTabs) {
             float labelWidth = visibleWidth / 2;
             float labelHeight = visibleHeight / 3;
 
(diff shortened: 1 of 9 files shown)


forge-gui-mobile/src/forge/screens/match/views/VManaPool.java
```diff
@@ -146,17 +146,19 @@ public void draw(Graphics g) {
...
             }
-            float w = image.getWidth() * h / image.getHeight();
+            float w2 = Forge.altZoneTabs && "Horizontal".equalsIgnoreCase(Forge.altZoneTabMode)
+                ? image.getWidth() * h * 0.7f / image.getHeight() : image.getWidth() * h / image.getHeight();
+            float w =  w2;
             while (w > getWidth()) {
                 h /= 2;
-                w = image.getWidth() * h / image.getHeight();
+                w = w2;
             }
```
Jetz72, line 155: I'm pretty sure this condition will never change. `w` will always be the same and `w2` is only defined outside it.

PR #8720, opened 2025-09-16
ColorSet: store EnumSet and toString as Final

Hanmac:
part of #8682

ColorSet doesn't need ManaCostShard

the EnumSet and the String representation are created in constructor and marked as final

### DIFF
--- a/forge-core/src/main/java/forge/card/ColorSet.java
+++ b/forge-core/src/main/java/forge/card/ColorSet.java
@@ -112,7 +112,7 @@ public final class ColorSet implements Comparable<ColorSet>, Iterable<Byte>, Ser
 
     private ColorSet(Collection<Color> ordered) {
         this.orderedShards = ordered;
-        this.myColor = ordered.stream().map(Color::getColorMask).reduce((byte)0, (a, b) -> a | b);
+        this.myColor = ordered.stream().map(Color::getColorMask).reduce((byte)0, (a, b) -> (byte)(a | b));
         this.orderWeight = this.getOrderWeight();
         this.enumSet = EnumSet.copyOf(ordered);
         this.desc = ordered.stream().map(Color::getShortName).collect(Collectors.joining());
--- a/forge-core/src/main/java/forge/card/ColorSet.java
+++ b/forge-core/src/main/java/forge/card/ColorSet.java
@@ -20,7 +20,6 @@ package forge.card;
 import com.google.common.collect.UnmodifiableIterator;
 import forge.card.MagicColor.Color;
 import forge.card.mana.ManaCost;
-import forge.card.mana.ManaCostShard;
 import forge.util.BinaryUtil;
 
 import java.io.Serializable;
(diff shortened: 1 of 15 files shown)


Hanmac:
ugh the test was Green for me on Desktop :/

Hanmac:
Ugh, back to `ExceptionInInitializerError` i couldn't reproduce on Desktop

Hanmac:
@tool4ever finally found my problem (the Localizer was used too early)

What do you think of that changes?
no need to have extra list for ColorSet and Shards/Colors if they are always the same.
Also, init all 32 ColorSet objects early to be used by these Filters.

Hanmac:
@Jetz72 your opinion to the changes?

(the Localizer did bite me in the butt because I didn't notice the reason why)

forge-core/src/main/java/forge/card/ColorSet.java
```diff
@@ -41,27 +40,97 @@
...
+    private final Set<Color> enumSet;
+    private final String desc;
 
     public static final ColorSet ALL_COLORS = fromMask(MagicColor.ALL_COLORS);
-    private static final ColorSet NO_COLORS = fromMask(MagicColor.COLORLESS);
+    public static final ColorSet NO_COLORS = fromMask(MagicColor.COLORLESS);
 
-    private ColorSet(final byte mask) {
-        this.myColor = mask;
+    private ColorSet(Collection<Color> ordered) {
```
Jetz72: You could probably make this take a varargs `Color...` and trim out the `Arrays.asList` calls above. Just need to swap in `Arrays.stream` below.

forge-core/src/main/java/forge/card/ColorSet.java
```diff
@@ -41,27 +40,97 @@
 public final class ColorSet implements Comparable<ColorSet>, Iterable<Byte>, Serializable {
     private static final long serialVersionUID = 794691267379929080L;
 
+    // needs to be before other static
+    private static final ColorSet[] cache = new ColorSet[MagicColor.ALL_COLORS + 1];
+    static {
```
Jetz72, line 45: Which other static is this comment referring to?

Hanmac: The statics like ALL_Color and NO_Color that calls fromMask

Jetz72:
Definitely onboard with dropping the `ManaCostShard` stuff in this class. 

This reminds me that DeckRecognizer has a bunch of helper methods for colors that seem appropriate to move into the color class. Worth considering.

forge-core/src/main/java/forge/card/ColorSet.java
Hanmac: done

Hanmac:
> This reminds me that DeckRecognizer has a bunch of helper methods for colors that seem appropriate to move into the color class. Worth considering.

i'm on my way for this (see the other MR)

Hanmac:
Next smaller MR will be helper function for ColorSet like:

`ColorSet.contains(MagicColor.Color)` using the orderedShards Collection
and `ColorSet.combine(ColorSet)` that combines one ColorSet with another one

Hanmac:
Also, i saw `AEtherFilter` way too often in the crash logs, i want to make it better too

PR #8724, opened 2025-09-16
AetherFilter: use Predicate for Caption

Hanmac:
* AEtherFilter predicates now create the Label dynamically
* makes chain for FilterButton > AEtherFilter > Predicate<PaperCard> 
so the FilterButton itself can be used as card filter

### DIFF
--- a/forge-gui-mobile/src/forge/screens/planarconquest/ConquestAEtherScreen.java
+++ b/forge-gui-mobile/src/forge/screens/planarconquest/ConquestAEtherScreen.java
@@ -100,10 +100,10 @@ public class ConquestAEtherScreen extends FScreen {
     }
 
     private void updateFilteredPool() {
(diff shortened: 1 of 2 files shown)


Hanmac:
@Jetz72 your opinion to this cleanup?

Jetz72:
I think it's a little strange to make a UI element implement Predicate, but it's a private class so it's probably fine.

The 32 elements in the filter enum for all the different color pairings scream that something stands to be cleaned up there. I think generally we should only need one "all possible combinations of colors" enumeration across the codebase and if any others are needed they should just programmatically derive from that. But the fact that it's bundled in with other filter options suggests that slimming it down would require a pretty substantial rewrite. Not sure if that's worth the trouble.

Hanmac:
@Jetz72 yeah the AetherFilter is kinda funky and maybe should be removed/refactored later

Jetz72 (approved):

PR #8726, opened 2025-09-17
update setlookup files

### LINKED ISSUE #8721: Adventure mode shops require images in legacy folders
For whatever reason adventure mode shops only use old set codes to find card images if the set has them.

For example if I find a Wall of Razors from Stronghold in a shop (official code STH, but has legacy Forge Code2 SH defined in its editions file), the shop will not display an image unless it is in the SH folder. Images in the STH folder aren't found.

This is entirely at odds with the rest of Forge which handles the new set codes in image folder names just fine. If I buy the Wall of Razors the image in the STH folder displays in my inventory and presumably in battle. Cards from newer sets without legacy Code2 entries in their editions files display fine in shops as well.

I'm on the 9/14 snapshot, playing on Windows desktop. I have "use card variants from all sets" on, and I only noticed the issue after I enabled it. Not sure if whatever is causing the bug is confined to that option or if I was just lucky before - I didn't play very long with it off.

kevlahnota:
- closes #8721


Hanmac:
@kevlahnota why are these files needed or can they be removed?

kevlahnota:
The user want to use newer set codes as folder names instead of legacy set code. With these it should work example the set code for stronghold, it can look at the sth and sh folder. Though I guess maybe he used a different program to download bulk images instead of the built in downloader of forge.

Hanmac:
@tehdiplomat @Eradev we should look into the Definition of the Edition Files and see if these extra setcode lookup can be removed

kevlahnota:
> @tehdiplomat @Eradev we should look into the Definition of the Edition Files and see if these extra setcode lookup can be removed

Its better if we pass all the available setcodes when fetching the image so we dont need the extra files. If scryfall updates the setcode, only the edition files need the update and just parse those updated setcodes.

Hanmac:
yes, i want to get rid of setlookup mechanic if possible

that also bugged me when i was trying my #7771 ImageKey thoughts

kevlahnota:
> yes, i want to get rid of setlookup mechanic if possible
> 
> that also bugged me when i was trying my #7771 ImageKey thoughts

Can we pass the imagekey as a record to get the key, code1, code2 or anything needed so it doesn't need parsing and save computation?

Hanmac:
> > yes, i want to get rid of setlookup mechanic if possible
> > that also bugged me when i was trying my #7771 ImageKey thoughts
> 
> Can we pass the imagekey as a record to get the key, code1, code2 or anything needed so it doesn't need parsing and save computation?

That is planned, but the ImageKey Record isn't finished yet

i still think that setLookup can be removed before that

tehdiplomat:
Yea i thought we already had all of this information on the edition itself? I don't relaly like having 60 files that already exist somewhere.

PR #8730, opened 2025-09-17
Fix Self-Destruct

tool4ever:


forge-gui/res/cardsfolder/s/self_destruct.txt
```diff
@@ -1,10 +1,9 @@
 Name:Self-Destruct
 ManaCost:1 R
 Types:Instant
-A:SP$ Pump | ValidTgts$ Creature.YouCtrl | AILogic$ PowerDmg | TgtPrompt$ Select target creature you control | RememberTargets$ True | SubAbility$ DBDealDamage1 | StackDescription$ None | SpellDescription$ Target creature you control deals X damage to any other target and X damage to itself, where X is its power.
-SVar:DBDealDamage1:DB$ DealDamage | ValidTgts$ Any | TargetUnique$ True | AILogic$ PowerDmg | NumDmg$ X | DamageMap$ True | SubAbility$ DBDealDamage2 | DamageSource$ Remembered
-SVar:DBDealDamage2:DB$ DealDamage | NumDmg$ X | Defined$ Remembered | SubAbility$ DBDamageResolve | DamageSource$ Remembered
-SVar:DBDamageResolve:DB$ DamageResolve | SubAbility$ DBCleanup
-SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
-SVar:X:Remembered$CardPower
+A:SP$ DealDamage | ValidTgts$ Creature.YouCtrl | TgtPrompt$ Select target creature you control | AILogic$ PowerDmg | NumDmg$ X | DamageMap$ True | SubAbility$ DBDealDamage2 | DamageSource$ ThisTargetedCard | SpellDescription$ Target creature you control deals X damage to any other target and X damage to itself, where X is its power.
```
Hanmac, line 4: Doesn't this flip the Damage order?
It should first Damage the other Target,
And then Damage to itself?

tool4ever: but it should happen simultaneously anyway
that's why they just fill the table
and then when doing replacements you might choose which affected first, if it even matters?

forge-game/src/main/java/forge/game/Game.java
```diff
@@ -880,8 +882,6 @@ public void onPlayerLost(Player p) {
                         }
                         triggerList.put(c.getZone().getZoneType(), null, c);
                         getAction().ceaseToExist(c, false);
-                        // CR 603.2f owner of trigger source lost game
```
Hanmac, line 883: What about the delayed trigger? Should they still happen or not?

tool4ever: they shouldn't be cleared like that
`TriggerHandler.onPlayerLost` will take care of which ones are invalid

PR #8736, opened 2025-09-17
Fix Trial of Agony can target creatures controlled by two different opponents

CTimmerman:
Fixes #8729.

PR #8739, opened 2025-09-18
Remove TargetsFromSingleZone

### LINKED ISSUE #8729: AI using trial of agony can target creatures own by two different opponents
**Describe the bug**
Trial of agony requires the user to target two creatures owned by one opponent. When this card is used by the AI, they are able to target creatures owned by two different opponents. The effect then resolving with almost double the effect where two creatures are being dealt damage and the in game log does not show a creature being targeted as unable to be blocked (see screenshot of in game log). This issue is not present when cast by human play. Only creatures owned by a single opponent are able to be targeted as expected.

**To Reproduce**
AI use trial of agony while two opponents of that user each has a creature on board.

**Expected behavior**
Trial of agony should only be able to target two creatures owned by the same opponent

Image

Image

 - OS: Windows
 - Version 2.0.05-SNAPSHOT-07.26

tool4ever:
The logic was basically the same as TargetsWithSameController but more flawed - so we can avoid some duplicated code.

Closes #8736
Closes #8729


Hanmac (approved):

Hanmac:
whats about the other removed stuff in AI? is this also unused?

tool4ever:
yea, we just check `canTarget` below now anyway
so it doesn't make sense to have bad copied engine logic sitting there anymore

PR #8740, opened 2025-09-18
Conquest: make ColorButton instead of ColorFilter

Hanmac:
This makes an extra ColorButton that can doesn't need all the ColorSets as extra enum Options

i don't know yet if i should specialize the other buttons too, i will think about it


Hanmac:
The other Enums will be separated, so there is no need for extra Filter classes when the Enum itself can handle that

Hanmac:
@tool4ever @Jetz72 you might look this out

Hanmac:
@kevlahnota your opinion on this cleanup?

kevlahnota:
> @kevlahnota your opinion on this cleanup?

I think its good.

Hanmac:
`public static ColorSet[] allOptions() { return cache; }` right now is bit-ordered, not in the orderWight Order
that's why the ColorButton needs to order them again.

There are only so few spaces, where I would need different ColorSets to be ordered.
(Conquest ColorButton is one of them)

That matters when i later turn ColorSet into an Enum

Hanmac:
@tehdiplomat what do you think? can this be merged?

Hanmac:
Blocked by #8757

Hanmac:
@kevlahnota @tool4ever this MR should be ready too

Hanmac:
@tool4ever rebased and ready for review

PR #8742, opened 2025-09-18
Refactor "Unique Cards Only" Setting for Mobile Editor

Jetz72:
Previously, when this option was set, the deck editor would handle it by fetching an abridged card pool from `FModel` and passing that off to the item manager. This caused problems with the "add additional copies" and "replace variant" options on cards already in the deck, since if "Unique Cards" was toggled on in the catalog page, it could only see the unique-only pool.

This fixes that by moving all the unique-filtering logic over to the same place where `ItemManager` applies its other filters, no longer actually reducing the card pool. The logic for doing so was transplanted directly from the desktop version - 

https://github.com/Card-Forge/forge/blob/ba4ee98c3bdd19189f58e22751a969cf894cd494/forge-gui-desktop/src/main/java/forge/itemmanager/ItemManager.java#L997-L1021

https://github.com/Card-Forge/forge/blob/ba4ee98c3bdd19189f58e22751a969cf894cd494/forge-gui-desktop/src/main/java/forge/itemmanager/CardManager.java#L57-L124

Not a perfect solution, but it seems to work, and parity between the two UI is nice when it's possible. Ideally desktop and mobile could inherit these methods from some parent interface, but that feels like a bigger project.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
+++ b/forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
@@ -59,7 +59,7 @@ public class AdventureDeckEditor extends FDeckEditor {
         }
 
         @Override
-        public ItemPool<PaperCard> getCardPool(boolean wantUnique) {
+        public ItemPool<PaperCard> getCardPool() {
             return Current.player().getCards();
         }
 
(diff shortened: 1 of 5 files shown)


Hanmac:
Might need to check if we still need `getUniqueCardsNoAlt`

Hanmac:
@kevlahnota we need to look at `getUniqueCardsNoAlt` if its still needed

Jetz72:
> Might need to check if we still need `getUniqueCardsNoAlt`

I'm having trouble grasping what we'd need `CardDB.getAllCards` for (without `NoAlt`). From what I can tell, CardDB's `allCardsByName` list contains duplicate entries for PaperCards with more than one name, so they can be looked up. The `getAllCards` and `streamAllCards` family of methods derives their output from the values of this map. But without the keys, that just means they're seeing a collection of all cards where DFCs and split cards have duplicate entries, right?

Unique cards reduces it to one card per name, but doesn't look like it would resolve the multiple names per card situation until the `NoAlt` variants.

Hanmac (approved):

PR #8746, opened 2025-09-19
updateRarityFilterOdds

### LINKED ISSUE #8728: Loading Planar Conquest when not needed
I saw this happen while doing my ColorSet changes

FModel init the `ConquestController` and updates the Rarity there:
https://github.com/Card-Forge/forge/blob/dbf9568866c35eb2ced25cbcdf38da2a164cf6f8/forge-gui/src/main/java/forge/model/FModel.java#L250

Because I had a problem with Constructor (Localizer problem), it got reported all the way back.

But the underlying Problem is this:
* right now, there is only GUI for mobile part (android/mobile-dev)
that means the data is unnecessary for desktop (and test?)
* it is called for each test, even if the test isn't about planar conquest.
If we removed the extra calls, it could make the tests faster.

kevlahnota:
- closes #8728

### DIFF
--- a/forge-gui/src/main/java/forge/model/FModel.java
+++ b/forge-gui/src/main/java/forge/model/FModel.java
@@ -220,8 +220,6 @@ public final class FModel {
         for (final CardBlock b : blocks) {
             magicDb.getBlockLands().add(b.getLandSet().getCode());
         }
-        questPreferences = new QuestPreferences();
-        conquestPreferences = new ConquestPreferences();
         netPreferences = new ForgeNetPreferences();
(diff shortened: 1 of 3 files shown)


forge-gui/src/main/java/forge/model/FModel.java
```diff
@@ -466,10 +463,17 @@ public static IStorage<CardBlock> getBlocks() {
     }
 
     public static QuestPreferences getQuestPreferences() {
+        if (conquestPreferences == null)
```
Hanmac: Wrong var

PR #8757, opened 2025-09-21
ColorSet: turn Set into an Enum

Hanmac:
Part of #8682 

This turns ColorSet into an Enum with all 32 Combinations


kevlahnota:
are those ColorSet are saved in a file or only used on network game?  (hence they represent byte)? As long as they don't introduce problem, this is good.

tool4ever:
this looks off:
image

Hanmac:
> this looks off: image

fixed for now
the Dialog needs to be updated later anyway

forge-core/src/main/java/forge/card/ColorSet.java
```diff
@@ -346,23 +307,13 @@ public boolean hasGreen() {
...
-
-    /*
-     * (non-Javadoc)
-     *
-     * @see java.lang.Object#toString()
-     */
-    @Override
-    public String toString() {
-        return desc;
+        return (byte)ordinal();
```
Hanmac, line 316: @tool4ever i don't know if we still need `ColorSet.getColor` if it's just ordinal cast into byte

Hanmac:
@tool4ever ready to review?

PR #8764, opened 2025-09-22
FSkinProp: unify Enum to SkinProp places

Hanmac:
@Jetz72 moved the logic to common ground

i need to test with mobile-dev against side-effects

### DIFF
--- a/forge-gui-mobile/src/forge/screens/match/views/VPlayerPanel.java
+++ b/forge-gui-mobile/src/forge/screens/match/views/VPlayerPanel.java
@@ -9,6 +9,7 @@ import com.badlogic.gdx.utils.Align;
 
 import forge.Forge;
 import forge.Graphics;
+import forge.assets.FSkin;
 import forge.assets.FSkinColor;
 import forge.assets.FSkinColor.Colors;
 import forge.assets.FSkinFont;
@@ -19,6 +20,7 @@ import forge.game.card.CounterEnumType;
 import forge.game.player.PlayerView;
 import forge.game.zone.ZoneType;
 import forge.localinstance.properties.ForgePreferences.FPref;
+import forge.localinstance.skin.FSkinProp;
 import forge.menu.FMenuBar;
 import forge.menu.FMenuItem;
 import forge.menu.FPopupMenu;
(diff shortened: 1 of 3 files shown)


forge-gui-mobile/src/forge/screens/match/views/VPlayerPanel.java
```diff
@@ -139,23 +141,8 @@ public void addZoneDisplay(ZoneType zoneType) {
...
-            case PlanarDeck -> FSkinImage.PLANAR;
-            case SchemeDeck -> FSkinImage.SCHEME;
-            case AttractionDeck -> FSkinImage.ATTRACTION;
-            case ContraptionDeck -> FSkinImage.CONTRAPTION;
-            case Ante -> FSkinImage.ANTE;
-            case Junkyard -> FSkinImage.JUNKYARD;
-            default -> FSkinImage.HDLIBRARY;
-        };
+    public static FSkinImageInterface iconFromZone(ZoneType zoneType) {
+        return FSkin.getImages().get(FSkinProp.iconFromZone(zoneType, Forge.hdbuttons));
```
Jetz72, line 145: Any idea if it'd be sorely missed if we just deprecated or removed low definition icons here? Is it just there to support older UI skins or is there some other reason for it?

Jetz72:
Seems like a good place to move this logic. Could possibly support others in a similar manner, like `ManaCostShard` and `DeckSection`. It'll need some helper method for routing these if you want to fit multiple types of enums and their icons in the same list view, but it seems like the nicest option. Could optionally give all these methods same name as overloads, e.g. `iconFrom(ZoneType)` and `iconFrom(MagicColor.Color)`.

forge-gui-mobile/src/forge/screens/match/views/VPlayerPanel.java
Hanmac: there are some fallbacks that would disable the hdbuttons

Hanmac:
> Seems like a good place to move this logic. Could possibly support others in a similar manner, like `ManaCostShard` and `DeckSection`. It'll need some helper method for routing these if you want to fit multiple types of enums and their icons in the same list view, but it seems like the nicest option. Could optionally give all these methods same name as overloads, e.g. `iconFrom(ZoneType)` and `iconFrom(MagicColor.Color)`.

ManaCostShard already has their logic with SHARD_IMG and/or MANA_IMG

DeckSection i can add too

PR #8766, opened 2025-09-23
Fix Lady Octopus, Inspired Inventor

squee1968:


Hanmac:
@squee1968 need to rebase your changes.

The file got moved to the folders

Issue #8646, opened 2025-09-05
CantBeTargetedBy: Unify Logic with Message


Hanmac:
SpellAbility::canTarget
https://github.com/Card-Forge/forge/blob/15d4e1ac20e96d48c7b19341bcd7f02f9c2e66cb/forge-game/src/main/java/forge/game/spellability/SpellAbility.java#L1378

InputSelectTargets::onCardSelected
https://github.com/Card-Forge/forge/blob/15d4e1ac20e96d48c7b19341bcd7f02f9c2e66cb/forge-gui/src/main/java/forge/gamemodes/match/input/InputSelectTargets.java#L163

Share a lot of similar checks with TargetRestriction and similar flags

I try to move the whole logic into a `SpellAbility::cantTargetMsg` function that returns a String if the target would be fail
so `canTarget` returns true if the String is null

Affected from this are these StaticAbilties:
* StaticAbilityCantTarget
* StaticAbilityCantAttach

i want them to have functions to return the First StaticAbility where the condition is true.
so the `cantTargetMsg` can be which Static/Keyword is causing it to not be targeted Shroud/Hexproof/Protection

Hanmac:
@tool4ever for better Desc on `TgtPrompt` and `ValidTgts`, i think we should add something like `ValidTgtsDesc`.
that works like in-between them:

* ValidTgts: `Creature.YouCtrl`
* ValidTgtsDesc: `creature you control`
* TgtPrompt: `Select target creature you control`

so in theory, we might get rid of TgtPrompt for most of the cases.

PR #8772, opened 2025-09-24
cantBeEnchantedByMsg and cantAttach StaticAbility

Hanmac:
Part of #8646

will be used for HumanGUI to return if Aura can target something

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -7162,7 +7162,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
     }
 
     @Override
-    public final String cantBeEnchantedByMsg(final Card aura) {
+    protected final String cantBeEnchantedByMsg(final Card aura) {
         if (!aura.hasKeyword(Keyword.ENCHANT)) {
             return "No Enchant Keyword";
         }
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -7171,24 +7171,32 @@ public final boolean canBeControlledBy(final Player newController) {
...
             String k = ki.getOriginal();
             String m[] = k.split(":");
             String v = m[1];
-            if (!isValid(v.split(","), aura.getController(), aura, null)) {
-                return false;
-            }
-            if (!v.contains("inZone") && !isInPlay()) {
-                return false;
+            if (!isValid(v.split(","), aura.getController(), aura, null) || (!v.contains("inZone") && !isInPlay())) {
+                String desc;
```
tool4ever: hmn, these lines would now exist 4 times (one more each in Card and CardState)
can't the KW provide it somehow?

Hanmac: Not yet, for that the Keyword needs to be extended

Hanmac: Done something that can be extended later

tool4ever (approved):

Hanmac:
i don't know if there is a better way for this:

```java
        if (attach.isAura()) {
            String msg = cantBeEnchantedByMsg(attach);
            if (msg != null) {
                return msg;
            }
        }
        if (attach.isEquipment()) {
            String msg = cantBeEquippedByMsg(attach, sa);
            if (msg != null) {
                return msg;
            }
        }
        if (attach.isFortification()) {
            String msg = cantBeFortifiedByMsg(attach);
            if (msg != null) {
                return msg;
            }
        }
```

but might be optimized later

PR #8774, opened 2025-09-24
Edition updates: AA3, AA4, ECL, SLD

paulsnoops:


forge-gui/res/editions/Arena Anthology 3.txt
```diff
@@ -0,0 +1,33 @@
+[metadata]
+Code=AA3
+Date=2025-09-23
+Name=Arena Anthology 3
+Type=Online
+ScryfallCode=AA3
```
Hanmac, line 6: ScryfallCode = Code isn't needed

PR #8793, opened 2025-09-27
Card: addColor use internal Record

Hanmac:
Part of #8682

More changes after #8757


Hanmac:
@tool4ever what do you think? That should be small enough?

(i moved the CardColor as InternalRecord because it isn't used elsewhere)

Issue #8670, opened 2025-09-08
Property Cleanup Meta


Hanmac:
Remove of "Not"/"non" Properties in favor of "!" negation

Card:

- [x] NotTriggered
- [x] notnamed
- [ ] notTributed
- [ ] NotDefined
- [ ] notExertedThisTurn
- [x] notattacking
- [x] notblocking
- [ ] nonChosenCard
- [ ] `sharesNameWith NonToken` ?
- [ ] `doesNotShareNameWith`
- [ ] `greatestCMC_NonLandPermanent` ?

Player:

- [ ] NonActive

CardState:

- [ ] `non` + ColorWord
- [ ]  `non` + StringType (SubTypes)
- [ ] ChosenType + `chosenType.startsWith("Non")`
- [ ] IsNotChosenType/IsNotChosenType2

SpellAbility:

- [x] nonManaAbility

Hanmac:
`sharesNameWith` Property should be refactored in using getDefinedCards list

PR #8804, opened 2025-09-28
Property cleanup

Hanmac:
Part of #8670

removed some "non" properties in favor of "!" negation


forge-gui/src/main/java/forge/gui/card/CardScriptParser.java
```diff
@@ -485,7 +485,7 @@ private static boolean isValidLegal(final String valid) {
             "wasNotCast", "ChosenType", "IsNotChosenType", "IsCommander",
             "IsRenowned");
     private static final Set<String> VALID_EXCLUSIVE_STARTSWITH = ImmutableSortedSet
-            .of("named", "notnamed", "OwnedBy", "ControlledBy",
+            .of("named", "!named", "OwnedBy", "ControlledBy",
```
tool4ever: just remove it

Issue #5634, opened 2024-07-17
Spells are incorrectly displayed on the stack


MorihiroS:
**Describe the bug**
Some spells have incorrect text when they're displayed in the stack:

- [ ] Assassin's Trophy (It says that a player who cast the spell (not the one who had their permanent broken!!) is looking for the "Basic.Land" card);
- [ ] All is Dust (It says "Sacrifice permanents"");
- [ ] Recross thе Paths (It says "NICKNAMEreveals...", without space before "reveals");
- [ ] Surgical Extraction (It says "CARDNAME (XX) until end of turn. Search...")

**To Reproduce**
1. AI/Player casts Assassin's Trophy/All is Dust/Recross thе Paths/Surgical Extraction
2. See wrong figure

**Expected behavior**
Assassin's Trophy (It should say that a player who had their permanent broken may looking for the basic land card.);
All is Dust (It should say "Sacrifice all colored permanents" or "Each player sacrifices all permanents they control that are one or more colors");
Recross thе Paths (It should say "NICKNAME reveals...");
Surgical Extraction (It should say like "NICKNAME choose target CARDNAME (XX) in a graveyard. Search..." )

**Additional context**
Similar with #4318

**Desktop**
 - OS: Windows 10
 - Java Version: 1.8.0_202-b08
 - Forge Version: 1.6.64-SNAPSHOT-07-13

**Screenshots**

![trophy](https://github.com/Card-Forge/forge/assets/115870280/96a0cc1c-f4d7-4eb5-bfd9-0d07b33e9cf9)
![all is dust](https://github.com/user-attachments/assets/45ea782b-b4c9-4ea4-ba2e-467741ab83d2)
![recr1](https://github.com/user-attachments/assets/f54f7720-f081-411b-9cd2-4f91b48385a0)
![surg](https://github.com/user-attachments/assets/394ca4fe-df1e-49c6-8e67-c0eb2ed4b698)

MorihiroS:
Also problem with activated ability of Nahiri, the Harbinger:

![Nahiri](https://github.com/user-attachments/assets/85ab9c83-09dd-4eb3-b0c0-e0ac7e3131c6)

Hanmac:
some of them can be fixed with using `ChangeTypeDesc`

MorihiroS:
Add Tear Asunder and Sleight of Hand

MorihiroS:
[...]

PR #8809, opened 2025-09-29
Change type desc

Hanmac:
Part of #5634 

i was looking for "Land.Basic" effects first, and then removed the ChangeNum default value

### DIFF
--- a/forge-gui/res/cardsfolder/a/anchor_to_reality.txt
+++ b/forge-gui/res/cardsfolder/a/anchor_to_reality.txt
@@ -1,7 +1,7 @@
 Name:Anchor to Reality
 ManaCost:2 U U
 Types:Sorcery
-A:SP$ ChangeZone | Cost$ 2 U U Sac<1/Artifact;Creature/artifact or creature> | Origin$ Library | Destination$ Battlefield | ChangeType$ Equipment,Vehicle | ChangeTypeDesc$ Equipment or Vehicle card | ChangeNum$ 1 | RememberChanged$ True | SubAbility$ DBScry | SpellDescription$ Search your library for an Equipment or Vehicle card, put that card onto the battlefield, then shuffle.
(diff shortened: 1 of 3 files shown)


forge-gui/res/cardsfolder/a/averna_the_chaos_bloom.txt
```diff
@@ -3,5 +3,5 @@ ManaCost:G U R
 Types:Legendary Creature Elemental Shaman
 PT:4/2
 R:Event$ Cascade | ValidPlayer$ You | ActiveZones$ Battlefield | ReplaceWith$ DBLand | ReplacementResult$ Updated | Description$ As you cascade, you may put a land card from among the exiled cards onto the battlefield tapped. (Do this after the last card is exiled but before deciding whether to cast a nonland card.)
-SVar:DBLand:DB$ ChangeZone | Hidden$ True | Origin$ Exile | ChooseFromDefined$ ReplacedCards.Land | ChangeNum$ 1 | Destination$ Battlefield | Tapped$ True
+SVar:DBLand:DB$ ChangeZone | Hidden$ True | Origin$ Exile | ChooseFromDefined$ ReplacedCards.Land | Destination$ Battlefield | Tapped$ True
```
tool4ever: just find/replace removing it everywhere will break some cards lol

maybe the behaviour needs to be synchronized, then "search...any number" scripts don't need to use extra SVar

pretty sure we had an issue for that somewhere

Hanmac: ugh yeah, i really need to add that special "anyNumber"

Hanmac:
@tool4ever may be easy stuff like this should be auto generated too?
* `Mountain,Plains` → `ChangeTypeDesc$ Mountain or Plains`
* `ChangeType$ Creature,Planeswalker | ChangeNum$ 2` → `ChangeTypeDesc$ creature and/or planeswalker`

Hanmac:
`Count$DifferentCardNames_` needs to be reworked into handlePaid,
like `Valid Creature.YouCtrl$DifferentCardNames`

but that will be another MR

tool4ever (approved):

PR #8830, opened 2025-10-04
Rework sacrifice cost text

Eradev:
Working on #5916

### DIFF
--- a/forge-game/src/main/java/forge/util/MessageUtil.java
+++ b/forge-game/src/main/java/forge/util/MessageUtil.java
@@ -153,15 +153,27 @@ public class MessageUtil {
                 }
 
                 // Prefixes
-                if (pOrS.equals("Other")) {
-                    if (hasOther) {
+                switch (pOrS) {
+                    case "Other" -> {
+                        if (hasOther) {
+                            continue;
+                        }
+
+                        currentTerm += pluralize ? "other " : "another ";
+                        prefixes.append(currentTerm);
+                        hasOther = true;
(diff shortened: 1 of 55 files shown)


forge-game/src/main/java/forge/util/MessageUtil.java
```diff
@@ -97,4 +97,190 @@ public static String mayBeYou(Player player, Object what) {
     public static String mayBeYou(PlayerView player, Object what) {
         return what == null ? "(null)" : what == player ? Localizer.getInstance().getMessage("lblYou") : what.toString();
     }
+
+    public static String complexTargetTypesToString(String types, boolean pluralize) {
+        String[] parts = types.split(";");
```
Hanmac: i would do the split outside, because then we could use it for Valid that uses `,` as Split

Eradev: I've made it so you can pass the split regex as parameter.

Hanmac: The parameter `types` is only used to be split into `String[] parts`
wouldn't it be better to pass `String[] parts` directly as parameter?

Eradev: Good point. I'll change it next time.

Hanmac:
@Eradev i did something for this in: #9063

`Lang.getInstance.buildValidDesc` could be extended to support stuff like "you control" or something

PR #8839, opened 2025-10-05
Add AtOppEOT logic to a couple lands that the AI misuses

### LINKED ISSUE #8836: AI Exudes Harmless "Toxins"
**Describe the bug**
I've noticed the AI has a preference for choosing 0 when playing Scavenger Regent's Exude Toxin adventure. This is silly because it results in the AI wasting mana in order to shuffle 1 of its cards from its hand into its library to no good effect (in general).

**Expected behavior**
IDK. I've come to expect the AI to make silly moves here & there but this 1 seems especially egregious when it occurs.

I realize (@ least dimly) what a daunting task it must be to train the AI to play magical cards. I think I heard some1 liken a similar activity to training retarded children in the intricacies of rocket science. I get the feeling AI issues frequently go unaddressed since there can be so many unintended consequences for "simple" fixes that end up creating more problems than solving. There probably aren't that many devs that wrestle w/ the AI logic. That said, I think I'd like to point out another couple of cards the AI seems to have difficulty w/ here rather than creating an individual issue for each 1.

2 lands come to mind: The Gold Saucer & Bucolic Ranch. When I see the AI play either land, I know it's going to be quite difficult to lose that game. The AI will activate either land as soon as it possibly can & @ every available opportunity. This results in a lot of "wasted" mana since the AI is usually stacking up a bunch of cards in hand that would make better use of that mana. I'm pretty sure there are other lands like this & a possible fix probably involves changing how the AI prioritizes activated abilities of lands (or just these specific lands).

Ragged Playmate seems to get activated far more than is profitable for the AI to do so. If the AI has a target for this ability, it seems to choose it &, like the lands mentioned above, it seems to do this before practically any other considerations (I frequently see the AI activate during upkeep, spending mana before it's seen its new draw). I've seen the AI target the same creature w/ 2 Playmates when that play is redundant & wasteful; I think I've even seen 2 Playmates target each other. I've seen the AI activate, removing the Playmate as a possible blocker on my turn & then decide against sending their unblockable attack because the AI is facing a lethal alpha strike on my turn (a strike which usually wouldn't have been lethal if they hadn't tapped their Playmate).

IDK if it's productive to bring these things up but I did notice the mods labeling some threads w/ an "AI" tag so I assume it's OK to raise these issues here.

 - OS: Win 10
 - 09.23 snap

Agetian:
Fixes #8836

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/FlipACoinAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/FlipACoinAi.java
@@ -18,9 +18,7 @@ public class FlipACoinAi extends SpellAbilityAi {
     protected AiAbilityDecision checkApiLogic(Player ai, SpellAbility sa) {
         if (sa.hasParam("AILogic")) {
             String ailogic = sa.getParam("AILogic");
-            if (ailogic.equals("Never")) {
-                return new AiAbilityDecision(0, AiPlayDecision.CantPlayAi);
-            } else if (ailogic.equals("PhaseOut")) {
+            if (ailogic.equals("PhaseOut")) {
                 if (!ComputerUtil.predictThreatenedObjects(sa.getActivatingPlayer(), sa).contains(sa.getHostCard())) {
                     return new AiAbilityDecision(0, AiPlayDecision.CantPlayAi);
                 }


forge-ai/src/main/java/forge/ai/ability/PeekAndRevealAi.java
```diff
@@ -28,6 +28,11 @@ protected AiAbilityDecision checkApiLogic(Player aiPlayer, SpellAbility sa) {
             if (aiPlayer.getGame().getPhaseHandler().getPhase().isBefore(PhaseType.MAIN2)) {
                 return new AiAbilityDecision(0, AiPlayDecision.CantPlayAi);
             }
+        } else if ("AtOppEOT".equals(logic)) {
```
tool4ever: I have recently pulled AtOppEOT into the parent so you no longer need to paste it into each API ;)
same is true for the other one after you switch the template into this (newer) variant

Agetian: Oh, very nice, wasn't aware of that! That's actually a really welcome change XD
I tried updating things to make use of it, seem to work fine, but please take a look if I did everything correctly.

forge-ai/src/main/java/forge/ai/ability/FlipACoinAi.java
```diff
@@ -12,10 +12,10 @@
...
     /* (non-Javadoc)
-     * @see forge.card.abilityfactory.SpellAiLogic#canPlayAI(forge.game.player.Player, java.util.Map, forge.card.spellability.SpellAbility)
+     * @see forge.card.abilityfactory.SpellAiLogic#checkApiLogic(forge.game.player.Player, java.util.Map, forge.card.spellability.SpellAbility)
      */
     @Override
-    protected AiAbilityDecision canPlay(Player ai, SpellAbility sa) {
+    protected AiAbilityDecision checkApiLogic(Player ai, SpellAbility sa) {
         if (sa.hasParam("AILogic")) {
             String ailogic = sa.getParam("AILogic");
             if (ailogic.equals("Never")) {
```
tool4ever: thanks
actually now this "Never" branch is also obsolete

Agetian: Awesome, tweaked too :)

tool4ever (approved):

PR #8840, opened 2025-10-05
Show Mayhem in flashback zone

Hanmac:
@tool4ever this shows Mayhem in the Flashback zone

or does it need better logic?


forge-game/src/main/java/forge/game/zone/PlayerZone.java
```diff
@@ -56,7 +56,7 @@ public boolean test(final Card c) {
 
             boolean graveyardCastable = c.hasKeyword(Keyword.FLASHBACK) ||
                     c.hasKeyword(Keyword.RETRACE) || c.hasKeyword(Keyword.JUMP_START) || c.hasKeyword(Keyword.ESCAPE) ||
-                    c.hasKeyword(Keyword.DISTURB);
+                    c.hasKeyword(Keyword.DISTURB) || c.hasKeyword(Keyword.MAYHEM);
```
tehdiplomat, line 59: For Mayhem, should we be checking if the card has been discarded this turn?

PR #8842, opened 2025-10-05
Dungeon completed

Hanmac:
unify logic with Dungeon Completed


tool4ever (approved):
one day we might unify more of these `doXMath`

Hanmac:
> one day we might unify more of these `doXMath`

I wanted to chain them at the PlayerProperty, but the format failed me 😓

PR #8849, opened 2025-10-06
Flavor Names

### LINKED ISSUE #8281: Broader support for Flavor Names
Currently, cards like "Mind Flayer, the Shadow" aka "Arvinox, the Mind Flail" use a property called "AltName". This allows cards that have been retroactively renamed by their Universes Within reprintings to be looked up by either name. However, it has some organizational issues and limitations. In particular, the alternate name's entry in the rulesByName and faceByName tables points to different `CardRules` and `CardFace` objects from the original name. It's also currently hard-coded to work with the Universes Within edition, can't support more than one alternate name per set of card rules, and needs an extra flag on the `Card::getName` method to normalize renamed cards to their oracle name.

This doesn't currently cause any significant issues, but with the upcoming organizational catastrophe that is [Through the Omenpaths](https://magic.wizards.com/en/news/announcements/through-the-omenpaths-and-digital-universes-beyond-updates), we might take the opportunity to improve on this system a bit.

We could switch to an alternate `FlavorName` property, and leverage the functional variant system to apply it on a print-by-print basis. So for instance we could write:

```
Name:Arvinox, the Mind Flail
Variant:SLD340:FlavorName:Mind Flayer, the Shadow
```

And we'd change the old SLD entry to:
```
340 M Arvinox, the Mind Flail @Isis $SLD340
```

As an alternative, we might skip the entry in the card script and apply the name directly through the edition entry. However, we may want to reexamine our syntax if we do so, possibly adding a key/value system for arbitrary properties (e.g. `123 C Foo @Artist {FlavorName: Bar}`, we could also add `VariantName` under the same roof). If we did this, we'd still probably generate the functional variants within the associated `CardRules`, so that cards with multiple faces can distribute the flavor names to those respective faces.

The proposed changes under the hood:
* The `Name` field always reflects the card's official oracle name; the `FlavorName` is always applied through a functional variant and applies only to printings using that variant. It does not overwrite the actual card name.
* Similar to AltNames, flavor names get recorded in the CardDb's name lists for lookup purposes. However, these would point to the same `CardRules` and `CardFace` instances as the originally named ones, rather than copies with modified names.
* Most searches and filters by card name can also match a card's flavor name.
* A flavor name is relayed to the in-game Card, stored as a separate property.
* `Card::getName` returns the card's current oracle name as modified by any in-game effects. Any game rules that care about card names will continue to use this.
* `Card`, `CardView`, and related State classes get a `getDisplayName` method. This will usually work the same as `getName`, but can instead supply the card's flavor name if it has one and no effects are altering the card's base name. This is used for most user-facing text, particularly "CARDNAME" and "NICKNAME" substitutions.
* Similarly, `PaperCard::getDisplayName` and `PaperCard::getDisplayOracle` can supply the flavor name for use in the deck builder.

Jetz72:
Adds support for cards having flavor names using the functional variant framework and the `FlavorName:` property in the script. This replaces the old `AltName` functionality, whereby a variant had two separate `CardRules` objects and the CardDB could be consulted to determine if they should be treated identically. Now a single `CardRules` object can handle all of them. 

Searching for a flavor name in the editor will find only prints matching that flavor name, whereas searching the Oracle name will find all prints. Prints with flavor names can be included in translation files. Unlike regular variants which are entered in translation data as `Card Name $Variant`, variants with flavor names simply use the flavor name as a key.

Some new methods to be aware of:

* `PaperCard.getDisplayName` - Gets the untranslated name or flavor name of a particular print to display in the UI. Should only be used for user-facing stuff; `getName` will still return the oracle name. `PaperCard.hasFlavorName` can indicate whether or not there's any distinction between the two.
* `Card.getDisplayName` and `Card.getName` function similarly - the former is for user facing UI things, the latter for game rules. A card's display name will be its flavor name if it has one, unless some effect is altering the card's name.
* As a view-facing class, `CardView` behaves differently from the above two. `CardView.getName` supplies a display name, and it now has `CardView.getOracleName` to supply an unaltered name for rules (and AI) purposes.
* `CardDB.allRulesByName`, `allFacesByName`, and `normalizedNames` will be able to map flavor names to the respective rules, faces, and names, but this will not fetch ones with the matching flavor names.
* `PaperCard.getAllSearchableNames` supplies a list of names a print can be found under. Contains the main face Oracle name, alt and combined name for cards that have alternate faces, flavor names for all of the above for cards that have them, and translated names of all of the above if the user is using another language.

This also implements a change to the variant syntax in edition files. They can now take arbitrary keys and values using JSON-object syntax. Old: `100 Card Name @Artist $VariantName`; New: `100 Card Name @Artist ${"variant": "VariantName"}`. Planning to use this for a separate PR to support some playtest cards with name collisions. We can also use this for future print-specific metadata we might want to add (e.g. has a horizontal layout, no Scryfall image available, foil-only...).

Potential future improvements:
* Subtext in the card detail UIs to indicate that a card is using a flavor name, and what its Oracle name is. Several of them are pretty crowded but an easy one might be adding some note to the rules text displays in-game.
* User preferences for disabling flavor names.
* Some parts of the code currently make the assumptions that we won't have A) a flavor name on the base rules of a card rather than a functional variant or B) a multi-faced card with a flavor name on one face but not one on the other. This could be generalized if needed.
* Decide if we want Universes Within cards to supplant the Oracle, script, and primary names of the original Universes Beyond cards, in which case we should rename the scripts and swap the names and flavor names.
* The searchable name code could be optimized a bit by sharing data for different prints across `CardRules`, but that'd get a lot more complicated to maintain. Currently it caches the searchable names on the `PaperCard`.

To-do list remaining before this PR is ready for review:
* ~~Remove some stuff I accidentally left in for that playtest card name collision project.~~
* ~~Think I need to update `PaperCard.getTranslationKey`.~~
* ~~Test deck importing. Might need to modify the card request pipeline so it can map flavor names to a print with the specified variant.~~
* ~~Gonna try to generate a map of flavor names to corresponding variant names and use that while processing Edition Entries to automatically apply the appropriate variant. Would make it unnecessary to tack `${"variant": "UniversesWithin"}` onto entries in the edition file; just using the flavor name would be enough.~~
* ~~Figure out if the `CardDb.getAllCards` family of methods need to just be replaced with their corresponding `NoAlt` methods and the originals removed. (See [here](https://github.com/Card-Forge/forge/pull/8742#issuecomment-3361234111))~~ (Probably, but in a different PR.)
* ~~Explore options for supporting "FlavorName" as a parameter in the edition entry. Might end up with three ways to achieve the same thing here but oh well.~~
* ~~Add flavor name support to tokens?~~ (Later)
* ~~Searches for Oracle text currently find all variants even if the search text only appears in a single variant. Need to fix that.~~

Closes #8281. Enables implementation of OM1.


Hanmac:
@Jetz72 you forgot to commit `addAuxiliaryScripts` ?

Hanmac:
im unsure about this:

```
Name:Lucille
Variant:UniversesWithin:FlavorName:Gisa's Favorite Shovel
```

wouldn't it bet better to define the OtherName in the Edition?

so it gets defined in the PaperCard, and that overwrites the Name from CardRules?

or would that be problematic for Transforming cards? :thinking:

Jetz72:
> wouldn't it bet better to define the OtherName in the Edition?

I was thinking about doing it that way, just moving the entire flavor name over to the edition entry. The issue is that it looks like full support of OM1 and future Omenpath sets may also call for flavored Oracle text, keywords, and type-lines, e.g. The Void Stone. Automated mappings can cover the type line, "CARDNAME" replacers, and in-game keywords, but applying it to paper cards in the editor and inventories gets tricky. Better that we can just have both versions of the Oracle text available. And writing all the flavored Oracle text into the edition files would be a bit wordy.

Hanmac:
@Jetz72 how are you planning the FlavorName mapping for multiple of the same set?

Command Tower has 5 different flavor names in SLD:
https://scryfall.com/search?q=%21%22Command+Tower%22+include%3Aextras+is%3Aflavorname&unique=art&as=grid&order=released


Najeela has 4 different flavor names:
https://scryfall.com/search?as=grid&order=released&q=%21%22Najeela%2C+the+Blade-Blossom%22+include%3Aextras&unique=prints

Jetz72:
Added the ability to just write a flavor name as an entry in an edition file, and it'll automatically apply the appropriate variant.

I might still explore options for applying a flavor name as a parameter to the edition entry. I can't dispute it's a more elegant place for it than the card script. But if we want to support the flavored oracle text (And I imagine people who prefer the Omenpaths cards will ask for it), then writing long lines of rules text in the edition file gets a bit messy. 

What might work best is if the parameter in the edition file acted as a shorthand for defining the respective variant in the rules object. Tacking `${"flavorName": "SpongeBob SquarePants"}` on to the end of the edition entry could instruct the CardDB to generate a new variant on the CardRules object in question. Not sure off the top of my head how easy it'd be to mutate the rules object like that. But that'd essentially leave us with two ways to define flavor names - either directly in the rules file for cards that have flavored oracle or effect text, or just in the edition file for cards that don't.

I also realized that there are a couple tokens that also use flavor names ([e.g.](https://scryfall.com/card/sld/1681/marit-lage)), and adding some variant support to those would be nice. I could do it similarly to cards, adding variant support to token rules objects. Or I could have the card that spawns the token apply the flavor name if that card is itself a flavored variant. Weighing options there.

> how are you planning the FlavorName mapping for multiple of the same set?

There isn't any upper bound on the number of variants a card can have. Was experimenting with this earlier today, using `Variant:SLD7030:FlavorName:Master Emerald Shrine`, `Variant:SLD710:FlavorName:Cybertron`, etc. Then can just use the respective name in the edition file.

Jetz72:
I've added a way to generate flavor name variants using only the edition entry, by adding the parameter "flavorName", e.g.
```
1612 R Tovolar, Dire Overlord @Zoltan Boros ${"flavorName": "NOT A WOLF // KING OF WOLVES"}
```
When loading an edition entry this way, it'll search for a variant with a matching flavor name, and generate a new one if it can't be found. It has rudimentary support for replacing the card's name within the generated variant's oracle text, but can't do nicknames or flavored ability names or anything more elaborate.

This makes the application of flavor names a lot more elegant on the surface level since you only have to change one resource file rather than synchronizing two. But it's a bit less elegant on the back end because edition entries are loaded after rules, meaning we're now modifying a `CardRules` object and the `CardDB` caches past their initialization. These alterations are well ahead of any point where the UI would start reading and searching through this data, but it does make the pipeline a bit messier. I'm kinda on the fence about this trade-off. Any input regarding this, @Hanmac, @tehdiplomat, @tool4ever? Still time to revise stuff.

Hanmac:
> meaning we're now modifying a CardRules object and the CardDB caches past their initialization.

I think this is still in the process of loading the game Data, so I allow it

Jetz72:
Looks like the deck importer can find the appropriate rules when given a flavor name, but it has no baring on what print is loaded. It'd be nice if it preferred the print using the matching name, but it becomes a question of whether a flavor name in a CardRequest should take precedence over an art or edition preference. I'm leaning towards "yes". 

Might just do the easy solution for flavoring tokens, i.e. `FlavorName$` as a script parameter on the parent card. Or I may just leave it off entirely, since it doesn't look like Through the Omenpaths currently has a token set. Eventually we may want to implement a "preferred token" parameter on the edition entry, eventually piped all the way through to token generation and used to choose a print. At that point we'd be able to match SLD cards to their designated tokens, and it'd be worth adding flavor name support on the token edition entries.

Hanmac:
imo CardRequest needs to be refactored more, but that's an issue for another time

Jetz72:
Think this feature is ready for review. To recap, there are currently there are multiple ways to specify a flavor name:
* Manually, by writing `Variant:Foo:FlavorName:Loret Ipsum` in the card script, and adding `${"variant": "Foo"}` to the end of the edition entry. You'll also want to add `Variant:Foo:Oracle:When Loret Ipsum enters...` if the flavor name would appear in the Oracle text, or if it would otherwise be changed.
* By lookup, again by using `Variant:Foo:FlavorName:Loret Ipsum` in the card script, but simply using "Loret Ipsum" as the name in the edition file.
* Automatically, by putting `${"flavorName": "Loret Ipsum"}` at the end of the edition entry. 

The third method is the easiest, but besides a simple find/replace for the card name, it won't be able to make any changes to flavor the Oracle text, such as for ability words. They all function the same under the hood once the CardDB is loaded; the latter two are just shortcuts for the first.

This PR contains examples of each method - manual for some Universes Within cards and automatic for some Secret Lair items. This isn't necessarily intended as a precedent, though. I'd recommend resorting to automatic in most cases except when other text changes are needed, or if a flavor name is used in more than one set. But I'll leave it up to whoever moves forward with implementing either Through the Omenpaths or the flavor names for existing cards. It would also be worth deciding whether cards that started in Secret Lair and later gained Universes Within printings should be changed to use the UW name as their script/Oracle names.

Hanmac:
@Jetz72 you probably need to rebase your Branch because of this:
"This branch cannot be rebased due to conflicts"



@tehdiplomat can you take a look at this when this is ready?

Hanmac:
I still need to think about if the new FlavorName does deprecate the `getFunctionalVariant` 🤔 but that might be for another MR

Jetz72:
> you probably need to rebase your Branch because of this:
> "This branch cannot be rebased due to conflicts"

I'm guessing that's just from my own fork's master branch being a couple days behind. Merging master into this branch seemed to work without conflict.

> I still need to think about if the new FlavorName does deprecate the `getFunctionalVariant` 🤔 but that might be for another MR

Can you elaborate on this? Do you mean the edition entry parameters, or something else?

Hanmac:
> > I still need to think about if the new FlavorName does deprecate the `getFunctionalVariant` 🤔 but that might be for another MR
> 
> Can you elaborate on this? Do you mean the edition entry parameters, or something else?

Maybe i got just confused that `String NO_FUNCTIONAL_VARIANT` still exist and is used 🤷‍♂️

Jetz72:
> Maybe i got just confused that `String NO_FUNCTIONAL_VARIANT` still exist and is used 🤷‍♂️

The vast majority of PaperCards will not use any functional variant, i.e. `NO_FUNCTIONAL_VARIANT`, i.e. an empty string. We could use null instead of an empty string for that. It was just originally implemented that way to align with similar placeholders like `NO_COLLECTOR_NUMBER` and `NO_ARTIST_NAME`.

Hanmac:
> > Maybe i got just confused that `String NO_FUNCTIONAL_VARIANT` still exist and is used 🤷‍♂️
> 
> The vast majority of PaperCards will not use any functional variant, i.e. `NO_FUNCTIONAL_VARIANT`, i.e. an empty string. We could use null instead of an empty string for that. It was just originally implemented that way to align with similar placeholders like `NO_COLLECTOR_NUMBER` and `NO_ARTIST_NAME`.

yeah, and that might need to be refactored.

i meant this part for example:

```java
        //Cards with flavor names will use that flavor name as their translation key. Other variants are just appended as a suffix.
        if(!NO_FUNCTIONAL_VARIANT.equals(getFunctionalVariant()) && getAllFaces().stream().noneMatch(pc -> pc.getFlavorName() != null))
            return getName() + " $" + getFunctionalVariant();
        return getDisplayName();
```

the slight difference between variant and flavorName

Jetz72:
Was probably a subtle misnomer there to use `pc -> pc.getFlavorName()` since `pc` in that context would actually be a `CardFace`. I'm open to refactoring `NO_FUNCTIONAL_VARIANT` to null, though that might be better as a separate PR. At some point in the future I'd also like to raise `@Nullable` and `@NonNull` annotations for consideration; for this and other places where it'd be easy to compare a function's return or a variable's value to the wrong thing.

forge-core/src/main/java/forge/card/CardDb.java
```diff
@@ -455,71 +509,28 @@ public void initialize(boolean logMissingPerEdition, boolean logMissingSummary,
...
-            //also include specialize faces
-            for (ICardFace face : paperCard.getRules().getSpecializeParts().values()) allCardsByName.put(face.getName(), paperCard);
-        }
+        //Card may have multiple names. Add it under all of them.
+
+        List<ICardFace> allFaces = paperCard.getAllFaces();
+        Set<String> namesToAdd = new HashSet<>();
+        allFaces.stream().map(ICardCharacteristics::getName).forEach(namesToAdd::add);
+        allFaces.stream().map(ICardFace::getFlavorName).filter(Objects::nonNull).forEach(namesToAdd::add);
+        namesToAdd.remove(mainName);
```
tool4ever, line 531: just remove the early return above so all names can be handled here?

then you don't even need to build extra Set only to add them to the real field directly after?

forge-core/src/main/java/forge/card/CardRules.java
```diff
@@ -464,6 +469,58 @@ public Set<String> getSupportedFunctionalVariants() {
...
+        }
+        String variantName = suggestedVariantName != null ? suggestedVariantName : "FlavorName" + flavorName.hashCode();
+        if(supportedFunctionalVariants.contains(variantName))
+            variantName = variantName + flavorName.hashCode();
+
+        CardFace variantMain = ((CardFace) mainPart).getOrCreateFunctionalVariant(variantName);
+        variantMain.setFlavorName(nameParts[0]);
+        //Rudimentary name replacement. Can't do nicknames, pronouns, ability words, or flavored keywords. Need to define variants manually for that.
+        if(mainPart.getOracleText().contains(mainPart.getName()))
+            variantMain.setOracleText(mainPart.getOracleText().replace(mainPart.getName(), nameParts[0]));
```
tool4ever, line 508: something doesn't seem to work with this:
image

forge-core/src/main/java/forge/card/CardDb.java
Jetz72: The weird structure here was an attempt to make that method as performant as I reasonably could given that it has to be called tens of thousands of times, and iterating over faces to collect names isn't necessary for probably 99% of PaperCards. 

I went with creating a set of all names and then removing the main name from it so that in the case of a split-card (where the main name is the combined name for multiple faces), it'd be there under both the name of the first face and the combined name without creating duplicate entries for non-split cards other circumstances.

I could probably retool it a bit so it reads more coherently though.

forge-core/src/main/java/forge/card/CardRules.java
Jetz72: Ah, that's because that card doesn't use that. The Universes Within cards were among the first ones I tested the system on. I did them the manual way, adding a variant for their names to their scripts but neglecting to add the variant oracle text. I could go through and fix their scripts, or I could convert them to use the automatic flavor names method. Just need to settle on which implementation to use. I'm leaning toward manual for cases where it looks like a flavor name may be printed more than once (e.g. Universes Within prints), and automatic for one-offs.

forge-core/src/main/java/forge/card/CardDb.java
tool4ever: oh I see, well in that case this seems fine :)

forge-core/src/main/java/forge/card/CardRules.java
tool4ever: oops right, I somehow missed you mentioning that 😅
while I'd prefer having cleaner scripts this doesn't seem important enough to worry blocking about

tool4ever (approved):
LGTM

Hanmac:
@Jetz72 seems Pattern Groups either aren't supported 
https://developer.android.com/reference/java/util/regex/Pattern#group-name
> The constructs and APIs are available since API level 26. 

or something else is funky with the Regex that causes it to fail on android
I don't have a way to test this against Android API versions right now
The exception is pointing at `(?<name>[^@$]*)` group, for whatever reason
(i will add the Exception later)

So, can you remove the named pattern groups for now?

PR #6467, opened 2024-10-30
Add a toggle to reenable declaring combatant ordering

tehdiplomat:

PR #8862, opened 2025-10-08
Fix AI not ordering combatants for damage in non-legacy mode

### LINKED ISSUE #8837: Please Remove the Player's "Pseudo-Banding"
The AI's combat dmg assignment is, in large part, directed by the player's assignment of blockers when it comes to multi-blocking singular attackers. I've noticed the AI will always assign dmg to the 1st blocker assigned & only direct excess dmg to additional blockers in the order those blockers were assigned. This results in overly beneficial combat trades for the player once the player understands how the AI assigns combat dmg.

For example, say the AI is attacking me w/ a 6/6 & I have a 1/2 & a 5/5. If I assign the 1/2 & then the 5/5 as blockers, the end result is a dead 6/6, a dead 1/2, & a 5/5 w/ 4 dmg on it; I effectively traded my 1/2 for their 6/6. Had I assigned the 5/5 & then the 1/2, the end result would be a dead 6/6, a dead 5/5, & a 1/2 w/ 1 dmg on it; the choice of which of my 2 creatures survived is essentially up to me when it should be the opponent's. IRL, against a human opponent, the human would almost invariably choose to kill the 5/5 w/ their 6/6 rather than trade their big boy for a 1/2.

Due to the way the AI assigns combat dmg, the player can garner huge advantages on most board states involving creatures. This places a pressure on deck design for creature-heavy decks &, specifically, deck designs that can efficiently convert mana into stat pts. I do like playing low- or no-creature decks from time to time but it can be hard to justify doing that when the win-rate for playing efficient creature decks is so high. There is a sort of damper on creative deck design when playing a mono-green deck full of dumb, vanilla fatties can almost always win.

You can argue I don't have to exploit the AI when I understand its behavior; I can simply assign my blockers to its benefit rather than detriment. I say that is a non-argument. We have an AI for a reason. If you want to play the game like that, why not take over ALL of the AI's decisions & play both decks @ once against each other?? Also, the temptation to exploit is too great for a person of my low virtue, especially when there's something important on the line like digital bragging rights or imaginary credits I can spend @ the imaginary store.

I'd like to see the AI given more/any agency when it comes to distributing combat dmg to multiple blockers. 1 possible implementation would be to have the AI prioritize blockers w/ a higher mana value. In my above example, the 5/5 would almost invariably have a higher CMC than the 1/2. If the AI were using this system of priority, the 5/5 would end up dead regardless of the order in which I assigned blockers. I'm sure there would be ways to exploit such a system but, as it stands, the current system is constantly & easily exploitable by a player that understands how the AI assigns combat dmg. Heck, even a random re-ordering of blocker assignment prior to dmg using the current logic would probably be better than what we have; @ least you couldn't reliably cheese the AI if you didn't know exactly how it was going to assign dmg.

1 thing I've noticed is when the player assigns combat dmg using the "Auto" button, it follows the AI's logic. Any solution to this problem might affect the player's auto-assignment of combat dmg. I think it'd be fine to leave it as it is but might be better to update the logic for both functions to match. As it is, the "Auto" button has screwed me @ least once when it killed the "wrong" creature in a multi-block.

There are no possible alternatives. The current state of affairs cannot stand. Something must be done!

Agetian:
Fixes #8837 

I'm not sure if orderAttackers should also use this logic (line 553), difficult to imagine when that would be relevant, exactly.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/CountersMoveAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/CountersMoveAi.java
@@ -427,7 +427,7 @@ public class CountersMoveAi extends SpellAbilityAi {
             boolean isNegative = ComputerUtil.isNegativeCounter(cType, src);
             List<Card> filteredTgtList;
             filteredTgtList = isNegative ? CardLists.filterControlledBy(tgtCards, ai.getOpponents()) :
-                CardLists.filter(tgtCards, CardPredicates.isControlledByAnyOf(ai.getAllies()).or(CardPredicates.isController(ai)));
+                CardLists.filterControlledBy(tgtCards, ai.getYourTeam());
 
             if (!filteredTgtList.isEmpty()) {
                 List<Card> best = CardLists.filter(filteredTgtList, card -> {


forge-game/src/main/java/forge/game/combat/Combat.java
```diff
@@ -492,7 +492,7 @@ public void orderBlockersForDamageAssignment() { // this method performs control
 
     /** If there are multiple blockers, the Attacker declares the Assignment Order */
     public void orderBlockersForDamageAssignment(Card attacker, CardCollection blockers) { // this method performs controller's role
-        if (blockers.size() <= 1 || !this.legacyOrderCombatants) {
+        if (blockers.size() <= 1 || (!this.legacyOrderCombatants && !playerWhoAttacks.isAI())) {
```
tool4ever: we really don't need to pollute engine with AI check here (it's against the rules anyway)
AI can just do it all internally in `distributeAiDamage`

Agetian: I can do it inside distributeAiDamage, but that'll involve a possible expensive creation of a new CardCollection, not sure how to work around that tbh

Agetian: Example implementation committed (it works, but a new card collection object is needed on every call)

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
```diff
@@ -2046,8 +2046,15 @@ public static Map<Card, Integer> distributeAIDamage(final Player self, final Car
             // TODO sort remaining tramplers for DamageDone triggers
         }
 
-        if (block.size() == 1) {
-            final Card blocker = block.getFirst();
+        // Order the combatants in preferred order in case legacy ordering is disabled
+        final boolean legacyOrderCombatants = self.getGame().getRules().hasOrderCombatants();
+        final CardCollection orderedBlockers = new CardCollection(block);
+        if (!legacyOrderCombatants) {
+            ComputerUtilCard.sortByEvaluateCreature(orderedBlockers); // assume sorted in case the legacy option is enabled
```
tool4ever: reuse `AiBlockController.orderBlockers`

Agetian: Done :+1:

tool4ever: wrong implementation:
that function isn't void for a reason

forge-game/src/main/java/forge/game/combat/Combat.java
tool4ever: well while most users will probably play with the option turned off still better style to do it inside the if

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
```diff
@@ -2046,8 +2046,15 @@ public static Map<Card, Integer> distributeAIDamage(final Player self, final Car
             // TODO sort remaining tramplers for DamageDone triggers
         }
 
-        if (block.size() == 1) {
-            final Card blocker = block.getFirst();
+        // Order the combatants in preferred order in case legacy ordering is disabled
+        final boolean legacyOrderCombatants = self.getGame().getRules().hasOrderCombatants();
+        final CardCollection orderedBlockers = new CardCollection(block);
+        if (!legacyOrderCombatants) {
+            AiBlockController.orderBlockers(attacker, orderedBlockers); // assume sorted in case the legacy option is enabled
```
tool4ever: since this function handles both you need to call `orderAttackers` when inside that case instead
(though arguably it's rarer and not really different logic right now)

forge-game/src/main/java/forge/game/combat/Combat.java
Agetian: Probably not going to touch Combat.java in this PR, especially considering I'm not the best at knowing how to properly alter the game mechanics files...

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
Agetian: Ah, lol, I was wondering why it actually worked the way it does, looks like it modifies orderedBlockers as a side effect due to the evaluate call, but then does more stuff and returns the final version as a result. Modified :+1:

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
Agetian: How do I determine this from here? Am I correct assuming that if the AI is the blocker, then orderAttackers would need to be called? This one is slightly tricky to get my head around :/

Agetian: (I guess checking if "defender" and "self" are equal could be one way to test if the AI is the blocker inside this call, but not sure)

forge-game/src/main/java/forge/game/combat/Combat.java
tool4ever: no, I meant you are creating a new CardCollection when it isn't always needed...

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
Agetian: Also, need some help with understanding what that other case actually is :)
Am I correct if I assume that "ordering attackers" implies a case where I attack the AI and it blocks with several things? (doesn't make much sense as "ordering attackers" then, but I can't imagine what other case would fit :( )

forge-game/src/main/java/forge/game/combat/Combat.java
Agetian: Hmm, yes, that's exactly what worried me, but I can't figure out how to do this properly :/
I mean, if I declare orderedBlockers inside the "if", it won't always exist outside of it for the sake of the checks below, and I'll have to do something like "if orderedBlockers is null, use block, otherwise use orderedBlockers", not sure if that's optimal. And there is no implicit conversion between CardCollectionView and CardCollection, too.

Agetian: Maybe I can create it inside the if and get a view and replace whatever's in "block" for this purpose, let me try it

Agetian: Ok, if "block" becomes mutable, that makes it possible to work around creating a separate CardCollection unless needed.

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
tool4ever: imo always best to follow the rules so I grabbed the old ones for you:

> 509.2 Second, for each attacking creature that’s become blocked, the active player announces that creature’s damage assignment order, which consists of the creatures blocking it in an order of that player’s choice. (During the combat damage step, an attacking creature can’t assign combat damage to a creature that’s blocking it unless each creature ahead of that blocking creature in its order is assigned lethal damage.) This turn-based action doesn’t use the stack.

> 509.3 Third, for each blocking creature, the defending player announces that creature’s damage assignment order, which consists of the creatures it’s blocking in an order of that player’s choice. (During the combat damage step, a blocking creature can’t assign combat damage to a creature it’s blocking unless each creature ahead of that blocked creature in its order is assigned lethal damage.) This turn-based action doesn’t use the stack.

Agetian: Thanks, though I wish I knew how to properly translate this into code in this method :) Which, sadly, I don't...
Especially considering that we're talking about the case where the legacy rules don't apply, and as far as I understand the new rules, it's possible to deal damage to any creature in a stack of blocking creatures - so, if an attacked creature is blocked by a number of creatures, the attacking player is now able to split the damage whichever way he pleases, not necessarily relying on a particular ordering that the other player sets?

Agetian: At any rate, I really welcome code improvement commits into this branch from anyone who can help, otherwise the AI risks being as stupid as it is right now for a while :(

Agetian: Maybe @tehdiplomat can help here too, as the author of the breaking commit #6467 ? As of right now, the combat AI is really screwed up tbh :) Thanks in advance!

Agetian: As far as I got with this:
* I can see that this method (distributeAIDamage) is called via a call to assignCombatDamage for the AI from two code paths - one is called "assignBlockersDamage" (where "blockers" are ordered), the other is called "assignAttackersDamage" (where "attackers" are ordered)
* It's not immediately apparent how to derive the calling code path from inside distributeAIDamage, as the parameters themselves don't immediately indicate whether the attackers or the blockers are being parsed.
* There probably is some kind of a lightweight and easy way to do this (following the rules outlined above, if relevant - still not sure to what extent they're relevant under the conditions where damage assignment can be done freely to any creature among the blockers), but it eludes me.

Given the extremely tight work schedule that I have, I don't think I can crack this one on my own. Thus, yes, I'll definitely need some help with this.

tool4ever: hmn thanks for trying anyway
guess I need to investigate if it's safe enough to merge in its current state for the time being

Agetian: No problem, if I have a sudden insight into this, I'll definitely improve it.
Initial tests seem to indicate that this is alright (better than what exists at the moment, anyway), but I haven't tested any corner cases and I still don't have a test case for "ordering attackers", so yeah, more testing is definitely welcome. Hopefully we can improve it to be more rule-correct / expand it to account for the second use case as well, but that'll probably require some additional help from someone else able and willing to join in...

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
```diff
@@ -2046,6 +2046,12 @@ public static Map<Card, Integer> distributeAIDamage(final Player self, final Car
             // TODO sort remaining tramplers for DamageDone triggers
         }
 
+        // Order the combatants in preferred order in case legacy ordering is disabled
+        if (!self.getGame().getRules().hasOrderCombatants()) {
+            // TODO: sometimes, orderAttackers needs to be called instead - differentiate this here somehow
```
tehdiplomat: I have a feeling the names of `attacker` and `block` are misnomers here, so we should be determining if the params are attacking or blocking

```suggestion
            if (combatant.isAttacking()) { 
                opposedCombatants = AiBlockController.orderBlockers(combatant, new CardCollection(opposedCombatants));
            } else {
                 opposedCombatants = AiBlockController.orderAttackers(combatant, opposedCombatants);
            }
```

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
tehdiplomat: Oh yea looks like you guys talked this through. I think my suggestion works, along with renaming the params. Thanks for spotting this one. I noticed it a few weeks ago, but didn't realize it was related to the rules change.

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
Agetian: Agree with this change, @tehdiplomat , committed! Thanks a lot for help! :+1:

forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
Agetian: Yes, agreed! No prob at all, thank you for your help and input!! Much appreciated!

Agetian: Tried to update the code to match the parameter names , seems like it works but probably needs some extra testing :/

PR #8867, opened 2025-10-08
Check for Snow lands and Wastes when generating random decks

Eradev:


forge-gui-mobile/src/forge/adventure/util/CardUtil.java
```diff
@@ -482,176 +451,185 @@ public static Deck generateDeck(GeneratedDeckData data, CardEdition starterEditi
...
                 landTypes.add("Plains");
-            if(swamp>0)
+            if (swamp > 0)
                 landTypes.add("Swamp");
-            if(forest>0)
+            if (forest > 0)
                 landTypes.add("Forest");
-            cards.addAll(generateDualLands(landTypes,neededDualLands));
 
+            cards.addAll(generateDualLands(landTypes, neededDualLands));
```
Hanmac, line 523: @Eradev should the `generateDualLands` function check for `useSnowLands` too?

Eradev: I didn't put it in generateDualLands as it could severely limit the possibilities.

tehdiplomat:
In the future can you commit styling changes as a different commit? Its really hard to tell what actually changed in this PR.

PR #8895, opened 2025-10-10
15 TMT/TMC cards

Fulgur14:


forge-gui/res/cardsfolder/upcoming/casey_jones_jury_rig_justiciar.txt
```diff
@@ -0,0 +1,7 @@
+Name:Casey Jones, Jury-Rig Justiciar
+ManaCost:1 R
+Types:Legendary Creature Human Berserker
+PT:2/1
```
tool4ever, line 4: KW missing

forge-core/src/main/java/forge/card/CardRules.java
```diff
@@ -348,7 +351,7 @@ public boolean canBePartnerCommander() {
         }
         return canBeCommander() && (hasKeyword("Partner") || !this.partnerWith.isEmpty() ||
                 hasKeyword("Friends forever") || hasKeyword("Choose a Background") ||
-                hasKeyword("Partner - Father & Son") || hasKeyword("Partner - Survivors") ||
+                hasKeyword("Partner—Father & Son") || hasKeyword("Partner—Survivors") || hasKeyword("Partner—Character select") ||
```
tool4ever: why switch to the wrong formatting here?

forge-gui/res/cardsfolder/upcoming/donnie_april_adorkable_duo.txt
```diff
@@ -0,0 +1,9 @@
+Name:Donnie & April, Adorkable Duo
+ManaCost:4 U
+Types:Legendary Creature Mutant Ninja Human Turtle
+PT:3/3
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigCharm | TriggerDescription$ When NICKNAME enters, ABILITY
+SVar:TrigCharm:DB$ Charm | MinCharmNum$ 1 | CharmNum$ 2 | Choices$ DBDraw,DBChangeZone | AdditionalDescription$ or both. Each mode must target a different player.
+SVar:DBDraw:DB$ Draw | NumCards$ 2 | ValidTgts$ Player | TargetUnique$ True | SpellDescription$ Target player draws two cards.
+SVar:DBChangeZone:DB$ ChangeZone | ValidTgts$ Player | TargetUnique$ True | Hidden$ True | Mandatory$ True | ChangeType$ Artifact.TargetedPlayerOwn,Instant.TargetedPlayerOwn,Sorcery.TargetedPlayerOwn | ChangeTypeDesc$ artifact, instant or sorcery | Chooser$ Targeted | ChangeNum$ 1 | Origin$ Graveyard | Destination$ Hand | SpellDescription$ Target player returns an artifact, instant, or sorcery card from their graveyard to their hand.
```
tool4ever: bad Chooser$

forge-core/src/main/java/forge/card/CardRules.java
Fulgur14: I *think* I eventually left it as it was, the reason is that I wasn't 100% sure if the long dash would work in the keyword. I didn't want to break anything.

forge-gui/res/cardsfolder/upcoming/donnie_april_adorkable_duo.txt
Fulgur14: In what way? I took this code from Obscura Confluence, which uses "Chooser$ Targeted" in this way.

forge-core/src/main/java/forge/card/CardRules.java
tool4ever: then please revert/align

forge-gui/res/cardsfolder/upcoming/donnie_april_adorkable_duo.txt
tool4ever: _Obscura Confluence_ can only target one player anyway

Fulgur14: Ah, I understand now. Would this work? I can't find the example of proper syntax.

forge-core/src/main/java/forge/card/CardRules.java
Fulgur14: Currently, "Partner - Survivors" etc. are there. I made two changes to the file in a single commit, that might cause it to not display correctly, but when you open it, it's there... Should it be changed?

tool4ever: this line is exactly like Github displays
Hanmac will eventually switch it to a more parameterized version anyway

Fulgur14: When I open it for editing, it looks like this:
" if (hasKeyword("Partner - Survivors") && b.hasKeyword("Partner - Survivors")) {
            legal = true; // The Last of Us Secret Lair gimmick partner commander
        }"

tool4ever: that's not the line I've commented on here...

Fulgur14: Ah, finally found it, sorry :)

Issue #8901, opened 2025-10-11
AI helps opponent create an army


CTimmerman:
AI doesn't properly assess the danger of Darien, King of Kjeldor; Kazuul, Tyrant of the Cliffs etc.

Patch coming up.

PR #8902, opened 2025-10-11
Fix AI helps opponent create an army (#8901)

CTimmerman:
Only decisively attack a player who generates tokens on being attacked or damaged.

image
Now Darien & Kazuul mostly just delay the inevitable instead of turning the tide.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -974,6 +974,16 @@ public class AiAttackController {
             return aiAggression;
         }
 
+        // Only do decisive attacks against token-generating players
+        if (!bAssault && defender instanceof Player) {
+            Player opponent = (Player)defender;
+            if (CardLists.count(ai.getCardsIn(ZoneType.Battlefield), CardPredicates.nameEquals("Rabble Rousing"))
+                - CardLists.count(opponent.getCardsIn(ZoneType.Battlefield), CardPredicates.nameEquals("Darien, King of Kjeldor"))
+                - CardLists.count(opponent.getCardsIn(ZoneType.Battlefield), CardPredicates.nameEquals("Kazuul, Tyrant of the Cliffs")) < 0) {
+                    return aiAggression;
+                }
+        }
+
         if (bAssault && defender == defendingOpponent) { // in case we are forced to attack someone else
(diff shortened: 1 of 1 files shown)


forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -850,6 +850,23 @@ public final int declareAttackers(final Combat combat) {
         // Determine who will be attacked
         GameEntity defender = chooseDefender(combat, bAssault);
 
+        // Avoid creating an enemy army
+        // Darien, King of Kjeldor; Kazuul, Tyrant of the Cliffs etc
+        if (!bAssault && defender instanceof Player) {
```
tool4ever: this is too early and will lead to illegal plays in presence of attack requirements

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -850,6 +850,23 @@ public final int declareAttackers(final Combat combat) {
...
+            for (Card c : ((Player)defender).getCardsIn(ZoneType.Battlefield)) {
+                for (Trigger t : c.getTriggers()) {
+                    if ((TriggerType.Attacks.equals(t.getMode())
```
tool4ever: this seems too broad and also overlaps partly with more precise checks in `doRevengeOfRavensAttackLogic` and `isEffectiveAttacker` anyway

while it's usually preferable to try and parse it by code alone I wonder if a name-based check will be cleaner here
- or if there are some more cards maybe `AiCombatAggressionMod` SVar approach would even be beneficial

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -974,6 +974,16 @@ && isEffectiveAttacker(ai, attacker, combat, finalDefender)) {
             return aiAggression;
         }
 
+        // Only do decisive attacks against token-generating players
+        if (!bAssault && defender instanceof Player) {
+            Player opponent = (Player)defender;
+            int gains = 0;
+            if (ai.isCardInPlay("Rabble Rousing")) gains += 1;
```
tool4ever: should at least count the actual amount of these cards being present

tool4ever (dismissed):
good enough for now

PR #8908, opened 2025-10-12
Limit edition range for Adventure drafts

tehdiplomat:


Agetian:
Looks good, but needs a conflict resolution, it seems...

Jetz72:
Oh, I didn't notice this was open when I added it in #9096. Close as redundant then?

jumpinjackie:
I noticed the predicates in both are different.

This one, each predicate subsequently negates the preceding one, while the one in #9096 each predicate just negates the "standard" set predicate, so it looks like the "vintage" predicate will still have possibility of "modern" sets in it, and the "modern" predicate will still have possibility of "pioneer" sets in it. etc.

Jetz72:
Yes, I discussed the reasoning behind doing it that way in the other PR. We can still change it if we want, though if anything, it might be easier to drop the random format filters in favor of a single weighted selection that we can bump up and down based on edition characteristics.

jumpinjackie:
Ah, point taken.

PR #8915, opened 2025-10-12
TMT: Add Sneak

Hanmac:
Add Sneak Keyword

TODO:

- [ ] For `Leonardo, Sewer Samurai` i need to add Keyword Formating
- [x] For `Raphael's Technique`, i wish the Each Player may Discard their Hand would be coded without GenericChoice.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/raphaels_technique.txt
+++ b/forge-gui/res/cardsfolder/upcoming/raphaels_technique.txt
@@ -2,10 +2,7 @@ Name:Raphael's Technique
 ManaCost:4 R R
 Types:Instant
 K:Sneak:2 R
-A:SP$ GenericChoice | TempRemember$ Chooser | ShowChoice$ ExceptSelf | Defined$ Player | Choices$ Discard,No | SubAbility$ DBDiscard | SpellDescription$ Each player may discard their hand and draw seven cards.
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/spellability/SpellAbilityRestriction.java
```diff
@@ -337,6 +337,11 @@ public final boolean checkTimingRestrictions(final Card c, final SpellAbility sa
                 return false;
             }
         }
+        if (sa.isSneak()) {
+            if (!game.getPhaseHandler().is(PhaseType.COMBAT_DECLARE_BLOCKERS)) {
```
tehdiplomat, line 341: Any reason this is defined here instead of during the creation of the SA?

Hanmac: In case a spell with Valid Phases gains that Keyword somehow (I know Wizards are going to do that)

Hanmac:
Sneak on a Creature seems to be like Ninjitsu, but still a Spell:

https://markrosewater.tumblr.com/post/797115474836094976/if-i-cast-a-creature-using-its-sneak-cost-does-it

Hanmac:
@tool4ever i added Sneak part for Creatures,
someone else might add Script for `Leonardo, Sewer Samurai`

tool4ever:
LGTM, just need scripts finalized

PR #8916, opened 2025-10-13
Proposition: Put some expensive queries into cache

Eradev:
I've played against the Nightmare deck, and it was very laggy as it tried to count its Swamps over and over again. This should alleviate the problem by caching some expensive queries, which are cleared every time the zones are changed. Even if it happens often, it will still only calculate it once per reset rather than 3-4 times.

What do you think?

### DIFF
--- a/forge-game/src/main/java/forge/game/zone/Zone.java
+++ b/forge-game/src/main/java/forge/game/zone/Zone.java
@@ -68,7 +68,7 @@ public class Zone implements java.io.Serializable, Iterable<Card> {
     }
 
     protected void onChanged() {
-        game.clearPlayersCache();
+        game.clearShortLivedCaches();
     }
 
     public Player getPlayer() { // generic zones like stack have no player associated
--- a/forge-game/src/main/java/forge/game/Game.java
+++ b/forge-game/src/main/java/forge/game/Game.java
@@ -825,7 +825,7 @@ public class Game {
     }
 
     public void onPlayerLost(Player p) {
-        clearPlayersCache();
+        clearShortLivedCaches();
         //set for Avatar
         p.setHasLost(true);
         // Rule 800.4 Losing a Multiplayer game
@@ -1381,9 +1381,10 @@ public class Game {
     }
 
     /**
-     * Reset players' cache.
+     * Reset short lived caches that should not persist between turns or phases.
      */
-    public void clearPlayersCache() {
-        getPlayers().forEach(p -> p.resetCache());
+    public void clearShortLivedCaches() {
+        getPlayers().forEach(p -> p.clearCache());
+        phaseHandler.clearCache();
     }
 }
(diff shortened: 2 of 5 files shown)


Hanmac:
Calculate Amount needs some refactoring anyway because right now you need to do "put into cache" at multiple locations


Might refactor it into a Supplier with `computeIfAbsent`

forge-game/src/main/java/forge/game/zone/Zone.java
```diff
@@ -68,6 +68,7 @@ public Zone(final ZoneType zone0, Game game0) {
     }
 
     protected void onChanged() {
+        game.resetCache();
```
tool4ever: I don't think this will work
while zone change is one of the more common ways for the board state to change there are just too many others, e.g. using _Orcish Farmer_ (though technically we might see this one because it already uses effect card)
by the time you're done hooking this up to all possible places I'd expect you end up with way more than 100 calls

Eradev: ~~I believe that tapping does trigger a change as well.~~

EDIT: Now it does.

Eradev:
> Calculate Amount needs some refactoring anyway because right now you need to do "put into cache" at multiple locations
> 
> Might refactor it into a Supplier with `computeIfAbsent`

There are some instances you don't want to cache (like the random one) so that's why it is targeted ATM. I could add a utility function to prevent some code repeat.

Hanmac:
While looking at #8842 i noticed, that Player Properties can't be chained endlessly in xCount

it really didn't like this, what i tried:
`PlayerCountPropertyYou$DungeonsCompleted$DifferentCardNames/Twice`

Also for another issue, we need to think about how stuff like `/Twice` or `/Min` are applied.

forge-game/src/main/java/forge/game/zone/Zone.java
tool4ever: how...?

Eradev: After testing, Zone already did trigger when tapping. Both the Battlefield and Graveyard apparently.

image

Eradev:
As it is right now, it mostly helps with decks with multiple cards that checks for the same values, or have copies of itself (such as Relentless Rats decks). But I do see multiple cache hits when playing a single copy of Wight of Precinct Six.

I thought about invalidating cache values depending on the updated zone(s), but I'm not sure about the added complexity.

forge-game/src/main/java/forge/game/zone/Zone.java
tool4ever: hmn I see
so reset will happen every mainLoopStep at least

tool4ever:
your assumption of same string = same value is on shaky ground :P
what about something like _Acidic Soil_ with:
``SVar:X:Count$Valid Land.RememberedPlayerCtrl``

Eradev:
> your assumption of same string = same value is on shaky ground :P what about something like _Acidic Soil_ with: `SVar:X:Count$Valid Land.RememberedPlayerCtrl`

I don't think it'll matter for spells as they'll trigger cache reset when the stack gets changed.

But I see your point that remembered player can be different. I'll see if I can do anything about it.

PR #8926, opened 2025-10-14
Fix Burning-Rune Demon able to find only one card

tool4ever:


Jetz72:
If the user selects fewer than the number required, will this just cause the effect to fizzle? Wonder if it makes sense to supply a minimum to `chooseCardsForZoneChange` when `Exactly` is set.

tool4ever:
fizzling only happens when all targets become illegal
right now `allowMultiSelect` will not allow the cases that use it, but I'll think about how much effort changing that would require

PR #8930, opened 2025-10-15
Clean up Ruin Grinder

tool4ever:


Hanmac:
`Defined$ Player.IsRemembered`
to
`Defined$ Remembered`

Might be better?

Hanmac:
i think this looks good? (hadn't yet time to test it myself)

i will update my MR #8915 after this

PR #8931, opened 2025-10-15
Update the_lost_and_the_damned.txt

Fulgur14:
Added a corner case excluded by the latest fix. (Not sure if it's even possible, but might be in the future.)

EDIT: Theoretically would cover situation where opponent plays Dryad Arbor when under the effect of Gather Specimens, but that would normally only happen when the opponent is controlled :)

EDIT2: Or, grabbing Dryad Arbor with Treacherous Urge.


Hanmac:
@tool4ever you should check out this corner fix

tool4ever (approved):
good catch

it's a bit messy having this lead to a second trigger but I wasn't fully convinced using `ThisTurnEnteredFrom` instead is really always safe in some weird scenario

PR #8941, opened 2025-10-17
KeywordCollection: Use linkedHashSetValues to prevent duplicates

Hanmac:
As mentioned there: https://github.com/Card-Forge/forge/pull/8587#issuecomment-3415238141

KeywordInterfaces should not be duplicates anyway


tool4ever (approved):
this already stops the doubled Herigast emerge

PR #8960, opened 2025-10-19
ColorSet: add combine function

Hanmac:
Part of #8682 

This part replaces the color in CardState with ColorSet

and adds `ColorSet.combine` as helper to combine multiple ColorSets

### DIFF
--- a/forge-game/src/main/java/forge/game/card/token/TokenInfo.java
+++ b/forge-game/src/main/java/forge/game/card/token/TokenInfo.java
@@ -63,7 +63,7 @@ public class TokenInfo {
         String[] types = null;
         String[] keywords = null;
         String imageName = null;
-        ColorSet color = ColorSet.C;
+        ColorSet color = null;
         for (String info : tokenInfo) {
             int index = info.indexOf(':');
             if (index == -1) {


forge-game/src/main/java/forge/game/card/CardFactory.java
```diff
@@ -343,8 +343,7 @@ private static Card readCard(final CardRules rules, final IPaperCard paperCard,
             card.setManaCost(combinedManaCost);
 
             // Combined card color
-            final byte combinedColor = (byte) (rules.getMainPart().getColor().getColor() | rules.getOtherPart().getColor().getColor());
-            card.setColor(combinedColor);
+            card.setColor(ColorSet.combine(rules.getMainPart().getColor(), rules.getOtherPart().getColor()));
```
tool4ever: can just call `rules.getColor` anyway?
same with the ManaCost above?

forge-game/src/main/java/forge/game/card/token/TokenInfo.java
```diff
@@ -115,7 +118,7 @@ private Card toCard(Game game, int id) {
         c.setName(name);
         c.setImageKey(ImageKeys.getTokenKey(imageName));
 
-        c.setColor(color.isEmpty() ? manaCost : color);
+        c.setColor(color == null ? ColorSet.fromManaCost(new ManaCost(new ManaCostParser(manaCost))) : color);
```
tool4ever, line 121: seems like this fallback would no longer work correctly since that's not your new default value

Hanmac: We really need to see to try to remove this TokenInfo thing

it seems it is only used in some places of the AI
(GameState and GameCopier)

PR #8962, opened 2025-10-20
Corrected Traumatic Prank's mana cost.

Darthseid:
It was previously 3R. You can see the actual cost and effects here: https://scryfall.com/card/ysnc/12/traumatic-prank


Hanmac:
@Darthseid did the rebalance got reverted?
The ManaCost got changed for `May 30, 2023`

Hanmac:
@paulsnoops can you take a look at these Arena rebalanced cards?

paulsnoops:
Looks like that was part of these cards that got reverted when rotation happened: https://www.reddit.com/r/MagicArena/comments/16b18t8/a_bunch_of_cards_were_unnerfed_for_rotation_to/

We never did get around to updating them

PR #8967, opened 2025-10-20
AbilityUtils: refactor Party types

### LINKED ISSUE #8963: Party Types cleanup
Right now, there are multiple places that use `"Cleric", "Rogue", "Warrior", "Wizard"` for the Party Types

move them to the CardType Constants like the Outlaw Types

and maybe update the checks to not call `getAllCreatureTypes` if able

Hanmac:
Closes #8963 

i did try to cleanup the Party Calc Loop
trying to avoid to iterate over all the cards multiple times.

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -2603,8 +2603,6 @@ public class AbilityUtils {
         }
 
         if (sq[0].contains("Party")) {
-            int partySize = 0;
-
             Set<String> chosenParty = Sets.newHashSet();
             int wildcard = 0;
             ListMultimap<String, Card> multityped = MultimapBuilder.hashKeys().arrayListValues().build();
@@ -2616,11 +2614,13 @@ public class AbilityUtils {
                     continue;
                 }
                 CardTypeView type = card.getType();
-                // cards with all creature types will just return full list
                 Set<String> creatureTypes;
+
+                // extra logic for "all creature types" cards
                 if (type.hasAllCreatureTypes()) {
+                    // one of the party types could be excluded, so check each of them separate
                     creatureTypes = CardType.Constant.PARTY_TYPES.stream().filter(p -> type.hasCreatureType(p)).collect(Collectors.toSet());
-                } else {
+                } else { // shortcut for others 
                     creatureTypes = type.getCreatureTypes();
                     creatureTypes.retainAll(CardType.Constant.PARTY_TYPES);
                 }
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -2596,59 +2597,58 @@ public static int xCount(Card c, final String s, final CardTraitBase ctb) {
...
+            partySize = Math.min(chosenParty.size() + wildcard, 4);
 
             if (partySize < 4) {
-                partyTypes.removeAll(chosenParty.keySet());
+                partyTypes.removeAll(chosenParty);
 
                 // Here I'm left with just the party types that I haven't selected.
-                for (Card multi : multityped.keySet()) {
-                    Set<String> types = multityped.get(multi);
+                for (Set<String> types : multityped) {
```
tool4ever: hmn not a huge fan of combinatorics but I strongly suspect there are corner cases where the logic still fails 🤔

- Wizard
- Cleric Warrior
- Cleric Wizard

partySize should be 3, but the code only results in 2 since it would count Cleric first already for the second creature

now this wouldn't be a problem if the order of the multityped wasn't always fixed, which might give you an idea on how to tweak it

Hanmac: i will think about it

i probably can filter/group the `multityped` list via still missing creature types

so because "Wizard" gets picked first, it will be sorted like this, "Cleric ~~Wizard~~", "Cleric Warrior"
and then Cleric will be picked next

Hanmac: @tool4ever something as example i'm going to refactor soon:

```java

import java.util.*;
import java.util.Map.Entry;
import java.util.regex.Matcher;
import java.util.regex.Pattern;

public class Main {
            public static final Set<String> PARTY = new HashSet(Arrays.asList(
                "Cleric",
                "Rogue",
                "Warrior",
                "Wizard"));
    public static void main(String[] args) {
        System.out.println("Hello, World!");
        System.out.println(Runtime.version());
        Map<String, Set<String>> testData = new LinkedHashMap<>();
        //testData.put("A", new HashSet(Arrays.asList("Human", "Wizard")));
        testData.put("C", new HashSet(Arrays.asList("Human", "Cleric", "Wizard")));
        testData.put("B", new HashSet(Arrays.asList("Human", "Cleric", "Warrior")));
        //testData.put("D", new HashSet(Arrays.asList("Human", "Rogue", "Warrior")));
        //testData.put("D", new HashSet(Arrays.asList("Human", "Cleric", "Warrior")));
        System.out.println(testData);

        Set<String> chosenParty = new HashSet();
        int wildcard = 0;
        Map<String, List<String>> multi = new LinkedHashMap();
        Set<String> chosenMulti  = new HashSet();
        for (Map.Entry<String, Set<String>> e : testData.entrySet()) {
            Set<String> types = e.getValue();
            types.retainAll(PARTY);

            switch (types.size()) {
                case 4:
                    wildcard++;
                    break;
                case 1:
                    chosenParty.addAll(types);
                    break;
                default:
                    for (String t : types) {
                        multi.computeIfAbsent(t, v -> new ArrayList<>()).add(e.getKey());
                    }
            }
        }

        multi.keySet().removeAll(chosenParty);

        // sort multi list by number of entries

        multi.entrySet().stream()
        .sorted(Map.Entry.<String, List<String>>comparingByValue(
            Comparator.<List<String>>comparingInt(Collection::size)))
        //.sorted(Map.Entry.<>comparingByValue(Comparator<>.comparingInt​(Collection::size)))
        .forEach( e -> {
            e.getValue().removeAll(chosenMulti);
            System.out.println(e.getValue());
            if (e.getValue().size() > 0) {
                chosenParty.add(e.getKey());
                chosenMulti.add(e.getValue().get(0));
            }
        });

        multi.keySet().removeAll(chosenParty);

            System.out.println(chosenParty);
            System.out.println(wildcard);
            System.out.println(multi);
            System.out.println(chosenMulti);

    }
}

```

i would like it if i can use comparingByValue without the whole types

Hanmac: @tool4ever can you checkout the new version and what you think of it?

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -2601,61 +2603,65 @@ public static int xCount(Card c, final String s, final CardTraitBase ctb) {
...
-                        chosenParty.put(type, card);
+                case 4:
+                    wildcard++;
+                    break;
+                case 1:
+                    chosenParty.addAll(creatureTypes);
+                    break;
+                default:
+                    for (String t : creatureTypes) {
+                        multityped.put(t, card);
```
tool4ever, line 2639: doesn't putAll work there?

Hanmac: The mapping isn't card -> creatureTypes,
but type -> cards

PR #8970, opened 2025-10-21
Implemented Commander variant of Adventure Mode

jgroehl:
There was an open project on the Wiki for this - here is an implementation for it.
I tested it through some play and it works and at least for me is a lot of fun!
Through a short discussion with @kazyfreak [here](https://github.com/jgroehl/forge/issues/1), I reverted any enforced restrictions I wanted to put in place initially. 
At the end of the day, the player can now have a commander, is supplied with a starting commander + deck, and has full deckbuilding flexibility.

Please let me know if you think and if this could be included into `main`!

Summary of code changes:
- New Adventure Scene: 
  - Added "Commander" option.
  - Supplies player with 15 (random) startup commanders.
  - Communicates the chosen commander to the world creation.
  - Added a small card preview (hijacking the RewardActor) so the player can see the chosen card.
- World Creation
  - logical flow fork if a commander has been chosen
- Adventure Player
  - Now stores a flag whether Adventure is played in commander mode.
- Config (most of the added lines of code)
  - added a function to heuristically add a somewhat functional random deck pile for the chosen starting commander.
- Duel Scene
  - Started in commander mode if adventure is in the commander variant
- Reward Screen
  - Added a convenience function where cards that are already possessed are flagged for autosell automatically.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
@@ -500,7 +500,7 @@ public class ComputerUtilCard {
         }
 
         List<Card> lands = CardLists.filter(list, CardPredicates.LANDS);
-        if (lands.size() > 6) {
+        if (lands.size() > 6 || lands.size() == Iterables.size(list)) {
             return getWorstLand(lands);
         }
 
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
@@ -487,14 +487,14 @@ public class ComputerUtilMana {
         String originalProduced = manaProduced.toString();
 
         if (originalProduced.isEmpty()) {
-            return manaProduced.toString();
+            return originalProduced;
         }
 
         // Run triggers like Nissa
         final Map<AbilityKey, Object> runParams = AbilityKey.mapFromCard(hostCard);
         runParams.put(AbilityKey.Activator, ai); // assuming AI would only ever gives itself mana
         runParams.put(AbilityKey.AbilityMana, saPayment);
-        runParams.put(AbilityKey.Produced, manaProduced.toString());
+        runParams.put(AbilityKey.Produced, originalProduced);
         for (Trigger tr : ai.getGame().getTriggerHandler().getActiveTrigger(TriggerType.TapsForMana, runParams)) {
             SpellAbility trSA = tr.ensureAbility();
             if (trSA == null) {
@@ -1292,11 +1292,11 @@ public class ComputerUtilMana {
      * @return ManaCost
      */
     public static ManaCostBeingPaid calculateManaCost(final Cost cost, final SpellAbility sa, final boolean test, final int extraMana, final boolean effect) {
-        Card card = sa.getHostCard();
+        Card host = sa.getHostCard();
         Zone castFromBackup = null;
-        if (test && sa.isSpell() && !card.isInZone(ZoneType.Stack)) {
-            castFromBackup = card.getCastFrom();
-            card.setCastFrom(card.getZone() != null ? card.getZone() : null);
+        if (test && sa.isSpell() && !host.isInZone(ZoneType.Stack)) {
+            castFromBackup = host.getCastFrom();
+            host.setCastFrom(host.getZone() != null ? host.getZone() : null);
         }
 
         Cost payCosts;
@@ -1319,7 +1319,7 @@ public class ComputerUtilMana {
                 final int multiplicator = Math.max(xCounter, 1);
                 manaToAdd = extraMana * multiplicator;
             } else {
-                manaToAdd = AbilityUtils.calculateAmount(card, sa.getParamOrDefault("XAlternative", "X"), sa) * xCounter;
+                manaToAdd = AbilityUtils.calculateAmount(host, sa.getParamOrDefault("XAlternative", "X"), sa) * xCounter;
             }
 
             if (manaToAdd < 1 && payCosts != null && payCosts.getCostMana().getXMin() > 0) {
@@ -1331,7 +1331,7 @@ public class ComputerUtilMana {
             if (xColor == null) {
                 xColor = "1";
             }
-            if (card.hasKeyword("Spend only colored mana on X. No more than one mana of each color may be spent this way.")) {
+            if (host.hasKeyword("Spend only colored mana on X. No more than one mana of each color may be spent this way.")) {
                 xColor = "WUBRGX";
             }
             if (xCounter > 0) {
@@ -1356,14 +1356,14 @@ public class ComputerUtilMana {
                 mCost = ManaCost.combine(mCost, mkCost);
                 ManaCostBeingPaid mcbp = new ManaCostBeingPaid(mCost);
                 if (!canPayManaCost(mcbp, sa, sa.getActivatingPlayer(), true)) {
-                    sa.getHostCard().setSVar("NumTimes", "Number$" + i);
+                    host.setSVar("NumTimes", "Number$" + i);
                     break;
                 }
             }
         }
 
-        if (test && sa.isSpell()) {
-            sa.getHostCard().setCastFrom(castFromBackup);
+        if (test && sa.isSpell() && !host.isInZone(ZoneType.Stack)) {
+            host.setCastFrom(castFromBackup);
         }
 
         return manaCost;
@@ -1596,6 +1596,7 @@ public class ComputerUtilMana {
                 }
 
                 // don't use abilities with dangerous drawbacks
+                // TODO this has already been checked earlier
                 AbilitySub sub = m.getSubAbility();
                 if (sub != null && !SpellApiToAi.Converter.get(sub).chkDrawbackWithSubs(ai, sub).willingToPlay()) {
                     continue;
--- a/forge-ai/src/main/java/forge/ai/ability/AnimateAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/AnimateAi.java
@@ -411,15 +411,13 @@ public class AnimateAi extends SpellAbilityAi {
         }
 
         if (logic.equals("ValuableAttackerOrBlocker")) {
-            if (ph.inCombat()) {
-                final Combat combat = ph.getCombat();
-                for (Card c : list) {
-                    Card animated = becomeAnimated(c, sa);
-                    boolean isValuableAttacker = ph.is(PhaseType.MAIN1, ai) && ComputerUtilCard.doesSpecifiedCreatureAttackAI(ai, animated);
-                    boolean isValuableBlocker = combat != null && combat.getDefendingPlayers().contains(ai) && ComputerUtilCard.doesSpecifiedCreatureBlock(ai, animated);
-                    if (isValuableAttacker || isValuableBlocker)
-                        return new AiAbilityDecision(100, AiPlayDecision.WillPlay);
-                }
+            final Combat combat = ph.getCombat();
+            for (Card c : list) {
+                Card animated = becomeAnimated(c, sa);
+                boolean isValuableAttacker = ph.is(PhaseType.COMBAT_BEGIN, ai) && ComputerUtilCard.doesSpecifiedCreatureAttackAI(ai, animated);
+                boolean isValuableBlocker = combat != null && combat.getDefendingPlayers().contains(ai) && ComputerUtilCard.doesSpecifiedCreatureBlock(ai, animated);
+                if (isValuableAttacker || isValuableBlocker)
+                    return new AiAbilityDecision(100, AiPlayDecision.WillPlay);
             }
         }
 
--- a/forge-core/src/main/java/forge/card/CardDb.java
+++ b/forge-core/src/main/java/forge/card/CardDb.java
@@ -389,6 +389,7 @@ public final class CardDb implements ICardDatabase, IDeckGenPool {
             String normalizedFlavorName = cr.getDisplayNameForVariant(variantName);
             if(!flavorNameMappings.containsKey(normalizedFlavorName)) {
                 flavorNameMappings.put(normalizedFlavorName, variantName);
+                flavorNameMappings.put(cr.getName(), IPaperCard.NO_FUNCTIONAL_VARIANT);
                 rulesByName.put(normalizedFlavorName, cr);
                 cacheFlavorName(cr.getMainPart().getFunctionalVariant(variantName));
                 if(cr.getOtherPart() != null)
--- a/forge-game/src/main/java/forge/game/GameActionUtil.java
+++ b/forge-game/src/main/java/forge/game/GameActionUtil.java
@@ -879,14 +879,12 @@ public final class GameActionUtil {
         } else if (abMana.isComboMana()) {
             // amount is already taken care of in resolve method for combination mana, just append baseMana
             sb.append(baseMana);
+        } else if (StringUtils.isNumeric(baseMana)) {
+            sb.append(amount * Integer.parseInt(baseMana));
         } else {
-            if (StringUtils.isNumeric(baseMana)) {
-                sb.append(amount * Integer.parseInt(baseMana));
-            } else {
-                sb.append(baseMana);
-                for (int i = 1; i < amount; i++) {
-                    sb.append(" ").append(baseMana);
-                }
+            sb.append(baseMana);
+            for (int i = 1; i < amount; i++) {
+                sb.append(" ").append(baseMana);
             }
         }
         return sb.toString();
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -3419,7 +3419,6 @@ public class AbilityUtils {
     }
 
     public static int playerXProperty(final Player player, final String s, final Card source, CardTraitBase ctb) {
-
         final String[] l = s.split("/");
         final String m = CardFactoryUtil.extractOperators(s);
 
--- a/forge-game/src/main/java/forge/game/ability/effects/ManaReflectedEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/ManaReflectedEffect.java
@@ -103,12 +103,10 @@ public class ManaReflectedEffect extends SpellAbilityEffect {
                     return "0";
                 } else if (colors.size() == 1) {
                     baseMana = MagicColor.toShortString(colors.iterator().next());
+                } else if (colors.contains("colorless")) {
+                    baseMana = MagicColor.toShortString(player.getController().chooseColorAllowColorless(Localizer.getInstance().getMessage("lblSelectManaProduce"), sa.getHostCard(), ColorSet.fromNames(colors)));
                 } else {
-                    if (colors.contains("colorless")) {
-                        baseMana = MagicColor.toShortString(player.getController().chooseColorAllowColorless(Localizer.getInstance().getMessage("lblSelectManaProduce"), sa.getHostCard(), ColorSet.fromNames(colors)));
-                    } else {
-                        baseMana = MagicColor.toShortString(player.getController().chooseColor(Localizer.getInstance().getMessage("lblSelectManaProduce"), sa, ColorSet.fromNames(colors)));
-                    }
+                    baseMana = MagicColor.toShortString(player.getController().chooseColor(Localizer.getInstance().getMessage("lblSelectManaProduce"), sa, ColorSet.fromNames(colors)));
                 }
             } else {
                 colorMenu = ColorSet.fromMask(mask);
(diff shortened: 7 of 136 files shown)


forge-gui-mobile/src/forge/deck/FDeckEditor.java
```diff
@@ -1084,7 +1085,7 @@ public boolean isLimitedEditor() {
         return editorConfig.isLimited();
     }
     public boolean isCommanderEditor() {
-        return editorConfig.hasCommander();
+        return AdventurePlayer.current().isCommanderMode() || editorConfig.hasCommander();
```
Jetz72: I think this will crash the deck editor outside Adventure mode. Either way, Adventure specific stuff should be in `AdventureDeckEditor`. Specifically you'll want to make sure `AdventureEditorConfig` (or an another subclass of `DeckEditorConfig`) is supplying the appropriate values from its methods.

forge-gui-mobile/src/forge/adventure/util/Config.java
```diff
@@ -255,10 +259,160 @@ public Deck starterDeck(ColorSet color, DifficultyData difficultyData, Adventure
                         return CardUtil.getDeck(entry.value, false, false, "", false, false);
                     }
                 }
+            case Commander:
+                return buildCommanderPile(commander);
```
tool4ever: should use `DeckgenUtil.generateCommanderDeck` instead and extend from that if needed

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
```diff
@@ -49,10 +49,13 @@ public class AdventurePlayer implements Serializable, SaveFileContent {
 
     // Deck data
     private Deck deck;
-    private final ArrayList<Deck> decks = new ArrayList<Deck>(MIN_DECK_COUNT);
+    private final ArrayList<Deck> decks = new ArrayList<>(MIN_DECK_COUNT);
     private int selectedDeckIndex = 0;
     private final DifficultyData difficultyData = new DifficultyData();
 
+    // Commander mode
+    private boolean commanderMode;
```
Jetz72: I'm not sure I like hinging so much behavior on a single boolean flag. If hypothetically we ever wanted to add Oathbreaker Adventure or some other weird format, we'd have to generalize this. Maybe store and reference the player's `DeckFormat` or `GameFormat` instead? Or store their `AdventureMode` and derive deck and game format from that.

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
```diff
@@ -334,6 +348,14 @@ public void setColorIdentity(ColorSet set) {
         this.colorIdentity = set;
     }
 
+    private Deck newDeck(String deckname){
+        if (commanderMode) {
+            Deck n_deck = new Deck(deckname);
+            n_deck.putSection(DeckSection.Commander, new CardPool());
```
Jetz72: Was something breaking if you didn't manually initialize this? Most places use `Deck::getOrCreate(DeckSection)` and should be able to handle its absence.

forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
```diff
@@ -197,6 +197,8 @@ public void enter() {
         String isDeckMissingMsg = "";
         if (eventData != null && eventData.eventRules != null) {
             mainGameType = eventData.eventRules.gameType;
+        } else if (AdventurePlayer.current().isCommanderMode()){
+            mainGameType = GameType.Commander;
```
Jetz72, line 202: You'll also want to update `DuelScene::applyAdventureDeckRules` to handle missing commanders, illegal commanders, mismatched color identity, and any other commander deck rules.

forge-gui-mobile/src/forge/adventure/util/Config.java
jgroehl: Thanks for the swift response, @tool4ever!
Oh, I did not actually notice that there was a Commander deck generator already available.

I have refactored my gen script into the DeckgenUtil class.

Would you prefer for me to completely replace with the already existing generator?

Here are my thoughts behind the code I wrote:
The starting deck is contains a healthy amount of ramp and draw effects and low mana value creatures to have a shot at winning duels early on, even with a high mana value commander.
It also explicitly does not generate and legendaries and only commons/uncommons, such that the player has more of an upgrading experience and high-impact cards are not part of the starter deck.

What do you think?

Jetz72 (commented):
The biggest hurdle here is that Adventure mode isn't really balanced for Commander. Items and encounters and bosses are designed with 40-60 card constructed decks with limited inventory in mind. How thoroughly have you tested it with balance in mind?

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
jgroehl: Yes, I like this! Thanks for the pointer!

forge-gui-mobile/src/forge/adventure/util/Config.java
tool4ever: no problem, thank you too for contributing :)

yea I meant a bit more than just moving the code...

the Deckgen logic does a lot of calculations for the final deck, so a full alternative system seems unneeded

I do think your beginner-friendly approach with the filtered preselection makes sense though
which should all be quite possible and probably involves some more interaction with _CardThemedCommanderDeckBuilder_

I suggest walking through the top call chain a bit so you can get a feel to how it's connected and where you best hook into

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
jgroehl: Well, I thought it would, but I just tested it without the weird wrapper function and it works just fine, as you said. Thanks!

forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
jgroehl: That all works through line 211 here, since the DeckFormat is adjusted properly in the Enum:

        applyAdventureDeckRules(mainGameType.getDeckFormat());

forge-gui-mobile/src/forge/deck/FDeckEditor.java
jgroehl: Oh, I fully agree, I hadn't actually realised I was changing the core class outside of the adventure package.
I have overwritten the isCommanderEditor function in the ´AdventureDeckEditor´ now. It returns true if the gamemode is commander, and otherwise it forwards to ´super´.

forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
jgroehl: Thanks for the amazing feedback @Jetz72!
I have pushed the suggested changes! The code is a lot cleaner.

jgroehl: Regarding the balancing: From the playtesting I have done so far, I find the experience much more rewarding, actually! 100 singletons gives you a lot of variance - it does not become a given that you can defeat any boss, just because you assemble the first cards for an infinite loop etc.

It is a much more massive undertaking assembling a good multi-colour commander deck, since you need to travel around a lot chasing for cards that work!

I have also not found it to be problematic that the opponents do not use a commander.

forge-gui-mobile/src/forge/adventure/util/Config.java
jgroehl: Haha, I thought as much :D
I'll look into it! Thanks a lot again - I wasn't expecting such quick & good feedback!

forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
Jetz72: > That all works through line 211 here, since the DeckFormat is adjusted properly in the Enum:

That method is currently only set up to handle deck size restrictions and duplicate cards. In the deck editor, we can just run the `DeckFormat.getDeckConformanceProblem` to verify that the deck matches the game rules. However, the user can ignore that and exit the editor with an invalid deck. If they do, `applyAdventureDeckRules` is responsible for applying corrections before they enter into a game, trimming illegal or excess cards and padding smaller decks with filler. It'll need its own handling for commander rules violations. Maybe give them an "Atogatog" if they're missing a commander and remove any cards that are outside their color identity when cutting other excess cards.

jgroehl: I guess this boils down to: do we want to "hard-enforce" the commander rules, or is it up to the players to see the warnings, but then do as they please? 🤔
Happy to go either way tbh..

Jetz72: Upon review I think that method is meant to be gated behind a check for `FModel.getPreferences().getPrefBoolean(FPref.ENFORCE_DECK_LEGALITY)`. Anyone who wants to play with freeform decks should be able to just disable that and then do so.

But for everyone else, yes I think it'd be sensible to make sure their deck meets the format criteria. We could maybe ignore the case where they have no commander since that's just a strictly worse deck 99.999% of the time. But as for the color identity rule and illegal commanders, I'd rather not leave it to the honor system to enforce it.

kazyfreak:
In adventure mode currently its just warning flags (running decks with more than 4 cards) etc. I would be under the opinion have the warning flag (it's already there with the commander ruleset) but just let the player do what they want. No extra coding enforcement and if they want to run 200 cards go for it. Even running with out a commander the game still plays just the player won't be able to cast a commander. Same with color etc. Less is more in my opinion. The flag will just throw a warning during every deck edit.

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
```diff
@@ -896,12 +916,20 @@ public void onClose(final Consumer<Boolean> canCloseCallback) {
...
-            String warning = localizer.getMessage("lblAdventureDeckError", deckError);
-            FOptionPane.showConfirmDialog(warning, localizer.getMessage("lblInvalidDeck"), false, result -> resolveClose(canCloseCallback, result == true));
-            return;
+        String deckError;
+        if (!(getEditorConfig() instanceof ShopConfig))
+        {
+            if (AdventurePlayer.current().getAdventureMode() == AdventureModes.Commander)
+                deckError = GameType.Commander.getDeckFormat().getDeckConformanceProblem(getDeck());
+            else
+                deckError = GameType.Adventure.getDeckFormat().getDeckConformanceProblem(getDeck());
```
Jetz72: I think what you'd want to do is acquire `DeckFormat.Commander` via `AdventureEditorConfig.getDeckFormat`, then just simplify this to `getEditorConfig().getDeckFormat().getDeckConformanceProblem(...`. The editor config's deck format is used for various other bits of editor logic.

forge-gui-mobile/src/forge/adventure/util/Config.java
jgroehl: It will be a lot of work and tinkering to get this implementation of the commander deckbuilder to behave as I envisioned it. I think I would want to tackle that through a different enhancement after finalising the commander game mode contribution, using the already existing Deckgen utility. Sounds good?

tool4ever: that is fine too

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
jgroehl: Great shout! I refactored this into the ´AdventureEditorConfig´ class!

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
```diff
@@ -896,12 +916,17 @@ public void onClose(final Consumer<Boolean> canCloseCallback) {
...
         }
 
-        String deckError = GameType.Adventure.getDeckFormat().getDeckConformanceProblem(getDeck());
-        if (deckError != null) {
-            //Allow the player to close the editor with an invalid deck, but warn them that cards may be swapped out.
-            String warning = localizer.getMessage("lblAdventureDeckError", deckError);
-            FOptionPane.showConfirmDialog(warning, localizer.getMessage("lblInvalidDeck"), false, result -> resolveClose(canCloseCallback, result == true));
-            return;
+        String deckError;
+        if (!(getEditorConfig() instanceof ShopConfig))
```
jgroehl, line 920: This, btw, fixed a deck error message when exiting the Inn shop while having an unconforming deck...

forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
jgroehl: This is now handled exactly as the standard adventure modes, thus introducing minimal changes to enable commander mode.

jgroehl:
@Jetz72 @tool4ever Thanks for your feedback again 🙏.
I've implemented your suggestions, making the code edits quite minimal. The only larger chunk of code is the convenience function to automatically toggle cards for autosale on the reward screen that the player already had in their collection.
I've also cleaned up unused imports and accidentally introduced line breaks, etc. 
Would really appreciate another review!
Cheers 😄

forge-gui-mobile/src/forge/adventure/util/Config.java
jgroehl: Opened issue for this: #8984

jgroehl:
Uff, I hope I've got them all now - Question for the future @tool4ever: how can I run all these checks locally?

tool4ever:
`mvn test` for the full suite
single ones depending on how your IDE works...

forge-gui-mobile/src/forge/adventure/scene/NewGameScene.java
```diff
@@ -233,6 +288,31 @@ public void act(float delta) {
...
+
+        Predicate<PaperCard> commander_filter = card -> {
+            ICardFace face = card.getRules().getMainPart();
+            boolean isLegendaryCreature = face.getType().isCreature() && face.getType().isLegendary();
+            boolean sharesColorInIdentity = face.getColor().sharesColorWith(_color);
+
+            return isLegendaryCreature && sharesColorInIdentity;
+
+        };
+        List<PaperCard> commanderCandidates = new ArrayList<>(cardDb.streamUniqueCardsNoAlt().filter(commander_filter).toList());
```
tool4ever: use ``FModel.getCommanderPool()``

your're also not really checking color identity

forge-gui-mobile/src/forge/adventure/scene/NewGameScene.java
```diff
@@ -446,10 +533,31 @@ private void showModeHelp() {
                 }
                 break;
             case Chaos:
-                summaryText.append("Mode: Chaos\n\nYou (and all enemies) will receive a random preconstructed deck.\n\nWarning: This will make encounter difficulty vary wildly from the developers' intent");
+                summaryText.append("""
+                        Mode: Chaos
+
```
tool4ever: does hard converting these linebreaks work for all platforms?

jgroehl: AFAIK, they are part of the Java language specification since Java 15 and thus not platform-dependent.
But this is an "unnecessary" alteration regarding the objective of this issue, so I am happy to revert these and align the added text.

forge-gui-mobile/src/forge/adventure/scene/NewGameScene.java
jgroehl: Super! Will do.

In an early iteration, I took the theme colour into account; now I am just adding a list of 15 random commanders, where three each touch on one of the colours - makes sense to go back to that!
Could I get a quick opinion on which option you would personally prefer:

1. Hide colour identity from the new game view and give a long list of random commanders (-> the commanders' colour identity will be the deck colour identity).
2. Keep the 5 single colour identities and generate N candidates per identity that update when changing the colour identity. Here, each of the N suggestions has to have the current colour in its identity.
3. Expand the colour choices to all 31 colour identities when changing to commander mode and supply a single random commander for each.

I think that option 3 is the most intuitive, and option 2 limits the number of button clicks.

Jetz72: I worry that offering random commander candidates before the adventure starts just encourages rerolling by backing out and starting over. Perhaps a fixed list of like 2 or 3 starter commanders with simple themes for each mono-color?

tool4ever: hmn I think not having all commanders available initially could be motivating if there's some kind of gameplay way to earn them

but I'd sure feel annoyed as a player knowing they're all in the pool and the interface just forces me to spam buttons until it finally rolls one I want to start with

kazyfreak: How I rolled the version I'm testing (since its got a slightly different scope) is that I hard coded the 5 single color, 10 duels and triomes into the color picker (adding in symbols etc). Then linked those to a precon. They are staples, easy to adjust offers a few alternate commanders to build on, and as balanced as your going to get before players start customizing. That gives them something to start with that is viable, but tons of upgrades and they can pick colors or themes they are interested in if they want to go find a different commander elsewhere.

I also plan on coding enemies to have a % chance to drop what commander they are using. That way players can farm them if they want.

Again different game but I thought I'd offer my two cents.

I did have a slight issues I had to rebuild since the commander match wasn't an adventure mode. Some logic changes I did to my own forked version (Just things to keep an eye on)

Added Commander deck editor support with validation | AdventureDeckEditor.java (lines 67-91, 916-923) |
Added Commander detection to adventure battles | DuelScene.java (lines 201-204) |
Simplified detection logic + added debug output (deck editor not showing tab) | AdventureDeckEditor.java (lines 67-94) |
Fixed validation persistence + added commander menu items | AdventureDeckEditor.java (lines 614-620, 919-925) |
Fixed GameRules to use detected mainGameType for match logic | DuelScene.java (line 370)
Fixed match exit screen to use AdventureWinLose for all GameTypes in adventure mode | ViewWinLose.java (lines 89-92) |Fixed IndexOutOfBoundsException when loading maps with missing enemies (biome bounds check) | MapStage.java (lines 490-496) | *This is because we're not using all the common enemies bc they aren't commander*

jgroehl: Great, thanks for the quick feedback!
Based on your suggestions, I would tend to agree with @Jetz72's opinion.
To start off, let's offer a single commander precon for each monocolor, but implemented in a way that it is as simple as adding more precons to eventually enable more choices in the future.

-> This way, we enable a balanced start to the game, and encourage the player to build a collection during the playthrough for building stronger decks!
-> game design-wise, I'd want to go for weaker decks that rely on playing on-curve creatures to win the game, and restrict the 99 to commons/uncommons. -> from the beginning, upgrades will feel meaningful

@kazyfreak thanks for sharing the changes you made - that is mostly in line with the changes I made in the PR! In the proposed version, enemies remain untouched - they play their usual decks, and it is only the player who has access to the command zone.

jgroehl: @kazyfreak do you mind sharing the deck lists for the monocolor starter decks you built?

jgroehl:
I've built the starter precons around the uncommon TBD demigods, as they inherently reward playing a creature-heavy style.
Have not included other strong synergies on purpose, but the decks will receive some tweaking through playtesting over the weekend.

You can find the current lists here:
https://archidekt.com/folders/1243459

forge-gui-mobile/src/forge/adventure/scene/NewGameScene.java
jgroehl: I've reverted to the way it was before, just in case.

jgroehl:
I believe this is now good to go. I've playtested a bit more, and it all seems to make sense.
Thanks everyone for the feedback so far - happy to make more changes if you find more things out of place!

The changes are now quite minimal to enable the adventure player to run the playthrough with a commander deck.
The only really optional change is the autosell toggle for duplicates.
The starter decks can all win games against the early enemies, but have no chance against the bosses -> I think this is a good balance to get this game mode started!

In the future, it would be great to add specific features to this game mode, i.e. difficulty-specific starter decks, special enemies that also have commanders, dedicated shops for commander-related cards, 4-player EDH tournaments in the capital, EDH-specific gear rewards, changes to rewards in EDH settings (no card drops more than once), etc, etc :) Happy to support these developments if you think these would make sense!

tool4ever:
can't playtest but code LGTM 👍

jgroehl:
Since I have changed the data save by saving and loading an additional flag, it would be fantastic if someone else apart from me could test that the changes do not break the save file when loading. They did not for me for the save file I created before the changes, but it would be great to double-check!!

Agetian (approved):

Agetian:
Looks good to me! :)

Zara7hus7ra:
sol ring in starter deck seems a little bit broken?

Issue #8925, opened 2025-10-14
AI Can't Cast Gelatinous Genesis


bmlaw:
**Describe the bug**
AI Can't Cast Gelatinous Genesis

**To Reproduce**
1. Give AI a deck full of green mana and Gelatinous Genesis.
2. Wait.

**Expected behavior**
At some point, the AI should cast the only spells it has.

**Screenshots**
<img width="3030" height="1998" alt="Image" src="https://github.com/user-attachments/assets/28a1162e-d526-4096-98f8-fbce1458e3e8" />

Valerie is AI. 12 Mox Emeralds should provide more than enough mana to cast the spell.

Eduardo is me passing the turn.

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 2.0.0.7 snapshot 10.10

**Additional context**
Gelatinous Genesis shows up in the ooze_monogreen deck in Adventure as well as several decks in quest and conquest.

dracontes:
Tested a few relevant cards: X spells that create either X tokens or an X/X token. The AI readily plays the first: [Doppelgang](https://scryfall.com/card/mkm/198/doppelgang), [Awaken the Woods](https://scryfall.com/card/bro/170/awaken-the-woods), etc. However, the AI won't play cards of the second group, when it by all means could: for instance, [Wurmcalling](https://scryfall.com/card/tsp/234/wurmcalling) and [Slime Molding](https://scryfall.com/card/mm3/137/slime-molding).

JoeMoCode:
Hey there! I been playing around in this codebase and think I fixed this issue. Not sure the extent of testing you do, but I opened the PR if you'd like to see it.

PR #8972, opened 2025-10-21
Fixing issue 8925 - AI has trouble casting spells that create X-tough…

JoeMoCode:
…ness creatures.

Issue: https://github.com/Card-Forge/forge/issues/8925

Tested with the gelatinous genesis card and confirmed AI now can cast it.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/TokenAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/TokenAi.java
@@ -69,17 +69,20 @@ public class TokenAi extends SpellAbilityAi {
 
         Card actualToken = spawnToken(ai, sa);
 
-        if (actualToken == null || (actualToken.isCreature() && actualToken.getNetToughness() < 1)) {
-            // planeswalker plus ability or sub-ability is useful
-            return pwPlus || sa.getSubAbility() != null;
-        }
-
         String tokenAmount = sa.getParamOrDefault("TokenAmount", "1");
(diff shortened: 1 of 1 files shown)


forge-ai/src/main/java/forge/ai/ability/TokenAi.java
```diff
@@ -69,15 +69,19 @@ protected boolean checkPhaseRestrictions(final Player ai, final SpellAbility sa,
...
-
         String tokenAmount = sa.getParamOrDefault("TokenAmount", "1");
         String tokenPower = sa.getParamOrDefault("TokenPower", actualToken.getBasePowerString());
         String tokenToughness = sa.getParamOrDefault("TokenToughness", actualToken.getBaseToughnessString());
 
+        // Don't check toughness yet if token has variable P/T based on X
+        // (the check will happen after X is calculated)
+        boolean hasVariablePT = "X".equals(tokenAmount) || "X".equals(tokenPower) || "X".equals(tokenToughness);
+
+        if (!hasVariablePT && (actualToken == null || (actualToken.isCreature() && actualToken.getNetToughness() < 1))) {
```
tool4ever: thanks for contributing

I think it's probably ok for AI to behave somewhat less greedy like this

you can also reuse your new boolean in L86
and maybe change its name to make it slightly more precise

tool4ever (approved):

Hanmac:
Might need some more fixes later, (see X=xPaid) below, but that can be done later

PR #8976, opened 2025-10-22
Revert to numbered capturing groups in CardEdition regexes

### LINKED ISSUE #8975: New Android version 2.0.07-10.27 hangs on checking for update
**Describe the bug**
Game doesn't start, stuck on "Checking for updates" screen.  swiping back displays animated logo, and game shuts down.

**To Reproduce**
1. Update
2. run, 
3. download assets
4. restart
5. gets stuck there.

Tried closing and clearing cache, same problem.
Tried re downloading and installing(update)

**Expected behavior**
Should automatically bring the main menu

Image

 - Device: Moto G stylus 5G 2023
 - OS: Android
 - Version 14

Jetz72:
Fixes incompatibility with older Android SDKs. Closes #8975.

We really need an Android Build workflow, though I dunno if it could have caught this.

### DIFF
--- a/forge-core/src/main/java/forge/card/CardEdition.java
+++ b/forge-core/src/main/java/forge/card/CardEdition.java
@@ -613,25 +613,25 @@ public final class CardEdition implements Comparable<CardEdition> {
                 So, untill support for those devices is officially dropped,
                 we'll have to suffice with numbered groups.
                 We are looking for:
-                    * cnum - grouping #1
-                    * rarity - grouping #2
-                    * name - grouping #3
-                    * artist name - grouping #4
-                    * extra parameters - grouping #5
+                    * cnum - grouping #2
+                    * rarity - grouping #4
+                    * name - grouping #5
+                    * artist name - grouping #7
+                    * extra parameters - grouping #9
                 */
                 // Collector numbers now should allow hyphens for Planeswalker Championship Promos
-                "(?:^(.?[0-9A-Z-]+\\S*[A-Z]*)\\s)?(?:([SCURML])\\s)?([^@$]+)(?: @([^$]*))?(?: \\$\\{(.+)})?$"
+                "(^(.?[0-9A-Z-]+\\S*[A-Z]*)\\s)?(([SCURML])\\s)?([^@$]+)( @([^$]*))?( \\$\\{(.+)})?$"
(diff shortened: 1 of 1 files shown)


Hanmac:
While we are at, can we change the Name Group from `([^@]*)` to `([^@]+)`?
It doesn't make much sense that it would match an empty name?

Jetz72:
Sure. Also realized I missed one of the named groups.

forge-core/src/main/java/forge/card/CardEdition.java
```diff
@@ -608,17 +608,30 @@ public static class Reader extends StorageReaderFolder<CardEdition> {
...
+                "(?:^(.?[0-9A-Z-]+\\S*[A-Z]*)\\s)?(?:([SCURML])\\s)?([^@$]+)(?: @([^$]*))?(?: \\$\\{(.+)})?$"
+                //"(?:^(?<cnum>.?[0-9A-Z-]+\\S*[A-Z]*)\\s)?(?:(?<rarity>[SCURML])\\s)?(?<name>[^@$]*)(?: @(?<artist>[^$]*))?(?: \\$\\{(?<params>.+)})?$"
         );
 
         public static final Pattern TOKEN_PATTERN = Pattern.compile(
                 /*
-                 * cnum - grouping #2
-                 * name - grouping #3
-                 * artist name - grouping #5
+                 * cnum - grouping #1
```
tool4ever: why did the group numbers change compared to the previous version?
if the old ones had 5/9 I see no reason why it's less now

Hanmac: `(?:(?<rarity>[SCURML])\\s)?` uses non-matching groups for the outer one, so they don't get indexed

`(?:([SCURML])\\s)?` the outer is still not indexed

Jetz72: I'd been using some non-capturing groups for groupings that aren't used in the script, unlike the original which used capturing groups for everything.

Buuuuut.... Since you mentioned it, I'm not actually sure Android will support those either. I'll revert them just to be safe. Good call out.

Hanmac:
@tool4ever @Jetz72 the non-capturing shouldn't be needed to reverted
but @tool4ever was too trigger-happy

Jetz72:
If you're sure about that, it'd be a nice little cleanup. I just wanted to be on the safe side.

PR #8995, opened 2025-10-25
Fix Multiple Replacement Effects not working correctly

Hanmac:
fixes if one Updated applies first (Double Lifegain), and then gets replaced by a normal one (Lose Life Instead)


Agetian:
Looks good to me, can't think of a scenario where this may have a negative effect / break something

PR #8997, opened 2025-10-25
Move wiki as-is into project

phughk:
Migrate the existing forge wiki project files into the main forge project.

There is no template rendering or publishing to github wiki. There could be a post-merge hook to push to the wiki project after merging to the main repo.


tool4ever:
This PR is redundant: #8320

Issue #8595, opened 2025-09-01
Corruption of Towashi + ResolvedLimit


Hanmac:
@tool4ever can you check how `ResolvedLimit` works with multiple Triggers having the Same SpellAbility?

Like `Corruption of Towashi`:
https://github.com/Card-Forge/forge/blob/ed3ea99e1aeb170f6e02aaa9c88c25512bcdba7a/forge-gui/res/cardsfolder/c/corruption_of_towashi.txt#L6-L7

Can we clean this up via ResolvedLimit?

----

Probably not, because Trigger always sets itself in `setOverridingAbility`
https://github.com/Card-Forge/forge/blob/ed3ea99e1aeb170f6e02aaa9c88c25512bcdba7a/forge-game/src/main/java/forge/game/trigger/Trigger.java#L613-L616

and there are some places where the SpellAbility needs to know its trigger, :/

tool4ever:
Also relevant for _Victor, Valgavoth's Seneschal_

PR #9003, opened 2025-10-26
Card/edition fixes

Eradev:
Fixes https://github.com/Card-Forge/forge/issues/8595

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -1861,6 +1861,12 @@ public class AbilityUtils {
                 }
 
                 if (sq[0].equals("ResolvedThisTurn")) {
+                    if (sa.isTrigger()) {
+                        Trigger trigger = sa.getTrigger();
+
+                        return doXMath(trigger.getResolvedThisTurn(trigger.getHostCard().getOwner()), expr, c, ctb);
+                    }
+
                     return doXMath(sa.getResolvedThisTurn(), expr, c, ctb);
                 }
 
--- a/forge-gui/res/cardsfolder/v/victor_valgavoths_seneschal.txt
+++ b/forge-gui/res/cardsfolder/v/victor_valgavoths_seneschal.txt
@@ -2,14 +2,11 @@ Name:Victor, Valgavoth's Seneschal
 ManaCost:1 W B
 Types:Legendary Creature Human Warlock
 PT:3/3
-T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Enchantment.YouCtrl | TriggerZones$ Battlefield | Execute$ TrigSurveil | TriggerDescription$ Eerie — Whenever an enchantment you control enters and whenever you fully unlock a Room, surveil 2 if this is the first time this ability has resolved this turn. If it's the second time, each opponent discards a card. If it's the third time, put a creature card from a graveyard onto the battlefield under your control.
-T:Mode$ FullyUnlock | ValidCard$ Card.Room | ValidPlayer$ You | Secondary$ True | Execute$ TrigSurveil | TriggerZones$ Battlefield | TriggerDescription$ Eerie — Whenever an enchantment you control enters and whenever you fully unlock a Room, surveil 2 if this is the first time this ability has resolved this turn. If it's the second time, each opponent discards a card. If it's the third time, put a creature card from a graveyard onto the battlefield under your control.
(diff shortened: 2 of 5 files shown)


Hanmac (dismissed):

Hanmac:
Because the trigger is two parts, the Activation Limit might not work 

See #8595

Eradev:
> Because the trigger is two parts, the Activation Limit might not work
> 
> See #8595

Isn't that why we check for X and Y for this case? ActivationLimit just prevents it from activating multiple times if it didn't have time to set X/Y to 1.

As for the "card enters" part, it'll prevent it from triggering multiple times if there something that duplicates ETB abilities.

Hanmac:
Oasis only once each turn on the trigger should work for the entire part

So "etb or something else" should still only trigger once,  if you do something else in the same turn it enters

I don't know right now if ActivationLimit can handle that 🤔

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -1861,6 +1861,12 @@ public static int xCount(Card c, final String s, final CardTraitBase ctb) {
                 }
 
                 if (sq[0].equals("ResolvedThisTurn")) {
+                    if (sa.isTrigger()) {
+                        Trigger trigger = sa.getTrigger();
+
+                        return doXMath(trigger.getResolvedThisTurn(trigger.getHostCard().getOwner()), expr, c, ctb);
```
tool4ever, line 1867: why only owner?
anyway this case is different, the activator doesn't matter

forge-gui/res/cardsfolder/v/victor_valgavoths_seneschal.txt
```diff
@@ -2,14 +2,11 @@ Name:Victor, Valgavoth's Seneschal
...
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Enchantment.YouCtrl | TriggerZones$ Battlefield | LinkedTrigger$ Eerie | ResolvedLimit$ 3 | Execute$ TrigSurveil | TriggerDescription$ Eerie — Whenever an enchantment you control enters and whenever you fully unlock a Room, surveil 2 if this is the first time this ability has resolved this turn. If it's the second time, each opponent discards a card. If it's the third time, put a creature card from a graveyard onto the battlefield under your control.
+T:Mode$ FullyUnlock | ValidCard$ Card.Room | ValidPlayer$ You | Secondary$ True | LinkedTrigger$ Eerie | ResolvedLimit$ 3 |Execute$ TrigSurveil | TriggerZones$ Battlefield | TriggerDescription$ Eerie — Whenever an enchantment you control enters and whenever you fully unlock a Room, surveil 2 if this is the first time this ability has resolved this turn. If it's the second time, each opponent discards a card. If it's the third time, put a creature card from a graveyard onto the battlefield under your control.
```
tool4ever, line 6: there's no ResolvedLimit on this trigger

forge-gui/res/cardsfolder/v/victor_valgavoths_seneschal.txt
```diff
@@ -2,14 +2,11 @@ Name:Victor, Valgavoth's Seneschal
...
-SVar:DBLog:DB$ StoreSVar | SVar$ X | Type$ CountSVar | Expression$ X/Plus.1
-SVar:X:Number$1
-T:Mode$ Phase | Phase$ Cleanup | TriggerZones$ Battlefield | Execute$ DBCleanup | Static$ True
-SVar:DBCleanup:DB$ StoreSVar | SVar$ X | Type$ Number | Expression$ 1
+SVar:DBChangeZone:DB$ ChangeZone | Origin$ Graveyard | Destination$ Battlefield | ChangeType$ Creature | ChangeNum$ 1 | Mandatory$ True | GainControl$ True | ConditionCheckSVar$ X | ConditionSVarCompare$ EQ3 | SelectPrompt$ Select a creature card in a graveyard | Hidden$ True
+SVar:X:Count$ResolvedThisTurn
 DeckNeeds:Type$Enchantment
 Oracle:Eerie — Whenever an enchantment you control enters and whenever you fully unlock a Room, surveil 2 if this is the first time this ability has resolved this turn. If it's the second time, each opponent discards a card. If it's the third time, put a creature card from a graveyard onto the battlefield under your control.
```
tool4ever, line 12: thanks for the idea
I think it's not good enough unfortunately:
if this card loses all abilities while the trigger is on the stack your logic would fail

@Hanmac
right now I wonder if multimode isn't better if we can figure out how to deal with param collision? 🤔

Eradev: That's fine. I'll close this to keep a trace, and opened another PR #9069 with the other fixes.

PR #9009, opened 2025-10-27
Restore check to treat empty room as dynamic state

tool4ever:


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -557,18 +557,16 @@ public boolean setState(final CardStateName state, boolean updateView, boolean f
         boolean needsTransformAnimation = transform || rollback;
         // faceDown has higher priority over clone states
         // while text change states doesn't apply while the card is faceDown
-        if (state != CardStateName.FaceDown) {
+        if (state != CardStateName.FaceDown && state != CardStateName.EmptyRoom) {
```
Hanmac, line 560: Wouldn't that break the clone states again?

Empty Room should be got from the Clone State

tool4ever: this doesn't provide the right state
that happens later in L584 with `getState`

Hanmac: Then what is this CloneState at thus position for, if the state gets later from getState anyway? 🤔

tool4ever: as far as I can tell it's just a sanity check that's basically the same logic as `hasState`

not sure when it would be needed, I don't think I ever saw that `RuntimeException` on Sentry?

maybe you can clean that up some more later, but it shouldn't matter for now 🤷‍♂️

PR #9025, opened 2025-10-28
TLA cards, 28th October, Batch 3

Fulgur14:
This file defines the 'Invasion Tactics' enchantment card with its properties and effects.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/azula_cunning_usurper.txt
+++ b/forge-gui/res/cardsfolder/upcoming/azula_cunning_usurper.txt
@@ -3,9 +3,8 @@ ManaCost:2 U B B
 Types:Legendary Creature Human Noble Rogue
 PT:4/4
 K:Firebending:2
-T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigChooseCard | TriggerDescription$ When NICKNAME enters, target opponent exiles a nontoken creature they control, then they exile a nonland card from their graveyard.
-SVar:TrigChooseCard:DB$ ChooseCard | ValidTgts$ Opponent | Choices$ Creature.!token+TargetedPlayerCtrl | ChoiceTitle$ Choose a nontoken creature you control | Mandatory$ True | SubAbility$ DBExile1
-SVar:DBExile1:DB$ ChangeZone | Defined$ ChosenCard | Origin$ Battlefield | Destination$ Exile | SubAbility$ DBExile2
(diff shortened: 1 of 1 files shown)


forge-gui/res/cardsfolder/upcoming/azula_cunning_usurper.txt
```diff
@@ -0,0 +1,10 @@
+Name:Azula, Cunning Usurper
+ManaCost:2 U B B
+Types:Legendary Creature Human Noble Rogue
+PT:4/4
+K:Firebending:2
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigExile | TriggerDescription$ When NICKNAME enters, target opponent exiles a nontoken creature they control, then they exile a nonland card from their graveyard.
+SVar:TrigExile:DB$ ChangeZone | Origin$ Battlefield | Destination$ Exile | DefinedPlayer$ Targeted | ValidTgts$ Opponent | Chooser$ Targeted | ChangeType$ Creature.!token | ChangeNum$ 1 | IsCurse$ True | Mandatory$ True | SubAbility$ DBExile
```
tool4ever: was this tested?

Fulgur14: Hm, now exile works from battlefield but not from graveyard for some reason?

tool4ever: extra ChooseCard shouldn't be needed

Fulgur14: Yes, I looked at Grave Birthing, so not sure what happened...

Fulgur14: OK, does this Azula still have problems?

tool4ever: yes, get rid of ChooseCard

Fulgur14: Like this? I think this is close to how it was in the beginning...

tool4ever: it works now?

Fulgur14: I'm as surprised as you are, but yes -- cards are exiled correctly and they can be cast just fine :)

PR #9030, opened 2025-10-28
Energybending (TLA) [Accidentally put it into master, so I'll leave it as a solo file.]

Fulgur14:


forge-gui/res/cardsfolder/upcoming/energybending.txt
```diff
@@ -0,0 +1,6 @@
+Name:Energybending
+ManaCost:2
+Types:Instant
+A:SP$ AnimateAll | ValidCards$ Land.YouCtrl | Types$ AllBasicLandType | SubAbility$ DBDraw | SpellDescription$ Lands you control gain all basic land types until end of turn. Draw a card.
```
tool4ever: I don't think you tested this :p

Fulgur14: OK, so... how is this different from Prismatic Omen? That uses "AllBasicLandType". Does it need to spell them out?

tool4ever: it's a hardcoded placeholder so it doesn't work everywhere just like that

tool4ever: that doesn't work because of text changes...

Fulgur14: You mean with text change effects? Because basic functionality does work...

tool4ever: yes

tool4ever: Hanmac says Effect + static for now

Fulgur14: Hm, can you give me an example of a card done in this way, please?

tool4ever: Hellish Rebuke

PR #9032, opened 2025-10-29
Heist fix

### LINKED ISSUE #8769: Card Exiled with Heist are exiled face up and are visible to other players
**Describe the bug**
Keyword Heist from MTGA exiles card from target opponent's library, however it exiles them faceup, while it should facedown.

1. Cast any card which heists opponent's library
2. Heist library
3. Let player check the exile and the cards should be there, visible to them

Reminder text from mtg wiki:

 Heist (Look at three random nonland cards from target opponent's library. Exile one of them face down. You may cast that card for as long as it remains exiled, and you may spend mana as though it were mana of any type to cast that spell.)

OS: Windows
Browser: chrome
Version 2.0.0.6-SNAPSHOT-09.22

Eradev:
Fixes #8769

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/HeistEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/HeistEffect.java
@@ -1,6 +1,5 @@
 package forge.game.ability.effects;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/ability/effects/HeistEffect.java
```diff
@@ -38,17 +37,19 @@ public void resolve(SpellAbility sa) {
             List<Card> choices = Aggregates.random(CardLists.getNotType(target.getCardsIn(ZoneType.Library), 
                 "Land"), 3);
             if (choices.isEmpty()) continue; //nothing to heist
-            Card chosenCard = player.getController().chooseSingleCardForZoneChange(ZoneType.Exile, 
-                new ArrayList<ZoneType>(Arrays.asList(ZoneType.Exile)), sa, new CardCollection(choices), 
+            Card chosenCard = player.getController().chooseSingleCardForZoneChange(ZoneType.Exile,
+                new ArrayList<>(List.of(ZoneType.Exile)), sa, new CardCollection(choices),
```
tool4ever: if you don't need mutable no reason to construct another list
otherwise the extra call would still be redundant
also shouldn't that be Library?

tool4ever (approved):

PR #9042, opened 2025-10-30
Remove conjured cards from reward pool

Eradev:
It didn't apply the filter when there were other restrictions.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/data/RewardData.java
+++ b/forge-gui-mobile/src/forge/adventure/data/RewardData.java
@@ -106,11 +106,10 @@ public class RewardData implements Serializable {
         if (configData.allowedEditions != null && configData.allowedEditions.length > 0) {
             allCards = IterableUtil.filter(allCards, PaperCardPredicates.printedInAnyEditions(configData.allowedEditions));
         } else if (configData.restrictedEditions != null && configData.restrictedEditions.length > 0) {
-            allCards = IterableUtil.filter(allCards, PaperCardPredicates.onlyPrintedInEditions(configData.restrictedEditions).negate());
+            allCards = IterableUtil.filter(allCards, PaperCardPredicates.isObtainableNotRestricted(configData.restrictedEditions));
+        } else {
+            allCards = IterableUtil.filter(allCards, PaperCardPredicates.isObtainableAnyEdition());
         }
-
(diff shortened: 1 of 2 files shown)


forge-gui-mobile/src/forge/adventure/data/RewardData.java
```diff
@@ -107,10 +107,11 @@ static private void initializeAllCards(){
             allCards = IterableUtil.filter(allCards, PaperCardPredicates.printedInAnyEditions(configData.allowedEditions));
         } else if (configData.restrictedEditions != null && configData.restrictedEditions.length > 0) {
             allCards = IterableUtil.filter(allCards, PaperCardPredicates.onlyPrintedInEditions(configData.restrictedEditions).negate());
-        } else {
-            allCards = IterableUtil.filter(allCards, PaperCardPredicates.isObtainableAnyEdition());
         }
 
+        // Remove unobtainable cards (ie. Conjure-only)
+        allCards = IterableUtil.filter(allCards, PaperCardPredicates.isObtainableAnyEdition());
```
tool4ever: this seems unneeded
the other two Predicates also check for `isCardObtainable` anyway...

Eradev: > Total cards before obtainability filter: 31689
Total cards after obtainability filter: 31680
Total removed cards: 9
Removed: Hag of Twisted Visions
Removed: Hag of Ceaseless Torment
Removed: Hag of Scoured Thoughts
Removed: Hag of Inner Weakness
Removed: Hag of Dark Duress
Removed: Hag of Syphoned Breath
Removed: Hag of Mage's Doom
Removed: Hag of Death's Legion
Removed: Hag of Noxious Nightmares

Jetz72: Yeah, looks like the `restrictedEditions` branch has that predicate negated, so it won't remove cards that exclusively appear as unobtainable entries in restricted sets. But it might be a bit more straightforward if we ran the card pool through a single filter specifically for restricting editions instead of applying both `onlyPrintedInEditions.negate` and `isObtainableAnyEdition`, iterating over the printings of each card an extra time.

Jetz72 (approved):

PR #9050, opened 2025-10-31
Houston mulligan feature

### LINKED ISSUE #8868: Houston Mulligan?
If anyone's interested in adding it, do you think we could get the Houston Mulligan as an option?  My playgroup's been using it to speed up the start of games and I'd like to see how consistently this performs with my decks.

FYI Houston Mulligan:
Draw an opening hand of 10 and choose 3 to place at the bottom of your deck to sculpt your starting 7.

earno12:
Fixes #8868

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiCostDecision.java
+++ b/forge-ai/src/main/java/forge/ai/AiCostDecision.java
@@ -853,6 +853,12 @@ public class AiCostDecision extends CostDecisionMakerBase {
     }
 
     @Override
+    public PaymentDecision visit(CostBlight cost) {
+        // This tells the AI: "Treat this like placing counters"
+        return this.visit((CostPutCounter) cost);
+    }
+
+    @Override
     public boolean paysRightAfterDecision() {
         return false;
     }
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
@@ -72,7 +72,7 @@ public class ComputerUtilCard {
      * @param list
      */
     public static void sortByEvaluateCreature(final CardCollection list) {
-        list.sort(ComputerUtilCard.EvaluateCreatureComparator);
+        list.sort(getCachedCreatureComparator().reversed());
     }
 
     /**
@@ -544,7 +544,10 @@ public class ComputerUtilCard {
         return null;
     }
 
-    public static final Comparator<Card> EvaluateCreatureComparator = (a, b) -> evaluateCreature(b) - evaluateCreature(a);
+    public static Comparator<Card> getCachedCreatureComparator() {
+        Map<Card, Integer> cache = new IdentityHashMap<>();
+        return Comparator.comparing(c -> cache.computeIfAbsent(c, creatureEvaluator));
+    }
     public static final Comparator<SpellAbility> EvaluateCreatureSpellComparator = (a, b) -> {
         // TODO ideally we could reuse the value from the previous pass with false
         return ComputerUtilAbility.saEvaluator.compareEvaluator(a, b, true);
--- a/forge-ai/src/main/java/forge/ai/SpecialCardAi.java
+++ b/forge-ai/src/main/java/forge/ai/SpecialCardAi.java
@@ -1305,7 +1305,7 @@ public class SpecialCardAi {
     // Phyrexian Dreadnought
     public static class PhyrexianDreadnought {
         public static CardCollection reviseCreatureSacList(final Player ai, final SpellAbility sa, final CardCollection choices) {
-            choices.sort(Collections.reverseOrder(ComputerUtilCard.EvaluateCreatureComparator));
+            choices.sort(ComputerUtilCard.getCachedCreatureComparator());
             int power = 0;
             List<Card> toKeep = Lists.newArrayList();
             for (Card c : choices) {
--- a/forge-core/src/main/java/forge/card/CardChangedType.java
+++ b/forge-core/src/main/java/forge/card/CardChangedType.java
@@ -19,7 +19,12 @@ package forge.card;
 
 import java.util.Set;
 
-public record CardChangedType(CardType addType, CardType removeType, boolean addAllCreatureTypes, Set<RemoveType> remove) {
+import com.google.common.collect.Lists;
+
+import forge.card.CardType.CoreType;
+import forge.util.IterableUtil;
+
+public record CardChangedType(CardTypeView addType, CardTypeView removeType, boolean addAllCreatureTypes, Set<RemoveType> remove) implements ICardChangedType {
 
     public final boolean isRemoveSuperTypes() {
         return remove.contains(RemoveType.SuperTypes);
@@ -33,6 +38,7 @@ public record CardChangedType(CardType addType, CardType removeType, boolean add
         return remove.contains(RemoveType.SubTypes);
     }
 
+    @Override
     public final boolean isRemoveLandTypes() {
         return remove.contains(RemoveType.LandTypes);
     }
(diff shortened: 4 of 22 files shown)


forge-game/src/main/java/forge/game/Match.java
```diff
@@ -69,7 +71,32 @@ public void addGamePlayed(Game finished) {
...
+            Game game = new Game(modifiedPlayers, currentRules, this);
+
+            for (RegisteredPlayer rp : players) {
+                rp.setStartingHand(7);
+            }
+
+            return game;
+
+        } else {
+            for (RegisteredPlayer rp : players) {
```
tehdiplomat: Why is this loop needed? Also, I don't like that we're using raw values in this function. It feels like it will cause problems with Vanguard cards that modify starting hand size.

forge-game/src/main/java/forge/game/Match.java
```diff
@@ -69,7 +71,32 @@ public void addGamePlayed(Game finished) {
     }
 
     public Game createGame() {
-        return new Game(players, rules, this);
+        final GameRules currentRules = this.getRules();
+
+        MulliganDefs.MulliganRule selectedRule = StaticData.instance().getMulliganRule();
+
+        if (selectedRule == MulliganDefs.MulliganRule.Houston) {
+            List<RegisteredPlayer> modifiedPlayers = new ArrayList<>();
```
tehdiplomat: This block feels awkward. I wonder if it would be better if we split starting hand size from maxHandSize.  And then we could just have a single call to the mulligan for modifying starting hand size.

forge-game/src/main/java/forge/game/mulligan/HoustonMulligan.java
```diff
@@ -0,0 +1,58 @@
...
+                    (SpellAbility)null
+            );
+        }
+
+        kept = true;
+    }
+
+    @Override
+    public boolean canMulligan() {
+        return !kept && timesMulliganed == 0;
```
tehdiplomat: Shouldn't this just be `return false`

forge-game/src/main/java/forge/game/mulligan/HoustonMulligan.java
```diff
@@ -0,0 +1,58 @@
...
+            return;
+        }
+
+        kept = true;
+        timesMulliganed = 1;
+    }
+
+    @Override
+    public void mulligan() {
+        timesMulliganed++;
```
tehdiplomat: Based on our conversation, this function should never be reached. And probably should throw an error if we get here?

The tucking contents of this function should be in `afterMulligan()`

forge-game/src/main/java/forge/game/mulligan/HoustonMulligan.java
```diff
@@ -0,0 +1,58 @@
...
+    private static final int STARTING_DRAW_SIZE = 10;
+    private static final int TUCK_COUNT = 3;
+
+    public HoustonMulligan(Player p, boolean firstMullFree) {
+        super(p, false);
+    }
+
+    @Override
+    public int handSizeAfterNextMulligan() {
+        return STARTING_DRAW_SIZE;
```
tehdiplomat: This isn't really relevant anymore.

forge-game/src/main/java/forge/game/mulligan/MulliganService.java
```diff
@@ -80,13 +94,9 @@ private void runPlayerMulligans() {
                 }
 
                 allKept = false;
-
                 mulligan.mulligan();
             }
-        } while (!allKept);
 
-        for (AbstractMulligan mulligan : mulligans) {
-            mulligan.afterMulligan();
```
tehdiplomat, line 89: Why did this go away?

forge-game/src/main/java/forge/game/Match.java
tehdiplomat: Something like:
`
for (RegisteredPlayer rp : players) {
    rp.setStartingHand(selectedRule.getModifiedHandSize(rp.getStartingHand());
}
`

forge-game/src/main/java/forge/game/mulligan/HoustonMulligan.java
earno12: Changed to return false

forge-game/src/main/java/forge/game/mulligan/MulliganService.java
earno12: Readded those lines.

forge-game/src/main/java/forge/game/mulligan/HoustonMulligan.java
earno12: resolved this in latest push

forge-game/src/main/java/forge/game/mulligan/HoustonMulligan.java
earno12: moved tucking logic to afterMulligan() and removed the override mulligan()

forge-game/src/main/java/forge/game/Match.java
earno12: refactored code in Match class and implemented definition for getModifiedHandSize in MulliganDefs class.

forge-game/src/main/java/forge/game/Match.java
earno12: removed the troubling code from Match class

forge-core/src/main/java/forge/MulliganDefs.java
```diff
@@ -9,10 +9,40 @@
...
 public class MulliganDefs {
     public enum MulliganRule {
-        Original,
-        Paris,
-        Vancouver,
-        London
+        Original {
+            @Override
+            public int getModifiedHandSize(int startingHandSize) {
+                return startingHandSize;
```
tehdiplomat: SHouldn't this function just be in AbstractMulligan once and overridden once in your new Mulligan class?

earno12: Yeah you are probably right. I'll go and make the changes.

earno12: I have made the requested changes in my latest push.

forge-game/src/main/java/forge/game/mulligan/HoustonMulligan.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public boolean canMulligan() {
+        return false;
+    }
+
+    @Override
+    public int getModifiedHandSize(int startingHandSize) {
+        return 10;
```
tehdiplomat: `return startingHandSize + TUCK_COUNT`

forge-game/src/main/java/forge/game/mulligan/HoustonMulligan.java
```diff
@@ -0,0 +1,58 @@
...
+
+        for (final Card c : cardsToTuckDown) {
+            player.getGame().getAction().moveToLibrary(
+                    c,
+                    -1,
+                    (SpellAbility)null
+            );
+        }
+
+        if (player != null) {
```
tehdiplomat: I mean.. i don't generally have problems with guard clauses, this one isn't going to do a whole lot, since we're already referencing player multiple times before this point.

forge-game/src/main/java/forge/game/Match.java
tehdiplomat: Did you? It still seems like its here.

forge-game/src/main/java/forge/game/mulligan/HoustonMulligan.java
earno12: updated

forge-game/src/main/java/forge/game/Match.java
earno12: no this is outdated code

forge-game/src/main/java/forge/game/mulligan/HoustonMulligan.java
earno12: removed lines 25 to 27

earno12: oh I see what you commented on now. I was looking at an outdated view of the comments. I have removed lines 38 to 40.

forge-game/src/main/java/forge/game/mulligan/HoustonMulligan.java
```diff
@@ -0,0 +1,51 @@
...
+        );
+
+        for (final Card c : cardsToTuckDown) {
+            player.getGame().getAction().moveToLibrary(
+                    c,
+                    -1,
+                    (SpellAbility)null
+            );
+        }
+
```
earno12, line 40: @tehdiplomat I have to readd the logic: `if (player != null) {player.setMaxHandSize(7);}` here. I tried to remove, but the game allows the player to have more than 7 cards in hand without other natural modifiers like Reliquary Tower or Thought Vessel.

earno12: Okay I have it fixed now in latest push.

forge-game/src/main/java/forge/game/Match.java
tehdiplomat: You keep resolving this, but this IS showing up in your PR.

earno12: How can I resolve this? Do I need to take out ' if (selectedRule == MulliganDefs.MulliganRule.Houston)' on line 80 or is the for loop 'for (RegisteredPlayer rp : players) {' on line 82 problematic? I will fix it, but I just need your guidance on what is wrong still.

forge-game/src/main/java/forge/game/Match.java
earno12, line 1: @tehdiplomat I have removed the logic from Match.java. It has been relocated to MulliganService.java. Any feedback is much appreciated.

tehdiplomat:
Hey @earno12 gonna spin your PR off into a new branch in case there's any other touchups for it. Thanks for digging through most of it: https://github.com/Card-Forge/forge/pull/9382

PR #9058, opened 2025-11-01
Fix reflected only checking activated mana abilities

tool4ever:
https://apps.magicjudges.org/forum/topic/25604/

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardUtil.java
+++ b/forge-game/src/main/java/forge/game/card/CardUtil.java
@@ -307,9 +307,7 @@ public final class CardUtil {
             for (final Card c : cards) {
                 abilities.addAll(c.getSpellAbilities());
                 for (Trigger trig : c.getTriggers()) {
-                    if (trig.ensureAbility() != null) {
-                        abilities.add(trig.getOverridingAbility());
-                    }
+                    abilities.add(trig.ensureAbility());
                 }
             }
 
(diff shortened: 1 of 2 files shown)


forge-game/src/main/java/forge/game/card/CardUtil.java
```diff
@@ -304,20 +305,28 @@ private static Set<String> getReflectableManaColors(final SpellAbility abMana, f
         } else if (reflectProperty.equals("Produce")) {
             final FCollection<SpellAbility> abilities = new FCollection<>();
             for (final Card c : cards) {
-                abilities.addAll(c.getManaAbilities());
+                abilities.addAll(c.getSpellAbilities());
+                for (Trigger trig : c.getTriggers()) {
+                    if (trig.getOverridingAbility() != null) {
```
Hanmac: Use ensureAbility there?

Hanmac: @tool4ever i meant more like `abilities.add(trig.ensureAbility)` because there shouldn't be a trigger without it?

tool4ever: right, seems that way

Hanmac (approved):
I need to refactor Reflected anyway later

tool4ever:
yea I noticed there's a lot of different functions for predicting mana, maybe I can clean them up a bit later

PR #9066, opened 2025-11-02
11 TLA leaked cards

Fulgur14:


forge-gui/res/cardsfolder/upcoming/destined_confrontation.txt
```diff
@@ -0,0 +1,10 @@
+Name:Destined Confrontation
+ManaCost:2 W W
+Types:Sorcery
+A:SP$ RepeatEach | RepeatPlayers$ Player | RepeatSubAbility$ ChooseCrtr | SubAbility$ SacAllOthers | SpellDescription$ Each player chooses any number of creatures they control with total power 4 or less, then sacrifices all other creatures they control.
```
tool4ever: should work fine without RepeatEach

Fulgur14: This card is basically reprint of Slaughter the Strong, just with higher mana cost, so I simply reused that code.

tool4ever: then you can fix both ;)

Fulgur14: Wait, isn't the problem the WithTotalPower$ option? If I use similar structure to, say, Cataclysm, wouldn't it limit the selection to 4 power total?

Fulgur14: It's late here -- if this is correct, I'll change Slaughter the Strong tomorrow.

tool4ever: I fixed it for you

Fulgur14: Thanks :)

forge-gui/res/cardsfolder/upcoming/zukos_conviction.txt
```diff
@@ -0,0 +1,10 @@
+Name:Zuko's Conviction
+ManaCost:B
+Types:Instant
+K:Kicker:4
+A:SP$ ChangeZone | ValidTgts$ Creature.YouOwn | TgtPrompt$ Choose target creature card in your graveyard | TargetMin$ X | TargetMax$ X | Origin$ Graveyard | Destination$ Hand | SubAbility$ DBChangeZone | SpellDescription$ Return target creature card from your graveyard to your hand. If this spell was kicked, instead put that card onto the battlefield tapped.
```
tool4ever: can be scripted cleaner, see _From Father to Son_

Fulgur14: Tapped$ True won't cause any problems if the card goes to hand, right?

tool4ever: no, we try to ignore illegally scripted plays

PR #9068, opened 2025-11-03
Allow NameCard | ValidCards$ to filter on colors

Eradev:


Hanmac:
@tool4ever might be a better idea to turn this into Properties like with Card and CardState?

jamincollins:
## Findings

#### 1. Null safety in hasColor
The hasColor method calls input.getColor().hasAnyColor(desiredColor) without checking if input.getColor() returns null. ICardFace.getColor() may return null for cards without color data. Test case will throw NPE.

Test case:
```java
ICardFace face = mockFace();
when(face.getColor()).thenReturn(null);
CardFacePredicates.valid("Card.White").test(face); // NullPointerException
```
Impact: Potential crash when filtering cards with null color.

#### 2. MagicColor.fromName with invalid input
hasColor passes colorName substring directly to MagicColor.fromName. If the input string contains color name as substring (e.g., "WhiteKnight" contains "White"), contains check will match incorrectly and fromName will parse partial name, returning 0 (colorless). This causes false negatives.

Test case:
```java
ICardFace face = mockFaceWithColor(MagicColor.WHITE);
boolean result = CardFacePredicates.valid("Creature.WhiteKnight").test(face);
// m contains "White" via contains(), colorName = "WhiteKnight", fromName returns 0
// hasColor returns false, predicate incorrectly rejects valid card
```
Impact: Incorrect filtering due to substring matching.

#### 1. Null ICardFace in ValidPredicate.test
ValidPredicate.test does not guard against null input. CardFacePredicates.valid may be used in streams where null faces appear. The test method accesses input.getType() without null check.

Test case:
```java
CardFacePredicates.valid("Card.Creature").test(null); // NullPointerException at input.getType()
```
Impact: NPE risk in batch processing.

#### 2. Color name parsing with 'non' prefix
The code extracts colorName via m.substring(mustHave ? 0 : 3). For "nonWhite", mustHave=false, colorName = m.substring(3) = "White". Works. But for "nonWhiteBlue" (invalid but possible), colorName = "WhiteBlue", fromName will parse first char 'W' + 'h' ? Actually fromName with length>2 will lowercase and switch, returning 0. Unclear behavior.

Test case:
```java
CardFacePredicates.valid("Card.nonWhiteBlue").test(face); // ambiguous parsing
```
Impact: Undefined behavior for malformed modifiers.

#### 3. ManaCost null in hasManaCost
hasManaCost calls input.getManaCost().getShortString() without null check on getManaCost(). Could NPE.

Test case:
```java
ICardFace face = mockFaceWithNullManaCost();
CardFacePredicates.valid("Card.ManaCost{1}").test(face); // NPE
```
Impact: Crash on cards with missing mana cost.

#### 1. Substring matching causes false positives
The contains check for White/Blue/Black/Red/Green matches any string containing those substrings. Example: "nonWhite" contains "White", triggers color handling. But "WhiteKnight" also contains "White", triggering color handling incorrectly as property.

Test case:
```java
// Property "WhiteKnight" should be checked via hasProperty, not as color
CardFacePredicates.valid("Creature.WhiteKnight").test(face); // Misinterpreted as color filter
```
Impact: Backwards incompatible change to existing valid strings.

#### 2. Missing case-insensitive handling
MagicColor.fromName lowercases input, but the contains check is case-sensitive. If valid string uses lowercase "white", contains("White") will not match, falling through to hasProperty which will fail.

Test case:
```java
CardFacePredicates.valid("Card.white").test(face); // contains fails, hasProperty fails
```
Impact: Color filtering only works with capitalized names, inconsistent with other parts.

#### 1. hasProperty recursion with non prefix
hasProperty recursively calls itself for "non" prefix. With color filtering now short-circuiting before hasProperty, property names starting with "non" that are also color names may be misrouted. Example "nonWhite" is handled as color, not property.

Test case:
```java
// If a card type property named "White" exists, nonWhite would be misinterpreted
CardFacePredicates.valid("Card.nonWhite").test(face);
```
Impact: Interaction ambiguity between property and color filtering.

#### 2. ValidCards string parsing order
The valid string format is Type.Property1+Property2... Color checks happen before hasProperty fallback. If a property name contains color substring, it will be interpreted as color filter instead of property.

Test case:
```java
CardFacePredicates.valid("Creature.Nonland").test(face); // "Nonland" contains "non", but contains("White") false, falls to hasProperty
```
Impact: Potential misclassification of properties.

### Performance and security
#### 1. Repeated MagicColor.fromName parsing
hasColor calls MagicColor.fromName for every card test. fromName does string operations and switch. In large card lists (e.g., library of 10k cards), this adds overhead. Could cache parsed colors.

Test case:
```java
List<ICardFace> library = ... 10k cards;
library.stream().filter(CardFacePredicates.valid("Card.White")).count(); // Repeated parsing
```
Impact: Performance degradation in bulk filtering.

#### 2. Integer.parseInt without validation
cmcEQ parsing uses Integer.parseInt without try-catch. Malformed cmcEQ value will throw NumberFormatException, crashing predicate evaluation.

Test case:
```java
CardFacePredicates.valid("Card.cmcEQabc").test(face); // NumberFormatException
```
Impact: Unchecked exception propagates to caller.

## Verdict
Request Changes
The PR introduces color filtering but has multiple safety and correctness issues: null safety gaps, substring matching causing false positives/negatives, case sensitivity mismatch, and potential NPEs. Needs robust parsing and defensive checks.

## Recommended Fixes
1. Add null guards for input, input.getColor(), input.getManaCost() in ValidPredicate.test and helper methods.
2. Replace contains() checks with exact token matching or regex word boundaries to avoid substring false positives.
3. Normalize valid string to lowercase before parsing, or make color checks case-insensitive.
4. Parse color modifiers with explicit prefix matching (e.g., m.equals("White") or m.equals("nonWhite")) rather than contains().
5. Wrap Integer.parseInt in try-catch and return false on malformed input.
6. Consider pre-parsing valid string into structured filter objects to avoid repeated parsing.

---

Eradev:
Many of the comments are tied to simply bad scripts. I don't think we want to mute those errors.

tool4ever:
yea this LLM is still a bit crazy in the context it assumes relevant:
> library of 10k cards

Issue #1182, opened 2022-07-23
Refactor ChangeZone that target multiple types


Northmoc:
These cards currently won't interact properly with Quintorius, Fuming Effigy, etc. – 
Crop Sigil, Pull from the Deep, Retrieve, Reconstruct History, Restoration Specialist, Shreds of Sanity

I envision two solutions:
 - modifying targeting to allow for targeting min-max for each value in a list
 - using multiple pumps for Targeting then doing one ChangeZone with `Defined$ Targeted`

_Originally posted by @Northmoc in https://github.com/Card-Forge/forge/pull/1174#discussion_r928122956_

tool4ever:
Maybe doing it with ChangeZoneResolve would be good enough?

tool4ever:
Some other tricky ones because they search mix of hidden and public zones at once:
- Ecological Appreciation
- Journey for the Elixir
- Nissa's Encouragement

PR #9078, opened 2025-11-04
Avatar Destiny (TLA)

Fulgur14:


forge-gui/res/cardsfolder/upcoming/avatar_destiny.txt
```diff
@@ -0,0 +1,16 @@
+Name:Avatar Destiny
+ManaCost:2 G G
+Types:Enchantment Aura
+K:Enchant:Creature.YouCtrl:creature you control
+SVar:AttachAILogic:Pump
+S:Mode$ Continuous | Affected$ Creature.EnchantedBy | AddPower$ X | AddToughness$ X | AddType$ Avatar | Description$ Enchanted creature gets +1/+1 for each creature card in your graveyard and is an Avatar in addition to its other types.
+T:Mode$ ChangesZone | ValidCard$ Card.AttachedBy | Origin$ Battlefield | Destination$ Graveyard | Execute$ TrigMill | TriggerDescription$ When enchanted creature dies, mill cards equal to its power. Return this card to its owner's hand and up to one creature card milled this way to the battlefield under your control.
+SVar:TrigMill:DB$ Mill | NumCards$ Y | RememberMilled$ True | SubAbility$ DBChangeZoneHand
+SVar:DBChangeZoneHand:DB$ ChangeZone | Origin$ Graveyard | Destination$ Hand | ChangeZoneTable$ True | SubAbility$ DBChangeZoneBattlefield
```
tool4ever: like I said not supported yet
maybe @Hanmac wants to pick it up?

forge-gui/res/cardsfolder/upcoming/avatar_destiny.txt
```diff
@@ -0,0 +1,16 @@
...
+SVar:DBChangeZoneBattlefield:DB$ ChangeZone | Hidden$ True | Origin$ Graveyard,Exile | Destination$ Battlefield | ChangeType$ Creature.IsRemembered | GainControl$ True | SelectPrompt$ Select up to one creature card | SubAbility$ DBResolve
+SVar:DBResolve:DB$ ChangeZoneResolve | SubAbility$ DBCleanup
+SVar:DBCleanup:DB$ Cleanup | ClearChosenColor$ True
```
tool4ever: wrong cleanup

forge-gui/res/cardsfolder/upcoming/avatar_destiny.txt
tool4ever: otherwise remove the bad parts
and it can be done via #1182 later

Fulgur14: OK, so you mean do it with two moves for now?

tool4ever: you could use some other helper but I don't think it's worth it 🤷‍♂️

PR #9081, opened 2025-11-04
TLA cards, 4th November, batch 2

Fulgur14:


forge-gui/res/cardsfolder/upcoming/earthbender_ascension.txt
```diff
@@ -0,0 +1,13 @@
...
+SVar:TrigEarthbend:DB$ Earthbend | Num$ 2 | SubAbility$ DBSearch
+SVar:DBSearch:DB$ ChangeZone | Origin$ Library | Destination$ Battlefield | Tapped$ True | ChangeType$ Land.Basic | ChangeTypeDesc$ basic land
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Land.YouCtrl | TriggerZones$ Battlefield | Execute$ TrigPutCounterQuest | TriggerDescription$ Landfall — Whenever a land you control enters, put a quest counter on this enchantment. When you do, if it has four or more quest counters on it, put a +1/+1 counter on target creature you control. It gains trample until end of turn.
+SVar:TrigPutCounterQuest:DB$ PutCounter | Defined$ Self | CounterType$ QUEST | CounterNum$ 1 | RememberAmount$ True | SubAbility$ TrigImmediateTrig
+SVar:TrigImmediateTrig:DB$ ImmediateTrigger | Execute$ TrigPutCounterP1P1 | TriggerAmount$ Count$RememberedNumber | IsPresent$ Card.Self+counters_GE4_QUEST | SubAbility$ DBCleanup | TriggerDescription$ When you do, if it has four or more quest counters on it, put a +1/+1 counter on target creature you control. It gains trample until end of turn.
```
tool4ever: you can try but I think such nested intervening needs to be split into two conditions?

Fulgur14: I'm mainly a bit confused about the TriggerAmount$ -- it's just if the variable is greater than 0, the trigger can trigger?

I could probably rewrite it to use a single SVar check that combines both conditions. Would that be better?

Fulgur14: OK, good news -- this actually works. Bad news - when I tried it, the fourth counter triggered four times. It seems "TriggerAmount$ Count$RememberedNumber" remembers the total amount of quest counters and triggers once for each, but if it does... wouldn't every single card that uses this be implemented wrong? Would adding Limit Max fix this?

tool4ever: you have currently scripted it to always create the reflexive trigger, it just waits until the counter number is right
so just add it as condition too

Fulgur14: OK, sorry, can you explain this a bit more, please?

tool4ever: `ConditionPresent$ Card.Self+counters_GE4_QUEST`

tool4ever: nooo, the IsPresent is also needed because it's "intervening if" 😱

Issue #8942, opened 2025-10-17
Hardcoded "gains this card’s other abilities" doesn't work with added abilites


Jetz72:
Currently, [Symbiote Spider-Man](https://scryfall.com/card/spm/156/symbiote-spider-man)'s phrasing "gains this card’s other abilities" is just hard-coded to add a copy of the card's other printed ability. Since this effect happens while the card is freshly exiled from the graveyard, it usually won't have any other abilities. However, perpetual effects can throw this expectation out the window. This interaction can be done in arena, where the perpetual added ability is copied by this effect as well:

<img width="381" height="529" alt="Image" src="https://github.com/user-attachments/assets/d9a9bc3e-bfea-4d72-8ddd-355ceee0c516" />

<img width="339" height="469" alt="Image" src="https://github.com/user-attachments/assets/516547d8-f29a-4c65-bf08-6edbceda7d5b" />

<img width="656" height="539" alt="Image" src="https://github.com/user-attachments/assets/53bd1a20-54ed-4725-83dc-c29004fe1f68" />

(Images provided by Atticus Von Shwoll on Discord)

Smogbelcher Chariot gives perpetual keywords to Skv'x the Augmenter (i.e. Symbiote Spider-Man), and they were successfully copied onto a Skeleton token.

Might work to give `AnimateEffect` a parameter to indicate "copy all other abilities from this card". I'm not sure how this should interact with abilities added temporarily by ongoing effects or external static abilities though.

AtticusVonShwoll:
One important thing to mention is that "Copy all other abilities from this card" apply to any abilities the card has at that moment depending on where the card is. When it comes to Symbiote Spiderman(Skv'x) the ability only abilities that would transfer would be ones that are applying to the card when its in exile since the "last known information" thing does not apply as the text for the card says "it gains this card's other abilities" so it keeps checking while its in exile, which is why perpetual effects continue to work.

PR #9094, opened 2025-11-06
Trigger: reuse Execute for multiple Trigger

### LINKED ISSUE #8595: Corruption of Towashi + ResolvedLimit

### LINKED ISSUE #9091: Multiple Trigger combined with same Execute
This is an Issue for me to track the Problem that cards with complex triggers like this:

https://github.com/Card-Forge/forge/blob/227a1df06c06dfa99182cd5c5954e2a23c7c1ea9/forge-gui/res/cardsfolder/v/victor_valgavoths_seneschal.txt#L5-L7

should reuse the `Execute` SpellAbility

CardFactory is called there:
https://github.com/Card-Forge/forge/blob/227a1df06c06dfa99182cd5c5954e2a23c7c1ea9/forge-game/src/main/java/forge/game/card/CardFactory.java#L387-L388

and the Trigger will be built there:
https://github.com/Card-Forge/forge/blob/227a1df06c06dfa99182cd5c5954e2a23c7c1ea9/forge-game/src/main/java/forge/game/trigger/TriggerHandler.java#L143-L146

And in the Trigger, the SpellAbility will be built there:
https://github.com/Card-Forge/forge/blob/42cb1f55e1809ed9ff7d52acf6310ae55a0ca8ba/forge-game/src/main/java/forge/game/trigger/Trigger.java#L613-L616

in theory, i just need to see for intrinsic Trigger, where SVarHolder = CardState

Hanmac:
Closes #9091
Closes #8595

### DIFF
--- a/forge-gui/res/cardsfolder/c/corruption_of_towashi.txt
+++ b/forge-gui/res/cardsfolder/c/corruption_of_towashi.txt
@@ -3,12 +3,8 @@ ManaCost:4 U
 Types:Enchantment
 T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigIncubate | TriggerDescription$ When CARDNAME enters, incubate 4. (Create an Incubator token with four +1/+1 counters on it and "{2}: Transform this artifact." It transforms into a 0/0 Phyrexian artifact creature.)
 SVar:TrigIncubate:DB$ Incubate | Amount$ 4
-T:Mode$ Transformed | ValidCard$ Permanent.YouCtrl+inZoneBattlefield | Execute$ TrigDraw | TriggerZones$ Battlefield | OptionalDecider$ You | CheckSVar$ X | SVarCompare$ LT1 | TriggerDescription$ Whenever a permanent you control transforms or a permanent you control enters transformed, you may draw a card. Do this only once each turn.
-T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Permanent.Transformed+YouCtrl | OptionalDecider$ You | TriggerZones$ Battlefield | Execute$ TrigDraw | Secondary$ True | CheckSVar$ X | SVarCompare$ LT1 | TriggerDescription$ Whenever a permanent you control transforms or a permanent you control enters transformed, you may draw a card. Do this only once each turn.
-SVar:TrigDraw:DB$ Draw | SubAbility$ DBLog
-SVar:DBLog:DB$ StoreSVar | SVar$ X | Type$ Number | Expression$ 1
-SVar:X:Number$0
-T:Mode$ Phase | Phase$ Cleanup | TriggerZones$ Battlefield | Execute$ DBCleanup | Static$ True
-SVar:DBCleanup:DB$ StoreSVar | SVar$ X | Type$ Number | Expression$ 0
+T:Mode$ Transformed | ValidCard$ Permanent.YouCtrl+inZoneBattlefield | Execute$ TrigDraw | TriggerZones$ Battlefield | OptionalDecider$ You | ResolvedLimit$ 1 | TriggerDescription$ Whenever a permanent you control transforms or a permanent you control enters transformed, you may draw a card. Do this only once each turn.
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Permanent.Transformed+YouCtrl | OptionalDecider$ You | TriggerZones$ Battlefield | Execute$ TrigDraw | Secondary$ True | ResolvedLimit$ 1 | TriggerDescription$ Whenever a permanent you control transforms or a permanent you control enters transformed, you may draw a card. Do this only once each turn.
+SVar:TrigDraw:DB$ Draw
 DeckHas:Ability$Counters|Token & Type$Incubator|Artifact|Phyrexian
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/trigger/Trigger.java
```diff
@@ -611,7 +611,11 @@ public final Object clone() {
     public SpellAbility ensureAbility(final IHasSVars sVarHolder) {
         SpellAbility sa = getOverridingAbility();
         if (sa == null && hasParam("Execute")) {
-            sa = AbilityFactory.getAbility(getHostCard(), getParam("Execute"), sVarHolder);
+            if (this.isIntrinsic() && sVarHolder instanceof CardState state) {
```
tool4ever, line 619: can you make the intrinsic check smarter somehow?
otherwise it breaks again if a _Clone_ copies these cards :/

tool4ever: maybe Table instead so it can be stored/compared by State too? 🤔

Hanmac: @tool4ever Clone shouldn't be affected by this, i think
because when Clone copies a Trigger, it already has the SpellAbility set

Or do you think the Clone breaks that the two Trigger point at the same SpellAbility?
That's something i would need to debug

tool4ever: hmn yes, looks like `Trigger.copy` needs some detection logic

tool4ever:
```
            if (tr.isIntrinsic()) {
                Trigger t = tr.copy(card, lki);
                t.overridingAbility = null;
                t.ensureAbility(tr.getCardState());
                triggers.add(t);
            }
```

something like this seems to work in ``CardState.copy``
but needs cleaner implementation

Hanmac: `t.overridingAbility = null;` i get a feeling that this might break some cards

i think the cleanest way would be to pass the new CardState in copyFrom there to Trigger.copy:
https://github.com/Card-Forge/forge/blob/9248deebdba3de0bd56365af96a54ec1dead1950/forge-game/src/main/java/forge/game/card/CardState.java#L761

Hanmac:
@tool4ever we might also need to take a look at this issue #8942
(even though it's Animate)

forge-game/src/main/java/forge/game/trigger/Trigger.java
Hanmac: in `CardState.copyFrom` i should probably copy `abilityForTrigger` first

then i might know which of them can reuse the copied value

Hanmac: @tool4ever i can't test it right now, but does this work?

(i might refactor the copy abilities more to always use CardState instead of Card 🤔 )

Hanmac:
i will need to add `Oasis of Renewal` to this MR too

forge-game/src/main/java/forge/game/trigger/Trigger.java
tool4ever: well _Clone_ counting seems fine this way

but still fails if _Idris_ gains it

Hanmac: Idris should be fixed for now

tool4ever:
@Hanmac
> did you see my ping in #9095?
> need to decide if you want to handle that special case too

Hanmac:
@tool4ever the problem with planetarium is a bit deeper, it probably would be better to use Remember for the PlayAbility

Hanmac:
@tool4ever if the planetarium is handled differently, can this be merged?

or do we need something for Animate first? https://github.com/Card-Forge/forge/issues/8942

forge-game/src/main/java/forge/game/card/CardState.java
```diff
@@ -951,6 +957,14 @@ public SpellAbility getCloakUp() {
         return cloakUp;
     }
 
+    public SpellAbility getAbilityForTrigger(String svar) {
+        return abilityForTrigger.computeIfAbsent(svar, s -> AbilityFactory.getAbility(getCard(), s, this));
+    }
+
+    public boolean hasAbilityForTrigger(String svar) {
```
tool4ever: function no longer used

forge-game/src/main/java/forge/game/trigger/Trigger.java
tool4ever: negative, it looks like it can still collide "easily":
image

- I exiled both Victor
- cast the enchantment
- the second trigger already forced the discard

:/

Hanmac: found and fixed, i think (HashTable vs TreeTable)

tool4ever (approved):
it's somewhat complicated
maybe a few comments might be good later

PR #9095, opened 2025-11-06
Planetarium of Wan Shi Tong (TLA) -- should work with #9091

Fulgur14:


tool4ever:
Technically this card might be even more problematic since previously the "you may" parts were always before the first instruction:
So it could be argued you can look multiple times until you decide to cast...

I doubt they bother with that in the Release Notes so I might just approve it later anyway 🤷‍♂️

Fulgur14:
We can ask later if not clear, but I think even the looking is included.

tool4ever:
Ugh it seems I was both wrong and right:

> Once you choose to cast the top card of your library, Planetarium of Wan Shi Tong's ability won't trigger again that turn.

@Hanmac
I suppose this means we'd have to track each subability too 🤔
but the problem is those Optional$ happen too late as part of the effect

Hanmac:
@tool4ever probably better to use Remember on the Play Effect 🤔

tool4ever:
@Fulgur14
ok, looks like for now you'll still have to stick with Static cleanup (ChangesZone + TurnBegin)

Hanmac:
I wouldn't use StoreSVar, but just use Remembered

Fulgur14:
Like this?

Hanmac:
Add this to the Triggers to stop them when they already applied this turn:
`PresentDefined$ Remembered | IsPresent$ Card | PresentCompare$ EQ0`

to follow this:
> Once you choose to cast the top card of your library, Planetarium of Wan Shi Tong's ability won't trigger again that turn.

PR #9096, opened 2025-11-06
Minor Adventure Updates

### LINKED ISSUE #8822: Adventure Mode: Inn hosted draft events look seriously biased towards newer sets
**Describe the bug**

In my current Adventure mode playthrough, in the 60-80 towns I have visited (didn't actually count, but it sure felt like this number), not a single inn hosted a Mirage, Tempest or Urza Block Draft event. It's almost always some newer or standard legal set. Once in a blue moon, I may get something old like an Invasion Block Draft, but I have yet to encounter a Draft event for a block older than that!

Is there a way to bias draft events set selection towards certain sets/blocks in the adventure mode config? Or at least not make it supremely biased towards newer/standard legal sets?

**To Reproduce**

Try to find a town inn that hosts a Mirage, Tempest or Urza Block Draft event. Give up.

**Expected behavior**

Either I can configure what sets/blocks appear in hosted draft events or make the random set selection not so weighted towards recent sets.

**Screenshots**

N/A

 - OS: Windows 10
 - Browser N/A
 - Version 2.0.07-SNAPSHOT-10.01

**Additional context**

Jetz72:
Add support for `allowedEvents` to the config per user request and to better support custom Adventure worlds. It can override most other filters.

Removed standard sets from the set pool filter when a non-standard pool is selected. Closes #8822.

Expanded the doppelganger anti-cheese to consider distinct cards and not just total paper cards.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/data/AdventureEventData.java
+++ b/forge-gui-mobile/src/forge/adventure/data/AdventureEventData.java
@@ -197,7 +197,7 @@ public class AdventureEventData implements Serializable {
         ConfigData configData = Config.instance().getConfigData();
         Predicate<CardEdition> filter = CardEdition.Predicates.CAN_MAKE_BOOSTER;
 
-        if(configData.allowedEvents != null) {
+        if(configData.allowedEvents != null && configData.allowedEvents.length > 0) {
             Set<String> allowedEvents = Set.of(configData.allowedEvents);
             filter = filter.and(q -> allowedEvents.contains(q.getCode()));
         }
@@ -210,7 +210,7 @@ public class AdventureEventData implements Serializable {
                 Set<String> restrictedEvents = Set.of(configData.restrictedEvents);
                 filter = filter.and((q) -> !restrictedEvents.contains(q.getCode()));
             }
-            if (configData.allowedEditions != null) {
+            if (configData.allowedEditions != null && configData.allowedEditions.length > 0) {
                 Set<String> allowed = Set.of(configData.allowedEditions);
                 filter = filter.and(q -> allowed.contains(q.getCode()));
             } else if(configData.restrictedEditions != null) {


Arthur-F:
Wouldn't something like this make more sense so as to not warp the probabilities of getting newer packs for #8822? In theory this should make pack distribution obey the expected probabilities on AventureEventData, no?


```

    // Era 1: Standard sets (This is the base, no change)
    Predicate<CardEdition> filterEraStandard = filterStandard;

    // Era 2: Pioneer-exclusive sets
    // (In Pioneer) AND (NOT in Standard)
    Predicate<CardEdition> filterEraPioneer = filterPioneer.and(filterStandard.negate());

    // Era 3: Modern-exclusive sets
    // (In Modern) AND (NOT in Pioneer)
    Predicate<CardEdition> filterEraModern = filterModern.and(filterPioneer.negate());

    // Era 4: Vintage-exclusive sets ("The Rest")
    // (In Vintage) AND (NOT in Modern)
    Predicate<CardEdition> filterEraVintage = filterVintage.and(filterModern.negate());
```

Jetz72:
That was originally how I was going to do it, but then I realized "Vintage but not Modern" and "Modern but not Pioneer" are relatively small and fixed pools, whereas "Pioneer but not Standard" is larger and will expand with every rotation. We'd end up with the opposite problem, where the much older sets get over-represented. 

I figure what we want is Standard to be a bit more common for how few sets it has (but not as much as before), and older eras like Vintage-only to be a bit rarer to see. So I just cut Standard out of the non-Standard pools to put an upper bound on how common they are.

Arthur-F:
> That was originally how I was going to do it, but then I realized "Vintage but not Modern" and "Modern but not Pioneer" are relatively small and fixed pools, whereas "Pioneer but not Standard" is larger and will expand with every rotation. We'd end up with the opposite problem, where the much older sets get over-represented.
> 
> I figure what we want is Standard to be a bit more common for how few sets it has (but not as much as before), and older eras like Vintage-only to be a bit rarer to see. So I just cut Standard out of the non-Standard pools to put an upper bound on how common they are.

That makes sense, good catch! The only thing that feels a little bad to me is how it skews the percentages for each set representation , since some of them double/triple/quadruple dip.
I suppose the optimal solution would be if we had weighted probabilities based on the amount of cards/sets per format, but your approach already works to mitigate the problem and is much better than the current implementation in my opinion.

forge-gui-mobile/src/forge/adventure/data/AdventureEventData.java
```diff
@@ -175,17 +175,17 @@ private static CardBlock pickCardBlockByFormat(AdventureEventController.EventFor
...
     public static Predicate<CardEdition> selectSetPool() {
-        // Should we negate any of these to avoid overlap?
         final int rollD100 = MyRandom.getRandom().nextInt(100);
         Predicate<CardEdition> rolledFilter;
         if (rollD100 < 30) {
             rolledFilter = filterStandard;
         } else if (rollD100 < 60) {
-            rolledFilter = filterPioneer;
+            // Remove standard from older pools because its representation is already inflated.
+            rolledFilter = filterPioneer.and(filterStandard.negate());
```
tehdiplomat, line 184: I could have sworn we already did this. Sigh.

PR #9114, opened 2025-11-08
Search fixes

Eradev:


docs/_sidebar.md
```diff
@@ -1,26 +1,26 @@
-### [Forge Wiki Home](Home.md)
+# [Forge Wiki Home](Home.md)
```
tool4ever: Appreciate the wiki touches
I also thought it weird to start with the third heading but when it's in the Wiki it looked way too huge imo

docs/User-Guide.md
```diff
@@ -52,44 +61,50 @@ _**Download and unpack/install the package to their own new folder!**_
...
+
+## Gameplay
 
-# Gameplay
+### Full Control
 
-## Full Control
+Right click/long tap on your player avatar:
 
-Right click/long tap on your player avatar:<br />
```
tool4ever, line 94: FYI I'm actually using the html breaks in some places to only wrap to new line
your change will insert a full empty one between instead (which I think often looks ugly)

but I'm curious if you think it's better like this anyway? (for the rare chance someone only views it in text editor?!)

docs/Creating-a-custom-Set.md
```diff
@@ -1,3 +1,5 @@
+# Creating a custom set
```
tool4ever: I have usually decided to avoid adding the same title because in the Wiki it actually generates one from the filename first already

so it looks a bit weird if you have both under each other :/

what do you think is better?

docs/User-Guide.md
Eradev: Oh, then we can use something other than `<br>`. I personally prefer not having html tags inside .md files when possible.

https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#line-breaks

docs/Creating-a-custom-Set.md
Eradev: It breaks the title hierarchy. I can live without it if you don't mind.

docs/User-Guide.md
tool4ever: true, I guess those spaces would be a good alternative too :)

docs/_sidebar.md
Eradev: Again, it breaks the title hierarchy. Usually I'd advocate for not using wrong tags for display reasons, but it's a minor issue  and I can revert it.

tool4ever: for this I'd insist since it takes way too much space otherwise, especially if you browse on mobile
if there's a better way to increase the text size slightly feel free to switch to that though
thanks!

tool4ever (approved):

PR #9115, opened 2025-11-08
User guide update

tool4ever:


docs/User-Guide.md
Eradev, line 1: Ideally, a markdown should only have one main title (`#`) as it acts like a `<H1>` HTML tag.

I'm not sure how much importance you want to give to accessibility and SEO for these files VS how they display though.

tool4ever: imo it looks weird to have all relevant topics nested directly :/
in both wiki:
image

or IntelliJ:
image

Eradev: Maybe we should separate them into smaller, distinct pages then?

tool4ever: maybe?
I admit to preferring reading somewhat longer documents without having to switch around too much but there might be more arguments for a different approach :)

PR #9120, opened 2025-11-09
Initial checkin for Waterbending

tehdiplomat:


forge-game/src/main/java/forge/game/cost/CostWaterbend.java
```diff
@@ -0,0 +1,22 @@
...
+public class CostWaterbend extends CostPartMana {
+    private int maxWaterbend;
+
+    public CostWaterbend(final ManaCost cost) {
+        super(cost, null);
+
+        maxWaterbend = cost.getGenericCost();
+    }
+
+    public int getMaxWaterbend() {
```
Hanmac: I would use getAmount there? 🤔

tehdiplomat: Hmm but I'm returning an integer not a String?

Hanmac: It's still an Amount, I just thought a more generic name would be better

Hanmac: Amount might not work there, put away for later days

Hanmac (approved):

PR #9127, opened 2025-11-10
More waterbending cards

Fulgur14:
Aang, Swift Savior might require airbending update to accomodate working with spells.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/hama_the_bloodbender.txt
+++ b/forge-gui/res/cardsfolder/upcoming/hama_the_bloodbender.txt
@@ -2,7 +2,7 @@ Name:Hama, the Bloodbender
 ManaCost:2 UB UB UB
 Types:Legendary Creature Human Warlock
 PT:3/3
-T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigMill | TriggerDescription$ When NICKNAME enters, target opponent mills three cards. Exile up to one noncreature, nonland card from that player's graveyard. For as long as you control Hama, you may cast the exiled card during your turn by waterbending {X} rather than paying its mana cost, where X is its mana value. (While paying a waterbend cost, you can tap your artifacts and creatures to help. Each one pays for {1}.)
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigMill | TriggerDescription$ When NICKNAME enters, target opponent mills three cards. Exile up to one noncreature, nonland card from that player's graveyard. For as long as you control NICKNAME, you may cast the exiled card during your turn by waterbending {X} rather than paying its mana cost, where X is its mana value. (While paying a waterbend cost, you can tap your artifacts and creatures to help. Each one pays for {1}.)
 SVar:TrigMill:DB$ Mill | NumCards$ 3 | ValidTgts$ Opponent | TgtPrompt$ Select target opponent to mill | SubAbility$ DBChangeZone
 SVar:DBChangeZone:DB$ ChangeZone | Optional$ True | Origin$ Graveyard | Destination$ Exile | DefinedPlayer$ Targeted | Chooser$ You | ChangeType$ Card.nonCreature+nonLand | ChangeNum$ 1 | IsCurse$ True | RememberChanged$ True | SubAbility$ DBEffect
 SVar:DBEffect:DB$ Effect | StaticAbilities$ STPlay | RememberObjects$ Remembered | Duration$ AsLongAsControl | ExileOnMoved$ Exile | SubAbility$ DBCleanup


forge-gui/res/cardsfolder/upcoming/crashing_wave.txt
```diff
@@ -0,0 +1,8 @@
+Name:Crashing Wave
+ManaCost:U U
+Types:Sorcery
+S:Mode$ RaiseCost | ValidCard$ Card.Self | Activator$ You | Type$ Spell | Cost$ Waterbend<X> | EffectZone$ All | Description$ As an additional cost to cast this spell, waterbend {X}. (While paying a waterbend cost, you can tap your artifacts and creatures to help. Each one pays for {1}.)
+A:SP$ Tap | TargetMin$ 0 | TargetMax$ X | TgtPrompt$ Select up to X target creatures to tap | ValidTgts$ Creature | SubAbility$ DBPutCounter
+SVAr:DB$ PutCounter | Choices$ Creature.tapped+OppCtrl | ChoiceTitle$ Choose any number of tapped creatures your opponents control to distribute counters to | CounterType$ STUN | CounterNum$ 3 | ChoiceAmount$ 3 | MinChoiceAmount$ 0 | DividedAsYouChoose$ 3
```
tool4ever: typo

forge-gui/res/cardsfolder/upcoming/spirit_water_revival.txt
```diff
@@ -0,0 +1,12 @@
+Name:Spirit Water Revival
+ManaCost:1 U U
+Types:Sorcery
+S:Mode$ OptionalCost | EffectZone$ All | ValidCard$ Card.Self | ValidSA$ Spell | Cost$ Waterbend<6> | Description$ As an additional cost to cast this spell, you may waterbend {6}. (While paying a waterbend cost, you can tap your artifacts and creatures to help. Each one pays for {1}.)
+A:SP$ ChangeZoneAll | ChangeType$ Card.YouOwn | Origin$ Graveyard | Destination$ Library | Shuffle$ True | CheckSVar$ X | SubAbility$ DBDraw | SpellDescription$ Draw two cards. If this spell's additional cost was paid, instead shuffle your graveyard into your library, draw seven cards, and you have no maximum hand size for the rest of the game.
+SVar:DBDraw:DB$ Draw | NumCards$ Y | SubAbility$ DBEffect
+SVar:DBEffect:DB$ Effect | StaticAbilities$ STHandSize | Duration$ Permanent | CheckSVar$ X | SubAbility$ DBExile | SpellDescription$ You have no maximum hand size for the rest of the game.
```
tool4ever: you're using restriction instead of condition again

forge-gui/res/cardsfolder/upcoming/spirit_water_revival.txt
```diff
@@ -0,0 +1,12 @@
...
+SVar:STHandSize:Mode$ Continuous | Affected$ You | SetMaxHandSize$ Unlimited | Description$ You have no maximum hand size for the rest of the game.
+SVar:DBExile:DB$ ChangeZone | Origin$ Stack | Destination$ Exile
```
tool4ever, line 9: description missing

forge-gui/res/cardsfolder/upcoming/spirit_water_revival.txt
Fulgur14: BTW, would anything break if that syntax was unified? :( Is there any situation where both could be used, with different meanings?

tool4ever: That's like asking if `Defined$` and `ValidTgts$` can be unified
I don't know how else to explain it anymore 😢

forge-gui/res/cardsfolder/upcoming/crashing_wave.txt
tool4ever: you only fixed half

forge-gui/res/cardsfolder/upcoming/spirit_water_revival.txt
Fulgur14: I know of cards that use Defined$ and ValidTgts$ in a single ability, though.

tool4ever: well that's on top
but it's not about combining both, how else is the engine supposed to know which one you mean

forge-gui/res/cardsfolder/upcoming/crashing_wave.txt
tool4ever: image

forge-gui/res/cardsfolder/upcoming/spirit_water_revival.txt
Fulgur14: That's what I'm asking about -- in what situation you could use both. But I guess it's similar reason as to why continuous effects use AddPower$ and AddToughness$ while Pump uses NumAtt$ and NumDef$?

tool4ever: no, that could be unified
I'm not sure any official card needs both together right now but that doesn't matter because **they mean different things**

Fulgur14: So what is the difference in meaning? "Check this once" vs. "Check this continuously"?

tool4ever: Restrictions happen here:
> 601.2. To cast a spell is to take it from where it is (usually the hand), put it on the stack, and pay its costs, so that it will eventually resolve and have its effect. Casting a spell includes proposal of the spell (rules 601.2a-d) and determination and payment of costs (rules 601.2f-h). To cast a spell, a player follows the steps listed below, in order. A player must be legally allowed to cast the spell to begin this process (see rule 601.3). If a player is unable to comply with the requirements of a step listed below while performing that step, the casting of the spell is illegal; the game returns to the moment before the casting of that spell was proposed (see rule 732, "Handling Illegal Actions").

but Conditions:
> 608.2h. If an effect requires information from the game (such as the number of creatures on the battlefield), the answer is determined only once, when the effect is applied. If the effect requires information from a specific object, including the source of the ability itself, the effect uses the current information of that object if it's in the public zone it was expected to be in; if it's no longer in that zone, or if the effect has moved it from a public zone to a hidden zone, the effect uses the object's last known information. See rule 113.7a. If an ability states that an object does something, it's the object as it exists--or as it most recently existed--that does it, not the ability.

Fulgur14: OK, I'll try to remember it. I keep making this mistake, I know...

forge-gui/res/cardsfolder/upcoming/hama_the_bloodbender.txt
```diff
@@ -0,0 +1,11 @@
+Name:Hama, the Bloodbender
+ManaCost:2 UB UB UB
+Types:Legendary Creature Human Warlock
+PT:3/3
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigMill | TriggerDescription$ When NICKNAME enters, target opponent mills three cards. Exile up to one noncreature, nonland card from that player's graveyard. For as long as you control Hama, you may cast the exiled card during your turn by waterbending {X} rather than paying its mana cost, where X is its mana value. (While paying a waterbend cost, you can tap your artifacts and creatures to help. Each one pays for {1}.)
```
tool4ever: NICKNAME

Hanmac:
@tool4ever can these be ready soon?

tool4ever:
right now I'm not sure when I have enough time to add the support from IDE
ideally another coder can handle it too

tehdiplomat (approved):

PR #9129, opened 2025-11-10
Secret of Bloodbending and support

tool4ever:


Hanmac:
@tool4ever i still think that this should be inside the CostWaterbend:
`triggerElementalBend(TriggerType.Waterbend);`

tool4ever:
hmn I don't think there's a way to reach cost parts like that again but feel free to try 🤔

Hanmac:
> hmn I don't think there's a way to reach cost parts like that again but feel free to try 🤔

Doesn't CostWaterbend.payAsDecided get called? Or is Waterbend merged with the generic CostPartMana?

tool4ever:
yea it can get merged

PR #9137, opened 2025-11-11
Add variant names for OM1

paulsnoops:


tool4ever:
Are the images grabbed correctly from Scryfall?

paulsnoops:
> Are the images grabbed correctly from Scryfall?

Yeah image fetching works fine. The only thing I've noticed not implemented is the card detail referencing the original card but that can come later. Like this:

image

Jetz72:
Ah, right. When you add the flavor name via the edition file, it also tacks on a `Variant:whatever:Oracle:` parameter that has the original name find/replaced with the flavor name. I can add in some logic so it can also do that for flavor names specified in the script. 

One issue that emerges with automating the Oracle text though is any differences besides the name. Keywords (e.g. "Web-slinging" -> "Enweb") will probably need a mapping defined separately, so that we can also replace them in-game. Flavor words (e.g. "Lizard Formula" and "Venom Blast") aren't included on the Omenpaths prints, but I don't know if there's a reliable way to remove those automatically. They also appear in `TriggerDescription` parameters for their corresponding abilities. And I don't think there are any disparities like this for this set in particular, but they have started using pronouns for legendary cards, and if they ever use a different one for a Universes Within printing, that'll probably just need to be entered in manually.

Hanmac:
@Jetz72 "automating the Oracle text"
we are going to update the Oracle Text only for the Spider-Man cards, not for the OmenPath ones.
So we don't care about Enweb

as for Flavor words, i wanted to make them displayed differently, but we don't really have a great renderer for _italic_

Jetz72:
I'll just do the name replacement for now and we can worry about the more complicated bits later. Universes Within names for keywords might be a future headache for us, if for instance they print a new card with a UW equivalent of "Firebending" or "The Ring Tempts You", and yet rule that those mechanics still interact with triggers on cards like Avatar Aang or Call of the Ring.

Issue #9136, opened 2025-11-11
Typo: Stack/Log message for 'chosen source' effect lacks a space after player name


Vhati:
**Describe the bug**

When using damage prevention and such, there's a prompt to choose the source of anticipated damage. The message reflecting that decision in the Log/Stack has a typo. It mashes together the player name and the next word.

[ChooseSourceEffect.java](https://github.com/Card-Forge/forge/blob/a2b324c9ce12bc3a3ca54fbaf11c42439a3f94eb/forge-game/src/main/java/forge/game/ability/effects/ChooseSourceEffect.java#L24)

> protected String getStackDescription(SpellAbility sa) {
> &ensp;&ensp;final StringBuilder sb = new StringBuilder();
> &ensp;&ensp;sb.append(Lang.joinHomogenous(getTargetPlayers(sa)));
> &ensp;&ensp;sb.append(**"chooses a source."**);

Before "chooses" here, there ought to be a space character.

[joinHomogenous()](https://github.com/Card-Forge/forge/blob/a2b324c9ce12bc3a3ca54fbaf11c42439a3f94eb/forge-core/src/main/java/forge/util/Lang.java#L73) just builds a "this *and* that" phrase when there are multiple players. It wouldn't add a trailing space.

For comparison, in the message for [DamagePreventEffect.java](https://github.com/Card-Forge/forge/blob/a2b324c9ce12bc3a3ca54fbaf11c42439a3f94eb/forge-game/src/main/java/forge/game/ability/effects/DamagePreventEffect.java#L70), those successive sentence fragments have a leading space.

**To Reproduce**

1. Cast "Rune of Protection: Red", and activate it with red permanents in play.
2. Choose a source from the list prompt.
3. See the log: "Rune of Protection: Red (8) - **InsertName**chooses a source."

IIRC the typoed message would appear on the stack, too, if activation were in response to immanent direct damage already on the stack.

**Operating System**
2.0.07-snapshot-10.25 on Android 15.

PR #9142, opened 2025-11-11
Issue #9136 fix

aryst0krat:
Quick little fix for a formatting issue


aryst0krat:
Connected to the wrong issue while looking into multiple issues

PR #9160, opened 2025-11-14
StaticAbility: add getTimestamp using getLayerTimestamp

Hanmac:
Part for #7378

Adds way to have StaticAbilities with different Timestamp than its HostCards LayerTimestamp

### DIFF
--- a/forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java
+++ b/forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java
@@ -92,17 +92,16 @@ public final class StaticAbilityContinuous {
         final List<Player> affectedPlayers = StaticAbilityContinuous.getAffectedPlayers(stAb);
         final Game game = hostCard.getGame();
 
-
-        // nothing more to do
-        if (stAb.hasParam("Affected") && affectedPlayers.isEmpty() && affectedCards.isEmpty()) {
-            return affectedCards;
-        }
-
         final StaticEffect se = game.getStaticEffects().getStaticEffect(stAb);
         se.setAffectedCards(affectedCards);
         se.setAffectedPlayers(affectedPlayers);
         se.setParams(params);
         se.setTimestamp(stAb.getTimestamp());
+
+        // nothing more to do
+        if (stAb.hasParam("Affected") && affectedPlayers.isEmpty() && affectedCards.isEmpty()) {
+            return affectedCards;
+        }
 
         String addP = "";
         int powerBonus = 0;


forge-game/src/main/java/forge/game/staticability/StaticAbility.java
```diff
@@ -583,6 +583,13 @@ public int compareTo(StaticAbility arg0) {
         .result();
     }
 
+    public long getTimestamp() {
+        if (hasParam("Timestamp")) {
+            return Long.valueOf(getParam("Timestamp"));
```
tool4ever, line 588: technically you already handled "command zone pollution" with `setRenderForUI`
but it might still be better for performance if we avoid extra Card objects anyway 🤔

so without effect as hostCard it will need to use the affected Card I guess?
need to think about any problems that might cause...

forge-game/src/main/java/forge/game/staticability/StaticAbility.java
```diff
@@ -583,6 +583,13 @@ public int compareTo(StaticAbility arg0) {
         .result();
     }
 
+    public long getTimestamp() {
```
tool4ever, line 586: imo already refactor `Card.suspectedEffect` this way here
so we know if it works before adding extra code for it

tool4ever: but actually we need to be careful where we could even use this?
https://www.reddit.com/r/mtgrules/comments/1bvz35u/question_about_suspected_creatures/

and that would be the same for keywordCounters

Hanmac: yes, all my other stuff i did before, like Suspected could be cleanup more

it's so that while the hidden static can't be removed by "remove abilities", the gained keywords and traits can be removed

forge-game/src/main/java/forge/game/staticability/StaticAbility.java
Hanmac: Yes, this part is only for stuff that affects the card itself (counters, suspected, maybe unearth)

it's also for preventing problems on LKI copies, where i would have needed to copy the effects too

forge-game/src/main/java/forge/game/staticability/StaticAbility.java
tool4ever: If you don't want to touch suspected here this should just be merged into keywordCounterStatic branch only

Hanmac: i will touch suspected later, but first i wanted to merge this one as startup
(this is only a small MR for later changes)

forge-game/src/main/java/forge/game/staticability/StaticAbilityContinuous.java
```diff
@@ -93,18 +92,18 @@ public static CardCollectionView applyContinuousAbility(final StaticAbility stAb
         final List<Player> affectedPlayers = StaticAbilityContinuous.getAffectedPlayers(stAb);
         final Game game = hostCard.getGame();
 
-        final StaticEffects effects = game.getStaticEffects();
```
tool4ever, line 96: this actually looks like a functional change and against the rules:
> 613.6. If an effect should be applied in different layers and/or sublayers, the parts of the effect each apply in their appropriate ones. If an effect starts to apply in one layer and/or sublayer, it will continue to be applied to the same set of objects in each other applicable layer and/or sublayer, even if the ability generating the effect is removed during this process.

if you don't create the StaticEffect at the first affected layer you risk others changing the objects so that it suddenly applies to them this way

Hanmac: Isn't that `starts to apply` means that the effect would do something?

This code change also is not for the layers, but if the StaticAbility would affect nothing (affectedCards is empty),
then no StaticEffect object is created. (i wanted to try to clean up this more, but more for later)

if the static ability would have previous affected cards, then affectedCards isn't empty

tool4ever: ugh then its judge territory and needs more research first

right now I'd bet that "starts to apply" just means it had to be calculated even if the end result is doing nothing

your change would effectively mean that a static would only apply partially to some objects and I've **never** seen an example for that, e.g.:

> All blue creatures are Wizards and get +1/+1.

on Layer 4 you control no ~~Wizards~~ blue creatures
> All creatures are blue.

Layer 5
> Layer 7

now suddenly all creatures get +1/+1 💣

the rule is there to prevent such messy interactions

Hanmac: @tool4ever wouldn't work anyway because:
> 613.1d Layer 4: Type-changing effects are applied. These include effects that change an object’s card type, subtype, and/or supertype.
> 613.1e Layer 5: Color-changing effects are applied.

meaning turning something blue wouldn't make them apply for the type change effect anyway

tool4ever: I know, only for the +1/+1
that's the functional change you are making!

Hanmac: i will revert that part for now and consult with Rules Judges.

As for Static Abilities, i will try to do some `grep` later, but as far as i see, they can only check for something what already applies in their layer.


>  Kormus Bell `All Swamps are 1/1 black creatures that are still lands.` can only check for Type changes.

Applies in Layer 4, 5 and 7b.
So their check can only look for stuff that is calculated in Layer 4 (the Swamp type via dependency)


> Dralnu's Crusade `All Goblins are black and are Zombies in addition to their other creature types.`

Applies in layer 4 and 5. (the Layer7c ability is extra static)
So their check can only look for stuff that is calculated in Layer 4 (Goblin via dependency)

tool4ever: dependencies and the rule I quoted are different things 🤷‍♂️

Hanmac: @tool4ever and like i said, a static ability cant look for a value that isn't calculated yet (your example with turning blue creatures into wizards)

tool4ever: hmn, I think you might have missed my point?

- with your change:
all creatures could get +1/+1
- without:
only creatures that are already blue when that static in Layer 4 applies get +1/+1

buuut...
I realize now the code in `checkStaticAbilities` stores them via `affectedPerAbility` anyway
sorry, somehow I thought I refactored it so it would reuse the values stored in `StaticEffect` instead

which should mean there is no functional change either way 🤔

Hanmac:
> I still don't like adding new code without using it but it's small enough I guess

i will use it as soon as i get home

PR #9162, opened 2025-11-14
Update planetarium_of_wan_shi_tong.txt

Fulgur14:
I realized that it does need ChangesZone to handle the case where it's blinked or something.


Hanmac (approved):

tool4ever:
I'd prefer it cleans itself up when it already leaves the battlefield

Fulgur14:
Hm, couldn't that cause problems? If you get two triggers on stack at the same time, resolve one, and then get rid of Planetarium, the other trigger wouldn't remember that it was used.

The current implementation still has corner cases like that, but harder to achieve (blinking Planetarium with trigger on the stack would reset it).

tool4ever:
no, that's not how it works under the hood
it would only trigger with "new" LKI host anyway so the only ingame card it can cleanup is the new object in the zone it moved to

PR #9168, opened 2025-11-15
Airbend spells

tehdiplomat:


forge-game/src/main/java/forge/game/ability/effects/AirbendEffect.java
```diff
@@ -55,21 +56,27 @@ public void resolve(SpellAbility sa) {
...
-            if (!gameCard.canExiledBy(sa, true)) {
-                continue;
-            }
             handleExiledWith(gameCard, sa);
 
             Map<AbilityKey, Object> moveParams = AbilityKey.newMap();
             AbilityKey.addCardZoneTableParams(moveParams, triggerList);
 
-            Card movedCard = game.getAction().exile(gameCard, sa, moveParams);
+            SpellAbilityStackInstance si = null;
```
tool4ever, line 64: this doesn't work because it's targeting a spell, not a card

tehdiplomat: This is exactly how we have Venser Spell Savant coded?

tool4ever: nope, like I said spell targets are obtained from `getTargetSpells`

tehdiplomat: Yes. But what you are saying isn't actually true. The way I did it, is exactly whast we're doing for Venser.

tehdiplomat: The reason is because the targeting system can't work when crossing some Cards and some Spells.

tool4ever: should work now

PR #9171, opened 2025-11-16
Initial version of architecture doc

phughk:
Introduce an architecture doc that people can use to navigate the code quickly with.

It's work-in-progress, happy to merge as-is and follow up with small PRs or keep this "open" while I work on it (I don't have much spare time sadly).

Also includes changes for helping people make custom builds to play around with.

### DIFF
--- a/docs/_sidebar.md
+++ b/docs/_sidebar.md
@@ -49,6 +49,7 @@
   - [Snapshots & Releases](Development/Snapshots-and-Releases.md)
   - [Android Builds](Development/Android-Builds.md)
   - [Dev Mode](Development/DevMode.md)
+  - [Architecture](Development/Architecture.md)
   - [Ownership](Development/ownership.md)
   - [Docker Container](docker-setup.md)
 
--- a/docs/Development/Architecture.md
+++ b/docs/Development/Architecture.md
@@ -1,0 +1,182 @@
+# Architecture
+
+The following guide describes each module in the project, it's purpose, and any particular files and objects of interest.
+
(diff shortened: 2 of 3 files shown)


docs/_sidebar.md
```diff
@@ -63,4 +64,4 @@
 - [Missing Cards in Forge](Missing-Cards-in-Forge.md)
   - [Un‐cards, Playtest Cards, and Other Funny Cards](Un‐cards,-Playtest-Cards,-and-Other-Funny-Cards.md)
 
-- [Credit & Thanks](Credit-and-Thanks.md)
\ No newline at end of file
+- [Credit & Thanks](Credit-and-Thanks.md)
```
phughk, line 67: This is probably line endings change to CRLF, though I don't know. If it were introducing a new line we would see a character at the end of this line change.

tool4ever:
Thanks, sounds good

I'm not sure yet how it should connect with
https://github.com/Card-Forge/forge/blob/master/CONTRIBUTING.md
(I don't like too much duplicated information, though I guess a split between "overview" and "full details" could be useful)

phughk:
I still have this pr in mind and hope to return to it within the upcoming month or 2. I have also looked at the already existing section with similar coverage and hope to merge both. The primary goal is to give people informative pointers on where to find code and understand capabilities.

phughk:
Closing, since I don't think it's that necessary now with claude etc.
An alternative approach may be using https://openspec.dev/ , but that is up to the team.

Issue #9132, opened 2025-11-10
AI Uses Repopulate Pointlessly


aryst0krat:
**Describe the bug**
The AI will cast Repopulate even when you don't have a single creature in your graveyard.

**To Reproduce**
1. Play vs an AI with Repopulate in their deck.
2. Play until they have the ability to cast Repopulate before you've put any creatures in the graveyard.
3. AI will play Repopulate without any creatures to target.

**Expected behavior**
The AI should not use the spell without any targets. _Ideally_ it should also wait for targets with some priority, but that's more a potential strength increase to the AI's difficulty than a bug to be fixed.

**Screenshots**

<img width="2560" height="1440" alt="Screenshot of a game that shows the graveyard being empty of creatures while Repopulate is being cast" src="https://github.com/user-attachments/assets/fa7e345f-a2a8-4857-89a7-bea75c08b0d1" />

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: 2.0.07-SNAPSHOT-11.09

**Additional context**
This is especially egregious because Repopulate can be cycled if they don't have any other valid play.

PR #9181, opened 2025-11-16
Fix AI use of Repopulate

benhunter:
Fixes https://github.com/Card-Forge/forge/issues/9132


forge-gui/res/cardsfolder/r/repopulate.txt
```diff
@@ -1,7 +1,7 @@
 Name:Repopulate
 ManaCost:1 G
 Types:Instant
-A:SP$ ChangeZoneAll | Origin$ Graveyard | Destination$ Library | ValidTgts$ Player | TgtPrompt$ Select target player | ChangeType$ Creature | Shuffle$ True | SpellDescription$ Shuffle all creature cards from target player's graveyard into that player's library.
+A:SP$ ChangeZoneAll | Origin$ Graveyard | Destination$ Library | ValidTgts$ Player | TgtPrompt$ Select target player | ChangeType$ Creature | Shuffle$ True | AILogic$ ManifestCreatsFromGraveyard | SpellDescription$ Shuffle all creature cards from target player's graveyard into that player's library.
```
tool4ever: thanks, I suppose this does pass the test but just applying the logic from _Ghastly Conscription_ doesn't really make sense since the cards are pretty different in their usage (e.g. the AI might now use this to save their opponent who's getting milled out :P)

you'd have to debug `ChangeZoneAllAi` to find out where the wrong play decision comes from
since there already appears to be some general logic for Graveyard as origin in place ideally that can be extended

as coding custom AILogic$ is usually reserved for more exotic cases (also depends a bit how many closely-related cards exist)

tehdiplomat: So why did you end up using this AILogic? `ManifestCreatsFromGraveyard` I guess this at least slightly improves the play of Repopulate I wonder if it really should be using its own AILogic, something like "Replenish my library with something good. Or prevent my opponent for utilizing their graveyard as a resource"

benhunter: I think the wrong play decision comes from not having any other logic that checks whether the graveyard has creatures in it. 

Maybe that logic could go here [1] and check whether the Origin contains any ChangeType?

[1] : https://github.com/Card-Forge/forge/blob/d1f461196ab50f840aa4ec8cb09c9b7d576f936d/forge-ai/src/main/java/forge/ai/ability/ChangeZoneAllAi.java#L181

benhunter: `ManifestCreatsFromGraveyard` was the only place I found a check for creatures in graveyard - I see that it only targets opponents, whereas Repopulate could be useful on your own graveyard.

I think there's also an edge case where you may want to force a library shuffle after the top card is revealed (even when no creatures are in the graveyard).

tool4ever: right, it already uses the "ChangeType" param to find the best target before
but that check doesn't really take that into account

this leads to the question if there's a case where that behavior could be desirable 🤔
if there's no good reason for it often the answer is just that this API just wasn't fleshed out enough yet
(like how it currently ignores targeting yourself anyway)

tool4ever: oh yea using shuffle to screw up opponents primed top deck would be advanced play :)
I'm not sure we've done that before in any other effects that could but feel free to try
the main problem is the AI doesn't have that great of memory helpers to know an opponent did something like that though

benhunter:
I reverted the AILogic on Repopulate and went for a more general, but simple solution in `ChangeZoneAllAi` to check that at least 1 ChangeType is affected. The current code does not attempt to target the AI's own graveyard and doesn't address situations where the AI may not want to modify that graveyard (this is how the AI was already performing).

benhunter:
I looked into a larger refactor of `ChangeZoneAllAi` but the if logic is quite complex - I'm afraid more changes there could have significant affects on how the AI plays other cards.

A new AILogic might be a better approach to address Repopulate specifically.

forge-gui/res/cardsfolder/r/repopulate.txt
benhunter: Would you mind taking a look at the updated fix?

benhunter: Would you mind taking a look at the updated fix? @tehdiplomat

forge-ai/src/main/java/forge/ai/ability/ChangeZoneAllAi.java
```diff
@@ -177,8 +177,11 @@ protected AiAbilityDecision canPlay(Player ai, SpellAbility sa) {
                 if (oppList.isEmpty()) {
                     return new AiAbilityDecision(0, AiPlayDecision.CantPlaySa);
                 }
-                Player oppTarget = Collections.max(oppList, AiPlayerPredicates.compareByZoneValue(sa.getParam("ChangeType"), origin, sa));
-                if (!oppTarget.getCardsIn(ZoneType.Graveyard).isEmpty()) {
+                String changeType = sa.getParam("ChangeType");
+                Player oppTarget = Collections.max(oppList, AiPlayerPredicates.compareByZoneValue(changeType, origin, sa));
+                int countChangeType = AbilityUtils.filterListByType(oppTarget.getCardsIn(ZoneType.Graveyard), changeType, sa).size();
+                // Assumes the SpellAbility is only useful when 1 or more ChangeType will change zones
+                if (!oppTarget.getCardsIn(ZoneType.Graveyard).isEmpty() && countChangeType > 0) {
```
tool4ever: now the first part of the expression is kind of redundant

benhunter: Great point - I removed the redundant graveyard check.

PR #9185, opened 2025-11-17
Add basic Earthbend AI logic.

### LINKED ISSUE #9177: AI does not use earthbend.
Earthbend should target a land and force it to become a creature, but AI does not execute it.

Agetian:
Fixes #9177.
May be improved later, as it's very basic at the moment.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/GameState.java
+++ b/forge-ai/src/main/java/forge/ai/GameState.java
@@ -963,6 +963,8 @@ public abstract class GameState {
             spellDef = spellDef.substring(0, spellDef.indexOf("->")).trim();
         }
 
+        spellDef = spellDef.replace("^", ":"); // alternate marker for when : is the name of the card
+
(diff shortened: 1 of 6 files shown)


Agetian:
Btw, any ideas on what considerations would be good for the AI here? Maybe I'll improve this later

Jetz72:
> Btw, any ideas on what considerations would be good for the AI here? Maybe I'll improve this later

`getBestLandToAnimate` already de-priotitizes Creature lands and it's possible to give an already animated land extra counters with Earthbend, so you can probably ditch the non-creature part of the filter.

If we have any logic for evaluating ETB value for flickers and such, would be good to apply it here. The respawn will re-trigger them and nobody wants extra bounce-land ETBs.

If there aren't plenty of lands in play, prioritize fetch-lands like Evolving Wilds, because the respawn effect lets you double dip on those.

PR #9193, opened 2025-11-17
Add TLE Ruthless Waterbender

paulsnoops:
This one was missing


paulsnoops:
> weird, feels like I'm remembering this name from an earlier review but I guess my brain is just mixing up some of this card soup :D

This one was assigned to Sol in the projects but missing and his PR had Ruinous Waterbending instead, maybe it was this because the names are similar? :D

PR #9204, opened 2025-11-18
Update Shorikai, Genesis Engine Oracle Text

BigCrunch22:
Since the EOE change where Vehicles and Spacecrafts can be commanders, the part of Shorikai at the bottom became redundant and was removed. Update it to match its current Oracle text.

Also removed the keyword that makes it a commander since I think it's redundant with the current changes already implemented.


tool4ever (approved):
thanks

PR #9213, opened 2025-11-20
Autosell quest rewards

### LINKED ISSUE #9214: Smithing crash
**Describe the bug**
Attempting to smith any card crashes to desktop.

**To Reproduce**
Go to any capital.
Visit the smith.
Try to smith using any filters.

**Expected behavior**
No crash.

**Screenshots**

Arch Linux

**Additional context**
This is due to a missing check introduced with the counter showing owned copies. The check on line 935 of RewardActor needs to check RewardScene.Type, or at least that it isn't null.

KarlMHanson:
I'm not sure if there's a historical reason for why Auto-Sell isn't available for quest rewards, or if that's a distinction made on purpose. Regardless, the user experience of having to go into your collection to flag your rewards for sale while trying to remember what you just got is rather miserable.

This change simply enables auto-sell check boxes and button on quest reward screens.

EDIT:
~~Will conflict with #9211.~~

~~Also fixes #9214. If that fix is desired without this change, I can make a separate PR for just that.~~

~~Needs minor work to not interfere with #9211.~~

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/util/RewardActor.java
+++ b/forge-gui-mobile/src/forge/adventure/util/RewardActor.java
@@ -90,7 +90,7 @@ public class RewardActor extends Actor implements Disposable, ImageFetcher.Callb
     private boolean hover, hasbackface;
     boolean loaded = true;
     boolean alternate = false, shown = false;
-    boolean isRewardShop, showOverlay, isLoot, isQuestReward;
+    boolean isRewardShop, showOverlay, canAutoSell;
     TextraLabel overlayLabel;
     int artIndex = 1;
     String imageKey = "";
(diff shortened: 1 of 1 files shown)


KarlMHanson:
#9211 is a better fix for #9214 than the band-aid fix included here. This will need an adjustment or two once that gets merged.

forge-gui-mobile/src/forge/adventure/util/RewardActor.java
```diff
@@ -206,7 +206,7 @@ else if (Forge.getCurrentScene() instanceof UIScene) {
     }
 
     public void setAutoSell(boolean sell) {
-        if (!isLoot)
+        if (!isLoot && !isQuestReward)
```
Jetz72: Is this `isQuestReward` field ever doing anything different than `isLoot`? Could maybe just combine them into a single `canAutoSell` field and `||` the initial assignment together.

KarlMHanson: Yeah, that might be better. I'll look into it.

KarlMHanson: Done.

PR #9218, opened 2025-11-21
Cleanup wastetown magetower text

aryst0krat:
Mostly capitalizing and making the language a bit more more natural


aryst0krat:
I've noticed while making these changes that action dialogue choices are sometimes just like "Leave" but are sometimes stylized like "(Leave)"

I'd love to change all of one to the other way around, but I'd prefer to have some kind of guidance on it first.

aryst0krat:
Going to wrap a bunch of other text cleanup into the same pull request, since I've noticed many more errors

PR #9222, opened 2025-11-22
Fix flicker

Eradev:
Fix flicker when switching music and currentTrack was null. (ie. when entering town)


Jetz72 (approved):
Good catch, looks like I missed one.

PR #9224, opened 2025-11-22
PlayerControllerHuman: use translated MagicColor enum

Hanmac:
Part of #8682 and #8690 

uses Enum for HumanGui
Uses Translation Name for Colors
Uses Icon on MobileGUI

doesn't change AI parts (yet)

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/gui/ListChooser.java
+++ b/forge-gui-desktop/src/main/java/forge/gui/ListChooser.java
@@ -40,10 +40,13 @@ import javax.swing.event.ListSelectionListener;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 
+import forge.card.MagicColor;
+import forge.localinstance.skin.FSkinProp;
 import forge.toolbox.FList;
 import forge.toolbox.FMouseAdapter;
 import forge.toolbox.FOptionPane;
 import forge.toolbox.FScrollPane;
+import forge.toolbox.FSkin;
 import forge.util.Localizer;
 
 /**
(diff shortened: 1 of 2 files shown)


forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1834,11 +1833,11 @@ public List<AbilitySub> chooseModeForAbility(final SpellAbility sa, List<Ability
...
     @Override
-    public List<String> chooseColors(final String message, final SpellAbility sa, final int min, final int max,
-                                     List<String> options) {
-        options = options.stream().map(DeckRecognizer::getLocalisedMagicColorName).collect(Collectors.toList());
-        List<String> choices = getGui().getChoices(message, min, max, options);
-        return choices.stream().map(DeckRecognizer::getColorNameByLocalisedName).collect(Collectors.toList());
+    public ColorSet chooseColors(final String message, final SpellAbility sa, final int min, final int max,
+                                     ColorSet options) {
+        List<MagicColor.Color> choices = Lists.newArrayList(options.getOrderedColors());
+        choices = getGui().getChoices(message, min, max, choices);
```
tool4ever: wouldn't this still need the translation function like below too?

Hanmac:
need to find a way to make Icon on Desktop

Hanmac:
Found a way to add Icons for Desktop,
might need something to get better FontSize

tool4ever (approved):

Hanmac:
there are some eggs i want to address later.

* like `(click, then CTRL/CMD + click)` part should only be added for Desktop gui, the mobile one has checkboxes
* or the List Render on needs to differ between different type of Objects

PR #9239, opened 2025-11-24
Remove Unearth Effect from being rendered

Hanmac:


Hanmac:
i was thinking about using hidden static for this, but it is also connected to this `addLeaveBattlefieldReplacement`

PR #9244, opened 2025-11-24
Few more tweaks to Adventure audio

Jetz72:
Fixes an issue where a playlist would advance to the next track in the middle of the previous one. Apparently in some audio implementations, pausing a track won't stop it from firing its completion listener.

Fixes a boolean logic order of operations error if `setBackgroundMusic` is called on `null`.

Smooths out the music transitions when entering and leaving points of interest.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/stage/GameHUD.java
+++ b/forge-gui-mobile/src/forge/adventure/stage/GameHUD.java
@@ -30,6 +30,9 @@ import com.badlogic.gdx.utils.viewport.ScalingViewport;
 import com.github.tommyettinger.textra.TextraButton;
 import com.github.tommyettinger.textra.TextraLabel;
 import com.github.tommyettinger.textra.TypingLabel;
+
+import java.util.EnumSet;
+
 import forge.Forge;
 import forge.adventure.character.CharacterSprite;
 import forge.adventure.data.AdventureQuestData;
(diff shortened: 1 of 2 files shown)


forge-gui-mobile/src/forge/adventure/stage/GameHUD.java
```diff
@@ -515,10 +518,25 @@ public void updateBGM() {
...
+    private static final EnumSet<MusicPlaylist> PLAYLIST_OVERWORLD = EnumSet.of(MusicPlaylist.WHITE, MusicPlaylist.BLUE, MusicPlaylist.BLACK, MusicPlaylist.RED, MusicPlaylist.GREEN, MusicPlaylist.COLORLESS);
+
     void changeBGM(MusicPlaylist playlist) {
-        if (playlist != SoundSystem.instance.getCurrentPlaylist()) {
-            targetPlaylist = playlist;
+        MusicPlaylist currentPlaylist = SoundSystem.instance.getCurrentPlaylist();
+        if (playlist == currentPlaylist) {
+            return;
+        }
+        //If we're going from an interior to an exterior, skip the fade out.
```
tehdiplomat, line 528: Won't this skip exterior to interior too? Is that wanted?

Jetz72: Yeah. After playing with it for a bit I think the fade-out fade-in is too slow when doing map transitions, especially because the visuals transition much more abruptly. It only really feels right when crossing over regions in the overworld. 

Maybe if I took the time to implement proper crossfade, and if we could eliminate the main-thread hang due to the auto-save, it'd be worth considering. But that seems like a lot of trouble for a maybe.

tehdiplomat: Thats fine, this comment just doesn't match the conditional below it thats all.

Jetz72: Ah. I'll revise it if I do any further work in the area.

PR #9245, opened 2025-11-25
Add back m4a to supported audio formats.

Jetz72:
When I split the check for if an audio format was supported into a per-UI interface method, I'd left this off from mobile because it wasn't listed in lwjgl's supported formats. I assumed it was only there because it worked on Desktop. But according to user reports, it worked regardless.

### DIFF
--- a/forge-gui-android/src/forge/app/Main.java
+++ b/forge-gui-android/src/forge/app/Main.java
@@ -75,6 +75,7 @@ import java.io.*;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Date;
+import java.util.Set;
 
 public class Main extends AndroidApplication {
     private AndroidAdapter Gadapter;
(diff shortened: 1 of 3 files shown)


tool4ever:
I have to say that's something a Dev should confirm first, especially if the library itself claims otherwise.
A user might very well just have a .*mp3 renamed to the wrong extension but can't tell the difference, and that's the only reason it works...

Jetz72:
Good thinking, and an interesting development now that I've looked into it more: When running the mobile view on desktop, the supported formats are indeed limited to `.mp3`, `.wav`, and `.ogg`. In fact, it only checks by extension, so there's no way an MP3 renamed to M4A could have ever loaded.

*However*, on Android, the file is passed to the platform's MediaPlayer API without any such check, which looks like it can support [a plethora of types](https://developer.android.com/media/platform/supported-formats). I think I might route this logic over the the Device Adapter to resolve further.

Jetz72:
Alright, I threw a bunch of common audio formats that supposedly work on Android into an Android-specific implementation and put LWJGL-Desktop back the way it was. The full set of formats that actually are supported will vary between versions of Android and/or based on whatever audio codecs the device included with its build of the OS. These ones should be universal, but if it turns out one of them isn't supported, the worst that'll happen is that it doesn't play anything and logs an error.

 There are probably much better ways of testing that Android can play an audio format - reading the mime type and checking it against the `MediaCodecList` API seems promising - which would let us support basically any file the device supports. But I hit a wall getting Android to generate a debug build on my setup, and I'm not confident I can implement something like that blindly without breaking Android's audio all over again. I think this should suffice for now.

Issue #9259, opened 2025-11-28
Network + Translatable Enum = Error


Hanmac:
Color now uses TranslatedName for Display function:
https://github.com/Card-Forge/forge/blob/0ab879bac08743443bf98c07ececc61a7bc1b250/forge-gui/src/main/java/forge/player/PlayerControllerHuman.java#L1836-L1839

but that isn't serializable there:
https://github.com/Card-Forge/forge/blob/0ab879bac08743443bf98c07ececc61a7bc1b250/forge-gui/src/main/java/forge/gamemodes/net/server/NetGuiGame.java#L249-L251

This says, the Function should also be `Serializable`:
https://www.baeldung.com/java-serialize-lambda#bd-2-how-to-serialize-lambda

so the parameter can be used for network, but I'm not 100 sure that solves our problems

another solution is to try to tweak the GUI for Desktop, so `getTranslableName` isn't needed

PR #9262, opened 2025-11-29
HumanGUI check for ITranslatable without transformer function

Hanmac:
Partial fix of #9259

This bypasses the problem with Function not serializable


tool4ever (approved):
Crash fixed, I guess mobile GUI handles it differently and doesn't need to be touched here

Hanmac:
> Crash fixed, I guess mobile GUI handles it differently and doesn't need to be touched here

mobile already has an extra renderer
https://github.com/Card-Forge/forge/blob/2c33b1fa6608b8271d5fc258115948b574ad951d/forge-gui-mobile/src/forge/toolbox/FChoiceList.java#L689-L693

but thats another piece i need to clean up for later reason when i want to have different types in the same list

PR #9263, opened 2025-11-29
Fix GUI Functions for display needs to be serializable

Hanmac:
Other part of #9259

This should cause all Display Functions to be serializable

i added a new Interface because it was very ugly to mark a Type as both while extending a class

### DIFF
--- a/forge-gui/src/main/java/forge/itemmanager/AdvancedSearch.java
+++ b/forge-gui/src/main/java/forge/itemmanager/AdvancedSearch.java
@@ -1271,17 +1271,17 @@ public class AdvancedSearch {
 
     private static abstract class CustomListEvaluator<T extends InventoryItem, V> extends FilterEvaluator<T, V> {
         private final Collection<V> choices;
-        private final FSerializableFunction<V> toShortString, toLongString;
+        private final FSerializableFunction<V, String> toShortString, toLongString;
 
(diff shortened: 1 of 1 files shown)


Jetz72:
Can you actually serialize and deserialize a function this easily without any issues?

Hanmac:
@Jetz72 it seems https://www.baeldung.com/java-serialize-lambda#bd-2-how-to-serialize-lambda

i was trying to do it without extra interface, it worked for the functions, but for some classes it doesn't like it.
That's why i added that interface.

Jetz72:
The fact that it's hard-coded to return a String suggests we should probably use a less general name than "Serializable Function". `DisplayFunction`? `TextFunction`? `ListItemTextFunction`? 

Broadly speaking it feels a bit janky that we have to pass a function through such a long pipeline in the first place. Ideally the UI would be responsible for recognizing what kind of list it's been given and decide for itself how to display it, without explicit instructions from the game logic. We already do that with some object types like Cards. It'd be nice to just trim these function signatures down a bit. But that might be a lot of trouble for little practical benefit.

Hanmac:
@tool4ever your opinion about different Template parameter `<T, V>` instead, or a different name?

tool4ever:
yea either another generic param or a name like "...StringMapper"?

PR #9271, opened 2025-11-30
Update flexible_waterbender.txt

Akse1ot:
Fix script Flexible Waterbender: replace SetPower$/SetToughness$ with Power$/Toughness$ so the ability works correctly


squee1968:
Thank you for fixing my error.

tehdiplomat (approved):

PR #9280, opened 2025-12-03
Bump tag when building daily snapshots

Jetz72:
An attempt to fix the daily snapshots being buried halfway down the releases page. The release date seems to be tied to the date of the tag's commit, so I think updating it will bump the release up the list.

 I admittedly haven't tested this, since I'd have to hack up the script and plug in a lot of config data to run it in my own fork. But the commands for updating the tag seem to work when I tried them locally. Can revert and refire if the build fails.

### DIFF
--- a/.github/workflows/snapshot-both-pc-android.yml
+++ b/.github/workflows/snapshot-both-pc-android.yml
@@ -22,7 +22,7 @@ jobs:
       packages: write
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
       - name: Set up JDK 17
         uses: actions/setup-java@v3
         with:
@@ -111,6 +111,13 @@ jobs:
           ls
         env:
           GITHUB_TOKEN: ${{ github.token }}
+
+      - name: Bump Snapshot Tag
+        run: |
+          git fetch origin tag "daily-snapshots"
+          git fetch origin master
+          git tag -f daily-snapshots origin/master
+          git push origin -f daily-snapshots
 
       - name: Upload snapshot to GitHub Prerelease
         uses: ncipollo/release-action@v1


Hanmac:
See also https://github.com/ncipollo/release-action/issues/440 

also this comment to better delete the snapshot release:
https://github.com/ncipollo/release-action/issues/440#issuecomment-2408924270

tehdiplomat:
Yea at the time i just wanted things working, but had planned to come back to deleting/recreating. It ends up creating 10 minute windows where no snapshots are available (although that might happen anyway?). The ftp uploader might fail in cases where we delete and recreate since it expects to find the old package to delete during the upload process. Happy to try this and go the other route if its not working.

Hanmac:
i don't know if we can use gh in the action for this:
`gh release delete your-release-name --cleanup-tag --yes`

Jetz72:
I figured deleting the tag and release completely and then recreating them would be the most surefire way. I was interested in seeing if something less drastic would suffice, though. The [github docs](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release) mention:
> The `created_at` attribute [of a release] is the date of the commit used for the release, and not the date when the release was drafted or published.

Doesn't sound like anyone in that comment thread tried just moving the tag, and I struggled to find a definitive yay or nay elsewhere on whether this would work, but it sounds like it should.

Jetz72:
Documenting outcome for future historians, web-crawlers harvesting LLM training data, and so I'm not a hypocrite after I just mentioned how hard it was to find a definitive answer on whether or not this works: Adjusting the tag to point to the latest master branch commit before updating the release assets *does* move the release to the top of the releases page, but *does not* update the date displayed on the release. If not tagged as a prerelease, I suspect it'd also place it as the latest release on the repo homepage.

PR #9282, opened 2025-12-03
Add option to mass-select targets for proliferate

TrevorHayes:
I have been playing a proliferate deck which was very tedious since the game asks you to select many targets every time you proliferate (and you're nearly always simply selecting all your cards). This update changes the proliferate "select targets" prompt to replace the "cancel" button (which is unnecessary since you can simply not choose any targets) with a button that cycles between selecting all your own cards, selecting all valid targets (your cards plus all opponents cards plus all players), and deselecting all targets.
The mass-select button is triggered by a flag so it could easily be added to other effects.
I have played many commander and adventure battles testing this button.

### DIFF
--- a/forge-game/src/main/java/forge/game/spellability/SpellAbility.java
+++ b/forge-game/src/main/java/forge/game/spellability/SpellAbility.java
@@ -67,8 +67,6 @@ import forge.game.trigger.TriggerType;
 import forge.game.trigger.WrappedAbility;
 import forge.game.zone.ZoneType;
 
-//only SpellAbility can go on the stack
-//override any methods as needed
 /**
  * <p>
  * Abstract SpellAbility class.
@@ -92,7 +90,6 @@ public abstract class SpellAbility extends CardTraitBase implements ISpellAbilit
 
     private int id;
 
-    // choices for constructor isPermanent argument
     private String originalDescription = "", description = "";
     private String originalStackDescription = "", stackDescription = "";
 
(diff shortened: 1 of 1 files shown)


forge-gui/res/cardsfolder/r/resourceful_defense.txt
TrevorHayes, line 1: This isn't relevant to the UI changes; it's just a card I was playing with and it's very confusing since it asks you which cards to move counters to twice so I updated the prompt text. Maybe there's a better way to do that; I don't know card scripting.

forge-game/src/main/java/forge/game/spellability/SpellAbility.java
```diff
@@ -150,6 +150,7 @@ public static class EmptySa extends SpellAbility {
 
     private boolean isCopied = false;
     private boolean mayChooseNewTargets = false;
+    private boolean useMassSelect = false; // Should the UI show "select mine/select all" function
```
tool4ever: nice idea
can you just replace this logic with an ApiType check in the input class instead?
we try to keep the rules engine clean of such GUI stuff

TrevorHayes: Makes sense. I have update the logic as you suggest.

Hanmac:
i merge this for now, and do some more cosmetic changes later (like using Translatable interface)

PR #9283, opened 2025-12-04
CoreType: add icons and make translateable

### LINKED ISSUE #9236: Translate CoreType and SuperType with Icons
* make both CoreType and SuperType Translatable Enums
* And Icons to them if able (Creature has Icon, while Kindred has not)

The AdvancedSearch Filter uses `CardType.getCombinedSuperAndCoreTypes`, split it into two Filter for better Translation
and remove `CardType.getCombinedSuperAndCoreTypes` because it wouldn't be used anymore.

As for the Icons:

https://github.com/Card-Forge/forge/blob/4e08529c2449d8a9279808c4bf80130491f453e1/forge-gui/src/main/java/forge/itemmanager/SItemManagerUtil.java#L42-L49

* this already has some icons for some types, but the labels are translated in plural, not singular as I want. Move Icon function into something like this: https://github.com/Card-Forge/forge/blob/4e08529c2449d8a9279808c4bf80130491f453e1/forge-gui/src/main/java/forge/localinstance/skin/FSkinProp.java#L711
* The StatTypes should be split apart like the ConquestFilter

Hanmac:
Closes #9236 

i still need to add SuperTypes back again

also some Types doesn't have an Icon, so i either need Icon for them, or get a better empty Icon

### DIFF
--- a/forge-core/src/main/java/forge/card/CardType.java
+++ b/forge-core/src/main/java/forge/card/CardType.java
@@ -108,14 +108,16 @@ public final class CardType implements Comparable<CardType>, CardTypeView {
         }
     }
 
-    public enum Supertype {
-        Basic,
-        Elite,
-        Host,
-        Legendary,
-        Snow,
-        Ongoing,
-        World;
+    public enum Supertype implements ITranslatable {
+        Basic("lblBasic"),
+        Elite("lblElite"),
+        Host("lblHost"),
+        Legendary("lblLegendary"),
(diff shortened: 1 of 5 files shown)


Hanmac:
@kevlahnota I would like to have an "empty" icon, so in the ChoiceList, there is an empty place

Hanmac:
There might be more icons added later

Also, the Mobile Item Renderer needs a clean-up

PR #9291, opened 2025-12-05
Update Oracle Text and Descriptions: Letter A

BigCrunch22:
Update the Oracle text and descriptions of cards that start with the letter A to match their current given text.

I was going to do a more in-depth edit of the descriptions, but that will take much longer since I'm doing this manually. I might make a draft of it in the future as well as continue with the other letters.

I was doing my best to not mess up the descriptions since I was skimming it instead of seeing it in-game, but I can't guarantee 100% that nothing potentially got messed up.

Ignore the weird things below; I messed up my copy.


Jetz72:
@Hanmac Worth considering #7401 with regard to this? Think that being on the table is why we haven't yet gone through with this cleanup.

Hanmac:
> @Hanmac Worth considering #7401 with regard to this? Think that being on the table is why we haven't yet gone through with this cleanup.

I will probably do soon, but not thus weekend

Issue #8996, opened 2025-10-25
Attack multiple opponents


CTimmerman:
The latest Git version overkills one opponent instead of multiple. Also the triggers are a bit slow to resolve despite the delay being 0.001 and the process not being debugged in this case.

<img width="1680" height="1050" alt="Image" src="https://github.com/user-attachments/assets/192518d4-0143-4f42-ad89-767aeaad50ac" />

CTimmerman:
AI still does not attack multiple opponents and instead often dies after swinging everything at just one.

PR #9308, opened 2025-12-10
Don't overextend attack

CTimmerman:
The first step to attacking multiple opponents (#8996) is not overextending on one. See this patch saving player one:

image

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -38,6 +38,7 @@ import forge.game.spellability.SpellAbilityPredicates;
 import forge.game.staticability.StaticAbility;
 import forge.game.staticability.StaticAbilityAssignCombatDamageAsUnblocked;
 import forge.game.staticability.StaticAbilityMode;
+import forge.game.staticability.StaticAbilityMustAttack;
 import forge.game.trigger.Trigger;
 import forge.game.trigger.TriggerType;
 import forge.game.zone.ZoneType;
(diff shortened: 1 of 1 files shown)


forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -820,6 +820,26 @@ public final int declareAttackers(final Combat combat) {
 
         final boolean bAssault = doAssault();
 
+        // Don't overkill too much / keep blockers for multiplayer or unexpected fog.
```
tool4ever, line 829: this is certainly a good side goal for attack declaration
- but the logic in `doAssault` is way too complex to just plainly remove attackers in response
- among other things you're now ignoring alternative win conditions and combat legality from attack requirements
- the TODO in L991 already suggests some of these factors

CTimmerman:
Now checking isGoaded and entitiesMustAttack (which should be enough on its own imo):
image

CTimmerman:
The issue this patch fixes is about using too many attackers. It appears that the current doAssault adds all of them.

As commented, the current patch checks for needing to attack (using isGoaded and static check, but not the singular rememberedtoattack check in the existing code; i'd make shouldAttack a single function checking all potential causes) and leaves 2 extra attackers beyond the 20 life overkill (more than enough for shared infect, but an edge case would be a bunch of mites sorted last. To avoid edge case code i'd just Monte Carlo or neural net the options.).

A smaller team sounds unable to swing for the win so should not return true in doAssault, unless there's a big enough trampler in there, which should make the rest not needed for the attack. I'll see about modifying doAssault and leave the messy history.

PR #9315, opened 2025-12-11
Mana: turn into a record

Hanmac:
part of #8682 

Mana class was already unchangeable, so turning it into a record doesn't hurt

add MagicColor.Color.fromName and uses match pattern for some of the selects

### DIFF
--- a/forge-core/src/main/java/forge/card/MagicColor.java
+++ b/forge-core/src/main/java/forge/card/MagicColor.java
@@ -88,29 +88,15 @@ public final class MagicColor {
     }
 
     public static String toShortString(final byte color) {
-        return switch (color) {
-            case WHITE -> "W";
-            case BLUE -> "U";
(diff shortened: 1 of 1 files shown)


forge-core/src/main/java/forge/card/MagicColor.java
```diff
@@ -88,25 +88,25 @@ public static String toShortString(final String color) {
...
     public static String toShortString(final byte color) {
-        switch (color) {
-            case WHITE: return "W";
-            case BLUE:  return "U";
-            case BLACK: return "B";
-            case RED:   return "R";
-            case GREEN: return "G";
-            default:    return "C";
-        }
+        return switch (color) {
```
tool4ever: is this better?
`return Color.fromByte(color).getShortName();`

tool4ever (approved):

PR #9320, opened 2025-12-12
ECL: add Blight Effect

### LINKED ISSUE #9318: Blight Effect
New Effect:
> each opponent blights 1. (They each put a -1/-1 counter on a creature they control.)

Points:
* DefinedPlayer
* Amount Num
* They choose A creature they control, and THEY put N -1/-1 counters on it.
* This inside a Playerloop
* And at the End of this, the Counters are placed the cards at once.

Extra:
+ maybe some AI logic for choosing creatures?

Hanmac:
Closes #9318


Agetian:
Probably needs at least a basic form of AI that will target the worst stuff available, otherwise looks good to me :) might be able to look into the AI slightly later, not sure if I can pencil it in before Christmas :(

Hanmac:
@tool4ever @Agetian i added minimal BlightAi support for later

PR #9321, opened 2025-12-12
Fixes multiple allied lands in a single booster pack in Avatar booster draft

solstheman:


forge-gui/res/editions/Avatar The Last Airbender.txt
```diff
@@ -11,7 +11,7 @@ Prerelease=6 Boosters, 1 RareMythic+
 BoosterBox=30
 
 [Common]
-Base=Common:fromSheet("TLA cards")
+Base=Common:!fromSheet("TLA allied lands"):fromSheet("TLA cards")
```
tehdiplomat, line 14: Hmm.. this would work, but i think the allied lands are supposed to be L rarity.

solstheman: I think L is reserved for basic lands (to differentiate between a rare land and a common land). See Kaldheim's snow lands, for example.

solstheman: Or more recently, Final Fantasy does it this way.

Jetz72: https://gatherer.wizards.com/TLA/en-us/267/boiling-rock-prison
> Rarity: Land

https://gatherer.wizards.com/TLA/en-us/277/rumble-arena
> Rarity: Common

Expecting any long-term consistency with regard to Magic sets will quickly drive you to insanity.

solstheman: Got it, by "FF does it this way" I meant both sets filter their non-basic common lands the way I do it in this PR.

tehdiplomat: Yes I think that's set incorrect too. I'll merge for now, but we should probably fix it without needing it.

Hanmac:
@tehdiplomat i fixed the Rarity of these lands to "L"
so this change can probably be reverted

PR #9323, opened 2025-12-12
Keep track of perpetual mana cost changes for display

TrevorHayes:
Currently perpetual changes to P/T are reflected in hand but mana cost changes are not. This change tracks perpetual mana cost reductions so they can be displayed in hand (and other zones).

Normally this just means reducing (or increasing) the generic mana cost value, but for spells with X in the cost decreasing the cost contributes to "X" so for those spells I have added the generic reduction as an additional mana symbol after the regular mana costs but with a yellow highlight. For example, a Fireball spell reduced by 2 mana would display as XR2 where the 2 is outlined in yellow. This should be fairly intuitive since the player sees it appear when they reduce the cost. I experimented with displaying it as XR-2 (which is what it says in the card text box) but that takes up another symbol's worth of room and looks funny. I tried adding "-" to the generic cost symbol but that was hard to read.

I have tested decreasing costs and increasing costs with regular spells, spells with both X and regular generic costs like X2B (Ingenious Mastery), spells that set the mana cost to a specific value (thought partition; this already worked but I didn't break it),  and spells that have changed zones before and after cost reductions.
fireball_cost_reduct

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardState.java
+++ b/forge-game/src/main/java/forge/game/card/CardState.java
@@ -255,6 +255,13 @@ public class CardState implements GameObject, IHasSVars, ITranslatable {
 
         // This doesn't work on hybrid generic costs
         int newGeneric = manaCost.getGenericCost() - genericCostAdjustment;
+        // Apply negative cost adjustments to cards with an X cost only.
+        // While this could be done for regular cards as well (to potentially offset other cost-increasing effects)
+        // it would rarely matter and would require changing the "no_cost" flag in ManaCost from -1
+        // otherwise a -1 generic adjustment will get interpreted as "no_cost"
+        if (manaCost.getShardCount(ManaCostShard.X) == 0) {
+            newGeneric = Math.max(0, newGeneric);
+        }
 
         // Replace the original generic mana cost with the adjusted value
         perpetualAdjustedManaCost = new ManaCost(


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -437,6 +437,7 @@ public void updateAbilityTextForView() {
     }
 
     public void updateManaCostForView() {
+        currentState.setManaCost(this.getManaCost());
```
Hanmac: I'm pretty sure that breaks perpetual effects

TrevorHayes: I haven't found any issues with it but that code is only called by "addChangedManaCost" and "removeChangedManaCost". I added this line for Thought Partition (not sure what other cards directly change the mana cost; maybe there's other cases I missed?)
I did test that "decrease mana cost by 1" and "set mana cost to 5" (applied in either order) resulted in a displayed mana cost of 4 which matches what the actual cast cost ends up being when you cast it.

Hanmac: It messes with `Card.getOriginalManaCost` and in result, messes with `Card.getManaCost`

TrevorHayes: Yes; I'm not sure why "getOriginalManaCost" uses CardState at all. Shouldn't "original" be a static value of the card itself? I haven't found any cases where this code doesn't result in the correct values but I see your point - I put that line in to fix constant-value perpetual effects but I think what you're saying is that the perpetual calculation routine should be checking the changedCardManaCost tree rather than updating the state manaCost value. I'll work on that.

TrevorHayes: I have updated the code to not set the manaCost value here. Thanks!

TrevorHayes: Anything else I can do here?

tehdiplomat:
I Don't have any major problems with this, but i think it should still be an option for people to disable if they don't want it:

image

forge-core/src/main/java/forge/card/mana/ManaCost.java
```diff
@@ -91,14 +91,18 @@ private void sealClass(List<ManaCostShard> shards0) {
      */
     public ManaCost(final IParserManaCost parser) {
         final List<ManaCostShard> shardsTemp = Lists.newArrayList();
+        boolean xMana = false;
         while (parser.hasNext()) {
             final ManaCostShard shard = parser.next();
             if (shard != null && shard != ManaCostShard.GENERIC) {
+                if (shard.toString().equals("{X}")) {
```
tool4ever: this change shouldn't be needed imo

forge-game/src/main/java/forge/game/card/CardState.java
```diff
@@ -207,9 +213,81 @@ public final ManaCost getManaCost() {
...
+            return;
+        }
+        ManaCost manaCost = getCard().getManaCost();
+        // I don't know why refetching the card data like in this next line is necessary but in some cases (when the cost reduction
+        // wasn't added when the card was in the current zone, I think) the static abilities are missing. There's probably
+        // a better way to do this.
+        for (final StaticAbility stAb : getCard().getGame().getCardsInGame().get(getCard()).getStaticAbilities()) {
+//        for (final StaticAbility stAb : getStaticAbilities()) { // For some reason the current state sometimes doesn't have static abilities
+            // Only collect perpetual cost changes
+            if ("Card.Self".equals(stAb.getParam("ValidCard")) && "DBCleanup".equals(stAb.getParam("SubAbility"))) {
```
tool4ever: statics don't even have subabilities

forge-game/src/main/java/forge/game/card/CardState.java
```diff
@@ -207,9 +213,81 @@ public final ManaCost getManaCost() {
...
+        // Separate abilities to apply them in proper order
+        if (getCard() == null || getCard().getGame() == null) {
+            setPerpetualAdjustedManaCost(manaCost);
+            return;
+        }
+        ManaCost manaCost = getCard().getManaCost();
+        // I don't know why refetching the card data like in this next line is necessary but in some cases (when the cost reduction
+        // wasn't added when the card was in the current zone, I think) the static abilities are missing. There's probably
+        // a better way to do this.
+        for (final StaticAbility stAb : getCard().getGame().getCardsInGame().get(getCard()).getStaticAbilities()) {
```
tool4ever: there is `Game.getCardState`
though it might not be a bug if it's called from LKI or something

TrevorHayes:
> I Don't have any major problems with this, but i think it should still be an option for people to disable if they don't want it:

Ah, I like that as a place to put it rather than the general settings. I'll have to track down the other perpetual changes that are already displayed and change that code as well, I suppose.

forge-core/src/main/java/forge/card/mana/ManaCost.java
TrevorHayes: Not sure what you mean; this is there so that X spells that have perpetual cost reductions can be displayed with them (as fireball example). Is there a better way to detect that a spell has an X in the cost?

forge-game/src/main/java/forge/game/card/CardState.java
TrevorHayes: Not sure what you mean; both of these checks were required to avoid crashing on calls on some cards. Yes, some don't have the "SubAbility" parameter but the way the check is written, returning null is fine and evaluates to false.

tool4ever: ok, let me phrase it this way - your whole PR works for exactly one card:
https://github.com/search?q=repo%3ACard-Forge%2Fforge+%2Fmode.*cost.*card.self.*dbcleanup%2F&type=code
and the only reason it does is because that's a scripting error

forge-game/src/main/java/forge/game/card/CardState.java
TrevorHayes: OK; I switched to using that but I still don't know why it's needed at all.

forge-game/src/main/java/forge/game/card/CardState.java
TrevorHayes: Well; that's embarrassing. I had checked other cost change abilities but then didn't retest them after noticing this one thing that seemed to separate the ability I was mostly testing with from other effects. I notice "Lumbering Lightshield" also has this sub-ability, but that's a cost-raising card. Should that sub-ability be removed from those two cards?
This logic error has been addressed: I couldn't find anything to differentiate these types of cost reductions so now it just relies on ParseInt throwing NFE for non-numeric "amount" values.

TrevorHayes:
> I Don't have any major problems with this, but i think it should still be an option for people to disable if they don't want it:

I have added the suggested menu option as "Perpetual Changes to Mana Cost" and verified that it works. However, I have not added translations to the localization files as I'm not confident Google Translate would use the correct terms. What's usually done for localization? I don't know how much this matters; I notice lots of labels that don't have localization and fall back to the English strings.

forge-game/src/main/java/forge/game/card/CardState.java
tool4ever: yea feel free to clean up the script
if you look closely you'll see that _Lumbering Lightshield_ is different though

forge-game/src/main/java/forge/game/card/CardState.java
```diff
@@ -210,9 +215,90 @@ public final ManaCost getManaCost() {
...
+        final List<StaticAbility> reduceAbilities = Lists.newArrayList();
+        // Separate abilities to apply them in proper order
+        if (getCard() == null || getCard().getGame() == null) {
+            setPerpetualAdjustedManaCost(manaCost);
+            return;
+        }
+        ManaCost manaCost = getManaCost();
+        // I don't know why refetching the card data like in this next line is necessary but in some cases (e.g. when
+        // the cost reduction wasn't added when the card was in the current zone) the static abilities are missing.
+        for (final StaticAbility stAb : getCard().getGame().getCardState(getCard()).getStaticAbilities()) {
```
tool4ever: the engine really shouldn't perform such expensive operations when the chance it's irrelevant is this high

forge-game/src/main/java/forge/game/card/CardState.java
TrevorHayes: I guess because the DBCleanup is on the DBAnimate ability not on the static it triggers.

forge-game/src/main/java/forge/game/card/CardState.java
TrevorHayes: True; and it's not necessary. I have removed the call from CardState.setManaCost() so it only triggers when a card has a perpetual mana cost effect applied.

forge-game/src/main/java/forge/game/card/CardState.java
tool4ever: yea that's how you chain abilities

forge-game/src/main/java/forge/game/card/CardState.java
tool4ever: I meant more if the GUI even needs it or at the very least checking for `hasPerpetual()`

TrevorHayes: This is only called by Card.calculatePerpetualAdjustedManaCost() which itself is only called by AnimateEffectBase.doAnimate() at the moment a card has a perpetual effect that raises or lowers the cost added to it (AnimateEffectBase lines 229-234).

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
```diff
@@ -225,6 +228,9 @@ public String getDescription() {
                 addedStaticAbilities, removeAbilities, timestamp, 0);
             if (perpetual) {
                 c.addPerpetual(new PerpetualAbilities(timestamp, changes));
+                if (costChange) {
+                    c.calculatePerpetualAdjustedManaCost();
```
tool4ever, line 228: try moving the logic to `PerpetualAbilities.applyEffect`
otherwise I don't see how it'll get refreshed when the card moves zones

forge-game/src/main/java/forge/game/card/CardState.java
```diff
@@ -213,6 +218,87 @@ public final void setManaCost(final ManaCost manaCost0) {
...
+    /**
+     * Calculate and save the value of the mana cost adjusted by any perpetual raise/lower cost
+     * effects for display.
+     */
+    public void calculatePerpetualAdjustedManaCost() {
+        final List<StaticAbility> raiseAbilities = Lists.newArrayList();
+        final List<StaticAbility> reduceAbilities = Lists.newArrayList();
+        // Separate abilities to apply them in proper order
+        if (getCard() == null || getCard().getGame() == null) {
+            setPerpetualAdjustedManaCost(manaCost);
```
tool4ever: might as well just keep it null here?
then you don't even need the helper

forge-game/src/main/java/forge/game/card/CardState.java
```diff
@@ -213,6 +218,87 @@ public final void setManaCost(final ManaCost manaCost0) {
...
+            int genericCostAdjustment;
+            int remainingGenericCostAdjustment;
+            if (genericCost + totalGenericCostAdjustment < 0) {
+                genericCostAdjustment = -genericCost;
+                remainingGenericCostAdjustment = totalGenericCostAdjustment + genericCost;
+            } else {
+                genericCostAdjustment = totalGenericCostAdjustment;
+                remainingGenericCostAdjustment = 0;
+            }
+            if (genericCostAdjustment != 0) {
```
tool4ever: imo this math block should be cleaned up a bit
certainly parsing mana string twice can be avoided

forge-core/src/main/java/forge/card/mana/ManaCost.java
tool4ever: hmn ok
but just check for `ManaCostShard.X`?

forge-game/src/main/java/forge/game/card/CardState.java
tool4ever: I'll try to debug after the rest is sorted out
we're getting there :)

forge-core/src/main/java/forge/card/mana/ManaCost.java
TrevorHayes: Ah, yes. I will change that.

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
TrevorHayes: Good catch; that had been taken care of when the calculation was done whenever a card was created but now that it's not, you're right that it has stopped working. I miss unit tests... but I don't have any idea how to make unit tests for this stuff.
I can't just move the logic because then it wouldn't update the cost when a card changes without changing zones but I have moved the logic to CardTraitChanges and then call it from both AnimaeEffectBase and PerpetualAbilities.

forge-game/src/main/java/forge/game/card/CardState.java
TrevorHayes: Changed.

forge-game/src/main/java/forge/game/card/CardState.java
TrevorHayes: I have added comments and fixed it so that the double parsing is avoided but don't see any obvious way to remove other calculations.

tool4ever: well I have saved 34 lines for you lol

in addition I'm only using one loop instead of three + bailing out directly if there is no generic/X shard to reduce

I hope you agree it's much easier to read now...

forge-game/src/main/java/forge/game/card/CardState.java
tool4ever: couldn't reproduce
I suspect my LKI guess was right?

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
Hanmac: i will need to refactor it in #9680 anyway later

tehdiplomat:
@TrevorHayes Looks like your translations are broken:

```
java.util.MissingResourceException: Can't find resource for bundle java.util.PropertyResourceBundle, key lblShowCardPerpetualManaCostOverlays
	at java.base/java.util.ResourceBundle.getObject(ResourceBundle.java:564)
	at java.base/java.util.ResourceBundle.getString(ResourceBundle.java:521)
	at forge.util.Localizer.getMessage(Localizer.java:87)
	at forge.util.Localizer.getMessage(Localizer.java:80)
	at forge.screens.settings.SettingsPage.<init>(SettingsPage.java:597)
	at forge.screens.settings.SettingsScreen.<init>(SettingsScreen.java:54)
	at forge.screens.settings.SettingsScreen.show(SettingsScreen.java:26)
	at forge.screens.home.HomeScreen.lambda$new$5(HomeScreen.java:112)
	at forge.toolbox.FButton.tap(FButton.java:236)
	at forge.Forge$MainInputProcessor.tap(Forge.java:1366)
	at forge.toolbox.FGestureAdapter.touchUp(FGestureAdapter.java:245)
	at forge.toolbox.FGestureAdapter.touchUp(FGestureAdapter.java:177)
	at com.badlogic.gdx.InputEventQueue.drain(InputEventQueue.java:70)
	at com.badlogic.gdx.backends.lwjgl3.DefaultLwjgl3Input.update(DefaultLwjgl3Input.java:190)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Window.update(Lwjgl3Window.java:414)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:181)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:153)
	at forge.app.GameLauncher.<init>(GameLauncher.java:139)
	at forge.app.Main.main(Main.java:30)
INVALID PROPERTY: 'lblShowCardPerpetualManaCostOverlays' -- Translation missing from en_US
java.util.MissingResourceException: Can't find resource for bundle java.util.PropertyResourceBundle, key lblShowCardPerpetualManaCostOverlays
	at java.base/java.util.ResourceBundle.getObject(ResourceBundle.java:564)
	at java.base/java.util.ResourceBundle.getString(ResourceBundle.java:521)
	at forge.util.Localizer.getMessage(Localizer.java:102)
	at forge.util.Localizer.getMessage(Localizer.java:80)
	at forge.screens.settings.SettingsPage.<init>(SettingsPage.java:597)
	at forge.screens.settings.SettingsScreen.<init>(SettingsScreen.java:54)
	at forge.screens.settings.SettingsScreen.show(SettingsScreen.java:26)
	at forge.screens.home.HomeScreen.lambda$new$5(HomeScreen.java:112)
	at forge.toolbox.FButton.tap(FButton.java:236)
	at forge.Forge$MainInputProcessor.tap(Forge.java:1366)
	at forge.toolbox.FGestureAdapter.touchUp(FGestureAdapter.java:245)
	at forge.toolbox.FGestureAdapter.touchUp(FGestureAdapter.java:177)
	at com.badlogic.gdx.InputEventQueue.drain(InputEventQueue.java:70)
	at com.badlogic.gdx.backends.lwjgl3.DefaultLwjgl3Input.update(DefaultLwjgl3Input.java:190)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Window.update(Lwjgl3Window.java:414)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:181)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:153)
	at forge.app.GameLauncher.<init>(GameLauncher.java:139)
	at forge.app.Main.main(Main.java:30)
java.util.MissingResourceException: Can't find resource for bundle java.util.PropertyResourceBundle, key nlShowCardPerpetualManaCostOverlays
	at java.base/java.util.ResourceBundle.getObject(ResourceBundle.java:564)
	at java.base/java.util.ResourceBundle.getString(ResourceBundle.java:521)
	at forge.util.Localizer.getMessage(Localizer.java:87)
	at forge.util.Localizer.getMessage(Localizer.java:80)
	at forge.screens.settings.SettingsPage.<init>(SettingsPage.java:598)
	at forge.screens.settings.SettingsScreen.<init>(SettingsScreen.java:54)
	at forge.screens.settings.SettingsScreen.show(SettingsScreen.java:26)
	at forge.screens.home.HomeScreen.lambda$new$5(HomeScreen.java:112)
	at forge.toolbox.FButton.tap(FButton.java:236)
	at forge.Forge$MainInputProcessor.tap(Forge.java:1366)
	at forge.toolbox.FGestureAdapter.touchUp(FGestureAdapter.java:245)
	at forge.toolbox.FGestureAdapter.touchUp(FGestureAdapter.java:177)
	at com.badlogic.gdx.InputEventQueue.drain(InputEventQueue.java:70)
	at com.badlogic.gdx.backends.lwjgl3.DefaultLwjgl3Input.update(DefaultLwjgl3Input.java:190)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Window.update(Lwjgl3Window.java:414)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:181)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:153)
	at forge.app.GameLauncher.<init>(GameLauncher.java:139)
	at forge.app.Main.main(Main.java:30)
INVALID PROPERTY: 'nlShowCardPerpetualManaCostOverlays' -- Translation missing from en_US
java.util.MissingResourceException: Can't find resource for bundle java.util.PropertyResourceBundle, key nlShowCardPerpetualManaCostOverlays
	at java.base/java.util.ResourceBundle.getObject(ResourceBundle.java:564)
	at java.base/java.util.ResourceBundle.getString(ResourceBundle.java:521)
	at forge.util.Localizer.getMessage(Localizer.java:102)
	at forge.util.Localizer.getMessage(Localizer.java:80)
	at forge.screens.settings.SettingsPage.<init>(SettingsPage.java:598)
	at forge.screens.settings.SettingsScreen.<init>(SettingsScreen.java:54)
	at forge.screens.settings.SettingsScreen.show(SettingsScreen.java:26)
	at forge.screens.home.HomeScreen.lambda$new$5(HomeScreen.java:112)
	at forge.toolbox.FButton.tap(FButton.java:236)
	at forge.Forge$MainInputProcessor.tap(Forge.java:1366)
	at forge.toolbox.FGestureAdapter.touchUp(FGestureAdapter.java:245)
	at forge.toolbox.FGestureAdapter.touchUp(FGestureAdapter.java:177)
	at com.badlogic.gdx.InputEventQueue.drain(InputEventQueue.java:70)
	at com.badlogic.gdx.backends.lwjgl3.DefaultLwjgl3Input.update(DefaultLwjgl3Input.java:190)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Window.update(Lwjgl3Window.java:414)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:181)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:153)
	at forge.app.GameLauncher.<init>(GameLauncher.java:139)
	at forge.app.Main.main(Main.java:30)
```

TrevorHayes:
> @TrevorHayes Looks like your translations are broken:

I know there aren't translations (that was mentioned in the code review explicitly) because I don't think you can just plug "perpetual mana cost" into Google translate and get a correct output. But when I tested other missing translations they seemed to fall back to the English text so it seemed to be kinda normal to be missing some of that stuff.
So - this is a stack trace but what is the actual program behavior? Is this causing an error? How did you generate the issue? 
I'm away for the next week so I won't be able to fix anything immediately.

tehdiplomat:
The issue isn't that there aren't other language translations. The issue is that the english translations are looking for a different phrase than what are provided in the properties file.

TrevorHayes:
Ah; I see. Is this causing a user-facing issue or just generating an error in the logs?

tool4ever:
fixed

PR #9325, opened 2025-12-13
Create illusory_thoughts.txt

Ellios77:
Adding the missing card from Drake Stone set.

### DIFF
--- a/forge-gui/res/cardsfolder/i/illusory_thoughts.txt
+++ b/forge-gui/res/cardsfolder/i/illusory_thoughts.txt
@@ -1,0 +1,9 @@
+Name:Illusory Thoughts
+ManaCost:U U
+Types:Sorcery
+A:SP$ ChangeZone | Origin$ Hand | Destination$ Library | LibraryPosition$ 0 | ValidTgts$ Opponent | ChangeType$ Card | ChangeNum$ 2 | IsCurse$ True | RememberChanged$ True | SubAbility$ DBChangeZoneLibrary | SpellDescription$ Look at target opponent’s hand and put up to two cards from it on top of their library. Search that player’s library for that many cards. That player puts those cards into his or her hand, then shuffles their library.
+SVar:DBChangeZoneLibrary:DB$ ChangeZone | Origin$ Library | Destination$ Hand | ChangeType$ Card | ChangeNum$ X | ChangeType$ Card | Chooser$ You | DefinedPlayer$ ParentTarget | SubAbility$ DBShuffle
+SVar:DBShuffle:DB$ Shuffle | Defined$ Targeted | SubAbility$ DBCleanup
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+SVar:X:Remembered$Amount
+Oracle:Look at target opponent’s hand and put up to two cards from it on top of their library. Search that player’s library for that many cards. That player puts those cards into his or her hand, then shuffles their library.


jamincollins:
## Findings
#### 1. Fixed ChangeNum instead of up to two cards
The card text requires “up to two cards”. The implementation uses ChangeNum$ 2 which forces exactly two cards to be moved from hand to library. If the opponent has fewer than two cards in hand, ChangeZoneEffect will attempt to move two cards and will either error or select fewer cards silently, breaking the “up to” wording.

Test case:
```java
// Setup: opponent has 1 card in hand
Player opponent = game.getPlayer("opponent");
opponent.getHand().add(new Card("Test"));
SpellAbility sa = new SpellAbility(illusoryThoughts, player);
sa.setTarget(opponent);
sa.resolve();
// Expected: opponent chooses 0-1 cards, actual: engine attempts to move 2 cards
```

Impact: Non-compliant with official rules, potential crash on empty hand.

#### 2. Duplicate ChangeType parameter in DBChangeZoneLibrary
SVar:DBChangeZoneLibrary contains `ChangeType$ Card | ChangeNum$ X | ChangeType$ Card`. The duplicate ChangeType key is parsed by AbilityUtils.filterListByType which reads sa.getParam("ChangeType") returning the last value. While it currently resolves to Card, the duplicate is undefined behavior and may break if parser is changed to first-wins or throws on duplicate.

Test case:
```java
SpellAbility sa = CardFactoryUtil.parseSVar("DBChangeZoneLibrary", illusoryThoughts);
String ct = sa.getParam("ChangeType");
// ct is "Card" today, but duplicate key is present in source
```

Impact: Fragile data, future parser changes could change semantics.

#### 3. IsCurse$ True on a Sorcery
IsCurse is intended for Curse subtype abilities. Illusory Thoughts is a Sorcery. Setting IsCurse$ True may cause the ability to be treated as a curse for UI filtering, logging, or replacement effects, and deviates from Head Games which uses IsCurse$ True because the effect is modeled as a curse-like hand manipulation. The flag is misleading for a non-curse card.

Test case:
```java
SpellAbility sa = illusoryThoughts.getSpellAbility(0);
assert !sa.isCurse(); // currently true due to param
```

Impact: Inconsistent metadata, potential misclassification.

#### 4. RememberChanged amount with fixed ChangeNum
RememberChanged$ True records the number of cards moved. With ChangeNum$ 2 fixed, Remembered$Amount will always be 2 even when the opponent has <2 cards, causing DBChangeZoneLibrary to search for 2 cards from library and attempt to return 2 cards to hand. This can cause the search to fail or return fewer cards than remembered.

Test case:
```java
// Opponent hand size =1
// First ChangeZone moves 1 card, Remembered$Amount =1? Actually ChangeNum fixed to 2 -> engine may still move 1
// DBChangeZoneLibrary uses ChangeNum$ X where X = Remembered$Amount -> expects 2 cards from library
```

Impact: Mismatch between cards moved and cards searched, game state desync.

### Performance and security
#### 5. Curly quotes in SpellDescription and Oracle
The SpellDescription and Oracle use Unicode curly apostrophes ’. Forge's parser and localization systems expect ASCII apostrophes. This can cause string comparison failures, incorrect display, and potential parsing errors in card factory.

Test case:
```java
String oracle = illusoryThoughts.getParam("Oracle");
assert oracle.contains("opponent’s"); // contains U+2019
// Later lookup with ASCII ' fails
```

Impact: Display/lookup inconsistencies, potential silent failures.

## Verdict
Request Changes
The card implementation has multiple rule compliance and data integrity issues: fixed ChangeNum instead of up to two, duplicate ChangeType parameter, misuse of IsCurse flag, and potential mismatch between remembered amount and actual cards moved. Curly quotes also risk parsing issues.

## Recommended Fixes
1. Change ChangeNum$ 2 to optional selection: use ChangeNum$ 2 with Optional$ True or ChangeNumDesc$ up to two and allow player choice.
2. Remove duplicate ChangeType$ Card from SVar:DBChangeZoneLibrary.
3. Remove IsCurse$ True unless explicitly needed for UI; verify with Head Games/Jester's Mask pattern.
4. Ensure RememberChanged amount reflects actual cards moved, not fixed 2.
5. Replace curly quotes with ASCII apostrophes in SpellDescription and Oracle.

---

PR #9332, opened 2025-12-14
Add adventure world-specific battle and menu background support

### LINKED ISSUE #9299: World-Specific Battle Backgrounds
## Summary
Allow Adventure worlds to provide custom battle background images, with fallback to default skin images.

## Problem
Battle backgrounds are loaded exclusively from `forge-gui/res/skins/default/`. All adventure worlds share the same images, preventing visual customization to match each world's theme.

## Proposed Solution
Check for world-specific backgrounds before falling back to skin folder:

1. `[WorldName]/skins/adv_bg_match.jpg` (world-specific)
2. `skins/[current-skin]/adv_bg_match.jpg` (skin)
3. `skins/default/adv_bg_match.jpg` (fallback)

This mirrors the existing pattern for maps, decks, sprites, and other world resources.

## Implementation
Modify `FSkinTexture.load()` to check `Config.getFile("skins/" + filename)` before `FSkin.getSkinFile()`.

**File:** `forge-gui-mobile/src/forge/assets/FSkinTexture.java`

Maybe it's also possible to have a world-based background for the main menu?

vanja-ivancevic:
- Allow adventure worlds to provide custom battle backgrounds by placing files in their `skin/` directory, using the same pattern that `SoundSystem` uses to allow custom music.
- The texture loading order for adventure backgrounds (ADV_BG_*) would be:
  1. Adventure world skin directory (e.g., `Shandalar Old Border/skin/adv_bg_swamp.jpg`)
  2. Common adventure skin directory (`common/skin/`)
  3. Default skin directory (fallback)

Closes #9299


forge-gui-mobile/src/forge/assets/FSkinTexture.java
```diff
@@ -116,14 +117,55 @@ public void load() {
         load("");
     }
 
+    private boolean isAdventureBackground() {
+        return name().startsWith("ADV_");
+    }
```
Jetz72, line 129: You might check it against a constant EnumSet instead of inferring behavior from the symbol name.

forge-gui-mobile/src/forge/assets/FSkinTexture.java
```diff
@@ -116,14 +117,55 @@ public void load() {
...
+
+    private FileHandle getAdventureBackgroundFile() {
+        String adventureDirectory = GuiBase.getAdventureDirectory();
+        if (adventureDirectory == null || adventureDirectory.isEmpty()) {
+            return null;
+        }
+
+        // Check adventure-specific skin directory first
+        String adventureSkinPath = adventureDirectory + "skin/" + filename;
```
Jetz72: I'd recommend adding something to `ForgeConstants` like `ADVENTURE_SKIN_DIR = "skin" + PATH_SEPARATOR` so it's consistent with others.

Jetz72 (commented):
Does the background correctly update if you change planes without restarting the application? I think `load` is only invoked when the texture first appears, and I don't know if it's unloaded after that point.

forge-gui-mobile/src/forge/assets/FSkinTexture.java
vanja-ivancevic: Maybe the prefix string approach might be better because it's self-maintaining. If a new ADV_BG_* is added to the enum later, it automatically gets the correct loading behavior without updating a separate constant.

forge-gui-mobile/src/forge/assets/FSkinTexture.java
vanja-ivancevic: Good idea. I'll do that.

vanja-ivancevic:
> Does the background correctly update if you change planes without restarting the application? I think `load` is only invoked when the texture first appears, and I don't know if it's unloaded after that point.

It doesn't automatically update without restarting. However, changing planes prompts the user to restart anyways.

forge-gui-mobile/src/forge/assets/FSkinTexture.java
vanja-ivancevic: Done — added SKIN_DIR to ForgeConstants. Is that fine?

Jetz72: Little ambiguous since we also use "skins/" for regular application skins, but it should be fine for now. Might be worth consolidating those directory names someday.

forge-gui-mobile/src/forge/assets/FSkinTexture.java
Jetz72: I kinda have reservations about representing adventure regions with their own entries in this enum in the first place, but if anyone ever gets around to cleaning that up, I guess it won't matter either way.

Jetz72:
I'd like to think we'll eventually get rid of the need to restart to change planes. I don't like adding to the list of problems that will need to be identified and solved in order to achieve that. Could just use a static method to clear out all the adventure background textures and invoke it when switching planes. (Which would be another use case for that EnumSet just sayin' 😝)

vanja-ivancevic:
Great point. Added `invalidateAdventureTextures()` which resets the cached textures when switching planes - called alongside `setAdventureDirectory()` in all four places. Also switched to using an EnumSet as you suggested. Would this work?

vanja-ivancevic:
Sorry to bother you, did you have a chance to look at this?

Jetz72:
Not yet - been away from the project while traveling the past few weeks. Hoping to catch back up on things sometime this week.

vanja-ivancevic:
Thank you, no problem! Take your time :)

Jetz72 (approved):

vanja-ivancevic:
Do I need to still click something? "Update with merge commit" or "Update with rebase"?

Jetz72:
Nah, just when I'm the sole reviewer I tend to leave my approvals for a bit to give anyone else a chance to step in and veto if they want, or do the merge themselves if they agree. In this case I also approved it a bit before we were gonna do a release and we try to hold off on anything nonessential the immediate lead-up to those.

PR #9339, opened 2025-12-18
Remove load from paper card call during game state copying

timboode:
# Problem

The AI simulation system was experiencing significant performance bottlenecks due to GameCopier.createCardCopy() calling Card.fromPaperCard() for every card copied during simulation. This method re-parses card definitions from scratch by invoking CardFactory.getCard(), which is computationally expensive when performed repeatedly across thousands of simulated game states. Profiling identified this as the most critical bottleneck in the /forge-ai module.

# Solution

Implemented a new optimized code path via CardCopyService.copyStatsToGame() that efficiently copies card statistics between game instances without re-parsing card definitions. The method creates a minimal card shell using Card.newCard() and then copies all necessary state from the source card using the existing CardState.copyFrom() infrastructure. The new path preserves all card properties including types, abilities, counters, zone information, and controller/owner relationships while avoiding the expensive parsing overhead.

# Testing

Added 13 comprehensive unit tests to GameSimulationTest.java covering various card scenarios: basic creatures, creatures with counters, lands, artifacts, enchantments, planeswalkers, multi-faced cards (MDFCs, transform, adventure), and tokens. These tests verify that copied cards maintain correct properties and that the simulation produces identical results compared to the original implementation.


tool4ever:
imo no reason to keep this open 🤷‍♂️
looks like this was just some drive by PR attempt?

PR #9348, opened 2025-12-20
Fix AI not casting Bestow

tool4ever:
The api checks were failing, so we pipe the altHost from canPlay into it instead

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -834,7 +834,9 @@ public class AiController {
     }
 
     private AiPlayDecision canPlayAndPayFor(final SpellAbility sa) {
-        Card altHost = null;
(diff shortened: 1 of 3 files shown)


Hanmac:
the `updateKeyword` should probably be part of `getAlternateHost` ?

also check out undoAlternateHost in #6992

tool4ever:
> the `updateKeyword` should probably be part of `getAlternateHost` ?

it's disabled there since it would cause view update
but maybe flickering is no longer a problem since it's LKI anyway?
technically only `updateKeywordsCache` is needed

> also check out undoAlternateHost in #6992

I don't see the point?
just leave the LKI alone after using it

Hanmac:
i just think that updateKeywords should be called in getAlternateHost
because that would be cleaner (on the LKI host of cause)

then you don't need to make extra checks for bestow in the AI code

Also this MR should probably be on top of #6992

PR #8434, opened 2025-08-18
YEOE: Cloudsculpt Armorer

dracontes:
~Tested card scripts for:~
- ~[Cosmic Sovereign](https://www.reddit.com/r/MagicAlchemy/comments/1mr2sap/yeoe_cosmic_sovereign/)~
- ~[Drix Interlacer](https://www.reddit.com/r/MagicArena/comments/1mr2l5m/yeoe_drix_interlacer/)~
- ~[Gravitic Herald](https://www.reddit.com/r/MagicAlchemy/comments/1mr3v13/yeoe_gravitic_herald/)~

Needing engine update:
- [Cloudsculpt Armorer](https://www.reddit.com/r/MagicAlchemy/comments/1mr2gre/yeoe_cloudsculpt_armorer/)
=> The second triggered ability needs a new trigger type, something which is frankly beyond my expertise to set up. It's currently using `CounterTypeAddedAll` as a germane placeholder for testing purposes.

PR #9353, opened 2025-12-22
Added Missing Alchemy Card - Cosmic Sovereign

Darthseid:
Cosmic Sovereign was never ported to this game. I took the time to write its rules and upload the art to custom cards. It is already included in the Edge of Eternities Set.


paulsnoops:
Thanks for your contribution! However, we already have a PR open for this card #8434

Also, we don't host images for real cards

PR #9358, opened 2025-12-23
GameAction: reduce View Updates per Layer

Hanmac:
This should reduce unnecessary updates to the GUI

### DIFF
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -949,7 +949,7 @@ public class GameAction {
 
     public void ceaseToExist(Card c, boolean skipTrig) {
         if (c.isInZone(ZoneType.Stack)) {
-            c.getGame().getStack().remove(c);
+            game.getStack().remove(c);
         }
 
         final Zone z = c.getZone();
@@ -1087,19 +1087,6 @@ public class GameAction {
 
         // remove old effects
         game.getStaticEffects().clearStaticEffects(affectedCards, affectedPerLayer);
-        CardCollection affectedKeywordsBefore = new CardCollection();
-        if (affectedPerLayer.containsKey(StaticAbilityLayer.TEXT)) {
-            affectedKeywordsBefore.addAll(affectedPerLayer.get(StaticAbilityLayer.TEXT));
-        }
-        if (affectedPerLayer.containsKey(StaticAbilityLayer.TYPE)) {
-            // setting Basic Land Type case
-            affectedKeywordsBefore.addAll(affectedPerLayer.get(StaticAbilityLayer.TYPE));
-        }
-        if (affectedPerLayer.containsKey(StaticAbilityLayer.ABILITIES)) {
-            affectedKeywordsBefore.addAll(affectedPerLayer.get(StaticAbilityLayer.ABILITIES));
-        }
-        affectedKeywordsBefore.forEach(Card::updateKeywordsCache);
-
 
         // search for cards with static abilities
         final FCollection<StaticAbility> staticAbilities = new FCollection<>();
@@ -1109,7 +1096,7 @@ public class GameAction {
             dependencies = HashBasedTable.create();
         }
 
-        game.forEachCardInGame(new Visitor<Card>() {
+        game.forEachCardInGame(new Visitor<>() {
             @Override
             public boolean visit(final Card c) {
                 // need to get Card from preList if able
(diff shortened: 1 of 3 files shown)


forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -1116,6 +1116,7 @@ public boolean visit(final Card c) {
 
         staticAbilities.sort(effectOrder);
 
+        final Map<StaticAbilityLayer, List<Card>> affectedPerLayer = Maps.newHashMap();
```
tool4ever: you can't only check the new affected
old ones (removal) might also need to be updated
that's why it will get ugly fast

Hanmac: Shouldn't the old ones be cleaned by StaticEffect remove method?

tool4ever: no, their big view updates should also be skipped so it's not done twice for most cases

Hanmac: But this is not part of this PR?

tool4ever: then your PR will break it...

Hanmac: should be fixed

in the very far future, the StaticEffect class can be removed anyway when everything uses Layers

Hanmac:
@tool4ever is this ready?

tool4ever:
I should have time to check tomorrow

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -1223,13 +1229,37 @@ public boolean visit(final Card c) {
...
+        if (affectedPerLayer.containsKey(StaticAbilityLayer.MODIFYPT)) {
+            affectedPT.addAll(affectedPerLayer.get(StaticAbilityLayer.MODIFYPT));
+        }
+        /*
+        if (affectedPerLayer.containsKey(StaticAbilityLayer.SWITCHPT)) {
+            affectedPT.addAll(affectedPerLayer.get(StaticAbilityLayer.SWITCHPT));
         }
+        //*/
+        affectedPT.forEach(Card::updatePTforView);
+        affectedKeywords.forEach(Card::updateKeywords);
```
tool4ever, line 1262: unfortunately this causes "May be played by" to not appear

Hanmac: i moved May be played to NonAbilityText
other stuff probably should be moved there too

but MayPlay is on my short list anyway to be refactored or removed

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -2410,7 +2418,21 @@ public final String getNonAbilityText() {
             sbLong.append("\r\n");
         }
         sb.append(sbLong);
-        return CardTranslation.translateMultipleDescriptionText(sb.toString(), this);
```
tool4ever, line 2413: why remove translation?

Hanmac: because this part is for hidden keywords, and this should not have worked anyway

forge-game/src/main/java/forge/game/StaticEffect.java
```diff
@@ -268,27 +279,34 @@ final CardCollectionView remove(List<StaticAbilityLayer> layers) {
...
-                    affectedCard.removeCantHaveKeyword(getTimestamp());
+                    abilitiesChanged |= affectedCard.removeCantHaveKeyword(getTimestamp());
                 }
 
                 affectedCard.removeChangedSVars(getTimestamp(), ability.getId());
 
                 // need update for clean reapply
-                affectedCard.updateKeywordsCache(affectedCard.getCurrentState());
+                if (abilitiesChanged) {
+                    affectedCard.updateKeywordsCache(affectedCard.getCurrentState());
```
tool4ever: could probably also pull out the cache rebuild after removing everything first (same for type)?

forge-game/src/main/java/forge/game/GameAction.java
tool4ever: updates still missing (MayPlay isn't a hidden KW)

forge-game/src/main/java/forge/game/StaticEffect.java
Hanmac: what exactly do you mean there? the keywords need to be cleaned before static abilities apply

tool4ever: you will rebuild the cache for each effect you remove
but it should be good enough to remove ALL first, then only ONCE

forge-game/src/main/java/forge/game/GameAction.java
Hanmac: fixed

forge-game/src/main/java/forge/game/StaticEffect.java
tool4ever: though might need some switch because for dependencies we're removing single ones only 🤔

Hanmac: removed from static Effect cleanup and added in between ClearStatic and StaticAbilities

tool4ever (approved):

PR #9363, opened 2025-12-24
CardState: store changed CardType

Hanmac:
this should reduce the amount of times when CardTypeWithChanges gets calculated

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -591,6 +591,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
         currentStateName = state;
         currentState = getState(state);
 
+        updateTypeCache();
         if (updateView) {
             updateStateForView();
             view.updateNeedsTransformAnimation(needsTransformAnimation);
@@ -4268,6 +4269,8 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
             changed = true;
         changedCardTypes.clear();
 
+        this.updateTypeCache();
+
         return changed;
     }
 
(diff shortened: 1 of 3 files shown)


forge-game/src/main/java/forge/game/card/CardState.java
```diff
@@ -720,6 +738,7 @@ public final void copyFrom(final CardState source, final boolean lki, final Card
         // Makes a "deeper" copy of a CardState object
         setName(source.getName());
         setType(source.type);
+        this.changedType = source.changedType;
```
tool4ever: is this line for LKI?
but it would be wrong for normal copy effects to just keep all changes...

Hanmac: CardState.copyFrom is called in GameCopier and LKI only

tool4ever: @Hanmac
no, you mean `Card.copyFrom`

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -598,6 +598,7 @@ public boolean setState(final CardStateName state, boolean updateView, boolean f
             if (game != null) {
                 // update Type, color and keywords again if they have changed
                 if (!changedCardTypes.isEmpty()) {
+                    updateTypeCache();
```
tool4ever: needs to happen even when view is skipped...

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4316,6 +4319,7 @@ public final void addChangedCardTypesByText(final CardType addType, final long t
         this.changedTextColors.addEmpty(timestamp, staticId);
         this.changedTextTypes.addEmpty(timestamp, staticId);
 
+        this.updateTypeCache();
```
tool4ever: maybe move to below function instead
you're missing at least `addChangedTextTypeWord`

tool4ever (approved):

PR #9374, opened 2025-12-28
Cleanup redundant field

tool4ever:


forge-game/src/main/java/forge/game/card/CardFactory.java
```diff
@@ -301,8 +299,10 @@ public static SpellAbility buildBasicLandAbility(final CardState state, byte col
         return sa;
     }
 
-    private static Card readCard(final CardRules rules, final IPaperCard paperCard, int cardId, Game game) {
+    private static Card readCard(final IPaperCard paperCard, int cardId, Game game) {
         final Card card = new Card(cardId, paperCard, game);
+        CardRules rules = paperCard.getRules();
+        card.updateRulesView(rules);
```
Hanmac: why the param there?
shouldn't you be able to call getRules inside the function?

PR #9378, opened 2025-12-28
Added Auntie Ool, Cursewretch card script to upcoming folder.

earno12:
Requesting to add the card script for Auntie Ool, Cursewretch.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/auntie_ool_cursewretch.txt
+++ b/forge-gui/res/cardsfolder/upcoming/auntie_ool_cursewretch.txt
@@ -1,0 +1,10 @@
+Name:Auntie Ool, Cursewretch
+ManaCost:1 B R G
+Types:Legendary Creature Goblin Warlock
+PT:4/4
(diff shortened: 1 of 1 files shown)


forge-gui/res/cardsfolder/upcoming/auntie_ool_cursewretch.txt
```diff
@@ -0,0 +1,10 @@
+Name:Auntie Ool, Cursewretch
+ManaCost:1 B R G
+Types:Legendary Creature Goblin Warlock
+PT:4/4
+K:Ward:Blight<2>
+T:Mode$ CounterAdded | ValidCard$ Creature.YouCtrl | TriggerZones$ Battlefield | CounterType$ M1M1 | Execute$ TrigDraw | TriggerDescription$ Whenever one or more -1/-1 counters are put on a creature you control, draw a card.
```
tool4ever, line 6: splitting it over two triggers means you're not checking the conditional part meant for resolving only then, so I will go with the version in #9379 instead

PR #9383, opened 2025-12-29
Added castability highlighting for cards in hand

jtomse:
This adds my number one favorite QoL feature from other MTG implementations (DOTP, Magic Arena, etc.) - castability highlighting.  Or in other words, the game uses a thin yellow border to highlight cards that you are able to cast at any given moment based on your available mana, timing, etc.  This does not highlight all *playable* cards, only cards that have a mana cost.  The option is toggled off by default.  Makes looking at hand and deciding what you can cast much easier to parse. 

Details:
Added a private helper method (canCastCard()) that checks if a card in hand can be cast with available mana. The method:
  1. Verifies the card is in hand
  2. Gets the actual Card object from the game by matching ID
  3. Retrieves all spell abilities for the card
  4. For each spell ability, checks if it can be played (timing/restrictions)
  5. Verifies the player can pay the mana cost using ComputerUtilMana.canPayManaCost()

If a card is castable, a thin yellow border is drawn using the same logic as the current green "hover" border

Also added a descriptive toggle in the in-game Preferences menu (towards the bottom of the "Graphics Options" section) to allow users to activate this new functionality as desired.  Been meaning to get around to adding this for prob 10 years now lol.

All feedback welcomed and appreciated


forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
```diff
@@ -299,6 +301,13 @@ protected final void paintComponent(final Graphics g) {
             g2d.fillRoundRect(cardXOffset - n2, (cardYOffset - n2) + offset, cardWidth + (n2 * 2), cardHeight + (n2 * 2), cornerSize + n2, cornerSize + n2);
         }
 
+        // Yellow border for cards in hand that can be cast with available mana
+        if (canCastCard() && isPreferenceEnabled(FPref.UI_SHOW_CASTABLE_BORDER)) {
```
tool4ever: you should change the order so the expensive check isn't run when the preference is off

forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
```diff
@@ -352,6 +361,80 @@ protected final void paintComponent(final Graphics g) {
...
+        }
+
+        final var game = gameView.getGame();
+            if (game == null) {
+                return false;
+        }
+
+        // Find the actual Card object in the player's hand by matching ID
+        Card actualCard = null;
+        for (final var player : game.getPlayers()) {
```
tool4ever: `Game.findByView`

forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
jtomse: Thanks, good catch, updated

forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
jtomse: Replaced nested loop with findByView call instead as suggested - thanks!

forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
```diff
@@ -352,6 +361,71 @@ protected final void paintComponent(final Graphics g) {
...
+    private boolean canCastCard() {
+        try {
+            // Only check cards in hand
+            if (card == null || !ZoneType.Hand.equals(card.getZone())) {
+                return false;
+            }
+
+            // Get the game view to access the actual Game object
+            final var gameView = matchUI.getGameView();
+                if (gameView == null) {
```
tehdiplomat: Can you fix the spacing here. It looks kinda weird.

forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
```diff
@@ -352,6 +361,71 @@ protected final void paintComponent(final Graphics g) {
...
+                // First check if the spell can be played (timing, restrictions, etc.)
+                if (!sa.canPlay(true)) {
+                    continue;
+                }
+
+                // Set the activating player for the spell ability
+                sa.setActivatingPlayer(player);
+
+                // Check if the player can actually pay the mana cost from available sources
+                if (ComputerUtilMana.canPayManaCost(sa, player, 0, false)) {
```
tehdiplomat, line 415: A bit of a misnomer here, since the AI not being able to pay the mana cost doesn't mean its not able to be paid. Also there could be  circumstances where there's no legal targets, or other costs are unable to be paid that this would bypass.

jtomse: You're right, it's really to show if you have the available mana to cast something so you don't have to count in your head.  But it doesn't mean there's necessarily a valid target, or that you have enough life to pay the additional cost , or that playing the card won't trigger a game loss, etc.  It also doesn't highlight cards with no mana cost that can be played (lands, zero CMC, etc.).  It really is just a helper so you don't have to do a manual mana check/count each turn to see if you finally have enough mana of the right kind to cast that creature you've been waiting to cast, or if you've tapped 2 Plains and 3 Islands this turn, do you still have enough available to cast your other UW.  Do you think renaming the feature is in order? Something other than castability highlighting like "Mana available" or somthing like that? Lmk what you think.

MostCromulent:
@jtomse just thought it might be helpful to draw your attention to https://github.com/Card-Forge/forge/pull/9643.

That PR has very different functionality but it involves solving a similar "what actions can the player take on their turn?" heuristic as you are with your card highlighting feature. There may be something of use in there for you. 

(The card highlighting would definitely be a QOL improvement on top of what I'm working on there!)

tool4ever:
Superseded by #10680

PR #9387, opened 2025-12-30
Support for custom alternative cost targeting

Eradev:
Added further support for custom cards targeting using Named$. 

Now supports naming AlternativeCost

```text
S:Mode$ AlternativeCost | Named$ Showcase | ValidSA$ Spell.Self | EffectZone$ All | Cost$ 3 U | Description$ Showcase {3}{U} (Then if you cast this for its showcase cost, create a 1/1 red Bard creature token that can't block and exile this. Whenever it attacks, you may cast a free copy of this.)
```

Allowing targeting like the following

```text
T:Mode$ SpellCast | ValidActivatingPlayer$ You | ValidCard$ Card | Condition$ FromNamedAbilityShowcase | Execute$ TrigToken | TriggerDescription$ Whenever you cast a spell for its showcase cost, create a 2/2 blue Drake creature token with flying.
```

### DIFF
--- a/forge-game/src/main/java/forge/game/trigger/Trigger.java
+++ b/forge-game/src/main/java/forge/game/trigger/Trigger.java
@@ -446,21 +446,6 @@ public abstract class Trigger extends TriggerReplacementBase {
             if (game.getCombat().getAttackersAndDefenders().values().containsAll(attacker.getOpponents())) {
                 return false;
             }
-        } else if (condition.startsWith("FromNamedAbility")) {
-            var rest = condition.substring(16);
-            final SpellAbility trigSA = (SpellAbility) runParams.get(AbilityKey.Cause);
-
-            if (trigSA != null && trigSA.getName().equals(rest)) {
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/trigger/Trigger.java
```diff
@@ -450,9 +450,17 @@ public boolean meetsRequirementsOnTriggeredObjects(Game game, final Map<AbilityK
             var rest = condition.substring(16);
             final SpellAbility trigSA = (SpellAbility) runParams.get(AbilityKey.Cause);
 
-            if (trigSA != null && !trigSA.getName().equals(rest)) {
-                return false;
+            if (trigSA != null && trigSA.getName().equals(rest)) {
```
tool4ever: imo this extra condition can be removed since you already have access to `NamedAbility` via ForgeScript

tool4ever (approved):

PR #9389, opened 2025-12-30
DelayedReveal with multiple zone

### LINKED ISSUE #9385: Some Surgical Extraction Effects Don't Let You Search Hand
As title says, Ancient Vendetta (and some other cards with the same effect like The Rise of Sozin iirc) doesn't allow you to look at opponent's hand for some reason despite you searching their hand as part of the effect.

Use latest snapshot on desktop, cast Ancient Vendetta and target opponent and choose a card in their library and/or graveyard but not currently in their hand iirc.

You should still be able to see your opponent's hand during this, but for some reason it does not allow you.

Image

tool4ever:
Closes #9385


tool4ever:
ugh, I didn't think mobile had more complicated logic for this
I'll see if I can figure it out, otherwise I might just have to revert the scripts :/

Hanmac (approved):

PR #9395, opened 2025-12-31
Use UNKNOWN folder if set is ???

Eradev:
For those removed Alchemy cards and Arena ones no present on Scryfall.

### DIFF
--- a/forge-core/src/main/java/forge/util/ImageUtil.java
+++ b/forge-core/src/main/java/forge/util/ImageUtil.java
@@ -106,7 +106,7 @@ public class ImageUtil {
 
         CardRules card = cp.getRules();
         String edition = cp.getEdition().equals(CardEdition.UNKNOWN_CODE)
-                ? "UNKNOWN"
+                ? CardEdition.UNKNOWN_SET_NAME
                 : cp.getEdition();
         s.append(toMWSFilename(nameToUse));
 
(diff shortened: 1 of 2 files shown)


forge-core/src/main/java/forge/util/ImageUtil.java
```diff
@@ -104,7 +105,9 @@ public static String getImageRelativePath(PaperCard cp, String face, boolean inc
         StringBuilder s = new StringBuilder();
 
         CardRules card = cp.getRules();
-        String edition = cp.getEdition();
+        String edition = cp.getEdition().equals(CardEdition.UNKNOWN_CODE)
```
tehdiplomat, line 108: Should we just update `UNKNOWN_CODE`edition  name to be `UNKNOWN` instead of `Undefined`?

Eradev: It is not undefined. The code is ???.

tehdiplomat: I didn't say the code. I said the name. 

```
public static final CardEdition UNKNOWN = new CardEdition("1990-01-01", UNKNOWN_CODE, "??", Type.UNKNOWN, "Undefined", FoilType.NOT_SUPPORTED, new EditionEntry[]{});
```

PR #9396, opened 2025-12-31
Expose LandType count

Eradev:
Expose LandType count for custom cards.


Jetz72:
If someone wants to patch up the merge conflict, seems like a straightforward enough add. Guess it just fell through the cracks.

tool4ever:
I was mostly thinking about our approach to adding stuff for custom cards - if there was a way to make this more generic
(or at least add a comment so it doesn't get cleaned up)

Jetz72:
I'm not sure there's a good way to make it generic without refactoring a bunch of card scripts. Could just search for the substring "Type" then inspect whatever's before it but that's kinda ugly. I'll just slap on a comment. But I do think this is worth extra consideration in future cases when we clean up unused APIs.

PR #9404, opened 2026-01-02
VentureEffect: move Dungeon to Tokens

### LINKED ISSUE #8374: Fixed dungeons images fetch
> Fixed dungeons images fetch

I need to fix that in a different MR
Because the ImageFetch is the wrong place to get the Dungeon type

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/issues/7921#issuecomment-3186901920_

Part of this Issue to be addressed:
* DUNGEONS("dungeons") section isn't really needed? Or is it?
* Or should they be moved to Other section?
* (Dungeon use T on their Card, i might move the Dungeons to TokenScripts?)

Venture Effect uses to iterate over all Variant Cards
https://github.com/Card-Forge/forge/blob/599a6290680f9b154dbeaef64ba37b992dce707d/forge-game/src/main/java/forge/game/ability/effects/VentureEffect.java#L48-L58
if we are specifically looking for Dungeons, we might use better ways than that?

I would like to update the pattern and tokenPattern to use named captures:
```java
"^(?:(?<cnum>.?[0-9A-Z-]+\\S?[A-Z]*)\\s)?(?:(?<rarity>[SCURML])\\s)?(?<name>[^@\\$]+?)(?: @(?<artist>[^$]+?))?(?: \\$(?<vname>.+))?$"
```
if we change the Regexp Pattern, we might look into this Issue with FlavorNames #8281

Hanmac:
Closes #8374

moves Dungeons into Token part, because they have type T
That also fixes the Imagefetching

### DIFF
--- a/forge-core/src/main/java/forge/card/CardEdition.java
+++ b/forge-core/src/main/java/forge/card/CardEdition.java
@@ -142,7 +142,6 @@ public final class CardEdition implements Comparable<CardEdition> {
         PRERELEASE_PROMO("prerelease promo"),
         BUNDLE("bundle"),
         BOX_TOPPER("box topper"),
-        DUNGEONS("dungeons"),
         JUMPSTART("jumpstart"),
         REBALANCED("rebalanced"),
         ETERNAL("eternal"),


Hanmac:
I probably need to see if the Card Faces really support Netwirk play, but that's for later

tool4ever:
ugh I knew I forgot to check something:
> io.netty.handler.codec.EncoderException: java.io.NotSerializableException: forge.card.CardFace

PR #9412, opened 2026-01-03
Adds flying animation for flying creatures

jtomse:
This adds my second favorite QoL feature from other MTG implementations (DOTP, Magic Arena, etc.) - a subtle floating animation for flying creatures. The option is toggled off by default. Added a descriptive toggle in the in-game Preferences menu (towards the bottom of the "Graphics Options" section) to allow users to activate this new functionality as desired. Not only makes parsing game-states with many creatures easier, it looks cool too!
 
All feedback welcomed and appreciated


forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
```diff
@@ -112,12 +114,18 @@ public class CardPanel extends SkinnedPanel implements CardContainer, IDisposabl
...
     private int cardXOffset, cardYOffset, cardWidth, cardHeight;
+    private int baseY = -1;
+    private float flyingPhase = 0f;
     private boolean isSelected;
     private boolean hasFlash;
     private CachedCardImage cachedImage;
 
     private static Font smallCounterFont;
     private static Font largeCounterFont;
+    private static final WeakHashMap<CardPanel, Void> allPanels = new WeakHashMap<>();
```
tool4ever, line 125: looks nice indeed
I'm mostly wondering why you went with tracking all panels across each thread on your own instead of accessing them via a `PlayArea` instance?
Not saying rework is absolutely necessary since for an optional feature I'd consider it isolated enough already but still seems a bit weird...

jtomse: Agreed, still learning the codebase - makes much more sense to use PlayArea.getCardPanels and make sure only battlefield instances are registered.  Closing this PR and will open new with refactored code.  Appreciate you pointing me in the right direction.

jtomse:
Will refactor - please see last comment.

PR #9413, opened 2026-01-04
Some cleanup

tool4ever:


forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -2920,16 +2920,19 @@ private void addCardToZone(ZoneType zone, final boolean repeatLast, final boolea
...
-                cardFaceView = new CardFaceView(CardTranslation.getTranslatedName(cardFace.getDisplayName()), cardFace.getName());
-                choices.add(cardFaceView);
+            final CardFaceView f;
+            if (repeatLast) {
+                f = lastAdded;
+            } else {
+                List<CardFaceView> choices = new ArrayList<>();
+                for (ICardFace cardFace : carddb.getAllFaces()) {
+                    choices.add(
+                            new CardFaceView(CardTranslation.getTranslatedName(cardFace.getDisplayName()), cardFace.getName())
```
Hanmac, line 2930: Probably need to fix this, the order is off

tool4ever: yea I still need to test if it would use the translated Name without
but I can't just switch it because that would also mess with selecting flavor names :/
and previously they got sorted together too, not sure yet if I like that or not

Hanmac: With another change I flipped name vs display name, that's why it was wrong order

But also I added `CardFaceView::new` to go directly from CardFace to View

See my fix there
https://github.com/Card-Forge/forge/commit/28f090362f1e3b15c6d5c0481202b4d39fb97af2

PR #9421, opened 2026-01-05
Added flying animation for flying creatures

jtomse:
This is a refactored version of my previous attempt.  Adds my second favorite QoL feature from other MTG implementations (DOTP, Magic Arena, etc.) - a subtle floating animation for flying creatures. The option is toggled off by default. Added a descriptive toggle in the in-game Preferences menu (towards the bottom of the "Graphics Options" section) to allow users to activate this new functionality as desired. Not only makes parsing game-states with many creatures easier, it looks cool too!

Please let me know if this attempt looks better - thanks!


forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
```diff
@@ -1157,6 +1172,22 @@ private boolean showAbilityIcons() {
...
     }
 
+    boolean shouldFloat() {
+        if (!FModel.getPreferences().getPrefBoolean(FPref.UI_ANIMATE_FLYING)) {
+             return false;
+        }
+        if (!card.getCurrentState().hasFlying()) {
+            return false;
+        }
+        if (card == null || isAnimationPanel || !displayEnabled) {
```
tehdiplomat: Should probably do your null check before you check if that card has flying.

MostCromulent:
I tested this out and the floating animation seemed to cause distortion on the card image and text of the flyer, in a way I thought was a bit distracting. 

Not sure if this bothers anyone else. I guess one answer is you can always turn it off if you don't like it.

jtomse:
Hmm..I don't get any distortion on my end - it's crystal clear.  Maybe due to monitor refresh rate?  Mine is set to 165 Hz.  Is yours different?  The animation plays at 20 FPS, so if others are getting distortion or tearing I could increase it.  Just trying to keep resource usage as low as possible.  Maybe I went too low.

jamincollins:
## Findings

#### 1. shouldFloat checks card.getZone() == Battlefield but does not verify controller or visibility
The shouldFloat() method in CardPanel returns true only if card is non-null, hasFlying, not animation panel, displayEnabled, and zone equals Battlefield. It does not verify that the card is visible to the local player or that it is actually on the battlefield for the current view. In multiplayer, cards from other players are displayed but may be in hidden zones or face-down. If a card is face-down in play, hasFlying may be unknown.

Test case:
```java
CardPanel panel = new CardPanel();
panel.card = mockCardWithZoneBattlefieldAndHasFlyingTrueButFaceDown();
boolean floats = panel.shouldFloat();
```
If face-down creature with flying is revealed later, the animation may start/stop incorrectly.

Impact: animation may run for cards that should not be animated, violating expected UI behavior. MTG rule compliance not directly impacted but user experience degraded.

#### 2. baseY update threshold of 5 pixels causes drift accumulation
In setCardBounds, baseY is updated only when !shouldFloat || Math.abs(y - baseY) > 5. If a floating card is nudged by layout changes smaller than 5px repeatedly, baseY remains stale, causing cumulative offset drift.

Test case:
```java
CardPanel panel = new CardPanel();
panel.baseY = 100;
panel.flyingPhase = 0;
// layout moves panel to y=102, shouldFloat true
panel.setCardBounds(0, 102, 200, 280);
// baseY stays 100 because |102-100|=2 <=5
// floatOffset applied relative to 100, visual position drifts
```

Impact: visual jitter / drift over time, especially with frequent re-layouts.

#### 3. shouldFloat accesses card.getCurrentState().hasFlying() without null guard on getCurrentState()
The code checks `card == null` then calls `!card.getCurrentState().hasFlying()`. If card.getCurrentState() returns null, NPE occurs.

Test case:
```java
Card card = mockCardWithNullCurrentState();
panel.card = card;
boolean floats = panel.shouldFloat(); // NPE at getCurrentState()
```

Impact: crash when card state not initialized.

#### 4. flyingPhase overflow wraps but does not handle negative increments
flyingPhase is incremented by FLYING_SPEED = 0.12f each tick and wrapped when >= FLYING_PHASE_MAX. If flyingPhase ever becomes negative due to external modification, sin will still work but wrap never resets.

Test case:
```java
panel.flyingPhase = -0.5f;
panel.flyingPhase += 0.12f; // stays negative
```

Impact: potential slow drift if phase initialized incorrectly.

#### 5. ANIMATION_DELAY_MS = 50ms assumes 20 FPS, no frame-rate independence
Timer fires every 50ms regardless of system load. Under load, timer may coalesce, causing phase jumps.

Test case:
```java
// System under load, Timer events delayed to 200ms intervals
panel.flyingPhase += 0.12f * 4; // jump
```

Impact: animation stutters.

#### 6. UI_ANIMATE_FLYING default false preserves backwards compatibility but preference change requires new match
The note says Requires starting a new match. However, startAnimationTimerIfNeeded is called only in PlayArea constructor. If preference toggled mid-match via preferences UI, timer is not started/stopped until next match.

Test case:
```java
FModel.getPreferences().setPref(FPref.UI_ANIMATE_FLYING, "true");
playArea = new PlayArea(...); // already constructed, no registration
```

Impact: user expectation mismatch; preference not applied until restart.

#### 7. WeakHashMap registeredPlayAreas may lose entries under GC pressure
registeredPlayAreas uses WeakHashMap with PlayArea keys. If PlayArea is only weakly reachable, GC may collect it while timer is running, causing ConcurrentModification during iteration.

Test case:
```java
// PlayArea reference held only by WeakHashMap
System.gc();
updateFlyingAnimationForAllPlayAreas(); // keySet changes mid-iteration
```

Impact: intermittent missed updates or NullPointerException.

#### 8. updateFlyingAnimationForAllPlayAreas calls setCardBounds which triggers layout, causing re-entrant updates
setCardBounds calls setBounds which may trigger repaint and layout. Updating all panels in timer tick may cause cascading layout recalculations.

Test case:
```java
for each panel:
    panel.setCardBounds(...);
    // triggers doLayout -> may reposition other panels -> baseY stale
```

Impact: performance degradation, possible layout thrash.

#### 9. shouldFloat condition uses card.getZone() equality, but zone may change between ticks without baseY reset
If card moves from Battlefield to Graveyard, shouldFloat becomes false, baseY is updated on next setCardBounds because !shouldFloat triggers baseY = y. However, if card moves off battlefield while floating, the last floating offset remains until next setCardBounds.

Test case:
```java
panel.shouldFloat() true -> floating
card moves to Graveyard
timer tick still processes panel because playArea not updated
panel.shouldFloat() now false -> baseY updated
```

Impact: visual snap.

#### 10. PlayArea constructor registers only if zone == Battlefield at construction time
If PlayArea is reused or zone changes, registration is not updated.

Test case:
```java
PlayArea pa = new PlayArea(player, ZoneType.Hand);
pa.zone = ZoneType.Battlefield; // later
// not registered, animation never runs
```

Impact: missing animation for valid cases.

### Performance and security

#### 11. Animation timer runs at 20 FPS for all registered PlayAreas, even with zero flying creatures
updateFlyingAnimationForAllPlayAreas iterates all panels for each PlayArea even if shouldFloat is false for all.

Test case:
```java
PlayArea with 200 cards, none flying
Timer tick still loops 200 panels
```

Impact: unnecessary CPU usage, especially in large games.

#### 12. Static Timer is shared across all PlayAreas, stop condition based on registeredPlayAreas emptiness but timer may be stopped while new PlayArea registers
Race condition: startAnimationTimerIfNeeded checks animationTimer != null, then creates. Concurrent construction may create two timers.

Test case:
```java
Thread 1: PlayArea ctor, animationTimer == null -> create timer
Thread 2: PlayArea ctor, animationTimer == null (check before start)
```

Impact: duplicate timers.

## Verdict
Request Changes
The flying animation feature introduces multiple edge-case risks including null safety on card state, drift accumulation from baseY threshold, potential GC-related loss of registered PlayAreas, and re-entrant layout during animation ticks. The timer lifecycle is not robust to preference changes mid-match and may leak or duplicate timers under concurrency.

## Recommended Fixes
1. Add null guards for card.getCurrentState() in shouldFloat and handle face-down cards.
2. Remove baseY threshold or reset baseY on any position change when shouldFloat is true; consider using absolute bounds instead of offset accumulation.
3. Make animation frame-rate independent using delta time.
4. Register PlayArea on zone change and unregister when zone leaves Battlefield.
5. Use CopyOnWrite or synchronized snapshot for registeredPlayAreas iteration.
6. Start/stop timer on preference change via listener rather than only at construction.
7. Guard setCardBounds re-entrancy or use separate animation offset without triggering full layout.

---

Issue #6954, opened 2025-02-01
AI casting other player's spells with as foretold


wenasi:
**Describe the bug**
The AI can cast spells from opponents

**To Reproduce**
1. Disable Deck Conformity checks for easier testing. The bug originally happened to me in a Commander game with deck conformity checked.
2. Create a constructed deck with only Islands and As Foretolds
3. Create a constructed deck with only Reliquary Towers and Lightning Bolts. It helps putting some 3 toughness creature in to encourage the AI to use the lightning bolt
4. Give the bot the As Foretold deck, and use the Lightning Bolt deck yourself
5. After there is at least 1 time counter on As Foretold, the bot will occasionally cast a Lightning Bolt from you.

**Expected behavior**
The AI should only be able to cast spells it can legally cast with As Foretold

**Screenshots**

![Image](https://github.com/user-attachments/assets/970c9b90-08a9-4b85-b226-fb19dff9090f)
![Image](https://github.com/user-attachments/assets/b6eb1f40-d378-4fbd-be94-6a6a0782354d)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version 2.0.02-SNAPSHOT-01.31

**Additional context**
The Screenshots show the bot casting Lightning Bolt (11), with me as the owner. Notably 11 was not one of the Lightning Bolts I had in hand. My Graveyard was empty when the Lightning Bolt was cast. After it resolved, it did go into my graveyard though

After further testing I did see the number of cards in my deck drop by 1 when it cast something from me, so it does just cast the spell straight from the deck

I tested putting a few different cards into my deck. A later test had Lighning Bolt, Hop to It, Blood for the Blood God, Cruel Ultimatum and Endless One in it, while the deck for the AI still only had Islands and As Foretold in it.
The AI passed various turns without casting any of my spells, never cast Hop to It, but did cast Lightning Bolt, Cruel Ultimatum and Blood for the Blood God. 

I tested a round with Zur the Enchanter in my deck, which the AI also didn't cast

tool4ever:
It looks like it fails to find the CardPlayOption there:
https://github.com/Card-Forge/forge/blob/d0bd80f1582f0ac2f744174cfc276fb2851cae2d/forge-game/src/main/java/forge/game/spellability/SpellAbilityRestriction.java#L255

probably because it uses LKI host? You could try and grab it directly from the SA there but that might lead to other side effects.

MayPlay just needs its real rework :/

Hanmac:
[...]

Issue #4700, opened 2024-02-16
May Play needs to Refactored into Different Static Abilities


Hanmac:
in short summary, we need to split the MayPlay logic between multiple effects that may or may not be combined:
* Effects that alter the Cost like `Rooftop Storm`, that should be use a Static instead of a Keyword like Alternative Cost:0
* Effects that allow you to cast/play stuff from zones but without setting a cost, like `Augur of Autumn`
* Effects that combine both, but for these, the other effects can't be used too like `Bolas's Citadel`
* The Goal also was to reduce the use of Continuous for LKI stuff

i think my first attempt to split was 3+ years ago:
https://github.com/Card-Forge/forge/compare/mayPlayAlternativeCost

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/issues/4696#issuecomment-1948867789_

tool4ever:
@Hanmac
some steps I could try:
1. cleanup _Maestros Ascendancy_ + _Falco Spara_
they are special since they use additional cost + casting permission but the engine doesn't leave them as `basicSpell`

2. refactor remaining MayPlay*ManaCost$ into AlternativeCost static
I'm not sure yet if we need a third static, but if we do this first it should be easier to rework the rest with a MayPlay static and finally eliminate the `checkStaticAbilities()` calls

Hanmac:
one thing i sadly noticed that it doesn't really would reduce checkStaticAbilities that much,
because for checking the Spell, it needs to run them again anyway
maybe that can be optimized later



but i once made a List to separate them between:
* AlternativeZone (maybe additional Cost?)
* AlternativeCost
* And some kind of combinations of both

We need Static List, because AlternativeCost + AlternativeZone can be combined, like this:
`private List<StaticAbility> mayPlayList = Lists.newArrayList();`

tool4ever:
It's worth noting there are also cases where we need a stack copy to apply statics without any keyworded alternatives involved

E.g. _Spawnsire of Ulamog_ should allow you to cast any creature type spell while controlling _Maskwood Nexus_ (which doesn't affect sideboard)

Hanmac:
Because it says: "Eldrazi spells", yes it needs to use Stack Copy

Hanmac:
@tool4ever can we try to do the next step?
> Effects that allow you to cast/play stuff from zones but without setting a cost, like `Augur of Autumn`

Hanmac:
[...]

PR #9427, opened 2026-01-06
Update Fires of Invention

Eradev:


Hanmac:
Technically, you could use it to cast spells you don't control.

But only in combination with other MayPlay Effects

tool4ever:
imo better to try my suggestion in #6954

Hanmac:
@tool4ever we probably need a cleaner refactor with #4700

PR #9429, opened 2026-01-06
ImageKeys: Fix image loading for sets with modern folder names

### LINKED ISSUE #9408: Adventure mode shops fail to load card images on first visit for sets with legacy Code2 values
**Describe the bug**
In Adventure mode, roughly 1/3 of card images fail to display when first entering a shop. Closing and reopening the shop causes all images to load correctly. This only affects cards from older sets that have legacy Code2 values defined in their edition files (e.g., Darksteel DST→DS, Apocalypse APC→AP, Legions LGN→LE).

Adventure mode shops appear to look in the legacy code folder (e.g., DS/) on first load, rather than the modern code folder (e.g., DST/). If images are stored only in the modern folder (as Scryfall Scrapper and presumably other third-party downloaders do), the first lookup fails. The setlookup files exist but don't seem to work unless the legacy folder actually exists on disk.

Evidence:
Cards from newer sets (no Code2) load fine on first visit
Cards from older sets (with Code2) consistently fail on first visit
Creating symlinks from legacy codes to modern folders (e.g., DS → DST) fixes the issue completely
A second computer with nearly empty legacy folders (like 1 image each) did not exhibit the bug

**To Reproduce**
1. Fresh Forge install with images downloaded via Scryfall Scrapper (uses modern set codes like DST/, APC/, LGN/)
2. Start new Adventure mode game
3. Visit shops - cards from older sets show blank
4. Close and reopen shop - images appear

**Expected behavior**
The setlookup system should find images in modern code folders even when legacy folders don't exist.

**Screenshots**

 - OS: Windows 11
 - Browser: Desktop app, but using Chrome
 - Version 2.0.08-SNAPSHOT-01.01

N/A

**Additional context**
Workaround:
Create symlinks from legacy codes to modern code folders. (From Powershell:)
cmd /c mklink /D LE LGN

Dragynrain:
## Summary
When images are downloaded via Scryfall, they are stored in folders using modern set codes (e.g., DST for Darksteel). However, the image lookup system only checked legacy Code2 folders (e.g., DS), causing images to not load on first access in Adventure mode shops.

This fix extends `hasImage()` to check multiple folder locations:
1. Legacy Code2 folder (e.g., DS/) - checked first, wins for duplicates
2. Modern code folder (e.g., DST/)
3. Alias folder (if defined)

The fix also handles mixed setups where images may be split across multiple folders, merging contents into the cache while preserving priority for existing legacy folder images.

Fixes #9408

## Test Plan
- [x] Tested with only modern code folders (DST/) - images load on first shop visit
- [x] Tested with symlinks from legacy to modern folders - still works
- [x] Tested ~15 Adventure mode shops with older sets - all loaded correctly


Dragynrain:
Anything I can do to get this moving along?

Hanmac:
I don't really like adding more fallbacks

I would prefer to clean up with a more cleaner design (see my other issues)

Jetz72:
I think supporting fallback paths for the game's resources makes sense regardless of how it's organized later. If we change the storage location for images at a later date, we'd want a fallback regardless so that we don't force every user to redownload all their card images. Some users have countless images accumulated over time, others manually install hi-res image packs, others still use custom card images to reskin their cards. At minimum it'd be good to have fallback support, and ideally we'd want fallback support that migrates files to the new storage location automatically.

Dragynrain:
Anything I can do to move this along?

Hanmac:
The bigger question is why `getSetFolder` is used to return `getCode2ByCode` and if the Code2 there should not be used anymore

jamincollins:
## Findings
#### 1. Alias and code folder check skips duplicate folder names but does not verify CardEdition mapping consistency
The code builds foldersWithImages by checking setFolder (Code2), then ed.getCode(), then ed.getAlias(). The condition `!alias.equalsIgnoreCase(setFolder) && (code == null || !alias.equalsIgnoreCase(code))` prevents adding duplicate folder names to the list, which is good. However, if a set has both code and alias pointing to same physical folder as setFolder, the logic may still create inconsistent editionAlias map earlier. The editionAlias map is populated from ed.getAlias() and ed.getCode() without checking if those folders actually contain images. This can lead to stale alias entries causing hasImage to believe an edition has images when it does not.

Test case:
```java
// Simulate CardEdition with alias "DST" and code "DS" where setFolder is "DS"
// but only alias folder exists, not setFolder
PaperCard pc = new PaperCard();
pc.setEdition("DST");
String setFolder = ImageKeys.getSetFolder("DST"); // returns "DS"
CardEdition ed = StaticData.instance().getEditions().get(setFolder);
boolean editionHasImage = ImageKeys.editionImageLookup.get(pc.getEdition());
assert editionHasImage == true; // may be true due to alias folder detection
```

Impact: Incorrect positive hasImage result may cause UI to assume image exists, leading to missing image placeholders later.

#### 2. f.list() may return null and files array not checked for empty
In the new loop, `String[] files = f.list(); if (files == null) continue;` guards against null. Good. However, the code then iterates over files without checking if filename is null. `f.list()` can return empty array, fine. More concerning: filename.split("\\.")[0] will throw ArrayIndexOutOfBounds if filename is empty string. Unlikely but possible with malformed files.

Test case:
```java
File f = new File("/tmp/emptyfile");
File[] files = f.listFiles(); // returns null if not directory
// filename = ""
String cardName = filename.split("\\.")[0]; // throws
```

Impact: NPE or AIOOBE during startup scanning, potentially aborting cache preload.

#### 3. cachedCards key collision when different folders contain same filename
The key is built as `setFolder + "/" + filename.replace(".fullborder", ".full").replace(".jpg", "").replace(".png", "")`. The folder used in the File value is the actual folder, but key uses setFolder. If two different physical folders (code vs alias) contain the same filename, the first wins for cachedCards. This is intended but the setFolderContent HashSet merges names, so hasImage will return true if any folder has the card. However, the cached file path points to the first folder, which may be legacy Code2, not modern folder. If legacy folder is missing the specific card but modern folder has it, the cachedCards entry will be missing because first folder wins for that key, but setFolderContent will contain the name, leading to hasImage true but getImage returns null.

Test case:
```java
// folder A (Code2) has "LightningBolt.jpg", folder B (modern) has "LightningBolt2.jpg"
// key for both uses setFolder, first folder wins
ImageKeys.cachedCards.put("DS/LightningBolt", fileA);
ImageKeys.cachedCards.put("DS/LightningBolt2", fileB); // fine
// If fileA missing but fileB present for same card name, hasImage true, getImage fails
```

Impact: Silent image load failure for mixed folder setups.

#### 4. editionImageLookup populated before verifying folder content actually contains relevant cards
editionHasImage is set to `!foldersWithImages.isEmpty()`. It does not verify that the folders contain any images for the specific set. If a folder exists but is empty or contains only non-image files, editionHasImage will be true incorrectly. FileUtil.isDirectoryWithFiles may check for files but not for image files.

Test case:
```java
// CACHE_CARD_PICS_DIR + code is a directory with only .txt files
boolean has = FileUtil.isDirectoryWithFiles(path); // true
// editionHasImage set true, but later loop skips non jpg/png
```

Impact: Backwards compatibility break: previously editionHasImage false for empty folders, now true.

#### 5. editionAlias populated unconditionally before folder check, causing stale alias entries
The editionAlias map is populated at lines 419-432 before foldersWithImages check. If ed is null, editionAlias not updated. But if ed exists, alias/code added to editionAlias regardless of whether those folders contain images. Later code may rely on editionAlias for other lookups. The PR does not update editionAlias usage elsewhere, but changing hasImage logic to consider multiple folders may cause inconsistent state with other parts of ImageKeys that expect editionAlias to reflect actual image availability.

Test case:
```java
// StaticData has edition with alias, but alias folder removed
// editionAlias still contains alias, causing other code to try loading from missing folder
```

Impact: Interaction with ImageUtil and other image resolution paths may cause unnecessary file checks.

### Performance and security
#### 6. Multiple FileUtil.isDirectoryWithFiles checks per edition cause I/O amplification
For each edition first accessed, code now performs up to three directory existence checks (setFolder, code, alias) plus file listing for each existing folder. Previously only one check. For large sets, this increases startup I/O. No caching of folder existence results beyond editionImageLookup.

Test case:
```java
ImageKeys.hasImage(pc);
// With 500 editions, 3x checks -> noticeable delay
```

Impact: Startup performance regression.

#### 7. Path traversal risk via filename
Filename comes from directory listing, not user input, so risk low. However, key construction uses setFolder + "/" + filename.replace... No sanitization for ".." in filename. If an attacker can write files to image cache directory, they could create ".." filenames leading to path confusion in cachedCards File objects.

Test case:
```java
// Malicious file named "../secret.png" in cache dir
String key = setFolder + "/" + filename.replace...
File value = new File(CACHE_CARD_PICS_DIR + folder + "/" + filename);
// value points outside intended dir
```

Impact: Potential path traversal if cache dir is writable.

## Verdict
Request Changes

The PR introduces multi-folder image lookup to support modern set folder names, but several edge cases and correctness issues remain. The logic for merging folder contents and caching file paths can lead to hasImage returning true while actual image retrieval fails, especially in mixed legacy/modern setups. Null safety and I/O amplification also need attention.

## Recommended Fixes
1. Verify folder content contains at least one image file before marking editionHasImage true; filter FileUtil check to ensure jpg/png presence.
2. Guard filename.split and cardName extraction against empty strings; skip malformed filenames.
3. Ensure cachedCards key maps to correct folder path when first folder lacks specific card but later folder has it; consider per-folder caching or fallback lookup.
4. Move editionAlias population after confirming folder existence to avoid stale aliases.
5. Cache directory existence results to avoid repeated I/O; consider lazy loading.

---

PR #9437, opened 2026-01-07
Add ante support to Adventure mode

vanja-ivancevic:
## Summary

- Adds **Play for Ante** and **Match Ante Rarity** checkboxes to Adventure mode settings
- Adventure mode now respects the global ante preferences (previously hardcoded as disabled)
- Won cards are added to collection, lost cards are removed after each match

## Details

Ante was already fully implemented in the game engine and used by Quest mode. Adventure mode simply hardcoded `rules.setPlayForAnte(false)`. This PR makes it configurable via the existing global preferences.

### Changes

**SettingsScene.java**
- Added two checkboxes for `UI_ANTE` and `UI_ANTE_MATCH_RARITY` preferences (uses existing localized strings)

**DuelScene.java**
- `setPlayForAnte()` and `setMatchAnteRarity()` now read from preferences instead of hardcoded values
- Added ante result processing in `GameEnd()` to transfer cards between player collection based on match outcome

Works well in testing.

image

### DIFF
--- a/forge-gui/res/adventure/Shandalar Old Border/maps/map/cave/cave_multilevel/cave_16BR3U.tmx
+++ b/forge-gui/res/adventure/Shandalar Old Border/maps/map/cave/cave_multilevel/cave_16BR3U.tmx
@@ -72,7 +72,7 @@
     "type": "item",
     "count": 1,
     "probability": 0.1,
-    "itemName": "Spine Whip"
+    "itemName": "Spine Staff"
   },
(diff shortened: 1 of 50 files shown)


tehdiplomat (dismissed):

vanja-ivancevic:
Last update: made the ante card filter conditional on the ante setting. So ante cards should available in shops/rewards when ante is enabled

PR #9462, opened 2026-01-08
Spry and Mighty (ECL)

Fulgur14:
It doesn't seem that LeastPower exists in this context, but I think it should, as it's an obvious extension. This construction should allow to choose two creatures at once and still get the correct difference. I'm putting it alone since it might be complicated.

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -3676,6 +3676,7 @@ public class AbilityUtils {
             return diffCMC.size();
         }
 
+        // shortcut to filter from Defined directly
         if (def.startsWith("Valid")) {
             final String[] splitString = def.split("/", 2);
             String valid = splitString[0].substring(6);


forge-gui/res/cardsfolder/upcoming/spry_and_mighty.txt
```diff
@@ -0,0 +1,10 @@
+Name:Spry and Mighty
+ManaCost:4 G
+Types:Sorcery
+A:SP$ ChooseCard | Choices$ Creature.YouCtrl | ChoiceZone$ Battlefield | SubAbility$ DBDraw | RememberChosen$ True | MinAmount$ 2 | Amount$ 2 | SpellDescription$ Choose two creatures you control. You draw X cards and the chosen creatures get +X/+X and gain trample until end of turn, where X is the difference between the chosen creatures' powers.
+SVar:DBDraw:DB$ Draw | NumCards$ X | SubAbility$ DBPump
+SVar:DBPump:DB$ Pump | Defined$ Remembered | NumAtt$ +X | NumDef$ +X | SubAbility$ DBCleanup
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+SVar:X:Count$Valid Creature.IsRemembered$GreatestPower/Minus.Y
```
tool4ever: you can use RememberedFirst & RememberedLast

Fulgur14: Hm, those are barely used. I can't glean the syntax for this case...

tool4ever: ah sorry, I need to think of something else

tool4ever: while you could also just plainly chain two ChooseCard I think maybe extra support will be good future-proofing

Fulgur14: Yes, I quite like this solution :)

Issue #9453, opened 2026-01-08
Keyword texts generated by Type


Hanmac:
Related to #9418 

Hexproof from uses `Hexproof:<Valid>:<Desc>`
Currently, there is:
* `Hexproof:Card.Red:red` to make `Hexproof from red`
* `Hexproof:Card.Instant:instants` to make `Hexproof from instants`

In case of ColorWords, CardTypes (and SubTypes?), i want to simplify the logic so that it can work from Quality, and generate the Valid and Desc from there.

There is some logic already in KeywordWithType:
https://github.com/Card-Forge/forge/blob/58f56db850627e3b53d16503599be19a0930b59a/forge-game/src/main/java/forge/game/keyword/KeywordWithType.java#L34-L49

What stuff i need:
* the part for the new `getTitle` function returning `Hexproof from <>` that is currently created in Card: (this might be plural for CardTypes)
https://github.com/Card-Forge/forge/blob/bdb3dd15642708f9b7f08be50aebd755180db12f/forge-game/src/main/java/forge/game/card/Card.java#L2551-L2562
* the part of Valid and ValidDesc, currently only used for Enchant stuff, but might be used for KeywordFactory
 https://github.com/Card-Forge/forge/blob/58f56db850627e3b53d16503599be19a0930b59a/forge-game/src/main/java/forge/game/card/Card.java#L7143-L7144
* the part used in reminder text, might be different from the ValidDesc (see Outlaw)

The `Valid:Desc` variant should still work for more complex ones, but `Hexproof:red` and `Hexproof:Artifact` should be enough for simple versions.
This might help other Keywords, too like Protection.

In later work, we might use this to differ if a keyword is affected by TextChanges instead of creating a new one.

Hanmac:
Next one is `KeywordWithCostAndType` with "Splice onto" and TypeCycling
i will do some special logic with Affinifycycling to get "card with affinity" in the reminder text

Hanmac:
Looks like `UndefinedKeyword` is a leftover from a fallback and should be able to be removed

Hanmac:
@tool4ever now that all KeywordInterface have `getTitle`, we might remove some of the `instanceof` casts in Card.java for the Keyword Description

Hanmac:
I think that's Completed for now

Hanmac:
Found some more keyword cases i want to address:

`Avenger of the Fallen` needs Desc for Mobilize:
[...]

PR #9480, opened 2026-01-10
Emerge: refactor keyword

Hanmac:
Part of #9453

Makes Emerge extend from KeywordWithCost as default, and adds only Type as Optional

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -2575,14 +2575,8 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
                         }
                     }
                     sbLong.append("\r\n");
-                } else if (keyword.startsWith("Emerge")) {
-                    final String[] k = keyword.split(":");
-                    sbLong.append(k[0]);
-                    if (k.length > 2) {
-                        sbLong.append(" from ").append(k[2].toLowerCase());
-                    }
-                    sbLong.append(" ").append(ManaCostParser.parse(k[1]));
-                    sbLong.append(" (").append(inst.getReminderText()).append(")");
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/keyword/Emerge.java
```diff
@@ -1,13 +1,48 @@
...
+import forge.game.cost.Cost;
+
+public class Emerge extends KeywordWithCost {
+    protected String type = null;
+    protected String descType = null;
+
+    public String getValidType() { return type == null ? "Creature" : type; }
+    public String getTypeDescription() { return descType; }
+
+    public String getTitle() {
```
tool4ever, line 15: you can probably cleanup `keywordsToText ` with this?

forge-game/src/main/java/forge/game/keyword/Emerge.java
```diff
@@ -1,13 +1,48 @@
 package forge.game.keyword;
 
-public class Emerge extends KeywordWithCostAndType {
+import java.util.Locale;
+
+import forge.card.CardType;
+import forge.game.cost.Cost;
+
+public class Emerge extends KeywordWithCost {
```
tool4ever, line 8: you forgot to map it?

forge-game/src/main/java/forge/game/keyword/Emerge.java
Hanmac: that's the point of #9453 just doing smaller steps

forge-game/src/main/java/forge/game/keyword/Emerge.java
Hanmac: the diff is ugly, but Emerge class was already mapped, i just changed the parent class

PR #9483, opened 2026-01-11
Utilize remove from deck via ante results when tearing up a blacker l…

tehdiplomat:
…otus

This also fixes ante being correctly removed from Adventure decks when losing cards.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
+++ b/forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
@@ -1540,27 +1540,37 @@ public class AdventurePlayer implements Serializable, SaveFileContent {
         return copiesUsed;
     }
 
-    public void removeLostCardFromPools(PaperCard card, int leftInPool) {
+    public void removeLostCardFromPools(PaperCard card) {
         if (card.isVeryBasicLand() && !card.isFoil()) {
             return;
         }
 
-        // This is all copied from the QuestController's similar method.
+        int leftInPool = Current.player().getCards().count(card) - 1;
+
         for (final Deck deck : decks) {
-            int cntInMain = deck.getMain().count(card);
-            int cntInSb = deck.has(DeckSection.Sideboard) ? deck.get(DeckSection.Sideboard).count(card) : 0;
-            int nToRemoveFromThisDeck = cntInMain + cntInSb - leftInPool;
+            int cntInDeck = deck.count(card);
+            int nToRemoveFromThisDeck = cntInDeck - leftInPool;
             if (nToRemoveFromThisDeck <= 0) {
-                continue; // this is not the deck you are looking for
+                continue;
             }
 
-            int nToRemoveFromSb = Math.min(cntInSb, nToRemoveFromThisDeck);
-            if (nToRemoveFromSb > 0) {
-                deck.get(DeckSection.Sideboard).remove(card, nToRemoveFromSb);
-                nToRemoveFromThisDeck -= nToRemoveFromSb;
-                if (nToRemoveFromThisDeck <= 0) {
-                    continue; // done here
+            for(DeckSection section : DeckSection.values()) {
+                if (section == DeckSection.Main) {
+                    continue; // handled later
                 }
+                int cntInSection = deck.get(section).count(card);
+                int nToRemoveFromSection = Math.min(cntInSection, nToRemoveFromThisDeck);
+                if (nToRemoveFromSection > 0) {
+                    deck.get(section).remove(card, nToRemoveFromSection);
+                    nToRemoveFromThisDeck -= nToRemoveFromSection;
+                    if (nToRemoveFromThisDeck <= 0) {
+                        break;
+                    }
+                }
+            }
+
+            if (nToRemoveFromThisDeck <= 0) {
+                continue;
             }
 
             deck.getMain().remove(card, nToRemoveFromThisDeck);
--- a/forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
+++ b/forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
@@ -118,9 +118,8 @@ public class DuelScene extends ForgeScene {
                     Current.player().addCard(card);
                 }
                 for (PaperCard card : anteResult.lostCards) {
-                    int remaining = Current.player().getCards().count(card) - 1;
-                    // Clean up existing decks to remove
-                    Current.player().removeLostCardFromPools(card, remaining);
+                    // We could clean this up by trying to combine all the lostCards into a mapping, but good enough for now
+                    Current.player().removeLostCardFromPools(card);
                 }
             }
         } catch (Exception e) {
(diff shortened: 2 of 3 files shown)


forge-game/src/main/java/forge/game/Match.java
```diff
@@ -397,18 +396,22 @@ private void executeAnte(Game lastGame) {
                 outcome.addAnteWon(registered, gainedPaperOwnership);
             }
 
+            if (!getRules().useAnte()) {
+                continue;
+            }
```
Jetz72, line 401: I assume this check is to prevent ownership changes for any cards that somehow ended up anted in a game with no ante - possibly if some ante card ends up in the game via something like Discord, Lord of Disharmony. But it doesn't look like it'll stop a GainOwnership effect like Tempest Efreet. Maybe throw a `useAnte` check in `Player.changeOwnership`?

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
```diff
@@ -1540,6 +1540,33 @@ public int getCopiesUsedInDecks(PaperCard card) {
...
+            }
+
+            int nToRemoveFromSb = Math.min(cntInSb, nToRemoveFromThisDeck);
+            if (nToRemoveFromSb > 0) {
+                deck.get(DeckSection.Sideboard).remove(card, nToRemoveFromSb);
+                nToRemoveFromThisDeck -= nToRemoveFromSb;
+                if (nToRemoveFromThisDeck <= 0) {
+                    continue; // done here
+                }
+            }
```
Jetz72, line 1570: Might want to generalize the sections it searches, rather than just main and side. With Commander format in Adventure, the tragic scenario where your commander somehow gets lost to an ante would remove it from the collection but not the deck list.

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
```diff
@@ -1540,6 +1540,33 @@ public int getCopiesUsedInDecks(PaperCard card) {
...
+
+            deck.getMain().remove(card, nToRemoveFromThisDeck);
+        }
+        Current.player().getCards().remove(card, 1);
```
Jetz72, line 1578: Why only remove one when the method supposedly supports removing multiple?

forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
```diff
@@ -111,21 +111,16 @@ public void GameEnd() {
...
+            GameOutcome.AnteResult anteResult = hostedMatch.getGame().getOutcome().getAnteResult(humanPlayer);
+            if (anteResult != null) {
+                for (PaperCard card : anteResult.wonCards) {
+                    Current.player().addCard(card);
+                }
+                for (PaperCard card : anteResult.lostCards) {
+                    int remaining = Current.player().getCards().count(card) - 1;
+                    // Clean up existing decks to remove
+                    Current.player().removeLostCardFromPools(card, remaining);
                 }
```
Jetz72, line 123: You might include a limit that prevents them from losing more copies of a card than they had in the deck they played. Otherwise it could remove copies of placeholder cards (i.e. Wastes) from their inventory when they weren't in the deck.

As an alternative, we could instead add `isCollectible` to `PaperCardFlags` then set that when padding out the deck, to be passed to the in-game card. Or we could move the functionality for padding Adventure decks to inside the game setup so we can flag the generated cards there. Those seem like overkill for something like this though.

forge-game/src/main/java/forge/game/Match.java
tehdiplomat: This was moved from line 93 of this same file, where it used to wrap the entire execute ante call. Now that it's also being used to remove Blacker Lotus we're just shortcircuiting the call.

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
tehdiplomat: Mostly because the anteResult that we're looping over is an array and not a CardPool or mapping, so each card only represents a single thing to be removed.

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
tehdiplomat: I didn't add ante to Adventure mode, so I'm not sure how that interaction was added, but IIRC ante is only pulled from your library. I guess its possible to lose your commander that was present in a different deck though. That would definitely be a sad panda.

forge-game/src/main/java/forge/game/Match.java
Jetz72: Yeah, but it looks like `Player.changeOwnership` (used by `OwnershipGainEffect`, and in turn used by a handful of cards with "remove if not using ante rules" on them) puts cards in the same lostOwnership and gainedOwnership lists that you're utilizing for Blacker Lotus.

So with this implementation, ante disabled, and some method to leak ante-only cards into the game regardless, you'd get:
* Blacker Lotus - Ownership removed regardless, as it should be.
* Jeweled Bird (Normal ante-only card) - Moves cards to Ante zone; does not actually change ownership. Seems sensible.
* Tempest Efreet (`GainOwnership` ante-only card) - Previously, this would exchange "ownership" only within the game and cards would return to their true owners afterward. Now it will exchange ownership outside the game as well, as if ante was enabled.

This is the card technically working as written, but seems inconsistent when the other ante-only cards don't result in actual ownership changes when ante is disabled.

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
Jetz72: A misclick or having insufficient life when it gets targeted by a Bronze Tablet would do it. It's also possible to have your commander in your library if it gets sent there and you choose not to send it back to the command zone for whatever reason.

forge-game/src/main/java/forge/game/Match.java
tehdiplomat: Both of those cards get removed from decks if ante is not enabled, so I'm not sure it's relevant?

Jetz72: The same can be said of the rest of this method - if ante isn't enabled nothing should be in anyone's ante zones. I assumed you kept the `useAnte` check in order to plan around the edge case where some kind of ante card or effect makes it into a game where ante is disabled - either via some kind of Wish or Conjure effect, something custom, some weird Adventure circumstance, for the sake of future-proofing, or due to 'Discord, Lord of Disharmony' who I'm pretty sure can bring them into any game

tehdiplomat: Added the check for ownership permanence into Player.changeOwnership only if ante is on

forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
tehdiplomat: I don't think this is blocking, I'm not super concerned about that scenario.

Jetz72 (approved):
Seems good to me; I can't think of any other weird edge cases worth considering right now.

Issue #9484, opened 2026-01-11
Brutal Cathar + Slaughter Specialist.


FrostBlizzaga:
**Describe the bug**
Quite simply, when Brutal Cathar has exiled an opposing Slaughter Specialist, the Slaughter Specialist receives a +1/+1 counter when the Brutal Cathar is killed. This should be impossible because of Brutal Cathar's ruling that neither the exiled creature, nor the Brutal Cathar would exist on the field at the same time to stop creatures that enter as a copy of another creature for example.

**To Reproduce**
1. This bug happened in an Innistrad Midnight Hunt draft in MTG Forge Adventure mode.

**Expected behavior**
When the Brutal Cathar is killed, the slaughter specialist should not receive a counter when it re-enters.

**Screenshots**

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/6845eeb2-476b-4d53-9fce-045d33b9bf2e" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/20d43a3b-286b-4125-a790-fd9f484d3b00" />

**Desktop (please complete the following information):**
 - OS: Windows 10 home 
 - OS Build 19045.6466
 - Browser: Chrome
 - Version: 22H2

PR #9491, opened 2026-01-11
Only register LTB trigger that are in lastStateBattlefield

tool4ever:
Part for #9484


Hanmac (approved):

Hanmac:
@tool4ever is this ready?

tool4ever:
It fixes the trigger part.

The other aspect is more tricky, I might try to come up with a follow-up later:
if _Clone_ is exiled and returns while a boardwipe is resolving its RE can still copy permanents that didn't LTB (in the engine) yet.

Hanmac:
We might write this for later

PR #9495, opened 2026-01-12
Add console command to reveal cave and dungeon names

Kiousu:
Summary
  - Add a new adventure console command: `reveal caves dungeons`.
  - Marks all cave/dungeon POIs as visited so their names show up in the map “Names” view.


forge-gui-mobile/src/forge/adventure/stage/ConsoleCommandInterpreter.java
```diff
@@ -338,6 +338,27 @@ private ConsoleCommandInterpreter() {
             System.out.println("POI Names - Types\n" + String.join("\n", poiNames));
             return "POI lists dumped to stdout.";
         });
+        registerCommand(new String[]{"reveal", "caves", "dungeons"}, s -> {
```
nickxiv: This could be rewritten to take in the PointOfInterest type as an argument instead of being hardcoded to caves and dungeons. This way we could reveal cities and capitals as well.

Kiousu: I've updated the function so that it takes POI type as an argument, but Towns and Capitals don't show names on the map, so while it will "reveal" them, there is no update on the map. The command now covers Castles, though.

forge-gui-mobile/src/forge/adventure/stage/ConsoleCommandInterpreter.java
```diff
@@ -338,6 +340,50 @@ private ConsoleCommandInterpreter() {
...
+                }
+                if (!normalized.isEmpty()) {
+                    types.add(normalized);
+                }
+            }
+            if (types.isEmpty()) {
+                return "Please specify at least one PointOfInterest type (e.g. \"reveal cave\", \"reveal dungeon\", \"reveal town\", \"reveal capital\").";
+            }
+            int revealed = 0;
+            for (PointOfInterest poi : save.getWorld().getAllPointOfInterest()) {
```
nickxiv, line 370: This looks like it's currently iterating over every POI, regardless of if it matches the type passed in. Could enhance performance by filtering these down to just the ones that match the passed in type(s).

Kiousu: I can refactor this for readability, but that won’t change performance: we still need to inspect all POIs at least once without a precomputed type index. To actually reduce iteration cost, we’d need to add a typed lookup/index in world POI storage, which is a larger design change. If that already exists and I missed it, please let me know.

forge-gui-mobile/src/forge/adventure/stage/ConsoleCommandInterpreter.java
```diff
@@ -338,6 +340,50 @@ private ConsoleCommandInterpreter() {
...
+                return "Please specify at least one PointOfInterest type (e.g. \"reveal cave\", \"reveal dungeon\", \"reveal town\", \"reveal capital\").";
+            }
+            HashSet<String> types = new HashSet<>();
+            for (String type : s) {
+                String normalized = type.toLowerCase();
+                if (normalized.endsWith("ies")) {
+                    normalized = normalized.substring(0, normalized.length() - 3) + "y";
+                } else if (normalized.endsWith("s") && normalized.length() > 1) {
+                    normalized = normalized.substring(0, normalized.length() - 1);
+                }
```
Jetz72, line 358: I think including logic for normalizing plurals and synonyms is a bit excessive for a developer console.

Kiousu: I don’t see this as adding “smart” logic to the console — just making it less brittle.
The previous review itself referred to the town POI as a “city,” which highlights the core issue: even contributors don’t reliably remember the exact internal identifiers. Requiring exact string matches makes the console fragile and forces people to look up data definitions. 
The normalization here is intentionally minimal (simple plurals and a few explicit aliases like city → town). It’s not general NLP — just a thin layer to prevent avoidable friction.

Jetz72: > The previous review itself referred to the town POI as a “city,” which highlights the core issue: even contributors don’t reliably remember the exact internal identifiers. Requiring exact string matches makes the console fragile and forces people to look up data definitions.

The previous review wasn't from a contributor. As far as I can tell, that was their first comment on any Forge PR or issue. 

I generally agree that having to look these kinds of parameters up is annoying, but haven't you resolved that by including the list of examples in your error message?

Kiousu: I see. My mistake.

That’s fair. I'll remove the normalization when I get the chance.

PR #9496, opened 2026-01-12
Fix AI unalives self with Shanid (#9123)

### LINKED ISSUE #9123: AI unalives self with Shanid
Image
It could have won. Sad.

CTimmerman:
Fixes #9123 by checking for _Shanid, Sleepers' Scourge_'s life loss and card draw before playing legendary lands or casting legendary spells.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -44,6 +44,7 @@ import forge.game.cost.*;
 import forge.game.keyword.Keyword;
 import forge.game.mana.ManaCostBeingPaid;
 import forge.game.phase.PhaseType;
+import forge.game.player.GameLossReason;
 import forge.game.player.Player;
 import forge.game.player.PlayerActionConfirmMode;
 import forge.game.player.PlayerCollection;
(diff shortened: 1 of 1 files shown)


forge-ai/src/main/java/forge/ai/AiController.java
```diff
@@ -987,6 +998,9 @@ public AiPlayDecision canPlaySa(SpellAbility sa) {
             }
         }
         if (sa.isSpell()) {
+            if (card.getType().isLegendary() && wouldLoseWithShanid()) {
+                return AiPlayDecision.CurseEffects;
+            }
             return canPlaySpellOrLandBasic(card, sa);
```
tool4ever, line 1004: you're pretty close to the right util methods already:
https://github.com/Card-Forge/forge/blob/a3b6b15c17dbd5aefe28e68fe495dd1093943841/forge-ai/src/main/java/forge/ai/AiController.java#L1003-L1010

please modify those instead:
that way you should also be able to leverage `findSubAbilityByType` and avoid ugly hardcoded logic

CTimmerman: So add lifeloss to `getDamageForPlaying` or make a new similar function and also for drawing? `getLossForDrawing`?

tool4ever: check both methods, you'll see lifeloss is already handled there

and there is no loss for drawing here, better not mix up too much or you risk making things hard to merge again

CTimmerman: You told me not to rebase, and I've only added one commit to an existing MR of mine. Keep your vagueness and DIY. It works for me now and my branch works for my decks. I'm tired of waiting for even approved MRs.

PR #9500, opened 2026-01-12
handlePaid: count any different values

tool4ever:


forge-game/src/main/java/forge/game/card/CardUtil.java
```diff
@@ -314,26 +314,24 @@ private static Set<String> getReflectableManaColors(final SpellAbility abMana, f
...
+                if (ab.isSpell() || ab.isLandAbility()) {
                     continue;
                 }
                 if (maxChoices == colors.size()) {
                     break;
                 }
 
+                if (!parents.contains(ab.getHostCard())) {
+                    parents.add(ab.getHostCard());
+                }
```
Hanmac, line 326: @tool4ever you should do this as extra MR
you are mixing too many changes into one again

PR #9503, opened 2026-01-12
KeywordWithCostAndType: change Splice and TypeCycling

Hanmac:
Part of #9453


tool4ever:
The reminder now uses "one" instead of "a"
and for the Affinity one it's missing either

tool4ever:
Splice now says: "card spell"

PR #9505, opened 2026-01-13
Fix Ashling, the Limitless

tehdiplomat:
AffectedZone$ Stack seems to be too late to work for the engine?

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/ashling_the_limitless.txt
+++ b/forge-gui/res/cardsfolder/upcoming/ashling_the_limitless.txt
@@ -2,7 +2,7 @@ Name:Ashling, the Limitless
 ManaCost:2 R
 Types:Legendary Creature Elemental Sorcerer
 PT:2/3
-S:Mode$ Continuous | Affected$ Elemental.Permanent+wasCastFromYourHandByYou | AffectedZone$ Stack | AddKeyword$ Evoke:4 | Description$ Elemental permanent spells you cast from your hand have evoke {4}. (If you cast a spell for its evoke cost, it's sacrificed when it enters.)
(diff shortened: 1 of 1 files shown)


Hanmac:
She needs an Engine update anyway.
> Elemental permanent spells you cast from your hand gain evoke {4} **as you cast them**. (If you cast a spell for its evoke cost, it's sacrificed when it enters.)

[Update bulletin](https://magic.wizards.com/en/news/announcements/lorwyn-eclipsed-update-bulletin)

forge-gui/res/cardsfolder/upcoming/ashling_the_limitless.txt
```diff
@@ -2,7 +2,7 @@ Name:Ashling, the Limitless
 ManaCost:2 R
 Types:Legendary Creature Elemental Sorcerer
 PT:2/3
-S:Mode$ Continuous | Affected$ Elemental.Permanent+wasCastFromYourHandByYou | AffectedZone$ Stack | AddKeyword$ Evoke:4 | Description$ Elemental permanent spells you cast from your hand have evoke {4}. (If you cast a spell for its evoke cost, it's sacrificed when it enters.)
+S:Mode$ Continuous | Affected$ Permanent.Elemental | AffectedZone$ Hand | AddKeyword$ Evoke:4 | Description$ Elemental permanent spells you cast from your hand have evoke {4}. (If you cast a spell for its evoke cost, it's sacrificed when it enters.)
```
tool4ever: this would also affect your opponents

Hanmac:
It still isn't perfect, because now it works when it is cast by opponent

Hanmac:
@tool4ever we need to debug why Henzie Blitz works, while this didn't

Issue #9504, opened 2026-01-13
Crash when opening custom cube dialog


vkordic:
**Describe the bug**
When trying to start a Custom Cube draft, the program crashes and needs to be killed through Task Manager.
It does not happen every time.

This is the stack trace:
```
Language 'en-US' loaded successfully.
(ThreadUtil first call): Running on a machine with 16 cpu core(s)
Firdoch Core has neither ManaCost nor Color
Read cards: 32158 archived files in 1 ms (25 parts) using thread pool
Upcoming set Marvel Super Heroes (MSH) dated in the future. All `upcoming` cards will be added to this set with unknown rarity.
The card A-Baba Lysaga, Night Witch was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card A-Divide by Zero was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card A-Esika's Chariot was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card A-Faceless Haven was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card A-Fires of Invention was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card A-Goldspan Dragon was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card A-Hullbreaker Horror was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card A-Kumano Faces Kakkazan was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card A-Luminarch Aspirant was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card A-Monster Manual was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card A-The Meathook Massacre was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card Blinding Radiance was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card Goblin Bruiser was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
The card Ogre Painbringer was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder. 
[...]

PR #9542, opened 2026-01-16
Fix race condition in CardEdition.getCardInSet()

fediazgon:
The `getCardInSet()` method uses lazy initialization for `cardsInSetLookupMap` without synchronization. When `BoosterDraft.loadCustomDrafts()` processes multiple draft files in parallel using `CompletableFuture.supplyAsync()`, multiple threads can call `getCardInSet()` on the same `CardEdition` instance simultaneously. 

```java
List<CompletableFuture<?>> futures = new ArrayList<>();                                                                                                                                                  
for (final String element : dList) {                                                                                                                                                                     
    if (element.endsWith(FILE_EXT)) {                                                                                                                                                                    
        futures.add(CompletableFuture.supplyAsync(()-> { // HERE                                                                                                                                                 
            final List<String> dfData = FileUtil.readFile(ForgeConstants.DRAFT_DIR + element);                                                                                                           
            queue.add(CustomLimited.parse(dfData, FModel.getDecks().getCubes()));                                                                                                                        
            return null;                                                                                                                                                                                 
        }).exceptionally(ex -> {                                                                                                                                                                         
            ex.printStackTrace();                                                                                                                                                                        
            return null;                                                                                                                                                                                 
        }));                                                                                                                                                                                             
    }                                                                                                                                                                                                    
}                                                                                                                                                                                                        
CompletableFuture.allOf(futuresArray).join(); 
```
                                                                                                                                                                                                           
 This causes `TreeMap` internal structure corruption with errors like: 
                                                                                                                                 
 ``` java                                                                                                                                                                                                    
 Caused by: java.lang.NullPointerException: Cannot read field "left" because "r" is null                                                                                                                 
     at java.base/java.util.TreeMap.rotateLeft(TreeMap.java:2578)                                                                                                                                        
     at java.base/java.util.TreeMap.fixAfterInsertion(TreeMap.java:2645)                                                                                                                                 
     ...                                                                                                                                                                                                 
     at forge.card.CardEdition.getCardInSet(CardEdition.java:437)                                                                                                                                        
 ```                                                                                                                                                                                                     
The solution is to apply double-checked locking pattern.

Closes https://github.com/Card-Forge/forge/issues/9504.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/BlightAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/BlightAi.java
@@ -1,23 +1,105 @@
 package forge.ai.ability;
 
-import java.util.Map;
-import java.util.Optional;
-
+import forge.ai.AiAbilityDecision;
+import forge.ai.AiPlayDecision;
 import forge.ai.ComputerUtilCard;
 import forge.ai.SpellAbilityAi;
+import forge.game.ability.AbilityUtils;
 import forge.game.card.Card;
+import forge.game.card.CardCollection;
+import forge.game.card.CardLists;
 import forge.game.card.CounterEnumType;
+import forge.game.keyword.Keyword;
 import forge.game.player.Player;
+import forge.game.player.PlayerCollection;
+import forge.game.player.PlayerPredicates;
 import forge.game.spellability.SpellAbility;
 import forge.util.StreamUtil;
 
+import java.util.Map;
+import java.util.Optional;
+
 public class BlightAi extends SpellAbilityAi {
 
-    protected Card chooseSingleCard(Player ai, SpellAbility sa, Iterable<Card> options, boolean isOptional, Player targetedPlayer, Map<String, Object> params) {
-        Optional<Card> filtered = StreamUtil.stream(options).filter(c -> !c.canReceiveCounters(CounterEnumType.M1M1)).findAny();
-        if (filtered.isPresent()) {
-            return filtered.get();
+    @Override
+    protected AiAbilityDecision checkApiLogic(Player ai, SpellAbility sa) {
+        return canPlayWithTargeting(ai, sa, /*mandatory=*/ false);
+    }
+
+    @Override
+    protected AiAbilityDecision doTriggerNoCost(final Player ai, final SpellAbility sa, final boolean mandatory) {
+        return canPlayWithTargeting(ai, sa, mandatory);
+    }
+
+    private AiAbilityDecision canPlayWithTargeting(Player ai, SpellAbility sa, boolean mandatory) {
+        if (!sa.usesTargeting()) {
+            return new AiAbilityDecision(100, AiPlayDecision.WillPlay);
         }
+
+        sa.resetTargets();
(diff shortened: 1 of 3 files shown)


Hanmac:
i wonder if it would be better to have `cardsInSetLookupMap` be created in the constructor
using Stream and [Multimaps::toMultimap](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Multimaps.html#index-java.lang.Iterable-com.google.common.base.Function-)

```java
cardsInSetLookupMap = cardStream.collect(
  toMultimap(
    e -> e.name,
    e -> e,
    MultimapBuilder.treeKeys(String.CASE_INSENSITIVE_ORDER).arrayListValues()::build)
  )
);
```

And getAllCardsInSet() can just return `cardsInSetLookupMap.values()` ?

fediazgon:
> i wonder if it would be better to have `cardsInSetLookupMap` be created in the constructor
> using Stream and [Multimaps::toMultimap](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Multimaps.html#index-java.lang.Iterable-com.google.common.base.Function-)

Ok, I thought this lazy initialization was important but it seems, during card database initialization, every card iterates through every edition calling getCardInSet(). This means the lookup map gets initialized for all editions during startup anyway. Let's initialize it in the constructor then.

Fixed in https://github.com/Card-Forge/forge/commit/3454c225b5f8e77f88357d223fc91c948e79a507

Hanmac (dismissed):

Hanmac:
@tool4ever @tehdiplomat what do you think?
or should that optimized more?

Hanmac:
@fediazgon 
i don't know if we should do this in this MR or a later one,
but the Constructors might be clean-up more

Currently, it has 3 Constuctors:
* the big one that takes ListMultimaps:
* the extra one that takes native Array
* and the one that takes Datum Values (and a native Array)

The first one is the one used in the Reader for reading the Edition files:
https://github.com/Card-Forge/forge/blob/2b563fbe353b6e6e397588542e70f9798c2ae348/forge-core/src/main/java/forge/card/CardEdition.java#L770

the second one isn't used external, but only by the third one

the third one is used for
* the Unknown one:
https://github.com/Card-Forge/forge/blob/2b563fbe353b6e6e397588542e70f9798c2ae348/forge-core/src/main/java/forge/card/CardEdition.java#L276
* some user bucket:
https://github.com/Card-Forge/forge/blob/2b563fbe353b6e6e397588542e70f9798c2ae348/forge-core/src/main/java/forge/card/CardEdition.java#L919-L920

Each one with an empty card list → that means, the third constructor can be updated to remove that unused extra Param, and might remove the second constructor too.

(i might try a smaller MR that can be merged faster)

fediazgon:
I think having this one as its own change allows us to link this PR in the future in case there are other race conditions lying around as a way of documenting what to do. If we add to many changes it can be easy to miss the point of what the problem and the fix were. Also, this is currently a big issue for anyone that is playing custom cube drafts since, when I tested it, it was failing 1/4 times.

fediazgon:
@Hanmac Let's merge yours and this one. We can also just merge this one since it includes both commits. We will just show as 2 contributors for the same (squashed) commit.

PR #9553, opened 2026-01-16
Does Machines and The Last Ronin (TMT)

Fulgur14:


Fulgur14:
Not sure if the "EQ4" in The Last Ronin (taken from Kraven's Last Hunt) is correct. What if something like The Water Crystal increases the number of milled cards? Should the immediate trigger check for GE4 instead?

forge-gui/res/cardsfolder/upcoming/the_last_ronin.txt
```diff
@@ -0,0 +1,14 @@
...
+SVar:DBImmediateTrig:DB$ ImmediateTrigger | Execute$ TrigChangeZone | ConditionDefined$ Remembered | ConditionPresent$ Card | ConditionCompare$ EQ4 | SubAbility$ DBCleanup | TriggerDescription$ When you do, return target creature card from your graveyard to your hand.
+SVar:TrigChangeZone:DB$ ChangeZone | ValidTgts$ Creature.YouOwn | TgtPrompt$ Select target creature card in your graveyard | Origin$ Graveyard | Destination$ Hand
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+SVar:DBEffect:DB$ Effect | Triggers$ TriggerAttack | SpellDescription$ Whenever a creature you control attacks alone this turn, put three +1/+1 counters on it. It gains trample, lifelink, and indestructible until end of turn.
+SVar:TriggerAttack:Mode$ Attacks | ValidCard$ Creature.YouCtrl | Alone$ True | Execute$ TrigPutCounter | TriggerZones$ Battlefield | TriggerDescription$ Whenever a creature you control attacks alone this turn, put three +1/+1 counters on it. It gains trample, lifelink, and indestructible until end of turn.
```
tool4ever: bad TriggerZones

Fulgur14: Does that mean Origin of the Hidden Ones is wrong?

tool4ever: just remove both, the engine takes care of it anyway

forge-gui/res/cardsfolder/upcoming/the_last_ronin.txt
```diff
@@ -0,0 +1,14 @@
+Name:The Last Ronin
+ManaCost:4 B G
+Types:Enchantment Saga
+K:Chapter:3:DBDestroyAll,DBMill,DBEffect
+SVar:DBDestroyAll:DB$ DestroyAll | ValidCards$ Creature | SpellDescription$ Destroy all creatures.
+SVar:DBMill:DB$ Mill | Defined$ You | NumCards$ 4 | RememberMilled$ True | SubAbility$ DBImmediateTrig | SpellDescription$ Mill four cards. When you do, return target creature card from your graveyard to your hand.
+SVar:DBImmediateTrig:DB$ ImmediateTrigger | Execute$ TrigChangeZone | ConditionDefined$ Remembered | ConditionPresent$ Card | ConditionCompare$ EQ4 | SubAbility$ DBCleanup | TriggerDescription$ When you do, return target creature card from your graveyard to your hand.
```
tool4ever: to include changes from mill replacements you can use TriggerAmount$ instead

Fulgur14: Like this?

PR #9558, opened 2026-01-17
Fix for refundPayment when multiple costs

fediazgon:
In cards like [Champion of the Weird](http://scryfall.com/card/ecl/95/champion-of-the-weird) with multiple costs, the player might start paying some costs but then it might cancel others (e.g., Bligh 2, cancel Pay 1 life). In this scenario, Blight 2 is not refunded because Blight 2 is not an undoable cost.

I don't think rolling back an ability should care about undoable costs, since we are just trying to reset the game to a state previous to ability being used.

### DIFF
--- a/forge-game/src/main/java/forge/game/cost/CostPayment.java
+++ b/forge-game/src/main/java/forge/game/cost/CostPayment.java
@@ -123,8 +123,10 @@ public class CostPayment extends ManaConversionMatrix {
     public final void refundPayment() {
         Card sourceCard = this.ability.getHostCard();
         for (final CostPart part : this.paidCostParts) {
-            if (part.isUndoable()) {
-                part.refund(sourceCard);
+            part.refund(sourceCard);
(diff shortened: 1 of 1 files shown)


tool4ever:
We've more or less accepted that manually reverting more complex costs is a lost cause.
As soon as something messes with the outcome (e.g. in this case a _Vorinclex, Monstrous Raider_) the path to the past becomes way too uncertain.

There's kind of a workaround but among other things it costs some performance: `EXPERIMENTAL_RESTORE_SNAPSHOT`
See also this issue: #4437

The other role for `isUndoable()` is `IGameController.undoLastAction()`, but that's only relevant after something resolved (just mana abilities to be precise).

Ok - with that background info for you out of the way I don't mind merging this:
imo costs that are always undoable should just not override `refund()` anyway

tool4ever (approved):

Hanmac:
i want to maybe reuse the PaymentDescision to undo the costs, but that might be for much later

PR #9564, opened 2026-01-18
Don & Leo, Problem Solvers (TMT)

Fulgur14:
I'm putting this separately because it's a bit of an experiment and I have to leave for the Prerelease soon, so I don't have time to test it right now.

I was thinking that maybe we could do simultaneous zone changes with several different targeting types like this; I'm not satisfied with implementation of, say, Pull from the Deep, beause it seems it would cause extra triggers for Attuned Hunter and similar cards.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/don_leo_problem_solvers.txt
+++ b/forge-gui/res/cardsfolder/upcoming/don_leo_problem_solvers.txt
@@ -5,7 +5,7 @@ PT:4/6
 K:Vigilance
 T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You | TriggerZones$ Battlefield | Execute$ TrigPump | TriggerDescription$ At the beginning of your end step, exile up to one target artifact you control and up to one target creature you control. Then return them to the battlefield under their owners' control.
 SVar:TrigPump:DB$ Pump | ValidTgts$ Artifact.YouCtrl | TargetMin$ 0 | TargetMax$ 1 | Select target artifact you control | SubAbility$ DBChangeZone
-SVar:DBChangeZone:DB$ ChangeZone | ValidTgts$ Creature.YouCtrl | TargetMin$ 0 | TargetMax$ 1 | Defined$ ThisDefinedAndTgts$ ParentTarget | TgtPrompt$ Select target creature you control | Origin$ Battlefield | Destination$ Exile | RememberChanged$ True | SubAbility$ DBReturn
+SVar:DBChangeZone:DB$ ChangeZone | ValidTgts$ Creature.YouCtrl | TargetMin$ 0 | TargetMax$ 1 | ThisDefinedAndTgts$ ParentTarget | TgtPrompt$ Select target creature you control | Origin$ Battlefield | Destination$ Exile | RememberChanged$ True | SubAbility$ DBReturn
 SVar:DBReturn:DB$ ChangeZone | Defined$ Remembered | Origin$ All | Destination$ Battlefield | SubAbility$ DBCleanup
 SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
 Oracle:Vigilance\nAt the beginning of your end step, exile up to one target artifact you control and up to one target creature you control. Then return them to the battlefield under their owners' control.


forge-gui/res/cardsfolder/upcoming/don_leo_problem_solvers.txt
```diff
@@ -0,0 +1,11 @@
+Name:Don & Leo, Problem Solvers
+ManaCost:3 WU WU
+Types:Legendary Creature Mutant Ninja Turtle
+PT:4/6
+K:Vigilance
+T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You | TriggerZones$ Battlefield | Execute$ TrigPump | TriggerDescription$ At the beginning of your end step, exile up to one target artifact you control and up to one target creature you control. Then return them to the battlefield under their owners' control.
+SVar:TrigPump:DB$ Pump | ValidTgts$ Artifact.YouCtrl | TargetMin$ 0 | TargetMax$ 1 | Select target artifact you control | SubAbility$ DBChangeZone
+SVar:DBChangeZone:DB$ ChangeZone | ValidTgts$ Creature.YouCtrl | TargetMin$ 0 | TargetMax$ 1 | Defined$ Targeted | TgtPrompt$ Select target creature you control | Origin$ Battlefield | Destination$ Exile | RememberChanged$ True | SubAbility$ DBReturn
+SVar:DBReturn:DB$ ChangeZone | Defined$ Remembered | Origin$ All | Destination$ Battlefield | GainControl$ True | SubAbility$ DBCleanup
```
tool4ever: no GainControl

forge-gui/res/cardsfolder/upcoming/don_leo_problem_solvers.txt
```diff
@@ -0,0 +1,11 @@
+Name:Don & Leo, Problem Solvers
+ManaCost:3 WU WU
+Types:Legendary Creature Mutant Ninja Turtle
+PT:4/6
+K:Vigilance
+T:Mode$ Phase | Phase$ End of Turn | ValidPlayer$ You | TriggerZones$ Battlefield | Execute$ TrigPump | TriggerDescription$ At the beginning of your end step, exile up to one target artifact you control and up to one target creature you control. Then return them to the battlefield under their owners' control.
+SVar:TrigPump:DB$ Pump | ValidTgts$ Artifact.YouCtrl | TargetMin$ 0 | TargetMax$ 1 | Select target artifact you control | SubAbility$ DBChangeZone
+SVar:DBChangeZone:DB$ ChangeZone | ValidTgts$ Creature.YouCtrl | TargetMin$ 0 | TargetMax$ 1 | Defined$ Targeted | TgtPrompt$ Select target creature you control | Origin$ Battlefield | Destination$ Exile | RememberChanged$ True | SubAbility$ DBReturn
```
tool4ever: use `ThisDefinedAndTgts` instead

tool4ever: try `ThisDefinedAndTgts$ ParentTarget`

tool4ever: looks like invalid syntax now

Fulgur14: OK, so what was wrong with Defined$ Targeted in the first place? I thought "Targeted" will take all targets, including the parent ones?

tool4ever: not every effect is prepared to handle Defined and ValidTgts at the same time

Fulgur14: Hm, so what should I do now?

tool4ever: test it

Fulgur14: OK, I tested it and it works -- The War Doctor correctly only gets one counter when the ability fires.

Issue #9532, opened 2026-01-15
Turn CardTrait Changes and Keyword Changes into Interfaces/records


Hanmac:
Similar to CardType changes in #9375

Especially the stuff with `updateBasicLandAbilities` should be moved into CardState.
https://github.com/Card-Forge/forge/blob/2c0c65677a50475c5d67bbc2dd759da2d88a4d15/forge-game/src/main/java/forge/game/card/Card.java#L3522-L3536

And the two places that check for `hasRemoveIntrinsic`, should be handled by a different kind of extra class (second implementation of the new Interface)
https://github.com/Card-Forge/forge/blob/2c0c65677a50475c5d67bbc2dd759da2d88a4d15/forge-game/src/main/java/forge/game/card/Card.java#L5023-L5032

I don't know yet if i want to move the implementation of the changes into the new classes/records like i did with the CardType changes:
https://github.com/Card-Forge/forge/blob/2c0c65677a50475c5d67bbc2dd759da2d88a4d15/forge-game/src/main/java/forge/game/card/Card.java#L3466-L3483

https://github.com/Card-Forge/forge/blob/2c0c65677a50475c5d67bbc2dd759da2d88a4d15/forge-game/src/main/java/forge/game/card/Card.java#L7150-L7155

https://github.com/Card-Forge/forge/blob/2c0c65677a50475c5d67bbc2dd759da2d88a4d15/forge-game/src/main/java/forge/game/card/Card.java#L7193-L7198

https://github.com/Card-Forge/forge/blob/2c0c65677a50475c5d67bbc2dd759da2d88a4d15/forge-game/src/main/java/forge/game/card/Card.java#L7216-L7221

If possible, i want KeywordChanges/KeywordCollection to inherit the CardTraitChanges Interface, because Keywords have Card Traits inside.

PR #9570, opened 2026-01-19
MagicColor: add LandType to Color

Hanmac:
Part of #8682 and #9532 

This adds the basic land type to ColorEnum
to be used for later changes

### DIFF
--- a/forge-core/src/main/java/forge/card/MagicColor.java
+++ b/forge-core/src/main/java/forge/card/MagicColor.java
@@ -154,16 +154,16 @@ public final class MagicColor {
         COLORLESS(Constant.COLORLESS, MagicColor.COLORLESS, "C", null, "lblColorless");
 
         private final String name, shortName, symbol;
-        private final String basicLand;
+        private final String basicLandType;
         private final String label;
         private final byte colormask;
 
-        Color(String name0, byte colormask0, String shortName, String basicLand, String label) {
+        Color(String name0, byte colormask0, String shortName, String basicLandType, String label) {
             name = name0;
             colormask = colormask0;
             this.shortName = shortName;
             symbol = "{" + shortName + "}";
-            this.basicLand = basicLand;
+            this.basicLandType = basicLandType;
             this.label = label;
         }
 
@@ -196,8 +196,8 @@ public final class MagicColor {
             return shortName;
         }
 
-        public String getBasicLand() {
-            return basicLand;
+        public String getBasicLandType() {
+            return basicLandType;
         }
 
         @Override


Jetz72:
Perhaps call the field "basicLandType" if Colorless doesn't point to Wastes? Might make sense at some point to split some of these constructor parameters out to EnumMaps. The signature is getting a little cluttered.

fediazgon:
Is there any rationale with not making these `enums`? There are a bunch of places in the code where we do things like:

```java
saMana.hasParam("ReplaceOnly")
```

where this could have been an enum. If you need a display name, you can also do:

```java
public enum BasicLandType {
    PLAINS("Plains"),
    ISLAND("Island"),
    SWAMP("Swamp"),
    MOUNTAIN("Mountain"),
    FOREST("Forest");

    private final String displayName;

    BasicLandType(String displayName) {
        this.displayName = displayName;
    }

    public String getDisplayName() {
        return displayName;
    }
}
```

Jetz72:
> Is there any rationale with not making these `enums`? There are a bunch of places in the code where we do things like:
> 
> ```java
> saMana.hasParam("ReplaceOnly")
> ```

I'm not sure the example here illustrates the potential upside from such an enum. I don't think we have too many parameters or fields that we'd want to be exclusive to basic land types.

fediazgon:
The main upside I see is discoverability. With something like saMana.hasParam(Param.REPLACE_ONLY), autocomplete makes it much easier to see what parameters are valid. With raw strings, unless you already know the exact value, it’s hard to tell what hasParam can take without searching the codebase.

Note: this is just an example to explain why I think `String basicLand` should also be an enum.

tehdiplomat:
Who is going to be responsible for converting all of the scripts into enums? And then keeping those up to date when new scripts are written?

Hanmac:
the enum discussion is not part of this MR

Hanmac:
Also part of #9532
I want to clean up the part of `updateBasicLandAbilities` using Color→LandType

Hanmac:
@Jetz72 renamed to `basicLandType`

Hanmac:
Ping @tool4ever for your opinion

tool4ever:
Well kind of difficult to consider such refactor when it's still only at the state that adds more unused code 🤷‍♂️

personally I'd prefer to just reuse `CardType.Constant.BASIC_TYPES` if possible

Hanmac:
> personally I'd prefer to just reuse `CardType.Constant.BASIC_TYPES` if possible

but that's the point, i want to get rid of these Constants in the future.
(especially the Snow one)

Also `CardName VS BasicLandType`
this new info is about the LandType, not the CardName

PR #9565, opened 2026-01-18
Fix BlightAi not setting target opponent on some abilities.

fediazgon:
Added checkApiLogic to properly target opponents with blightable creatures.

Closes https://github.com/Card-Forge/forge/issues/9554

Tested with 

```txt
[metadata]
Name:New Puzzle
URL:https://www.cardforge.org
Goal:Win
Turns:1
Difficulty:Easy
Description:Win this turn.
[state]
turn=1
activeplayer=p1
activephase=UPKEEP
p0life=20
p0landsplayed=0
p0landsplayedlastturn=0
p0numringtemptedyou=0
p0speed=0
p0hand=
p0library=Swamp|Set:ECL|Art:1
p0graveyard=
p0battlefield=Blooming Stinger|Set:EOE|Art:1;Blooming Stinger|Set:EOE|Art:1
p0exile=
p0command=
p0sideboard=
p1life=20
p1landsplayed=0
p1landsplayedlastturn=0
p1numringtemptedyou=0
p1speed=0
p1hand=
p1library=Swamp|Set:ECL|Art:1
p1graveyard=Krenko's Command|Set:J25|Art:1
p1battlefield=Champion of the Weird|Set:ECL|Art:1
p1exile=
p1command=
p1sideboard=
```

PR #9571, opened 2026-01-19
Blossombind fixes

tool4ever:
> If the cost of an ability or an additional cost of a spell requires untapping or putting counters on a creature enchanted with Blossombind, that cost can't be paid. If a resolving spell or ability says that a player may untap or put counters on that creature, that player can't choose to do so.

### DIFF
--- a/forge-game/src/main/java/forge/game/GameActionUtil.java
+++ b/forge-game/src/main/java/forge/game/GameActionUtil.java
@@ -932,14 +932,6 @@ public final class GameActionUtil {
         return completeList;
     }
 
-    public static void checkStaticAfterPaying(Card c) {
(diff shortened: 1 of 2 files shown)


Hanmac:
@tool4ever BlightEffect / Blight Cost probably needs the same update

there you can somewhat remove parts of #9565 

Like for chooseCard, you shouldn't be able to choose a card you can't put counters on.
(the list should already be filtered by the Effect, right?)

tool4ever:
I'm still waiting on CR update for that
last few days the Wizards link is just 404 🤦‍♂️

it matters if the full wording will be "choose a creature you control, then put..." vs. the reminder
I don't think it's likely but then again such things happened before 🙈

Issue #9479, opened 2026-01-10
AI can't play Karn's Bastion


eBaug1:
The AI will use the Proliferate ability of this land during upkeep as soon as it has the mana for it, even if it makes sense or not, and will do so indefinently is my experience. It will only play other spell if it has enough mana after activating this ability. No mention at the start of the game that the AI has issues with this card.

<img width="1920" height="1024" alt="Image" src="https://github.com/user-attachments/assets/f72853b4-5dc2-45a6-83ea-573b495a35b2" />

fediazgon:
"Makes sense" is very subjective. In this case is proliferating to increase the counters in Llanowar Reborn. Looking at the AI code, it should never proliferate if no counters are added. I can try to improve it a little bit so it takes into account the mana cost.

PR #9573, opened 2026-01-20
Fix AI proliferate to consider mana cost vs value

fediazgon:
AI now calculates the total value of proliferating based on what's on the battlefield (planeswalkers worth 3, creatures with positive counters worth 1, opponent poison worth 2, etc.) and only activates proliferate if the value meets or exceeds the mana cost. This prevents the AI from wasting mana on low-value proliferates like spending 4 mana on Karn's Bastion just to add one +1/+1 counter.

Closes https://github.com/Card-Forge/forge/issues/9479

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/CountersProliferateAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/CountersProliferateAi.java
@@ -4,6 +4,7 @@ import com.google.common.collect.Lists;
 import forge.ai.*;
 import forge.game.GameEntity;
 import forge.game.card.*;
+import forge.game.player.GameLossReason;
 import forge.game.player.Player;
 import forge.game.spellability.SpellAbility;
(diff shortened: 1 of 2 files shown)


forge-ai/src/main/java/forge/ai/ability/CountersProliferateAi.java
Hanmac, line 67: Add check if opponent has 9 counters and can lose by Poison `o.cantLoseCheck(GameLossReason.Poisoned)`
if yes, then the calculation rating should return 100 (or more)

fediazgon: Fixed in https://github.com/Card-Forge/forge/pull/9573/commits/091bc99f366b5b96da18e169dcef202d9c6966f7

Hanmac (approved):

Issue #9574, opened 2026-01-20
Dependency Layer 3


Hanmac:
Stuff that affects Text Change Layer: 

* CardType
  * TypeWordChange
  * StaticLayer GainTextOf

* TextBox (CardTraits)
  * TypeWordChange
  * ColorWordChange
  * StaticLayer GainTextOf
  * TextBoxExchange

* Color
  * StaticLayer GainTextOf
  * Incorporate

* ManaCost
  * TypeWordChange (over additional Cost, which should be traits)
  * ColorWordChange (over additional Cost, which should be traits)
  * StaticLayer GainTextOf
  * Incorporate

* PT
  * StaticLayer GainTextOf
  * TextBoxExchange

First Section, CardType Changes:

For the small example of CardType Changes, i added #9375 to be to handle the TypeWordChange as Map.
But in reality, it should have handled them as Word per Word changes.

Right now, it `addChangedCardTypesByText` calls `addEmpty` on both TypeWordChange and ColorWordChange, even if the color words shouldn't be affected.
That is because it was made for `GainTextOf` before `TextBoxExchange` was introduced.

As a first attempt, I'm going to refactor the CardTypeChanges to not rely on the finished TypeWordChange Map, but single Word → Word changes instead. Then the `GainTextOf` Effect can better reset the entire CardType without the need for extra `addEmpty`, and it can be handled by `addChangedCardTraitsByText` for now.

Hanmac:
I will try to combine TypeWordChange and ColorWordChange into a single Object instead of new Map<String, String>

then this combined Text Change object should handle the WordChangedType, "Island to Mountain"
but also text changes from the CardTraits

the bigger problem right now is to inject the text box changes

PR #9577, opened 2026-01-20
Card: Combine Card type Changes by text

Hanmac:
Part of #9574

this handles both Card Type changes By Text (Shapeshifter and TypeWord) into one TreeMap
(to be updated for dependency)

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -4463,7 +4463,6 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
 
     public final void addNewPTByText(final Integer power, final Integer toughness, final long timestamp, final long staticId) {
         newPTText.put(timestamp, staticId, Pair.of(power, toughness));
-        updatePTforView();
     }
     public final boolean removeNewPTbyText(final long timestamp, final long staticId) {
         return newPTText.remove(timestamp, staticId) != null;
@@ -4989,7 +4988,6 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
         this.changedTextColors.addEmpty(timestamp, staticId);
         this.changedTextTypes.addEmpty(timestamp, staticId);
         updateChangedText();
-        updateAbilityTextForView();
     }
 
     public final CardTraitChanges addChangedCardTraits(Collection<SpellAbility> spells, Collection<SpellAbility> removedAbilities,
--- a/forge-game/src/main/java/forge/game/GameAction.java
+++ b/forge-game/src/main/java/forge/game/GameAction.java
@@ -1234,6 +1234,7 @@ public class GameAction {
         if (affectedPerLayer.containsKey(StaticAbilityLayer.TEXT)) {
             affectedPerLayer.get(StaticAbilityLayer.TEXT).forEach(Card::updateNameforView);
             affectedKeywords.addAll(affectedPerLayer.get(StaticAbilityLayer.TEXT));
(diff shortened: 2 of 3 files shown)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4993,6 +4984,11 @@ public final void addChangedCardTraitsByText(Collection<SpellAbility> spells,
         changedCardTraitsByText.put(timestamp, staticId, new CardTraitChanges(
             spells, null, trigger, replacements, statics, true, false
         ));
+
+        // setting card type via text, does overwrite any other word change effects?
+        this.changedTextColors.addEmpty(timestamp, staticId);
```
tool4ever, line 4988: why move this here?
imo too late now (cache update already happened)

Hanmac: Different cache

CardType cache doesn't use `changedTextTypes` anymore

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -5536,10 +5532,15 @@ public final void removeChangedTextColorWord(final Long timestamp, final long st
      */
     public final void addChangedTextTypeWord(final String originalWord, final String newWord, final Long timestamp, final long staticId) {
         changedTextTypes.add(timestamp, staticId, originalWord, newWord);
+        changedCardTypesByText.put(timestamp, staticId, new WordChangedType(originalWord, newWord));
+        updateTypeCache();
```
tool4ever: cache update is redundant

Hanmac: adding to changedCardTypesByText does need to update the TypeCache

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4210,18 +4203,16 @@ public void setChangedCardKeywords(Table<Long, Long, KeywordsChange> changedCard
...
 
-    public final void addChangedCardTypesByText(final CardType addType, final long timestamp, final long staticId) {
-        changedCardTypesByText.put(timestamp, staticId, new CardChangedType(addType, null, false,
-                EnumSet.of(RemoveType.SuperTypes, RemoveType.CardTypes, RemoveType.SubTypes)));
-
-        // setting card type via text, does overwrite any other word change effects?
-        this.changedTextColors.addEmpty(timestamp, staticId);
-        this.changedTextTypes.addEmpty(timestamp, staticId);
-
-        this.updateChangedText();
```
tool4ever, line 4221: updateChangedText isn't called anywhere else
need to have it somewhere
maybe in Continuous instead?

forge-game/src/main/java/forge/game/card/Card.java
tool4ever: if you call `updateChangedText` anyway you will update it ~10 lines later
makes no sense

forge-game/src/main/java/forge/game/card/Card.java
Hanmac: > updateChangedText isn't called anywhere else

updateChangedText called by:
* addChangedTextColorWord
* removeChangedTextColorWord
* addChangedTextTypeWord
* removeChangedTextTypeWord
* updateCloneState

tool4ever: I mean `addChangedCardTypesByText` is only for the Shapeshifter
and now it's not getting updated

Hanmac: yes?
The Shapeshifter adds a CardType change with this one
and it calls updateTypeCache

updateChangedText needed to be called by addChangedCardTraitsByText

PR #9590, opened 2026-01-22
Implement AI profile picker per AI player

benhunter:
This PR implements an AI profile selector in the lobby that allows the human player to choose the AI profile for each AI opponent. Currently the game uses the same AI profile for all AI opponents in multiplayer games.

image

Human players will not see the AI profile selector:
image

This feature will be disabled by default until it is ready for full integration. It can be enabled from the Game Settings under Preferences > Advanced Settings > Experimental AI Picker.

image

The AI profile picker is not yet implemented for the mobile GUI.

This work will be helpful for future efforts to support more AI options (including alternate AI systems such as model-based).


forge-gui/pom.xml
```diff
@@ -82,6 +82,18 @@
             <artifactId>jetty-servlet</artifactId>
             <version>${jetty.version}</version>
         </dependency>
+        <dependency>
```
tool4ever: imo test should be moved to desktop module like all others so such dependencies can stay in just one POM

Agetian: Yes, agreed, otherwise a nice contribution!

benhunter: That makes sense.

I attempted to move the test to forge-gui-desktop, but could not get the static mocks to work in that package. I tried a couple other approaches, but the static mocks are difficult. It could be a test config issue or some version incompatibility. The test packages look a bit dated and maybe could use an overhaul.

Regardless, I removed the test and test dependencies from forge-gui.

thelamesaucegit:
Would be excellent to expand this to include a new sim command flag in the CLI to specify AI profiles per player in simulated matches. 
I've just started investigating what it would take to implement this within my own fork of the code, but if progress is already being made in this respect I may just hold off.

Jetz72:
> I've just started investigating what it would take to implement this within my own fork of the code, but if progress is already being made in this respect I may just hold off.

I'm not aware of any further efforts on this front currently. Feel free to submit a PR.

PR #9591, opened 2026-01-22
Simplify LobbyPlayer.equals() return logic

benhunter:


Jetz72:
Could simplify it even further with an `Objects.equals`.

Hanmac:
Also, `Objects.hash` for the `hashCode` function.
(but i can do that later)

benhunter:
> Could simplify it even further with an `Objects.equals`.

Great point - I removed the override altogether. Tests pass and game runs as expected.

Hanmac:
> > Could simplify it even further with an `Objects.equals`.
> 
> Great point - I removed the override altogether. Tests pass and game runs as expected.

No, we meant use [Objects.equals](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-)

For `Objects.equals(this.name, other.name)`

benhunter:
> > > Could simplify it even further with an `Objects.equals`.
> > 
> > 
> > Great point - I removed the override altogether. Tests pass and game runs as expected.
> 
> No, we meant use [Objects.equals](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-)
> 
> For `Objects.equals(this.name, other.name)`

Right, I see the difference. IntelliJ warned about requiring a class check, so the code would look something like this:

```
    @Override
    public boolean equals(Object obj) {
        if (obj.getClass() != this.getClass()) {
            return false;
        }
        return Objects.equals(this, obj);
    }
```

Does this behave differently (for our purpose) from allowing `Object.equals()` to run without override?

Hanmac:
The difference is avatarIndex and the other properties should be ignored

benhunter:
> The difference is avatarIndex and the other properties should be ignored

I think it's there now :)

Issue #9583, opened 2026-01-21
isManaAbility for CardTraits, because of Triggered mana abilities


Hanmac:
While looking at this part for #9532
https://github.com/Card-Forge/forge/blob/2c0c65677a50475c5d67bbc2dd759da2d88a4d15/forge-game/src/main/java/forge/game/card/Card.java#L3467-L3476
and thinking on how to optimize it, like using a `Predicate`, 
I noticed that it doesn't respect "Triggered mana abilities",

so the two cards that use it (maybe unify the property name?)
* `Town-Razer Tyrant` with `RemoveNonManaAbilities` 
* `Blood Sun` with `ExceptManaAbilities`

Should, in the rare chance that something with a "Triggered mana ability" somehow got turned into a non-basic land,
have them keep these Triggers.

Hanmac:
done with the MRs for now

PR #9592, opened 2026-01-22
Trigger: add isManaAbility

Hanmac:
Part of #9583

to be used later in ChangedCardTraits

### DIFF
--- a/forge-game/src/main/java/forge/game/spellability/SpellAbility.java
+++ b/forge-game/src/main/java/forge/game/spellability/SpellAbility.java
@@ -403,10 +403,11 @@ public abstract class SpellAbility extends CardTraitBase implements ISpellAbilit
             return false; //Loyalty ability, not a mana ability.
         }
         // CR 605.1b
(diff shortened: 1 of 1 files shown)


Hanmac:
@tool4ever please take a look at this so i can fix the CardTraitChanges over the weekend

forge-game/src/main/java/forge/game/spellability/SpellAbility.java
```diff
@@ -402,7 +403,10 @@ public final boolean isManaAbility() {
             return false; //Loyalty ability, not a mana ability.
         }
         // CR 605.1b
-        if (isTrigger() && this.getTrigger().getMode() != TriggerType.TapsForMana && this.getTrigger().getMode() != TriggerType.ManaAdded) {
+        if (isTrigger() && !TriggerType.TapsForMana.equals(getTrigger().getMode()) && !TriggerType.ManaAdded.equals(getTrigger().getMode())) {
+            return false;
+        }
+        if (!isActivatedAbility()) {
```
tool4ever: imo better if you just commit to this branch

merging it like this would mean breaking the current way it checks triggers, e.g. _Tyvar the Bellicose_ will no longer see all

Hanmac: fixed

PR #9599, opened 2026-01-22
Tie equipment loadouts to active deck

vanja-ivancevic:
- Equipment loadouts are now tied to each deck
- Switching decks automatically saves current equipment and restores the new deck's loadout
- Backward compatible with existing saves (loadouts start empty, get populated on first deck switch)

Works well so far in testing.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
+++ b/forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
@@ -72,7 +72,7 @@ public class AdventurePlayer implements Serializable, SaveFileContent {
     private final ArrayList<ItemData> inventoryItems = new ArrayList<>();
     private final Array<Deck> boostersOwned = new Array<>();
     private final HashMap<String, Long> equippedItems = new HashMap<>();
-    private final HashMap<Integer, HashMap<String, Long>> deckLoadouts = new HashMap<>();
+    private final ArrayList<HashMap<String, Long>> deckLoadouts = new ArrayList<>();
     private final List<AdventureQuestData> quests = new ArrayList<>();
     private final List<AdventureEventData> events = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


Jetz72:
Seems like the kind of feature that should have to be enabled by a setting. There are users who keep 10+ decks, and if one of those people gets a better set of boots, it'll be a bit annoying having to go through and update the loadout for each of them. Having it default to off would also ensure the user understands that equipment is now tied to their active deck. Otherwise I suspect we'd encounter plenty of users that load up a game, switch decks, wander into battle without realizing all their equipment fell off, then file a bug report because they don't understand why it happened.

vanja-ivancevic:
Great point! I will update it as a setting, default "off".

vanja-ivancevic:
Hey, any chance we could revisit this? Thanks a lot :)

Jetz72:
Got a merge conflict to resolve. 

Should it clear the stored loadout when the corresponding deck is deleted? Or copy it when a deck is copied?

vanja-ivancevic:
> Got a merge conflict to resolve.
> 
> Should it clear the stored loadout when the corresponding deck is deleted? Or copy it when a deck is copied?

Rebased onto master and resolved the merge conflict in `SettingsScene.java` (just needed both `disableCrackedItems` and `bindEquipmentLoadoutsToDecks` checkboxes).

Also added a new commit to handle deck loadouts on delete, copy, and clear:

- delete deck: clears the stored loadout for that deck (index shifting handled automatically)
- Copy deck: copies the source deck's loadout to the new slot
- Clear deck: clears the stored loadout
- Fixed a bug where `deckLoadouts` wasn't cleared on game reload, which would cause stale loadout data to accumulate

Switched `deckLoadouts` from `HashMap<Integer, ...>` to `ArrayList` that mirrors the `decks` list, which keeps the two in sync naturally.

PR #9604, opened 2026-01-24
Add ECL to Lorwyn quest world

GreySimphonic:
Someone tried to start Lorwyn in Quest Commander subformat and couldn't get a commander. This would expand the reward pool for regular Lorwyn players, allow for more Lorwyn commanders, and if I understand the files correctly leave the original classic fights and challenges intact (just expand their reward possibilities, maybe).

Side note: I only intended to change line 10, and I'm not familiar with how to do public PRs. 70% of the reason I opened this was to try to do a tiny PR and get a feel for it. I have more substantial ones in mind for the future, potentially. And this one was done in the GH GUI, and I wonder if that's what changed line 40 (EOF difference?).


GreySimphonic:
It was raised on Discord that throwing a new set blindly into a group of older sets might throw off the intended balance. I can't speak to that one way or the other, but it sounds reasonable, and like maybe this isn't a good idea. I can't judge that well for myself, going to wait and see if anyone else has feedback, but it seems like there's a good chance the Lorwyn world won't get all the Lorwyn cards (at least for now, without a more substantial PR, maybe).

PR #9613, opened 2026-01-26
remove DeckRecords

benhunter:
Cleanup dead code.


Jetz72:
@tehdiplomat Any idea what this was supposed to be? Looks like it wasn't even used in the commit that introduced it.

tehdiplomat:
Oh i looked this up the other day. Looks like it got added during the times of a PlanarConquest. My guess is the thought was to track how you did vs another deck. Maybe DrDev ended up doing it a different way. @Jetz72

PR #9614, opened 2026-01-26
KeywordChanges: use ICardTraitChanges

Hanmac:
This uses the KeywordChange on the Player,
to be used as CardTraitChange on the Keyword Object,
instead of creating a new one.

### DIFF
--- a/forge-game/src/main/java/forge/game/player/Player.java
+++ b/forge-game/src/main/java/forge/game/player/Player.java
@@ -989,7 +989,9 @@ public class Player extends GameEntity implements Comparable<Player> {
             }
         }
         KeywordsChange cks = new KeywordsChange(kws, removeKeywords, false);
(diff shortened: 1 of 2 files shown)


forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -989,10 +989,7 @@ public final void addChangedKeywords(final List<String> addKeywords, final List<
             }
         }
         KeywordsChange cks = new KeywordsChange(kws, removeKeywords, false);
-        if (!cks.getAbilities().isEmpty() || !cks.getTriggers().isEmpty() || !cks.getReplacements().isEmpty() || !cks.getStaticAbilities().isEmpty()) {
```
tool4ever, line 992: not every KW has traits, so the check made sense?

Hanmac: resolved

PR #9617, opened 2026-01-27
Improving some mana payment aspects for AI payment

tehdiplomat:


Agetian (dismissed):

tool4ever (changes requested):
ok so signets might be usable but it breaks another test?
sorry but I really disagree with creating code paths to fix one thing while making other parts stop working on top of adding an additional variable to bug reports...
and the way this duplicates all the code kind of makes it impossible to identify the additions :/
could you please provide some details so we can understand what changes are happening here?

PR #9620, opened 2026-01-27
Keywords now implement ICardTraitChanges

Hanmac:
Keywords now have "hasTraits" and also implement ICardTraitChanges


tool4ever (approved):

Hanmac:
Found that getChangedCardKeywordsList didn't use the CardState

PR #9633, opened 2026-01-28
YECL: Limitless Rekindling

dracontes:
[As revealed recently](https://www.reddit.com/r/MagicAlchemy/comments/1qoxa32/limitless_rekindling/).
Needs support that feels beyond my current expertise. Tested with `ValidCards$ Instant` to determine that the resolution process does work as expected. 
As far as I can determine `ValidCards$` in the context of  `NameCard` lines doesn't parse comma splits or set codes. Looking into the code of `ChooseCardNameEffect.java` (lines [36-39](https://github.com/Card-Forge/forge/blob/a57b4631adc9abb26358341ca4c5c8937e758b87/forge-game/src/main/java/forge/game/ability/effects/ChooseCardNameEffect.java#L36) and [96-109](https://github.com/Card-Forge/forge/blob/a57b4631adc9abb26358341ca4c5c8937e758b87/forge-game/src/main/java/forge/game/ability/effects/ChooseCardNameEffect.java#L96) ) and comparing with `SacrificeAllEffect.java` (lines [45-52](https://github.com/Card-Forge/forge/blob/a57b4631adc9abb26358341ca4c5c8937e758b87/forge-game/src/main/java/forge/game/ability/effects/SacrificeAllEffect.java#L45); which  is an effect that can parse commas and set codes in `ValidCards$` ) I clearly notice these classes handle `ValidCards$` in ways that seem difficult to reconcile. Still, I followed the latter effect's lead in laying out the putative script herein.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/limitless_rekindling.txt
+++ b/forge-gui/res/cardsfolder/upcoming/limitless_rekindling.txt
@@ -1,0 +1,10 @@
+Name:Limitless Rekindling
+ManaCost:3 R R
+Types:Kindred Sorcery Elemental
+K:Storm
+A:SP$ NameCard | AtRandom$ True | ValidCards$ Instant.setECL,Sorcery.setECL | SubAbility$ DBConjure | StackDescription$ REP Conjure_{p:You} conjures & you may_{p:You} may | SpellDescription$ Conjure a random instant or sorcery card from the Lorwyn Eclipsed expansion into exile. Until end of turn, you may cast that card without paying its mana cost.
+SVar:DBConjure:DB$  MakeCard | Name$ ChosenName | Conjure$ True | Zone$ Exile | RememberMade$ True | SubAbility$ DBEffect
+SVar:DBEffect:DB$ Effect | RememberObjects$ RememberedCard | StaticAbilities$ STPlay | SubAbility$ DBCleanup | ForgetOnMoved$ Exile
+SVar:STPlay:Mode$ Continuous | MayPlay$ True | MayPlayWithoutManaCost$ True | Affected$ Card.IsRemembered+nonLand | AffectedZone$ Exile | Description$ Until end of turn, you may cast the exiled card without paying its mana cost.
+SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
+Oracle:Conjure a random instant or sorcery card from the Lorwyn Eclipsed expansion into exile. Until end of turn, you may cast that card without paying its mana cost.\nStorm


jamincollins:
## Findings
#### 1. ValidCards set filter not supported in random NameCard path
Description
The card uses ValidCards$ Instant.setECL,Sorcery.setECL with AtRandom$ True. The random NameCard path in ChooseCardNameEffect.java line 113-115 uses CardFacePredicates.valid(valid) which only checks type and ManaCost/cmcEQ via hasProperty that delegates to input.getType().hasStringType(v). Set filtering (setECL) is handled by CardProperty.cardHasProperty for Card objects, not ICardFace. Therefore set filter will be ignored and random selection may pick cards from any set.
Test case:
```java
// ChooseCardNameEffect.java:113-115
Predicate<ICardFace> cpp = CardFacePredicates.valid("Instant.setECL");
chosen = StaticData.instance().getCommonCards().streamAllFaces()
        .filter(cpp).collect(StreamUtil.random()).map(ICardFace::getName).orElse("");
// Expected: only ECL instants; Actual: all instants because setECL not checked
```
Impact: Card violates intended restriction, may conjure non-ECL cards.

#### 2. Comma-separated ValidCards not parsed for random path
Description
ValidCards$ Instant.setECL,Sorcery.setECL is comma-separated. ChooseCardNameEffect does not split ValidCards for random path; CardFacePredicates.valid receives whole string "Instant.setECL,Sorcery.setECL". ValidPredicate splits on first dot only, k[0] = "Instant". The remaining part is "setECL,Sorcery.setECL" which is split on + only, so "setECL,Sorcery.setECL" is treated as a single property. hasProperty will call input.getType().hasStringType("setECL,Sorcery.setECL") which returns false, filtering out all cards.
Test case:
```java
Predicate<ICardFace> cpp = CardFacePredicates.valid("Instant.setECL,Sorcery.setECL");
boolean ok = cpp.test(face); // face is Instant from ECL
// Expected true; Actual false because property check fails
```
Impact: Random conjure will always return empty, card does nothing.

#### 3. Set filter via ValidCards uses CardProperty not CardFacePredicates
Description
Set filtering via property starts with "set" is implemented in CardProperty.cardHasProperty (line 1959) which requires Card object and looks up PaperCard by name. For ICardFace random selection, CardFacePredicates.valid never delegates to CardProperty, so set restriction is silently dropped. This is a mismatch between Card-based and Face-based filtering.
Impact: Inconsistent behavior between player-choice and random modes.

#### 4. RememberMade and StaticAbilities may reference wrong card if name collides across sets
Description
MakeCard uses Name$ ChosenName and RememberMade$ True. If chosen name exists in multiple sets, the NameCard effect will conjure the first matching face found by StaticData.instance().getCommonCards().streamAllFaces() filtered by predicate (which ignores set). The RememberMade card will be created with Name$ ChosenName, which resolves via getFaceByName, picking first match. May conjure wrong set version.
Impact: May conjure unintended printing, breaking set restriction.

### Performance and security
#### 5. Random selection over all faces without early set filter
Description
streamAllFaces() loads all card faces into memory and filters via predicate that cannot filter by set. For large data sets this is inefficient, and the predicate is evaluated for every face. With set filter ignored, the search space is larger than intended.
Impact: Performance degradation and incorrect results.

## Verdict
Request Changes
The PR introduces a card that relies on set filtering via ValidCards in random NameCard selection. Current implementation does not support set filtering in CardFacePredicates, and comma-separated ValidCards is not parsed correctly, leading to either silent failure or empty selection.

## Recommended Fixes
1. Change ValidCards to use separate NameCard calls or implement set filtering in CardFacePredicates.valid for ICardFace, e.g., by checking face.getPaperCard().getEdition() or similar.
2. Use two separate ValidCards entries or split comma-separated values before passing to CardFacePredicates.valid, or use ValidCards$ Instant+setECL and duplicate for Sorcery.
3. Alternatively, change AtRandom$ True to use ChooseFromDefinedCards with a pre-filtered list of ECL instants/sorceries, ensuring set restriction is enforced.

---

Issue #5112, opened 2024-04-22
Highlight Auto Payment


tehdiplomat:
Sometimes its unclear what the AI will try to auto pay with before it does so. It would be nice to highlight the auto pay feature so you can know whether or not to utilize it.

Hanmac:
Apropos Auto Payment, sometimes the AI prefers to pay life for _Forgotten Monument_ given ability
instead of using the Mana abilities of _Hidden Courtyard_ instead. (it might depend on the order of the lands)

tehdiplomat:
@Hanmac I think we can just make test cases for things like that and then fix them

tool4ever:
the way I'd envision this:
- 3 level of "AI assisted playing" as user preference
  - **Expert**
  Off
  - **Casual**
  "Highlight entities that could become next user action" (default)
    - this payment idea
    - #9383 already adds some support before that step
    - #9863 also related
    - #10533
  -   **Beginner**
  like level 2 but now the highlighting is limited further based on AI suggestions for good plays
      - this can also include #3210
      - maybe something like "yield until AI thinks I should act" based on #9643
      - potentially even different border colors to reflect differences in `AiAbilityDecision` scoring
      - also `UI_OVERLAY_DRAFT_RANKING`
- when a fresh install is detected (no previous user directory?) show a small onboarding dialog on first start
  - later more preferences can be grouped into appropriate levels with respective defaults 🥳
  - option to experience each level inside game

tehdiplomat:
Yea that sounds like a good full project idea. For expert mode, does that mean the auto-tapper would be disabled? The deck I'm running now seems to prefer tapping my creature lands over a painland (even if it can tap the painland for colorless), because i guess its valuing the optionality of the painland over the combat prowess of a 1/1? 

I haven't made any real progress over any of these ideas, so if Claude wants a shot at them that should be fine too.

PR #9643, opened 2026-01-30
Expanded yield system / auto-pass options

### LINKED ISSUE #3699: Auto Yield does not work in multiplayer
**Describe the bug**
Auto Yield does not work in multiplayer.

**Expected behavior**
In single player vs AI, the Forge interface permits the player to select Auto Yield for stack items. The Auto Yield works as expected, resolving future stack items generated by that ability without player intervention. In multiplayer, I am able to select "Auto Yield" for stack items, but all stack items still require my specific approval. The auto yield does not work.

 - OS: Windows 11
 - Java Version: JRE 8.0.381
 - Forge Version: 1.6.58-SNAPSHOT-08.26

MostCromulent:
Screenshot 2026-04-17 165223
Screenshot 2026-04-17 165214

                                                          
---

**Summary**      

Inspired by recent discussion on the Forge Discord, this PR adds an expanded yield system to reduce click fatigue in multiplayer games and speed up play. Changes are extensively documented in **[Documentation](https://github.com/MostCromulent/forge/blob/NetworkPlay/dev/.documentation/YieldRework-Documentation.md).** Fully backwards compatible with current functionality: the feature is **disabled by default** and must be explicitly enabled in preferences. 

**Problem**

In multiplayer Magic games (3+ players), the current priority system requires excessive clicking:
- Dozens of priority passes every turn in a 4-player game.
- Players must manually pass priority even when they have no possible actions.
- This creates click fatigue and slows down gameplay significantly (particularly if players are inattentive and don't immediately respond to prompts).

**Solution**

Expanded yield options that give players more flexibility to control yield timing and conditions. The aim is to help players spend less time responding to unnecessary priority prompts and let them get to their desired phase of action quicker:

Particularly helpful for _that one friend who won't pay attention to priority prompts in Commander._

- **Yield Options Panel**: Dockable panel with extended yield buttons and visual feedback  (blue=ready, red=active)                                                                        
- **Six Yield Modes**: Next Phase, Before Combat, End Step, End Turn, Your Next Turn, Stack Clears                                     
- **Smart Suggestions**: Prompts to auto-yield when player cannot respond (no mana, no actions, can't respond to stack). Configurable through checkbox options in game taskbar menu.                            
- **Configurable Interrupts**: Automatically cancels yield when attacked, targeted, mass removal cast, etc. Interrupt conditions are highly configurable through checkbox options in game taskbar menu.
- **F-Key Hotkeys**: F1-F6 for yield modes, ESC to cancel.
                                                                            
**Key Design Decisions**

- **GUI-layer only**: No changes to game engine ~~or network protocol~~ (**Edit**: though see comments below about network implementation to avoid desync).
- **Feature-gated**: Default OFF, requires explicit enable + restart
- **Backward compatible**: Existing Ctrl+E behavior unchanged when feature disabled                         

**Testing**

Changes have been tested in local play (thanks to _PotionSTAT_ for testing and feedback). ~~As changes are purely on the GUI layer and do not involve network logic, there is no reason to think testing would not replicate in network use cases.~~ (**Edit:** see comments about network play below).

---
Code authored by Claude Code (Opus 4.5) under human direction and review.


DOCUMENTATION.md
Jetz72, line 1: This file shouldn't be added to the main repo.

forge-gui/src/main/java/forge/gamemodes/match/input/InputPassPriority.java
```diff
@@ -176,4 +288,118 @@ public boolean selectAbility(final SpellAbility ab) {
...
+        }
+
+        if (game.getPhaseHandler().getPlayerTurn().equals(player)) {
+            return false;
+        }
+
+        return !hasAvailableActions(game, player);
+    }
+
+    private boolean hasAvailableActions(Game game, Player player) {
```
tool4ever: some of the design in this PR seems a bit weird to me
but I'm still thinking about how to best phrase those cases

so let's start with an easy one for now:
this function is identical to `canRespondToStack`

DOCUMENTATION.md
tool4ever: it's good for following the AI reasoning process at this stage

but yea when it gets finalized for merge some of the content should land in `docs/` for end users (Wiki)
and non trivial code aspects would be fine as Javadoc only

MostCromulent:
Testing by _PotionSTAT_ has revealed an issue with network play integration not realised before PR. 

YieldController was using gameView.getGame() which returns an unsynchronized dummy object for network clients. Have refactored to use network-safe GameView properties and added StackItemView.getApiType() for mass removal detection. Will also check implications for smart prompts.

MostCromulent:
**Re: Network Code Changes**                                                  
                                                                            
While the yield system was designed to avoid network protocol changes, in testing discovered that smart suggestions in InputPassPriority had a network compatibility issue requiring new TrackableProperties. This is because the logic was checking the `Game` object on the server which doesn't exist on network clients.                   
                                                                            
The solution: Add two TrackableProperties (HasAvailableActions, WillLoseManaAtEndOfPhase) to PlayerView. These piggyback on existing GameView serialization—no protocol changes required. The host calculates these values and they're automatically included when GameView is sent to clients. 

The yield system remains client-local; this ensures the data for yield suggestions is synchronized, while yield decisions stay independent per client. 

The alternative to making these network changes would have been implementing a more basic heuristic for the yield prompts which would have been a worse and experience for the player.

**Edit:  Claude AI's explanation for why server needs to know client yield mode state**

The server needs to track the client's yield mode to detect end conditions because the client doesn't have access to the same context as the server, which has the authoritative game state and can reliably detect when yield end conditions are met across all players.

When client activates a yield mode it uses `notifyYieldModeChanged` to notify the server. The server then monitors game state and checks yield end conditions in `YieldController.java`. When end condition is met the server calls `clearYieldMode()` which triggers `syncYieldModeToClient()` to notify the client.

Both server and client stay in sync about yield state. If only the client tracked it, multiplayer edge cases could cause the client to yield too long or not long enough.

DOCUMENTATION.md
MostCromulent: Have committed draft wiki documentation for review.

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -1759,6 +1759,64 @@ public final ManaPool getManaPool() {
...
+
+    /**
+     * Check if this player has any available actions (playable spells/abilities).
+     * Used for smart yield suggestions in network play.
+     *
+     * Note: This uses a heuristic for mana checking since CostPartMana.canPay()
+     * always returns true. We estimate available mana from floating mana plus
+     * untapped mana sources and compare to spell CMCs.
+     */
+    public boolean hasAvailableActions() {
```
tool4ever: I was actually considering a way where the feature doesn't need its own checkbox but the changes in this class alone make that impossible:
this costs way too much performance and certainly should only be performed when it's actually needed

DOCUMENTATION.md
tool4ever: seems appropriate to me

forge-game/src/main/java/forge/game/player/Player.java
MostCromulent: Do you have a preferred way of dealing with this? I raised with the AI coder and its suggested response was:

**Option 1**:  Change to only compute hasAvailableActions() if the smart prompts are enabled by the user; or

**Option 2**: Adopt a less accurate heuristic with lower performance cost:

```
Remove:                                                      
  1. hasAvailableActions() method from Player.java             
  2. updateAvailableActionsForView() method from Player.java   
  3. The TrackableProperty from PlayerView.java                
  4. The call in PhaseHandler.passPriority()                   
                                                               
  Replace with: Simpler heuristics using existing View data    
  (like shouldShowNoManaPrompt already does):                  
                                                               
  private boolean shouldShowStackYieldPrompt() {               
      // ... existing null/stack checks ...                    
                                                               
      // Simple heuristic: no mana = can't respond to stack    
      return !hasManaAvailable(pv);                            
  }                                                            
                                                               
  private boolean shouldShowNoActionsPrompt() {                
      // ... existing null/turn checks ...                     
                                                               
      // Simple heuristic: empty hand + no mana = nothing to do
      FCollectionView<CardView> hand = pv.getHand();           
      boolean emptyHand = (hand == null || hand.isEmpty());    
      return emptyHand || !hasManaAvailable(pv);               
  }                                                            
                                                               
  Trade-offs:                                                  
  - Pro: Zero performance cost - uses already-synced View      
  properties                                                   
  - Pro: No new TrackableProperties needed                     
  - Con: Less accurate - misses free spells, 0-cost abilities, 
  etc.                                                         
  - Con: Won't detect "has cards but can't afford any of them"
```

MostCromulent: When you say "the feature doesn't have its own checkbox" do you mean expanded yields are just integrated by default rather than being an experimental option in the Preferences menu?

tool4ever: the code should probably also be moved to the View classes to keep the engine clean

forge-gui/src/main/java/forge/gamemodes/match/input/InputPassPriority.java
```diff
@@ -327,7 +326,24 @@ private GameView getGameView() {
     }
 
     private PlayerView getPlayerView() {
-        return getController().getPlayer().getView();
+        // For network clients, we need to get the PlayerView from the GameView
```
tool4ever: I still need to investigate this, seems rather hacky

- [x] but if it's needed it should reuse `lookupPlayerViewById`

MostCromulent: I'll address the duplicate function issue now, and can come back to this if you find a better approach.

forge-game/src/main/java/forge/game/player/Player.java
tool4ever: > When you say "the feature doesn't have its own checkbox" do you mean expanded yields are just integrated by default rather than being an experimental option in the Preferences menu?

yea that's what I meant but I feel like this is turning more into a power user feature anyway and it might be better for the new user experience having the GUI less crowded at first
this also relates to how it intersects with the existing system, but I'll make a separate comment on that later

based on that Option 1 probably makes more sense, if the heuristics are broadened it might also degrade usefulness too much imo

thanks

forge-gui/src/main/java/forge/localinstance/properties/ForgePreferences.java
```diff
@@ -286,6 +301,13 @@ public enum FPref implements PreferencesStore.IPref {
         SHORTCUT_MACRO_RECORD ("16 82"),
         SHORTCUT_MACRO_NEXT_ACTION ("16 50"),
         SHORTCUT_CARD_ZOOM("90"),
+        SHORTCUT_YIELD_UNTIL_NEXT_PHASE("112"),         // F1 key
```
tool4ever: I don't mind a few more shortcuts but right now F1 is already hardcoded to Help so this doesn't even work
while some more flexibility in the system might be nice, considering this PR size already it shouldn't be done here

forge-gui/src/main/java/forge/gamemodes/match/input/InputPassPriority.java
```diff
@@ -176,4 +314,100 @@ public boolean selectAbility(final SpellAbility ab) {
         }
     	return false;
     }
+
+    // Smart yield suggestion helper methods
+
+    private boolean isExperimentalYieldEnabled() {
+        return FModel.getPreferences().getPrefBoolean(FPref.YIELD_EXPERIMENTAL_OPTIONS);
```
tool4ever, line 374: given that this feature access is desktop only right now you should also check for mobile GUI
otherwise I'd imagine this will still get enabled and cause some confusion (e.g. for users who switch between them to play adventure)
if you want to support both better leave it for a follow-up, I don't think I'd survive the code growth from that alone... 😅

MostCromulent: Fixed. I wouldn't want to inflict that on the devs, I've already made your lives hard enough.

forge-game/src/main/java/forge/game/player/Player.java
MostCromulent: Noting your feedback about keeping the core engine clean, I think there are two issues:

1. Broad issue of where the yield logic should live and to what extent it interacts with the engine;
2. Downstream of that, what specifically to do about `hasAvailableActions()`.

I asked the AI to generate a refactor plan to move the yield logic to the GUI layer as much as possible. It would be good to get your feedback on this before I implement it to make sure it's doing what you're after.

```
 Proposed: Move Yield Logic from Engine to GUI Layer                          
                                                                              
 Architectural Issue                                                          
                                                                              
 The YieldRework branch has added some yield-related code to the core engine  
 layer (forge-game). Per dev feedback, GUI concerns should stay in the GUI    
 layer (forge-gui) to keep the engine clean.                                  
                                                                              
 Current Engine Layer Pollution                                               
 File: PhaseHandler.java                                                      
 Lines: 1167-1171                                                             
 What It Does: Triggers updateAvailableActionsForView() on every priority pass                                                                         
 ────────────────────────────────────────                                     
 File: Player.java                                                            
 Lines: 1773-1820                                                             
 What It Does: hasAvailableActions() computation + view update method         
 ────────────────────────────────────────                                     
 File: PlayerView.java                                                        
 Lines: 559-565                                                               
 What It Does: Tracker getter/setter (this is fine - it's the sync mechanism) 
 What Should Stay vs Move                                                     
                                                                              
 Keep in Engine (forge-game):                                                 
 - TrackableProperties (HasAvailableActions, WillLoseManaAtEndOfPhase,        
 ApiType) - network sync mechanism                                            
 - PlayerView/GameView getters - reading tracker values                       
 - View update setter methods - storing values for sync                       
 - Player.hasAvailableActions() computation logic - needs game state access   
                                                                              
 Move to GUI (forge-gui):                                                     
 - The TRIGGER for when to compute (currently in PhaseHandler)                
 - Decision logic for when computation is needed (based on yield preferences) 
                                                                              
 Proposed Pattern                                                             
                                                                              
 Engine Layer                             GUI Layer                                
 ────────────────                    ─────────────────────────                
 Game state changes                                                           
     │                                                                        
     ▼                                                                        
 Priority passes to player                InputPassPriority.showMessage()          
     │                                        │                                    
     │                                        ▼                                    
     │                                    Checks if computation needed:            
     │                                    - Experimental yields enabled?           
     │                                    - Specific suggestion prefs enabled?     
     │                                        │                                    
     │                                        ▼ (if needed)                        
     │                                      Calls                                    
 player.updateAvailableActionsForView()                                       
     │                                        │                                    
     ▼                                        ▼                                    
 Tracker syncs to clients ◄─────-──────── Result stored in tracker                 
                                                                              
 The key change: Engine fires events/provides state. GUI decides when to      
 compute expensive operations.                                                
                                                                              
 Questions for Devs                                                           
                                                                              
 1. Is this the right pattern? GUI layer triggers computation, engine         
 provides the methods but doesn't auto-trigger them.                          
 2. Where should the trigger live? Options:                                   
   - InputPassPriority.showMessage() - right before smart suggestions are     
 checked                                                                      
   - Event listener pattern - GUI listens for priority change events          
   - Other location?                                                          
 3. Should computation logic also move? Player.hasAvailableActions() does the 
  expensive work (iterates cards, calls getAllPossibleAbilities). It needs    
 game state access, so it lives in Player.java. Is that acceptable, or should 
  this logic also move to GUI layer somehow?     
```

tool4ever: I'm currently not sure if always pushing from engine to GUI is ideal but that's what we're usually doing

So a few more calls like that are fine 🤷‍♂️

My comment didn't mean any module switching but just having those helper methods inside PlayerView instead
that way they're already closer to their intended usage

MostCromulent: Understood, sorry for confusion - PlayerView move and making `hasAvailableOptions()` computation conditional on experimental yield preference being enabled are now implemented in [cbdc373](https://github.com/Card-Forge/forge/pull/9643/commits/cbcd3732a9c320ed9f0cca79d4d99e7aa53e3560).

DOCUMENTATION.md
MostCromulent: Have removed DOCUMENTATION.md from this branch - it is available [at this link](https://github.com/MostCromulent/forge/blob/NetworkPlay/dev/.documentation/YieldRework-Documentation.md) for reference if needed.

forge-game/src/main/java/forge/game/player/Player.java
tool4ever: - that commit did not clean up Player class
- we definitely don't want another field in GameRules class
  - should just query the PlayerController / the GUI connected to it so it works on per Player:
    - at least locally already so it skips AI
    - better for future work given that partial network support is already included (which is nice but I haven't even gotten around to look at that part yet)

MostCromulent: No problem, see 2nd attempt on latest commit.

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -1759,6 +1759,7 @@ public final ManaPool getManaPool() {
     }
     public void updateManaForView() {
         view.updateMana(this);
+        view.updateWillLoseManaAtEndOfPhase(this);
```
tool4ever: if we keep this it also needs to be behind a preference guard

MostCromulent: Done.

forge-gui-desktop/src/main/java/forge/control/KeyboardShortcuts.java
```diff
@@ -113,6 +114,106 @@ public void actionPerformed(final ActionEvent e) {
...
+        };
+
+        /** Yield until your next turn (experimental, 3+ players only). */
+        final Action actYieldUntilYourNextTurn = new AbstractAction() {
+            @Override
+            public void actionPerformed(final ActionEvent e) {
+                if (!Singletons.getControl().getCurrentScreen().isMatchScreen()) { return; }
+                if (matchUI == null || matchUI.getCurrentPlayer() == null) { return; }
+                if (!FModel.getPreferences().getPrefBoolean(FPref.YIELD_EXPERIMENTAL_OPTIONS)) { return; }
+                if (matchUI.getPlayerCount() >= 3) {
```
tool4ever: is yet another interface method really needed?
instead of just doing:
`matchUI.getGameView().getPlayers().size()`

to me it looks like this will just lead to the same GUI object with extra steps via `YieldCallback` anyway

forge-gui-desktop/src/main/java/forge/control/KeyboardShortcuts.java
```diff
@@ -113,6 +114,106 @@ public void actionPerformed(final ActionEvent e) {
...
+        /** Yield until next phase (experimental). */
+        final Action actYieldUntilNextPhase = new AbstractAction() {
+            @Override
+            public void actionPerformed(final ActionEvent e) {
+                if (!Singletons.getControl().getCurrentScreen().isMatchScreen()) { return; }
+                if (matchUI == null || matchUI.getCurrentPlayer() == null) { return; }
+                if (!FModel.getPreferences().getPrefBoolean(FPref.YIELD_EXPERIMENTAL_OPTIONS)) { return; }
+                matchUI.setYieldMode(matchUI.getCurrentPlayer(), YieldMode.UNTIL_NEXT_PHASE);
+                if (matchUI.getGameController() != null) {
+                    matchUI.getGameController().selectButtonOk();
```
tool4ever: we can't just blindly press ok, could cause unwanted actions
`passPriority()`?

tool4ever:
last cleanup looks rather good, I'll try to get back to this soonish

MostCromulent:
Screenshot 2026-02-27 141406

FYI, I have now added an "Auto-Pass If No Actions" toggle. This may end up being the most useful feature in this PR.

If this is activated and you have no actions to take (e.g. out of mana / no activatable abilities / no playable cards) then it will immediately auto-pass until you can take action. The end result is very much like the flow-of-play on Arena, which moves pretty quickly and only gives priority when there is some useful action to take.

Obviously this relies heavily on how good the heuristic is for "do you have an available action?". Some info about that below if it's helpful / if anyone has suggestions on how it can be refined.

Note this heuristic isn't new for the "Auto-Pass" toggle, it was already in the PR for the smart yield prompts but has been repurposed for this functionality as well.

> ### The available-actions heuristic
> 
> `PlayerView.updateHasAvailableActions(Player)` determines whether the player can take any meaningful game action. It builds on the engine's existing `canPlay()` validation and fills two gaps that the engine doesn't check.
> 
> #### What the engine already checks (via `canPlay()`)
> 
> The heuristic calls `card.getAllPossibleAbilities(player, true)` for each card. This method internally calls `canPlay(true)` on each ability, filtering out anything that fails. By the time the heuristic sees an ability, the engine has already validated:
> 
> - **Timing** - sorcery-speed restrictions, instant-speed permissions, split second on stack
> - **Zone legality** - the card is in a zone it can be played from
> - **Non-mana costs** - sacrifice, tap, discard, life payment, exile (each `CostPart` subclass implements its own `canPay()`)
> - **Static restrictions** - "can't cast" effects, protection, rule-of-law limits
> - **Alternative costs** - flashback, adventure, foretell, optional costs
> 
> #### What the engine does NOT check
> 
> `CostPartMana.canPay()` is hardcoded to `return true` - the engine never pre-computes whether mana costs are payable. It lets the player attempt to pay and rolls back if they can't. Target legality is also not checked by `canPlay()`. The heuristic fills both gaps:
> 
> **Mana affordability (`canAffordSpell`):**
> 1. Builds a mana profile of the player's available resources:
>    - Counts floating mana in pool
>    - Counts each untapped permanent with a mana ability as +1 mana
>    - Builds a byte mask of producible colors from floating mana and from each mana ability's `canProduce()` output (sources that produce "Any" set all color bits)
> 2. For each ability, checks:
>    - Is CMC - available mana? (total amount check)
>    - Is `(colorProfile & ~availableColors) == 0`? (every colored pip in the cost must be a color the player can produce)
>    - Free abilities (no mana cost) always pass
> 
> Note, normal gameplay AI players use `payComputerCosts()` which uses `ComputerUtilMana` to simulate payment — but that's a full mana-payment solver, not a cheap check. It's too expensive to call speculatively for every ability across every card just to decide whether to auto-pass. That's why a more lightweight heuristic is needed.
>
> **Target availability (`hasValidTargets`):**
> - Non-targeted abilities always pass
> - Targeted abilities: calls `TargetRestrictions.getAllCandidates(sa, true)` to check if at least one legal target exists on the battlefield/stack/etc.
> 
> #### What the heuristic scans
> 
> Three categories of actions, returning `true` on the first match:
> 
> 1. **Hand - spells**: each spell ability that passed `canPlay(true)` is checked for mana affordability + valid targets
> 2. **Hand - land plays**: if `getAllPossibleAbilities` returns a land ability, the engine has already confirmed the player hasn't used their land drop and timing is correct. No further checks needed.
> 3. **Battlefield - non-mana activated abilities**: same mana + target check, excluding mana abilities (tapping a land isn't a meaningful game action)
> 
> If nothing matches, `hasAvailableActions` is set to `false` and auto-pass fires.
> 
> ### Conservative by design
> 
> The heuristic prefers false positives (saying the player has actions when they don't) over false negatives (auto-passing when the player could act). Known cases where it over-estimates:
> 
> - ~~**Mana count is approximate**: each untapped source counts as 1 mana regardless of actual output (Sol Ring counts as 1, not 2)~~
> - ~~**Hybrid/Phyrexian mana**: `getColorProfile()` returns all colors in the cost, so a hybrid {W/U} spell appears to require both W and U~~
> - **X costs**: CMC treats X as 0, so X spells always appear affordable
> - **Non-mana costs**: not re-checked (already validated by the engine's `canPlay()`, but some edge cases like "sacrifice a creature" may have changed since `canPlay()` ran)
> - ~~**Other zones**: does not check graveyard (flashback), exile (adventure/foretell), or command zone ΓÇö these are potential future improvements~~
> 
> In all these cases, the player gets prompted rather than auto-passed, which is the safe direction.
> 
> ### Server-side execution
> 
> `updateHasAvailableActions()` runs in `PlayerView` with a full `Player` reference. `mayAutoPass()` runs in `YieldController` via `AbstractGuiGame`. Both execute on the server, even for network clients -the remote client only activates the preference toggle, the server evaluates the game state.

(**Edit:** have updated above to reflect improvements in heuristic).

MostCromulent:
@tool4ever would be helpful to get your view on this:

I had the AI consider what the performance impact would be of perfecting the "has available actions" heuristic discussed above to eliminate false positives. (I can easily image the user complaints on the Discord about being interrupted when yielding when they didn't actually have anything to play...)

It's response was this would involve adopting the `ComputerUtilMana` mana solver which is approximately 28x slower per spell than the current heuristic. However the real cost of the whole thing is in the `getAllPossibleAbilities` + `canPlay()` calls that both approaches **already** share, with that computation preference gated unless features on this branch are enabled.

Analysis here:

>**getAllPossibleAbilities cost**  
>
> The following are rough estimates based on reading the code paths, not measured. I am counting method calls, field accesses, and loop iterations as "ops" — e.g., a for loop over 7 lands with 3 checks each = ~21 "ops." It's napkin math to get order-of-magnitude comparisons, not benchmarked numbers.
>
> Per card, the method does:
> 
> 1. **Collect abilities** - iterate `getSpellAbilities()`, call `GameActionUtil.getAlternativeCosts()` for each. Typically 1-3 abilities per card, each alt-cost check is lightweight. ~10-20 ops per card.
> 
> 2. **`canPlay(true)` on each ability** - this is the expensive inner call. For each ability it checks:
>    - `Spell.canPlay()` / `AbilityActivated.canPlay()`: split second check, zone check, timing check via `SpellAbilityRestriction.canPlay()` (6-8 boolean checks), then `CostPayment.canPayAdditionalCosts()` which iterates all `CostPart` objects calling `canPay()` on each
>    - **But `CostPartMana.canPay()` returns `true` unconditionally** - so the mana check is literally free
>    - Non-mana costs (sacrifice, tap, discard, life, exile) each do a real check - typically 1-3 non-mana cost parts per ability
>    - Optional costs: if `canPlay()` fails, tries each `OptionalCostValue` (flashback, etc.)
>    - ~50-100 ops per ability
> 
> 3. **Total per card:** ~60-120 ops (1-3 abilities ├ù ~50-100 each)
> 
> **For a typical priority pass (5 hand + ~10 battlefield cards = 15 cards):**
> - `getAllPossibleAbilities` across all cards: **~900-1800 ops**
> - Current mana heuristic on the results: **~15 ops**
> - ComputerUtilMana on the results: **~420 ops per spell that passes canPlay**
> 
> So the ratio is:
> 
> | Step | Ops | % of total (with heuristic) | % of total (with ComputerUtilMana) |
> |------|-----|----|----|
> | `getAllPossibleAbilities` (all cards) | ~1200 | **98%** | ~70% |
> | Mana check (heuristic) | ~15 | 2% | - |
> | Mana check (ComputerUtilMana) | ~420 | - | ~30% |
> 
> **Bottom line:** `getAllPossibleAbilities` dominates the cost regardless. Swapping the heuristic for `ComputerUtilMana` would increase total time by roughly **30-35%**, not the 28x it appears in isolation. The mana check is a small fraction of the overall scan because `canPlay()` is already doing most of the heavy lifting per ability.
> 
> That said — the heuristic is essentially free in comparison, and its false positives are safe (player gets prompted, never auto-passed incorrectly). The question is whether the 30% slowdown is worth eliminating those false positives.
> 
> Given this runs once per priority pass and the whole thing is already sub-millisecond, it's probably fine either way.

So I guess the question is: from your POV is that an acceptable performance cost of perfecting the heuristic, noting this is an optional feature + most of the cost is already incurred + it's preference gated anyway?

If we were to do that, AI suggests the updateHasAvailableActions call would move from PlayerView to somewhere in forge-gui (e.g., InputPassPriority or YieldController) where it has access to both Player and ComputerUtilMana. The result would then get back into the TrackableProperty for network sync by calling playerView.set(TrackableProperty.HasAvailableActions, result) from the GUI layer.

tool4ever:
nah, using ComputerUtilMana *that way* is almost certainly a bad idea:
- there are still lots of mana abilities AI can't use, so you'd essentially only be trading some false positive for probably more false negatives
- that also leads to some slippery slope where the AI is potentially providing an intransparent form of gameplay help for the player (the sa<>payment decision carries a lot of "should" in addition to "can")

but I would still expect to see `getAvailableManaEstimate` playing a role in favor of reinventing more heuristics...

autumnmyst:
I would move the interrupt settings to be as screen that you can open from a button right next to the yield durations, I think it's something you might want to change a lot depending on the game state.

MostCromulent:
@autumnmyst:

Screenshot 2026-03-01 114241

Screenshot 2026-03-01 114234

autumnmyst:
> @autumnmyst:
> 
> Screenshot 2026-03-01 114241 Screenshot 2026-03-01 114234

This looks nice, a couple comments:

1. Change "opponent casts a spell" -> "opponent casts a spell or activated ability", you could in theory separate these, but they are similar enough that I would group them since they encompass the event "opponent chose to do something", and we still have triggered abilities as a separate interrupt.
2. I wonder how necessary "Combat" and "End Step" duration options are for yielding when we have the phase marking system. I think you could have them like this in (more/less) ascending duration order: "Clear Stack", "Next Phase", "Next Turn", and "Your Turn"

The reason for 2 being that "next phase" should probably act as "next marked phase" and should continue autopassing until it reaches that players next marked phase. If a player wants to pass until combat or end step, they can simply unmark all phases except those and it will pass until then, making the separate options redundant. (This might be how it already works, but the alternative interpretation is that it autopasses the current stack and passes the prio on the empty stack, then breaks for the next phase regardless of mark. I don't think a single phase yield warrants its own option since if a player can simply mark the very next phase if they want to pass a single phase, so it's functionally redundant)

The only down side of this that the player must remove all phase marks between now and when they want to pass, which might be cumbersome on their turn. I addressed this in my autopass implementation by having an "autopass until next marked phase after current turn" which lets the player autopass past all the marks on their own turn and pass to any phase on another players turn, including combat or end step.

At the end of the day, the most flexible solution would probably be to have a separate type of phase mark specifically for yielding, so one could keep all their existing phase marks and autopass through them to a specific "yield marked" phase (I would represent this as an orange mark, applied by shift-clicking the phases, or something along those lines).

autumnmyst:
Also quick question: Do your networking changes also address the fact that auto-yielding doesn't work for remote clients? This is probably a good time to fix that existing bug since you're already making changes to networking. Also do the same for auto accept/deny triggered choices.

MostCromulent:
> Change "opponent casts a spell" -> "opponent casts a spell or activated ability"

It actually already works this way (includes spells and activated abilities, but not triggered abilities which are the subject of a seperate interrupt setting). I have updated the UI to make this clearer. 

> "next phase" should probably act as "next marked phase" and should continue autopassing until it reaches that players next marked phase. 

Can confirm this is how it works. The player's phase markers are the foundation this build on top of, so it should always respect phase selections and move to the player's next nominated phase. It shouldn't ever stop the player at a phase they haven't marked unless an interrupt condition fires.

> I wonder how necessary "Combat" and "End Step" duration options are for yielding when we have the phase marking system.

> the most flexible solution would probably be to have a separate type of phase mark specifically for yielding

I had actually considered something like this at an early stage in development.

However this PR is already pretty big and adding that would involve a heap more UI changes and wiring. Unless there was a strong view from the devs that something like that should be incorporated immediately, I think it would be best to get current functionality implemented and then that could be looked at as a future improvement.

While "Combat" and "End Step" yield buttons may be unnecessary with that kind of system in place, in the absence of that system I think they're still a useful shortcut to get where the user wants to go and should be retained for time being. That's particularly to avoid the issue you identify where users would need to constantly fiddle with their phase markers.

> Do your networking changes also address the fact that auto-yielding doesn't work for remote clients? This is probably a good time to fix that existing bug since you're already making changes to networking. Also do the same for auto accept/deny triggered choices.

Good catch! Have applied the same pattern this PR already uses to sync yield modes over network to fix auto-yields as well.

MostCromulent:
Updated summary of network architecture following recent pass. 

Network play is still being tested (and will need to be retested after delta merge); I note that local network testing (host + client on same device) is a bit challenging here because both instances read from the same Forge preferences so hard to test mismatched yield settings. Will need to test using separate devices. 

> ### Network yield architecture
> 
> #### Execution model
> 
> Each remote player has **two `YieldController` instances**: one on the client (UI feedback) and one on the host's `NetGuiGame` proxy (which `mayAutoPass` consults). All `PlayerControllerHuman` / `InputPassPriority` logic runs host-side — `InputPassPriority` never runs on the client.
> 
> #### Protocol
> 
> | Direction | Method | Purpose |
> |---|---|---|
> | Client → Server | `notifyYieldStateChanged(player, mode, prefs)` | Yield mode + `YieldPrefs` snapshot. Sent on game open, mode change, and settings change. |
> | Server → Client | `syncYieldMode(player, mode)` | Host clears client's yield (end condition fired). |
> 
> `setHostYieldEnabled(boolean)` greys out client buttons and sends a warning to chat when host has Advanced Yield Settings turned off. The host is authoritative on whether the feature is available — all yield evaluation (interrupt checks, smart-suggestion scans, per-player state bookkeeping) runs host-side, so the host controls the compute cost.
> 
> #### Lifecycle
> 
> 1. **Game start:** Client seeds `notifyYieldStateChanged(player, NONE, prefs)`. Host stores the snapshot in its `NetGuiGame` proxy.
> 2. **Mode change:** Client updates local UI (button highlighted), sends mode + current prefs — prefs ride along on every message so the host's copy stays fresh. Host validates (rejects if experimental off, warns via chat) and stores in proxy's `YieldController`.
> 3. **Per priority pass:** Host's `mayAutoPass` checks yield condition (from host's authoritative `GameView`) and interrupt prefs (from client's `YieldPrefs` snapshot, not host's local prefs). If yield holds, host silently passes — no `updateButtons` sent. Yield is the absence of prompts. If yield ends, host sends `syncYieldMode(NONE)` so client UI clears, then falls through to normal `updateButtons` prompt. `InputPassPriority` routes its auto-pass check through `gui.getRemoteYieldPrefs()` for remote players rather than reading host prefs.
> 4. **Pref change mid-match:** Same message, existing mode + new prefs. Host updates its stored snapshot.
> 5. **Game end:** All `YieldController` state dropped with the GUI instance. No persistence.

autumnmyst:
Testing this right now, feels pretty solid to play with especially the "Auto-Pass If No Actions" (APINA) feature. A few things I noticed though:
1. APINA blows by declaring attackers even if you could declare attackers, definitely address this one.
2. The Yield Interrupt Setting "When you can declare blockers" is redundant, as even when unchecked, the game always allows you to declare blockers if you can block (and it doesn't actually break any yields as fast as I can tell). I would either make it skip blockers automatically or remove this interrupt setting.
3. The "At beginning of combat" interrupt setting is also a but unnecessary in my mind, since you already have a yield until combat as a duration option. Breaking the yield at the beginning of each combat is effectively just passing until combat. I would remove this interrupt setting.

the rest are minor, but still good QOL
4. It's kind of hard to tell when APINA is on or off, it just shows the button a little darker, but I suggests you have it say "Auto-Pass If No Actions: OFF" and "Auto-Pass If No Actions: **ON**" describing the current toggle state.
5. Slight visual bug when turning APINA from the off to on state: when returning to the yields tab after toggling on, it will always appear lighter in color or "on" until hovered over, at which point it will appear correctly.

autumnmyst:
Just patched the declare attackers issue locally, my method was to check in the actual attack declaration logic (not the DA phase itself) for whether specifically APINA was the cause of the pass which:
1. Ensures the actual declare attackers priority round is properly passed if there are no actions
2. Allows other forms of yielding to correctly skip over attacker declaration even when one could declare attackers (such as end turn, pass until next turn, etc.)

MostCromulent:
Main outstanding design issue I could do with some help on is how to get the `hasAvailableActions` heuristic as accurate as possible without compromising performance too much, since Auto-Pass relies on this and I think will end up being the most commonly used feature in this PR.

I think the objective is: ideally it should be perfectly accurate, but if we have to accept some inaccuracy we would rather it generate false positives (prompts player even if no action can be taken) than false negatives (skips player's turn even if they could have taken action).

An architecturally clean solution here has benefits for other PRs in a similar space like https://github.com/Card-Forge/forge/pull/9383 and maybe https://github.com/Card-Forge/forge/issues/5112.

Some options from the AI here: **[has-available-actions-heuristic.md](https://github.com/user-attachments/files/26616098/has-available-actions-heuristic.md)** 

If something like Option B there sounds feasible I could do that as a separate preparatory PR.

**Edit:** @tool4ever do you have a view on this?

autumnmyst:
> Main outstanding design issue I could do with some help on is how to get the `hasAvailableActions` heuristic as accurate as possible without compromising performance too much, since Auto-Pass relies on this and I think will end up being the most commonly used feature in this PR.
> 
> I think the objective is: ideally it should be perfectly accurate, but if we have to accept some inaccuracy we would rather it generate false positives (prompts player even if no action can be taken) than false negatives (skips player's turn even if they could have taken action).
> 
> An architecturally clean solution here has benefits for other PRs in a similar space like #9383 and maybe #5112.
> 
> Some options from the AI here: **[has-available-actions-heuristic.md](https://github.com/user-attachments/files/26616098/has-available-actions-heuristic.md)**
> 
> If something like Option B there sounds feasible I could do that as a separate preparatory PR.

I can take a look at this more closely this weekend, but from just playing a few games against the AI this morning I was very happy with the heuristic in general. The only thing I really noticed was it counted some generic mana tapping abilities like from trioms as possible actions, but something like that would likely require a lot of categorization work as to what mana abilities are worth considering (untapped basic lands are still ignored if you couldn’t afford to do anything with the mana)

autumnmyst:
@MostCromulent I agree with working on the heuristics in a separate PR. This one is already quite large. I think overall you have this in a pretty good state from my testing + the heuristic sounds like a more isolated component of it all.

forge-game/src/main/java/forge/game/player/PlayerView.java
```diff
@@ -531,6 +552,145 @@ void updateMana(Player p) {
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        // Check hand for playable spells that we can afford
+        for (Card card : p.getCardsIn(ZoneType.Hand)) {
+            for (SpellAbility sa : card.getAllPossibleAbilities(p, true)) {
```
tool4ever, line 604: after new APINA architecture should consider if sorting by mana value would potentially allow faster result for performance

forge-ai/src/main/java/forge/ai/AvailableActions.java
```diff
@@ -0,0 +1,74 @@
...
+                if (checkTimeout(deadlineNanos, timeoutMs)) return true;
+                if (!sa.isManaAbility() && canAfford(sa, player) && hasValidTargets(sa)) {
+                    return true;
+                }
+            }
+        }
+
+        for (ZoneType zone : new ZoneType[]{ZoneType.Graveyard, ZoneType.Exile, ZoneType.Command}) {
+            for (Card card : player.getCardsIn(zone)) {
+                for (SpellAbility sa : card.getAllPossibleAbilities(player, true)) {
```
tool4ever: potentially checking Flashback zone instead would be a free improvement?
need to ponder about using it as full source of truth though - might make things a bit unbalanced with regards to the deepness of the performed logic for your own hand 🤔
or just as a filtered input for the loop...

forge-ai/src/main/java/forge/ai/AvailableActions.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private static boolean canAfford(SpellAbility sa, Player player) {
+        if (sa.getPayCosts() == null || !sa.getPayCosts().hasManaCost()) {
+            return true;
+        }
+        return ComputerUtilMana.canPayManaCost(sa, player, 0, false);
+    }
+
+    private static boolean hasValidTargets(SpellAbility sa) {
```
tool4ever: bad logic, use `ComputerUtilAbility.isFullyTargetable`

MostCromulent:
@tool4ever would you rather that I include mobile implementation in this PR (diff should just be UI elements and wiring rather than new core logic) or wait till we land this and then bring mobile support in as a separate PR? I'm happy to play it however you prefer.

tool4ever:
given PR size starting with one GUI will avoid adding even more time for review

autumnmyst:
@MostCromulent I'm noticing the "Next Turn", "Your Turn", "Combat", and "End Step" yield setting sometimes never interrupt when Autopass: ON and skip your turn over and over even if you could do something. Looks like it's because APINA triggers first and prevents the yield from stopping.

Also, the yields should be reset after each game. Should be a quick fix by adding a `yieldStates.clear()` when the game ends.

Both of these issues combined means if you lose a game while passing and start a new one it just skips all your turns!

forge-gui/src/main/java/forge/gamemodes/match/YieldController.java
```diff
@@ -0,0 +1,640 @@
...
+
+    public boolean mayAutoPass(PlayerView player) {
+        player = TrackableTypes.PlayerViewType.lookup(player);
+        if (autoPassUntilEndOfTurn.contains(player)) {
+            return true;
+        }
+        if (isAutoPassingNoActions(player)) {
+            return true;
+        }
+        return shouldAutoYieldForPlayer(player);
```
autumnmyst: Fix for yield not breaking during APINA (`autoPassUntilEndOfTurn` is already checked in `shouldAutoYieldForPlayer`)
```suggestion
        // Yield first so its stop-point logic runs even when APINA would short-circuit.
        if (shouldAutoYieldForPlayer(player)) {
            return true;
        }
        return isAutoPassingNoActions(player);
```

autumnmyst:
1. HasAvailableActions only updates when APINA is on, but the yield suggestions need that value updated even when not autopassing
2. I'm noticing that when turning autopass OFF -> ON it passes once even if the player has actions, much like if the player pressed "Ok." Is this the desired behavior? It seems it could lead to frustration if they only meant to turn the toggle on for later.

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1498,7 +1528,17 @@ public List<SpellAbility> chooseSpellAbilityToPlay() {
                 player.getName(), getGame().getPhaseHandler().getPhase(), getGame().isGameOver());
         final MagicStack stack = getGame().getStack();
 
+        if (FModel.getPreferences().getPrefBoolean(FPref.YIELD_EXPERIMENTAL_OPTIONS)
+                && FModel.getPreferences().getPrefBoolean(FPref.YIELD_AUTO_PASS_NO_ACTIONS)) {
```
autumnmyst: For 1:
```suggestion
        if (FModel.getPreferences().getPrefBoolean(FPref.YIELD_EXPERIMENTAL_OPTIONS)) {
```

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1498,7 +1528,17 @@ public List<SpellAbility> chooseSpellAbilityToPlay() {
                 player.getName(), getGame().getPhaseHandler().getPhase(), getGame().isGameOver());
         final MagicStack stack = getGame().getStack();
 
+        if (FModel.getPreferences().getPrefBoolean(FPref.YIELD_EXPERIMENTAL_OPTIONS)
```
autumnmyst, line 1536: May also want to gate this scan on `!shouldAutoYieldForPlayer(player)`. When a yield mode will cause a pass, `mayAutoPass()` short-circuits before reading `hasAvailableActions`, so refreshing it is wasted work. Interrupts should still fire since `shouldAutoYieldForPlayer` handles that and returns false, so the scan runs for that case.

MostCromulent:
> 2\. I'm noticing that when turning autopass OFF -> ON it passes once even if the player has actions, much like if the player pressed "Ok." Is this the desired behavior? It seems it could lead to frustration if they only meant to turn the toggle on for later.

This is intentional, but I'm open-minded on whether it's best UI design.

Basic idea is that if user is activating auto-pass on their priority it indicates they want to yield. This is same reason activating always-yes or always-no on a stack trigger doesn't just silently toggle the preference, it also applies it and passes priority at the same time.

The alternative has its downsides too. If you activate auto-pass after already receiving priority it can't recalculate APINA (or at least when I looked into this it would have resulted in some very messy code which wasn't worth it) - so you have situation where you activate auto-pass, you may have no actions available, but it just sits there leaving you with priority. That's probably even more confusing.

mermerico:
I love this PR but I found the settings a bit confusing. Here was my effort to reorganize. Let me know what you think and I can upload a patchset.

image

image

image

image

autumnmyst:
@mermerico my concern here is that “intelligent” and “advanced” aren’t mutually exclusive. Often times you will sit with autopass on but then might draw an instant you don’t intend to play immediately, and will then use the yield options to pass forward to a certain point (while having interrupts to stop that yielding), all the while keeping autopass on. For legacy behavior everything is behind a flag in preferences and, if using this system, I wonder how necessary having a legacy setting is since you would just flip the flag (end turn + auto yields still work like how they do in current forge regardless)

autumnmyst:
The right-click yields are really cool. Just a couple suggestions:

1. Have some way to yield just the stack (the suggestions are good, but won't appear if you can respond). Maybe right-clicking the current phase with a non-empty stack could just yield that stack and nothing else, but right clicking current phase with empty stack yields everything until that phase comes around again (like it does currently)?
[A really advanced version of this could let you right click a spell/ability on the stack, and yield to specifically *that* spell/ability, but you would need to deal with those objects getting deleted/manipulated which could get messy, so unless it's trivial probably save for later PR.]

2. Have a checkbox at the bottom of yield interrupts for "Autopass respects interrupts" (unchecked by default). From playing a few games, I'm noticing it's exceedingly rare that I want to actually hold prio when I have no actions (honestly I haven't found a single reason to do so against the computer, since there's no point in bluffing counterspells etc.). I'm finding that interrupts are mainly used for stopping the fast-forward, since I might have an instant I want to cast only in case of an interrupt. However, if I don't have anything, autopass respecting interrupts means it stops me on all my interrupts even if I couldn't do anything about it. I'm suggesting leaving this unchecked by default since (from my experience) the only reason to hold when you have nothing is: 1. if things are playing out too quickly to follow (in which case you could also just turn off autopass / turn on resolve delay) 2. To bluff a counterspell in multiplayer. Both of these I'd consider extenuating enough circumstances to warrant opening the menu and checking the box.

MostCromulent:
> The right-click yields are really cool. Just a couple suggestions:

@autumnmyst both implemented. 👍

autumnmyst:
I found and fixed a small bug as well as moving interrupts to being event-driven. I figured since I'd already implemented a fix I would just make a PR into your branch: https://github.com/MostCromulent/forge/pull/18

forge-gui/src/main/java/forge/gamemodes/net/ProtocolMethod.java
```diff
@@ -98,7 +102,12 @@ public enum ProtocolMethod implements IHasForgeLog {
     setShouldAlwaysAcceptTrigger  (Mode.CLIENT, Void.TYPE, Integer.TYPE),
     setShouldAlwaysDeclineTrigger (Mode.CLIENT, Void.TYPE, Integer.TYPE),
     setShouldAlwaysAskTrigger     (Mode.CLIENT, Void.TYPE, Integer.TYPE),
-    setUiShouldSkipPhase          (Mode.CLIENT, Void.TYPE, PlayerView.class, PhaseType.class, Boolean.TYPE);
+    setYieldMarker           (Mode.CLIENT, Void.TYPE, PlayerView.class, PhaseType.class),
+    clearYieldMarker         (Mode.CLIENT, Void.TYPE),
+    setStackYield            (Mode.CLIENT, Void.TYPE, Boolean.TYPE),
+    setYieldInterruptPref    (Mode.CLIENT, Void.TYPE, forge.localinstance.properties.ForgePreferences.FPref.class, Boolean.TYPE),
```
tool4ever: imo AI shouldn't try to avoid import lines this way

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1474,15 +1502,17 @@ public CardCollectionView tuckCardsViaMulligan(CardCollectionView hand, int card
...
-            if (CombatUtil.validateAttackers(combat)) {
-                return; // don't prompt to declare attackers if user chose to
-                // end the turn and not attacking is legal
+            if (isAutoPassingNoActions()) {
+                // Don't cancel — APINA resumes on subsequent priority passes
+                if (!CombatUtil.canAttack(attackingPlayer)) {
+                    return;
+                }
+            } else {
+                // Yield mode (EOT, next turn, etc.) — intentionally skip attackers
```
tool4ever: I don't understand why the legality check was removed
this seems wrong

tool4ever: what about the old auto pass cancel:
do we feel it's now more desired to keep passing even if we had to declare attackers...?

MostCromulent: No I think you're right: once the interrupt fires we want to cancel the yield. If the user has APINA enabled then it can take care of things from there.

forge-gui/src/main/java/forge/gamemodes/net/ProtocolMethod.java
```diff
@@ -98,7 +103,12 @@ public enum ProtocolMethod implements IHasForgeLog {
     setShouldAlwaysAcceptTrigger  (Mode.CLIENT, Void.TYPE, Integer.TYPE),
     setShouldAlwaysDeclineTrigger (Mode.CLIENT, Void.TYPE, Integer.TYPE),
     setShouldAlwaysAskTrigger     (Mode.CLIENT, Void.TYPE, Integer.TYPE),
-    setUiShouldSkipPhase          (Mode.CLIENT, Void.TYPE, PlayerView.class, PhaseType.class, Boolean.TYPE);
+    setYieldMarker           (Mode.CLIENT, Void.TYPE, PlayerView.class, PhaseType.class),
```
tool4ever, line 106: imo getting too fragmented / inconsistent :/

should probably have a generic method that can transfer any FPref

MostCromulent:
Closing this PR - replaced by https://github.com/Card-Forge/forge/pull/10606.

PR #9652, opened 2026-02-02
KeywordWithType: add getTitle

Hanmac:
Part of #9453

mosty for refactoring `Hexproof:Color`
Also Keyword Counters now use `KeywordInst.getTitle`

the other WithType might be updated later

(and maybe even Protection too)

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -2537,7 +2537,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
                 } else if (keyword.startsWith("Kicker")) {
                     sbLong.append(kickerDesc(keyword, inst.getReminderText())).append("\r\n");
                 } else if (keyword.startsWith("Trample:")) {
-                    sbLong.append("Trample over planeswalkers").append(" (").append(inst.getReminderText()).append(")").append("\r\n");
+                    sbLong.append(inst.getTitle()).append(" (").append(inst.getReminderText()).append(")").append("\r\n");
                 } else if (keyword.startsWith("Hexproof:") && inst instanceof Hexproof hexproof) {
                     final String[] k = keyword.split(":");
                     sbLong.append(hexproof.getTitle());
--- a/forge-game/src/main/java/forge/game/keyword/Trample.java
+++ b/forge-game/src/main/java/forge/game/keyword/Trample.java
@@ -2,6 +2,13 @@ package forge.game.keyword;
 
 public class Trample extends KeywordWithType {
     @Override
+    public String getTitle() {
+        if (!type.isEmpty()) {
+            return "Trample over planeswalkers";
+        }
+        return "Trample";
+    }
+    @Override
     protected String formatReminderText(String reminderText) {
         if (!type.isEmpty()) {
             return "This creature can deal excess combat damage to the controller of the planeswalker it's attacking.";


forge-game/src/main/java/forge/game/keyword/KeywordWithTypeInterface.java
```diff
@@ -0,0 +1,6 @@
+package forge.game.keyword;
+
+public interface KeywordWithTypeInterface {
```
tool4ever, line 3: aren't we always naming them I*?

Hanmac: because my naming schemata is stupid
like KeywordInterface and KeywordInstance

we could clean that up later

forge-gui/res/cardsfolder/upcoming/mondo_gecko.txt
```diff
@@ -3,7 +3,7 @@ ManaCost:1 U U
 Types:Legendary Creature Lizard Mutant
 PT:2/3
 A:AB$ ChooseColor | Cost$ 1 Discard<1/Card> | Defined$ You | SubAbility$ DBAnimate | SpellDescription$ Until end of turn, CARDNAME becomes the color of your choice and gains hexproof from that color.
-SVar:DBAnimate:DB$ Animate | Defined$ Self | Colors$ ChosenColor | Keywords$ Hexproof:Card.ChosenColor:chosenColor | OverwriteColors$ True
```
tool4ever, line 6: image

Hanmac: fixed via Pump and DefinedKW

tool4ever: ah this was broken on master too :)

forge-gui/res/cardsfolder/v/volatile_stormdrake.txt
```diff
@@ -3,7 +3,8 @@ ManaCost:1 U
 Types:Creature Drake
 PT:3/2
 K:Flying
-K:Hexproof:Activated,Triggered:activated and triggered abilities
```
tool4ever, line 6: imo this still only matches CR 702.11d and so should be kept as one

forge-game/src/main/java/forge/game/card/CounterKeywordType.java
```diff
@@ -35,12 +38,8 @@ public String getCounterOnCardDisplayName() {
     }
 
     private String getKeywordDescription() {
-        if (keyword.startsWith("Hexproof:")) {
-            final String[] k = keyword.split(":");
-            return "Hexproof from " + k[2];
-        }
-        if (keyword.startsWith("Trample:")) {
-            return "Trample over Planeswalkers";
```
tool4ever, line 43: Trample doesn't have this title

Hanmac: fixed

forge-gui/res/cardsfolder/v/volatile_stormdrake.txt
Hanmac: imo its a bug

Matt says:
> Volatile Stormdrake's more economical template. Those two qualities can't overlap, unlike colors, so there's no risk of confusion. Something for the rules team to consider, I suppose.

https://bsky.app/profile/wotcmatt.bsky.social/post/3mdsz5gpyys2n

PR #9655, opened 2026-02-02
keywordsToText cleanup

tool4ever:


Hanmac:
there might be more cleanup later (see Protection)

PR #9663, opened 2026-02-03
YECL: Opulent Clomper

dracontes:
[As revealed recently](https://scryfall.com/card/yecl/15/opulent-clomper). Needs support for the "it becomes a random color that _**it isn't**_" part of the triggered ability. I reckoned it would be some middle term between the `Excluded$` and ` ColorsFrom$` parameters available for `ChooseColor` lines. Tried to split the difference between those two in Java with no success. I did lay out a suggestion for name and syntax that does so in any case.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/opulent_clomper.txt
+++ b/forge-gui/res/cardsfolder/upcoming/opulent_clomper.txt
@@ -1,0 +1,11 @@
+Name:Opulent Clomper
+ManaCost:1 W1 G
+Types:Creature Ouphe
+PT:*/*
+S:Mode$ Continuous | CharacteristicDefining$ True | SetPower$ X | SetToughness$ X | Description$ Vivid — CARDNAME's power and toughness are each equal to the number of colors among permanents you control.
+SVar:X:Count$Valid Permanent.YouCtrl$Colors
+T:Mode$ Phase | Phase$ Upkeep | ValidPlayer$ You | TriggerZones$ Battlefield | IsPresent$ Card.Self+!AllColors | Execute$ TrigChooseColor | TriggerDescription$ At the beginning of your upkeep, if this creature isn't all colors, it becomes a random color that it isn't in addition to its other colors.
+SVar:TrigChooseColor:DB$ ChooseColor | Random$ True | ExcludeColorsFrom$ Valid Card.Self | SubAbility$ DBAnimate
+SVar:DBAnimate:DB$ Animate | Colors$ ChosenColor | Duration$ Permanent | SubAbility$ DBCleanup
+SVar:DBCleanup:DB$ Cleanup | ClearChosenColor$ True
+Oracle:Vivid — Opulent Clomper's power and toughness are each equal to the number of colors among permanents you control.\nAt the beginning of your upkeep, if this creature isn't all colors, it becomes a random color that it isn't in addition to its other colors.
--- a/forge-gui/res/cardsfolder/upcoming/opulent_clomper.txt
+++ b/forge-gui/res/cardsfolder/upcoming/opulent_clomper.txt
@@ -1,5 +1,5 @@
 Name:Opulent Clomper
-ManaCost:1 W1 G
+ManaCost:1 G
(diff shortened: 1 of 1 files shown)


forge-gui/res/cardsfolder/upcoming/opulent_clomper.txt
```diff
@@ -0,0 +1,11 @@
+Name:Opulent Clomper
+ManaCost:1 W1 G
```
squee1968: cost is 1 G

jamincollins:
## Findings
#### 1. ExcludeColorsFrom DSL keyword is card-file only, no Java implementation exists
Description
PR #9663 adds Opulent Clomper card file with SVar TrigChooseColor using ExcludeColorsFrom$ Valid Card.Self. This is a card DSL keyword defined in card .txt files only. No Java implementation exists for ExcludeColorsFrom - the keyword is interpreted by the card script parser at load time, not by Java code. The agent previously looped searching forge-game/src/main/java for ExcludeColorsFrom, which will always be empty.

Test case:
```java
// Verify DSL keyword handling
// ExcludeColorsFrom is parsed in CardScriptParser/DB parsing, not as Java method
String cardFile = readFile("forge-gui/res/cardsfolder/upcoming/opulent_clomper.txt");
// Contains "ExcludeColorsFrom$ Valid Card.Self"
// No Java file contains "ExcludeColorsFrom" - verified via grep
```

Impact: Verification loop for Java implementation is unnecessary - DSL keywords are card-file defined.

#### 2. Random color selection may produce no valid color
Description
SVar TrigChooseColor uses ChooseColor | Random$ True | ExcludeColorsFrom$ Valid Card.Self. If creature already has all five colors, ExcludeColorsFrom excludes all colors, leaving no valid choice for random selection. The DB parser may return null or default.

Test case:
```java
// Simulate all colors present
Card self = new Card();
self.addColor(MagicColor.W);
self.addColor(MagicColor.U);
self.addColor(MagicColor.B);
self.addColor(MagicColor.R);
self.addColor(MagicColor.G);
// ExcludeColorsFrom excludes all colors
// Random choice from empty set → undefined behavior
```

Impact: Potential null choice or exception when creature is already five-colored.

#### 3. Vivid power/toughness calculation with zero colors
Description
SVar:X:Count$Valid Permanent.YouCtrl$Colors sets X to count of colors among permanents you control. If player controls no permanents, X=0, PT=0/0. In MTG, 0/0 creatures die immediately.

Test case:
```java
// Player controls no permanents
int colors = countColors(player.getPermanents());
// colors = 0
// Creature PT = 0/0 → state-based action removes it
```

Impact: Card may be self-destructive on empty board.

#### 4. Card file uses upcoming folder, not yet in main set
Description
File placed in forge-gui/res/cardsfolder/upcoming/. The card will not be loaded in standard set data until moved. Ensure card data is consistent with YECL set.

Test case:
// Verify file path matches PR description
String path = "forge-gui/res/cardsfolder/upcoming/opulent_clomper.txt";
// Check YECL set code present in metadata

Impact: Card may not be discoverable if set code mismatch.

#### 5. Upkeep trigger may cause infinite loop with color change
Description
Trigger at upkeep: if creature isn't all colors, it becomes random color it isn't. Changing colors may trigger continuous effects, potentially causing re-evaluation loop with Vivid PT calculation.

Test case:
// Creature gains color → PT recalculated → state-based check → upkeep trigger re-fires?

Impact: Potential infinite loop or excessive recalculation.

## Verdict
Request Changes
PR adds card DSL only, no Java changes needed. The verification loop for Java implementation of ExcludeColorsFrom was incorrect - DSL keywords are card-file defined. Need to verify edge case where ExcludeColorsFrom excludes all colors, and ensure Vivid PT calculation handles zero colors gracefully.

## Recommended Fixes
1. Document that ExcludeColorsFrom is DSL-only, no Java search needed.
2. Add guard for random color selection when no colors remain to choose.
3. Verify Vivid PT calculation handles zero colors (0/0 creature death).
4. Test upkeep trigger with all-colors case to prevent loop.

---

PR #9668, opened 2026-02-03
YECL: Another 4 cards

dracontes:
As revealed recently. Tested to satisfaction:
- [Adaptive Armorer](https://scryfall.com/card/yecl/1/adaptive-armorer)
- [Adherent's Heirloom](https://scryfall.com/card/yecl/29/adherents-heirloom)
- [Tend to the Kiln](https://scryfall.com/card/yecl/13/tend-to-the-kiln)
- [Thoughtweft's Call](https://scryfall.com/card/yecl/3/thoughtwefts-call)

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/adaptive_armorer.txt
+++ b/forge-gui/res/cardsfolder/upcoming/adaptive_armorer.txt
@@ -6,7 +6,7 @@ K:Vigilance
 T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigDraft | TriggerDescription$ When this creature enters, draft a card from CARDNAME's spellbook and put it onto the battlefield. When an Equipment enters this way, attach it to target creature you control.
 SVar:TrigDraft:DB$ Draft | Spellbook$ Amorphous Axe,Citizen's Crowbar,Cloudsteel Kirin,Conqueror's Flail,Fireshrieker,Fishing Pole,Krovod Haunch,Leech Gauntlet,Lion Sash,Mace of the Valiant,Maul of the Skyclaves,Shield of the Realm,Sigiled Sword of Valeron,Thran Power Suit,Thunder Lasso | RememberDrafted$ True | SubAbility$ DBPut
 SVar:DBPut:DB$ ChangeZone | Origin$ Hand | Destination$ Battlefield | Defined$ Remembered | SubAbility$ DBTrigger
-SVar:DBTrigger:DB$ ImmediateTrigger | Execute$ TrigAttach | Defined$ Remembered | ConditionDefined$ Remembered | ConditionPresent$ Card.Equipment | TriggerDescription$ When an Equipment enters this way, attach it to target creature you control.
-SVar:TrigAttach:DB$ Attach | ValidTgts$ Creature.YouCtrl | Object$ Remembered | TgtPrompt$ Select target creature you control | SubAbility$ DBCleanup
+SVar:DBTrigger:DB$ ImmediateTrigger | Execute$ TrigAttach | Defined$ Remembered | ConditionDefined$ Remembered | RememberObjects$ Remembered | ConditionPresent$ Card.Equipment | TriggerDescription$ When an Equipment enters this way, attach it to target creature you control.
+SVar:TrigAttach:DB$ Attach | ValidTgts$ Creature.YouCtrl | Object$ DelayTriggerRememberedLKI | TgtPrompt$ Select target creature you control | SubAbility$ DBCleanup
 SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
 Oracle:Vigilance\nWhen this creature enters, draft a card from Adaptive Armorer's spellbook and put it onto the battlefield. When an Equipment enters this way, attach it to target creature you control.


forge-gui/res/cardsfolder/upcoming/adaptive_armorer.txt
```diff
@@ -0,0 +1,12 @@
+Name:Adaptive Armorer
+ManaCost:2 W W
+Types:Creature Giant Warrior
+PT:4/3
+K:Vigilance
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigDraft | TriggerDescription$ When this creature enters, draft a card from CARDNAME's spellbook and put it onto the battlefield. When an Equipment enters this way, attach it to target creature you control.
+SVar:TrigDraft:DB$ Draft | Spellbook$ Amorphous Axe,Citizen's Crowbar,Cloudsteel Kirin,Conqueror's Flail,Fireshrieker,Fishing Pole,Krovod Haunch,Leech Gauntlet,Lion Sash,Mace of the Valiant,Maul of the Skyclaves,Shield of the Realm,Sigiled Sword of Valeron,Thran Power Suit,Thunder Lasso | Zone$ Battlefield | RememberDrafted$ True | SubAbility$ MoveToPlay
+SVar:MoveToPlay:DB$ ChangeZone | Hidden$ True | Origin$ All | Destination$ Battlefield | Defined$ Remembered | ConditionDefined$ Remembered | ConditionPresent$ Card.Equipment | SubAbility$ DBTrigger
```
tool4ever: what's going on here?

dracontes: Lifted that bit from [Masterpiece Vault](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/m/masterpiece_vault.txt) which in turn I also scripted but, looking over the files in the archives, I'm not sure where I lifted that code from originally (if at all). Best I can figure about a year hence is that I got a bit hung up on the "When an Equipment enters this way" part of the trigger's resolution process.  I take it the conditions should be moved to the `ImmediateTrigger` line and adjusted accordingly.

dracontes: Delving a bit more: seems like I took my lead from [Pharika's Spawn](https://scryfall.com/card/thb/112/pharikas-spawn), namely the replacement effect when it enters escaped.

tool4ever: ok, homework for you :P
where do the cards ETB from / are we sure "Draft" doesn't always put them into the Hand first?

it matters if e.g. _Ranar_ should trigger from _Celestial Vault_

dracontes: As far as it's described on [MTGWiki](https://mtg.wiki/page/Draft_(keyword_action)), the cards are added the hand first. Which likely means the solution is a bit more complicated than paring things down. Is the property `ThisTurnEntered_Battlefield_from_Hand` adequate?

tool4ever: Hmn?
I'm mostly saying that means all Draft scripts using Zone$ are wrong
otherwise why would you even use extra ChangeZone effect here?

dracontes: After a while wondering why variants of what I proposed weren't working, I tested if Forge's game engine worked as one would expect in this regard. The test card is appended for convenience. 

[makers_flow.txt](https://github.com/user-attachments/files/25053987/makers_flow.txt)

I tested [Garruk, Wrath of the Wilds](https://scryfall.com/card/ymid/47/garruk-wrath-of-the-wilds) [-1] loyalty ability which didn't trigger the test card. I looked up gameplay footage of a[ Garruk, Wrath of the Wilds Brawl deck](https://youtu.be/vENppwSISIw?t=151) for the same game action and the drafted card does visually go to the hand before then being put onto the battlefield.
I also tested Masterpiece Vault as I had scripted it in Forge, and the ChangeZone$ effect turns out to not do anything useful as the test card doesn't trigger for it. 

image

Something to be solved on the script or on the Java side of things?

tool4ever: you could either refactor all Draft scripts that use Zone$ param by removing that and using ChangeZone instead
I suppose it'd keep both these game actions clearer apart
Or you just clean up this + the other card and an Alchemy dev like @Northmoc can touch the effect code later

dracontes: In the interest of having things working as expected, and to have seek and draft follow a similar pattern (without the Zone$ parameters), I'll do the script refactor.

forge-gui/res/cardsfolder/upcoming/adaptive_armorer.txt
```diff
@@ -0,0 +1,12 @@
+Name:Adaptive Armorer
+ManaCost:2 W W
+Types:Creature Giant Warrior
+PT:4/3
+K:Vigilance
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigDraft | TriggerDescription$ When this creature enters, draft a card from CARDNAME's spellbook and put it onto the battlefield. When an Equipment enters this way, attach it to target creature you control.
+SVar:TrigDraft:DB$ Draft | Spellbook$ Amorphous Axe,Citizen's Crowbar,Cloudsteel Kirin,Conqueror's Flail,Fireshrieker,Fishing Pole,Krovod Haunch,Leech Gauntlet,Lion Sash,Mace of the Valiant,Maul of the Skyclaves,Shield of the Realm,Sigiled Sword of Valeron,Thran Power Suit,Thunder Lasso | RememberDrafted$ True | SubAbility$ DBPut
+SVar:DBPut:DB$ ChangeZone | Origin$ Hand | Destination$ Battlefield | Defined$ Remembered | SubAbility$ DBTrigger
+SVar:DBTrigger:DB$ ImmediateTrigger | Execute$ TrigAttach | Defined$ Remembered | ConditionDefined$ Remembered | ConditionPresent$ Card.Equipment | TriggerDescription$ When an Equipment enters this way, attach it to target creature you control.
```
tool4ever: we don't cleanup only after immediate trigger, they should use their own RememberObjects instead

PR #9669, opened 2026-02-04
Fix Brutal Cathar trigger timing and add regression canaries(#9484)

Kiousu:
Fixes ChangesZone trigger timing for battlefield-origin events so newly-entered objects don’t retroactively observe older leaves/dies events (Brutal Cathar / Slaughter Specialist case).

  Also adds focused regression canaries in GameSimulationTest for:

  - exile-and-return timing (Brutal Cathar, Fiend Hunter, Banisher Priest),
  - normal simultaneous death observer behavior,
  - command-zone/leave watcher behavior (Oubliette),
  - replacement + leaves-battlefield path (Rest in Peace + Skyclave Apparition),
  - LKI counter-dependent dies behavior (Jotun Owl Keeper).

  No functional changes outside trigger timing guard + tests.

### DIFF
--- a/forge-gui-desktop/src/test/java/forge/ai/simulation/GameSimulationTest.java
+++ b/forge-gui-desktop/src/test/java/forge/ai/simulation/GameSimulationTest.java
@@ -24,16 +24,6 @@ import java.util.List;
 import java.util.Map;
 
 public class GameSimulationTest extends SimulationTest {
-    private int countCreaturesControlledBy(Player p) {
-        int count = 0;
-        for (Card c : p.getCardsIn(ZoneType.Battlefield)) {
-            if (c.isCreature()) {
-                count++;
-            }
-        }
-        return count;
-    }
-
     private Map<AbilityKey, Object> destroyParams(Game game) {
         Map<AbilityKey, Object> params = AbilityKey.newMap();
         params.put(AbilityKey.LastStateBattlefield, game.copyLastStateBattlefield());
@@ -1643,7 +1633,7 @@ public class GameSimulationTest extends SimulationTest {
         Game afterCast = sim.getSimulatedGameState();
 
         Player simOpponent = afterCast.getPlayers().get(0);
-        AssertJUnit.assertEquals(1, countCreaturesControlledBy(simOpponent));
+        AssertJUnit.assertEquals(1, simOpponent.getCreaturesInPlay().size());
     }
 
     @Test


forge-gui-desktop/src/test/java/forge/ai/simulation/GameSimulationTest.java
```diff
@@ -1354,6 +1400,475 @@ public void testKalitasNumberOfTokens() {
...
+
+        addCards("Plains", 3, catharController);
+        Card specialist = addCardToZone("Slaughter Specialist", opponent, ZoneType.Battlefield);
+        addCardToZone("Raging Goblin", opponent, ZoneType.Battlefield);
+        Card brutalCathar = addCardToZone("Brutal Cathar", catharController, ZoneType.Hand);
+
+        game.getPhaseHandler().devModeSet(PhaseType.MAIN1, catharController);
+        SpellAbility catharSa = brutalCathar.getFirstSpellAbility();
+        AssertJUnit.assertNotNull(catharSa);
+        catharSa.setTargetCard(specialist);
```
tool4ever: hmn thanks, but you picked a pretty difficult problem for your first PR...

anyway I thought I already fixed the trigger handling (and only left the issue open for the replacement effect part):
while I see this test fails without the engine change I suspect that's only because you're not providing the LastStateBattlefield correctly when destroying 🤷‍♂️

also here a target is set for the permanent spell which makes no sense (it's possible the AI targets the trigger by itself later though)
this kind of leads me to believe this is generated by an AI once again fooling itself? 👎

Kiousu: Thanks for the feedback, absolutely my mistake.

I think that the tests should be valid now. At the very least, I think I've fixed the two issues that you've pointed out. Dropping the guard I added causes the following tests to fail:
  testBanisherPriestReturnDoesNotSeePastOpponentDeath
  testBrutalCatharDoesNotTriggerReturnedSlaughterSpecialistFromItsDeath
  testBrutalCatharReturnDoesNotSeeSimultaneousOtherCreatureDeaths

If I’m still missing something and the tests are still wrong, that’s on me.

Kiousu: I had the most terrible thought. What if I misread/misremembered the card?

In a case of tragic serendipity, the test was right for the wrong reason.

tool4ever: I'm trying to recreate `testBrutalCatharReturnDoesNotSeeSimultaneousOtherCreatureDeaths` ingame and it seems to work fine:
image

so unfortunately I still don't trust those tests 🤔

Kiousu: Started with the test, got the test wrong. The issue was always solved. Sorry for wasting time.

PR #9642, opened 2026-01-30
Network multiplayer optimization (delta sync + testing)

MostCromulent:
> **Full Disclosure Up Front:** I have zero coding ability. All code has been written by Claude Code AI based on natural language instruction. However that doesn't mean I have just accepted the first thing it spat out. I understand people may be skeptical about the "vibe coding" aspect of this, which is why I have attempted to verify the effectiveness of the code as much as I can through extensive automated testing, and why I have provided comprehensive documentation so there is full transparency for reviewers. If the devs have technical implementation questions or feedback I am going to have to rely on Claude to help me respond to that. I'm sorry if this is a mess; ultimately I am trying to help contribute to Forge in the best way I can. Yes, a comprehensive network play overhaul was probably overambitious as a first project by a non-coder.

**Overview**

This PR attempts to introduce major improvements to Forge's network play functionality. 

These features and their implementation are comprehensively documented in **[Branch Documentation](https://github.com/MostCromulent/forge/blob/NetworkPlay/dev/.documentation/NetworkPlay.md)**.  In summary:
- [**Delta Synchronisation Protocol**:](https://github.com/MostCromulent/forge/blob/NetworkPlay/dev/.documentation/NetworkPlay-delta.md) The current network protocol transfers the entire game state to each player on every update. This is replaced with a new protocol which, after initialisation, only transfers changed properties in the game state on update. If there is a sync error defaults back to the current protocol to restore the full game state. Testing indicates this process massively reduces total bandwidth usage by around ~99.5% compared to status quo. (An average of 2.39 MB per game using new protocol vs 578.58 MB per game using current protocol across an automated 100-game test). 
- [**Comprehensive Test Infrastructure**:](https://github.com/MostCromulent/forge/blob/NetworkPlay/dev/.documentation/Testing-Consolidated.md) Automated headless network testing and analysis tools specifically developed to verify and debug the delta synchronisation protocol. Includes the ability to headlessly run large scale parallel AI vs AI games to completion using full network infrastructure, with comprehensive debug logging.

- ~~[**Reconnection support:**](https://github.com/MostCromulent/forge/blob/NetworkPlay/dev/.documentation/NetworkPlay.md#feature-2-reconnection-support) Allows disconnected players to reconnect to games with automatic state recovery. If a player cannot reconnect within a timeout period then they will be automatically handed to AI control so the game can continue. The host may manually skip the timeout and immediately hand control to AI.~~ **(Edit: Reconnect support moved to seperate PR #9708)**
[...]

PR #9671, opened 2026-02-05
Network multiplayer optimization (UI improvements)

MostCromulent:
**THIS PR HAS BEEN SPLIT FROM #9642.**

> [Network UI Improvements:](https://github.com/MostCromulent/forge/blob/NetworkPlay/dev/.documentation/NetworkPlay.md#feature-4-network-play-ui-improvements) Better feedback during network operations including the prompt window identifying which player currently has priority and how long the game has been waiting for them to take an action; e.g. "Waiting for MostCromulent... (32 seconds)".

**Dependency**
This branch is **has been refactored** to be **fully independent** from other features in #9642.

**Features**
  Waiting timer and connection error improvements for network games:

  *   Show player name in waiting messages ("Waiting for Bob..." instead of generic "Waiting for opponent...")
  *   Add elapsed time counter to waiting prompts (e.g., "Waiting for Bob... (1:23)")
  *   Replace generic "Unable to connect" with more detailed connection error messages (connection refused, unknown host,
  timeout, no route to host)
       * "Unknown host. Please check the hostname or IP address is correct."
       * "Connection timed out. The server may be unreachable or behind a firewall."
       * "No route to host. Please check your network connection."

  **Files Changed**

  | File | Module | Description |
  |------|--------|-------------|
  | `CMatchUI.java` | forge-gui-desktop | Waiting timer, player name prompt messages, view-safe priority detection |
  | `InputLockUI.java` | forge-gui | Network-aware waiting messages with player name and elapsed time |
  | `NetConnectUtil.java` | forge-gui | Detailed connection error messages for common failure types |
  | `CSubmenuOnlineLobby.java` | forge-gui-desktop | Handle `CONN_ERROR_PREFIX` for detailed error display |
  | `OnlineLobbyScreen.java` | forge-gui-mobile | Handle `CONN_ERROR_PREFIX` for detailed error display |
  | `ForgeConstants.java` | forge-core | `CONN_ERROR_PREFIX` constant |
  | `en-US.properties` | forge-gui | 9 new localization strings for waiting and error messages |

---
Screenshot 2026-02-05 064449
Screenshot 2026-02-05 064401

### DIFF
--- a/forge-gui-mobile/src/forge/screens/match/MatchScreen.java
+++ b/forge-gui-mobile/src/forge/screens/match/MatchScreen.java
@@ -378,7 +378,7 @@ public class MatchScreen extends FScreen {
             if (devMenu.isVisible()) {
                 try {
                     //rollbackphase enable -- todo limit by gametype?
-                    devMenu.getChildAt(2).setEnabled(game.getPlayers().size() == 2 && game.getStack().size() == 0 && !GuiBase.isNetworkplay(null) && game.getPhase().isMain() && !game.getPlayerTurn().isAI());
+                    devMenu.getChildAt(2).setEnabled(game.getPlayers().size() == 2 && game.getStack().size() == 0 && !GuiBase.isNetworkplay(MatchController.instance) && game.getPhase().isMain() && !game.getPlayerTurn().isAI());
                 } catch (Exception e) {/*NPE when the game hasn't started yet and you click dev mode*/}
             }
         }


forge-gui/src/main/java/forge/gamemodes/match/input/InputLockUI.java
```diff
@@ -48,21 +55,125 @@ public InputUpdater(final int idxCall) {
 
         @Override
         public void run() {
-            if ( ixCall != iCall.get() || !isActive()) {
+            if (ixCall != iCall.get() || !isActive()) {
                 return;
             }
             FThreads.invokeInEdtLater(showMessageFromEdt);
```
tool4ever, line 54: when is this relevant in netplay? somehow I never reached a breakpoint here... 🤔

I don't understand why there should be two timer systems:
If one input isn't enough, then why isn't it only handled in AbstractGuiGame instead so it works for mobile too?

MostCromulent: Ignore most recent commit, I'll rework this.

MostCromulent: Actually, can I confirm: 

I understand ordinarily display wouldn't be in `AbstractGuiGame`, however Claude AI's initial approach in [9cfa358](https://github.com/Card-Forge/forge/pull/9671/commits/9cfa3586354be25f31b77bc7c0c91b2499b237f0) was to consolidate both timing and display in `AbstractGuiGame` on the basis waiting for opponent prompt is already dealt with in line 479-480 (albeit briefly).

From your perspective is that the preferable approach, or should it be attempting to keep the timing in `AbstractGuiGame` and move the rest elsewhere? E.g. Have `AbstractGuiGame` deal with the timer then `CMatchUI` and `MatchController` can each format their own platform-specific waiting messages.

My assumption is better to split them out given current waiting prompt is ~1-2 lines in `AbstractGuiGame` and therefore trivial, while Claude's consolidated approach would take up a lot more space.

Sorry for ignorance or if this is a stupid question, I just want to figure out the correct principle to apply here and for future reference.

tool4ever: what's the argument for "platform-specific waiting messages"?

it looked like a step in the right direction before revert

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
```diff
@@ -963,7 +970,190 @@ public void popupMenuCanceled(PopupMenuEvent e) {
...
+            cPrompt.setMessage(enhancedMessage);
+        }
+    }
+
+    /**
+     * Enhance generic "Waiting for opponent" and "Yielding" messages with the actual player name.
+     * This is used on the host side where messages aren't sent through NetGuiGame.
+     */
+    private String enhanceWaitingMessageForHost(final String message, final PlayerView forPlayer) {
+        if (!GuiBase.isNetworkplay() || message == null || message.isEmpty()) {
```
tool4ever: relying on `GuiBase.isNetworkplay` is actually rather negative because it only uses a static field:
this means mixing different lobby types could change the value and lead to weird results
I'd rather not increase that problem's exposure...

would be interesting to hear some options to fix it, might even be worth its own PR first if too complex :/

forge-gui/src/main/java/forge/gamemodes/match/input/InputLockUI.java
MostCromulent: tbh it would just be duplication, i think only argument for it is to try and avoid too much of AbstractGuiGame being taken up with the timer/display logic. 
I'll revert the revert and we can go from there.

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
MostCromulent: ## Claude's suggestions:

```
There are around 17 usages across the codebase. They fall into distinct categories:      

 What each usage actually asks:
                                                                                             
  Category: "Is there a remote player?"      
  Files: AbstractGuiGame (x2), MatchScreen (disable rollback)                                                               
  What it really means: Behavior differs because a human is on the other end
  ────────────────────────────────────────
  Category: "Must I serialize across the wire?"
  Files: PlayerControllerHuman (x3), InputBase (x2)
  What it really means: Can't send engine objects directly, must use views
  ────────────────────────────────────────
  Category: "Should I skip EDT assertions?"
  Files: FThreads (x1)
  What it really means: Network threading model differs from local
  ────────────────────────────────────────
  Category: "Should I process this event?"
  Files: FControlGameEventHandler (x1)
  What it really means: Combat update only needed for remote refresh
  ────────────────────────────────────────
  Category: "Should I do extra UI refresh?"
  Files: MatchController (x4), MatchScreen (x1), VCardDisplayArea (x1)
  What it really means: Client needs explicit refresh that local gets for free
  Options for a global fix

 Option A: Per-instance flag on AbstractGuiGame

  Add boolean networkGame to AbstractGuiGame, set during construction or game setup that 
records whether this specific game instance is a network. Every call site that currently has
  access to an IGuiGame (most of them) calls gui.isNetworkGame() instead of GuiBase.isNetworkplay().

  - Fixes the core problem: the flag is per-game-instance, not global
  - Minimal structural change — just moving from static to instance
  - Call sites that don't have a gui reference (FThreads, FControlGameEventHandler) need one 
  threaded through, which is slightly messier
  - Doesn't solve the semantic problem — it's still a boolean that means different things at different call sites

The information already exists at the right moment. GameLobby has allowNetworking 
(set in the constructor — ServerGameLobby passes true, local lobbies pass false). It calls 
getGui(index) per slot, which already branches on LobbySlotType.LOCAL vs LobbySlotType.REMOTE.
The IGuiGame instances are created right there, and the lobby knows whether it's a network lobby.

Option B:

  Instead of one boolean that means 5 different things, replace each category with a method that asks the actual question:

  - gui.hasRemotePlayers() — for timer display, rollback disable
  - gui.requiresViewSerialization() — for the PlayerControllerHuman/InputBase serialization checks
  - gui.requiresExplicitRefresh() — for the mobile refresh calls
  - Remove the FThreads check entirely (it's suppressing a bug, not solving one)
  - FControlGameEventHandler check gets pushed into the event handler registration (only subscribe in network games)

  Each method can default to false in AbstractGuiGame and return true in the appropriate subclass. No static field, no single boolean overloaded with multiple meanings. Call sites become self-documenting.

  - Most correct solution — each call site says what it actually needs
  - Larger diff, but each change is mechanical
  - Enables future scenarios (e.g., spectator mode, replay mode) where some of these are true and others aren't
  - The capability methods could initially all delegate to a single networkGame field internally, making this a
  refactor-only change with identical behavior
```

tool4ever: right, thanks

so option A doesn't sound too bad, only thing I'm worried about is the usage in `assertExecutedByEdt`
I think getting the GUI from all places there might get too ugly...? 😅

but imo we can cheat a bit here (for now):
it is only a safeguard after all and if we bail out as long as there's at least one network game negative impact for the engine should be avoided
(and honestly this is already somewhat future-proofing since running multiple games this way is currently desktop only and probably more for power user)

so I'd like `GuiBase.isNetworkplay` to look like this:
```
    public static boolean isNetworkplay(IGuiGame game) {
        if (game != null) {
            // query AbstractGuiGame implementation if provided
            return game.isNetGame();
        }
        // both IGuiBase implementations should have (at least indirect) access to matches to check all available IGuiGame?
        return getInterface().hasNetGame();
    }
```

so FThreads could just use the slightly less accurate fallback
hope that makes sense

MostCromulent: I think so, see latest commit for Claude's implementation.

tool4ever: Problem A:
to me it looks like `NetGuiGame` is only held by server and for some reason it's only working on remote client side now
that's why the net property probably needs to get set manually into the IGuiGame by the lobby when creating on either side?

Problem B:
seems remote clients won't use HostedMatch, so it's a bit more tricky to check than I first imagined
I have implemented a roundabout way for desktop to show what I had in mind, but still need to consider it some more

forge-gui/src/main/java/forge/gamemodes/match/input/InputLockUI.java
tool4ever: sigh, the problem is with this change it will somehow end up sending each timer update over the network
we really don't want constant extra traffic from counting
so I guess it needs to be split up a bit more after all:
I'd expect that "waiting" messages are only sent once and then each side does their clock enhancement locally

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
tool4ever: I'd also like to point out that maybe this part doesn't need to be limited to netplay 🤔
given that we have AI timeout settings too?

and I could even imagine a future where some special enemies in adventure mode limit the humans allowed thinking time (a bit like blitz chess which seems popular enough)
though that would require more follow up to this so the time allowance tracking doesn't always get reset - so ignore this part for now

MostCromulent: > Problem A:
> to me it looks like `NetGuiGame` is only held by server and for some reason it's only working on remote client side now
> that's why the net property probably needs to get set manually into the IGuiGame by the lobby when creating on either side?

Claude developed a workaround for this on most recent commit:

```
 Also fix host not showing network-specific messages. A network match has
  three IGuiGame contexts: host's local GUI (CMatchUI), server-side proxy
  (NetGuiGame), and remote client. Only NetGuiGame returned true from
  isNetGame() — the host's CMatchUI inherited the default false, so it
  showed "Waiting for opponent" and never triggered showWaitingTimer.
  Fix: AbstractGuiGame now tracks a networkGame field with setNetworkGame(),
  called by FServerManager.getGui() when creating the host's local GUI.
```

tool4ever: looks like it works now!

forge-gui/src/main/java/forge/gamemodes/match/input/InputLockUI.java
tool4ever: cleaned it up some more, I'm hopeful for the current state

- **in the future** imo the new ProtocolMethod should just be a generic "send unlocalized prompt call"
- and the timer can just be started when `lblWaitingForPlayerWithTime` is received
- then more methods can switch to that approach which will allow remote players to handle the localization on their end instead of the host translating for everybody :)

tool4ever (approved):
LGTM

PR #9676, opened 2026-02-06
Restore default screen orientation autodetection

### LINKED ISSUE #9654: Portrait mode UI not working in most recent nightly snapshots (2.0.10) desktop version.
After downloading the most recent snapshot 2.0.10, the Portrait UI is no longer functional, it uses a stretched landscape mode in both adventure and classic modes through the adventure exe. confirmed portrait mode was working in 2.0.08 & 2.0.09

**To Reproduce**
1. Go to forge-adventure.exe on a portrait monitor
2. Click on either adventure or classic 
3. See error

**Expected behavior**
Use the proper Portrait UI that existed in in previous version when detected a monitor that is taller than it is wider

**Screenshots**

Image
Image
Image

 - OS: Windows 11

**Additional context**
i have tried on a normal monitor in portrait mode + a tablet in portrait mode using a virtual display and the bug exists on both

krafczyk:
Fixes #9654 

* When the user doesn't pass 'portrait', 'landscape', or 'height/width', fall back to the original behavior auto detecting monitor orientation.
* Use monitor aspect ratio to inform width/height when only one is set via command line.
* Add helper method to signal to Forge that the user is overriding the default orientation.
* More comprehensive height/width logic


forge-gui-mobile-dev/src/forge/app/GameLauncher.java
```diff
@@ -58,28 +59,60 @@ public GameLauncher(final String versionString, final String[] args) {
...
         }
 
-        // Check if the manually supplied window size indicates portrait mode
-        if ((windowHeight > 0) && (windowWidth > 0)) {
-            if (windowHeight > windowWidth) {
-                isPortrait = true;
+        boolean hasAnyDimArg = (widthArg != null || heightArg != null);
+        boolean hasBothDims = (widthArg != null && heightArg != null);
+
+        boolean manualWindowSize = hasAnyDimArg;
```
tool4ever: should clean up some of these booleans
these command line arguments don't really need their own ones if you're just going to reassign them to another anyway
just makes things messier

forge-gui-mobile-dev/src/forge/app/GameLauncher.java
```diff
@@ -58,28 +59,60 @@ public GameLauncher(final String versionString, final String[] args) {
...
+
+            // If user explicitly overrode orientation (portrait/landscape or both dims), normalize by swapping
+            if (overrideOrientation) {
+                if (isPortrait && windowHeight < windowWidth) {
+                    int tmp = windowHeight; windowHeight = windowWidth; windowWidth = tmp;
+                } else if (!isPortrait && windowWidth < windowHeight) {
+                    int tmp = windowHeight; windowHeight = windowWidth; windowWidth = tmp;
+                }
             }
         }
```
tool4ever: imo if the size from settings is used it shouldn't be messed with
or at the very least only if the orientation arguments are actually provided
otherwise you're making it more difficult to freely combine both parts in case the screen size is big enough for either mode

forge-gui-mobile-dev/src/forge/app/GameLauncher.java
krafczyk: I'm not entirely sure what you mean here. All of these booleans help clarify how specific choices of arguments are mapped to 'user intent' For example, `overrideOrientation` gets set if portrait, landscape, or the user passed 'both dims' hence `hasBothDims`. I think if we remove `hasBothDims`, that makes the intent of the code less clear doesn't it?

It is more lines of code, but the benefit is clearer intent.

forge-gui-mobile-dev/src/forge/app/GameLauncher.java
krafczyk: I don't know what you mean. Almost the entire section above is only run if the user passes a command line argument related to window size. In such a case, it makes sense to believe the user intends to override settings doesn't it?

You can see around line 128 the only situation when setting values are silently swapped is if the stored setting dimensions don't match `isPortrait`. I believe that was the original behavior.

tool4ever: incorrect

before your (initial) PR:
can freely enter any values in settings.json and get the desired window size to test

now:
always enforces landscape size this way

PR #9680, opened 2026-02-06
Perpetual: use StaticAbilities

### LINKED ISSUE #9241: Perpetual Effect into Hidden Statics
Perpetual Effects should be changed into Hidden Statics, so while still applying, they use Layer System.

The outliner is ManaCost changes for `Thought Partition` and Incorporate.
It needs to apply before Layer 4 (Type-change)
someone might justify it as a text-change? 
`Volrath’s Shapeshifter` already uses addChangedManaCost too

Also, their Timestamp might need a reset into negative or zero on zone change:

> (2) Once an object changes zones, the timestamp of the perpetual effect changes to, effectively, "negative infinity". That is, earlier than any other non-perpetual timestamp 

https://www.reddit.com/r/MagicArena/comments/17pb9y9/comment/k8a98g3/

Some Perpetual Ability Effects, have LosePerpetual as its trigger.
So i need to get from the Trigger to the HiddenStatic to the Perpetual Effect

Also, Commander can clean up their Perpetual #4135 
> you may choose to remove all perpetual effects on that Commander

Should probably be done via EmptySA?

Affected Layers:

* Layer 3: PerpetualManaCost and PerpetualIncorporate
* Layer 4: PerpetualTypes
* Layer 5: PerpetualColors and PerpetualIncorporate
* Layer 6: PerpetualAbilities and PerpetualKeywords
* Layer 7b: PerpetualNewPT
* Layer 7c: PerpetualPTBoost

Hanmac:
Closes #9241 

still WIP

need to refactor the other ones into Statics

Also, the removePerpetual Effect for this needs to be deeper refactored:

> They perpetually gain “When you cast this spell, create a Treasure token and this spell perpetually loses this ability.”

Because right now, we can't handle ` Runeblade Raiser ` with 
> When this creature dies, return it to the battlefield under its owner’s control. It perpetually loses this ability.

In theory, the CardTraits Changes need a way to remove a specific Trigger (per Id?)

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -4677,7 +4677,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
             if (!lki && oldTimestamp > 0) {
                 // change zone sets the timestamp of negative infinite, but still relative
                 // use current game timestamp to shift these timestamps back
-                st.putParam("Timestamp", String.valueOf(oldTimestamp - currentGameTimestamp));
+                st.putParam("Timestamp", String.valueOf(Long.MIN_VALUE + oldTimestamp));
             }
             this.perpetualEffects.put(e.getKey(), st);
         }


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4653,16 +4659,24 @@ public final void removePerpetual(final long timestamp) {
...
+        if (!lki) {
             for (PerpetualInterface p : perpetual) {
                 p.applyEffect(this);
             }
         }
+
+        for (Map.Entry<PerpetualInterface, StaticAbility> e : oldCard.perpetualEffects.entries()) {
+            StaticAbility st = e.getValue().copy(this, lki);
+            if (!lki) { // change zone sets the timestamp of negative infinite, but -1 is good enough
+                st.putParam("Timestamp", String.valueOf(-1));
```
tool4ever: if you really want to do it correctly this still needs to preserve the relative order

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -4630,15 +4630,21 @@ public final boolean hasIntensity() {
     }
 
     private List<PerpetualInterface> perpetual = new ArrayList<>();
+    private Multimap<PerpetualInterface, StaticAbility> perpetualEffects = MultimapBuilder.hashKeys().arrayListValues().build();
```
tool4ever, line 4633: remove List above

forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
```diff
@@ -92,39 +92,36 @@ public static void doAnimate(final Card c, final SpellAbility sa, final Integer
...
+                c.addPerpetual(new PerpetualManaCost(manaCost), timestamp);
             }
         }
         
         if (!addType.isEmpty() || !removeType.isEmpty() || addAllCreatureTypes || !remove.isEmpty()) {
             if (perpetual) {
-                c.addPerpetual(new PerpetualTypes(timestamp, addType, removeType, remove));
+                c.addPerpetual(new PerpetualTypes(timestamp, addType, removeType, remove), timestamp);
             }
             c.addChangedCardTypes(addType, removeType, addAllCreatureTypes, remove, timestamp, 0, true, false);
```
tool4ever, line 107: shouldn't all be `else` now?

forge-game/src/main/java/forge/game/card/perpetual/PerpetualInterface.java
```diff
@@ -1,8 +1,11 @@
 package forge.game.card.perpetual;
 
 import forge.game.card.Card;
+import forge.game.staticability.StaticAbility;
 
 public interface PerpetualInterface {
-    long getTimestamp();
-    void applyEffect(Card c);
+    default long getTimestamp() { return -1; }
```
tool4ever, line 7: reminder to remove this later

forge-game/src/main/java/forge/game/card/Card.java
Hanmac: i added oldTimestamp - currentGameTimestamp
that should shift them into negative

or does this need more fine tuning?
(i think if it changes zones again and again, they should be shifted more back, right?)

forge-game/src/main/java/forge/game/card/Card.java
Hanmac: not all are transformed yet, the list will be removed later

(i don't know yet if the new one needs to be a Multimap or if a List is enough)

Hanmac:
@tool4ever over the weekend i might need some help to find a way to remove Triggers via ChangedCardTraits and Static Ability

especially this one: https://scryfall.com/card/ytdm/8/runeblade-raiser

forge-game/src/main/java/forge/game/card/Card.java
tool4ever: I'll do some testing for it when the PR is more ready

tool4ever: I see two problems:
- without extra logic this would just reverse the order
  - A: 2 & B: 3 => A: -3 & B: -2
- when only changing a single card this way its timestamps could still grow smaller than those of perpetual effects from other cards in a way that switches the order

can't we just add `Long.MIN_VALUE` the first time it changes?

PR #9689, opened 2026-02-07
Rename Adventure arena function from 'loose' to 'lose'

aryst0krat:
The function to handle losses in the Adventure Mode arenas had a function called 'loose' instead of 'lose', I've just fixed that.


tehdiplomat:
I'll merge but there's no value to this pr

aryst0krat:
Very little, sure, but it's a typo that affects clarity/searchability. I could have waited to collect it with others but who knows how long it might have taken to find more if I wasn't actively looking for them 🤷

I'll try not to make a habit of submitting a PR for a single typo though.

PR #9692, opened 2026-02-07
Beta release of Realm of Legends adventure mode world

VigilantSera:
Only should have changes in the plane folder.


tehdiplomat:
@VigilantSera Let us know if you have time to update the filenames. Since a fair amount of adventure users are Android and it may cause issues there. We can fast follow the renames if it makes sense also, but better to not get people whining about something we know might cause problems.

tehdiplomat (approved):

Issue #4745, opened 2024-02-26
Outpost Siege loses track of "can be cast" when casting fails


Samthere:
Windows, current snapshot.

I have `Outpost Siege` [FRF] in Khans mode, which puts a card into exile and allows it to be played that turn. I started casting the exiled card, then cancelled. I'm no longer able to cast the card.

Presumably this is some tracking effect/element which is removed at the start of casting, instead of the end.

tool4ever:
I want to do a big refactor to check all ExileOnMoved usage but that's something for my backlog

PTheocharis:
Hello, I am a beginner in coding and I want to start improving. My class has a project to contribute to an Open source project, do you think I can try to handle this task?

tool4ever:
Always happy for help
you need more details?

tehdiplomat:
I wonder if this is resolved with the undo restore experimental functionality that was recently added.

harshitissm:
Hi, I am starting my open-source contribution journey and came across this issue which was market as good for newcomers.

Would anyone explain me this issue?

tool4ever:
We have multiple params for effect cleanup, the logic is defined here:
https://github.com/Card-Forge/forge/blob/2042aaa03393b243c5b84f868aaa0c50d3f5f124/forge-game/src/main/java/forge/game/ability/effects/EffectEffect.java#L228-L235

ExileOnMoved$ might be a bit quicker but is only for single affected cards and causes problems with our standard way of doing stack rollback

So basically all effects that involve some kind of play permission should avoid it. Depending on the amount it might even make sense to implement heuristics for auto-detection 🤔

pachecoberlin:
I made a puzzle starting at untap phase with Outpost Siege [FRF] on battlefield, so i can choose Khans. I can confirm that addExileOnMovedTrigger is called, while resolving in upkeep.
I have one card in library, which is exiled and can be cast and canceled several times in both main phases. 
So i cannot reproduce. Any other ideas or steps on how to test/reproduce would be greatly appreciated.

tool4ever:
@dracontes 
I'm trying to document these params usage in the related wiki section

Can you help search if there's a card where `ExileOnMoved$` is required instead of `ForgetOnMoved$`?
so probably something that's limited to a single use (ideally nothing with stack as that complicates things)

dracontes:
> Can you help search if there's a card where `ExileOnMoved$` is required instead of `ForgetOnMoved$`?

[...]

PR #9703, opened 2026-02-09
Refactor: Hidden keywords to StaticAbilityMode (Issue #3307)

calaespi:
Refactors hardcoded hidden keywords (MustBeBlocked, CantGainControl, etc.) to use type-safe StaticAbilityMode enum. Includes comprehensive tests in HiddenKeywordsMetaTest.java.


forge-game/src/main/java/forge/game/spellability/AbilityActivated.java
```diff
@@ -87,8 +89,18 @@ public boolean canPlay() {
...
+        
+        boolean cantBeActivated = c.hasKeyword("CARDNAME's activated abilities can't be activated.");
+        if (!cantBeActivated) {
+            for (final StaticAbility sa : c.getStaticAbilities()) {
+                if (sa.checkConditions(StaticAbilityMode.CantBeActivated)) {
+                    cantBeActivated = true;
+                    break;
+                }
+            }
+        }
```
Hanmac, line 92: this is already handled there:
https://github.com/Card-Forge/forge/blob/5b5de38976d99e8a4e0e0a6d76a7d967b4c21b64/forge-game/src/main/java/forge/game/spellability/AbilityActivated.java#L103-L107

Hanmac:
You probably should check out the other StaticAbility classes, like `StaticAbilityCantTransform`:
https://github.com/Card-Forge/forge/blob/master/forge-game/src/main/java/forge/game/staticability/StaticAbilityCantTransform.java

using something like `ValidCard` and others can be used by other cards

forge-gui/res/cardsfolder/l/lure.txt
```diff
@@ -3,5 +3,6 @@ ManaCost:1 G G
 Types:Enchantment Aura
 K:Enchant:Creature
 SVar:AttachAILogic:Pump
-S:Mode$ Continuous | Affected$ Creature.EnchantedBy | AddHiddenKeyword$ All creatures able to block CARDNAME do so. | Description$ All creatures able to block enchanted creature do so.
+S:Mode$ Continuous | Affected$ Creature.EnchantedBy | AddStaticAbility$ SMustBlock | Description$ All creatures able to block enchanted creature do so.
+SVar:SMustBlock:Mode$ MustBeBlockedByAll
 Oracle:Enchant creature\nAll creatures able to block enchanted creature do so.
```
Hanmac, line 7: If `MustBeBlockedByAll` is done as a static, it should not be added to the enchanted creature anymore.

Notice how the Aura doesn't say: `enchanted creature gains`

Hanmac:
I did the CountersRemain as extra MR: #9709

calaespi:
Merged master (incorporating StaticAbilityCountersRemain) and refactored CantGainControl, LethalDamageByPower, and BounceAtUntap to use dedicated StaticAbility classes as requested. Also removed the redundant RetainCounters mode.

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -6025,7 +6025,9 @@ public final int getTotalDamageDoneBy() {
 
     // this is the amount of damage a creature needs to receive before it dies
     public final int getLethal() {
-        if (hasKeyword("Lethal damage dealt to CARDNAME is determined by its power rather than its toughness.")) {
+        boolean lethalByPower = hasKeyword("Lethal damage dealt to CARDNAME is determined by its power rather than its toughness.") || StaticAbilityLethalDamageByPower.isLethalDamageByPower(this);
```
tool4ever: @Hanmac 
we need a better solution imo
this method is called so many times, it'd be terrible to check it like this for a **single** card

don't want another *Topsy Turvy* situation like #8513

calaespi:
Refactored `LethalDamageByPower` to use a cached boolean flag in `Card.java`, avoiding the expensive iteration over all static abilities in `getLethal()`. Also removed the now unused `StaticAbilityLethalDamageByPower` class. Verified with tests (Zilortha test passes, existing failure in Skullbriar is unrelated).

forge-game/src/main/java/forge/game/card/Card.java
tool4ever: I'd prefer the new approach?

calaespi: Just to clarify: the current implementation already uses the cached flag approach (checking `!lethalDamageByPower.isEmpty()`) to avoid the expensive iteration. I assume this is the "new approach" you are referring to?

tool4ever: yes thanks
just need @Hanmac to vote on it

Hanmac: 👍 on that part

If the card gets updated, the keyword can be removed

Also remove the static mode that was added because of it

forge-gui/res/cardsfolder/l/lure.txt
tool4ever: imo at least all MustBeBlockedBy variants should be handled together
but maybe remove it from this one so we can merge in smaller steps
and you have the static already as part of #4745 instead

forge-game/src/main/java/forge/game/ability/effects/ControlGainEffect.java
```diff
@@ -87,7 +88,13 @@ protected String getStackDescription(SpellAbility sa) {
     }
 
     private static void doLoseControl(final Card c, final Card host, final long tStamp) {
-        if (null == c || c.hasKeyword("Other players can't gain control of CARDNAME.")) {
+        if (null == c) {
+            return;
+        }
+        
+        boolean cantGainControl = c.hasKeyword("Other players can't gain control of CARDNAME.") || StaticAbilityCantGainControl.cantGainControl(c);
```
tool4ever, line 95: this seems fine
but there is another place with the KW
remove checking for it from both

forge-gui-desktop/src/test/java/forge/gamesimulationtests/HiddenKeywordsMetaTest.java
```diff
@@ -0,0 +1,232 @@
...
+        Player p1 = game.getPlayers().get(0);
+        
+        Card bear = createCard("Grizzly Bears", p1);
+        p1.getZone(ZoneType.Battlefield).add(bear);
+        bear.setGameTimestamp(game.getNextTimestamp());
+        bear.setSickness(false);
+        
+        // Simulate Detain effect
+        // Detain adds "CARDNAME can't attack or block." and "CARDNAME's activated abilities can't be activated."
+        bear.addIntrinsicKeyword("CARDNAME can't attack or block.");
```
tool4ever, line 98: shouldn't test for outdated KW

calaespi:
I have merged the latest changes from upstream/master and resolved the conflicts in AbilityActivated.java. 
I preserved the `isDetained()` check from upstream while keeping the removal of the redundant keyword loop.
The PR should be conflict-free now.

forge-game/src/main/java/forge/game/staticability/StaticAbilityCountersRemain.java
tool4ever, line 1: reverting too much now...

forge-gui-desktop/src/test/java/forge/gamesimulationtests/Issue4745Test.java
```diff
@@ -0,0 +1,89 @@
+package forge.gamesimulationtests;
```
tool4ever, line 1: not sure what you're doing but you keep mixing your PR
this is from #9717

scripts/compile-and-test-pr.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
tool4ever: sigh, stop adding random stuff

calaespi: Sorry! I love scripts that give me multiple commands at once. I'll make a note not to do it again.

jamincollins:
## Findings

#### 1. StaticAbilityMustBeBlockedByAll checks attacker against itself
The `mustBeBlockedByAll` method iterates over cards in STATIC_ABILITIES_SOURCE_ZONES including the attacker itself, then calls `applyMustBeBlockedByAll` which only checks `ValidCard` against attacker. This means a static ability defined on the attacker card will be found when iterating over the attacker's own static abilities, causing self-check loops. The method also uses a misleading `host` variable that is never used.

Test case:
```java
// Simulate attacker with MustBeBlockedByAll static ability on itself
Card attacker = new Card(...);
attacker.addStaticAbility(createStaticAbility(StaticAbilityMode.MustBeBlockedByAll, "ValidCard", "Creature"));
Card blocker = new Card(...);
boolean result = StaticAbilityMustBeBlockedByAll.mustBeBlockedByAll(attacker, blocker);
// result should be true if ability matches, but the iteration includes attacker itself
// leading to ambiguous host resolution
```

Impact: Potential double-evaluation and confusion about host resolution; may cause incorrect blocking decisions when ability is on the attacker itself.

#### 2. StaticAbilityBounceAtUntap iterates all static abilities without zone filtering
`shouldBounceAtUntap` iterates over all cards in STATIC_ABILITIES_SOURCE_ZONES (Battlefield, Graveyard, Exile, Command, Stack) and checks every static ability for mode BounceAtUntap. This is overly broad - abilities in Graveyard/Exile/Stack should not trigger during untap step. The method also lacks a host card check.

Test case:
```java
// Card in graveyard with BounceAtUntap static ability
Card graveyardCard = game.getCardsIn(ZoneType.Graveyard).get(0);
StaticAbility bounceAb = graveyardCard.addStaticAbility(StaticAbilityMode.BounceAtUntap);
Card target = new Card(...);
boolean shouldBounce = StaticAbilityBounceAtUntap.shouldBounceAtUntap(target);
// Returns true even though ability source is in graveyard
```

Impact: Incorrect bounce triggers from abilities in zones where they shouldn't apply, violating MTG timing rules.

#### 3. LethalDamageByPower uses Set<Long> without timestamp collision handling
`Card.addLethalDamageByPower` adds timestamp to Set<Long>. Multiple static abilities with same timestamp (e.g., from same source) will only be stored once, but removal via `removeLethalDamageByPower` removes all instances of that timestamp. If two different abilities share timestamp, removal of one removes both.

Test case:
```java
Card c = new Card(...);
long ts1 = 100L;
c.addLethalDamageByPower(ts1);
c.addLethalDamageByPower(ts1); // duplicate, Set prevents duplicate
c.removeLethalDamageByPower(ts1);
// isLethalDamageByPower() returns false even if another ability with same ts should persist
```

Impact: Incorrect lethal damage calculation when multiple abilities share timestamps.

#### 4. CantGainControl check iterates over all static abilities without zone validation
Similar to BounceAtUntap, `StaticAbilityCantGainControl.cantGainControl` iterates over STATIC_ABILITIES_SOURCE_ZONES including Graveyard/Exile/Stack. Static abilities from cards in graveyard should not prevent control gain.

Test case:
```java
Card sourceInGraveyard = game.getCardsIn(ZoneType.Graveyard).get(0);
sourceInGraveyard.addStaticAbility(StaticAbilityMode.CantGainControl, "ValidCard", "Creature");
Card target = new Card(...);
boolean cant = StaticAbilityCantGainControl.cantGainControl(target);
// Returns true incorrectly
```

Impact: False positives for control gain restriction.

#### 5. StaticAbilityMustBeBlockedByAll.mustBeBlockedByAll null attacker/blocker
Method does not null-check attacker or blocker before accessing attacker.getGame(). NPE possible if called with null.

Test case:
```java
boolean result = StaticAbilityMustBeBlockedByAll.mustBeBlockedByAll(null, blocker);
// NullPointerException at attacker.getGame()
```

Impact: Crash during combat resolution.

#### 6. CombatUtil.canBlock returns false for mustBeBlockedByAll without checking blocker validity
The check `if (StaticAbilityMustBeBlockedByAll.mustBeBlockedByAll(attacker, blocker)) return false;` is placed before validating blocker is capable of blocking. If attacker is null or blocker null, NPE.

Impact: Potential crash.

#### 7. Card.isLethalDamageByPower returns true if any timestamp present, even if ability expired
Set is never cleared except via explicit removal. If timestamp removal fails due to exception, Set grows indefinitely.

Impact: Memory leak and incorrect state.

#### 8. Me the Immortal and Skullbriar changed from static ability to keyword without migration
Card definitions changed from `S:Mode$ CountersRemain` to `K:Counters remain on CARDNAME...`. Existing game state with active CountersRemain static abilities will not be recognized by new keyword-based system, causing counter loss on zone change.

Test case:
```java
// Load game with Me the Immortal with counters, change zone
// Old static ability would preserve counters, new keyword may not trigger
```

Impact: Backwards compatibility break for saved games.

#### 9. Lure card changed from AddHiddenKeyword to MustBeBlockedByAll mode
Card definition changed, but existing tests may expect hidden keyword behavior. Hidden keyword removal logic in StaticEffect.java only removes AddHiddenKeyword, not MustBeBlockedByAll.

Test case:
```java
// Remove static effect for Lure
// StaticEffect.remove() checks hasParam("AddHiddenKeyword") for hidden keywords
// MustBeBlockedByAll param not removed, leading to stale state
```

Impact: Inconsistent state cleanup.

#### 10. StaticAbilityMustBeBlockedByAll.applyMustBeBlockedByAll uses isValid with attacker as reference
`blocker.isValid(stAb.getParam("ValidBlocker"), attacker.getController(), attacker, stAb)` passes attacker as reference object. The `isValid` method may expect different reference semantics.

Test case:
```java
// ValidBlocker param "Creature" should check blocker type, not reference attacker
// Passing attacker as reference may cause incorrect validation
```

Impact: Incorrect blocking restrictions.

#### 11. StaticAbilityBounceAtUntap bypasses keyword check
Untap phase previously used keyword "During your next untap step..." to find bounce targets. New method adds static ability check but does not remove keyword check, causing duplicate processing.

Impact: Double bounce.

#### 12. Outpost Siege ForgetOnMoved change may break existing tests
Changed `ExileOnMoved$ Exile` to `ForgetOnMoved$ Exile`. Tests expecting ExileOnMoved trigger may fail.

Impact: Test failures.

### Performance and security
#### 13. Linear scan over all static abilities on each check
`mustBeBlockedByAll`, `cantGainControl`, `shouldBounceAtUntap` each iterate over all cards in STATIC_ABILITIES_SOURCE_ZONES and all static abilities per card. Called during combat/blocking checks which happen frequently. O(n*m) complexity.

Test case:
```java
// 100 cards on battlefield, each with 5 static abilities
// CombatUtil.canBlock called 1000 times per turn
// 100 * 5 * 1000 = 500,000 iterations
```

Impact: Performance degradation in large games.

## Verdict
Request Changes
The refactor introduces multiple correctness issues around zone validation, null safety, and backwards compatibility. Static ability checks iterate over inappropriate zones and lack null guards. Card definition changes from static abilities to keywords may break saved games. Performance concerns with linear scans during combat.

## Recommended Fixes
1. Add zone filtering to StaticAbility helper classes to only check Battlefield sources
2. Add null checks for attacker/blocker/card parameters
3. Use AbilityUtils to check ability applicability with proper zone validation
4. Preserve backwards compatibility for CountersRemain cards during migration
5. Add unit tests for null inputs and zone edge cases
6. Optimize static ability checks with caching

---

calaespi:
Superseded by a narrower rewrite on current master: https://github.com/Card-Forge/forge/pull/11805

That PR keeps only `CantGainControl` + `LethalDamageByPower` (as discussed with @Hanmac / @tool4ever), drops Lure/`MustBeBlockedByAll`, CountersRemain card flips, Outpost Siege/ForgetOnMoved, and unrelated test harness churn.

Closing this one in favor of #11805.

calaespi:
Closed in favor of #11805 (smaller, rebased rewrite).

PR #9707, opened 2026-02-10
Update dream_harvest.txt

dracontes:
[It came to my attention](https://bsky.app/profile/wotcmatt.bsky.social/post/3mbwhad3myk2a) that [Dream Harvest ](https://scryfall.com/card/ecl/216/dream-harvest) should work like [Tasha's Hideous Laughter](https://scryfall.com/card/afr/78/tashas-hideous-laughter). Tested to satisfaction.

In my rework of the script I hewed closer to Tasha's, which does result in as many command zone objects remembering exiled cards as there are opponents. My attempt to address that, linking the `Effect` line to the `RepeatEach` line, resulted in the same issue as described in a bug report on the Discord Server: only the first opponent in line gets the exile effect in full force while the remaining only exile one card.

### DIFF
--- a/forge-gui/res/cardsfolder/d/dream_harvest.txt
+++ b/forge-gui/res/cardsfolder/d/dream_harvest.txt
@@ -1,8 +1,12 @@
 Name:Dream Harvest
 ManaCost:5 UB UB
 Types:Sorcery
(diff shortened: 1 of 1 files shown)


Hanmac (approved):

dracontes:
> the script should already work fine, I introduced the extra param just for it to not have it messy like this

Checked on the most recent build. It does work as expected. My bad on not doing so sooner.

PR #9712, opened 2026-02-10
Various Tweaks - Adventure Event Ante and Keyboard Input

### LINKED ISSUE #9578: Add Page-up/down scroll functionality in inventory/deck.
Currently it's cumbersome to use the mouse wheel to scroll through the inventory, especially as the inventory increases to contain thousands of cards.

While it has been suggested that a scroll tab would be difficult to implement, I suggest instead that you could use the Page-Up/Down keys to scroll through the inventory. 

An alternative would be if you could select a card and use the basic arrow keys to navigate through the inventory.

Jetz72:
* Fixes an internal NPE that appeared [here](https://github.com/Card-Forge/forge/issues/9000#issuecomment-3448928169) (though obviously the performance hang with large numbers of tokens isn't going anywhere any time soon).
* Add support for Page Up/Down and the arrow keys to the mobile deck editor. Closes #9578.
* As part of the above - Previously arrow keys would always be treated as a fling gesture in the LibGDX port. Now they're passed to a keyDown event if the fling event goes unhandled. It's possible there are side effects since this tweak applies across all UIs, but I didn't encounter any in my testing.
* Ante results in adventure events now apply to the player's current event deck instead of their inventory.
* Ante results in adventure matches with multiple games now apply results from all games instead of just the last one.

### DIFF
--- a/forge-gui/src/main/java/forge/sound/EventVisualizer.java
+++ b/forge-gui/src/main/java/forge/sound/EventVisualizer.java
@@ -12,7 +12,6 @@ import forge.gui.events.UiEventBlockerAssigned;
 import forge.gui.events.UiEventNextGameDecision;
 import forge.util.TextUtil;
 
-import java.util.Collection;
 import java.util.Objects;
 
 import com.google.common.collect.Multimap;


Hanmac:
i picked the NPE fix

what also could have worked:
`player.equals(event.result().getWinningLobbyPlayer());`
because player should not be null?

Jetz72:
> what also could have worked:
> `player.equals(event.result().getWinningLobbyPlayer());`
> because player should not be null?

Wasn't sure at the time if player would be null while spectating a game with only AI players. Decided using `Objects.equals` was easier than digging through the code to find out if it actually worked that way.

PR #9713, opened 2026-02-10
Update abilities for Quintorius, History Chaser

tehdiplomat:


Hanmac:
i really need to remove the need for `Planeswalker$ True` in the future

tehdiplomat:
> i really need to remove the need for Planeswalker$ True in the future

Just inherit by loyalty costs instead?

Hanmac:
> > i really need to remove the need for Planeswalker$ True in the future
> 
> Just inherit by loyalty costs instead?

That's the plan for the Future

PR #9716, opened 2026-02-10
Remove redundant TriggerRemembered

tool4ever:
- this is mostly for _Ratonhnhaké:ton_, so that in case his tokens get doubled each reflexive trigger can only attach to the one from the associated instance.
- because triggers that remember something are always created fresh it seems rather pointless to copy them into SA

### DIFF
--- a/forge-game/src/main/java/forge/game/CardTraitBase.java
+++ b/forge-game/src/main/java/forge/game/CardTraitBase.java
@@ -747,9 +747,10 @@ public abstract class CardTraitBase implements GameObject, IHasCardView, IHasSVa
     }
 
     public List<Object> getTriggerRemembered() {
-        if (this instanceof SpellAbility sa) {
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/CardTraitBase.java
```diff
@@ -746,6 +746,13 @@ protected void copyHelper(CardTraitBase copy, Card host, boolean keepTextChanges
         copy.keyword = this.keyword;
     }
 
-    abstract public List<Object> getTriggerRemembered();
+    public List<Object> getTriggerRemembered() {
+        if (this instanceof SpellAbility sa) {
+            return sa.getTrigger().getTriggerRemembered();
```
Hanmac, line 751: Check if getTrigger isn't null

Hanmac:
@tool4ever is this MR ready?

tool4ever:
probably, just a couple of things I'll check later

PR #9717, opened 2026-02-10
Fix Outpost Siege interaction and update Card lethal logic (Issue #4745)

calaespi:
This PR fixes an issue with Outpost Siege where the exiled card interaction was using 'ExileOnMoved' instead of 'ForgetOnMoved', causing incorrect behavior when the card left the battlefield. It also updates Card.java to correctly check for 'LethalDamageByPower' static abilities in the getLethal() method.

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -6037,17 +6037,7 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
 
     // this is the amount of damage a creature needs to receive before it dies
     public final int getLethal() {
-        boolean lethalByPower = hasKeyword("Lethal damage dealt to CARDNAME is determined by its power rather than its toughness.");
-        if (!lethalByPower) {
-            for (final StaticAbility sa : this.getStaticAbilities()) {
-                if (sa.checkConditions(StaticAbilityMode.LethalDamageByPower)) {
-                    lethalByPower = true;
-                    break;
-                }
-            }
-        }
-
-        if (lethalByPower) {
+        if (hasKeyword("Lethal damage dealt to CARDNAME is determined by its power rather than its toughness.")) {
             return getNetPower();
         }
         return getNetToughness();
--- a/forge-game/src/main/java/forge/game/staticability/StaticAbilityMode.java
+++ b/forge-game/src/main/java/forge/game/staticability/StaticAbilityMode.java
@@ -53,8 +53,6 @@ public enum StaticAbilityMode {
 
     // StaticAbilityCombatDamageToughness
     CombatDamageToughness,
-    // StaticAbilityLethalDamageByPower
-    LethalDamageByPower,
 
     // StaticAbilityColorlessDamageSource
     ColorlessDamageSource,


tehdiplomat:
Why are three different "issues" being lumped together in a single PR?

tool4ever:
thanks for contributing
I'd recommend focusing on #9703 before doing more statics so you can better understand how they're supposed to work first

forge-game/src/test/java/forge/game/ability/ForgetOnMovedTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import java.util.ArrayList;
+
+public class ForgetOnMovedTest {
+
+    @BeforeClass
+    public void initLocalization() {
+        Localizer.getInstance().initialize("en-US", "/Users/calaespi/Desktop/Proyectos/Personales/forge/forge-gui/res/languages");
```
tool4ever: the way some of this is coded doesn't seem fully human 🤖

calaespi: Haha, I assure you I am human! 🤖 -> 👨‍💻. I had a hardcoded path for local testing and forgot to revert it. Fixed now to use a relative path logic.

calaespi:
@tehdiplomat I apologize for lumping the changes. They were discovered while working on the Outpost Siege interaction and felt related enough to include together. I will make sure to separate unrelated changes in future PRs.

@tool4ever Thanks for the recommendation on #9703, I will definitely check it out to better understand the static abilities. I have also pushed a fix for the hardcoded path you noticed.

calaespi:
Summary of this push: The branch no longer compiled because StaticAbilityCountersRemain had been removed but was still used in GameAction.java. I restored StaticAbilityCountersRemain.java and the CountersRemain value in StaticAbilityMode so that “counters remain” (e.g. Skullbriar) still works and the project compiles. I also removed the duplicate LethalDamageByPower entry in StaticAbilityMode (it was declared twice, which is invalid in Java) and removed unused imports in AbilityActivated.java (StaticAbility and StaticAbilityMode). Compilation and the relevant tests (ForgetOnMovedTest, HiddenKeywordsMetaTest, LureTest, LureMechanicTest, Issue4745Test) pass locally. Ready for review.

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -6026,7 +6026,17 @@ public final int getTotalDamageDoneBy() {
 
     // this is the amount of damage a creature needs to receive before it dies
     public final int getLethal() {
-        if (hasKeyword("Lethal damage dealt to CARDNAME is determined by its power rather than its toughness.")) {
+        boolean lethalByPower = hasKeyword("Lethal damage dealt to CARDNAME is determined by its power rather than its toughness.");
```
tool4ever: please remove this because it's already part of your other PR
and in that other PR everything you're doing here needs to be removed

tool4ever (changes requested):

forge-game/src/main/java/forge/game/card/Card.java
calaespi: Done. I’ve removed the lethal / LethalDamageByPower changes from this PR: reverted getLethal() in Card.java to the original keyword-only check and removed the LethalDamageByPower enum from StaticAbilityMode.java

PR #9718, opened 2026-02-10
Add ForgetOnMoved unit test and init localization (Pull Request #1)

calaespi:
This PR adds a unit test to verify that 'ForgetOnMoved' correctly excludes Stack, Exile, and Exiled zones. It also initializes Localizer/Lang to prevent NullPointerExceptions during GameType initialization, ensuring better stability for tests and gameplay.


tool4ever:
this is already in #9717
please avoid overlapping PR

PR #9724, opened 2026-02-12
Add options to set the user/cache dirs

RossBrunton:
This adds two new environment variables, `FORGE_USER_DIR` and `FORGE_CACHE_DIR` for overriding the default locations for user data.

For example, forge can be ran using:

```bash
FORGE_USER_DIR="$HOME/.local/state/forge" FORGE_CACHE_DIR="/tmp/forge" forge
```

To use alternate directories.

There should be no change in behaviour when neither of these vars are set.

### DIFF
--- a/docs/User-Guide.md
+++ b/docs/User-Guide.md
@@ -106,6 +106,9 @@ If you are using the Mac OS X version of Forge then you will find the forge.prof
 /Contents/Resources/Java/
 and you will find the file.
 
+The userDir's location may be changed by setting the environment variable "FORGE_USER_DIR" to the desired path. Likewise, the cache dir can be changed using the
+"FORGE_CACHE_DIR" environment variable.
+
 ## Import Data
 If you have a directory full of deck files, you can use the Import Data dialog to copy or move them to the appropriate directory. The dialog gives you a full listing of all file copy/move operations, so you can see what will happen before you click 'Start Import'.
 
--- a/forge-gui/src/main/java/forge/localinstance/properties/ForgeProfileProperties.java
+++ b/forge-gui/src/main/java/forge/localinstance/properties/ForgeProfileProperties.java
@@ -172,29 +172,52 @@ public class ForgeProfileProperties {
 
         final String fallbackDataDir = TextUtil.concatNoSpace(homeDir, "/.forge");
 
-        if (StringUtils.containsIgnoreCase(osName, "windows")) {
-            // the split between appdata and localappdata on windows is relatively recent.  If
-            // localappdata is not defined, use appdata for both.  and if appdata is not defined,
-            // fall back to a linux-style dot dir in the home directory
+        // Data Dir
+        String dataDir = null;
+        if (System.getenv().containsKey("FORGE_USER_DIR")) {
+            dataDir = System.getenv().get("FORGE_USER_DIR");
+        }
+        else if (StringUtils.containsIgnoreCase(osName, "windows")) {
+            // If appdata is not defined, fall back to a linux-style dot dir in the home directory
             String appRoot = System.getenv().get("APPDATA");
             if (StringUtils.isEmpty(appRoot)) {
                 appRoot = fallbackDataDir;
             }
-            String cacheRoot = System.getenv().get("LOCALAPPDATA");
-            if (StringUtils.isEmpty(cacheRoot)) {
-                cacheRoot = appRoot;
-            }
-            // the cache dir is Forge/Cache instead of just Forge since appRoot and cacheRoot might be the
-            // same directory on windows and we need to distinguish them.
-            return Pair.of(appRoot + File.separator + "Forge", cacheRoot + File.separator + "Forge" + File.separator + "Cache");
+            dataDir = appRoot + File.separator + "Forge";
         }
         else if (StringUtils.containsIgnoreCase(osName, "mac os x")) {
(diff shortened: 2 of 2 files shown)


tool4ever:
hmn what's the point?
using the preferences file should already cover this a bit more easily

RossBrunton:
I'm perhaps blind, but I can't see any documentation about a preferences file nor do I see a preferences file in the published build.

There's preferences information in the user dir, but that means that the user dir can't be relocated.

tool4ever:
it's right above the point in the guide you've edited

RossBrunton:
Had another look at this and it does look like that file can change the user data dir. For some reason I thought that file actually lived in the user dir and only controlled where the card images where stored. Sorry about that.

However, the properties file does have a few limitations:
* It has to be hardcoded in the forge package, so individual users can't change it unless they download their own copy of forge. If installed via a package manager, the installed package needs to be modified which is messy.
* As far as I can tell, it doesn't support the expansion of `$USER` or `~` (Or, presumably, `%APPDATA%` on Windows), meaning that the username has to be hardcoded.

With these two issues, I still think it makes sense to allow the user to override the locations using environment variables.

jamincollins:
## Findings

### Edge Cases and Null Safety
#### 1. FORGE_USER_DIR and FORGE_CACHE_DIR environment variable empty string handling
The code checks `System.getenv().containsKey("FORGE_USER_DIR")` and then immediately uses `System.getenv().get("FORGE_USER_DIR")` without checking for null or empty string. If the environment variable is set to an empty string, `dataDir` becomes empty string, which will cause `getDir` to canonicalize to an empty absolute path and then fail on file operations later.

Test case:
```java
// Simulate empty FORGE_USER_DIR
String dataDir = System.getenv().containsKey("FORGE_USER_DIR") ? System.getenv().get("FORGE_USER_DIR") : null;
// dataDir == "" -> File("").getAbsolutePath() returns working directory, leading to unintended data loss risk
```

Impact: Empty env var leads to data directory pointing to current working directory or empty string, causing file writes to wrong location and potential data loss.

### Backwards Compatibility
#### 2. DataDir calculation changes Windows fallback order
Previously, Windows fallback used `APPDATA` for both data and cache, with `LOCALAPPDATA` only for cache. Now dataDir is derived independently, and if `APPDATA` is missing, fallbackDataDir is used for dataDir, but cacheDir logic still references `dataDir` when `LOCALAPPDATA` is missing. This could change cache location for users who previously relied on `APPDATA` fallback.

Test case:
```java
// User with missing APPDATA env var previously got dataDir = fallbackDataDir, cacheRoot = fallbackDataDir
// Now dataDir = fallbackDataDir, cacheDir = dataDir + "/Cache" -> fallbackDataDir/Cache
// Previously cache was fallbackDataDir/Forge/Cache -> path change
```

Impact: Cache location moves for edge case users with missing APPDATA, breaking backwards compatibility.

#### 3. Cache dir depends on dataDir when LOCALAPPDATA missing
Line 206: `cacheDir = dataDir + File.separator + "Cache"` makes cache path dependent on user-configured dataDir. If user sets `FORGE_USER_DIR` to a custom path on Windows without `LOCALAPPDATA`, cache will be created under the custom user dir instead of a system cache location.

Test case:
```java
System.getenv().put("FORGE_USER_DIR", "/custom/forge");
System.getenv().remove("LOCALAPPDATA");
// cacheDir becomes "/custom/forge/Cache" instead of expected system cache
```

Impact: Cache pollutes user data directory, mixing user data with cache, and may cause permissions issues on shared systems.

### Performance and Security
#### 4. Environment variable injection without validation
FORGE_USER_DIR and FORGE_CACHE_DIR values are taken directly from environment without sanitization. Path traversal or non-existent paths are accepted.

Test case:
```java
System.getenv().put("FORGE_USER_DIR", "../../../tmp/evil");
// Data files written to unintended location
```

Impact: Security risk if environment is controlled by untrusted process; data may be written outside expected sandbox.

### Logic Correctness
#### 5. Missing trailing separator for env var paths
`getDefaultDirs` returns dataDir and cacheDir without trailing separator. Later code adds separator when building defaults, but `getDir` canonicalizes via `new File(retDir).getAbsolutePath()` and then ensures trailing separator. If env var value lacks trailing separator, it works, but if env var is set to path with trailing separator, `getAbsolutePath` may normalize it. However, the bigger issue: `getDefaultDirs` previously returned paths without trailing separator, and the change preserves that, but the code at line 225 adds separator for comparison: `defaults.getLeft() + File.separator`. If env var is set to a path that already ends with separator, comparison may be inconsistent.

Test case:
```java
// FORGE_USER_DIR="/opt/forge/" vs "/opt/forge"
// getAbsolutePath normalizes, but save() comparison may differ
```

Impact: Inconsistent property saving logic, may cause unnecessary profile file writes.

## Verdict
Request Changes
The PR adds useful environment variable overrides but introduces edge cases around empty env vars, Windows fallback cache relocation, and lack of validation. Needs defensive handling for empty strings, explicit cache path independence, and documentation of security considerations.

## Recommended Fixes
1. Treat empty FORGE_USER_DIR/FORGE_CACHE_DIR as unset: check for null and empty after getenv.
2. Preserve original cache location semantics on Windows when LOCALAPPDATA missing; do not make cache depend on dataDir unless explicitly intended.
3. Validate env var paths: reject empty, relative paths, or paths containing ".." before use.
4. Ensure consistent trailing separator handling for env var inputs.
5. Add unit tests for env var override scenarios including empty values, missing APPDATA, and cross-platform defaults.

---

PR #9725, opened 2026-02-12
Consolidate and update Network Play user documentation

MostCromulent:
My attempt at updating the Network Play wiki documentation in light of #9708 merge and consolidating the current 3 network play wiki files into a single self-contained guide, including essential content from all current files.


tool4ever (approved):
thanks, good call

I cleaned up some redundancy and kept one of the images as I felt it might still help some people

PR #9726, opened 2026-02-12
CloneAi: fix Mandatory Trigger

Hanmac:
This should fix `Silent Hallcreeper` ignoring the last trigger

* we probably should be able to remove the AI logic like `CloneAttacker`


Agetian:
Looks like there are some failing build checks, but otherwise looks good to me :)

Hanmac:
the AI logic should be clean up more, but that might be done later :tm:

tool4ever (approved):

Issue #9729, opened 2026-02-12
replace own MapOfLists with guava Multimap


Hanmac:
MapOfLists is only used in very few places with these two subclasses:
* EnumMapOfLists
* HashMapOfLists

all of them use ArrayList as Collection

might be cleaner to replace them with Multimap
https://github.com/google/guava/wiki/NewCollectionTypesExplained#multimap

PR #1830, opened 2022-11-08
Improve isCardAddedThisTurn

tool4ever:
This fixes previous zone changes e.g.:

- Destroy your creature
- Return it to your hand
- Discard it
- _Brought Back_ is still able to target it

PR #9732, opened 2026-02-12
Zone: use Multimap instead of MapOfLists

Hanmac:
Part of #9729

* uses ListMultimap instead of MapOfLists
* also use simpler from Zone check: https://github.com/Card-Forge/forge/pull/1830#issuecomment-1308376093

### DIFF
--- a/forge-game/src/main/java/forge/game/zone/Zone.java
+++ b/forge-game/src/main/java/forge/game/zone/Zone.java
@@ -20,11 +20,10 @@ package forge.game.zone;
 import java.util.*;
 import java.util.function.Predicate;
 
-import org.testng.collections.Maps;
-
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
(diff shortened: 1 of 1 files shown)


Hanmac:
@tool4ever there is [Multimaps#invertFrom](https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/Multimaps.html#invertFrom(com.google.common.collect.Multimap,M))

which would create a `Multimap<Card, ZoneType>`, but i haven't looked in which order the ZoneTypes are added

but more interesting would have been: [LinkedListMultimap](https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/LinkedListMultimap.html)

```
Multimap<K, V> multimap = LinkedListMultimap.create();
multimap.put(key1, foo);
multimap.put(key2, bar);
multimap.put(key1, baz);
```
> ... the iteration order for [Multimap.keys()](https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/Multimap.html#keys()) is ``[key1, key2, key1]`

but i don't think i can use my `enumKeys` on this one? 🤔

PR #9734, opened 2026-02-12
QuestEventDuelManager: replace MapOfLists with ListMultimap

Hanmac:
Part of #9729

* replaced MapOfLists with ListMultimap
* combined the getRandomOpponent Data into QuestDuel


Hanmac:
i wanted to combine more stuff, like the getRandomOpponent functions

but maybe later

PR #9738, opened 2026-02-13
Add logic so that the AI will play Sylvan Library

marthinwurer:
The goal is to let the AI use Sylvan Library if it's in a premade deck. The AI should just use it for card filtering, like Mirri's Guile. I have a test for casting, and a test for not paying life to keep more cards.

Disclaimer: this PR was mostly written by AI (Claude Code). I have reviewed the generated code. Let me know if any changes are needed!


forge-gui-desktop/src/test/java/forge/ai/AIIntegrationTests.java
```diff
@@ -113,4 +113,74 @@ public void testDoesNotCastRepopulateWhenNoCreaturesInOpponentGraveyard() {
 
         AssertJUnit.assertTrue("Repopulate must still be in hand", hand.contains(repopulate));
     }
+
+    @Test
+    public void testSylvanLibraryCanBeCast() {
```
tool4ever: this test is close to useless

marthinwurer: This was originally added because the AI wouldn't cast it at all because of the AI:RemoveDeck option in the card file. I can remove it entirely or add it to the other test, but it's not entirely useless because the previous behavior was just not casting it at all.

tool4ever: yea that's the intended behaviour
I don't think we'll need a test to confirm that for every card

marthinwurer: Removed.

tool4ever (commented):
mostly not in favor of a merge:
- not much investigation work seems to have been made, e.g. as long as the AI decides to filter only (which isn't even guaranteed) you're still using very simple heuristics which can easily cause more harm than good - especially in the early game
- so proving AI doesn't kill itself this way feels like roughly 10% of things to consider for this card
- I don't think that's an acceptable threshold, users will usually expect a bit more - especially when considering such an iconic card we'd just get more tickets for obvious failures

PR #9743, opened 2026-02-13
Fix Argentum Masticore to remember discarded card.

adrianrlaw:
Argentum Masticore failed to remember the discarded card when its UnlessCost is paid, resulting in the only valid targets being permanents with mana value zero. This PR makes Argentum Masticore remember the card discarded for its UnlessCost so that the card's mana value can be used for targeting.


tool4ever:
thanks for the find but it looks like this worked before I did #9716
need to debug what's going wrong :/

adrianrlaw:
I think regardless, it makes more sense for the remembered object to come from the UnlessCost being handled rather than what triggered the discard trigger for the destroy subability.

tool4ever:
it does not because if you counter the reflexive trigger you'd also prevent the cleanup now

adrianrlaw:
Can confirm is fixed in #9745

PR #9746, opened 2026-02-14
Add keyboard shortcuts cheatsheet to desktop match UI

MostCromulent:
## Summary                                                                                                                                        
- Adds a keyboard shortcuts cheatsheet to the desktop UI, accessible via configurable key (default "H" for "Hotkeys") during matches and from Help > Keyboard Shortcuts in the menu bar.
- Configurable shortcuts are shown with in-place rebinding (click a field, press a new key to replace the binding)
- Configurable shortcuts auto-populate by iterating KeyboardShortcuts.getCachedShortcuts(), so if a new SHORTCUT_* FPref entry is added to Forge in future it will automatically appear in the dialog without needing changes to the dialog code.
- Hard coded shortcuts (F1, F11, Ctrl+Z, Ctrl+Q, Ctrl+A, Ctrl+E, Ctrl+T, Ctrl+O) are hardcoded in the dialog because they are defined across separate menu classes (GameMenu, LayoutMenu, CardOverlaysMenu, HelpMenu) with no shared registry, so there is no list to iterate.

Untitled

Screenshot 2026-02-14 131807

<sup>(**Note:** In images above Concede, Alpha Strike, and End Turn appear in both sections. That is because they currently exist as both configurable shortcuts (though default for concede is non-functional — assembleKeystrokes treats CTRL as modifier rather than key press). The fixed Ctrl+Q/A/E menu accelerators are separate bindings tied to the menu bar items. Both systems trigger the same game actions. This is **existing Forge behavior**, not anything added in this PR - while outside scope of this PR, it may be worth streamlining this in future.)</sup>

---

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/screens/home/settings/VSubmenuPreferences.java
+++ b/forge-gui-desktop/src/main/java/forge/screens/home/settings/VSubmenuPreferences.java
@@ -608,6 +608,7 @@ public enum VSubmenuPreferences implements IVSubmenu<CSubmenuPreferences> {
                     FModel.getPreferences().setPref(prefKey, getCodeString());
                     FModel.getPreferences().save();
                     shortcut0.attach();
+                    FControl.instance.getForgeMenu().refresh();
                     KeyboardShortcutField.this.setBackground(Color.white);
                 }
             });


MostCromulent:
While we're here -  Concede, Alpha Strike, and End Turn exist twice: as _both_ hard-coded shortcuts and configurable shortcuts. It appears the hard coding may be so that the shortcut appears in the match UI game menu.

Wouldn't it be better to remove the hard-coding, make them configurable but with the current hard-coded keys as defaults, and just make the menu dynamically reflect whatever the assigned shortcut is?

tool4ever:
I can't think of a good reason why the hardcoded ones would need to continue existing.
Most likely they were around before the configurable ones and never got integrated (perhaps because the menu key display causes conflicts if not handled?)
Otherwise a streamlining seems desirable...

MostCromulent:
Fixed. Only hard-coded shortcuts remaining are Help and Full Screen.

Screenshot 2026-02-22 091057

tehdiplomat:
This changed my alpha strike from `A` to `Ctrl+A` I bet a lot of people are going to complain about that change as it speeds the game up a lot. (Your original screenshot even shows the command as `A`)

MostCromulent:
> This changed my alpha strike from `A` to `Ctrl+A` I bet a lot of people are going to complain about that change as it speeds the game up a lot. (Your original screenshot even shows the command as `A`)

I think this is a result of discussion above about migrating redundant dual hardcoded + configurable shortcuts into just configurable ones; the preference migration logic kept the hard-coded form if the user's configurable one was the the default (e.g. hard-coded was CTRL+A while configurable default was just A).

In any event I think @tool4ever has removed the preference migration entirely as of today's snapshot to avoid issues.

PR #9748, opened 2026-02-14
Dev menu improvements: "Add Token to Battlefield" + search + quantity select

MostCromulent:
Screenshot 2026-02-14 172821
Screenshot 2026-02-14 172838
Screenshot 2026-02-14 180209
Screenshot 2026-02-14 180223

3 additions to dev menu functionality to make it easier to work with tokens:

---

**Add Token to Battlefield:** Adds a new "Add Token to Battlefield" option to the dev panel, wired through the same CDev/VDev/DevModeMenu/IDevModeCheats pattern as existing cheats. Tokens are deduplicated by name and lazy-loaded from TokenDb on first use. Creature tokens prompt for summoning sickness (or auto-set it for tokens with haste).

**Quantity selector for card additions:** The existing "Add Card to Zone" dev cheats (hand, battlefield, library, graveyard, exile) now prompt for a quantity after selecting a card, allowing multiple copies to be added at once. Prompts that depend on the card (summoning sickness, top/bottom of library) are asked once and reused for all copies. The entire quantity loop runs within a single invoke() call to avoid race conditions from concurrent zone modifications.

**ListChooser search field**: Adds an inline search/filter text field to the ListChooser dialog **for lists with more than 25 items**, matching the existing mobile ListChooser behavior. Search results prioritize starts-with matches (sorted by name length for closer matches first), followed by contains matches. 

This has broader application than just the dev menu, but there are few dialogs that often have more than 25 entries so in practice this is only likely to affect:
    - Dev menu card/token selection (10,000+ card faces, 1,000+ tokens)
    - Card naming during gameplay (e.g. Pithing Needle, Cabal Therapy — all legal card names)

In any event this aligns with list functionality in mobile, and is probably a QOL improvement anyway.

---


Hanmac:
probably need to add PT to the Name select if its not a named token
(not ending with "Token")

dracontes:
I have to say, the search field on the "Add Card to [Zone]" prompt has been on my personal wishlist for a fair bit. Tested out that part of this PR and it works as well as I'd want and expect 👍 

I did test the "Add Token to Battlefield" part and searching for either named or game-defined noncreature tokens turns up this, which I don't think is at all the expected behavior.

image

Also I think "Name the token" at the top of the prompt would be more apt.

ETA: And perhaps have the token search ignore dungeon cards.

image

MostCromulent:
> I did test the "Add Token to Battlefield" part and searching for either named or game-defined noncreature tokens turns up this, which I don't think is at all the expected behavior.

> Also I think "Name the token" at the top of the prompt would be more apt.

> ETA: And perhaps have the token search ignore dungeon cards.

All 3 issues addressed on latest commit (code was previously trying to assign power/toughness to non-creature tokens). 

Also improved the de-duplication logic so it doesn't collapse same-named creature tokens into one entry if they have different colors or different abilities. Will now disambiguate same-name tokens by P/T, color and, if necessary, oracle text.

Screenshot 2026-02-16 190752 Screenshot 2026-02-16 190806

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -2851,23 +2876,35 @@ public void addTokenToBattlefield() {
...
 
-            // Display name with P/T in brackets
-            final FSerializableFunction<PaperToken, String> displayFn = pt -> {
-                ICardFace face = pt.getRules().getMainPart();
-                return pt.getName() + " (" + face.getIntPower() + "/" + face.getIntToughness() + ")";
+            // Find which short labels are ambiguous so we can append oracle text only for those
+            final Map<String, Integer> labelCounts = new HashMap<>();
+            for (final PaperToken pt : choices) {
+                labelCounts.merge(tokenLabel(pt), 1, Integer::sum);
+            }
```
Hanmac: In theory, each token whose name ends with `" Token"` can have duplicated names
each other one is a "Named Token" and should not be ambiguous

dracontes:
> All 3 issues addressed on latest commit (code was previously trying to assign power/toughness to non-creature tokens).

These issues have been adequately addressed.
 
> Also improved the de-duplication logic so it doesn't collapse same-named creature tokens into one entry if they have different colors or different abilities. Will now disambiguate same-name tokens by P/T, color and, if necessary, oracle text.

I was going to ask about this, so good show on it for the most part. I did notice that while the` \tokenscripts` folder shows two Pest tokens, to wit, those produced by [Nuisance Engine](https://scryfall.com/card/mrd/221/nuisance-engine) and the [Strixhaven ones](https://scryfall.com/card/tstx/5/pest), the prompt search only turns up the latter.

image

Jetz72:
Any thoughts on adding this to the mobile/adventure UI's dev menu as well?

MostCromulent:
> Any thoughts on adding this to the mobile/adventure UI's dev menu as well?

Done - now applies to mobile UI as well.

MostCromulent:
> I did notice that while the` \tokenscripts` folder shows two Pest tokens, to wit, those produced by [Nuisance Engine](https://scryfall.com/card/mrd/221/nuisance-engine) and the [Strixhaven ones](https://scryfall.com/card/tstx/5/pest), the prompt search only turns up the latter.

Root cause is that the new addTokenToBattlefield() method called tokenDb.allTokensByName, which loads tokens listed in edition files. There are 114 token scripts that aren't assigned to an edition so don't get picked up in the list. The Nuisance Engine Pest is one of those.

I've changed it to call TokenDb.rulesByName instead, which is edition agnostic and contains all the tokens.

(Note it seems tokens like Nuisance Engine pest without an assigned edition are missing auto-download card art because they are from older sets without printed token cards on Scryfall).

dracontes:
> I've changed it to call TokenDb.rulesByName instead, which is edition agnostic and contains all the tokens.
> 
> (Note it seems tokens like Nuisance Engine pest without an assigned edition are missing auto-download card art because they are from older sets without printed token cards on Scryfall).

All working fine now. Admittedly the lack of token images doesn't concern me too much as I haven't downloaded images for Forge in a long while: Forge's built-in card rendering is enough for me.

PR #9750, opened 2026-02-14
Fix Twilight Diviner

tool4ever:


Hanmac:
i did the same fix with https://github.com/Card-Forge/forge/tree/zoneFixLKI

PR #9751, opened 2026-02-14
Group identical permanents into numbered stacks

### LINKED ISSUE #2556: Treasure tokens with counter instead of separate cards
Screenshot_20230225_014644_Forge
As you can see in the screenshot above, I was at over 300 treasure tokens at this point. I tried to get another combat step but the game froze. The additional combat would have put me to something close to 4000 treasures. I wouldn't even know how the game should display all those as separate tokens. A possible solution could be that if you get over a certain amount of treasures, instead of separate tokens, it just displays one treasure token with a counter of how many tokens you have.

MostCromulent:
**BEFORE (Default settings)**
image
**AFTER (Group all permanents)**
Screenshot 2026-04-03 114709

### Summary

Adds a "Stack/Group Permanents" submenu to the Game menu with four options:
- **Default** - existing behavior (lands/tokens/artifacts/enchantments stacked with fan-out, creatures unstacked)
- **Stack Creatures** - existing behavior from "Stack Creatures" checkbox, also stacks creatures 
- **Group Creatures/Tokens (NEW)** - identical creatures and tokens are shown as a single compact pile of up to 4 visible cards, with a xN count badge on top
- **Group All Permanents (NEW)** - groups all identical permanents with xN count badges

Screenshot 2026-04-03 114351

### Grouping behavior
- Identical permanents (same name, P/T, counters, abilities, tapped state, damage, no attachments) are collapsed into overlapping stacks of up to 4 visible cards with a xN badge on top.
- When any criterion changes (one gets tapped, takes damage, gains a counter), that permanent separates from its group into a separate file
- Blockers assigned to different attackers are kept in separate piles for visual separation; attackers with different blockers are also kept in separate piles so you can easily track group combat assignments.
- Grouping respects the existing "Tokens in separate row" preference. 
- UI updates immediately when grouping preference changed, no need to restart match. 
- Grouping is **UI layer only** - as far as game engine is concerned all grouped objects are still processed separately. (Note this means e.g. selecting a pile of Treasure Tokens will still prompt you 1-by-1 to sacrifice and select mana color, rather than processing whole batch together. Batch processing could be improved in future but requires engine changes beyond scope of this PR).

### Badge interaction
- **Left-click badge** - select all cards in the group (e.g. tap all lands for mana)
- **Left-click badge on tapped group** - undo last action per card (reverse a batch mana tap)
- **Left-click individual card** in a group - split it out for individual declaration/selection; re-click to merge back
- **Right-click badge** -prompt for how many to declare as attackers/blockers, remove from combat, or select for sacrifice/targeting

Screenshot 2026-04-03 115046

Screenshot 2026-04-03 115113

Screenshot 2026-04-03 122949

---


paulsnoops:
Looks cool! Does it interfere or replace this?

image

MostCromulent:
> Does it interfere or replace this?

It respects the token row preference (i.e. it will group both creatures and tokens but place them in separate rows).

It overrides the existing stack preference if its enabled, since it's kind of built on top of it. The preferences could probably be integrated together since they are doing a similar sort of thing, so you end up with a "Group and Stack Permanents" option with:
- **Off:** default behaviour _without_ the existing stack preference.
- **Stack:** existing stack preference (no grouping).
- **Group creatures/tokens:** new functionality.
- **Group all permanents:** new functionality.

Is the "group" vs "stack" naming intuitive? When I think of a "stack" I think more along the lines of what this PR does where it puts everything in one big pile than what the existing stack preference does, but I don't want to go and re-label existing functionality since people are used to it.

MostCromulent:
I note current default behavior with the existing "stack creatures" preference turned on is that creatures + enchantments + artifacts are in stacks of 4, while lands + tokens are in stacks of 5.

Is there a reason for that difference?

If we integrate the stack/group functions together should we replace with a default stack size of 5 for everything, or should we flow this on and maintain the 4 vs 5 stack size for the new grouping function as well?

MostCromulent:
You can now right-click the number badge and it will prompt how many creatures/tokens from the pile you want to declare, rather than needing to click them individually.

Screenshot 2026-02-15 101856
Screenshot 2026-02-15 101909

paulsnoops:
> > Is the "group" vs "stack" naming intuitive? When I think of a "stack" I think more along the lines of what this PR does where it puts everything in one big pile than what the existing stack preference does, but I don't want to go and re-label existing functionality since people are used to it.

The only thing I'd say here is the use of the word stack (do we already use it in this context?) as MtG uses that specifically in the rules meaning a different thing, I don't know if it's really an issue though @tehdiplomat what do you think?

paulsnoops:
> I note current default behavior with the existing "stack creatures" preference turned on is that creatures + enchantments + artifacts are in stacks of 4, while lands + tokens are in stacks of 5.
> 
> Is there a reason for that difference?
> 
> If we integrate the stack/group functions together should we replace with a default stack size of 5 for everything, or should we flow this on and maintain the 4 vs 5 stack size for the new grouping function as well?

Not sure honestly, I'd assume because most cards have a maximum of 4 in a deck so there was no reason to be any more, and tokens/lands were just set that way, I don't think it matters from a rules point of view

Madwand99:
Just want to say that I love this PR and hope it gets through. I was considering implementing something similar but hopefully that won't be necessary now. I *did* end up implementing the part that separates out attackers and blockers from stacks, so there is some overlap there, but perhaps it's OK to separate the functionality into different PRs for easier reviewing.

MostCromulent:
@Madwand99 Thanks for testing! Did you notice any issues when trying it?

I haven't updated this PR in a while just because I've been focussed on others, but it seemed to be in a pretty stable state without obvious bugs. 

If that was your view too I can probably just mark it ready for review.

Madwand99:
I can't say I've been able to test thoroughly yet (Jhoira AI killed itself on my Impact Tremors), but I can say that in my tests it was a relief having 40 tokens displayed so compactly.

MostCromulent:
I'm going to mark this ready for review - haven't noticed any obvious bugs, and it would be good to get some fresh eyes on it for testing.

tool4ever:
> > I note current default behavior with the existing "stack creatures" preference turned on is that creatures + enchantments + artifacts are in stacks of 4, while lands + tokens are in stacks of 5.
> > Is there a reason for that difference?
> > If we integrate the stack/group functions together should we replace with a default stack size of 5 for everything, or should we flow this on and maintain the 4 vs 5 stack size for the new grouping function as well?
> 
> Not sure honestly, I'd assume because most cards have a maximum of 4 in a deck so there was no reason to be any more, and tokens/lands were just set that way, I don't think it matters from a rules point of view

I wonder if it looks nicer if both have the same default value 🤔
Either way a preference for user to adjust maximum might be neat?

forge-gui-desktop/src/main/java/forge/view/arcane/PlayArea.java
```diff
@@ -660,10 +709,105 @@ public final void mouseLeftClicked(final CardPanel panel, final MouseEvent evt)
...
+            if (p.getCard() != null) {
+                selected.add(p.getCard());
+            }
+        }
+        List<CardView> others = selected.size() > 1 ? selected.subList(1, selected.size()) : null;
+        boolean isCombat = alreadyInCombat
+                || activateDesc.equals(loc.getMessage("lblAttackWithCard"))
+                || activateDesc.equals(loc.getMessage("lblBlockWithCard"));
+        if (alreadyInCombat) {
+            MouseTriggerEvent rightClickTrigger = new MouseTriggerEvent(3, evt.getX(), evt.getY());
```
tool4ever: relying on these synthetic clicks seem rather shaky, what's happening here and what are the alternatives?

MostCromulent: > **What they do / why they matter**
> 
> In this codebase, left-click means "declare/select" and right-click means "undeclare/cancel" — `InputAttack` and `InputBlock` read the mouse button off the trigger event to decide which one. The badge dialog complicates that, because the user has already right-clicked the badge before any game action is taken; that gesture is just opening a count picker, not yet expressing declare-vs-undeclare. By the time we know which action to perform, the original right-click no longer carries the right meaning, so we hand the input handler a fabricated `MouseTriggerEvent` whose button code matches the chosen action: button=3 to undeclare, button=1 to declare.
> 
> Concretely, the two checks that matter are `InputAttack.onCardSelected` line 175 and `InputBlock.onCardSelected` line 110 — both gate on `triggerEvent.getButton() == 3` to take the undeclare branch. So `alreadyInCombat` synthesizes button=3 to undeclare, the in-combat declare branch synthesizes button=1 to declare, and the non-combat branch synthesizes button=1 just for consistency.
> 
> Redundancy I found while investigating: the `alreadyInCombat` synthetic (`new MouseTriggerEvent(3, evt.getX(), evt.getY())`) is redundant — `evt` is already a right-click, so `new MouseTriggerEvent(evt)` would do. The non-combat synthetic is also redundant — non-combat inputs (`InputSelectTargets`, `InputPayMana`, etc.) don't read the button at all. I grepped all `ITriggerEvent.getButton()` consumers in `forge-gui` and `forge-gui-desktop`; the two `==3` checks above are the only ones. So "button=1" vs "button=3" is the entire protocol; everything else ignores it. 
>
>The in-combat declare synthetic is the only site that's actually load-bearing: the user right-clicked the badge but we want to *declare* (not undeclare), so we have to hand the handler a button=1 event to reach the declare branch. Without it, right-clicking the badge to declare would silently take the undeclare path on a stack with no declared attackers and no-op.
> 
> **Proposed changes**
> 
> Use `new MouseTriggerEvent(evt)` for both the `alreadyInCombat` and non-combat branches — drops two synthetic constructions. Keep the synthetic `button=1` for the in-combat declare branch and add a comment explaining why we're flipping the button code. 
> 
> Out of scope here, but the cleaner long-term fix is to add explicit `declare(card, others)` / `undeclare(card, others)` entry points on `InputAttack`/`InputBlock` so the badge handler bypasses the mouse-button channel entirely. InputAttack already has private declareAttacker / undeclareAttacker helpers. The expensive part is reaching them from the badge handler. The handler talks to the game via getMatchUI().getGameController().selectCard(...) — it doesn't hold a direct reference to the active input. So you'd need new methods on the controller interface (declareCards(card, others) / undeclareCards(card, others) or a single batched selectCards(card, others, BatchAction)), InputProxy dispatch for them, and — per the codebase's ProtocolMethod rule —corresponding entries in the network protocol enum so net games see them.

tool4ever: hmn alright, then only the minor cleanup sounds ok for this isolated situation

MostCromulent:
> I wonder if it looks nicer if both have the same default value 🤔
> Either way a preference for user to adjust maximum might be neat?

@tool4ever have implemented this. Stack depth is unified across all card types, but subject to a user-facing "max stack depth" preference (default 4, range 1-10).

Screenshot 2026-04-30 070506

MostCromulent:
Resolved merge conflicts with https://github.com/Card-Forge/forge/pull/10538

**Group permanents ON / separate combatants OFF** - already groups same attacker/blocker sets into piles
Screenshot 2026-05-02 070228

**Group permanents ON / separate combatants ON** - splits the piles out individually.
Screenshot 2026-05-02 070240

Also added the 'separate combatants' preference to in-match menu underneath 'put tokens in separate row' preference; seems like they belong together and are sort of thing you might want to change mid-match.

tool4ever (approved):

PR #9752, opened 2026-02-14
Realm of legends updates - deck, item, and map fixes

VigilantSera:


dracontes:
Comparing files on my side and it turns out there's two lines with `Alena, Kessig Trapper` on `\Realm of Legends\decks\legends\dragonhawk.dck` (lines 9 and 16).

PR #9756, opened 2026-02-15
Optional Smart Token Art Selection

clairchiara:
When token art for a specific set is missing, Forge currently selects a random token art. This PR introduces an optional deterministic fallback strategy based on set release order.
When art isn’t found for a given set, it goes through every set in release order and tries to find the latest art released before the set or, if still not found, the earliest art released after the set.


forge-core/src/main/java/forge/token/TokenDb.java
```diff
@@ -84,14 +87,48 @@ protected PaperToken addTokenInSet(CardEdition edition, String name, CardEdition
...
+        PaperToken coreExpansionReprintToken = smartFallbackToken(name, realEdition, true);
+        return coreExpansionReprintToken != null ? coreExpansionReprintToken : smartFallbackToken(name, realEdition, false);
+    }
+
+    private PaperToken smartFallbackToken(String name, CardEdition realEdition, boolean onlyCoreExpansionOrReprint) {
+        String latestFound = null;
+        boolean pastRealEdition = false;
+        final EnumSet<CardEdition.Type> coreExpansionOrReprint = EnumSet.of(CardEdition.Type.CORE, CardEdition.Type.EXPANSION);
+        coreExpansionOrReprint.addAll(CardEdition.Type.REPRINT_SET_TYPES);
+        for (CardEdition edition : this.editions.getOrderedEditions(false)) {
```
Hanmac, line 124: No need for extra function.
`editions` is already an `CardEdition.Collection` extending `Iterable<CardEdition>`

clairchiara: The goal is to find the closest edition to realEdition rather than stop at the first result (which would be the earliest release)

Hanmac: @clairchiara no i mean this:

`edition : this.editions.getOrderedEditions(false)` returns Sorted Collection, but not reversed

Wouldn't `edition : this.editions` return the same? No need for extra `getOrderedEditions` with different parameter?
it is already an Iterable

clairchiara: `this.editions` isn't sorted as far as I can tell

clairchiara:
Messed up the rebase…

clairchiara:
It doesn’t seem that this is required anymore (https://github.com/Card-Forge/forge/pull/10518)

Issue #9739, opened 2026-02-13
Remove MapToAmount


Hanmac:
These extra classes should not be needed

in most cases, they can be replaced with grouping and counting

Hanmac:
[GlobalAttackRestrictions](https://github.com/Card-Forge/forge/blob/master/forge-game/src/main/java/forge/game/combat/GlobalAttackRestrictions.java) and its internal class GlobalAttackRestrictionViolations have many unused code
like `MapToAmount<GameEntity> defenderMax` doesn't make much sense if it is just used as a `Map<GameEntity,Long>` anyway
`GlobalAttackRestrictionViolations` all properties are unused, and `isViolated` is just calculated anyway
`returnQuickly` is always true, the method with false isn't called

kojotak:
My attempt to remove MapToAmount related classes: https://github.com/Card-Forge/forge/pull/9744

Hanmac:
Completed

PR #9757, opened 2026-02-15
AttackConstraints: rework playerRequirements as Multimap

Hanmac:
Part of #9739 

Uses `Multimap<GameEntity, StaticAbility>` as better List.

(in theory, countViolations could return the StaticAbility it is violating)

### DIFF
--- a/forge-game/src/main/java/forge/game/combat/AttackConstraints.java
+++ b/forge-game/src/main/java/forge/game/combat/AttackConstraints.java
@@ -323,8 +323,10 @@ public class AttackConstraints {
                 bestMatch.requirements += playerReq.getValue().size();
                 usedAttackers.add(bestMatch.attacker);
                 // recalculate remaining requirements
+                playerReqs.values().removeAll(playerReq.getValue());
+            } else {
+                playerReqs.removeAll(playerReq.getKey());
             }
-            playerReqs.removeAll(playerReq.getKey());
         }
         if (!usedAttackers.isEmpty()) {
             // order could have changed


forge-game/src/main/java/forge/game/combat/AttackConstraints.java
```diff
@@ -310,35 +310,28 @@ private List<Attack> getSortedFilteredRequirements() {
...
+                    .max(Comparator.comparing(e -> e.getValue().size())).orElse(null);
             // find best attack to also fulfill the additional requirements
-            Attack bestMatch = Iterables.getLast(IterableUtil.filter(result, att -> !usedAttackers.contains(att.attacker) && att.defender.equals(playerReq.getLeft())), null);
+            Attack bestMatch = result.stream().filter(att -> !usedAttackers.contains(att.attacker) && att.defender.equals(playerReq.getKey())).findFirst().orElse(null);
             if (bestMatch != null) {
-                bestMatch.requirements += playerReq.getRight();
+                bestMatch.requirements += playerReq.getValue().size();
                 usedAttackers.add(bestMatch.attacker);
                 // recalculate remaining requirements
-                playerReqs.removeIf(s -> s.contains(playerReq.getLeft()));
```
tool4ever, line 328: this refactor doesn't work:
when a viable Attack is found you need to remove all other requirements which the GameEntity can satisfy
`playerReqs.values().removeAll(playerReq.getValue());`
and the call below can stay "else"

tool4ever (approved):

PR #9760, opened 2026-02-15
GameEvent refactor

MostCromulent:
@tool4ever here is the GameEvent refactor.

This has been done **[based on the refactor plan](https://github.com/MostCromulent/forge/blob/NetworkPlay/dev/.documentation/eventrefactor.md)** which has been updated to reflect implementation. See [implementation notes section](https://github.com/MostCromulent/forge/blob/NetworkPlay/dev/.documentation/eventrefactor.md#implementation-notes-deviations-from-plan) for deviations from original plan.

I have not had a chance to test in depth yet - but unit test is succesful and local network testing indicates client is at least playing sounds now.

In relation to subgame events Claude ran into an issue, summarised below:

> Subgame Event Reclassification — Options
> 
> The Problem
> 
> The event refactor (Step 5) added blanket forwarding of all GameEvents to remote clients via handleGameEvent(). The two subgame events (GameEventSubgameStart, GameEventSubgameEnd) contain Game references — not serializable, and their handlers do host-only GUI wiring (subscribing to EventBus, switching GameView, iterating PlayerControllerHuman). Forwarding these to a remote client would fail at serialization and is conceptually wrong — they're local lifecycle signals.
> 
> Per your earlier feedback, these should be reclassified as UiEvent to create a clean dividing line: GameEvent = networked game state, UiEvent = local GUI concerns.
> 
> The Constraint
> 
> SubgameEffect (in forge-game) creates and fires both events. UiEvent is defined in forge-gui. forge-game cannot depend on forge-gui — the dependency goes the other way. There is no existing precedent for forge-game code firing UiEvents; all 3 current UiEvent types are created exclusively in forge-gui input handlers.
> 
> Option A: Filter in Forwarding Code
> 
> Skip subgame events in the HostedMatch.receiveGameEvent() forwarding loop with an instanceof check.
> 
> - Pros: 2-line change, zero architectural disruption
> - Cons: No type-system enforcement — the dividing line is a runtime check, not a compile-time guarantee. Future GameEvent types that shouldn't be forwarded would need additional filters.
> 
> Option B: LocalGameEvent Marker Interface in forge-game
> 
> Create a new interface LocalGameEvent extends GameEvent in forge-game. Subgame events implement LocalGameEvent instead of GameEvent directly.
> Forwarding code skips any LocalGameEvent instance.
> 
> - Pros: Type-system enforced dividing line within forge-game. Any future host-only events can implement LocalGameEvent. No module boundary issues.
> - Cons: Requires removing subgame visit methods from IGameEventVisitor and updating all 5 visitor implementations. A new concept (LocalGameEvent) that's effectively a subset of GameEvent — not as clean as GameEvent vs UiEvent.
> 
> Option C: Replace Events with Callback Hooks
> 
> Remove the subgame events entirely. SubgameEffect calls hooks on Game or Match (like the existing startGameHook/endGameHook pattern in
> HostedMatch). HostedMatch registers the hooks during game setup.
> 
> - Pros: Cleanest separation — subgame lifecycle exits the event system entirely. No forwarding concern. Follows existing precedent (startGameHook, endGameHook).
> - Cons: Most invasive. Requires refactoring SubgameEffect to use callbacks, and moving the GUI wiring logic from event visitors into hook implementations. Changes the execution model (hooks are synchronous callbacks vs EventBus dispatch).

### DIFF
--- a/forge-game/src/main/java/forge/game/spellability/SpellAbilityView.java
+++ b/forge-game/src/main/java/forge/game/spellability/SpellAbilityView.java
@@ -6,7 +6,6 @@ import com.google.common.collect.Maps;
 
 import forge.game.card.CardView;
 import forge.game.card.IHasCardView;
-import forge.game.player.PlayerView;
 import forge.trackable.TrackableObject;
 import forge.trackable.TrackableProperty;
 import forge.trackable.Tracker;
@@ -34,7 +33,7 @@ public class SpellAbilityView extends TrackableObject implements IHasCardView {
         updateHostCard(sa);
         updateDescription(sa);
         updatePromptIfOnlyPossibleAbility(sa);
-        // Note: updateIsSpell, updateIsTrigger, updateActivatingPlayer are NOT called
+        // Note: updateIsSpell NOT called
         // here because subclasses (e.g. WrappedAbility) may not be fully initialized yet
         // during super() construction. These are updated lazily in SpellAbility.getView().
     }
@@ -77,20 +76,6 @@ public class SpellAbilityView extends TrackableObject implements IHasCardView {
     }
     void updateIsSpell(SpellAbility sa) {
         set(TrackableProperty.SA_IsSpell, sa.isSpell());
-    }
-
-    public boolean isTrigger() {
-        return get(TrackableProperty.SA_IsTrigger);
-    }
-    void updateIsTrigger(SpellAbility sa) {
-        set(TrackableProperty.SA_IsTrigger, sa.isTrigger());
-    }
-
-    public PlayerView getActivatingPlayer() {
-        return get(TrackableProperty.ActivatingPlayer);
-    }
-    void updateActivatingPlayer(SpellAbility sa) {
-        set(TrackableProperty.ActivatingPlayer, PlayerView.get(sa.getActivatingPlayer()));
     }
 
     @Override
--- a/forge-gui/src/main/java/forge/gamemodes/match/AbstractGuiGame.java
+++ b/forge-gui/src/main/java/forge/gamemodes/match/AbstractGuiGame.java
@@ -36,7 +36,6 @@ public abstract class AbstractGuiGame implements IGuiGame, IMayViewCards {
     private final Map<PlayerView, IGameController> gameControllers = Maps.newHashMap();
     private final Map<PlayerView, IGameController> originalGameControllers = Maps.newHashMap();
     private boolean gamePause = false;
-    private boolean gameSpeed = false;
     private PlaybackSpeed playbackSpeed = PlaybackSpeed.NORMAL;
     private String daytime = null;
     private boolean ignoreConcedeChain = false;
@@ -331,14 +330,13 @@ public abstract class AbstractGuiGame implements IGuiGame, IMayViewCards {
         return gamePause;
     }
 
-    public boolean isGameFast() {
-        return gameSpeed;
-    }
-
-    public void setgamePause(boolean pause) {
+    public void setGamePause(boolean pause) {
         gamePause = pause;
     }
 
+    public PlaybackSpeed getGameSpeed() {
+        return playbackSpeed;
+    }
     public void setGameSpeed(PlaybackSpeed speed) {
         playbackSpeed = speed;
     }
@@ -915,6 +913,16 @@ public abstract class AbstractGuiGame implements IGuiGame, IMayViewCards {
         return showConfirmDialog(message, title, yesButtonText, noButtonText, true);
     }
 
+    private FControlGameEventHandler localEventHandler;
+
+    @Override
+    public void handleGameEvent(GameEvent event) {
+        if (localEventHandler == null) {
+            localEventHandler = new FControlGameEventHandler(this);
+        }
+        localEventHandler.receiveGameEvent(event);
+    }
+
     @Override
     public void notifyStackAddition(GameEventSpellAbilityCast event) {
     }
@@ -935,16 +943,6 @@ public abstract class AbstractGuiGame implements IGuiGame, IMayViewCards {
 
     @Override
     public void updateTurn(PlayerView player) { }
-
-    private FControlGameEventHandler localEventHandler;
-
-    @Override
-    public void handleGameEvent(GameEvent event) {
-        if (localEventHandler == null) {
-            localEventHandler = new FControlGameEventHandler(this);
-        }
-        localEventHandler.receiveGameEvent(event);
-    }
 
     @Override
     public void updatePlayerControl() {
--- a/forge-gui/src/main/java/forge/gui/control/FControlGameEventHandler.java
+++ b/forge-gui/src/main/java/forge/gui/control/FControlGameEventHandler.java
@@ -4,7 +4,6 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
 import com.google.common.eventbus.Subscribe;
-import forge.game.card.Card;
 import forge.game.card.CardView;
 import forge.game.event.*;
 import forge.game.player.Player;
@@ -159,39 +158,17 @@ public class FControlGameEventHandler extends IGameEventVisitor.Base<Void> {
         return null;
     }
 
-    private Void processCard(final Card card, final Set<CardView> list) {
-        synchronized (list) {
-            list.add(card.getView());
-        }
-        return processEvent();
-    }
     private Void processCard(final CardView cardView, final Set<CardView> list) {
         synchronized (list) {
             list.add(cardView);
         }
         return processEvent();
     }
-    private Void processCards(final Collection<Card> cards, final Set<CardView> list) {
-        if (cards.isEmpty()) { return null; }
-
-        synchronized (list) {
-            for (final Card c : cards) {
-                list.add(c.getView());
-            }
-        }
-        return processEvent();
-    }
-    private Void processCardViews(final Collection<CardView> cards, final Set<CardView> list) {
+    private Void processCards(final Collection<CardView> cards, final Set<CardView> list) {
         if (cards.isEmpty()) { return null; }
 
         synchronized (list) {
             list.addAll(cards);
-        }
-        return processEvent();
-    }
-    private Void processPlayer(final Player player, final Set<PlayerView> list) {
-        synchronized (list) {
-            list.add(player.getView());
         }
         return processEvent();
     }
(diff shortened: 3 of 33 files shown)


forge-gui/src/main/java/forge/gui/control/FControlGameEventHandler.java
```diff
@@ -386,30 +390,31 @@ public Void visit(final GameEventCombatEnded event) {
...
     }
 
     @Override
     public Void visit(final GameEventCardChangeZone event) {
         if (GuiBase.getInterface().isLibgdxPort()) {
-            updateZone(event.from());
-            return updateZone(event.to());
+            final PlayerView owner = event.card() != null ? event.card().getOwner() : null;
+            updateZone(owner, event.from());
+            return updateZone(owner, event.to());
```
Hanmac: The Old Zone might not the controllers one

For example, if I take control over an opponents' monster, and then it dies,
then my Battlefield and opponents Graveyard need to be updated.

(the Event Card might need to be LKI? Or extra LKI object)

I'm okay with making a small record `ZoneView(PlayerView p, ZoneType type)`

forge-gui/src/main/java/forge/gui/control/FControlGameEventHandler.java
```diff
@@ -186,17 +196,22 @@ private Void updateZone(final Player p, final ZoneType z) {
         }
         return processEvent();
     }
+    private Void updateZone(final PlayerView p, final ZoneType z) {
```
Hanmac, line 184: First `updateZone(Player p, final ZoneType)` function should be updated to call the second one, or be removed

forge-gui-desktop/src/test/java/forge/game/event/GameEventSerializationTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Tests that GameEvent records survive Java serialization round-trips.
+ * This validates the core assumption of the event refactor: events use
+ * view-typed fields (CardView, PlayerView, etc.) that are Serializable,
+ * so they can be sent over the network via the existing Netty pipeline.
+ *
+ * Note: Events using ZoneType (e.g. GameEventCardChangeZone) cannot be
+ * tested here because ZoneType's static initializer requires the Localizer,
+ * which is not available in unit tests without full GUI bootstrapping.
```
Hanmac: I made a PR #9762 to make the ZoneType translation Lazy

forge-gui/src/main/java/forge/gamemodes/net/ProtocolMethod.java
```diff
@@ -40,19 +41,14 @@ public enum ProtocolMethod {
     alertUser           (Mode.SERVER, Void.TYPE),
     updatePhase         (Mode.SERVER, Void.TYPE, Boolean.TYPE),
```
tool4ever: aren't Phase + Turn also handled by their GameEvents?

forge-gui/src/main/java/forge/gamemodes/match/AbstractGuiGame.java
```diff
@@ -922,7 +924,32 @@ public void notifyStackRemoval(GameEventSpellRemovedFromStack event) {
     }
 
     @Override
-    public void handleLandPlayed(Card land) {
+    public void handleLandPlayed(CardView land) {
+    }
+
+    @Override
+    public void handleGameEvent(GameEvent event) {
+        SoundSystem.instance.receiveEvent(event);
```
tool4ever: isn't FControlGameEventHandler still missing?

tool4ever: oh I guess it will be added after the Subgame conversion, too bad I forgot the module dependency issue :(

tool4ever: ok, what if we have a `PlayerController.fireUiEvent(GameEvent ev)` method that SubgameEffect can call instead?

it would call `fireEvent` only for the Match visitor?
and Game no longer needs to subscribe to Match visitor then?

forge-game/src/main/java/forge/game/event/GameEventCardChangeZone.java
```diff
@@ -2,14 +2,17 @@
...
 import forge.util.TextUtil;
 
-public record GameEventCardChangeZone(CardView card, ZoneType from, ZoneType to) implements GameEvent {
+public record GameEventCardChangeZone(CardView card, ZoneView from, ZoneView to) implements GameEvent {
 
     public GameEventCardChangeZone(Card card, Zone zoneFrom, Zone zoneTo) {
-        this(CardView.get(card), zoneFrom == null ? null : zoneFrom.getZoneType(), zoneTo == null ? null : zoneTo.getZoneType());
+        this(CardView.get(card),
+             zoneFrom == null ? null : new ZoneView(PlayerView.get(zoneFrom.getPlayer()), zoneFrom.getZoneType()),
+             zoneTo == null ? null : new ZoneView(PlayerView.get(zoneTo.getPlayer()), zoneTo.getZoneType()));
```
Hanmac: You should add a `getView()` function to Zone class

forge-gui/src/main/java/forge/gamemodes/match/AbstractGuiGame.java
MostCromulent: > isn't FControlGameEventHandler still missing?

See Claude's implementation in https://github.com/Card-Forge/forge/pull/9760/commits/ebb28c94f968ba4c8de0fc3dedc0e23679428a7a.

FWIW this is way beyond my technical understanding, I'm **[basically just this](https://pbs.twimg.com/media/GKT6T6kWQAAP4W4?format=png&name=900x900)** between you and the AI here.

forge-gui/src/main/java/forge/sound/EventVisualizer.java
```diff
@@ -181,110 +181,97 @@ public SoundEffectType visit(final GameEventLandPlayed event) {
...
-                    if(!"C".equals(currManaColor)) {
-                        fullManaColors.append(currManaColor);
-                    }
-                }
-
-            }
-            // No interest if "colors together" or "alternative colors" - only interested in colors themselves
-            fullManaColors = new StringBuilder(TextUtil.fastReplace(fullManaColors.toString()," ", ""));
+            // Determine mana colors this land can produce using view-level info
+            forge.card.ColorSet mana = land.getCurrentState().origProduceMana();
```
Hanmac: The Land Sound fallback from 3color to 2color to 1color doesn't make much sense for me
i will refactor and remove it in a different MR

i will just make it fallback to OtherLand

forge-game/src/main/java/forge/game/spellability/SpellAbilityView.java
```diff
@@ -68,6 +72,27 @@ void updatePromptIfOnlyPossibleAbility(SpellAbility sa) {
...
+
+    public boolean isTrigger() {
+        return get(TrackableProperty.SA_IsTrigger);
+    }
+    void updateIsTrigger(SpellAbility sa) {
+        set(TrackableProperty.SA_IsTrigger, sa.isTrigger());
+    }
+
+    public PlayerView getActivatingPlayer() {
+        return get(TrackableProperty.SA_ActivatingPlayer);
```
tool4ever: there's no reason not to reuse some Property if they mean the same thing right?
e.g. in this case ActivatingPlayer from StackItemView

forge-gui/src/main/java/forge/gui/control/FControlGameEventHandler.java
```diff
@@ -192,11 +216,8 @@ public Void visit(final GameEventTurnPhase ev) {
         needPhaseUpdate = true;
         needSaveState = !"dev".equals(ev.phaseDesc());
 
-        Player ap = ev.playerTurn();
-        boolean refreshField = !ap.getTokensInPlay().isEmpty() || (FModel.getPreferences().getPrefBoolean(FPref.UI_STACK_CREATURES) && !ap.getCreaturesInPlay().isEmpty());
```
tool4ever, line 196: shouldn't it have been possible to preserve the conditional update?

forge-gui/src/main/java/forge/gamemodes/match/AbstractGuiGame.java
tool4ever: haha, feel free to test for any regression in netplay
(e.g. cards not refreshing properly for clients)

code wise this already looks pretty good

forge-gui-desktop/src/test/java/forge/game/event/GameEventSerializationTest.java
```diff
@@ -0,0 +1,227 @@
...
+import forge.game.zone.ZoneType;
+import forge.game.zone.ZoneView;
+import org.testng.annotations.Test;
+
+import java.io.*;
+import java.util.List;
+
+import static org.testng.Assert.*;
+
+/**
```
tool4ever, line 20: I'm not really in favor of keeping such low value tests
this basically just proves the Java contract "if something is Serializable, can it be serialized"?

now if there was some way we could dynamically iterate all GameEvent to detect if someone introduced a non-serializable field that would be quite helpful. But not sure there's a way, maybe with Reflection?

MostCromulent: See https://github.com/Card-Forge/forge/pull/9760/commits/2f829a4661b7b6ca0c2093d821d6f4b7c860fdd0

> How the reflection-based test was validated:                                                                        
>                               
>   1. Positive case (normal run): Test discovers 50+ GameEvent classes from the classpath JAR, checks all record component types, passes with exclusions for GameEventSubgameStart and GameEventSubgameEnd (which intentionally hold Game references).                                                                                                   
>   2. Negative case (exclusions removed): Temporarily cleared the exclusion set — test correctly fails with:           
>   GameEventSubgameEnd.maingame has non-serializable type: forge.game.Game
>   GameEventSubgameStart.subgame has non-serializable type: forge.game.Game
>   3. Discovery sanity check: The test asserts eventClasses.size() > 10 to guard against silent classpath scanning failures (which happened during development — the initial directory-only scanner found 0 classes because forge game is packaged as a JAR).

MostCromulent:
I'll swap this from a draft PR to a real PR so you can decide to deal with it however/whenever you want.

forge-gui-desktop/src/test/java/forge/game/event/GameEventSerializationTest.java
tool4ever: ahh very cool :)

PR #9768, opened 2026-02-16
CardView: rework BasicLand Watermark Properties

Hanmac:
Part of #8682 

i noticed that the check for `addAbilityFactoryAbilities` was messing with getCardForUI, where `getManaAbilities` was empty.
so it failed for `Wastes`, somehow

(we need to refactor getCardForUI in the future anyway)


Hanmac:
@tool4ever @tehdiplomat "addAbilityFactoryAbilities" for getCardForUi causes problems

PR #9781, opened 2026-02-17
EventVisualizer: simlifiy LandSound Effects

Hanmac:
See https://github.com/Card-Forge/forge/pull/9760/changes#r2816312088

the SoundFile names doesn't reflect the order of Shards the in the ColorSet 🤔

### DIFF
--- a/forge-gui/src/main/java/forge/sound/EventVisualizer.java
+++ b/forge-gui/src/main/java/forge/sound/EventVisualizer.java
@@ -198,26 +198,26 @@ public class EventVisualizer extends IGameEventVisitor.Base<SoundEffectType> imp
             case G -> SoundEffectType.GreenLand;
 
             case WU -> SoundEffectType.WhiteBlueLand;
-            case WG -> SoundEffectType.WhiteGreenLand;
-            case WR -> SoundEffectType.WhiteRedLand;
-            case BW -> SoundEffectType.BlackWhiteLand;
+            case GW -> SoundEffectType.WhiteGreenLand;
+            case RW -> SoundEffectType.WhiteRedLand;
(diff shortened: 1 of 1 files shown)


Jetz72:
Does anyone ever use the support for 3-color land sound effects? I'd struggle to even think of 10 distinct sounds which would fit the 2-color lands. Trilands with the same colors don't seem to have any thematic through-line.

Hanmac:
@tehdiplomat your opinion? keep or remove it?

tehdiplomat:
Yea, feels like a waste of our package size to have so many different sounds. 1 for each color, and one for generically multi colored should be plenty.

Hanmac:
> Yea, feels like a waste of our package size to have so many different sounds. 1 for each color, and one for generically multi colored should be plenty.

We don't have these Files
The Sound System checks for these paths just in case

Jetz72:
It's currently only a consideration of what sound types to support for anyone running or making custom sound sets. I don't know if any existing sound sets make use of either the dual land or triome sounds. They aren't officially used, but the extra `SoundEffectType`s aren't really harming anything besides some extra clutter in the codebase. I think we could justifiably remove them, justifiably leave them alone, or justifiably remove only the triome ones since they seem far harder to choose appropriate sounds for and apply to far fewer cards.

I think we also have support for per-card sound effects via `SoundEffectType.ScriptedEffect`. Applies either if the sound file is named after the card or if the card script has a certain SVar set. Seems cool but also apparently unused, and it's kinda hard to think of a good use case for it. Maybe someone could do something with it if we broadened it to look for file names that match subtypes? Something for a separate PR, though.

tool4ever (approved):

PR #9783, opened 2026-02-17
Add logic for activated ability of Emry, Lurker of the Loch

NathanRoy1:
Improves AI evaluation for cast-from-graveyard abilities such as Emry, Lurker of the Loch

This change improves EffectAi and related AI logic so the AI:

- Only activates those abilities when there is a legal, castable target
- Respects turn structure, combat phases, and flash timing
- Avoids tapping abilities at invalid times (opponent’s turn without flash, insufficient mana)


forge-ai/src/main/java/forge/ai/ability/EffectAi.java
```diff
@@ -408,6 +416,91 @@ protected AiAbilityDecision checkApiLogic(final Player ai, final SpellAbility sa
                 }
             }
 
+
+            boolean hasMayPlayFromGrave = false;
+            for (String st : sa.getParam("StaticAbilities").split(",")) {
+                Map<String, String> stParams = FileSection.parseToMap(sa.getSVar(st.trim()), FileSection.DOLLAR_SIGN_KV_SEPARATOR);
```
tool4ever: you can just combine this with the loop above

forge-ai/src/main/java/forge/ai/ability/EffectAi.java
```diff
@@ -408,6 +416,91 @@ protected AiAbilityDecision checkApiLogic(final Player ai, final SpellAbility sa
...
+                if ("True".equalsIgnoreCase(stParams.getOrDefault("MayPlay", "False"))
+                        && "Graveyard".equalsIgnoreCase(stParams.getOrDefault("AffectedZone", ""))) {
+                    hasMayPlayFromGrave = true;
+                    break;
+                }
+            }
+            if (hasMayPlayFromGrave) {
+                String validTgts = sa.getParamOrDefault("ValidTgts", "Card");
+                CardCollection validInGrave = CardLists.getValidCards(
```
tool4ever: use something like `CardUtil.getValidCardsToTarget` instead

forge-ai/src/main/java/forge/ai/ability/EffectAi.java
```diff
@@ -408,6 +416,91 @@ protected AiAbilityDecision checkApiLogic(final Player ai, final SpellAbility sa
...
+                        new CardCollection(ai.getCardsIn(ZoneType.Graveyard)),
+                        validTgts, ai, sa.getHostCard(), sa);
+
+                // Timing context
+                boolean isAiMainPhase = phase.isPlayerTurn(ai)
```
tool4ever: I'm honestly not sure this will lead to better logic than just doing an api check which would happen if the hardcoded "ReplaySpell" logic above was used

maybe just refactor it so your autodetect uses that approach?

NathanRoy1:
Alright i just made the changes requested. I kept the code that manages the flash effect since without it , the AI does not see that there is a possible blocker in his graveyard. For the code that checks if its able to block there is probably a method somewhere that does what I'm doing. If so,  please let me know

PR #9805, opened 2026-02-19
Improve skipping mechanism for cards without collector numbers

tehdiplomat:
Force Sleep when querying multiple urls in a row


Hanmac (approved):

Hanmac:
@tehdiplomat might debug why it is `0` instead of `NO_COLLECTOR_NUMBER` 🤔

tehdiplomat:
@Hanmac it's how someone added it to the edition file.  https://github.com/Card-Forge/forge/blob/master/forge-gui/res/editions/Marvel%20Super%20Heroes%20Commander.txt#L16

Issue #9705, opened 2026-02-09
Quality of life features to the cards originally from MTG Arena


Svyatoslav28597:
Hi! I've noticed that some things regarding digital only cards from arena are absent in forge.
The main things are:
1) Mtg Arena displays flavor and reminder text in a rather unique way, despite it being absent on the cards it is presented in a little pop-up window. Due to absense of this feature all Digital only cards and rebalances lack Flavor Texts, and they can't be viewed. Just to clarify, none of those flavor texts are on scryfall and the easiest way to view them is to use mtg wiki. 

<img width="1402" height="1079" alt="Image" src="https://github.com/user-attachments/assets/960b359b-e397-45d9-848a-c2823491d20e" />

2) Certain digital mechanics like Conjure/Specialize/Spellbook create add additional cards to the game, however you  can't view this cards before the effects tooks off. Arena adresses this by having an addittional window with all the cards. This also can help with certain cards like Garth, One-Eyed who use similar functions. Currently the only way to see what those cards do is to google them.

<img width="1680" height="735" alt="Image" src="https://github.com/user-attachments/assets/e9338dfb-17e0-4ce6-ab99-82e9730fad6a" />

It would be nice to have visual interface similar to the arena, It would make the usage of such cards way more fluid and intuitive. While it may took quite some time for visual rework, in the meantime flavor texts can be added to the usual descriptions of cards in forge, unless this indirectly breaks some other things.

Alternatively you can use fandmade "rendered as printed" cards from acommunnity which turn traditional digital prints into more "paper like" including having written flavor. However this is a unofficial project and it may not be the best way to fix this issue.

MostCromulent:
I'll take a look at this.

MostCromulent:
So... making an Arena-style hover tooltip that shows a zoomed image of the selected card / related cards that it creates or transforms into / keyword definitions is relatively straightforward. 

Including the flavour text is a dealbreaker though because this isn't automatically stored in Forge's card data, so it can't be easily extracted to the tooltip. You would basically have to go and retrospectively add a flavour text field to every card in Forge to make it work, which is definitely not worth the payoff.

I'll work on an implementation of the hover tooltip but I'm leaving the flavour text out.

Svyatoslav28597:
Alternative workaround for flavor texts is adding it only to the cards which appear exclusively on the arena, and that would be...  863 cards if "game:arena -game:paper -is:reprint" on scryfall includes all of them

[...]

PR #9806, opened 2026-02-19
Add card info tooltips with keyword explanations and related cards

MostCromulent:
## Summary

Adds Arena-style card info popups to desktop UI including selected card, related cards (tokens created, spellbooks etc), and keyword explanations. Includes menu options to enable/disable each section and scale card image size. Addresses https://github.com/Card-Forge/forge/issues/9705. 

(I can imagine some people choosing to use this instead of default card view/card details panels).

Screenshot 2026-02-24 203735

Screenshot 2026-02-19 145001
Screenshot 2026-02-19 144825
Screenshot 2026-02-19 144922

## Features
                                                                                                                      
- Adds a hoverable card info popup that appears when mousing over cards during a match, showing the card image, keyword ability and keyword action explanations with reminder text, and related card images (tokens, spellbook, meld pairs) grouped by category
- Related cards supports 'creates, 'transforms into', 'other face', 'meld', 'flips into', 'specialises into', 'partner', and 'spellbook'.
- Keyword actions are defined in a new KeywordAction enum (forge-game) covering all 67 entries from MTG comprehensive rules section 701, making them reusable across modules in future
- Popup is constrained to the Forge window bounds, enforces single-instance display, and includes user preferences for image size and enable/disable toggle under Game menu

## New files

- forge-game/.../keyword/KeywordAction.java — enum of all MTG keyword actions with reminder text
- forge-gui-desktop/.../view/arcane/CardInfoPopup.java — popup window with pill-panel styling
- forge-gui-desktop/.../match/menus/CardInfoPopupMenu.java — Game menu submenu for popup settings

## Note

Original feature request in https://github.com/Card-Forge/forge/issues/9705 also wanted to include flavour text for alchemy cards. This is not currently included in Forge card data so cannot be easily extracted. I have not included in this PR as I don't think significant extra diff to achieve this is worth the benefit.

---


dracontes:
Some notes testing this out for a bit:
- There's a clear overlap with being able to roll the mouse wheel up on hovering over a card to show an enlarged view of the card image. In fact these two ways to getting a detailed view of the card image do conflict a fair bit. 
I do agree the additional information in question is useful to have closer at hand.
Though, personally, much because I think it's best if we don't hew _too_ close to Arena, my preference would be for the additional information to be added to the wheel roll overlay instead. As it is something people should already use, it dovetails nicely.
Or, at least, the wheel roll overlay should have this additional information as well and users should be able to choose between or against the two options.

hoveroptions-1

- In any case there's some issues with margins and not particularly informative text for some abilities, and for some reason the generated images for associated cards (namely spellbooks) are just the boilerplate blank frame one.

hoveroptions-2

hoveroptions-3

hoveroptions-4

MostCromulent:
Understand what you say about avoiding too much similarity. 

I'm happy to re-work this to integrate with the scroll-up card zoom view if that's the consensus, though that perhaps makes the functionality less obvious to the kind of less experienced users who the tooltips and related cards are most useful for?

(Also personally I basically never use the scroll-up function, hence why I missed the conflict when zoomed in. I should at least be able to sort that out.)

Will look into the other issues. The image downloader for the related cards box was at least working for me earlier today.

dracontes:
> I'm happy to re-work this to integrate with the scroll-up card zoom view if that's the consensus, though that perhaps makes the functionality less obvious to the kind of less experienced users who the tooltips and related cards are most useful for?

Well, the existing functionality is similar, in broad strokes, to how things work on the Android version: there ~swiping~ tapping brings up (and dismisses) the card detail, which can then be swiped between two display formats.

> Will look into the other issues. The image downloader for the related cards box was at least working for me earlier today.

Just to be clear: I don't download images for Forge. I find the built-in card renderer/image generator sufficient for my usage. The problem is that those related cards aren't rendering as expected.

A few more notes:
- The tooltip popup flickers something fierce if you hover the cursor on the name bar of a tapped card. It happens even after moving the cursor well into the card's footprint, where you'd expect the cursor to "lock in" as it were.
- A few more issues with displayed text:

hoveroptions-6

hoveroptions-5

- Also seems like the issue with the margins of the bounding box for the text is at least intermittent. Seems like it fixed itself on my side. ETA: Or not, see below,

hoveroptions-7

- Also also, seems like the popup isn't all that consistent picking up on granted abilities: Mary Jane in the screenshot above has flying and intimidate.

Hanmac:
in theory, SpellAbilityEffect represents KeywordActions
maybe they can be used too 🤔 

for KeywordInterfaces, checkout `KeywordInterface#getTitle` too (see TypeCycling)

MostCromulent:
@dracontes, I'll test this more extensively tomorrow but latest commits should hopefully address many of the issues you had above including card rendering and initial implementation of seperate preference toggles to enable/disable showing related cards/keyword tips in the card zoom view. 

Still working on getting it to more accurately pick up granted abilities/solve a few layout problems.

Screenshot 2026-02-19 201757
Screenshot 2026-02-19 230838
Screenshot 2026-02-19 224830
Screenshot 2026-02-19 223720

forge-gui-desktop/src/main/java/forge/view/arcane/CardInfoPopup.java
```diff
@@ -0,0 +1,1100 @@
...
+    }
+
+    /**
+     * Scan oracle text for keyword actions that aren't already shown as keyword
+     * abilities, and append them to the keyword list.
+     */
+    public static void addKeywordActions(final List<KeywordData> result,
+                                           final String oracleText,
+                                           final Set<String> existingNames,
+                                           final String cardName) {
```
Jetz72: I think the logic for determining what keyword actions a card utilizes should be in either the main GUI module or the game module. It'd be good to have on hand in case we wanted to add similar functionality to the mobile UI (e.g. along the side in the zoomed in view).

forge-game/src/main/java/forge/game/keyword/KeywordAction.java
```diff
@@ -0,0 +1,134 @@
...
+ * sacrifice, etc.) that every player knows — UI code may choose to omit these from
+ * tooltips to avoid clutter.</p>
+ */
+public enum KeywordAction {
+    // 701.2 – 701.13: Basic game actions
+    ACTIVATE("Activate", "Put an activated ability on the stack.", true),
+    ATTACH("Attach", "Move an Aura, Equipment, or Fortification onto another object or player.", true),
+    BEHOLD("Behold", "Reveal a card with the required quality from your hand, or choose a permanent you control with that quality.", false),
+    CAST("Cast", "Put a spell on the stack.", true),
+    COUNTER("Counter", "Remove a spell or ability from the stack. It doesn't resolve and none of its effects happen.", true),
```
Jetz72: Would it be worth the trouble at this stage to use translation keys instead of hard-coded strings? I know keyword abilities are kinda in the same boat, though.

Jetz72 (commented):
Ideally we could identify KeywordActions by inspecting the card's abilities and their parameters rather than the Oracle text. But I'm not sure how well that works with everything. Something like Scry might be a straightforward spot check for a `ScryEffect` or a `ReplaceScry`, but I think there are some that are implemented purely through card script.

forge-game/src/main/java/forge/game/keyword/KeywordAction.java
MostCromulent: No trouble, it's just some grunt work for the AI doing the coding. See https://github.com/Card-Forge/forge/pull/9806/commits/5db88c655742c4ab3e041d1d8571b5defb858c9 / https://github.com/Card-Forge/forge/pull/9806/commits/59b01150266724640b34b46514947a490bde48dc

MostCromulent: Happy to do the same for the Keyword abilities as a seperate PR if you want?

MostCromulent:
> Ideally we could identify KeywordActions by inspecting the card's abilities and their parameters rather than the Oracle text. But I'm not sure how well that works with everything. Something like Scry might be a straightforward spot check for a ScryEffect or a ReplaceScry, but I think there are some that are implemented purely through card script.

@Jetz72, I've asked the AI to assess the feasability of this for each of the KeywordActions. Here's the response:

> Direct ApiType match (34):
>   Scry, Surveil, Mill, Fight, Goad, Investigate, Explore, Connive, Discover, Cloak, Manifest Dread, Amass, Learn,
>   Incubate, Ring Tempts You, Venture, Vote, Clash, Detain, Regenerate, Manifest, Meld, Set In Motion, Abandon,        
>   Proliferate, Planeswalk, Open an Attraction, Assemble, Villainous Choice, Time Travel, Endure, Airbend, Earthbend,
>   Blight                                                                                                              
>                              
>   K: keyword definition — no SpellAbility to inspect (3):
>   - Monstrosity → K:Monstrosity:4:5 B G (keyword definition, not an effect)
>   - Adapt → K:Adapt:1:2 G U (same pattern)
>   - Exert → K:Exert (keyword, triggers are separate)
> 
>   Ambiguous ApiType — overloaded or flag-dependent (4):
>   - Transform → SetState (also used for flip cards, MDFCs)
>   - Convert → SetState (same as Transform)
>   - Populate → CopyPermanent with Populate$ True flag
>   - Bolster → PutCounter with Bolster$ True flag
> 
>   No ApiType exists (8):
>   Behold, Fateseal, Roll to Visit, Collect Evidence, Suspect, Forage, Harness, Waterbend
> 
>   Indistinguishable from generic effects (1):
>   - Support → PutCounter with specific targeting (no flag — identical to any "put +1/+1 counters on target creatures"
>   ability)
> 
>   Totals: 34 detectable via effects, 16 still need Oracle text fallback. Plus CardView doesn't expose ability data
>   today — all 50 would require plumbing game-layer data into the view layer.
> 
>That means there would still need to be a hybrid approach with Oracle text fallback rather than a clean replacement. There's also an architectural question since the detection currently runs against CardStateView (which only exposes strings) rather than the game Card object — threading ability data to the view layer would be a larger change.
> 
> Given the text approach seems to work, I'd lean toward keeping it as-is for this PR and potentially revisiting if we build richer ability metadata into CardView in the future. But happy to refactor if you feel the hybrid approach is worth it now.

Jetz72:
Well, 34 out of 50 (not counting the 17 trivial ones) isn't too bad of a start. Digging into it a bit more -

* As it points out, Forge treats Monstrosity and Adapt as keyword abilities (which it probably shouldn't but not a huge issue). If we were to refactor these we'd probably just skip them in the keyword action search and let the ability keyword write the description.
* Populate and Bolster would have to be done by inspecting parameters on their respective methods. Roll to Visit Your Attractions is done as an action via an optional parameter on `RollDice`. 
* Suspect and Harness would have to be detected via the `Attributes$` parameter on `AlterAttribute`
* Fateseal and Support seem indistinguishable. If we went this route, we'd probably resolve this by adding marker parameters to the cards that use these keyword actions. Currently a total of 21, across both keywords.
* Transform and Convert would be indistinguishable purely from the ability scripts. We could introduce marker parameters to separate out the 15 convert cards, but I'm not sure it'd even be worth it. As far as I can tell, these actions are exactly the same thing, rules-wise.  I don't think non-transforming MDFCs use `SetState`. 
* Feels a bit strange that Flip isn't considered a keyword action. It'd be set apart by the `Mode$` parameter.
* Exert's API is `CostExert`. It evidently wasn't looking at cost APIs, which is why it also missed Behold, Collect Evidence, Forage, and Waterbend. Scry and Mill can also occur as costs.
* On a similar note, it'd also need to inspect Trigger and Replacement APIs. Plenty of these actions like Scry have effects which either trigger when you do them or modify them via REs. 

I think we could get away without a hybrid approach, but it'd still be a pretty huge overhaul to what you currently have. Each keyword would need one or more cues to check against an ability, trigger, or replacement effect, for their API, parameters, sub-abilities, and any associated costs. Instead of threading ability data to the view layer we'd probably just compute the keyword action list on the game layer. Possibly within the abilities/triggers/REs themselves before compiling them in the `Card[State]View` class to hand off to the UI...

On second thought, if we ever wanted tooltips to work outside of a game we'd have to do it even earlier, with less tools. Possibly while as part of the `CardFace`/`CardRules`? And then we'd need to either link those keyword abilities to their ability string or generate them a second time while in-game so that we can handle cases when they're moved between cards....

...I think it's probably fine for now. I don't think the current implementation will usually be an issue as long as the card name isn't included in the text matching.

forge-gui/res/languages/en-US.properties
```diff
@@ -3353,4 +3364,140 @@ lblRepair=Repair
...
+lblKwActionSupport=Support
+lblKwActionSupportReminder=Put a +1/+1 counter on each of up to N target creatures.
+lblKwActionMeld=Meld
+lblKwActionMeldReminder=Exile two specific cards and combine them into one oversized card on the battlefield.
+lblKwActionExert=Exert
+lblKwActionExertReminder=An exerted creature won''t untap during your next untap step.
+lblKwActionExplore=Explore
+lblKwActionExploreReminder=Reveal the top card of your library. Put it into your hand if it''s a land. Otherwise, put a +1/+1 counter on this creature, then you may put the card back or into your graveyard.
+lblKwActionAssemble=Assemble
+lblKwActionAssembleReminder=Put a Contraption you own from outside the game onto the battlefield on one of your sprockets.
```
Jetz72: Assemble pulls the top card from the Contraption deck, not one of your choice from outside the game. 

On a related note, cranking a Contraption is also sorta qualifies as a keyword ability among un-cards, but I'm not sure if there's any useful description for it. It essentially does nothing except signal "Whenever you crank this contraption" triggers to fire.

forge-gui/res/languages/en-US.properties
```diff
@@ -3353,4 +3364,140 @@ lblRepair=Repair
...
+lblKwActionFateseal=Fateseal
+lblKwActionFatesealReminder=Look at the top N cards of an opponent''s library, then put any number on the bottom and the rest on top in any order.
+lblKwActionClash=Clash
+lblKwActionClashReminder=Each clashing player reveals the top card of their library, then puts it on the top or bottom. You win if your card''s mana value is higher.
+lblKwActionPlaneswalk=Planeswalk
+lblKwActionPlaneswalkReminder=Move to a new plane by turning over the next card in the planar deck.
+lblKwActionSetInMotion=Set in motion
+lblKwActionSetInMotionReminder=Turn a scheme face up and follow its instructions.
+lblKwActionAbandon=Abandon
+lblKwActionAbandonReminder=Turn a face-up ongoing scheme face down and put it on the bottom of its owner''s scheme deck.
```
Jetz72: I don't think the text "set in motion" will match any Scheme except "All in Good Time". The description also doesn't make mention of taking it off the top of the Scheme deck.

However, it may be easiest to just flag Set In Motion and Abandon as basic game actions. Their meaning is unclear in a vacuum, but within the Archenemy format they're fairly intuitive.

dracontes:
Good show on what you've addressed so far.

Gave the more recent build a spin. A few more notes for your consideration.

- I'm able to dismiss the zoomed view of a card by rolling down on the mouse wheel or clicking when hovering anywhere on the screen but a region including the information detail boxes and an area under those that is as tall as the card image and as wide as the detail boxes. I think you should be able to do so anywhere on the screen. 

hoveroptions-8

- The right card faces aren't showing up for specialize cards.

hoveroptions-14

- The flickering of the tooltip popup  on a tapped card is still an issue. After a bit more testing, I hazard it's because the tooltip popup is linked to the card's footprint while it's untapped. Then when the card is tapped the popup covers the top region of the card, getting between the cursor and the card which leads to the flickering when you're hovering the cursor over that region of a tapped card.

- For the tooltip popup, I think related cards/tokens could be shown a bit bigger in relation to the card image. Large enough to be readable at least.

hoveroptions-22

- The game engine is a bit confused about cards that have the string 'planeswalk' in their text

hoveroptions-24

- Some abilities/concepts for which I feel the information detail needs some attention.
‣ The explanatory text for abilities that involve values could be handled more uniformly. I feel that if a digit is involved then it should also show up on the box's heading. So 'earthbend 2', 'scry 2'. If the number is spelled out the explanatory text can more boilerplate like for milling.

image

hoveroptions-10

‣ A few instances explanatory text needing adjustment:

image

‣ I feel like devotion could use explanatory text 

hoveroptions-21

‣More of a navigational warning that [partner](https://mtg.wiki/page/Partner) is a bit messy right now on Forge as concerns display during a match. 

hoveroptions-23

MostCromulent:
@dracontes thanks for the testing feedback, very helpful. Most of those issues should now be fixed.

I'm still dealing with a problem where it won't generate related cards for specialize cards (i.e. your Lae'zel, Githyanki Warrior example where it just shows the base card 5 times). It seems to work properly for MDFCs and spellbook cards but not specialisations; something to do with how the relationships are represented in the card data.

(**Edit**: this seems to be fixed.)

> For the tooltip popup, I think related cards/tokens could be shown a bit bigger in relation to the card image. Large enough to be readable at least. 

This should now render up to 2 related cards side by side in the hover toolip at whatever size you set as your preference in the menu. If there are more than 2 related cards (spellbooks, many tokens etc) it will include a line below the 2 prompting you to zoom to see the rest. The zoom menu will include a scrollbar letting you view all of them at your selected size.

> I feel like devotion could use explanatory text

This has been added. A nice QOL todo for devotion/affinity etc would be for the tooltip to calculate your current value and include it at the end e.g. "(Your devotion to black is currently [X].)" On my list for once core functionality is working. 

(**Edit**: this is now in.)

Hanmac:
> As it points out, Forge treats Monstrosity and Adapt as keyword abilities (which it probably shouldn't but not a huge issue). If we were to refactor these we'd probably just skip them in the keyword action search and let the ability keyword write the description.

i will do some thinking and refactoring about Monstrosity and Adapt (i try to avoid turning it into a own SpellAbilityEffects if able)
maybe i can reuse the logic i did for Earthbend to predefine the other Params like CounterType

Adapt #9854
Monstrosity #9859
Support #9870
Bolster #9872

MostCromulent:
I've added an option to show overlays on the hover tooltip of the selected card. This just follows the user's existing overlay preferences. 

It occurs to me there may be situations where you want different overlays between the two (i.e. you want to see the PT overlay in the tooltip but you don't need the name because the tooltip is already readable) - but keeping two different sets of overlay preferences + menu options to support that also seems like overkill? Would welcome view on whether the juice is worth the squeeze.

This duplicates some of the existing overlay logic in the new `KeywordInfoUtil` - there's a potential consolidation opportunity here (~300 line reduction across two files) to extract the ability icon rendering into a shared utility for both the tooltips and the normal card overlays, which I'm happy to do, but does involve messing with existing overlay code otherwise untouched by this PR.

Hanmac:
I will do some cleanup with Landwalk and Affinity later that day

MostCromulent:
> It occurs to me there may be situations where you want different overlays between the two (i.e. you want to see the PT overlay in the tooltip but you don't need the name because the tooltip is already readable)

Having now tested this more thoroughly, I definitely think users will want granular control of the overlays between cards on field vs hover tooltips. Have consolidated this into a shared dialog menu:

Screenshot 2026-02-24 201119

MostCromulent:
@Jetz72 if you're still comfortable with the KeywordActions detection working based the oracle text (at least for time being), I've tested this and it seems to be functioning properly and ready for review. (With as many cards as MTG has I expect there will be edge cases where certain syntax still trips it up, but I haven't encountered any glaring errors in latest testing).

One outstanding design question is what should the default preference for this be? I think the keyword tooltips in particular are very useful , but they don't make a heap of sense without the card popup next to them - so I lean to enabling both by default but this isn't a super strong view. If it's all off by default I think there's a risk it just gets missed in the menu settings.

dracontes:
Good show so far 👍 

A few more comments after a bit more testing.

- Getting the same hover tooltip flickering happening on the game log for a similar reason as on the battlefield: the tooltip is displayed by hovering anywhere on a particular log message (area highlighted in red on the screenshot), but the tooltip's card image is linked to the footprint of the card image on the log message. Maybe here the card image should render outside of the log's footprint entirely just for clarity, like it does for the stack.

hoveroptions-a-1

- I appreciate that you did enough to disentangle the zoom-in and tooltip. I do think that having the zoom-in deal with the full display of large spellbooks exclusively is a smart solution. Still, I do feel the checkboxes should be available for users to disable all of it at their convenience, if not for any other reason. Even agreeing that these features should be set by default to the configuration you find most useful, so people are at least aware of them, I have had to field questions on Discord regarding disabling the overlays for instance.

- Cards like [Garth One-Eye](https://scryfall.com/card/mh2/197/garth-one-eye) should display a similar information box as cards with spellbooks

hoveroptions-a-11

- I feel like "the Ring tempts you" should show the respective emblem though seeing as there's no script for it might be a tall order to have it display for me (unless the card image already does).

hoveroptions-a-10

- I think if there's "Mill 2" there should be "Mill 1."

hoveroptions-a-5

- Issues with information text display for sundry abilities:
‣ Craft is a bit messy
hoveroptions-a-6

‣ I would expect "Amass [creature type] [N]"
hoveroptions-a-7

‣ A bit confused/confusing with equip abilities of a different character
hoveroptions-a-8

MostCromulent:
@dracontes thanks for feedback, this is very helpful and I will investigate.

MostCromulent:
> I appreciate that you did enough to disentangle the zoom-in and tooltip. I do think that having the zoom-in deal with the full display of large spellbooks exclusively is a smart solution. Still, I do feel the checkboxes should be available for users to disable all of it at their convenience 

I think we have that here - unless there was different enable/disable functionality you wanted?

Untitled

MostCromulent:
@dracontes all of the issues above should now be fixed.

**Game log overlap** - now renders outside of the log entirely.
Screenshot 2026-02-25 214044

**Garth One-Eye / Cards naming other cards**
Screenshot 2026-02-25 214302
Screenshot 2026-02-25 214318

**The Ring Tempts You** - took some slightly ugly Ring-related hard coding within the tooltip file, but have got this working. Have also got it to track your current ring level in the header and tooltip. 
Screenshot 2026-02-25 213805

**Mill**
Screenshot 2026-02-25 214730

**Craft**
Screenshot 2026-02-25 214640

**Amass**
Screenshot 2026-02-25 215838


**Equip**
Screenshot 2026-02-25 214558

Hanmac:
Devotion is a bit tricky

because the gods have "Devotion for X and Y",
where Hybrid Mana only counts once, not twice

would be best if you check for SVar X = contains Devotion like this:
https://github.com/Card-Forge/forge/blob/682bb198b79ec7824b8f9161e20fff7a60b018d8/forge-gui/res/cardsfolder/i/iroas_god_of_victory.txt#L7

dracontes:
> I think we have that here - unless there was different enable/disable functionality you wanted?

I apologize: I somehow must have convinced myself during the testing I did for my 2026-02-21 review of that version of the PR there was a checkbox implemented to disable the zoom-in view altogether, either from the Game dropdown menu or on Preferences. Just fired up that version of this PR and I can't find it.

Still, I would like that to be available for the reasons I stated. if it isn't too much trouble to implement.

MostCromulent:
> Devotion is a bit tricky [...] would be best if you check for SVar X = contains Devotion like this:

@Hanmac have re-worked this on latest commits, seems to be working well.

>  checkbox implemented to disable the zoom-in view altogether, either from the Game dropdown menu or on Preferences.

@dracontes have re-worked master enable/disable hotkeys and preferences similar to CTRL-O for overlays.

Screenshot 2026-02-26 070026

Screenshot 2026-02-26 070148

dracontes:
> have re-worked master enable/disable hotkeys and preferences similar to CTRL-O for overlays.

I'm sorry, I should have been clearer: what I am asking for is a checkbox to disable the image itself showing up when you scroll up when hovering a card. Could certainly be that checkbox you set up is what's reworked to do that.

Ran down the discussion, verifying relevant items. Here are some salient enough issues. The rest were addressed well enough.

- Like for other, similar instances, if there's "monstrosity 5" there should be "monstrosity X".

image

- Likewise for collect evidence.

image

- I'm pretty sure the underlined text isn't true: [the keyword actions transform and convert are interchangeable](https://mtg.wiki/page/Convert).

image

- Some confusion with trample and trample over planeswalkers

hoveroptions-b-5

- Seems like the layout here could be a bit more compact

hoveroptions-b-6

- ETA: The text for [waterbend](https://mtg.wiki/page/Waterbend) is a bit off, making it sound more like convoke than it actually is.

image

MostCromulent:
> what I am asking for is a checkbox to disable the image itself showing up when you scroll up when hovering a card. 

I think the CTRL-I shortcut should do what you're looking for: completely disable the tooltip functionality in zoom view. Disabling the "Enable Card Information Tooltips in Zoom View" preference will do the same thing.

> Here are some salient enough issues. The rest were addressed well enough.

These should now be fixed.

tool4ever (changes requested):
huge PR, so definitely not before next release

forge-core/src/main/java/forge/util/ImageUtil.java
```diff
@@ -260,6 +260,13 @@ public static String getScryfallDownloadUrl(PaperCard cp, String face, String se
             faceParam = (face.equals("back") && cp.getRules().getSplitType() != CardSplitType.Flip
                     ? "&face=back"
                     : "&face=front");
+        } else if (cp.getRules().getSplitType() == CardSplitType.Specialize) {
```
tool4ever, line 263: is this enough to make downloading work during gameplay?
if yes, please prepare for shrinking by forwarding the part as PR to master

dracontes:
> I think the CTRL-I shortcut should do what you're looking for: completely disable the tooltip functionality in zoom view.

It doesn't disable the zoom view which is what I suspect is a good option to have. I'll note it down as a future feature request in any case.

A few more noteworthy items found while testing this version of the PR:
- Support showing up in weird places, these two cards:

hoveroptions-c-8

- Another situation that could use a more compact layout, 

hoveroptions-c-2

- Some confusion with manifest and manifest dread.

hoveroptions-c-6

- Escape's header needs an adjustment.

hoveroptions-c-7

- As well as craft. I feel the the ability should be written out in full in the header with the requited (sub)types. I do notice the explanatory text needs a touch up as well.

hoveroptions-c-9

- Probably something for the future, likely already planned. Certainly note it down that I'd like to see the ability explanations extended to the permanents that grant those abilities to other objects, namely Equipment and Auras. 

hoveroptions-c-5

tool4ever:
I should probably clarify my labeling isn't meant to prevent this PR from progressing further...

*However* these past experiences have shown an in-depth code review of this size isn't very enjoyable as unfortunately the time required doesn't grow linearly due to having to keep more context in mind, that is:
`2000 LoC PR >> 1000 PR + 1000 PR`

And I think as another "natural" factor AI quality also goes down a bit after a certain size, so 1000+ LoC is also more likely to contain errors or bloat rather than smaller parts combined - especially if it evolved over multiple iterations rather than a mostly pure one-shot from design spec. :/

Admittedly the current PR scope might not be the most favorable slicing target, but imo it should still at least be evaluated to increase likelihood of a successful merge.

MostCromulent:
@tool4ever all good, have been focused on other things so this has been on the backburner anyway - it's more of a "nice to have" than an essential. totally agree with what you say about review challenges.

some rough thoughts on how this could be broken up into more manageable bundles for review.

1. vertical slice: hover images + simple tooltips for existing keywords only; just displaying the keyword text
2. expand to include plain text of ability keywords
3. expand to add "related card" tooltips and hover images
4. expand keyword tooltips to include more dynamic content: affinity count, card types in graveyard, etc.

(as an aside, this was one of my earliest projects and the quality of the AI guidelines I'm using now + my own experience managing it is _a lot_ better than when this was done, so would probably benefit from a refactor anyway)

tool4ever:
hmn maybe MVP approach around 3:
I'd guess for more users the actually missing completely from GUI part of "seeing Spellbook-like contents without alt-tabbing to Scryfall" would be the most helpful?
(though personally I dislike Alchemy and it's not that many cards, so my review priority would also still be rather low tbh)

Svyatoslav28597:
With the release of ALchemy Strixhaven it seems that scryfall started adding  flavor text to the alchemy cards. I am wondering if this is enought to allow this menu display flavor texts from the Alchemy?

Jetz72:
> With the release of ALchemy Strixhaven it seems that scryfall started adding flavor text to the alchemy cards. I am wondering if this is enought to allow this menu display flavor texts from the Alchemy?

Forge doesn't currently track or support tracking flavor text for any cards. It'd be an interesting thing to support, and it's not the only useful thing we could get if we started fetching print metadata from Scryfall, but it doesn't currently fit neatly into any of our existing data pipelines. I suspect if we did include it, we'd probably just throw it in the normal card text areas below a separator.

Hanmac:
@MostCromulent does the new KeywordView Record helps you with some of the stuff?

Title and ReminderText might be enough?
Cost and Amount stuff is part of the Title and you probably don't need them extra?

MostCromulent:
@Hanmac yep, should be helpful in avoiding some of the more complex oracle parsing this was doing before. we're looking to do a staged set of refactored PRs on this per https://github.com/Card-Forge/forge/pull/9806#issuecomment-4412030490 so wont get to the more complex stuff for a while, but the records should either be complete or very solid foundation for what we need.

jamincollins:
## Findings

#### 1. KeywordInfoUtil annotateAffinity typeParam null fallback uses display name
Description: annotateAffinity in KeywordInfoUtil.java falls back to extracting matchType from display name when kw.typeParam is null. The extraction logic uses lowerName.replaceAll... which may strip mana symbols but also may remove critical characters used for matching. This could cause incorrect affinity counting for keywords with complex type parameters.
Test case:
```java
// Simulate KeywordInfoUtil.annotateAffinity with typeParam null
KeywordInfoUtil.KeywordData kw = new KeywordInfoUtil.KeywordData("Affinity for artifacts", "reminder", null);
PlayerView controller = ...; // mock
String lowerName = kw.name.toLowerCase().replaceAll("\\{[^}]+}", "").trim();
// lowerName becomes "affinity for artifacts"
// fallback matchType extraction from display name may fail for "Affinity for artifact tokens"
```
Impact: Affinity counts may be zero for cards with non-standard type params, breaking tooltip accuracy.

#### 2. Protection.parse strips leading 'from ' and colon but does not validate input
Description: Protection.parse sets fromWhat = details.startsWith("from ") ? details.substring(5) : details; then strips colon-delimited part. No validation for empty details, resulting in empty fromWhat and title "Protection from ". This could produce malformed UI strings.
Test case:
```java
Protection prot = new Protection();
prot.parse(""); // fromWhat = "", title returns "Protection from "
prot.parse("from "); // fromWhat = "", title malformed
```
Impact: UI shows empty protection keyword, confusing users.

#### 3. KeywordWithType.getTitle LANDWALK concatenation may produce incorrect capitalization
Description: LANDWALK case builds title as Character.toUpperCase(descType.charAt(0)) + descType.substring(1) + "walk". If descType is empty, charAt(0) throws StringIndexOutOfBoundsException. If descType is null, NPE. No guard for empty/null.
Test case:
```java
KeywordWithType kw = new KeywordWithType();
kw.parse(""); // descType = ""
String title = kw.getTitle(); // throws StringIndexOutOfBoundsException
```
Impact: Crash when parsing malformed keyword data.

#### 4. KeywordInfoUtil.sortByOracleText position lookup returns -1 for missing keywords
Description: findKeywordPosition returns -1 if keyword not found in oracle text. Sorting with -1 values may place keywords unpredictably. No handling for both positions -1.
Test case:
```java
List<KeywordData> keywords = Arrays.asList(kw1, kw2);
sortByOracleText(keywords, "Some text without keywords");
// posA=-1, posB=-1 => comparator returns 0, stable sort preserved but order undefined
```
Impact: Keyword order may not match user expectation for cards with keywords not mentioned in oracle text.

#### 5. CardInfoPopup image loading may cause deadlock on EDT
Description: CardInfoPopup likely uses SwingImageFetcher on EDT for image loading with callbacks. If callback triggers UI updates synchronously, deadlock possible. No explicit off-EDT loading verification.
Test case:
```java
// Simulate popup showing during UI freeze
SwingUtilities.invokeLater(() -> showPopup());
Thread.sleep(100);
// Image load completes on EDT, blocking repaint
```
Impact: UI freeze during popup display.

#### 6. Protection change breaks existing getProtectionValid logic
Description: Protection.java removed static getProtectionValid method used by game logic. Old code paths expecting that method will fail with NoSuchMethodError. No deprecation shim.
Test case:
```java
// Existing code:
String valid = Protection.getProtectionValid("Protection:Red", true);
// Throws NoSuchMethodError
```
Impact: Game logic for protection validation may break, affecting combat resolution.

#### 7. KeywordInfoUtil.buildKeywords deduplicates by name but not by keyword enum
Description: buildKeywords uses addedNames set to deduplicate, but keywords with same display name but different enum instances may be incorrectly merged. No check for Keyword enum identity.
Test case:
```java
// Two different Keyword entries with same name string
String keywordKey = "Affinity for artifacts,Affinity for artifacts";
List<KeywordData> result = KeywordInfoUtil.buildKeywords(keywordKey, new HashSet<>());
// May still process duplicate
```
Impact: Duplicate keyword entries in tooltip.

#### 8. CardOverlaySettingsDialog preferences save on every toggle without debouncing
Description: CardOverlaySettingsDialog registers action listener that saves prefs on every checkbox toggle. Rapid toggling causes multiple disk writes.
Test case:
```java
FCheckBox cb = new FCheckBox();
cb.addActionListener(e -> { prefs.setPref(...); prefs.save(); });
// User clicks rapidly 10 times -> 10 disk writes
```
Impact: Performance degradation and potential file I/O contention.

#### 9. ImageCache getCardOriginalImageInfo change may break callers expecting void
Description: ImageCache.getCardOriginalImageInfo changed signature? Actually method now returns Pair but callers may expect old behavior. No compile error if changed to return type, but callers using return value may get unexpected pair.
Test case:
```java
// Old caller:
ImageCache.getCardOriginalImageInfo(key, true, card);
// Now returns Pair, ignored, but may cause unchecked conversion warnings
```
Impact: Subtle bugs if return value used incorrectly.

### Performance and security
#### 10. KeywordInfoUtil.annotateKeywordCounts iterates over all keywords per card view
Description: annotateKeywordCounts loops over keywords and for each calls controller queries (hasStringType, getPermanentCount, etc.). No caching of controller queries, leading to O(N*M) queries per popup.
Test case:
```java
// Card with 5 keywords, each annotate triggers 5 controller queries
// Popup shown 100 times => 2500 queries
```
Impact: UI lag during tooltip display.

#### 11. CardInfoPopup caches images without size limit
Description: CardInfoPopup likely caches CardImage instances in a map without eviction. Unbounded cache growth on large games.
Test case:
```java
// Open popup for 1000 different cards
Map<Card, Image> cache = new HashMap<>();
// Memory grows unbounded
```
Impact: OutOfMemoryError in long sessions.

## Verdict
Request Changes
The PR introduces valuable card info tooltips but contains several logic correctness and safety issues: Protection refactor removes critical game logic method without shim, KeywordWithType.getTitle lacks null/empty guards, annotateAffinity fallback is fragile, and performance concerns exist around unbounded caching and repeated preference saves. These must be addressed before merge.

## Recommended Fixes
1. Add null/empty guards in KeywordWithType.getTitle LANDWALK case and fallback for AFFINITY pluralisation
2. Restore Protection.getProtectionValid as deprecated shim delegating to new parsing logic or migrate callers
3. Add validation in Protection.parse to reject empty details and default to safe value
4. Implement debounce for preference saves in CardOverlaySettingsDialog (e.g., save on dialog close)
5. Add cache size limit and eviction policy for CardInfoPopup images
6. Cache controller queries in KeywordInfoUtil.annotateKeywordCounts to avoid repeated lookups
7. Add unit tests for KeywordInfoUtil edge cases: empty keyword string, null cardView, missing typeParam

---

PR #9809, opened 2026-02-19
Fix card name parsing in getCardByName for allCardVariants mode

vanja-ivancevic:
## Summary

Follow-up fix for #9798. When `useAllCardVariants` is enabled, `getCardByName()` and `getCardByNameAndEdition()` receive card names in `"Name|SET"` format (e.g., `"Verdeloth the Ancient|INV"`) from reward/shop JSON data. However, unlike `CardDb.getCard()` which parses this format via `CardRequest.fromString()`, `CardDb.getAllCards()` does a direct map lookup by card name — so `"Verdeloth the Ancient|INV"` matches nothing, and every card reward silently falls back to Wastes.

**Changes:**
- Parse the `|SET` suffix in `getCardByName()` before passing to `getAllCards()`/`getUniqueCardsNoAlt()`
- When an edition is specified, prefer cards from that edition (falling back to any allowed printing if unavailable)
- Same name parsing in `getCardByNameAndEdition()` for its `getAllCards()` lookup

**Not affected:** Callers passing plain card names (from `PaperCard.getCardName()`, land generation, SpellSmith) — no `|` separator means no parsing occurs, identical behavior to before.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/data/RewardData.java
+++ b/forge-gui-mobile/src/forge/adventure/data/RewardData.java
@@ -6,6 +6,7 @@ import forge.ImageKeys;
 import forge.StaticData;
 import forge.adventure.util.*;
 import forge.adventure.world.WorldSave;
+import forge.card.CardDb;
 import forge.card.CardEdition;
 import forge.deck.Deck;
 import forge.item.PaperCard;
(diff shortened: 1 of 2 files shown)


Hanmac:
its more like a general problem, like why for a function like this: `getCardByNameAndEdition` the Set is still part of the name and not the edition? Shouldn't it be handled before that?

vanja-ivancevic:
Thanks for the feedback. To confirm, are you suggesting we parse the "`Name|SET`" format at the call sites (in `RewardData.generate())` using `CardRequest.fromString() `before calling `getCardByName/getCardByNameAndEdition`, so those utility methods only ever receive clean card names?

vanja-ivancevic:
I have refactored and tested it like this. It works well too. What do you think?

Jetz72:
Might be more utility in implementing a `CardUtil.getCard(CardRequest/String cardRequest)` method that applies whatever filters and fallbacks are necessary. Then you could specify specific prints, foiling, and flags instead of just sets. But this should be alright for now.

vanja-ivancevic:
Thanks for the input. Sounds better long term. I'll work on it in another PR.

PR #9819, opened 2026-02-20
Check for error codes when downloading images

tehdiplomat:


Jetz72 (approved):
If it works, a good follow up would be to switch the boolean `disableScryfallDownload` kill-switch to a 30 second timeout or so.

PR #9822, opened 2026-02-20
Fixup some debugging code for scryfall issues

tehdiplomat:


forge-gui/src/main/java/forge/util/ImageFetcher.java
```diff
@@ -199,10 +199,6 @@ public void fetchImage(final String imageKey, final Callback callback) {
             if (useArtCrop) {
                 filename = TextUtil.fastReplace(filename, ".full", ".artcrop");
             }
-            if (ImageKeys.missingCards.contains(filename)) {
```
tehdiplomat, line 202: I added this earlier today thinking it was problematic, but it prevents shops from downloading art. I'm not sure why thats true, but it seems like it is.

Jetz72: I can't tell just from a glance if that field is supposed to be "cards we definitely don't have the image file for right now" or "cards that we don't have the image file for and never will".

tehdiplomat: Yea this whole area needs some love. Ideally, we would just have a cache where we dropped the name in the cache as soon as it was requested, Go fetch the image once only, and then replace the cache if the image was downloaded.

PR #9824, opened 2026-02-20
Fix NPE in ProtocolMethod.checkReturnValue on null returns

MostCromulent:
Fixes crash reported on Discord network play channel.

image

## Summary
- Add null check in `checkReturnValue()` to prevent NPE in diagnostic code
- This method was intentionally softened from a throw to a warning log, but a null return value caused the log statement itself to NPE on `value.getClass()`
- The fix restores the intended warn-and-continue behavior so the actual caller can handle the null (or fail with a more informative error closer to the real problem)

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/net/ProtocolMethod.java
+++ b/forge-gui/src/main/java/forge/gamemodes/net/ProtocolMethod.java
@@ -176,7 +176,6 @@ public enum ProtocolMethod {
             return;
         }
         if (value == null) {
-            System.err.printf("IllegalStateException: Protocol method %s: null returned by client, expected %s  (ProtocolMethod.java)%n", name(), getReturnType().getName());
             return;
         }
         if (!ReflectionUtil.isInstance(value, returnType)) {


Hanmac:
i'm pretty sure that some methods, like `chooseSpellAbilityToPlay` should be allowed to return null
so these cases should not do a warning

MostCromulent:
> i'm pretty sure that some methods, like `chooseSpellAbilityToPlay` should be allowed to return null so these cases should not do a warning

AI analysis below. Does this approach work for you?

```
At least three IGuiGame methods declared in the protocol can return null during normal gameplay:
- getAbilityToPlay, returns null when user cancels or no abilities available
- ChooseSingleEntityForEffect, returns null when isOptional=true and user cancels
- showInputDialog, returns null when user clicks cancel.

These are normal user-driven outcomes, not error conditions. A System.err warning for each one 
creates log noise and could mislead anyone reading logs into thinking something is wrong.

The fix should be adjusted to silently accept null rather than printing a warning. The method already 
knows which return type is expected, but it doesn't know whether null is a valid value for that specific 
method — and for at least three methods it is. Since this is a validation/diagnostic method 
(not business logic), the cleanest fix is to simply return early on null without the System.err message:

  if (value == null) {
      return;
  }

This still prevents the NPE (the original bug), stops treating legitimate nulls as errors, and keeps 
the method focused on type-checking non-null values. If we wanted to be more precise, we could 
annotate which ProtocolMethod entries allow null returns and only warn on the others, but that 
would probably be over-engineering for a diagnostic helper.

The type-mismatch warning (line 184) remains untouched and still catches actual errors.
```

Hanmac:
I think just returning on null could be enough

The cases there null should not happen (like return list or fixed values) might be refactored later?

Nullable return types might be wrapped in an `Optional<T>`, but I don't know if this Enum would like that 🤔

tool4ever:
I might be overthinking this but looks to me like normally null shouldn't be a problem:
image

will try to investigate why it was able to get a null from the primitive method

MostCromulent:
If it helps, from the Discord screenshots it looked like there was a version mismatch going on as well underneath the NPE - might be a product of the pre/vs post GameEvent refactor?

PR #9825, opened 2026-02-20
Fix TargetingPlayer abilities failing when AI plays against human

vanja-ivancevic:
## Summary
- Fix AI-activated TargetingPlayer abilities (e.g. Preacher, Arena) that fail repeatedly when played against a human opponent
- Name the AI evaluation thread `"Game AI Eval"` so Forge's threading model correctly identifies it
- Extend the timeout when a human TargetingPlayer is detected, so the human has time to choose targets

## Problem
When the AI evaluates activated abilities with `TargetingPlayer` (e.g. Preacher, Arena), it calls `chooseTargetsFor()` on the human player's controller from a worker thread. This causes two problems:

1. The worker thread was unnamed (e.g. `Thread-42`), so `ThreadUtil.isGameThread()` misidentified it as the GUI thread. This caused `InputSyncronizedBase.awaitLatchRelease` to throw `IllegalStateException`.

2. Even with the thread correctly identified, the AI evaluation runs with a 5-second timeout (`AI_TIMEOUT`). The human player's UI prompt blocks the worker thread, and if the human doesn't respond in 5 seconds the evaluation times out.

Both failures are caught by `FutureTask.get()` so the game doesn't crash, but the AI never activates the ability even though it wants to. This repeats every turn the AI has an untapped TargetingPlayer card with valid targets.

Triggered abilities with TargetingPlayer (e.g. The Abyss) are unaffected since triggers run on the main game thread.

## Fix
1. Name the AI evaluation thread `"Game AI Eval"` so `ThreadUtil.isGameThread()` returns true, preventing the `IllegalStateException`.

2. Before starting the evaluation thread, scan the candidate spells for a human `TargetingPlayer`. If found, use `Integer.MAX_VALUE` for the timeout instead of `AI_TIMEOUT`, so the human has unlimited time to choose targets. All existing AI evaluation and targeting logic is preserved.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
+++ b/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
@@ -863,11 +863,8 @@ public class PlayerControllerAi extends PlayerController {
             SpellAbility cur = sa;
             while (cur != null) {
                 if (cur.usesTargeting() && cur.hasParam("TargetingPlayer")) {
-                    Player targetingPlayer = AbilityUtils.getDefinedPlayers(cur.getHostCard(), cur.getParam("TargetingPlayer"), cur).get(0);
                     cur.clearTargets();
-                    cur.setTargetingPlayer(targetingPlayer);
-                    // TODO AI won't really understand it should make a bad choice
-                    targetingPlayer.getController().chooseTargetsFor(cur);
+                    cur.getTargetingPlayer().getController().chooseTargetsFor(cur);
                     // there's a chance a target gets selected that makes the cost unaffordable
                     if (!ComputerUtilCost.canPayCost(sa, sa.getActivatingPlayer(), false)) {
                         sa.resetTargets();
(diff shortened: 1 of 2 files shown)


tool4ever (changes requested):
not really in favor of this approach, as it now skips any logic for evaluating the targets
so the AI will just become 100% eager with these
ideally this param would be detected and timeout increased when this happens

good catch with the thread name though, will def cherry pick this

vanja-ivancevic:
Thanks for the review. Reworked per your feedback. Now only `AiController.java` is changed:

- Thread naming fix ("Game AI Eval") - same as before
- Before starting the evaluation thread, scan the candidate spells for a human `TargetingPlayer`. If found, use `Integer.MAX_VALUE` for the timeout instead of `AI_TIMEOUT`, so the human has unlimited time to choose targets (otherwise the game becomes unrecoverable on timeout).

Tested briefly and it works.

tool4ever:
ok sorry - I have chosen to reimplement a cleaner variant of the previous version 😅
this was mostly motivated because I realized the check for being able to pay the cost only happened afterwards
thanks for tackling this

tool4ever (approved):

vanja-ivancevic:
That's cool :). Thanks for fixing it!

PR #9827, opened 2026-02-20
Adventure: Show ante results popup after battle

vanja-ivancevic:
- After a game with ante enabled, a popup now shows each card won or lost, with the full card image rendered and tappable to zoom
- Won cards include an "Auto-Sell" button showing the gold value, which marks the card for auto-sell (same behavior as the reward scene toggle)
- Supports multiple cards from ownership-change effects (Bronze Tablet, Darkpact, etc.) by showing one popup per card sequentially
- Refactored the post-game popup chain (ante results -> boss dialogue -> exit) for clearer control flow

Works well in testing so far.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
+++ b/forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
@@ -7,6 +7,9 @@ import com.google.common.collect.ImmutableList;
 import forge.Forge;
 import forge.Graphics;
 import forge.LobbyPlayer;
+import forge.card.CardRenderer;
+import forge.card.CardRenderer.CardStackPosition;
+import forge.card.CardZoom;
 import forge.adventure.character.EnemySprite;
 import forge.adventure.character.PlayerSprite;
 import forge.adventure.data.*;
@@ -20,6 +23,7 @@ import forge.adventure.util.Current;
 import forge.assets.FBufferedImage;
 import forge.assets.FSkin;
 import forge.deck.*;
+import forge.game.card.CardView;
 import forge.game.GameRules;
 import forge.game.GameType;
 import forge.game.player.Player;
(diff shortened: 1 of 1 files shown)


forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
```diff
@@ -174,6 +207,66 @@ private FOptionPane createFOption(String message, String title, FBufferedImage i
...
+    }
+
+    private void showAnteCardsSequentially(List<PaperCard> cards, int index, boolean won, Runnable onDone) {
+        if (index >= cards.size()) {
+            onDone.run();
+            return;
+        }
+        PaperCard card = cards.get(index);
+        Runnable next = () -> showAnteCardsSequentially(cards, index + 1, won, onDone);
+        showAnteCardPopup(won ? "Won from Ante" : "Lost to Ante", card, won, next);
```
Jetz72: With the recent support for Blacker Lotus, a card lost might not necessarily be due to ante. Maybe use more general titles, like "Gained Card" and "Lost Card"?

vanja-ivancevic: Rebased onto latest master to pick up #9712. The popups now show for event matches as well. Auto-sell button is disabled during events since won cards go to the event deck sideboard rather than the player's collection. 

Also renamed the titles to "Card Gained" / "Card Lost" to cover Blacker Lotus and other non-ante ownership changes.

What do you think?

Jetz72:
I'm a bit iffy on the way the daisy-chained runnables are constructed, but we can cross that bridge if we ever need to add another. There's a comment above `removeLostCardFromPools` that looks like it was lost in the shuffle, and a merge conflict that needs resolving, but overall seems good to me.

vanja-ivancevic:
Rebased and restored the lost comment above `removeLostCardFromPools`.

Noted on the chaining - one option would be merging the won/lost card lists into a single pass to remove the `showAnteResults` indirection, but happy to leave it as-is for now if you're satisfied with the current state.

PR #9832, opened 2026-02-20
Refactor Soul Sculptor

tool4ever:


forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
```diff
@@ -953,8 +953,6 @@ public void run() {
             });
```
Hanmac, line 953: Simple GameCommand can be replaced with functions like this:
`game.getEndOfTurn().addUntil( () -> game.getEndOfTurn().addUntil(until));`

(if the until parameter is made final)

tool4ever: then how would it get a fixed serialVersionUID?

Hanmac: It's not needed in this case

(also while GameCommands are serializable, i don't see where that is used)

Hanmac: example where i used that already with DetainEffect:
https://github.com/Card-Forge/forge/blob/4cdd44a63db4a15fa2a0fe83f526f62be11205e2/forge-game/src/main/java/forge/game/ability/effects/DetainEffect.java#L22

PR #9834, opened 2026-02-20
TMNT spoilers, 20th February

Fulgur14:


forge-gui/res/cardsfolder/upcoming/ooze_spill.txt
```diff
@@ -0,0 +1,5 @@
+Name:Ooze Spill
```
tool4ever, line 1: missing ManaCost seems your favorite this set :P

forge-gui/res/cardsfolder/upcoming/stomped_by_the_foot.txt
```diff
@@ -0,0 +1,7 @@
+Name:Stomped by the Foot
+Types:Instant
```
squee1968, line 2: missing mana cost

PR #9843, opened 2026-02-20
Move keyword ability text to localization keys

MostCromulent:
## Summary

Per discussion with @Jetz72 on #9806: Moves keyword ability display names and reminder text from hardcoded Java strings in `Keyword.java` to `en-US.properties` localization keys. Follows the same pattern already used by `KeywordAction` in that PR.

- 199 display name entries (`lblKwAbility*`) and 199 reminder text entries (`lblKwAbility*Reminder`)
- No wording changes — exact same text, just externalized
- `getDisplayName()` and `getReminderText()` both go through Localizer
- The English `displayName` field is retained because `smartValueOf()` uses it to parse keyword names from card scripts — it's not used for display

---


forge-game/src/main/java/forge/game/keyword/Keyword.java
```diff
@@ -2,225 +2,241 @@
...
-        reminderText = reminderText0;
         displayName = displayName0;
+        this.translationKey = "lblKwAbility" + toCamelCase(name());
+    }
+
+    private static String toCamelCase(String enumName) {
+        StringBuilder sb = new StringBuilder();
+        boolean capitalize = true;
+        for (char c : enumName.toCharArray()) {
+            if (c == '_') {
```
Jetz72: If this is private, hard-coded to start capitalized, and only used to get the translation key, may as well fold in the `lblKwAbility` prefix and make it a dedicated `getTranslationKey` method?

forge-game/src/main/java/forge/game/keyword/Keyword.java
```diff
@@ -271,9 +287,13 @@ public static KeywordInterface getInstance(String k) {
...
     }
 
+    public String getDisplayName() {
+        return Localizer.getInstance().getMessage(translationKey);
+    }
+
     @Override
     public String toString() {
-        return displayName;
+        return getDisplayName();
```
Hanmac: This could be problematic
I had other enums crash on "toString" because Localizer was not defined yet

🤔

Hanmac:
@tool4ever @tehdiplomat what are the others' opinion in extracting the Title/Translation strings?

The thing that is bugging me are, that we pass the reminder texts though a Formatter twice.
- First, the Formatter for Locale without parameters (need to see if it would complain? The other formatter does)
- Second, the Translated Text are filled with the Values for the formatted text.

Like it would probably be better to pass the parameters directly into the Locale Formatter?
(or is the problem there that, like for `Amount`, the stuff for `%d` gets replaced to `%s` for X?)

A bigger case are now the Mixmatched stuff with Keywords that use Colors or CardTypes/CreatureTypes.
Like for example, "Enchant creature", the "Enchant" part gets translated, the "Creature" part does not. (but could?)
(and let's not get started with the whole "creature you control" stuff, or gender-specific translations)

Also, some cases might need extra love, like "Hexproof from <>", where the "from" might not be translated yet.

tool4ever (changes requested):
Honestly I'm not sure this much hardcoding is solving the problem adequately:
FWIW `keywordsToText` already leverages `CardTranslation` and while that approach adds a bit of heuristics Levenshtein Distance seems pretty solid in general and can probably be tweaked further...

the only thing missing there is currently it's only looking at intrinsic ones? giving it some kind of fallback strategy support seems like a more valueable direction to me if possible 🤔

PR #9850, opened 2026-02-21
Improve MagicStack stack removal logic and stack notifications via SwingWorkers

adrianrlaw:
As a combo player, spamming spells/abilities and abusing the stack is my MO. Taking this playstyle to Forge however has caused some noticeable performance problems that I suspected to be programmatic.

This PR aims to provide two sets of improvements:

**1) Centralize stack instance removal in `MagicStack`.**

- The stack resolution path had duplicated logic for removing stack instances, specifically when resolving copied spells/abilities, resulting in double removals.

**2) Rework stack notification UI behavior to be more performant.**

 - Move notification gating off the EDT into background workers.
   - Previously, gating was handled by passing a recursive call to `notifyStackAddition()` into `SwingUtilities.invokeLater()`.
 - Serialize notifications with a lock/condition on stack index.
 - Ensure proper cleanup (and cancellation) of outstanding stack notifications at end of the game procedures.
   - There was no systematic way to cancel or clean up in-flight notifications when a game was finished.

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
+++ b/forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
@@ -24,6 +24,9 @@ import java.awt.image.BufferedImage;
 import java.util.*;
 import java.util.List;
 import java.util.Map.Entry;
+import java.util.concurrent.Callable;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.RunnableFuture;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
@@ -169,6 +172,7 @@ public final class CMatchUI
     private int nextNotifiableStackIndex = 0;
     private final Lock stackIndexNotifyUpdate = new ReentrantLock();
     private final Condition updatedNotifiableStackIndex = stackIndexNotifyUpdate.newCondition();
+    private final List<RunnableFuture<Void>> stackNotificationWorkerList = new ArrayList<RunnableFuture<Void>>();
 
     public CMatchUI() {
         this.view = new VMatchUI(this);
@@ -818,6 +822,19 @@ public final class CMatchUI
     @Override
     public void finishGame() {
         FloatingZone.closeAll(); //ensure floating card areas cleared and closed after the game
+
+        // Clean up all workers on the stack
+        for (RunnableFuture<Void> worker : stackNotificationWorkerList)
+            worker.cancel(true);
+
+        // Reset nextNotifiableStackIndex
+        stackIndexNotifyUpdate.lock(); // Acquire lock
+        nextNotifiableStackIndex = 0;
+        updatedNotifiableStackIndex.signalAll(); // signal all workers waiting on condition
+        stackIndexNotifyUpdate.unlock(); // Release lock
+
+        stackNotificationWorkerList.clear();
+
         final GameView gameView = getGameView();
         if (hasLocalPlayers() || gameView.isMatchOver()) {
             new ViewWinLose(gameView, this).show();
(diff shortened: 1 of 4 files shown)


forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
```diff
@@ -1378,15 +1423,24 @@ public void notifyStackAddition(GameEventSpellAbilityCast event) {
...
             }
             // In any case, I have to increase the counter
+            stackIndexNotifyUpdate.lock(); // Acquire lock
             nextNotifiableStackIndex++;
-        } else {
-            // Not yet time to show the modal - schedule the method again, and try again later
-            Runnable tryAgainThread = () -> notifyStackAddition(event);
-            GuiBase.getInterface().invokeInEdtLater(tryAgainThread);
+            // TODO: signalAll() may be unnecessary since we only increment in this worker.
+            updatedNotifiableStackIndex.signalAll(); // Notify threads waiting on updatedNotifiableStackIndex
```
tool4ever, line 1433: this seems like so much work for such a basic thing:
what stops us from just storing each GameEvent in a List and then displaying the modals until it's empty again?

adrianrlaw: I'm not sure why `nextNotifiableStackIndex` is being tracked but I'm assuming it's to help with ordering the modals (please let me know and correct me if I'm wrong). With this assumption in mind, sure we can probably keep each GameEvent in a List and iterate through it to find which GameEvent has a stack index that matches, but that sounds expensive and we'd have to wait on a GameEvent to be inserted into the list again anyways when it ends up empty (which also requires condition checks and locking to be thread safe if we want to avoid locking up the EDT thread). Also, with the previous implementation, we'd just re-queue our `notifyStackAddition()` to the back of the EDT queue recursively, which with my quick adventures with Google is supposedly discouraged. SwingWorker is supposed to already kinda do this on its own without hogging up EDT time. The ultimate goal here, I guess, is reducing the cpu time spent on checking `nextNotifiableStackIndex`. _(Note: before coming up with these changes, I generated a flame graph to see where the slowdown occured. notifyStackAddition() was by far the biggest hog for some reason.)_
That said, the most important change here is fixing the double-removal issue in MagicStack.

adrianrlaw: On a side note, a function recursing over a condition without modifying it on its own (in the branch where it recurses) is very scary to me.

tool4ever: if the goal is to always display the topmost stack entry you could just grab them from the end of the list 🤔
and when you receive a GameEvent while the list isn't empty the method should just return early (the previous call can just loop on those additional ones) 🤷‍♂️

adrianrlaw: I'm not very experienced with Java, so forgive me for having difficulty envisioning the implementation. What happens when the stack grows to 100, 200, or maybe even 1000 (Storm is probably Wizard's funniest mistake)? Do we want to loop over a large list of game events looking for the one we want to show on the EDT thread or let each event handle itself (running and cleanup)? Also, what if Wizards cooks up something that changes the order of things resolving on the stack (not that they will but I can imagine them doing this, or maybe they already have)? How do we handle cancellation?

tool4ever: hmn, I admit to never really having used this feature, so maybe I misunderstood how it works 😅
when the stack gets filled with multiple entries is it supposed to only show you a modal for the top - and only when that leaves the stack you'd get one for the next below etc.?

adrianrlaw: I think I'm starting to understand what this section of code is trying to do but I may be reading this wrong. So ya know when there's like a trigger that's supposed to reveal some sort of information when it goes on the stack or when the AI is supposed to reveal or make public some sort of game object as part of a choice or additional cost? This part of the code is supposed to catch and handle that and notify the player that some sort of information was revealed. Think AI sacrifices a facedown card as part of an additional cost sort of thing.

tool4ever: not really, we have other ways to reveal stuff?
Nobody has worked on this feature in a long time, it doesn't even exist on mobile and tbh personally I don't see the point of showing a huge window with redundant information but I thought you were finding it useful :P
If you don't even have its preference enabled but it was still causing significant CPU time something is seriously wrong...

adrianrlaw: the double removal in MagicStack was probably the biggest problem tbh...

PR #9854, opened 2026-02-21
CardFactoryUtil: unkeyword Adapt

Hanmac:
This turns Adapt Keyword back into an activated Ability with special Param (that already existed in the code)

i also moved some Param shenanigans from the AbilityFactory into the SpellAbilityEffect helper

that place can also create SpellDescription if needed (works for basic ones, not Charm Effects)


Hanmac:
(Bolster and Support could also benefit from some cleanup)

tool4ever:
@Hanmac
I wonder if it's even cleaner if you create a shallow subclass `AdaptEffect extends CountersPutEffect`...
so you don't even need the param 🤔

Hanmac:
> @Hanmac I wonder if it's even cleaner if you create a shallow subclass `AdaptEffect extends CountersPutEffect`... so you don't even need the param 🤔

my current reduction is like this: `PutCounter | Adapt$ 3` (where CounterNum is set by Adapt)

PR #9863, opened 2026-02-21
feat: make "Auto" mulligan button smarter

BlythT:
Currently it discards the first N cards from your
hand. If your hand is sorted by mana value, this
is often all lands.

Using the existing ScoreHand AI, we can compare
all combinations of hands with N cards removed
and keep the best one; intelligently discarding
lands or permanents based on colours, flood vs
screw, etc.


forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -2067,18 +2067,118 @@ public static boolean playImmediately(Player ai, SpellAbility sa) {
...
+
     public static int scoreHand(CardCollectionView handList, Player ai, int cardsToReturn) {
         // TODO Improve hand scoring in relation to cards to return.
         // If final hand size is 5, score a hand based on what that 5 would be.
         // Or if this is really really fast, determine what the 5 would be based on scoring
         // All of the possibilities
 
-        final AiController aic = ((PlayerControllerAi)ai.getController()).getAi();
+        int mulliganThreshold = 4; // Sensible default for humans
+        if (ai.getController() instanceof PlayerControllerAi) {
```
tehdiplomat: Probably need to rename `ai` to `player` if we allow non-ai Players into this function

BlythT: Added that: https://github.com/Card-Forge/forge/pull/9863/changes/323d394e0dcb340d62ca91ccb6c46196115d1687

forge-gui/src/main/java/forge/gamemodes/match/input/InputLondonMulligan.java
```diff
@@ -81,18 +82,14 @@ protected final void onOk() {
 
     @Override
     protected final void onCancel() {
-        int cardsLeft = toReturn - selected.size();
-        int count = 0;
-        for(Card c : player.getZone(ZoneType.Hand).getCards()) {
+        // Use the AI to determine the best cards to return to the library
+        CardCollectionView toMulligan = ComputerUtil.chooseBestCardsToReturn(player, toReturn);
```
tool4ever: nice idea
but shouldn't this still pass `cardsLeft`?
otherwise if the user does a partial selection before pressing auto you could return more than legal

BlythT: Fair point, I didn't realize you could do partial selection auto mulligans. I'll work that in to the design

BlythT: Applied a fix here: https://github.com/Card-Forge/forge/pull/9863/changes/a91678a1da7c038ddd57e8d2415472bb49ae4074

jamincollins:
## Findings
#### 1. combinationsWithRemoved deduplication by card name discards distinct physical cards
The combinatorial generator deduplicates combinations using only `Card::getName`, so two copies of the same named card are treated as interchangeable. For hands containing multiple distinct instances with different game state (counters, attachments, etc.), the algorithm can skip evaluating a genuinely different kept set.

Test case:
```java
Card bolt1 = createCard("Lightning Bolt", p);
Card bolt2 = createCard("Lightning Bolt", p);
bolt1.addCounter(CounterType.CHARGE); // different state
Card island = createCard("Island", p);
CardCollectionView hand = new CardCollectionView(Arrays.asList(bolt1, bolt2, island));
List<CardCollectionView> combos = ComputerUtil.combinationsWithRemoved(hand, 1);
// Expected 3 distinct removals, dedup by name yields 2
assertEquals(3, combos.size());
```
Impact: Sub-optimal mulligan choices when cards share names but differ in game-relevant state; MTG rules require per-instance selection.

#### 2. chooseBestCardsToReturn throws on empty hand with cardsToReturn > 0
`chooseBestCardsToReturn` validates `cardsToReturn > hand.size()` before generating combinations, but an empty hand with `cardsToReturn = 0` is handled, while `cardsToReturn > 0` throws `IllegalArgumentException`. During mulligan, a player could theoretically have an empty hand due to effects, leading to an unhandled exception in UI flow.

Test case:
```java
Player p = game.getPlayers().get(0);
CardCollectionView emptyHand = CardCollection.EMPTY;
try {
    ComputerUtil.chooseBestCardsToReturn(p, emptyHand, 1);
    fail("Expected IllegalArgumentException");
} catch (IllegalArgumentException e) {
    // Expected
}
```
Impact: Uncaught exception crashes the mulligan UI instead of gracefully returning no cards.

#### 3. scoreHand assumes non-null ManaCost
`scoreHand` filters cards with `c.getManaCost().getCMC()` and `chooseBestCardsToReturn` sums `c.getManaCost().getCMC()` without null checks. Cards without a mana cost (tokens, some special cards) can cause NPE.

Test case:
```java
Card token = new Card(); // getManaCost() returns null
CardCollectionView hand = new CardCollection();
hand.add(token);
int score = ComputerUtil.scoreHand(hand, player, 0); // NPE
```
Impact: Runtime NPE during mulligan evaluation.

#### 4. scoreHand mulliganThreshold hard-coded for humans changes AI scoring semantics
`scoreHand` now uses `mulliganThreshold = 4` for non-AI players, whereas previously the threshold was always read from `AiProps.MULLIGAN_THRESHOLD`. This changes scoring for human players who may be using the same method for AI decisions, breaking backwards compatibility.

Test case:
```java
Player human = game.getPlayer(0); // controller instanceof PlayerControllerHuman
int thresholdUsed = 4; // hard-coded
// Previously threshold would be AiProps.MULLIGAN_THRESHOLD if AI, else undefined
```
Impact: Human hands are scored with a fixed threshold, diverging from previous AI behavior and potentially changing mulligan recommendations.

#### 5. InputLondonMulligan.onCancel does not guard against chooseBestCardsToReturn exception
`InputLondonMulligan.onCancel` calls `ComputerUtil.chooseBestCardsToReturn` without try/catch. If the AI throws, the mulligan UI will fail to complete, leaving the game in an inconsistent state.

Test case:
```java
// Simulate remainingToReturn > unselectedCards.size()
int remainingToReturn = 5;
CardCollectionView unselected = ... // size 3
ComputerUtil.chooseBestCardsToReturn(player, unselected, remainingToReturn); // throws
```
Impact: Uncaught exception prevents mulligan completion, side effect on game flow.

### Performance and security
#### 6. combinationsWithRemoved generates O(C(n,k)) candidates without size guard
For a hand of size n with k cards to return, the generator evaluates `C(n, n-k)` combinations. With `handSize=7` this is fine, but the method has no guard for larger hands (e.g., after draw effects). `C(20,10)=184,756` combinations will cause UI freeze.

Test case:
```java
CardCollectionView largeHand = createHandOfSize(20);
ComputerUtil.combinationsWithRemoved(largeHand, 10);
// duration grows factorially
```
Impact: Potential denial-of-service / UI freeze for edge cases with large hands.

## Verdict
Request Changes
The smart auto mulligan improves UX but introduces null-safety gaps, unchecked exceptions, and performance risks that need hardening before merge. The deduplication strategy also risks sub-optimal choices.

## Recommended Fixes
1. Guard `chooseBestCardsToReturn` and `scoreHand` against null ManaCost and empty hands; return empty collection gracefully.
2. Add a size guard in `combinationsWithRemoved` to fall back to heuristic selection when hand size exceeds a threshold (e.g., >12).
3. Use card identity, not just name, for deduplication, or remove deduplication entirely for mulligan hands.
4. Wrap AI call in `InputLondonMulligan.onCancel` with try/catch to prevent UI crash.
5. Make `scoreHand` threshold configurable for human players or document the hard-coded 4.

---

PR #9866, opened 2026-02-22
Fix ConcurrentModificationException in CMatchUI card iteration during network play

MostCromulent:
## Summary

Fixes a `ConcurrentModificationException` thrown during network play when the EDT iterates zone card collections concurrently with Netty IO thread updates.

**The bug:** `refreshField()`, `setSelectables()`, and `clearSelectables()` in `CMatchUI` iterate the live `FCollection` (ArrayList-backed) returned by `PlayerView.getCards()`. During network play, the Netty IO thread applies incoming game state updates via `TrackableCollectionType.copyChangedProps()`, which mutates these same collections with `remove()`/`add()` calls. The ArrayList iterator detects the modification and throws `ConcurrentModificationException`, aborting card rendering mid-loop.

**Stack trace from user report:**
```
java.util.ConcurrentModificationException
    at java.base/java.util.ArrayList$Itr.checkForComodification
    at java.base/java.util.ArrayList$Itr.next
    at forge.screens.match.CMatchUI.updateCards(CMatchUI.java:542)
    at forge.screens.match.CMatchUI.lambda$refreshField$4(CMatchUI.java:610)
```

**The fix:** Call `threadSafeIterable()` (existing method on `FCollectionView` that creates a snapshot copy) before iterating zone collections on the EDT. Five call sites changed, no new code introduced.

This is a pre-existing bug (the vulnerable code dates to 2021, commit `2c8a658270`) but only manifests in network play where the Netty IO thread runs independently of the EDT.

---

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
+++ b/forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
@@ -574,10 +574,10 @@ public final class CMatchUI
         FThreads.invokeInEdtNowOrLater(() -> {
             for (final PlayerView p : getGameView().getPlayers()) {
                 if (p.getCards(ZoneType.Battlefield) != null) {
-                    updateCards(p.getCards(ZoneType.Battlefield).threadSafeIterable());
+                    updateCards(isNetGame() ? p.getCards(ZoneType.Battlefield).threadSafeIterable() : p.getCards(ZoneType.Battlefield));
                 }
                 if (p.getCards(ZoneType.Hand) != null) {
-                    updateCards(p.getCards(ZoneType.Hand).threadSafeIterable());
+                    updateCards(isNetGame() ? p.getCards(ZoneType.Hand).threadSafeIterable() : p.getCards(ZoneType.Hand));
                 }
             }
             FloatingZone.refreshAll();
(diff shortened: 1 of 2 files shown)


Hanmac (dismissed):

tool4ever:
@MostCromulent
wouldn't `MatchController` also need these changes?

MostCromulent:
> wouldn't `MatchController` also need these changes?

Yes, my oversight sorry. Now included.

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
```diff
@@ -607,7 +607,7 @@ public void refreshField() {
         FThreads.invokeInEdtNowOrLater(() -> {
             for (final PlayerView p : getGameView().getPlayers()) {
                 if (p.getCards(ZoneType.Battlefield) != null) {
-                    updateCards(p.getCards(ZoneType.Battlefield));
+                    updateCards(p.getCards(ZoneType.Battlefield).threadSafeIterable());
```
tool4ever: I feel like especially here we're at a very busy code path:
so copying the Collection locally for no reason might cost too much performance?

would it be better to first check `isNetGame()` if it's needed?

PR #9868, opened 2026-02-22
Fix batching for network game events to reduce redundant GameView serialization

MostCromulent:
@tool4ever: issue with GameEvent refactor discovered while investigating user reports of recent increased network lag, and AI solution below:

---
## Summary

PR #9760 (GameEvent refactor) changed how game events reach remote clients. Previously, the server's `FControlGameEventHandler` batched events into coalesced UI updates and `NetGuiGame` sent ~3-5 protocol messages per game action. After the refactor, a `GameEventForwarder` sends every individual `GameEvent` immediately via `NetGuiGame.handleGameEvent()`, which calls `updateGameView()` + `send(handleGameEvent, event)` per event.

`updateGameView()` serializes and sends the **complete `GameView`** over the network every time it's called. A single game action that fires N events now generates 2N network messages (N full GameView serializations + N event messages) instead of the ~4-6 messages it sent before.

There are ~200 `fireEvent` call sites across the game engine, and events frequently fire in bursts — for example, combat resolution, stack processing, or phase transitions can fire 10+ events in rapid succession. That means 10+ redundant full GameView serializations for a single game action.

This also affects the delta sync branch (NetworkPlay/main), though the cost is much smaller there — `updateGameView()` only sends changed properties and skips entirely when nothing has changed, so redundant calls are cheap rather than expensive. Still, batching eliminates unnecessary delta-sync comparisons and empty-check overhead.

## Fix

- **Batch events in `GameEventForwarder`** using EDT coalescing (the same `invokeInEdtLater` gate pattern used by `FControlGameEventHandler` before the refactor). Events that fire in rapid succession are collected and flushed as a single batch.
- **Add `handleGameEvents(List<GameEvent>)` to `IGuiGame`** with a default implementation that delegates to `handleGameEvent` one-by-one — no impact on local play, desktop GUI, or mobile.
- **Override in `NetGuiGame`** to call `updateGameView()` once, then send each event individually. This reuses the existing `ProtocolMethod.handleGameEvent` protocol method, so no client-side changes are needed.

For a game action firing 10 events, network messages drop from 20 (10 full GameView + 10 event) to 11 (1 GameView + 10 event). The full GameView serialization is the dominant cost, so this eliminates the main source of redundant network traffic.

---

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/net/server/NetGuiGame.java
+++ b/forge-gui/src/main/java/forge/gamemodes/net/server/NetGuiGame.java
@@ -299,6 +299,14 @@ public class NetGuiGame extends AbstractGuiGame {
     }
 
     @Override
+    public void handleGameEvents(List<GameEvent> events) {
+        updateGameView();
+        for (GameEvent event : events) {
+            send(ProtocolMethod.handleGameEvent, event);
+        }
+    }
+
+    @Override
     public boolean isNetGame() { return true; }
 
     @Override
--- a/forge-gui/src/main/java/forge/gui/control/GameEventForwarder.java
+++ b/forge-gui/src/main/java/forge/gui/control/GameEventForwarder.java
@@ -2,12 +2,39 @@ package forge.gui.control;
 
 import com.google.common.eventbus.Subscribe;
 import forge.game.event.GameEvent;
+import forge.gui.GuiBase;
 import forge.gui.interfaces.IGuiGame;
+
+import java.util.ArrayList;
+import java.util.List;
 
 public class GameEventForwarder {
     private final IGuiGame gui;
+    private final List<GameEvent> pendingEvents = new ArrayList<>();
+    private volatile boolean flushQueued = false;
+
     public GameEventForwarder(IGuiGame gui) { this.gui = gui; }
 
     @Subscribe
-    public void receiveGameEvent(GameEvent ev) { gui.handleGameEvent(ev); }
+    public void receiveGameEvent(GameEvent ev) {
+        synchronized (pendingEvents) {
+            pendingEvents.add(ev);
+        }
+        if (!flushQueued) {
+            flushQueued = true;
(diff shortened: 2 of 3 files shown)


forge-gui/src/main/java/forge/gamemodes/net/server/NetGuiGame.java
```diff
@@ -298,6 +298,14 @@ public void handleGameEvent(GameEvent event) {
         send(ProtocolMethod.handleGameEvent, event);
     }
 
+    @Override
+    public void handleGameEvents(List<GameEvent> events) {
+        updateGameView();
+        for (GameEvent event : events) {
```
tool4ever, line 304: this is unfortunate - need risk assessment:
1. L303 will essentially send the newest state of the Game before events (with potential Views from the short past) are transferred, right?
2. I'd say crashing seems unlikely since the IDs would most likely still exist after (and if not such case be guarded)?
3. so might only cause rare visual errors that sort itself out quickly? that might be acceptable 🤔

MostCromulent: Correct on all three:

>1. Yes — the GameView sent at L303 reflects the state after all batched events have been applied to the game model. The client then receives that "ahead" GameView followed by the individual events.
>2. This is the same pattern as the single-event path (L296-298) which already sends the current GameView before each event. The batch version just avoids doing N redundant updateGameView() calls — it's actually less out-of-sync than the single path was, since in the single path each successive updateGameView() was also "ahead" of the remaining events in the batch.
>3. Crash risk: very low. GameView references are by ID, and game objects aren't removed between the view snapshot and event processing. Any ID present in an event will still resolve on the client.
>4. Visual risk: minor and self-correcting. The client may briefly show a state slightly ahead of the event animations, but the next updateGameView() call (which happens frequently) reconciles everything. This is the existing behavior for single events already — batching doesn't make it worse.

PR #9872, opened 2026-02-22
CountersPutEffect: keywordAction Bolster

Hanmac:
I asked Judges, the `canReceiveCounters` shouldn't be for the choices there

i also added Reminder Text for the others

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
@@ -211,6 +211,9 @@ public class CountersPutEffect extends SpellAbilityEffect {
 
             CardCollection choices = CardLists.getValidCards(game.getCardsIn(choiceZone), sa.getParam("Choices"),
                     activator, card, sa);
+            if (counterType != null && !sa.hasParam("SkipReceiveCounters")) {
+                choices = CardLists.filter(choices, CardPredicates.canReceiveCounters(counterType));
+            }
 
             // TODO might use better message
             String title = Localizer.getInstance().getMessage("lblChooseaCard") + " ";
@@ -748,6 +751,7 @@ public class CountersPutEffect extends SpellAbilityEffect {
             sa.putParam("CounterNum", n);
             sa.putParam("Choices", "Creature.leastToughnessControlledByYou");
             sa.putParam("ChoiceTitle", Localizer.getInstance().getMessage("lblChooseACreatureWithLeastToughness"));
+            sa.putParam("SkipReceiveCounters", "True");
             if (!sa.hasParam("SpellDescription")) {
                 StringBuilder sb = new StringBuilder("Bolster");
                 sb.append(" ").append(n);


forge-game/src/main/java/forge/game/ability/effects/CountersPutEffect.java
```diff
@@ -221,9 +211,6 @@ protected void resolvePerType(SpellAbility sa, Player placer, CounterType counte
 
             CardCollection choices = CardLists.getValidCards(game.getCardsIn(choiceZone), sa.getParam("Choices"),
                     activator, card, sa);
-            if (counterType != null) {
```
tool4ever, line 224: that change is still wrong

this rule applies:
> 608.2d. If an effect of a spell or ability offers any choices other than choices already made as part of casting the spell, activating the ability, or otherwise putting the spell or ability on the stack, the player announces these while applying the effect. The player can't choose an option that's illegal or impossible...

Otherwise "Sacrifice a creature" would just allow you to pick something that can't be sacrificed too 🤯

there is nothing special about counters this way
the only reason why it shouldn't filter for e.g. Bolster is because it has "Action A: Choose" followed by "Action B: Do stuff with Choice"

tool4ever: this is also "madwarper confirmed":
https://www.reddit.com/r/magicTCG/comments/1qgtm74/comment/o0j52h6/

Hanmac: Again, Blight doesn't use CounterPutEffect, and Blight specific rules doesn't apply there

tool4ever: nothing specific for this:
> 701.68a. To "blight N" means to put N -1/-1 counters on a creature you control.

implementing it as `BlightEffect` and not like Bolster is just our own choice and makes no difference

Hanmac: @tool4ever you overlooked the next rule:
> 701.68b If a player is given the choice to blight but is unable to put N -1/-1 counters on a creature they control (usually because they control no creatures), they can’t choose to blight. 

Blight has this special rule. Bolster does not.

tool4ever: so what? it's not directly relevant for what we're discussing:
by default the keyword actions "Bolster N" and "Blight N" are both defined as mandatory

**if a game effect forces a player to do either they must do as much as possible**

again, the only reason you can possibly choose creatures affected by `CantPutCounter` is because Bolster is split into two actions

Hanmac: I already messaged you about that

I asked judges yesterday, and they confirmed that you can choose a creature you can't put counters on for Bolster

tool4ever: > I already messaged you about that
> 
> I asked judges yesterday, and they confirmed that you can choose a creature you can't put counters on for Bolster

yes, I know and I agree there

but you can't make this change for every `Choices$` effect:
e.g. _Choking Miasma_ with "If this spell was kicked, put a +1/+1 counter on a creature you control"

these ones are like Blight - you can't make an impossible choice!

PR #9879, opened 2026-02-22
Minor Netplay cleanup

tool4ever:
Revert 050ecb5

@MostCromulent
I can't think of any reason why only Mobile would still need this?
seemed to work fine anyway when I tested on that GUI

### DIFF
--- a/forge-game/src/main/java/forge/game/GameView.java
+++ b/forge-game/src/main/java/forge/game/GameView.java
@@ -112,22 +112,6 @@ public class GameView extends TrackableObject {
         set(TrackableProperty.PlayerTurn, PlayerView.get(phaseHandler.getPlayerTurn()));
     }
 
-    public void updateNeedsPhaseRedrawn(PlayerView p, PhaseType ph) {
-        set(TrackableProperty.PlayerTurn, p);
-        set(TrackableProperty.Phase, ph);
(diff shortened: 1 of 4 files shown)


MostCromulent:
AI analysis:

> Why mobile only? The theory would be a threading gap — mobile uses Libgdx (separate render thread) while desktop uses Swing (single EDT). But updatePhase() on mobile is already invoked via invokeInEdtLater(), which posts to the Libgdx thread. So updatePhase() already runs on the render thread and directly sets phaseLabel.setActive(true). The NeedsPhaseRedrawn mechanism then does the identical thing again on the next frame.    
>                                                                                                                                                   
>   The problematic side effect: updateNeedsPhaseRedrawn() writes back to GameView, overwriting PlayerTurn and Phase from the client side. That's the wrong direction — the client shouldn't be mutating the game's view state. This could mask or introduce subtle sync issues during network play.
> 
>   Is it still necessary? Almost certainly not:
>   - updatePhase() already does the work directly — the flag just repeats it
>   - Desktop handles the same call chain without any equivalent mechanism
>   - The writeback to GameView is a liability, not a benefit

tool4ever:
ok, happy our overlord agrees with me ;)

if any user complains we'll know where to look

thanks

PR #9845, opened 2026-02-21
Add life total change logging to game log

TytaniumDev:
I'm building a tool that uses headless Forge simulations to play commander games and then analyze the logs (more detail in #9844 ). Player life changes are currently not logged, which makes it hard to visualize game states at a high level.

This PR adds life logging. The code was primarily written by Claude but I've verified it and made sure tests were written to validate all the scenarios I could think of.

## Summary

- `GameEventPlayerLivesChanged` is already fired on every life total change but had no visitor in `GameLogFormatter` — the event was silently discarded
- Adds new `GameLogEntryType.LIFE` enum value, replacing the `// Where's life loss?` comment that was next to `DAMAGE`
- Adds `visit(GameEventPlayerLivesChanged)` to `GameLogFormatter` producing log lines like: `Life: Ai(1)-Doran Big Butts 40 -> 37`
- Adds localized message key `lblLogPlayerLifeChange` to all language files (English fallback for non-English locales)
- Adds 6 targeted tests in `GameLogLifeChangeTest`

No behavior changes to existing log entries.

Closes #9844

### Combat damage → life change

Two Nest Robbers (2/1 haste) attack an opponent at 4 life:

```
[COMBAT]       p1 assigned Nest Robber (1) and Nest Robber (2) to attack p2.
[COMBAT]       p2 didn't block Nest Robber (1).
               p2 didn't block Nest Robber (2).
[DAMAGE]       Nest Robber (1) deals 2 combat damage to p2.
[DAMAGE]       Nest Robber (2) deals 2 combat damage to p2.
[LIFE]         Life: p2 4 -> 0
```

### Lifegain from spell resolution

AI casts Lightning Helix (3 damage + 3 life) targeting opponent at 3 life:

```
[STACK_ADD]    p1 cast Lightning Helix targeting [p2]
[DAMAGE]       Lightning Helix (3) deals 3 non-combat damage to p2.
[LIFE]         Life: p2 3 -> 0
[LIFE]         Life: p1 20 -> 23
[STACK_RESOLVE] Lightning Helix (3) - Lightning Helix (3) deals 3 damage to p2. p1 gains 3 life.
```

### Existing tests (63/63 passing, no regressions)

```
[INFO] Tests run: 63, Failures: 0, Errors: 0, Skipped: 0
       -- in forge.ai.simulation.GameSimulationTest
```

### Full build with Checkstyle (all 13 modules passing)

```
[INFO] Reactor Summary for Forge Parent 2.0.11-SNAPSHOT:
[...]

PR #9889, opened 2026-02-23
Better game log UI: inline card images, improved verbosity presets

MostCromulent:
Screenshot 2026-02-23 182606 Screenshot 2026-02-23 182420

Screenshot 2026-02-23 190417

This PR makes four changes to improve the in-match log panel, particularly on desktop.

1. **Inline Card Images** — Log entries that reference a card now show a miniature card image (50x70px) on the left, modelled on the existing stack panel's `StackInstanceTextArea` pattern. Uses `CachedCardImage` for async loading. Togglable via Layout > View > Log Detail Level > Show Card Images.

2. **Card Hover** — Hovering over a card-linked log entry updates the card detail panel and card view panel, same as hovering over stack items. Face-down cards are guarded (no detail shown).

3. **Clearer log verbosity settings**  — Currently the log verbosity setting exposes all 18 internal `GameLogEntryType` values in a flat list. Most players don't know what "STACK_ADD" or "ZONE_CHANGE" means, and the ordinal cutoff filter means you can't include damage without also including everything above it.

PR replaces the 18-item list with three intuitive presets:

| Preset | Includes |
|--------|----------|
| **Low** | Turns, damage, mulligans, outcomes |
| **Medium** | Low + spells, combat, lands, zone changes |
| **High** | Everything including phases and mana abilities |

Old preference values are migrated automatically. 
The verbosity setting only affects the in-game UI log panel, not text game log output.

4. **Zone change logging** — Creatures dying (Battlefield → Graveyard) and being exiled (Battlefield → Exile) are now logged as `ZONE_CHANGE` entries, visible at Medium and above.

Desktop gets all features 1 - 4, as 1 and 2 are based on existing stack/card view functionality. 
Mobile gets features 3 and 4 only.

---

### DIFF
--- a/forge-gui-mobile/src/forge/screens/settings/SettingsPage.java
+++ b/forge-gui-mobile/src/forge/screens/settings/SettingsPage.java
@@ -9,6 +9,7 @@ import forge.StaticData;
 import forge.adventure.util.Config;
 import forge.ai.AiProfileUtil;
 import forge.assets.*;
+import forge.game.GameLogEntryType;
 import forge.game.GameLogVerbosity;
 import forge.gui.GuiBase;
 import forge.localinstance.properties.ForgeConstants;
@@ -25,9 +26,11 @@ import forge.screens.match.MatchController;
 import forge.sound.MusicPlaylist;
 import forge.sound.SoundSystem;
 import forge.toolbox.FCheckBox;
+import forge.toolbox.FDialog;
 import forge.toolbox.FGroupList;
 import forge.toolbox.FList;
 import forge.toolbox.FOptionPane;
+import forge.toolbox.FScrollPane;
 import forge.util.Utils;
 
 import java.util.*;
@@ -351,6 +354,31 @@ public class SettingsPage extends TabPage<SettingsScreen> {
             Forge.getLocalizer().getMessage("cbpGameLogEntryType"),
             Forge.getLocalizer().getMessage("nlGameLogEntryType"),
             GameLogVerbosity.class), 3);
+        lstSettings.addItem(new Setting(FPref.DEV_LOG_ENTRY_TYPE,
+            Forge.getLocalizer().getMessage("lblCustomLogSettings"),
+            Forge.getLocalizer().getMessage("lblLogVerbosityCustom")) {
+                @Override
+                public boolean isEnabled() {
+                    return GameLogVerbosity.fromString(
+                        FModel.getPreferences().getPref(FPref.DEV_LOG_ENTRY_TYPE))
+                            == GameLogVerbosity.CUSTOM;
+                }
+                @Override
+                public void select() {
+                    if (isEnabled()) {
+                        new CustomLogCategoriesDialog().show();
+                    }
+                }
+                @Override
+                public void drawPrefValue(Graphics g, FSkinFont font, FSkinColor color,
+                                          float x, float y, float w, float h) {
+                    String display = isEnabled()
+                        ? FModel.getPreferences().getCustomLogTypes().size()
+                          + "/" + GameLogEntryType.values().length
+                        : "N/A";
+                    g.drawText(display, font, color, x, y, w, h, false, Align.right, false);
+                }
+            }, 3);
         lstSettings.addItem(new BooleanSetting(FPref.LOAD_CARD_SCRIPTS_LAZILY,
             Forge.getLocalizer().getMessage("cbLoadCardsLazily"),
             Forge.getLocalizer().getMessage("nlLoadCardsLazily")), 3);
@@ -697,6 +725,7 @@ public class SettingsPage extends TabPage<SettingsScreen> {
 
         public abstract void select();
         public abstract void drawPrefValue(Graphics g, FSkinFont font, FSkinColor color, float x, float y, float width, float height);
+        public boolean isEnabled() { return true; }
     }
 
     private class BooleanSetting extends Setting {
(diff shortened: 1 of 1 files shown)


forge-gui/src/main/java/forge/localinstance/properties/ForgePreferences.java
```diff
@@ -338,6 +340,37 @@ public String getDefault() {
     /** Instantiates a ForgePreferences object. */
     public ForgePreferences() {
         super(ForgeConstants.MAIN_PREFS_FILE, FPref.class);
+        migrateLogVerbosity();
+    }
+
+    /** Migrate old GameLogEntryType values to GameLogVerbosity presets. */
+    private void migrateLogVerbosity() {
```
tool4ever, line 365: sorry, don't quite agree with this change:
power users should have fine-grained control (piloting different decks may make other information important)
if anything the preference could be enhanced so it's not that additive combo way but lets you toggle each type on its own (with revised beginner friendly default)

ping @tehdiplomat for additional input?

MostCromulent: No problem, I could revise so you've got a Low / Medium / High / Custom preference where selecting custom reveals an additional submenu with individual checkboxes for every reporting item so you have completely fine-grained control?

tool4ever: that sounds like the best of both words :)

forge-game/src/main/java/forge/game/GameLogEntry.java
```diff
@@ -2,16 +2,23 @@
...
     public final String message;
     public final GameLogEntryType type;
-    // might add here date and some other fields
+    public final transient CardView sourceCard;
 
     GameLogEntry(final GameLogEntryType type0, final String messageIn) {
+        this(type0, messageIn, null);
+    }
+
+    GameLogEntry(final GameLogEntryType type0, final String messageIn, final CardView sourceCard) {
```
Hanmac: While touching GameLogEntry, maybe change it to a `record`?

forge-gui/src/main/java/forge/localinstance/properties/ForgePreferences.java
MostCromulent: Done.

Screenshot 2026-02-23 200249

forge-game/src/main/java/forge/game/GameLog.java
```diff
@@ -68,7 +69,22 @@ public List<GameLogEntry> getLogEntries(final GameLogEntryType logLevel) { // nu
     
         for (int i = log.size() - 1; i >= 0; i--) {
             GameLogEntry le = log.get(i);
-            if (logLevel == null || le.type.compareTo(logLevel) <= 0) {
+            if (logLevel == null || le.type().compareTo(logLevel) <= 0) {
+                result.add(le);
+            }
+        }
```
Hanmac, line 75: @tool4ever we need to find a better way to handle these For loops
probably using a different List (LinkedList?)
And using Stream + Predicate Filter?

Also, the entries are stored in →, getting reversed ←, and after the function, they might be reversed again →
might not be the best way?

forge-gui/src/main/java/forge/localinstance/properties/ForgePreferences.java
paulsnoops: Good change but please can the capitalisation be consistent? 😅

Jetz72: Is there a good way to include the custom options in the preferences menus? I dunno if we have a checkbox multiselect preference for desktop or mobile/adventure.

tool4ever: imo it's acceptable if we don't try to squeeze everything in there
if the user wants to customize it chances are he's ingame already anyway

tool4ever: that said, it looks like the AI forgot to add such menu options for mobile?

forge-game/src/main/java/forge/game/GameLog.java
tool4ever: let's avoid adding more complexity via this PR
but the good news is VLog uses `Lists.reverse` from Guava
that method looks nice since it follows the View approach, so it should be cheap enough

but most of the code is stuck with `Collections.reverse` (used 45 times)
imo I could look for some more busy places to swap that out - performance should like that?

forge-gui/src/main/java/forge/localinstance/properties/ForgePreferences.java
Jetz72: >imo it's acceptable if we don't try to squeeze everything in there
>if the user wants to customize it chances are he's ingame already anyway

I don't think that's necessarily true. There's a good chance the user sees the options menu on the sidebar, finds the option there, and never realizes it can also be configured through the in-game menu. In that case, they'd see an option for "custom" but not have any idea how to actually customize it.

And yeah the mobile should have the options for customizing the custom mode somewhere. Doesn't necessarily need to be in-game. Bare minimum, could just make choosing "Custom" open a standard modal multi-select prompt, or add a button to open one.

tool4ever: well in that case maybe for this iteration just have the text explain to user where to look 🤷‍♂️
though admittedly the mobile approach has kept the ingame UI clean of such individual parts so far 🤔
(could also try to hide "Custom" in preferences but that might cause trouble with overwriting)
oh well guess either way there'll still be some more code needed 😅

MostCromulent: Alright, I have re-worked this so the custom settings are set through a popup dialog which is shared between the in-match UI and the preferences menu. The button is greyed out unless 'Custom' is selected.

The dialogue auto-populates based on available options rather than being hard-coded, so will automatically adapt if additional logging options are added. (e.g. https://github.com/Card-Forge/forge/pull/9845)

Screenshot 2026-02-24 064609
Screenshot 2026-02-24 064526
Screenshot 2026-02-24 064547

Also implemented for mobile:

Screenshot 2026-02-24 070027

Screenshot 2026-02-24 070107

tool4ever (approved):
another nice one :)

PR #9894, opened 2026-02-23
Fix arena item rewards always yielding the same item

vanja-ivancevic:
- Three reward branches in `RewardData.generate()` (`item`, `cardPackShop`, `landSketchbookShop`) bypassed the local `rewardRandom` and called `WorldSave.getWorld().getRandom()` directly
- For arena loot where `useSeedlessRandom=true`, this used the deterministic world RNG (re-seeded on every map load) instead of a fresh `Random` instance
- Capital arena item rewards resolved to the same item on every visit because the world random reached the same state each time
- Replace all three direct world-random calls with `rewardRandom`, matching every other branch in the method
- No behavioral change for shops (`useSeedlessRandom=false`), where `rewardRandom` already references the world random

(Bug reported by @CrystalCreation on 21.2.26 in the Old Border channel)

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/data/RewardData.java
+++ b/forge-gui-mobile/src/forge/adventure/data/RewardData.java
@@ -257,7 +257,7 @@ public class RewardData implements Serializable {
                 case "item":
                     if(itemNames!=null) {
                         for (int i = 0; i < count + addedCount; i++) {
-                            String itemName = itemNames[WorldSave.getCurrentSave().getWorld().getRandom().nextInt(itemNames.length)];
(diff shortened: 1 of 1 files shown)


vanja-ivancevic:
The `cardPackShop` and `landSketchbookShop` changes are no-ops for current callers. Shops pass `useSeedlessRandom=false`, so `rewardRandom` already points to the world random (same object reference, line 167). 

Included them so every branch in the switch consistently uses `rewardRandom` and the `useSeedlessRandom` flag remains the single point of control for RNG selection, matching the intent of the comment on line 168.

PR #9895, opened 2026-02-23
Fix block sound regression

tool4ever:


Hanmac:
i wonder why the Values looks like this to require that extra checks 🤔 
and if we should refactor/change them

tool4ever:
it looked like your Multimap refactor caused some slightly functional changes in areas which were using empty collections
 - which it doesn't support
in the end a further investigation didn't seem worth the time to fix it, but feel free to try

PR #9898, opened 2026-02-23
Fix Avatar: The Last Airbender rarity for common land

Simisays:
Found this while working on my Adventure cardprice rework. Coudn't understand why Misty Palms Oasis price wasn't calculated the way commons are supposed to be. Turns out the edition was faulty :)


paulsnoops:
@Hanmac changed these in December, I think maybe to work with the boosters?

Simisays:
After looking into it, I think they are still "common" as per game rules but in draft sets they have the basic land slot for draft balance. This is why common lands in Commander sets have C but these same lands have L in draft sets. 

It's no big deal though for adventure, i'll close this for now

PR #9913, opened 2026-02-25
Attack block req

### LINKED ISSUE #1424: Attack and Block Restrictions should use Static Abilities
* Block Restrictions in PhaseHandler should use Static Ability, the remaining blockers can be used as variable to the checks

* for Attack Restrictions, it also should use a Static and should not use AttackRestrictionType Enum.
Especially `AttackRestrictionType.NEED_BLACK_OR_GREEN` is problematic because of color text changes.

### LINKED ISSUE #6000: AttackRestriction should be moved to StaticAbilityCantAttackBlock
AttackRestriction needs to be reworked too, they should not rely on AttackRestrictionType enum.

- [ ] CARDNAME can only attack alone.
- [ ] CARDNAME can't attack unless a creature with greater power also attacks.
- [ ] CARDNAME can't attack unless a black or green creature also attacks. (especially this one, because Text Change)
- [ ] CARDNAME can't attack or block alone.
- [ ] CARDNAME can't attack unless at least two other creatures attack.

The parts that were using AttackRestriction/Type could use StaticAbility List instead.

Hanmac:
Closes #1424
Closes #6000

Still Draft/WIP

This is going to replace AttackRestrictionType in the future

right now, it renames:
* AttackRestrict to AttackRestrictNum
* BlockRestrict to BlockRestrictNum


Hanmac:
@tool4ever your opinion? does it go into the right direction?

forge-gui/res/cardsfolder/b/bright_palm_soul_awakener.txt
```diff
@@ -6,7 +6,7 @@ K:Backup:1:BackupAbility
 SVar:BackupAbility:DB$ Animate | Triggers$ AttackTrig | sVars$ AE
 SVar:AttackTrig:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigDoubleCounters | TriggerDescription$ Whenever this creature attacks, double the number of +1/+1 counters on target creature. That creature can't be blocked by creatures with power 2 or less this turn.
 SVar:TrigDoubleCounters:DB$ MultiplyCounter | ValidTgts$ Creature | CounterType$ P1P1 | SubAbility$ DBEffect
-SVar:DBEffect:DB$ Effect | StaticAbilities$ BlockRestrict | RememberObjects$ Targeted
+SVar:DBEffect:DB$ Effect | StaticAbilities$ BlockRestrictNum | RememberObjects$ Targeted
```
tool4ever: bad change

tool4ever:
maybe? the big challenge is adapting the logic in `AttackConstraints` and that's not been done yet

Hanmac:
> maybe? the big challenge is adapting the logic in `AttackConstraints` and that's not been done yet

i started with one of them today

tool4ever:
yes, I see it but I meant that one is also not complete yet

Hanmac:
First test with `Scarred Puma` seems to work, will do the rest later

tool4ever:
uhm what you need a review for - want a proof that all the code you just removed from `getLegalAttackers()` without replacement served a purpose...?

Hanmac:
> uhm what you need a review for - want a proof that all the code you just removed from `getLegalAttackers()` without replacement served a purpose...?

Maybe, i think the code needs to be refactored anyway.
But i have no idea right now where to start with this.

tool4ever:
Imo the problem is this:
right now the logic was stripped down to only handle their direct attack restriction part
but since these also interact with other attackers you also need to consider if they have any requirements

> 508.1d. The active player checks each creature they control to see whether it's affected by any requirements (effects that say a creature attacks if able, or that it attacks if some condition is met). If the number of requirements that are being obeyed is fewer than the maximum possible number of requirements that could be obeyed without disobeying any restrictions, the declaration of attackers is illegal...

or in other words:
in addition to the (obvious) answer "currently not possible" we also need "could be possible"

Hanmac:
@Agetian probably needs to help me with the Logic for this

Agetian:
I might be able to take a look into this, hopefully closer to the weekend :)

Agetian:
I agree that this needs the proper AI handling before merge though, currently it's missing quite a good deal of restriction considerations on the AI part. I'll see what I can do, not sure how well I can figure it out though, seems like a pretty good chunk of code that was rewired ^^;

Hanmac:
> I might be able to take a look into this, hopefully closer to the weekend :)

It's more about the piece that @tool4ever was talking about, finding out if there is a legal way for them to attack or not

Hanmac:
@MostCromulent this MR also might need some AI help

especially the Check with "could be possible"
(that's the part to check if there is a different Attack configuration that would satisfy more attack requirements without breaking attack requirements)

making TestCases for this might also help too

MostCromulent:
> this MR also might need some AI help

@Hanmac here's very high-level AI options outline based on review of the PR. Let me know if there's something in here you would like AI to flesh out in more detail for possible implementation.

**[pr9913-attack-block-req-options.md](https://github.com/user-attachments/files/27709062/pr9913-attack-block-req-options.md)**

PR #9916, opened 2026-02-25
Allow commander to be any creature if not enforcing deck conformity

dkrako:
Deactivating deck conformity check did not have any effect on the available commanders.

This change allows any creature to be set as commander.


Jetz72:
Huh. Guess I added this option for partner commanders but not for the main one? Seems valid to me.

dkrako:
Thanks for the feedback. Are there any tests that I should add?

I didn't check if the issue also exists on the desktop app.

PR #9919, opened 2026-02-25
Add Max Deck Count to config.json

WSpadaro:
Allow setting a custom max deck count from the config.json. This will still be bound between 20 and 99. Defaults to 20 if value is missing.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
+++ b/forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
@@ -569,11 +569,8 @@ public class AdventurePlayer implements Serializable, SaveFileContent {
             }
         }
 
-        // Load Max Deck Count and grab it from Config file, use the larger one
-        if (data.containsKey("maxDeckCount"))
-            this.maxDeckCount = data.readInt("maxDeckCount");
-        // Sanity Check Grab the larger of the save or config setting, then make sure the number is not insane and make sure it is at least 20
-        this.maxDeckCount = Math.max(Math.min(Math.max(this.maxDeckCount, Config.instance().getConfigData().maxNumberOfDecks), 99), 20);
+        // Set max deck count to either the value in the config or the current player deck count and then ensure it is bound by 20 and 99
+        this.maxDeckCount = Math.min(Math.max(Math.max(data.containsKey("deckCount") ? data.readInt("deckCount") : 20, Config.instance().getConfigData().maxNumberOfDecks), 20), 99);
 
 
         // Load decks
@@ -754,7 +751,6 @@ public class AdventurePlayer implements Serializable, SaveFileContent {
         data.store("lifeLoss", this.difficultyData.lifeLoss);
         data.store("spawnRank", this.difficultyData.spawnRank);
         data.store("rewardMaxFactor", this.difficultyData.rewardMaxFactor);
-        data.store("maxDeckCount", this.maxDeckCount);
 
         data.store("name", name);
         data.store("heroRace", heroRace);


forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
```diff
@@ -103,6 +103,8 @@ public PlayerStatistic getStatistic() {
 
     public int getDeckCount() { return decks.size(); }
 
+    public int getMAX_DECK_COUNT() { return MAX_DECK_COUNT; }
```
Jetz72: I know we don't have a style guide but ALL_CAPS_WITH_UNDERSCORES is generally the style for a static constant. If we're changing that we should change the style of the name to reflect that it's a local variable. And `getMAX_DECK_COUNT` is right out.

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
```diff
@@ -562,6 +569,15 @@ public void load(SaveFileData data) {
             }
         }
 
+        // Load Max Deck Count or grab it from Config file if missing from save
+        if (data.containsKey("maxDeckCount"))
+            this.MAX_DECK_COUNT = data.readInt("maxDeckCount");
+        else
+            this.MAX_DECK_COUNT = Config.instance().getConfigData().maxNumberOfDecks;
+        // Sanity Check make sure the number is not insane and make sure it is at least 20
+        this.MAX_DECK_COUNT = Math.max(Math.min(this.MAX_DECK_COUNT, 99), 20);
```
Jetz72: I know it solves the problem of not purging decks if the config change gets unceremoniously reset by an update but I can't shake the feeling that storing the limit in the save file is a bit awkward.

Perhaps the max deck count could be set using something like `min(max(max(20, config's maxNumberOfDecks), save data's deckCount), 99)`? Uses whatever's currently in the file or whatever number of decks you currently have, whichever is higher. With hard bounds of 20-99.

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
WSpadaro: Yeah, it actually irked me leaving it as CAPS and not switching to camelCase. When adding it to config I named it: maxNumberOfDecks. Not sure why I was hesitant to rename the existing variable. Will get that swapped.

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
WSpadaro: After some thought I realized, besides letting players override max deck limit, this also allows each plane to set their own max deck count in their config. Would there ever be a reason that a creator of a plane would want to limit the player to less than 20 decks? I.E. Would someone ever make a plane where they would want to force the player to use only 1 deck (or any number less than 20)?

Jetz72: Probably not since you can just export the deck and import a new one, or in the worst case remake it manually. It wouldn't present a challenge so much as an annoyance. The setting would be a knob to that can be turned for technical purposes, not game-design purposes.

The only exception I can think of might be if we had an Adventure plane or mode with deckbuilding rules based on [The Beleaguered Boxer](https://scryfall.com/card/unk/RZ16/the-beleaguered-boxer), but that'd require so many other changes under the hood to work.

tool4ever: I think this does make sense as an optional gameplay mechanic available to plane creators:
e.g. backpack item rewards to increase your deck limit
preparing for a larger dungeon and having to decide which decks to carry for switching while out of town would be rewarding strategic planning 🧙
but unless I'm mistaken for that to work it would first need to be split into a stash + carry limit?

Jetz72: We've considered locking in decks while in dungeons in the past. I always envisioned it as a single deck for the whole dungeon plus the ability to swap cards using the deck's sideboard. If we allowed multiple decks in the dungeon or while outside towns though, I think it'd still make sense to let them build the normal amount in town and just lock in a few in areas when the restriction is in place.

WSpadaro: As far as the change you suggested. It changes the behavior from:
You can change max deck count when starting new save or new game plus -> You can change max deck count when starting a new save or new game plus OR you can change change it whenever you want but only if you are increasing it and not decreasing it.
Makes it slightly more usable at the cost of simplicity imo. It does not matter to me. I'll make that change and get it checked in.

Jetz72: What I meant was leaving the `maxDeckCount` out of the save entirely, and instead when you load the save it uses `dynamicDeckCount` as a lower bound on the limit (in conjunction with 20). In other words: you can change the max deck count whenever you want, but it won't go lower than the number of decks you have until you delete those decks and restart.

WSpadaro: That makes sense to use dynamicDeckCount as the lower bound to ensure no decks are lost upon config change.

I added it to the save because I saw that the other other config options used in AdventurePlayer, the difficulty settings, are initially loaded from config.json but then stored and loaded from the save. The commenter's rational was:
//Previously these were not being read from or written to save files, causing defaults to appear after reload

This holds true for max deck count, so I followed the same behavior. Without saving it to the save file, the defaults will appear after reload (following an update). It makes sense to me to keep it in the save for QoL sake, not having to go back into the config after each update but if it is not wanted in the save, I will remove it.

Jetz72: The config file isn't intended as a secondary options menu. I don't think we want to make a habit of preserving user-edits to it. If we did, I think we'd instead do it by way of supporting config overrides from a file in the user's custom content directory.

WSpadaro: maxDeckCount has been removed from the save. It is now only based on the greater of the config file or current deck count and then clamped by 20 and 99.

PR #9923, opened 2026-02-26
Deck importer improvements - Commander detection heuristic / misc fixes

MostCromulent:
## Summary

Several improvements to the deck importer:

1. **Commander detection heuristics:** Two-tier heuristic for automatically detecting and assigning to Commander format. 
   - When Commander is auto-detected, the options panel is expanded and the format dropdown is automatically enabled and set to Commander so the detection is visible and the **user can choose to override it**. 
   - When accepting a deck with Commander format from a non-Commander editor, automatically switch to the Commander editor.
   - Auto-detect criteria:
        - Explicit headers  (`Commander`, `//Commander`, `CM:`) - auto-detect as Commander
        - Headerless heuristic (98-102 singleton cards + blank-line separator) - identify commander from cards after the last blank line (common Commander export format in plain text e.g. Moxfield), validated with `DeckSection.Commander.validate()`. The blank-line heuristic guards against false positives on other 100-card singleton formats (e.g. Canadian Highlander) by requiring the trailing card(s) to be valid commander candidates (legendary creatures, eligible planeswalkers).

3. **Companion section**: Map `Companion` header to Sideboard (Forge auto-detects companions at game start). Without this, companion cards fall into the preceding section (typically Commander), creating a bogus second commander.

4. **Card name regex**: Added `?` to support cards like `Continue?`

5. **Foil/etched markers**: Strip Moxfield `*F*`/`*E*` suffixes so cards are recognized

6.  **Deck name with space separator**: Accept `Name Turtle Power!` format

## Known limitations

- **Mobile**: The shared controller correctly identifies commanders and builds the deck with a Commander section, but the mobile import dialog (`FDeckImportDialog`) does not have format selection UI or editor switching. Commander cards are silently dropped when importing into a Constructed editor on mobile. This is a pre-existing limitation, not a regression.

---

### DIFF
--- a/forge-core/src/main/java/forge/deck/DeckRecognizer.java
+++ b/forge-core/src/main/java/forge/deck/DeckRecognizer.java
@@ -490,8 +490,6 @@ public class DeckRecognizer {
             "avatar", "commander", "companion", "schemes",
             "conspiracy", "planes", "deck", "dungeon",
             "attractions", "contraptions"};
-    private static final CharSequence[] IGNORABLE_SECTION_NAMES = {
-            "maybeboard", "maybe", "considering", "tokens"};
 
     private static CharSequence[] allCardTypes(){
         List<String> cardTypesList = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


tool4ever:
does this need to be revived or just not worth it?

MostCromulent:
imo the worthwhile ideas from this one are incorporated here while dropping the more marginal stuff:

https://github.com/Card-Forge/forge/pull/10517

PR #9935, opened 2026-02-26
Add network integration test to CI

MostCromulent:
## Summary

Cherry-picked from #9642 and adapted for the master network protocol following discussion with @tool4ever.

- `TestTrueNetworkTraffic` runs an AI vs AI game over the real TCP network stack using minimal 10-card land-only decks
- AI takes turns playing lands only and game ends due to decking out.
- Catches serialization failures, protocol mismatches, pipeline exceptions, and silent client disconnections
- Runs in CI without `-Drun.stress.tests=true` — adds ~1 second to the test suite

This PR also adds send error tracking to `RemoteClient`, allowing the test to detect Netty pipeline exceptions (e.g. `EncoderException`) that were previously caught and swallowed silently. The test asserts zero send errors after game completion.

## Note on master at time of this PR

At the time of this PR, master has a race condition described in #9911 — the event batching flush runs on the EDT while the game thread continues modifying shared objects, causing `ArrayIndexOutOfBoundsException` during serialization. 

This test _**will intentionally fail**_ on that code, consistently producing 24-37 send errors per game in local testing.

Once #9911 is merged, the test passes with zero send errors. This has been verified locally by temporarily applying the #9911 fix and confirming the test passes cleanly. The test should catch similar regressions in future.

---

### DIFF
--- a/forge-gui-desktop/src/test/java/forge/net/TestTrueNetworkTraffic.java
+++ b/forge-gui-desktop/src/test/java/forge/net/TestTrueNetworkTraffic.java
@@ -43,15 +43,13 @@ public class TestTrueNetworkTraffic {
             if (!(GuiBase.getInterface() instanceof HeadlessGuiDesktop)) {
                 GuiBase.setInterface(new HeadlessGuiDesktop());
             }
-            if (FModel.getPreferences() == null) {
-                FModel.initialize(null, preferences -> {
-                    preferences.setPref(FPref.LOAD_CARD_SCRIPTS_LAZILY, false);
-                    preferences.setPref(FPref.UI_LANGUAGE, "en-US");
-                    preferences.setPref(FPref.ENFORCE_DECK_LEGALITY, false);
-                    FModel.getNetPreferences().setPref(ForgeNetPreferences.FNetPref.UPnP, "NEVER");
-                    return null;
-                });
-            }
+            FModel.initialize(null, preferences -> {
+                preferences.setPref(FPref.LOAD_CARD_SCRIPTS_LAZILY, false);
+                preferences.setPref(FPref.UI_LANGUAGE, "en-US");
+                preferences.setPref(FPref.ENFORCE_DECK_LEGALITY, false);
+                FModel.getNetPreferences().setPref(ForgeNetPreferences.FNetPref.UPnP, "NEVER");
+                return null;
+            });
 
             initialized = true;
         }


forge-gui-desktop/src/test/java/forge/net/TestTrueNetworkTraffic.java
```diff
@@ -0,0 +1,261 @@
...
+            Assert.assertNotNull(game, "Game should exist after startGame");
+
+            // 10. Convert remote player to AI on server side
+            // Find the player whose GUI is NetGuiGame (the remote slot)
+            for (Player p : game.getPlayers()) {
+                IGuiGame gui = hostedMatch.getGuiForPlayer(p);
+                if (gui instanceof NetGuiGame) {
+                    System.out.println("[TestTrueNetworkTraffic] Converting remote player '" + p.getName() + "' to AI");
+
+                    // Clear InputQueue BEFORE swapping controller to unblock game thread
```
tool4ever: shouldn't we just reuse and test `FServerManager.convertToAI` here?
also the `clearInputs()` order is different here - that needs analysis

forge-gui-desktop/src/test/java/forge/net/TestTrueNetworkTraffic.java
```diff
@@ -0,0 +1,261 @@
...
+                FModel.initialize(null, preferences -> {
+                    preferences.setPref(FPref.LOAD_CARD_SCRIPTS_LAZILY, false);
+                    preferences.setPref(FPref.UI_LANGUAGE, "en-US");
+                    return null;
+                });
+            }
+
+            // Set UPnP to NEVER to avoid dialog on server start
+            originalUPnP = FModel.getNetPreferences().getPref(ForgeNetPreferences.FNetPref.UPnP);
+            FModel.getNetPreferences().setPref(ForgeNetPreferences.FNetPref.UPnP, "NEVER");
```
tool4ever: why do we end up needing manual resets instead of passing them like above via `initialize`?
seems unjustified to me...

forge-gui-desktop/src/test/java/forge/net/TestTrueNetworkTraffic.java
```diff
@@ -0,0 +1,215 @@
...
+            initialized = true;
+        }
+    }
+
+    @Test(timeOut = 60000, description = "True network traffic test with remote client")
+    public void testTrueNetworkTraffic() throws Exception {
+        System.out.println("[TestTrueNetworkTraffic] Starting true network traffic test...");
+
+        // Temporarily disable deck legality for 10-card decks
+        originalLegality = FModel.getPreferences().getPrefBoolean(FPref.ENFORCE_DECK_LEGALITY);
```
tool4ever: what about this preference, move it too?
if we ever need a (net) test for deck legality putting that into its own class should be enough to handle both?

tool4ever (approved):

PR #9950, opened 2026-02-28
Add precon starter deck selection mode

vanja-ivancevic:
- Add Precon game mode that lets players browse and select from ~107 preconstructed decks at game start (for old border)
- Decks are grouped by edition (Mirage, Tempest, Urza's Saga, etc.) with an "All Editions" option
- Repurposes existing color/edition UI selectors as deck/set pickers when Precon mode is active
- Feature is gated on the presence of a `decks/starter/precon/ `folder, so vanilla and other mods are unaffected
- Standardize precon deck filenames to `SET - Deck Name.dck `format for edition auto-detection

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
+++ b/forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
@@ -299,6 +299,10 @@ public class AdventurePlayer implements Serializable, SaveFileContent {
         return adventureMode;
     }
 
+    public boolean isCommanderMode() {
+        return adventureMode != null && adventureMode.isCommanderLike();
+    }
+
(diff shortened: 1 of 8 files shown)


vanja-ivancevic:
Added a follow-up for Commander support:

- new `Commander Precon` Adventure mode
- uses `decks/starter/commanderprecon/` with the same picker flow as `Precon`
- runs under Commander rules, so duel/editor behavior stays correct
- plane folder takes priority, with fallback to common if missing or empty

Agetian:
This is great! I actually was about to start my own implementation and even had a rudimentary prototype, when I realized that you already completed this and even submitted it. No idea how this slipped my attention before, lol.

PR #9952, opened 2026-02-28
Fix Specialize card image download from Scryfall

MostCromulent:
Cherry-picked from https://github.com/Card-Forge/forge/pull/9806 as requested by @tool4ever.

## The Bug

Two issues prevented Specialize face images (e.g. HBG Alchemy cards) from downloading from Scryfall:

1. **Wrong image key format:** `Card.setImageKey(IPaperCard, CardStateName)` used `getCardXSpecImageKey()` which returns a file-path key (`HBG/Lae'zel, Blessed Warrior.full`). The image pipeline expects a lookup key in `c:Name|SET|art$wspec` format — without the `c:` prefix and `$wspec` suffix, `ImageFetcher` can't detect the specialize face or resolve the PaperCard. `CardFactory` already does this correctly; `setImageKey` just wasn't consistent with it.

2. **Missing Scryfall collector number suffix:** `getScryfallDownloadUrl()` had no handling for `CardSplitType.Specialize`. Scryfall uses collector numbers with a color letter suffix (e.g. `2w`, `2u`, `2b`, `2r`, `2g`) but the code never appended it.

## Testing

Tested locally with Lae'zel, Githyanki Warrior (HBG) — deleted cached images, specialized into white, confirmed the image auto-downloads from Scryfall with the correct `2w` collector number.

---

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/ImageCache.java
+++ b/forge-gui-desktop/src/main/java/forge/ImageCache.java
@@ -32,7 +32,6 @@ import java.util.concurrent.TimeUnit;
 
 import javax.imageio.ImageIO;
 
-import forge.item.IPaperCard;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 
@@ -47,7 +46,7 @@ import forge.game.card.CardView;
 import forge.game.player.PlayerView;
 import forge.gui.FThreads;
 import forge.gui.GuiBase;
-import forge.item.PaperCard;
+import forge.item.IPaperCard;
 import forge.item.InventoryItem;
 import forge.localinstance.properties.ForgeConstants;
 import forge.localinstance.properties.ForgePreferences;


forge-core/src/main/java/forge/util/ImageUtil.java
```diff
@@ -284,6 +291,17 @@ public static String getScryfallTokenDownloadUrl(String collectorNumber, String
                 langCode, versionParam, faceParam);
     }
 
+    private static String specFaceToCollectorSuffix(String face) {
```
tool4ever, line 294: @Hanmac
can MagicColor be used for this?

Hanmac: Could it theory 🤔

MostCromulent: Have updated and confirmed this works.

Hanmac: @tool4ever probably meant the enum

Hanmac (dismissed):

Hanmac:
@tool4ever good for now,
Image keys need to be reworked later anyway

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -6360,19 +6360,19 @@ public final void setImageKey(final IPaperCard ipc, final CardStateName stateNam
             return;
         switch (stateName) {
             case SpecializeB:
-                setImageKey(ipc.getCardBSpecImageKey());
+                setImageKey(ipc.getImageKey(false) + ImageKeys.SPECFACE_B);
```
tool4ever: this somehow keeps the old image when a card specializes but only on mobile
for reference here's the old PR that added this method: #3097

tool4ever: there must be something wrong between what each GUI expects from the ImageKey? 🤔
- AI can try to figure out if a smaller cleanup fix (but no GUI branching) can resolve this
- but if it's a deeper problem it should be solved via issue from @Hanmac and just revert this part for now :/

MostCromulent: Have found the issue - desktop ImageCache.java at line 213 onwards has a specific fix for the specialisation cards which is missing in the mobile implementation. AI duplicated that code in mobile, tested, and it updates correctly.
I'll commit a centralised GUI-agnostic fix shortly.

MostCromulent: Actually, given both desktop and mobile already have their own separate image cache implementations is the correct answer to just duplicate the fix on mobile? I'll commit what I've got now and you can have a look. Centralising this would require a bigger refactor of the existing desktop implementation.

> The centralized method collapses to:
> 
> ```java
> // In ImageUtil (forge-core)
> public static String resolveCardImageKey(String imageKey) {
>     String face = "";
>     if (imageKey.endsWith(ImageKeys.BACKFACE_POSTFIX)) {
>         face = "back";
>         imageKey = imageKey.substring(0, imageKey.length() - ImageKeys.BACKFACE_POSTFIX.length());
>     } else if (imageKey.endsWith(ImageKeys.SPECFACE_W)) {
>         face = "white";
>         imageKey = imageKey.substring(0, imageKey.length() - ImageKeys.SPECFACE_W.length());
>     } // ... etc for U, B, R, G
> 
>     PaperCard card = getPaperCardFromImageKey(imageKey);
>     if (card == null) return null;
>     return getImageKey(card, face, true);
> }
> ```
> 
> **Complications by site:**
> 
> | Site | Can use `resolveCardImageKey` directly? | Issue |
> |---|---|---|
> | Mobile `getImage()` | Yes | Clean replacement |
> | Mobile `imageKeyFileExists()` | Yes | Clean replacement |
> | Desktop `ImageCache` | Mostly | Also needs `ipc` (PaperCard) for art crop logic at line 262 |
> | Desktop `FImageUtil` | No | Passes raw key WITH suffix to `getPaperCardFromImageKey` (looks buggy), plus unique post-processing at lines 119-125 |
> 
> The desktop `ImageCache` needs both the resolved path AND the `ipc` PaperCard. FImageUtil is a separate beast - it looks like it may have a latent bug (passing unsanitized keys to the DB lookup).

tool4ever: imo we can proceed with the current version for now even if it's somewhat messy
from my testing this means we can instead get rid of the method linked to this comment, so that would be a good trade-off

MostCromulent: AI suggested that method is still used elsewhere and couldn't be cleanly removed:

> There are still active callers in 4 files beyond the definitions:
> 
> | File | Usage |
> |---|---|
> | `forge-gui/ImageFetcher.java` | Lines 181-193 - used when **downloading** specialize images |
> | `forge-gui-mobile/ImageCache.java` | Lines 228-240 and 300-312 - the code **we just added** in this PR |
> | `forge-gui-desktop/FImageUtil.java` | Lines 100-112 - desktop image **rendering** |
> | `forge-gui-desktop/ImageCache.java` | Lines 236-248 - desktop image **cache lookup** |
> 
> The mobile `ImageCache` calls are ones we introduced in this PR to fix the mobile display. But there are 3 other call sites (ImageFetcher, desktop FImageUtil, desktop ImageCache) that are **not part of our PR** and still rely on these methods.
> 
> Removing the methods would require refactoring all of those callers to use the suffix-based approach instead - which significantly expands the scope of this PR.

However it then found a way to significantly consolidate, see latest commit:

> Every call site was doing the same thing:
> 
> 1. Decode the suffix to a color name (`"white"`, `"blue"`, etc.)
> 2. Use that color name to pick which `getCardXSpecImageKey()` to call
> 3. That method internally just calls `ImageUtil.getImageKey(card, "white", true)` (or whichever color)
> 
> So we were translating suffix -> color string -> method call -> same color string -> `ImageUtil.getImageKey`. The middle step was pointless. Now we just go suffix -> color string -> `ImageUtil.getImageKey` directly.
> 
> That let us delete the 5 per-color methods from the interface, from `PaperCard` (including 5 cached fields), and from `PaperToken` (5 stubs that returned the base image anyway). About 130 lines gone, replaced by one-liners at each call site.

Have tested and seems to work fine on both desktop and mobile.

tool4ever: I actually only meant this `setImageKey` overload but think seems to work out even better :)
if the caching was even worth it for such exotic mechanic we can always do it in a cleaner way later

tool4ever (dismissed):

Hanmac (dismissed):

forge-gui-desktop/src/main/java/forge/ImageCache.java
```diff
@@ -207,7 +207,7 @@ private static Pair<BufferedImage, Boolean> getOriginalImageInternal(String imag
             return Pair.of(null, false);
         }
 
-        IPaperCard ipc = null;
+        PaperCard ipc = null;
```
Hanmac: Why the type change?

Hanmac:
Closes #7532

PR #9953, opened 2026-02-28
Allow for quick building decks for limited UI

tehdiplomat:


tool4ever:
something wrong with the language files?

PR #9962, opened 2026-03-01
Limit AI to cast Green Sun's Zenith for all its mana

### LINKED ISSUE #6720: AI casts Green Sun's Zenith for all its mana
**Describe the bug**
AI casts Green Sun's Zenith for all its mana even if its green creatures have much lower mana values

**To Reproduce**
1. The highest mana value of an AI's creature is 4.
2. AI casts Green Sun's Zenith for all its mana (for example, x=5)

**Expected behavior**
AI should cast Green Sun's Zenith with an X value equal to the mana value of the green creature it's looking for, or 

**Additional context**
At least, AI should cast Green Sun's Zenith with an X value that doesn't increase the maximum mana value of green creatures in its deck.

**Desktop**
OS: Windows 10
Java Version: JDK 17.0.13
Forge Version: 2.0.01-SNAPSHOT-12.23

**Screenshots**

zen1
zen2

kojotak:
An attempt to fix #6720 by allowing to limit the maximum mana cost X. It works for instant/sorcery, but doesn't work for a creature (`Rocco, Cabaretti Caterer`). Is there a better way, for example using `SVar:AIPreference:AIMaxTgtCost`?

An example of Green sun's zenith cast with X=2, which is the max CMC of all creatures in a library:

image

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
@@ -660,15 +660,13 @@ public class ComputerUtilCost {
 
             if (sa.hasParam("AIMaxTgtCost")) {
                 String value = sa.getParam("AIMaxTgtCost");
-                if ("MaxCMC".equals(value) && sa.hasParam("ChangeType") && sa.hasParam("Origin")) {
-                    String filter = sa.getParam("ChangeType");
-                    if (filter.contains("cmcLEX")) {
-                        filter = filter.replace("X", "" + val);
-                    }
-                    Collection<Card> originCards = ai.getCardsIn(ZoneType.valueOf(sa.getParam("Origin")));
-                    Collection<Card> filteredCards = CardLists.getValidCards(originCards, filter, ai, source, sa);
-                    val = ObjectUtils.min(val, filteredCards.stream().mapToInt(Card::getCMC).max().orElse(val));
+                String svar = source.getSVar(value);
+                if (svar.contains("YouOwn")) {
+                    // limit the CMC to available mana
+                    svar = svar.replace("YouOwn", "YouOwn+cmcLE" + val);
                 }
+                int calculated = AbilityUtils.calculateAmount(source, svar, sa);
+                val = ObjectUtils.min(val, calculated);
             }
         }
 
--- a/forge-gui/res/cardsfolder/f/finale_of_devastation.txt
+++ b/forge-gui/res/cardsfolder/f/finale_of_devastation.txt
@@ -1,7 +1,8 @@
 Name:Finale of Devastation
 ManaCost:X G G
 Types:Sorcery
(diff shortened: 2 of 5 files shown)


forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
```diff
@@ -657,6 +657,19 @@ public static int getMaxXValue(SpellAbility sa, Player ai, final boolean effect)
 
         if (root.costHasManaX()) {
             val = ComputerUtilMana.determineLeftoverMana(root, ai, effect);
+
+            if (sa.hasParam("AIMaxTgtCost")) {
```
tool4ever, line 661: too unflexible / think bigger - just allow scripters to provide the maximum value for good measure:
`SVar:Y:Count$ValidLibrary Creature.YouOwn$GreatestCardManaCost`

forge-gui/res/cardsfolder/f/finale_of_devastation.txt
```diff
@@ -1,7 +1,7 @@
 Name:Finale of Devastation
 ManaCost:X G G
 Types:Sorcery
-A:SP$ ChangeZone | ChangeType$ Creature.YouCtrl+cmcLEX | Hidden$ True | Origin$ Library | OriginAlternative$ Graveyard | Destination$ Battlefield | ShuffleNonMandatory$ True | SubAbility$ DBPump | SpellDescription$ Search your library and/or graveyard for a creature card with mana value X or less and put it onto the battlefield. If you search your library this way, shuffle.
+A:SP$ ChangeZone | ChangeType$ Creature.YouCtrl+cmcLEX | Hidden$ True | Origin$ Library | OriginAlternative$ Graveyard | Destination$ Battlefield | ShuffleNonMandatory$ True | SubAbility$ DBPump | AIMaxTgtCost$ MaxCMC | SpellDescription$ Search your library and/or graveyard for a creature card with mana value X or less and put it onto the battlefield. If you search your library this way, shuffle.
```
tool4ever: this one should probably consider an additional SVar in case you'd get really close to 10 and the secondary effect is worth it

kojotak:
This will take me more time than expected  🙈 "Much to learn, I have" 

I still can not figure it out how to properly script the condition of limiting the library search by available mana - so there is an ugly `if(svar) replace` hack in https://github.com/Card-Forge/forge/pull/9962/changes/66adbd5648c06b1c0b6a79f5f07255a9f571c059 . And speaking of `Finale of Devastation` - I have yet to debug it why ai does not consider the pump effect at all (it casts it in main2 without any combat advantage).

tool4ever:
hmn, this looked ok now mostly:
- you don't have to try and inject the mana there - if it's less than the maximum ChangeZoneAi logic should just take over to see if there's still something fetchable like in other cases too
- probably just missing PlayMain1 SVar

kojotak:
> if it's less than the maximum ChangeZoneAi logic should just take over to see if there's still something fetchable like in other cases too

Yes, I agree, but this will fix the issue #6720 only if available mana is > most expensive creature in library. That's the situation from the screenshot of this PR (available mana = 5, max payable CMC = 2, X is set to 2 ✅). When the library contains creature with higher CMC (for example Apex devastor CMC=10), then ComputerUtilCost resolves X to 5, all available mana is used and creature with only CMC = 2 is put into battlefield. 

> * probably just missing PlayMain1 SVar

Thanks, I've finally found that too, but without luck. AI is still casting it in main2 despite having mana and multiple creatures in play (which can benefit from the +X/+X).

I don't give up. Yet 😉

Issue #9721, opened 2026-02-11
FR: Option to not stack/overlap cards in hand


GamesByFelix:
**Is your feature request related to a problem? Please describe.**
I play the desktop version on a big screen and would like to always see the full cards in my hand. However depending on the amount of cards, the GUI sometimes stacks two cards on top of each other, obscuring the lower one.

Here's a screenshot of the undesired effect:
<img height="400" alt="Image" src="https://github.com/user-attachments/assets/8ad05e4f-ad64-453e-8b8d-ccb00e3137bc" />

**Describe the solution you'd like**
I'd like an option to always scale the cards in a way that they are fully visible and don't overlap. It would be okay for me to have scrollbar, if it's not possible to scale all cards inside the view properly.

**Describe alternatives you've considered**
Resizing the hand view sometimes fixes this, but it's not a great workaround.

Issue #9861, opened 2026-02-21
Network multiplayer development meta issue


MostCromulent:
Adding this as a public meta list for anyone else who wants to contribute to network play development.

### Potential New Features
- [ ] Configurable player turn timer; priority passes if turn time limit exceeded (potentially redundant now we have AFK timer)
- [ ] Add support for draft and sealed modes.
- [ ] Add support for a public server browser, rather than connecting by direct IP address only??
- [ ] Connection quality indicator?
- [ ] Integrate chat into desktop UI rather than being a floating window (relatively simple on match screen, would require redesign of lobby screen - if done, should be optional as some users prefer floating window) - (dockable windows already implemented in https://github.com/Card-Forge/forge/pull/9969), can adopt same method)

### Fixes / QoL improvements
- [ ] Purge hidden information from sending to clients (need to reuse some kind of placeholder CardView?)
should further reduce traffic especially for FullStatePacket + in the early game when most cards are still in libraries
also reduces cheating potential (though not that much of a priority)
- [ ] Fix https://github.com/Card-Forge/forge/pull/9802 regression - reconnected players will not see the gamelog but did prior to this change
- [ ] Expand AFK timer to cover other `sendAndWait()` AFK situations which are currently not null safe
- [ ] Add red/amber styling for server chat warning messages: e.g. UPNP failure / disconnect / AFK timeout warning.

## PR IN PROGRESS:
- [ ] Delta serialization rather than sending entire game state every update - merged but not yet enabled - https://github.com/Card-Forge/forge/pull/9642
- [ ] Improved yield options to speed up 3+ player games - https://github.com/Card-Forge/forge/pull/9643
- [ ] Misc UI improvements - https://github.com/Card-Forge/forge/pull/9899

## COMPLETE:
- [X] Improved server messages in chat window - https://github.com/Card-Forge/forge/pull/9660
- [X] Version mismatch warning in chat - https://github.com/Card-Forge/forge/pull/9698
- [X] Reconnect support - https://github.com/Card-Forge/forge/pull/9708
- [X] Sort player fields in turn order in desktop UI - https://github.com/Card-Forge/forge/pull/9723
- [X] Clients process events based on GameEvent forwarding rather than protocol methods - https://github.com/Card-Forge/forge/pull/9760
- [X] Fix phase stop preferences in multiplayer - https://github.com/Card-Forge/forge/pull/9820
[...]

PR #9969, opened 2026-03-02
Improved card zone view - dockable windows, zone banners, user prefs, hand UI options

MostCromulent:
## Summary

Improved zone view functionality including dockable floating windows, user-friendly banners to identify zone origin of cards, better detection logic for playable zone cards, and improved hand display options.

## Features

**1. Dockable floating zone windows:** Zone windows opened from buttons under the character portrait can be dragged and docked in UI panels. Once docked, they can split or move between panels just like any other tab.

**2. Improved "Playable Zone Cards" window:** renamed version of the "Flashback" window which includes user-friendly banners identifying the zone origin of each playable non-hand card. Also improved detection logic for which cards appear in this window, including static abilities like on _Squee, The Immortal_

https://github.com/user-attachments/assets/cfeddf25-da0d-4d5d-a3a7-f542956529fb

**3. User preference for opening zones in tabbed or windowed mode:** right-clicking the zone button under character portrait presents a radial menu to select whether to open in window or tab by default when pressing button. Preference is set per-zone and persists across sessions. Saves you needing to dock window every time if you're always intending to use it in tabbed mode.

Screenshot 2026-03-07 190103

**4. Improved hand appearance options:** Hand Panel options in the Layout menu including front-ending the existing 'order by CMC option', disabling overlapping cards in hand, and fixing number of cards per row allowing hand to spill into multiple rows with vertical scrollbar if necessary. Incidentally addresses feature request in https://github.com/Card-Forge/forge/issues/9721

(This is really an artifact of the earlier iteration of this PR which included zone cards in the hand UI; could be cherrypicked to a separate PR but it's already implemented here and I think still a worthy UI improvement).

Screenshot 2026-03-07 190039

**Default**
Screenshot 2026-03-07 191054

**Default + Prevent Card Overlap**
Screenshot 2026-03-07 191112

**Max cards per row: 7**
Screenshot 2026-03-07 191151

---

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/menus/LayoutMenu.java
+++ b/forge-gui-desktop/src/main/java/forge/menus/LayoutMenu.java
@@ -238,8 +238,8 @@ public final class LayoutMenu {
         final FScreen screen = Singletons.getControl().getCurrentScreen();
         if (screen != null && screen.isMatchScreen()) {
             final IVTopLevelUI view = screen.getView();
-            if (view instanceof VMatchUI) {
-                for (final VHand h : ((VMatchUI) view).getControl().getHandViews()) {
+            if (view instanceof VMatchUI vmu) {
+                for (final VHand h : vmu.getControl().getHandViews()) {
                     h.getLayoutControl().updateHand();
                 }
             }
--- a/forge-gui-desktop/src/main/java/forge/screens/match/controllers/CHand.java
+++ b/forge-gui-desktop/src/main/java/forge/screens/match/controllers/CHand.java
@@ -116,21 +116,16 @@ public class CHand implements ICDoc {
 
         synchronized (ordering) {
             ordering.clear();
-
-            final boolean orderByCmc = FModel.getPreferences().getPrefBoolean(FPref.UI_ORDER_HAND);
+            ordering.addAll(cards);
 
             // Sort hand by CMC/color at the UI layer. This duplicates the game-layer sort in
             // PlayerZone.onChanged(), but is necessary because network clients only have CardViews
             // (no access to the game model), and because toggling the preference mid-game needs
             // to take effect immediately without waiting for a zone change event.
-            if (orderByCmc) {
-                final List<CardView> sorted = new ArrayList<>(cards);
-                sorted.sort(Comparator.comparingInt((CardView cv) -> cv.getCurrentState().getManaCost().getCMC())
+            if (FModel.getPreferences().getPrefBoolean(FPref.UI_ORDER_HAND)) {
+                ordering.sort(Comparator.comparingInt((CardView cv) -> cv.getCurrentState().getManaCost().getCMC())
                         .thenComparing(cv -> cv.getCurrentState().getColors().getOrderWeight())
                         .thenComparing(cv -> cv.getCurrentState().getName()));
-                ordering.addAll(sorted);
-            } else {
-                ordering.addAll(cards);
             }
         }
 
@@ -143,8 +138,7 @@ public class CHand implements ICDoc {
                 cardPanel = new CardPanel(matchUI, card);
                 cardPanel.setDisplayEnabled(false);
                 placeholders.add(cardPanel);
-            }
-            else {
+            } else {
                 cardPanel.setCard(card); //ensure card view is updated
             }
             cardPanels.add(cardPanel);


Jetz72:
Doesn't the UI have a dedicated "Flashback Zone" that's meant to serve this purpose exactly? I dunno if it's as broadly functional as it should be, though.

MostCromulent:
There's currently this button which opens a floating window:

Screenshot 2026-03-03 063112

But IMHO the implementation in this PR is more intuitive:
- Keeps all the playable cards immediately visible in the same place.
- The labelling clearly identifies which zone the card is being played from (Flashback window doesn't visibly distinguish between cards in exile vs graveyard for example).
- I find new players in particular often don't realise the zone labels underneath the player portrait are actually clickable buttons which open floating windows, they look like baked-in UI elements.

In any event, if there's some duplication with Flashback window, this is optional and I would have thought pretty harmless?

Jetz72:
I do like your approach aesthetically. However, when working on a project the age and size of Forge, a new feature that aims to serve the same purpose as an existing (albeit flawed) one makes the codebase, the UI, and the overall project design messier. Sometimes it's fine to keep both, other times the new one should completely supplant the old, other times still it'd be better to work potential improvements into the groundwork that's already established. I'm not yet sure which category this should fall into, but I'm skeptical that it's the first one.

> I find new players in particular often don't realise the zone labels underneath the player portrait are actually clickable buttons which open floating windows, they look like baked-in UI elements.

This seems like a separate UI design problem to solve rather than a reason to work around those buttons. Being able to view cards in various zones is a fairly fundamental game function the player should be familiar with.

MostCromulent:
Understand. Let me know which direction you want me to go once you've had a chance to fully consider it.

(**Edit:** If I had to choose between them I'd rather have this and get rid of the flashback window - but I would say that! 😅)

tool4ever:
> This seems like a separate UI design problem to solve rather than a reason to work around those buttons. Being able to view cards in various zones is a fairly fundamental game function the player should be familiar with.

I'm also leaning in this direction. Rationale:
- I don't like creating this kind of "superzone" which could usually require multiple view adjustments across longer games to keep it from becoming too flooded (even if they're reached more quickly from menu)
- I worry about causing "preference fatigue" for newer users based on our already mostly growing amount and then introducing options that overlap complexity with the existing GUI on top

Onboarding experience for the current approach seems easy enough to improve here:
1. find better name instead of Flashback so collision with the Keyword doesn't cause wrong expectations
2. add section to User Guide explaining virtual zone concept
3. teach usage in the "tutorial puzzle" (long term goal) :)

That said, I understand opening/closing the floating window could also be considered annoying.
*If possible* I'd rather see a solution where the dynamic behaviour is extended, i.e. convert from floating to tabbed to arrange it like you want:
- this would also turn it into a multi-purpose feature (e.g. when playing with "top of library revealed" cards you could then do the same with your library) 💡
- I'm not sure how easy the Layout classes could deal with that though - if it becomes too messy maybe introducing some kind of "Wildcard" tab for holding floating windows would be a better scoped alternative? 🤔

And I also like the visuals - can keep these either way. 👍

Jetz72:
> If possible I'd rather see a solution where the dynamic behaviour is extended, i.e. convert from floating to tabbed to arrange it like you want

Saw a feature request the other day for the ability to dock the floating windows as tabs/panels in the main layout. Frankly the prospect of digging into Swing to that extent terrifies me but maybe the AI would be good at it?

> find better name instead of Flashback so collision with the Keyword doesn't cause wrong expectations

"Playable"? "May-Play"? 

I dunno if the current Flashback zone covers all zones, but if not, expanding that functionality is a given. I also like the visual groupings by actual zone. 

One question to address is whether it should only show things that are truly playable in the current state, or all things that an effect is granting permission to cast. So for instance, if you're allowed to play lands from your graveyard but you've already played your land for turn, should it still show lands? Same thing if you've impulse-drawn an instant when your opponent has a card that prevents instants from being played. I'd lean toward showing it anyway, but it might depend on the intention here.

MostCromulent:
> Saw a feature request the other day for the ability to dock the floating windows as tabs/panels in the main layout. 

I saw that as well and thought it might be a good middle ground. I'll get the AI to look into how complex that would be to implement. 

> One question to address is whether it should only show things that are truly playable in the current state, or all things that an effect is granting permission to cast. 

My strong view is it should operate on same principle as the hand UI: show everything thats capable of being cast and it's then up to the player to work out whether it can actually be played with the given game state. 

- It's more confusing to have a UI zone where cards appear and disappear based on evaluation of whether they are actually playable (possibly several times within the same turn), when this isn't how any of the hand UI or other zone UI windows work. 

- I'm pretty sure implementing this correctly would require solving the "identify playable cards to player" logic with a high degree of accuracy. That sort of logic is being worked on in a few other PRs including advanced yield options and that one with the playable card highlighting, but IMHO it would be a significant complexity overlay to include just for this UI change.

MostCromulent:
**Dockable zone windows are now implemented.** 

AI did a much better job on this out-of-the-gate than I expected. It is surprisingly smooth:

- Floating zone windows can now be dragged and dropped as tabs on main UI panels.
- Once docked, you can re-arrange and split these tabs like any other part of the desktop layout
- Remove a docked zone window by right clicking tab and selecting "undock", this will restore it to floating window.
- Have renamed "Flashback" to "Playable Zone Cards" (this seems intuitive but open to suggestions on something better)
- Have also updated the logic for that zone so it captures static effects that make cards castable like _Squee, The Immortal_.
- Have completely removed initial implementation of having playable zone cards in main hand UI - they are now only accessed through either the "Playable Zone Cards" window or one of the zone-specific windows.
- When clicking zone button under character portrait to show/hide zone window, it will remember whether it was in docked or windowed mode prior to being hidden.

https://github.com/user-attachments/assets/28cada61-4c6a-4507-adda-b2a62a8de9e6

(Video is more stuttery than it is in reality).

MostCromulent:
Potential improvement: add right click option on zone buttons below player portrait to choose between opening window or opening tab?

tool4ever:
yea I guess why not, there's nothing there right now anyway

seems to work pretty good indeed :)

forge-gui/res/languages/en-US.properties
```diff
@@ -2882,6 +2882,10 @@ lblDisconnect=Disconnect
 lblCardName=Card Name
 lblPowerOrToughness=Power/Toughness
 lblAbilityIcon=Ability Icon
+lblOrderHandByCMC=Order Hand by CMC and Color
```
tool4ever: avoid duplicating:
this is already "nlOrderHand"

forge-gui-desktop/src/main/java/forge/view/arcane/HandArea.java
```diff
@@ -48,6 +50,18 @@ public HandArea(final CMatchUI matchUI, final FScrollPane scrollPane) {
 
         this.setDragEnabled(true);
         this.setVertical(true);
+        try {
+            final int maxCards = Integer.parseInt(FModel.getPreferences().getPref(FPref.UI_HAND_MAX_CARDS_PER_ROW));
```
tool4ever: use `getPrefInt` to avoid additional error handling here

MostCromulent:
Future expansion opportunity based on dockable window functionality (not for this PR): 

I previously suggested in https://github.com/Card-Forge/forge/issues/9861 making the online chat a UI panel rather a floating window. tool4ever pointed out some prefer to keep window. This provides a consistent way to do best of both worlds.

PR #7981, opened 2025-07-09
Fix Meld cards fetch

Eradev:
Fixes #7979 and minor fixes to other sets.

PR #9972, opened 2026-03-02
Fix Meld cards coll number

### LINKED ISSUE #9970: Forge doesn't download images for "Second parts" of the meld cards
Meld as a mechanic requires 2 cards, who together transform in the single powerful cards. 
Each meld pair has two parts, main one with meld condition(Urza, Gisela, Hanweir Battlements) and the second one with the reminder text "melds with X" (Bruna, powerstones, hanweir garrison)
Forge downloads images for "melded" cards and first part of the meld pair, however it doesn't dowload images for the cards with reminder text from the second part

1. Be sure to not have images downloaded prior
2. Find any "second part of the meld pair" card
3. See error

Forge should download images for them just like for any other cards 

 - OS: windows
 - Browser chrome
 - Version 2.0.11 snapshot 03.01

Both cards in hand, yet forge doesn't download image for one of them:
Image

Eradev:
Fixes #9970

### DIFF
--- a/forge-gui/res/editions/Eldritch Moon.txt
+++ b/forge-gui/res/editions/Eldritch Moon.txt
@@ -28,7 +28,7 @@ TokenFallbackCode=SOI
 12 C Wretched Gryff @Darek Zabrocki
 13 U Blessed Alliance @Johann Bodin
 14 C Borrowed Grace @Volkan Baǵa
-15a R Bruna, the Fading Light @Clint Cearley
+15 R Bruna, the Fading Light @Clint Cearley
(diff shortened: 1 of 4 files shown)


Hanmac:
Do we need to revert some other changes made by #7981?

Eradev:
> Do we need to revert some other changes made by #7981?

No. It seems they only changed some IDs this time.

Hanmac:
https://scryfall.com/card/sir/115/graf-rats page is broken for scryfall, but that's not our fault 😅

PR #9978, opened 2026-03-03
Added Psychic Frog AI logic for blocking and attacking

NathanRoy1:
Implements SpecialCardAi.PsychicFrog to handle both activated abilities of Psychic Frog.
Ability 1 — Discard a card: put a +1/+1 counter on Psychic Frog

Activates during combat when the Frog is blocking and the counter would prevent it from dying or allow it to kill an attacker it otherwise couldn't
Activates proactively at end of turn when 2+ safe discard candidates are available, or when hand size exceeds the maximum hand size limit

Ability 2 — Exile three cards from graveyard: Psychic Frog gains flying until end of turn
Activates on opponent's turn to intercept unblocked flyers, preferring blocks where the Frog survives and kills the attacker
Will use any graveyard cards  if life is in danger
Activates on the AI's turn to attack unblocked when the opponent has no flying/reach blockers, with a higher priority score when the attack is lethal

### DIFF
--- a/docs/User-Guide.md
+++ b/docs/User-Guide.md
@@ -237,11 +237,14 @@ Underneath each player's battlefield portrait are zone buttons (hand, library, g
 **Display modes:** Right-click a zone button to toggle how it opens:
 
 - **Open in Window** — Opens a floating window that can be moved and resized independently. Right-click the window's title bar to toggle alphabetical sorting.
-- **Add Tab to Hand Panel** — Docks the zone as a tab alongside your hand panel, integrated into the main layout. The tab can be dragged to other cells like any other panel.
+- **Add Tab to Hand Panel** — Docks the zone as a tab alongside your hand panel, integrated into the main layout.
 
 Your preference is saved per zone and remembered separately for your own zones and opponent zones.
 
 **Drag-to-dock:** A floating zone window can also be docked by dragging its title bar over any cell's tab bar — a blue highlight shows where it will land. Once docked, right-click the tab for an option to undock it back to a floating window.
+
+> [!TIP]
+> Enable the Layout->View->New card count menu option to display a delta number in the tab header which may help you more quickly identify changes in complex boardstates.
 
 ## Auto-Sort Multiplayer Fields
 In multiplayer games (3+ players), Forge can automatically arrange opponent battlefield panels in turn order. This is controlled via the **Layout > View** menu during a match.
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
@@ -657,7 +657,7 @@ public class ComputerUtilCost {
 
         if (root.costHasManaX()) {
             val = ComputerUtilMana.determineLeftoverMana(root, ai, effect);
-
+            // TODO already calculate once before previous method to avoid running more expensive checks above this highest possible maximum
             if (sa.hasParam("AIMaxTgtCost")) {
                 String value = sa.getParam("AIMaxTgtCost");
                 String svar = source.getSVar(value);
--- a/forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
@@ -344,10 +344,8 @@ public class ChangeZoneAi extends SpellAbilityAi {
 
         String type = sa.getParam("ChangeType");
         if (type != null && type.contains("X") && sa.getSVar("X").equals("Count$xPaid")) {
-            // Set PayX here to maximum value.
             final int xPay = ComputerUtilCost.getMaxXValue(sa, ai, sa.isTrigger());
             sa.setXManaCostPaid(xPay);
-            type = type.replace("X", Integer.toString(xPay));
         }
 
         for (final Player p : pDefined) {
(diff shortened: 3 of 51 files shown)


forge-gui/res/cardsfolder/p/psychic_frog.txt
```diff
@@ -4,7 +4,7 @@ Types:Creature Frog
 PT:1/2
 T:Mode$ DamageDone | ValidSource$ Card.Self | ValidTarget$ Player,Planeswalker | CombatDamage$ True | Execute$ TrigDraw | TriggerDescription$ Whenever CARDNAME deals combat damage to a player or planeswalker, draw a card.
 SVar:TrigDraw:DB$ Draw | Defined$ You | NumCards$ 1
-A:AB$ PutCounter | Cost$ Discard<1/Card> | CounterType$ P1P1 | CounterNum$ 1 | SpellDescription$ Put a +1/+1 counter on CARDNAME.
-A:AB$ Pump | Cost$ ExileFromGrave<3/Card> | Defined$ Self | KW$ Flying | SpellDescription$ CARDNAME gains flying until end of turn.
+A:AB$ PutCounter | Cost$ Discard<1/Card> | CounterType$ P1P1 | CounterNum$ 1 | AILogic$ PsychicFrog | SpellDescription$ Put a +1/+1 counter on CARDNAME.
```
tool4ever, line 7: hi, thanks for contributing

imo these abilities aren't really unusual enough to warrant their own logic:
- technically Pump/PutCounter should both already have some logic for combat tricks that could be improved (so more cards can benefit)
- so it's probably more about when paying those cost parts is worth it, for that check `ComputerUtil.getCardPreference` for more flexible approach

NathanRoy1: If I understand correctly, it’s not that you think the logic isn’t worth implementing. Rather, you’d prefer to integrate this behavior directly into PumpAi and PutCounterAi so that a larger range of cards can benefit from the improved logic.

Given that, what would you prefer for this PR? Should it be closed for now or would you like this implementation to remain as a temporary solution until the more generalized logic is added?

tool4ever: Well if you want to rework it feel free to just add further commits

But if you'd rather start fresh later closing might be better

NathanRoy1: I should be able to rework it in a short delay so I'll leave it open for now

Agetian: I actually kind of like this "isolated special logic" approach, at least until this can be generalized/rewritten (SpecialCardAi was created with that in mind - a place to put specialized/tricky logic until it can be made generic in the proper API-based AIs). If this can't be rewritten in a reasonable time frame to be generic, I vote for integrating this version for the time being, as it seems to perform well. The only thing I'd also vote for is that it's cleaned up a bit if possible (it reformats some stuff and adds some extraneous empty lines in some places unrelated to the PR - I'll mark those places separately in the review)

forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
```diff
@@ -302,8 +302,11 @@ protected AiAbilityDecision checkApiLogic(Player ai, final SpellAbility sa) {
             return doChargeToOppCtrlCMCLogic(ai, sa);
         } else if (logic.equals("TheOneRing")) {
             return SpecialCardAi.TheOneRing.consider(ai, sa);
+        }else if ("PsychicFrog".equals(logic)) {
+            return SpecialCardAi.PsychicFrog.considerCounterAbility(ai, sa , ph);
         }
 
+
```
Agetian: Unneeded extra empty line

forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
```diff
@@ -302,8 +302,11 @@ protected AiAbilityDecision checkApiLogic(Player ai, final SpellAbility sa) {
             return doChargeToOppCtrlCMCLogic(ai, sa);
         } else if (logic.equals("TheOneRing")) {
             return SpecialCardAi.TheOneRing.consider(ai, sa);
+        }else if ("PsychicFrog".equals(logic)) {
+            return SpecialCardAi.PsychicFrog.considerCounterAbility(ai, sa , ph);
```
Agetian: ai, sa, ph
(without the extra space after "sa")

forge-ai/src/main/java/forge/ai/ability/PumpAi.java
```diff
@@ -61,7 +61,7 @@ protected boolean checkAiLogic(final Player ai, final SpellAbility sa, final Str
 
     @Override
     protected boolean checkPhaseRestrictions(final Player ai, final SpellAbility sa, final PhaseHandler ph,
-            final String logic) {
+                                             final String logic) {
```
Agetian, line 64: Substandard style (we usually don't format it like that)

forge-ai/src/main/java/forge/ai/ability/PumpAi.java
```diff
@@ -347,7 +349,7 @@ protected AiAbilityDecision checkApiLogic(Player ai, SpellAbility sa) {
     }
 
     private boolean pumpTgtAI(final Player ai, final SpellAbility sa, final int defense, final int attack, final boolean mandatory,
-    		boolean immediately) {
+                              boolean immediately) {
```
Agetian, line 352: Substandard style (we usually don't format it like that)

Agetian:
It also looks like the compile is failing due to a missing import in SpecialCardAi?

Agetian:
> It also looks like the compile is failing due to a missing import in SpecialCardAi?

Fixed this. Should now compile.

PR #9992, opened 2026-03-05
CardFactoryUtil: use Adventure Image

Hanmac:
this uses the "On an Adventure" Helper Image for the Effect in Command

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/EffectEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/EffectEffect.java
@@ -121,7 +121,11 @@ public class EffectEffect extends SpellAbilityEffect {
 
         String name = sa.getParam("Name");
         if (name == null) {
-            name = hostCard + (sa.hasParam("Boon") ? "'s Boon" : "'s Effect");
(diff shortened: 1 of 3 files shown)


forge-game/src/main/java/forge/game/ability/effects/EffectEffect.java
```diff
@@ -121,7 +121,11 @@ public void resolve(SpellAbility sa) {
 
         String name = sa.getParam("Name");
         if (name == null) {
-            name = hostCard + (sa.hasParam("Boon") ? "'s Boon" : "'s Effect");
+            if (sa.hasParam("Adventure")) {
```
tool4ever, line 124: can we check for the CardState instead of extra param?

Hanmac: Doesn't work for cards that do its own effects too
Like:
> Exile the top two cards of your library. Until your next end step, you may play those cards.

I want the image only for the Adventure part

Jetz72:
I think Plot also has a marker we could add. Companion has one too but I don't think that one has accurate text on it...

Issue #8770, opened 2025-09-24
No MapMulti on Android


Hanmac:
https://github.com/Card-Forge/forge/blob/ab3b00cb656a85881c0a72beae5496bb134a0590/forge-gui/src/main/java/forge/gamemodes/planarconquest/ConquestUtil.java#L151

```
NoSuchMethodError: No interface method mapMulti(Ljava/util/function/BiConsumer;)Ljava/util/stream/Stream; in class Ljava/util/stream/Stream; or its super classes (declaration of 'java.util.stream.Stream' appears in /system/framework/core-oj.jar) at forge.gamemodes.planarconquest.ConquestUtil.getBasicLandSets(ConquestUtil.java:151)
```

PR #9997, opened 2026-03-06
CostRemoveAnyCounter: cleanup description

Hanmac:
* remove `YouCtrl` from the Filter, the List is already from the payer.
* some Stream counting for `getMaxAmountX`
* use `payCostFromSource` that can be later extended for #7401 
* try to generate the Desc from Type in case of simple `Creature` or simple double type `Creature;Planeswalker`

### DIFF
--- a/forge-game/src/main/java/forge/game/cost/CostRemoveAnyCounter.java
+++ b/forge-game/src/main/java/forge/game/cost/CostRemoveAnyCounter.java
@@ -80,13 +80,10 @@ public class CostRemoveAnyCounter extends CostPart {
         if (this.counter != null) {
             return validCards.stream().mapToInt(c -> c.canRemoveCounters(this.counter) ? c.getCounters(this.counter) : 0).sum();
         }
-        return validCards.stream().mapMultiToInt((c, s) -> {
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/cost/CostRemoveAnyCounter.java
```diff
@@ -65,24 +71,22 @@ public CostRemoveAnyCounter(final String amount, final CounterType counter, fina
...
+        CardCollectionView validCards;
+        if (payCostFromSource()) {
+            validCards = new CardCollection(source);
+        } else {
+            validCards = CardLists.getValidCards(payer.getCardsIn(ZoneType.Battlefield), this.getType().split(";"), payer, source, ability);
+        }
+        if (this.counter != null) {
+            return validCards.stream().mapToInt(c -> c.canRemoveCounters(this.counter) ? c.getCounters(this.counter) : 0).sum();
+        }
+        return validCards.stream().mapMultiToInt((c, s) -> {
```
tool4ever: seems unsafe? #8770 :/

Hanmac: Ugh, I will think about it 🤔

Issue #9985, opened 2026-03-04
Commander deck "Toothy, Imaginary Friend" has > 1 copy of "Pir, Imaginative Rascal"


rpherbig:
**Describe the bug**
When I try to make a Commander game with the deck "Toothy, Imaginary Friend"  selected, I get an error (see screenshot).

**To Reproduce**
1. Go to Hme
2. Select format check box for "Commander"
3. Select the AI player
4. Give them the deck "Toothy, Imaginary Friend"
5. Hit Start

**Expected behavior**
The game starts

**Screenshots**

<img width="2560" height="1400" alt="Image" src="https://github.com/user-attachments/assets/0d487cee-ca30-4b34-a3f1-aca0c96a065c" />

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: n/a
 - Version: 2.0.10-SNAPSHOT-02.17

**Smartphone (please complete the following information):**
n/a

**Additional context**
n/a

kojotak:
~I have found the bug, but I don't know how to fix it 🙈~ . The problem seems to be in DeckgenUtil.java:720 in

```
                if (partners.size() > 0) {
                    selectedPartner = partners.get(MyRandom.getRandom().nextInt(partners.size()));
                    preSelectedCards.removeAll(StaticData.instance().getCommonCards().getAllCards(selectedPartner.getName()));
                }
```

The `selectedPartner` is `Pir, Imaginative Rascal` from `BBD` edition and `N.A.` collectorNumber. But on the next line, the `getAllCards` finds two Pirs - one from `SLD` with collectorNumber `1050`, second from `BBD` with collectorNumber `11`. Here lies the problem - the `preSelectedCards` removes nothing, since the `equals` is comparing collectorNumbers. The result is, that `Pir` is added into deck twice: first time as a partner and second time into main deck.

rpherbig:
@kojotak I just ran into a similar error with the Commander deck "Ludevic, Necro-Alchemist": the problematic card was "Kydele, Chosen of Kruphix".

Will your PR also fix that one?

kojotak:
@rpherbig yes, I'm able to start a commander game with generated deck for `Ludevic, Necro-Alchemist`:

<img width="742" height="543" alt="Image" src="https://github.com/user-attachments/assets/e304aa05-0ece-44ef-84cb-4542db753f0c" />

Still not sure whether #10000 fixes the cause or the effect.

rpherbig:
[...]

PR #10000, opened 2026-03-06
Fixed duplicates in commander generated decks

kojotak:
Possible fix for #9985 

I don't know why, but `Pir, Imaginative Rascal` loaded from `CardRelationMatrixGenerator.cardPools` has collectorNumner `N.A.`. That is not the case if loading from `StaticData.instance().getCommonCards()` (this returns 11 and 1050). The `N.A.` collectorNumber results in `equals` method returning false in an attempt to remove this card from the deck, causing `Pir, Imaginative Rascal` beeing duplicated (in main deck and as a commander partner).

With this change, the card generated deck for `Toothy, Imaginary Friend` can be loaded:

image


Agetian (approved):

Hanmac:
@Agetian might need to look into why it got the wrong CN?

Agetian:
Yeah, not sure why it doesn't return the proper CN :/ A deeper look into this might be needed but not sure who would be best suited for the job?

kojotak:
I went a little bit deeper. `PaperCard` without the CN is loaded from the `deckgendecks/Commander.dat` file, not from the `res/editions/*.txt` files. You can check that using breakpoint in `CardThemedMatrixIO#loadMatrix` inspecting the matrix hashMap. Don't know, yet, why the serialized card is without CN nor how to repair (rebuild) the dat file.

Agetian:
Oh, hmm, I'll see if I can find an opportunity to regenerate that LDA file , along with the two other ones that desperately need updating.

kojotak:
Shouldn't the `Commander.dat` file contain just card names?  The problem will arise again when some card will have different name/edition/collector number/flags/foil/artIndex (according to `equals` method). The `PaperCard` instance (stored in `CardRelationMateixGenerator.cardPools` ) is used only for `DeckFormat.isLegalCard()` check, but this can be delayed after loading the whole card from `StaticData` . The `generateRandomCommanderDeck()` method would be a little bit slower in exchange for smaller memory footprint of `cardPools` (and dat files).

Agetian:
It's a LDA model for building Commander decks based on training data, it should be regularly updated, ideally, but I'm not sure how and who did it before (I do know how the other LDA models were generated and I update those from time to time). The card information should be a separate thing, yes, and I don't understand why CN is inferred specifically from Commander.dat and not just from the card DB :/

rpherbig:
any updates on this?

tool4ever:
the data has been regenerated
but we should still check again

jamincollins:
## Findings

#### 1. Potential duplicate card instances from getCard vs direct reference
Description:
The PR changes `preSelectedCards.add(pair.getKey())` to `preSelectedCards.add(StaticData.instance().getCommonCards().getCard(pair.getKey().getName()))` in generateRandomCommanderDeck for matrix-based commander deck generation. The change is intended to avoid duplicates caused by the CardRelationMatrixGenerator.cardPools containing PaperCard instances that may be different object references than those in CommonCards, leading to duplicate logical cards after removal operations. However, `getCard(String)` returns a PaperCard instance based on current art preference and edition filtering, which may be a different object than the one from the matrix pool. The preSelectedCards list is later filtered by `preSelectedCards.removeAll(StaticData.instance().getCommonCards().getAllCards(selectedPartner))` which relies on equals/hashCode semantics. If PaperCard.equals is based on identity or mutable state, using getCard may introduce inconsistency where removal of partner cards may fail to remove all duplicates, or may remove more than intended.

The code later does `preSelectedCards.removeAll(StaticData.instance().getCommonCards().getAllCards(commander))` at line 720, which uses getAllCards returning multiple prints of the same card name. Mixing instances from cardPools vs CommonCards could lead to asymmetric equals behavior.

Test case:
```java
// Simulating the scenario
PaperCard matrixCard = CardRelationMatrixGenerator.cardPools.get("Commander").get(commanderName).get(0).getKey();
PaperCard commonCard = StaticData.instance().getCommonCards().getCard(matrixCard.getName());

List<PaperCard> preSelected = new ArrayList<>();
preSelected.add(matrixCard);
preSelected.add(commonCard);

// If equals is based on name, both are considered same; if based on identity, they are different
boolean same = matrixCard.equals(commonCard);
boolean removed = preSelected.removeAll(StaticData.instance().getCommonCards().getAllCards(matrixCard.getName()));
// removal may leave matrixCard if equals is identity-based
```

Impact: Deck generation may still contain duplicates or fail to remove partner/commander cards consistently, leading to illegal decks.

#### 2. Null return from getCard for unknown names
Description:
`StaticData.instance().getCommonCards().getCard(pair.getKey().getName())` is called without null check. If the matrix pool contains a PaperCard whose name is not present in CommonCards (e.g., deprecated card, name mismatch, or card removed from DB), `getCard` returns null per CardDb.tryGetCard logic. Adding null to preSelectedCards will cause NPE later when `format.isLegalCard(pair.getKey())` is already checked, but `preSelectedCards` now contains null which later iterates in `preSelectedCards.removeAll(...)` and `for (PaperCard c : preSelectedCards)` loops, causing NullPointerException.

The original code added `pair.getKey()` directly, which was guaranteed non-null from the pool. The new code introduces a lookup failure path.

Test case:
```java
PaperCard matrixCard = new PaperCard("Nonexistent Card");
PaperCard commonCard = StaticData.instance().getCommonCards().getCard(matrixCard.getName());
assert commonCard == null;
// Adding to preSelectedCards leads to NPE later in limitCardsToCommanderBracket or removal loops
```

Impact: Crash during deck generation for commanders with stale matrix entries.

#### 3. getCard returns different print with different flags/art
Description:
`getCard(name)` uses default art preference and may return a PaperCard with different artIndex or flags than the matrix entry. Subsequent code `preSelectedCards.removeAll(StaticData.instance().getCommonCards().getAllCards(selectedPartner))` uses `getAllCards(name)` which returns all prints. If the matrix entry had a specific print with a particular artIndex or set code, the replacement card may have different legality characteristics (e.g., different edition legality, different card rules if reprint changed rules text). Format legality check `format.isLegalCard(pair.getKey())` is performed on original matrix card, but the added card is from CommonCards which may have different legality.

Test case:
```java
PaperCard matrixCard = ...; // from pool, edition = "M10"
PaperCard commonCard = StaticData.instance().getCommonCards().getCard(matrixCard.getName()); // returns "2XM" print
boolean legalMatrix = format.isLegalCard(matrixCard); // true
boolean legalCommon = format.isLegalCard(commonCard); // may be false if 2XM not legal in format
```

Impact: Illegal cards may be added to preSelectedCards, violating format legality.

#### 4. Loss of weight information from CardRelationMatrixGenerator
Description:
The matrix pool entries are `Map.Entry<PaperCard, Integer>` where Integer is connection count used for weighting. The PR changes addition to use `getCard(name)` which discards the weight. The code previously shuffled the list but did not use weights; however, the change may affect future weighting logic if prepareWeightedRandomizedCardPool is added. More importantly, using getCard may return a PaperCard instance that is shared across multiple references, causing unintended mutation side-effects if PaperCard is mutable.

Impact: Potential shared mutable state across decks, leading to cross-deck contamination.

#### 5. Duplicate removal via removeAll relies on equals consistency
Description:
Later code does `preSelectedCards.removeAll(StaticData.instance().getCommonCards().getAllCards(selectedPartner))` and `preSelectedCards.removeAll(StaticData.instance().getCommonCards().getAllCards(commander))`. If preSelectedCards contains instances from CommonCards via getCard, removal will work via equals. However, if some entries remain from the Brawl branch where `getUniqueByName` is used, those instances may be different objects with same name but different identity. Mixing instance sources creates inconsistent removal behavior.

Test case:
```java
List<PaperCard> list = new ArrayList<>();
list.add(StaticData.instance().getCommonCards().getCard("Lightning Bolt"));
list.add(StaticData.instance().getCommonCards().getUniqueByName("Lightning Bolt"));
Collection<PaperCard> toRemove = StaticData.instance().getCommonCards().getAllCards("Lightning Bolt");
list.removeAll(toRemove); // may leave one instance if equals is identity
```

Impact: Commander card may remain in preSelectedCards, resulting in duplicate commander in deck.

### Performance and security
#### 6. Repeated getCard lookups cause repeated DB scans
Description:
For each potential card in the matrix pool (potentially hundreds per commander), the code calls `StaticData.instance().getCommonCards().getCard(name)`. This performs a full DB lookup via CardRequest parsing and edition scanning. For multiple commander generations, this causes repeated scans of the card DB, increasing latency. Caching the mapping from name to PaperCard would be more efficient.

Test case:
```java
for (int i=0;i<1000;i++) {
    StaticData.instance().getCommonCards().getCard("Lightning Bolt");
}
// Duration grows linearly with lookups
```

Impact: Deck generation performance degradation, especially for AI opponents generating many decks.

## Verdict
Request Changes
The change addresses duplicate instances but introduces null safety risks, potential legality mismatches, and performance overhead. The root cause of duplicates should be addressed by ensuring consistent PaperCard identity across the system or by using name-based deduplication rather than mixing instance sources. The current fix is fragile and may cause NPEs or illegal decks.

## Recommended Fixes
1. Add null check after getCard and skip null entries to avoid NPE.
2. Verify that getCard returns a card with same legality as matrix entry; optionally compare edition legality before adding.
3. Consider using getUniqueByName for consistency with Brawl branch, or normalize all entries to name-based deduplication via Set<String>.
4. Cache getCard results in a Map<String, PaperCard> to avoid repeated DB lookups.
5. Ensure PaperCard.equals/hashCode is based on name+set+artIndex to make removeAll work reliably across instances.

---

PR #10004, opened 2026-03-07
Add network diagnostics logging and migrate to tinylog

MostCromulent:
## Summary

Existing logging uses `System.out.println` (no timestamps) and provided no performance metrics, making it impossible to identify bottlenecks from log files alone.

Migrates all network `System.out`/`System.err` logging to tinylog (`Logger`) and adds targeted performance diagnostics with thresholds to avoid log noise during normal play.

### 1. tinylog Migration (all network classes)
Every `System.out.println` and `System.err.println` call in the network package is replaced with the appropriate `Logger.info`, `Logger.warn`, or `Logger.error` call. This ensures every log line gets a timestamp, log level, and thread name — critical for correlating events between server and client logs.

**Files:** `GameProtocolHandler`, `FGameClient`, `FServerManager`, `GameClientHandler`, `ProtocolMethod`, `TrackableTypes`

### 2. Protocol Processing Duration (`GameProtocolHandler`)
Measures wall-clock time for `method.invoke()` execution and EDT queue delay (time between message arrival on the IO thread and actual processing on the EDT). Logged only when either exceeds 50 ms — a threshold chosen as the point where per-message delay becomes perceptible. Normal protocol calls complete in single-digit milliseconds, so this fires only on genuine bottlenecks.

### 3. IO-Thread `beforeCall()` Timing (`GameProtocolHandler`)
Measures time spent in `beforeCall()`, which runs on Netty's single IO thread and includes `updateTrackers()` and `replicateProps()`. Logged only above 50 ms. This is the most likely cause of the reported multi-second delays — `beforeCall()` blocks ALL subsequent network messages, so if it's slow, every message in the pipeline backs up. Normal execution is sub-millisecond.

### 4. `send()` Duration (`RemoteClient`)
Measures how long `writeAndFlush().sync()` blocks the calling thread. Logged only above 50 ms. Detects TCP back-pressure or slow client acknowledgment causing the server game thread to stall. Under normal network conditions this completes near-instantly.

### 5. Serialization Size Logging (`CompatibleObjectEncoder` / `CompatibleObjectDecoder`)
Logs compressed frame size and decode duration for messages exceeding 20 KB compressed. Normal `setGameView` packets are ~8–15 KB compressed in 2-player games, so the 20 KB threshold fires selectively — primarily on 4-player games (~15–25 KB compressed) and anomalously large 2-player states. Decode duration is also logged above 50 ms to catch deserialization bottlenecks.

### 6. Event Type Identification (`NetGuiGame`)
Logs the class names of game events in each batch sent to the client. Correlates server-side event batches with client-side processing delays to identify which game actions trigger the heaviest network traffic.

---

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/net/server/NetGuiGame.java
+++ b/forge-gui/src/main/java/forge/gamemodes/net/server/NetGuiGame.java
@@ -322,10 +322,8 @@ public class NetGuiGame extends AbstractGuiGame {
     @Override
     public void handleGameEvents(List<GameEvent> events) {
         if (paused) { return; }
-        String eventTypes = events.stream()
-                .map(e -> e.getClass().getSimpleName())
-                .collect(Collectors.joining(", "));
-        Logger.info("Sending batch of {}: [{}]", events.size(), eventTypes);
+        Logger.info("Sending batch of {}: [{}]", () -> events.size(),
+                () -> events.stream().map(e -> e.getClass().getSimpleName()).collect(Collectors.joining(", ")));
         sender.write(ProtocolMethod.setGameView, getGameView());
         sender.send(ProtocolMethod.handleGameEvents, events);
     }


forge-gui/src/main/java/forge/gamemodes/net/server/NetGuiGame.java
```diff
@@ -321,7 +322,10 @@ public void handleGameEvent(GameEvent event) {
     @Override
     public void handleGameEvents(List<GameEvent> events) {
         if (paused) { return; }
-        Logger.info("Sending batch of {} game events", events.size());
+        String eventTypes = events.stream()
+                .map(e -> e.getClass().getSimpleName())
+                .collect(Collectors.joining(", "));
+        Logger.info("Sending batch of {}: [{}]", events.size(), eventTypes);
```
tool4ever: should such string operation be done with lazy logging?
https://tinylog.org/logging/#lazy-logging

MostCromulent: Wouldn't this only apply if the particular log event is conditional, i.e. there are circumstances it might fire and circumstances it might not - GameEvent batching is guaranteed to happen so I think that means this will always fire even with lazy logging enabled?

**Edit:** I guess this might apply if we moved this out of INFO level logging to DEBUG or TRACE, in which case depending on log level preference it wouldn't fire and you can save the computation? But for present purposes we want this logging in the production code so we can debug user logs.

tool4ever: yes, if you refactor it into Lambda we can later connect it with performance setting or something and save the string building that way

PR #10007, opened 2026-03-07
Affinity: extra class for the special cases

Hanmac:
Some cleanup for the bigger Keyword MR


forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -3824,23 +3824,12 @@ public void resolve() {
...
-            String desc;
-            if (k.length > 2) {
-                String typeText = k[2];
-                if (typeText.contains(" with "))
-                    desc = typeText.substring(typeText.indexOf(" with ") + 6);
-                else
-                    desc = typeText + "s";
-            } else
-                desc = CardType.getPluralType(t);
+        if (keyword.startsWith("Affinity") && inst instanceof Affinity affinity) {
```
tool4ever: I wonder if using `inst.getKeyword() == Keyword.AFFINITY` more wouldn't be cleaner?

PR #10013, opened 2026-03-08
Cleanup counting and grouping

Hanmac:
this replaces:
* the `map.put(key, map.get(old) + value` code with `map.merge(key, value, Integer::sum)`
* it also replaces many counting functions with grouping Collectors if able

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
@@ -755,8 +755,8 @@ public class ComputerUtilCard {
             return "";
         }
 
-        return list.stream().map(Card::getName)
-                .collect(Collectors.groupingBy(Function.identity(), Collectors.counting()))
+        return list.stream()
+                .collect(Collectors.groupingBy(Card::getName, Collectors.counting()))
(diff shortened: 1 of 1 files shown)


forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -2372,9 +2370,11 @@ public static String chooseSomeType(Player ai, String kindOfType, SpellAbility s
                 }
 
                 if (reducingCost) {
-                    List<String> valid = Lists.newArrayList(validTypes);
-                    valid.remove("Land"); // Lands don't have costs to reduce
-                    chosen = ComputerUtilCard.getMostProminentCardType(ai.getAllCards(), valid);
+                    List<CardType.CoreType> valid = validTypes.stream().map(s -> CardType.CoreType.valueOf(s)).collect(Collectors.toList());
```
tool4ever, line 2373: hmn imo this logic shouldn't only handle coretypes 🤔

but if it's not yet supported better to have the fallback below pick from validTypes instead

tool4ever: hmn nevermind should be safe I guess

PR #10016, opened 2026-03-08
More spring cleaning

tool4ever:
Removing lots of redundant string ops

### DIFF
--- a/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
+++ b/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
@@ -73,7 +73,7 @@ public class PlayerControllerAi extends PlayerController {
     }
 
     public void setupAutoProfile(Deck deck) {
-        pilotsNonAggroDeck = deck.getName().contains("Control") || Deck.getAverageCMC(deck) > 3;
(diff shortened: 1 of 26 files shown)


Hanmac:
technically, `Localizer.getInstance().getMessage(` gets back to MessageFormat
and this reacts a bit different when giving Numbers or Strings :thinking: 

but that should only matter for either numbers beyond 1000 or floating point numbers.

tool4ever:
I checked and all arguments are just converted normally there first anyway:
https://github.com/Card-Forge/forge/blob/dd8803bd74ab79a48f65f03140bcd302e22b2f02/forge-core/src/main/java/forge/util/Localizer.java#L128

so it should be 100% equivalent imo

Hanmac:
> so it should be 100% equivalent imo

inside MessageFormat it looks like this:
```java
                } else if (obj instanceof Number) {
                    // format number if can
                    subFormatter = NumberFormat.getInstance(locale);
                } else if (obj instanceof Date) {
                    // format a Date if can
                    subFormatter = DateFormat.getDateTimeInstance(
                             DateFormat.SHORT, DateFormat.SHORT, locale);//fix
                } else if (obj instanceof String) {
                    arg = (String) obj;

                } else {
                    arg = obj.toString();
                    if (arg == null) arg = "null";
                }
```

so if we would ever need that, we might need to change the Localizer

PR #10018, opened 2026-03-08
Add GameEventProxy to reduce network event serialization size

MostCromulent:
@tool4ever latest salvo in our war to win the GameEvent refactor.

## Summary

- Introduces `GameEventProxy`, which wraps `GameEvent` objects by replacing `CardView` and `PlayerView` references with lightweight ID markers before network serialization, preventing Java serialization from expanding the full game state object graph in each event batch
- Increases the network event flush interval from 50ms to 500ms to reduce the number of batches sent per game
- ~~Adds a unit test that discovers all `GameEvent` record types via reflection and verifies each can be serialized by the proxy~~

## How it works

`GameEventProxy` uses Java's `ObjectOutputStream.replaceObject()` / `ObjectInputStream.resolveObject()` hooks to automatically substitute `TrackableObject` references with `IdRef(typeTag, id)` markers during local byte-array serialization. This handles all nesting (records, collections, maps) automatically — no per-event-type handling needed.

**Server side:** In `NetGuiGame.handleGameEvents()`, each `GameEvent` is serialized into a byte array with CardView/PlayerView references replaced by IdRef markers. The resulting `GameEventProxy` is just a `byte[]` wrapper — Netty encodes it without traversing the game state object graph.

**Client side:** In `GameClientHandler.beforeCall(handleGameEvents)`, proxy byte arrays are deserialized with IdRef markers resolved back to `TrackableObject` instances from the client's `Tracker`. To ensure all objects are available for resolution, `updateObjLookup()` is called synchronously on the IO thread when `setGameView` arrives (since `copyChangedProps()` runs asynchronously on the EDT and may not have completed yet).

## Why only CardView and PlayerView?

Only these two types are replaced with ID markers. Other `TrackableObject` types serialize normally (they are small and don't cause graph expansion):

- **StackItemView** — Ephemeral. Only exists on the stack while a spell is resolving. With the 500ms flush interval, spells often resolve before the batch sends. The `GameView` reflects current state, so the StackItemView may no longer be present for `updateObjLookup()` to register.
- **SpellAbilityView** — Not reachable from `GameView`'s property graph. No `TrackableProperty` uses `SpellAbilityViewType`, so `updateObjLookup()` on the incoming `GameView` never discovers these objects.
- **CombatView** — Not referenced by any `GameEvent` record.
- **CardStateView** — Nested inside `CardView`; not independently referenced in events.

If wrapping fails for any individual event, the original event is sent as a fallback.

~~## Test~~

~~`GameEventProxyTest` uses Guava's `ClassPath` to discover all `GameEvent` record classes in `forge.game.event`, constructs each with default values, and verifies `GameEventProxy.wrap()` succeeds. This catches non-serializable fields in new or modified events at build time rather than at runtime during network play.~~

## Local testing

Verified with manual network games (server + client on localhost):
- Zero proxy-related errors (no "Failed to wrap", "Failed to unwrap", or "Could not resolve" in logs)
- `handleGameEvents` messages no longer appear in the encoder size log (below the 20KB reporting threshold), confirming the proxy is effective
- Only `setGameView` messages (~20KB compressed) appear in the encoder log — unchanged from before
- Game plays through combat, spells, and zone changes with no visual issues on the client side
- Unit test passes

---

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/net/client/FGameClient.java
+++ b/forge-gui/src/main/java/forge/gamemodes/net/client/FGameClient.java
@@ -4,9 +4,9 @@ import com.google.common.collect.Lists;
 import forge.game.player.PlayerView;
 import forge.gamemodes.net.CompatibleObjectDecoder;
 import forge.gamemodes.net.CompatibleObjectEncoder;
+import forge.gamemodes.net.IHasNetLog;
 import forge.gamemodes.net.ReplyPool;
 import forge.gamemodes.net.event.*;
-import org.tinylog.Logger;
 import forge.gui.interfaces.IGuiGame;
 import forge.interfaces.ILobbyListener;
 import io.netty.bootstrap.Bootstrap;
@@ -23,7 +23,7 @@ import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
-public class FGameClient implements IToServer {
+public class FGameClient implements IToServer, IHasNetLog {
     static final int HEARTBEAT_INTERVAL_SECONDS = Integer.getInteger("forge.net.heartbeatInterval", 15);
     private final IGuiGame clientGui;
     private final String hostname;
@@ -73,13 +73,13 @@ public class FGameClient implements IToServer {
                 try {
                     ch.sync();
                 } catch (final InterruptedException e) {
-                    Logger.error(e, "Client channel interrupted");
+                    netLog.error(e, "Client channel interrupted");
                 } finally {
                     group.shutdownGracefully();
                 }
             }).start();
         } catch (final InterruptedException e) {
-            Logger.error(e, "Client connect interrupted");
+            netLog.error(e, "Client connect interrupted");
         }
     }
 
@@ -93,7 +93,7 @@ public class FGameClient implements IToServer {
         if (disconnectSimulated) {
             return;
         }
-        Logger.info("Client sent {}", event);
+        netLog.info("Client sent {}", event);
         channel.writeAndFlush(event);
     }
 
@@ -103,7 +103,7 @@ public class FGameClient implements IToServer {
      * will detect the silence and close the connection.
      */
     public void simulateDisconnect() {
-        Logger.info("[simulateDisconnect] Suspending all network writes.");
+        netLog.info("[simulateDisconnect] Suspending all network writes.");
         disconnectSimulated = true;
         // Remove the IdleStateHandler to stop heartbeats, and add an outbound
         // handler that drops ALL writes (including game replies that bypass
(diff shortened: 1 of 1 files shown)


forge-gui-desktop/src/test/java/forge/net/GameEventProxyTest.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Verifies that all GameEvent record types can be serialized by GameEventProxy.
+ * This catches non-serializable fields in new or modified events, which would
+ * otherwise only fail at runtime during network play.
+ */
+@Test(groups = { "UnitTest" })
+public class GameEventProxyTest {
```
tool4ever: this seems rather redundant to `GameEventSerializationTest`?

forge-gui/src/main/java/forge/gamemodes/match/HostedMatch.java
```diff
@@ -226,7 +226,7 @@ public void startGame() {
                 gui.setOriginalGameController(p.getView(), humanController);
 
                 if (gui instanceof forge.gamemodes.net.server.NetGuiGame ngg) {
-                    forge.gui.control.GameEventForwarder forwarder = new forge.gui.control.GameEventForwarder(gui);
+                    forge.gui.control.GameEventForwarder forwarder = new forge.gui.control.GameEventForwarder(gui, 500_000_000L);
```
tool4ever: don't think we need this overload right now...

forge-gui/src/main/java/forge/gamemodes/net/GameEventProxy.java
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * Wraps a {@link GameEvent} by serializing it into a byte array with all
+ * {@link TrackableObject} references replaced by lightweight {@link IdRef}
+ * markers. On the client side, the proxy is unwrapped by resolving each
+ * IdRef from the client's {@link Tracker}.
+ *
+ * <p>This avoids Java serialization expanding TrackableObject references into
+ * the full game state object graph when events are sent over the network.
+ */
+public class GameEventProxy implements Serializable {
```
tool4ever: let's cherry pick `IHasNetLog` from delta branch into this
then we can already log here with the right tag :)

MostCromulent: Done + have added the NETWORK tag console writer to tinylog config - this will be overridden with the more detailed log file writer when delta lands, unless you want all of that infrastructure included now.

forge-gui/src/main/java/forge/gui/control/GameEventForwarder.java
```diff
@@ -45,4 +65,16 @@ public void flush() {
...
+
+    private void cancelScheduledFlush() {
+        if (scheduledFlush != null) {
+            scheduledFlush.cancel(false);
+            scheduledFlush = null;
+        }
+    }
+
+    public void shutdown() {
+        cancelScheduledFlush();
```
tool4ever, line 77: doesn't this method need hookup somewhere?

MostCromulent: Added - https://github.com/Card-Forge/forge/pull/10018/commits/df48d33d37ba70b880cbcddc2ea5267228679eb4

PR #10026, opened 2026-03-09
Fix null CardView NPEs from EventProxy deserialization

MostCromulent:
## Summary

- Fix NPE in `EventVisualizer.visit(GameEventZone)` when the event's card is null
- Fix NPE in `CMatchUI.updateCards` when iterating card collections containing null elements

Both bugs share the same root cause: `GameEventProxy` replaces `CardView` references with ID markers during network serialization. On the client, these IDs are resolved from the `Tracker`. When a card no longer exists in the Tracker — e.g. a token created by Rite of Replication and immediately destroyed by the Legend Rule — the resolution returns `null`, which Java's deserialization writes back into the event's fields and collections.

Before `GameEventProxy`, events were serialized with full `CardView` objects inline, so even destroyed cards deserialized as complete (non-null) objects. The proxy's ID-based approach is more efficient but introduces this new failure mode for ephemeral cards.

The fix adds null guards at the two crash sites:
- `EventVisualizer.visit(GameEventZone)`: skip the event if `card` is null
- `FControlGameEventHandler.processCard/processCards`: filter null entries before adding to the UI update sets

Dropping nulls is safe — these represent cards the client can't display anyway (they aren't in the GameView), so skipping their UI refresh has no visible effect.

---

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/match/HostedMatch.java
+++ b/forge-gui/src/main/java/forge/gamemodes/match/HostedMatch.java
@@ -206,7 +206,6 @@ public class HostedMatch {
                 }
             }
             p.updateAvatar();
-            //sleeve
             p.getLobbyPlayer().setSleeveIndex(rp.getPlayer().getSleeveIndex());
             if (p.getLobbyPlayer().getSleeveIndex() == -1) {
                 if (iPlayer < sleeveIndices.length) {
@@ -217,8 +216,7 @@ public class HostedMatch {
             }
             p.updateSleeve();
 
-            if (p.getController() instanceof PlayerControllerHuman) {
-                final PlayerControllerHuman humanController = (PlayerControllerHuman) p.getController();
+            if (p.getController() instanceof PlayerControllerHuman humanController) {
                 final IGuiGame gui = guis.get(p.getRegisteredPlayer());
                 humanController.setGui(gui);
                 gui.setGameView(null); //clear out game view first so we don't copy into old game view
@@ -234,7 +232,7 @@ public class HostedMatch {
                 }
                 playersPerGui.put(gui, p.getView());
 
-                if (gameControllers != null ) {
+                if (gameControllers != null) {
                     LobbySlot lobbySlot = getLobbySlot(p.getLobbyPlayer());
                     gameControllers.put(lobbySlot, humanController);
                 }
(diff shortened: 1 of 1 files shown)


forge-gui/src/main/java/forge/gui/control/FControlGameEventHandler.java
```diff
@@ -167,7 +168,9 @@ private Void processCards(final Collection<CardView> cards, final Set<CardView>
         if (cards.isEmpty()) { return null; }
 
         synchronized (list) {
-            list.addAll(cards);
+            for (CardView c : cards) {
+                if (c != null) list.add(c);
```
tool4ever: I think this really needs a better low level fix - trying to put null checks everywhere will surely come back to curse us (besides expecting null inside Collection is really not our convention!)

I **hope** it can work like this:
- skip events if one field deserializes completely null
- filter null from deserialized collection

MostCromulent: Do you have preference from below?

> **Option A: Drop the whole event on any resolution failure**
> 
> Pros:
> - Simple - one flag, one check, done
> - No event-type-specific logic
> - Guarantees no null TrackableObject reaches any consumer
> 
> Cons:
> - Loses the entire event even if only one element in a collection failed. E.g., a "blockers declared" event with 5 cards where 1 failed to resolve - you lose all 5 instead of just the 1
> - More events dropped than strictly necessary
> 
> **Option B: Filter nulls from collections, drop events with null scalar fields**
> 
> Pros:
> - Preserves partial information ΓÇö the 4 good blockers still produce sound/UI effects
> - Closer to tool4ever's literal wording
> 
> Cons:
> - Harder to implement generically at the deserialization layer - `resolveObject` works on individual objects, it doesn't know if they're inside a collection or a scalar field
> - Would likely need post-processing after deserialization: walk the event's fields, filter nulls from any collections, reject if any scalar TrackableObject field is null
> - Event records are immutable - you can't mutate a collection inside a record after construction. You'd need to reconstruct the event or use mutable collections
> - More code, more surface area for bugs
> 
> **Recommendation:** Option A is the right call for now. The events affected are overwhelmingly simple (one card, one zone). Collection-heavy events like `GameEventBlockersDeclared` failing to resolve one blocker is an edge case of an edge case. Start simple, revisit if it causes problems.

tool4ever: hmn ok A it is then if AI thinks B would get too messy

also consider if it's worth performing lookup on host side first as sanity check - if it fails remote would probably fail too so we could save traffic?

MostCromulent: Done on https://github.com/Card-Forge/forge/pull/10026/commits/49e1bfbc7c919f05dd07a7d1773e486ba3861b18

PR #10029, opened 2026-03-10
Craft: rework Desc and reminder

Hanmac:


tool4ever:
this one's wrong now:
image

PR #10037, opened 2026-03-11
Card zone view - misc QOL improvements

MostCromulent:
Follow-up to #9969 to add a handful of quality-of-life fixes:

- **Separate own/opponent preferences** - "Open in Window" / "Add Tab" choices are now stored independently for your zones and opponent zones . Previously a single preference controlled both, so choosing tab mode for e.g. your graveyard would also force tab mode for all opponents' graveyards when you may want those in windows instead.
- **Prevent duplicate hand tabs** - Clicking the hand zone button under your portrait when tab mode is active now selects the player's existing VHand tab instead of creating a second VZone tab with identical content. When "Open in Window" is selected, the floating window still opens as before.
- **Opponent tab labels** - Docked zone tabs for opponents are prefixed with the player name (e.g. "Bob Graveyard (3)").
- **Hand tab card count** - VHand panel tab now shows the current card count (e.g. "Player Hand (7)"), matching the pattern used by docked zone tabs.
- **Localization keys** - Undock, Sort by Name, and Unsort were previously hardcoded English strings, now use localization keys.
- **User documentation** - Added "Viewing Cards in Different Zones" subsection to Layout in `docs/User-Guide.md` documenting zone features including configurable display modes and drag-to-dock.

---

🤖  Generated with [Claude Code](https://claude.com/claude-code)


forge-gui/res/languages/en-US.properties
```diff
@@ -2895,6 +2899,7 @@ lblNoPlayerForEDocID=NO PLAYER FOR {0}
 #VHand.java
 lblYourHand=Your Hand
 lblPlayerHand={0} Hand
+lblPlayerHandN={0} Hand ({1})
```
tool4ever: this would be redundant to `lblPlayerHandN`

forge-gui/res/languages/en-US.properties
```diff
@@ -2887,6 +2887,10 @@ lblPreventCardOverlap=Prevent Card Overlap
 lblHandPanel=Hand Panel
 lblOpenInWindow=Open in Window
 lblAddTabToHandPanel=Add Tab to Hand Panel
+lblUndock=Undock
+lblSortByName=Sort by Name
+lblUnsort=Unsort
+lblPlayerZoneN={0} {1} ({2})
```
tool4ever: that's not really a localization string
should avoid adding trivial hardcoding

forge-gui-desktop/src/main/java/forge/screens/match/views/VHand.java
```diff
@@ -135,4 +137,10 @@ public HandArea getHandArea() {
         return VHand.this.hand;
     }
 
+    public void updateTabLabel(final int count) {
+        if (player == null) return;
+        tab.setText(Localizer.getInstance().getMessage("lblPlayerHandN", player.getName(), String.valueOf(count)));
```
tool4ever: I like the idea of using the label for some additional information
but at the same time the card number is already part of the player details panel so this feels quite redundant

so we should rather look for something else that'd add more value?

MostCromulent: For present purposes this is just making the tab for the player hand consistent with tabs for other docked zones, which already include card count.

If we're going to abandon that here should we also be removing it for the other zones? Do you have a view on what could more helpfully be included?

Speaking for myself, I find the card count in the tab header useful - if I've got the zone docked as a tab in the hand panel I'm more likely to look to the count there than in the details panel + seeing the count change in the tab is a good reminder that eg there's a new flashback or other card I'm able to play from a non-hand zone.

tool4ever: yea consistency is good - just feeling that the redundancy would be more obvious now with the hands being tabbed by default
hmn but I guess we need to find something not problematic for hidden zones 🤔

what about counting "X new" (or even better a delta number) (`enteredThisTurn`) instead?

this would basically support your reasoning but convey that information more directly instead of having to essentially track against old card totals? 💡

thinking about it that sounds like more helpful to me, especially in complex multiplayer boardstates to let you more quickly identify which opponent zones had changes you might have missed otherwise without closer log study (e.g. after yielding too much :D)
(not sure how easily this also applies to the Flashback zone since its more generated, maybe it'd have to stay an exception for now or can actually be supported with reasonable amount of extra code...)

also probably needs to be turned into a menu option then?

MostCromulent: See implementation in latest commit:

> **Delta tracking approach:** Shows "+N new" on zone/hand tabs that aren't currently visible, counting cards that arrived since the tab was last viewed. The baseline resets whenever the tab is the selected (active) tab in its cell, so switching to a tab clears its delta. This is purely client-side - no engine changes needed.
> 
> Considered tracking `enteredThisTurn` but went with view-based tracking because:
> - "New since you last looked" is more useful than "new this turn" - if you check a zone mid-turn the delta clears, and in long multiplayer turns you don't lose the count at turn boundaries
> - No game engine changes required - avoids adding new trackable properties, protocol methods, and network traffic for what is a minor GUI feature
> - Avoids edge cases around turn-change timing (draw-for-turn happening before the first UI refresh of the new turn)
> 
> The tradeoff is that it's count-based, not entry-based: if 2 cards enter and 1 leaves while you're not looking, the delta shows +1 rather than +2. For the intended use case (noticing zone changes) this is sufficient and additional code complexity is not justified.

PR #10040, opened 2026-03-11
CountersPutAi: fix AI because Choices are before Target

Hanmac:
fix Role Model choosing useless Keyword counters


Agetian:
Looks good, can't think of anything this can break off the top of my head :+1:

PR #10042, opened 2026-03-12
Adventure mode: make map overlay toggle sticky

vanja-ivancevic:
When exploring, it helps to have an overview of which POIs have been visited. The "names" toggle allows you to do this, but it's very annoying to have to re-toggle every time you open the map.

- The map view overlay toggle (details/events/reputation/names) now remembers the last selected mode
- Re-opening the map restores the previous overlay instead of always resetting to the default state
- Tracked as an instance field on the singleton `MapViewScene`, so it persists for the session

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/scene/MapViewScene.java
+++ b/forge-gui-mobile/src/forge/adventure/scene/MapViewScene.java
@@ -166,18 +166,26 @@ public class MapViewScene extends UIScene {
     }
 
 
+    private void setOverlayButtonStates(int mode) {
+        String[] buttons = {"details", "events", "reputation", "names"};
+        // Each mode shows only the *next* button in the cycle
+        // mode 0 (none/names): show "details"
+        // mode 1 (details):    show "events"
+        // mode 2 (events):     show "reputation"
+        // mode 3 (reputation): show "names"
+        int activeIndex = mode; // the button to show (wraps: 0->details, 1->events, 2->reputation, 3->names)
+        for (int i = 0; i < buttons.length; i++) {
+            TextraButton btn = ui.findActor(buttons[i]);
+            if (btn != null) {
+                btn.setVisible(i == activeIndex);
+                btn.setDisabled(i != activeIndex);
+            }
+        }
+    }
+
     public void details() {
         lastOverlayMode = 1;
-        TextraButton detailsButton = ui.findActor("details");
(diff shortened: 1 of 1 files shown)


forge-gui-mobile/src/forge/adventure/scene/MapViewScene.java
```diff
@@ -46,6 +46,7 @@ public class MapViewScene extends UIScene {
     private final float maxZoom = 1.2f;
     private final float minZoom = 0.25f;
     private Set<PointOfInterest> bookmark;
+    private int lastMode = 0; // 0=none, 1=details, 2=events, 3=reputation
```
Jetz72: Maybe call it `lastOverlayMode` or something for clarity?

forge-gui-mobile/src/forge/adventure/scene/MapViewScene.java
```diff
@@ -370,6 +375,11 @@ public void enter() {
             questButton.setDisabled(labels.isEmpty());
             questButton.setVisible(!labels.isEmpty());
         }
+        // Restore last overlay mode
+        if (lastMode == 1) details();
+        else if (lastMode == 2) { details(); events(); }
+        else if (lastMode == 3) { details(); events(); reputation(); }
+
```
Jetz72, line 344: Why run all the previous ones? Is it just so the weirdly-implemented cycling buttons end up in the right states?

vanja-ivancevic: Yes, exactly. Each method hides its own button and shows the next one in the cycle, so chaining is needed to get all button states correct. 

A cleaner approach would be to extract a helper like `setOverlayButtonStates`(int mode) that directly sets all four buttons' visibility for the target mode, then call only the target method (e.g. just `reputation`()) without needing the intermediate steps. Happy to refactor if you'd prefer that.

Jetz72: Yeah that'd be a step in the right direction.

vanja-ivancevic: Extracted a `setOverlayButtonStates` helper so the restore no longer needs to chain through intermediate modes. Works well in testing. Feel free to remove the comments if you prefer.

PR #10045, opened 2026-03-12
Sealed Deck event in Adventure Mode

Agetian:
This idea has been floating around in my mind for quite a while, but I had neither the time nor the knowledge to figure this out until recently. This is an attempt to lay the foundations for a possible implementation of a new event type in Adventure mode - a Sealed Deck event which works within the official Sealed Deck parameters (6 booster packs - or whatever the code returns for the Sealed booster pack count), and plays similar to the Draft format otherwise (8 players in a tournament, single elimination format).

**Important note:** I'm not an expert on UI or mobile Forge intrinsics, so I had to use **AI assistance** to figure out how the code works and how to properly integrate the Sealed Deck events into the existing framework. Some of the methods were written with AI assistance as well and then inspected and tweaked where needed. Therefore, **even though I went over every single portion of the code and rewrote a good deal of it that looked out of place or too intrusive, this blueprint contains elements of AI-assisted code which is not the most efficient or proper way of implementing this.** As such, **help is desperately wanted and needed** here to help both clean this up and modify it so that it's up to par with the existing code so that it can be merged if this feature would be desired upstream.

The current blueprint was tested in action in various conditions and seems to work. The following basics are tested:
- Generation of the Sealed Deck events.
- Applying for a Sealed Deck event.
- Using a Silver Challenger Coin to apply for the event.
- Creating a deck.
- Exiting the deck editor with a valid deck and with an invalid/incomplete deck.
- Starting the event.
- Going through the rounds.
- Receiving tier 0 (loss) and tier 3 (champion) rewards.

Note the following implementation specifics:

- Currently, the application price is $6000 (assuming that $3000 in the draft is the price of the three booster packs, thus, $1000 per pack - thus, for a Sealed which is typically 6 booster packs, $6000 seems fair). Since this is double the price of the Draft event, I also doubled the price in Shards (thus, 100 shards compared to 50 shards for the Draft event).
- Silver Challenger Coin, which has been unused for a few years, is used as an alternative means of entering the tournament. This means that by default, you're getting two of those at the start of the Adventure - would that be fair or is that too much? Currently I changed it to 1 Silver coin at the start of Adventure, seems like a better choice.
- The ultimate (Champion) reward is currently the entire Sealed pool (90 cards), as if you purchased the actual 6 boosters at the price of $1000 per booster, plus the other rewards (which means 2 boosters on top of that + the redeemable Coin). Is that fair game or is that too overpowered? Personally I think it's fine to award the full Sealed pool, just not sure if the extra boosters push this over the top or not.
- Currently, the event is set to be rather rare (25% of the time Sealed, 75% of the time Draft). Should this be tweaked to be more or less common? Should this be configurable per Adventure world? (e.g. in Old Border Shandalar, Sealed Deck events could be more common if desired, because back in the old days, this limited format was more prevalent?). Not sure how difficult it would be to do this, if needed.

**I really hope that someone can collaborate here and submit code modifications and changes where needed to complete this project properly.** @Jetz72, if you're available at some point and have some time, can you please assist? 

Thank you very much in advance to anyone who would be willing to contribute to this. Please feel free to commit directly to this branch if you have working changes that would be an improvement.


Jetz72:
Can dig into this more later, but I like the sound of it from a glance.

> Also, is it OK to reuse the Silver Challenger Coin for the purpose of entering the Sealed Deck event, thus relegating it of its previously intended possible use? I know that it was planned for some kind of a Constructed event, but that was never finished and there is no indication that it will ever be, so I thought that using the available coin would be fair.

Seems like a valid option. An alternative would be to make it cost two gold event tokens. I guess it depends if you see Sealed events as a league below Draft events? The tiering nature of the coins seems a bit questionable to me in general the more I think about it.

> Currently, the event is set to be rather rare (25% of the time Sealed, 75% of the time Draft). Should this be tweaked to be more or less common? Should this be configurable per Adventure world? (e.g. in Old Border Shandalar, Sealed Deck events could be more common if desired, because back in the old days, this limited format was more prevalent?). Not sure how difficult it would be to do this, if needed.

Probably wouldn't hurt to have more event-related knobs for designers to turn inside the config file. There's also Jumpstart which has a decreasing likelihood over time, but not sure how we'd represent that as a setting.

> * Code wise, some of the changes are most definitely hacky. For instance, AdventureDeckEditor has a method called isDrafting which is specific to booster drafts, and which overrides the method in FDeckEditor. However, I have to update this method with some Sealed-specific tweaks, thus making Sealed Deck events function as a form of "hacked" Draft events, to an extent. I couldn't figure out how to do this better (and I can't rename the method since it's an override), the AI couldn't suggest a workable framework for this either without building an overly complex and convoluted uh... something. Therefore, there are some FIXME comments in the code where it feels like changes are needed but I'm not sure how to do this better than it is currently written.

Why do you need `isDrafting` to be true? The editor doesn't draft cards in this format. I'd think Sealed would be close enough in nature to Jumpstart that you could get away without changing anything in the deck editor at all.

Agetian:
Thanks a lot for your input, @Jetz72, that's actually already very helpful! :)

Yes, reusing the Silver Coin makes sense to me when it comes to "tiering" these, at least in a sense - I don't think we can truly "tier" them by the outcome (rewards etc.), they're just different, but we can "tier" them by complexity, with Jumpstart being the simplest one (you don't need to even build the deck, only choose the packs), Sealed being the advanced one (you need to build the deck, but you build it from the packs that you're provided with), and Draft being the most complex one (you need to not only build the deck, but draft the "building blocks" [cards] one by one with a certain strategy and synergy in mind). Thus, I think this matches the Bronze-Silver-Gold "tiers" quite nicely :)

I agree that some of those options (for all three event types) would be nice to be made configurable in the json file, though currently I'm not sure what exactly and how to implement here, so this might be a thing to consider at the end of the project, maybe even as a separate PR :)

Oh, that's actually so true about the Drafting/isDrafting part, when I started thinking about how to design this, I assumed that Sealed Deck tournaments are kind of like "drafts without the drafting", which is why I tried to reuse the drafting mechanics and had to work around the specifics (in particular, isDrafting needs to be "true" in order for the system to correctly advance the event status at the needed times). I'll see if I can rebuild this with the Jumpstart-like approach in mind, that might be cleaner (if I can figure out how :))  If I produce anything workable, I'll update this PR accordingly.

Agetian:
Ok, tried to rewrite it in a less intrusive way so that the Drafting part isn't touched/tainted with the Sealed-related code :)

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
```diff
@@ -657,6 +665,32 @@ public void completeDraft() {
         }
     }
 
+    public void completeSealed() {
+        AdventureEventData currentEvent = getCurrentEvent();
+        if (currentEvent == null)
+            return;
```
Jetz72: Is this now unused?

Agetian: Umm no, it's called from onClose in order to properly transition from one event state to another and do other wrap up stuff :)

Agetian:
Tested Jumpstart, Sealed, and Draft events to ensure that they all work with this code change.

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
Jetz72: Ah, I dunno how I missed that whole block of code while looking over it. Sorry for the confusion.

Agetian: No prob at all, thank you for looking into this, it's pretty complex for me tbh, so it's great to have a second pair of eyes ^^;

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
```diff
@@ -910,6 +943,35 @@ public void onClose(final Consumer<Boolean> canCloseCallback) {
...
+                        false, result -> {
+                    if (result) {
+                        currentEvent.eventStatus = AdventureEventController.EventStatus.Abandoned;
+                        resolveClose(canCloseCallback, true);
+                    } else {
+                        canCloseCallback.accept(false);
+                    }
+                });
+                return;
+            } else {
```
Jetz72: Not sure this validation or the option to abandon the event is needed here; unlike draft, where the draft manager will break if we try to leave with a half-backed draft and can't be written to a save, it should be fine to leave at this stage of a sealed or Jumpstart event. Since we're not using `isDrafting`, it'll fall into the regular validation logic below, checking it against `DeckFormat.Limited` (i.e. 40 cards). If they ignore the warning and start a game, their deck will just get padded with Wastes over in `DuelScene.applyAdventureDeckRules`.

I suppose it'd also be possible to just run the appropriate validator when they click the start game button and refuse to let them into the game with an unfinished deck. That might let us simplify things a ton since we could just skip over the started step entirely when starting a Sealed event.

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
```diff
@@ -657,6 +665,31 @@ public void completeDraft() {
...
+        currentEvent.isDraftComplete = true;
+
+        if (currentEvent.draftedDeck == null) {
+            currentEvent.draftedDeck = new Deck();
+            CardPool pool = currentEvent.registeredDeck.get(DeckSection.Sideboard);
+            if (pool != null) {
+                currentEvent.draftedDeck.getOrCreate(DeckSection.Sideboard).addAll(pool);
+                currentEvent.draftedDeck.setName("Sealed Pool Cards");
+            }
+        }
```
Jetz72: Don't you already save the card pool to `draftedDeck` over in `AdventureEventData.generateSealedPool`?

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
```diff
@@ -657,6 +665,31 @@ public void completeDraft() {
...
+
+        if (currentEvent.eventStatus == AdventureEventController.EventStatus.Entered) {
+            currentEvent.eventStatus = AdventureEventController.EventStatus.Ready;
+        }
+
+        if (allowAddBasic()) {
+            showAddBasicLandsDialog();
+        }
```
Jetz72: Whichever way we go with validation, doesn't showing the basic lands dialog at the end of the deck building process mean they'll only see it after they've already finished their deck (presumably with all the lands they want included).

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
Agetian: Yeah, I noticed that part before (getting the deck padded with Wastes), though I'm not particularly sure which way would be best here: to allow the player to play with an unfinished deck without a warning (which may end up being slightly confusing as to why they have Wastes in their deck all of a sudden, or maybe if they misclicked expecting the game to allow them to add basic lands, or whatever), or be "on the safe side" and warn against submitting an incomplete deck - for a newcomer, the current approach may actually be slightly more intuitive, but I'm not sure :/ I'm fine with having it either way, I agree that making it simpler (with no warning), we can streamline the process quite a bit. I guess we can ask for more opinions, perhaps?

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
Agetian: Yeah, it does feel a bit annoying, I copied it from the Draft portion which has this comment:

            //Might be annoying if you haven't pruned your deck yet, but best to remind player that
            //this probably needs to be done since it's there since it's not normally part of Adventure

Not sure if it's best to keep it or not to keep it - currently it's consistent with Draft and it does remind the player to pay attention to it, but it's annoying for someone who's played for a while already :/ What do you say? Remove it from Sealed?

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
Agetian: Aye, good catch :) Thanks! I removed it and tested to ensure everything still works just fine.

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
Jetz72: I think the only reason we can't just skip the "Entered" phase (where in other events you'd either be drafting cards or choosing your Jumpstart pieces) is that it'd currently make it easily possible to accidentally start a match with a 0 card deck. If we do the validation in EventScene when the user clicks the button to start a match, we can ensure that this won't ever happen. They'll get an error directing them to click the edit deck button and the confusion is averted. Then we won't need to hook the editor into the event state machine.

Agetian: Ah, sounds like a fair plan, yeah! Not sure if I fully understand how to properly do this though... If you'd be willing to commit the change, please feel free to commit right into this branch, I would really appreciate it!

Agetian:
> I can take a look at putting event probabilities in the config next time I'm dabbling in that area.

Thanks a lot, I appreciate this as well, sounds like a great change for the different emergent worlds we already have in Adventure mode :+1:

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
Jetz72: In the draft flow, it'll appear after they've selected their last card, before assembling their final deck. But in this iteration of Sealed, it'd appear only after your deck has passed validation. If they haven't added any basic lands yet, they would have had to construct an unplayable 40 card deck without any basic lands. Even if we move the dialog ahead of validation, it'd still only appear when the user clicked to exit the editor, which they'd usually only do when they're finished building their deck.

We could consider ways to make the add basic lands dialog more apparent. The question of how to do it occasionally crops up on Discord. I'm not sure if it'd be worth the trouble yet, but it may be possible to insert "Basic Lands" as one of the main tabs in the editor alongside "Main" and "Sideboard". A much lazier solution would be to just dump a pile of basics into their card pool. Maybe after saving their `draftedDeck` so it doesn't pollute their rewards.

Agetian: True! :) Alrighty, I removed the Add Basic Lands dialog.

Yeah, somehow this probably needs to be just made "more visible" in general, I remember watching ProJared's playthrough of Adventure on YouTube (a really great one, btw, it's super long), he was a beginner when we just started, and it was painful to see him go from village to village looking for basic Forests to buy since he wasn't aware of the Add Basic Lands functionality at all (outside of the draft context, just in the base Adventure ^^; )

Agetian:
Alrighty, fixed a couple issues (namely, the sealed pool being regenerated every time the deck editor was entered) and rewrote the code to be simpler when it comes to the deck editor (so that AdventureDeckEditor is only minimally touched for the needed stuff)

forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
Agetian: Tried my best to simplify it in vein of our discussion, probably the best I can do, would appreciate some help if more modifications would be needed :)

Agetian:
Alright, this should be ready in its basic implementation, at least as far as I can take it and test it. I suggest that we try to merge this in and update in the followup PRs as needed.

Issue #9044, opened 2025-10-30
AI targets itself with both the chosen modes of Shadrix Silverquil's ability


dave-lovell:
**Describe the bug**
Shadrix Silverquil reads: '... you may choose two. Each mode must target a different player.' But the AI chooses two and targets itself with both modes.

**To Reproduce**
1. Play a 2-player game against AI with Shardix Silverquil
2. The AI will misplay Shadrix's ability

**Expected behavior**
I expect the AI to target itself with one ability, and the other player with the other ability.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: Chrome
 - Version: Unsure, sorry

PR #10053, opened 2026-03-13
Set unique target players for CharmAi

kojotak:
Possible fix for #9044. 

AI no longer targets the same player. Same issue as with `Shadrix Silverquill` can be observed with `Chaos Balor`. It works, but I'm not sure, whether `CharmAi` is the right place (and time) to set targets. 

image

image


Agetian:
Are there cards with CharmAi that can legitimately target the same player and you can choose more than one option? If there are, would this break them, especially in 1v1 matches where this behavior would be viable?

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
```diff
@@ -79,6 +80,22 @@ protected AiAbilityDecision checkApiLogic(Player ai, SpellAbility sa) {
...
+        if (chosenList.size() > 1) {
+            // Each mode must target a different player.
+            long uniquePlayerTargets = chosenList.stream()
+                    .filter(ab -> "True".equals(ab.getParam("TargetUnique")))
+                    .map(ab -> ab.getParam("ValidTgts")).filter(Objects::nonNull)
+                    .filter(tgt -> "Player".equals(tgt) || "Opponent".equals(tgt)).count();
+            if (uniquePlayerTargets == num) {
+                Player nextPlayer = ai;
+                for (AbilitySub choosen : chosenList) {
+                    choosen.resetTargets();
```
tool4ever: hmn this is a tricky situation:
I don't think trying to retroactively fix the situation is viable enough (right now it could still perform illegal targeting and chances are this approach just ends up producing dumb plays anyway)

What should happen instead is manually building out the SA chain inside each choices loop so that their API checks can just naturally skip already targeted entries 🪄

Though it looks to me like this might risk `CharmEffect.chainAbilities` creating an infinite loop, so probably also requires some additional reset logic before finishing 🤔

kojotak:
> Are there cards with CharmAi that can legitimately target the same player and you can choose more than one option? If there are, would this break them, especially in 1v1 matches where this behavior would be viable?

I was worried about it too, therefore the proposed change checks for `TargetUnique$ True` and `ValidTgts` `Player`/`Opponent`. So for example `Primal Command` is unaffected (even if AI will choose the two modes with `ValidTgts$ Player` - they are not `TargetUnique`)

---

I tried to check charms with `CanRepeatModes$ True` but those are not supported yet (for example `Verdant Confluence`)

```
// TODO unused for now, the AI doesn't know how to effectively handle repeated choices
boolean allowRepeat = sa.hasParam("CanRepeatModes");
```

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
kojotak: > I don't think trying to retroactively fix the situation is viable enough

You are right. I've discovered some invalid targeting today as well 😞. 

I was considering adding new parameter (`TargetDifferentPlayer$ True`) representing `Each mode must target a different player.` (as a hint for targeting) but it duplicated the `TargetUnique$ True` (for each mode). 

I will try to take a look at "SA chain", because that's something new for me.

kojotak:
Refactoring from the last week:

- introduction of new keyword EachModeMustTargetADifferentPlayer$
- it helps to separate "custom AI logic", so other charms are not affected
- list of choices is build sequentially (instead of doing choices for current AI player and then changing targets with uncertain results)

tool4ever:
I'm still thinking about a more robust approach for this 🤞

PR #10066, opened 2026-03-15
Bug fix: Enforce allowedEditions in adventure card resolution

vanja-ivancevic:
- Adventure mode card pool (`uniqueCardsByName`) picks printings based on which set images are downloaded, not which editions are allowed by the plane config
- When players have modern set images cached, cards in shops, rewards, and player inventory can show modern-frame printings (e.g. Commander Masters Sol Ring instead of Revised)
- Fix: enforce allowedEditions filtering in `getFullCardPool()`, `getCardByName()`, and sanitize card pools on save load to fix existing saves
- Affects any adventure plane that uses `allowedEditions` in config.json

Bug reported by many players who see new-border card variants after updating Forge snapshots. Clearing image cache temporarily fixes it because `hasImage()` likely re-evaluates, but the cards return on next update.


Agetian:
Looks good to me, could prolly use another look too just in case :)

Agetian (approved):

forge-gui-mobile/src/forge/adventure/util/CardUtil.java
```diff
@@ -824,8 +836,72 @@ public static PaperCard getCardByNameAndEdition(String cardName, String edition)
...
+            if (card != null && !allowed.contains(card.getEdition())) {
+                toReplace.add(Map.entry(card, entry.getValue()));
+            }
+        }
+        for (Map.Entry<PaperCard, Integer> entry : toReplace) {
+            PaperCard replacement = ensureAllowedEdition(entry.getKey());
+            int count = entry.getValue();
+            pool.remove(entry.getKey(), count);
+            pool.add(replacement, count);
+        }
```
Jetz72, line 898: I think the replacement logic here will clear the no-sell flag on Jumpstart cards in players' inventories.

forge-gui-mobile/src/forge/adventure/player/AdventurePlayer.java
```diff
@@ -624,6 +625,7 @@ public void load(SaveFileData data) {
                     decks.add(new Deck(data.readString("deck_name_" + i)));
                 }
                 CardPool mainCards = CardPool.fromCardList(Lists.newArrayList((String[]) data.readObject("deck_" + i)));
+                CardUtil.sanitizeCardPool(mainCards);
```
Jetz72: Should this also be applied to sideboards and commanders? If they have an undesired print in those, you may end up with a disparity between the deck and the inventory.

vanja-ivancevic:
Thanks very much for the insightful review. I agree that it should be a console command to "sanitize". Updated based on feedback:

- Removed automatic save sanitization from `AdventurePlayer` load. Existing saves are no longer modified on load.
- Added `sanitize editions` console command instead, so players can opt-in to cleaning up wrong-edition cards in their collection and all deck sections (main, sideboard, commander, attractions, contraptions).
- `sanitizeCardPool` now preserves `noSell` flags when replacing cards.
- The prevention layer (`getFullCardPool` and `getCardByName`) remains, ensuring no new wrong-edition cards enter the game going forward.

PR #10067, opened 2026-03-15
Feature: Added support for custom number of players for smaller cubes

ke7aqm:
Added support in .draft file to set the NumPlayers for cubes smaller than 360 that can't support that many players.
Added new NumPlayer field to all current cubes to default to 8. Need to find out if there is a place in the documentation that needs to be updated as well.

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
+++ b/forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
@@ -391,7 +391,19 @@ public class BoosterDraft implements IBoosterDraft {
 
         final SealedTemplate tpl = draft.getSealedProductTemplate();
 
-        // TODO Determine PodSize from custom draft
+        int players = draft.getNumPlayers();
+        final int minPlayers = 2;
+        final int maxPlayers = 8;
+
+        // Check if numPlayers is set in draft file, if not default to 8.
+        if (players == 0) {
+            players = N_PLAYERS;
+        } else {
+            players = Math.max(players, minPlayers);
+            players = Math.min(players, maxPlayers);
+        }
+
+        setPodSize(players);
 
         final UnOpenedProduct toAdd = new UnOpenedProduct(tpl, dPool);
         toAdd.setLimitedPool(draft.isSingleton());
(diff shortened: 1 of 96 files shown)


ke7aqm:
Edit: Set merge for rebase.

The commit looks scarier than it is, since there are 94 files that have been modified, but 92 of them are .draft files where I added the new NumPlayers variable to each .draft in the repo for reference. I don't know if it should be documented somewhere else, but I wanted to make sure a user could see that the variable was there to use. If the is a better place or way to document, I'm glad to add it.

forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
```diff
@@ -391,7 +391,22 @@ private void setupCustomDraft(final CustomLimited draft) {
 
         final SealedTemplate tpl = draft.getSealedProductTemplate();
 
-        // TODO Determine PodSize from custom draft
+        int players = draft.getNumPlayers();
+        final int minPlayers = 2;
+        final int maxPlayers = 8;
+
+        // Check if numPlayers is set in draft file, if not default to 8.
+        if (players == 0)
```
Agetian: Stylistically we generally use this form of if:

if (players == 0) {
...
} else {
...
}

Otherwise, looks good to me

ke7aqm: Good to know. I am more of a C/C++ programmer and don't generally do much with Java. That styling has always bothered me, so I default to what I did. I can change it though, if it is preferred. Should I update it?

Agetian: Please do, it would be good for consistency :) Yeah, I know what you mean, I also do a lot of coding in a variety of languages, it's a bit confusing sometimes as to what style I should use :)

ke7aqm: Done. :)

ke7aqm: Should I also add a Twobert that is 180 cards with a .draft file to match and show some of the options? Maybe Ryan Overturf's Original Recipe Twobert? Or is it good the way it is?

Agetian: Indeed, would be good to have some functional cubes to go with this feature :)

PR #10080, opened 2026-03-17
Fix Finality/Shield/Stun counters invalidating "no abilities"

tool4ever:


Hanmac:
Can we clean this up with `updateReplacementEffects`? As extra param?

forge-game/src/main/java/forge/game/card/CardState.java
```diff
@@ -749,7 +751,7 @@ public FCollectionView<ReplacementEffect> getReplacementEffects() {
             result.add(defenseRep);
         }
 
-        card.updateReplacementEffects(result, this);
+        card.updateReplacementEffects(result, this, rulesHost);
```
Hanmac, line 755: do there:
```
if (rulesHost) {
    return result;
}
```

then we can cleanup the extra Read_Ahead check, right?

Agetian (dismissed):

forge-game/src/main/java/forge/game/card/CardState.java
tool4ever: ok, good point

Hanmac (approved):

Hanmac:
another way would be to hide them via "hiddenStatics" that do create ReplacementEffects
But that might be a cleanup for another round

PR #10083, opened 2026-03-17
Rudimentary AI for Station

Agetian:
Not sure how to make it smarter, but it's better than nothing :)


Agetian:
Ugh this is actually pretty bad , I'll rethink this tomorrow morning

PR #10087, opened 2026-03-18
Rudimentary Station AI logic, take two

Agetian:
A very basic AI which will activate the ability when it's reasonably safe to do so (AI is at decent life after the possible attack, or it doesn't have any possible blockers, or there's nothing to block). Needs to be improved, but not sure how. Probably chooseTapType needs to be updated as well so that what exactly is tapped is more carefully selected (and something is kept back).

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -3453,6 +3453,9 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
             if ((sa.isAdventure() || sa.isOmen()) && !getCurrentStateName().equals(sa.getCardStateName())) {
                 continue;
             }
+            if (sa.isLandAbility()) {
+                continue;
+            }
             if (!(sa instanceof SpellPermanent && sa.isBasicSpell())) {
                 return false;
             }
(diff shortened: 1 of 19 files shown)


forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
```diff
@@ -302,6 +302,8 @@ protected AiAbilityDecision checkApiLogic(Player ai, final SpellAbility sa) {
             return doChargeToOppCtrlCMCLogic(ai, sa);
         } else if (logic.equals("TheOneRing")) {
             return SpecialCardAi.TheOneRing.consider(ai, sa);
+        } else if (amountStr.equals("StationX") && source.hasKeyword(Keyword.STATION)) {
```
tool4ever: just check `sa.isKeyword`

Agetian: Nice, thanks! :) Done!

Hanmac:
Maybe: make station more likely if it's a (untapped) Spacecraft? And it becomes a creature?

Agetian:
Yeah, I thought about that tbh, but not sure how to do it... The problem is, it'll need synchronization between what is chosen to be tapped for what Station, and that'll need a rather complex modification to both chooseTapType and willPayCosts that I can't quite figure out :( I'll think about it some more though. May need to do this as a followup PR unless some bright ideas come to mind ^^;

Basically, we'll need to determine that whatever we tap will be enough to get the Station to the value where it becomes a creature.

PR #10098, opened 2026-03-19
Remove duplicate Commander mode selection in Adventure new game screen

Agetian:
Commander mode is already defined for situations where config.json specifies starter Commander decks in the code block above. Thus, this block appears to be superfluous and adds the Commander mode to the selection menu twice (moreover, it would add Commander mode even for worlds that do not specify any commander starters)


Agetian:
Already covered in another PR

PR #10099, opened 2026-03-19
Cache AI combat predictions to reduce timeout during spell evaluation

vanja-ivancevic:
Here's a heavily vibe-coded performance optimization that I think is worth chewing through. Testing is working well so far. I would love to get some feedback and opinions on this, since it's not really my area at all.

- Cache `predictNextCombatsRemainingLife` results on AiController, scoped to each evaluation pass (`chooseSpellAbilityToPlay`, `canPlayFromEffectAI`, `doTrigger`). Within a single evaluation, the 13 callers of `aiLifeInDanger` across DamageDealAi, FogAi, ChangeZoneAi, SpecialCardAi, etc. share cached results instead of each running a full O(N*M) blocking simulation. Only caches when `excludedBlockers == null`.
- Add a short-circuit in `predictNextCombatsRemainingLife`: if total attacker power + life payment < AI life, skip the blocking simulation entirely (AI survives even with zero blocks). Correctly bails out for infect, double strike, and MustBeBlocked creatures.
- Restructure `getDamagingSAToChain` to run cheap validation checks (sub-abilities, divided damage) before the expensive `aiLifeInDanger` call. The danger check now only runs when the random chance check fails and needs to be overridden.
- Cache fields are volatile for thread safety on the FutureTask timeout/cancel path.
- 30 tests validate cached results match uncached across diverse boards: empty, token swarms, flyers vs ground, double strike, first strike, infect, deathtouch, trample, vigilance, shadow, reach vs flying, protection, anthems, life payment boundaries, nested scopes, excluded blockers, opponent's turn, exact lethal boundary.

Benchmarked improvement (AI turn time):
- 15+10 creatures: 1582ms to 313ms (5x)
- 30+20 creatures: 10233ms to 4922ms (2x)
- 20+15 mixed: 1426ms to 362ms (4x)
- 30+15 counters: 8778ms to 92ms (95x)

TODO for future work:
- `AiAttackController` blocker loop (line 423) calls `predictNextCombatsRemainingLife` once per potential attacker with different `excludedBlockers` each time. This cache can't help there. With 30 creatures that's 30 full simulations, which is why 50+30 boards still timeout. Incremental blocking evaluation or a cheaper "is it safe to attack with this creature" heuristic could address this.
- `getDamagingSAToChain` card/ability search iterates all cards in hand, battlefield, and external zones for every damage spell. Caching the search result or pre-filtering candidates per evaluation pass could reduce this overhead.


Agetian:
This is actually very important (due to some calculations taking a lot of time, especially on Android), but needs thorough testing. I'll try to test this a bunch over the weekend and see how far I can get. Otherwise, would need help with reviewing, the more we look through it, the better imo.

forge-gui-desktop/src/test/java/forge/ai/PredictCombatCacheTest.java
```diff
@@ -0,0 +1,563 @@
...
+        aic.beginPredictCombatCacheScope();
+        int cachedNotSerious;
+        int cachedSerious;
+        int cachedNotSerious2;
+        int cachedSerious2;
+        try {
+            cachedNotSerious = ComputerUtil.predictNextCombatsRemainingLife(ai, false, false, 0, null);
+            cachedSerious = ComputerUtil.predictNextCombatsRemainingLife(ai, true, false, 0, null);
+            // These should come from cache
+            cachedNotSerious2 = ComputerUtil.predictNextCombatsRemainingLife(ai, false, false, 0, null);
```
tool4ever: appreciate the initiative but hate that "your" AI floods us with 30+ tests yet confidently claims it would cache these when the call order would obviously invalidate previous results...

besides there is already some support in the engine for this via `AiController.getPredictedCombat()`:
should extend/unify that approach by a more flexible helper class instead of creating parallel systems

vanja-ivancevic: Thank you for the direct feedback! That makes a lot of sense. I'm gonna pass this on from the reworking and go test a bit more in the build.

Reworked the caching to extend the existing `getPredictedCombat()` pattern on `AiController` instead of the parallel scope-based system:

- Two lazy Integer fields, reset alongside `predictedCombat` in `chooseSpellAbilityToPlay()`
- `aiLifeInDanger` routes through cache when `payment == 0;` calls with `excludedBlockers` or non-zero payment always compute fresh
- Removed the scope brackets, volatile fields, depth counter, and test files
- Kept the short-circuit optimization and `DamageDealAi` reorder as standalone improvements
- Tested locally with 4 semantic tests (not committed) that verify: cache returns stale value after board changes, invalidation works on new priority, non-zero payment bypasses cache, and serious/non-serious are cached independently.

tool4ever: Here's a more flexible solution: #10149

vanja-ivancevic: Looks good - works well in testing too. Good speed improvements. Happy to close this PR in favor of #10149 if you'd prefer to go with `AiCache`.

tool4ever: I might cherry pick the DamageDealAi change
the other shortcut I'll need to investigate more deeply later

vanja-ivancevic: Great. Looking forward to seeing it live in action :). Thanks again for your efforts on this.

PR #10103, opened 2026-03-20
Fix Scryfall image fallback skipping requested card's own edition

vanja-ivancevic:
- The set-lookup fallback path skipped the requested card's own print and only tried alternate editions, which were not filtered by allowedEditions, resulting in wrong-frame art (e.g. 9ED art for an S00 card)
- Always try the requested card's Scryfall URL first before falling back to other prints
- In Adventure mode, filter fallback candidates through allowedEditions when configured

### DIFF
--- a/forge-gui-mobile/src/forge/util/LibGDXImageFetcher.java
+++ b/forge-gui-mobile/src/forge/util/LibGDXImageFetcher.java
@@ -2,7 +2,10 @@ package forge.util;
 
 import com.badlogic.gdx.files.FileHandle;
 import forge.Forge;
+import forge.adventure.data.ConfigData;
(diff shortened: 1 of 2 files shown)


Hanmac:
IMO we should check out if we REALLY need `hasSetLookup` anymore or if we can remove them

vanja-ivancevic:
Agreed it's worth revisiting, but that would be a larger refactor across ImageKeys, StaticData, and RewardActor. I just quickly wanted to fix the issue that's allowing new border cards to creep in to the old border shandalar.

vanja-ivancevic:
I mean to say, I don't know the history, etc. and how removing it might affect edge cases.

PR #10105, opened 2026-03-20
Adventure Mode: allow/disallow genetic AI decks on a per-world basis

Agetian:
Allows to specify an option in config.json that disallows genetic AI (e.g. needed for Old Border Shandalar because all genetic AI decks in Forge are very modern new border decks)


Agetian:
Taken care of in another PR

PR #10107, opened 2026-03-21
Old Border Shandalar: Add enableGeneticAI config option, remove Standard/Constructed modes

vanja-ivancevic:
- Add `enableGeneticAI` boolean to `ConfigData` (defaults `true` for backward compatibility)
- Check the flag in `DuelScene` for arena battles, Hard/Insane enemies, and missing deck fallback
- Set `enableGeneticAI: false` in Old Border Shandalar config.json to prevent modern card decks from replacing curated old-border enemy decks on higher difficulties
- Remove `starterDecks` and `constructedStarterDecks` from Old Border difficulties (replaced by the new Precon starter mode)

On Hard+ difficulty, genetic AI and LDA deck generation would give enemies Modern/Legacy/Standard competitive decks with 2020+ cards, completely breaking the old-border-only experience. Arena battles were also hardcoded to always use genetic AI decks regardless of difficulty. This adds a per-world config option so realm modders can opt out.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/data/ConfigData.java
+++ b/forge-gui-mobile/src/forge/adventure/data/ConfigData.java
@@ -29,4 +29,5 @@ public class ConfigData {
     public String[] restrictedEvents;
     public String[] allowedEvents;
     public String[] allowedJumpstart;
+    public boolean enableGeneticAI = true;
 }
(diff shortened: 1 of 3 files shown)


Jetz72:
Late response because I didn't realize I was on page 3 of the github when I first saw this, but wouldn't it make sense to have the genetic AI toggle be a setting enabled in the hard and insane difficulty definitions of the config, rather than a single global one for the plane?

Agetian:
> Late response because I didn't realize I was on page 3 of the github when I first saw this, but wouldn't it make sense to have the genetic AI toggle be a setting enabled in the hard and insane difficulty definitions of the config, rather than a single global one for the plane?

This would definitely make it more granular and easier to configure if someone wants to have the genetic AI on Insane but not on Hard, don't think we currently have a case for that in the developed worlds, but in the future it could be useful, yeah. Might be worth changing to that as time permits.

PR #10116, opened 2026-03-21
CardState: fix Printed PT on Vehicle Backside

Hanmac:
fixes Spring-loaded Sawblades not displaying PT when not animated


Agetian:
:+1:

PR #10124, opened 2026-03-21
Add auto-yield until end of turn

calebgmahala:
## Summary
- Adds an "Auto-Yield Until End of Turn" feature that automatically passes priority for the remainder of the current turn, similar to F6 in MTGO
- Includes keyboard shortcut support and localization strings for all supported languages

## Questions
- How are we managing localization's? Information on this should probably be documented alongside the translation files
- What should the default keybinds be for this?
- When auto-yielding, the player will skip their own attack declaration and pass with no attackers. The existing auto-pass has this same behavior. On MTGO, auto-pass does not skip over your attackers step, but yield until end step does. Do we have a preference for this? My thought is that we should mimic MTGO (To do this we should log a separate issue for changing the `autoPassUntilEndOfTurn` to stop in attackers)

## Testing Plan
- [ ] Passes priority through all opponent stack items
- [ ] Cancels in cleanup
- [ ] Cancels after match concludes
- [ ] Skips over players combat when no combat restrictions (Same as auto pass until end of turn)
- [ ] Stops in players combat when player has combat restrictions (Same as auto pass until end of turn)
- [ ] Skips over opponents combat when no valid blockers (Same as auto pass until end of turn)
- [ ] Stops in opponents combat when player has valid blockers (Same as auto pass until end of turn)
- [ ] All of the above works over local net play (2 instances on same machine connected via localhost
- [ ] All of the above works over internet net play (2 separate machines on different networks)

### DIFF
--- a/docs/User-Guide.md
+++ b/docs/User-Guide.md
@@ -237,11 +237,14 @@ Underneath each player's battlefield portrait are zone buttons (hand, library, g
 **Display modes:** Right-click a zone button to toggle how it opens:
 
 - **Open in Window** — Opens a floating window that can be moved and resized independently. Right-click the window's title bar to toggle alphabetical sorting.
-- **Add Tab to Hand Panel** — Docks the zone as a tab alongside your hand panel, integrated into the main layout. The tab can be dragged to other cells like any other panel.
+- **Add Tab to Hand Panel** — Docks the zone as a tab alongside your hand panel, integrated into the main layout.
 
 Your preference is saved per zone and remembered separately for your own zones and opponent zones.
 
 **Drag-to-dock:** A floating zone window can also be docked by dragging its title bar over any cell's tab bar — a blue highlight shows where it will land. Once docked, right-click the tab for an option to undock it back to a floating window.
+
+> [!TIP]
+> Enable the Layout->View->New card count menu option to display a delta number in the tab header which may help you more quickly identify changes in complex boardstates.
 
 ## Auto-Sort Multiplayer Fields
 In multiplayer games (3+ players), Forge can automatically arrange opponent battlefield panels in turn order. This is controlled via the **Layout > View** menu during a match.
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
@@ -657,7 +657,7 @@ public class ComputerUtilCost {
 
         if (root.costHasManaX()) {
             val = ComputerUtilMana.determineLeftoverMana(root, ai, effect);
-
+            // TODO already calculate once before previous method to avoid running more expensive checks above this highest possible maximum
             if (sa.hasParam("AIMaxTgtCost")) {
                 String value = sa.getParam("AIMaxTgtCost");
                 String svar = source.getSVar(value);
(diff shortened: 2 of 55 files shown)


forge-gui-mobile/src/forge/screens/match/MatchScreen.java
```diff
@@ -650,6 +650,9 @@ public boolean keyDown(int keyCode) {
                     return true;
                 }
                 break;
+            case Keys.NUM_6: //yield until end of turn
```
calebgmahala, line 653: I'm not sure what we want this to be. Looking for suggestions

forge-gui/res/languages/en-US.properties
calebgmahala, line 1: Unsure how we handle translations. I had claude construct the translations for the other languages. Please let me know if we have a specific process for managing translations (we may want to add a readme / agents file so that changes can be made correctly)

forge-gui/src/main/java/forge/localinstance/properties/ForgePreferences.java
```diff
@@ -296,6 +296,7 @@ public enum FPref implements PreferencesStore.IPref {
         SHORTCUT_UNDO ("17 90"),
         SHORTCUT_CONCEDE ("17 81"),
         SHORTCUT_ENDTURN ("17 69"),
+        SHORTCUT_YIELD_UNTIL_EOT ("54"), // 6
```
calebgmahala, line 299: Similar to the above comment. I'm unsure what we want the default hotkey for this to be

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
calebgmahala, line 2409: I really don't like this method. I feel like we should split it out into pass and auto pass. Just my personal opinion :)

forge-gui/src/main/java/forge/gui/interfaces/IGuiGame.java
```diff
@@ -263,12 +263,23 @@ default void handleGameEvents(List<GameEvent> events) {
 
     void autoPassUntilEndOfTurn(PlayerView player);
 
-    boolean mayAutoPass(PlayerView player);
+    /**
+     * Returns true if the player should automatically pass priority,
+     * whether due to auto-pass (cancelled by opponent spells) or
+     * auto-yield (persists through opponent spells until end of turn).
+     */
+    boolean shouldAutoPassPriority(PlayerView player);
```
calebgmahala, line 271: I renamed this method. I think this aligns more with what it is actually doing.

It's used both in the pre-existing autoPassUntilEndOfTurn and in the new autoYieldUntilEndOfTurn

forge-gui/src/main/java/forge/localinstance/properties/ForgePreferences.java
calebgmahala: Also, I think it would be beneficial to add same line comments describing what the decimal key code maps to

tool4ever:
we already have a PR for this which will be more advanced: #9643

calebgmahala:
> we already have a PR for this which will be more advanced: #9643

oh fantastic! I'll take a peek at it then

Issue #10076, opened 2026-03-17
Counter Type Meta Issue


Hanmac:
- [x] use Translatable Interface for better name. Might Translate them later in the Future.
- [x] Icon for Keyword Counter used in Choice Dialog.
- [ ] Static ability for +X/+Y Counter, because dependency: #5642
  - [ ] Allow Custom variants?
- [ ] Use List File for Custom Counters, might move some Enum Counters there that aren't used in Code
  - [ ] Add Icons for Custom Counters, examples there: https://mtg.wiki/page/User:RivalRowan/MTGO_counters

The Name only really has two modes, might just generate directly:
- P1P1 to `+1/+1` for PTCounter
- Capitalize for anything else.

Types of CounterTypes:
* `CounterEnumType`: the ones that are REALLY Hard Coded in the Java Code (has Color)
* `CounterCustomType`: the ones that will be defined in a List like CreatureTypes. (has Color)
* `CounterKeywordType`: the ones that are Keywords by the rules, and fallback to anything else. (has Keyword Icon)
* `CounterPTType`: the ones that are like P1P1, currently EnumType

Possible CounterCustomType List:
`KeyName=Desc,R,G,B`

```
CAGE=CAGE,155,155,155
```
(Dialog-Name will be Cage)

----

For Cleanup and Custom Counters: List of Counters that are used in Code like P1P1 and M1M1, separate them between Rules Code and AI code:

Rules:
  - SHIELD, STUN and FINALITY Counter for their Replacement Effects
  - MANABOND for their Static
  - DEFENSE for Battle/Siege
  - LORE for Sagas
  - LOYALTY for Planeswalker
  - DREAM for some (not more than 7) rules
  - TIME for TimeTravel.
  - POISON, ENERGY, EXPERIENCE, RAD (and Ticket) for Player

AI:
  - QUEST for Proto-Sagas (and some legendary?)
  - FADE for Fading
  - TIME for Vanishing
  - CHARGE for Station
  - KI for some Legendary
  - POISON, ENERGY, EXPERIENCE, RAD and Ticket for Player
  - ICE for darkDepths and Storage Lands
  - CAGE for MairsilThePretender
  - BURDEN for TheOneRing
- AI uses some counter types in the `isNegativeCounter` and `isUselessCounter` functions. Helper might be enough for these cases?

Jetz72:
> Static ability for +X/+Y Counter
> 
>   Allow Custom variants?

[...]

PR #10136, opened 2026-03-23
DevCheats: make CounterType Lists dynamic

Hanmac:
Part of #10076

This makes PlayerControllerHuman not depend on `CounterEnumType` and have it return all fixed `CounterKeywordType` and all other types added there.

The returned list might be extended with other `CounterType` variants.

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CounterEnumType.java
+++ b/forge-game/src/main/java/forge/game/card/CounterEnumType.java
@@ -18,8 +18,6 @@
 
 package forge.game.card;
 
-import com.google.common.collect.ImmutableList;
-
 import java.util.Locale;
 
 /**
@@ -541,11 +539,6 @@ public enum CounterEnumType implements CounterType {
     public static CounterEnumType getType(final String name) {
         final String replacedName = name.replace("/", "").replaceAll("\\+", "p").replaceAll("\\-", "m").toUpperCase(Locale.ROOT);
         return Enum.valueOf(CounterEnumType.class, replacedName);
-    }
-
-    protected static final ImmutableList<CounterEnumType> values = ImmutableList.copyOf(values());
-    public static ImmutableList<CounterEnumType> getValues() {
-        return values;
     }
 
     @Override
--- a/forge-game/src/main/java/forge/game/card/CounterType.java
+++ b/forge-game/src/main/java/forge/game/card/CounterType.java
@@ -19,7 +19,7 @@ public interface CounterType extends Serializable {
     }
     static List<CounterType> getValues() {
         List<CounterType> result = Lists.newArrayList();
-        result.addAll(CounterEnumType.getValues());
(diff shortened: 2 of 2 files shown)


forge-game/src/main/java/forge/game/card/CounterEnumType.java
```diff
@@ -543,7 +543,10 @@ public static CounterEnumType getType(final String name) {
         return Enum.valueOf(CounterEnumType.class, replacedName);
     }
 
-    public static final ImmutableList<CounterEnumType> values = ImmutableList.copyOf(values());
+    protected static final ImmutableList<CounterEnumType> values = ImmutableList.copyOf(values());
+    public static ImmutableList<CounterEnumType> getValues() {
+        return values;
+    }
```
Hanmac: @tool4ever what would be better?
it gets added to a dynamic List in `CounterType::getValues` anyway?

tool4ever: you don't need to worry:
the JVM knows `values()` should not be backed by the real enum

(though I guess technically ImmutableList enforces the idea)

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -2709,8 +2709,8 @@ public void modifyCountersOnPermanent(boolean subtract) {
             }
             final Card card = gameCacheCounters.get(cv);
 
-            final ImmutableList<CounterType> counters = subtract ? ImmutableList.copyOf(card.getCounters().keySet())
-                    : ImmutableList.copyOf(CounterEnumType.values);
+            final List<CounterType> counters = subtract ? ImmutableList.copyOf(card.getCounters().keySet())
```
tool4ever, line 2712: maybe use ordered Set so you can add existing ones first?

forge-game/src/main/java/forge/game/card/CounterEnumType.java
Hanmac: The bigger question was more if `CounterEnumType::getValues` is needed anyway,
if i just add them only in `CounterType::getValues`
and this function is only called for the PlayerControllerHuman?

So just call `result.addAll(Arrays.asList(CounterEnumType.values()));` and skip the ImmutableList?

> Arrays.asList returns a mutable list while the list returned by List.of is [immutable](https://docs.oracle.com/javase/9/docs/api/java/util/List.html#immutable)

Or a normal `List.of` is fine too?

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
Hanmac: you mean to comment this to CounterKeywordType right?

Agetian:
Looks good to me :)

PR #10137, opened 2026-03-23
Add AI logic for Rogue's Passage and similar cards

Agetian:
Might be a bit slow on busy battlefields due to having to call doesCreatureAttackAI, but tries to postpone that particular call as the last resort option, even after calling doesSpecificCreatureAttackAI, which only checks one particular creature (the unblockable copy).

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/EffectAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/EffectAi.java
@@ -330,10 +330,16 @@ public class EffectAi extends SpellAbilityAi {
                         if (!CombatUtil.canBeBlocked(card, ai.getOpponents().getCreaturesInPlay(), phase.getCombat())) {
                             continue;
                         }
+                        if (card.getNetPower() >= ai.getWeakestOpponent().getLife() && ai.getWeakestOpponent().canLoseLife() && !ai.getWeakestOpponent().cantLoseForZeroOrLessLife()) {
+                            // try to finish off the opponent with an unblockable creature
+                            sa.getTargets().add(card);
+                            return new AiAbilityDecision(100, AiPlayDecision.WillPlay);
+                        }
                         final Card copy = CardCopyService.getLKICopy(card);
                         String cantBeBlocked = "Mode$ CantBlockBy | ValidAttacker$ Creature.Self";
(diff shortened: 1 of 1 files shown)


forge-ai/src/main/java/forge/ai/ability/EffectAi.java
Hanmac, line 447: @Agetian can you check out if you can get this Generic Section working?
If yes, we don't need extra AI logic

(might split cantBlock and cantBlockBy)

Agetian: I'll take a look ☺️

Agetian: Hmm, I'm not sure if putting this kind of logic into the generic pool is viable tbh, I mean, it's physically possible, but it will require running those combat checks that are rather expensive, so they will still need to be gated behind the AILogic flag I think... For now, I'd say a better (less expensive/faster) approach would be a targeted logic, but if at some point we optimize the combat simulation enough that it doesn't bog the game down, we can generalize this

Agetian: Maybe we can avoid the calls if we somehow reliably detect this particular ability, but I can't quite figure out how :(
I think I'll merge this in for now, since it's considerably better than AILogic Pump for the occasion, but if you can help later by figuring out a more generic approach that won't need AILogic at all, I'd really appreciate it :)

Agetian: Currently the generic logic bails at failing to see whether the given creature will attack (canAttack = false, with other params also false) - at that point, to see if it's OK to activate this ability, we'd need to run the combat simulation to determine that it will attack given unblockable

Hanmac: i might refactor the generic logic later

Agetian: Sure thing, always appreciate your help! :+1: 
I added some additional considerations, maybe those would be possible to include in the more generic approach (e.g. dealing lethal with an unblockable, maybe avoiding unfavorable trades somehow)

Issue #9918, opened 2026-02-25
KeywordView Records


Hanmac:
add KeywordView Records with only String parameters
so KeywordInstances can be turned into Views and Stored in CardStateView

this might help other stuff like #9806

Hanmac:
@MostCromulent would such records in the CardStateView help with your Card Info Tooltip?

MostCromulent:
@Hanmac - AI analysis is it's an architectural improvement but not functionally essential:

> #### The tooltip in # 9806 follows this flow:
> 
> 1. Gets `CardStateView.getKeywordKey()` -> a comma-separated string like `"Flying,Haste,Protection from Blue"`
> 2. **Re-parses** each token back into a `KeywordInterface` via `Keyword.getInstance(token)` in `KeywordInfoUtil`
> 3. Extracts `getTitle()`, `getReminderText()`, type params from those instances
> 4. Creates `KeywordData` objects (a simple value class with `name`, `reminderText`, `typeParam`)
> 5. Does additional processing: count annotation (Devotion, Affinity, Domain), keyword action extraction from oracle text, deduplication, sorting
> 
> #### Would KeywordView Records help # 9806?
> 
> **Short answer: marginally, and it's not a prerequisite.**
> 
> **Benefits:**
> - Eliminates the round-trip of "keyword ->string -> re-parse to keyword ->extract display data." With records in CardStateView, structured data would be directly available
> - For **network play**, the client currently can't extract reminder text without the full keyword parsing infrastructure. Records would provide that data pre-computed
> - PR # 9806 already has `KeywordData` which is functionally identical to what a `KeywordView` record would be - this validates the concept
> 
> **Limitations:**
> - The current `Keyword.getInstance()` re-parsing works reliably; the records would be a cleanliness improvement, not a bug fix
> - Most of # 9806's complexity is in **post-processing** that happens after keyword extraction: count annotation (devotion/affinity/domain counts), keyword action scanning from oracle text, deduplication/merging, conjugation matching etc. None of that goes away with records.
> - Adding records to `CardStateView` as trackable properties means touching the delta sync serialization layer (`TrackableTypes`) where keyword serialization is currently a **TODO stub** - that's non-trivial plumbing
> 
> #### Priority assessment
> 
> | Factor | Rating | Notes |
> |--------|--------|-------|
> | Benefit to # 9806 | Low | Already works fine with string parsing |
[...]

PR #10153, opened 2026-03-24
CounterKeywordType: add Icon

Hanmac:
Part of #10076 

- [x] Cleanup CounterEnum Constructor
- [x] makes CounterType translatable with `getUntranslatedName` → `toString`
- [x] Adds Icons for Keyword Enum + Str (that can be used for other Keyword Icons later too)
- [x] CounterKeywordType now stores the Keyword Enum too. (this could be cleanup later with #9918)
- [x] Desktop Icons are added.
- [x] Mobile needs extra Renderer yet.

Maybe:
- [x] Fix the indent of elements without Icon. For Desktop, I might need to create an extra empty Image/Icon)
- [ ] use https://docs.oracle.com/javase/8/docs/api/java/awt/Color.html as Foreground Color for the TextColor

Current state on Desktop:
Bildschirmfoto vom 2026-03-24 21-32-25

### DIFF
--- a/forge-gui-mobile/src/forge/assets/FSkinImageInterface.java
+++ b/forge-gui-mobile/src/forge/assets/FSkinImageInterface.java
@@ -4,8 +4,8 @@ import com.badlogic.gdx.graphics.Pixmap;
 import com.badlogic.gdx.graphics.g2d.TextureRegion;
 
 public interface FSkinImageInterface extends FImage {
-	public void load(Pixmap preferredIcons);
-	public TextureRegion getTextureRegion();
+    public void load(Pixmap preferredIcons);
+    public TextureRegion getTextureRegion();
     public float getNearestHQWidth(float baseWidth);
     public float getNearestHQHeight(float baseHeight);
 }
--- a/forge-gui-mobile/src/forge/toolbox/FChoiceList.java
+++ b/forge-gui-mobile/src/forge/toolbox/FChoiceList.java
@@ -29,6 +29,8 @@ import forge.card.CardZoom.ActivateHandler;
 import forge.card.MagicColor;
 import forge.card.mana.ManaCost;
 import forge.game.card.CardView;
+import forge.game.card.CounterKeywordType;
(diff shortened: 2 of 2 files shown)


forge-gui-desktop/src/main/java/forge/gui/ListChooser.java
```diff
@@ -409,6 +410,11 @@ public Component getListCellRendererComponent(final JList<? extends T> list, fin
             if (value instanceof CardType.CoreType c) {
                 defRenderer.setIcon(fromSkinProp(FSkinProp.iconFromCoreType(c)));
             }
+            if (value instanceof CounterKeywordType c) {
+                if (c.isKeywordCounter()) {
+                    defRenderer.setIcon(fromSkinProp(FSkinProp.iconFromKeyword(c.type(), c.keyword())));
+                }
+            }
             return result;
         }
```
Jetz72, line 423: Could we extract a common interface here? `IHasIcon` or something? I think I looked at interfaces in conjunction with `FSkinProp` in the past but I don't recall the exact context.

Hanmac: There is `IHasSkinProp` but that doesn't work for them because the Icons and the Interface are defined in forge-GUI, while Color and the others are defined in forge-Core or forge-Game

Jetz72: Ah right, that was the issue. Could possibly work around that by having core only supply an icon key/ID which gets translated as an actual icon by `FSkinProp`. Wouldn't be the only place where the core module supplies UI-oriented info. 

But on the other hand, between `FSkinProp`, and the mobile UI's `FSkinImage` and `CardFaceSymbols`, we already have a handful of lists of definitions/constants representing individual images. Adding another one into the chain might be a bit excessive.

Hanmac:
The `iconFromKeyword` function can be later extended for the GUI to use for the Keyword Icons too
(if #9918 would be available)

PR #10155, opened 2026-03-24
Prompt for token owner when running the "Add Token to Battlefield" action in dev mode menu.

### LINKED ISSUE #10154: The "Add Token to Battlefield" dev mode action always adds tokens under your control

In my [fan-made custom set for Expedition 33](https://github.com/jumpinjackie/forge-e33-mod) we have a mechanic called Luminous which is like Metalcraft, but counts a certain type of token instead of counting artifacts.

Testing Luminous cards in this set from an AI opponent's context (to verify the Luminous bonus is kicking in when certain cards are resolved by the AI opponent) has been a bit cumbersome because currently the "Add Token to Battlefield" dev mode menu always adds tokens to the battlefield *under your control*. 

Things would've been simpler if the "Add Token to Battlefield" action prompted for the token owner, the same way the "Add Card to Battlefield" action does.

Currently, the workaround of setting up token board state for the opponent has been an awkward process of:

 1. Adding the tokens to the battlefield (on your side)
 2. Add X copies of [Donate](https://scryfall.com/card/uds/31/donate) or [Harmless Offering](https://scryfall.com/card/fdn/625/harmless-offering) to your hand.
 3. Resolve however many copies of Donate/Harmless Offering to donate the required number of tokens to the opponent.

**Additional context**

jumpinjackie:
This PR updates the "Add Token to Battlefield" dev mode action to prompt for the token owner first.

Prompt logic taken from the handler of the "Add Card to Battlefield" action.

Non-english string translations are a "best guess" autocompletion from copilot.

Fixes #10154

### DIFF
--- a/forge-gui/res/languages/de-DE.properties
+++ b/forge-gui/res/languages/de-DE.properties
@@ -1448,6 +1448,7 @@ lblFileNotFound=Datei nicht gefunden
 lblPutCardInWhichPlayerZone=Lege Karte bei welchem Spieler in {0}?
 lblPutCardInWhichPlayerBattlefield=Lege die Karte bei welchem Spieler auf das Spielfeld?
 lblPutCardInWhichPlayerPlayOrStack=Die Karte soll für welchen Spieler ins Spiel/auf den Stapel?
+lblPutTokenInWhichPlayerBattlefield=Lege Token bei welchem Spieler auf das Spielfeld?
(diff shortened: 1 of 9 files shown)


Hanmac:
imo you also could have reused `lblPutCardInWhichPlayerBattlefield`
🤷

jumpinjackie:
Didn't know how pedantic the card vs token distinction needed to be. Played it safe.

Hanmac:
I'm not against it really, i myself just try to avoid extra language strings if able

PR #10158, opened 2026-03-24
Large adventure (main) update

Simisays:
- First implementation of the card price list (disabled in other planes) for cards and boosters
- First implementation of wins based enemy spawn
- new enemies
- Castle overhaul, with new item rewards, new look and various fixes and rebalances
- Possibility to make rewards and enemies spawn only on insane
- Possibility for items to start in hand at the beginning of the game
- Kiora and cheff's boss overhaul
- Fix for the maps in old border world not loading in Tiled program
- reorganisation of the beast sprite maps
- removal of unused decks and sprites
- expanded tileset
- Possibility for rewardchoice (currently only used for the mox reward) 
- New contraption and attraction shops (+ new sprite)
- Enemies will now move away from you after certain wins


Agetian:
It looks like the build is failing for some reason, might look into it tomorrow morning (could be as simple as some extra imports)

Simisays:
Yeah I think something messed up with merging with main branch. I think it's fixed

forge-gui-mobile/src/forge/adventure/data/BiomeData.java
```diff
@@ -100,20 +102,32 @@ public EnemyData getExtraSpawnEnemy(float difficultyFactor) {
     }
 
     public EnemyData getEnemy(float difficultyFactor) {
-        //todo: implement difficultyFactor
         float totalDistribution = 0.0f;
-        for (EnemyData data : enemyList) {
-            totalDistribution += data.spawnRate;
+        difficultyFactor = Current.player().getStatistic().rank();      //// conpare difficulty data to on how many wins you have on your save
```
Agetian, line 106: Comment tweak:

// compare difficulty....

forge-gui-mobile/src/forge/adventure/player/PlayerStatistic.java
```diff
@@ -30,6 +30,23 @@ public int totalWins()
         }
         return wins;
     }
+    public float rank()   ///// used for overworld enemy spawn
```
Agetian, line 33: Comment tweak:

// used for ...

forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
```diff
@@ -294,17 +294,22 @@ void addEffects(RegisteredPlayer player, Array<EffectData> effects) {
         int extraManaShards = 0;
         Array<IPaperCard> startCards = new Array<>();
         Array<IPaperCard> startCardsInCommandZone = new Array<>();
+        Array<IPaperCard> startCardsInHand = new Array<>();
+
```
Agetian, line 298: Extra empty line

forge-gui-mobile/src/forge/adventure/scene/DuelScene.java
```diff
@@ -294,17 +294,22 @@ void addEffects(RegisteredPlayer player, Array<EffectData> effects) {
...
 
         for (EffectData data : effects) {
             lifeMod += data.lifeModifier;
             changeStartCards += data.changeStartCards;
             startCards.addAll(data.startBattleWithCards());
             startCardsInCommandZone.addAll(data.startBattleWithCardsInCommandZone());
+            startCardsInHand.addAll(data.startBattleWithCardsInHand());
+
```
Agetian, line 306: Extra empty line

Agetian:
Overall looks good to me, great job! I'll tweak the things I noticed in a followup PR

Agetian:
Hmm, it looks like for some reason, this PR affects e.g. Old Border Shandalar saves - when I load a saved game from earlier, it tells me that data migration is complete, but I lose the entire inventory and all the equipment after the migration. Do you know why? Is there any way to quickly patch this?

Agetian:
Looks like it has to do with this exception related to the new field, not yet sure how to fix:

[Invalid Class Exception]
Overriding serialized class version mismatch: local serialVersionUID = 3793811462061217889 stream serialVersionUID = -1350202758899129612
java.lang.ClassCastException: cannot assign instance of java.lang.String to field forge.adventure.data.EffectData.startBattleWithCardInHand of type [Ljava.lang.String; in instance of forge.adventure.data.EffectData

Agetian:
Ugh, probably needs some kind of a custom deserialization routine in order to load the older saves?.. :/ I'm not sure tbh

Agetian:
I tried to fix the items not deserializing correctly and the inventory becoming empty on data migration.
Like, if I modify EffectData.java to use the old serial ID, it'll deserialize correctly and load the inventory and stuff, but I'm not sure if jury-rigging the ID is the right way to go, and a more complex solution is definitely outside my area of expertise :/  Do you have any ideas, @tool4ever @Simisays @Jetz72 ?

Here's what works (at the top of EffectData):

    @Serial
    private static final long serialVersionUID = -1350202758899129612L;

But this seems like an ugly hack to me.

Agetian:
Here's the "simple but ugly(?)" solution:
https://github.com/Card-Forge/forge/pull/10163

I noticed that we do set serial IDs in many places in the code, so maybe this actually works? :/

Jetz72:
Setting the ID might be a viable solution, but at the same time this was waaaaay too short notice to merge a PR of this scale. Haven't yet had time to look it over. How confident are we that nothing else is broken on previous saves? Still time to revert before snaps fire if we want to do some more testing...

Agetian:
I also thought about it, but don't think a full revert is warranted at the moment tbh (I may be wrong).
When it comes specifically to the structure of the saved game, there's only one field added to EffectData, which is what was causing the deserialization error above. I don't see anything else in the code that would specifically break the saves, and I've been testing multiple saves (Old Border, Realm of Legends) on different levels of advancement for the entire day, no issues so far, neither on load nor further down the road. Of course, additional investigation is highly welcome, as this commit is way too large to go over.

Main world could potentially be more troublesome as this update changes a lot about it - not necessarily save-related, just in general - although I didn't notice any oddities in my experiments (tried a few saves dating back to 2023-2025, the oldest ones required actual data migration, and they preserved the decks and the inventory after this fix). That said, if there's something in here that might necessitate a restart, then it'll prolly be something about the main world (and in that case, it's most likely going to be warranted due to the scale of the changes).

Thus, when it comes to testing:
- Multiple tests seem to indicate this does not break Old Border Shandalar or Realm of Legends saves, allowing to continue the old saves seamlessly, with progress preserved (decks, cards, inventory, selected equipment).
- Multiple tests seem to indicate this allows to continue the main Shandalar adventure, with saves dating back to as far as 2023, preserving cards, decks, and inventories to the best of my knowledge.
- I've been playing Old Border Shandalar and standard Shandalar after this change on two different saves (post-upgrade) for about 6-7 hours today and haven't experienced any oddities or crashes.

That's about as far as I could take it.
I don't think there's anything immediately apparent that necessitates a revert. If anything does happen and we start having reports about things crashing or not loading saves correctly, we can revert on notice imo. The saves are not automatically overwritten and borked - if anything is off, you're basically fine until you overwrite the save. I suggest we warn the Adventure players about potential issues and ask them to report anything odd that might happen (and *not* overwrite their saves if it does), and we can react accordingly. I simply don't see how else we can playtest the change of this level thoroughly enough, and we can't really delay it for too long, as it affects files in other worlds too (Old Border and Realm of Legends), which may create tough PR conflicts further down the road if we keep the PR floating around for long enough. Imo, being reactive (seeing if there's anything subtly broken and patching it up quickly if possible - and if it turns out to be impossible, then reverting and returning this PR back to the drawing board) is more productive. This is just my $0.02 based on my testing, of course. Any other opinions are welcome.

Jetz72:
Fair enough. I'd still like to take a look at generalizing the card-starting-in-zone systems sometime, and I haven't had time to thoroughly look over the card price stuff, but as long as saves aren't broken I suppose any other issues are fixable.

Agetian:
Ok, fingers crossed... I'll post a message in Discord channels about this.

Issue #8718, opened 2025-09-15
Put Edition files in a .zip archive; add version checking to zipped resources


Jetz72:
Users frequently update Forge by installing a new version overtop their old one. Merging the resource directory in this way can cause issues if files have been renamed or removed from directories that parse every file inside them. In addition, users frequently neglect to merge or overwrite certain directories while doing this, meaning the versions of their engine and resources end up mismatched.

The `editions` directory is a common victim of this, with users missing certain editions or indefinitely retaining old copies of renamed ones. The `cardsfolder` is more resistant to this user behavior, since it behaves as a single file to be replaced in its entirety. Applying this behavior to card editions would mitigate potential user error.

What would go a step further though is to add metadata to these archives, containing the Forge version number they were distributed with. This would likely be in the form of a `version.txt` file included in the file (or maybe a `YOU DO NOT NEED TO EXTRACT THIS.txt` file if we want to resolve another potential user misunderstanding at the same time). Forge could check this file and display a warning message on startup if it doesn't match the current version of Forge.

PR #10162, opened 2026-03-25
Hide incomplete planes in Adventure mode until they're further developed / finished

Agetian:
Alternatively, these can be hidden behind a dev mode flag, but I think this should be good for the time being.

If any of them are de facto abandoned and have no hope of ever being finished, they can be removed. Unless it's established for sure, probably it's best to keep them hidden (they don't take much space anyway) so that they can be picked up and continued by the original developer or by another interested party in the future.

### DIFF
--- a/forge-gui/res/adventure/Realm of Legends/world/items.json
+++ b/forge-gui/res/adventure/Realm of Legends/world/items.json
@@ -1039,7 +1039,7 @@
     "effect": {
         "lifeModifier": -6,
         "startBattleWithCard": [
-        "NirkanaCutthroat|ROE"
+        "Nirkana Cutthroat|ROE"
      ]
     }
   },


Jetz72:
If anyone updates their game by extracting the new files to the same directory (as many people do), they'll end up with both copies in their files and the old ones won't be hidden. Any work-in-progress planes we stash in the "incomplete" directory won't be possible for devs to test unless they're moved outside of it. Showing them in dev mode would be better but I'm not sure shoving them in a separate directory is a good solution.

Agetian:
Hmm yeah, agreed, I might redo this in a way that allows to hide them with the ready flag

Jetz72:
Easiest short-term solution would be just to hard-code a blacklist of plane names in the files, and only load those if dev mode is on. Might also have to handle the case where someone is already on one of these planes when they start up the newly updated version of Forge?

Agetian:
> If anyone updates their game by extracting the new files to the same directory (as many people do), they'll end up with both copies in their files and the old ones won't be hidden.

Btw, this is a very bad upgrade practice that we should absolutely discourage imo, I mean, a lot of files are shuffled around, especially these days, in the Adventure folder (with map, item, and enemy deck changes - some decks are added, some are removed, etc.). Upgrading by just overwriting an existing installation without even deleting the older "res" folder is definitely going to get someone in trouble ^^;

Jetz72:
It is, and it happens every now and then when someone has a broken outdated edition file kicking around. We could reduce the potential issues somewhat by archiving portions of the resource files (#8718), but for now it's something we have to keep in mind when renaming or moving resource files.

PR #10163, opened 2026-03-25
Fix migrating Adventure data losing inventory after the recent main world update

Agetian:
This works as an (ugly hack type of?) a solution so that the inventory data is not lost on load. If a more comprehensive solution is needed, it's outside of my area of expertise, so help is welcome...

We do set serial IDs in many places in the code though, so maybe this is viable?

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/data/EffectData.java
+++ b/forge-gui-mobile/src/forge/adventure/data/EffectData.java
@@ -6,6 +6,7 @@ import forge.item.PaperCard;
 import forge.item.PaperToken;
 import forge.model.FModel;
 
+import java.io.Serial;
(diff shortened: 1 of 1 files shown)


Agetian:
Tested it with a few Old Border saves and it worked correctly in every instance.

Agetian:
Tested this extensively today on multiple saves, doesn't seem to cause any issues but allows to load (and then save correctly) the older saved game, thus upgrading it to correctly have the updated property set. Will merge this in - if there's a better solution, we can revert this and implement a different one later, specifically after letting the players upgrade their saves (otherwise there will be lots of reports later on the updated snapshot that the old saves stopped working / are missing items)

PR #10176, opened 2026-03-26
Key cards

tehdiplomat:


Agetian:
AI-wise this looks good to me

forge-gui-desktop/src/main/java/forge/screens/deckeditor/controllers/ACEditorBase.java
```diff
@@ -525,6 +525,77 @@ private void addItem(final String verb, final String dest, final boolean toAlter
...
+            java.util.List<String> keyCards = new java.util.ArrayList<>(deck.getKeyCards());
+            String cardName = card.getName();
+            
+            boolean isKeyCard = false;
+            for (String keyCardName : keyCards) {
+                if (keyCardName.equalsIgnoreCase(cardName)) {
+                    isKeyCard = true;
+                    break;
+                }
+            }
```
Hanmac: I don't like the IgnoreCase thing, I wish that the key list would just have forced correct case

forge-core/src/main/java/forge/deck/Deck.java
```diff
@@ -594,6 +594,17 @@ public String getAiHint(String name) {
...
+        String hint = getAiHint("KeyCards");
+        if (hint == null || hint.isEmpty()) {
+            return List.of();
+        }
+        return List.of(hint.split(";")).stream()
+                .map(String::trim)
+                .filter(s -> !s.isEmpty())
+                .collect(Collectors.toList());
+    }
+
```
Hanmac, line 625: Add "addKeyCard"/"removeKeyCard" directly into Deck class, so the all the other classes doesn't have to handle to set it themselves.

(and it's handled in a unified location)

forge-gui-desktop/src/main/java/forge/screens/deckeditor/controllers/ACEditorBase.java
tehdiplomat: If the key cards are only created via the UI it would be fine, but as you've seen with the new adventure worlds, thats not how people are making deck files.

forge-core/src/main/java/forge/deck/Deck.java
tehdiplomat: Are you imaging this being its own param(i.e. not extracted from AIHints)? Or just its own method functions to handle it?

Hanmac: These could still be AIHints (for now)
But having methods to handle them would cleanup some code

(Including the case insensitive part)

tehdiplomat: The more the thought about it the more i thought it made more sense to be its own header type too.

forge-core/src/main/java/forge/deck/Deck.java
```diff
@@ -619,6 +648,18 @@ public UnplayableAICards getUnplayableAICards() {
         return unplayableAI;
     }
 
+    public void setAiHint(String hintType, String hintValue) {
```
tehdiplomat, line 651: I left this for now. But I have some other ideas for how it might be useful. (I.e. mulliganing to get a key card)

PR #10177, opened 2026-03-26
Fix the Sealed Deck event reward showing 0 cards

Agetian:


forge-gui-mobile/src/forge/adventure/data/AdventureEventData.java
```diff
@@ -747,9 +747,15 @@ public void giveRewards() {
...
-                rewards[3].cardRewards = new Deck[]{draftedDeck};
-            }
+            rewards[3] = new AdventureEventReward();
+            rewards[3].minWins = 3;
+            rewards[3].maxWins = 3;
+            draftedDeck.setName("Sealed Card Pool");
+            draftedDeck.setComment("Prize for placing 1st overall in sealed event");
+            CardPool sideboard = draftedDeck.get(DeckSection.Sideboard);
+            draftedDeck.getMain().addAll(sideboard);
+            draftedDeck.get(DeckSection.Sideboard).clear();
```
Jetz72: Doesn't rewarding a deck already give you the cards in every zone?

Agetian: It does, but it only counts the ones in the main for the purpose of displaying the number of cards in the pool :)

Agetian: Oh, this actually gives me an idea, give me a sec... Maybe this can be fixed higher up the chain where the deck is stored for reward

Agetian: Kk committed a better fix, this is more appropriate imo :) Thanks!

Jetz72: Could probably just fix it so it counts cards in all zones? But that sounds like a separate issue.

Agetian: Hmm yeah, I might look into that, if I figure it out I'll submit a followup PR :)

Agetian: Probably counting all the zones won't always be viable btw - in Draft, for instance, only the Main cards are counted because only the drafted deck is awarded, not the entire draft pool (but in Sealed, we're awarding the entire thing)

Jetz72: In draft, `draftedDeck` (which is awarded) is just a copy of `registeredDeck` taken right after the draft phase completes. Means any basic lands they add don't get added into the reward pool. It can have cards in either the main deck or sideboard since the player can draft cards to either zone.

I think it might be easier to follow the logic if we renamed `draftedDeck` to `rewardDeck` and always use it to hold the cards the player is awarded if they win the event.

Agetian: Yeah, that's true!

Agetian: Hmm, maybe rewardPack? What do you think? It's not always a real "deck" (as in the sealed, for instance), but it does consist of cards and it's opened after the event as a pack.

Jetz72: Could use whatever. `rewardPool` maybe? `rewardDeck` still works since it's typed as a deck. As a broader change it'd probably make sense to be able to supply rewards as `CardPools` but that seems like a lot of work for little benefit.

Agetian: Yep, agreed, will prolly use Deck

PR #10183, opened 2026-03-26
Rename variable for the reward pool

Agetian:


Agetian:
> Might be good to make it so Jumpstart defines a rewardDeck too for consistency. Wonder if there's enough common ground between the `EventScene.selectedJumpstartPackIsLast` and `AdventureDeckEditor.completeDraft` to consolidate into a common method that registers the player's deck and advances the event status.

Yeah, sounds like a plan. I'll see if I can figure that out, too :)

PR #10189, opened 2026-03-27
Adventure event deck validation; other cleanup

Jetz72:
* Attempting to enter an event match with an invalid deck will now display an error.
* Moved around some logic for how the event states advance for Sealed, and for how the `metaDraftTable` is made visible.
* Cleaned up some of the code for opening the deck editor from the event screen. Deleted `DraftScene` entirely because it serves the same purpose as `DeckEditScene`.
* Updated the base deck editor so opening it with cards in the sideboard, none in the main board, and no draft pack page will default to opening the sideboard first. Simplified config code elsewhere accordingly.


Agetian:
Looks and works great, thank you! :)

PR #10195, opened 2026-03-27
Enable and rebalance Draft and Sealed for small old sets in Adventure mode

Agetian:
- Enable Draft and Sealed for small old sets in Adventure mode (such as The Dark, Antiquities, etc.)
- Rebalance draft packs for smaller old sets for a more interesting and less bland draft/sealed experience based on suggestions from Discord and practical testing.

This is a little bit on the experimental side, though I admit this does provide additional variety and make the Draft/Sealed experience for these sets less stall-y and more diverse in Adventure. Opinions are welcome whether we mainline this and whether we need to further tweak the configurations for the boosters as well.


Agetian:
ARN and ATQ ideally need to go paired with 2ED (if we follow the release timeline), but that'll prolly make them inaccessible due to Power Nine. Not sure if we should enable those in Adventure or keep them true-to-timeline (2ED) and thus off limits in Adventure.

vanja-ivancevic:
I like this very much. Thank you for doing it. Will you push it?

Agetian:
Most likely yes, though I'm still not sure if this is best applied globally or somehow locally to Old Border Shandalar. Is it possible to apply this to the Old Border world in case we decide to go that way?
My tests indicate this is fine in general though

Jetz72:
Perhaps this is something that could take advantage of the block override feature @vanja-ivancevic introduced in #10398?

Agetian:
Indeed, that was exactly my hope. Would this be difficult to convert into a world-specific override?

Agetian:
Looks straightforward at first glance, will likely try it tomorrow morning

vanja-ivancevic:
Global path is probably easier, no? Would it bother vanilla players if old drafts show up?

Agetian:
Prolly won't bother anyone very much, but I'm slightly concerned that these tweaked definitions will become our global ones even outside of Adventure mode :/ Probably not a huge concern, but still... I'd probably still prefer these to at least be local to Adventure mode where it's most relevant. The min cards per pack restriction should probably be relaxed globally though, I don't see anything wrong with that, especially since earlier sets did use fewer than 11 cards per pack and there were legit custom early sealed and early draft formats that involved more than 4 draft packs (back in the early years of Magic, true official drafting wasn't really a thing yet...)

Agetian:
Ugh, something weird happened with this branch, I'll start a new one.

Agetian:
Started a new one: https://github.com/Card-Forge/forge/pull/10707

PR #10199, opened 2026-03-28
CardFactoryUtil: Megamorph as ReplacementEffect

### LINKED ISSUE #10165: Megamorph should be cleanup as ReplacementEffect
https://github.com/Card-Forge/forge/blob/a6a40b1d7165bff1e03e7ef77fec39a9101fb4a6/forge-game/src/main/java/forge/game/ability/effects/SetStateEffect.java#L206-L208

Rules:
> 702.37b Megamorph is a variant of the morph ability. “Megamorph [cost]” means “You may cast this card as a 2/2 face-down creature with no text, no name, no subtypes, and no mana cost by paying {3} rather than paying its mana cost” and **“As this permanent is turned face up, put a +1/+1 counter on it if its megamorph cost was paid to turn it face up.”** A megamorph cost is a morph cost

* ReplaceTurnFaceUp needs Cause Param
https://github.com/Card-Forge/forge/blob/a6a40b1d7165bff1e03e7ef77fec39a9101fb4a6/forge-game/src/main/java/forge/game/card/Card.java#L905-L909

then `Mega$` Param can be removed

it shouldn't be able to get more "Megamorph" instances, but just in case, we probably need to check if the ReplacementEffect and the SpellAbility are from the same Keyword?

Hanmac:
Closes #10165

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
@@ -680,6 +680,7 @@ public class ComputerUtilCost {
 
         if (root.costHasManaX()) {
             val = ComputerUtilMana.determineLeftoverMana(root, ai, effect);
+            // TODO find a way to consider lower value due to Ward
             if (sa.hasParam("AIXMax")) {
                 sa.setXManaCostPaid(val);
                 int calculated = AbilityUtils.calculateAmount(source, sa.getParam("AIXMax"), sa);
--- a/forge-ai/src/main/java/forge/ai/CreatureEvaluator.java
+++ b/forge-ai/src/main/java/forge/ai/CreatureEvaluator.java
@@ -37,6 +37,8 @@ public class CreatureEvaluator implements Function<Card, Integer> {
         int power = c.getNetCombatDamage();
         final int toughness = c.getNetToughness();
 
+        // TODO getKeyCards
+
         // TODO replace with ReplacementEffect checks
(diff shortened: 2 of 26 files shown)


Hanmac:
I need to debug why it fails for Manifest now 🤔

forge-game/src/main/java/forge/game/ability/effects/SetStateEffect.java
```diff
@@ -224,7 +218,6 @@ public void resolve(final SpellAbility sa) {
                 }
             }
         }
-        table.replaceCounterEffect(game, sa, true);
```
tool4ever, line 227: hmn, this would have been an easy fix here:
https://magicjudge.tumblr.com/post/138508619577/regarding-your-recent-answer-about-doubling-season

(related to #3512)

Hanmac: Okay, so what should be the fix?

tool4ever: Easiest for now could be if
`CountersPutEffect:replaceCounterEffect` does effect = `!sa.isKeyword(Keyword.MEGAMORPH)`
? 🤔

Hanmac: Maybe check if the cause is Static? 🤔

Edit: CountersPutEffect doesn't check for cause

Hanmac: should be fixed?

PR #10202, opened 2026-03-28
Fix CountersProliferateAi

### LINKED ISSUE #10200: Staff Of Compleation
Staff of compleation does 5 things, but in game it only taps to proliferate which also costs life, when it hits the board, its pays life and taps, then it it can untap itself for mana, and do it all over again when there will be no counters on the board to proliferate so its just wasting life.

tool4ever:
Closes #10200


Agetian (dismissed):

Agetian:
Looks like one of the tests is failing, otherwise should be good 😊

PR #10216, opened 2026-03-30
SOS cards from story, 30th March

Fulgur14:


forge-gui/res/cardsfolder/upcoming/together_as_one.txt
```diff
@@ -0,0 +1,8 @@
+Name:Together as One
+ManaCost:6
+Types:Sorcery
+A:SP$ Draw | NumCards$ X | ValidTgts$ Player | SubAbility$ DBDamage | SpellDescription$ Converge — Target player draws X cards, Together as One deals X damage to any target, and you gain X life, where X is the number of colors of mana spent to cast this spell.
```
tool4ever: CARDNAME

PR #10219, opened 2026-03-30
Player Hidden Keywords related to SearchLibrary

kojotak:
An attempt to refactor 3 hidden player keywords related to "search library" from #4760 :

- "LimitSearchLibrary"
- "CantSearchLibrary"
- "Spells and abilities you control can't cause you to search your library."

Tested using `Stoneforge Mystic`, `Mindlock Orb`, `Ashiok, Dream Render`, `Aven Mindcensor`. Example of applying library search limit of 4 cards:

image


forge-game/src/main/java/forge/game/staticability/StaticAbilitySearchLibrary.java
```diff
@@ -0,0 +1,52 @@
...
+import static forge.game.staticability.StaticAbilityMode.LimitSearchLibrary;
+import static forge.game.staticability.StaticAbilityMode.CantCauseToSearchLibrary;
+
+
+public class StaticAbilityCantSearchLibrary {
+
+    /**
+     * @return maximum number of cards which can be fetched from a library considering its size and search limit or null if there is no limit
+     */
+    public static Integer limitSearchLibraryConsideringSize(Player player) {
```
kojotak, line 18: 1. I should have probably named this class `StaticAbilitySearchLibrary`. 
2. Not sure about this method name - it's not related to static ability, but it's always used with the value of the limit in order to prevent "card overflow"

forge-gui/res/cardsfolder/l/leonin_arbiter.txt
```diff
@@ -2,7 +2,7 @@ Name:Leonin Arbiter
 ManaCost:1 W
 Types:Creature Cat Cleric
 PT:2/2
-S:Mode$ Continuous | Affected$ Player | AddKeyword$ CantSearchLibrary | IgnoreEffectCost$ 2 | Description$ Players can't search libraries. Any player may pay {2} for that player to ignore this effect until end of turn.
+S:Mode$ CantSearchLibrary | ValidPlayer$ Player | IgnoreEffectCost$ 2 | Description$ Players can't search libraries. Any player may pay {2} for that player to ignore this effect until end of turn.
```
tool4ever: This doesn‘t work

Hanmac: @kojotak check "Volrath's Curse" for both Continuous and other modes

in this case, it should be something like this:
`Mode$ Continuous,CantSearchLibrary | Affected$ Player | ValidPlayer$ Player | IgnoreEffectCost$ 2`

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -3723,11 +3723,11 @@ public boolean canSearchLibraryWith(SpellAbility sa, Player targetPlayer) {
             return true;
         }
 
-        if (hasKeyword("CantSearchLibrary")) {
+        if (StaticAbilityCantSearchLibrary.cantSearchLibrary(this)) {
             return false;
         }
         return targetPlayer == null || !targetPlayer.equals(sa.getActivatingPlayer())
-                || !hasKeyword("Spells and abilities you control can't cause you to search your library.");
+                || !StaticAbilityCantSearchLibrary.cantCauseToSearchLibrary(this);
```
tool4ever: this should just use the same method above so it can be more dynamic

forge-gui/res/cardsfolder/a/aven_mindcensor.txt
```diff
@@ -4,5 +4,5 @@ Types:Creature Bird Wizard
 PT:2/1
 K:Flash
 K:Flying
-S:Mode$ Continuous | Affected$ Player.Opponent | AddKeyword$ LimitSearchLibrary | Description$ If an opponent would search a library, that player searches the top four cards of that library instead.
+S:Mode$ LimitSearchLibrary | ValidPlayer$ Player.Opponent | LimitNum$ 4 | Description$ If an opponent would search a library, that player searches the top four cards of that library instead.
```
tool4ever, line 7: technically this is a replacement effect
not sure yet if the static approach is better for both

forge-gui/res/cardsfolder/l/leonin_arbiter.txt
kojotak: Thanks, good hint. It was also neccessary to check `getIgnoreEffectPlayers` in `cantSearchLibrary` so it finally works.

Tested:

-  payed  2R for ignore effect
-  2W for Stoneforge mystic - confirm trigger
-  select an equipment from library

image

forge-game/src/main/java/forge/game/player/Player.java
kojotak: Do you mean:

- remove the `cantCauseToSearchLibrary()` method completely
- add check for `CantCauseToSearchLibrary` mode into `cantSearchLibrary()`?

tool4ever: it's just a slightly more specific variant, not a different mode
we usually have one entry point, so scripts can just use params for their needs

kojotak: Removed the unnecessary mode in https://github.com/Card-Forge/forge/pull/10219/commits/f7b33c7cbdd8d8213690e3e107f51c14ad08c168 but this may break something else, because the `if` is no longer specific to `Ashiok` only.

I was thinking about an alternative: add `ValidActivatingPlayer$ Opponent` for `Ashiok` card, move the `targetPlayer vs activatingPlayer` check into new `cantCauseToSearchLibrary()` method.

forge-gui/res/cardsfolder/a/aven_mindcensor.txt
kojotak: So it should look like `Kenessos, Priest of Thassa` with Scry?

forge-gui/res/cardsfolder/a/ashiok_dream_render.txt
```diff
@@ -2,7 +2,7 @@ Name:Ashiok, Dream Render
 ManaCost:1 UB UB
 Types:Legendary Planeswalker Ashiok
 Loyalty:5
-S:Mode$ Continuous | Affected$ Opponent | AddKeyword$ Spells and abilities you control can't cause you to search your library. | Description$ Spells and abilities your opponents control can't cause their controller to search their library.
+S:Mode$ CantSearchLibrary | ValidPlayer$ Opponent | Description$ Spells and abilities your opponents control can't cause their controller to search their library.
```
Hanmac: We need to add ValidCause to check for Spells and Abilities and check if they are controlled by the same player as the one that gets their library searched

forge-gui/res/cardsfolder/a/ashiok_dream_render.txt
```diff
@@ -2,7 +2,7 @@ Name:Ashiok, Dream Render
 ManaCost:1 UB UB
 Types:Legendary Planeswalker Ashiok
 Loyalty:5
-S:Mode$ CantSearchLibrary | ValidPlayer$ Opponent | Description$ Spells and abilities your opponents control can't cause their controller to search their library.
+S:Mode$ CantSearchLibrary | ValidPlayer$ Opponent  | ValidCause$ SpellAbility.OppCtrl | Description$ Spells and abilities your opponents control can't cause their controller to search their library.
```
Hanmac: not enough, needs to be the same Player, 
checkout Params like "BlockerRelative" for similar stuff

kojotak: Oh, multi player situation, right🤔? 

P1 owns Ashiok, P2 causes to P3 to search his library. This should work, since P2 != P3. P2 can not search its own library. But P2 can force P1 to search his library 🙈.

forge-gui/res/cardsfolder/a/aven_mindcensor.txt
tool4ever: it should be fine since no other cards currently use it:
we can't switch to RE fully anyway because of the IgnoreEffectCost variant :/

forge-gui-desktop/src/test/java/forge/ai/ability/StaticAbilitySearchLibraryTest.java
```diff
@@ -0,0 +1,168 @@
...
+        ashiok.setCounters(CounterEnumType.LOYALTY, 5);
+        Card aridMesa = addCard("Arid Mesa", p1);
+
+        game.getAction().checkStateEffects(true);
+
+        assertFalse(p1.canSearchLibraryWith(findSearchLibraryAbility(aridMesa), p1));
+    }
+
+    @Test
+    public void testAshiok_opponentCanCauseOtherOpponentToSearchTheirLibrary() {
```
kojotak, line 113: `ValidCause$ SpellAbility.OppCtrl` is not needed for this test to pass, so there is a probably an error in there

forge-gui/res/cardsfolder/a/ashiok_dream_render.txt
kojotak: I've tested this a little bit and concluded, that there should be just 2 conditions necessary. ValidPlayer and ValidCause, but not this "simple" one (with 1 parameter - the cause), but "relative-like" (with 2 parameters - cause and context - given card causing the search effect, instead of the default context card - source of static abbility) - I've called it `ValidPlayerCauseRelative` - similar to `BlockerRelative`.

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -3722,11 +3722,7 @@ public boolean canSearchLibraryWith(SpellAbility sa, Player targetPlayer) {
             return true;
         }
 
-        if (hasKeyword("CantSearchLibrary")) {
-            return false;
-        }
-        return targetPlayer == null || !targetPlayer.equals(sa.getActivatingPlayer())
-                || !hasKeyword("Spells and abilities you control can't cause you to search your library.");
+        return !StaticAbilitySearchLibrary.cantSearchLibrary(this, sa);
```
tool4ever, line 3725: I'm a bit skeptical about the usefulness of these tests in general
(you're not even using `targetPlayer` any longer)

kojotak: I agree. These tests help me to understand and experiment with Forge engine and give me some (false?) confidence. I can easily remove them if the rest is OK.

tool4ever: Let's try to break it down:
*Ashiok* needs the following information
a) SA controller
b) Searcher
c) Searched

then the condition has to check `a = b = c = Opponent` (all the same player)

I have a hard time seeing that this refactor has this equivalently covered yet

PR #10220, opened 2026-03-30
Momir use random land art

clairchiara:
Fix bug with Momir always using land art from the last set to generate the deck
+
make the lands more randomised and not all from the same set


forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
```diff
@@ -338,6 +338,7 @@ protected final void updateView(final boolean fullUpdate) {
 
     /** Returns a runnable to start a match with the applied variants if allowed. */
     public Runnable startGame() {
+        FModel.getBlocks(); // Ensure blocks are loaded for Momir Basic and MoJhoSto deck generation
```
Jetz72: I'm not sure if this is the most appropriate place to apply this workaround, or if it's the most appropriate workaround. Anyone else have any thoughts on this, @Hanmac, @tool4ever?

tool4ever: Seems this got broken by https://github.com/Card-Forge/forge/pull/8750, so thanks for the find

imo just needs to be refactored to have `getBlockLands()` lazy initialize itself

clairchiara: It's tricky because neither FModel nor CardBlock are accessible from GameType or StaticData, but I could move it inside this block so it's only loaded when the deck needs to be autogenerated
`else if (autoGenerateVariant != null) {`

Sorry I'm not really familiar with the codebase yet

Agetian:
Not sure how to improve this, but looks good to me otherwise. Whether a Momir deck looks better with lands from different sets or from a single set is a good question, but I guess I'm kinda ambivalent here - different sets provide more variety, while same set provides more consistency (avoiding mixture of old border and new border sets, for instance). I guess it's fine either way :)

forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
```diff
@@ -458,6 +458,7 @@ public Runnable startGame() {
                 else if (autoGenerateVariant != null) {
                     Deck autoDeck = autoGenerateVariant.autoGenerateDeck(rp);
                     deck = new Deck();
+                    FModel.getBlocks(); // Ensure blocks are loaded for Momir Basic and MoJhoSto deck generation
```
Jetz72: Aren't you now calling it after `autoGenerateDeck`?

forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
Jetz72: If we don't mind dropping the requirement that the lands be from a card block, we could just do something like: `Aggregates.random(StaticData.getCommonCards().getAllCards("Forest"), 12)`. That'd sidestep the preloading issue, and then we could probably clean up `getBlockLands` altogether since I think it's only used for this.

forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
clairchiara: Uh, you’re right. Weird because I swear it’s working fine, but I’ll move it up just in case.

forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
clairchiara: > If we don't mind dropping the requirement that the lands be from a card block, we could just do something like: `Aggregates.random(StaticData.getCommonCards().getAllCards("Forest"), 12)`. That'd sidestep the preloading issue, and then we could probably clean up `getBlockLands` altogether since I think it's only used for this.

That’s a good idea, I’ll see if It works.

clairchiara: It seems to work just fine, so I switched to it and removed all `getBlockLands()` related code including the additional `FModel.getBlocks()` call

Jetz72 (approved):

PR #10221, opened 2026-03-30
Ante tweaks

vanja-ivancevic:
- Fix Timmerian Fiends and Tempest Efreet card scripts to use CorrectedSelf instead of Self when referencing the sacrificed host card. Self returns a stale reference to the original card object, while CorrectedSelf resolves to the actual game state copy in the graveyard. This fixes ownership changes and zone moves not applying to the correct card after sacrifice.
- Log mid-game ante zone additions in the existing GameEventCardChangeZone handler (e.g. Contract from Below, Demonic Attorney).
- Tested with Timmerian Fiends, Bronze Tablet, Contract from Below, and Demonic Attorney. Additional testing welcome.


forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -3360,6 +3360,13 @@ public void changeOwnership(Card card) {
         }
         card.setOwner(this);
 
+        // Also update the game state copy if the card has been zone-changed
+        // (e.g. a sacrificed card whose copy exists in the graveyard)
+        Card gameCard = card.getGame().getCardState(card, null);
```
tool4ever: all these engine changes don't make sense:
it's probably just the script doing things in the wrong order

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -3390,10 +3397,10 @@ public void destroyPhysicalCard(Card card) {
         card.getOwner().lostOwnership.add(card);
     }
 
-    public CardCollectionView getLostOwnership() {
+    public List<Card> getLostOwnership() {
         return lostOwnership;
     }
-    public CardCollectionView getGainedOwnership() {
+    public List<Card> getGainedOwnership() {
```
tool4ever: there's surely a smarter way than tracking all changes when only the last is relevant...

forge-game/src/main/java/forge/game/player/Player.java
vanja-ivancevic: Thanks for the review! Good call on the script approach. Reverted the engine changes. The fix is now script-only: changed Timmerian Fiends and Tempest Efreet to use `CorrectedSelf` instead of `Self` for the post-sacrifice ownership change and zone move. `Self` returns a stale reference to the original card object, while `CorrectedSelf` resolves to the actual graveyard copy via `game.getCardState()`.

forge-game/src/main/java/forge/game/GameAction.java
```diff
@@ -579,6 +579,12 @@ private Card changeZone(final Zone zoneFrom, Zone zoneTo, final Card c, Integer
         // play the change zone sound
         game.fireEvent(new GameEventCardChangeZone(c, zoneFrom, zoneTo));
 
+        // log when a card is added to ante mid-game (e.g. Contract from Below, Demonic Attorney)
+        if (zoneTo.is(ZoneType.Ante) && zoneFrom != null && !zoneFrom.is(ZoneType.Ante)) {
+            game.fireEvent(new GameEventAddLog(GameLogEntryType.ANTE,
```
tool4ever: this doesn't really seem different from the event above

vanja-ivancevic: The GameEventCardChangeZone above only logs Battlefield -> Graveyard/Exile transitions (GameLogFormatter line 297 filters everything else). Library -> Ante moves are silently dropped, so the explicit ANTE log entry is the only way mid-game ante additions show up in the game log.

tool4ever: ok, then please just tweak that instead of firing another event

vanja-ivancevic: Good point, moved the ante logging into the existing GameEventCardChangeZone handler in GameLogFormatter instead.

PR #10224, opened 2026-03-31
Chaos mode in Realm of Legends - generate valid Commander decks

### LINKED ISSUE #10214: Realm of Legends: Remove Chaos Option
Currently, the 'chaos' deck option is broken on realm of legends since it requires 60 card decks. I only left it in because it currently can't be removed, but users are picking it and complaining the game is broken.

If someone could figure out how to remove it as an option for Realm of Legends, that'd be great.

MostCromulent:
## Summary

Closes #10214

Addresses @VigilantSera request in Card-Forge/forge#10214 — Chaos mode generates random 60-card preconstructed decks, incompatible with Realm of Legends' 98-card minimum deck size.

- Adds `allowChaos` config flag to `ConfigData` (default `true`), consistent with the existing `enableGeneticAI` flag pattern
- Gates the Chaos mode in `NewGameScene` behind this flag, matching how Precon/CommanderPrecon modes are conditionally added
- Sets `allowChaos: false` in Realm of Legends' `config.json`

## Notes

- **Backwards compatible:** The flag defaults to `true`, so all other adventures are unaffected. Existing Realm of Legends saves are also unaffected — the flag only controls the new game screen mode picker, not gameplay or save loading.
- **Consistent with existing config options:** `allowChaos` follows the same pattern as `enableGeneticAI` — a public boolean on `ConfigData` with a default value, read directly in `NewGameScene`.

---

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/util/AdventureModes.java
+++ b/forge-gui-mobile/src/forge/adventure/util/AdventureModes.java
@@ -42,8 +42,8 @@ public enum AdventureModes {
     }
 
     public boolean isCommanderLike() {
-        if (this == Commander || this == CommanderPrecon) return true;
-        return this == Chaos && "Commander".equalsIgnoreCase(Config.instance().getConfigData().chaosDeckFormat);
+        boolean isChaosCommander = this == Chaos && "Commander".equalsIgnoreCase(Config.instance().getConfigData().chaosDeckFormat);
+        return this == Commander || this == CommanderPrecon || isChaosCommander;
     }
 
     public boolean usesFolderDeckPicker() {


Agetian:
Looks good to me

Jetz72:
Would a valid alternative be to just tweak Chaos to generate a valid 100 card random commander deck? I believe we have code for generating those somewhere. Could use the config to specify a deck format for the chaos deck instead of making it a toggle.

Agetian:
That would indeed be a better (and preferred) implementation 👍

forge-gui-mobile/src/forge/adventure/util/AdventureModes.java
```diff
@@ -42,7 +42,8 @@ public Array<String> getModes() {
     }
 
     public boolean isCommanderLike() {
-        return this == Commander || this == CommanderPrecon;
+        if (this == Commander || this == CommanderPrecon) return true;
+        return this == Chaos && "Commander".equalsIgnoreCase(Config.instance().getConfigData().chaosDeckFormat);
```
Agetian: Maybe can be simplified to a single return statement? If a single line is too long, a boolean can be introduced determining the Chaos Commander setting, and then a return with three "or"ed statements (this == Commander || this == CommanderPrecon || isChaosCommander)

Issue #7793, opened 2025-06-04
Adventure Mode on mobile: disabling vibration on encountering enemies


Nitroxylin:
**Is your feature request related to a problem? Please describe.**
While it is possible to disable vibrations on life loss (although the setting for doing so is placed in a non-obvious place), there doesn't seem to be a similar checkbox for disabling vibration on encountering enemies. The buzz can be annoying.

**Describe the solution you'd like**
A similar checkbox in the settings.

povanspring:
Being able to disable this for PC controller would be nice too.

tool4ever:
@phaelex
I have revived this issue instead

imo the `EventVisualizer` should be involved there
and probably have a slider for strength

while currently only the mobile port supports vibration maybe makes sense to implement the main entry point in forge-gui for future extension?

@MostCromulent
something you might be interested in
(so remote players could vibrate?)

phaelex:
I'd like to request being able to disable controller vibration entirely, not just as a granular option for each event that causes the vibration to occur (encountering enemies, picking up shards/gold/chests, etc).

PR #10255, opened 2026-04-02
Add centralised haptic feedback controls

MostCromulent:
**Closes https://github.com/Card-Forge/forge/issues/7793**

Addresses all feature requests in that discussion: separate intensity control for device and controller, per-event vibration toggles,  implemented in forge-gui for future extension.

## Summary

- Introduces `HapticEngine` to handle all vibration dispatch with preference toggles and intensity scaling
- Replaces scattered `Gdx.input.vibrate()` / `Controllers.getCurrent().startVibration()` calls across adventure and match code with `HapticEngine.vibrate()` calls
- Adds per-event enable/disable toggles (life loss, poison, long press, enemy encounter, adventure reward, shop action) and separate device/controller intensity sliders (0-100%)
- Adds `vibrate()` / `vibrateController()` to `IGuiBase` with no-op defaults, implemented in `GuiMobile`

## Details

Previously, vibration was handled inline at each callsite - `VPlayerPanel.update()` computed life/poison deltas and called `Gdx.input.vibrate()` directly, adventure stages duplicated controller-check boilerplate, and the only user setting was a single "vibrate on life loss" toggle.

Now `HapticEngine` subscribes to `GameEventPlayerLivesChanged` and `GameEventPlayerPoisoned` via EventBus, so match vibration logic is decoupled from the GUI panel. This gives flexibility to expand to other event types in future without the need for additional call sites.

Adventure and UI callsites (which do not rely on EventBus) use static `HapticEngine.vibrate(pref, ...)` methods that check the per-event toggle and apply intensity scaling before dispatching to the platform.

On mobile settings are exposed in both the main settings page and the adventure settings scene.

Screenshot 2026-04-03 083123

Screenshot 2026-04-03 083012

---


forge-gui/src/main/java/forge/haptic/HapticEngine.java
```diff
@@ -0,0 +1,75 @@
...
+    public static void vibrate(FPref pref, int baseDeviceMs, int baseControllerMs, float controllerAmplitude) {
+        if (baseDeviceMs <= 0 && baseControllerMs <= 0) return;
+        if (!FModel.getPreferences().getPrefBoolean(pref)) return;
+        int deviceIntensity = getIntensity(FPref.UI_VIBRATE_DEVICE_INTENSITY);
+        if (baseDeviceMs > 0 && deviceIntensity > 0) {
+            GuiBase.getInterface().vibrate(scale(baseDeviceMs, deviceIntensity));
+        }
+        int controllerIntensity = getIntensity(FPref.UI_VIBRATE_CONTROLLER_INTENSITY);
+        if (baseControllerMs > 0 && controllerIntensity > 0) {
+            GuiBase.getInterface().vibrateController(scale(baseControllerMs, controllerIntensity), controllerAmplitude);
```
tool4ever: why does this scale the time instead of modifying controllerAmplitude?

forge-gui/src/main/java/forge/haptic/HapticEngine.java
```diff
@@ -0,0 +1,75 @@
...
+        if (intensity <= 0) return;
+        GuiBase.getInterface().vibrate(scale(baseDeviceMs, intensity));
+    }
+
+    public static void vibrate(FPref pref, int baseDeviceMs, int baseControllerMs, float controllerAmplitude) {
+        if (baseDeviceMs <= 0 && baseControllerMs <= 0) return;
+        if (!FModel.getPreferences().getPrefBoolean(pref)) return;
+        int deviceIntensity = getIntensity(FPref.UI_VIBRATE_DEVICE_INTENSITY);
+        if (baseDeviceMs > 0 && deviceIntensity > 0) {
+            GuiBase.getInterface().vibrate(scale(baseDeviceMs, deviceIntensity));
```
tool4ever: the device path also has methods available to get rid of this intensity workaround

forge-gui/src/main/java/forge/haptic/HapticEngine.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public static void vibrate(FPref pref, int baseDeviceMs) {
+        if (baseDeviceMs <= 0) return;
+        if (!FModel.getPreferences().getPrefBoolean(pref)) return;
+        int intensity = getIntensity(FPref.UI_VIBRATE_DEVICE_INTENSITY);
+        if (intensity <= 0) return;
+        GuiBase.getInterface().vibrate(baseDeviceMs, intensity * 255 / 100);
+    }
+
+    public static void vibrate(FPref pref, int baseDeviceMs, int baseControllerMs, float controllerAmplitude) {
```
tool4ever: I'm not sure it still makes sense to pass `controllerAmplitude` here

imo the engine should only modify the time, while the intensity should always be what the user feels comfortable with?

forge-gui/src/main/java/forge/haptic/HapticEngine.java
```diff
@@ -0,0 +1,71 @@
...
+            }
+        }
+    }
+
+    private boolean isLocalPlayer(forge.game.player.PlayerView player) {
+        LobbyPlayer local = GamePlayerUtil.getGuiPlayer();
+        return local != null && player.isLobbyPlayer(local);
+    }
+
+    public static void vibrate(FPref pref, int baseDeviceMs) {
```
tool4ever: this can be cleaned up by being an overloaded version of the one below

forge-gui-mobile/src/forge/GuiMobile.java
```diff
@@ -371,4 +372,18 @@ public boolean hasNetGame() {
...
+    public void vibrate(int milliseconds, int amplitude) {
+        if (milliseconds > 0 && amplitude > 0) {
+            Gdx.input.vibrate(milliseconds, amplitude, true);
+        }
+    }
+
+    @Override
+    public void vibrateController(int milliseconds, float amplitude) {
+        if (milliseconds > 0 && Controllers.getCurrent() != null && Controllers.getCurrent().canVibrate()) {
+            Controllers.getCurrent().startVibration(milliseconds, amplitude);
```
Jetz72, line 386: Does `Controllers.getCurrent()` start returning null if we go back to interacting with the game via mouse+keyboard or via touch screen? Or is there a chance I get into a battle and suddenly the controller sitting on my desk starts vibrating?

forge-gui-mobile/src/forge/screens/settings/SettingsPage.java
```diff
@@ -641,12 +641,38 @@ public void select() {
...
+            intensityOptions), 6);
         lstSettings.addItem(new BooleanSetting(FPref.UI_VIBRATE_ON_LONG_PRESS,
             Forge.getLocalizer().getMessage("lblVibrateAfterLongPress"),
             Forge.getLocalizer().getMessage("nlVibrateAfterLongPress")), 6);
+        lstSettings.addItem(new BooleanSetting(FPref.UI_VIBRATE_ON_LIFE_LOSS,
+            Forge.getLocalizer().getMessage("lblVibrateWhenLosingLife"),
+            Forge.getLocalizer().getMessage("nlVibrateWhenLosingLife")), 6);
+        lstSettings.addItem(new BooleanSetting(FPref.UI_VIBRATE_ON_POISON,
+            Forge.getLocalizer().getMessage("lblVibrateOnPoison"),
+            Forge.getLocalizer().getMessage("nlVibrateOnPoison")), 6);
```
Jetz72: Weird thought so I don't mind if you'd prefer to leave it this way, but this feels like a pretty narrow scope for a menu option. Wouldn't be technically accurate, but do you think anyone would notice or care if this setting was just rolled into the life loss vibration toggle?

forge-gui/src/main/java/forge/haptic/HapticEngine.java
```diff
@@ -0,0 +1,67 @@
...
+        if (baseDeviceMs <= 0 && baseControllerMs <= 0) return;
+        if (!FModel.getPreferences().getPrefBoolean(pref)) return;
+        int deviceIntensity = getIntensity(FPref.UI_VIBRATE_DEVICE_INTENSITY);
+        if (baseDeviceMs > 0 && deviceIntensity > 0) {
+            GuiBase.getInterface().vibrate(baseDeviceMs, deviceIntensity * 255 / 100);
+        }
+        int controllerIntensity = getIntensity(FPref.UI_VIBRATE_CONTROLLER_INTENSITY);
+        if (baseControllerMs > 0 && controllerIntensity > 0) {
+            GuiBase.getInterface().vibrateController(baseControllerMs, controllerIntensity / 100f);
+        }
```
Jetz72: Why separate the paths for vibrating the device and a controller, instead of vibrating the device as a fallback when the user doesn't have an active controller? Is there a setup where the user might want both to vibrate at once?

forge-gui-mobile/src/forge/screens/settings/SettingsPage.java
MostCromulent: Fair point - this is actually how it works already. I separated poison/life loss for correctness on this PR but agree poison is a narrow enough case they can be dealt with together.

forge-gui-mobile/src/forge/GuiMobile.java
MostCromulent: Have done some testing on this. `Controllers.getCurrent()` returns the most recently used controller and does not go null when switching to keyboard+mouse or touchscreen - so the controller sitting on your desk will vibrate.

See what you think of solution in latest commit:

> `HapticEngine.vibrate(pref, durationMs)` is the single entry point for all haptic feedback. It checks the event's preference toggle and the intensity slider, then routes:
> 
> 1. **Controller path** - if a controller is connected AND was the last-used input device, vibrate it via `Controllers.getCurrent().startVibration()`. Works on any platform, subject to SDL2 support (`canVibrate()`).
> 
> 2. **Device path** - otherwise, vibrate the device motor via `Gdx.input.vibrate()`. Skipped on desktop because there's no motor, and the call can leak to connected controllers through OS audio routing.
> 
> 3. **No-op** - desktop with no active controller, or intensity set to 0.
> 
> "Last-used input" is a boolean flag set `true` by controller button/axis events, and `false` by touch/click/keyboard events. Passive mouse cursor movement is ignored (fires constantly on desktop), as is minor joystick drift.

forge-gui/src/main/java/forge/haptic/HapticEngine.java
MostCromulent: Have implemented on latest commit. Hard to imagine a setup where user wants both - we can implement it this way and just see if anyone complains? :man_shrugging:

PR #10258, opened 2026-04-03
Show commander tax in perpetual mana cost overlay

MostCromulent:
## Summary
- Commander tax isn't technically a perpetual effect, but the perpetual overlay was added to answer "what does this card actually cost?" — commander tax is the most common reason a commander's cost differs from its printed cost, so it fits naturally here and a separate overlay setting isn't justified
- When the "perpetual mana cost" overlay is enabled, commanders in the command zone now display their mana cost including accumulated commander tax ({2} per previous cast from command zone)
- The tax is layered dynamically on top of any existing perpetual cost adjustments in the getter, keeping the perpetual calculation pipeline unchanged

## Design notes

**Scope:** The overlay shows permanent/intrinsic modifiers only (perpetual effects + commander tax), not external battlefield-state cost modifiers like Foundry Inspector. Similarly, cards that specifically offset commander tax — like Myth Unbound ({1} less per cast) or Liesa, Shroud of Dusk (pay life instead of tax) — are not reflected in the overlay. This is an existing limitation of the perpetual overlay — it has never accounted for external modifiers — and commander tax doesn't change that.

**View update trigger:** The mana cost overlay reads from a cached trackable property, so it needs an explicit `updateManaCostForView()` call to refresh. Three locations were considered:
- `Player.incCommanderCast()` — card is on the stack when this fires, so the zone-gated getter returns base cost (no effect)
- `GameAction.changeZone()` — multiple branches need the trigger, and `setPerpetual()` fires before the zone is set
- `Card.setZone()` — fires after zone assignment, single point for all transitions, gated behind `isCommander()`, consistent with the 50+ existing `view.update*()` calls in `Card.java`

`Card.setZone()` was chosen for simplicity and completeness.

---

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -6929,6 +6929,9 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
         if (currentZone == zone) { return; }
         currentZone = zone;
         view.updateZone(this);
+        if (isCommander()) {
+            updateManaCostForView();
+        }
     }
 
     public boolean isInZone(final ZoneType zone) {
--- a/forge-game/src/main/java/forge/game/card/CardState.java
+++ b/forge-game/src/main/java/forge/game/card/CardState.java
@@ -44,6 +44,7 @@ import forge.game.spellability.SpellPermanent;
 import forge.game.staticability.StaticAbility;
 import forge.game.staticability.StaticAbilityMode;
 import forge.game.trigger.Trigger;
+import forge.game.zone.ZoneType;
 import forge.util.CardTranslation;
 import forge.util.ITranslatable;
 import forge.util.IterableUtil;
@@ -273,7 +274,24 @@ public class CardState implements GameObject, IHasSVars, ITranslatable {
     }
 
     public ManaCost getPerpetualAdjustedManaCost() {
-        return perpetualAdjustedManaCost == null ? getManaCost() : perpetualAdjustedManaCost;
+        ManaCost base = perpetualAdjustedManaCost == null ? getManaCost() : perpetualAdjustedManaCost;
+
+        // Layer commander tax on top of perpetual adjustments when in command zone
+        Card card = getCard();
+        if (card != null && card.isCommander() && card.getZone() != null
+                && card.getZone().is(ZoneType.Command)) {
+            Player owner = card.getOwner();
+            if (owner != null) {
+                int tax = owner.getCommanderCast(card) * 2;
+                if (tax > 0) {
+                    int newGeneric = base.getGenericCost() + tax;
+                    String shards = base.getShortString().replaceFirst("" + base.getGenericCost(), "").trim();
+                    String costStr = newGeneric + (shards.isEmpty() ? "" : " " + shards);
+                    return new ManaCost(costStr);
+                }
+            }
+        }
+        return base;
     }
 
     public final ColorSet getColor() {


jamincollins:
## Findings

#### 1. Commander tax string manipulation may corrupt mana cost with multi-digit generic costs
The `getPerpetualAdjustedManaCost()` method extracts shards by stripping the generic cost via `replaceFirst("" + base.getGenericCost(), "")`. This is fragile: if generic cost is 10 and short string is "10 R", `replaceFirst("10", "")` yields " R" – OK. If cost is 1 and short string is "10 R", `replaceFirst("1", "")` yields "0 R" because the first "1" is removed from "10". Similarly for generic cost 2 and string "12 R", it would corrupt. The code assumes the generic cost appears only as a standalone prefix, but `replaceFirst` matches the first occurrence anywhere in the string.

Test case:
```java
ManaCost base = new ManaCost("10 R");
int generic = base.getGenericCost(); // 10
String shards = base.getShortString().replaceFirst("" + generic, "").trim(); // "R" correct

ManaCost base2 = new ManaCost("12 R");
int generic2 = 2;
String shards2 = base2.getShortString().replaceFirst("" + generic2, "").trim(); // "1 R" – wrong, should be "12 R" with generic 2
```
Impact: Incorrect mana cost display for commanders whose base cost contains a digit matching the first digit of generic cost, e.g., 12 R with tax 2 → displayed as 4 R instead of 4 R? Actually generic becomes 2+tax and shards corrupted.

Recommended fix: Parse shards via `ManaCost` API rather than string replace.

#### 2. Commander tax only applied when tax > 0
The method returns early only when tax > 0, otherwise returns base. This means when tax is zero, base is returned correctly. However, the method also constructs a new `ManaCost` only when tax > 0, which is fine but inconsistent with potential zero-tax display expectations. More importantly, the code does not update the perpetualAdjustedManaCost field; it returns a transient ManaCost each call, which may be acceptable for overlay but causes repeated parsing.

Test case:
```java
Card commander = ...;
commander.setZone(commandZone);
Player owner = commander.getOwner();
owner.getCommanderCast(commander); // returns 0
ManaCost cost = commander.getCurrentState().getPerpetualAdjustedManaCost();
// returns base, no tax added – correct
```
Impact: None for zero tax, but the string manipulation still executes only when tax>0.

#### 3. Null safety on card.getZone() and card.getOwner()
The check `card.getZone() != null && card.getZone().is(ZoneType.Command)` prevents NPE on zone. However `card.getOwner()` can be null for cards without owner (e.g., tokens, emblems). The code checks `owner != null` before using it. Good.

Potential edge: `card.getZone()` returns Zone object whose `is()` may be called on a zone from a different game instance after snapshot restore? Unclear.

Test case:
```java
Card c = new Card();
c.setCommander(true);
CardState s = c.getCurrentState();
s.setCard(c);
// card.getZone() == null
ManaCost cost = s.getPerpetualAdjustedManaCost(); // returns base, no NPE
```

#### 4. Generic cost may be negative for X costs
`base.getGenericCost()` can be negative for X-cost cards with reduction? In `calculatePerpetualAdjustedManaCost`, generic cost may be reduced below zero but kept for X costs. The tax addition `base.getGenericCost() + tax` could make negative + positive, still potentially negative. Then `new ManaCost(costStr)` parses string like "-1 1"? The constructor via String may parse incorrectly.

Test case:
```java
ManaCost base = new ManaCost("X");
base.getGenericCost(); // ?
```
Impact: Unclear if X costs appear in commanders. If so, tax addition to generic may produce invalid cost string.

#### 5. PerpetualAdjustedManaCost field not updated with tax
The tax is layered only in `getPerpetualAdjustedManaCost()` getter, not stored in `perpetualAdjustedManaCost`. This means any code that directly reads `perpetualAdjustedManaCost` field (e.g., via reflection or future code) will miss tax. The view update in `setZone` calls `updateManaCostForView()` which sets TrackableProperty.ManaCost via `getPerpetualAdjustedManaCost()`, so view is correct. However, if `calculatePerpetualAdjustedManaCost()` is called later and overwrites `perpetualAdjustedManaCost`, the tax will be reapplied on top of already-taxed base? Actually base is chosen as perpetualAdjustedManaCost if set, then tax added. So tax is applied each getter call, which is correct but could double-apply if base already includes tax from previous call? No, base is perpetualAdjustedManaCost which is set by calculatePerpetualAdjustedManaCost, not including tax. Good.

Potential issue: `calculatePerpetualAdjustedManaCost` sets `perpetualAdjustedManaCost` via string replacement similar to tax code, using `manaCost.getShortString().replace("" + manaCost.getGenericCost(), "")`. Same string fragility as above.

#### 6. updateManaCostForView called on every zone change for commanders
`Card.setZone` now calls `updateManaCostForView()` if `isCommander()`. This triggers view update on every zone change, even when commander tax is unchanged and card is not in command zone. Could cause unnecessary UI updates.

Test case:
```java
card.setZone(battlefield);
// isCommander true → updateManaCostForView called
// getPerpetualAdjustedManaCost will check zone is Command, so tax not applied, but view still updated
```
Impact: Minor performance / flicker.

#### 7. Tax applied only when card in Command zone, not other command zones like SchemeDeck?
`ZoneType.Command` check uses `card.getZone().is(ZoneType.Command)`. `PART_OF_COMMAND_ZONE` includes Command, SchemeDeck, PlanarDeck, AttractionDeck, ContraptionDeck, Junkyard. If commander is in SchemeDeck (e.g., in Oathbreaker?), tax would not be applied. Is that desired? Probably tax should apply for any command zone.

Test case:
```java
card.setZone(new Zone(ZoneType.SchemeDeck, game));
ManaCost cost = state.getPerpetualAdjustedManaCost();
// tax not applied
```

### Performance and security

#### 8. Repeated string parsing in getter
`getPerpetualAdjustedManaCost()` is called frequently by view updates. Each call does `base.getShortString()`, `replaceFirst`, `trim`, and `new ManaCost(costStr)` which parses string via `ManaCostParser`. This is expensive for UI rendering and may cause GC churn.

Test case:
```java
for (int i=0; i<1000; i++) {
    state.getPerpetualAdjustedManaCost();
}
```
Impact: Potential UI stutter when displaying many commanders.

## Verdict
Request Changes
The PR implements commander tax display in perpetual mana cost overlay, but the implementation uses fragile string manipulation to reconstruct mana costs, which can corrupt costs with multi-digit generic values and X costs. The tax logic also has limited zone coverage and repeated parsing overhead.

## Recommended Fixes
1. Replace string replaceFirst logic with ManaCost API manipulation: create a new ManaCost via parser or use ManaCostShard list to add generic cost.
2. Extend zone check to `isPartOfCommandZone()` or check `ZoneType.PART_OF_COMMAND_ZONE` for broader command zone coverage.
3. Cache the taxed ManaCost or compute tax lazily and avoid repeated parsing in getter.
4. Ensure setZone update is only triggered when zone actually changes to Command or tax-relevant zone.
5. Add unit tests for multi-digit generic costs, X costs, and hybrid mana.

---

MostCromulent:
Each suggestion checked against the branch.

|---|---|---|
| 1 | Replace `replaceFirst` with `ManaCost` API manipulation | The corruption case does not exist. `getShortString()` derives the printed number from the same call the PR uses — `int generic = getGenericCost();` then `if (generic > 0) sb.append(generic);` (`forge-core/.../ManaCost.java:299-305`) — so the leading digits always are the generic cost and the match sits at index 0. The `"12 R"` example hardcodes a generic of 2 for a cost whose `getGenericCost()` returns 12. With a generic of 0 the only string containing `"0"` is the `isZero()` output `"0"` itself, since no shard short string carries a digit except twobrid (`2/W`), which can only follow the generic. With a negative generic the reduction is appended at the end and `ManaCostParser.next()` parses the sign by design. |
| 2 | Broaden the zone check to `PART_OF_COMMAND_ZONE` | That set also covers `SchemeDeck`, `PlanarDeck`, `AttractionDeck`, `ContraptionDeck` and `Junkyard`, which hold schemes, planes and attractions rather than commanders. Commander tax is a cost for casting from the command zone, so `ZoneType.Command` is the correct scope and broadening it would apply tax where no commander sits. |
| 3 | Cache the taxed cost to avoid repeated parsing | The getter is not on the render path. `CardView.updateManaCost` is its only caller, reached from `updateManaCostForView()` at four event sites; the UI reads the cached `TrackableProperty` rather than recomputing per frame. |
| 4 | Trigger the `setZone` update only on relevant zone changes | `setZone` already returns early on `if (currentZone == zone)` before reaching the new call (`Card.java:6929`), so the update fires only on an actual zone change. Restricting it further to the command zone would leave stale tax displayed after a commander moves out of it, which is the case the call exists to cover. |
| 5 | Add unit tests for multi-digit generic, X and hybrid costs | Each named case is settled by construction rather than by chance, as covered under 1. Exercising `getPerpetualAdjustedManaCost` also needs a full `Card`, `CardState`, `Player`, `Zone` and `Game` fixture, unlike the pure-value style of the existing `ManaCostBeingPaidTest`. |

No changes applied.

---

PR #10263, opened 2026-04-03
Hovering over deck name in lobby displays comment from .dck file

agylesox:
Purpose is to show the comment from the .dck file on mouse hover.
This was split out of PR #10232 since it can exist independently.

Screenshot 2026-04-03 at 8 10 40 AM


Jetz72 (approved):
Worth thinking about if there's a way to display this info in the LibGDX UI for mobile and adventure. But seems good.

PR #10275, opened 2026-04-04
Add pick two option to draft file

ke7aqm:
- Added code to support Pick Two draft in custom cube
- Added example cube (Absolute Beginners Cube)
- Updated all cube .draft files to include the new PickTwo field

### DIFF
--- a/forge-gui/res/draft/cube_skiera.draft
+++ b/forge-gui/res/draft/cube_skiera.draft
@@ -6,3 +6,4 @@ LandSetCode:M11
 Booster: 16 Any
 NumPacks: 3
 NumPlayers: 8
+PickTwo: False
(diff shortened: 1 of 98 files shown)


forge-gui/res/draft/cube_skiera.draft
```diff
@@ -6,3 +6,4 @@ LandSetCode:M11
 Booster: 16 Any
 NumPacks: 3
 NumPlayers: 8
+PickTwo:False
```
Agetian: Stylistically, can we keep the space after : for the new option for consistency?

ke7aqm: I can definitely do that. I did notice that the Singleton boolean didn't have the space, which is why I kept it that way, but I can add it and update it.

Jetz72:
Should this maybe be an optional parameter, rather than included as explicitly set to false in every cube file that doesn't use this modifier?

ke7aqm:
That is definitely an option. I have chosen to explicitly include it so it is exposed as an option when someone looks at other .draft files to add their own cubes. If there is a place that has documentation for all of the options, we could just add it there, but for now, I'd rather just update all of the .draft files to include it for future reference to those adding custom cubes.

PR #10279, opened 2026-04-04
Added lands to library

AncoronDS:
Added lands to library so AI doesn't draw from empty library


Agetian:
Looks good, the original puzzle description doesn't mention these lands, but without them, the AI simply draws into an empty library when any instant/sorcery is cast, so it was most certainly implied

Issue #10233, opened 2026-04-01
SOS: Prepared


Hanmac:
> *(While it’s prepared, you may cast a copy of its spell. Doing so unprepares it.)*

If a permanent with a prepare spell becomes prepared, a copy of its prepare spell appears in exile.

Ruling:
* If you control a prepared creature and another player gains control of it, the player who controls the prepared creature is the one who now can cast the copy of the prepare spell from exile.

It should not count for exiled *cards*.

Some Prepared Spells are existing cards, some are not.
We will see if they print Tokens for the Spells.

~~What we could do:~~
~~make a StaticAction on the Prepared Card to copy and cast the spell from exile.
That might help us from keeping the Spell Copy in Exile.~~

[Rulings](https://magic.wizards.com/en/news/feature/secrets-of-strixhaven-release-notes):

- Preparation cards can only be cast with their base characteristics.
- A preparation card is a creature card in every zone. For example, while it's in your graveyard, Emeritus of Ideation is a blue creature whose mana value is 5. It can't be the target of Zealous Lorecaster's triggered ability ("When this creature enters, return target instant or sorcery card from your graveyard to your hand.").
- As an effect causes a creature with a prepare spell to become prepared (including effects that state that a creature "enters prepared"), that creature's controller creates a copy of that creature's prepare spell in exile. That copy remains in exile for as long as that permanent is on the battlefield and is prepared. That creature's controller may cast that copy as long as it remains in exile. As they cast it, that creature stops being prepared.
- Only the current controller of a prepared creature can cast the copy of its prepare spell in exile. It doesn't matter who owns that creature or who controlled it when it became prepared.
- While copies of spells in zones other than the stack cease to exist when state-based actions are checked, this does not apply to copies of prepare spells in exile that were created as a creature became prepared.
- Being prepared isn't a copiable value. If a permanent becomes a copy of a prepared creature, it won't be prepared. That permanent will have the alternative characteristics, though, and some other effect could cause it to become prepared later.
- If a prepared creature loses all abilities, it won't stop being prepared, and nothing will happen to its alternative characteristics or to the copy of its prepare spell in exile.
- If a prepared creature stops being a creature, it will still be prepared, and the copy of its prepare spell will remain in exile. That permanent's controller will still be able to cast it. The same is true if a prepared creature becomes a copy of something else, even if it's then a permanent without a prepare spell.
- A creature without a prepare spell can't become prepared.
[...]

PR #10288, opened 2026-04-04
Placeholder Faces / CopyFaceFrom

Jetz72:
Adds the syntax `CopyFaceFrom:<card face name>` to card scripts. Doing this will inherit the entire card face defined in a separate card script. A couple playtest split cards are included which use and demonstrate this feature, but it primarily enables Prepare creatures to reference existing spells (#10233).

Updates some parts of the code that use `getAllCards(cardName)` to use `getAllCardsNoAlt(cardName)`, now that the distinction is more meaningful. Added a doc comment on each with detailed examples, but in essence, the former will find any card that has the given name on any of its faces, and the latter will only match a full card name.

### DIFF
--- a/forge-core/src/main/java/forge/card/CardRules.java
+++ b/forge-core/src/main/java/forge/card/CardRules.java
@@ -81,14 +81,7 @@ public final class CardRules implements ICardCharacteristics {
         addsWildCardColor = false;
         setColorID = 0;
 
-
-        //calculate color identity
-        byte colMask = calculateColorIdentity(mainPart);
-
-        if (otherPart != null) {
-            colMask |= calculateColorIdentity(otherPart);
-        }
-        colorIdentity = ColorSet.fromMask(colMask);
+        colorIdentity = calculateColorIdentity(this);
     }
 
     void reinitializeFromRules(CardRules newRules) {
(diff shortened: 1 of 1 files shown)


forge-core/src/main/java/forge/card/CardRules.java
```diff
@@ -506,6 +526,41 @@ public String getDisplayNameForVariant(String variantName) {
         return variantName;
     }
 
+    /**
+     * A card has placeholder faces if its script uses `CopyFaceFrom` to reference another card.
+     * These will be filled in via `supplyPlaceholderFaces` after all scripts have been processed.
+     */
+    /* package */ boolean hasPlaceholderFaces() {
```
Hanmac, line 535: should these "package" be removed and be public instead?

forge-core/src/main/java/forge/card/CardRules.java
```diff
@@ -506,6 +526,41 @@ public String getDisplayNameForVariant(String variantName) {
...
+                this.otherPart = face;
+        }
+        this.allFaces = Collections.unmodifiableList(newFaceList);
+
+        //Recalculate color identity.
+        byte colMask = calculateColorIdentity(mainPart);
+        if (otherPart != null) {
+            colMask |= calculateColorIdentity(otherPart);
+        }
+        this.colorIdentity = ColorSet.fromMask(colMask);
```
Hanmac: maybe move this part to extra function?

forge-core/src/main/java/forge/card/CardRules.java
Jetz72: I don't think there's any system outside of CardDB's initialization that would have any reason to use those methods. `CardRules.placeholderFaces` gets deleted after that point anyway, so `hasPlaceholderFaces` will invariably return false and `supplyPlaceholderFaces` will invariably do nothing. As noted, I'd have made `getPreInitName` package private as well, but `StaticData` needed it as well for its initial cataloging of the `CardRules`.

forge-core/src/main/java/forge/card/CardRules.java
Jetz72: I was thinking about doing that at some point, but moved on to something else before I actually did it. Good call.

Hanmac:
@tool4ever @tehdiplomat can this be merged as requirement for #10233 ?

PR #10291, opened 2026-04-05
feat(draft): replace Cogwork Librarian opt-out dialog with opt-in button

billyhewlett:
## Problem

When a player holds Cogwork Librarian during a cube draft, the current
implementation prompts "Draft an extra pick with Cogwork Librarian?" (Yes/No)
on every single pick until the player chooses to use it. The player must
actively click "No" on each pack they want to draft normally, which is
disruptive when they are waiting for the right pack to use the ability.

Additionally, there is a pre-existing bug: if a player activates Cogwork
and then drafts a card that has a draft action (e.g., Lore Seeker), the
extra pick is silently dropped. `draftCard()` ends with a hardcoded
`return true` when any draft action runs, overriding the `passPack = false`
set by Cogwork's handler.

## Changes

**`LimitedPlayer.java`**
- Add `cogworkLibrarianActivatedByUI` public flag, consumed by
  `handleCogworkLibrarian()` instead of showing a dialog. Defaults to
  `false`, so the ability is not used unless the player explicitly opts in.
- Add `hasCogworkLibrarianAvailable()` predicate so the UI can query
  pending state without touching internal flags.
- Replace `handleCogworkLibrarian()` dialog prompt with flag-read-and-clear.
- Fix `draftCard()` to `return passPack` instead of hardcoded `return true`,
  so Cogwork combined with draft-action cards (Lore Seeker, etc.) works
  correctly.

**`FDeckEditor.java` (mobile)**
- Add a persistent `btnUseLibrarian` button in `DraftPackPage`, hidden by
  default. Appears above the card list when Cogwork is pending.
- Override `doLayout()` to reserve space for the button when visible.
- Button tap sets `cogworkLibrarianActivatedByUI = true` and hides itself.
- Button is hidden at start of `moveCard()` and re-evaluated in `refresh()`
  after each pack loads.

## Behavior Change

| Before | After |
|--------|-------|
| Dialog on every pick while holding Cogwork; must click "No" to skip | Button visible while Cogwork is pending; tap once to use it |
| Extra pick dropped when combined with draft-action cards | Extra pick correctly retained |

## Testing

Draft with a cube containing Cogwork Librarian. Verify:
1. Button appears when Cogwork is in hand.
2. Tapping it grants the extra pick for that pick only.
3. Not tapping it skips the pick silently — no dialog.
4. Drafting Lore Seeker while Cogwork is pending correctly grants both effects.

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/limited/LimitedPlayer.java
+++ b/forge-gui/src/main/java/forge/gamemodes/limited/LimitedPlayer.java
@@ -55,6 +55,9 @@ public class LimitedPlayer {
 
     private int playerFlags = 0;
 
+    /** Set by the draft UI when the player taps "Use Cogwork Librarian" before a pick. */
+    public boolean cogworkLibrarianActivatedByUI = false;
+
     private final List<PaperCard> faceUp = Lists.newArrayList();
     private final List<PaperCard> revealed = Lists.newArrayList();
     private final Map<String, List<String>> noted = new HashMap<>();
@@ -380,7 +383,7 @@ public class LimitedPlayer {
             }
         }
 
-        return true;
+        return passPack;
     }
 
     public void addLog(String message) {
@@ -688,10 +691,18 @@ public class LimitedPlayer {
         return true;
     }
 
+    /** Returns true if the Cogwork Librarian extra-draft opportunity is currently pending. */
+    public boolean hasCogworkLibrarianAvailable() {
+        return (playerFlags & CogworkLibrarianExtraDraft) == CogworkLibrarianExtraDraft;
+    }
+
     public boolean handleCogworkLibrarian(DraftPack pack, PaperCard drafted) {
-        if(pack.isEmpty())
-            return false;
-        return !Objects.equals(SGuiChoose.one("Draft an extra pick with Cogwork Librarian?", Lists.newArrayList("Yes", "No")), "No");
+        if (pack.isEmpty()) return false;
+        // Opt-in: the UI sets cogworkLibrarianActivatedByUI via a button tap before the pick.
+        // If the player did not tap the button this pick, default to not using it.
+        boolean activated = cogworkLibrarianActivatedByUI;
+        cogworkLibrarianActivatedByUI = false; // consume the flag
+        return activated;
     }
 
     public boolean handleAgentOfAcquisitions(DraftPack pack, PaperCard drafted) {
--- a/forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
+++ b/forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
@@ -798,6 +798,13 @@ public class AdventureDeckEditor extends FDeckEditor {
             event.getDraft().setLogEntry(this.draftLog);
             deckHeader.initDraftLog(this.draftLog, this);
         }
+        if (event.getDraft() != null) {
+            deckHeader.initCogworkButton(() -> {
+                event.getDraft().getHumanPlayer().cogworkLibrarianActivatedByUI = true;
+                deckHeader.btnCogwork.setVisible(false);
+                deckHeader.revalidate();
+            });
+        }
     }
 
     public AdventureDeckEditor(Deck deckToPreview) {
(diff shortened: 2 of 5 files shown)


forge-gui/src/main/java/forge/gamemodes/limited/LimitedPlayer.java
```diff
@@ -688,10 +691,18 @@ public boolean handleLeovoldsOperative(DraftPack pack, PaperCard drafted) {
...
+
     public boolean handleCogworkLibrarian(DraftPack pack, PaperCard drafted) {
-        if(pack.isEmpty())
-            return false;
-        return !Objects.equals(SGuiChoose.one("Draft an extra pick with Cogwork Librarian?", Lists.newArrayList("Yes", "No")), "No");
+        if (pack.isEmpty()) return false;
+        // Opt-in: the UI sets cogworkLibrarianActivatedByUI via a button tap before the pick.
+        // If the player did not tap the button this pick, default to not using it.
+        boolean activated = cogworkLibrarianActivatedByUI;
+        cogworkLibrarianActivatedByUI = false; // consume the flag
```
tool4ever, line 704: since it's missing wiring for Desktop it'd break there...

jamincollins:
## Findings
#### 1. Cogwork Librarian opt-in defaults to skip extra draft
The handleCogworkLibrarian method now returns the value of cogworkLibrarianActivatedByUI which defaults to false. Previously the method prompted the user with SGuiChoose.one and defaulted to yes if user did not select No. Changing to opt-in means if the player does not tap the button before the pick, the extra draft is silently skipped even though the ability is active. This changes game rules: Cogwork Librarian's ability is optional but the default should be to ask, not to silently skip.

Test case:
```java
LimitedPlayer p = new LimitedPlayer(0, draft);
p.playerFlags |= 1 << 3; // CogworkLibrarianExtraDraft
p.cogworkLibrarianActivatedByUI = false; // player never tapped button
boolean used = p.handleCogworkLibrarian(pack, card);
assert !used; // extra draft is lost, ability wasted
```
Impact: Ability is effectively disabled unless UI button is pressed. If UI fails to show button or player misses it, extra draft is lost without prompting.

#### 2. cogworkLibrarianActivatedByUI flag not reset on pack change
The flag is consumed in handleCogworkLibrarian by setting it to false, but if the player taps the button and then changes pick or cancels, the flag remains consumed. Also if multiple packs have the ability pending, the flag may be consumed too early.

Test case:
```java
p.cogworkLibrarianActivatedByUI = true;
p.handleCogworkLibrarian(pack1, card1); // consumes flag
// player taps button again for next pick but flag already false
p.handleCogworkLibrarian(pack2, card2); // returns false unexpectedly
```
Impact: Race condition between UI button visibility and draft timing.

#### 3. draftCard return value changed from always true to passPack
The draftCard method now returns passPack instead of always true. This changes contract for callers expecting boolean indicating success. If any caller relies on return true meaning card drafted, it may misinterpret passPack false as failure.

Test case:
```java
boolean result = player.draftCard(card, DeckSection.Sideboard);
// previously result always true if card not null
// now result may be false when extra draft consumes pick
```
Impact: Backwards compatibility risk.

#### 4. Button visibility logic depends on cogworkLibrarianActivatedByUI
FDeckEditor.refresh checks !getDraftPlayer().cogworkLibrarianActivatedByUI to decide visibility. If flag is true from previous activation but ability not yet processed, button may be hidden incorrectly.

Test case:
```java
player.cogworkLibrarianActivatedByUI = true;
player.hasCogworkLibrarianAvailable(); // true
// button visibility = true && !true => false, button hidden while waiting
```
Impact: UI flicker or button hidden at wrong time.

### Performance and security
#### 5. Public mutable flag exposes state to UI
cogworkLibrarianActivatedByUI is public boolean field mutated from UI thread. No synchronization; may cause race with draft thread.

Test case:
```java
// UI thread sets flag to true
// draft thread reads and consumes flag
// no volatile/happens-before guarantee
```
Impact: Possible stale read under concurrent access.

## Verdict
Request Changes
The opt-in change removes the user prompt and defaults to skipping the ability, changing game rules and risking lost extra drafts. The return value change to passPack also alters contract. Needs explicit prompt fallback and proper flag lifecycle.

## Recommended Fixes
1. Keep opt-in button but fallback to prompt if flag not set when ability is pending, preserving rule compliance.
2. Reset cogworkLibrarianActivatedByUI after each pick regardless of activation, and guard button visibility against stale flag.
3. Restore draftCard return semantics or document change.
4. Make cogworkLibrarianActivatedByUI volatile or encapsulate with synchronized access.

---

PR #10292, opened 2026-04-05
fix(draft): Lore Seeker adds cube booster from surplus in cube drafts

billyhewlett:
## Problem

In cube drafts, Lore Seeker's effect ("after you draft this, you may add a
booster pack to the draft") calls `addBooster(CardEdition)`, which opens a
set-selection dialog and injects a booster from an arbitrary retail set.
This breaks the cube draft's card pool contract: players draft a card from
outside the cube.

## Background

In a cube draft, the number of cards used equals (players × packs × pack size).
The remainder of the cube's card pool sits unused after booster initialization.
For example, a 540-card cube with 8 players drafting 3 packs of 15 uses
8 × 3 × 15 = 360 cards, leaving 180 surplus. Lore Seeker should draw its
booster from this surplus rather than from a retail set.

## Changes

**`UnOpenedProduct.java`**
Add `getRemainingCards()`: returns all cards still in the print sheets after
booster draws. Enables callers to inspect surplus without modifying state.

**`BoosterDraft.java`**
- Retain a reference (`cubeProduct`) to the cube's `UnOpenedProduct` during
  `setupCustomDraft()`.
- After `initializeBoosters()` distributes all packs, snapshot remaining
  cards into a shuffled `cubeSurplus` list.
- Override `addBooster()` (no-arg) to draw a pack from `cubeSurplus`.
  Returns `null` when surplus is exhausted.

**`IBoosterDraft.java`**
Add `default DraftPack addBooster()` returning `null` as the base case for
non-cube draft implementations.

**`LimitedPlayer.java`**
In `addSingleBoosterPack()`, try `draft.addBooster()` first. If it returns
a pack (cube surplus available), use it. Fall back to the existing
set-selection dialog only for non-cube drafts or when surplus is exhausted.

## Non-cube drafts

Unaffected. `addBooster()` returns `null` for all non-`BoosterDraft`
implementations, and the fallback path runs as before.

## Testing

Draft with a cube containing Lore Seeker. Verify:
1. Drafting Lore Seeker adds a pack drawn from cube cards (not a retail set dialog).
2. After surplus is exhausted, falls back to the set-selection dialog.
3. Non-cube drafts with Lore Seeker continue to show the set dialog.

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
+++ b/forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
@@ -74,6 +74,10 @@ public class BoosterDraft implements IBoosterDraft {
     static final List<CustomLimited> customs = new ArrayList<>();
     protected LimitedPoolType draftFormat;
 
+    // Surplus cube cards left over after boosters are distributed; used by Lore Seeker
+    private UnOpenedProduct cubeProduct = null;
+    private List<PaperCard> cubeSurplus = null;
+
     protected final List<IUnOpenedProduct> product = new ArrayList<>();
     public static void initializeCustomDrafts() {
         loadCustomDrafts();
@@ -407,6 +411,7 @@ public class BoosterDraft implements IBoosterDraft {
 
         final UnOpenedProduct toAdd = new UnOpenedProduct(tpl, dPool);
         toAdd.setLimitedPool(draft.isSingleton());
+        this.cubeProduct = toAdd; // retain reference to capture surplus after initialization
         for (int i = 0; i < draft.getNumPacks(); i++) {
             this.product.add(toAdd);
         }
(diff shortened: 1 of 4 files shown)


forge-gui/src/main/java/forge/gamemodes/limited/BoosterDraft.java
```diff
@@ -501,9 +506,28 @@ public void initializeBoosters() {
...
     }
 
+    /** Returns a DraftPack drawn from cube surplus cards (for Lore Seeker in cube drafts).
+     *  Returns null if this is not a cube draft or surplus is exhausted. */
+    @Override
+    public DraftPack addBooster() {
+        if (cubeSurplus == null || cubeSurplus.isEmpty()) {
+            return null;
+        }
+        int packSize = Math.min(currentBoosterSize > 0 ? currentBoosterSize : 15, cubeSurplus.size());
```
tehdiplomat, line 525: If currentBoosterSize is 0 or less you are just gonna return the entire remaining cube?

PR #10343, opened 2026-04-09
Extend IdRef serialization from game events to protocol method args

MostCromulent:
@tool4ever re-targeting this at master now that delta is merged. 

Previous PR discussion should be preserved here: https://github.com/MostCromulent/forge/pull/17

---

## Summary

Protocol methods serialize CardView/PlayerView references as full object graphs — every `selectCard`, `selectPlayer`, and similar call sends the entire trackable object tree. This replaces that with lightweight IdRef(typeTag, id) markers at the Netty encoder/decoder level, and consolidates all TrackableObject serialization logic into a single `TrackableSerializer` class.

- **Server→client and client→server** protocol args are replaced with IdRef markers by the encoder and resolved from the Tracker by the decoder
- **DeltaPacket events** are wrapped with IdRef replacement at packet construction time and unwrapped after state application on the client (avoids a decoder timing issue where new objects aren't in the tracker yet during deserialization)
- **GameEventProxy removed** — its IdRef/StaleCardRef replacement logic is folded into TrackableSerializer; its event byte-array wrapping is replaced by `wrapEvents`/`unwrapEvents`
- **findByView fallback** added in `Game.java` — when a zone-specific search misses (e.g. IdRef resolved from a tracker that wasn't updated after a zone change), falls back to global ID search

`setGameView` and `openView` are excluded from replacement since they carry the full state that populates the client's Tracker.

## Design decisions

**Fallback search vs tracker update (findByView):** When the server tracker holds a stale CardView with an old zone (because `updateObjLookup` skips objects whose ID is already registered), `findByView` searches the wrong zone. Rather than modifying the tracker's `!hasObj` guard (shared by delta sync and view updates, high blast radius), a fallback global search is added — it only triggers on a miss, matching the existing null-zone fallback path.

**Event wrap/unwrap timing:** Events in DeltaPacket can't use encoder-level replacement because the decoder runs before `applyDelta` creates new objects in the tracker. Instead, events are serialized to byte arrays with IdRef replacement at packet construction time (server-side), and deserialized with resolution after state application (client-side), when the tracker is fully populated.

**Server vs client encoder modes:** The server encoder uses a full Tracker (set by `RemoteClientGuiGame.setGameView`) for verified replacement with StaleCardRef detection. The client encoder does simple IdRef replacement without a tracker — using one would create StaleCardRef markers that the server resolves as detached CardViews, breaking game object identity.

---


tool4ever:
Planning to merge in around two weeks if nothing else comes up

forge-game/src/main/java/forge/game/Game.java
```diff
@@ -694,8 +694,17 @@ public Card findByView(CardView view) {
...
+            if (p != null) {
+                visit.visitAll(p.getZone(view.getZone()));
+            }
+        } else {
+            forEachCardInGame(visit);
+        }
+        // Zone-specific search may miss if the view has stale zone info
+        // (e.g. IdRef resolved from a tracker that wasn't updated after a
+        // zone change). Fall back to global search.
+        if (visit.getFound() == null) {
```
tool4ever, line 707: need to investigate if we'd also hit this locally:
i.e. if there are situations (viewing face down stuff etc.) where we don't want to always find newer instance!

MostCromulent: AI analysis:

> I don't think the fallback can actually return a "different" Card instance, but it's worth walking through why.
> 
> **Lookup is purely by ID.** `CardIdVisitor` (Game.java:677) matches on `card.getId()` only — no zone/visibility/face-down logic. The fallback is semantically equivalent to `findById(view.getId())`, which already exists and is used elsewhere in this file.
> 
> **Card IDs are unique among Cards reachable by `forEachCardInGame`.** That walker visits Graveyard, Hand, Library, Battlefield, MeldedCards, Exile, Command, InboundTokens, Stack. The only construct in the codebase that creates a second Card with the same ID is `CardCopyService.getLKICopy` (CardCopyService.java:230), and LKI snapshots live in transient trigger/replacement maps — never in zones — so the walker never encounters them. Face-down is a property toggle on a single Card instance, not a distinct Card, so face-down/face-up doesn't produce two ID-matching candidates either.
> 
> So when both paths succeed, they return the same Card. The only behavioral difference is `null` (old) vs. the real Card (new) when the view's zone is stale — never *wrong Card* vs. *right Card*.
> 
> **Local play:** The miss path shouldn't normally fire locally because the local UI's CardView is the live tracker's CardView and stays in sync via `TrackableProperty.changed`. The bug this fallback addresses is network-specific (server tracker holds a CardView with a stale zone because `updateObjLookup` skips already-registered IDs). If local ever does hit it, the outcome is the same — find the Card by ID — just via the longer path.
> 
> One potential downside is that silent recovery masks bugs. If local play ever starts hitting the fallback, nobody notices. This could be addressed by adding a one-line `WARN` log inside the fallback branch so we'd see it in logs if it ever fires locally.

tool4ever (approved):

forge-game/src/main/java/forge/game/Game.java
```diff
@@ -694,8 +694,17 @@ public Card findByView(CardView view) {
...
         } else if (view.getController() != null && view.getZone() != null) {
-            visit.visitAll(getPlayer(view.getController()).getZone(view.getZone()));
-        } else { // fallback if view doesn't has controller or zone set for some reason
+            Player p = getPlayer(view.getController());
+            if (p != null) {
+                visit.visitAll(p.getZone(view.getZone()));
+            }
+        } else {
+            forEachCardInGame(visit);
+        }
```
Hanmac, line 703: @MostCromulent in bad case, forEachCardInGame is called twice, remove this part

PR #10348, opened 2026-04-10
Documentation update for delta sync + minor testing simplification

MostCromulent:
- Update Network Play wiki page to include network logs info
- Add Network Testing page to Development section of Wiki outlining network testing tools
- Minor network testing simplification: testConfigurableSequential and testConfigurableParallel use the same -Dtest.2pGames/3pGames/4pGames properties as the comprehensive tests via ComprehensiveTestExecutor.fromSystemProperties(). This aligns with description in Network Testing wiki doc.

### DIFF
--- a/docs/Network-Play.md
+++ b/docs/Network-Play.md
@@ -16,20 +16,16 @@
 > [!CAUTION]
 > **Network play is a work-in-progress.** You will encounter bugs. When they appear, they will likely be mid-game and require restarting both the client and host. Please report bugs and issues to help improve stability.
 
-> [!CAUTION]
-> **There is no built-in matchmaking.** Network play is designed for playing against people you know and manually sharing connection details.
-
-> [!TIP]
-> **Cross-platform play is supported.** 
-
 > [!TIP]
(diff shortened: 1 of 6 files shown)


docs/Network-Play.md
```diff
@@ -8,15 +8,19 @@
...
 ---
 
 # Status & Support
 
 > [!CAUTION]
 > **Network play is a work-in-progress.** You will encounter bugs. When they appear, they will likely be mid-game and require restarting both the client and host. Please report bugs and issues to help improve stability.
 
-Cross-platform play is supported.  
-There is no built-in matchmaking. Network play is designed for playing against people you know and manually sharing connection details.
+> [!CAUTION]
```
tool4ever: I don't think we need to flood the reader quite this much with callouts:
This is basic info, not really something dangerous or especially useful...

MostCromulent: Have re-arranged this a bit in most recent commit:
- "there will be bugs" caution and Discord link stay as callouts at the top of page
- cross-platform support moved into the body of the text in the requirements table
- "No matchmaking" tip added to the quick start section, where it probably makes more sense in context.

But let me know if that's not what you were after.

tool4ever (approved):

PR #10353, opened 2026-04-10
devmode: planewalkers ETB with loyalty counters

kojotak:
I was a little bit puzzled, when a planewalker added from devmode disappeared after the first phase change (because of missing loyalty counters). There is already a dedicated dev mode button for adding counters, but I think that this small code change is worth saving users those extra clicks.

Still doesn't work for `Dakkon, Shadow Slayer` 😞.


tool4ever:
hmn ETB replacements are disabled on purpose
try the cast spell button instead

PR #10354, opened 2026-04-10
Fix sendAndWait() timeout NPE for idle network players 

MostCromulent:
Addresses issue in https://github.com/Card-Forge/forge/issues/9861

### Summary

In network multiplayer, `ReplyPool.get()` uses a `CompletableFuture` to block the game thread while waiting for a remote player's response. It had a hardcoded 5-minute timeout — if a player didn't respond in time (e.g., went AFK mid-game), `GameProtocolSender.sendAndWait()` silently returned null. Many callers in `RemoteClientGuiGame` don't handle null (`getAbilityToPlay`, `assignCombatDamage`, `getChoices`, `order`, etc.), so the null propagates into the game engine and causes an NPE. Local and AI games are unaffected — `sendAndWait()` is only used for human players in network multiplayer.

This PR removes the timeout so `future.get()` blocks until the player responds or disconnects. The timeout was added in 2015 as a generic safety valve in the original network implementation (1 minute), then bumped to 5 minutes in 2018 when players hit it during normal play. It predates the heartbeat mechanism, which now provides a tighter safety net: clients send a heartbeat every 15 seconds, and the server's `IdleStateHandler` closes the connection after 45 seconds of silence, triggering `cancelAll()` to unblock any waiting futures. The timeout is redundant for detecting dead connections and harmful for live ones.

### Follow-up PR

The existing timer mechanism is clearly unsafe and this addresses the immediate NPE.

It would still be good to have an optional turn timer for multiplayer, configurable by the host. Hitting the timeout should cause a priority yield rather than returning null and prompting an NPE. A TODO is added at the natural hook point (`RemoteClientGuiGame.sendAndWait()`). Will address in a separate future PR with GUI considerations.

---


forge-gui/src/main/java/forge/gamemodes/net/ReplyPool.java
```diff
@@ -27,13 +25,13 @@ public void complete(final int index, final Object value) {
         }
     }
 
-    public Object get(final int index) throws TimeoutException {
+    public Object get(final int index) {
         final CompletableFuture future;
```
tool4ever, line 29: I'm pretty sure `getAbilityToPlay` should be fine?
so we could at least add overload here when it's that ProtocolMethod and connect it to a simple AFK timeout preference
(and just catch the Exception directly)

MostCromulent: On looking into this it turns out the `sendAndWait(getAbilityToPlay)` situation is a subset of AFK scenarios.

> When a remote player sits on priority doing nothing, the server isn't blocked in `sendAndWait` — it's blocked in `InputPassPriority.showAndWait()`, which waits on a `CountDownLatch` in `InputSyncronizedBase.awaitLatchRelease()`. The client drives the game forward by sending async fire-and-forget actions (`passPriority`, `selectCard`, etc.) that release the latch on the server side. `getAbilityToPlay` is only invoked later, as a sub-prompt after a player has actively clicked a card with multiple abilities to pick from. So an AFK timeout on `sendAndWait` alone would only catch the "clicked a card then walked away" case, not the more common "never clicked anything" case.

In planning implementation for your comment above the AI ended up developing a solution that covers both the `sendAndWait` and `InputPassPriority.showAndWait()` paths.

This is broader than what you strictly asked for, but the AFK timeout preference and timer was already wired up for the `sendAndWait(getAbilityToPlay)` suggestion so this avoids applying that preference only to a niche situation which doesn't cover the main scenario players would probably call AFK.

> Arm an `AfkTimeout` at `InputPassPriority.showAndWait()` (the real blocking point) whenever `FServerManager.isHosting()` is true — symmetric for the local host and any remote player, either of whom can hold the game up. On timeout the server broadcasts an auto-pass message, calls `input.stop()` to release the latch, and also calls `ReplyPool.cancelAll()` on the player's channel — which cascades the fix into the original niche `getAbilityToPlay` case for free, so nothing is lost from your suggestion. If the player stays flagged, subsequent waits drop from the full preference value to a 10-second repeat and broadcast nothing, so phases fly through until the player takes any action. Setting is added as `NET_AFK_TIMEOUT` prefs entry (minutes, default 5, 0 = disabled) with settings UI on desktop and mobile. Scope kept narrow: only priority passes are covered. Extending to other server-side waits (`assignCombatDamage`, `getChoices`, `order`, etc.) is blocked on those methods not being null-safe — documented as a warning in the `armAfkTimeout` Javadoc so the next person to touch this knows why.

This also prints timeout warnings to multiplayer chat so players have visibility when they're about to hit the timer (30 seconds before the first auto-pass).

So long story short: implementation in latest commit is broader than you asked for, but at this stage the AI has already done the work so figured I should at least provide it for review rather than throw it out.

Obviously if you want this wound back to just the `sendAndWait` scenario can still do that.

forge-gui/src/main/java/forge/gamemodes/match/input/InputPassPriority.java
```diff
@@ -51,6 +53,19 @@ public InputPassPriority(final PlayerControllerHuman controller) {
         super(controller);
     }
 
+    @Override
+    public void showAndWait() {
+        final FServerManager server = FServerManager.getInstance();
+        final AfkTimeout timeout = server != null
```
tool4ever, line 59: need to check that this Input really belongs to the network game
(host might start another local one to pass time while waiting for this friend to finish dinner)

MostCromulent:
`sendAndWait`timeout/NPE would happen after 5 mins, but our AFK timer captures broader scenario - so is a behaviour change compared to status quo. 

Should production default preference be AFK timer disabled, or 5 mins?

tool4ever (approved):
seems ok to have it enabled unless user feedback will somehow be against it

Issue #10345, opened 2026-04-10
Auto-yield does not work for clients


DawnTheDurr:
In multiplayer, auto-yields only work for the host of the match, not any of the clients. This includes always yes and always no for optional cards

**Expected behavior**
Right clicking a card in stack and turning on auto-yield should auto pass priority like it does for hosts.
Always yes/no should also auto pass for optional effects

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Browser: Firefox
 - Version 2.0.12-snapshot-04.09 (but has been present since i started using forge at 2.0.11)

MostCromulent:
Bug report is a duplicate of https://github.com/Card-Forge/forge/issues/3699, fix already in progress in https://github.com/Card-Forge/forge/pull/9643

@tool4ever if you want I can just cherrypick the auto-yield fix from that PR (commits https://github.com/Card-Forge/forge/commit/bd8343b113298b85d2756fc81c6b95e4647a941d and https://github.com/Card-Forge/forge/pull/9643/changes/4846a37a0906507a0510e8c3201337f83628df2e) since its a discrete issue with pretty small diff? Only included it there because we were already working in that space when issue was raised.

tool4ever:
yea makes sense to do that
ideally some of the extra calls can be moved into the existing ones though 🤔

PR #10355, opened 2026-04-10
Add network sync for auto-yield and trigger accept/decline

### LINKED ISSUE #3699: Auto Yield does not work in multiplayer

MostCromulent:
Cherry-picked from https://github.com/Card-Forge/forge/pull/9643 with architectural changes per [tool4ever's feedback](https://github.com/Card-Forge/forge/issues/10345#issuecomment-4226941034).

Closes https://github.com/Card-Forge/forge/issues/3699

### Summary

Client-side auto-yield and trigger always-yes/always-no settings had no effect for remote network clients because preferences were only stored locally. This adds `notifyAutoYieldChanged` and `notifyTriggerChoiceChanged` protocol methods to sync these settings from client to server.

Rather than adding separate notification calls at each GUI call site (VStack, VAutoYields, VGameMenu across desktop and mobile), notifications are folded into `AbstractGuiGame`'s existing setter methods (`setShouldAutoYield`, `setShouldAlwaysAcceptTrigger`, etc.). An `instanceof NetGameController` check ensures notifications only fire on the client side, which also avoids re-entrancy when the server-side `PlayerControllerHuman` applies the received state back through the same setters.

Trigger choices are carried as a `TriggerChoice` enum (ASK / ALWAYS_YES / ALWAYS_NO) rather than bare ints.

---


forge-gui/src/main/java/forge/interfaces/IGameController.java
```diff
@@ -52,4 +53,9 @@ public interface IGameController {
      */
     void requestResync();
 
+    /** Notify server that auto-yield was toggled for an ability key. */
+    default void notifyAutoYieldChanged(String key, boolean autoYield) { }
```
tool4ever: I'm wondering if it wouldn't be better to move the whole implementation from IGuiGame to this interface / PlayerControllerHuman as it seems weird to basically duplicate the set-methods:
- at the same time relocating where the data is stored would lead to slightly different functionality in how yields are grouped...?
- but that might actually be another desirable improvement - e.g. if two humans play locally against each other they probably have rather different opinions about what should be yielded instead of sharing them? 🤔

---
I think the approach for the new YieldController already always separates each player:
a) so that might enable it to integrate this part more easily
b) or be an argument for keeping this feature on its current GUI level (related to my comment on that PR about the usefulness of having the preference gate)

tool4ever:
yea, looks better now imo

I believe this also fixed the "disable all" not working on mobile, since that was for some reason interacting with the previously unused controller field

possibly we can still reduce the footprint a bit by refactoring Accept/Decline/Ask methods pattern to the earlier approach, but maybe it's better to wait with that until we have a better understanding on how it should integrate with the more extended yield framework in the future :thinking:

PR #10357, opened 2026-04-11
Secrets draft

tehdiplomat:


Agetian:
Are the rankings available already or it's too early for that?

tehdiplomat:
@Agetian Yea i was having some trouble pulling it last night, somehow some funky file characters were in the base ranking file, but i just pushed it now.

PR #10361, opened 2026-04-11
Add snow-covered basic lands option to Add Basic Lands dialog

vanja-ivancevic:
- Adds a "Snow-Covered" checkbox next to the Land Set selector in the Add Basic Lands dialog (both desktop and mobile)
- Appears only when the limited pool contains Ice Age (including Ice Age block drafts with Alliances/Homelands sideboosters, as long as ICE itself is in the pool)
- Modern Horizons and Masters Edition II are intentionally excluded — they already distribute snow basics through their pack slots, so the UI workaround is unnecessary there
- The checkbox is auto-disabled when the currently selected dropdown edition is not Ice Age, and auto-unchecks if the user navigates away
- Constructed deckbuilders and any other caller of the existing constructors see no behavior change; the new behavior is opt-in via a new constructor parameter passed by \`CEditorLimited\` (desktop) and \`FDeckEditor.showAddBasicLandsDialog\` (mobile)
- Motivation: in Ice Age, snow basic lands historically only shipped in the 60-card starter decks and never in boosters, leaving "snow matters" cards practically useless in current Forge limited tournaments

### DIFF
--- a/forge-core/src/main/java/forge/card/CardEdition.java
+++ b/forge-core/src/main/java/forge/card/CardEdition.java
@@ -1072,27 +1072,4 @@ public final class CardEdition implements Comparable<CardEdition> {
         }
         return true;
     }
-
-    /**
-     * True when this edition contains all five snow-covered basic lands at
-     * BasicLand rarity. Sets like Coldsnap or Kaldheim print snow basics at
-     * common rarity (where they appear in boosters as part of the normal pack
-     * distribution) and are intentionally excluded - the limited deckbuilder
-     * only needs to surface snow basics for sets like Ice Age or Modern
-     * Horizons where they were treated as basics but never appeared in packs.
-     */
-    public boolean hasSnowBasicLands() {
-        for (String landName : MagicColor.Constant.SNOW_LANDS) {
-            boolean foundAtBasicRarity = false;
-            for (EditionEntry entry : this.getCardInSet(landName)) {
-                if (entry.rarity() == CardRarity.BasicLand) {
-                    foundAtBasicRarity = true;
-                    break;
-                }
-            }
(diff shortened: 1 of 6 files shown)


tehdiplomat:
No. In limited formats snow lands were never free.

vanja-ivancevic:
> No. In limited formats snow lands were never free.

Please see the discussion in the old-border chat: https://discord.com/channels/267367946135928833/1474576995211018473

For Ice-Age, there is currently no way to get snow-covered lands in sealed/draft in forge. In historical paper mtg during ICE block, players would've had to get the SC lands from outside if they were playing booster drafts. The SC lands only showed up in the 60-card starter packs.

tehdiplomat (changes requested):
This doesn't represent how these limited environments actually functioned. Snow lands needed to be drafted in MH or KHM which is why they are so valuable if you want a snow themed deck. If our Sealed decks aren't generating with Snow lands for IA we should just fix that.

vanja-ivancevic:
> This doesn't represent how these limited environments actually functioned. Snow lands needed to be drafted in MH or KHM which is why they are so valuable if you want a snow themed deck. If our Sealed decks aren't generating with Snow lands for IA we should just fix that.

Ok, i'm not familiar with the modern formats. I'll change it to affect just ICE then.

Agetian:
I think it's fine for ICE since it's probably about the only way that format can be "fixed", and yeah, it definitely shouldn't be applied to the other formats with snow-covered lands, as the newer ones make it relevant in the drafting phase itself :)

Agetian (approved):

vanja-ivancevic:
Closing this in favor of a different approach

PR #10368, opened 2026-04-12
Fix card loading crash

autumnmyst:
#10288 added a post-loop supplyPlaceholderFaces step for CopyFaceFrom cards, but two gaps cause StaticData init to NPE leaving commonCards null and breaking desktop tests whenever a placeholder card lands in a funny/silver-bordered set.
1. CardRules.getType() is not pre-init safe. While getPreInitName()/isVariant() both tolerate a null mainPart, getType() results in StaticData.<init>'s card.getType().isBasicLand() to NPE. Fixed by returning an empty CardType when mainPart is null, and falling back to mainPart.getType() if otherPart is null in COMBINE mode.
2. CardDb constructor skips placeholder for filtered cards. The filteredCards continue jumps over needsPlaceholderFaces.add(rule), so filtered placeholder cards never get their faces supplied but they're still in rulesByName, and initialize -> addSetCard -> cr.getName() later crashes. Fix: collect placeholder rules before the filter check.


Hanmac (approved):

tool4ever:
good catch before this landed in snapshot!

PR #10372, opened 2026-04-12
fix(draft): conspiracy cards work in custom cube drafts

billyhewlett:
## Problem

Conspiracy cards (Backup Plan, Double Stroke, Advantageous Proclamation,
etc.) are silently ignored when drafting a custom cube that includes them.
Three separate gaps in the code combine to produce this behaviour:

1. **Cards never enter the draft pool** — `CustomLimited.parse()` builds
   the card pool from `deckCube.getMain()` only, ignoring
   `DeckSection.Conspiracy`. Conspiracy cards therefore never appear in
   any pack.

2. **Effects never trigger** — `GauntletMini.startRound()` creates
   `RegisteredPlayer` objects but never calls `assignConspiracies()`.
   The Load Draft screen's single-match path already calls it correctly;
   Gauntlet mode (the default for Draft) was the missing path.

3. **Cards can't be moved into the Conspiracy section** — In
   `FDeckEditor`, `CatalogPage.onCardActivated()` and `buildMenu()` call
   `getPageForSection(destination)` without `forceCreateIfAbsent=true`.
   The Conspiracy tab is never created by default, so `destinationPage`
   is `null` and the card silently fails to move into
   `DeckSection.Conspiracy`. `assignConspiracies()` therefore finds
   nothing even when fix 2 is applied.

## Changes

**`CustomLimited.java`** (`forge-gui`)
- When a `DeckSection.Conspiracy` section is present, merge it into the
  card pool alongside `deckCube.getMain()`.

**`GauntletMini.java`** (`forge-gui`)
- Call `pl.assignConspiracies()` for each player in `startRound()`,
  mirroring the call already present in `LoadDraftScreen`.

**`FDeckEditor.java`** (`forge-gui-mobile`)
- Pass `forceCreateIfAbsent=true` to `getPageForSection()` in both
  `onCardActivated()` (tap) and `buildMenu()` (long-press) inside
  `CatalogPage`. The Conspiracy tab is now created on demand when the
  player first drafts a conspiracy card.

## Testing

Draft a cube containing at least one Backup Plan and one Double Stroke.
Verify:
1. Backup Plan and Double Stroke appear in packs during the draft.
2. After drafting Backup Plan, it appears in the Conspiracy section of
   the deck builder (tap or long-press to move it).
3. Start a game — Backup Plan draws an extra opening hand to choose from.
4. Double Stroke's hidden agenda prompt appears at game start.


forge-game/src/main/java/forge/game/extrahands/BackupPlanService.java
```diff
@@ -26,9 +26,16 @@ public boolean initializeExtraHands() {
...
             return multipleHands;
         }
+        final int handSize = player.getMaxHandSize();
         for(PlayerZone extraHand : player.getExtraZones()) {
             if (extraHand.getZoneType() == ZoneType.ExtraHand) {
-                player.drawCards(7, extraHand);
+                // Only draw an extra hand if there are enough cards in the library.
+                // Multiple Backup Plans in test decks (or edge cases) could exhaust
+                // the library and incorrectly trigger a Turn-0 game loss.
+                if (player.getZone(ZoneType.Library).size() < handSize) {
```
tool4ever: why go against the rules? 👎
> If, through a combination of Backup Plan and/or other conspiracies, you have fewer cards in your deck than you must draw as the game begins, you'll lose the game when state-based actions are checked during the upkeep of the first turn.

billyhewlett: I have removed the Backup Plans change.  I updated the PR description.

forge-gui-mobile/src/forge/deck/FDeckEditor.java
```diff
@@ -1806,7 +1806,7 @@ else if(parentScreen.shouldEnforceConformity()) //If we have a commander, filter
         @Override
         protected void onCardActivated(PaperCard card) {
             DeckSection destination = DeckSection.matchingSection(card);
-            final DeckSectionPage destinationPage = parentScreen.getPageForSection(destination);
+            final DeckSectionPage destinationPage = parentScreen.getPageForSection(destination, true);
```
Jetz72: Using `forceCreateIfAbsent` would cause the section to be created even if the deck format does not allow that section. Limited formats should already support the Conspiracy section, by way of `DeckEditorConfig.getExtraSections`. Should be in the `hiddenExtraSections` list, no?

billyhewlett: Updated to follow the same pattern as setSelectedSection(): check hiddenExtraSections first, call showExtraSectionTab() to promote it if present, then retrieve the page normally without forceCreateIfAbsent.

Jetz72: `DeckSectionPage.updateCaption` should already cause a hidden tab to appear when a card is added to it?

Jetz72 (approved):

PR #10392, opened 2026-04-13
Fix Android crash: guard ProcessHandle in NetworkLogWriter

### LINKED ISSUE #10391: Java crash after last update
After entering the game, I entered Shandalar and find an enemy. The ante goes fine, but as soon as we both draw our starting hand the game crashes and gives a massive Java script error.

**To Reproduce**
As far as I can tell anything done to challenge any NPC is met with this crash. This is on a backed up save file, one that I made right before this update (in case this exact thing happened and I needed to reinstall forge, I didn't want to lose my progress). I Uninstaller and reinstalled, then restored my 2 saves. Both experience crashes, as well as if I start a new game. Unknown if i do a fresh reinstall and start a fresh game if that would make a difference. 

Some notable settings:

Im using Genetic AI
Ante is on
Basic lands are allowed
All variants are selected, and rebalenced cards are also allowed.

**Expected behavior**
A massive Java script error. See attached screenshot. I am unable to view the full crash error in any way.

**Screenshots**

 - Device: [Galaxy S22
 - OS: [unk]
 - Version [unk]
 - not tech literate,  sorry :l

Image

MostCromulent:
`ProcessHandle.current().pid()` is unavailable on Android's ART runtime. After #10377 registered `NetworkLogWriter` in the Android service file, the writer is now discovered on Android — but `writeSystemInfoHeader()` crashes with `NoClassDefFoundError` on the first `netLog.trace()` call. Guard the PID line so it's skipped on Android; the rest of the log header writes normally.

Built and tested on Android emulator — confirmed matches start without crash and network logs are written.

**Note:** `NetworkLogWriter` is registered unconditionally via `tinylog.properties` and `netLog.trace()` calls fire in shared code paths (`InputSyncronizedBase`, `PlayerControllerHuman`), so the writer creates log files even for local AI games. This PR addresses the immediate crash; as follow-up should investigate whether tinylog writers can be registered conditionally so network logging only activates for network games.

---

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/net/NetworkLogWriter.java
+++ b/forge-gui/src/main/java/forge/gamemodes/net/NetworkLogWriter.java
@@ -106,12 +106,14 @@ public class NetworkLogWriter extends AbstractFormatPatternWriter {
 
     private void writeSystemInfoHeader(BufferedWriter writer, String key) {
         try {
-            long pid = ProcessHandle.current().pid();
             StringBuilder sb = new StringBuilder();
             sb.append("=".repeat(80)).append("\n");
             sb.append("Network Debug Log Started\n");
             sb.append("Log file key: ").append(key).append("\n");
-            sb.append("PID: ").append(pid).append("\n");
+            try {
+                long pid = ProcessHandle.current().pid();
+                sb.append("PID: ").append(pid).append("\n");
+            } catch (NoClassDefFoundError e) { /* unavailable on Android */ }
             try {
                 String hwInfo = GuiBase.getHWInfo()
                         .replace("##########################################", "");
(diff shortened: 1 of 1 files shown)


forge-gui/src/main/java/forge/gamemodes/net/NetworkLogWriter.java
```diff
@@ -106,12 +106,14 @@ private BufferedWriter createWriter(String key) {
...
-            long pid = ProcessHandle.current().pid();
             StringBuilder sb = new StringBuilder();
             sb.append("=".repeat(80)).append("\n");
             sb.append("Network Debug Log Started\n");
             sb.append("Log file key: ").append(key).append("\n");
-            sb.append("PID: ").append(pid).append("\n");
+            try {
+                long pid = ProcessHandle.current().pid();
+                sb.append("PID: ").append(pid).append("\n");
+            } catch (NoClassDefFoundError e) { /* unavailable on Android */ }
```
tool4ever: if we know it's unavailable we really don't need to force an exception into the JVM on startup at all
or is this before `GuiBase.isAndroid` is valid?

MostCromulent: Crash happens on match start `soGuiBase.isAndroid` can be used.

tool4ever (approved):

PR #10397, opened 2026-04-14
Try to get find cards in same set first

clairchiara:
If card is not found with artIndex and/or collectorNumber, try again to find it in the same set without them. Only then fallback to full database if still not found.

_Example: the 40K commander precon "Ruinous Powers" contains `Mountain|40K|3`. Although the 40K set has Mountains, it does not have one at artIndex 3. Currently this causes another Mountain to be loaded according to user art preferences, either from LEA or the latest set (eg TMT). A more logical behavior would be to load another Mountain from 40K instead, as these are available._

### DIFF
--- a/forge-core/src/main/java/forge/card/CardDb.java
+++ b/forge-core/src/main/java/forge/card/CardDb.java
@@ -672,12 +672,28 @@ public final class CardDb implements ICardDatabase, IDeckGenPool {
             CardEdition edition = editions.get(reqEditionCode.toUpperCase());
 
             PaperCard cardFromSet = this.getCardFromSet(request.cardName, edition, request.artIndex, request.collectorNumber, request.isFoil);
-            if(cardFromSet != null && request.flags != null)
+            if (cardFromSet != null && request.flags != null) {
                 cardFromSet = cardFromSet.copyWithFlags(request.flags);
+            }
 
-            if (cardFromSet != null)
+            if (cardFromSet != null) {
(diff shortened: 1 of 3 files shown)


Jetz72:
I'd think if the card exists in the set but we've requested an invalid art index or collector number, that's a sign the problem is primarily with the request. In this case, it seems like that deck should be using art index 2. Note how they both have text in the bottom left indicating which deck they're meant for:

Art index 1 - [CN315](https://scryfall.com/card/40k/315/mountain)
Art index 2 - [CN316](https://scryfall.com/card/40k/316/mountain)

Falling back to the same set upon a failed request does make a kind of sense, but I think it'd make it harder to catch other mistakes like these.

clairchiara:
@Jetz72 
That makes sense, I thought about adding a warning log in the console but it seems to happen for _a lot_ of cards...

Jetz72:
> That makes sense, I thought about adding a warning log in the console but it seems to happen for _a lot_ of cards...

The warning might be more appropriate, then, so we can track down those invalid card requests. If a card request comes with a specific set and collector number or art index, and the card isn't there, then either the request is erroneous or the edition is missing an entry. Either way, picking the next closest card with no warning means some issue gets covered it.

Eradev:
ArtIndex should be replaced by their collector numbers instead. Everywhere that's still using ArtIndex is just not updated.

ie. 8 Mountain|40K|[316]

Issue #8126, opened 2025-07-19
Modularize Custom Editions


Jetz72:
Right now, custom editions are structured as `custom/cards/SET/[card script text files]` and `custom/editions/editionFile.txt`. (I think there's also `custom/tokens/SET/...`?) The only way to add images that I'm aware of is manually adding them to Forge's image cache. This in particular makes installing and uninstalling custom sets cumbersome (and on Android, impossible without a powerful file manager).

We could reorganize this, so that each custom edition has a singular directory. `custom/cardSets/SET/` could act as a root folder for a single edition, containing: 
* `/edition.txt` - Main edition file describing prints and metadata.
* `/cardScripts/` and `/tokenScripts/` - Folders containing all card and token scripts for the cards in this set.
* `/cardImages/` and `/tokenImages/` - For image files, stored by name and/or collector number. We might make the `.fullborder` clause of the names optional while doing this. Should probably figure this out once ImageKeys are done being revised.
* `/TypeLists.txt`, `/blocks.txt`, /cubes/[deck lists]` - Various other things that might be added to Forge's default data and resources from this edition. We might even support edition directories that don't contain cards or edition data.

That would make it simpler to install them, but there's room for other improvements that build on top of this structure, such as:
* Allow both `/SET/` and `/SET.zip` to function similarly, so you can either drop the zip file in the directory and be done with it, or leave them as folders so they're easier to edit.
* Wrap the custom cards and images used by adventure and quest into modules that behave similarly.
* Add an in-game menu for enabling and disabling custom sets.
* Create a script structure for defining custom keywords or state-based actions that unwrap as abilities.
* Split Un-cards and Unknown Event cards into custom editions? This would let users disable them and keep their creature types from appearing in game. Potential complications though are Steamflogger Boss and Discord, Lord of Disharmony. The former is a legal card in an un-set, the latter is an un-card in a mostly legal set.
* Split Mystery Booster playtest cards into a custom edition? Similar problem if we want to try and keep the rest of the Mystery Booster sets intact.

Eradev:
There's also a need to think about how to handle cards with the same name as existing ones, but not using the same card rule.

Jetz72:
Hmm. That is a problem we've seen with the Pick Your Poison and Red Herring playtest cards. One option would be to rework the card database to catalog rules from different modules independently from one another, so the paper cards loaded from that edition only use the rules defined in that set. 

[...]

PR #10398, opened 2026-04-14
Add plane-scoped adventure data overrides (blocks + editions)

vanja-ivancevic:
Adventure mods can drop partial `blockdata/blocks.txt` and `editions/*.txt` into their adventure folder. Entries are held in a plane-scoped `AdventureOverrides` registry; nothing mutates upstream singletons, so classic-mode screens accessed mid-Adventure keep reading unmodified upstream data.

- Overlay files parse through the existing `CardBlock.Reader` / `CardEdition.Reader`, syntax matches upstream.
- `CardBlock` gains a neutral `setBoosterResolver` extension hook; overlay blocks install a resolver that reads the registry, so `BoosterDraft.setupPacks` picks up the plane-local template without touching `StaticData.getBoosters()`.
- Adventure consumers (`AdventureEventData`, `AdventureEventController`) query the registry with upstream fallback.
- Classic-mode callers are unchanged.

Demonstrated with two Shandalar Old Border overlays (drop Coldsnap from the Ice Age block, add a snow-covered basic slot to ICE boosters):

**`Shandalar Old Border/blockdata/blocks.txt`**
```
Ice Age, 3/6/ICE, ICE ICE ALL
```

**`Shandalar Old Border/editions/Ice Age.txt`**
```
[metadata]
Code=ICE
Booster=10 Common, 3 Uncommon, 1 Rare, 1 name("Snow-Covered Plains|ICE", "Snow-Covered Island|ICE", "Snow-Covered Swamp|ICE", "Snow-Covered Mountain|ICE", "Snow-Covered Forest|ICE")
```

### DIFF
--- a/forge-core/src/main/java/forge/item/generation/BoosterGenerator.java
+++ b/forge-core/src/main/java/forge/item/generation/BoosterGenerator.java
@@ -636,15 +636,7 @@ public class BoosterGenerator {
                 List<PaperCard> srcList = new ArrayList<>();
 
                 for (String cardName: cardNames) {
-                    // "Card Name|SET" pins a specific printing; bare name uses the default.
-                    int pipeIdx = cardName.indexOf('|');
-                    if (pipeIdx > 0) {
-                        srcList.add(StaticData.instance().getCommonCards().getCard(
-                                cardName.substring(0, pipeIdx).trim(),
-                                cardName.substring(pipeIdx + 1).trim()));
-                    } else {
-                        srcList.add(StaticData.instance().getCommonCards().getCard(cardName));
-                    }
+                    srcList.add(StaticData.instance().getCommonCards().getCard(cardName));
                 }
 
                 src = srcList;
--- a/forge-core/src/main/java/forge/item/generation/BoosterGenerator.java
+++ b/forge-core/src/main/java/forge/item/generation/BoosterGenerator.java
@@ -636,7 +636,15 @@ public class BoosterGenerator {
                 List<PaperCard> srcList = new ArrayList<>();
 
                 for (String cardName: cardNames) {
-                    srcList.add(StaticData.instance().getCommonCards().getCard(cardName));
+                    // "Card Name|SET" pins a specific printing; bare name uses the default.
+                    int pipeIdx = cardName.indexOf('|');
+                    if (pipeIdx > 0) {
+                        srcList.add(StaticData.instance().getCommonCards().getCard(
+                                cardName.substring(0, pipeIdx).trim(),
+                                cardName.substring(pipeIdx + 1).trim()));
+                    } else {
+                        srcList.add(StaticData.instance().getCommonCards().getCard(cardName));
+                    }
                 }
 
                 src = srcList;
--- a/forge-gui-mobile/src/forge/adventure/util/Config.java
+++ b/forge-gui-mobile/src/forge/adventure/util/Config.java
@@ -554,7 +554,7 @@ public class Config {
     }
 
     public void loadResources() {
-        AdventureOverrides.instance().load(prefix, FModel.getMagicDb().getEditions());
+        AdventureOverrides.instance().load(prefix, FModel.getMagicDb().getEditions(), configData);
         RewardData.getAllCards();//initialize before loading custom cards
         final CardRules.Reader rulesReader = new CardRules.Reader();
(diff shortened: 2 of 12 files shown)


forge-core/src/main/java/forge/item/generation/BoosterGenerator.java
```diff
@@ -636,7 +636,15 @@ public static PrintSheet makeSheet(String sheetKey, Iterable<PaperCard> src) {
...
-                    srcList.add(StaticData.instance().getCommonCards().getCard(cardName));
+                    // "Card Name|SET" pins a specific printing; bare name uses the default.
+                    int pipeIdx = cardName.indexOf('|');
+                    if (pipeIdx > 0) {
+                        srcList.add(StaticData.instance().getCommonCards().getCard(
+                                cardName.substring(0, pipeIdx).trim(),
+                                cardName.substring(pipeIdx + 1).trim()));
+                    } else {
+                        srcList.add(StaticData.instance().getCommonCards().getCard(cardName));
+                    }
```
Jetz72: Shouldn't `getCard` already be applying `CardRequest` syntax to its search? I think if you split it up here, it'll end up joined right back together before processing the request anyway.

forge-gui-mobile/src/forge/adventure/util/Config.java
```diff
@@ -555,6 +555,17 @@ public String getCommanderPreconDeckPath(int deckIndex) {
...
+        // Called here (after FModel.initialize) so MagicDb and preferences are available.
+        // Absent files fall through; errors are logged but don't break adventure startup.
+        File blocksOverlay = new File(prefix + "blockdata/blocks.txt");
+        if (blocksOverlay.isFile()) {
+            CardBlock.applyAdventureOverrides(blocksOverlay, FModel.getBlocks(), FModel.getMagicDb().getEditions());
+        }
+        File editionsOverlay = new File(prefix + "editions");
+        if (editionsOverlay.isDirectory()) {
+            CardEdition.applyAdventureOverrides(editionsOverlay, FModel.getMagicDb().getEditions());
+        }
```
Jetz72: Won't this mutate Forge's data outside Adventure if they've gone into Adventure mode and then back to classic?

Jetz72 (commented):
I feel it might be safer to give Adventures the means to supply fantasy blocks, print sheets, and custom boosters in an additive way instead of overwriting the main resources. Maybe coupled with a `restrictedBlocks` parameter in the config so you can subtract them as well?

vanja-ivancevic:
Thanks for the review. Addressed both inline points:

- **`name(|SET)` split**: redundant, `CardDb.getCard(String)` already parses it. Reverted in 10f8210.
- **Mid-session mutation leak**: rewrote in 63726ee. Overrides now live in a plane-scoped `AdventureOverrides` registry; classic-mode code keeps reading unmodified upstream data. `CardBlock` gets a neutral `setBoosterResolver` hook that overlay blocks install so `BoosterDraft` picks up the plane-local template without touching `StaticData`.

On your top-level ask (additive + `restrictedBlocks`): the rewrite nets out partway, nothing mutates upstream and pure additions work, but overlay still replaces upstream blocks by matching name. Happy to go fully additive with a `restrictedBlocks` config if you'd prefer, or is the non-mutating shape acceptable as-is?

Jetz72:
I like the idea of Adventure keeping its custom data separate from the main pool. Even the custom adventure cards don't do this, instead just writing themselves into the game's main card database. Wonder if we could fold those into this kind of Adventure-specific database extension down the line. If we're ever to get to the point where we can switch planes without restarting the application, mutating the base game's data seems like something to avoid.

Eventually, maybe as part of #8126, I'm hoping we can make the ability to composite data from multiple modules (e.g. Adventure Planes, custom content packages) standardized across more of Forge's data sets (e.g. the `CardDB`, edition data, blocks, boosters, formats/legalities, the list of card subtypes...). So we could just load or unload an Adventure plane's content and `StaticData`/`CardDb`/`CardEdition` would update the data they supply accordingly. Dunno if this gets us closer to that but it should work for now.

For now, I'd vote to keep the Adventure Registry for custom blocks and boosters, but generally use it to add definitions rather than shadow existing ones. Then we suppress the ones they're replacing via config like other restrictions.

vanja-ivancevic:
Pushed fa4ec9c + 62c1fb3: overlay blocks are now purely additive. Name collisions with upstream are skipped with a warning, and a new `restrictedBlocks` config entry suppresses upstream blocks at the Adventure layer. Old Border's overlay is renamed to `Ice Age (Old Border)` and its config lists the upstream `Ice Age` under `restrictedBlocks`.

Jetz72 (approved):

PR #10410, opened 2026-04-15
Post-match lobby cleanup and network concede fixes

MostCromulent:
## Summary

Addresses Phase 0.5 of the network draft/sealed implementation plan — hardening existing network match lifecycle before adding new game modes. Also fixes several pre-existing network concede bugs observed during testing. See discussion at [Card-Forge/forge#9861 (comment)](https://github.com/Card-Forge/forge/issues/9861#issuecomment-4234120552) and [follow-up](https://github.com/Card-Forge/forge/issues/9861#issuecomment-4245231008).

## Issues addressed

- **No lobby cleanup after network match** — the previous match's state lingered in the lobby after returning. Added `onMatchOver` callback to reset lobby state (null hostedMatch, un-ready slots, clear player GUIs, push lobby update to clients).

- **Client concede dialog never appeared** — `instanceof PlayerControllerHuman` excluded `NetGameController`, so the concede confirmation was silently skipped. Broadened the check to include network controllers.

- **Client concede force-quit both players** — client sent a premature `nextGameDecision(QUIT)` before the server processed the concede, ending the match for both. Network games now let the server drive the game-end flow.

- **Game thread blocked after concede** — only the conceding player's input latch was released, but the game thread could be waiting on the other player's input queue. Now releases all players' input latches.

- **Remote client never saw win/lose screen** — game-end events were buffered in `GameEventForwarder` and never flushed. Added explicit flush after the game loop exits.

- **NPEs from late-arriving messages** — protocol messages arriving after cleanup caused `NullPointerException` on the server. Added null guards in the protocol handler with null reply for non-void methods.

## Why these changes are safe

- **Lobby cleanup** only runs when a match ends via `NextGameDecision.QUIT` — the same path that already calls `endCurrentGame()`. It adds cleanup *after* existing teardown, doesn't change teardown order.
- **Concede dialog fix** only affects network clients. Local games still go through the `PlayerControllerHuman` path with the per-player outcome check. No behavior change for single-player or local multiplayer.
- **Skipping premature QUIT** only applies to network games (`isNetGame()` guard). Local games still send `nextGameDecision` synchronously as before.
- **Input latch release** is idempotent — `CountDownLatch.countDown()` on an already-released latch is a no-op.
- **Protocol null guards** only fire after cleanup when controllers are already gone. During normal gameplay, `getToInvoke()` and `getClient()` always return non-null.

---

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
+++ b/forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
@@ -527,6 +527,7 @@ public abstract class GameLobby implements IHasGameType {
         //if above checks succeed, return runnable that can be used to finish starting game
         return () -> {
             hostedMatch = GuiBase.getInterface().hostMatch();
+            hostedMatch.setOnMatchOver(this::onMatchOver);
             hostedMatch.startMatch(GameType.Constructed, variantTypes, players, guis);
 
             for (final Player p : hostedMatch.getGame().getPlayers()) {
@@ -540,6 +541,18 @@ public abstract class GameLobby implements IHasGameType {
 
             onGameStarted();
         };
+    }
+
+    protected void onMatchOver() {
+        hostedMatch = null;
+        gameControllers.clear();
+        for (int i = 0; i < getNumberOfSlots(); i++) {
+            final LobbySlot slot = getSlot(i);
+            if (slot != null) {
+                slot.setIsReady(false);
+            }
+        }
+        updateView(true);
     }
 
     public final static class GameLobbyData implements Serializable {
--- a/forge-gui/src/main/java/forge/gamemodes/match/HostedMatch.java
+++ b/forge-gui/src/main/java/forge/gamemodes/match/HostedMatch.java
@@ -56,6 +56,7 @@ public class HostedMatch {
     public HashMap<LobbySlot, IGameController> gameControllers = null;
     private Runnable startGameHook = null;
     private Runnable endGameHook = null;
+    private Runnable onMatchOver = null;
     private final List<PlayerControllerHuman> humanControllers = Lists.newArrayList();
     private Map<RegisteredPlayer, IGuiGame> guis;
     private int humanCount;
@@ -81,6 +82,7 @@ public class HostedMatch {
         startGameHook = hook;
     }
     public void setEndGameHook(Runnable hook) { endGameHook = hook; }
+    public void setOnMatchOver(Runnable callback) { onMatchOver = callback; }
 
     private static GameRules getDefaultRules(final GameType gameType) {
         final GameRules gameRules = new GameRules(gameType);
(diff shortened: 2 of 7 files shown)


forge-gui/src/main/java/forge/gamemodes/match/HostedMatch.java
```diff
@@ -364,6 +377,18 @@ public void endCurrentGame() {
...
+                    for (PlayerControllerHuman allHc : humanControllers) {
+                        allHc.getInputQueue().deleteObserver(fwd);
+                    }
+                }
+            }
+        }
+
         for (final PlayerControllerHuman humanController : humanControllers) {
             if (humanController.getGui() instanceof forge.gamemodes.net.server.RemoteClientGuiGame ngg) {
                 ngg.shutdownForwarder();
```
tool4ever, line 388: can you just pull this shutdown call into the loops above now?

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -3344,6 +3344,15 @@ public void concede() {
         if (player != null) {
             player.concede();
             getGame().getAction().checkGameOverCondition();
+            if (getGame().isGameOver()) {
+                // Release all input latches — the game thread may be blocked
+                // on a different player's input queue than the one who conceded
+                for (Player p : getGame().getPlayers()) {
+                    if (p.getController() instanceof PlayerControllerHuman pch) {
+                        pch.getInputQueue().onGameOver(true);
```
tool4ever, line 3352: what about the same call on `FControlGameEventHandler`:
does this fire earlier and the other become redundant (or still needed for remote client)?

MostCromulent: > Not redundant — keep the loop, but clarify code comment. 
> 
> At HostedMatch.java:238-244 — the subscription is asymmetric: 
> - Local human controller → FControlGameEventHandler subscribed (which unlocks the queue at line 132) 
> - Remote client (server-side proxy, RemoteClientGuiGame) → only GameEventForwarder subscribed, no FControlGameEventHandler 
>  
> So FControlGameEventHandler.onGameOver covers local humans only. For remote-client controllers on the server, no event handler unlocks their InputQueue — the game thread can block on a remote client's latch forever without the explicit loop. Also, FControlGameEventHandler routes the unlock through invokeInEdtLater (async), while the explicit call releases synchronously on the concede thread. 
> 
> InputQueue.onGameOver(true) iterates inputStack and releases latches — idempotent on already-released latches, so overlap with FControlGameEventHandler for local humans is harmless.
> 
> Should tighten the comment in PlayerControllerHuman.concede() to something like "remote-client controllers on the server have no FControlGameEventHandler, so their queues won't be released by the normal event path."

forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
```diff
@@ -542,6 +543,18 @@ else if (autoGenerateVariant != null) {
         };
     }
 
+    protected void onMatchOver() {
+        hostedMatch = null;
+        gameControllers.clear();
+        for (int i = 0; i < getNumberOfSlots(); i++) {
+            final LobbySlot slot = getSlot(i);
+            if (slot != null) {
+                slot.setIsReady(false);
```
tool4ever: let's not break local play like that:
Image

MostCromulent: Oops. 😬
Should now be fixed.

PR #10413, opened 2026-04-15
Working card image downloader

phughk:
Screenshot 2026-04-15 at 15 43 54

Add a section below "Download Skins" called "Download Card Images".

Screenshot 2026-04-15 at 15 45 03

The menu from "Download Card Images" allows set selection, but also scryfall-lang filter. The values are updated dynamically. Downloading images does store the images, but does not refresh the stored/cached information currently. To see up-to-date cached image information after download, the game must be restarted.

Screenshot 2026-04-15 at 15 47 06


phughk:
Notably, the CDN is not used for this bulk downloading. The internal rate limiter seems to be sufficient. If we were to improve this, using the scryfall.io links from the scryfall dataset is necessary. Currently they are not available AFAICT.

tool4ever:
In general I like the idea, preparing for offline holidays etc.

Agetian:
Yeah, looks good to me too :)

Agetian (approved):

tool4ever:
This seems to increasingly cause users to get blocked
please check the limits are still in line with new Scryfall policy or we might need to disable it

phughk:
I've seen the above and will try to tackle it this week 👍

ghost:
Here are responses from Scryfall dev Scarlett: On the Scryfall Discord: https://discord.com/channels/291498816459243521/361685936179904513/1513482345707999313

> Scryfall is heavily cached via Cloudflare, and all images are hosted on a CDN that is essentially free. Whenever someone makes a new request we haven't seen before, or the cached data is old, the server generates new data and that gets passed to Cloudflare and the user, and _that part_ costs bandwidth. It's pretty small, with Cloudflare absorbing almost all our traffic and essentially making it free, but at our scale it adds up.
> In any case if they need to query data on a scale that passes our rate limits, yes they'll need their own server to serve their own data. E.g. Moxfield serves their data from their own servers because of the scale they serve it at; they just periodically update from our servers. For the images they can just use our `*.scryfall.io`, so there won't be any need to mirror that.
> Ok, `/cards/:set/:cn` is not an endpoint we changed the rate limit on. There's no problem there. If it worked historically and was respecting the rate limits, it should work now.

> I remember we did however block some applications recently that were just hammering the API for card data without regard to rate limit and functionally DDoSing us over the thousands of users running that software simultaneously. We heard back from a range of XMage players who turned out to have been blocked. This could have been impacted too. A popular Tabletop Simulator addon also got hit because TTS apparently doesn't actually give them any avenue to properly rate limit requests, at which point that's a “host your own server” thing.

> There's a possibility it was not respecting the rate limits. I see what could be seven simultaneous requests to the same API endpoint, and I don't see timestamps on those lines, but it's possible those are not 50ms apart. This would have lead to the application eventually getting blocked from our API.

> **If they are requesting data on large numbers of cards, then rather than requesting the image one by one, they should be making `/cards/collection` requests for 75 cards per request. They'll get back data and can use that data to directly access (non-rate-limited) CDN URLs. Once they have those, they're in the clear to request cards at whatever rate they like.** (my emphasis)

> `api.scryfall.com` is rate limited.
`*.scryfall.io` is our CDN. All card images and all bulk data will be there. You can access it without rate limit.

> You need to get card image URIs from our API, but once you have them you can just request them freely. For example for kickoff celebrations you need to make one (1) request to <https://api.scryfall.com/cards/4e5590e1-0ac0-4bdd-815b-136bf24ced03?format=json>, and then you get back some image_uris and can store and use those URIs freely.

> If you know the Scryfall ID already (it seems they're going by set/cn only) you can make a guess about where the image will be. For example, let's take a look at kickoff celebrations:
```json
{
  "object": "card",
  "id": "4e5590e1-0ac0-4bdd-815b-136bf24ced03",
  "oracle_id": "94868faf-9e88-4221-8db6-3ad10aa02da5",
  // ... skipping a few lines ...
  "image_uris": {
    "small": "https://cards.scryfall.io/small/front/4/e/4e5590e1-0ac0-4bdd-815b-136bf24ced03.jpg?1738356433",
    "normal": "https://cards.scryfall.io/normal/front/4/e/4e5590e1-0ac0-4bdd-815b-136bf24ced03.jpg?1738356433",
    "large": "https://cards.scryfall.io/large/front/4/e/4e5590e1-0ac0-4bdd-815b-136bf24ced03.jpg?1738356433",
    "png": "https://cards.scryfall.io/png/front/4/e/4e5590e1-0ac0-4bdd-815b-136bf24ced03.png?1738356433",
    "art_crop": "https://cards.scryfall.io/art_crop/front/4/e/4e5590e1-0ac0-4bdd-815b-136bf24ced03.jpg?1738356433",
    "border_crop": "https://cards.scryfall.io/border_crop/front/4/e/4e5590e1-0ac0-4bdd-815b-136bf24ced03.jpg?1738356433"
  },
```
> You can see that the image_uri corresponds to the ID. So if you know the ID (`4e5590e1-0ac0-4bdd-815b-136bf24ced03`) you can guess that the normal image will be here:
`https://cards.scryfall.io/normal/front/{first letter}/{second letter}/{id}.jpg`
This is not _future proof,_ we make no promises this will remain the case, so it's only a good idea if you're ready for it to break abruptly someday. But if you need to skip an API call and have the print ID, you can do this. That said, if all you know is set/cn, you have to make an API call anyway to find out the ID so you might as well just use the image_uris we give you.

The ?1738356433 at the end of the URIs is a cache-breaking timestamp.

> If we change an image in the same spot, the location will not change but the timestamp might.

ghost:
Another response from Scryfall dev Corey:

> This app needs to download one of our bulk files at boot. The direct image URLs are provided. 

(https://scryfall.com/docs/api/bulk-data)

> Hammering ?format=image endpoints is not the correct way to download tons of images
Also the collection endpoint isn’t appropriate either

PR #10423, opened 2026-04-15
Fix commander CardView desync on remote client

### LINKED ISSUE #10406: Cannot target any commanders with effects
**Describe the bug**
Can cast your commander as normal but attempting to target any commander or declare your own as an attacker does not do anything, none of the commanders when testing had hexproof or shroud

**To Reproduce**
1. Attempt to target any commander with any target effect or select your own as a declared attacker
2. Commander will remain untargeted

**Expected behavior**
Commander should attack or be targeted by the effect 

 - OS: Arch Linux
 - Browser Firefox
 - Version 2.0.12-SNAPSHOT-04.14
 
**Additional context**
Issue started in latest snapshot, unsure if issue was in the previous one. Happening in multiplayer games and haven't tested in singleplayer

MostCromulent:
@tool4ever - closes Card-Forge/forge#10406

**[Full root cause analysis, log analysis, and investigation results are in the attached document.](https://github.com/user-attachments/files/26765008/commander-desync-investigation.md)**

## Summary

Fixes a regression from #9642 where commanders on the remote client display incorrect state with delta-sync disabled:
- No summoning sickness icon when commander enters the battlefield
- Attacking commander appears untapped
- Cannot select commander as blocker (clicks ignored)

## Root Cause

`TrackableCollectionType.copyChangedProps` unconditionally copies properties from incoming CardViews onto the tracker's canonical instance. The `Commander` property (ordinal 210) is a cross-zone reference list that holds stale CardView copies (e.g. `zone=Stack`) sharing IDs with zone collections. Because it's processed before `Battlefield` (ordinal 216), it overwrites the canonical instance with stale state. The Battlefield pass becomes a self-copy no-op, so the corruption persists.

## Fix

Restore the pre-#9642 skip for `CardViewCollectionType` and `StackItemViewListType`: swap to the tracker instance for identity consistency (required by delta sync) but skip per-item property copying. This is safe for all card types — regular cards exist in exactly one zone, so the skip is a no-op (the tracker instance already has current properties from `refreshTrackerCardViews`). Commander damage and cast counts are `IntegerMapType` properties on PlayerView, not CardView collections, so they are unaffected.

## Testing

- **Manual (non-delta):** 2-player Commander, host commander displays correctly on client. Zero errors.
- **Manual (delta sync on):** 2-player Commander, same correct behavior. 6 checksums verified, zero mismatches.
- **Automated batch (delta sync on):** 10/10 games passed (3 Commander + 7 Constructed, 2p/3p/4p). 127 checksums verified across 2643 deltas, zero mismatches. Bandwidth savings: 91.0% overall, 94.6% state-only.

## Architectural note (true root cause)

`Commander` and `Flashback` are typed as `CardViewCollectionType` — the same type as actual zones — but have different semantics. They are cross-zone reference lists holding duplicate CardView copies of cards that also exist in their real zone collections. This type mismatch is the true root cause — `copyChangedProps` cannot distinguish zone collections (authoritative) from cross-zone references (stale copies). A cleaner model would store card IDs rather than full CardView copies, eliminating this class of bugs entirely. This fix is correct and minimal; the architectural debt remains.

---

### DIFF
--- a/forge-game/src/main/java/forge/trackable/TrackableTypes.java
+++ b/forge-game/src/main/java/forge/trackable/TrackableTypes.java
@@ -104,7 +104,11 @@ public class TrackableTypes {
                     if (newObj != null) {
                         T existingObj = from.getTracker().getObj(itemType, newObj.getId());
                         if (existingObj != null) {
-                            existingObj.copyChangedProps(newObj);
+                            // Skip CardView collections — cross-zone refs like Commander hold stale copies
+                            if (prop.getType() != TrackableTypes.CardViewCollectionType &&
+                                    prop.getType() != TrackableTypes.StackItemViewListType) {
+                                existingObj.copyChangedProps(newObj);
+                            }
                             newCollection.add(existingObj);
                         } else {
                             from.getTracker().putObj(itemType, newObj.getId(), newObj);


forge-game/src/main/java/forge/trackable/TrackableTypes.java
```diff
@@ -104,7 +104,11 @@ protected void copyChangedProps(TrackableObject from, TrackableObject to, Tracka
                     if (newObj != null) {
                         T existingObj = from.getTracker().getObj(itemType, newObj.getId());
                         if (existingObj != null) {
-                            existingObj.copyChangedProps(newObj);
+                            // Skip CardView collections — cross-zone refs like Commander hold stale copies
+                            if (prop.getType() != TrackableTypes.CardViewCollectionType &&
```
tool4ever, line 108: I might try and restore more of the old behaviour to also take care of Bug 1

tool4ever (approved):

PR #10432, opened 2026-04-17
Prepared spell

Hanmac:
Part of #10233 

Start of the stuff needed for SplitType, CardState and Attribute

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/AlterAttributeEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/AlterAttributeEffect.java
@@ -2,6 +2,7 @@ package forge.game.ability.effects;
 
 import java.util.Map;
 
+import forge.card.CardStateName;
 import forge.game.GameLogEntryType;
 import forge.game.GameType;
 import forge.game.ability.AbilityKey;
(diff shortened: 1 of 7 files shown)


forge-game/src/main/java/forge/game/ability/effects/AlterAttributeEffect.java
```diff
@@ -55,6 +56,12 @@ public void resolve(SpellAbility sa) {
 
                         c.getGame().fireEvent(new GameEventCardPlotted(c, sa.getActivatingPlayer()));
                         break;
+                    case "Prepare":
+                        if (activate && !gameCard.hasState(CardStateName.PreparedSpell)) {
+                            continue;
+                        }
+                        gameCard.setPrepared(activate);
```
tehdiplomat, line 63: Is this where we're planning on spawning the copy of the prepared spell if it has become prepared?

tehdiplomat (approved):

forge-game/src/main/java/forge/game/ability/effects/AlterAttributeEffect.java
Hanmac: i don't have a plan for this yet, this will be later MR

Jetz72: Inside of `setPrepared` might be a better fit?

PR #10433, opened 2026-04-17
More preparing

tool4ever:


Hanmac:
i might exp with other GamePieceType like CopiedSpell for this types too

Jetz72:
Based on the wording they've used so far, I'd guess they'll be updating 704.5e instead of 704.5d (i.e. the one for copied spells/cards vs the one for tokens). I don't think we currently check that SBA?

tool4ever:
ok, seems to work in general now :)

Hanmac (approved):

Hanmac:
> Based on the wording they've used so far, I'd guess they'll be updating 704.5e instead of 704.5d (i.e. the one for copied spells/cards vs the one for tokens). I don't think we currently check that SBA?

we can update the code later for 704.5e for better handling of CopiedSpell

PR #10435, opened 2026-04-17
Fix mobile "Always No" stack-menu item ending the turn

MostCromulent:
**Bug (reported on Discord).** On mobile, tapping "Always No" on a Mana Vault upkeep trigger "works the same way as if the 'End the Turn' button was pressed" — but only when no triggers other than Mana Vault are on the stack.

**Investigation.** Reported after installer 04.16; initially suspected #10355 which landed in that build. Mobile diagnostic instrumentation ruled that out — the PR only renamed `gui.*` → `controller.*` for the relevant setters. The offending call path is unchanged on mobile since Feb 2015.

**Root cause.** Tapping "Always No" does two things: (1) toggle the preference flag, and (2) immediately apply the new preference to the trigger currently on the stack, so the user isn't forced to separately tap through whatever prompt is pending. Step (1) is a simple state change. Step (2) is where the bug lives: the handler calls `selectButtonCancel()`, which simulates pressing the Cancel button on whatever input state is currently active. That was the wrong tool — Cancel means different things in different states. On a trigger-confirm prompt Cancel means "decline the trigger"; on a priority-hold prompt Cancel means "end the turn." While a Mana Vault trigger sits on the stack during upkeep, the active input is `InputPassPriority`, so the simulated Cancel press lands on `InputPassPriority.onCancel` and triggers `autoPassUntilEndOfTurn()` — the End-Turn behavior the user observed.

**Fix.** Swap the Cancel-button simulation for `controller.passPriority()` in the Always-Yes/No handlers — the primitive that explicitly names "pass priority once so the stack resolves with the new preference in effect." It's self-guarded (`instanceof InputPassPriority` at `PlayerControllerHuman.java:2412`): it only dispatches when the player is actually holding priority, and when it does dispatch it cannot end the turn. In `MatchScreen` Ctrl+N / Ctrl+Y, drop the redundant `selectButton*` lines — the trailing `controller.passPriority()` already covers the intended advance.

**Why it's safe.**
- `passPriority()` cannot route to `InputPassPriority.onCancel`. The End-Turn path is structurally unreachable.
- Preference-flag behavior (`confirmTrigger` reads at resolution) unchanged.
- Three-axis UI preserved: Always Yes/No governs the resolution answer, Auto-Yield governs priority auto-pass. Desktop untouched.

---


forge-gui-mobile/src/forge/screens/match/views/VStack.java
```diff
@@ -309,8 +309,8 @@ protected void buildMenu() {
                                             else {
                                                 controller.setShouldAlwaysAcceptTrigger(triggerID);
                                                 if (stackInstance.equals(gameView.peekStack())) {
-                                                    //auto-yes if ability is on top of stack
-                                                    controller.selectButtonOk();
+                                                    //pass priority so trigger resolves with auto-yes flag
+                                                    controller.passPriority();
```
tool4ever: this was originally written with the idea in mind that you'd open the menu while being asked already...

which could be the more likely scenario since it may not immediately be clear if a new trigger contains an optional part that's supported by the feature?

and indeed there used to be a check for InputConfirm, but it seems to have been lost

MostCromulent: I think this only affects what happens on your input when you _first_ set the yield preference; subsequent triggers are controlled by whether Auto-Yield is enabled to effectively be hands free.

So issue is: what do we want the behaviour to be in mana vault example?

Option A - restore old InputConfirm guard - preference is set but the guarded selectButtonCancel() no-ops and user has to manually pass priority. Adds a click.
Option B - current PR solution - preference is set and priority passes
Option C - don't have it do _anything_ when you set the yield other than toggle the preference. If you want it to pass priority automatically then you need to enable Auto-Yield as well. (This is at least consistent with behaviour on every subsequent trigger). 

It's a niche enough situation I have no issue just putting the guard back in, but raise it for consideration.

tool4ever: I don't see why we'd want to pass priority in this case which wouldn't save a click when used after that - the way it should mirror yield behaviour is more about the context:
that is, if it's being changed while you're in the Input that would be skipped it's clearly reasonable

MostCromulent: See latest commit, I think this is what you're after?

> The original FControl.getInputQueue().getInput() instanceof InputConfirm guard (from eb3f6684109, June 2014) was lost during the September/October 2014 GUI refactor (b608d7e7a36 → 3538823e11f), which moved mobile VStack off direct FControl access and onto IGameController / GameView. That refactor left VStack with no way to inspect the active input — the input queue lives behind IGameController, and for network clients it lives on the server entirely — so the 2014 check cannot simply be reinstated in place.
> 
> This commit restores the same guard as a server-side check, exposed through two new IGameController methods: selectButtonOkIfInputConfirm() and selectButtonCancelIfInputConfirm(). PlayerControllerHuman runs the instanceof InputConfirm check before invoking selectButtonOk/Cancel; NetGameController forwards via ProtocolMethod so the same check runs on the server for remote clients. VStack calls the new methods in place of the passPriority() calls from the first iteration of this PR.

tool4ever: looks way too roundabout?
why even implement new method when it's identical to existing one:
image

but honestly what stops us from just adding the logic into `setShouldAlwaysAccept/DeclineTrigger` instead of inventing even more ProtocolMethods? 🤔

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -3573,12 +3573,18 @@ public boolean shouldAlwaysDeclineTrigger(final int trigger) {
     public void setShouldAlwaysAcceptTrigger(final int trigger) {
         if (isRemoteClient()) remoteTriggerDecisions.put(trigger, Boolean.TRUE);
         else localStore().setTriggerDecision(trigger, AutoYieldStore.TriggerDecision.ACCEPT);
+        if (inputQueue.getInput() instanceof InputConfirm) {
```
tool4ever: should still include the previous stack check ☝️

tool4ever (approved):

PR #10438, opened 2026-04-18
Abide by art preferences when generating random pool

clairchiara:
When generating random decks, get cards according to user's art preferences

### DIFF
--- a/forge-core/src/main/java/forge/deck/generation/DeckGeneratorBase.java
+++ b/forge-core/src/main/java/forge/deck/generation/DeckGeneratorBase.java
@@ -63,7 +63,13 @@ public abstract class DeckGeneratorBase {
     protected abstract float getSpellPercentage();
 
     public DeckGeneratorBase(IDeckGenPool pool0, DeckFormat format0, Predicate<PaperCard> formatFilter0) {
-        pool = new DeckGenPool(format0.getCardPool(pool0).getAllCards(formatFilter0));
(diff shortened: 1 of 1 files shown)


forge-core/src/main/java/forge/deck/generation/DeckGeneratorBase.java
```diff
@@ -63,13 +63,25 @@ public abstract class DeckGeneratorBase {
...
 
     public DeckGeneratorBase(IDeckGenPool pool0, DeckFormat format0, Predicate<PaperCard> formatFilter0) {
-        pool = new DeckGenPool(format0.getCardPool(pool0).getAllCards(formatFilter0));
+        final Iterable<PaperCard> allCardsFiltered = format0.getCardPool(pool0).getAllCards(formatFilter0);
+        final List<PaperCard> allCardsPreferredEditions = new ArrayList<>();
+        for (PaperCard c : allCardsFiltered) {
+            PaperCard paperCard = pool0.getCard(c.getName());
+            allCardsPreferredEditions.add(paperCard);
+        }
+        pool = new DeckGenPool(allCardsPreferredEditions);
```
Jetz72, line 72: The way this is done means that even if the user doesn't have a preferred print for a card, the pool will still be overwritten with copies of the first print the `CardDb` has for that card's name. 

It also seems weird to request a list of cards from the card pool, and then turn around and ask that same card pool for each card from that list individually. Suggests that this is probably an issue that should be tackled elsewhere?

PR #10439, opened 2026-04-18
Support |SET pinning in booster name() predicate

vanja-ivancevic:
- Booster templates using `name("Card|SET")` in a non-main-code position silently produced zero cards for that slot. The slot parser routed `name(...)` through `buildExtraPredicate`, which used `PaperCardPredicates.names(...)` to compare operand strings against `card.getName()`. Names carrying a `|SET` suffix never matched, so the slot dropped.
- First surfaced on Shandalar Old Border's ICE booster, whose snow-covered basic land slot is pinned to the ICE printing: `1 name("Snow-Covered Plains|ICE", ...)`. Packs opened 14 cards instead of 15, with no snow basic.
- Fix resolves pinned names via `StaticData.getCommonCards().getCard(name)` (same call the mainCode branch uses) and filters with a set-contains predicate. Plain names without `|` keep the existing `PaperCardPredicates.names` path, so current callers (`Final Fantasy.txt`, `Oath of the Gatewatch.txt`) are unaffected.
- Benefits every booster-generation path since they all flow through `BoosterGenerator.getBoosterPack`: shop packs, draft packs, sealed packs, event rewards.

### DIFF
--- a/forge-core/src/main/java/forge/item/generation/BoosterGenerator.java
+++ b/forge-core/src/main/java/forge/item/generation/BoosterGenerator.java
@@ -724,7 +724,24 @@ public class BoosterGenerator {
             } else if (operator.startsWith("name(")) {
                 operator = StringUtils.strip(operator.substring(4), "() ");
                 String[] cardNames = TextUtil.splitWithParenthesis(operator, ',', '"', '"');
-                toAdd = PaperCardPredicates.names(Lists.newArrayList(cardNames));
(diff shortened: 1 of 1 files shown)


Jetz72:
Isn't there a `fromSets(` filter you can use a few lines down?

vanja-ivancevic:
Good catch, `fromSets()` already does the job. Closing in favor of #NEW which fixes the overlay data instead.

PR #10445, opened 2026-04-18
feat(net): show all network interfaces when hosting a match

RafaelHGOliveira:
## Summary

When hosting a multiplayer match, the current dialog only shows **one** local IP (whatever `getLocalAddress()` picks as routable) plus the external WAN IP. This is a problem when the host wants to play with friends over a VPN or overlay network — **Tailscale, Hamachi, ZeroTier**, etc. — because the IP the friends actually need is almost never the \"routable\" one chosen automatically.

This PR replaces that single-line prompt with a **per-interface list**: every active non-loopback IPv4 address is shown with a friendly interface name and its own Copy button. The primary routable interface is marked with ★.

Host dialog showing all network interfaces

## Why

- **Tailscale / Hamachi / ZeroTier / VPN users**: the overlay network's IP lives on a separate virtual NIC. With the old dialog, the host had no way to see — much less copy — that address from inside Forge, and had to fall back to `ipconfig` / `ifconfig` / the VPN app's own UI just to share the right address with friends.
- **Multi-homed machines** (Ethernet + Wi-Fi, multiple ZeroTier networks, corporate VPN, etc.): picking the right IP was guesswork.
- The backend (`FServerManager.getAllLocalAddresses()`) already enumerates every interface with friendly names recognizing Hamachi (`ham*`), ZeroTier (`zt*`), Tailscale (CGNAT `100.64.x.x` range), and common VPN patterns — this PR just wires that data into the desktop host flow.

## Changes

- `forge-gui-desktop/.../CSubmenuOnlineLobby.java` — adds `showServerAddressesDialog()` called after `host()` succeeds; replaces the single-URL `NetConnectUtil.copyHostedServerUrl()` call in the desktop path. `copyHostedServerUrl()` is kept intact (still used by mobile and the in-lobby Copy URL button).
- `forge-gui/res/languages/en-US.properties` — three new i18n keys: `lblInterface`, `lblAddress`, `lblChooseAddressToCopy`.


tool4ever:
nice idea, though it looks like your AI got confused with what's available outside of your fork ;)

RafaelHGOliveira:
Thanks for the heads-up! You're right — `getAllLocalAddresses()` was only in my fork. I've amended the commit to include the method directly in `FServerManager.java`, so the PR is now self-contained and should build cleanly against upstream.

tool4ever (approved):

PR #10448, opened 2026-04-18
Restore mobile auto-yield persistence across matches

### LINKED ISSUE #10442: Add option to make auto-yields permanent
**Is your feature request related to a problem? Please describe.*
The auto-yield system only works in a single match or game, having to re-enable the same auto-yield toggles every time is tiresome. Auto-yields used to remain between matches but I believe some recent snapshot update changed or removed that behavior.

Example of described issue:
-Auto-yield selection at the end of a match:
Image

-Auto-yield selection empty at the start of a later match:
Image

**Describe the solution you'd like*
A simple option to make auto-yield selection permanent, remaining static between matches or even saved between game sessions.

MostCromulent:
Screenshot 2026-04-19 074435

## Summary

Closes #10442.

**The bug** — Auto-yield state used to persist across matches for mobile users, but it appears this was an architectural accident. Before #10355, auto-yields were instance fields on `AbstractGuiGame`. On mobile, `MatchController` is a singleton (`public static final MatchController instance = new MatchController()`), so its `autoYields` set survived between matches. On desktop, `CMatchUI` is created fresh per match, so the same code never persisted yields there. No user-facing preference, and behaviour was asymmetric between platforms.

#10355 moved auto-yield state from `AbstractGuiGame` to `PlayerControllerHuman` (per-player ownership — to prevent hotseat cross-contamination and to simplify network sync). This was architecturally correct, but eliminated mobile's accidental singleton-backed persistence. Mobile users noticed.

## The fix

`PlayerControllerHuman` and `NetGameController` now hold yield state in two parallel stores:

- **Static shared store** — used when `isLibgdxPort()` AND a new preference `UI_PERSIST_AUTO_YIELDS_ACROSS_MATCHES` is enabled (default `true`). Lives for the Forge session.
- **Per-instance store** — used otherwise; dies with the controller at match end (original post-#10355 behaviour).

A selector (`useSharedYieldStore()`) picks between them on each access; all getters/setters route through it. The new toggle is exposed on the mobile Settings screen next to `UI_AUTO_YIELD_MODE`.

### Why it's safe

- **Desktop is fully unchanged.** The `isLibgdxPort()` gate hard-blocks the shared-store path regardless of the preference value. Desktop users who've never seen persistence won't start seeing it, even if the pref is flipped on via `forge.preferences`.
- **Network isolation is preserved.** Remote `PlayerControllerHuman` instances (identified via `gui instanceof RemoteClientGuiGame`) always use per-instance fields. The #10355 invariant that "Player A's yields don't affect Player B's server-side view" still holds.
- **`clearAutoYields()` only clears per-instance state** so `HostedMatch.endCurrentGame()`'s end-of-match cleanup no longer wipes the session-persistent shared store. Explicit user actions (removing a yield in the dialog, toggling "Disable All") still work because they go through `setShould*` which route through the selector.

## Scope / open design questions

This is deliberately a minimal fix for the regression — it restores previous mobile behaviour, nothing more. PR could be amended to go further and address:

1. **Desktop persistence.** Should desktop users also be able to opt into cross-match persistence? Today they cannot — the platform gate is a hard block. If we wanted to allow it, we'd need to decide between a single cross-platform pref and platform-scoped prefs.
2. **Cross-session persistence.** `autoYields` keys are stable ability text and could be serialised to `forge.preferences` so yields survive Forge restarts, not just session lifetime. `triggersAlwaysAccept` cannot be persisted this way (trigger IDs are runtime-assigned).

---


tool4ever (changes requested):
I think I actually misunderstood the issue and didn't even consider the different behaviour on mobile.

But AI missed the bigger problem - it doesn't even transfer between games of the same match now on Desktop either.
And that's something we certainly need depending how this feature is configured:
Image

Should probably just revise it to support these variants uniformly on both platforms 🤔

- this additional "Each Session" is certainly interesting for Desktop too
- converting that into full persistence ("Each Install") **instead** could be even more helpful
- should probably still keep the Match-boundary around for cases where user is involved in multiple ones simultaneously
- might even bring back the old per-UI style as option?

imo AI can have some freedom on relocating how we store this data if it helps with a cleaner architecture

sorry for the extra work

MostCromulent:
Alright, Take 2 on this. See what you think of latest commit.

New "Each Session" and "Each Install" yield persistence options added across-platform and integrated into existing drop-down menus on both desktop and mobile, replacing implementation on first commit.

**[Architectural reference document here.](https://github.com/user-attachments/files/26868134/auto-yield-persist-architecture.md)**

Screenshot 2026-04-19 203804

tool4ever:
I still need to test a bit but this looks pretty neat now - I'd imagine some combo players will be happy to manage these files for their decks 👍

And while I originally thought "Each Session" might not be needed if you can just use "Each Install" and then clear when you want to start fresh it seems the way AI implemented it makes both useful - since you can switch between them if you're doing some testing session and don't want to modify your more regular defaults :)

tool4ever (approved):

PR #10452, opened 2026-04-19
Guard against redundant GameView sends in non-delta protocol

MostCromulent:
@tool4ever potential mitigation for non-delta bandwidth increase since the merge.

---

## Summary
Many `RemoteClientGuiGame` protocol methods prepend a `setGameView` to their send path to ensure the client sees current state. That preamble is redundant when an event batch is about to flush on the same call — the flush already emits a full `GameView` to the client. Add a guard that skips the preamble in that case. Benefits all 20 call sites of `updateGameView()`. Delta path untouched.

## The issue

`RemoteClientGuiGame` calls `updateGameView()` before many protocol methods so the client sees current state before the next operation. `updateGameView()` itself calls `flushPendingEvents()` internally — and when events are pending, the flush fires `handleGameEvents`, which emits a full `setGameView` + `applyDelta(events)` batch to the client.

At that point the client has everything it needs: a fresh full `GameView` and the events explaining how the state got there. The `updateGameView()` preamble's *own* follow-up `setGameView` is a second, byte-identical snapshot of the same `GameView` reference on the same thread. Pure duplicate.

11 callers pre-date the delta-sync PR (`setGameView`, `openView`, `showPromptMessage`, `setPanelSelection`, `setCard`, `openZones`, etc.). The delta PR added 9 more for input prompts (`confirm`, `getChoices`, `showOptionDialog`, etc.). The additional 9 callers added by the delta-sync PR increase the blast radius of this issue, but all 20 produce this duplicate when events are pending.

## How the delta path already handles this

```java
flushPendingEvents();
DeltaPacket delta = syncManager.collectDeltas(gameView);
if (!delta.isEmpty()) {                 // ← guard
    sender.send(ProtocolMethod.applyDelta, delta);
}
```

The `!delta.isEmpty()` check does double duty: if events were pending, `flushPendingEvents()` just bundled events + state into an `applyDelta` and drained the dirty bits, so `collectDeltas()` returns empty and the guard skips. If nothing changed, same outcome. Either way, no duplicate.

## Why the non-delta path fails to guard

```java
if (!useDeltaSync || !initialSyncSent) {
    if (flush) send(ProtocolMethod.setGameView, gameView, -1L);   // no guard
    else       sender.write(ProtocolMethod.setGameView, gameView, -1L);
    return;
}
```

`send()` calls `flushPendingEvents()` internally, which drains the batch and emits `setGameView` (wire # 1) + `applyDelta` (wire # 2). Control returns to `send()`, which emits a second `setGameView` (wire # 3 ) carrying the same state. The following `sendAndWait(prompt, …)` emits the prompt (wire # 4). Wire # 3 is the redundancy.

## Fix

```java
if (!useDeltaSync || !initialSyncSent) {
    // Batch flush already emits setGameView; skip the duplicate.
    if (forwarder != null && forwarder.hasPendingEvents()) {
        flushPendingEvents();
        return;
    }
    if (flush) send(ProtocolMethod.setGameView, gameView, -1L);
    else       sender.write(ProtocolMethod.setGameView, gameView, -1L);
    return;
}
```

Plus one method on `GameEventForwarder`:

```java
public boolean hasPendingEvents() { return !pendingEvents.isEmpty(); }
```

The non-delta path can only mirror the "batch will handle it" half of the delta guard.

Because the fix lives inside `updateGameView()`, all 20 callers benefit, including the 11 that pre-date the delta PR.

## Why this is safe

- **Delta path unchanged.** The guard sits entirely inside the `if (!useDeltaSync || !initialSyncSent)` branch; the resync path and delta branch below are identical to master.
- **No recursion.** `GameEventForwarder.flush()` clears `pendingEvents` *before* calling `handleGameEvents`, so when the batch re-enters `updateGameView(false)`, `hasPendingEvents()` returns false and the guard is skipped. A secondary `flushing` flag in `flushPendingEvents()` provides belt-and-braces.
- **No content difference.** The `setGameView` emitted by the batch (`sender.write`) and the one skipped by the guard (`sender.send`) pass the same `gameView` reference and seq on the same thread with no intervening mutation.
- **Initial sync preserved.** When `!initialSyncSent`, the guard still routes to a full `setGameView` — via batch when events pending, directly when not.
- **Forward-compatible.** When delta sync is re-enabled, the guard becomes dead code that the delta branch's existing check subsumes. No cleanup required.

---


forge-gui/src/main/java/forge/gamemodes/net/server/RemoteClientGuiGame.java
tool4ever: maybe also branch with `hasPendingEvents()` here - seems waste of CPU to check whole gameView for dirtyProps again if we'd know they would be empty anyway?

(the comment implies delta could change even if events got flushed but that's no longer true?)

tool4ever (approved):

PR #10464, opened 2026-04-19
fix(Dracogenesis): restrict MayPlay to Dragons the controller owns

RafaelHGOliveira:
## Summary
Restricts Dracogenesis's `MayPlay` continuous effect to Dragon cards the controller owns (`Affected$ Dragon.YouOwn`).

## Why
Without the `.YouOwn` restriction, Dracogenesis matched Dragon cards in **all players' zones**, allowing the controller to cast opponents' Dragons for free. The card text says "You may cast Dragon spells" — the fix scopes the effect to cards the casting player owns.

## Test plan
- [ ] Cast Dracogenesis; confirm you can cast your own Dragons for free from hand, graveyard, library, exile, and command zone.
- [ ] Confirm Dracogenesis no longer grants MayPlay on opponents' Dragon cards.

### DIFF
--- a/forge-gui/res/cardsfolder/d/dracogenesis.txt
+++ b/forge-gui/res/cardsfolder/d/dracogenesis.txt
@@ -1,7 +1,7 @@
 Name:Dracogenesis
 ManaCost:6 R R
 Types:Enchantment
-S:Mode$ Continuous | Affected$ Dragon | MayPlay$ True | MayPlayWithoutManaCost$ True | MayPlayDontGrantZonePermissions$ True | AffectedZone$ Hand,Graveyard,Library,Exile,Command | Description$ You may cast Dragon spells without paying their mana costs.
+S:Mode$ Continuous | Affected$ Dragon.YouOwn | MayPlay$ True | MayPlayWithoutManaCost$ True | MayPlayDontGrantZonePermissions$ True | AffectedZone$ Hand,Graveyard,Library,Exile,Command | Description$ You may cast Dragon spells without paying their mana costs.
 SVar:NonStackingEffect:True
 AI:RemoveDeck:Random
 Oracle:You may cast Dragon spells without paying their mana costs.


forge-gui/res/cardsfolder/d/dracogenesis.txt
```diff
@@ -1,7 +1,7 @@
 Name:Dracogenesis
 ManaCost:6 R R
 Types:Enchantment
-S:Mode$ Continuous | Affected$ Dragon | MayPlay$ True | MayPlayWithoutManaCost$ True | MayPlayDontGrantZonePermissions$ True | AffectedZone$ Hand,Graveyard,Library,Exile,Command | Description$ You may cast Dragon spells without paying their mana costs.
+S:Mode$ Continuous | Affected$ Dragon.YouOwn | MayPlay$ True | MayPlayWithoutManaCost$ True | MayPlayDontGrantZonePermissions$ True | AffectedZone$ Hand,Graveyard,Library,Exile,Command | Description$ You may cast Dragon spells without paying their mana costs.
```
tool4ever, line 4: that's not what the card says...

RafaelHGOliveira: So this card should allow you to cast dragons even from another player's graveyard? 🤔

Jetz72: On its own it shouldn't enable you to cast dragons from any graveyard. If you combined it with some other effect that allows you to cast cards from an opponent's graveyard, it would allow you to cast a dragon in their graveyard for free. That's what `MayPlayDontGrantZonePermissions` is supposed to be for. Are you sure nothing else was granting the ability to cast from their graveyard when you originally experienced this?

Hanmac: Technically yes, if something else would allow to be cast from that Zone

RafaelHGOliveira: Ok, maybe it was skill issue lol

PR #10467, opened 2026-04-20
Update CostExile.java

Akse1ot:
Fixes CostExile validation for +withSharedCardType so the cost works correctly when selecting three or more cards that share a common card type.


Akse1ot:
I implemented this for my fork; it may also be useful to someone in the main branch.

Hanmac (approved):

PR #10468, opened 2026-04-20
Update HumanCostDecision.java

Akse1ot:
Adds a UI filter to HumanCostDecision that prevents incompatible cards from being selected when paying an exile cost with +withSharedCardType, especially for selections of three or more cards.

### DIFF
--- a/forge-gui/src/main/java/forge/player/HumanCostDecision.java
+++ b/forge-gui/src/main/java/forge/player/HumanCostDecision.java
@@ -374,32 +374,6 @@ public class HumanCostDecision extends CostDecisionMakerBase {
 
     // Inputs
 
-    private boolean sharesCommonCoreType(final Iterable<Card> cards) {
-        boolean foundAny = false;
-
-        for (CardType.CoreType coreType : CardType.CoreType.values()) {
-            boolean allHaveType = true;
-            boolean sawCard = false;
-
-            for (final Card card : cards) {
-                sawCard = true;
-                if (!card.getType().hasType(coreType)) {
-                    allHaveType = false;
-                    break;
-                }
-            }
-
-            if (sawCard) {
-                foundAny = true;
-                if (allHaveType) {
-                    return true;
-                }
-            }
-        }
-
-        return !foundAny;
-    }
-
     private PaymentDecision exileFromSame(final CostExile cost, final CardCollectionView list, final int nNeeded, final List<Player> payableZone) {
         if (nNeeded == 0) {
             return PaymentDecision.number(0);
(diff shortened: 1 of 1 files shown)


Akse1ot:
I implemented this for my fork; it may also be useful to someone in the main branch.

forge-gui/src/main/java/forge/player/HumanCostDecision.java
```diff
@@ -468,18 +503,58 @@ private PaymentDecision exileFromMiscZone(final CostExile cost, final int nNeede
...
+            return null;
+        }
+
+        final InputSelectCardsFromList inp = new InputSelectCardsFromList(controller, mandatory ? nNeeded : 0, nNeeded, typeList, ability) {
+            private static final long serialVersionUID = 1L;
+
+            @Override
+            protected boolean onCardSelected(final Card c, final List<Card> otherCardsToSelect, final ITriggerEvent triggerEvent) {
+                for (final Card selectedCard : this.selected) {
+                    if (!sharesAnyCoreType(selectedCard, c)) {
```
Hanmac: Use `CardType#sharesCardTypeWith`

Or better `Card#sharesCardTypeWith`

Akse1ot: Updated to use Card#sharesCardTypeWith and removed the custom helper.

forge-gui/src/main/java/forge/player/HumanCostDecision.java
```diff
@@ -468,18 +494,58 @@ private PaymentDecision exileFromMiscZone(final CostExile cost, final int nNeede
...
+            return null;
+        }
+
+        final CardCollection chosen = new CardCollection(inp.getSelected());
         if (chosen.size() < nNeeded) {
             return null;
         }
-        if (sharedType) {
-            if (!chosen.get(1).sharesCardTypeWith(chosen.get(0))) return null;
+        if (!sharesCommonCoreType(chosen)) {
```
Hanmac: @tool4ever can this really happen?
The `InputSelectCardsFromList` should prevent wrong targets anyway?

@Akse1ot so `sharesCommonCoreType` check can be removed?

Akse1ot: No, I don’t think it can be removed. The InputSelectCardsFromList filter only enforces pairwise sharesCardTypeWith checks against the already selected cards. For three or more cards that still allows combinations where every pair shares some card type, but the full set has no single common card type. sharesCommonCoreType is still needed as the final full-set validation.

Hanmac: `(!selectedCard.sharesCardTypeWith(c))`

this line already check that each selected card shares a card type with each other card

To a scenario where A, B, C only partially share some types can't happen

Akse1ot: If I understood the current logic correctly, the selection check is still pairwise via `sharesCardTypeWith`, so a case like this seems possible:

A = Artifact Creature  
B = Artifact Land  
C = Land Creature

Each pair shares a card type, but the full set does not share a single common card type. That was the reason I kept the final full-set validation.

If I misunderstood your intent or how this flow actually behaves here, then I apologize, and we can treat the issue as resolved.

Hanmac: It is still more clean if you take the first item `chosen.get(0)` and then run `sharesCardTypeWith` against it

Akse1ot: Updated as suggested. The selection filter now validates each additional card against the first selected card, and the extra final check was removed.

Hanmac (dismissed):

forge-gui/src/main/java/forge/player/HumanCostDecision.java
```diff
@@ -459,32 +459,68 @@ private PaymentDecision exileFromTop(final CostExile cost, final int nNeeded) {
...
     }
 
-    private PaymentDecision exileFromMiscZone(final CostExile cost, final int nNeeded, final CardCollection typeList, final boolean sharedType) {
-        // when it's always a single triggered card getting exiled don't act like it might be different by offering the zone for choice
-        if (cost.zoneRestriction == -1 && ability.isTrigger() && nNeeded == 1 && typeList.size() == 1) {
-            if (confirmAction(cost, Localizer.getInstance().getMessage("lblExileConfirm", typeList.getFirst().getTranslatedName()))) {
-                return PaymentDecision.card(typeList.getFirst());
-            }
-            return null;
+private PaymentDecision exileFromMiscZone(final CostExile cost, final int nNeeded, final CardCollection typeList, final boolean sharedType) {
```
Agetian: Formatting seems slightly off (no indentation now), otherwise lgtm

Hanmac (approved):

Issue #10429, opened 2026-04-17
Add support for online draft and sealed modes


MostCromulent:
## The Plan

**[Current design document](https://github.com/user-attachments/files/26805528/2026-04-12-network-draft-sealed-design.md)** as modified by **[discussion here](https://github.com/Card-Forge/forge/issues/9861#issuecomment-4245231008)**

## Roadmap

**Phase 1 - Draft/Sealed online implementation**

- [X] Phase 0.5: network lifecycle hardening - https://github.com/Card-Forge/forge/pull/10410
- [X] Phase 1A: Online draft/sealed mode for desktop - https://github.com/Card-Forge/forge/pull/10474
- [ ] Phase 1B: Extend online/draft sealed mode to mobile - https://github.com/Card-Forge/forge/pull/10806

**Side Quest - Phase 1.5??**
- [ ] Move conspiracy draft to abstract network-safe framework rather than hard-coding

**Phase 2 - Tournament mode**

- [ ] Refine design plan for tournament mode
- [ ] Phase 2A: Implement tournament mode for local play
- [ ] Phase 2B: Extend tournament mode to online

PR #10474, opened 2026-04-20
Add draft and sealed modes to online multiplayer (desktop)

MostCromulent:
> Apologies to the devs for another huge PR - I think this one is worth it though! :pray:

**Phase 1A of #10429 - DESKTOP ONLY** 

Mobile support will come in follow-up PR once we are happy with architecture and UI flow on this.

**[Architectural reference document here. (Updated 23/04/2026)](https://github.com/user-attachments/files/26985807/network-draft-sealed-architecture-v2.md)**

## Basic workflow

Screenshot 2026-04-21 064312

1. **Lobby mode.** The online lobby gains a Constructed / Limited mode selector. In Limited mode the host creates an event via a short wizard (Draft or Sealed → pool type → block/set sub-dialogs → pick timer + disconnect grace window for draft). Event config is displayed in a UI panel and visible to all players before they ready up.
2. **Pod assembly.** Participants are populated from lobby slots; remaining seats are auto-filled with AI for draft; seats are shuffled before start.
3. **Run the event.**
   - **Draft** — all human hosts and clients draft simultaneously using the normal draft UI. AI seats pick in between human rounds. Each pack is routed to the correct seat via DraftPackArrivedEvent. A floating picker overlay shows current draft status. Progress of the draft is logged in the Event Log. If a human hits the pick timeout, they will automatically select the first card on their screen and pass their pack.
   - **Sealed** — host generates pools; each client receives their sealed pool and is dropped into the deck editor to build.

Screenshot 2026-04-21 061041
Screenshot 2026-04-21 062702

4. **Deckbuilding.** On draft/sealed completion every participant's pool is saved as a deck on their local device. Each deck is tagged with metadata including event id, format, product (block + set combo), and timestamp. Players are taken immediately to offline deck builder. Players can also access decks later under a shared Draft/Sealed Decks section in multiplayer menu. 
5. **Match setup** 
   - If playing immediately after draft, lobby will automatically filter deck selector to match the metadata for the event. The host can disable this with "Only allow decks from this event" checkbox.
   -  Alternatively if playing games in separate session from draft, host can load a previous event from the "Setup Event" button, which will let them choose any event matching metadata in decks saved on their device.
   -  Standard match procedure follows.

---


forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
```diff
@@ -97,6 +100,30 @@ RemoteClient getClient(final Channel ch) {
...
+     * Send an event to the given slot. If the slot is a remote client, sends
+     * the NetEvent over the wire; otherwise invokes {@code localAction} on the
+     * local lobby listener (the host's own path).
+     */
+    public void sendToSlot(int slotIndex, NetEvent remoteEvent, Consumer<ILobbyListener> localAction) {
+        RemoteClient client = getClientBySlotIndex(slotIndex);
+        if (client != null) {
+            client.send(remoteEvent);
+        } else if (lobbyListener != null) {
+            localAction.accept(lobbyListener);
```
tool4ever: isn't this Consumer param redundant to `dispatchToLocalListener`?

forge-gui/res/languages/en-US.properties
```diff
@@ -2980,6 +2984,62 @@ lblOnline=Online
 lblShowChatPanel=Show Chat Panel
 lblOpenNetworkLogs=Open Network Logs
 lblDisconnect=Disconnect
+lblShowDraftPanel=Show Draft Panel
+#VLobby.java (network draft/sealed)
+lblNetworkLobbyMode=Mode:
+lblNetworkModeConstructed=Constructed
```
tool4ever: most of these type strings already exist
should scan file and avoid easy duplicates

tool4ever:
Not sure there was a good reason to add another 100 lines to a monster PR for essentially a redundant display of the already digital information? 😛

Anyway I did some smaller tests, mostly to confirm this shouldn't break any existing stuff and looks like I won't need much more time with the code either, so a merge might be possible soon 🥳

MostCromulent:
In terms of UI flow for starting draft through to playing match do you think this is intuitive? 

I considered whether the draft setup info should be through some sort of floating dialog displayed to all players but seemed best to embed in lobby page as much as possible.

tool4ever (dismissed):
seems solid enough as foundation:
if anything I'm a bit more skeptical on a few architecture details and ideally some more technical debt can be taken care of before going for mobile while collecting user feedback - unless that support will suddenly be in really high demand 😆

MostCromulent:
Just found an NPE - hold off merging until I can push a fix.

MostCromulent:
> I'm a bit more skeptical on a few architecture details and ideally some more technical debt can be taken care of before going for mobile

Anything in particular you want me to focus on for follow-up?

tool4ever (approved):

tool4ever:
hmn not enough time for a deep dive for now, but it felt weird having to track nesting like this:
image

PR #10487, opened 2026-04-22
Fix AI target selection for Saskia curse ability

Madwand99:
Fixes an issue where Saskia (and other cards like True-Name Nemesis) could not select AI players as valid targets in multiplayer games.

Root cause:
AI player selection logic excluded certain player types incorrectly.

Fix:
Updated ChoosePlayerAi logic to include all valid opponents.

Validation:
- Added unit test: testCurseChoiceUsesPreferredOpponentFromAllPlayers
- Verified in-game behavior manually

### DIFF
--- a/forge-gui-desktop/src/test/java/forge/ai/ability/ChoosePlayerAiTest.java
+++ b/forge-gui-desktop/src/test/java/forge/ai/ability/ChoosePlayerAiTest.java
@@ -2,58 +2,29 @@ package forge.ai.ability;
 
 import com.google.common.collect.Lists;
 import forge.ai.AITest;
-import forge.ai.LobbyPlayerAi;
 import forge.ai.SpellAbilityAi;
 import forge.ai.SpellApiToAi;
-import forge.deck.Deck;
 import forge.game.Game;
-import forge.game.GameRules;
-import forge.game.GameStage;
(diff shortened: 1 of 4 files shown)


forge-gui-desktop/src/test/java/forge/ai/ability/ChoosePlayerAiTest.java
```diff
@@ -0,0 +1,69 @@
...
+import forge.game.player.Player;
+import forge.game.player.RegisteredPlayer;
+import forge.game.spellability.SpellAbility;
+import org.testng.AssertJUnit;
+import org.testng.annotations.Test;
+
+import java.util.List;
+
+public class ChoosePlayerAiTest extends AITest {
+    private Game initAndCreateThreePlayerGame() {
```
tool4ever: no reason to duplicate this method instead of moving it to parent class

Madwand99: Fixed by moving initAndCreateThreePlayerGame into AITest.  DamageDealAiTest.java‎ had some code that duplicated initAndCreateThreePlayerGame so I removed the redundancy.

tool4ever (approved):
thanks

PR #10493, opened 2026-04-22
Funny card loading fix

autumnmyst:
When funny cards are disabled, Forge is filtering card names rather than printings via a single funnyCards ban set. Any legal card sharing a name with a funny-set entry is banned too. Concretely: Fast // Furious (Modern Horizons 2, legal) is unavailable because Unknown Event contains a joke card with the same name. Forge's script implements the legal MH2 version, but the name match drops it before the script ever loads.

Fix: Since we don't even support the script for the funny version, fix is just replacing it with a comment to prevent interference with legal card loading.

### DIFF
--- a/forge-gui/res/editions/Unknown Event.txt
+++ b/forge-gui/res/editions/Unknown Event.txt
@@ -446,7 +446,7 @@ RB14 R Another Night in Vegas @
 RR14 R Brutal Command @
 UU14a U Tax Draw @
 UU14b U Tax Sweeper @
-CR15a C Fast // Furious @
+# CR15a C Fast // Furious @  -- disabled: collides with MH2 "Fast // Furious"; only rules script is the MH2 legal version, so loading this printing would attach wrong rules to joke art.
 RB15 R Monster Mash-Up @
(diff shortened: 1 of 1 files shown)


autumnmyst:
There are far more elaborate fixes for this that would allow for the unknown card to load once we have the script for it, but honestly it's already very messy to have 2 cards with the same name loaded at once. In many cases, a funny set with a same-name card should be legal as it's supposed to be the same card, but this is the single case where a funny card sharing a name should **not** be legal, so I think this is the best approach for now.

Oddly enough I believe we did have the proper rules text for this card before but later removed it? I remember making a PR for this issue before but now it's cropped up again.

Jetz72:
We discussed it over in #8520, we have support for same-name-different-script and I even had that working at one point. The complication was same-name-different-legality. Lot of the format legality logic work by card name, because as long as a card was printed in an allowed set, any other printing with the same name is normally allowed. We'd need to switch that to handle legality by `CardRules`.

Jetz72 (approved):

PR #10504, opened 2026-04-23
Fix for Maze of Ith in multiplayer

Madwand99:
In multiplayer games, the AI uses Maze of Ith and similar cards indiscriminately, even protecting other players at the cost of itself. Maze’s pre-existing script is broad: ValidTgts$ Creature.attacking, so all attacking creatures are targetable. The AI logic in UntapAi.java looks at all attackers during an opponent’s turn, instead of only attackers whose defending player is the AI. This is not unique to Maze: The same pattern could affect Maze of Shadows, Spires of Orazca, and targeted AILogic$ Fog effects like Shimian Night Stalker / Turn the Tables style cards. It is not a rules/card-script bug; it is AI target-selection being too multiplayer-naive.

Fix: I filtered defensive combat targets to only creatures attacking the AI or something defended by the AI, using combat.getDefenderPlayerByAttacker(card) == ai.

Changed:

UntapAi.java (line 162): filters RemoveBestAttacker untap targets.
UntapAi.java (line 374): filters Maze/Maze of Shadows targets.
EffectAi.java (line 140): filters targeted fog-style effects.
RemoveFromCombatAi.java (line 24): drawback check now looks for attackers actually aimed at the AI.

I tested in multiplayer and the fix works.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/RemoveFromCombatAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/RemoveFromCombatAi.java
@@ -4,6 +4,8 @@ package forge.ai.ability;
 import forge.ai.AiAbilityDecision;
 import forge.ai.AiPlayDecision;
 import forge.ai.SpellAbilityAi;
+import forge.game.card.Card;
+import forge.game.combat.Combat;
 import forge.game.player.Player;
 import forge.game.spellability.SpellAbility;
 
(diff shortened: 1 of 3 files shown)


forge-ai/src/main/java/forge/ai/ability/RemoveFromCombatAi.java
```diff
@@ -20,7 +22,16 @@ public AiAbilityDecision chkDrawback(Player aiPlayer, SpellAbility sa) {
         // AI should only activate this during Human's turn
 
         if ("RemoveBestAttacker".equals(sa.getParam("AILogic"))) {
-            boolean result = aiPlayer.getGame().getCombat() != null && aiPlayer.getGame().getCombat().getDefenders().contains(aiPlayer);
```
tool4ever, line 23: the parent has already handling setting up a viable target
so you can just directly make your decision

tool4ever (approved):

PR #10505, opened 2026-04-23
Closing the UI tab in multiplayer after losing fixed

Madwand99:
Fixed an issue in multiplayer matches where, after the human player had already lost, closing the match tab would only remove the UI while the remaining AI players kept playing in the background. This happened because the close/concede flow saw that no local human still needed to concede and treated the match as safe to close, even though the game itself was still running.

The fix updates AbstractGuiGame.concede() so that, in local non-network games, if all remaining active players are AI and no local human is still alive, Forge now explicitly ends the game with GameEndReason.AllHumansLost on the game thread. That allows the normal game-end cleanup path to run and prevents off-screen AI-vs-AI play from continuing after the tab is closed.

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/match/AbstractGuiGame.java
+++ b/forge-gui/src/main/java/forge/gamemodes/match/AbstractGuiGame.java
@@ -2,6 +2,7 @@ package forge.gamemodes.match;
 
 import com.google.common.collect.*;
 import forge.game.GameEntityView;
+import forge.game.GameEndReason;
 import forge.game.GameLog;
 import forge.game.GameView;
 import forge.game.card.CardView;
@@ -375,6 +376,9 @@ public abstract class AbstractGuiGame implements IGuiGame, IMayViewCards {
                     return false;
                 }
             } else {
+                if (!ignoreConcedeChain && !isNetGame() && forceEndGameForRemainingAIs()) {
+                    return false;
+                }
                 return !ignoreConcedeChain;
             }
             if (isNetGame()) {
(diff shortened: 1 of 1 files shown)


forge-gui/src/main/java/forge/gamemodes/match/AbstractGuiGame.java
```diff
@@ -409,6 +413,38 @@ public boolean concede() {
         }
     }
 
+    private boolean forceEndGameForRemainingAIs() {
+        final GameView gameView = getGameView();
+        if (gameView == null || gameView.isGameOver()) {
+            return false;
```
tool4ever, line 419: the parent already checked GameOver, so this is redundant

forge-gui/src/main/java/forge/gamemodes/match/AbstractGuiGame.java
```diff
@@ -409,6 +413,38 @@ public boolean concede() {
...
+        }
+
+        boolean hasRemainingLocalHuman = false;
+        boolean hasRemainingAi = false;
+        for (PlayerView player : gameView.getPlayers()) {
+            if (!player.getHasLost()) {
+                if (isLocalPlayer(player) && !player.isAI()) {
```
tool4ever: seems useless to check for humans, they would have been conceded by parents logic before we ever reach this?

Madwand99: Fixed!

forge-gui/src/main/java/forge/gamemodes/match/AbstractGuiGame.java
Madwand99: Fixed!

tool4ever (approved):

PR #10506, opened 2026-04-23
Persistent Always-Yes / Always-No trigger preferences

MostCromulent:
## Summary

Extends the tiered Auto-Yield architecture from #10448 to repeatedly-firing optional triggered abilities. Pressing Y/N on an optional trigger prompt can now persist that decision across future firings of the same trigger, at a tier chosen by the user.

Triggers are governed by a new `UI_AUTO_TRIGGER_MODE` preference that is **separate from** the existing yield-mode preference, so users can tier triggers and yields independently (e.g. Per Card for yields, Per Ability (Install) for triggers).

Both desktop and mobile get an **Auto-Triggers** menu mirroring the existing Auto-Yields menu, with its own listing dialog and a separate Disable-All toggle. Most of the diff is this UI work — new `VAutoTriggers` views on both platforms and menu/hotkey/preferences wiring to mirror the existing Auto-Yields dialogs.

Persistence follows the same four-tier hierarchy as yields:
- **Per Card (Each Game)** — the default; preserves today's per-instance, per-game behaviour
- **Per Ability (Each Match)**
- **Per Ability (Each Session)**
- **Per Ability (Each Install)** — written to the existing \`auto-yields.dat\` file (schema is additive, so existing installs load unchanged)

Default is **Per Card (Each Game)**, so there is **no behavioural change** unless the user opts into a longer-lived tier.

## Design notes

- \`Trigger.getStableKey()\` / \`getYieldKey()\` produce user-decision keys paralleling \`SpellAbility.toUnsuppressedString()\` / \`yieldKey()\`.
- Public API mirrors yields: \`setShouldAlways{Accept,Decline,Ask}Trigger(String, boolean)\`.
- Tier is resolved per-side from the preference and does not cross the wire.
- Remote-client match startup replays persisted ACCEPT/DECLINE decisions via \`NetGameController.replayActiveTriggerDecisions\`, symmetric to \`replayActiveYields\`.
- Engine-internal integer trigger IDs (\`SpellAbility.getSourceTrigger()\`, \`MagicStack.hasStateTrigger(int)\`) are untouched — load-bearing for game-state queries, separate from the user-decision layer.
- Disable toggle is orthogonal to yields (separate menu item, separate flag).
- **Drive-by fix:** the existing yield-side \`setDisableAutoYields\` was client-local only — the server's \`remoteAutoYieldsDisabled\` flag never flipped on remote clients. Now wired through \`ProtocolMethod.setDisableAutoYields(Boolean)\` alongside the new trigger equivalent.

---


forge-game/src/main/java/forge/game/trigger/Trigger.java
```diff
@@ -149,6 +149,40 @@ public String toString(boolean active) {
         }
     }
 
+    /**
+     * Returns a stable key for persisting Always-Yes / Always-No decisions.
+     * Omits the per-game-state "triggerRemembered" suffix from toString so the
+     * key matches across game instances. Staleness across Forge versions or
+     * card-text edits is acceptable -- stale keys silently fail to match.
+     */
+    public final String getStableKey() {
```
tool4ever: just overload existing method

forge-gui/src/main/java/forge/localinstance/properties/ForgeConstants.java
```diff
@@ -352,6 +352,11 @@ public final class ForgeConstants {
     public static final String AUTO_YIELD_PER_ABILITY_SESSION = "Per Ability (Each Session)";
     public static final String AUTO_YIELD_PER_ABILITY_INSTALL = "Per Ability (Each Install)";
 
+    public static final String AUTO_TRIGGER_PER_CARD = "Per Card (Each Game)";
```
tool4ever: I wonder if this doesn't overcomplicate things too much:
do we really expect many users would need these on different levels given how closely related they are? 🤔

forge-gui-desktop/src/main/java/forge/screens/match/VAutoTriggers.java
```diff
@@ -0,0 +1,126 @@
...
+import forge.toolbox.FButton;
+import forge.toolbox.FCheckBox;
+import forge.toolbox.FList;
+import forge.toolbox.FOptionPane;
+import forge.toolbox.FScrollPane;
+import forge.util.Localizer;
+import forge.view.FDialog;
+
+@SuppressWarnings("serial")
+public class VAutoTriggers extends FDialog {
```
tool4ever: and with the same reasoning do we really want a separate window?
wouldn't it be easier if user doesn't have to switch around and if a trigger yield exists too just see its decision grouped with it...?

forge-gui/src/main/java/forge/localinstance/properties/ForgeConstants.java
MostCromulent: Yeah I was originally cautious you might want persistent yields but per-card triggers if you wanted to time combos in a particular way or change mid match. 

But thinking about it more: in that case you just don't use the always-yes/always-no settings, or you clear the ones you don't want in the menu. 

So I think fine to consolidate triggers and yields into the one preference hierarchy + menu. I'll refactor.

MostCromulent:
Have consolidated yield and triggers into same menu structure. 

Also added search filter since plausibly if you have "Per Install" preference you'll end up with a lot of saved decisions.

Screenshot 2026-05-05 064145

forge-game/src/main/java/forge/trackable/TrackableProperty.java
```diff
@@ -257,6 +257,7 @@ public enum TrackableProperty {
     //StackItem
     Key(TrackableTypes.StringType),
```
tool4ever, line 258: why are we not simply reusing this one?

forge-game/src/main/java/forge/game/trigger/Trigger.java
```diff
@@ -140,7 +150,7 @@ public String toString(boolean active) {
                     desc = TextUtil.fastReplace(desc, "EFFECTSOURCE", getHostCard().getEffectSource().getDisplayName());
             }
             sb.append(desc);
-            if (!this.triggerRemembered.isEmpty()) {
+            if (includeRemembered && !this.triggerRemembered.isEmpty()) {
```
tool4ever: wouldn't this part fit better into `WrappedAbility.getStackDescription` anyway?

forge-game/src/main/java/forge/game/spellability/SpellAbility.java
```diff
@@ -994,10 +994,10 @@ public void resetOnceResolved() {
         }
     }
 
-    // key for autoyield - the card description (including number) (if there is a card) plus the effect description
+    // key for autoyield - the card name (if there is a card) plus the effect description
     public String yieldKey() {
         if (getHostCard() != null) {
-            return getHostCard().toString() + ": " + toUnsuppressedString();
+            return getHostCard().getName() + ": " + toUnsuppressedString();
```
tool4ever: what's happening here?!
if we'd always drop the ID now `abilitySuffix` becomes useless...

forge-gui-desktop/src/main/java/forge/screens/match/views/VStack.java
```diff
@@ -301,15 +301,15 @@ private final class AbilityMenu extends JPopupMenu {
...
 
-        private Integer triggerID = 0;
+        private String triggerYieldKey = "";
+        private boolean triggerAbilityScope;
 
         public AbilityMenu(){
             jmiAutoYield = new JCheckBoxMenuItem(Localizer.getInstance().getMessage("cbpAutoYieldMode"));
             jmiAutoYield.addActionListener(arg0 -> {
                 final String key = item.getKey();
                 final boolean autoYield = controller.getMatchUI().getGameController().shouldAutoYield(key);
```
tool4ever: these can also be replaced with the new fields...?

forge-gui/src/main/java/forge/interfaces/IGameController.java
```diff
@@ -58,18 +58,20 @@ public interface IGameController {
...
     boolean getDisableAutoYields();
     void setDisableAutoYields(boolean disable);
 
     // Trigger accept/decline preferences
-    boolean shouldAlwaysAcceptTrigger(int trigger);
-    boolean shouldAlwaysDeclineTrigger(int trigger);
-    void setShouldAlwaysAcceptTrigger(int trigger);
-    void setShouldAlwaysDeclineTrigger(int trigger);
-    void setShouldAlwaysAskTrigger(int trigger);
+    boolean shouldAlwaysAcceptTrigger(String key);
```
tool4ever: how much could we reduce the code if we'd just consolidate the setters for Yield/Accept/Decline/Ask + their 3 getters into one for each which uses an int built from bit flags?

MostCromulent: > The interface bloat is real — the five trigger-decision methods on IGameController (3 setters + 2 getters) all funnel into AutoYieldStore's existing tri-state TriggerDecision enum (ASK / ACCEPT / DECLINE). The state lives as one value in storage; the interface just unspools it into method names.
> 
> One nuance on the literal bit-flag suggestion: the three states are mutually exclusive (you can't be "always-accept and always-decline"), so flags imply orthogonality that isn't there. The closer fit is to lift the existing TriggerDecision enum into IGameController:
> 
> ```java
> TriggerDecision getTriggerDecision(String key);
> void setTriggerDecision(String key, TriggerDecision decision, boolean isAbilityScope);
> ```
> 
> Yield is orthogonal (yield-priority on any SA, vs. auto-respond to optional-trigger prompts), so merging it into the same call would lose that distinction; better to leave the yield methods alone.
> 
> **Code-saved analysis**
> 
> The IGameController interface goes from 5 methods to 2, saving 3 lines. The forwarding bodies in PlayerControllerHuman, NetGameController, and YieldController save roughly 29 lines combined. The toggle pattern at call sites (6 sites at about 3 lines each) saves another 18 or so. The wire layer (YieldUpdate, server-side handlers) is neutral.
> 
> Net estimate is roughly 50 lines across about 10 files, with around 15 edit sites touched. The win is the API shape, not the LoC.

tool4ever: ok, sounds good

tool4ever (approved):

PR #10507, opened 2026-04-23
Release interchangeable blockers as a group in notNeededAsBlockers

MostCromulent:
See https://github.com/Card-Forge/forge/pull/10507#issuecomment-5723093240

---

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -425,11 +425,95 @@ public class AiAttackController {
             // try to use strongest as attacker first
             CardLists.sortByPowerDesc(blockers);
 
-            for (Card c : blockers) {
+            // Batch contiguous interchangeable blockers (see isAiEquivalentBlocker) in
+            // a single predict call — reduces N per-blocker predicts to 1, or
+            // 1 + log2(N) on the bisect path, for groups of identical tokens
+            int i = 0;
+            outerLoop:
+            while (i < blockers.size()) {
+                Card c = blockers.get(i);
                 if (vigilantes.contains(c)) {
                     // TODO predict the chance it might die if attacking
+                    i++;
                     continue;
                 }
+
+                int groupEnd = i + 1;
+                while (groupEnd < blockers.size()
+                        && !vigilantes.contains(blockers.get(groupEnd))
+                        && isAiEquivalentBlocker(c, blockers.get(groupEnd))) {
+                    groupEnd++;
+                }
+
+                if (groupEnd - i > 1) {
+                    int groupSize = groupEnd - i;
+                    for (int j = 0; j < groupSize; j++) notNeededAsBlockers.add(blockers.get(i + j));
+                    int lifeAll = ComputerUtil.predictNextCombatsRemainingLife(ai, playAggro, pilotsNonAggroDeck, 0, notNeededAsBlockers);
+
+                    if (lifeAll == Integer.MIN_VALUE) {
+                        for (int j = 0; j < groupSize; j++) notNeededAsBlockers.remove(blockers.get(i + j));
+                        int lo = 0, hi = groupSize - 1;
+                        int safeK = 0;
+                        int safeLife = lastAcceptableBaselineLife;
+                        while (lo <= hi) {
+                            int mid = (lo + hi) / 2;
+                            for (int j = 0; j < mid; j++) notNeededAsBlockers.add(blockers.get(i + j));
+                            int probeLife = ComputerUtil.predictNextCombatsRemainingLife(ai, playAggro, pilotsNonAggroDeck, 0, notNeededAsBlockers);
+                            for (int j = 0; j < mid; j++) notNeededAsBlockers.remove(blockers.get(i + j));
+                            if (probeLife != Integer.MIN_VALUE) {
+                                safeK = mid;
+                                safeLife = probeLife;
+                                lo = mid + 1;
+                            } else {
+                                hi = mid - 1;
+                            }
+                        }
+                        if (safeK > 0) {
+                            boolean tradeOk = true;
+                            if (pilotsNonAggroDeck) {
+                                int dmgPerToken = perAttackerDmg(c, thresholdMod);
+                                int totalDelta = Math.abs(safeLife - lastAcceptableBaselineLife);
+                                if (totalDelta > dmgPerToken * safeK) {
+                                    tradeOk = false;
+                                }
+                            }
+                            if (tradeOk) {
+                                for (int j = 0; j < safeK; j++) notNeededAsBlockers.add(blockers.get(i + j));
+                                if (pilotsNonAggroDeck) lastAcceptableBaselineLife = safeLife;
+                            }
+                        }
+                        break outerLoop;
+                    }
+
+                    if (pilotsNonAggroDeck) {
+                        int dmgPerToken = perAttackerDmg(c, thresholdMod);
+                        int totalDelta = Math.abs(lifeAll - lastAcceptableBaselineLife);
+                        if (totalDelta > dmgPerToken * groupSize) {
+                            // aggregate trade fails — retry per-token to match baseline's partial-accept
+                            for (int j = 0; j < groupSize; j++) notNeededAsBlockers.remove(blockers.get(i + j));
+                            for (int j = 0; j < groupSize; j++) {
+                                Card cj = blockers.get(i + j);
+                                notNeededAsBlockers.add(cj);
+                                int lifej = ComputerUtil.predictNextCombatsRemainingLife(ai, playAggro, pilotsNonAggroDeck, 0, notNeededAsBlockers);
+                                if (lifej == Integer.MIN_VALUE) {
+                                    notNeededAsBlockers.remove(cj);
+                                    break outerLoop;
+                                }
+                                if (Math.abs(lifej - lastAcceptableBaselineLife) > dmgPerToken) {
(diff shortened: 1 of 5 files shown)


Hanmac:
The fast checkout methods currently might ignore `hasRemoveIntrinsic` 

also, the order of the checks might need to be reversed from the layer, to get early return true
- Layer 6 check for new abilities
- Layer 4 check for remove Intrinsic, **if true, return empty**
- Layer 3 check for text changes
- Layer 1 check for default values (LeftState/RightState only if current state is default)

tool4ever:
- Fix 1: sounds ok, will test that later - though 100+ new lines there will make understanding the logic a bit more complicated 😅
- Fix 2+3: should be safe
- Fix 4-6: seems more messy, maybe it'd be better to extend the keywordsCache approach instead so it covers all traits - then instead of yes/no for updating you'll be able to reuse each List multiple times afterwards
@Hanmac didn't we have an issue for that lying around somewhere?

Hanmac:
> * then instead of yes/no for updating you'll be able to reuse each List multiple times afterwards

i mean there is `ICardTraitChanges`, which passes the List through as parameter:

```java
    default List<SpellAbility> applySpellAbility(List<SpellAbility> list) { return list;}
    default List<Trigger> applyTrigger(List<Trigger> list) { return list;}
    default List<ReplacementEffect> applyReplacementEffect(List<ReplacementEffect> list) { return list;}
    default List<StaticAbility> applyStaticAbility(List<StaticAbility> list) { return list;}
```

I might do `KeywordCollection` implement `ICardTraitChanges`, so it can iterate over its keywords internally

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -494,6 +570,52 @@ public final List<Card> notNeededAsBlockers(final List<Card> currentAttackers, f
         return notNeededAsBlockers;
     }
 
+    /**
+     * Whether two blockers are indistinguishable to
+     * {@link ComputerUtil#predictNextCombatsRemainingLife}'s combat math.
+     * Used by {@link #notNeededAsBlockers} to batch predict calls over
+     * groups of identical tokens.
+     */
+    public static boolean isAiEquivalentBlocker(Card a, Card b) {
```
tool4ever, line 590: 1. I don't think "indistinguishable" is strictly true, e.g. `AiBlockController#changesPTWhenBlocked` is based on a trigger check. Though covering only the more common cases seems enough for this heuristic
2. However I'd still like to see some consideration for the *opportunity cost* introduced this way - yes people who encounter lots of token creatures currently complain, but some noise has to be taken into account. Therefore - for this to be worth it - the additional runtime of adding these attacker grouping checks for games with almost no tokens needs to be acceptable...
3. I'm not even convinced this grouping is currently needed:
isn't the remaining life way more often heavily influenced by the amount of remaining creatures (for chump blocking) or their total toughness (against tramplers)?
So just always using binary search (maybe with a bias towards smaller boardstates) makes more sense? 🤔

MostCromulent: Have run some tests on this, results below:

> **Method**: 
>Tested two variations against linear baseline (the code prior to this PR with no algorithm enabled):
-**H003**: the shipped fix in this PR - equivalence-class grouping + batch binary search in `notNeededAsBlockers`
-**H003b**: the proposed simplification — drop the equivalence-class detection entirely and just binary-search the whole sorted blocker list. 
>
>Built H003b on top of the testbed branch and ran (a) an in-flight verify that runs both algorithms on identical pre-decision state at every `notNeededAsBlockers` call and compares release sets; (b) bucketed per-call ms-by-blocker-count timing on a 20-game precon batch + 10-game Commander batch with three variants: linear baseline, shipped H003, and H003b.
> 
> **Decision-making**: 1,313 in-flight verify calls across both formats, 2 divergences (99.85% agreement). Where they diverge, H003b is strictly more conservative — declines an aggressive trade H003 takes. Both choices are reasonable AI play; the grouping isn't producing weird decisions.
> 
> **Performance** at 20 random precon games, fixed seed:
> 
> |                          | Linear (no algorithm) | H003 (PR) | H003b (simplification)  |
> |--------------------------|-----------------------|-------------|---------|
> | Batch wall               | 197.6s                | **131.0s**  | 202.0s  |
> | ms/call N ≤ 5 (95% of decisions) | 68 ms         | **29 ms**   | 42 ms   |
> | ms/call N6-15 (5% of decisions)  | 198 ms        | 203 ms      | **144 ms** |
> 
>Note: each of these tests kept the other fixes in this PR turned on; only H003/H003b was isolated.
>
> The opportunity-cost concern — that grouping checks add runtime in non-token games — turns out empirically negative: H003 is ~34% wall-faster than the same code without it even in random precon games where token swarms don't appear. H003b is essentially tied with linear at the batch level (slight regression); it only beats H003 on the rare wide-board calls.
> 
> **Why**: even N ≤ 5 "heterogeneous" precon boards routinely have partial duplicates — two of the same vanilla creature drawn off a 60-card deck, two basic-stat tokens from a single ETB. H003's grouping engages on those runs and batch-probes a 2-duplicate group in 1 predict call instead of 2. H003b can't recognise duplicates so it always does log(N) binary-search probes + up to N damage-trade walks, which on small boards comes out to *more* predict calls than even plain linear iteration. The grouping mechanism isn't gratuitous overhead; it's a net wall-saver even where there are no token swarms.
> 
> **Recommendation**: keep H003 as designed. The simpler H003b doesn't recover the wins H003 gets from batch-collapsing partial duplicates on small boards, and it's roughly tied with no-algorithm-at-all on the aggregate workload. The ~50 LOC of equivalence-class machinery is doing real work in real games, not just on token swarms.

The wall clock times here aren't particularly helpful, the per call times are the key measure. Wall times get noisy over a long test cycle (even with seeded starts) because a faster algorithm means AI has more time before it hits its thinking time cap, this means it has time to make a different decision, that decision then cascades into a different game state which can mean a longer game length.

forge-game/src/main/java/forge/game/card/CardState.java
```diff
@@ -783,6 +824,31 @@ public FCollectionView<ReplacementEffect> getReplacementEffects(boolean rulesHos
...
     }
+
+    /**
+     * True when {@link #getReplacementEffects(boolean)} would produce an
+     * empty collection. The checks below must mirror every RE source in
+     * that method and in {@link Card#updateReplacementEffects} — adding a
+     * new RE path there requires a matching early-out here.
+     */
+    private boolean canUseReplacementEffectsFastPath(boolean rulesHost) {
+        if (!replacementEffects.isEmpty()) return false;
```
tool4ever, line 835: looks a bit annoying to maintain, maybe the PR I've mentioned above which just caches everything instead is a better direction 🤔

MostCromulent:
Most of what this PR proposed has since landed in master or is covered by open PRs. Only the `notNeededAsBlockers` change is still outstanding, so the branch has been rebased onto current master and cut back to that one fix.

## The rebase

| Change in this PR | Status | Covered by |
|---|---|---|
| `Card.getChangedCardTraitsList` empty short-circuit | in master | #11389 |
| `Card.hasRemoveIntrinsic` fast path | in master | #11389, in a narrower form that guards on `changedCardTypes` alone per CR 305.7 |
| `Card.hasNoTraitOverlays` helper | dropped | only ever served the three `CardState` fast paths below |
| `CardState.getTriggers` empty-source fast path | superseded | #11931, #11366 |
| `CardState.getStaticAbilities` empty-source fast path | superseded | #11931, #11366 |
| `CardState.getReplacementEffects` empty-source fast path | superseded | #11931 (#11366 deliberately leaves replacement effects out) |
| `KeywordCollection` trait bitmask | superseded | #11931 bumps its version inside `KeywordInstance.createTraits`; with #10516 landed, keyword-side caching belongs in `applyTrigger` / `applyStaticAbility` / `applyReplacementEffect` |
| `CardFactoryUtil.setupKeywordedAbilities` cache invalidation | superseded | #11931, which closes the same staleness window one layer down |
| `AiAttackController.notNeededAsBlockers` grouped release | remains | — |

Both caching PRs cover the empty case as well as the general one: a card with no sources rebuilds an empty list once per version rather than on every call, so an empty-source early-out on top of either buys nothing.

## The outstanding fix

`notNeededAsBlockers` adds one potential blocker at a time and runs `predictNextCombatsRemainingLife` after each, so a board of N interchangeable creatures costs N full combat predictions to answer a question that has the same answer for every member of the run. A run is now released in one prediction, bisected when releasing all of it would be fatal, and retried one at a time when the run as a whole is a bad damage trade.

Measured by calling the method directly on a fixed board, median of nine runs:

| Board | before | after |
|---|---|---|
| 15 different creatures | 178 ms | 174 ms |
| 10 identical creatures | 90 ms | 25 ms |
| 20 identical creatures | 580 ms | 64 ms |
| 40 identical creatures | 4492 ms | 198 ms |
| 12 identical plus 15 different | 883 ms | 509 ms |

The released set is the same in every one of those, and on the board with nothing to group the two builds sit inside each other's run-to-run spread. The 40-creature case is the point of the change: one call there costs most of the default five-second AI decision budget.

Whole games were checked too, with the AI decision timeout raised so that neither build abandons an evaluation the other finishes. On an ordinary deck there is no material difference in per-game time. On a goblin-token deck the games are faster: over fifteen seeded games per build, a mean of 5932 ms before and 3792 ms after, and 270 ms against 184 ms per turn.

MostCromulent:
Closing this PR as I found a rules correctness issue I need to work on. Will bring in a fresh PR when that is sorted.

PR #10514, opened 2026-04-24
Improve multiplayer AI interaction decisions

Madwand99:
Summary:

This PR improves multiplayer AI behavior around counterspells and bounce/removal-style interaction.

The main behavior change is that the AI is less likely to spend interaction on fights between other opponents. In multiplayer games, CounterAi now avoids countering targeted spells, opposing counterspells, or predicted threats that only affect opponents, while still countering broad effects like board wipes when they threaten the AI's own permanents.

ChangeZoneAi also avoids using bounce as a "save" spell for an opponent's permanent that is already being removed by another spell on the stack.

Supporting fixes:
- Threat prediction now reads target choices from the stack instance when available, instead of relying only on the SpellAbility object.
- Non-battlefield zone changes are treated as threats only for in-play permanents, and only when the source controller is hostile to the permanent controller.
- GameSnapshot now restores spell-on-stack targets, so copied counterspells keep targeting the spell they were countering.
- AITest's three-player setup helper now supports simulation and configurable AI seating.

Tests:

Ran: mvn -pl forge-gui-desktop -am -Dtest=forge.ai.ability.CounterAiTest,forge.ai.ability.ChangeZoneAiTest -Dsurefire.failIfNoSpecifiedTests=false test

Tested many games of Commander vs 3 Baral decks. Before changes the AI wasted spells all the time. After the changes, games are now more challenging and there are no WTF moments.


forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
```diff
@@ -1005,6 +1005,13 @@ private static boolean isPreferredTarget(final Player ai, final SpellAbility sa,
                 }
                 // bounce opponent's stuff
                 list = CardLists.filterControlledBy(list, ai.getOpponents());
+                if (!game.getStack().isEmpty()) {
+                    // Avoid "saving" an opponent's permanent that another stack effect is already removing.
+                    list.removeAll(ComputerUtil.predictThreatenedObjects(ai, null).stream()
```
tool4ever: should probably refactor this method because it at least partly assumes that the threat prediction is for objects controlled by the provided player

forge-ai/src/main/java/forge/ai/ability/CounterAi.java
```diff
@@ -22,12 +23,106 @@
...
 import forge.util.collect.FCollectionView;
 
 public class CounterAi extends SpellAbilityAi {
 
+    private TargetChoices getTargetChoicesOnStack(Game game, SpellAbility stackSpell) {
+        if (stackSpell == null) {
+            return null;
+        }
+
+        final SpellAbilityStackInstance si = game.getStack().getInstanceMatchingSpellAbilityID(stackSpell);
```
tool4ever: you don't need such detours (you'd miss subability effects this way too)

forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
Madwand99: Done! I refactored the callsite to use a narrower helper instead of interpreting generic threatened GameObjects locally. ChangeZoneAi now calls ComputerUtil.predictThreatenedCards(...), which makes the intent clearer and removes the card-filtering/casting logic from the caller. I can refactor this further if desired, the current version is simply an abstraction.

forge-ai/src/main/java/forge/ai/ability/CounterAi.java
Madwand99: I removed the SpellAbility -> stack lookup by id -> target choices detour and switched the multiplayer checks to operate on the actual SpellAbilityStackInstance. Those checks now also walk subabilities, so they won’t miss relevant targeting that lives below the root ability.

forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
tool4ever: no, putting it in an extra method is still just bloat - why do you think you need to cast type here?

tbh I may try and cherry-pick the best parts from this PR again when I find the time
might be easier than trying to make your AI clean it up 🤔

Madwand99: I think I've cleaned up this issue. I'll do some additional cleanup to see if I can resolve the other issues.

jamincollins:
## Findings
#### 1. GameSnapshot reverse stack iteration may break target resolution order
The `copyStack` method now iterates `fromGame.getStack().reverseIterator()` while building `copiedSpells` mappings incrementally. Targets that reference SpellAbilities deeper in the stack are copied before those targets are mapped, causing `copiedSpells.get(targetSa.getId())` to return null and silently drop the target.

Test case:
```java
Game game = initAndCreateThreePlayerGame(true);
Player caster = game.getPlayers().get(0);
Player counter = game.getPlayers().get(1);
Card spell = addCardToZone("Timber Wolves", caster, ZoneType.Hand);
Card counterspell = addCardToZone("Counterspell", counter, ZoneType.Hand);

SpellAbility spellSa = spell.getFirstSpellAbility();
spellSa.setActivatingPlayer(caster);
ComputerUtil.handlePlayingSpellAbility(caster, spellSa, null);

SpellAbility counterSa = counterspell.getFirstSpellAbility();
counterSa.setActivatingPlayer(counter);
counterSa.getTargets().add(spellSa);
ComputerUtil.handlePlayingSpellAbility(counter, counterSa, null);

GameSnapshot snap = new GameSnapshot(game);
Game copy = snap.makeCopy();
SpellAbility copiedCounter = copy.getStack().get(copy.getStack().size()-1).getSpellAbility();
assert copiedCounter.getTargets().getFirstTargetedSpell() != null : "Counter target lost after snapshot";
```
Impact: AI simulation using snapshots will see counters without targets, leading to incorrect decision evaluation.

#### 2. ComputerUtil.predictThreatenedObjects playerObjects filter short-circuits empty results too early
Filtering `objects` to only those controlled by `aiPlayer` and returning early when `playerObjects.isEmpty()` prevents evaluation of threats to the player object itself when `objects` contains only the player. The early return returns `threatened` which may be empty even though the player object should be evaluated.

Test case:
```java
Player ai = game.getPlayers().get(1);
SpellAbility sa = createDamageAllSpell(targetingPlayer = ai);
List<GameObject> threatened = ComputerUtil.predictThreatenedObjects(ai, null);
assert threatened.contains(ai) : "Player object should be considered threatened";
```
Impact: Damage-to-player threats may be missed, causing AI to incorrectly value protection.

#### 3. ChangeZoneAi removes threatened objects without null check on game.getStack()
The new block `if (!game.getStack().isEmpty()) { for (Player opponent : ai.getOpponents()) { list.removeAll(ComputerUtil.predictThreatenedObjects(opponent, null)); } }` assumes `predictThreatenedObjects` never returns null. If `predictThreatenedObjects` returns null due to unexpected state, `removeAll` will NPE.

Test case:
```java
// Simulate corrupted game state where predictThreatenedObjects returns null
Player opponent = ai.getOpponents().iterator().next();
try {
    List<GameObject> threatened = ComputerUtil.predictThreatenedObjects(opponent, null);
    assert threatened != null : "predictThreatenedObjects must not return null";
} catch (Exception e) {
    // NPE expected
}
```
Impact: NullPointerException during bounce decision in multiplayer.

#### 4. CounterAi.isOpponentCounteringAnotherOpponent assumes getActivatingPlayer() non-null
The method accesses `currentAbility.getActivatingPlayer()` and `targetedSpell.getActivatingPlayer()` without null checks before calling `isOpponentOf`. If a SpellAbility on stack has no activating player (e.g., copy, trigger), NPE occurs.

Test case:
```java
SpellAbility orphan = new SpellAbility(...);
orphan.setActivatingPlayer(null);
assert isOpponentCounteringAnotherOpponent(ai, orphan) == false : "Should handle null activating player";
```
Impact: Crash when evaluating counter decisions on stack with orphaned abilities.

#### 5. GameSnapshot.mapCopiedSpellAbilities recursive mapping duplicates IDs on self-mapping
The initial seeding `mapCopiedSpellAbilities(toEntry.getSpellAbility(), toEntry.getSpellAbility(), copiedSpells)` maps an ability to itself. Later `mapCopiedSpellAbilities(origSa, newSa, copiedSpells)` may overwrite with new mapping. If `origSa.getId() == newSa.getId()` due to ID reuse, the map will point to the wrong instance.

Test case:
```java
SpellAbility orig = ...;
SpellAbility copy = ...;
copy.setId(orig.getId()); // simulate ID collision
Map<Integer, SpellAbility> map = new HashMap<>();
GameSnapshot.mapCopiedSpellAbilities(orig, copy, map);
assert map.get(orig.getId()) == copy : "Mapping should point to copy";
```
Impact: Target resolution may resolve to original ability instead of copy, corrupting simulation state.

#### 6. ComputerUtil zone change threat now includes non-Exile destinations, may over-flag self-blink
The condition changed from `Destination.equals("Exile")` to `!Battlefield && !Stack`. Self-blink spells moving to hand then back may be flagged as threats to own permanents, causing AI to waste resources.

Test case:
```java
SpellAbility blink = createChangeZoneSpell(destination="Hand");
Card ownCreature = addCardToBattlefield(ai);
assert !ComputerUtil.predictThreatenedObjects(ai, null).contains(ownCreature) : "Self-blink should not threaten own creature";
```
Impact: AI may incorrectly protect against own blink effects.

#### 7. CounterAi threatensAny uses top-only prediction, may miss threats deeper in stack
`threatensAny` calls `predictThreatenedObjects(player, null, true)`, which only evaluates top of stack. A threat buried under counterspell may be missed, leading to premature counter decisions.

Test case:
```java
// Stack: Wrath at bottom, Counterspell on top
assert threatensAny(ai.getOpponents()) == false : "Should see wrath threat under counter";
```
Impact: AI may incorrectly decide not to counter because top-only view misses real threat.

### Performance and security
#### 8. GameSnapshot copyTargets recurses without cycle detection
`copyTargets` recurses into `subAbility` without checking for cycles. Malformed abilities with circular subAbility references will cause StackOverflowError.

Test case:
```java
SpellAbility a = new SpellAbility();
SpellAbility b = new SpellAbility();
a.setSubAbility(b);
b.setSubAbility(a);
// copyTargets should detect cycle
```
Impact: Denial of service via crafted card data causing stack overflow.

## Verdict
Request Changes
The PR improves multiplayer AI decision making but introduces target resolution ordering bugs in GameSnapshot, null safety gaps in CounterAi and ChangeZoneAi, and potential over-flagging of zone changes. The reverse stack iteration in snapshot copying risks silent loss of SpellAbility targets, which undermines simulation accuracy.

## Recommended Fixes
1. Process stack copies in original order and pre-populate copiedSpells map before copying targets, or iterate forward with a two-pass mapping.
2. Guard `predictThreatenedObjects` against null returns and ensure player object threats are evaluated even when `playerObjects` is empty.
3. Add null checks for `getActivatingPlayer()` in `isOpponentCounteringAnotherOpponent` and `predictThreatenedObjects`.
4. Add cycle detection to `copyTargets` and `mapCopiedSpellAbilities`.
5. Narrow zone change threat condition to exclude self-controlled zone changes and verify destination semantics.

---

PR #10521, opened 2026-04-25
Improve AI land destruction targeting with contextual priority

Madwand99:
Adds a contextual land-removal scorer so the AI values land destruction targets by actual threat, including high-mana lands, mana-boosting Auras, manlands, utility lands, dangerous named lands, Homeward Path only when relevant, and tempo-aware use of sacrifice lands like Strip Mine, Wasteland, and Tectonic Edge. Multiplayer-aware so priority for lands can be different depending on how many players exist.

Related to #5604

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
@@ -38,6 +38,7 @@ import forge.game.combat.CombatUtil;
 import forge.game.cost.Cost;
 import forge.game.cost.CostPayEnergy;
 import forge.game.cost.CostRemoveCounter;
+import forge.game.cost.CostSacrifice;
 import forge.game.cost.CostUntap;
 import forge.game.keyword.Keyword;
 import forge.game.keyword.KeywordCollection;
@@ -51,11 +52,19 @@ import forge.game.spellability.SpellAbility;
 import forge.game.staticability.StaticAbility;
 import forge.game.staticability.StaticAbilityMode;
 import forge.game.trigger.Trigger;
+import forge.game.trigger.TriggerType;
 import forge.game.zone.MagicStack;
 import forge.game.zone.ZoneType;
 import forge.item.PaperCard;
 
 public class ComputerUtilCard {
+    private static final List<String> DANGEROUS_LANDS_TO_REMOVE = Arrays.asList(
+            "Dark Depths",
+            "Glacial Chasm",
+            "Valakut, the Molten Pinnacle",
+            "Maze of Ith"
+    );
+
     public static Card getMostExpensivePermanentAI(final CardCollectionView list, final SpellAbility spell, final boolean targeted) {
         CardCollectionView all = list;
         if (targeted) {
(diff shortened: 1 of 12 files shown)


forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
```diff
@@ -258,6 +267,229 @@ public static Card getBestLandAI(final Iterable<Card> list) {
                 .orElseGet(() -> Aggregates.random(bLand)); // random tapped land of least represented type
     }
 
+    public static Card getBestLandToRemoveAI(final Player ai, final Iterable<Card> list, final SpellAbility removal) {
```
tool4ever, line 281: Should be built upon `LandEvaluator` imo

Madwand99: Done! Great idea.

forge-ai/src/main/java/forge/ai/ability/DestroyAi.java
```diff
@@ -425,23 +425,34 @@ public boolean doLandForLandRemovalLogic(SpellAbility sa, Player ai, Card tgtLan
         boolean canColorLock = (oppSkippedLandDrop || oppLands.size() > 3)
                 && tgtLand.isBasicLand() && CardLists.count(oppLands, CardPredicates.nameEquals(tgtLand.getName())) == 1;
 
-        // Non-basic lands are currently not ranked in any way in ComputerUtilCard#getBestLandAI, so if a non-basic land is best target,
-        // consider killing it off unless there's too much potential tempo loss.
-        // TODO: actually rank non-basics in that method and then kill off the potentially dangerous (manlands, Valakut) or lucrative
-        // (dual/triple mana that opens access to a certain color) lands
-        boolean nonBasicTgt = !tgtLand.isBasicLand();
+        int targetPriority = ComputerUtilCard.evaluateLandRemovalPriority(ai, tgtLand, sa);
+        boolean mediumPriorityTgt = targetPriority >= 50;
```
tehdiplomat: We should probably put evaluation thresholds near the evaluation functions as a constant, rather than just inline here.

forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
```diff
@@ -258,6 +287,207 @@ public static Card getBestLandAI(final Iterable<Card> list) {
...
+            // were already scored above; this catches external animation.
+            score += isAttackingAi(land, ai) ? 140 : 55;
+        }
+
+        if (isKnownDangerousLand(land)) {
+            // High priority floor: oddball lands whose danger is hard to infer
+            // from their generic ability shape, like Dark Depths or Glacial
+            // Chasm. Do not add here, because some of these also get a good
+            // baseline score from LandEvaluator.
+            score = Math.max(score, 170);
```
tool4ever: we shouldn't hardcode names, just add support for `AIEvaluationModifier` following the CreatureEvaluator

forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
tool4ever: the Comparator doesn't really combine both values this way

try to move most of the logic into it so other callers can benefit too

Madwand99: I removed the extra `evaluateLand` tiebreaker, since the removal priority score already includes the LandEvaluator baseline and `thenComparingInt` was only acting as a tie-breaker rather than combining the values.

I looked at moving more logic into LandEvaluator, but I think most of this targeting logic is too contextual for land play/evaluation decisions. Homeward Path depends on whether the AI controls stolen permanents, Strip Mine/Wasteland-style lands depend on whether the AI has a high-priority land to protect and whether the opposing land is tapped, manlands depend heavily on combat context, and the Aura cases depend on what the Aura removed or whether it boosts mana.

For that reason I kept LandEvaluator as the baseline and layered the removal-specific logic in `evaluateLandRemovalPriority`, with card-script `AILandRemovalMinScore` only affecting removal targeting rather than general land play decisions.

forge-ai/src/main/java/forge/ai/ability/DestroyAi.java
```diff
@@ -425,23 +425,34 @@ public boolean doLandForLandRemovalLogic(SpellAbility sa, Player ai, Card tgtLan
...
 
-        boolean timingCheck = canManaLock || canColorLock || nonBasicTgt;
+        boolean timingCheck = canManaLock || canColorLock || mediumPriorityTgt;
         boolean tempoCheck = numLandsOTB >= amountNoTempoCheck
                 || ((numLandsInHand >= amountLandsInHand || isHighPriority) && ((numLandsInHand + numLandsOTB >= amountNoTimingCheck) || timingCheck));
 
+        // Tectonic Edge, Strip Mine, and Wasteland should not cash in a large
+        // share of the AI's own mana base for a merely medium utility target.
+        boolean sacrificesSourceLand = sa.getHostCard().isLand()
+                && sa.getPayCosts() != null
```
tool4ever: use `ComputerUtilCost.isSacrificeSelfCost`

forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
```diff
@@ -258,6 +278,197 @@ public static Card getBestLandAI(final Iterable<Card> list) {
...
+    private static boolean aiControlsStolenCreature(final Player ai) {
+        for (Card creature : ai.getCreaturesInPlay()) {
+            if (!creature.getOwner().equals(ai)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static boolean hasManaSubAbility(final SpellAbility ability) {
```
tool4ever: reuse `SpellAbility.findSubAbilityByType`

forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
tool4ever: ok, I can agree with those arguments :)

forge-ai/src/main/java/forge/ai/ability/DestroyAi.java
Madwand99: Done!

forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
Madwand99: Done!

tool4ever (approved):

PR #10533, opened 2026-04-26
Auto-call coin flips for win/lose outcomes

RafaelHGOliveira:
## Summary

Adds a new preference **Auto-Call Coin Flips** (default **on**) that skips the Heads/Tails prompt for coin flips where only the win/lose outcome matters (e.g. Krark's Thumb, Frenetic Efreet, Mana Clash, Goblin Bomb).

The mathematical outcome is unchanged: any call on a fair coin has 50/50 odds, so auto-calling Heads simply removes a redundant click that the player would have to make every flip. Cards where Heads vs Tails actually matters (no `call`) still prompt as before.

## Motivation

In games with cards like Krark's Thumb, the player is asked twice per flip — once for the call, then once to pick which of the two results to keep. The call is mechanically meaningless on its own (50/50), and the player almost always picks Heads followed by the Heads result. This is repetitive in long games, especially Commander matches. Power players who want full manual control can disable the preference.

## Behavior

- `chooseBinary(HeadsOrTails)` — returns `true` (Heads) when the preference is enabled.
- `chooseFlipResult(..., call)` — when `call == true` (the Krark-style replacement effect picks one of two flips to keep), returns `true` to keep the matching result and guarantee the called side. When `call == false` (Win/Lose-the-flip choice that doesn't depend on the call), the player is still prompted.

## Implementation

| File | Change |
|------|--------|
| `ForgePreferences.java` | New `UI_AUTO_CALL_COIN_FLIP` pref, default `true` |
| `PlayerControllerHuman.java` | Pref check in `chooseBinary` and `chooseFlipResult` |
| `VSubmenuPreferences.java` | New checkbox in Gameplay Preferences |
| `CSubmenuPreferences.java` | Bind checkbox to pref |
| `en-US.properties` | Three i18n keys (`cbAutoCallCoinFlip`, `nlAutoCallCoinFlip`) |

## Test plan

- [x] `mvn -B compile -pl forge-gui-desktop -am -DskipTests` — clean build
- [ ] Cast Mana Clash with pref **on** — flips resolve without prompts, damage applied normally
- [ ] Cast Mana Clash with pref **off** — Heads/Tails prompt appears for each flip
- [ ] Cast something with Krark's Thumb in play, pref **on** — single auto-resolved flip, replacement effect picks the better result automatically
- [ ] Cast Frenetic Efreet, pref **on** — auto-resolves
- [ ] Cards with NoCall semantics (e.g. cards that ask "did you flip Heads or Tails?") — still prompt the player regardless of pref


jamincollins:
## Findings
#### 1. chooseFlipResult auto-return can declare a win on two Tails flips
Description: In PlayerControllerHuman.chooseFlipResult, the PR adds an early return when `call && pref`:
```java
if (call && FModel.getPreferences().getPrefBoolean(FPref.UI_AUTO_CALL_COIN_FLIP)) {
    return true;
}
```
This returns `true` unconditionally, meaning the controller always claims the Heads result was kept. FlipCoinEffect then computes `wonOrHeads = result == choice`. With auto-call, `choice` is also forced to `true` via `chooseBinary`. Therefore `wonOrHeads` is always `true`, even when both of Krark's Thumb's two flips are Tails. The game will log a win and fire `TriggerType.FlippedCoin` with `Result=true` despite the actual random outcomes.

The comment claims "picking the matching result guarantees a 'win'", but the code does not verify that a matching result actually exists. The correct behavior is to keep a Heads flip if one exists, otherwise keep Tails and lose. The current simplification removed the `boolean[] results` parameter in commit 0baf6730857, so the controller no longer knows the actual flips.

Test case:
```java
// Simulate Krark's Thumb with auto-call enabled
PlayerControllerHuman ctrl = ...;
FModel.getPreferences().setPref(FPref.UI_AUTO_CALL_COIN_FLIP, "true");
SpellAbility sa = mockSpellAbility();
Player flipper = mockPlayer();
// Two flips both Tails
boolean[] actualFlips = new boolean[]{false, false};
// Controller is called with call=true
boolean chosen = ctrl.chooseFlipResult(sa, flipper, true);
// chosen is true, but no Heads flip exists
// FlipCoinEffect will compute wonOrHeads = true == true = true → false win
assert !chosen; // should reflect that no Heads exists
```
Impact: Players can be granted false wins on replacement-effect coin flips, breaking MTG rules and producing incorrect game state and triggers.

#### 2. chooseBinary early return bypasses InputConfirm and recordConfirm
Description: `chooseBinary` for `HeadsOrTails` returns `true` immediately when the preference is enabled, skipping `InputConfirm.confirm` and `recordConfirm`. The original path records the confirmation for replay/undo and UI feedback. Skipping it means auto-called flips are not recorded, which can desynchronise replay logs and confuse the confirm history.

Test case:
```java
// Verify recordConfirm is called for manual call
PlayerControllerHuman ctrl = ...;
FModel.getPreferences().setPref(FPref.UI_AUTO_CALL_COIN_FLIP, "false");
ctrl.chooseBinary(sa, "Call?", BinaryChoiceType.HeadsOrTails, null);
// recordConfirm should be invoked

FModel.getPreferences().setPref(FPref.UI_AUTO_CALL_COIN_FLIP, "true");
ctrl.chooseBinary(sa, "Call?", BinaryChoiceType.HeadsOrTails, null);
// recordConfirm is NOT invoked → replay mismatch
```
Impact: Replay integrity and debug logs are incomplete for auto-called flips.

#### 3. Preference default true changes existing user experience without migration notice
Description: `UI_AUTO_CALL_COIN_FLIP` is registered with default `"true"`. Existing users upgrading will silently have all Heads/Tails prompts skipped. The UI description says "Automatically call Heads...", but there is no onboarding prompt or migration notice. Users who relied on manual calls for superstition or for specific card interactions will be surprised.

Impact: Unexpected behavior change on upgrade; potential user confusion.

#### 4. No-call flips still prompt, but the preference name is misleading
Description: The preference is named "Auto-Call Coin Flips" and the UI text mentions win/lose outcomes. For `NoCall` flips, `chooseBinary` is not invoked, so the pref has no effect, which is correct. However `chooseFlipResult` is still called with `call=false` for NoCall, and the early return is guarded by `call && pref`, so it correctly falls through to UI. Edge case is fine, but the name suggests it applies to all flips, which may mislead.

Impact: Minor UX inconsistency.

#### 5. Desktop-only UI change; mobile settings missing
Description: The UI checkbox is added only to `forge-gui-desktop/src/main/java/forge/screens/home/settings/CSubmenuPreferences.java` and `VSubmenuPreferences.java`. There is no corresponding change for mobile or other platforms. Users on mobile cannot toggle the preference, so the feature is effectively forced on for them.

Impact: Feature inconsistency across platforms; mobile users cannot opt out.

#### 6. AI controller does not respect the preference, creating asymmetry
Description: `PlayerControllerAi.chooseFlipResult` always returns `true` for `call=true` and random for `call=false`. It does not read `UI_AUTO_CALL_COIN_FLIP`. In a mixed human/AI game, human auto-call behavior differs from AI, which is acceptable, but the preference is a UI preference only for humans. The asymmetry is fine, but the code comment in `PlayerControllerHuman` claims the auto-call is mathematically neutral, while AI already optimizes for win. No conflict, but worth documenting.

Impact: No bug, but inconsistency in reasoning.

### Performance and security
#### 7. Preference lookup on every coin flip
Description: Both `chooseBinary` and `chooseFlipResult` call `FModel.getPreferences().getPrefBoolean(FPref.UI_AUTO_CALL_COIN_FLIP)` on every invocation. This is a cheap map lookup, but it occurs in hot game logic paths. Caching the value in the controller instance would avoid repeated lookups. Not a performance issue today, but adds unnecessary overhead.

Impact: Negligible performance cost; minor code smell.

## Verdict
Request Changes
The PR introduces a useful UX improvement but contains a logic flaw in `chooseFlipResult` that can falsely declare wins when both Krark's Thumb flips are Tails, and it bypasses confirmation recording. The desktop-only UI change also limits the preference to one platform. Fix the win-detection logic and ensure consistent UI/UX before merging.

## Recommended Fixes
1. Make `chooseFlipResult` respect actual flip outcomes. Since the `boolean[] results` parameter was removed, either restore it for this decision or change the contract so `FlipCoinEffect` passes the needed information to decide whether a matching result exists. A minimal fix is to return `true` only if at least one Heads flip exists; otherwise return `false`. This requires re-introducing results awareness.
2. Preserve `recordConfirm` for auto-called flips by calling it with the implied choice before early return.
3. Add the preference toggle to mobile/desktop settings UI and ensure the default is documented in release notes.
4. Cache the preference value in `PlayerControllerHuman` on initialization to avoid repeated lookups.

---

PR #10536, opened 2026-04-26
Eliminate redundant protocol round-trip: cache UI skip-phase prefs

MostCromulent:
## What's the issue

In network multiplayer, **the host pings the remote client over the wire on every priority transition** to ask "is your UI configured to skip this phase?". Each call goes through `RemoteClientGuiGame.isUiSetToSkipPhase` → `sendAndWait(ProtocolMethod.isUiSetToSkipPhase, …)` — a full TCP round-trip. The host's game thread parks until the reply lands.

`PlayerControllerHuman.getActivePlayerInput` makes this call every time a remote human gets priority and the stack is empty — order-of-magnitude **~13 calls per turn for one remote human, ~26 for two**.

The data being asked for is **static across most of a match** — just the client's UI label state for "stop at this phase". There's no reason to round-trip it every time.

## What's the fix

The client tells the host its full skip-phase state once at match start, and pushes per-toggle deltas thereafter. The host caches it per `PlayerControllerHuman` and answers `isUiSetToSkipPhase` from local memory — zero round-trips on the hot path.

This is the same pattern recently established for auto-yield prefs (the controller-owns-prefs refactor): per-player preference state lives on `IGameController`, the host's `PlayerControllerHuman` keeps a remote-mirror map gated on `isRemoteClient()`, and `NetGameController` ships mutations via 1:1-named `Mode.CLIENT` protocol methods. Initial state is replayed at game start by the client (`replayActiveYields` for yields; `replayUiSkipPhases` for skip-phase here).

The only structural difference: yield prefs live in stores loaded before the game (so the seed fires from `FGameClient.setGameControllers`), whereas skip-phase prefs live in UI label state, which is populated by `actuateMatchPreferences` after game start. The seed therefore fires from the end of `actuateMatchPreferences` — the first point at which all label state and player views are settled.

### How it works

Three logical pieces:

1. **Host-side cache.** `PlayerControllerHuman` gains `Map<PlayerView, EnumSet<PhaseType>> remoteSkipPhases`. New `IGameController.isUiSetToSkipPhase` reads from it when serving a remote (`gui instanceof RemoteClientGuiGame`); the empty-cache default returns `false` (don't skip), matching a fresh UI's conservative default. The two `getActivePlayerInput` call sites swap from `getGui().isUiSetToSkipPhase(…)` to `this.isUiSetToSkipPhase(…)`.

2. **Client-side push.** New `Mode.CLIENT` `ProtocolMethod.setUiShouldSkipPhase(PlayerView, PhaseType, Boolean)` carries a single `(player, phase, shouldSkip)` mutation. `NetGameController.replayUiSkipPhases(allPlayers, isSkipped)` is a one-shot helper that pushes the full snapshot — only `shouldSkip == true` entries, since the host's empty cache already represents `false`.

3. **UI wiring.** `PhaseLabel` (desktop) and `VPhaseIndicator.PhaseLabel` (mobile) gain `setOnToggled(Runnable)`, fired post-flip. `CMatchUI.actuateMatchPreferences` (desktop) and `MatchController.actuateMatchPreferences` (mobile) wire per-`(player, phase)` callbacks at the end of label initialization, then fire the initial `replayUiSkipPhases` for any local `NetGameController`. Mid-game toggles ship one ~80-byte message each.

The old `Mode.SERVER ProtocolMethod.isUiSetToSkipPhase` is deleted. The `IGuiGame.isUiSetToSkipPhase` method stays — still used locally by `FControlGamePlayback`, and by the local-fallback branch inside `PlayerControllerHuman.isUiSetToSkipPhase`. All are local reads with no network cost.

## Why it's safe

- **Local play and hotseat are unchanged.** When `isRemoteClient()` is false, `isUiSetToSkipPhase` delegates to `getGui()` (label state, exactly as today) and `setUiShouldSkipPhase` is a no-op. The on-toggle push and the seed are gated on `instanceof NetGameController`, so local controllers (`PlayerControllerHuman`) skip them entirely. Existing hotseat semantics — both rows seeded from `PHASES_HUMAN`, independent toggles mid-match, last-write-wins on save — are preserved exactly.
- **Mind-slave correctness preserved.** Desktop's `CMatchUI.isUiSetToSkipPhase(turnPlayer, phase)` AND-combines the turn player's row with the master's row. When the master toggles their label, the cached answer for every player they control changes too. `pushSkipPhaseToControllers` iterates players and re-pushes any dependent — narrow edge case (Mindslaver, Sorin Markov ult) but the cache stays consistent with what the GUI would return locally.
- **Empty-cache default is the conservative choice.** Until the seed lands (a brief window at match start), the host returns "don't skip" → stop at every phase → user gets prompted as if they hadn't configured anything. No phase ever gets accidentally skipped before the snapshot arrives.

---

### DIFF
--- a/forge-gui-mobile/src/forge/screens/match/MatchController.java
+++ b/forge-gui-mobile/src/forge/screens/match/MatchController.java
@@ -39,7 +39,9 @@ import forge.game.player.PlayerView;
 import forge.game.spellability.SpellAbilityView;
 import forge.game.zone.ZoneType;
 import forge.gamemodes.net.NetworkGuiGame;
+import forge.gamemodes.net.client.NetGameController;
 import forge.gamemodes.match.HostedMatch;
+import forge.interfaces.IGameController;
 import forge.gui.FThreads;
 import forge.gui.GuiBase;
 import forge.gui.util.SGuiChoose;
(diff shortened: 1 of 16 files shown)


forge-gui-mobile/src/forge/screens/match/MatchController.java
```diff
@@ -562,11 +564,33 @@ private static void actuateMatchPreferences() {
...
+        for (VPlayerPanel panel : panels) allPlayers.add(panel.getPlayer());
+        for (IGameController c : instance.getOriginalGameControllers()) {
+            if (c instanceof NetGameController) {
+                ((NetGameController) c).replayUiSkipPhases(allPlayers, instance::isUiSetToSkipPhase);
+            }
+        }
+    }
+
+    private static void pushSkipPhaseToControllers(final PlayerView player, final PhaseType phase) {
+        final boolean shouldSkip = instance.isUiSetToSkipPhase(player, phase);
```
tool4ever: imo you can just adapt the mindslaver extra from desktop here too
so we then can share at least this method in the more fitting parent class

tool4ever (approved):

PR #10537, opened 2026-04-26
Fix UI bug causing opening hand to be invisible

Madwand99:
Fixes a desktop UI bug where a local player’s opening hand could be invisible during the mulligan prompt in multiplayer games.

The issue was easiest to reproduce in a 4-player game when the match entered the mulligan prompt while a different player’s field tab was selected. It could also be reproduced around conceding/restarting when the local player’s own tab was not selected. In that state, CHand.updateHand() tried to prepare the player’s hand display, but it returned early if the field/library-label layout anchor was not ready or the match screen was not currently active. Because that early return happened before the hand card panels were populated/revealed, the player’s hand tab could exist but show no visible cards.

There was a second wrinkle: newly created hand CardPanels start with displayEnabled=false, because normal card draw handling reveals them at the end of the draw animation. During mulligan/setup, if the animation cannot run because the layout anchor is unavailable, those placeholders stayed disabled. Later drawing a card triggered a normal hand update/animation path, which is why the UI partially or fully recovered after drawing.

This fix changes CHand.updateHand() so it always updates the hand contents first. Animation setup now happens afterward and is treated as optional. If the library label, field view, or screen state is not ready for animation, the new placeholders are revealed immediately using the existing non-animated Animation.moveCard(placeholder) path.

I also changed desktop hand lookup to find a VHand by its owning PlayerView instead of assuming the visible hand tab list has the same indexes as the full multiplayer player list. That avoids fragile behavior in multiplayer layouts where not every player necessarily has a visible hand document.

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
+++ b/forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
@@ -380,9 +380,12 @@ public final class CMatchUI
         return view.getHands();
     }
     public VHand getHandFor(final PlayerView p) {
-        final int idx = getPlayerIndex(p);
-        final List<VHand> allHands = getHandViews();
-        return idx < 0 || idx >= allHands.size() ? null : allHands.get(idx);
+        for (final VHand hand : getHandViews()) {
+            if (p.equals(hand.getPlayer())) {
+                return hand;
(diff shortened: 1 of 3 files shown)


tool4ever (approved):
thanks

PR #10538, opened 2026-04-27
Combat readability fix for stacked tokens

Madwand99:
Summary

Fixes combat readability for stacked token creatures. Previously, identical tokens could remain visually stacked after being blocked, so red blocking arrows pointed into an overlapping stack and made it hard to tell which specific token was blocked. This was especially painful when choosing combat tricks after blocks.

Fix

Blocked attacking tokens are now separated from their stacks during battlefield layout. The same logic also applies to stacked creatures when creature stacking is enabled. The layout checks both assigned blockers and planned blockers, so the separation works during block selection as well as after blocks are declared.

Combat updates now also trigger a tabletop relayout, ensuring the stack separation appears immediately when blocking changes.

To reduce visual clutter, separated blocked attackers are sorted left-to-right by the position of their blocker or average blocker position. This helps red blocking arrows line up more directly and avoids many unnecessary crossings when several tokens are blocked at once.

BlockingTokens


MostCromulent:
Haven't had a chance to test, but how does this handle multiple tokens assigned to block same attacker - does is stack those on top of each other in per-attacker piles?

I was dealing with some similar logic in https://github.com/Card-Forge/forge/pull/9751 (no issue with this PR as a standalone change, just noting the conceptual overlap on this part).

Madwand99:
This fix has no effect on how blocking with tokens works, although it probably should. I'll work on that.

Madwand99:
I've added code to separate out both blocking and attacking tokens.
SeparateBlockers2

forge-gui-desktop/src/main/java/forge/view/arcane/PlayArea.java
```diff
@@ -256,6 +264,91 @@ private CardStackRow collectAllCreatures(List<CardPanel> remainingPanels) {
...
+        if (hasAny(blockers)) {
+            return blockers;
+        }
+        final Iterable<CardView> plannedBlockers = combat.getPlannedBlockers(card);
+        return hasAny(plannedBlockers) ? plannedBlockers : null;
+    }
+
+    private static boolean containsCard(final Iterable<CardView> cards, final CardView card) {
+        if (cards == null) {
+            return false;
```
tool4ever: looks null safe to me, so no need for this helper

tool4ever (changes requested):
if it's working as you've intended it needs to be gated behind an opt-in preference

it can be a bit confusing when you're selecting different attackers for blocking and suddenly stuff moves around on both fields (sometimes even when you click on the same spot which feels weird)
and undeclare might take more effort since another card is now under your mouse...

Madwand99:
Updated this to make the combat stack behavior opt-in.

Changes made:

- Added a new preference, UI_SEPARATE_COMBAT_STACKS, defaulting to false.
- Added a desktop Preferences checkbox: “Separate Combatants from Stacks”.
- Gated the combat-based token/creature separation behind that preference.
- Gated the extra battlefield relayout during combat updates behind the same preference, so users who leave it disabled keep the previous stable battlefield behavior during blocker assignment.
- Kept the existing non-combat stack splitting behavior unchanged. Tokens still separate for existing reasons like different counters, P/T, text, attachments, stack size, etc.
- Added localized strings for the new preference.

Removed the extra helper from the prior review comment and use the combat collections directly.
With the preference enabled, stacked creatures that are attacking or blocking are separated and sorted by their combat counterparts to make red combat arrows easier to follow. With it disabled, the UI should behave as before.
SeparateTokensGUIOption

tool4ever (approved):

PR #10539, opened 2026-04-27
Token selection settings on desktop

clairchiara:
Add the token selection tickbox to desktop settings


forge-gui/src/main/java/forge/localinstance/properties/ForgePreferences.java
```diff
@@ -90,6 +90,7 @@ public enum FPref implements PreferencesStore.IPref {
         UI_OVERLAY_ABILITY_ICONS("true"),
         UI_OVERLAY_DRAFT_RANKING("true"),
         UI_ENABLE_ONLINE_IMAGE_FETCHER ("true"),
+        UI_ENABLE_SMART_TOKEN_ART ("false"),
```
tool4ever, line 79: let's not add preferences that aren't hooked up to anything...

clairchiara: Whoops, should be hooked up now I think

PR #10540, opened 2026-04-27
Remove dead-host setless URL entries for tokens now in TDLS

vanja-ivancevic:
Follow-up to #10518.

- Removes 5 setless URL entries from `token-images.txt` (`b_0_1_thrull.jpg`, `g_1_1_saproling.jpg`, `w_1_1_citizen.jpg`, `u_1_1_camarid.jpg`, `r_1_1_goblin.jpg`) that point at `downloads.cardforge.org`, a host with no DNS A record (NXDOMAIN). All 5 URLs return HTTP 000 (no connection possible) from any client today.
- These entries pre-date the TDLS edition added in #10518 and were intended as setless fallbacks for tokens that had no edition-specific registration. Now that TDLS provides edition-specific URLs for the same 5 scripts, the dead setless entries serve no purpose.
- The risk they pose: legacy clients (years ago, when the host was alive) downloaded these as loose `pics/tokens/<script>.jpg` files. When Forge now picks TDLS via the new era-match preference and the per-edition cache is empty, `ImageKeys.getImageFile` falls through its lookup chain to those loose files and serves stale modern-frame art. Removing the entries means fresh installs cannot produce the conflicting loose files in the first place.
- Existing affected users still need to delete loose `.jpg` files at the root of `pics/tokens/` to clear their stale cache. That is a one-time cleanup, not a code fix.


vanja-ivancevic:
Should we delete all dead cardforge URLs?

PR #10546, opened 2026-04-27
Add card art selection dialog to desktop deck editor

MostCromulent:
Screenshot 2026-04-28 072135

Screenshot 2026-04-28 072201

## Summary

Adds a Change Art dialog to the desktop deck editor (Constructed-style editors), mirroring the "Replace Card Variant" right-click option in the mobile deck builder.

Right-click a card in the deck pane to open a themed grid of all printings, pick one, swap in place. Foil preserved; current printing first, then newest set first.

Includes debounced set-name search and an art-style dropdown (Standard / Borderless / Full Art / Showcase / Extended / Retro / Promo). The style filter is driven by edition section data, with a new `CardEdition.getSectionForCollectorNumber` helper to map a printing to its section.

Note: card art is not exhaustively categorised in edition data, so some printings won't appear under the style you'd expect — e.g. a full-art printing may not show up under "Full Art" if its edition doesn't tag the section that way.

Async download for missing images.

---


forge-gui-desktop/src/main/java/forge/screens/deckeditor/controllers/CEditorConstructed.java
```diff
@@ -355,13 +359,16 @@ public static void buildRemoveContextMenu(EditorContextMenuBuilder cmb, DeckSect
...
+            addChangeArtEntryIfApplicable(cmb);
             break;
         case Sideboard:
             cmb.addMoveItems(localizer.getMessage("lblRemove"), localizer.getMessage("lblfromsideboard"));
             cmb.addMoveAlternateItems("Move", "to deck");
+            addChangeArtEntryIfApplicable(cmb);
             break;
         case Commander:
             cmb.addMoveItems(localizer.getMessage("lblRemove"), localizer.getMessage("lblascommander"));
+            addChangeArtEntryIfApplicable(cmb);
```
Jetz72: Why limit it to these three sections? [Having alternative artwork isn't exclusive to traditional cards](https://scryfall.com/search?q=(t%3Avanguard+or+t%3Ascheme+or+t%3Aplane)+illustrations>1&unique=art).

You might just do something similar to the `addMakeFoils` bit below.

Might consider using the same "Replace Variant" terminology, or we could change both to "Replace Print" - another use-case for this kind of feature is swapping Attraction cards to a print with the same art but a different set of lights.

forge-gui-desktop/src/main/java/forge/screens/deckeditor/ChangeArtDialog.java
```diff
@@ -0,0 +1,388 @@
...
+            final Graphics2D g2 = (Graphics2D) g.create();
+            g2.setRenderingHint(RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_BICUBIC);
+            g2.setRenderingHint(RenderingHints.KEY_RENDERING, RenderingHints.VALUE_RENDER_QUALITY);
+            g2.drawImage(source, x, y, displayW, displayH, null);
+            g2.dispose();
+        }
+    }
+
+    /** JPanel-based renderer: card image on top, text underneath. Used in HORIZONTAL_WRAP JList. */
+    private static final class GridCellRenderer extends JPanel implements ListCellRenderer<PaperCard> {
```
Jetz72: A grid view for card images might be useful enough to make this a stand-alone class.

Jetz72 (commented):
>Note: card art is not exhaustively categorised in edition data, so some printings won't appear under the style you'd expect — e.g. a full-art printing may not show up under "Full Art" if its edition doesn't tag the section that way.

I wonder if that'll end up producing many bug reports. At some point it'd be good to have granular tagging for prints that could operate on the edition, section, or individual print level.

MostCromulent:
> >Note: card art is not exhaustively categorised in edition data, so some printings won't appear under the style you'd expect — e.g. a full-art printing may not show up under "Full Art" if its edition doesn't tag the section that way.
> 
> I wonder if that'll end up producing many bug reports. At some point it'd be good to have granular tagging for prints that could operate on the edition, section, or individual print level.

It will definitely result in reports. 

Do you think better to remove the category dropdown for this PR and we can come back to it when we have some sort of comprehensive tagging solution?

Or include it and we can just improve on that over time?

tool4ever:
imo using the categories already available still helps

forge-gui-desktop/src/main/java/forge/screens/deckeditor/ChangeArtDialog.java
MostCromulent: Thinking forward, there are a few desktop match dialogs which could be adapted to use this. For example I think the current "your opponent has revealed the following cards in hand" notification just shows a small text list of the cards.

forge-gui-mobile/src/forge/deck/FDeckEditor.java
```diff
@@ -1862,7 +1862,7 @@ protected void buildMenu(final FDropDownMenu menu, final PaperCard card) {
                 //if card has more than one art option, add item to change user's preferred art
                 final List<PaperCard> artOptions = FModel.getMagicDb().getCommonCards().getAllCardsNoAlt(card.getName());
                 if (artOptions.size() > 1) {
-                    menu.addItem(new FMenuItem(Forge.getLocalizer().getMessage("lblChangePreferredArt"), Forge.hdbuttons ? FSkinImage.HDPREFERENCE : FSkinImage.SETTINGS, e -> {
```
Jetz72, line 1865: Setting preferred art is different from swapping the print of one or more cards in the deck. I guess desktop doesn't have this feature either?

forge-gui-desktop/src/main/java/forge/screens/deckeditor/ChangePrintingDialog.java
```diff
@@ -0,0 +1,196 @@
...
+        if (newCard.equals(oldCard)) {
+            return;
+        }
+        final CardManager deckManager = (CardManager) CDeckEditorUI.SINGLETON_INSTANCE
+                .getCurrentEditorController().getDeckManager();
+        deckManager.removeItem(oldCard, 1);
+        deckManager.addItem(newCard, 1);
+        CDeckEditorUI.SINGLETON_INSTANCE.getCurrentEditorController()
+                .getDeckController().notifyModelChanged();
+    }
```
Jetz72: I'm unsure if searching the deck editor out statically is better than replicating the blocking `dialog.show()` pattern that `AddBasicLandsDialog` uses, but I feel like they should be consistent with one another unless there's a good reason for them not to be.

forge-gui-mobile/src/forge/deck/FDeckEditor.java
MostCromulent: Have reverted changes to mobile. Desktop doesn't have preferred printings - may look at whether its worth implementing that later, but not for this PR.

Jetz72 (approved):

PR #10555, opened 2026-04-28
YieldController MVP — per-PCH yield state with unified wire envelope

MostCromulent:
## Summary

Spin-off MVP carved from PR #9643 (YieldRework). Refactors yield state into a per-`PlayerControllerHuman` `YieldController` with a unified `YieldUpdate` wire envelope. Deletes 8 yield-related "ugly netplay fields" from PCH; collapses 5 fragmented yield ProtocolMethod entries into 2.

Architecture doc attached: **[yield-controller-architecture.md](https://github.com/user-attachments/files/27182451/yield-controller-architecture.md)**

## Included

- New `YieldController` class — owns markers, stack-yield, legacy autopass, per-card/ability auto-yield, trigger decisions, skip-phase prefs (one instance per PCH).
- Unified `YieldUpdate` sealed envelope — single `sendYieldUpdate` / `applyYieldUpdate` ProtocolMethod pair carries every yield sync direction.
- Atomic `SeedFromClient(snapshot)` at game start — replaces N-message replay loops with one wire message per controlled player. Covers per-card/ability auto-yields, `autoYieldsDisabled`, skip-phase prefs.
- Phase markers (right-click desktop / long-press mobile on phase indicators) — auto-pass priority until the marked phase is reached, then clears.
- "Yield to entire stack" stack-item context menu entry — resolves the entire stack including post-cancel additions.
- Master interrupt semantics preserved (`MagicStack.add()` opponent-spell cancel clears legacy autopass; markers and stack-yield deliberately survive engine-driven cancels).
- All ungated — no experimental flag.

## Deferred to follow-up

- Smart suggestions in `InputPassPriority` (`PendingSuggestion` enum)
- Five interrupt prefs (`YIELD_INTERRUPT_ON_*`), `YIELD_AUTO_PASS_RESPECTS_INTERRUPTS`, `YIELD_AUTO_PASS_NO_ACTIONS`
- Speed Settings (`YIELD_SKIP_PHASE_DELAY`, `YIELD_SKIP_RESOLVE_DELAY`)
- Mass-destruction / opponent-targeting interrupt evaluator
- Host/client experimental-flag reconciliation
- `VYieldOptions` mobile dialog and `VYieldSettings` Speed Settings section

The remainder of YieldRework rebases onto this MVP and ships those features in a follow-up PR.

---


forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -3595,24 +3596,39 @@ private boolean isPromptingForTrigger(final int trigger) {
...
-            EnumSet<PhaseType> set = remoteSkipPhases.get(turnPlayer);
-            return set != null && set.contains(phase);
+            return yieldController.isSkippingPhase(turnPlayer, phase);
         }
         return getGui().isUiSetToSkipPhase(turnPlayer, phase);
     }
 
     @Override
     public void setUiShouldSkipPhase(final PlayerView turnPlayer, final PhaseType phase, final boolean shouldSkip) {
         if (!isRemoteClient()) return;
```
tool4ever: is this method even still reachable?

forge-gui-desktop/src/main/java/forge/toolbox/special/PhaseLabel.java
```diff
@@ -59,11 +80,67 @@ public void mouseExited(final MouseEvent e) {
...
+
+    public void setYieldMarked(final boolean b) {
+        if (this.yieldMarked != b) {
+            this.yieldMarked = b;
+            repaintOnlyThisLabel();
+        }
+    }
+
+    private void handleRightClick() {
+        if (phaseOwner == null || phaseType == null) {
```
tool4ever: if this can't be done the same way as `onToggled` might as well handle left clicks in this other direction too to unify? (same for mobile)

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -3501,9 +3506,9 @@ private AutoYieldStore.Tier activeTier() {
...
     public boolean shouldAutoYield(final String key) {
         if (isRemoteClient()) {
-            if (remoteAutoYieldsDisabled) return false;
-            if (remoteCardYields.contains(key)) return true;
-            return remoteAbilityYields.contains(AutoYieldStore.abilitySuffix(key));
+            if (yieldController.isAutoYieldsDisabled()) return false;
+            if (yieldController.shouldAutoYieldKey(key, false)) return true;
+            return yieldController.shouldAutoYieldKey(AutoYieldStore.abilitySuffix(key), true);
         }
         if (localStore().isDisabled()) return false;
```
tool4ever: imo more logic in these methods should be moved

the YieldController can use AutoYieldStore for both local and remote and decide which one it queries in a cleaner way

forge-gui-desktop/src/main/java/forge/toolbox/special/PhaseIndicator.java
```diff
@@ -14,18 +17,18 @@ public class PhaseIndicator extends JPanel {
...
-    private PhaseLabel lblBeginCombat = new PhaseLabel("BC");
-    private PhaseLabel lblDeclareAttackers = new PhaseLabel("DA");
-    private PhaseLabel lblDeclareBlockers = new PhaseLabel("DB");
-    private PhaseLabel lblFirstStrike = new PhaseLabel("FS");
-    private PhaseLabel lblCombatDamage = new PhaseLabel("CD");
-    private PhaseLabel lblEndCombat = new PhaseLabel("EC");
-    private PhaseLabel lblMain2 = new PhaseLabel("M2");
-    private PhaseLabel lblEndTurn = new PhaseLabel("ET");
-    private PhaseLabel lblCleanup = new PhaseLabel("CL");
+    private PhaseLabel lblUpkeep = new PhaseLabel("UP", PhaseType.UPKEEP);
```
Hanmac: The PhaseIndicator on Desktop probably should be refactored like the one on Mobile using a Map for all the PhaseTypes:
https://github.com/Card-Forge/forge/blob/b7eb1247858f3e547257a85389c0f837693add97/forge-gui-mobile/src/forge/screens/match/views/VPhaseIndicator.java#L25

forge-gui-desktop/src/main/java/forge/screens/match/views/VPrompt.java
```diff
@@ -65,37 +71,59 @@ public class VPrompt implements IVDoc<CPrompt> {
...
     private KeyAdapter buttonKeyAdapter = new KeyAdapter() {
         @Override
         public void keyPressed(final KeyEvent e) {
             if (e.getKeyCode() == KeyEvent.VK_ESCAPE) {
-                if (btnCancel.isEnabled()) {
-                    if (FModel.getPreferences().getPrefBoolean(FPref.UI_ALLOW_ESC_TO_END_TURN) || !btnCancel.getText().equals("End Turn")) {
-                        btnCancel.doClick();
+                CMatchUI ui = CMatchUI.getActive();
+                if (ui != null) {
+                    PlayerView local = ui.getCurrentPlayer();
```
tool4ever: this should also happen in mobiles `MatchController#keyDown`

maybe it can be turned into a helper in YieldController?

forge-gui/src/main/java/forge/gamemodes/net/client/NetGameController.java
```diff
@@ -204,31 +222,81 @@ public boolean shouldAlwaysDeclineTrigger(final int trigger) {
     @Override
     public void setShouldAlwaysAcceptTrigger(final int trigger) {
         yieldStore.setTriggerDecision(trigger, AutoYieldStore.TriggerDecision.ACCEPT);
-        send(ProtocolMethod.setShouldAlwaysAcceptTrigger, trigger);
```
tool4ever, line 207: same as above, we shouldn't need to store it in two places...

forge-gui-desktop/src/main/java/forge/screens/match/views/VPrompt.java
```diff
@@ -65,37 +71,59 @@ public class VPrompt implements IVDoc<CPrompt> {
...
     }
 
     private KeyAdapter buttonKeyAdapter = new KeyAdapter() {
         @Override
         public void keyPressed(final KeyEvent e) {
             if (e.getKeyCode() == KeyEvent.VK_ESCAPE) {
-                if (btnCancel.isEnabled()) {
-                    if (FModel.getPreferences().getPrefBoolean(FPref.UI_ALLOW_ESC_TO_END_TURN) || !btnCancel.getText().equals("End Turn")) {
-                        btnCancel.doClick();
+                CMatchUI ui = CMatchUI.getActive();
```
tool4ever: why is this not simply using `controller.getMatchUI()`?

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
```diff
@@ -1311,10 +1356,38 @@ private void actuateMatchPreferences() {
...
+                && phaseOwner.equals(existing.getPhaseOwner())
+                && phase == existing.getPhase();
+        if (clickedSameLabel) {
+            controller.sendYieldUpdate(new YieldUpdate.ClearMarker(local));
+        } else {
+            // Setting a marker implies we want to stop here — un-skip the cell
+            // so the marker can fire (skip-phase pref + marker would skip past).
+            label.setEnabled(true);
+            label.repaintOnlyThisLabel();
+            controller.sendYieldUpdate(new YieldUpdate.SetMarker(phaseOwner, phase));
```
tool4ever: in a follow-up we might consider avoiding messing with the stop itself

forge-gui/src/main/java/forge/gamemodes/net/client/NetGameController.java
```diff
@@ -186,49 +164,83 @@ public void clearAutoYields() {
...
-        send(ProtocolMethod.setUiShouldSkipPhase, turnPlayer, phase, shouldSkip);
+        yieldController.setSkipPhase(turnPlayer, phase, shouldSkip);
+        send(ProtocolMethod.sendYieldUpdate, new YieldUpdate.SetSkipPhase(turnPlayer, phase, shouldSkip));
+    }
+
+    @Override
+    public void applyYieldUpdate(final YieldUpdate update) {
+        if (update instanceof YieldUpdate.SetMarker u) {
+            yieldController.setMarker(u.phaseOwner(), u.phase());
+        } else if (update instanceof YieldUpdate.ClearMarker) {
```
tool4ever, line 195: I don't understand why the server would need to send these except for ClearMarker...

and even for that one can't the client just do it on his own via GameEventTurnPhase because he has the marker in his cache anyway? 🤷‍♂️

forge-gui-desktop/src/main/java/forge/screens/match/views/VPrompt.java
```diff
@@ -65,37 +68,42 @@ public class VPrompt implements IVDoc<CPrompt> {
...
 
     private KeyAdapter buttonKeyAdapter = new KeyAdapter() {
         @Override
         public void keyPressed(final KeyEvent e) {
             if (e.getKeyCode() == KeyEvent.VK_ESCAPE) {
-                if (btnCancel.isEnabled()) {
-                    if (FModel.getPreferences().getPrefBoolean(FPref.UI_ALLOW_ESC_TO_END_TURN) || !btnCancel.getText().equals("End Turn")) {
-                        btnCancel.doClick();
-                    }
+                CMatchUI ui = controller.getMatchUI();
```
tool4ever: this doesn't work cleanly:
if we're showing any other prompt (e.g. trigger confirm) ESC shouldn't just cancel the yields instead of what would be expected from the context ⚠️

forge-gui/src/main/java/forge/gamemodes/net/client/NetGameController.java
```diff
@@ -186,49 +164,83 @@ public void clearAutoYields() {
...
+     */
+    public void seedYieldStateOnHost(Map<PlayerView, EnumSet<PhaseType>> skipPhases) {
+        send(ProtocolMethod.sendYieldUpdate,
+                new YieldUpdate.SeedFromClient(yieldController.buildClientSnapshot(skipPhases)));
     }
 
-    @Override
     public void setUiShouldSkipPhase(final PlayerView turnPlayer, final PhaseType phase, final boolean shouldSkip) {
-        send(ProtocolMethod.setUiShouldSkipPhase, turnPlayer, phase, shouldSkip);
+        yieldController.setSkipPhase(turnPlayer, phase, shouldSkip);
```
tool4ever: client shouldn't need to store this extra at this point?

forge-gui/src/main/java/forge/gamemodes/net/client/NetGameController.java
tool4ever: @MostCromulent
doesn't seem to work yet, marker doesn't disappear

forge-gui-desktop/src/main/java/forge/screens/match/views/VPrompt.java
tool4ever: well now it seems almost impossible to reach `cancelGenericYields`?
(unless the button is disabled which wouldn't make sense to user either)

maybe it should just replace `InputLockUI#selectButtonCancel`? 🤔 (or wherever it would disarm the autopass following the button press)

forge-gui/src/main/java/forge/gamemodes/net/client/NetGameController.java
tool4ever: (I was actually testing in a local game)

MostCromulent: Replicated it + a client side bug (yield enabled correctly but chevron did not appear) in network play. Both should now be addressed.

forge-gui/src/main/java/forge/gamemodes/match/AbstractGuiGame.java
```diff
@@ -816,4 +802,27 @@ public void updateDependencies() {
...
+     * {@link YieldController#shouldAutoYield} priority loop — we just detect the transition.
+     */
+    public final void checkMarkerAutoClear() {
+        GameView gv = getGameView();
+        if (gv == null) return;
+        boolean any = false;
+        for (Map.Entry<PlayerView, IGameController> e : gameControllers.entrySet()) {
+            YieldController yc = e.getValue().getYieldController();
+            if (yc == null) continue;
+            boolean hadMarker = yc.getAutoPassUntilMarker() != null;
```
tool4ever: AI seems a bit eager to complicate stuff today:
what's the point of this boolean, isn't that exactly what the return value of the call below is for?

forge-gui-desktop/src/main/java/forge/toolbox/special/PhaseIndicator.java
```diff
@@ -1,192 +1,57 @@
...
-        return this.lblCombatDamage;
-    }
-
-    /** @return {@link javax.swing.JLabel} */
-    public PhaseLabel getLblFirstStrike() {
-        return this.lblFirstStrike;
+    public PhaseIndicator() {
+        this.setOpaque(false);
+        this.setLayout(new MigLayout("insets 0 0 1% 0, gap 0, wrap"));
+        addPhaseLabel("UP", PhaseType.UPKEEP,                      "htmlPhaseUpkeepTooltip");
```
tool4ever: can you refactor these existing language strings into one?
We should only need "Click to toggle. Right click to fast forward."
because we can then just reuse `PhaseType.nameForUi` to construct the whole tooltip

tool4ever (approved):
While I'm not fully happy with all data flows yet it's arguably cleaner than before

PR #10561, opened 2026-04-29
Fix encoder/game thread race; add ID fallback in event refs

MostCromulent:
## Background

Investigating user report of mid-game network freezes/disconnects on Android multiplayer. 

User logs from master showed `EncoderException: ArrayIndexOutOfBoundsException: length=247; index=247` on `setGameView` messages, surfaced by PR #10535's `Network send error` listener. These errors are caught in the encoder pipeline and the game continues, so they don't directly explain the game freeze. But they pointed at a real concurrency bug worth investigating.

To dig further, PR #10558 added supplementary instrumentation. Running manual and batch tests with that instrumentation in place revealed a second pattern that wasn't visible in either master or the user logs: `Could not resolve IdRef` warnings followed by NPEs in `EventVisualizer`/`GameLogFormatter`. The NPE is caught by `GameProtocolHandler` and the game continues — the user-visible cost is a transient UI issue such as a missing sound or game-log entry, not a crash.

Root cause of the user-reported freeze is still unknown, and these bugs aren't proven to cause them — but ruling them out either way is a correctness fix.

### 1. Encoder/game-thread race on the TrackableObject EnumMap

PR #10535 (`Async writes in RemoteClient: stop game thread blocking on slow consumers`) replaced synchronous `.sync()` writes with async listener-based ones. 

The change was correct but had a side effect: the Netty `MessageToByteEncoder` runs on the I/O event-loop thread, and pre-PR-10535 the `.sync()` implicitly guaranteed the game thread was parked while it ran. Post-PR-10535 the game thread continues mutating `TrackableObject` state via `set()` while the encoder serializes the same `EnumMap`. `EnumMap.writeObject` uses a size-then-iterate sequence; a concurrent `put`/`remove` makes iteration walk past the array bounds → AIOOBE.

Compounding: `RemoteClient.write` (used by the event-batching code path) had **no failure listener** — only `send` did. When the encoder threw on a `setGameView` queued via `write`, the failure was silent; the message was dropped on the wire, leaving the client's tracker missing the just-mutated state.

### 2. Event refs assume the receiver's tracker has every id

The wrap-events mechanism replaces nested `CardView`/`PlayerView` refs inside each `GameEvent` with compact wire markers (`IdRef`/`StaleCardRef`). This design predates the current code: the `IdRef` marker was introduced by PR #10018 (in `GameEventProxy`), `StaleCardRef` was added later by PR #9642 (delta sync), and PR #10343 consolidated both into `TrackableSerializer.wrapEvents` while removing `GameEventProxy`. Resolution requires the receiver's tracker to have the referenced id.

That assumption is solid for top-level protocol method args (`selectCard(cv)` — the server only invokes the method after registering the object) but **not** for refs inside events: events can carry CardViews that aren't reachable from `GameView` (transient zone-change copies, off-graph refs). On a tracker miss, resolution returns null and downstream visitors throw NPE on `card.getCurrentState()`. The NPE is caught by `GameProtocolHandler` and the game continues — the user-visible cost is a missing sound or game-log entry, not a crash.

`StaleCardRef` already carried `name + imageKey` for the specific case where the server's tracker held a *different* incarnation of the same id, but the general "tracker doesn't have this id at all" path was uncovered. PR #10343 didn't introduce this gap — it inherited it from the older `GameEventProxy` design.

### 1. Encode on the calling thread

`CompatibleObjectEncoder` exposes `encodeToBuf(msg, alloc)` — runs the encode logic synchronously on the calling thread and returns a fresh `ByteBuf`. `RemoteClient.send`/`write`/`sendAndWait` and `FGameClient.send` now call this before `channel.writeAndFlush(byteBuf)`. The network I/O remains async — only the encoding step moves.

The thread race becomes structurally impossible: the calling thread is the only mutator of the data being serialized, so reads are sequential with writes by definition. `RemoteClient.write` also gains a failure listener so any future encode failures surface instead of silently dropping the message.

Alternative option considered but rejected was to override `TrackableObject.writeObject` to take a brief synchronized snapshot of the `props` `EnumMap` and serialize the snapshot. That removes the EnumMap-level race without changing where encoding runs. However on testing the snapshot was shallow, and `EnumMap` values that are themselves mutable lists aren't covered. Snapshotting "deep enough" reduces to re-implementing serialization, with a fresh defect at every nested data structure. Solving the root cause of the race is the simpler architectural solution.

### 2. Event refs carry display fallback data

Replaced `IdRef`/`StaleCardRef` for event-internal refs with `EventCardRef(id, name, imageKey, preserveSnapshot)` and `EventPlayerRef(id, name, preserveSnapshot)`. `resolve()` tries the tracker first; on a miss, constructs a detached view from the embedded fields. The `preserveSnapshot=true` flag handles the previous `StaleCardRef` case (server's tracker holds a different incarnation than the event captured). `IdRef` is retained for top-level protocol method args, where tracker presence is guaranteed by construction.

---

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/net/CompatibleObjectEncoder.java
+++ b/forge-gui/src/main/java/forge/gamemodes/net/CompatibleObjectEncoder.java
@@ -86,7 +86,16 @@ public class CompatibleObjectEncoder extends MessageToByteEncoder<Serializable> 
         }
     }
 
-    // setGameView and openView carry full state to populate the receiver's tracker.
+    /**
+     * Determines whether TrackableObject references should be replaced with
+     * IdRef markers for this message. Excludes only:
+     * - setGameView/openView: carry full state to populate the client's Tracker
+     *
+     * applyDelta is NOT excluded: its property maps already use Integer IDs
+     * (via DeltaSyncManager.toNetworkValue), and bundled events are wrapped
+     * by TrackableSerializer.wrapEvents before entering the packet, so no
+     * raw TrackableObjects remain in the serialization graph.
+     */
     private static boolean shouldReplaceTrackables(Serializable msg) {
         if (msg instanceof GuiGameEvent event) {
             ProtocolMethod method = event.getMethod();
--- a/forge-gui/src/main/java/forge/gamemodes/net/CompatibleObjectEncoder.java
+++ b/forge-gui/src/main/java/forge/gamemodes/net/CompatibleObjectEncoder.java
@@ -5,6 +5,7 @@ import forge.gui.GuiBase;
 import forge.trackable.Tracker;
 import forge.util.IHasForgeLog;
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.ByteBufOutputStream;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.MessageToByteEncoder;
@@ -29,6 +30,23 @@ public class CompatibleObjectEncoder extends MessageToByteEncoder<Serializable> 
 
     @Override
     protected void encode(ChannelHandlerContext ctx, Serializable msg, ByteBuf out) throws Exception {
+        encodeInto(msg, out, this.tracker, this.byteTracker);
+    }
+
+    /** Caller passes the returned buffer to writeAndFlush, which takes ownership. */
+    public ByteBuf encodeToBuf(Serializable msg, ByteBufAllocator alloc) throws Exception {
+        ByteBuf out = alloc.buffer();
+        try {
+            encodeInto(msg, out, this.tracker, this.byteTracker);
+        } catch (Exception e) {
+            out.release();
+            throw e;
+        }
+        return out;
+    }
+
+    private static void encodeInto(Serializable msg, ByteBuf out, Tracker tracker,
+                                   NetworkByteTracker byteTracker) throws Exception {
         int startIdx = out.writerIndex();
         ByteBufOutputStream bout = new ByteBufOutputStream(out);
         ObjectOutputStream oout = null;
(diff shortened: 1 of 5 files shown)


forge-gui/src/main/java/forge/gamemodes/net/TrackableSerializer.java
```diff
@@ -130,6 +130,17 @@ static Object resolve(Object obj, Tracker tracker) {
             }
             return detached;
         }
+        if (obj instanceof EventPlayerRef ref) {
+            if (!ref.preserveSnapshot) {
+                PlayerView fromTracker = tracker.getObj(TrackableTypes.PlayerViewType, ref.id);
+                if (fromTracker != null) return fromTracker;
+            }
+            PlayerView detached = new PlayerView(ref.id, tracker);
```
tool4ever: do we really need to worry about non-existing PlayerView?
seems a bit too far fetched to future-proof against imo...

forge-gui/src/main/java/forge/gamemodes/net/CompatibleObjectEncoder.java
```diff
@@ -68,16 +86,7 @@ protected void encode(ChannelHandlerContext ctx, Serializable msg, ByteBuf out)
         }
     }
 
-    /**
```
tool4ever, line 71: I'm ok with Fix 1

I don't know why AI was so aggressive in removing JavaDoc from here on though, let's keep that intact unless something needs correction...

forge-gui/src/main/java/forge/gamemodes/net/TrackableSerializer.java
```diff
@@ -47,23 +37,33 @@ static final class IdRef implements Serializable {
...
+    static final class EventCardRef implements Serializable {
         private static final long serialVersionUID = 1L;
         final int id;
-        final String imageKey;
         final String name;
+        final String imageKey;
+        final boolean preserveSnapshot;
 
-        StaleCardRef(int id, String imageKey, String name) {
+        EventCardRef(int id, String name, String imageKey, boolean preserveSnapshot) {
```
tool4ever: could be turned into a record?

forge-gui/src/main/java/forge/gamemodes/net/CompatibleObjectEncoder.java
tool4ever: the JavaDoc removal applies to all parts

MostCromulent: Sorry, should be fixed now. (Guidelines try and train AI away from over-documenting things which is its natural inclination, but seems to have overcorrected here). I should have supervised more closely.

tool4ever: thanks, yea I agree and also dislike trivial ones but these seemed valueable enough given the surrounding complexity

tool4ever (approved):

PR #10567, opened 2026-04-30
Improve AI simulation stack resolution for multiplayer choices

Madwand99:
This PR narrows the earlier AI decision work (https://github.com/Card-Forge/forge/pull/10559) to the simulator stack-resolution fix plus focused sacrifice decision coverage.

GameSimulator.resolveStack() now preserves the simulated AI player’s controller while temporarily installing simulation controllers for every other player. Previously, simulated stack resolution only wrapped one opponent, usually the weakest opponent. That worked for many two-player cases, but in multiplayer a resolving spell or trigger can require choices from several players. If a non-wrapped player needed to choose cards, permanents, targets, or confirm an action, simulation could accidentally hit a normal/UI/test controller instead of an AI simulation controller.

The included regression test uses a three-player Innocent Blood scenario because it forces multiple players to make sacrifice choices during stack resolution. With the old behavior, the non-weakest third player’s controller is reached and the test fails. With the new behavior, all non-preserved players use simulation controllers and the simulated stack resolves cleanly.

This also brings over focused sacrifice tests showing deterministic, value-aware sacrifice choices:

- stable sacrifice choice across repeated simulation-controller calls
- legal-candidate filtering
- worst creatures selected first
- least valuable permanent selected from mixed permanent options


forge-ai/src/main/java/forge/ai/simulation/GameSimulator.java
```diff
@@ -288,7 +284,24 @@ public static void resolveStack(final Game game, final Player opponent) {
...
+
+    private static void runWithSimulationControllers(final Game game, final Player preservedPlayer,
+            final List<Player> players, final int index, final Runnable proc) {
+        if (index >= players.size()) {
+            proc.run();
+            return;
+        }
+
+        Player player = players.get(index);
+        if (player.equals(preservedPlayer)) {
```
tool4ever: imo just applying it when `!PlayerController.isAI()` should be enough

this also reduces the overhead slightly in case it's a game without any human

Madwand99: Done!

tool4ever: try to write less roundabout code:
- you can just consume the first entry of the players list you're passing instead of the extra index counting
- the preservedPlayer also serves no purpose now

Madwand99: Done! The stack resolver now only installs temporary simulation controllers for non-AI players. Existing AI controllers are left in place, which avoids unnecessary controller replacement in all-AI games while still preventing simulated stack resolution from reaching human/test/UI controllers.

I also simplified the helper:
- removed the preserved-player parameter, since `isAI()` now determines whether wrapping is needed
- changed `resolveStack(game, player)` to `resolveStack(game)`
- simplified the recursive helper to consume the local player list instead of carrying an index

The regression test now uses a non-AI test controller, matching the intended case: simulated stack resolution should wrap non-AI players before asking them to make choices.

forge-gui-desktop/src/test/java/forge/ai/sacrifice/AiSacrificeDecisionTest.java
```diff
@@ -126,11 +200,48 @@ public void testAiSacrificesToPillarTombsToAvoidLethalLifeLoss() {
...
+        AssertJUnit.assertTrue(message, player.getZone(zone).contains(card));
+    }
+
+    private CardCollectionView chooseWithSimulationController(Game game, Player ai, SpellAbility source,
+            int amount, CardCollectionView validTargets) {
+        PlayerControllerAi controller = new PlayerControllerAi(game, ai, ai.getLobbyPlayer());
+        controller.setUseSimulation(true);
+
+        final CardCollectionView[] chosen = new CardCollectionView[1];
+        ai.runWithController(() -> chosen[0] = ai.getController().choosePermanentsToSacrifice(
```
tool4ever: I think your AI is fooling you with these tests:
this method doesn't even have simulation support, so wrapping it doesn't really proof anything...
(and otherwise they'd be in the wrong package anyway)

should probably revert this class, it doesn't really provide much benefit anyway

Madwand99: Eh, I kind of fooled myself. Trying to figure out how to split up #10559 and made the wrong call on these tests. Reverted the tests.

forge-gui-desktop/src/test/java/forge/ai/simulation/GameSimulationTest.java
```diff
@@ -61,6 +62,44 @@ public void testActivateAbilityTriggers() {
...
+        addCard("Runeclaw Bear", opponent);
+        addCard("Runeclaw Bear", ai);
+        Card otherOpponentBear = addCard("Runeclaw Bear", otherOpponent);
+
+        Card spell = addCardToZone("Innocent Blood", opponent, ZoneType.Hand);
+        SpellAbility sa = spell.getFirstSpellAbility();
+        sa.setActivatingPlayer(opponent);
+        game.getStack().addAndUnfreeze(sa);
+
+        PlayerControllerForTests throwingController = new PlayerControllerForTests(game, otherOpponent, otherOpponent.getLobbyPlayer()) {
```
tool4ever, line 89: sigh, actually this test isn't proof either and it looks to me like the TODO was just outdated:
I spent the effort of actually debugging and thanks to `clonePlayer` I don't see how a GUI player should end up in a simulated game anyway...

Madwand99: I'll close this PR. If you have any feedback on how I can actually solve the issue of AI drawing cards into Xyris -> Impact tremors -> death, let me know.

PR #10569, opened 2026-04-30
Improve AI mode selection for modal spells

Madwand99:
This improves CharmAi mode selection for modal spells that can choose extra modes when doing so is useful and payable.

Previously, the AI could identify one good mode and stop there, even on cards like Crush Contraband where choosing both modes is free and beneficial. It also did not properly reason about extra mode costs for mechanics like Escalate, Entwine, and Spree/Tiered.

Changes included:

Let non-trigger modal spells add additional useful modes when MinCharmNum < CharmNum.

Treat candidate extra modes like effects the AI can evaluate in the moment via normal spell/trigger heuristics.

Add cost-aware mode checks for:
- Escalate: adds the Escalate cost once per mode beyond the first.
- Spree/Tiered: adds each selected mode’s ModeCost.
- Entwine: pays Entwine only when the full cost is payable and all modes look useful.

Avoid crashes from sub-mode AI that assumes full spell context by falling back safely during optional mode evaluation.

Added regression coverage for:
- Crush Contraband
- Collective Resistance with affordable/unaffordable Escalate
- Journey of Discovery with Entwine
- Requisition Raid with affordable/unaffordable Spree

Remaining known gap:

CanRepeatModes is still not properly handled by CharmAi. Cards like Confluences and Seasons that say “you may choose the same mode more than once” do not yet have smart repeat-mode selection. The current AI can choose multiple distinct good modes, but it does not intentionally rank and repeat the best mode when repetition would be optimal.


Hanmac:
Example of AI where the old code added extra Cost but shouldn't:

AI used `Lively Dirge`: `Return up to two creature cards with total mana value 4 or less from your graveyard to the battlefield.`
while not having any creature in GY.

Because of that `up to`, AI thinks that it might be good enough.

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
```diff
@@ -94,8 +99,6 @@ protected AiAbilityDecision checkApiLogic(Player ai, SpellAbility sa) {
     private List<AbilitySub> chooseOptionsAi(SpellAbility sa, List<AbilitySub> choices, final Player ai, boolean isTrigger, int num, int min) {
         List<AbilitySub> chosen = Lists.newArrayList();
         AiController aic = ((PlayerControllerAi) ai.getController()).getAi();
-        // TODO unused for now, the AI doesn't know how to effectively handle repeated choices
-        boolean allowRepeat = sa.hasParam("CanRepeatModes");
```
tool4ever, line 98: hmn, you are removing this comment - but at the same time not handling this case...

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
```diff
@@ -150,6 +153,18 @@ private List<AbilitySub> chooseOptionsAi(SpellAbility sa, List<AbilitySub> choic
                 }
             }
         }
+        if (!isTrigger && chosen.size() < num && min < num) {
+            choices.removeAll(chosen);
+            for (AbilitySub sub : choices) {
+                handleDependentModes(sa, chosen, sub);
+                if (isModeWorthAdding(aic, ai, sub) && canPayForAdditionalMode(sa, chosen, sub, ai)) {
```
tool4ever: why yet another loop, we've already checked all possible API variants above 👎

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
```diff
@@ -158,6 +173,64 @@ private List<AbilitySub> chooseOptionsAi(SpellAbility sa, List<AbilitySub> choic
...
+    }
+
+    private boolean isModeWorthAdding(AiController aic, Player ai, AbilitySub sub) {
+        sub.setActivatingPlayer(ai);
+        try {
+            if (AiPlayDecision.WillPlay == aic.canPlaySa(sub)) {
+                return true;
+            }
+        } catch (RuntimeException ex) {
+            // Some sub-mode AI assumes full spell timing context; fall back to trigger heuristics.
```
tool4ever: if it crashes here it's a bug:
let's not hide it this way...
besides these methods are used unguarded earlier anyway

Madwand99: Good point. Done!

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
Madwand99: This second pass is intentional because it is not repeating the same evaluation. The first pass uses canPlaySa() to find modes strong enough to make the spell playable. For optional extra modes, that is too strict: the AI may already be casting the spell, and a remaining mode only needs to be useful and payable as an add-on. That is why this pass uses isModeWorthAdding() instead. I added a comment to make that distinction clearer.

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
Madwand99: Whoops! My bad, fixed it.

Madwand99:
> Example of AI where the old code added extra Cost but shouldn't:
> 
> AI used `Lively Dirge`: `Return up to two creature cards with total mana value 4 or less from your graveyard to the battlefield.` while not having any creature in GY.
> 
> Because of that `up to`, AI thinks that it might be good enough.

I looked into this case. We could add a narrow guard in CharmAi before paying for an extra mode: for paid ModeCost modes that are hidden ChangeZone effects from a known zone like graveyard to battlefield, check whether there is at least one matching card before adding the mode.

That would stop the Lively Dirge case where the AI pays for the reanimation mode with no creatures in graveyard. However, after checking the current card scripts, this guard appears to only help Lively Dirge right now, so it feels a bit too card-shaped for this modal-spell PR.

I think the better long-term fix is in ChangeZoneAi or the hidden-origin ChangeZone evaluation: “up to” hidden moves should not be considered useful just because choosing zero is legal, especially when an extra cost is being paid. I’m inclined to leave that as a follow-up unless maintainers prefer the narrow modal guard here.

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
```diff
@@ -158,6 +175,48 @@ private List<AbilitySub> chooseOptionsAi(SpellAbility sa, List<AbilitySub> choic
         return chosen;
     }
 
+    private boolean canPayForAdditionalMode(SpellAbility sa, List<AbilitySub> chosen, AbilitySub sub, Player ai) {
+        Card source = sa.getHostCard();
+        if (source.hasStartOfKeyword("Spree") || source.hasStartOfKeyword("Tiered")) {
```
tool4ever: using this method just seems wrong:
we've already chained choices at this point, so why are we manually adding the raises - doesn't `CostAdjustment` still apply after so you're grossly miscalculating by doubling them? 🤔

Madwand99: Good point. It wasn't a miscalculation (our tests pass just fine) but it was structurally inefficient. Fixed!

forge-ai/src/main/java/forge/ai/ability/CharmAi.java
```diff
@@ -150,6 +152,19 @@ private List<AbilitySub> chooseOptionsAi(SpellAbility sa, List<AbilitySub> choic
                 }
             }
         }
+        if (!isTrigger && chosen.size() < num && min < num) {
+            // Optional extra modes can be worth adding even when canPlaySa() is too strict for a standalone mode.
+            choices.removeAll(chosen);
+            for (AbilitySub sub : choices) {
+                handleDependentModes(sa, chosen, sub);
+                if (aic.doTrigger(sub, false) && canPayForAdditionalMode(sa, chosen, sub, ai)) {
```
tool4ever: I think conceptually this should evaluate via `chkDrawbackWithSubs` instead and only when choices contains at least one SA found as main mode already

Madwand99: Good point, thanks. I updated this so the optional-mode pass only runs after the AI has already found at least one normal playable mode, so it won’t make a spell playable solely from fallback extra-mode logic.

I also changed the extra-mode evaluation from doTrigger(sub, false) to chkDrawbackWithSubs(...). That seems like a better fit here because these modes are being evaluated as additional payload on an already-playable modal spell, not as standalone trigger/no-cost abilities.

The cost check now temporarily chains the candidate modes and uses the normal ComputerUtilCost.canPayCost path, so Escalate/Spree/Tiered costs are handled by CostAdjustment instead of being manually reconstructed in CharmAi.

tool4ever (approved):

PR #10573, opened 2026-05-01
Update CopySpellAbilityEffect.java

Akse1ot:
Allows CopySpellAbility with Defined$ TriggeredSpellAbility to copy the source spell/ability from BecomesTarget trigger stored as AbilityKey.SourceSA, which had not been done before.

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/CopySpellAbilityEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/CopySpellAbilityEffect.java
@@ -73,7 +73,7 @@ public class CopySpellAbilityEffect extends SpellAbilityEffect {
 
         List<SpellAbility> tgtSpells = getTargetSpells(sa);
 
-        if (tgtSpells.isEmpty() && "TriggeredSpellAbility".equals(sa.getParam("Defined"))) {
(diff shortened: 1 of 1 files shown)


Akse1ot:
Not currently required by original cards, but may be useful in the future and for custom scripts.

tool4ever (changes requested):
nah, that doesn't make sense
you could just write "TriggeredSourceSA" in that case

Akse1ot:
Got it, that makes sense. I’ll switch this to Defined$ TriggeredSourceSA instead of making TriggeredSpellAbility fall back to SourceSA.

tool4ever:
your code just isn't needed...
this is all part of the normal handling:
https://github.com/Card-Forge/forge/blob/78edb77ed2b3c403e509349cb6013f571078007c/forge-game/src/main/java/forge/game/ability/AbilityUtils.java#L1245

Akse1ot:
Ah, I see. I misunderstood this. If TriggeredSourceSA is already handled by getTargetSpells(), then no CopySpellAbilityEffect change is needed for this part. I’ll remove the special handling and just update the script to use Defined$ TriggeredSourceSA.

PR #10576, opened 2026-05-01
Improve AI draw-punisher avoidance heuristics

Madwand99:
This narrows the previous AI decision work (#10559) to a lightweight heuristic for avoiding dangerous draw effects into opposing draw-punishers.

Changes include:

- Adds a static heuristic for detecting risky self-draw plays without running a generic safety simulation.
- Detects draw-heavy effects such as wheel-style hand redraws and permanent-triggered draw engines.
- Detects opposing draw-punisher triggers by behavior, not card name, including damage/life-loss/token payoff triggers.
- Applies the heuristic to normal AI spell selection, simulation picker candidates, and optional free-cast decisions.
- Preserves a commander exception so the AI does not refuse to cast its own commander from the command zone.
- Fixes Timetwister AI logic matching to handle the card script’s casing.
- Adds regression coverage for Wheel of Fortune, Timetwister, Echo of Eons, Mindmoil, Teferi’s Puzzle Box, Arjun, Xyris, Nekusar, and Kederekt Parasite cases.


forge-ai/src/main/java/forge/ai/simulation/SpellAbilityPicker.java
```diff
@@ -334,6 +334,9 @@ private AiPlayDecision canPlayAndPayForSim(final SpellAbility sa) {
         if (shouldWaitForLater(sa)) {
             return AiPlayDecision.AnotherTime;
         }
+        if (ComputerUtilCard.shouldAvoidDrawPunisher(player, sa)) {
```
tool4ever: nah, the whole point of simulation is to not rely on such heuristics

forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
```diff
@@ -892,6 +894,205 @@ public static boolean canBeBlockedProfitably(final Player ai, Card attacker, boo
         return ComputerUtilCombat.attackerWouldBeDestroyed(ai, attacker, combat);
     }
 
+    public static boolean shouldAvoidDrawPunisher(final Player ai, final SpellAbility sa) {
```
tool4ever, line 898: hmn, I'll have to think on this:
we usually don't attempt guessing this sort of two level deep
(e.g. we have helpers for direct damage from etb/casting but this is more "I might draw from it" + "then I might lose life from drawing")

I'm not convinced that while at first this might look good for a few specific cases it doesn't end up causing just as much problems at other times, especially when it's hooked up this high in the chain 🤔

Madwand99: Makes sense. This heuristic is intentionally trying to catch cases that are one step beyond direct “this spell hurts me”: “this play causes me to draw a lot” plus “an opponent has an active draw punisher.” That is why it catches things like Wheel effects into Nekusar/Xyris, but also why it is riskier than the usual direct ETB/cast damage checks.

I think there are a few possible directions:

1. Keep the current high-level hook. This catches the broadest set of cases, including future draw-engine permanents like Mindmoil, Teferi’s Puzzle Box, and Arjun. The downside is exactly what you pointed out: false positives are expensive because this blocks the play before normal AI evaluation can consider context.
2. Move the heuristic into narrower ability logic, e.g. ChangeZoneAllAi / wheel-style effects only. This is much safer and easier to reason about, but it would no longer catch permanents like Mindmoil/Puzzle Box/Arjun being cast into Xyris/Nekusar/Kederekt.
3. Split the heuristic into narrower cases: immediate wheel-style draw effects in the relevant ability AI, plus a separate permanent-spell check only when the card being cast has its own large self-draw trigger. That would still catch Mindmoil/Puzzle Box/Arjun, but avoid using this as a broad “before any spell” gate.

Happy to go with any of these options.

forge-ai/src/main/java/forge/ai/simulation/SpellAbilityPicker.java
Madwand99: Done!

forge-ai/src/main/java/forge/ai/ability/ChangeZoneAllAi.java
```diff
@@ -68,7 +68,7 @@ protected AiAbilityDecision canPlay(Player ai, SpellAbility sa) {
         
         if ("LivingDeath".equals(aiLogic)) {
             return SpecialCardAi.LivingDeath.consider(ai, sa);
-        } else if ("Timetwister".equals(aiLogic)) {
+        } else if ("Timetwister".equalsIgnoreCase(aiLogic)) {
```
tool4ever: I already fixed the script instead

Madwand99: Cool, removed the IgnoreCase.

tool4ever (commented):
Alright, sorry for the delay but amount of PR all waiting for feedback is at a record high :)

Now the bad news:
I think your conceptual parent of this PR was actually the better approach since simulation is already the generic solution for anything that involves checking any play for indirect results of such "2+ action chains".

There'd be way too many related cases such that trying to write heuristics for them too just ends up with spaghetti code and more performance penalties piling upon each other...

Therefore the right direction should be a trifecta of simulation speed, stability and clear architecture that allows mixing both - and all that again in way smaller steps.

I'm optimistic that with some other AI-based PR a few bigger bottlenecks are already easier to target and will eventually land in some form to have that task and other engine parts benefit nicely from it.

Madwand99:
> Alright, sorry for the delay but amount of PR all waiting for feedback is at a record high :)
> 
> Now the bad news: I think your conceptual parent of this PR was actually the better approach since simulation is already the generic solution for anything that involves checking any play for indirect results of such "2+ action chains".

No worries, I also thought that simulation was the right approach, this 2nd version was just me trying to accommodate requirements. What do you suggest I do? Close this PR for now and wait? Reopen the 1st PR but improve on it somehow? Something else?

tool4ever:
yea feel free to reopen though it should be with a clearer strategy

PR #10579, opened 2026-05-02
Extend server URL dialog to mobile

MostCromulent:
Screenshot 2026-05-02 100412

#10445 added a multi-interface server URL dialog on desktop. This brings the same to mobile.

What's new:
- Lists every reachable address (External + each local interface) with its own Copy button, instead of the previous 2–3-button External / Local / Close dialog
- Auto-copies and stars the previously remembered URL so repeat hosts on the same network don't need to pick the same row each time
- Scrollable when many interfaces are present
- Re-openable from the Online menu (new "Server URL" entry, enabled while hosting)

Auto-shown right after hosting, same as before.

Shared data prep extracted into `NetConnectUtil` so the desktop and mobile dialogs can't drift on the star/remember logic.

---

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/screens/home/online/CSubmenuOnlineLobby.java
+++ b/forge-gui-desktop/src/main/java/forge/screens/home/online/CSubmenuOnlineLobby.java
@@ -5,9 +5,7 @@ import java.awt.Toolkit;
 import java.awt.datatransfer.StringSelection;
 import java.net.BindException;
 import java.util.ArrayList;
-import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 
 import javax.swing.JMenu;
@@ -20,7 +18,6 @@ import net.miginfocom.swing.MigLayout;
 
 import forge.gamemodes.net.ChatMessage;
 import forge.gamemodes.net.NetConnectUtil;
-import forge.gamemodes.net.server.FServerManager;
 import forge.gui.FNetOverlay;
 import forge.gui.FThreads;
 import forge.gui.SOverlayUtils;
@@ -29,10 +26,8 @@ import forge.gui.framework.EDocID;
 import forge.gui.framework.ICDoc;
 import forge.gui.util.SOptionPane;
 import forge.localinstance.properties.ForgeConstants;
-import forge.localinstance.properties.ForgeNetPreferences;
 import forge.menus.IMenuProvider;
 import forge.menus.MenuUtil;
-import forge.model.FModel;
 import forge.screens.home.CHomeUI;
 import forge.screens.home.CLobby;
 import forge.screens.home.VLobby;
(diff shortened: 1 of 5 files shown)


tool4ever (dismissed):

jamincollins:
## Findings
#### 1. ServerAddressList collectHostedServerAddresses uses netPrefs.getPrefInt for port each call
CollectHostedServerAddresses calls netPrefs.getPrefInt(ForgeNetPreferences.FNetPref.NET_PORT) directly and builds URLs from it. If port preference changes between construction of ServerAddressList and display, URLs remain consistent within a single call. No rule violation, but the method is not defensive against concurrent preference change. Test case:
```java
// Verify port used is current at collection time
ForgeNetPreferences prefs = FModel.getNetPreferences();
prefs.setPrefInt(ForgeNetPreferences.FNetPref.NET_PORT, 8080);
NetConnectUtil.ServerAddressList list = NetConnectUtil.collectHostedServerAddresses();
assert list.urls.get(0).endsWith(":8080");
prefs.setPrefInt(ForgeNetPreferences.FNetPref.NET_PORT, 9090);
// list should still reflect 8080, not 9090
assert list.urls.get(0).endsWith(":8080");
```
Impact: Low. Collection is atomic for a single invocation; no observable MTG rule impact.

#### 2. Mobile OnlineLobbyScreen host flow invokes ServerAddressesDialog on EDT after background host
OnlineLobbyScreen.activateHost runs NetConnectUtil.host in background, then FThreads.invokeInEdtLater calls OnlineScreen.Lobby.update() and ServerAddressesDialog.show(). The dialog shows after host completes but update() may run before server is fully ready. Potential race where ServerAddressesDialog shows with stale addresses. Test case:
```java
// Simulate rapid host then dialog
OnlineLobbyScreen screen = new OnlineLobbyScreen();
screen.activateHost(); // triggers background host
// Immediately check if dialog shows before FServerManager reports hosting
assert OnlineScreen.ServerUrl.item.isEnabled() == false;
```
Impact: Medium. Could show empty address list on first host.

#### 3. collectHostedServerAddresses assumes FServerManager.getAllLocalAddresses returns non-null map
Iterates over FServerManager.getAllLocalAddresses().entrySet() without null check. If implementation returns null, NullPointerException. Test case:
```java
// Mock FServerManager.getAllLocalAddresses() to return null
try {
    NetConnectUtil.collectHostedServerAddresses();
    fail("Expected NPE");
} catch (NullPointerException e) {
    // confirmed
}
```
Impact: High. Crash on hosts with misconfigured network stack.

#### 4. ServerAddressesDialog.show copies to clipboard unconditionally when starIndex >=0
GuiBase.getInterface().copyToClipboard(addresses.urls.get(addresses.starIndex)) runs before dialog is shown. If clipboard access fails or user denies permission, exception propagates and dialog never shows. No try/catch. Test case:
```java
GuiBase mock = mock(GuiBase.class);
doThrow(new RuntimeException("clipboard denied")).when(mock).copyToClipboard(any());
ServerAddressesDialog.show(); // throws
```
Impact: Medium. Dialog fails to open on clipboard restriction.

#### 5. Desktop showServerAddressesDialog assumes addresses.urls non-empty for starIndex fallback
starIndex computed as urls.isEmpty()? -1 : 0, then if starIndex >=0 copyToClipboard(addresses.urls.get(starIndex)). If urls empty, starIndex = -1, copy skipped. But panel building loops over addresses.urls.size() which is 0, so panel empty. No guard for empty list before creating FOptionPane. Could show empty dialog. Test case:
```java
ServerAddressList empty = new ServerAddressList(new ArrayList<>(), new ArrayList<>(), -1);
CSubmenuOnlineLobby.showServerAddressesDialogWith(empty);
// Expect dialog with empty panel, no crash
```
Impact: Low. UX issue, no crash.

#### 6. NetConnectUtil.copyHostedServerUrl deprecated logic removed but still present
copyHostedServerUrl remains in NetConnectUtil with old SOptionPane flow. It is no longer called from mobile/desktop after PR changes, but still compiles and could be invoked elsewhere. If called, it copies external or internal URL but does not update remembered URL. Inconsistent state with new rememberCopiedServerUrl. Test case:
```java
NetConnectUtil.copyHostedServerUrl(); // old path
String remembered = FModel.getNetPreferences().getPref(ForgeNetPreferences.FNetPref.NET_LAST_COPIED_URL);
// remembered remains unchanged
```
Impact: Medium. Backwards compatibility risk for external callers.

#### 7. Desktop showServerAddressesDialog no longer persists netPrefs.save on copy
Old code called netPrefs.setPref and netPrefs.save after copy. New code delegates to NetConnectUtil.rememberCopiedServerUrl which saves. However, desktop copy button sets holder[0].setVisible(false) before save completes (save is synchronous). Acceptable. No regression.

#### 8. OnlineMenu.ServerUrl item enabled only when FServerManager.getInstance().isHosting()
ServerUrl.item.setEnabled(FServerManager.getInstance() != null && FServerManager.getInstance().isHosting()). If server starts asynchronously, menu may be enabled before addresses are populated, leading to empty dialog. Test case:
```java
FServerManager.getInstance().startServer();
OnlineScreen.ServerUrl.update();
assert OnlineScreen.ServerUrl.item.isEnabled() == true;
// collectHostedServerAddresses may return empty until network interfaces ready
```
Impact: Medium. Race between server start and address discovery.

#### 9. ServerAddressesDialog AddressList layout computes height based on labels.length
computeFullHeight uses labels.length for rowHeight multiplication. If addresses.urls.size() changes between construction and layout, mismatch. Not synchronized. Potential layout overflow. Test case:
```java
// Simulate address list change after construction
AddressList list = new AddressList(addresses, holder, localizer);
// mutate addresses.urls externally
// layout may exceed bounds
```
Impact: Low. Unlikely in practice.

### Performance and security
#### 10. ServerAddressesDialog copies URL to clipboard without validation
URL string built from externalAddress + ":" + port. externalAddress from FServerManager.getExternalAddress() could be spoofed or contain malicious payload (e.g., newline). Clipboard copy is safe but displaying URL in FLabel may allow UI injection if address contains control characters. Test case:
```java
// Inject address with newline
FServerManager mock = mock(FServerManager.class);
when(mock.getExternalAddress()).thenReturn("192.168.1.1\nmalicious");
// Dialog shows label with newline, potentially breaking layout
```
Impact: Low. UI robustness.

## Verdict
Request Changes
The PR introduces mobile server URL dialog parity with desktop, but introduces race conditions around server readiness, lacks null safety for network address discovery, and does not guard clipboard failures. These need hardening before merge.

## Recommended Fixes
1. Add null guard for FServerManager.getAllLocalAddresses() and getExternalAddress() in collectHostedServerAddresses, return empty list if null.
2. Wrap clipboard copy in try/catch with fallback to show dialog without auto-copy.
3. Defer ServerAddressesDialog.show() until FServerManager reports isHosting() and addresses non-empty, or retry with delay.
4. Remove or deprecate NetConnectUtil.copyHostedServerUrl to avoid inconsistent remembered URL state.
5. Add unit tests for empty address list and clipboard denial scenarios.

---

tool4ever (approved):

PR #10581, opened 2026-05-02
Mobile chat polish and match-screen unread indicator

MostCromulent:
Two improvements to the mobile chat experience.

## Match Chat tab unread indicator

Screenshot 2026-05-02 103010

- Red badge with unread count appears on the Chat tab while the dropdown is closed
- Brief red flash on each new message draws the eye even before you read the badge
- Includes system messages (disconnects, version mismatches, etc.) so the user doesn't miss server notices
- Both clear when the user opens the Chat tab

## Chat bubble styling

Screenshot 2026-05-02 103037

- Sender is now identified on every bubble
- Message text on top, smaller meta line ("Sender · HH:MM:SS") below
- Server messages now render in a blue bubble with white text and identify "Server" as the sender, instead of looking identical to the host's own chat

Mobile-only — desktop chat is unchanged.

---

### DIFF
--- a/docs/Development/ownership.md
+++ b/docs/Development/ownership.md
@@ -5,9 +5,9 @@ Please also try to check back a bit after your PR gets merged in case it causes 
 ## DevOps
 | Concept | Owners | Ancestors | Example tasks |
 | - | - | - | - |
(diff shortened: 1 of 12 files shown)


tool4ever (approved):
good idea

PR #10584, opened 2026-05-02
Move booster pics to forge-extras

Eradev:


paulsnoops (dismissed):

paulsnoops:
@tehdiplomat downloads.cardforge.org doesn't exist anymore. Should we move all of the booster stuff to forge-extras?

tehdiplomat:
Yea it would be nice to move them to forge-extras.

paulsnoops:
@Eradev do you happen to have the old images for boosters, fatpacks etc.?

Eradev:
I'm pretty sure I have most of the boosters. Not sure about fat packs. I'll have to check at home.

Agetian:
I have a bunch of stuff old and new, if you tell me where to send them I can send the full package

paulsnoops:
> I have a bunch of stuff old and new, if you tell me where to send them I can send the full package

If you make a PR here I will approve: https://github.com/Card-Forge/forge-extras/tree/main/images

Edit: Or you should have merge access

Agetian:
Yeah, I can merge there no prob, the only issue is, the package is rather rough, someone would need to give it a once-over and maybe sort the unneeded stuff or replace some of the things that may look odd :/ I mean, it was basically collected from a lot of locations and may have varying quality to it, but it's the only collection I'm aware of that has all the booster art prepared for the latest sets, achievements, etc. :/ What do we do with it all? I can drop it to a cloud storage folder and we can take a look and sort through it maybe?

Eradev:
Dumped what I had in https://github.com/Card-Forge/forge-extras/pull/27

forge-gui/res/lists/booster-images.txt
```diff
@@ -1,548 +1,468 @@
...
+YMKM.png
+YNEO.png
+ZEN_1.jpg
+ZEN_2.jpg
+ZEN_3.jpg
+ZEN_4.jpg
+ZEN_5.jpg
+ZNR_1.png
+ZNR_2.png
+ZNR_3.png
```
tehdiplomat, line 468: Are we losing information here?

forge-gui/src/main/java/forge/util/ImageFetcher.java
```diff
@@ -107,18 +107,20 @@ public void fetchImage(final String imageKey, final Callback callback) {
...
             final String filename = imageKey.substring(ImageKeys.BOOSTER_PREFIX.length());
             // Look up the download URL from booster-images.txt
-            for (String line : FileUtil.readFile(ForgeConstants.IMAGE_LIST_QUEST_BOOSTERS_FILE)) {
-                if (line.endsWith("/" + filename)) {
-                    downloadUrls.add(line);
+            final List<String> boosterFileContent = FileUtil.readFile(ForgeConstants.IMAGE_LIST_QUEST_BOOSTERS_FILE);
+
+            for (String line : boosterFileContent) {
+                if (line.startsWith(filename)) {
+                    downloadUrls.add(ForgeConstants.GITHUB_ASSETS_BASE + "/images/boosters/" + filename);
```
tehdiplomat: I'm not sure i'm a huge fan of removing the option for linking out to a different site if we have a full URL. Maybe if its _just_ the filename we prepend our URL, otherwise we use the old logic?

forge-gui/res/lists/booster-images.txt
Eradev: I've removed some images that were overlapping for some reasons.

forge-gui/src/main/java/forge/util/ImageFetcher.java
Eradev: Is it needed to have some images hosted on a different server if the collaborators can simply add them in forge-extras though?

tool4ever: I don't think we can start hosting every users custom content for them

tehdiplomat: @Eradev Can we restore the ability to download from absolute urls if they start with `http`?

Eradev: Yes. I've been busy. I'll try to work on it this weekend.

tehdiplomat: Gotcha. Just wanted to make sure you weren't waiting for us

tehdiplomat (approved):

PR #10585, opened 2026-05-02
Player: store Land Played as LKI

### LINKED ISSUE #10465: Handle Land Played This Turn via List for MayPlay uses
For #4700 part about `incMayPlayTurn`

I want Player::getLandPlayedThisTurn to return a `List<SpellAbility>` (mostly LandAbility)
maybe stored in Game like `SpellCastThisTurn` (but Lands can only be played by active turn player?)

https://github.com/Card-Forge/forge/blob/226cc45ea857f9054daa72c9ae3d219d362fc5a6/forge-game/src/main/java/forge/game/player/Player.java#L2245-L2255

While checking this code, I found this problem:
https://github.com/Card-Forge/forge/blob/226cc45ea857f9054daa72c9ae3d219d362fc5a6/forge-ai/src/main/java/forge/ai/ability/DestroyAi.java#L418-L419
LandsPlayedLastTurn isn't really correct for multiplayer, because it doesn't look at TargetPlayers last turn.

If I store a copy of the LandAbility, I might implement this in java:
https://github.com/Card-Forge/forge/blob/a81bd0f4a07c8099000b0a28907d1f4709a22c85/forge-gui/res/cardsfolder/v/visions_of_phyrexia.txt#L11-L12
https://github.com/Card-Forge/forge/blob/a81bd0f4a07c8099000b0a28907d1f4709a22c85/forge-gui/res/cardsfolder/s/spider_man_2099.txt#L13-L14

Hanmac:
Closes #10465

Stores Lands Played as LKI list
Reduces the need for some StoreSVar stuff

its a SpellAbility List so this might be later used for MayPlayLimit


forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -1627,6 +1627,7 @@ public final void shuffle(final SpellAbility sa) {
     }
 
     public final Card playLand(final Card land, SpellAbility cause) {
+        Card lki = CardCopyService.getLKICopy(land);
```
tool4ever: some checks will fail if you store LKI that's facedown

forge-ai/src/main/java/forge/ai/GameState.java
```diff
@@ -59,8 +59,6 @@ static class PlayerState {
         private String counters = "";
         private String manaPool = "";
         private String persistentMana = "";
-        private int landsPlayed = 0;
```
tool4ever, line 62: this seems too essential to lose support for

Hanmac: It isn't used anywhere
The puzzle all use 0 value

forge-game/src/main/java/forge/game/player/Player.java
Hanmac: done

forge-ai/src/main/java/forge/ai/GameState.java
tool4ever: well it's not only about GameState since you're also removing it from GameSimulator and rollback:
so you're essentially degrading 3 different features for what - future refactor potential? 👎

Land MayPlay count works fine right now and the field could surely stay on the StaticAbility if the approach gets adjusted 🤷‍♂️

Hanmac: Other properties like sacrificed Or Discarded this Turn aren't part of Rollback either

tool4ever: yea but land plays remaining is one of the most important ones lol

Hanmac:
@tool4ever calculate MayPlayLimit now, this somewhat matters for `The Fourth Doctor`,
because its ability doesn't restrict to "your turns"
```
Once each turn, you may play a historic land or cast a historic spell from the top of your library. When you do, create a Food token. 
```
which could be a corner case that your opponent gains control of it doing your turns, and then the MayPlayLimit should apply for them extra.

tool4ever:
hmn, maybe? though I don't believe I ever saw a ruling which clarifies that 🤔
but still need to find a way to restore the lost functionality:
what about a static placeholder SA that `setLandsPlayedThisTurn` can use to avoid paying even more copy costs...?

forge-game/src/main/java/forge/game/staticability/StaticAbility.java
```diff
@@ -552,15 +551,12 @@ public Set<StaticAbilityLayer> getLayers() {
...
-        return mayPlayTurn;
-    }
-
-    public void incMayPlayTurn() {
-        this.mayPlayTurn++;
-    }
-
-    public void resetMayPlayTurn() {
-        this.mayPlayTurn = 0;
+        final Player controller = this.getHostCard().getController();
```
tool4ever: not good enough in case the permission affects another player, e.g. *Ian Malcolm*:
then your check would always fail leading to unlimited number of usages :/

Hanmac:
I could ask Judges to confirm it for "the Fourth Doctor"

Part of the MayPlay changes I want to go is the remove of Continuous Effects if able




> but still need to find a way to restore the lost functionality

Where exactly do you want to use it? The Puzzle all start at clean phase with no Land played

> what about a static placeholder SA that `setLandsPlayedThisTurn` can use to avoid paying even more copy costs...?

The GameSnapshot thing needs to be reworked for Player objects anyway, but that's not part of this issue

tool4ever:
- a) just because we *currently* don't ship any puzzles/gamestates that alter the land play count doesn't mean there are no users who have some custom content that's relying on it
- b) when undo feature is active letting AP play another land as soon as you cancel an ability will also lead to many bug reports
- c) GameSimulator will also attempt illegal plays - I'm honestly surprised tests don't fail already but it seems we don't have one that checks against that

Hanmac:
> * when undo feature is active

@tehdiplomat need to help me with that Undo feature, but that is not part of this MR

Hanmac:
> c) GameSimulator will also attempt illegal plays - I'm honestly surprised tests don't fail already but it seems we don't have one that checks against that

Both of them, GameCopier and GameSimulator should have copied Game Stack `thisTurnCast` anyway, but didn't

Edit: also, `thisTurnCast` should probably be a List of castSA LKI instead of Card LKI, then it can be easier filtered against players

PR #10588, opened 2026-05-02
Rotate forge.log on startup; consolidate log rotation and export

MostCromulent:
## Summary

Currently `ExceptionHandler.registerErrorHandling()` deletes `forge.log` at every startup unless the file is locked by another running instance. On Android — where exporting a crash log requires restarting the app — that means the crash log gets wiped before the user can export it.

This PR rotates the previous `forge.log` to `forge.<timestamp>.log` at startup so it survives the restart, and prunes rotated backups to the most recent N entries where N is `MAX_LOG_FILES` (renamed from `NET_MAX_LOG_FILES` and moved from `server.preferences` to `forge.preferences` since it now governs both `forge.log` rotation and network log cleanup). The cap applies per category: up to N `forge.*.log` backups AND N network log entries.

Follows discussion on #10577.

## Changes

- `ExceptionHandler.registerErrorHandling()`: replaced the slot-finding loop with a one-shot rename of `forge.log` to `forge.<timestamp>.log` (timestamp from the file's `lastModified`).
- New `forge.error.LogRotation` utility: prunes `forge.*.log` rotated backups to `MAX_LOG_FILES`, called from `FModel.initialize()` once preferences load.
- Renamed `ForgeNetPreferences.FNetPref.NET_MAX_LOG_FILES` → `ForgePreferences.FPref.MAX_LOG_FILES`. Same default (10). Now governs both `forge.log` rotation and `NetworkLogConfig.cleanupOldLogs()`.
- Mobile: replaced the separate "Export Game Log" and "Export Network Logs" menu entries with one "Export Logs" that collects both `forge.*.log` (user dir) and `network-debug-*.log` (networklogs dir) into a single zip.

## Alternative considered: tinylog rolling-file writer

The architecturally cleaner option is to migrate `forge.log` to a tinylog rolling-file writer, which has built-in rotation with `policies = startup` and `backups = N`, and to cross-reference the count via `${forge.maxLogFiles}` system-property substitution. Rejected for this PR because Forge has ~1,000 `System.out.println` and `e.printStackTrace()` call sites whose output currently lands in `forge.log` via the `MultiplexOutputStream` redirect — including `BugReporter.reportException()`'s `ex.printStackTrace()`, which is the exact path crash traces take into the log. Migrating to tinylog without a stdout/stderr-to-tinylog bridge `OutputStream` would silently lose all of that content, and the bridge itself is easy to misuse (feedback loop with the `console` writer if not wired carefully). This is a real future option, but only safe to pursue as part of a broader logging-architecture migration that converts those call sites to `Logger.x()` calls and verifies content preservation.

---


forge-gui/src/main/java/forge/error/ExceptionHandler.java
```diff
@@ -53,19 +53,19 @@ public static void registerErrorHandling() {
         //initialize log file
         File logFile = new File(ForgeConstants.LOG_FILE);
 
-        int i = 0;
-        while (logFile.exists() && !logFile.delete()) {
```
tool4ever, line 57: needs to be refactored differently:
most common reason I can't delete old one is because I'm spawning a second process
in that case we can't just rotate on launch and need a new filename instead, so both instances continue logging into their own file

MostCromulent: could just have each instance write a timestamped log from the start rather than beginning with forge.log and renaming it?

or would you prefer to keep generic forge.log naming convention for most recent log?

tool4ever: hmn hard too judge if that would be too confusing for some users :)
I'm leaning towards not having to teach all existing users they need to start sorting?

either way `getMenuItem_OpenLogFile` needs to be fixed to open the real assigned one

MostCromulent: See what you think of latest commit.

> Addressed the multi-instance concern. New behavior:
> 
> - One Forge instance at a time owns `forge.log` via an advisory lock on a sibling `forge.log.lock` sentinel (held for JVM lifetime, OS-released on crash).
> - Secondary instances detect the lock contention at startup and write to `forge.<ts>-<N>.log` instead, with the slot claimed atomically via `File.createNewFile()`.
> - "Help → Open Log File" now opens whichever file this JVM is actually writing to.
> 
> Filename conventions, since they affect what the pruner touches:
> 
> - `forge.<ts>.log` — normal rotated backup. Pruned by mtime when there are more than `MAX_LOG_FILES`.
> - `forge.<ts>.<N>.log` — same-second collision fallback for rotation (back-to-back restarts within one second). Also pruned.
> - `forge.<ts>-<N>.log` — per-process active log of a secondary instance. NOT pruned, because some other JVM may currently be writing to it.
> 
> The dot-vs-dash separator is what lets the pruner distinguish the collision-fallback rotated file from a still-active secondary's log.
> 
> Also tightened a few things while in here: collapsed the new `LogRotation` utility into `ExceptionHandler`, dropped two now-unused parameters from the mobile Export Logs flow, and wrapped log-stream close in a `try/finally` so lock cleanup always runs.

tool4ever: tbh not a fan of either these dot vs. dash semantics or introducing an additional file

can we keep it simpler?
- try to rename all existing Forge\<N\>.log here into timestamped archive filename format so their number can be reduced after preferences are available
- those that fail their rename are still in progress
- grab free number for this run (probably based on highest still remaining so order is kept)

tool4ever (approved):

PR #10589, opened 2026-05-02
Carpet of Flowers: use ResolvedLimit

### LINKED ISSUE #10478: Multiple Carpet of Flowers issue
When a player controls a cloned carpet of flowers, the abilities of each one will incorrectly operate as if they're from the same permanent and thus only one can be used. It is unclear if multiple unique Carpet of Flowers cards would reproduce this issue.

**To Reproduce**
Have a carpet of flowers and copy it. If the original carpet's ability was used in the first main phase, the cloned carpet will not trigger in the second main phase. If both carpets trigger in a phase, the first will resolve normally, the second will do nothing on resolution.

**Expected behavior**
Each Carpet should work independently. The text 'if you haven't added mana with this ability' should not apply to ALL Carpets, because each one is a unique ability.

Hanmac:
Closes #10478

### DIFF
--- a/forge-gui/res/cardsfolder/c/carpet_of_flowers.txt
+++ b/forge-gui/res/cardsfolder/c/carpet_of_flowers.txt
@@ -1,15 +1,10 @@
 Name:Carpet of Flowers
 ManaCost:G
 Types:Enchantment
-T:Mode$ Phase | Phase$ Main1,Main2 | ValidPlayer$ You | CheckSVar$ CarpetX | SVarCompare$ EQ0 | OptionalDecider$ You | TriggerZones$ Battlefield | Execute$ TrigMana | TriggerDescription$ At the beginning of each of your main phases, if you haven't added mana with this ability this turn, you may add X mana of any one color, where X is the number of Islands target opponent controls.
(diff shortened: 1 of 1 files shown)


forge-gui/res/cardsfolder/c/carpet_of_flowers.txt
```diff
@@ -1,15 +1,10 @@
 Name:Carpet of Flowers
 ManaCost:G
 Types:Enchantment
-T:Mode$ Phase | Phase$ Main1,Main2 | ValidPlayer$ You | CheckSVar$ CarpetX | SVarCompare$ EQ0 | OptionalDecider$ You | TriggerZones$ Battlefield | Execute$ TrigMana | TriggerDescription$ At the beginning of each of your main phases, if you haven't added mana with this ability this turn, you may add X mana of any one color, where X is the number of Islands target opponent controls.
+T:Mode$ Phase | Phase$ Main1,Main2 | ValidPlayer$ You | OptionalDecider$ You | TriggerZones$ Battlefield | Execute$ TrigMana | ResolvedLimit$ 1 | TriggerDescription$ At the beginning of each of your main phases, if you haven't added mana with this ability this turn, you may add X mana of any one color, where X is the number of Islands target opponent controls.
```
tool4ever, line 4: this does not work
the count always happens

tool4ever: or wait, maybe I'm mistaken 😊

tool4ever (approved):

PR #10591, opened 2026-05-03
Field tab card-count diff and flash on hidden tab change

MostCromulent:
## Summary

#10037 added a "(N new)" suffix to zone and hand tab labels. This extends the same feature to player field tabs, and adds a brief red pulse on any tab header (field/hand/zone) when its diff increases while the tab is hidden — useful in 3+ player games where opponent fields share a tabbed cell.

Both behaviors are gated by the existing `UI_ZONE_TAB_NEW_COUNT` preference (Layout > View > "New Card Count in Tabs"), which now has a descriptive tooltip. Diff and flash logic is consolidated into a new `TabDiffTracker` helper.

---


forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
```diff
@@ -243,6 +243,7 @@ private void refreshAllViews() {
                     f.updateZones();
                     f.updateManaPool();
                     f.getTabletop().update();
+                    f.updateTabLabel(f.getTabletop().getCardPanels().size());
```
tool4ever: this might need to be routed differently, at least it doesn't seem to work on its own

tool4ever:
Switching does not seem to reset:
image

tool4ever (approved):

PR #10599, opened 2026-05-04
Better auto-sell

Eradev:
* No longer checks auto-sell for basic lands and cards that can have multiple copies of in a deck.
* (Non-commander) Checks auto-sell for cards you have four or more copies of.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/util/RewardActor.java
+++ b/forge-gui-mobile/src/forge/adventure/util/RewardActor.java
@@ -44,6 +44,7 @@ import forge.assets.ImageCache;
 import forge.card.CardImageRenderer;
 import forge.card.CardRenderer;
 import forge.card.CardSplitType;
+import forge.deck.DeckFormat;
 import forge.deck.DeckSection;
 import forge.game.card.CardView;
 import forge.gui.GuiBase;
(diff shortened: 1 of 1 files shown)


forge-gui-mobile/src/forge/adventure/util/RewardActor.java
```diff
@@ -1000,10 +1001,17 @@ public void act(float delta) {
...
                 autoSell.setVisible(true);
 
-                if (AdventurePlayer.current().isCommanderMode()) {
-                    PaperCard pc = reward.getCard();
-                    if (pc != null) {
+                PaperCard pc = reward.getCard();
+                if (pc != null &&
+                    !pc.getRules().getType().isBasic() &&
+                    !DeckFormat.canHaveAnyNumberOf(pc)
+                ) {
```
Jetz72: `DeckFormat::getMaxCardCopies(PaperCard)` should have all the logic you need to determine this on a per-card basis, as long as you can get the current deck format. Dunno if there's a dedicated getter for that but it's apparently trivial enough I just baked it into the deck editor:
https://github.com/Card-Forge/forge/blob/830957a357362faf49826ad0ad4eba80f83c24f3/forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java#L54-L57

Jetz72: `getMaxCardCopies` should also be able to cover `isBasicLand`, since `canHaveAnyNumberOf` checks for it.

Jetz72 (approved):

PR #10600, opened 2026-05-04
Make game ending idempotent

IkeM-L:
Game ending wasn't idempotent so something causing a lot of game end events would end it multiple times  (e.g. over 1000 triggers being put on the stack ends the game in a draw once for each trigger over 1000 put on the stack)  and break the UI. With this fix, we exit early if the game is already over.

### DIFF
--- a/forge-game/src/main/java/forge/game/Game.java
+++ b/forge-game/src/main/java/forge/game/Game.java
@@ -555,6 +555,11 @@ public class Game {
     }
 
     public synchronized void setGameOver(GameEndReason reason) {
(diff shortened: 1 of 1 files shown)


IkeM-L:
Tested by building for android and running on my phone, adding [[everlasting torment]] and [[flourishing defences]] to the battlefield, then creating 100 tokens and casting [[chain reaction]]. This puts 100,000 triggers on the stack and now correctly ends the game once.

tool4ever (approved):
thanks

PR #10601, opened 2026-05-04
Fix life swap with no life change and side effects

IkeM-L:
Life swap effects didn't happen if life total didn't change, which results in the wrong game state for cards like [[tree of perdition]] swapping power/toughness with a player who had life equal to their modified toughness, but which still needed their base power/toughness to be updated.

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/LifeExchangeVariantEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/LifeExchangeVariantEffect.java
@@ -70,10 +70,12 @@ public class LifeExchangeVariantEffect extends SpellAbilityEffect {
             return;
         }
 
-
-        if (pLife > num && p.canLoseLife()) {
+        // Likewise, we still need to do this even if the life total doesn't change
+        if (pLife >= num && p.canLoseLife()) {
             final int diff = pLife - num;
             final int lost = p.loseLife(diff, false, false);
+            source.addNewPT(power, toughness, timestamp, 0);
+            game.fireEvent(new GameEventCardStatsChanged(source));
 
             if (lost > 0) { // Run triggers if player actually lost life
                 final Map<Player, Integer> lossMap = Maps.newHashMap();
@@ -84,13 +86,11 @@ public class LifeExchangeVariantEffect extends SpellAbilityEffect {
         } else if (num > pLife && p.canGainLife()) {
             final int diff = num - pLife;
             p.gainLife(diff, source, sa);
+            source.addNewPT(power, toughness, timestamp, 0);
+            game.fireEvent(new GameEventCardStatsChanged(source));
         } else {
             // do nothing if they cannot lose/gain this amount of life
         }
-
-        // 701.12g: We need to do this regardless of whether the life total actually changed
-        source.addNewPT(power, toughness, timestamp, 0);
-        game.fireEvent(new GameEventCardStatsChanged(source));
     }
 
 }
--- a/forge-game/src/main/java/forge/game/ability/effects/LifeExchangeVariantEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/LifeExchangeVariantEffect.java
@@ -70,8 +70,7 @@ public class LifeExchangeVariantEffect extends SpellAbilityEffect {
             return;
         }
 
-        // Likewise, we still need to do this even if the life total doesn't change
-        if (pLife >= num && p.canLoseLife()) {
+        if (pLife > num && p.canLoseLife()) {
             final int diff = pLife - num;
             final int lost = p.loseLife(diff, false, false);
             source.addNewPT(power, toughness, timestamp, 0);
@@ -88,7 +87,13 @@ public class LifeExchangeVariantEffect extends SpellAbilityEffect {
             p.gainLife(diff, source, sa);
             source.addNewPT(power, toughness, timestamp, 0);
             game.fireEvent(new GameEventCardStatsChanged(source));
-        } else {
+        }
+        // We still need to do this if the life total doesn't change because the creature's base power/toughness might
+        else if(pLife == num){
+            source.addNewPT(power, toughness, timestamp, 0);
+            game.fireEvent(new GameEventCardStatsChanged(source));
+        }
+        else {
             // do nothing if they cannot lose/gain this amount of life
         }
     }
--- a/forge-game/src/main/java/forge/game/ability/effects/LifeExchangeVariantEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/LifeExchangeVariantEffect.java
@@ -62,9 +62,9 @@ public class LifeExchangeVariantEffect extends SpellAbilityEffect {
         } else if ("Toughness".equals(mode)) {
             num = source.getNetToughness();
             toughness = pLife;
+        } else {
+            return;
         }
-        // We cannot safely return early here as no net change in life total might
-        // not mean no net change in power or toughness if they are modified.
 
         if (!source.isInPlay()) {
             return;
--- a/forge-game/src/main/java/forge/game/ability/effects/LifeExchangeVariantEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/LifeExchangeVariantEffect.java
@@ -70,12 +70,10 @@ public class LifeExchangeVariantEffect extends SpellAbilityEffect {
             return;
         }
 
-        // Likewise, we still need to do this even if the life total doesn't change
-        if (pLife >= num && p.canLoseLife()) {
+
+        if (pLife > num && p.canLoseLife()) {
             final int diff = pLife - num;
             final int lost = p.loseLife(diff, false, false);
-            source.addNewPT(power, toughness, timestamp, 0);
-            game.fireEvent(new GameEventCardStatsChanged(source));
 
             if (lost > 0) { // Run triggers if player actually lost life
                 final Map<Player, Integer> lossMap = Maps.newHashMap();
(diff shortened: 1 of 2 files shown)


IkeM-L:
Tested by building for android, then adding tree of perdition to battlefield, putting 12 - 1/-1 counters on it, setting opponent's life total to one, activating tree, then checking if the tree dies.

forge-game/src/main/java/forge/game/ability/effects/LifeExchangeVariantEffect.java
```diff
@@ -62,15 +62,16 @@ public void resolve(SpellAbility sa) {
...
+        // We cannot safely return early here as no net change in life total might
+        // not mean no net change in power or toughness if they are modified.
 
         if (!source.isInPlay()) {
             return;
         }
 
-        if (pLife > num && p.canLoseLife()) {
+        // Likewise, we still need to do this even if the life total doesn't change
+        if (p.canLoseLife()) {
```
tool4ever: ??
now player will always lose life, even if is his life is the lower value already...

IkeM-L: I assumed that was already guarded against by the loseLife function and the line directly below?

```
 // Rule 118.4
// this is for players being able to pay 0 life nothing to do
// no trigger for lost no life
if (toLose <= 0 || !canLoseLife()) {
    return 0;
}
```

```
if (lost > 0) { // Run triggers if player actually lost life
```

I'm also struggling to come with a scenario where this could happen, I'm happy to re-create and test if you can.

Jetz72: > I'm also struggling to come with a scenario where this could happen, I'm happy to re-create and test if you can.

If the player had less life than Tree of Perdition's toughness? In which case it should register as gaining life, and fall into the `canGainLife` branch? As it stands now, it looks like no exchange can register as gaining life so long as the player is capable of losing life.

IkeM-L: Ah I see what you mean, I think I have a fix but I need to check it works

tool4ever: imo this whole change is wrong anyway
a player can't gain or lose 0 life, so no exchange should happen

Jetz72: Perhaps instead of changing the operator so the case where `pLife == num` falls into the lose-life branch, just extract the shared `addNewPT` line outside the if-blocks?

> imo th a player can't gain or lose 0 life, so no exchange should happen

I believe the case brought up in Discord was when Tree of Perdition had counters altering its toughness to be equal to the player's life total when it otherwise would not be.

e.g. Player with 20 life, tree with 13 base toughness and 7 +1/+1 counters. If I understand correctly, player health shouldn't change but the tree should be at 27 after the swap.

IkeM-L: I considered that, but wasn't sure what the intended behavior was if the player can't gain/lose life and the tree's modified toughness is the same as their current life total.

Jetz72: Yeah looks like you're right about that:

> 119.7. If an effect says that a player can’t gain life, that player can’t make an exchange such that the player’s life total would become higher; in that case, the exchange won’t happen. Similarly, [...]

In that case, guess the exchange is prevented if the player would lose life but they can't, or the player would gain life but they can't. But if the player's life wouldn't change, the new PT should still be set without regard for any restrictions on their life total changing?

IkeM-L: I have asked a judge at my FLGS and they confirmed this interpretation.

IkeM-L: Update: The judge now thinks he was wrong and 701.12g should make it so the tree's toughness changes regardless of whether the player's life total can.

Jetz72: Relevant rules:

> 701.12g A spell or ability may instruct a player to exchange two numerical values. In such an exchange, each value becomes equal to the previous value of the other. If either of those values is a life total, the affected player gains or loses the amount of life necessary to equal the other value. Replacement effects may modify this gain or loss, and triggered abilities may trigger on it. A player who can’t gain life can’t be given a higher life total this way, and a player who can’t lose life can’t be given a lower life total this way **(see rules 119.7–8)**. If either of those values is a power or toughness, a continuous effect is created setting that power or toughness to the other value (see rule 613.4b). This rule does not apply to spells and abilities that switch a creature’s power and toughness.

> 119.7. If an effect says that a player can’t gain life, that player can’t make an exchange such that the player’s life total would become higher; in that case, the exchange won’t happen. Similarly, if an effect redistributes life totals, a player can’t receive a new life total such that the player’s life total would become higher. In addition, a cost that involves having that player gain life can’t be paid, and a replacement effect that would replace a life gain event affecting that player won’t do anything.

> 119.8. If an effect says that a player can’t lose life, that player can’t make an exchange such that the player’s life total would become lower; in that case, the exchange won’t happen. Similarly, if an effect redistributes life totals, a player can’t receive a new life total such that the player’s life total would become lower. In addition, a cost that involves having that player pay life can’t be paid.

> 701.12a A spell or ability may instruct players to exchange something (for example, life totals or control of two permanents) as part of its resolution. When such a spell or ability resolves, if the entire exchange can’t be completed, no part of the exchange occurs.

701.12g describes the process and notes that restrictions on gaining/losing life prevent an exchange from respectively increasing or decreasing their life. But for what happens when such an exchange is attempted, it directs us to 119.7 and 119.8. Those two address this interaction much more specifically and are consistent with 701.12a.

IkeM-L: You're right. This leaves me unsure what the correct behaviour is if the player can't lose life but can gain life (or the inverse) and the exchange would not change their life total.

Jetz72: I think if their total would not change, neither restrictions on gaining nor losing life would matter. As I'd interpret it, not even "Your life total can't change" (e.g. Platinum Emperion) would prevent the player's life total from being "exchanged" with an identical number. It wouldn't register as a gain or loss in life, but you'd get all the side effects on the other end of the exchange.

IkeM-L: I think that makes the current state of the PR correct, but I don't know if I should seek out more judges to make sure!

Jetz72: Looks like in its current state if the player can't lose life and an exchange is attempted with equal life and toughness, it'll miss every branch and not trigger the side effects. Think you just need a third branch to handle the case where `pLife == num` that doesn't check the player's life restrictions and only fires the `addNewPT`.

IkeM-L: I was hoping for a more elegant solution, but you're right, this does just need a special case.

Jetz72: Looks good to me. I'll step back and give some space for other devs to weigh in on the rules or implementation though. This was originally @tool4ever's review comment before I jumped in and hijacked it, after all.

IkeM-L: Thank you so much for your time!

forge-game/src/main/java/forge/game/ability/effects/LifeExchangeVariantEffect.java
```diff
@@ -62,9 +62,9 @@ public void resolve(SpellAbility sa) {
         } else if ("Toughness".equals(mode)) {
             num = source.getNetToughness();
             toughness = pLife;
-        } else {
-            return;
         }
```
Hanmac, line 67: @tool4ever i think this return should be allowed if there is wrong mode

Or should it throw an Exception?

tool4ever: imo an Exception for scripting errors is a bit too extreme

this part didn't need to be changed anyway

Hanmac (approved):

forge-game/src/main/java/forge/game/ability/effects/LifeExchangeVariantEffect.java
tool4ever: I'm still a bit skeptical but should be enough research on this corner case for now

Hanmac:
@tool4ever i got an idea to reduce the amount of duplicated code:

```java
if (pLife > num && !p.canLoseLife()) {
    return;
}
if (num > pLife && !p.canGainLife()) {
    return;
}
int lost = 0;
if (pLife > num) {
    lost = p.loseLife(pLife - num, false, false);
} else if (num > pLife) {
    p.gainLife(num - pLife, source, sa);
}
source.addNewPT(power, toughness, timestamp, 0);
game.fireEvent(new GameEventCardStatsChanged(source));
if (lost > 0) { // Run triggers if player actually lost life
    final Map<Player, Integer> lossMap = Maps.newHashMap();
    lossMap.put(p, lost);
    final Map<AbilityKey, Object> runParams = AbilityKey.mapFromPIMap(lossMap);
    source.getGame().getTriggerHandler().runTrigger(TriggerType.LifeLostAll, runParams, false);
}
```

tool4ever:
well running static checks twice isn't good either, so you'd need a helper for avoiding that too
not sure if that will end up looking cleaner 🤷‍♂️

Hanmac:
> well running static checks twice isn't good either, so you'd need a helper for avoiding that too not sure if that will end up looking cleaner 🤷‍♂️

You were right, the second static check can be removed

PR #10602, opened 2026-05-04
Adventure: consult AdventureOverrides for sealed event boosters

vanja-ivancevic:
- `AdventureEventData.generateSealedPool` fetched booster templates directly from `StaticData.instance().getBoosters().get(setCode)`, bypassing the `AdventureOverrides` registry.
- Drafts and reward packs already use the override-first lookup via `AdventureEventController.generateBooster`. Sealed events did not, so plane-scoped overlays (e.g. Old Border Shandalar's 14+1 Ice Age booster) were silently ignored in sealed pools.
- Mirror the same override-first lookup pattern in both the human and AI pool branches of `generateSealedPool`.
- Verified locally that the override booster template for ICE produces 15-card packs with one snow-covered basic land per pack across 200 sample packs, drawing all five colors.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/data/AdventureEventData.java
+++ b/forge-gui-mobile/src/forge/adventure/data/AdventureEventData.java
@@ -3,7 +3,6 @@ package forge.adventure.data;
 import com.badlogic.gdx.scenes.scene2d.ui.Image;
 import com.badlogic.gdx.utils.Array;
 import forge.Forge;
-import forge.StaticData;
 import forge.adventure.character.EnemySprite;
 import forge.adventure.pointofintrest.PointOfInterestChanges;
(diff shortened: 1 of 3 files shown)


Jetz72:
Looking over your `AdventureOverrides`, looks like `getBoosterTemplate` is the only one of the methods there that doesn't automatically fall back to the base game's data if not overridden. Is it worth folding those `if (template == null)` branches into the method?

vanja-ivancevic:
Good catch, that's the cleaner shape. Folded the fallback into `AdventureOverrides.getBoosterTemplate` so it now matches `getBlock`/`allBlocks`. Each callsite drops to a single lookup, and the three duplicated `if (template == null)` blocks are gone.

Jetz72 (approved):

PR #10606, opened 2026-05-04
Yield Rework V2 — configurable interrupts, APINA, suggestions, settings UI

MostCromulent:
Refactor of #9643 (Yield Rework v1) on top of the per-PCH `YieldController` MVP shipped in #10555. 

Re-implements the deferred YieldRework feature set against the MVP's architecture (single `YieldUpdate` wire envelope, per-PCH controller, atomic seed) without adding new ProtocolMethod entries.

**[Architectural reference doc attached separately covers the design in detail.](https://github.com/user-attachments/files/27379620/YieldReworkV2_ArchitecturalOutline.md)**

## Deferred features included

- **Configurable interrupt system** — six interrupt prefs governing when an active autopass cancels: `OPPONENT_SPELL`, `ATTACKERS`, `TARGETING`, `MASS_REMOVAL`, `TRIGGERS`, `REVEAL`. First two default ON to preserve master's existing hardcoded interrupts; the rest default OFF.
- **APINA (Auto-Pass If No Actions)** — persistent toggle that auto-passes priority whenever no playable actions exist. Backed by a new `forge.ai.AvailableActions` heuristic with a per-PCH timeout budget. Separate `RESPECTS_INTERRUPTS` opt-in (default OFF).
- **Smart suggestions** — `InputPassPriority` offers Accept/Decline prompts when conditions match: stack-yield when there's no playable response to the stack, marker-on-upkeep when no actions exist on your turn. Each suggestion type has a configurable decline scope (`NEVER`/`ALWAYS`/`STACK`/`TURN`).
- **Speed settings** — opt-in skip of the inter-phase pause and the post-resolve pause during sustained auto-passes.
- **Per-PCH preference sync** — every yield FPref is overridable per-controller, propagated via two new `YieldUpdate` envelope variants. Atomic seed at game start carries the full overlay map in a single message.
- **Settings UIs** — desktop `VYieldSettings` dialog + `VYield` dock-tab panel (Auto-Pass toggle button + settings shortcut), mobile `VYieldOptions` scrollable dialog. Game-menu entries on both platforms. Three configurable desktop shortcuts (Ctrl+Y / F2 / ESC).

Hardcoded `autoPassCancel()` sites in `MagicStack`, `PCH.declareAttackers`, and `HostedMatch` deleted — the event-driven handler covers them, governed by per-PCH prefs. Default-ON pref values keep observable behavior identical for unconfigured users.

---


autumnmyst:
I notice you still have something checking state despite the yield interrupts. I don't think this is necessary and will just cause the state-scanner to re-prompt on every priority pass while the interrupt state persists. `shouldInterruptYield(player)` shouldn't need to check the state of the game to decide whether to yield, since we have event driven interrupts now. Also, should make the stack yield respect interrupts, so if things are added that should interrupt, they indeed interrupt. I don't think `isInterruptibleYieldActive()` is necessary, all yields should be interruptible.

autumnmyst:
Also, I recall we talked about removing the smart suggestion prompts? Are they actually useful enough to justify as a feature?

MostCromulent:
@autumnmyst 

> I notice you still have something checking state despite the yield interrupts. I don't think this is necessary and will just cause the state-scanner to re-prompt on every priority pass while the interrupt state persists. 

this was for APINA, have updated so interrupts apply on event driven basis there too. much cleaner architecture now

> Also, should make the stack yield respect interrupts, so if things are added that should interrupt, they indeed interrupt. I don't think isInterruptibleYieldActive() is necessary, all yields should be interruptible.

i didn't think we _want_ stack yield to be interrupted? i.e. when you click 'yield to entire stack' you're saying "I don't want to interact with this stack in any way / I have no instants I can play, don't bother me again".

e.g. you're in a commander game and your opponents are interacting on the stack but you don't want/cant participate. if stack yield interrupts you get bothered every time opponents interact when you've already opted out.

if we really we want this i'd suggest we add a 'Stack yield respects interrupt' option to the settings, same as the APINA respects, so at least its optional

> Also, I recall we talked about removing the smart suggestion prompts? Are they actually useful enough to justify as a feature?

in discussion on Discord, TRTs view was worth keeping because its useful for players who don't want full automation and want to use Forge to eg. train for playing paper games

autumnmyst:
I think we still want stack interrupts, since yielding the stack is saying “I don’t want to interact with the stack in its current state” but you yield and then say, a mass removal spell is added, you might want to reconsider. You can always just turn off interrupts too. I don’t really see stack yield as “special” in that it should ignore interruption.

Example: play a spell, 10 prowess triggers appear, you have a counterspell but you don’t want to counter your own spell, but that means it doesn’t autopass. You don’t want to click through each trigger, so you yield the stack. The next player  plays their own counterspell on your spell. You had targeting interruptions on, but it was ignored and you never had a chance to respond to them countering your spell.

forge-gui/src/main/java/forge/gamemodes/match/YieldUpdate.java
```diff
@@ -16,19 +17,24 @@ public sealed interface YieldUpdate extends Serializable
         permits YieldUpdate.SetMarker,
                 YieldUpdate.ClearMarker,
                 YieldUpdate.StackYield,
+                YieldUpdate.SetAutoPassUntilEndOfTurn,
```
tool4ever, line 20: can't this also be used instead of `ProtocolMethod.passPriorityUntilEndOfTurn`?

forge-gui/src/main/java/forge/gamemodes/match/YieldController.java
```diff
@@ -301,5 +365,259 @@ public void applyClientSeed(YieldStateSnapshot snap) {
         localStore.setTriggerDecisionsDisabled(snap.autoTriggersDisabled());
         skipPhases.clear();
         skipPhases.putAll(snap.skipPhases());
+        boolPrefOverrides.clear();
+        if (snap.boolPrefOverrides() != null) boolPrefOverrides.putAll(snap.boolPrefOverrides());
```
tool4ever: won't these be empty on game start?
so the host applies its own at first - that's gonna be a bit confusing

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -3533,33 +3550,55 @@ public boolean isUiSetToSkipPhase(final PlayerView turnPlayer, final PhaseType p
...
+        if (yieldController.apply(update)) {
+            // Refresh prompt so cancel button reflects "Yielding until X" instead of stale "End Turn".
             getGui().updateAutoPassPrompt();
         }
+        tryAutoPassNow();
+    }
+
+    @Override
+    public void setYieldBoolPref(final FPref pref, final boolean value) {
+        yieldController.setBoolPref(pref, value);
```
tool4ever: there's no reason these two local paths need to store?

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -3571,7 +3555,6 @@ public void setYieldStringPref(final FPref pref, final String value) {
      *  Same compute gating as {@link #chooseSpellAbilityToPlay} so the actions field is fresh. */
     private void tryAutoPassNow() {
         if (!(inputQueue.getInput() instanceof InputPassPriority)) return;
-        if (yieldController.isYieldActive()) return;
```
tool4ever: wouldn't it been better to move the check into branch below, so the calculation can be skipped if we're yielding anyway?

docs/advanced-yield-options.md
```diff
@@ -0,0 +1,137 @@
+## Contents
+- [How to Access](#how-to-access)
```
tool4ever, line 2: - still needs sidebar entry
- and the Auto-Yield section in user guide needs some alignment together with the new persistence states
- your call if/how you combine it

MostCromulent: See https://github.com/Card-Forge/forge/pull/10606/commits/9fc503439bceda7445ed7ff9e04787bc4154b955: I went with a short summary on the User Guide page + a link to the more detailed Advanced Yield Options if user wants more info. Let me know if you want any changes to that.

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -3533,33 +3533,54 @@ public boolean isUiSetToSkipPhase(final PlayerView turnPlayer, final PhaseType p
...
+            // Refresh prompt so cancel button reflects "Yielding until X" instead of stale "End Turn".
             getGui().updateAutoPassPrompt();
         }
+        tryAutoPassNow();
+    }
+
+    @Override
+    public void setYieldBoolPref(final FPref pref, final boolean value) {
+        // Dialog already wrote to FModel; APINA toggles may flip mayAutoPass
+        tryAutoPassNow();
```
tool4ever: isn't checking for `YIELD_AUTO_PASS_NO_ACTIONS` enough?

I think it would also reduce bloat a bit if we remove the String/Boolean distinction since internally they're always stored as Strings anyway...

so we should still be able to send them as Objects directly over the wire and only use `String.valueOf` once it reached host? (if really worth it)

forge-gui-desktop/src/main/java/forge/screens/match/views/VPrompt.java
```diff
@@ -74,8 +78,21 @@ public void setCardView(final CardView card) {
...
+            PlayerView local = matchUI == null ? null : matchUI.getCurrentPlayer();
+            if (ctrl != null && local != null && ctrl.getYieldController() != null
+                    && ctrl.getYieldController().isYieldActive()) {
+                ctrl.sendYieldUpdate(new YieldUpdate.ClearMarker(local));
+                ctrl.sendYieldUpdate(new YieldUpdate.StackYield(local, false, false));
+                ctrl.sendYieldUpdate(new YieldUpdate.SetAutoPassUntilEndOfTurn(local, false));
+                return;
+            }
+
+            if (btnCancel.isEnabled()
```
tool4ever: why are we now doing stuff when the button isn't enabled...
also didn't we change it in the MVP to just let the Input handle it?

forge-gui-desktop/src/main/java/forge/control/KeyboardShortcuts.java
```diff
@@ -289,12 +287,10 @@ public void actionPerformed(final ActionEvent e) {
...
                 IGameController ctrl = matchUI.getGameController();
-                PlayerView local = matchUI.getCurrentPlayer();
-                if (ctrl == null || local == null || ctrl.getYieldController() == null) { return; }
+                if (ctrl == null || ctrl.getYieldController() == null) { return; }
                 if (!ctrl.getYieldController().isYieldActive()) { return; }
-                ctrl.sendYieldUpdate(new YieldUpdate.ClearMarker(local));
-                ctrl.sendYieldUpdate(new YieldUpdate.StackYield(local, false, false));
-                ctrl.sendYieldUpdate(new YieldUpdate.SetAutoPassUntilEndOfTurn(local, false));
+                // InputLockUI.selectButtonCancel calls clearActiveYieldAndDispatch as a side effect.
+                ctrl.selectButtonCancel();
```
tool4ever: I suppose there might be situations where the user has to cancel yield during a mana payment 🤔
otherwise if he wants to hold priority after and he's not fast enough game will move on?

and the way this shortcut is assigned now doesn't really serve a purpose in its default form - and even if it gets reassigned it'd just be a weird ESC alternative...

I wonder if we should just merge it with the APINA toggle one?
so that if yields are active it disables those first too

might actually make more sense to have a sort of "stop any yield form" key available if user confuses which one he used

MostCromulent: Yeah I think issue is we're trying to overload the ESC button too much. Let me know what you think of latest commit.
-ESC keeps its master functionality; it will cancel an end-of-turn yield in same circumstances cancel button would today but there's no special yield-specific logic
-F2 (re-assignable) becomes a "panic button" to disable all yield states + disable APINA if its on.
-F3 (re-assignable) becomes the an APINA on/off toggle.

tool4ever: hmn I thought we might still only use a single key:
either disable yield/APINA or turn the latter on

but argueably not 100% symmetrical so if there are situations where you'd want to turn it on while already yielding...? 🤔

might be nitpicking but it's also a bit weird we're now introducing F-Keys to the configurable defaults
especially when the related "End Turn" is at Ctrl+E

forge-gui/res/languages/en-US.properties
```diff
@@ -1626,7 +1626,46 @@ lblWaitingForOpponent=Waiting for opponent...
 lblYieldingUntilEndOfTurn=Yielding until end of turn.\nYou may cancel this yield to take an action.
 lblYieldingUntilPhaseFmt=Yielding until {0}''s {1}.\nYou may cancel this yield to take an action.
 lblYieldingUntilStackClears=Yielding until stack clears.\nYou may cancel this yield to take an action.
-lblYieldToEntireStack=Yield to entire stack
+lblYieldToStack=Yield to stack
```
tool4ever, line 1625: `lblSHORTCUT_ENDTURN` needs update?

forge-gui-desktop/src/main/java/forge/control/KeyboardShortcuts.java
MostCromulent: sure, situations where you'd want to turn it on while yielding are probably niche - I'll combine and we can see what user feedback from testing says. 

I'll check other shortcuts for conflict in a bit but maybe something like Y for toggle, CTRL-Y to access options dialog.

forge-gui/res/languages/en-US.properties
```diff
@@ -517,7 +517,7 @@ lblSHORTCUT_SHOWCONSOLE=Match: show console panel
 lblSHORTCUT_SHOWDEV=Match: show dev panel
 lblSHORTCUT_UNDO=Match: undo last action
 lblSHORTCUT_CONCEDE=Match: concede game
-lblSHORTCUT_ENDTURN=Match: pass priority until EOT or next stack event
+lblSHORTCUT_ENDTURN=Match: pass priority until end of turn or interrupt
```
tool4ever, line 518: can't AI also update other languages when touching existing keys?

skipping new ones is acceptable since they should be easier for translators to identify (though it's been a while I've seen most actively updated)

MostCromulent: do you want me to delete the keys in other languages so they fall back to default, or try and get AI to update the text in those languages? all i could do to check that personally is really run it through google translate.

tool4ever: I feel like deleting here would make it look weird standing out given that it's part of a bigger group

just thought it'd be pretty easy for AI, even if native speaker might word it differently 🤷‍♂️

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1494,18 +1500,30 @@ public List<SpellAbility> chooseSpellAbilityToPlay() {
...
+        // Skip when already yielding — yield proceeds regardless of available-actions.
+        // Yield check first: it's a field read, vs needsAvailableActions which does 3 synced FPref reads.
+        if (!yieldController.isYieldActive() && needsAvailableActions()) {
+            long timeoutMs = computeAvailableActionsBudgetMs(getPlayer());
+            getPlayer().getView().setHasAvailableActions(AvailableActions.compute(getPlayer(), timeoutMs));
+        }
+
+        // Game-thread only; no cross-thread races on the YIELD_SUPPRESS_AFTER_END flag
+        boolean nowMayAutoPass = mayAutoPass();
+        yieldController.noteMayAutoPassResult(nowMayAutoPass);
```
tool4ever, line 1535: why is this method synchronized if this is the only place it gets called?

forge-gui/src/main/java/forge/gamemodes/match/YieldController.java
```diff
@@ -207,8 +262,23 @@ public void clearAutoYields() {
             localStore.clear();
             return;
         }
-        boolean matchOver = owner.getGame() == null || owner.getGame().getView().isMatchOver();
-        activeStore().onGameEnd(matchOver);
+        activeStore().onGameEnd(owner.getGame() == null || owner.getGame().getView().isMatchOver());
+    }
+
+    /** Clear all transient yield state so it doesn't carry into the next game of the match. */
+    public synchronized void resetForNewGame() {
```
tool4ever: I feel like AI got confused by the refactor:
is this method really useful now - new YieldController is created at game start anyway...?

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
MostCromulent: > The fields touched are read cross-thread:
> 
> - yieldJustEndedFlag is read+cleared by didYieldJustEnd() from InputPassPriority.showMessage() (line 87). showMessage() runs on the EDT — InputProxy.update() line 73 dispatches it via FThreads.invokeInEdtLater(showMessage).
> - noteMayAutoPassResult() is called from chooseSpellAbilityToPlay() on the game thread.
> - The synchronized pairs with the synchronized didYieldJustEnd() to provide a happens-before edge for the game→EDT visibility of yieldJustEndedFlag.
> 
> So synchronized is doing real work — but the inline comment "Game-thread only; no cross-thread races on the YIELD_SUPPRESS_AFTER_END flag" at PlayerControllerHuman.java:1510 is wrong/misleading about that. The flag IS read cross-thread; the synchronized is what makes it safe. The fix is to keep the synchronized and correct/remove the misleading comment.
> 
> (autoPassInterrupted and wasAutoPassingLastTick look game-thread-only, so they're not what carries the synchronization need — yieldJustEndedFlag is.)

tool4ever:
I'm seeing input getting stuck on netplay after I've played a land with the host (left side):
image

MostCromulent:
> I'm seeing input getting stuck on netplay after I've played a land with the host (left side):

I think latest commit fixes this, assuming below reflects same symptoms you had (this is how I managed to reproduce it):

- Host has APINA enabled, client does not.
- Host plays land in M1 with no other playable cards
- APINA engages and it passes priority to client, but host input prompt describes host as still having priority - so both player prompts are claiming priority at same time.

tool4ever (approved):

PR #10609, opened 2026-05-05
Properly scaled pixel art for UI elements

lemtom:
Several UI elements for the inn menu seem to have some sort of upscaling filter that resulted in weird artefacts, which stylistically clash with other UI elements. Given the graphical style of the objects, I assume some sort of pixelization filter was applied to them.

I've scaled these down and reworked them. The game scales them up again to the original size, but now the rendering is crisper and more coherent with the rest of the UI.

One actual change I did make is remove the dollar sign from the coins. It kinda broke my immersion, but I can add it again if wanted.

afbeelding


lemtom:
☹️

lemtom:
I understand that there's always work, but this is a PR that modifies six images…

PR #10611, opened 2026-05-05
Fix #9959: fixed AI behaviour concerning the card Champions of Tyr

### LINKED ISSUE #9959: AI refuses to cast Champions of Tyr
**Describe the bug**
AI refuses to cast Champions of Tyr.

**To Reproduce**
1. Create a deck with nothing but Mox Pearls and Champions of Tyr.
2. Play against an AI using this deck.
3. Pass turn forever.

**Expected behavior**
AI would play its spells at some point.

**Screenshots**

Image

 - OS: Windows 11
 - Version 2.0.10 snapshot 2.06

PedroCalhau:
The Ai would refuse to cast the Champions of Tyr card because it's behaviour wasn't defined in the card's script.

Added AI logic to the Champions of Tyr card script to ensure the AI can evaluate and cast the card appropriately.

Introduced new test cases to validate the AI's decision-making in various scenarios, including resource availability, card prioritization, and threat assessment.

### DIFF
--- a/forge-gui-desktop/src/test/java/forge/ai/simulation/SpellAbilityPickerSimulationTest.java
+++ b/forge-gui-desktop/src/test/java/forge/ai/simulation/SpellAbilityPickerSimulationTest.java
@@ -897,4 +897,94 @@ public class SpellAbilityPickerSimulationTest extends SimulationTest {
         game.getAction().checkStateEffects(true);
         AssertJUnit.assertNull(picker.chooseSpellAbilityToPlay(null));
     }
+
+    @Test
+    public void testChampionsOfTyrChosenFromMultipleCardsInHand() {
+        Game game = initAndCreateGame();
+        Player p = game.getPlayers().get(1);
+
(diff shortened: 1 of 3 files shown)


Jetz72:
I struggle to follow how adding `AILogic$ Always` onto several of the card's abilities can simultaneously solve an issue where the card is never cast by the AI, while also ensuring the AI can evaluate the card and play it only in an appropriate circumstance. The card doesn't seem complicated enough that it should need any hints directing the AI how to use it. Wonder what the root of the original issue is.

tool4ever:
Thanks for contributing:
in general we avoid tests that prove nothing new (these should already work on master + don't use heuristics engine anyway)

@Jetz72
I guess because of its complex nature EffectAi bails without specific logic - that shouldn't mean Always needs to be slapped on all three of these lines though...
Tbh quite a few cards do this and the real solution would be checking if those can be refactored into some minor code detection instead (I suspect most should) :/

PR #10612, opened 2026-05-05
Remember repeated simultaneous ability ordering decisions

Madwand99:
Adds an optional “Remember this decision” checkbox to the simultaneous ability ordering dialog. When checked, Forge stores the selected ordering for the current game and automatically reuses it when the exact same set of abilities occurs again.

Changes included:

- Adds a remembered-order result path to IGuiGame, CMatchUI, and GuiChoose.
- Adds the checkbox to DualListBox, positioned below the existing ordering controls.
- Stores remembered simultaneous ability orders in PlayerControllerHuman for the current game only.
- Keeps the existing previous-order preselection behavior for decisions that are not remembered.
- Skips the ordering prompt entirely when all simultaneous abilities are effectively identical and ordering would not matter.
- Adds localization for the new checkbox label across all UI language files.

### DIFF
--- a/forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
+++ b/forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
@@ -1946,24 +1946,53 @@ public class PlayerControllerHuman extends PlayerController implements IGameCont
         return PlaySpellAbility.payCostDuringAbilityResolve(this, player, cost, sa, null);
     }
 
-    // stores saved order for different sets of SpellAbilities
-    private final Map<String, List<Integer>> orderedSALookup = Maps.newHashMap();
+    // Stores previous ability orders for preselection or auto-apply in the current game.
+    private final Map<String, SavedAbilityOrder> orderedSALookup = Maps.newHashMap();
+
+    private record SavedAbilityOrder(List<Integer> order, boolean remember) {}
+
+    private static String getOrderLookupString(final SpellAbility sa) {
+        final String saStr = sa.toString();
+        final int idxAdditionalInfo = saStr.indexOf(" [");
+        return idxAdditionalInfo > 0 ? saStr.substring(0, idxAdditionalInfo) : saStr;
+    }
+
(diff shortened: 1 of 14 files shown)


forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1946,52 +1946,85 @@ public boolean payCostDuringRoll(final Cost cost, final SpellAbility sa) {
         return PlaySpellAbility.payCostDuringAbilityResolve(this, player, cost, sa, null);
     }
 
-    // stores saved order for different sets of SpellAbilities
-    private final Map<String, List<Integer>> orderedSALookup = Maps.newHashMap();
```
tool4ever, line 1950: hmn, I fail to see why the architecture needs to be extended this much:
you could have just turned this into a triple if you need an extra boolean 🤷‍♂️

though I have to wonder if getting rid of the manual confirm really matters that much? After all any change in triggers will require its own order anyway and not being able to bail out after turning one automatic feels bad too 🤔

it'd probably be more helpful if this could either partially reuse previous orders and/or use the new `PersistentYieldStore` to support fine-tuning across Forge sessions...

MostCromulent: I have a version of this same thing in development locally (was waiting for all the yield stuff to land before bringing it in). The way I implemented:
- Turned the existing 'preselect last order of abilities' preference into a combo-box with Off / Preselect / Auto hierarchy
- If auto is enabled, a 'Clear saved ability orders' button appears in game menu so you can bail out 

imo having some sort of auto option is useful and good QoL as long as you can get out of it

Madwand99: As @MostCromulent seems to have an upcoming PR that improves on this one, I'll close this one for now. All I want is to reduce some of the (many) extra clicks involved in certain game situations, I don't mind who does it.

PR #10614, opened 2026-05-05
Make multiplayer AI targeting threat-aware

Madwand99:
## Summary

Improves multiplayer AI target selection by adding a threat-aware opponent score and using it for combat and common removal decisions.

Previously, AI combat target selection mostly favored the weakest opponent, then fell back to weighted randomness when no opponent was near death. Removal targeting generally valued only the target permanent itself. In multiplayer games this could make AI decisions feel arbitrary from a human player’s perspective.

## Changes

- Added `ComputerUtil.evaluateOpponentThreatTo(ai, opponent)` to estimate how threatening each opponent is to a specific AI player.
- Updated AI combat defender selection to prefer opponents with a higher threat score while still valuing low life totals, monarch, and initiative.
- Added threat-aware removal target ranking in `ComputerUtilCard`.
- Applied the new removal ranking to common destroy, damage-based removal, exile, and bounce targeting paths.
- Cached per-opponent threat scores during removal target selection to avoid recalculating the same score for each permanent.

## Notes

The threat score is intentionally relative to the AI player evaluating it, not universal across all players. This lets the AI prefer opponents that are most dangerous to itself, rather than simply picking the strongest board in the abstract.


forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -2899,6 +2899,48 @@ public static Player evaluateBoardPosition(final List<Player> listToEvaluate) {
         return bestBoardPosition;
     }
 
+    public static int evaluateOpponentThreatTo(final Player ai, final Player opponent) {
```
tool4ever: some nice ideas here
though it'd be better this was an overloaded variant of `evaluateBoardPosition` so we don't implement another method that's very close in functionality...

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -202,33 +202,31 @@ public static Player choosePreferredDefenderPlayer(Player ai) {
...
 
-        if (defender.getLife() > 8) {
-            // TODO connect with evaluateBoardPosition and only fall back to random when no player is the biggest threat by a fair margin
-
-            List<Player> opps = Lists.newArrayList(ai.getOpponents());
+        Player bestDefender = ai.getWeakestOpponent();
+        int bestScore = Integer.MIN_VALUE;
+        for (Player opp : ai.getOpponents()) {
+            final int life = opp.getLife();
+            int score = ComputerUtil.evaluateOpponentThreatTo(ai, opp);
```
tool4ever: missing sanity check to avoid expensive method call:
with only one opponent result is always trivial

forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
```diff
@@ -567,6 +567,43 @@ public static Card getBestAI(final Iterable<Card> list) {
...
+        }
+        final Map<Player, Integer> threatCache = new IdentityHashMap<>();
+        return Aggregates.itemWithMax(list, c -> evaluateRemovalTargetPriority(ai, c, threatCache));
+    }
+
+    public static Card getBestCreatureRemovalTargetAI(final Player ai, final Iterable<Card> list) {
+        if (Iterables.size(list) == 1) {
+            return Iterables.get(list, 0);
+        }
+        final Map<Player, Integer> threatCache = new IdentityHashMap<>();
```
tool4ever: use `AiCache` for cross API support

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -2899,6 +2899,48 @@ public static Player evaluateBoardPosition(final List<Player> listToEvaluate) {
...
+        int rating = 0;
+        int attackPower = 0;
+
+        rating += opponent.getCardsIn(ZoneType.Hand).size() * 15;
+        rating += opponent.getLandsInPlay().size() * 8;
+
+        for (final Card c : opponent.getCardsIn(ZoneType.Battlefield)) {
+            if (c.isCreature()) {
+                rating += ComputerUtilCard.evaluateCreature(c) / 2;
+                if (CombatUtil.canAttackNextTurn(c, ai)) {
```
tool4ever: in theory should consider the opposite here too?
(how threatening is AI currently that the opponent would consider alpha striking it)
maybe better in a follow-up though, so could just leave comment like that for now...

forge-ai/src/main/java/forge/ai/ComputerUtil.java
Madwand99: Done!

forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
Madwand99: Done!

forge-ai/src/main/java/forge/ai/ComputerUtil.java
Madwand99: I've left a TODO note, though I think politics will be challenging to implement.

forge-ai/src/main/java/forge/ai/AiAttackController.java
Madwand99: Done!

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -207,16 +207,17 @@ public static Player choosePreferredDefenderPlayer(Player ai, boolean forCombatD
...
             final int life = opp.getLife();
             int score = ComputerUtil.evaluateBoardPosition(ai, opp);
-            score += Math.max(0, 40 - life) * 10;
-            if (life <= 8) {
-                score += (9 - life) * 100;
+            // round away slightly so a single land drop doesn't mean players with earlier turn order are predictably attacked
+            score = (int) Math.ceil(score / 10) * 10;
+            if (life > 0 && life < 20) {
+                // TODO commander damage
+                score += Math.max(0, (20 - life)^2);
```
Madwand99: ^ is bitwise XOR, not exponentiation. Is this intended?

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -2875,83 +2874,47 @@ public static boolean isUselessCounter(CounterType type, Card c) {
...
-            rating += 40;
+        if (ai == null) {
+            // non combat check takes life into account here
+            rating += opponent.getLife() * 3;
+        } else {
+            // TODO: Consider whether the opponent is likely to attack a bigger threat instead.
+            // This is hard to predict for human players and multiplayer politics.
+            int remainingLife = predictNextCombatsRemainingLife(ai, true, true, 0 , null, List.of(opponent));
+            if (remainingLife < ai.getLife()) {
+                rating += Math.abs(ai.getLife() - Math.max(-20, remainingLife))^2;
```
Madwand99: Same here, ^ is bitwise XOR, not exponentiation. Intended?

forge-ai/src/main/java/forge/ai/AiAttackController.java
tool4ever: ugh thanks, definitely not 🤦‍♂️

just wanted to turn the scaling less linear

Madwand99: Fixed!

tool4ever (approved):

PR #10617, opened 2026-05-06
CardDB cleanup; index cards by CardRules

Jetz72:
* Implements `allCardsByRules`, an alternative to `allCardsByName` that guarantees that cards don't appear in multiple entries due to things like prepared cards nesting one card within the faces of another. `allCardsByName` remains available for searching purposes.
* Migrates a bunch of CardDB's overloaded methods up to the interface as default methods.
* Removes the `noAlt` versions of the `getAllCards`/`getUniqueCards` methods when the name is not a parameter. The only real distinction between the two previously was that one version contained some unintended duplicate entries or was otherwise slightly wrong.
* Splits the `rulesByName` into `rulesByPrimaryName` and `rulesByAltName`. Allows a `noAlt` variant of `getRules`, though I chose to implement it as a parameter, which I might do for the others sometime too. 
* Switches `uniqueCardsByName` to `uniqueCardsByRules`, so that Emeritus of Ideation can no longer be chosen as the default printing of Ancestral Recall. 
* Swaps a few instances of name lookup to rules lookup. Going forward: if you want to search up other prints of a card, `getAllCards(CardRules)` or `getAllCards(PaperCard)` are preferable to `getAllCards(card.getName())`.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/ChooseCardNameAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/ChooseCardNameAi.java
@@ -102,7 +102,7 @@ public class ChooseCardNameAi extends SpellAbilityAi {
         final CardDb cardDb = StaticData.instance().getCommonCards();
 
         for (ICardFace face : faces) {
-            final CardRules rules = cardDb.getRules(face.getName());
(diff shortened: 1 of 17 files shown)


tool4ever (approved):
should be ok?

Hanmac (approved):

PR #10618, opened 2026-05-06
Lobby update fixes and null-card guard in CMatchUI.updateCards

autumnmyst:
- **`LobbySlot.apply`**: refactor the per-field `if (data.getX() != null) { setX(...); changed = true; }` cascade into `setIfChanged` / `setIntIfChanged` helpers that skip the write (and the `changed` flag) when the incoming value already matches the current one. Also routes `SchemeDeckName`, `AvatarVanguard`, `PlanarDeckName`, and `DeckName` through the same path; these were previously updated but never flipped `changed`, so listeners weren't notified.
- **`NetConnectUtil`**: drop the explicit `server.updateLobbyState()` call from the player-change listener. `server.updateSlot()` already routes through `updateLobbyState()` via the `IUpdateable` listener installed above, so the explicit call broadcast a duplicate `LobbyUpdateEvent` to every client on every slot change.
- **`FServerManager.broadcastReadyState`**: dedupe by slot index, tracking the last broadcast value per slot. Without this, `UpdateLobbyPlayerEvent`s with the same `getReady()` value (which fire routinely during normal lobby flow) repeated the chat line "Player X is ready" multiple times in a row.
- **`CMatchUI.updateCards`**: change `return` -> `continue` on null zone, and add a null check on the `CardView` itself. Null can flow in from a remote-side event whose IdRef failed to resolve in the tracker; without these guards the loop bailed out and dropped every later card in the same update batch from the visual update.

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
+++ b/forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
@@ -86,7 +86,6 @@ public final class FServerManager implements IHasForgeLog {
     });
 
     private final Map<Integer, RemoteClientGuiGame> playerGuis = new ConcurrentHashMap<>(); // Store RemoteClientGuiGame instances for reuse
-    private final Map<Integer, Boolean> lastBroadcastReadyState = new ConcurrentHashMap<>(); // dedupe ready-state chat broadcasts
 
     // Network byte tracking for monitoring actual bandwidth usage
     private final forge.gamemodes.net.NetworkByteTracker byteTracker =
@@ -401,15 +400,11 @@ public final class FServerManager implements IHasForgeLog {
         localLobby.applyToSlot(index, event);
 
         if (event.getReady() != null) {
-            broadcastReadyState(index, localLobby.getSlot(index).getName(), event.getReady());
+            broadcastReadyState(localLobby.getSlot(index).getName(), event.getReady());
         }
     }
 
-    private void broadcastReadyState(int index, String playerName, boolean isReady) {
-        Boolean prev = lastBroadcastReadyState.put(index, isReady);
-        if (prev != null && prev == isReady) {
-            return; // no transition; skip duplicate broadcast
-        }
+    private void broadcastReadyState(String playerName, boolean isReady) {
         int readyCount = 0;
         int totalPlayers = 0;
         for (int i = 0; i < localLobby.getNumberOfSlots(); i++) {
@@ -936,7 +931,7 @@ public final class FServerManager implements IHasForgeLog {
                     }
                 }
                 if (event.getReady() != null) {
-                    broadcastReadyState(client.getIndex(), client.getUsername(), event.getReady());
+                    broadcastReadyState(client.getUsername(), event.getReady());
                 }
                 // Return to prevent duplicate processing by LobbyInputHandler
                 return;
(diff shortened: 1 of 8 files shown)


MostCromulent:
> Null can flow in from a remote-side event whose IdRef failed to resolve in the tracker; without these guards the loop bailed out and dropped every later card in the same update batch from the visual update.

I assume this is based on a bug you experienced. Any chance you can post a copy of your network logs from when this happened? 

What did it manifest as in terms of visual glitch / how does it manifest with your fix in place?

In terms of root cause we can probably tolerate a transient visual glitch if it self corrects on next network packet, but if it's resulting in more serious visual desync will need further investigation.

forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
```diff
@@ -86,6 +86,7 @@ public final class FServerManager implements IHasForgeLog {
     });
 
     private final Map<Integer, RemoteClientGuiGame> playerGuis = new ConcurrentHashMap<>(); // Store RemoteClientGuiGame instances for reuse
+    private final Map<Integer, Boolean> lastBroadcastReadyState = new ConcurrentHashMap<>(); // dedupe ready-state chat broadcasts
```
tool4ever: you can't just cache all slots without refreshing in case client leaves when ready and then comes back or when returning after game end
otherwise some messages will get lost instead

autumnmyst: True, let me address that

autumnmyst:
> > Null can flow in from a remote-side event whose IdRef failed to resolve in the tracker; without these guards the loop bailed out and dropped every later card in the same update batch from the visual update.
> 
> I assume this is based on a bug you experienced. Any chance you can post a copy of your network logs from when this happened?
> 
> What did it manifest as in terms of visual glitch / how does it manifest with your fix in place?
> 
> In terms of root cause we can probably tolerate a transient visual glitch if it self corrects on next network packet, but if it's resulting in more serious visual desync will need further investigation.

I came across this issue from looking logs from intermittent crashes on specifically the remote client in online games. Basically if for whatever reason the IdRef fails to resolve, remote client gets multiple NPEs while trying to perform a visual update, so this skips over them instead of always trying to get the zone if `CardView` is null and stopping iteration if zone is null

forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
autumnmyst: Basically we want to rebroadcast even if the state doesn’t change for:
1. Client leaves and joins again
2. Game completes, client in lobby

So clear cache on client leave / game completion

tool4ever: I wonder if that's really the right approach though...
would probably be better to refactor it so only `isReadyUpdate` carries the flag

autumnmyst: done

tool4ever: well, something is wrong now:
enabling Dev Mode in mobile lobby is impossible

autumnmyst: Got it, going to test more thoroughly tonight

autumnmyst: > well, something is wrong now: enabling Dev Mode in mobile lobby is impossible

Okay just pushed a fix, there was an issue with me not updating dev mode before other checks that resulted in the UI not rendering it toggled when it actually was.

tool4ever (approved):

PR #10621, opened 2026-05-07
Improve AI restraint for costly land destruction abilities

Madwand99:
This updates DestroyAi so land-destruction abilities from lands are evaluated with the existing Strip Mine-style tempo logic even when the card script does not explicitly set AILogic.

The motivating case is Dust Bowl, whose script has no AILogic tag. Previously the AI could treat it like ordinary land removal and repeatedly spend {3}, tap Dust Bowl, and sacrifice another land to destroy low-impact nonbasic lands. With this change, Dust Bowl-like abilities are only accepted at that rate when the target is meaningfully valuable, contributes to a mana/color lock, or otherwise passes the existing land-destruction heuristics.

Changes:

- Added land-removal logic classification for GhostQuarter, LandForLand, and land-originated activated land-destroy abilities.
- Added a stricter cost gate for Dust Bowl-style effects that require extra mana or sacrificing another land.
- Preserved Ghost Quarter’s special caution, since the opponent may replace the destroyed land.
- Added regression tests covering Dust Bowl refusing low-priority nonbasics while still answering high-priority lands like Gaea’s Cradle.

Validation:

mvn -pl forge-gui-desktop -am -Dtest=forge.ai.ability.DestroyAiTest -Dsurefire.failIfNoSpecifiedTests=false test

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/DestroyAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/DestroyAi.java
@@ -19,6 +19,9 @@ import forge.game.zone.ZoneType;
 import forge.util.collect.FCollectionView;
 
 public class DestroyAi extends SpellAbilityAi {
+    private static final String LOGIC_GHOST_QUARTER = "GhostQuarter";
+    private static final String LOGIC_LAND_FOR_LAND = "LandForLand";
+
     @Override
     public AiAbilityDecision chkDrawback(Player ai, SpellAbility sa) {
         return checkApiLogic(ai, sa);
@@ -229,9 +232,10 @@ public class DestroyAi extends SpellAbilityAi {
                 } else if (CardLists.getNotType(list, "Land").isEmpty()) {
                     choice = ComputerUtilCard.getBestLandToRemoveAI(ai, list, sa);
 
-                    if ("LandForLand".equals(logic) || "GhostQuarter".equals(logic)) {
-                        // Strip Mine, Wasteland - cut short if the relevant logic fails
-                        if (!doLandForLandRemovalLogic(sa, ai, choice, logic)) {
+                    String landRemovalLogic = getLandRemovalLogic(sa, logic);
+                    if (landRemovalLogic != null) {
+                        // Strip Mine, Wasteland, Dust Bowl, and similar lands.
+                        if (!doLandForLandRemovalLogic(sa, ai, choice, landRemovalLogic)) {
                             return new AiAbilityDecision(0, AiPlayDecision.CantPlayAi);
                         }
                     }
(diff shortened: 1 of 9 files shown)


forge-ai/src/main/java/forge/ai/ability/DestroyAi.java
```diff
@@ -300,6 +304,40 @@ protected AiAbilityDecision checkApiLogic(final Player ai, final SpellAbility sa
         return new AiAbilityDecision(100, AiPlayDecision.WillPlay);
     }
 
+    private String getLandRemovalLogic(SpellAbility sa, String logic) {
+        if (LOGIC_GHOST_QUARTER.equals(logic)) {
+            return LOGIC_GHOST_QUARTER;
+        }
+        if (LOGIC_LAND_FOR_LAND.equals(logic) || isLandDestroyAbilityFromLand(sa)) {
```
tool4ever: if we have successful auto-detection there's no reason to keep the redundant manual tagging support

however I wonder if the heuristic you've introduced for that isn't still a bit too broad:
e.g. also including creatures with these abilities when they've gained land type

Madwand99: Good points. Fixed!

forge-ai/src/main/java/forge/ai/ability/DestroyAi.java
```diff
@@ -300,6 +303,34 @@ protected AiAbilityDecision checkApiLogic(final Player ai, final SpellAbility sa
         return new AiAbilityDecision(100, AiPlayDecision.WillPlay);
     }
 
+    private boolean shouldApplyLandRemovalLogic(SpellAbility sa, String logic) {
+        return LOGIC_GHOST_QUARTER.equals(logic) || isLandDestroyAbilityFromLand(sa);
+    }
+
+    private boolean isLandDestroyAbilityFromLand(SpellAbility sa) {
+        Cost cost = sa.getPayCosts();
+        return sa.isActivatedAbility()
```
tool4ever, line 311: try code a bit less redundant for basic Magic rules:
sa.isActivatedAbility() **&& !sa.isSpell()**
it's highly unlikely this would ever be possible together in the future

forge-ai/src/main/java/forge/ai/ability/DestroyAi.java
tool4ever: still need to remove the field representing it
and especially from the scripts that used it, so it doesn't look like logic is missing + it doesn't get copied around by contributors for new cards

forge-ai/src/main/java/forge/ai/ability/DestroyAi.java
Madwand99: Done!

forge-ai/src/main/java/forge/ai/ability/DestroyAi.java
Madwand99: Done!

tool4ever (approved):

PR #10629, opened 2026-05-08
Enable delta sync

### LINKED ISSUE #10532: Huge Network latency issues
Hello. A friend and I have been playing this project for a long time. Usually it's bearable but still quite laggy... however. Recently with the new Strixhaven Update it's been even worse. We can't even go through a turn without it taking forever to go through actions. And each time our Latency would spike to 5,000 To 1,145 especially when actions happen. Even the bots take way too long to do any actions. 

My internet speed is between 400-500 MBPS and I have zero issues with any other thing other than Forge. And my friend has similar speeds if not better.

I Will add that we use Radmin VPN to connect Peer-to-Peer but unsure if that changes anything. Any info on if this could be fixed would be greatly appreciated. Thank you.

tool4ever:
Closes #10532


MostCromulent:
👍I should have decent amount of time available this week if we need to hotfix.

PR #10630, opened 2026-05-08
CounterCustomType: extra fallback CounterType

Hanmac:
Part of #10076

This splits the custom fallback from the `CounterKeywordType`
meaning that `CounterKeywordType` are always official Keywords.

While `CounterCustomType` are fallback if Counter isn't found.
(we might add logging later that it does a warning/error if CardRules isn't custom too)


Jetz72 (approved):
`isKeywordCounter` feels a tad redundant if we have to use `instanceof CounterKeywordType` in some places but I can see how it'll make things tidier in the long run.

PR #10633, opened 2026-05-09
Mobile lobby: drive panel count from lobby state, not dropdown

### LINKED ISSUE #10632: Netplay lobby screen
So when I go to play with friends that are on desktop and im on android if I dont join second in the lobby I cannot ready up or change my deck you have to join second to be able to do anything i provided two pictures of what it looks like when I play against 4 AI versus when I join friends that were already in a lobby.

Image

Image

MostCromulent:
Fixes #10632.

## Bug report

On Android, joining an online lobby hosted by a desktop player worked correctly only when the mobile user took slot 1 ("second to join, after the host"). If the mobile user joined as slot 2 or higher (i.e. another client was already in slot 1), they couldn't ready up or change deck — and at higher slot counts they couldn't see their own slot at all.

## Root cause

`forge-gui-mobile`'s `LobbyScreen.getNumPlayers()` returned `cbPlayerCount.getSelectedItem()` — the local player-count dropdown, which defaults to 2 and only changes when the user manually picks a different value. For a host this matched the lobby (the dropdown's `changedHandler` calls `lobby.addSlot()`/`removeSlot()`), but for a client the lobby's slot count comes from the server via `LobbyUpdateEvent` and the dropdown was never updated. So a joining client's panel-creation loop, scroll layout, and `updateDeck` early-return all clamped to 2, regardless of how many slots the server actually had.

A handful of pre-existing ordering issues in `update()` were latent until clients started exercising slots 2/3: deck-chooser populate callbacks fired before `playerPanels.add(panel)`, mid-update `panel.setMayEdit()` revalidates iterated `playerPanels` while it was still being built, and `playersScroll` never got a final layout pass after the server-driven slot additions.

## Fix

Mirrors desktop's `VLobby` architecture (where `activePlayersNum` is recomputed from `lobby.getNumberOfSlots()` at the top of every `update()` and the dropdown-equivalent — `addPlayerBtn` — calls `lobby.addSlot()` directly):

- `getNumPlayers()` returns `lobby.getNumberOfSlots()`; the dropdown becomes a host-side input and a cosmetic readout for clients.
- `update()` loop iterates `MAX_PLAYERS` and hides panels beyond `playerCount` (matches `VLobby.update` lines 260–323).
- `cbPlayerCount.changedHandler` now reads `cbPlayerCount.getSelectedItem()` directly as its target slot count, since `getNumPlayers()` would otherwise be degenerate against the lobby.
- Panel registration in `update()` adds to `playerPanels` and `playersScroll` *before* calling `panel.initialize(...)` — matches `VLobby:272-277`. Deck-chooser populate fires `onSelectionChange` synchronously, which can recurse into `updateDeck(i)` expecting `playerPanels.get(i)` to find the panel.
- Defensive bounds in `playersScroll.layoutAndGetScrollBounds` and `updateDeck` (`Math.min(getNumPlayers(), playerPanels.size())`-style guards), to absorb any other mid-update callback that races the panel list against the lobby.
- Single `playersScroll.revalidate()` at the end of `update()`, since the mid-loop revalidates inside `setMayEdit` are gated on `getHeight() > 0`, which is false for freshly created panels.
- For clients (`!lobby.hasControl()`), the dropdown's displayed value is set to the lobby's slot count at the end of `update()` so the user sees the correct "Players: N".

## User-facing result

A mobile player joining an online lobby now sees every slot the host configured, regardless of join order, with their own slot fully interactive (ready toggle, deck/avatar/name/team controls) and the "Players: N" readout reflecting the actual lobby size.

---

### DIFF
--- a/forge-gui-mobile/src/forge/screens/constructed/LobbyScreen.java
+++ b/forge-gui-mobile/src/forge/screens/constructed/LobbyScreen.java
@@ -82,7 +82,10 @@ public abstract class LobbyScreen extends LaunchScreen implements ILobbyView {
         protected ScrollBounds layoutAndGetScrollBounds(float visibleWidth, float visibleHeight) {
             float y = 0;
             float height;
-            for (int i = 0; i < getNumPlayers(); i++) {
+            // Bound by panels.size() — panel.setMayEdit can call revalidate
+            // mid-loop in LobbyScreen.update() before all slots have panels.
+            int count = Math.min(getNumPlayers(), playerPanels.size());
+            for (int i = 0; i < count; i++) {
                 height = playerPanels.get(i).getPreferredHeight();
                 playerPanels.get(i).setBounds(0, y, visibleWidth, height);
                 y += height;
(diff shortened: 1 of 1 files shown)


tool4ever (approved):
thanks

PR #10517, opened 2026-04-25
Improved UI flow for Commander deck creation

MostCromulent:
## Summary

Follows discussion on https://github.com/Card-Forge/forge/issues/10476

UI flow improvements for Commander deck creation on desktop:
- New **Format** dropdown in the deck panel that switches the active editor and carries the in-progress deck across formats.
- Right-click **"Set card as commander"** entry on the Card Catalog, Main, and Sideboard panes when the active format has a Commander section, with Partner-aware replacement.
- Deck importer routes the import to the editor matching the selected legality format, and auto-detects an explicit Commander section header in pasted lists.    

## Background

Currently the only way to create a Commander deck on desktop is to _**first**_ open the Commander deck list tab in the Deck Editor **_before_** creating or importing the deck. 

This UI is non-intuitive: when a user opens the deck editor it immediately shows the card catalog window and a new deck to start building. The dependency between tabs and deck format is not obvious or advertised. Users start making a deck on the default window only to discover it doesn't support Commander format, with no clear reason why or how to change it.

### 1. Format dropdown in deck panel header

Screenshot 2026-04-25 105308

A new `Format:` dropdown sits underneath the title row, listing `Constructed` / `Commander` / `Oathbreaker` / `Brawl` / `Tiny Leaders` — the five desktop editor `GameType`s.

Changing the dropdown swaps the active editor and **carries the in-progress deck across**. Switching from a commander format to Constructed moves the Commander section into Main so the legendary isn't lost. Selecting the format that's already active is a no-op.

The dropdown also stays in sync with editor changes driven from the deck-list tab row — opening a commander deck from `VCommanderDecks` updates the dropdown automatically.

### 2. "Set card as commander" right-click context entry

When the active format has a Commander section, right-clicking an eligible card in the Card Catalog, Main pane, or Sideboard pane exposes a "Set card as commander" entry (or "as oathbreaker" / "as signature spell" for Oathbreaker) similar to existing 'Add card to sideboard' context entry. Behaviour:

- Reuses the existing eligibility logic in `DeckFormat.isLegalCommander(rules)` — no new validators.
- **Partner-aware**: if a partner-eligible commander is already set and the new card pairs with it, both are kept; otherwise the existing commander is moved into Main and the new card replaces it.
- After promotion, the deck pane switches to the Commander section so the result is immediately visible.

### 3. Deck importer automatically routes to the correct editor

[...]

PR #10636, opened 2026-05-09
Add unified Deck Browser for home and deck editor

Madwand99:
This branch adds a shared Deck Browser experience for the desktop Home tab and Deck Editor.

Major changes:

- Replaces the Home tab deck list with a navigable Deck Browser that supports folders, .., virtual folders, generated deck options, net decks, archive decks, recursive search, refresh, and richer deck columns.
- Adds a Deck Browser tab to the Deck Editor and removes the old Commander/Oathbreaker/Brawl/Tiny Leaders deck tabs.
- Hooks Deck Editor browser navigation into the current deck save/load flow, including save target tracking by current folder.
- Reorganizes generated deck options into navigable virtual folders, with separate icons for folders, net folders, generated folders, and generated deck choices.
- Improves selection behavior so folder rows cannot be chosen as playable decks, restored/selected decks are highlighted and scrolled into view, and .. navigation preserves useful selection context.
- Adds refresh behavior for normal folders, net decks, and archive decks.
- Adds recursive Deck Browser search that runs on Enter and restores normal folder contents when canceled.
- Consolidates generated deck proxy behavior into GeneratedDeckProxy.
- Consolidates net archive deck storage/loading behavior through NetDeckStorageBase, reducing repeated archive-specific code.

Validation:

mvn -pl forge-gui-desktop -am -DskipTests compile passes.

DeckBrowser4

DeckBrowser3

DeckBrowser2

DeckBrowser1


tool4ever:
well this is huge...

can you spin off the `NetDeckStorageBase` refactor into its own PR for a first size reduction?

Madwand99:
> well this is huge...
> 
> can you spin off the `NetDeckStorageBase` refactor into its own PR for a first size reduction?

Yeah it's kept me pretty busy. And sure, will try that and see how it looks.

MostCromulent:
@Madwand99 currently the "format" of a deck (constructed/commander/oathbreaker) etc is determined by which deck browser panel you have open at the time the deck is created, so you have to be in the commander tab when you create your deck or else it doesn't get a command zone.

How does this work in your PR if we're collapsing those different browser panels into a general purpose one? i.e. how do you assign your deck to commander format here?

(FWIW I think the current "open format tab first, then create deck" approach is highly unintuitive hence https://github.com/Card-Forge/forge/pull/10517 - so I've got nothing against a unified approach, just want to make sure we don't lose functionality along the way).

Madwand99:
> How does this work in your PR if we're collapsing those different browser panels into a general purpose one? i.e. how do you assign your deck to commander format here?

@MostCromulent In the editor, the new Deck Browser is not just a passive file list. Navigating into a real format folder updates the Current Deck section’s active deck type. So entering the commander folder switches the editor to Commander handling, and new/saved decks from there get the command zone as expected. The browser is general-purpose UI, but the selected real folder still provides the format context.

tool4ever:
> > How does this work in your PR if we're collapsing those different browser panels into a general purpose one? i.e. how do you assign your deck to commander format here?
> 
> @MostCromulent In the editor, the new Deck Browser is not just a passive file list. Navigating into a real format folder updates the Current Deck section’s active deck type. So entering the commander folder switches the editor to Commander handling, and new/saved decks from there get the command zone as expected. The browser is general-purpose UI, but the selected real folder still provides the format context.

not sure this binding of folder name = variant type is flexible enough or wouldn't lead to some undesired results... 🤔

Madwand99:
> can you spin off the `NetDeckStorageBase` refactor into its own PR for a first size reduction?

@tool4ever Anything else needed?

Madwand99:
> not sure this binding of folder name = variant type is flexible enough or wouldn't lead to some undesired results... 🤔

@tool4ever Good point. I checked this and tightened it up. The mapping is not intended to be arbitrary “folder name equals variant”; it is based on known deck root paths such as the configured commander/oathbreaker/brawl/tiny leaders directories. Subfolders now inherit that explicit root context, including when opened from recursive/search results. Unknown custom folders still default to constructed rather than becoming variants by name.

MostCromulent:
> not sure this binding of folder name = variant type is flexible enough or wouldn't lead to some undesired results... 🤔

imo this is why ideally Forge shouldn't be enforcing a norm of save location = variant at all. 

variant type should just be assigned by user on a per deck basis with a format dropdown or checkbox in the editor

(not this PRs fault, it's just working with the current backend architecture)

paulsnoops:
image

Just a small thing but it seems to show the "folder/back to previous folder" counted as a deck
Custom user decks 2/2 when there's only 1

Madwand99:
@paulsnoops Fixed!

Madwand99:
I've added a fix for the issue in #10913, adding guards to prevent deck generation during sorting.

Madwand99:
Per a discussion with @tool4ever, I did a cleanup pass to reduce some of the Deck Browser bloat and make the architecture easier to review.

The main changes are:

- Extracted generated deck browser row/group logic into DeckBrowserGeneratedRows.
- Extracted net/archive loading and saved-state handling into DeckBrowserNetService.
- Grouped scattered deck browser navigation/search state into a BrowserState object.
- Centralized browser-row wrapping in DeckBrowserEntry.fromDeckProxy(...).
- Simplified DeckManager checks for “real deck row vs browser navigation row.”

The goal is to keep FDeckChooser as the UI coordinator while moving generated-deck and net/archive rules into smaller, named helpers.

forge-gui-desktop/src/main/java/forge/deckchooser/FDeckChooser.java
```diff
@@ -130,208 +240,1173 @@ public void setDeckSelectionCommand(final UiCommand command) {
...
+            decksComboBox.setText(netDeckCategory.getDeckType());
         }
-        syncUrlFieldWithSelectedDeck();
+        updateDecks(DeckProxy.getNetDecks(netDeckCategory));
     }
 
-    private void setDeckPoolWithConfig(final Iterable<DeckProxy> decks, final ItemManagerConfig config) {
-        updatingDeckPool = true;
-        try {
-            // Clear the old source before applying a new ItemManagerConfig; otherwise stale items can
```
tool4ever, line 319: you should really review your AI code so it doesn't remove important comments...

tool4ever:
I'll note that you're currently able to quickly switch between editing decks of each type via their tabs which wouldn't be possible now?

forge-gui-desktop/src/main/java/forge/deckchooser/FDeckChooser.java
Madwand99: I do, but I missed this one. Comment restored!

Madwand99:
> I'll note that you're currently able to quickly switch between editing decks of each type via their tabs which wouldn't be possible now?

This PR doesn’t remove the ability to switch deck type, but it moves it from separate tabs into folder navigation. Entering a real format folder in the Deck Browser updates the Current Deck section’s active deck type, so opening commander, oathbreaker, brawl, or tiny leaders switches the editor context accordingly. Adding #10517 before or after this PR is merged should improve the selection process even further.

Madwand99:
I've refactored the Deck Browser implementation to reduce chooser bloat and clarify responsibilities:
- Extracted browser location, generated-row, and net/archive behavior into focused helpers.
- Consolidated browser state and repeated row/proxy handling.
- Removed obsolete UI plumbing and duplicate logic.
- Preserved empty selections during navigation and inline actions.
- Hardened refresh and editor save-target behavior.

jamincollins:
## Findings
#### 1. DeckBrowserEntry.getAI returns empty deck for non-deck rows
Description: DeckBrowserEntry.getAI() returns EMPTY_DECK.getUnplayableAICards() for GENERATED_OPTION rows, which may hide AI-specific deck data for generated options. This could cause incorrect AI evaluation when browsing generated decks.
Test case:
```java
DeckBrowserEntry entry = DeckBrowserEntry.generatedOption("test", "", null);
Deck.UnplayableAICards ai = entry.getAI();
// ai should be null or properly reflect generated deck, not empty deck
```
Impact: AI deck evaluation may be incorrect for generated rows.

#### 2. DeckBase.hashCode uses name twice
Description: DeckBase.hashCode() returns (name.hashCode() * 17) + name.hashCode(), effectively 18 * hashCode. This violates typical hashCode contract but unlikely to cause immediate bug. Still inconsistent with equals.
Test case:
```java
DeckBase d1 = ...;
DeckBase d2 = ...;
assert d1.hashCode() != (d1.getName().hashCode() * 17 + d1.getName().hashCode());
```
Impact: Hash collisions possible.

#### 3. DeckStorage.setSourceFileName null safety
Description: DeckStorage.setSourceFileName is set from file.getName() without null check on file. If file is null, NPE.
Test case:
```java
File file = null;
Deck d = ...;
d.setSourceFileName(file.getName()); // NPE
```
Impact: Crash on null file.

#### 4. DeckBrowserEntry.unwrap returns null for non-deck rows
Description: unwrap returns null for non-deck rows, causing NPE in callers expecting non-null.
Test case:
```java
DeckBrowserEntry folder = DeckBrowserEntry.folder("x", "", null, null);
DeckProxy unwrapped = DeckBrowserEntry.unwrap(folder);
// unwrapped is null, caller may NPE
```
Impact: Potential NPE in DeckManager.isCommanderDeck.

#### 5. DeckBrowserEntry.getUniqueKey uses kind|path|name
Description: Unique key for non-deck rows uses kind|path|name which may collide if name contains pipe.
Test case:
```java
DeckBrowserEntry e1 = DeckBrowserEntry.folder("a|b", "p", null, null);
DeckBrowserEntry e2 = DeckBrowserEntry.folder("a", "b|p", null, null);
// Same unique key possible
```
Impact: Data collision.

#### 6. DecksComboBox.setDisplayedDeckType sets text directly
Description: setDisplayedDeckType sets text via super.setText(deckType.toString()) when not in model, bypassing selection event.
Test case:
```java
combo.setDisplayedDeckType(DeckType.CUSTOM_DECK);
// Text changes but selection event not fired
```
Impact: UI state mismatch.

### Performance and security
#### 7. DeckBrowserGeneratedRows.getDeckOptions loads formats repeatedly
Description: getDeckOptions calls FModel.getFormats().getStandard() etc on each invocation, causing repeated lookups.
Test case:
```java
for (int i=0;i<1000;i++) { DeckBrowserGeneratedRows.getDeckOptions(...); }
// Repeated format lookups
```
Impact: Performance degradation.

## Verdict
Request Changes
The PR introduces a unified Deck Browser but contains several null safety and data integrity issues that need addressing before merge.

## Recommended Fixes
1. Fix DeckBase.hashCode to use proper combination of fields.
2. Add null checks in DeckStorage and DeckBrowserEntry.unwrap callers.
3. Sanitize unique key generation for DeckBrowserEntry.
4. Ensure AI data for generated options is properly handled.
5. Optimize format lookups in DeckBrowserGeneratedRows.
---

PR #10640, opened 2026-05-09
Consolidate net archive deck storage classes

Madwand99:
This PR extracts the repeated net archive deck storage behavior into NetDeckStorageBase. The individual archive classes now mainly declare their archive list file and display prefix, while shared loading, deck type labeling, and selectAndLoad behavior live in one place.

This is intended as a small preparatory cleanup before the larger Deck Browser work. It reduces repeated code across the net archive implementations without changing the user-facing deck browser behavior.

In preparation for #10636


tool4ever (approved):
thanks

PR #10646, opened 2026-05-10
Card grid for desktop reveal and single-pick dialogs

MostCromulent:
### **Example using Inquisition of Kozilek**
Before:
Screenshot 2026-05-10 080841
After:
Screenshot 2026-05-10 092046

---

- #10546 introduced `CardImageGrid` as a general utility for the deck-editor art picker; this PR takes advantage of it for desktop reveal and single-pick dialogs, so reveal/discard/exile prompts show thumbnails instead of a names-only list.
- `CardImageGrid` is now generic over the item type via a `CellAdapter`; `forPaperCards` keeps the deck-editor art picker, `forCardViews` is the new in-match path.
- Non-`CardView` `ListChooser` dialogs (counter types, colors, modes, spell abilities) stay on the text list — there's no card art to render.
- Multi-pick `CardView` selection isn't part of `ListChooser` traffic at all — it routes through `DualListBox` (two-column ordering) or `InputSelectCardsFromList` (clicking cards on the battlefield), which are separate widgets with their own UX. Bringing the grid look to those is a larger change and out of scope here.
- Desktop GUI only — no protocol, serialization, or game-state changes; wire format and delta sync are untouched.

---

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/gui/ListChooser.java
+++ b/forge-gui-desktop/src/main/java/forge/gui/ListChooser.java
@@ -28,14 +28,17 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Objects;
 import java.util.function.Function;
 
 import javax.swing.AbstractListModel;
 import javax.swing.DefaultListCellRenderer;
 import javax.swing.ImageIcon;
+import javax.swing.JComponent;
 import javax.swing.JList;
 import javax.swing.JPanel;
 import javax.swing.ListCellRenderer;
+import javax.swing.ListModel;
 import javax.swing.ListSelectionModel;
 import javax.swing.SwingUtilities;
 import javax.swing.WindowConstants;
@@ -50,6 +53,7 @@ import com.google.common.collect.Lists;
 
 import forge.card.CardType;
 import forge.card.MagicColor;
+import forge.game.card.CardView;
 import forge.game.card.CounterKeywordType;
 import forge.game.card.CounterType;
 import forge.localinstance.skin.FSkinProp;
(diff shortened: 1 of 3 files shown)


MostCromulent:
Potential follow-up PR applying same visual concept to `DualListBox`, with drag-and-drop support:

Screenshot 2026-05-11 071756

tool4ever:
doesn't really convince me:
might be ok visually for ~3 cards but a bigger list? 😱

and the way it looks right now usability worsens without:
a) being able to see current Oracle
b) using keyboard to type your selection
c) much more of boardstate hidden behind dialog

If anything that *Inquisition of Kozilek* could instead let you pick from the players hand view directly (which would also align more with other GUI workflows)...

MostCromulent:
I guess my view is: it's a card game, should try and surface the cards rather than plain text as much as possible in UI widgets so long as it doesn't seriously compromise usability. 

I think some of those usability issues above can probably be surmounted:

- In my local duallistbox implementation it caps at (off the top of my head) 7 cards; any more than that just uses plain text. 
Could apply that to both widgets so you get card visualisation for most decisions involving small numbers of cards and the text list for large lists. 

- can incorporate search filter

- could have it scale widget size to some capped amount of window space so oracle/card image panels remain visible

- could add "view battlefield" button to minimise/restore widget 

Agree with your Kozilek discard suggestion though, I'll look into that further...

MostCromulent:
I'm going to close this one for now to get it off the open list, and think on this a bit more.

Based on https://github.com/Card-Forge/forge/pull/10660 a more fruitful path for exploration might be how FloatingZone could be re-used for other selection effects rather than re-inventing the wheel with some bespoke new widget. 

I'll show off a POC on the discord in future rather than going straight to PR.

PR #10648, opened 2026-05-10
Fix yield activation enabling cancel on forced input prompts

MostCromulent:
## Summary

Reported via Discord (Android, v2.0.13-SNAPSHOT-05.09): a Glacial Chasm ETB sacrifice trigger could silently no-op if the user clicked "Yield to entire stack" while the sacrifice prompt was showing, then pressed Cancel.

## Cause

`applyYieldUpdate` calls `updateAutoPassPrompt()` on every yield activation. That helper repaints the shared prompt panel — including re-enabling Cancel for "cancel yield". When the active input is a forced selection (e.g. `InputSelectCardsFromList` for "sacrifice a land", min=1) it had deliberately disabled Cancel; re-enabling it lets the user click what looks like "cancel yield" but is wired to the input's `onCancel`, which drops the mandatory selection with an empty result.

## Fix

Gate the repaint to passive inputs (`InputPassPriority`, `InputLockUI`, null). The yield flag still arms; the active prompt is left alone.

---

image

---


tool4ever (approved):
not 100% happy - logic looks a bit shaky if input roles get changed in the future

PR #10658, opened 2026-05-11
docs: add Realm of Legends plane and expand Shandalar Old Border section

vanja-ivancevic:
## Summary

Updates `docs/Adventure/Planes.md` (which mirrors the [Planes](https://github.com/Card-Forge/forge/wiki/Planes) wiki page)


MostCromulent:
May be worth moving the active planes to a section at the top of the page so you don't need to scroll past the "in development" Amonkhet, Innistrad etc planes which are hidden by default before you see them.

PR #10213, opened 2026-03-29
CounterListType: add List for easy extending counters

Hanmac:
Part of #10076 

adds an easy extendable List for Counters where colors can be defined

PR #10659, opened 2026-05-11
CounterType: reduce the use of CounterEnumType for low used ones

Hanmac:
Extracted from #10213 

it reduces the amount of `CounterEnumType` cases for `CounterType.getType`,
so the types can be extracted into the List


Hanmac:
@tool4ever this is cleanup for the other MR, reducing the amount where some one-off Counters are used by AI code

Jetz72 (approved):

PR #10615, opened 2026-05-06
Route libgdx-port checks through the target GUI

MostCromulent:
## Bug report

Resolves #10613 — On the mobile client against a desktop host, search-your-library effects (Demonic Tutor, Solemn Simulacrum's basic-land trigger, fetch lands) present prompt but no cards are visible.

## The general issue

Several UI decision points read `GuiBase.getInterface().isLibgdxPort()` to determine whether to route to desktop or mobile rendering paths.

This is a process-global static answering **"is this JVM libgdx?"** — when the right question is **"is the player whose UI will render this libgdx?"**. The two answers agree in single-player and same-platform netplay, but diverge in cross-platform netplay: a desktop host serving a mobile client returns `false` and steers the choice down a desktop-renderer code path the mobile client cannot honor.

### How it manifests in the tutor case

```
PlayerControllerHuman.chooseSingleEntityForEffect (line 585)
                        │
                        ▼
        ┌──────────────────────────────────────┐
        │ useSelectCardsInput(optionList, sa)  │
        │                                      │
        │  Reads GuiBase.getInterface()        │
        │       .isLibgdxPort()                │
        │  ── PROCESS-LOCAL on the HOST ──     │
        │                                      │
        │  Library zone qualifies only when    │
        │  !isLibgdxPort()  (PCH.java:459)     │
        └──────────────┬───────────────────────┘
                       │
        ┌──────────────┴───────────────┐
        │                              │
   returns TRUE                  returns FALSE
        │                              │
        ▼                              ▼
   PATH A                          PATH B
   InputSelectEntitiesFromList     getGui().chooseSingleEntityForEffect
        │                              │
        │ setSelectables ──┐           │ → GameEntityPicker / SGuiChoose
        │ tempShowZones ───┤           │   (a self-contained dialog —
        │ updateZones      │           │    does NOT use tempShowZones)
        │ hideZones ───────┘           │
        ▼                              ▼
   Renders cards INSIDE the         Renders cards INSIDE the picker
   matching floating zone           dialog. Works regardless of
   on the player's screen.          whether floating zones exist.
   ── REQUIRES tempShowZones
      to be implemented ──
```

Desktop host → check returns false → PATH A → host serializes `tempShowZones` over the wire to the mobile client. Mobile's `MatchScreen.tempShowZones` is an unimplemented stub (`// pfps needs to actually do something`), so cards never render. The same anti-pattern wrong-routes confirm dialogs, scry reorder, end-step discard, AI-skip reveals, and stack notifications.

[...]

PR #10660, opened 2026-05-11
Route opponent-hand reveal and discard through FloatingZone on desktop

MostCromulent:
**USING MIND WARP, X=2**

**BEFORE:**
Reveal step
Screenshot 2026-05-12 074617
Select step
Screenshot 2026-05-12 074624

**AFTER:**
Reveal step
Screenshot 2026-05-12 074032
Select step
Screenshot 2026-05-12 074044

---

## Summary

Per https://github.com/Card-Forge/forge/pull/10646#issuecomment-4416485496: `RevealYouChoose` / `LookYouChoose` discards (Inquisition, Mind Warp, etc.) currently surface as a names-list reveal dialog followed by a separate selection grid, when the opponent's hand could be the selection UI directly.

This routes both steps through the existing FloatingZone: reveal opens the hand face-up with a minimal OK acknowledgement; the pick uses `InputSelectCardsFromList` so legal targets are clicked in that same FloatingZone.

## Implementation

- `PlayerControllerHuman.reveal()` — desktop opponent-hand reveals swap `getGui().reveal()` (names list) for `tempShowZones` + `getGui().message()` (visual hand + minimal OK).
- `PlayerControllerHuman.chooseCardsToDiscardFrom()` — opponent-discard pick routes through `InputSelectCardsFromList`; tempShows the full hand for `Reveal*`/`Look*` modes without `RevealNumber` so revealed non-selectable cards stay visible. `RevealNumber` spells (Blackmail, Cabal Interrogator) keep the conservative tempShow-of-valid path so unrevealed cards stay face-down.
- `CMatchUI.hideZones()` — fix a pre-existing asymmetry: `tempShowZones` special-cases `ZoneType.Hand`, but `hideZones` only checked `FLOATING_ZONE_TYPES` (which omits Hand). The FloatingZone opens for an opponent's hand but never closed programmatically.

## Safety

Both PCH changes gate on `UI_SELECT_FROM_CARD_DISPLAYS && !getGui().isLibgdxPort()` — the same condition `useSelectCardsInput` already uses. Mobile and preference-off paths are unchanged. `isLibgdxPort()` is the per-GUI network-safe check from #10615. No engine, protocol, or `TrackableProperty` changes.

`RevealNumber` partial-reveal spells (e.g. Blackmail revealing 3 of 5) are rendered safely: `FloatingZone.getCards()` returns all hand cards, but each is drawn according to the existing `mayView` check — only `mayLookTemp`-flagged cards (the revealed subset) show their faces, the rest render as card backs. Non-revealed cards' identities stay hidden; only hand size is conveyed, which is already public info.

## Potential code cleanup (not in this PR)

`FPref.UI_SELECT_FROM_CARD_DISPLAYS` was added in 2019 (commit `09fc3ae60c2`) alongside the original `arrangeForScry` GUI introduction, which brought in-zone selection as a new UX. The preference has defaulted to `true` with no user-facing UI toggle ever since — effectively the active desktop UI path for 7+ years. 

The preference gates whether `InputSelect` extends beyond the local player's own hand and battlefield to other zones (opponent-hand FloatingZone, libraries, graveyards, etc.) — i.e. click a card in its zone display versus fall back to a names-list dialog. All three desktop call sites in `PlayerControllerHuman` pair it with `!getGui().isLibgdxPort()`. 

A separate cleanup PR could drop the `FPref` entry, simplify the three dual-gate sites to single `!isLibgdxPort()` checks, and collapse the now-dead fallback branches in `useSelectCardsInput` (the zone-set ternary) and `arrangeForScry` (the manual-order path).

---

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/view/arcane/FloatingZone.java
+++ b/forge-gui-desktop/src/main/java/forge/view/arcane/FloatingZone.java
@@ -673,6 +673,7 @@ public class FloatingZone extends FloatingCardArea {
             promptLabel.setText(FSkin.encodeSymbols(prompt, false));
             promptLabel.setVisible(true);
         } else {
+            promptLabel.setText("");
             promptLabel.setVisible(false);
         }
         window.revalidate();
--- a/forge-gui-desktop/src/main/java/forge/view/arcane/FloatingZone.java
+++ b/forge-gui-desktop/src/main/java/forge/view/arcane/FloatingZone.java
@@ -18,6 +18,8 @@
 package forge.view.arcane;
 
 import java.awt.Color;
+import java.awt.KeyboardFocusManager;
+import java.awt.event.KeyEvent;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.awt.event.MouseMotionAdapter;
@@ -26,13 +28,17 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 import javax.swing.BorderFactory;
 import javax.swing.ScrollPaneConstants;
+import javax.swing.SwingConstants;
 import javax.swing.SwingUtilities;
 import javax.swing.WindowConstants;
 import javax.swing.border.Border;
+import javax.swing.event.DocumentEvent;
+import javax.swing.event.DocumentListener;
 
 import forge.game.card.CardView;
 import forge.game.player.PlayerView;
@@ -52,9 +58,12 @@ import forge.model.FModel;
 import forge.screens.match.CMatchUI;
 import forge.screens.match.views.VHand;
 import forge.screens.match.views.VZone;
+import forge.toolbox.FLabel;
 import forge.toolbox.FMouseAdapter;
 import forge.toolbox.FScrollPane;
 import forge.toolbox.FSkin;
+import forge.toolbox.FTextField;
+import forge.toolbox.MouseTriggerEvent;
 import forge.toolbox.special.PlayerDetailsPanel;
 import forge.util.Localizer;
 import forge.util.collect.FCollection;
@@ -518,6 +527,16 @@ public class FloatingZone extends FloatingCardArea {
     protected boolean sortedByName = false;
     protected FCollection<CardView> cardList;
 
+    private final FTextField searchField = new FTextField.Builder()
+            .ghostText(Localizer.getInstance().getMessage("lblFilterByName"))
+            .build();
+    private final FLabel hotkeyHint = new FLabel.Builder()
+            .text(Localizer.getInstance().getMessage("lblHotkeySelectHint"))
+            .fontSize(11)
+            .fontAlign(SwingConstants.CENTER)
+            .build();
+    private String filter = "";
+
     private final Comparator<CardView> comp = (lhs, rhs) -> {
         if (!getMatchUI().mayView(lhs)) {
             return (getMatchUI().mayView(rhs)) ? 1 : 0;
@@ -538,6 +557,11 @@ public class FloatingZone extends FloatingCardArea {
             } else if (zone == ZoneType.Flashback) {
                 cardList.sort(ZONE_ORDER_COMPARATOR);
             }
+            if (!filter.isEmpty()) {
+                final String needle = filter.toLowerCase(Locale.ROOT);
+                cardList.removeIf(card -> !getMatchUI().mayView(card)
+                        || !card.getName().toLowerCase(Locale.ROOT).contains(needle));
+            }
             return cardList;
         } else {
             return null;
@@ -546,7 +570,15 @@ public class FloatingZone extends FloatingCardArea {
 
     private FloatingZone(final CMatchUI matchUI, final PlayerView player0, final ZoneType zone0) {
         super(matchUI, new FScrollPane(false, ScrollPaneConstants.VERTICAL_SCROLLBAR_AS_NEEDED, ScrollPaneConstants.HORIZONTAL_SCROLLBAR_NEVER));
-        window.add(getScrollPane(), "grow, push");
+        searchField.getDocument().addDocumentListener(new DocumentListener() {
+            @Override public void insertUpdate(final DocumentEvent e) { onFilterChanged(); }
+            @Override public void removeUpdate(final DocumentEvent e) { onFilterChanged(); }
+            @Override public void changedUpdate(final DocumentEvent e) { onFilterChanged(); }
+        });
+        window.add(searchField, "growx, wrap");
+        window.add(getScrollPane(), "grow, push, wrap");
+        hotkeyHint.setVisible(false);
+        window.add(hotkeyHint, "growx");
         window.setDefaultCloseOperation(WindowConstants.DISPOSE_ON_CLOSE); //pfps so that old content does not reappear?
         getScrollPane().setViewportView(this);
         setOpaque(false);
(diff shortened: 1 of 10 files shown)


MostCromulent:
Let me know if you want the `FPref.UI_SELECT_FROM_CARD_DISPLAYS` code cleanup suggested above as part of this PR.

tool4ever:
I think in general this seems better:
- now leverages the advantage of being able to inspect the boardstate, helping you make tough decisions
- the used dialog had incorrect implications about the order mattering (while rare it sometimes does but that decision is done afterwards by the owner!)

the only minor downside remaining is no longer being able to use the keyboard to progress:
- though arguably that was a rather fiddly process, I doubt many users were doing that over mouse unless they had a very good reason (technical/physical)...?
- but I think we could do much better and turn it into a real **speed-up for everyone** across **multiple use cases** 🧠

what if we (temporarily) assign the number keys to the first 9 selectables? might have to draw small labels on them for clearity or something - maybe based on Ctrl press 🤔
want to give that a shot? 🥳

(there's actually one place in `CMatchUi` that uses `VK_*` similarly already but for abilities)

MostCromulent:
> what if we (temporarily) assign the number keys to the first 9 selectables? might have to draw small labels on them for clearity or something - maybe based on Ctrl press 🤔
> want to give that a shot? 🥳

yeah I'll try that out - might also look at how seamlessly we could integrate a search filter at the top as well, since we also lose that compared being able to start typing in the previous list view

MostCromulent:
Screenshot 2026-05-12 213043

Should work for any card-select prompt that uses FloatingZone - opponent hand, graveyard, exile, library etc.

Does not apply for cases which don't use FloatingZone - own hand, battlefield, players, ephemeral non-zone cards (though as follow up I'm going to explore what we could do about the ephemerals; plausibly could generate a temp FloatingZone window containing the ephemeral cards to take advantage of same UX).

forge-gui-desktop/src/main/java/forge/view/arcane/FloatingZone.java
```diff
@@ -648,4 +719,60 @@ private void setPlayer(PlayerView player0) {
...
+            return false;
+        }
+        if (!e.isControlDown() || e.isAltDown() || e.isMetaDown()) return false;
+        final int digit = e.getKeyCode() - KeyEvent.VK_0;
+        if (digit < 1 || digit > 9) return false;
+        for (final FloatingZone fz : floatingAreas.values()) {
+            if (!fz.isVisible()) continue;
+            final CardPanel target = fz.findPanelByHotkeyDigit(digit);
+            if (target == null) continue;
+            fz.getMatchUI().getGameController().selectCard(target.getCard(), null,
```
tool4ever, line 787: I wonder if we could use 0 to just select the first X minimum needed to proceed
but might be too tricky to obtain here...?

tool4ever:
> Let me know if you want the `FPref.UI_SELECT_FROM_CARD_DISPLAYS` code cleanup suggested above as part of this PR.

hmn I'm undecided if users might prefer having the preference actually GUI-exposed now that this PR increases its usage 🤔
let's wait a bit and see if we'll get some reactions after merging on Discord?

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1144,7 +1162,20 @@ public CardCollectionView chooseCardsToDiscardFrom(final Player p, final SpellAb
         boolean optional = min == 0;
 
         if (p != player) {
-            tempShowCards(valid);
+            // Reveal/Look modes (no RevealNumber) publicly disclose the full hand — keep non-selectable revealed cards visible
+            final String discardMode = sa.getParam("Mode");
+            final boolean fullHandRevealedToChooser = discardMode != null
+                    && !sa.hasParam("RevealNumber")
+                    && (discardMode.startsWith("Reveal") || discardMode.startsWith("Look"));
```
Jetz72: Could the reveal-full-hand flag be a parameter computed at the call site? A while ago I was trying to cut back on the number of places where the `PlayerController` has to dissect a `SpellAbility` in order to decide how to behave. Looks like only `DiscardEffect` would need it - `BalanceEffect` and `ConniveEffect` both only use the player's own hand.

Jetz72:
> Screenshot 2026-05-12 213043

If the revealed zone is prompting the user to choose something (e.g. some number of cards to discard), maybe the prompt might replace the usual title text? Or does it already seem visible enough in the usual prompt space?

forge-gui-desktop/src/main/java/forge/view/arcane/FloatingZone.java
MostCromulent: Looks like this shouldn't be too hard but would require small protocol change.

Currently `InputSelectCardsFromList` doesn't broadcast the min/max amount to network clients, it just waits to receive the choice. Would need to add that to the protocol. This is what happens with list dialogs already so there's precedent for it.

Changes involved:

```
  1. IGuiGame.java — new interface method
  void setSelectionRange(int min, int max);
  Placed next to setSelectables(...).

  2. AbstractGuiGame.java — base impl (all concrete GUIs inherit it: CMatchUI, MatchController mobile, RemoteClientGuiGame,
  HeadlessNetworkGuiGame)
  private int selectionMin = 0, selectionMax = 0;

  public void setSelectionRange(int min, int max) { selectionMin = min; selectionMax = max; }
  public int getSelectionMin() { return selectionMin; }

  // inside clearSelectables():
  selectionMin = 0; selectionMax = 0;

  3. ProtocolMethod.java — register for the wire
  setSelectionRange (Mode.SERVER, Void.TYPE, Integer.TYPE, Integer.TYPE),

  4. InputSelectEntitiesFromList.java — publish min/max when a selection starts
  getController().getGui().setSelectables(vCards);
  getController().getGui().setSelectionRange(min, max);   // new

  5. FloatingZone.java — new Ctrl+0 branch in dispatchHotkey
  if (digit == 0) {
      final int need = fz.getMatchUI().getSelectionMin();
      if (need < 1) return false;
      int picked = 0;
      for (final CardPanel panel : fz.getCardPanels()) {
          if (picked >= need) break;
          if (!fz.getMatchUI().isSelectable(panel.getCard())) continue;
          fz.getMatchUI().getGameController().selectCard(panel.getCard(), null,
                  new MouseTriggerEvent(MouseEvent.BUTTON1, 0, 0));
          picked++;
      }
      return true;
  }
  (Logic folded into the existing digit block — we'd just relax the digit < 1 lower bound.)
```

MostCromulent:
> If the revealed zone is prompting the user to choose something (e.g. some number of cards to discard), maybe the prompt might replace the usual title text? Or does it already seem visible enough in the usual prompt space?

Yeah I think this is worthwhile, otherwise its not immediately obvious what you're expected to do in window unless you cross reference with the prompt.

Screenshot 2026-05-13 073019

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1140,11 +1154,20 @@ public CardCollectionView orderMoveToZoneList(final CardCollectionView cards, fi
...
+                                                       final CardCollection valid, final int min, final int max,
+                                                       final CardCollectionView visibleToChooser) {
         boolean optional = min == 0;
 
         if (p != player) {
-            tempShowCards(valid);
+            tempShowCards(visibleToChooser);
+            if (useSelectCardsInput(valid, sa)) {
+                final InputSelectCardsFromList inp = new InputSelectCardsFromList(this, min, max, valid, sa);
+                inp.setMessage(String.format(localizer.getMessage("lblChooseMinCardToDiscard"), optional ? max : min));
```
tool4ever, line 1165: imo it'd be better if we only use this specific part in the zone header, not the whole prompt text again

tool4ever:
Extra space isn't reclaimed when looking at zones normally after:
image

MostCromulent:
Potential follow-up: 

Forge currently processes a discard effect as two separate actions: reveal stage shows window with revealed cards, user clicks ok, then it repaints window for discard stage. I don't think there's any reason from a rules perspective why that two stage process is necessary? If the discard stage is able to display the revealed cards at same time (which it now does in this PR) this can potentially be consolidated into one action (at least on desktop). Would need to consider mobile UI.

tool4ever:
yea some API even call `tempShowCards` on their own
and there is also `DelayedReveal` which offers a more dedicated mechanism for this?

In general there should be substantial cleanup potential for redundant or almost useless code in these PlayerController methods, just need to carefully consider what each effect attempts to do...

and ideally some more Javadoc to explain less obvious concepts

tool4ever (approved):

PR #10661, opened 2026-05-11
Cancel End Turn auto-pass for playable post-blockers responses

Madwand99:
This branch fixes a UX issue where Forge could skip the legal priority window after blockers were declared. In Magic rules, players receive priority during the declare blockers step after blockers are assigned, so a defending player should be able to cast combat tricks or activate relevant abilities before combat damage.

The implementation updates human priority handling so End Turn auto-pass is interrupted during COMBAT_DECLARE_BLOCKERS only when it matters:

- the player has priority
- the player is actually being attacked
- the stack is empty
- the player has a payable non-mana spell or ability

This avoids waking unrelated players in multiplayer games and avoids stopping just because the player has mana abilities available. Spells that cost mana still count, but only if the mana cost can currently be paid.

Regression tests cover:

- mana abilities alone do not interrupt auto-pass
- unaffordable instant tricks do not interrupt auto-pass
- affordable instant tricks do interrupt auto-pass
- sorcery-speed cards do not interrupt during declare blockers
- uninvolved multiplayer opponents are not interrupted when another player is attacked

Verification:

mvn -pl forge-gui-desktop -am -Dtest=forge.player.PlayerControllerHumanTest -Dsurefire.failIfNoSpecifiedTests=false test

### DIFF
--- a/forge-gui-desktop/src/test/java/forge/player/PlayerControllerHumanTest.java
+++ b/forge-gui-desktop/src/test/java/forge/player/PlayerControllerHumanTest.java
@@ -1,0 +1,96 @@
+package forge.player;
+
+import forge.ai.simulation.SimulationTest;
+import forge.game.Game;
+import forge.game.card.Card;
+import forge.game.combat.Combat;
+import forge.game.phase.PhaseType;
+import forge.game.player.Player;
+import forge.game.zone.ZoneType;
+import org.testng.AssertJUnit;
+import org.testng.annotations.Test;
+
+public class PlayerControllerHumanTest extends SimulationTest {
+    @Test
+    public void declareBlockersAutoPassInterruptIgnoresManaAbilitiesOnly() {
+        Game game = initAndCreateGame();
+        Player player = game.getPlayers().get(1);
+
+        addCardToZone("Mountain", player, ZoneType.Battlefield);
+
+        setDeclareBlockersPriority(game, player);
+
+        AssertJUnit.assertFalse(PlayerControllerHuman.hasPlayableNonManaAbility(player));
+    }
+
+    @Test
+    public void declareBlockersAutoPassInterruptIgnoresUnaffordableInstantCombatTrick() {
+        Game game = initAndCreateGame();
+        Player player = game.getPlayers().get(1);
+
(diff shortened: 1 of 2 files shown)


MostCromulent:
1. there is already a PR under review (and nearing completion following coordination between multiple contributors) dealing with yields and auto-pass, including checking for available actions before offering priority and providing configurable interrupt options: https://github.com/Card-Forge/forge/pull/10606. this PR seems to duplicate that functionality with a less flexible and conflicting architecture

2. in any event this definitely doesn't need a unit test - see warning against AI coding agents introducing these here https://github.com/Card-Forge/forge/blob/master/CONTRIBUTING.md#using-ai-coding-agents

Madwand99:
I checked #10606 and agree that this branch overlaps with its yield/auto-pass architecture, so this implementation probably should not land independently if #10606 is expected to merge. However, I don’t think #10606 fully covers the specific behavior this branch is addressing. It interrupts on attackers declared, and available-action checks appear to be used for APINA/suggestions rather than as a DB/post-blockers End Turn interrupt gate. The desired behavior here is narrower: when the local player is attacked and End Turn would skip the declare blockers priority window, interrupt only if that player has a payable instant-speed action. I'll rework this as a follow-up on top of #10606 using its AvailableActions/YieldController infrastructure if and when that PR is merged, if needed. For now I'll close this.

MostCromulent:
welcome to suggest a PR once the yieldrework lands, though it seems the above might be conflating an interrupt condition vs APINA which are different concepts under that model

yield modes (end turn, clear stack, yield to phase marker) are the player saying "don't give me priority until X". these operate subject to configurable interrupt conditions (opponent casts spell / targets you with spell / etc), basically specific game events which are detected and give player priority back by cancelling yield state

APINA is a persistent toggle which reflects the player saying "don't bother me unless there is something useful i can actually do in my priority window".

so in your example of a player who wants to be interrupted if they can play a post-DB combat trick, you should just be able to enable APINA _without_ the end-turn yield and it will produce the desired result

i guess theoretically the post-DB check could be implemented as its own configurable interrupt condition but imo thats over-engineering: its basically introducing a new axis of interrupt preferences framed around "do an APINA check, but only for this one specific phase".

PR #10672, opened 2026-05-13
Add Commander bracket calculation and bracket views

### LINKED ISSUE #9112: Commander game changers
If this keeps evolving nicely Forge might try and use them for QoL features like calculating your deck bracket.

https://magic.wizards.com/en/news/announcements/commander-brackets-beta-update-october-21-2025

Madwand99:
This adds Commander bracket support to Forge deck browsing.

Implemented:

- Adds a CommanderBracketCalculator that reads bracket data from forge-gui/res/lists.
- Adds a Bracket column for Commander deck browsers, including local decks, Commander precons, Commander net decks, and the deck editor Commander browser.
- Adds a Bracket View mode that explains why a deck receives its suggested minimum bracket.
- Adds Bracket View support to the single-deck viewer opened from Commander deck selection.
- Localizes the bracket column, view name, and explanation text across Forge’s bundled language files.
- Uses Forge’s standard tiled background and text styling for the bracket explanation view.
- Adds a text-based B view button for Bracket View.

The current calculator rules cover:

- Mass Land Denial
- Chained Extra Turns
- Extra Turns
- Game Changers
- Early/Late 2-card combos

Validation:

mvn -pl forge-gui-desktop -am -DskipTests compile passes.

Thanks to Codex for help with code and translations


Jetz72:
The 5 new lists you've added are 5 additional sets of data we'd have to maintain in order for this feature to continue working right as new sets come out. It also looks like the only way it can mechanically differentiate bracket 1 and bracket 2 is that bracket 2 has one or more extra-turn effects?

Madwand99:
@Jetz72 Correct. I've written Python scripts to scrape EDHREC and scryfall to automate this which I can share if needed, but yes the lists need maintenance. And yes, there isn't much to differentiate bracket 1-2 or bracket 4-5 that can be detected by a script. This is by design of the bracket system, but it's OK because most casual play of Commander is in brackets 2-4 anyway.

tool4ever:
I think we'll just have to make it clear their updating is in no way an expectable part of normal prerelease versions.

But still having direct access to this deck strength methodology inside Forge seems worth it, especially because we'll later also be able to reuse these lists (especially gamechangers) for AI evaluations or difficulty scaling in adventure mode.

I'd love to have a more solid approach to obtain combos as part of some official export, but that can be revisited in the future if that situation changes.

Great work!

Jetz72:
I wouldn't mind seeing the scripts - it's certainly better if we can automate the process of updating these lists. Can include them in the PR via the `forge-gui/tools/` directory.

Madwand99:
I've fixed some formatting and added an extra test to detect future formatting issues. This is optional, I can cut it if needed, but it's been useful so far.

@Jetz72 Sounds good, I'll look into cleaning my scripts up and adding them.

tool4ever:
Imo marked text here is wrong/redundant:
image

Madwand99:
@tool4ever Thanks! Hopefully this fixes the redundant text.

tool4ever (dismissed):

Jetz72:
As a non-blocking future consideration, I don't think the lines you commented out in the lists will be preserved in a future run of the scraper script. Do they cause issues if they're not commented out?

Madwand99:
@Jetz72 I don't think so, or at least typos etc. weren't causing a problem before I commented them out. The test I added will catch them, though. Note that the script doesn't create a perfect version of these files, for mass land denial I added additional cards.

tool4ever (approved):

PR #10676, opened 2026-05-14
Add Ctrl+0 select-min hotkey for FloatingZone

### LINKED ISSUE #10669: Card selection hotkeys follow-up variant
Looks like this shouldn't be too hard but would require small protocol change.

Currently `InputSelectCardsFromList` doesn't broadcast the min/max amount to network clients, it just waits to receive the choice. Would need to add that to the protocol. This is what happens with list dialogs already so there's precedent for it.

Changes involved:

```
  1. IGuiGame.java — new interface method
  void setSelectionRange(int min, int max);
  Placed next to setSelectables(...).

  2. AbstractGuiGame.java — base impl (all concrete GUIs inherit it: CMatchUI, MatchController mobile, RemoteClientGuiGame,
  HeadlessNetworkGuiGame)
  private int selectionMin = 0, selectionMax = 0;

  public void setSelectionRange(int min, int max) { selectionMin = min; selectionMax = max; }
  public int getSelectionMin() { return selectionMin; }

  // inside clearSelectables():
  selectionMin = 0; selectionMax = 0;

  3. ProtocolMethod.java — register for the wire
  setSelectionRange (Mode.SERVER, Void.TYPE, Integer.TYPE, Integer.TYPE),

  4. InputSelectEntitiesFromList.java — publish min/max when a selection starts
  getController().getGui().setSelectables(vCards);
  getController().getGui().setSelectionRange(min, max);   // new

  5. FloatingZone.java — new Ctrl+0 branch in dispatchHotkey
  if (digit == 0) {
      final int need = fz.getMatchUI().getSelectionMin();
      if (need < 1) return false;
      int picked = 0;
      for (final CardPanel panel : fz.getCardPanels()) {
          if (picked >= need) break;
          if (!fz.getMatchUI().isSelectable(panel.getCard())) continue;
          fz.getMatchUI().getGameController().selectCard(panel.getCard(), null,
                  new MouseTriggerEvent(MouseEvent.BUTTON1, 0, 0));
          picked++;
      }
      return true;
  }
  (Logic folded into the existing digit block — we'd just relax the digit < 1 lower bound.)
```

_Originally posted by @MostCromulent in https://github.com/Card-Forge/forge/pull/10660#discussion_r3229996435_

MostCromulent:
Closes Card-Forge/forge#10669.

## Summary

Adds Ctrl+0 to FloatingZone: auto-picks the first N selectables for the active prompt, where N is the selection's required minimum. Useful for "discard X", "choose X targets", and similar.

`setSelectables` carries min/max alongside the card set (no new protocol method — matches the shape `chooseEntitiesForEffect` already uses). Picks are batched into a single `selectCard` so the server processes them atomically; separate per-pick messages race on shared input state (the server spawns a thread per inbound protocol method).

---


MostCromulent:
Further expansion opportunity 🧠 

We've applied the CTRL hotkeys to `FloatingZone` selections but that doesn't touch selection effects in the player's own `VHand`. We could extend same hotkey and quick-select functions to the hand as well.

forge-gui-desktop/src/main/java/forge/view/arcane/FloatingZone.java
```diff
@@ -807,9 +838,22 @@ private void assignOwnHotkeyDigits(boolean clear) {
                 panel.setHotkeyDigit(0);
             }
         }
-        if (!clear) {
+        if (clear) {
+            hotkeyHint.setVisible(false);
```
tool4ever: hmn, toggling the hint with the labels now looks rather weird since it will also move the panels around slightly

forge-gui/src/main/java/forge/gui/interfaces/IGuiGame.java
```diff
@@ -250,7 +250,15 @@ default <T> List<T> many(final String title, final String topCaption, final int
 
     void setHighlighted(GameEntityView pv, boolean b);
 
-    void setSelectables(final Iterable<CardView> cards);
+    default void setSelectables(final Iterable<CardView> cards) {
+        setSelectables(cards, 0, 0);
```
tool4ever: I can't think of a reasonable case to allow this...

forge-gui-desktop/src/main/java/forge/view/arcane/FloatingZone.java
MostCromulent: Testing pre-this PR, hint only appears first time you press CTRL but space is reserved. Just to confirm, is your preferred behavior that hint text is always visible but card number labels only appear when CTRL is pressed?

tool4ever: right now yes otherwise it would have to be mentioned at least somewhere like user guide or the hotkey cheat sheet
and even then it might just be better to always serve as reminder instead here 🤔
I mean it doesn't cost much space anyway, though could maybe hijack the search bar ghost text or something - but not sure that'd still look as clear? 🤷‍♂️

MostCromulent: Nah I think it's fine to include down the bottom as a hint, putting in search bar gets too messy. Just wanted to make sure that was intention. Will go ahead and implement on that basis.

forge-gui/src/main/java/forge/gamemodes/match/input/InputSelectTargets.java
```diff
@@ -413,9 +414,25 @@ private void removeTarget(final GameEntity ge) {
...
+     * the live remaining target requirement, not a stale snapshot from
+     * the constructor.
+     */
+    private void publishSelectables() {
+        final int effMin = numTargets != null ? numTargets : sa.getMinTargets();
+        final int effMax = numTargets != null ? numTargets : sa.getMaxTargets();
+        final int taken = sa.getTargets().size();
+        final int remainMin = Math.max(0, effMin - taken);
+        final int remainMax = Math.max(0, effMax - taken);
+        getController().getGui().setSelectables(CardView.getCollection(choices), remainMin, remainMax);
```
tool4ever: ok...so you want to extend scope from choices to targeting too
I guess that's fine so it covers FloatingZone cases more completely

but:
- doesn't seem to work yet, only select 1 card always
- can't the client already decide on its own how much selection progress is currently made anyway
- otherwise that's missing from choices too, since using the hotkey there would currently act as a toggle between partial selections

tool4ever:
> We've applied the CTRL hotkeys to `FloatingZone` selections but that doesn't touch selection effects in the player's own `VHand`. We could extend same hotkey and quick-select functions to the hand as well.

Support for docked tabs will certainly increase size a good amount, so better in a follow-up - maybe based on user feedback:
this could mean also aligning the behaviour so it only opens an extra floating window if no docked tab for that zone is already available? 🤔
In theory could also check if user has his own hand in floating mode and then just prefer that or something but maybe only if we ever actually allow hiding the tab

forge-gui-desktop/src/main/java/forge/view/arcane/FloatingZone.java
tool4ever: hmn, if the hint now wraps to two lines which is more likely the redrawing somehow seems to be way more noticeable when selecting :/

MostCromulent: Visual stutter seems to be downstream of redrawing the HTML text wrapping - have moved the CTRL+0 hint, when applicable, to a 2nd row instead (which you'd likely have with wrapping anyway) and looks like that solves it.

tool4ever:
Well I also got rid of the dynamic aspect since visually it could still shift around a few pixels + we have the prompt for that anyway so no reason for complicating things further
(yes choices prompt text isn't always dynamic yet but that can be fixed later instead)

tool4ever (approved):

PR #10680, opened 2026-05-15
Add playable card highlights

autumnmyst:
Adds a configurable outline on cards the local player can currently play or activate. Off by default; toggled under "Highlight Actionable Cards" in desktop preferences and mobile settings.

Sources the highlight set from the same `AvailableActions` heuristic APINA reads, so highlights and the auto-pass decision share one scan per priority window.

During cost payment the AI heuristic isn't meaningful, the player isalready paying, so highlights switch to marking any card with a playable mana ability, scanned across Hand/Battlefield/Graveyard/Exile/ Command. For declare-attackers and declare-blockers the heuristic isn't used either; combat legality is deterministic, so candidates come from `CombatUtil.canAttack` (undeclared, legal attackers) and `CombatUtil.canBlock` (defenders that can block at least one current attacker). Both combat sets refresh after each click so highlights track the in-progress declaration.

## Architecture

- `AvailableActions` (forge-ai) gains `collectActionable(player, timeoutMs)`, a full-scan sibling of `compute(...)`. Same `canPayManaCost` + `isFullyTargetable` predicates over Hand / Battlefield / Flashback. Timeout fallback over-highlights.
- `PlayerControllerHuman.chooseSpellAbilityToPlay` runs the scan once per priority window when APINA or highlights need it, caches the result, and derives APINA's `hasAvailableActions` from `!actionable.isEmpty()` against the same set.
- Per-input push API on PCH: `pushActionableCards`, `pushAttackerCandidates`, `pushBlockerCandidates`. Lifecycle: push on `showMessage`, clear on `onStop`, re-push after each click in combat and cost-payment inputs.
- GUI plumbing: `IGuiGame.setWeaklySelectable` / `clearWeaklySelectable`, stored on `AbstractGuiGame`. `RemoteClientGuiGame` ships the set over two new `ProtocolMethod` entries. `CardPanel` (desktop) and `CardRenderer` (mobile) draw a thin outline when the pref is on.
- Prefs: `UI_SHOW_ACTIONABLE_HIGHLIGHTS` (default off) and `UI_ACTIONABLE_HIGHLIGHT_COLOR` (default `66CCFF`). Color input is a case-insensitive 6-char RGB hex, normalized to uppercase on save.


forge-ai/src/main/java/forge/ai/AvailableActions.java
```diff
@@ -53,6 +56,72 @@ public static boolean compute(Player player, long timeoutMs) {
...
     }
 
+    /** Full-scan sibling of {@link #compute}. On timeout, remaining cards are
+     *  marked actionable so the user sees extra highlights rather than missing some. */
+    public static Set<CardView> collectActionable(Player player, long timeoutMs) {
+        long deadlineNanos = System.nanoTime() + timeoutMs * 1_000_000L;
+        Set<CardView> actionable = new HashSet<>();
+
+        for (Card card : sortedCardsIn(player, ZoneType.Hand)) {
+            if (checkTimeout(deadlineNanos, timeoutMs)) {
```
tehdiplomat: This feels very repetitive. Can we just have a single array that has all three zones and the first way to sort is based off the zone its in and a single loop.

forge-ai/src/main/java/forge/ai/AvailableActions.java
```diff
@@ -53,6 +56,72 @@ public static boolean compute(Player player, long timeoutMs) {
...
+    }
+
+    private static boolean cardHasActionableSpell(Card card, Player player) {
+        for (SpellAbility sa : card.getAllPossibleAbilities(player, true)) {
+            if (sa.isSpell()) {
+                if (canAfford(sa, player) && ComputerUtilAbility.isFullyTargetable(sa)) {
+                    return true;
+                }
+            } else if (sa.isLandAbility()) {
+                return true;
```
tehdiplomat: What if you've already played all your land drops this turn?

forge-ai/src/main/java/forge/ai/AvailableActions.java
```diff
@@ -53,6 +56,72 @@ public static boolean compute(Player player, long timeoutMs) {
...
+            if (!sa.isManaAbility() && canAfford(sa, player)
+                    && ComputerUtilAbility.isFullyTargetable(sa)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /** Timeout fallback: mark every scannable card so the user doesn't miss highlights. */
+    private static void addAllRemaining(Set<CardView> actionable, Player player) {
```
tehdiplomat: Does this add cards to the array that have already been added (and may have already been ruled as not actionable?)

forge-ai/src/main/java/forge/ai/AvailableActions.java
autumnmyst: Yes I agree.

There’s another thing I wanted to change, I want move this to reuse the APINA check flow but without the early break so as to iterate everything, since currently I basically have rewritten the same search process but this time I don’t exit early if highlights are on. Better to consolidate the code.

forge-ai/src/main/java/forge/ai/AvailableActions.java
autumnmyst: Yeah, wasn't sure how to handle timeout. Might be better to highlight everything to be consistent in showing timeout has been reached, but we might as well skip highlighting the things we know we can't play.

forge-ai/src/main/java/forge/ai/AvailableActions.java
autumnmyst: That's caught by getAllPossibleAbilities() since remove unplayable is true

mermerico:
I'm very excited for this, it's a big quality of life win! In my local prototype implementation I had a highlight specifically for the cards auto-pay is going to use. That made it easier to know when I could trust auto-pay and made me use it more often. Is that something that you would be interested in putting in this PR or should it be proposed as a separate PR?

tool4ever:
@mermerico
separate PR please, also see #5112 for the bigger picture

forge-gui-mobile/src/forge/card/CardRenderer.java
```diff
@@ -68,6 +68,22 @@ public enum CardStackPosition {
         BehindVert
     }
 
+    private static final Color DEFAULT_ACTIONABLE_HIGHLIGHT_COLOR = FSkinColor.fromRGB(0x66, 0xCC, 0xFF);
```
tool4ever: don't add extra default fields, just grab it with `getPrefDefault` instead

forge-ai/src/main/java/forge/ai/AvailableActions.java
```diff
@@ -2,66 +2,114 @@
...
         }
+        return actionable;
+    }
 
-        // Not sorted: activation costs are per-ability, not the permanent's CMC.
-        for (Card card : player.getCardsIn(ZoneType.Battlefield)) {
-            for (SpellAbility sa : card.getAllPossibleAbilities(player, true)) {
-                if (checkTimeout(deadlineNanos, timeoutMs)) return true;
-                if (!sa.isManaAbility() && canAfford(sa, player) && ComputerUtilAbility.isFullyTargetable(sa)) {
+    private static boolean cardHasActionableSpell(Card card, Player player) {
```
tool4ever: don't see the point for this extra method, should just be handled with the one below since that wouldn't need extra spell filtering anyway?
\+ you'd miss abilities like Channel this way

autumnmyst: Good point, I think this was also a gap in APINA too since I based this off the same per-zone predicates. For cards in hand, the check only returned true for spells and land abilities. Will fix this.

MostCromulent:
Does this work correctly for network play? 

Remote clients will need a way to tell server they have highlights enabled so server runs the longer AvailableActions scan, otherwise it will just use whatever the server hosts preference is.

I think preference probably needs to be threaded through same start of match client -> server prefs seed packet as yield settings in YieldController, and PlayerControllerHuman needs to consult the yield controller-backed per-player preference when doing the scan rather than host pref. 

I can deal with that in quick follow-up PR if we would rather just get this landed first.

tool4ever:
Could also look into using `JColorChooser` in a follow-up

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1522,17 +1522,127 @@ public void declareBlockers(final Player defender, final Combat combat) {
         getGui().updateAutoPassPrompt();
     }
 
+    private static final ZoneType[] ACTIONABLE_PAYMENT_ZONES = new ZoneType[] {
+        ZoneType.Hand, ZoneType.Battlefield, ZoneType.Graveyard, ZoneType.Exile, ZoneType.Command
```
tool4ever, line 1554: why are we checking all these other zones? it's nearly impossible for a mana ability to exist there

autumnmyst: Just to be comprehensive. Jack-o'-Lantern has a mana ability from the graveyard, and spirit guides have mana abilities in hand. At that point I figured we might as well check all zones for posterity.

forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
```diff
@@ -379,9 +379,23 @@ protected final void paintComponent(final Graphics g) {
             g2d.setColor(Color.WHITE);
             final int ins = 1;
             g2d.fillRoundRect(cardXOffset+ins, cardYOffset+ins, cardWidth-ins*2, cardHeight-ins*2, cornerSize-ins, cornerSize-ins);
+        } else if (isPreferenceEnabled(FPref.UI_SHOW_ACTIONABLE_HIGHLIGHTS) && matchUI.isWeaklySelectable(getCard())) {
```
tool4ever, line 385: did you really test this...?
I don't actually see the border ingame, maybe something got slightly moved some time ago to make those dimensions you copied from incorrect 🤔

autumnmyst:
> Does this work correctly for network play?
> 
> Remote clients will need a way to tell server they have highlights enabled so server runs the longer AvailableActions scan, otherwise it will just use whatever the server hosts preference is.
> 
> I think preference probably needs to be threaded through same start of match client -> server prefs seed packet as yield settings in YieldController, and PlayerControllerHuman needs to consult the yield controller-backed per-player preference when doing the scan rather than host pref.
> 
> I can deal with that in quick follow-up PR if we would rather just get this landed first.

That's true, good call. Currently host-off, client-on doesn't show anything to the client. All the other combinations should work though, since so long as the host has it on it will calculate highlights for each local player instance. If I can fix it with a small change I'll do it, but otherwise you're welcome to fix it in a follow up.

forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
autumnmyst: As in you don't see the highlights at all?

autumnmyst: If you have the preference on, you should see blue borders around the playable cards

tool4ever: well it works in mobile, maybe you only checked there unless something specific is required for it to not draw correctly?

autumnmyst: ~~Ah, it's the last change I made removing the hardcoded default vs preference default. I had changed my color from default so I didn't notice it, but if you never touch the color the default doesn't properly set in preferences. Should be a small fix~~

autumnmyst: Wait, actually no it should default just fine. Unless you manually delete the hex color from your preferences, then it won't default, but I think that's intentional to rely on the default not a hardcoded one.

tool4ever: why resolve, this isn't really fixed yet?

autumnmyst: Sorry, I thought from your discord msg that a render black borders bug was the culprit. If you use default settings do you see this issue? I'm not sure how to reproduce it for myself to debug.

tool4ever: oh right, I meant the preference with that name - it's actually right above your new one

while it is enabled by default and it's not really your code it would still be better to fix all broken borders in that method as a bonus

MostCromulent:
Should add a checkbox for Highlight actionable cards to the in-match Display submenu, so you don't always need to go to preferences to enable/disable the setting.

forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
autumnmyst: done, now highlight borders render when black borders disabled (and they look even better that way honestly) also other highlights like hover and selection highlights work.

autumnmyst:
> Should add a checkbox for Highlight actionable cards to the in-match Display submenu, so you don't always need to go to preferences to enable/disable the setting.

Done, added an in-game toggle under Display > Highlight Actionable Cards

One thing to note is it doesn't refresh upon toggling, the reason I didn't is 1. it's a very minor edge case and it will refresh on the next game action, a fix would be a lot of LoC added to the PR 2. I'm guessing whatever plumbing needs to be done for online games needs touch this path anyways, so it would be a lot of changes that would then need to be updated again in the fix for remote client highlights being independent of host highlights.

tool4ever (approved):
seems ok as groundwork

PR #10681, opened 2026-05-15
Dock rework: configurable layout, show/hide, new icons

MostCromulent:
Default view
Screenshot 2026-05-16 075351
In order: auto-pass, yield settings, end turn, alpha strike, auto-yield and triggers, targeting overlay, deck list

On right-click:
Screenshot 2026-05-16 075421

## Features

- Drag-and-drop to reorder dock buttons; right-click to hide/show/reset to defaults.
- Redesigned monochrome icons sourced from [game-icons.net](https://game-icons.net) (CC BY 3.0) — readable but fantasy-adjacent, avoids the Windows 98 look. Registered as new `ICO_DOCK_*` slots so the dock aesthetic doesn't bleed into mobile menus, the deck editor, or the desktop match-screen tab that share the original icon slots.
- Easy to extend: adding a new dock button = register new icon + one row in the `DockButtonId` enum + one command lambda in `CDock.initialize`.

## Technical implementation

- `VDock` is data-driven by a typed `DockButtonId` enum carrying icon, label key, and default visibility. One `EnumMap` holds the buttons; iteration drives build / right-click / drag / relayout. Commands wired via `Map.of(...)` in `CDock.initialize`.
- Layout (visibility + order) persists in FPref `UI_DOCK_LAYOUT`. New shipped buttons added in later builds insert after their predecessor in the saved layout.
- Drag-reorder uses a translucent `JLayeredPane.DRAG_LAYER` ghost; `FMouseAdapter`'s 3px threshold separates drag from click so reordering doesn't fire the button action.
- `DockButton.setActive(boolean)` paints a rounded tile background at runtime in a shared dock-highlight gold, so any active state renders identically without per-state sprite slots.
- Targeting arc state is pref-backed: `CDock.getArcState()` reads `UI_TARGETING_OVERLAY` on demand; one `refresh()` driven by pref reads replaces the previous per-button refresh methods. The dock targeting button swaps to `ICO_ARCSON` (filled) when fully on so the shape signals the state alongside the tile colour.

## Alternative considered: XML layout

An earlier development iteration stored the dock layout in a dedicated `dock.xml` (shipped + per-user copy, mirroring `match.xml`) with a nested `DockLayoutIO` class for parsing, serialization, merge-from-shipped, and debounced saves. It worked but added ~200 lines of XML I/O code (StAX reader/writer, event factories, attribute handling, file lifecycle) for a UX indistinguishable from just storing the layout as an FPref string. The current FPref-backed approach keeps the schema-evolution behaviour (new shipped buttons insert after their predecessor) in a few dozen lines of plain string parsing.

---


docs/Credit-and-Thanks.md
```diff
@@ -32,4 +32,5 @@ Some of the sprites and tiles that are used in Adventure mode come from external
 1. Castle = Horus - Brandon Fiechter & Derek Fiechter
 
 ## Other sprites and icons
-1. Clipboard icon by Dimitry Miroliubov at [Flaticon](https://www.flaticon.com/free-icon/clipboard_340068)
\ No newline at end of file
+1. Clipboard icon by Dimitry Miroliubov at [Flaticon](https://www.flaticon.com/free-icon/clipboard_340068)
+1. Dock-bar icons by [Lorc](https://lorcblog.blogspot.com/), [Delapouite](https://delapouite.com/), and Skoll, sourced from [game-icons.net](https://game-icons.net/), licensed under [CC BY 3.0](https://creativecommons.org/licenses/by/3.0/).
```
tool4ever: huh I guess GitHub auto increases those numbers so you don't have to count yourself? 😆
though it still looks weird in plain view imo

while we're here not sure if only dashes would be better or we should at least manually number them? 🤔

forge-gui/res/skins/default/sprite_icons.png
tool4ever, line 1: it seems like you're also overwriting multiple unrelated icons here...

MostCromulent: Hmm, you're right. Let me have another try at the sprite sheet.

forge-gui-desktop/src/main/java/forge/screens/match/controllers/CDock.java
```diff
@@ -42,118 +43,92 @@
...
+        targeting.setActive(arcs != ArcState.OFF);
+        targeting.setImage(FSkin.getIcon(arcs == ArcState.ON ? FSkinProp.ICO_ARCSON : FSkinProp.ICO_ARCSOFF));
+        targeting.setToolTipText(Localizer.getInstance().getMessage(targetingTooltipKey(arcs)));
     }
 
-    public void refreshAutoPassToggled() {
-        view.getBtnAutoPass().setToggled(FModel.getPreferences().getPrefBoolean(FPref.YIELD_AUTO_PASS_NO_ACTIONS));
+    private static String targetingTooltipKey(final ArcState arcs) {
+        return switch (arcs) {
+            case OFF -> "lblTargetingArcsOff";
```
tool4ever: should we remove these extra dock labels:
could easily reuse the already existing menu ones by just prepending with `lblTargetingArcs`...?

tool4ever (approved):

PR #10683, opened 2026-05-15
Option: Auto-apply remembered ability order

MostCromulent:
Screenshot 2026-05-16 092220

Forge prompts the user to order simultaneous abilities every time they fire, even after the user already ordered the same combo earlier in the game. 

The previous pref (Preselect Previous Ability Order) only prefilled the dialog — it still required confirmation. Players who'd decided their preferred order had no way to skip the repeat prompt.

This PR replaces the boolean pref with a tri-state Remember Ability Orders (Off / Preselect / Auto-apply). 
- Off always shows a fresh dialog
- Preselect keeps the old prefill behaviour
- Auto-apply builds on preselect by silently re-applying the saved order on repeat firings. 

A Clear Ability Orders entry in the Game menu (visible only in Auto mode) wipes the saved orders within the current Game. 

The read site is routed through YieldReworkV2's per-PCH pref overlay so remote clients in network play honour their own setting, not the host's.

Applies same concept as closed https://github.com/Card-Forge/forge/pull/10612 but with per-PCH architecture.

---


MostCromulent:
This is set up to apply the remembered order on a per-game basis with no multigame persistence. 

Simultaneous ability orders tend to develop over the course of a game as new cards enter the battlefield with different trigger abilities; at start of game you might have only 2 simultaneous triggers and that grows over time, with player having to re-confirm preferred order each time a new ability enters the mix. 

In practice the game eventually settles into a rhythm with the same triggers firing repeatedly; that's when this feature becomes useful and saves you time by applying saved order automatically.

I considered whether to wire this up to the full per-card / per-match / per-install persistence model + UI menu used for yes/no triggers and autoyields however imo its over-engineering for this use case.

Storing those different iterations and combos persistently and allowing user to configure saved ordering prefs for each possible iteration and combo they encountered is a UI nightmare and - even if solvable - the saved prefs and UI menu needed to control it would be a monster. imo not worth it.

tool4ever:
Thoughts:
- 1.a) the previous attempt did have a slight advantage by instead of combofying the preference using an individual checkbox while ordering so that you can have a mix of both (if that was a button could even include confirm so number of clicks stays the same)
- 1.b) though passing an individual flag this way might make it more complicated for netplay 🤔
- 2.a) persistence could still be nice - but only if we come up with some sort of algorithm which applies heuristics for reusing partial orders without an exact match 🚀
- 2.b) remember my talk about power users and not having to offer UI for *everything* - as long as files like `auto-yields.dat` are in text format with a non trivial amount of entries an external file editor will always be superior for fine tuning anyway

not sure either is worth the effort or this minor improvement is enough and we just take it...

feel free to make the final call though perhaps @Madwand99 also wants to chime in?

Madwand99:
I like the ability to choose to skip ability order decisions when possible. I agree with @MostCromulent that there is no need to remember beyond a given game. The main thing I'm not sure about is UI. First, the UI for applying ability orders is itself hard to understand. What does it mean? Which ability goes off first? (this is where an Undo would be handy). Then once I've figured that out, what would be easier: to go over to options, select Preselect or Auto-apply, come back, and continue the game, or maybe it's easier just to select a check-box on the auto-apply dialog to "Remember this decision" or similar? I don't have a good answer here.

MostCromulent:
> select Preselect or Auto-apply, come back, and continue the game, or maybe it's easier just to select a check-box on the auto-apply dialog to "Remember this decision" or similar? I don't have a good answer here.

I don't have any in-principle issue with a "Remember this decision" checkbox in the selector rather than a universal Auto-pass preference. Will look into network implications of passing preferences on a per-ability-order basis like this.

Zooming out a bit: what is the point of the "Preselect Previous Ability Order" preference at all? Since user is shown the ordering dialog anyway, why isn't this just the default ordering behavior and user gets the chance to re-order when they are re-shown the dialog if they don't want to keep previous selected order?

With that in mind (and subject to any adverse network issues) we could:
1. Retire the "Preselect Previous Ability Order" preference and just bake it in.
2. Add some sort of "Remember this decision" checkbox to the selection UI.
3. This gets rid of the entire Remember Ability Orders tri-state preference from the PR.

Madwand99:
> Will look into network implications of passing preferences on a per-ability-order basis like this.

Just curious, why does this preference need to be passed over the network anyway? Wouldn't such a preference be unique per each client? It's just automating the choice on the client side.

tool4ever:
it saves traffic if host knows he doesn't even have to ask client

MostCromulent:
Very brief high-level summary:

One would assume there are only two relevant network instances: the server and the client. Actually there are **three**: the server, a server-side proxy for the remote client, and the remote client itself.

In ordinary local games you as the user have access the full game state because the engine is running on your machine. In network multiplayer, only the server has access to the engine and not all of its information can be transferred over the network. This means remote clients have to be able to operate with a much more limited set of information.

This means **even for remote clients** most of the processing happens on the server-side remote proxy. Your local machine GUI basically sends the necessary inputs to the server, and your remote proxy on server side applies those inputs and the server sends the results back to your local machine.

This means that even for client-specific preferences (yield state, enabled phase stops, "remember this decision" checkbox in our example), network serialization is important because it's the only way your server-side remote proxy finds out what you want to do, other than you being asked to provide an input prompt every time.

tool4ever:
> Zooming out a bit: what is the point of the "Preselect Previous Ability Order" preference at all? Since user is shown the ordering dialog anyway, why isn't this just the default ordering behavior and user gets the chance to re-order when they are re-shown the dialog if they don't want to keep previous selected order?

well I guess if you're playing a theoretical deck where common trigger A before B and vice versa are both useful depending on context starting with a preselection adds an additional mouse click if you want to swap it around:
which I'd say is acceptable but at the same time wasn't the initial idea for this also only about saving a single click? 😛
though maybe the former is even less likely 🤷‍♂️

I suppose if this PR manages to offer a way to suppress the dialog completely that will be the best outcome - this will tie in nicely with not halting game if a player is auto-yielding

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1986,84 +1986,100 @@ public boolean payCostDuringRoll(final Cost cost, final SpellAbility sa) {
...
 
-            // if current SA isn't a trigger and it uses Targeting, try to show prompt
             if (currentSa.isTrigger()) {
                 needPrompt |= currentSa.getTrigger().hasParam("OrderDuplicates");
             } else if (currentSa.usesTargeting()) {
                 needPrompt = true;
             }
-            if (!needPrompt && !saStr.equals(firstStr)) {
-                // prompt by default unless all abilities are the same
+            if (!needPrompt && !saLookup.equals(firstLookup)) {
```
tool4ever: hmn sneaky logic change so only different triggering objects would no longer be enough to prompt...

tool4ever (approved):

PR #10693, opened 2026-05-17
Network cleanup

MostCromulent:
## Summary

A batch of cleanups to the netplay layer surfaced by a recent review of the package by Claude:

- Duplicate logic lifted into shared helpers — handleYieldMarkerToggle into AbstractGuiGame, hashing extracted from NetworkChecksumUtil, reconnect-helper iteration deduped, RemoteClientGuiGame ownership moved onto RemoteClient.
- API shapes tightened — setHighlighted now takes Iterable to mirror setSelectables; NetEvent.updateForClient defaults to no-op.
- Dead surface removed — empty marker interfaces (IToClient/IToServer), an unreachable branch in LobbyInputHandler, an unused FGameClient parameter, dead null guards in YieldController.
- Diagnostic correctness — clear lastChecksumBreakdown after onResyncRequested logs it, so a second resync doesn't log stale state.

See individual commit messages for the rationale behind each change.

---


forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
tool4ever: is there even a point to keep this extra handler:
seems like the logic would fit better together with `RegisterClientHandler` anyway?

forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
```diff
@@ -433,25 +431,40 @@ public IGuiGame getGui(final int index) {
...
-                if (client.getIndex() == index) {
-                    RemoteClientGuiGame newGui = new RemoteClientGuiGame(client);
-                    playerGuis.put(index, newGui);
-                    return newGui;
+            final RemoteClient client = findClientByIndex(index);
+            if (client != null) {
+                RemoteClientGuiGame gui = client.getGui();
+                if (gui == null) {
+                    gui = new RemoteClientGuiGame(client);
+                    client.setGui(gui);
```
tool4ever: if it's only used once and essential for operation just add it to constructor instead

forge-gui/src/main/java/forge/gamemodes/net/client/NetGameController.java
```diff
@@ -28,7 +29,7 @@ public class NetGameController implements IGameController {
     /** Source of truth for this client's yield state (auto-yields, trigger decisions, markers, skip-phase, etc.). */
     private final YieldController yieldController = new YieldController(null);
 
-    public NetGameController(final IToServer server) {
+    public NetGameController(final IRemote server) {
```
tool4ever: imo the interface split still brought some type safety?
because now this contructor would also accept `RemoteClient` :/

forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
```diff
@@ -1393,31 +1391,10 @@ private void actuateMatchPreferences() {
...
-        YieldMarker existing = controller.getYieldController().getAutoPassUntilMarker();
-        boolean clickedSameLabel = existing != null
-                && phaseOwner.equals(existing.getPhaseOwner())
-                && phase == existing.getPhase();
-        if (clickedSameLabel) {
-            controller.sendYieldUpdate(new YieldUpdate.ClearMarker(local));
-        } else {
-            // Setting a marker implies we want to stop here — un-skip the cell
-            // so the marker can fire (skip-phase pref + marker would skip past).
+        handleYieldMarkerToggle(phaseOwner, phase, () -> {
```
tool4ever: just call this method directly from the PhaseLabel classes now?

tool4ever (approved):

tool4ever:
I guess the room key thing only really makes sense if we ever end up with some sort of lobby directory...

PR #10701, opened 2026-05-17
Fix network CME from EDT walking game state

MostCromulent:
## Summary

User-reported host crash on local network play. The supplied logs show two distinct symptoms:

1. **Host crashes** mid-game with `ConcurrentModificationException`. The remote player is left looking at "Waiting for opponent…" indefinitely.
2. **Persistent client resync requests.** The client repeatedly detects sampled-checksum mismatches and asks the server for a full resync.

**Root cause (both symptoms).** Protocol methods on `RemoteClientGuiGame` are being invoked from the EDT, and those methods read the game state (walking the trackable graph to bundle dirty deltas with the protocol message). The graph is single-mutator (game thread) by design — no synchronisation. An EDT read concurrent with the game thread's mutation of `Tracker.delayedPropChanges` produced the CME through `Tracker.getDelayedPropsFor` → `mergeDelayedProps` → `collectDeltas`. The resync storm is the same race expressed differently: `collectDeltas` also computes a periodic sampled checksum from the same graph; a torn checksum from an EDT read drives the client to request a resync, which kicks off the next cycle. The crashes are the unlucky cases where the EDT read hits `ArrayList` iteration during concurrent modification.

The EDT entry point is `AbstractGuiGame.awaitNextInput`'s 250 ms Timer runnable, which calls three protocol methods on `RemoteClientGuiGame`: `showPromptMessage`, `updateButtons`, and `showWaitingTimer`.

**Fix.** Those three methods now use a new no-flush helper that ships the protocol message via the thread-safe Netty pipeline without walking the trackable graph. `showPromptMessage` additionally drops its explicit `updateGameView()` call. Their API contracts carry only player IDs, strings, and button flags — they never needed state bundling. Reachability from a non-game thread is itself the signal that the message doesn't depend on fresh state.

The three protocol-send helpers in `RemoteClientGuiGame` are also renamed so callsites can pick the right one by name:

| New | Was | Behaviour |
|---|---|---|
| `send` | (new) | Send only — no flush, safe from any thread |
| `flushAndSend` | `send` | Flush pending events, then send (game-thread-only) |
| `flushAndSendAndWait` | `sendAndWait` | Flush, send, block for reply (game-thread-only) |

## Why this is safe

- **EDT race eliminated.** The three methods no longer touch the trackable graph; no race possible.
- **Both observed bugs share a root cause** (EDT walking mutating state); removing the walk addresses both.
- **No semantic regression.** Args for the three methods are pure IDs/strings/flags — no state dependency. Any pending state events ride out on the next game-thread tick.
- **21 other callsites are pure mechanical renames** — same flush behaviour as before, under more honest names.

## Follow-up

This PR addresses the dangerous EDT cases — the paths where the race actually produced crashes and resync storms in the user-supplied logs. Further correctness work and broader cleanliness remain:

- **Latent races in adjacent non-game-thread paths.** `PlayerControllerHuman.tryUndoLastAction` (EDT) and `FServerManager.convertToAI`/`resumeAndResync` (Netty/Timer) all reach `handleGameEvents` or `updateGameView` from the wrong thread. Same shape as the race this PR fixes; not yet observed to crash, but the conditions exist.
- **Audit other `RemoteClientGuiGame` methods that pair `updateGameView()` with a protocol send** — some are state-dependent (keep the walk), others are UI-only (drop it). A per-method audit can move the don't-need-state ones onto the new no-flush `send` helper.
- **Audit `flushAndSend` callsites where `send` would suffice** — UI-only methods (`flashIncorrectAction`, `alertUser`, `enableOverlay`, …) don't need the flush.
- **Consider coalescing related UI updates into one protocol envelope.** The `awaitNextInput` runnable currently fires three independent messages for what is conceptually one UI state change; an envelope like `setAwaitingInputUI(prompt, buttons, waitingFor)` would atomically deliver the UI state with fewer round-trips and simpler ordering.

---

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/net/server/RemoteClientGuiGame.java
+++ b/forge-gui/src/main/java/forge/gamemodes/net/server/RemoteClientGuiGame.java
@@ -119,23 +119,34 @@ public class RemoteClientGuiGame extends NetworkGuiGame implements IHasForgeLog 
         }
     }
 
+    /**
+     * Send the protocol command directly, without flushing pending state.
+     * Safe from any thread. Use when the message carries only IDs and flags
+     * and doesn't depend on the client having fresh state first.
+     */
     private void send(final ProtocolMethod method, final Object... args) {
+        if (paused) { return; }
+        sender.send(method, args);
+    }
+
+    /**
+     * Send the protocol command, ordered after any pending events. Flushes
+     * the forwarder's event queue first (events ship their own bundled
+     * deltas), then sends. Game-thread-only. Use when this command depends
+     * on events the caller just fired arriving first.
+     */
+    private void flushAndSend(final ProtocolMethod method, final Object... args) {
         if (paused) { return; }
         flushPendingEvents();
         sender.send(method, args);
     }
 
     /**
-     * Send a protocol command without flushing pending state events.
-     * Use for commands reachable from non-game threads — the flush walks
-     * the trackable graph, which only the game thread may safely read.
+     * Flush pending state events, send the protocol command, and block
+     * until the client replies. Game-thread-only — both the flush and the
+     * blocking wait require it.
      */
-    private void forward(final ProtocolMethod method, final Object... args) {
-        if (paused) { return; }
-        sender.send(method, args);
-    }
-
-    private <T> T sendAndWait(final ProtocolMethod method, final Object... args) {
+    private <T> T flushAndSendAndWait(final ProtocolMethod method, final Object... args) {
         if (paused) { return null; }
         flushPendingEvents();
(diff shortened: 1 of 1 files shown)


MostCromulent:
@tool4ever re follow-up above:

Here is audit and proposed re-design of the helper methods: **[2026-05-18-remote-client-helper-audit.md](https://github.com/user-attachments/files/27936950/2026-05-18-remote-client-helper-audit.md)**

Could implement here if you prefer; otherwise we can get this PR out as a hotfix and I can roll that into either the network cleanup PR or a separate PR.

tool4ever (approved):
mapping for the new helpers looks good and can be done next

PR #10704, opened 2026-05-18
Fix: route hand reveal to prompt instead of popup

MostCromulent:
**Before**
image

**After**
Screenshot 2026-05-18 132045

- Per dev feedback on #10660, the FloatingZone reveal acknowledgement fired a modal `getGui().message()` dialog that interrupted input focus (Duress was the reported case).
- Swap the modal for the existing "OK / End Turn" prompt used during priority passes — OK acknowledges; End Turn auto-passes through end of turn, same semantic as the rest of the input bar.
- Uses `InputConfirm` directly with "OK"/"End Turn" labels — no existing `Input` covers this shape, and a single-call-site subclass would be over-engineering. Same call-site pattern as `chooseStartingPlayer` (PCH:833). End Turn calls `autoPassUntilEndOfTurn()` via `FThreads.invokeInEdtLater`, mirroring `InputPassPriority.onCancel`.
- Mobile unchanged: `useFloatingHandReveal` is gated on `!getGui().isLibgdxPort()`, so the libgdx port takes the existing names-list `else` branch.

---

### DIFF
--- a/forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
+++ b/forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
@@ -948,7 +948,12 @@ public class PlayerControllerHuman extends PlayerController implements IGameCont
                     getGui().updateZones(zonesToUpdate);
                     zonesShown[0] = getGui().tempShowZones(getLocalPlayerView(), zonesToUpdate);
                 });
-                getGui().message(fm);
+                final InputConfirm inp = new InputConfirm(this, fm,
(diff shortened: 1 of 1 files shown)


MostCromulent:
Will bring in reveal-stage skip for discard effects (i.e. you don't need the reveal because the revealed hand is visible when you select cards to discard anyway) as separate follow-up PR, still working out cleanest implementation. This solves immediate UI annoyance.

tool4ever (approved):

PR #10706, opened 2026-05-18
Fix: stop prompt eating player hand in spectator mode

MostCromulent:
## Summary

When spectating an AI vs AI match, the bottom hand cell would disappear and the prompt cell would expand to absorb the gap. `initMatch` set `allHands=true` only when every player was local, so spectators fell back to a per-player `mayViewAny` check — which returns false on an empty hand. Since `initMatch` runs before AI players have drawn their opening hand, no VHand instances get created, `loadLayout` silently drops the resulting `addDoc(VEmptyDoc)` calls, and `populate`'s gap-fill removes the empty hand cell.

Treat "no local players" as `allHands`. This reuses the same code path already exercised by hot-seat / pass-and-play matches (where every player is local), so the rendering and update logic is unchanged from a known-good case. It also doesn't widen card-visibility: spectators already pass the privacy model's vacuous "shown to any of zero viewers" check in `canBeShownToAny`, so they were always entitled to see hand contents — the bug was just that the cell hosting those hands never got built.

---

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/gui/framework/SRearrangingUtil.java
+++ b/forge-gui-desktop/src/main/java/forge/gui/framework/SRearrangingUtil.java
@@ -313,6 +313,26 @@ public final class SRearrangingUtil {
     /** The gap created by displaced panels must be filled.
      * from any side which shares corners with the gap.  */
     private static void fillGap() {
+        // First pass prefers a strategy that grows a player FIELD cell.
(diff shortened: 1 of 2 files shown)


tool4ever:
hmn, previously hands were still not displayed since spectators don't need to interact with them anyway
currently trying to find out why the match layout change can even cause this 🤔

tool4ever (approved):

PR #10707, opened 2026-05-18
Implement custom rebalanced draft formats for Old Border Shandalar

Agetian:
@vanja-ivancevic Can you please test that this works as desired in the Old Border world context? Also, are the names of the rebalanced draft blocks too long, perhaps? Maybe we should shorten them somehow? If you have ideas for a shorter title, please let me know :)


Agetian:
I'll try merging this in, it looks workable to me, with no other suggestions. If it ends up not working well for some reason in playtesting, we can update or revert.

PR #10709, opened 2026-05-18
Refresh SpellAbilityVariables after text change

tool4ever:


Hanmac (approved):

Hanmac:
I probably will think about if we could refactor these more 🤔

PR #10711, opened 2026-05-18
Add repeatable action macro recording for match play

Madwand99:
## Summary

This PR expands Forge’s match macro support from the older basic remembered-action behavior into a repeatable record/replay system suitable for common deterministic combo loops.

New behavior includes:

- Dock controls for recording and replaying macros
- Keyboard shortcuts for record, play once, and repeat playback
- Prompted repeat count when finishing a recording
- Replay support for:
  - card/player selection
  - activated abilities
  - priority passes
  - mana payments from pool
  - recorded future mana sources
  - color choices
  - mana-combination choices
  - modal choices
  - stack ordering
  - scry choices
  - confirmations
  - list/entity selections
- Safer cancellation when playback no longer matches the active prompt
- Optional replay debug logging with `-Dforge.macro.debug=true`
- Localized macro UI strings for the existing Forge language files

The local puzzle files used for testing are not included in this PR but are available on request.

## Testing

Build verification:

- `git diff --check`
- `mvn -pl forge-gui-desktop -am -DskipTests compile`

Manual validation was done with local puzzle scenarios covering a broad set of loop shapes and prompt types, including:

- Aggravated Assault + Reaver Cleaver
- Aurelia, the Warleader + Sword of Hearth and Home
- Sliver Queen + Ashnod's Altar + Impact Tremors
- Staff of Domination + Selvala, Heart of the Wilds
- Ghostly Flicker + Archaeomancer + Peregrine Drake
- Saffi Eriksdotter loop
- Retreat to Coralhelm / Scout Crab style land-entry loop
- Aluren + Cloudstone Curio + Kobolds
- Axebane Guardian + Umbral Mantle
- Basalt Monolith + Rings of Brighthearth
- Bloom Tender + Freed from the Real
- Devoted Druid + Vizier of Remedies
- Dockside Extortionist + Temur Sabertooth + Impact Tremors
- Earthcraft + Squirrel Nest
- Emiel the Blessed + Peregrine Drake + altar loop
- Exquisite Blood + Sanguine Bond
- Kitchen Finks + Vizier loops, including scry and sacrifice variants
- Goblin Bombardment loops
- Gravecrawler + altar + Impact Tremors
- Heliod + Walking Ballista
- Isochron Scepter + Dramatic Reversal
- Kiki-Jiki + Bell-Ringer
- Midnight Guard + Presence of Gond
- Mikaeus + Triskelion
- Mind Over Matter + Azami
- Niv-Mizzet + Curiosity
- Pili-Pala + Grand Architect
- Reveillark + Karmic Guide + altar
- Saheeli + Felidar Guardian + altar
- Tidespout Tyrant mana-rock loop
- Walking Ballista counter loop

These scenarios were used to exercise stack ordering, repeated priority passes, target replay, modal prompts, mana source replay, token identity changes, library searches, battlefield land selection, scry decisions, and replay cancellation behavior.

Thanks to Codex for help with code, translation, and extensive testing.


forge-gui/res/skins/default/rec-play.png
tool4ever, line 1: these don't really fit the dock icon look and maybe should just be added to the main spritesheet?

forge-gui/src/main/java/forge/sound/SoundSystem.java
```diff
@@ -481,4 +482,12 @@ private static List<String> collectProfiles(String subPath) {
     public void setIgnorePlayRequests(boolean ignorePlayRequests) {
         this.ignorePlayRequests = ignorePlayRequests;
     }
+
+    public boolean isSuppressSnapshotRestoredSound() {
+        return suppressSnapshotRestoredSound;
```
tool4ever: there's already an ignore method above
also if the macro causes snapshot restoring something seems wrong? 🤔

forge-gui-desktop/src/main/java/forge/view/KeyboardShortcutsDialog.java
```diff
@@ -55,17 +54,6 @@ public void mousePressed(final java.awt.event.MouseEvent e) {
                     field.setCodeString("");
                 }
             });
-            // Transfer focus away once a non-modifier key completes the binding.
```
tool4ever, line 58: I feel like random code removal would be best explained via its own commit

forge-gui/src/main/java/forge/player/RecordActionsMacroSystem.java
```diff
@@ -78,89 +269,944 @@ public void addRememberedAction(PlayerAction action) {
...
+        return sb.toString();
+    }
+
+    private String describeAction(final PlayerAction action) {
+        final StringBuilder sb = new StringBuilder(action.getClass().getSimpleName());
+        final GameEntityView view = action.getGameEntityView();
+        if (view != null) {
+            sb.append("(").append(view).append(")");
+        }
+        if (action instanceof ActivateAbilityAction activateAbilityAction) {
```
tool4ever: each action should just have its own method to provide this so you don't need all these manual checks

forge-gui/res/skins/default/rec-play.png
Madwand99: Done!

forge-gui/src/main/java/forge/sound/SoundSystem.java
Madwand99: Fixed! Note that I did this to remove the annoying "cancel" sound on some loops, but it's optional.

forge-gui-desktop/src/main/java/forge/view/KeyboardShortcutsDialog.java
Madwand99: This was not intended as unrelated cleanup. The focus-transfer behavior was moved into `KeyboardShortcutField` itself so shortcut capture behaves consistently in both the preferences pane and the shortcut dialog. After that, the dialog-local listener was redundant. Essentially, this change was needed so that it was possible to edit the shortcut keys (this might have been a preexisting bug but fixing it was needed for this feature to be fully functional).

forge-gui/src/main/java/forge/player/RecordActionsMacroSystem.java
Madwand99: Good call. Done!

forge-gui/res/skins/default/rec-play.png
tool4ever: getting better, but still slightly off + not transparent:
image

tool4ever: oh I guess you went for some different "pressed down" style for enabled
but wouldn't it look better if it was the same as the auto-pass one?

Madwand99: Yes, I went with a different style, for 2 main reasons:
1) I wanted 3 modes: on, off, and disabled. Current style only supports on/off as far as I can tell.
2) I find the yellow on/off style very ugly and hard to interpret/confusing. Buttons just make more sense to me.

MostCromulent: @Madwand99 the new dock icons are sourced from here: https://game-icons.net/. You should be able to find something in consistent style there.

The dock can support different icons to reflect different modes or states. See the targeting arrows button for an example of how this is coded: it has off (half icon, no background), hover (half icon, background), and always (full icon, background).

While I appreciate visual tastes will differ, we should probably try and maintain consistent visual style and UX in the dock as new buttons are added.

Madwand99: Sure, I replaced the Play icon. I wasn't able to find a Record icon that made sense to me so I kept the custom one.

I see what the targeting arrow does. It's actually not that different from what I'm doing with record/play, I'm just entirely relying on changes in the "icon" instead of adding yellow randomly. I do actually find the yellow makes the change to the targeting icon very hard to read. To my eyes, white+yellow appears very washed out as if the icon has been replaced by just a solid rectangle.

forge-game/src/main/java/forge/game/player/actions/StackOrderAction.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class StackOrderAction extends PlayerAction {
+    private final List<String> abilityDescriptions;
+
+    public StackOrderAction(final List<String> abilityDescriptions) {
+        super(null, "Order simultaneous abilities");
+        this.abilityDescriptions = new ArrayList<>(abilityDescriptions);
```
tool4ever: no reason to wrap this in a new object

---
the introduction of this action complicates the code quite a bit more:
but I guess there could be an argument that - even with the new improvements to skip the ordering dialog - complex combos could involve different orders between each step 🤔

In general instead of another huge size PR this could have been iterated over multiple ones with a cleaner architecture...
Anyway I'll be willing to merge as is mostly since this feature was in a pretty rough state and the changes are somewhat isolated

forge-gui/res/skins/default/rec-play.png
tool4ever: that's fair
personally I don't prefer either but I do think icons shouldn't use different styles

I wonder if it'd actually look better if all were done in this more button-like flavor instead 🤔
@MostCromulent ?

forge-game/src/main/java/forge/game/player/actions/StackOrderAction.java
Madwand99: Done!

Yeah if a combo is complex enough to need different orderings, then they would need to record the different orderings before playing. I haven't tested this corner case (there's an almost infinite number of things I haven't tested just due to the nature of the game).

I'd be happy to change my workflow in the future for PRs like this. I actually wasn't sure how big this was going to be until I was halfway done. If you like, I could try reaching out to you on Discord to discuss future ideas before I create a PR for them? If you have ideas for a cleaner architecture I could try to accommodate.

forge-gui/res/skins/default/rec-play.png
MostCromulent: I'm not wedded to the background highlight, it was just a convenience because there wasn't a clear alternative icon available to represent all available states while still intuitively representing the same function.

If we wanted to migrate current dock away from highlights we would need to find icons to represent Auto-pass ON / Auto-pass OFF, as well as another icon to represent Targeting Arcs completely off.

Perhaps something I can continue to look into as follow-up.

Madwand99: I could switch all the icons over to use my button system if desired, the code is already mostly there. Targeting arcs could I guess switch between two "pressed" states.

forge-game/src/main/java/forge/game/player/actions/StackOrderAction.java
tool4ever: sure let's try that next time 👍

forge-gui/res/skins/default/rec-play.png
tool4ever: It shouldn't need new icons because the "enabled" look is done by the class itself

so only the targeting arcs could keep their second one to switch from the standard to a third variant

MostCromulent: Just tested yours in-game. I'm not entirely convinced by the button aesthetic but that's more personal preference than any fundamental UX problem.

Potential middle-ground which avoids the highlighting, but takes inspiration from your play button:

Rather than representing toggle states with the highlighted background, indicate them by dimming/changing opacity of icon when disabled. That way we end up with:
- Auto-pass: off (dimmed), on (normal icon)
- Targeting arcs: off (dimmed half icon), hover (half icon), always (full icon).

MostCromulent: But if consensus is in favour of button aesthetic I'm not gonna go to the barricades over it. imo the flat icons just look a bit better/more modern.

Madwand99: I'm not against this for the other buttons, but it wouldn't work for the record/play icons. They should ideally be able to vary between on/off/disabled and I'm not sure I see a clear way to do that with the above limitation.

tool4ever: haha ok next idea:
maybe they don't have to be all identical after all?

-> what if we signal their capability by giving those that can be toggled the button style and the single press ones stay in their current form?

Madwand99: I'd be fine with that, but then the question should be asked: why not make them all buttons?

MostCromulent: I think we should aim for consistency one way or the other. 

I've mocked up the all-button look locally and it isn't actually too offensive visually. I'm fine with this.

Screenshot 2026-05-19 083443

My one additional suggestion would be rather than having the disabled "play macro" button as odd one out with no button styling, do something like apply the button styling but dim the icon to reflect disabled state. We can then apply same principle for disabled targeting arrows too, and get consistent styling across the board.

Screenshot 2026-05-19 085059

MostCromulent:
Worth considering whether [the wiki section on macros](https://github.com/Card-Forge/forge/blob/master/docs/User-Guide.md#repeatable-sequences-macros) needs an update to reflect changes here.

forge-gui/res/skins/default/rec-play.png
Madwand99: Looks good, I can implement this. One small change I might experiment with: disabled buttons might get a flat grey outline instead of the "pushed up" look.

Madwand99: Dockbuttons

I've made the changes: I think I like the new look. Also updated documentation.

MostCromulent: @Madwand99 have made a fix to the dock in https://github.com/Card-Forge/forge/pull/10718 and incorporated + registered your icons at same time; using same names and sprite locations. You should be able to just accept the master version of sprite sheet on a merge into this branch, will need to adjust your code for other changes in that fix.

tool4ever:
After this is merged two potential improvement directions:
a) while recording show user a live view of how the macro actions string is building for direct feedback if something was correctly registered or not
b) support storing and choosing from multiple recordings

Madwand99:
> After this is merged two potential improvement directions: a) while recording show user a live view of how the macro actions string is building for direct feedback if something was correctly registered or not b) support storing and choosing from multiple recordings

Nice! I like the first idea. The second idea might be tricky without some more work. As-is the currently recorded macros are designed to be somewhat fragile: if the game state changes they aren't usable any more. This is so loops can be interrupted if something unexpected happens. Because it isn't possible to "save" macros as-is, it wouldn't be practical to store multiple macros. I'm sure this is a limitation that could be overcome with more work and a lot more testing, though.

tool4ever:
hmn yea, it might mean it needs a second mode that's less strict with how actions identify objects 🤔
e.g. my deck has this infinite mana combo and the pieces are named X and Y, so having that macro available each time I'm playing it saves even more time...

tool4ever (approved):

PR #10716, opened 2026-05-18
Enable mass-select on remote-controlled human players

MostCromulent:
Follow-up to #10703.

## Summary

Mass-select for mass-search effects (Insidious Dreams, etc.) doesn't engage for remote-controlled humans on hosted games — they fall into the one-card-at-a-time picker instead. `ChangeZoneEffect.allowMultiSelect` gates on `decider.getController().isGuiPlayer()`, which on the server is true only for the host's local lobby player; a remote `PlayerControllerHuman` carries a different `LobbyPlayer` instance and returns false. Predates #10703 — that PR extended mass-select to mandatory searches and made the gap more visible.

The fix adds an `isHuman()` predicate to `PlayerController` (default false, overridden in `PlayerControllerHuman`) and swaps the gate to use it. Captures what the original code intended (`// limit mass selection to human players for now`) without conflating "human" with "host's local player."

## Alternative considered

Architecturally cleaner option: implement `PlayerControllerAi.chooseCardsForZoneChange` (currently a `// this isn't used` null stub) as a loop of `chooseSingleCardForZoneChange`, then drop the controller-type predicate from `allowMultiSelect` entirely. The gate disappears rather than getting refined — the engine calls `chooseCardsForZoneChange` uniformly and each controller fulfills the contract. Larger blast radius (touches `forge-ai` and the test controller, theoretical AI behavior to verify), so not adopted here.

---

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/ChangeZoneEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/ChangeZoneEffect.java
@@ -1561,7 +1561,7 @@ public class ChangeZoneEffect extends SpellAbilityEffect {
     }
 
     private static boolean allowMultiSelect(Player decider, SpellAbility sa) {
-        return decider.getController().isHuman()
+        return !decider.getController().isAI()
                 && !sa.hasParam("ShareLandType")
                 && !sa.hasParam("DifferentNames")
                 && !sa.hasParam("DifferentPower")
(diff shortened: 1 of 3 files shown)


forge-game/src/main/java/forge/game/ability/effects/ChangeZoneEffect.java
```diff
@@ -1561,7 +1561,7 @@ private static class HiddenOriginChoices {
     }
 
     private static boolean allowMultiSelect(Player decider, SpellAbility sa) {
-        return decider.getController().isGuiPlayer()        // limit mass selection to human players for now
+        return decider.getController().isHuman()
```
tool4ever: I'd prefer if we can just stick with `!isAI` instead of extra variant

tool4ever (approved):
technically we still don't want AI branching inside effects API
I might try and look for a cleaner option soon

PR #10718, opened 2026-05-19
Fix: restore ICO_ARCSHOVER for theme compatibility

MostCromulent:
## Summary

The dock rework retired `ICO_ARCSHOVER` and reused its sprite cell at `(400, 800)` for the new yield-settings glyph. Custom themes built around the previous layout still have their hover icon painted at that cell, so the dock's yield-settings button now renders the theme's legacy hover-arc icon, and `ICO_ARCSHOVER` itself is no longer drawn anywhere.

This PR restores the cell and wires it back into runtime use for backwards compatibility with custom themes. Themes will still need to update for new dock icons, but this addresses an avoidable breakage point.

## Changes
- Restore `ICO_ARCSHOVER` at `(400, 800)` as a duplicate of `ICO_ARCSOFF` in the default sheet — vanilla looks identical, themes can paint a distinct glyph there.
- Relocate `ICO_DOCK_SETTINGS` to `(560, 1760)` and extend `sprite_icons.png` from 680x1800 to 680x1840 to fit the new cell.
- Widen the per-cell theme-detection probe in `FSkin.testPreferredSprite` from ±2 to ±8 (clamped to half the smaller cell dimension) so themes with center-transparent icons are still detected.

## Sprite-sheet extension safety
- `FSkin.testPreferredSprite` already has a per-cell bounds check that falls back to the default sheet for any cell lying outside the active theme's sprite. Themes still at 680x1800 keep using their own art for every cell inside their bounds; only the relocated `ICO_DOCK_SETTINGS` falls back to default — and the default pixels at the new cell are byte-identical to what was at the cell's previous position.

---


MostCromulent:
@tool4ever I'm not sure how much we really care about this in the grand scheme of things, but it was raised as a user complaint on Discord and its an easy fix, so here we are.

We're running out of space on the sprite sheet and new dock buttons are being added in other PRs, so extending it is probably useful in any event.

tool4ever:
yea we can do this
you could already combine it here with the new icons from the other PR if you want or just wait

while a general sprite sheet clean up might eventually be desirable it certainly seems like a joyless task with needing to touch at least our self-hosted skins too 🤔

MostCromulent:
> yea we can do this you could already combine it here with the new icons from the other PR if you want or just wait
> 
> while a general sprite sheet clean up might eventually be desirable it certainly seems like a joyless task with needing to touch at least our self-hosted skins too 🤔

Yeah I'll add new icons from other PR here, then that PR can rebase on this.

MostCromulent:
AI hasn't been too bad at automating sprite sheet updates, so could potentially batch update the self-hosted skins at some point as part of broader cleanup.

tool4ever (approved):

PR #10719, opened 2026-05-19
DualCardBox widget for CardView prompts

MostCromulent:
Screenshot 2026-05-19 174050

## Summary

When the rules engine asks the player to partition or order a small CardView list (Doomsday's pile, Index's reorder, Sensei's Divining Top scry), the current widget is DualListBox — a pair of compact text lists that show only the card name, with no visuals and a select-then-click affordance for every move.

DualCardBox renders the same prompt as two card-image grids.

## When this is used

The new widget replaces DualListBox when **all** of the following hold; otherwise the existing DualListBox path is used unchanged.

- The prompt is an ordering/partition prompt (the existing `order(...)` entry point on `CMatchUI`).
- Every item in both the source and destination lists is a `CardView`.
- The combined size of source + destination is 9 cards or fewer (the Ctrl+1-9 hotkey ceiling).
- The prompt is not sideboarding mode.

## Features

- Two card-image grids with bulk-move buttons (`>>`, `<<`, `Auto`); `Auto` enabled only when emptying the pool would satisfy the prompt's min/max.
- FloatingZone-style Ctrl+1-9 quick-select hotkeys on the pool; Ctrl+0 selects the minimum needed.
- Enter accepts when OK is enabled; Esc clears the pool search filter.
- Click and hover on either pane drive the match-UI card preview.
- Destination pane supports drag-reorder with a visible insertion line.
- Cross-pane insertion line appears when dragging from pool to destination (handles empty-pane and append-at-end cases).
- Resizable dialog; panes relayout cards on resize.
- Centered grey-italic placeholder text in empty panes.

Also fixes a small FloatingZone glitch: Ctrl+digit badges now stay in sync when the zone's panel order changes mid-hold (previously they only repainted on Ctrl press/release).

---


MostCromulent:
Sharing for feedback/testing since this is in a hot path for UI prompts.

tool4ever:
First impression is pretty smooth

a) could have the right side not need half of the space by switching to an overlapping stack view, which might fit better with regards to the "Closest to top" label - though that'd leave them less exposed for dragging so I'm not sure if that really makes sense 🤷‍♂️

b) the reusage of the hotkey badges for the order seems reasonable but also a bit unfortunate, especially if you're in the middle of ordering so you see some digits twice. Granted the hint is on the left side but ideally you could use them for right side selection too - at least when left is drained? 🤔

Possible solution: assign numbers starting with the right side, continue with that count on the left after
this also guarantees all hotkeys can always affect a card (well limited to available cards being ordered) + using the same one twice moves the same card back instead of another one 💡

MostCromulent:
> Possible solution: assign numbers starting with the right side, continue with that count on the left after
> this also guarantees all hotkeys can always affect a card (well limited to available cards being ordered) + using the same one twice moves the same card back instead of another one 💡

Implemented on latest commit.

> could have the right side not need half of the space by switching to an overlapping stack view, which might fit better with regards to the "Closest to top" label

Something like below? I don't hate it and it doesn't interfere with drag-and-drop too much, though not sure whether the asymmetry is a good thing or bad thing; `DualListBox` has equal sized columns on both sides. Another option is keep even column width but still apply stacking to the destination panel.

image

tool4ever:
I guess with the current limit staying symmetrical aligns better with the parent, we can always iterate later

tool4ever (approved):

PR #10724, opened 2026-05-20
Yield Rework V2 minor follow-up fixes and reveal-suppress pref

autumnmyst:
- Auto-pass-no-actions with "respects interrupts" now properly breaks on opponent casts and attacker declarations. The event handler gated on isYieldActive() (explicit EOT/stack/marker yields only), so APINA was never told about these events and kept passing; gate on the broader shouldEvaluateInterrupts() instead.
- NPE guards: null phaseOwner/phase in setMarker (treated as a clear), null activator in onSpellAbilityCast, and null si()/negative targetId seen on reconnect-time event replays.
- resetForNewGame() clears per-tick yield flags between games of a match so game 2 doesn't inherit game 1's auto-pass / suggestion state.
- "Suppress reveals while yielding" skips the reveal()/notifyOfValue() modal when set, interrupt-on-reveal is off, and auto-pass is active.


autumnmyst:
I considered logging the reveals when suppressed, but that seemed like a gameplay change, as normally reveals are not logged. So I concluded no reveals should be logged, suppressed or otherwise. It's explicitly for speed, sacrificing that info.

Jetz72:
>I considered logging the reveals when suppressed, but that seemed like a gameplay change, as normally reveals are not logged. So I concluded no reveals should be logged, suppressed or otherwise. 

Is there a standing reason we wouldn't want them in the game log? Or is it just not included because nobody's bothered to include it thus far? My initial impression is that reveals would be a pretty sensible addition to the game log, whether or not the modal is skipped.

autumnmyst:
> > I considered logging the reveals when suppressed, but that seemed like a gameplay change, as normally reveals are not logged. So I concluded no reveals should be logged, suppressed or otherwise.
> 
> Is there a standing reason we wouldn't want them in the game log? Or is it just not included because nobody's bothered to include it thus far? My initial impression is that reveals would be a pretty sensible addition to the game log, whether or not the modal is skipped.

I agree with that. If that's a feature we're looking for it's a simple addition. I wasn't sure if remembering what was revealed manually was an important piece of gameplay, though personally I would rather log it.

MostCromulent:
> Is there a standing reason we wouldn't want them in the game log? Or is it just not included because nobody's bothered to include it thus far? My initial impression is that reveals would be a pretty sensible addition to the game log, whether or not the modal is skipped.

I suspect its just nobody has bothered to add it, and agree it would be good QoL inclusion. Only question is whether it creates any rules issue since in paper cards are just revealed temporarily but log appears as permanent record / you lose the gameplay challenge of having to remember revealed cards?

**Edit:** apparently MTGO displays revealed cards in some sort of log, and Arena keeps revealed cards visible until hand state changes. So adding to the log should be safe.

> If that's a feature we're looking for it's a simple addition. 

If consensus is to add reveal logging then imo:

1. It should produce a single log line with all of the cards + the zone they were revealed from, rather than flooding the log with 7 separate entries when a player's hand is revealed.
2. Should be wired up to the "Medium" log verbosity preset and added as a new checkbox to the custom log verbosity dialog; see here for precedent https://github.com/Card-Forge/forge/pull/9889

forge-gui/src/main/java/forge/gamemodes/match/YieldController.java
```diff
@@ -260,6 +266,22 @@ public void clearAutoYields() {
         activeStore().onGameEnd(owner.getGame() == null || owner.getGame().getView().isMatchOver());
     }
 
+    /** Reset per-game transient state between games of a match. {@link #clearAutoYields} drops
+     *  stored yields but per-tick flags (lastSeenStackNonEmpty, wasAutoPassingLastTick, the
+     *  yield-just-ended edge, autoPassInterrupted) would otherwise bleed into game 2 and
+     *  mis-fire suggestions at its first prompt. */
+    public synchronized void resetForNewGame() {
+        autoPassUntilEOT = false;
```
tool4ever: nothing against you personally but I'm getting somewhat tired of having to disprove AI claims to prevent excess code bloat...
in this case YieldController is a fresh object each game - which you could also have verified yourself first :/

forge-gui/src/main/java/forge/gui/control/FControlGameEventHandler.java
```diff
@@ -283,11 +283,16 @@ public Void visit(final GameEventSpellAbilityCast event) {
...
         YieldController yc = humanController.getYieldController();
-        if (!yc.isYieldActive()) return;
+        // isYieldActive() only covers explicit yields; APINA-with-respects-interrupts
+        // also wants to be told about casts so it can set autoPassInterrupted.
+        if (!yc.shouldEvaluateInterrupts()) return;
         GameView gv = matchController.getGameView();
         if (gv == null || gv.getGame() == null) return;
         // Look up the actual SpellAbilityStackInstance by id (host-side; client gv.getGame() is null).
+        // event.si() can be null on reconnect-time event replays.
+        if (event.si() == null) return;
```
tool4ever: what's this "event replay":
this object is an immutable record that only gets constructed in a single place and thread, so how would the field ever be null?!

forge-gui/src/main/java/forge/gamemodes/match/YieldController.java
autumnmyst: I apologize for this. I think it's actually my fault not the AI. This PR is basically a local set of fixes I've been adding to throughout the yield rework PRs, and this must have been a rebase mistake. It was present at some point but removed in https://github.com/Card-Forge/forge/pull/10606/changes/b03264c688ec9aee6d764cac4bd3817f35d82f44 and I just assumed it was a fix to [a bug I noticed back in the first iteration of the yield rework](https://github.com/Card-Forge/forge/pull/9643#issuecomment-4320685552) when it was part of AbstractGuiGame

forge-gui/src/main/java/forge/gui/control/FControlGameEventHandler.java
autumnmyst: removed

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -1790,7 +1795,9 @@ public void notifyOfValue(final SpellAbility sa, final GameObject realtedTarget,
         final String message = MessageUtil.formatNotificationMessage(sa, player, realtedTarget, value);
         if (sa != null && sa.isManaAbility()) {
             getGame().fireEvent(new GameEventAddLog(GameLogEntryType.LAND, message));
-        } else {
+        } else if (!shouldSuppressRevealModal()) {
```
tool4ever: waaaay too aggressive, not sure why you think this method is only relevant for revealing - stopping lots of other messages seems rather unexpected and with some being quite important, e.g. "Attack declaration invalid"

autumnmyst: reverted this so we just suppress reveals, can filter more carefully in a dedicated PR

autumnmyst:
> > Is there a standing reason we wouldn't want them in the game log? Or is it just not included because nobody's bothered to include it thus far? My initial impression is that reveals would be a pretty sensible addition to the game log, whether or not the modal is skipped.
> 
> I suspect its just nobody has bothered to add it, and agree it would be good QoL inclusion. Only question is whether it creates any rules issue since in paper cards are just revealed temporarily but log appears as permanent record / you lose the gameplay challenge of having to remember revealed cards?
> 
> **Edit:** apparently MTGO displays revealed cards in some sort of log, and Arena keeps revealed cards visible until hand state changes. So adding to the log should be safe.
> 
> > If that's a feature we're looking for it's a simple addition.
> 
> If consensus is to add reveal logging then imo:
> 
> 1. It should produce a single log line with all of the cards + the zone they were revealed from, rather than flooding the log with 7 separate entries when a player's hand is revealed.
> 2. Should be wired up to the "Medium" log verbosity preset and added as a new checkbox to the custom log verbosity dialog; see here for precedent [Better game log UI: inline card images, improved verbosity presets #9889](https://github.com/Card-Forge/forge/pull/9889)

Added a new reveal log, and added it to the medium verbosity

autumnmyst:
Closing in favor of #10968 for more granularity

PR #10726, opened 2026-05-20
Animate: RemoveThisAbility via Predicate

Hanmac:
For possible later cleanup in #9680


forge-game/src/main/java/forge/game/ability/effects/AnimateEffectBase.java
tool4ever: this shouldn't remove all KW

tool4ever (approved):

PR #10740, opened 2026-05-22
Enable ChangeZone in MayEffectFromOpeningDeck; update un-card list.

Jetz72:
Most of this is the un-card list update for the latest Unknown Event, plus some sets Black Lotus Lounge planes that I missed. Includes a slight tweak to `MayEffectFromOpeningDeck` that should allow `ChangeZone` effects to work using it without triggering a `ConcurrentModificationException`. Added for someone who was trying to implement "Buried Ogre".


Jetz72:
Added a fix to prevent Amass from crashing if given a creature type that doesn't have a defined army token script. The logic was already there to apply the type list and name anyway, so I had it fall back to a generic black army token script.

tehdiplomat (approved):

PR #3556, opened 2023-07-29
Card: add Table for SwitchPT

Hanmac:
Closes #3555 

- [x] remove Hidden Keyword `CARDNAME's power and toughness are switched`
- [ ] Extend the AI to check when to use that Ability
- [ ] Technically it should be its own layer, but we don't have a static ability for this?

AI cases:

- [ ]  In case of Attacking or Blocking to check if switched PT would be more in favor?
- [ ]  Use it as kill Spell if Creature has more damage than power or if power is negative? Need to calculate the right amount of PT after switch
- [ ]  AI probably need to respect isSwitchPT when checking for effects like +P/+T or -P/-T  ... especially if its like +X/-X or vise versa

PR #10744, opened 2026-05-22
Preserve hidden extrinsic keywords in Card.copyFrom (fixes P/T switch lost in LKI copies)

Neur0nz:
## Problem

Casting **Inside Out** (*switch target creature's power and toughness*) on a creature and then **Fling**ing it deals the creature's *un-switched* power instead of the switched value.

Minimal repro:
- Tireless Tribe (1/1); pump its toughness with its own ability a few times, then cast Inside Out on it → it shows e.g. `33/1` on the battlefield.
- Fling it. Fling deals `NumDmg$ X` with `SVar:X:Sacrificed$CardPower`, read from the sacrificed creature's last-known information — it deals **1** (the un-switched base power), not 33.

Any effect that reads a creature's last-known power after it leaves the battlefield is affected (Fling, Soul's Fire, Hail Storm, etc.).

## Root cause

`CardCopyService.getLKICopy` builds the last-known-information copy through `Card.copyFrom(Card)`. `copyFrom` copies every sibling continuous-effect keyword store — `changedCardKeywords`, colors, types, names, traits — but **omits `hiddenExtrinsicKeywords`**, the table written by `Card.addHiddenExtrinsicKeywords`. Inside Out's switch is a `KW$ HIDDEN ...` pump keyword, so it lands in that omitted table and is dropped from the LKI copy; `getNetPower()` on the copy then returns the un-switched value.

## Fix

Copy `hiddenExtrinsicKeywords` in `copyFrom`, alongside the other keyword stores. `getLKICopy` already rebuilds the keyword cache after `copyFrom`, so the copied keyword takes effect.

## Test

Adds `GameSimulationTest#testSwitchedPowerToughnessSurvivesLKICopy`: switches a Giant Spider's P/T with Inside Out (→ 4/2) and asserts `CardCopyService.getLKICopy(spider).getNetPower() == 4`. It **fails on master** (`expected:<4> but was:<2>`) and **passes with this change**.

## Note

`GameCopier` (forge-ai simulation) already copies `hiddenExtrinsicKeywords` explicitly right after calling `copyFrom`; that local loop is now redundant and could be removed in a follow-up — arguably evidence this copy belonged in `copyFrom` all along.

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -8108,6 +8108,18 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
 
         setChangedCardKeywords(in.getChangedCardKeywords());
 
+        // Hidden extrinsic keywords (e.g. Inside Out's "power and toughness are
+        // switched", added via Pump's KW$ HIDDEN) are continuous-effect keyword
+        // changes just like changedCardKeywords above. Copying them keeps the
+        // last-known-information copy (CardCopyService.getLKICopy) faithful so
+        // that effects reading a creature's last-known P/T after it leaves the
+        // battlefield -- e.g. Fling's Sacrificed$CardPower -- see the switched
+        // value instead of the unswitched base.
+        for (Table.Cell<Long, Long, List<String>> cell : in.hiddenExtrinsicKeywords.cellSet()) {
+            this.hiddenExtrinsicKeywords.put(cell.getRowKey(), cell.getColumnKey(),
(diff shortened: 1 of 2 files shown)


MostCromulent:
AI coding agents have a tendency to add unnecessary unit tests to prove their wiring works, this seems like it probably falls into that category. [See warning against this here.](https://github.com/Card-Forge/forge/blob/master/CONTRIBUTING.md#using-ai-coding-agents)

Hanmac:
probably better to revive one of my oldest MR #3556

tool4ever:
well the bug could manifest for other hidden KW too - though there aren't probably that many relevant left at this point...

PR #10746, opened 2026-05-23
Add dockable Macro Window with readable replay diagnostics

Madwand99:
Changes include:

- Added a Macro Window that opens when macro recording starts, can float or dock, and closes at match end.
- Shows recorded macro steps in a copyable, wrapping text view.
- Highlights the active replay step during playback.
- Adds a Macro log section for replay failures and diagnostic messages, shown only when needed.
- Improves recorded action descriptions so macro steps are more human-readable.
- Localizes Macro Window text and action descriptions across supported language files.
- Adds replay handling for several interactive prompts:
  - target selection
  - list selection
  - simultaneous ability ordering
  - optional confirms/declines
  - mana payment prompts
  - combat/phase advancement edge cases
- Improves loop replay behavior so macros can continue through pending stack/target prompts when appropriate.
- Keeps macro recording/playback state cleaned up when a match ends.

Thanks to Codex for help with code and translations.


forge-gui/res/languages/en-US.properties
```diff
@@ -530,6 +530,49 @@ lblMacroRecordStopTooltip=Stop recording macro and choose repeat count
...
 lblMacroPlaybackActiveTooltip=Macro playback is running
+lblMacroWindow=Macro Window
+lblMacroIdle=Macro idle
+lblMacroRecordingStatus=Recording macro
+lblMacroPlaybackStatus=Playing macro
+lblMacroReadyStatus=Macro ready
+lblMacroNoActionsRecorded=No recorded actions yet.
+lblMacroActionCount={0} recorded action(s)
+lblMacroLog=Macro log
+lblMacroPlaybackProgress={0} / {1}
```
tool4ever: we don't need trivial labels like this without anything to translate

forge-gui/res/languages/en-US.properties
```diff
@@ -530,6 +530,49 @@ lblMacroRecordStopTooltip=Stop recording macro and choose repeat count
...
+lblMacroActionDecline=Decline
+lblMacroActionChoiceFor={0} choice for {1}
+lblMacroActionFinishSelecting=Finish selecting targets
+lblMacroActionChooseManaCombination=Choose mana combination: {0}
+lblMacroManaAmount={0} {1}
+lblMacroManaAmountPlural={0} {1} mana
+lblMacroActionChooseMode=Choose mode: {0}
+lblMacroActionChooseModes=Choose modes:{0}
+lblMacroActionDuringPhase=during {0}
+lblMacroPhaseMain1=Main Phase 1
```
tool4ever: zone labels already exist

forge-gui/res/languages/en-US.properties
Madwand99: Agreed, fixed!

forge-gui/res/languages/en-US.properties
Madwand99: Fixed!

forge-gui-desktop/src/main/java/forge/control/KeyboardShortcuts.java
```diff
@@ -193,7 +193,7 @@ public void actionPerformed(final ActionEvent e) {
             }
         };
 
-        final Action actMacroRecord = macroAction(matchUI, ui -> ui.getGameController().macros().setRememberedActions());
+        final Action actMacroRecord = macroAction(matchUI, ui -> ui.getCDock().toggleMacroRecording());
```
tool4ever, line 196: this leads to redundant call of `refreshMacroButtons()`

Madwand99: Fixed!

forge-gui/src/main/java/forge/player/RecordActionsMacroSystem.java
```diff
@@ -904,27 +1140,49 @@ private boolean noRemainingTargetSelectionsBefore(final int endIndex) {
...
     }
 
     public boolean processAction(PlayerAction action) {
-        debug("try " + action.describe());
+        setActiveAction(action);
+        if (DEBUG) {
+            debug("try " + action.describe());
+        }
         final Input inp = playerControllerHuman.getInputProxy().getInput();
         if (action instanceof ActivateAbilityAction activateAbilityAction) {
```
tool4ever: Sigh, I feel like you **really** didn't need to have this PR extend the feature with another 1500+ Lines of Code all at once **again**...

Coincidentally that's also the size this class has reached now and the architecture really needs to get cleaned up so it doesn't bleed this much into all these GUI classes:
there are 9 instanceof checks for `PassPriorityAction` alone, that's a clear sign of a roundabout logic flow
like why can't the job of each action be consolidated into its class - right now you're completely ignoring `PlayerAction.run()` instead? 🤔

Madwand99: This is a great idea! I'm experimenting now with some code to move replay logic into the PlayerHandler classes. I'll need to do a bunch of tests to verify the logic is improved but I'll get that to you ASAP.

Madwand99: I agree with the direction, but I stopped short of moving replay execution into PlayerAction.run() because PlayerAction is in forge-game, while replay depends heavily on forge-gui input classes like InputSelectTargets, InputPayMana, InputPassPriority, and PlayerControllerHuman helpers. Moving that directly into PlayerAction would make forge-game depend on GUI-layer replay details.

Instead, I moved the pure/action-owned pieces into the action classes: selection classification, selected-card access, confirm prompt matching, and pass-priority phase/stack predicates. RecordActionsMacroSystem still owns the live input orchestration because it has to inspect and drive the current GUI input state.

If you have some ideas for how to move more code into the Action classes without running into the GUI issues I ran into, please let me know. Very interested in improving this code where it makes sense!

Madwand99: In an attempt to reduce lines of code in RecordActionsMacroSystem, I refactored the GUI-specific replay logic out of `RecordActionsMacroSystem` into `MacroActionReplayer`.

The intent is to keep `RecordActionsMacroSystem` focused on macro state and sequencing: recording, playback loops, retry/wait behavior, active step tracking, and macro window updates. `MacroActionReplayer` now owns the GUI/input-specific question of “can this recorded `PlayerAction` satisfy the current input?”, including card/player selection, confirms, pass priority, mana payments, targeting completion, and ability activation.

I chose this boundary instead of moving replay directly into the `forge-game` `PlayerAction` classes because replay currently depends on GUI-layer input types such as `InputSelectTargets`, `InputPayMana`, `InputPassPriority`, and `InputAttack`. Keeping that coupling in `forge-gui` avoids pushing GUI dependencies into `forge-game`, while still making the macro system easier to reason about and leaving room for a future replayable-choice API to replace the helper’s internals.

It should be noted that this separation does increase lines of code, but I think there is probably a benefit to doing it this way regardless. RecordActionsMacroSystem was getting a bit large.

Madwand99:
Closing this until #10930 is accepted.

PR #10747, opened 2026-05-23
Text Changes using Records

Hanmac:
Part of #8682 

> Would be cool if i could change changedTextColors Map<String,String> to something using Enum


tool4ever:
Well I don't know man - I think there's a point where abstraction becomes too much so it risks becoming a liability in the long term and I'm a bit concerned this direction is happening here...

Unless we're expecting Wizards to suddenly introduce way more text change variant mechanics I might have a hard time convincing myself that
- extra TrackableType + View
- two new records + interfaces each
- an ingame mechanic crossing another module boundary

are really useful/an improvement (instead of continuing with the approach that's directly based on a much easier to grasp String Map).

Maybe if I stare at it a while longer it'll eventually make more sense, though having to do that + this actually growing the code even with some cleanup aren't really promising signs. 🤷‍♂️

But that's just my opinion, I'll be interested what the others think on that.

Hanmac:
> an ingame mechanic crossing another module boundary

That's only because of WordChangedType, imo that could be moved to forge-game

Hanmac:
> extra TrackableType

I did some cleaning fir Trackable Type in #10751

Hanmac:
@tool4ever updated MR with rebase

PR #10751, opened 2026-05-23
TrackableType: cleanup default value, reduce need of extra classes

Hanmac:
This cleanup reduce the need to define extra classes just for `getDefaultValue`
(java makes them anyway under the hood)

- ~~add new TrackableValueType where you can give its default value, mostly variant of `0` or Empty View~~ 
- TrackableType now supports set default value
- TrackableObjectType has `null` as default
- ~~TrackableCollectionType has an empty collection (that isn't created each time)~~
- TrackableCollectionType also has `null` as default

### DIFF
--- a/forge-game/src/main/java/forge/trackable/TrackableTypes.java
+++ b/forge-game/src/main/java/forge/trackable/TrackableTypes.java
@@ -21,8 +21,10 @@ import forge.game.spellability.StackItemView;
 import forge.item.IPaperCard;
 
 public class TrackableTypes {
-    public static abstract class TrackableType<T> {
-        private TrackableType() {
+    public static class TrackableType<T> {
+        private T defaultValue;
+        private TrackableType(T defaultValue) {
+            this.defaultValue = defaultValue;
         }
 
         protected void updateObjLookup(Tracker tracker, T newObj) {
@@ -30,11 +32,15 @@ public class TrackableTypes {
         protected void copyChangedProps(TrackableObject from, TrackableObject to, TrackableProperty prop) {
             to.set(prop, from.get(prop));
         }
-        protected abstract T getDefaultValue();
+
+        protected T getDefaultValue() {
+            return defaultValue;
+        }
     }
 
-    public static abstract class TrackableObjectType<T extends TrackableObject> extends TrackableType<T> {
+    public static class TrackableObjectType<T extends TrackableObject> extends TrackableType<T> {
         private TrackableObjectType() {
+            super(null);
         }
 
         public T lookup(T from) {
@@ -73,10 +79,11 @@ public class TrackableTypes {
         }
     }
 
-    private static abstract class TrackableCollectionType<T extends TrackableObject> extends TrackableType<TrackableCollection<T>> {
+    private static class TrackableCollectionType<T extends TrackableObject> extends TrackableType<TrackableCollection<T>> {
         private final TrackableObjectType<T> itemType;
 
         private TrackableCollectionType(TrackableObjectType<T> itemType0) {
+            super(null);
             itemType = itemType0;
         }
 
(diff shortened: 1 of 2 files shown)


tool4ever (changes requested):
looks bad:
Image

forge-game/src/main/java/forge/trackable/TrackableTypes.java
```diff
@@ -119,32 +134,17 @@ protected void copyChangedProps(TrackableObject from, TrackableObject to, Tracka
...
-        public Float getDefaultValue() {
-            return 0f;
-        }
-    };
-    public static final TrackableType<String> StringType = new TrackableType<String>() {
-        @Override
-        public String getDefaultValue() {
-            return "";
+        protected TrackableCollection<T> getDefaultValue() {
+            return new TrackableCollection<>();
```
tool4ever: Can we try and keep that null too?
since this will get called a lot a new object each time seems rather wasteful

Hanmac: now all trackable types has default value

so the empty `new TrackableCollection<>()` is only created once (per type)

tool4ever (approved):

tool4ever (changes requested):
Imo it'd break (at least here) now, modifying the default value:
https://github.com/Card-Forge/forge/blob/908c095f9bd5df807aa1e2d7635f70be27548f0c/forge-game/src/main/java/forge/game/card/CardView.java#L1764-L1770

Hanmac:
wait, my fix doesn't work, need to do it another time

Edit:
I was thinking `computeIfAbsent` + `getDefaultValue`, but that doesn't work either 🤔

Hanmac:
@tool4ever your opinion on using:
`computeIfAbsent(key, (t) -> new TrackableCollection<CardView>())` ?

tool4ever:
Hmn, I think it's still just asking for trouble but not enough benefit:
this could easily introduce subtle bugs by new code since it'd be reasonable to expect that modifying the returned collection is safe :/

Besides what's the point of switching the default anyway? We know we'll still want to work with `null` values so the Maps can stay small...

Hanmac:
> Besides what's the point of switching the default anyway? We know we'll still want to work with `null` values so the Maps can stay small...

Because it's stupid that we need to differ between (empty) collection and `null` values.

(Even if we store empty collection as null behind)

Hanmac:
@tool4ever i can't really overwrite `FCollection.getEmpty` might need to remove that
(see can't override static methods)

but i would be okay if:
```java
public FCollectionView<CardView> getChosenCards() {
        return get(TrackableProperty.ChosenCards);
}
```

Would return empty Collection if able?
Then I don't need to mess with the default values,
while still be able to just iterate them in the GUI?

hm The GUI part doesn't need to know that it's a Trackable Collection inside 🤔

Hanmac:
Maybe related, but i noticed that the CardView helper functions that work with CardCollection stuff,
https://github.com/Card-Forge/forge/blob/50dfcde750ea2c0177541fbe173ade2c293ff10e/forge-game/src/main/java/forge/game/card/CardView.java#L1736-L1857

should be moved to GameEntityView to be used with `AttachedCards`
or maybe even directly into TrackableObject

tool4ever:
hmn, type safety can get a bit messy with these generics 🤔
maybe it should only present itself as Iterable to callers?
though I'm not certain that wouldn't just make it more messy somewhere else...

Hanmac:
> maybe it should only present itself as Iterable to callers?

yeah that what i was thinking too, the functions doesn't need to know that they are TrackableCollection on the inside, so i might not need a TrackableCollection.Empty variant

i don't have a problem with Tracker Properties to be null inside the Tracker,
but this functions setting the Game Properties to null when empty can be annoying,
i will try to debug how much it affects this

> hmn, type safety can get a bit messy with these generics 🤔

You mean for the addCard/removeCard ones?
i want to keep them that way for now, no need to make them Type generic.


i will try to make bit of the collection cleanup in another MR

Hanmac:
@tool4ever found the piece again i have been looking for:
https://github.com/Card-Forge/forge/blob/89af34e4a0c0d1395b4899139b12c65deb1efd79/forge-game/src/main/java/forge/game/player/PlayerView.java#L98-L100

Something like that on the View classes for the get methods, to return an empty List?
(I need to check if it needs to be an FCollection or if a normal Collection is enough)

Hanmac:
@tool4ever with #10801 I probably don't need to touch the default value for collection

tool4ever:
```
EDT > java.lang.NullPointerException: Cannot invoke "forge.util.collect.FCollectionView.size()" because "items" is null
	at forge.screens.match.views.VStack.updateStack(VStack.java:129)
	at forge.screens.match.controllers.CStack.update(CStack.java:45)
```

tool4ever (approved):

PR #10755, opened 2026-05-23
Let players ignore the illegal-deck warning

MostCromulent:
In response to a user request on Discord.

This adds the Deck Conformance preference as a checkbox to the home-screen Forge → Game menu, mirroring the deck editor's existing item and reusing the `ENFORCE_DECK_LEGALITY` preference so the two stay in sync. Also adds the existing legality hint to the deck editor's item, which previously had none.

---


forge-gui-desktop/src/main/java/forge/screens/home/sanctioned/ConstructedGameMenu.java
```diff
@@ -28,6 +28,7 @@ public static JMenu getMenu() {
         menu.add(getMenuItem_SingletonMode());
         menu.add(getMenuItem_ArtifactsMode());
         menu.add(getMenuItem_SmallCreaturesMode());
+        menu.add(getMenuItem_DeckConformance());
```
tool4ever: Besides missing a mobile variant I'd prefer avoiding this direction of menu growth if something more context-based can be found:
what if we change the stop dialog so it offers an "Ignore" button that lets you continue instead?

this would also allow it to work in kind of a hybrid fashion, so you know you'll still get warned but don't have to toggle anything for a quick playtest 💡

TheLastNarwhal: Thinking about it afterward, I felt like there wasn't much point in adding a toggle for Deck Conformance to a menu, since having to navigate a menu versus having to click over to Preferences is basically the same thing, or at least the same amount effort.

This is a good solution, probably better actually than having a toggle.

MostCromulent:
Implemented on latest commit:

Screenshot 2026-05-25 075349

tool4ever (approved):

PR #10759, opened 2026-05-24
Fixes Varchild's War Riders Cumulative Upkeep from giving tokens to each opponent

### LINKED ISSUE #3584: Varchild's War-Riders needs better Cumulative Upkeep Cost
Currently each Opponent will make X tokens.

djdaquin:
Varchild's War Riders' Cumulative Upkeep was not working properly and was giving X Survivors to each opponent. It is supposed to let you give out a total of X to any number of Opponents.


tool4ever (approved):
having the real Keyword used instead would only matter in some minor cases, so this compromise seems ok

Hanmac:
@tool4ever I will try it as cost again if I can prevent it from merging 🤔

PR #10760, opened 2026-05-24
Prevent crash when selling cards in commander mode

Dreavian:
When selling cards in commander mode, `parentScreen.getCommanderPage()` is null which causes a crash on line 1521 `FImage icon = parentScreen.getCommanderPage().icon;`. Exiting early seems correct, given the other condition + no needing commander options when selling cards.

Tested locally, and it fixes the crash in my save.

Issue #10761


Jetz72:
I'm curious why this hasn't been very widely reported ever since the implementation of commander mode in Adventure. Just looking over the code I'd expect this to happen universally. 

I think the root of the issue is that the shop UI probably shouldn't be registering as a commander deck editor in the first place. Might make sense for `AdventureDeckEditor.ShopConfig` to override `hasCommander` and just return false. Might make sense if `getDeckFormat` returned `null` since that editor doesn't technically have a deck to edit, though I dunno if the downstream code is prepared to handle that...

Dreavian:
Hi @Jetz72 , I thought it was weird that it wasn't reported either, but it seems like it is not just me based on the linked PR. 

I have made changes as you said around the shop config.

 I left your recommended `getDeckFormat` change alone, it seems like it won't handle null consistently, even though it seems it is null in some places.

With the new change the null check isn't required (as the shop config change makes `!parentScreen.isCommanderEditor()` true) but I have left the null check in place. It doesn't seem out of place given the same condition exists for `canBeMainCommander`, and it doesn't hurt for redundancy.

Jetz72 (approved):

Dreavian:
@Jetz72 / @tool4ever I am not sure why the tests didn't pass, they ran fine locally. I merged in master again, can one of you approve the workflow again to see if it was an intermittent thing?

Jetz72:
Yes, we were getting some sporadically failing tests across all PRs for a bit. Not your fault.

PR #10767, opened 2026-05-24
YSOS, batch 4: 4 cards

dracontes:
As revealed recently:
- [Inspiring Easel](https://scryfall.com/card/ysos/22/inspiring-easel): Tested to satisfaction.
- [Interdisciplinary Studies](https://scryfall.com/card/ysos/4/interdisciplinary-studies): Tested to satisfaction.
- [Paradigm Shifter](https://scryfall.com/card/ysos/23/paradigm-shifter): Tested to satisfaction.
- [Scalar Scholar](https://scryfall.com/card/ysos/25/scalar-scholar): Tested to satisfaction.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/interdisciplinary_studies.txt
+++ b/forge-gui/res/cardsfolder/upcoming/interdisciplinary_studies.txt
@@ -1,0 +1,6 @@
+Name:Interdisciplinary Studies
+ManaCost:2 U
+Types:Instant
+A:SP$ Seek | Num$ 1 | Type$ Card.MultiColor | SubAbility$ DBLearn | SpellDescription$ Seek a multicolored card.,,,,,,
+SVar:DBLearn:DB$ Learn | SpellDescription$ Learn.
+Oracle:Seek a multicolored card.\nLearn.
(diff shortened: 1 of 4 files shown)


forge-gui/res/cardsfolder/upcoming/interdisciplinary_studies.txt
```diff
@@ -0,0 +1,6 @@
+Name:Interdisciplinary Studies
+ManaCost:2 U
+Types:Instant
+A:SP$ Seek | Num$ 1 | Type$ Card.MultiColor | SubAbility$ DBLearn | SpellDescription$ Seek a multicolored card.,,,,,,
```
paulsnoops, line 4: ,,,,,, :D

tool4ever: that works as linebreak which should really be reworked at some point :/

paulsnoops (approved):

forge-gui/res/cardsfolder/upcoming/interdisciplinary_studies.txt
squee1968: When it does get reworked, what about making it something like "\l" (linebreak), similar to the way we use \n?

PR #10771, opened 2026-05-25
Consolidate ForgePreferences providing

tool4ever:
- Removed redundant commander legality check
- Fix sorting regression after #10640
- Cherry pick #10744


tool4ever:
Weird, locally tests were fine so that won't be fun debugging

MostCromulent:
> **Issue.** Two tests in forge-gui-desktop fail in CI with NullPointerExceptions (testTrueNetworkTraffic, testDeckLoaderCanLoadDeck) because the card database is never loaded — cardReader and commonCards are null. The tests bootstrap via TestUtils.ensureFModelInitialized(), which only calls FModel.initialize() when FModel.getPreferences() == null. On master that guard held before initialization because getPreferences() returned a plain field that stayed null until initialize() ran. This PR routes getPreferences() through GuiBase.getForgePrefs() (and previously through a memoized supplier), both of which return a non-null instance on first access. The guard is therefore always false, initialize() is never called from this path, and the card DB stays empty.
> 
> **Why it doesn't surface offline.** The card DB is a process-wide singleton, and the full suite shares one JVM. Several other test classes (AITest, SimulationTest, the forge.card.* mock tests) call FModel.initialize() unconditionally, so once any of them runs, the DB is seeded for every test that follows. The build configures Surefire with defaults — no testng.xml, no runOrder — so test classes execute in runOrder=filesystem order, i.e. the order the OS returns from the directory listing of compiled classes. That order differs between a local Windows/NTFS filesystem and the Ubuntu/ext4 CI runner. Locally, a seeding class happens to run before the network tests, so they piggyback on the loaded DB and pass; in CI the network tests run first against an empty DB and fail. The dependency is latent and only manifests under CI's ordering.
> 
> **Solution.** The fix belongs on the test side, since the network tests should be self-sufficient rather than relying on another class to seed the singleton. Change ensureFModelInitialized() to decide initialization from whether the card DB is actually loaded (e.g. FModel.getMagicDb() == null) or from an explicit initialized flag, rather than inferring it from getPreferences(). That makes the bootstrap independent of how getPreferences() is implemented and removes the ordering dependency, unblocking the consolidation change in this PR.

paulsnoops (approved):

tool4ever:
@MostCromulent
some other PR tests fail randomly, can you check again? hopefully it's just GitHub being messed up 🤔

PR #9899, opened 2026-02-23
Add Host/Join landing page to online lobby

MostCromulent:
Replaces the current "Start Server" entry point on both desktop and mobile with a consistent landing page with Host Game / Join Game buttons, warning about bugs, and link to network play wiki page.

Host game goes immediately to lobby screen with Server URL popup, Join game prompts for server IP address and port.

Screenshot 2026-02-24 074752
Screenshot 2026-02-28 142531

PR #10772, opened 2026-05-25
Remove unused GuiBase.isInterrupted flag

MostCromulent:
`GuiBase.isInterrupted()` was added in [d87d825](https://github.com/Card-Forge/forge/commit/d87d825bcee14fc9fc65a0d83049a248c1cdb904) (2020) to suppress an auto-connect prompt. Back then, activating the online lobby with no active lobby immediately popped a "connect to a server?" input dialog (`onActivate` → `NetConnectUtil.getServerUrl()`). On a dropped connection, teardown calls `closeConn` → `Forge.back()`, which re-activates the lobby screen and re-fired that prompt — so the flag was set during teardown and read once in `onActivate` to skip it.

## Why removing it is safe

- **The thing it guarded against is gone.** #9899 replaced the auto-connect prompt with a passive Host/Join landing page. `onActivate` no longer prompts on activation; with no lobby it just calls `revalidate()` to lay out the Host/Join buttons.
- **The post-disconnect path lands correctly without it.** `closeConn` nulls the lobby before `Forge.back()`, so the re-activated `onActivate` takes the `revalidate()` branch — the Host/Join page, which is exactly where a disconnected client should end up.
- **No other callers.** The field, its accessors, and the `setInterrupted(true)` in `NetConnectUtil` were the only references to it anywhere in the tree.

---

### DIFF
--- a/forge-gui-mobile/src/forge/screens/online/OnlineLobbyScreen.java
+++ b/forge-gui-mobile/src/forge/screens/online/OnlineLobbyScreen.java
@@ -15,7 +15,6 @@ import forge.gamemodes.net.OfflineLobby;
 import forge.gamemodes.net.client.FGameClient;
 import forge.gamemodes.net.server.FServerManager;
 import forge.gui.FThreads;
-import forge.gui.GuiBase;
 import forge.gui.interfaces.ILobbyView;
 import forge.gui.util.SOptionPane;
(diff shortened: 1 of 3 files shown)


tool4ever (approved):
the best kind of cleanup

PR #10774, opened 2026-05-25
Improve card-based Commander deck generation legality and weighting

Madwand99:
This PR tightens Commander-family generated deck legality and improves how card-based Commander generation uses Commander.dat.

Changes:

- Prevent Commander-family generated decks from adding duplicate non-basic card names across different printings.
- Treat selected commanders/partners/signature spells as already used when building the main deck, so they are not also added to the main deck.
- Disable the constructed-style extendPlaysets fallback for Commander-family builders, since filling gaps with duplicate non-land cards is illegal in singleton formats.
- Add format-aware copy-limit filtering to fallback full-pool fills.
- Change card-based Commander matrix selection from flat shuffle to weighted randomized ordering, so higher-frequency commander-card associations from Commander.dat are more likely to be considered earlier while still preserving generation variety.

Thanks to Codex for help with code and testing.


forge-gui/src/main/java/forge/deck/DeckgenUtil.java
```diff
@@ -819,6 +818,20 @@ else if (commander.getRules().canBePartnerCommander()) {
         return deck;
     }
 
+    private static List<Map.Entry<PaperCard, Integer>> getWeightedRandomizedCardPool(final List<Map.Entry<PaperCard, Integer>> potentialCards) {
+        final Map<Map.Entry<PaperCard, Integer>, Double> sortKeys = new IdentityHashMap<>();
+        for (final Map.Entry<PaperCard, Integer> cardEntry : potentialCards) {
+            sortKeys.put(cardEntry, getWeightedRandomSortKey(cardEntry));
+        }
+        potentialCards.sort(Comparator.comparingDouble(sortKeys::get).reversed());
```
tehdiplomat: Any particular reason why we're dividing by the weight and reversing rather than just multiplying by the weight?

Madwand99: Yes, this is intentional: log(random) / weight is a standard weighted sampling without replacement key. log(random) is negative, so larger weights produce keys closer to zero; reversing the sort places those higher-weight cards earlier while still allowing lower-weight cards to appear. Multiplying by weight would not preserve the same weighted sampling distribution, and with the current negative-log key would actually invert the intended bias unless the sort direction changed.

forge-gui/src/main/java/forge/gamemodes/limited/CardThemedDeckBuilder.java
```diff
@@ -604,7 +604,7 @@ public boolean test(PaperCard card) {
                         && !card.getRules().getMainPart().getType().isLand();
             }
         };
-        List<PaperCard> possibleList = Lists.newArrayList(pool.getAllCards(possibleFromFullPool));
+        List<PaperCard> possibleList = limitCopiesForFormat(Lists.newArrayList(pool.getAllCards(possibleFromFullPool)));
```
tool4ever: if `DeckGeneratorBase.maxDuplicates` isn't respected in some places that should be fixed instead of adding another parallel mechanism

Madwand99: Fair point. I updated code so the fallback list filtering uses the generator’s maxDuplicates for normal cards, while still preserving explicit card-specific format limits. This keeps the fallback path aligned with the generator duplicate policy instead of relying only on DeckFormat.

tool4ever: Hmn ok, I guess you'd want to use the card-specific DeckFormat method for the few DeckLimit mechanics:
in that case this should still be refactored uniformly so all places use the same logic...

Madwand99: Thanks, that makes sense. I refactored this so the generator is no longer using a one-off parallel copy limit.

The card-themed deck builder now routes staged card additions through shared helpers that respect DeckGeneratorBase.maxDuplicates for normal cards, while still preserving card-specific DeckFormat.getMaxCardCopies(card) exceptions like DeckLimit cards. I also applied that path uniformly across the relevant add flows: key cards, fallback/random cards, mana-curve cards, third-color additions, nonbasic lands, low-CMC filler, and playset extension.

While expanding my local legality test coverage, I found and fixed two related random Commander-family issues:

- Random Commander partner selection could choose the commander itself as its partner.
- Random Brawl generation could produce illegal colorless-commander decks, so those are skipped until that generator can build a legal colorless Brawl mana base.

I verified this with a local ignored integration test that generates decks across Commander/Oathbreaker card-based generators, sampled random Commander/Brawl generators, color generators, and Standard matrix/archetype generators, then checks each with Forge’s existing DeckFormat.getDeckConformanceProblem(deck).

tool4ever:
I've cherry picked some of the cleanup in DeckgenUtil to reduce the size of this a bit - for the logic changes I'll need more time since tbh your AI tends to often approach things sort of indirectly...

forge-gui/src/main/java/forge/deck/CommanderDeckGenerator.java
```diff
@@ -63,10 +63,16 @@ public static List<DeckProxy> getBrawlDecks(final DeckFormat format, boolean isF
         return uniqueCards.toFlatList().stream()
                 .filter(format.isLegalCardPredicate())
                 .filter(PaperCardPredicates.fromRules(CardRulesPredicates.CAN_BE_BRAWL_COMMANDER.and(canPlay)))
+                .filter(CommanderDeckGenerator::canGenerateBrawlDeck)
```
tool4ever: I don't see a reason for such sweeping removal:
`addWastesIfRequired()` seems to work fine

Madwand99: Fixed!

forge-gui/src/main/java/forge/gamemodes/limited/CardThemedDeckBuilder.java
```diff
@@ -737,11 +730,6 @@ private void addWastesIfRequired(){
      */
     private int[] calculateLandNeeds() {
         final int[] clrCnts = { 0,0,0,0,0 };
-        //Brawl allows colourless commanders to have any number of one basic land to fill out the deck..
-        if (format.equals(DeckFormat.Brawl) && keyCard.getRules().getColorIdentity().isColorless()){
```
tool4ever: sigh, why are you simply removing this code now?
this is clearly meant to support CR 903.12e

Madwand99: My bad. Hopefully the latest commit fixes this! Have to admit I'm not as familiar with Brawl.

tool4ever: I think the rule exists because Wastes haven't been Standard legal usually, but right now they are again
I need to debug why they get added on top, if that's actually the problem you're trying to address?

Madwand99: Added another fix for Brawl land generation. This was a way more subtle problem than I thought at first! The issue was that addLands() could add the chosen Brawl basic land type without fully decrementing landsNeeded, so Wastes could then be added on top afterward.

The problem I was originally trying to address was commander deck generation 1) generating many illegal decks and 2) not using probabilities for individual cards correctly, but decks for other formats got caught up in my tests. I think it's a good thing this is being looked at, regardless.

tool4ever: Right - not to sound too harsh but this is another example why your PR aren't easy to review as your fix still isn't correct, the comment just above even warns about this:
> // do not update landsNeeded until after the loop

And missing such "due dilligence" really makes me wary of the more complex parts like your weighting :/

But like you said it's still nice this got uncovered - I'll cherry pick this part too

forge-gui/src/main/java/forge/gamemodes/limited/CardThemedDeckBuilder.java
```diff
@@ -666,6 +659,7 @@ private void addLands(final int[] clrCnts) {
 
                 for (int j = 0; j < nLand; j++) {
                     deckList.add(getBasicLand(i));
+                    landsNeeded--;
```
tool4ever: I took some time to check that the weighting improves generation :)

after rebasing this line is redundant now

Madwand99:
Updated this branch after the related cleanup/weighting/Brawl fixes were split out and merged separately.

Removed from this PR:
- The `DeckgenUtil` weighted matrix sampling changes, since those are now covered by #11129.
- The AI `Collections.shuffle(..., MyRandom.getRandom())` cleanup, also now upstream.
- The Brawl/colorless/Wastes land-generation fixes, since those are now covered by #11061.
- Some leftover cleanup from the earlier broader version of this branch.

Still remaining in this PR:
- Commander card-based deck generation now avoids duplicate card names for singleton formats, while still respecting card-specific deckbuilding limits from `DeckFormat#getMaxCardCopies(card)`.
- The Commander card-based builder no longer fills short decks by extending playsets, since that can create illegal Commander-family decks.
- Shared card-themed generation paths now use one copy-limit helper based on `DeckGeneratorBase.maxDuplicates` plus format/card-specific copy limits, instead of adding a separate Commander-only duplicate filter at each call site.

The current PR scope should now be only the remaining legality safeguards for generated card-themed / Commander card-based decks.

forge-gui/src/main/java/forge/gamemodes/limited/CardThemedCommanderDeckBuilder.java
```diff
@@ -32,6 +35,8 @@ public CardThemedCommanderDeckBuilder(PaperCard commanderCard0, PaperCard partne
             this.aiPlayables = Lists.newArrayList(availableList);
         }
         this.availableList.removeAll(aiPlayables);
+        this.aiPlayables = uniqueCardNamesForSingletonDeck(aiPlayables);
```
tool4ever, line 38: I still fail to see why this should be needed

Madwand99: This was originally added to normalize the Commander card pool before generation, since `preSelectedCards` can contain duplicate card names from the matrix / printings and earlier add paths could select those duplicates.

That said, after the later refactor to enforce copy limits at the actual card-add points, I think this constructor-level filter is now mostly defensive/redundant for main deck legality. The more important pieces are the shared `addCardForGeneration` / `limitCopiesForGeneration` checks and the Commander override of `extendPlaysets()`.

I’m fine removing this line if we want the PR to rely only on the uniform add-time copy-limit logic. That said, I don't think the extra guard is causing any harm.

tool4ever: I was more thinking that even having duplicates at this point shouldn't happen already

Madwand99: Yes, duplicates shouldn't happen, but I see no reason not to put in a guard just in case. Happy to remove it if this is your preference.

tool4ever: huh?
it's not like we don't have control over the code before this constructor so if possible bugs should be fixed at the source rather than trying to account for it downstream
and if a commander matrix contains (illegal) duplicates that sounds like an error with the generation to me 🤷‍♂️

Madwand99: Agreed. I have the guard in because I can't guarantee what future versions of the .dat files will be; it is possible that errors will be made generating them. Happy to remove the guard if you prefer that.

jamincollins:
## Findings
#### 1. getMaxCopiesForGeneration returns maxDuplicates for Commander when singleton not enforced
getMaxCopiesForGeneration uses `formatMax == format.getMaxCardCopies() ? maxDuplicates : formatMax`. For Commander `format.getMaxCardCopies()` is 1. If the builder is instantiated without `setSingleton(true)`, `maxDuplicates` defaults to 4. The comparison `1 == 1` is true, so the method returns 4, allowing four copies of a card in a singleton format. The caller `canAddCardForGeneration` then permits illegal decks.

Test case:
```java
DeckFormat format = DeckFormat.Commander;
CardThemedDeckBuilder builder = new CardThemedDeckBuilder(new DeckGenPool(...), format);
// maxDuplicates defaults to 4, setSingleton never called
PaperCard card = ...; // normal non-land
int max = builder.getMaxCopiesForGeneration(card);
// max == 4, but Commander allows 1
```
Impact: Illegal decks can be generated if the singleton flag is not set before generation, violating CR 903.12e.

#### 2. uniqueCardNamesForSingletonDeck does not prevent duplicate basic lands from being offered to the generator
`uniqueCardNamesForSingletonDeck` filters by `format.getMaxCardCopies(card)`. For basic lands `getMaxCardCopies` returns `Integer.MAX_VALUE` via `canHaveAnyNumberOf`. The `currentCount < maxCopies` test is always true, so multiple PaperCard instances with the same basic land name are retained in `aiPlayables` and `availableList`. Later `addCards` skips lands entirely, but `addNonBasicLands` and mana curve logic can still be influenced by the inflated pool.

Test case:
```java
PaperCard plains1 = ...; PaperCard plains2 = ...; // same name
List<PaperCard> cards = List.of(plains1, plains2);
List<PaperCard> filtered = builder.uniqueCardNamesForSingletonDeck(cards);
// filtered.size() == 2, duplicates remain
```
Impact: Pool inflation and nondeterministic selection; legality is preserved but weighting is skewed.

#### 3. extendPlaysets no-op can leave decks undersized
`CardThemedCommanderDeckBuilder.extendPlaysets` is overridden to empty. Previously the base implementation filled gaps by duplicating cards. With singleton enforcement and no fallback, a limited card pool can result in `deckList` < `targetSize`. No error is raised; the deck is simply short.

Test case:
```java
// pool contains 70 unique legal cards, targetSize = 99
DeckGeneratorBase gen = new CardThemedCommanderDeckBuilder(cmdr, null, pool, false, DeckFormat.Commander);
gen.setSingleton(true);
CardPool deck = gen.getDeck(99, false);
// deck.countAll() < 99
```
Impact: Incomplete decks, degraded AI play.

#### 4. uniqueCardNamesForSingletonDeck assumes non-null keyCard and secondKeyCard
The method does `countsByName.put(keyCard.getName(), 1)` without null check. If `keyCard` is null, NPE. The constructor assumes non-null, but callers could reuse the builder.

Test case:
```java
CardThemedCommanderDeckBuilder b = new CardThemedCommanderDeckBuilder(null, null, list, false, DeckFormat.Commander);
// NPE at countsByName.put
```
Impact: Crash on malformed input.

#### 5. getMaxCopiesForGeneration fails for formats with generic max != 4
When `format.getMaxCardCopies()` is not 4, e.g., a custom format with max 3, `maxDuplicates` defaults to 4 for non-singleton. The equality check returns `maxDuplicates` for normal cards, exceeding format limits.

Test case:
```java
DeckFormat custom = ...; // getMaxCardCopies() == 3
builder.setSingleton(false); // maxDuplicates = 4
int max = builder.getMaxCopiesForGeneration(normalCard);
// max == 4 > 3
```
Impact: Illegal copies in non-standard formats.

#### 6. aiPlayables and availableList filtered independently
`uniqueCardNamesForSingletonDeck` is applied separately to `aiPlayables` and `availableList` with fresh count maps seeded only with commander/partner. A card name can appear once in each list, allowing up to two copies of the same card to be selected across the two pools during generation, even though the final deck must be singleton.

Test case:
```java
// card X in both aiPlayables and availableList
// both lists keep X because counts reset per list
```
Impact: Potential duplicate selection across generation phases.

#### 7. Color identity combination changed
Constructor now uses `ColorSet.combine` instead of bitmask OR. Semantically equivalent, but previous code combined keyCard with itself twice. The new code is clearer but changes order of operations for Oathbreaker where secondKeyCard is ignored for colors. No regression observed, but behavior diverges from previous implementation for edge cases where `ColorSet.combine` normalizes differently.

#### 8. DeckgenUtil weighted random sort changes selection order
`getWeightedRandomSortKey` uses `Math.log(random)/weight`. The previous code shuffled then filtered by legality. The new deterministic sort by weighted key changes the order in which `preSelectedCards` are populated, affecting downstream `limitCardsToCommanderBracket` and deck generation. The change is intentional for weighting, but the interaction with `format.isLegalCard` filter after sorting means low-weight illegal cards are still evaluated.

Test case:
```java
List<Map.Entry<PaperCard,Integer>> pool = ...;
List<Map.Entry<PaperCard,Integer>> sorted = DeckgenUtil.getWeightedRandomizedCardPool(pool);
// order differs from previous shuffle
```
Impact: Non-deterministic but different distribution; may mask legality bugs.

### Performance and security
#### 9. Repeated getDeckListCountsByName allocations
`addCards`, `addCreatures`, `limitCopiesForGeneration`, `addNonBasicLands` each call `getDeckListCountsByName`, which iterates over `deckList` and builds a new HashMap. For large decks this is O(n²) in the number of cards added.

Test case:
```java
// deckList grows to 99 cards, each add operation rebuilds map from scratch
```
Impact: Acceptable for current sizes but unnecessary allocation pressure.

## Verdict
Request Changes
The PR improves legality enforcement but introduces a critical bug in `getMaxCopiesForGeneration` that can allow illegal copy counts when `maxDuplicates` is not aligned with the format's generic max, and the `extendPlaysets` no-op risks undersized decks. Additional edge cases around null safety and duplicate filtering across pools need hardening before merge.

## Recommended Fixes
1. Make `getMaxCopiesForGeneration` return `Math.min(maxDuplicates, formatMax)` or derive `maxDuplicates` from the format's generic max on construction.
2. Guard `uniqueCardNamesForSingletonDeck` against null `keyCard`/`secondKeyCard` and consider sharing a single count map across `aiPlayables` and `availableList` to enforce global singleton.
3. Provide a fallback when `extendPlaysets` is no-op: either fail clearly or allow basic lands to fill gaps, and log when target size is not reached.
4. Ensure `setSingleton(true)` is enforced in all code paths creating `CardThemedCommanderDeckBuilder`, or make `maxDuplicates` default to `format.getMaxCardCopies()`.

---

Madwand99:
I removed the constructor-level `uniqueCardNamesForSingletonDeck` filtering. I agree that silently deduping inside `CardThemedCommanderDeckBuilder` is not the right place to handle malformed candidate input; the actual legality guard should be at card-add time.

I also tightened `getMaxCopiesForGeneration` so normal cards are capped by both the generator duplicate limit and the format’s generic max, while preserving card-specific `DeckFormat#getMaxCardCopies(card)` exceptions.

PR #10779, opened 2026-05-27
Commander draft

### LINKED ISSUE #10734: Commander Draft
There are three different sets that draft you immediately into a commander match. Let's see if we can add a way to let players and AI draft a commander deck (including the free commanders that help make commander decks more functional). 

Here's the general rules:

Packs are 20 cards usually pick 2. 
Deck size is 60 (so basically the same size as Brawl)
Singleton rule does not apply (since it's a limited format)
Players may always use up to two of The Prismatic Piper or the Faceless One as a commander, even if they didn't draft one. These cards MAY NOT be added for free to your main deck. 

If a bracket is formed, it'll be the same as the Conspiracy draft. 4 player first pod into a 2 player finals. 
Everyone starts at 40 life. The first pod is 4 players. 

=================================================================

903.13. [Commander Draft](https://mtg.fandom.com/wiki/Commander_draft)
903.13a Commander Draft is an option for a different style of Commander game. It consists of a draft (a style of limited play where players choose cards from sealed booster packs to build their decks) followed by a multiplayer game. The Commander Draft option uses Commander Legends booster packs by default.
903.13b A draft typically consists of three draft rounds. In each draft round, each player opens a booster pack, drafts two cards by placing them in a face-down pile in front of them, then passes the remaining cards to the next player. Each player then drafts two cards from the booster pack passed to them and passes the remaining cards. This procedure continues until all cards in that draft round have been drafted.
903.13c In the first and third draft rounds, booster packs are passed to each player’s left. In the second draft round, booster packs are passed to each player’s right.
903.13d During the draft, a player can look only at cards in the booster pack they are currently drafting from and cards they have already drafted. A player may not reveal drafted cards to other players unless an ability instructs them to.
903.13e After the draft is complete, the cards a player drafted become that player’s card pool. If the draft contained draft boosters from Commander Legends or Commander Masters, each player may add up to two cards named The Prismatic Piper to their card pool, but only if those cards are used as the player’s commander(s). If the draft contained draft boosters from Commander Legends: Battle for Baldur’s Gate®, each player may add up to two cards named Faceless One to their card pool, but only if those cards are used as the player’s commander(s).
903.13f Commander Draft deck construction follows the same rules as Commander deck construction (see rule 903.5) with three exceptions: (1) A player’s deck must contain at least 60 cards. There is no maximum deck size. (2) A player’s deck may include any number of cards from that player’s card pool with the same name. (3) If the draft contained draft boosters from Commander Masters, any card which can be a player’s commander by itself and whose color identity includes one or fewer colors is considered to have the partner ability for the purposes of deckbuilding. (See rule 702.124, “Partner.”)
903.13g Commander Draft games follow the same rules as Commander games. See rules 903.6–903.11.

tehdiplomat:
https://github.com/Card-Forge/forge/issues/10734


forge-core/src/main/java/forge/card/CardRules.java
```diff
@@ -335,6 +335,20 @@ public boolean canBeCommander() {
         return false;
     }
 
+    public boolean canBePartnerCommanders(CardRules b, boolean commanderDraft) {
+        if (commanderDraft) {
+            if (!canBeBackground() && !canBeCommander()) {
+                return false;
+            }
```
Jetz72: Think this would allow two backgrounds to partner with one another.

forge-gui-desktop/src/main/java/forge/screens/home/sanctioned/CSubmenuDraft.java
```diff
@@ -112,8 +113,13 @@ private void startGame(final GameType gameType) {
             return;
         }
 
+        // Detect commander draft: human deck has a Commander section with cards
+        final boolean isCommanderDraft = humanDeck.getDeck().has(DeckSection.Commander)
+                && !humanDeck.getDeck().get(DeckSection.Commander).isEmpty();
+        final GameType effectiveGameType = isCommanderDraft ? GameType.CommanderDraft : gameType;
```
Jetz72, line 117: Could this be recognized by way of the deck being tagged "CommanderDraft"?

forge-core/src/main/java/forge/card/CardRules.java
tehdiplomat: Ah yea, lemme fix that up

tehdiplomat:
I think this is working pretty well except the deck editor isn't updating the sideboards properly.

PR #10780, opened 2026-05-27
Update Commander deckgen matrix from EDHREC data

Madwand99:
## Summary

Replaces `forge-gui/res/deckgendecks/Commander.dat` with a newly generated Commander deckgen matrix based on EDHREC commander data. There are about 1000 commanders, compared to about 200 in the previous data set.

The source data was scraped from EDHREC’s Top Commanders, Past 2 Years dataset. The scrape collected raw commander/card association data first, then transformed it into Forge’s existing serialized matrix format.

For best results, combine with #10774.

Included are the Python+Java files I used to scrape EDHREC, and a README for using them.

Thanks to Codex for help with code and testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,3 +97,5 @@ __pycache__
 
 # Ignore Claude configuration
 .claude
+# EDHREC commander matrix tool intermediates
+forge-gui/tools/build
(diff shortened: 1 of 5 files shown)


tool4ever:
ugh guess I should have left a blocking comment here but might as well let users test the quality instead... 🤷‍♂️

still gonna ping @austinio7116 for potential input on future adjustments

Agetian:
It didn't hang or crash and seemed ro generate valid Commander decks, so it seemed alright to let this one be playtested in snapshots. If it ends up performing poorly we can promptly revert it.

PR #10783, opened 2026-05-27
Edition updates: A-F various fixes

paulsnoops:
Fixes for set names, release dates, collector numbers, rarities, artists, flavor names and the odd new card/variation here and there. Fixed ARC edition and moved schemes from ARC > OARC.


Jetz72:
How well do existing scheme decks hold up when the set code suddenly changes?

paulsnoops:
> How well do existing scheme decks hold up when the set code suddenly changes?

I've added an Alias as a fallback, works on testing with ARC schemes in my deck. Pretty sure it just finds the next set with the same name card anyway as I saw some DSC schemes in my deck even though they're listed as ARC.

Agetian (approved):

PR #10791, opened 2026-05-27
Fix off-thread trackable-graph walk in network showPromptMessage

MostCromulent:
## Summary

Network prompts walk the trackable graph before dispatch so the client has the entities the prompt references — but that walk is only safe on the game thread. 

- #10701 fixed a race where the off-thread await-input timer walked the graph: it made the message-only `showPromptMessage` a pure send with no walk, since a text-only prompt needs no graph state.
- #10776 then merged the message-only `showPromptMessage` and the card-bearing `showCardPromptMessage` into a single method with `syncAndSend` dispatch, with the message-only form left as a default that delegates to it with a null card. That re-introduced the original race: the await/auto-pass timer fires the message-only prompt on the EDT while the game thread is live, walking the graph concurrently with game-thread mutation. 

On the host this surfaced as ConcurrentModificationException in `Tracker.getDelayedPropsFor`, ArrayIndexOutOfBoundsException while serializing a resync's full state, and an unrecoverable client desync loop.

## Fix

`showPromptMessage` gates the graph walk on the `card` argument: a null card (the timer path, which runs while the game thread is live) sends without walking; a non-null card — which only originates from blocking input, with the game thread parked on its latch — keeps `syncAndSend` so the client still receives the referenced `CardView`.

---


tool4ever (approved):
ugh I definitely screwed that up 🤦‍♂️

PR #10792, opened 2026-05-27
Fix flaky test failures from card DB load racing a test timeout

MostCromulent:
## **Bug**

CI fails intermittently with ~20 test failures in `forge-gui-desktop` that disappear on re-run: one `ThreadTimeoutException` in `ReplacementHandlerTest.testPerpetualEntersTappedReplacementEffect`, followed by ~19 `RuntimeException: wrong token name:…` across `GameSimulationTest`, `AutoPaymentTest`, and `SpellAbilityPickerSimulationTest`. It reproduces on both the scheduled snapshot build and the push-triggered `Test with Java 21` workflow, on the same commit, with no code change between a failing run and a passing re-run.

## **Root cause**

The module's tests run in a single shared JVM, and the AI simulation tests share one lazily-loaded, memoized `StaticData` (the card and token databases). Whichever test first calls `FModel.getMagicDb()` pays the multi-second load. When that test carries `@Test(timeOut = 2000)` — as `testPerpetualEntersTappedReplacementEffect` does — a slow runner can exceed 2s mid-load, so TestNG interrupts the thread; `CardStorageReader.executeLoadTask` then swallows the `InterruptedException` and returns a partially-built database. Because the supplier is memoized, that broken `StaticData` with an empty token map is cached for the rest of the JVM, so every later token lookup throws `wrong token name`. One unlucky timeout cascades into ~20 failures.

The failure is intermittent because two independent things vary between runs. First, no `runOrder` is configured, so Surefire uses its default `filesystem` order — the order in which the OS enumerates the compiled test classes — which it documents as likely differing across systems and JVMs. Each CI run is a fresh checkout and compile on a freshly provisioned runner, so that order is not pinned and a different test can be the one that first triggers the lazy DB load. Second, the load only overruns the 2s budget on a slower runner. A run passes whenever an untimed test happens to trigger the load first, or the load completes in time — which is why re-running the job cleared it.

## **Fix**

The fix has two parts. `AITest` now performs the one-time `FModel.initialize()` in an untimed `@BeforeMethod` instead of inside `initAndCreateGame()` (which runs in the timed test body), so the database load can never be interrupted by a per-test timeout regardless of run order or runner speed. As defense in depth, `CardStorageReader.executeLoadTask` now propagates the `InterruptedException` (restoring the interrupt flag) instead of swallowing it, so an interrupted load fails loudly and is never cached as if it were complete.

---


Hanmac (approved):

tool4ever (approved):
still weird it just started to happen but analysis makes sense, so worth a shot

PR #10801, opened 2026-05-28
TrackableObject: View getter return empty List

Hanmac:
- [x] `updateAttachedCards` should probably be split of and call addCard/removeCard instead

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardView.java
+++ b/forge-game/src/main/java/forge/game/card/CardView.java
@@ -22,6 +22,9 @@ import forge.trackable.TrackableObject;
 import forge.trackable.TrackableProperty;
 import forge.trackable.Tracker;
 import forge.util.*;
+import forge.util.collect.FCollection;
+import forge.util.collect.FCollectionView;
+
 import org.apache.commons.lang3.StringUtils;
 
 import java.util.*;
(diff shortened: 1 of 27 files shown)


forge-game/src/main/java/forge/game/card/CardView.java
```diff
@@ -772,8 +771,8 @@ public PlayerView getEnchantedPlayer() {
         return null;
     }
 
-    public FCollectionView<CardView> getGainControlTargets() {
-        return get(TrackableProperty.GainControlTargets);
+    public List<CardView> getGainControlTargets() {
+        return Objects.requireNonNullElse(get(TrackableProperty.GainControlTargets), List.of());
```
tool4ever: hmn why not just `FCollection.getEmpty()`? List always has this strong implication of including duplicates

Hanmac: These are only for Iterating
And the return is immutable/unmodifiable anyway (`FCollection.getEmpty()` is that too)

imo there is no need for it to be `FCollectionView`, right?

tool4ever: well this is about semantics, returning a *View also makes immutability clearer than a plain `List` at least (unless you inspect code more closely or just crash at runtime lol)

on that note I still think it'd be safer again if such methods would only return Interfaces that don't expose extra mutator methods so then it wouldn't even compile - related to my comment here: https://github.com/Card-Forge/forge/issues/3397#issuecomment-4077459803

:/

Hanmac: First, i wanted to return just Collection instead, (that doesn't have mutators)
but this one wants to use `get(0)` on them (I noticed, i forgot to remove `!= null` check there)
https://github.com/Card-Forge/forge/blob/2b75c4b90c2c08c16269ec317d287c8fd324d2f7/forge-gui/src/main/java/forge/player/HumanCostDecision.java#L1509-L1510

i will re-think about using `FCollectionView` on them

Collection still has `remove` so it doesn't ease your problems

tool4ever (dismissed):
aside from my nitpick it seems to work

tool4ever (approved):

PR #10803, opened 2026-05-29
Add maximum Commander bracket controls for generated decks

Madwand99:
This adds support for limiting generated Commander decks to a selected maximum bracket.

Changes include:
- Added capped Commander deck generation overloads for both random Commander generators.
- Filter generated deck candidate cards incrementally with CommanderBracketCalculator.
- Added a Maximum Bracket dropdown for Commander lobbies, shown only when Commander is selected.
- Persisted the selected maximum bracket preference, defaulting to 5.
- Applied the selected bracket when starting games and when viewing generated Commander decks.
- Added Bracket: X to Commander match portrait tooltips, calculated once per deck.
- Updated the User Guide mention of avatar hover details.

Thanks to Codex for help with code and tests.


forge-gui/src/main/java/forge/deck/DeckgenUtil.java
```diff
@@ -726,14 +738,7 @@ else if (commander.getRules().canBePartnerCommander()) {
             case Brawl: //for Brawl - add additional filterprinted rule to remove old reprints for a consistent look
                 colorList = IterableUtil.filter(colorList,FModel.getFormats().getStandard().getFilterPrinted());
                 break;
-            case Oathbreaker:
```
tool4ever, line 729: when you refactor stuff so only a single case remains the switch itself can be cleaned up too

forge-gui-desktop/src/main/java/forge/screens/match/views/VField.java
```diff
@@ -406,6 +412,45 @@ public void updateDetails() {
         final boolean highlighted = isHighlighted();
         this.avatarArea.setBorder(highlighted ? borderAvatarHighlighted : borderAvatarSimple );
         this.avatarArea.setOpaque(highlighted);
-        this.avatarArea.setToolTipText(player.getDetailsHtml());
+        this.avatarArea.setToolTipText(getPlayerDetailsHtml());
```
tool4ever, line 418: there should be no reason to do this inside desktop module instead of the unified place

but either way this seems like the wrong approach:
- you can always view your own deck and check the info from there
- and for opponents in general you would be providing small but still illegal (hidden) information

if you feel more is still needed better rework outside of this PR and consider if keeping the cheating to Dev Mode wouldn't be better for clear distinction anyway

Madwand99: For opponents, knowing the bracket of their deck would not be considered cheating in Commander. Bracket conversations are intended to be a part of the pre-game conversation and all players SHOULD know the bracket of all other decks at the table.

forge-gui/src/main/java/forge/deck/DeckgenUtil.java
Madwand99: Flxed!

forge-gui-desktop/src/main/java/forge/screens/match/views/VField.java
tool4ever: 👎:

a) let's stop making such broad assumptions for all players. yes bracket conversation is recommended but maybe two friends would like to battle against each other with random generated decks without knowing their power levels...

b) you're aware Forge has more play modes than Constructed? I don't think e.g. in Adventure you want to soften the challenge like that, besides the questionable flavor of "Whenever you encounter a Dinosaur enemy you first have a pre-game conversation with him..."

Madwand99: Fair enough. I can remove the information in Adventure mode and when the bracket is unrestricted. However, if the bracket is restricted, the player should have some way of verifying that the restriction actually worked.

Madwand99: Added a limit on bracket display. It won't appear in Adventure mode, or when bracket is unrestricted i.e. when that pre-game conversation did not take place.

tool4ever: ok, thanks - that compromise seems fine :)

unfortunately the preferences call means adressing my first point by moving it to forge-gui becomes more tricky - but that can be tweaked when mobile support is added later

tool4ever:
A future follow-up might try the inverse direction: allow selecting a minimum target level

though that'd probably be a bit more complicated 🤔

Madwand99:
> A future follow-up might try the inverse direction: allow selecting a minimum target level
> 
> though that'd probably be a bit more complicated 🤔

In theory, it would be massively complicated to deliberately build for a certain bracket. However... the current system is not that far away. With #10774, the decks built are reasonably good quality. Bracket 4 performance is a bit harder to guarantee, and with bracket 5 you pretty much have to have human-built decks. I am currently planning to work with https://commanderbracket.app to improve bracket estimation, which should help a lot if cooperation works out.

forge-gui-desktop/src/main/java/forge/screens/home/VLobby.java
```diff
@@ -260,9 +274,14 @@ public VLobby(final GameLobby lobby) {
...
             pnlStart.setOpaque(false);
-            pnlStart.add(btnStart, "align center");
+            maximumCommanderBracket.addActionListener(e -> applyMaximumCommanderBracketToDeckChoosers());
+            maximumCommanderBracketFrame.add(newLabel("Maximum Bracket:"), "w " + START_ROW_LABEL_WIDTH + "px!, h 30px!");
+            maximumCommanderBracketFrame.add(maximumCommanderBracket, "w " + START_ROW_COMBO_WIDTH + "px!, h 30px!");
+            maximumCommanderBracketFrame.setOpaque(false);
+            addConstructedStartControls();
             // Start button event handling
             btnStart.addActionListener(arg0 -> {
+                updateCommanderGeneratedDeckSelections();
```
tool4ever: don't understand why you're adding an extra hook for your logic to this button

Madwand99: Good point. I changed this so Start no longer blindly regenerates generated Commander decks. We now track the maximum bracket used when the generated deck was stored, and only refresh it on Start if the selected bracket cap has changed since then. So the hook is just a stale-selection guard now, not an unconditional regeneration path.

tool4ever: random decks should already get refreshed on start, this is code bloat

Madwand99: I don’t think generated Commander decks are refreshed automatically on Start here. The chooser generates the deck when selected, then fireDeckChangeListener stores that deck in the lobby slot, and GameLobby.startGame() uses slot.getDeck(). This hook is only a stale-selection guard: if the user selected a generated Commander deck under one bracket cap, then changed the cap before pressing Start, we regenerate that stored deck once under the new cap. Otherwise it does nothing.

tool4ever: see `LocalLobby.onGameStarted()` implementation
it's a bit roundabout and could probably be reworked, but I'll remove this for now so we don't introduce duplicate handling

Madwand99:
These "Reduce bloat" changes have broken the PR. I can now select bracket 2, but play a game with bracket 4 generated decks. Can we revert please?

tool4ever:
hmn that's weird, I really don't want to pass the maximum around like you've done previously

grabbing the preference like the other deckgen options should work, I'll try to debug

Madwand99:
> hmn that's weird, I really don't want to pass the maximum around like you've done previously
> 
> grabbing the preference like the other deckgen options should work, I'll try to debug

Yeah, I think reading the preference in deckgen is fine and I agree that’s cleaner than passing the cap through the chooser/lobby. The issue I found is that changing the dropdown does not necessarily regenerate the already-stored deck for the current lobby slot.

Generated Commander decks are created when the deck selection command runs, then stored via fireDeckChangeListener. startGame() later uses slot.getDeck(). So if I select a generated Commander deck while the cap is 5, then change the cap to 2 and press Start, the stored deck can still be the cap-5 deck. The existing LocalLobby.onGameStarted() refresh happens after the match has already been built, so it affects the next lobby state rather than the game being launched.

I think the minimal fix is not to pass the cap around, but to refresh selected generated Commander decks when the Maximum Bracket dropdown changes. That way deckgen still reads the preference, but the stored lobby deck is no longer stale.

tool4ever:
Right, I expected that this should have been wired up correctly for these related options already:
image

but it looks like toggling them without selecting another deck would result in the same problem

so turns out we'll want a unified solution

tool4ever:
I believe it should work now? 😅

Madwand99:
> I believe it should work now? 😅

Yes, seems to work. I'm relieved we figured out a solution!

tool4ever (approved):

Issue #10750, opened 2026-05-23
Mana Record and MagicColor


Hanmac:
Part of #8682 

Big Mana Producers can cause Memory Crash
And also some cleanup for using Color Enum

- Mana Produced is a very long String "B B B", it should be refactored into a map `{B => 3}`
- then each Mana Produced by a SpellAbility create a Mana Record. If possible try to reuse existing Records. So it becomes `{Mana => 3}` #10811
- Same with ManaPool, instead of using byte, have it use Color, like `Color => Mana => 3` #10824
- adding 3 times Mana, caused 3 times `updateManaForView` and `new GameEventManaPool`, this is slowing down the system #10812

Guava [Multiset](https://github.com/google/guava/wiki/NewCollectionTypesExplained#multiset) would be nice to hide the `{X => 3}` structure 🤔

Hanmac:
Guava [Multiset](https://github.com/google/guava/wiki/NewCollectionTypesExplained#multiset) is very interesting idea.
Both for the Mana Produced by the Ability (currently String "B B B"),
but also for the Mana Records, (currently List<Mana>) that is going to be added to the mana pool.

Hm for ManaPool, i need to see if its worth it,
to have the floatingMana as `Map<Color, Multiset<Mana, N>>`
or if it could just be a `Multiset<Mana, N>` and have the color just be a filtered View

Hanmac:
@MostCromulent your (not AI) opinion on this?

> adding 3 times Mana, caused 3 times `updateManaForView` and `new GameEventManaPool`, this is slowing down the system

I think the GameEvent should only be once per color, or only once with a Set of Colors (or still a multiset?)

the update for the GUI at this, doesn't care about the colors
https://github.com/Card-Forge/forge/blob/2b75c4b90c2c08c16269ec317d287c8fd324d2f7/forge-gui/src/main/java/forge/gui/control/FControlGameEventHandler.java#L498-L500

tool4ever:
Right, doesn't make much sense to split mana produced from one game action into same pool as multiple events internally 🤔

Want me to refactor it into holding `List<Byte>`?

Hanmac:
@tool4ever I'm going to refactor the stuff over the weekend

especially I want to use `Multiset<Mana>` to prevent it from Crashing when making big chunk of Mana at the same Time

Hanmac:
I want to change the color from Byte (for ManaAtom) to MagicColor enum
but that would change the value for colorless I would need to update later.

PR #10811, opened 2026-05-30
AbilityManaPart: use Multiset for produced

Hanmac:
Part of #10750

This should help with producing big amount of Mana per ManaAbility,
it only creates one Mana Record instance per color type.

the multiset is still expanded when adding them to the ManaPool (for now)

I wanted to do the ManaPool too, but it got too complicated, so i do something smaller


tool4ever (approved):
as far as I can tell nothing seems to care about duplicate objects there

Hanmac:
Technically, the Mana records are still duplicated, 
just from `[M, M]` to `{M=>2}`

PR #10816, opened 2026-05-31
Online draft/sealed bug fixes and polish (desktop)

MostCromulent:
## Summary

A set of bug fixes and polish for network draft/sealed on desktop:

- **Sideboard cap in limited matches:** networked limited games launched as Constructed, so the deck-building sideboard dialog wrongly demanded the pool be cut to the 15-card Constructed sideboard limit. They now launch as a limited game type (no sideboard cap).
- **Edit button now loads the deck:** the per-row Edit button in the Net Event Decks list opened an empty editor because it used the wrong deck pool; it now opens the selected deck, matching double-click behaviour.
- **Correct editor per format:** draft pools open in the Draft deck editor and sealed pools in the Sealed editor, instead of everything opening in the Sealed editor.
- **Immediate event filter:** choosing a past event now filters the deck list to that event right away, instead of requiring the "only allow decks from this event" checkbox to be toggled off and back on.
- **Past events sorted newest first:** the load-past-event picker now lists events by date, newest to oldest.

---

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/itemmanager/DeckManager.java
+++ b/forge-gui-desktop/src/main/java/forge/itemmanager/DeckManager.java
@@ -8,7 +8,6 @@ import java.awt.RenderingHints;
 import java.awt.event.MouseEvent;
 import java.util.*;
 import java.util.Map.Entry;
-import java.util.function.Consumer;
 
 import javax.swing.JMenu;
 import javax.swing.JTable;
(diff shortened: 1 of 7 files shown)


forge-gui-desktop/src/main/java/forge/itemmanager/DeckManager.java
```diff
@@ -327,6 +334,10 @@ protected List<String> getFilteredSetCodesInCatalog(){
     }
 
     public void editDeck(final DeckProxy deck) {
+        if (editDeckCommand != null) {
+            editDeckCommand.accept(deck);
```
tool4ever: seems roundabout:
can't we just check `DeckProxy` eventTag directly to branch?

tool4ever (approved):

PR #10817, opened 2026-05-31
handleStayOpen complained in the IDE

Hanmac:
my IDE was complaining about that:

> The method handleStayOpen(JMenuItem, MouseEvent, Consumer<MouseEvent>) in the type MenuUtil is not applicable for the arguments (new JCheckBoxMenuItem(){}, MouseEvent, Consumer<MouseEvent>)


MostCromulent:
I'm not getting any complaint on IntelliJ - AI seems to think it's an Eclipse thing. 🤷‍♂️
In any event, seems safe to merge.

tool4ever:
same here - Eclipse giving me wrong warnings was also one of the reasons why I stopped using it :P

PR #10820, opened 2026-05-31
Set name in WrappedAbility

Eradev:
Forgot to carry on the name.


forge-game/src/main/java/forge/game/trigger/WrappedAbility.java
```diff
@@ -45,6 +45,9 @@ public WrappedAbility(final Trigger regtrig0, final SpellAbility sa0, final Play
         sa = sa0;
         sa.setTrigger(regtrig0);
         decider = decider0;
+
+        // Support for custom named abilities
+        setName(sa.getName());
```
tool4ever: should just use proxy-method like all other

tool4ever (approved):

PR #10822, opened 2026-05-31
Update MTGP_05.pzl

AncoronDS:
Fixed where you could win the puzzle by passing the turn.


Agetian:
Can you try also adding this:

activephaseadvance=UPKEEP

to see if it correctly moves to player's upkeep phase automatically from cleanup? (see e.g. PS_FIN4.pzl for an example of how it works)

AncoronDS:
I just tested it and it doesn't seem to make a difference. 

In both cases it pauses at cleanup if you set a toggle and in both cases it proceeds to upkeep if you don't.
Both in MTGP_05.pzl and in PS_FIN4.pzl.

Agetian:
Gotcha! I might need to look into that at some point. Thanks for your contribution!

PR #10824, opened 2026-06-01
ManaPool: use Multiset for floatingMana

Hanmac:
Part of #10750

Store Mana in the ManaPool as `{Mana => 3}`

### DIFF
--- a/forge-game/src/main/java/forge/game/mana/ManaPool.java
+++ b/forge-game/src/main/java/forge/game/mana/ManaPool.java
@@ -17,8 +17,11 @@
  */
 package forge.game.mana;
 
-import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.HashMultiset;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Multiset;
+import com.google.common.collect.Multisets;
+import com.google.common.collect.Sets;
 
 import forge.card.MagicColor;
 import forge.card.mana.ManaAtom;
@@ -37,6 +40,8 @@ import forge.game.spellability.SpellAbility;
 import forge.game.staticability.StaticAbilityUnspentMana;
 
 import java.util.*;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 /**
  * <p>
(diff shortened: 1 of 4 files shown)


tool4ever (commented):
Hmn, presenting me with another tricky refactor:

- Personally I'm somewhat skeptical the code really becomes cleaner by introducing these additional concepts into it

- But more importantly I'm also concerned this could degrade performance on a heavy workload class...
There seems to be way more filtering going on because you no longer store the color mapping - instead the HashMultiset will have lots of buckets based on the different Mana records

Sorry, but I'd prefer a second opinion here too before I can spend some time testing for regressions 🤔

Hanmac:
> * instead the HashMultiset will have lots of buckets based on the different Mana records

That shouldn't be much of a problem,
AbstractMapBasedMultiset does keep the total of the elements as extra variable.

https://github.com/google/guava/blob/a8e9533b499a446cccf7eff54e7775327fb119ec/android/guava/src/com/google/common/collect/AbstractMapBasedMultiset.java#L249-L251

so in case of `{A => 1, B => 2, C => 3}` it already knows the total size = `6`

and right now, there shouldn't be too much filtering going on

forge-game/src/main/java/forge/game/mana/ManaPool.java
```diff
@@ -48,30 +53,37 @@
...
     public final void addMana(final Iterable<Mana> manaList) {
-        Set<MagicColor.Color> colors = EnumSet.noneOf(MagicColor.Color.class);
-        for (final Mana m : manaList) {
-            floatingMana.put(m.getColor(), m);
-            colors.add(MagicColor.Color.fromByte(m.getColor()));
+        Set<MagicColor.Color> colors;
+
+        if (manaList instanceof Multiset<Mana> manaSet) {
+            colors = manaSet.elementSet().stream().map(m -> MagicColor.Color.fromByte(m.getColor())).collect(Collectors.toSet());
+            floatingMana.addAll(manaSet);
```
Hanmac, line 80: one reason why i don't have multiple lists per color, is that i can't use 
[Multisets.removeOccurrences](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Multisets.html#removeOccurrences-com.google.common.collect.Multiset-com.google.common.collect.Multiset-) on them.

tool4ever:
I can try and check some more later this week but right now it looks like everyone (and their mother) wants me to review their stuff 😅

PR #10825, opened 2026-06-01
Allow the grouping of identical cards

Eradev:
Added an option to group identical cards in the mobile deck editor.

image

image

### DIFF
--- a/forge-gui-mobile/src/forge/deck/FDeckEditor.java
+++ b/forge-gui-mobile/src/forge/deck/FDeckEditor.java
@@ -5,7 +5,6 @@ import com.badlogic.gdx.math.Vector2;
 import com.badlogic.gdx.utils.Align;
 import forge.Forge;
 import forge.Forge.KeyInputAdapter;
-import forge.adventure.scene.AdventureDeckEditor;
 import forge.Graphics;
 import forge.assets.*;
 import forge.card.CardEdition;
@@ -712,20 +711,10 @@ public class FDeckEditor extends TabPageScreen<FDeckEditor> {
     protected void toggleGroupIdenticalCards() {
         FModel.getPreferences().togglePrefBoolean(FPref.UI_GROUP_IDENTICAL_CARDS);
 
-        if(getCatalogPage() != null) {
-            getCatalogPage().scheduleRefresh();
-        }
-
-        if (getMainDeckPage() != null) {
-            getMainDeckPage().cardManager.updateView(false, null);
-        }
-
-        if (getSideboardPage() != null) {
-            getSideboardPage().cardManager.updateView(false, null);
-        }
-
-        if(this instanceof AdventureDeckEditor adventureEditor) {
-            adventureEditor.refresh();
+        for (TabPage<FDeckEditor> tabPage  : tabPages) {
+            if (tabPage instanceof CardManagerPage cmp) {
+                cmp.cardManager.updateView(false, null);
+            }
         }
     }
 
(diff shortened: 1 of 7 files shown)


tool4ever:
doesn't really make sense to me, why not just pile by name in that case...

Eradev:
> doesn't really make sense to me, why not just pile by name in that case...

Because in Commander you get a long horizontal scrolling. Not to mention in your collection.

It is also much easier to see which art you have instead of them being piled up.

Jetz72:
If you have both this and pile-by turned on, will the piles obscure the quantities?

> Because in Commander you get a long horizontal scrolling. Not to mention in your collection.

Perhaps at some point it'd make sense to implement wrapping for certain types of piles. It makes sense that piling by mana or power can extend out horizontally. Not so much when piling by name. 

This still seems like a nicer way of grouping a card pool by print, though. I wonder if there's a good way that it could be folded into the `ImageView`-exclusive settings instead of an editor-wide option that won't appear to do anything in the `ListView`.

Eradev:
It's just like card prices. They are hidden under the first of the pile, but displayed when selected.

image

Eradev:
I had thought about just modifying ImageView but it doesn't know if it is in a draft or maybe somewhere else where we wouldn't want to merge them. (Maybe other places where multi-selections are possible?)

forge-gui-mobile/src/forge/itemmanager/ItemManager.java
```diff
@@ -964,6 +965,21 @@ public boolean showPriceInfo() {
...
+    public void setAllowGroupIdenticalCards(boolean allowGroupIdenticalCards0) {
+        if(allowGroupIdenticalCards == allowGroupIdenticalCards0) {
+            return;
+        }
+            
+        allowGroupIdenticalCards = allowGroupIdenticalCards0;
+        if (pool != null) {
+            updateView(false, null);
+        }
+    }
```
Jetz72: If it's only meant to group cards and not items, would it make sense for the `ItemManager` to return false for `getAllowGroupIdenticalCards`, and shift the setter down to the `CardManager`? Or if this might later be expanded to other item types, might rename the symbol family to `allowGroupIdentical`.

Could alternatively add a flag to the `ItemManagerConfig` for whether to allow it, which might also handle the "`ImageView` not knowing the context" issue. It'd fit in with the others, though the list of boolean properties are already getting a little out of hand on those definitions.

forge-gui-mobile/src/forge/deck/FDeckEditor.java
```diff
@@ -703,6 +705,30 @@ protected void toggleConformity() {
...
+            getMainDeckPage().cardManager.updateView(false, null);
+        }
+
+        if (getSideboardPage() != null) {
+            getSideboardPage().cardManager.updateView(false, null);
+        }
+
+        if(this instanceof AdventureDeckEditor adventureEditor) {
+            adventureEditor.refresh();
+        }
```
Jetz72: Would it be simpler to just loop over the `tabPages`, check if they're `CardManagerPage`s, and if so call `updateView` on them?

forge-gui-mobile/src/forge/itemmanager/ItemManager.java
Eradev: I thought about the definition, but as you say, I found that it was already too much.

Jetz72 (approved):

PR #10827, opened 2026-06-01
forge-gui: add test harness and tests for pure-logic classes

allentiak:


forge-gui/src/main/java/forge/gamemodes/match/input/InputPayManaOfCostPayment.java
```diff
@@ -23,19 +24,30 @@ public InputPayManaOfCostPayment(final PlayerControllerHuman controller, ManaCos
         manaCost = cost;
         extraMatrix = matrix;
         applyMatrix();
-
-        // CR 118.3c forced cast must use pool mana
-        // TODO this introduces a small risk for illegal payments if the human "wastes" enough mana for abilities like Doubling Cube
```
tool4ever, line 28: why are you removing non-trivial comments?

allentiak: Thanks for catching that!
My bad. I'll fix it.

Hanmac:
imo there is no need for ForgePreferencesAdapter
and maybe IForgePreferences?

ForgePreferences already has a good base class like `PreferencesStore`
so wouldn't it be easier, to define an interface for `PreferencesStore` like IForgePreferences but with Template `<T>`?
(@tool4ever i might do something for this in a smaller MR)

then we might just pass FModel.getPreferences() to these functions?

forge-gui/src/main/java/forge/gamemodes/match/input/InputPayManaOfCostPayment.java
allentiak: Fixed.

allentiak:
@tool4ever, @Hanmac,

Thanks for the reviews!

@tool4ever 
I've just fixed the bug you discovered.

@Hanmac 
I will see about these classes and get back to you.

allentiak:
@Hanmac

I've just updated the PR.

On the one side, you're right: `ForgePreferencesAdapter` was unnecessary. My codebase knowledge is quite basic...
Therefore, I've removed it and had `ForgePreferences` implement `IForgePreferences` directly: all the required methods (`getPref`, `getPrefInt`, `getPrefBoolean`, `setPref`, `save`) are already declared `final` in `PreferencesStore`, so Java lets the inherited implementations satisfy the interface with no overrides. Production constructors now just pass `FModel.getPreferences()` at the call site.

On the other side, I have kept `IForgePreferences` as a narrow interface rather than going generic (`IPreferencesStore<T>`) for one hard reason: those same `final` methods on `PreferencesStore` can't be stubbed by Mockito's `mock-maker-subclass` — so without an interface, there's no way to inject a fake in tests. A generic interface would work too, but the narrow form is simpler and follows ISP: each class only sees the 3–4 methods it actually uses.

What do you think?

tool4ever (dismissed):
> My codebase knowledge is quite basic...

well I mean you usually only show up every few months at most :P
combining that with a 2000+ lines diff PR might not be the most sensible choice?

Anyway, in general:

a) this might offer some value but if it does it's covered under quite a bit of bloat...
I'd love if someone takes the time to write tests around more complicated interactions but this looks more like an attempt to generate as much pretty generic unit tests as possible - I'm not sure Forge as a project is a great fit for such kind of overhead :/

Examples like testing our own getters/setters seems way too self-absorbed:

```
    @Test
    public void reportNetGameAfterSetNetGame() {
        game.setNetGame();
        assertThat(game.isNetGame()).isTrue();
    }

    @Test
    public void startWithNormalGameSpeed() {
        assertThat(game.getGameSpeed()).isEqualTo(PlaybackSpeed.NORMAL);
    }

    @Test
    public void useTheEnumNameAsItsDisplayName() {
        for (MassSelectMode mode : MassSelectMode.values()) {
            assertThat(mode.getName()).isEqualTo(mode.name());
        }
    }

    @Test
    public void beInstantiableAsClearAbilityOrders() {
        assertThat(new YieldUpdate.ClearAbilityOrders()).isNotNull();
    }
```

b) I don't favor this approach of bending our code to be able to support everything being self-contained. Preferences alone are used in so many places, you'd have to add constructors for inserting your mocks almost everywhere! Such a mix of test and production code just invites technical debt on its own. 🤔

And in the end Forge is complex enough that simply running tests under a more complete environment could likely still yield more accurate results. 🤷‍♂️

sorry, in this state 👎 from me

allentiak:
@tool4ever 
Thanks for taking the time to review this PR.

I think you're right in that I didn't take the time to polish the code as much as I should have, and in that it was quite a big PR.

I will restart this in a smaller scale.
Is it OK if I let you know when I do?

tool4ever:
yea sure, I'll probably see it anyway
like I said I know more tests can be valuable :)

allentiak:
@tool4ever

As promised, I've narrowed this PR to just the test harness plus tests for pure-logic classes. I will continue the work on the other classes on a separate PR.

I went back through every test, and only kept those that exercise hand-written logic.

That being said, there are a few borderline tests. For now, I've kept those - adding comments explaining why I think they should be kept. I have no problem dropping any (of all) of them if you still find them redundant.

tool4ever:
Current heat wave means I'm less productive and the remaining time was spent on more important bugs so far.

allentiak:
Sure, no worries.

tool4ever (commented):
Ok, I'm not seeing any real showstoppers...
but tbh I'm still somewhat skeptical about the overhead of these (at least to me) mostly trivial tests 🤔

consider this a "neutral" OK - would prefer another Dev consider if there's more value

allentiak:
@tool4ever 
Thanks for the review.

These are *characterization* tests.
https://en.wikipedia.org/wiki/Characterization_test

These testes *have* to be thorough, since they exist for the sole purpose of protecting the current behaviour of the code.

Why adding these tests? Because it's the only way to enable an eventual *safe* refactoring of the production code - which, as I have stated since the start, is my ultimate goal. (And, even if that refactoring never took place, they would still protect the code's current behaviour - which is good by itself.)

allentiak:
I just updated the PR via GH's web interface to the latest changes in `master`.
No other changes were done.

Jetz72:
> Why adding these tests? Because it's the only way to enable an eventual _safe_ refactoring of the production code - which, as I have stated since the start, is my ultimate goal. (And, even if that refactoring never took place, they would still protect the code's current behaviour - which is good by itself.)

What parts of the code were you looking to refactor that touch on the same areas as these tests? I've been dabbling with test suites as well, specifically one that covers in-game rules and interactions. In that case, the interactions between abilities touch on several huge interconnected systems. Refactoring any one of those systems risks breaking stuff subtle enough that it doesn't get caught until a bug report comes in a few days after release. 

What kind of change are you expecting might break the interaction of "turn on dev mode -> dev mode is now turned on"? Or "a NullPointerException is thrown if you pass null as a parameter to a constructor whose first statement is to call `requireNonNull` on that same parameter"? The only reason I can imagine that behavior might change is if someone deliberately changed it.

allentiak:
@Jetz72 Thank you for your questions.
Right now I'm too swamped to write a short answer.
I will get back to you during next week.

allentiak:
Hi again, @Jetz72,

First of all, thank you for your patience.
I knew it would take me some time to write a comprehensive yet not-so-long answer to your 2 questions.
My answer is below.


> Q1 of 2: What parts of the code were you looking to refactor that touch on the same areas as these tests?

Currently, there is quite a lot of GUI logic intertwined with game logic all over this sub-project. My idea is to, eventually, decouple the GUI logic from the backend logic. This is just the first step into that direction. Even though the changes will take time to be implemented, the fact that the current behaviour will be preserved by these tests is a good thing by itself - it prevents accidental alterations to it.

### Illustrative example 

The single biggest and most entangled bridge class in the subproject is **`forge.player.PlayerControllerHuman`** (`forge-gui/src/main/java/forge/player/PlayerControllerHuman.java`, ~3,960 lines). 

This class sits structurally *on* the seam:

- It **extends `PlayerController`** — the backend/game-engine contract from `forge-game`. So the engine calls it as pure game logic.
- Its method bodies **reach into `IGuiGame` via `getGui()`** to drive the UI — **169 `getGui()` call sites** in this one file.

So every rules decision the engine delegates to a human is implemented here by (a) computing game state and (b) prompting the UI, in the same method. That's the mix I'm trying to segregate.

### Concrete code example

This is just one among many.

**`assignCombatDamage(...)` (line 302)** — backend rules logic and UI prompting interleaved in one method:
```java
// backend rules: trample / multi-blocker / planeswalker checks
if ((attacker.hasKeyword(Keyword.TRAMPLE) && defender != null) || (blockers.size() > 1) ...) {
    ...
    // UI call, mid-method:
    final Map<CardView, Integer> result = getGui().assignCombatDamage(
            vAttacker, gameCacheBlockers.getTrackableKeys(), damageDealt, vDefender, overrideOrder, maySkip);
    ...
    // backend: translate CardView results back into engine Card map
    map.put(gameCacheBlockers.get(e.getKey()), e.getValue());
}
```

### Other clear examples in the same file

- **`sideboard(...)` (line 254)** — validates a `Deck` (backend) then calls `getGui().sideboard(...)` / `getGui().showErrorDialog(...)`.
- **`getAbilityToPlay(...)` (line 213)** — reasons over `SpellAbility` list then `getGui().getAbilityToPlay(CardView.get(hostCard), ...)`.
- **`divideShield` (339)**, **`specifyManaCombo` (359)**, **`announceRequirements` (389)** — each convert `Card`/`GameEntity`/`SpellAbility` to `*View` objects and immediately call `getGui().assignGenericAmount(...)`.

The `CardView.get(card)` / `GameEntityView.get(...)` conversions scattered through these methods are the actual seam artifacts: game-domain objects being marshalled into view objects for the UI, inline.


> Q2 of 2: What kind of change are you expecting might break the interaction of "turn on dev mode -> dev mode is now turned on"?

As you can imagine, the level of structural changes I described above cannot be implemented recklessly. Since they could affect the current functionality of the code - something we must avoid at all costs, we need a protection mechanism for the current behaviour.

That's where these tests come into play. As I said [above](https://github.com/Card-Forge/forge/pull/10827#issuecomment-4860392770), these are *characterization* tests.
In this way they act as guardrails, protecting all the current behaviour of the code. (_More details in the Wikipedia link [above](https://github.com/Card-Forge/forge/pull/10827#issuecomment-4860392770).)_ This is what will make sure that any later refactoring preserves the current, working behaviour. And, of course "all current behaviour" also includes this part of it, no matter how silly it may seem.

Of course, these few tests are only a part of what's needed. I simply decided to start small.

@Jetz72 
Thanks again for your questions. I hope I have answered them to your satisfaction?

Jetz72:
The overreach of `PlayerController`s into other systems is an area of interest to me as well, and I agree it'll take a very steady hand and careful planning to move the logic elsewhere.

I'm familiar with characterization tests, but I've generally known them to span larger or more complicated functions. If the scope of the tests can be as narrow as the ones demonstrated here, and your intent to ensure consistent behavior is so absolute that ones such as these would be included anyway, then how large will the complete test suite be? I think if you don't compromise on either of those things, you're looking at hundreds or thousands of tests for any one of Forge's larger systems. 

If we put those in the project in the lead up to this refactoring you're looking to try and do, we'd have to run them all for every single PR that gets submitted until then. If anyone is working in the same area and they actually do want to change the behavior, they'll likely end up causing one of these tests to fail. If they weren't aware that the behavior these tests assert is just arbitrary and based on what the project did at the time you implemented them, they'll end up spending time trying to figure out "wait, why is it so important that it work *that* way specifically?" 

Have you considered maintaining a separate branch on your fork for a test suite without merging it into the main project? That way it can be as elaborate as you want, you can still use it to safely experiment with refactoring, and then changes could be cherry-picked over to Forge as appropriate?

allentiak:
@Jetz72
Thanks for your follow-up questions. I'm happy we're finally starting to move past the "what's-the-point-of-this?" stage.
I will reply to your new questions during next week.

allentiak:
@Jetz72
Something came up, so I'm afraid it will take me longer to answer this.
Hopefully, I should have time by the end of this week.
Thanks for your patience.

In the meantime, I saw that #10830 was merged.
That's great news!

allentiak:
@Jetz72 
Update: I'm afraid I won't be able to get back to retake this topic before the beginning of August 2026.
Thank you again for your understanding and patience on this.

Jetz72:
Awaiting a response on a time-insensitive PR like this one isn't a matter of patience. You can just respond when you have an answer and we'll get the notification. If it takes long enough that it's at risk of being closed for inactivity, the bot will post a warning first.

PR #10830, opened 2026-06-01
add IPreferences Interface

Hanmac:
Preferences Interface for #10827

@MostCromulent i'm not 100% sure if the YieldController changes work that way
or if i need to change the PlayerControllerHuman to IGameController
because `NetGameController` calls `new YieldController(null);`

but there are still some methods unimplemented for IGameController

(i can split up the YieldController changes into a different PR)

### DIFF
--- a/forge-gui/src/main/java/forge/localinstance/properties/AbstractPreferences.java
+++ b/forge-gui/src/main/java/forge/localinstance/properties/AbstractPreferences.java
@@ -35,11 +35,11 @@ import forge.util.TextUtil;
  * Loads preferred values when instantiated.
  * If a requested value is not present, default is returned.
  */
-public abstract class PreferencesStore<T extends Enum<T> & PreferencesStore.IPref> {
+public abstract class AbstractPreferences<T extends Enum<T> & IPreferences.IPref> implements IPreferences<T> {
     private final Map<T, String> preferenceValues;
     private final String filename;
 
-    public PreferencesStore(final String filename0, final Class<T> clasz) {
+    public AbstractPreferences(final String filename0, final Class<T> clasz) {
         preferenceValues = new EnumMap<>(clasz);
         filename = filename0;
 
@@ -68,8 +68,8 @@ public abstract class PreferencesStore<T extends Enum<T> & PreferencesStore.IPre
 
     protected abstract T[] getEnumValues();
     protected abstract T valueOf(String name);
-    protected abstract String getPrefDefault(T key);
 
+    @Override
     public void save() {
         BufferedWriter writer = null;
         try {
@@ -91,18 +91,17 @@ public abstract class PreferencesStore<T extends Enum<T> & PreferencesStore.IPre
         }
     }
 
+    @Override
     public final void reset() {
         this.preferenceValues.clear();
     }
 
+    @Override
     public final void setPref(final T q0, final String s0) {
         preferenceValues.put(q0, s0);
     }
 
-    public final void setPref(final T q0, final boolean val) {
-        setPref(q0, String.valueOf(val));
-    }
-
+    @Override
     public final String getPref(final T fp0) {
         String val;
 
@@ -114,22 +113,6 @@ public abstract class PreferencesStore<T extends Enum<T> & PreferencesStore.IPre
     
     public final void togglePrefBoolean(final T q0) {
         setPref(q0, !getPrefBoolean(q0));
-    }
-
-    public final int getPrefInt(final T fp0) {
-        try {
-            return Integer.parseInt(getPref(fp0));
-        } catch (NumberFormatException e) {
-            return Integer.parseInt(getPrefDefault(fp0));
-        }
-    }
-
-    public final boolean getPrefBoolean(final T fp0) {
-        return Boolean.parseBoolean(getPref(fp0));
-    }
-    
-    public final double getPrefDouble(final T fp0) {
-        return Double.parseDouble(getPref(fp0));        
     }
 
     public void setGameType(final T q0, final Set<GameType> gameTypes) {
@@ -178,9 +161,5 @@ public abstract class PreferencesStore<T extends Enum<T> & PreferencesStore.IPre
             result.add(GameType.Archenemy);
         else if (gameType.equals("Archenemy Rumble"))
             result.add(GameType.ArchenemyRumble);
-    }
-
-    public interface IPref  {
-        String getDefault(); // Common method for getting the default value
     }
 }
--- a/forge-gui/src/main/java/forge/localinstance/properties/AbstractPreferences.java
+++ b/forge-gui/src/main/java/forge/localinstance/properties/AbstractPreferences.java
@@ -110,10 +110,6 @@ public abstract class AbstractPreferences<T extends Enum<T> & IPreferences.IPref
 
         return val;
     }
-    
-    public final void togglePrefBoolean(final T q0) {
-        setPref(q0, !getPrefBoolean(q0));
-    }
 
     public void setGameType(final T q0, final Set<GameType> gameTypes) {
         String s0 = "";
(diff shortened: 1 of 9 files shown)


MostCromulent:
Claude says it will break multiplayer yields in this form:

> There are two separate things here — the behavior change the refactor introduces, and the `PlayerControllerHuman → IGameController` question Hanmac raised. Splitting them out:
> 
> **1. The regression (independent of the question)**
> 
> Converting `toggleAutoPassNoActions()` / `toggleAutoPassOrStopAll()` from static (taking `IGameController`) to instance methods routed through `owner` changes behavior for network clients.
> 
> The originals took an `IGameController` because the toggle is a controller-level dispatch, and `ctrl.setYieldPref(...)` is polymorphic:
> - `PlayerControllerHuman.setYieldPref` → handles it locally
> - `NetGameController.setYieldPref` → sends it to the host over the wire (`SetYieldPref` envelope)
> 
> On a network client, `getGameController()` returns the `NetGameController`, whose YieldController is built as `new YieldController(null)` — `owner == null` is the sentinel for the client's controller (no local PCH). So the instance version hits `if (owner == null) return false;` and returns early: the local pref isn't flipped and the `SetYieldPref` wire message isn't sent. The runtime toggle (`P` / menu mid-game) becomes a no-op on clients, and the menu checkbox doesn't update. The host-side proxy that decides auto-pass for that player reads `YIELD_AUTO_PASS_NO_ACTIONS` from its `prefOverrides`, populated only by inbound `SetYieldPref`/seed envelopes, so without the send the host keeps its own value. (Initial state still propagates via `seedYieldStateOnHost` at game start — it's specifically the runtime toggle that's affected.)
> 
> This is true regardless of how the `owner` type question is resolved — it comes from routing through `owner` (null on clients) instead of the controller.
> 
> **2. The question: should `owner` become `IGameController`?**
> 
> That change alone wouldn't fix the above, and it has a cost. Beyond the toggle, `YieldController` uses several PCH-only methods — `getGui`, `getLocalPlayerView`, `getGame`, `getLobbyPlayer`, `isRemoteClient` — none of which are on `IGameController`. Promoting `owner` to `IGameController` would mean adding all of those to the interface, which `NetGameController` (no local game state) can't meaningfully implement.
> 
> **Options**
> 
> 1. Keep the two `toggle*` methods static taking `IGameController` (revert just that slice). They don't need the instance `prefs` field — a UI toggle targets the local user's FModel pref, which is what they currently do. `owner` stays `PlayerControllerHuman`.
> 2. Keep the instance style but give `YieldController` a separate `IGameController` dispatch reference (distinct from `owner`), with `NetGameController` passing `this`. The toggle routes through that reference; `owner` stays `PlayerControllerHuman` for the game-state methods.

Hanmac:
reverted the YieldController for now

forge-gui/src/main/java/forge/localinstance/properties/AbstractPreferences.java
```diff
@@ -112,22 +111,6 @@ public final String getPref(final T fp0) {
         return val;
     }
```
allentiak, line 113: Just curious...
Why did you remove the setters and getters from the abstract class?

allentiak: Or is it that you not sure whether to use an abstract class or an interface?

Hanmac: They are done as "default" in the Interface,
so the Abstract one doesn't need to implement them

allentiak: I hadn't read the end part of the line! :sweat_smile:
I see now that the new abstract class implements the interface.

That being said... Why an extra indirection via the abstract class at all?
Can't you just use the interface and make other classes implement it?

allentiak: I mean - having both an interface and an abstract class with the same function is redundant in this case, right?

Hanmac: Its only about the type specific ones
getPrefBoolean -> getPref
setPref(boolean) -> setPref(string)

the other methods of the interface still need to be implemented

allentiak: I still don't see the reason why an abstract class is needed in this case when the interface can have any required methods with default implementations already in it...
Or am I missing something else here?

Hanmac: The interface just says that the methods exist
it doesn't handle how the values are stored
only the Abstract class says it uses EnumMap
and how it loads and saves the properties to the files

allentiak: Thanks for the explanation.

Whereas it clearly explains which responsibilities of the "Preferences" abstraction each construct (abstract class + interface) has, it does not reply my main question.

This is: **What is(are) the underlying reason(s) for which one single abstraction ("Preferences") has its responsibilities split into two separate, <ins>independently-inheritable</ins> constructs (an interface + a class)?**

I ask this because, unless there is either a logical or technical reason for this, I think that's only a disaster waiting to happen.
Why? Because, unless it is somehow technically possible to restrict the inheritance to one of the two constructs, there may be a case where another class or interface inherits from the wrong construct - something completely illogical.

Or is there a technical restriction for this division that I am not seeing here?

allentiak: I mean, I think it may make sense to create an abstraction for the "Preferences" behaviour.
I'm just wondering if this is the right approach to it :-)

Hanmac: The Interface exists so you can Mock it in tests

(or what I tried with YieldController, have it decorate the Options)

allentiak: Thanks. That answers _part_ of the question.

Then, again, why not just move everything into the interface, so it can be used for both the production and test code, and ditch the abstract class as a redundant construct?
(The "disaster-waiting-to-happen-due-to-possible-dual-inheritance-path" part of my previous answer still applies.)

Hanmac: The Interface only does half of the functions

Only the Abstract Class handles the real load/save

allentiak:
You have already said that.

I'm asking why splitting the functionality across the abstract class and
the interface. I still don't see any reason for that, and none of your
previous answers explain that single, crucial point.
Would you be so kind as to please explain why? (Please refer to my earlier
long answer as for why I'm asking this.)

Hanmac:
> You have already said that. I'm asking why splitting the functionality across the abstract class and the interface. I still don't see any reason for that, and none of your previous answers explain that single, crucial point. Would you be so kind as to please explain why? (Please refer to my earlier long answer as for why I'm asking this.)

Because the interface can't know how the Data is stored
(If at all)

For example, see Java HashMap

```java
public class HashMap<K,V>
extends AbstractMap<K,V>
implements Map<K,V>
```

The interface has the methods list
The AbstractMap most of the implementation
And HashMap the final part to make it work

allentiak:
Thanks for the example!
It was instrumental.
I asked a bot about it, and I think that I understand now.
It's due to a technical limitation of Java.

Lately, I had been working with Rust, which does away with class
inheritance at all - so I was not aware of this.

Now I see why it makes sense.
Yeah, go ahead.
I'll approve this PR as-is.

allentiak:
I see that this PR is still a draft.
Ping me as son at you mark it as ready for review, and I'll approve it.

allentiak:
I cannot approve it myself since I don't have write access to the repo.

Hanmac:
@tool4ever @tehdiplomat your review to this interface?

Jetz72 (approved):

PR #10833, opened 2026-06-02
Card cleanup: 2026-06-02

dracontes:
Dealing with capitalization issues in the context of `non[Type]` syntax.


Hanmac:
@tool4ever should `![Type]` be better? 🤔

tool4ever (approved):
imo this might be the exception to support "non" just so scripting it follows closer to the actual card text
and I'm not sure text changes would handle it otherwise

Hanmac:
> imo this might be the exception to support "non" just so scripting it follows closer to the actual card text and I'm not sure text changes would handle it otherwise

It looks for word boundaries, so `!Human` should be replaced by text changes

PR #10834, opened 2026-06-02
Fix ability-menu crash and positioning for floating-zone cards

MostCromulent:
## Bug report

Crash when casting a modal dual-faced card (e.g. Birgi) from the "Playable Zone Cards" window via a Wish, with the Sideboard window closed: `IllegalComponentStateException: component must be showing on the screen to determine its location`.

## Root cause

Cards in the "Playable Zone Cards" window report their real zone (Sideboard), so the face-selection popup was anchored to the card's home-zone panel. If that window was never opened the panel doesn't exist and the code already falls back safely; but once it had been opened then closed, the panel still existed yet was off-screen, so `JPopupMenu.show()` threw.

## Fix

- **Guard the anchor** — `getAbilityToPlay` now treats a non-showing panel like a missing one: the fallback condition becomes `panel == null || !panel.isShowing()`, so a hidden home-zone panel routes to the prompt area instead of into `JPopupMenu.show()`. This alone stops the crash for both the opened-then-closed floating window and docked-then-closed tab cases.
- **Anchor to the clicked panel** — `CardPanelContainer.mouseReleased` (the single dispatch point for every card area) records the clicked panel on `CMatchUI`, and `getAbilityToPlay` prefers it (when it's for the same card and on screen) over the zone-based `findCardPanel` lookup. The menu now opens in the window the card was actually clicked from, fixing the case where a card shown in multiple windows (e.g. both the Playable Zone Cards and Sideboard windows) opened its menu over the wrong one.

---

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
+++ b/forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
@@ -161,6 +161,7 @@ public final class CMatchUI
     private boolean allHands;
     private boolean showOverlay = true;
     private JPopupMenu openAbilityMenu;
+    private CardPanel lastClickedCardPanel;
 
     private IVDoc<? extends ICDoc> selectedDocBeforeCombat;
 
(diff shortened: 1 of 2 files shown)


tool4ever (approved):
thanks, good double fix

PR #10873, opened 2026-06-04
Prepared: Show linked exile cards as ghosts on their permanent

MostCromulent:
Screenshot 2026-06-04 211118

Screenshot 2026-06-04 210948

## Summary

A prepared spell's castable copy — and cards exiled "until this permanent leaves" (Banishing Light and similar) — live in the Exile zone, so players often don't realise the permanent is holding a card or that a prepared spell can be cast. This is a recurring source of confusion.

This adds a GUI-only convenience: a permanent that holds such a card shows a faded blue-grey "ghost" of it attached to the permanent. It's purely cosmetic — the real cards stay in Exile and the engine's zones are unchanged. Implemented on both the Swing and libgdx UIs.

## Key features
- **Gated behind a preference** (`UI_SHOW_LINKED_EXILE_CARDS`, default on), toggleable from the desktop Display menu, the desktop Preferences screen, and the mobile settings page.
- **Click a prepared-spell ghost to cast it** from its permanent; ghosts for "exile until ~ leaves" cards are informational only (hover/tap to inspect).
- **Shown only when a permanent holds exactly one linked card** — a deliberate choice so mass "exile until ~ leaves" effects (e.g. Temporary Lockdown) don't flood a permanent with a pile of ghosts. Those cards remain visible in the Exile zone.
- The link reuses existing view data (`UntilLeavesBattlefield`) plus one new `CardView` trackable property for prepared spells; no game-rules changes.

---


Hanmac:
@MostCromulent should probably add a text flag there for isPrepared?
https://github.com/Card-Forge/forge/blob/45e2f41ff6f59c5c252e937f9f6f5998245a175b/forge-game/src/main/java/forge/game/card/Card.java#L2887-L2910

hm i might move these part from Card::getAbilityText into a different function
maybe in CardView? https://github.com/Card-Forge/forge/blob/45e2f41ff6f59c5c252e937f9f6f5998245a175b/forge-game/src/main/java/forge/game/card/CardView.java#L907-L912

tool4ever (dismissed):

forge-gui-desktop/src/main/java/forge/view/arcane/PlayArea.java
```diff
@@ -1098,6 +1116,69 @@ private boolean doUpdateCard(final CardView card, boolean fromRefresh) {
...
+
+    // The one card this permanent holds in exile, or null if it holds none or more than one
+    private static CardView singleLinkedExileCard(final CardView card) {
+        final CardView prepared = card.getPreparedSpell();
+        final FCollectionView<CardView> untilLeaves = card.getUntilLeavesBattlefield();
+        if (prepared != null) {
+            return untilLeaves.isEmpty() ? prepared : null;
+        }
+        return untilLeaves.size() == 1 ? untilLeaves.iterator().next() : null;
+    }
```
Jetz72: I think you could reasonably allow up to the maximum stack depth instead of just one.

Jetz72 (commented):
Why does the mobile implementation use an enum to distinguish the types of ghost cards while the desktop one does not?

MostCromulent:
> Why does the mobile implementation use an enum to distinguish the types of ghost cards while the desktop one does not?

@Jetz72, Claude's explanation:

> The two platforms only carry the state they actually consult.
> 
> There are two kinds of ghost: an **actionable** one (a prepared spell — you can cast it) and an **informational** one (a card exiled until the permanent leaves, e.g. Banishing Light — there's nothing to do with it).
> 
> On desktop nothing reads that kind. A click on any ghost routes uniformly to `selectCard` and the engine decides what happens — a prepared spell casts, an informational card is a no-op — and the tint is identical for both. So a single `boolean ghost` is all desktop needs.
> 
> Mobile does need the kind, because of its zoom view. On mobile you open a card's zoom and can swipe up to "activate" it; for a prepared-spell ghost that swipe-up should cast, for an informational ghost it should do nothing. The catch is that the swipe-up doesn't act on the ghost panel directly — it looks the card up in the shared per-card `CardAreaPanel` cache, which is also what the Exile zone uses to display that card. Letting an informational card go through that path could poke the Exile-zone copy and misfire, so we suppress activate for the informational ghosts — and that guard has to know which kind a ghost is. Desktop has no swipe-up-from-zoom gesture, so it never hits this.
> 
> Unifying doesn't really pay off either way: adding the enum to desktop would be write-only state it never reads, and collapsing mobile to a boolean would need a second `isInfo` flag anyway — so the enum is the cleaner representation on the side that actually distinguishes the kinds.

MostCromulent:
> @MostCromulent should probably add a text flag there for isPrepared?

@Hanmac done.

forge-gui-mobile/src/forge/screens/match/views/VCardDisplayArea.java
```diff
@@ -339,6 +370,56 @@ public void updateCard(final CardView card) {
...
+                    break;
+                }
+                attachedPanels.add(newGhost(exiled, GhostKind.INFO));
+                cap--;
+            }
+        }
+
+        private static int maxStackDepth() {
+            try {
+                return Integer.parseInt(FModel.getPreferences().getPref(FPref.UI_MAX_STACK_DEPTH));
```
tool4ever: AI needs to remember `getPrefInt` should be used instead

tool4ever (approved):

MorihiroS:
@MostCromulent I tested it, this looks awesome, thank you!

PR #10897, opened 2026-06-05
Fix crash on game end from unfillable layout gap

MostCromulent:
Discord user reported `UnsupportedOperationException: Gap was not filled.` thrown from `SRearrangingUtil.fillGap()` via `FloatingZone.closeAll()` -> `CMatchUI.finishGame()` at the end of a game (triggered by docking AI commanders'' command zones). Because `closeAll()` is the first thing `finishGame()` does, the throw aborted it before the win/lose screen was shown, leaving the match UI stuck.

`fillGap()` tears down docked zone cells one at a time, and the intermediate layout can leave a gap the grow-a-neighbor algorithm can''t cover. Gap-filling is purely cosmetic and the caller removes the cell regardless, so this drops the fatal `throw` and lets `fillGap()` fall through instead.

---

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/gui/framework/SRearrangingUtil.java
+++ b/forge-gui-desktop/src/main/java/forge/gui/framework/SRearrangingUtil.java
@@ -317,7 +317,7 @@ public final class SRearrangingUtil {
         // Falls back to original priority order if none qualify.
         if (tryFillGap(true)) { return; }
         if (tryFillGap(false)) { return; }
-        throw new UnsupportedOperationException("Gap was not filled.");
+        // An unfillable gap is cosmetic and the caller removes the cell anyway, so fall through rather than throw
     }
 
     private static boolean containsField(final List<DragCell> cells) {


jamincollins:
## Findings

#### 1. fillGap now silently ignores unfillable gaps

Description:
The change removes the `throw new UnsupportedOperationException("Gap was not filled.")` from `SRearrangingUtil.fillGap()`. The commit message explains that at game end, FloatingZone.closeAll() tears down docked zone cells one at a time and intermediate layout can leave a gap the grow-a-neighbor algorithm can't cover. The throw propagated out of CMatchUI.finishGame() before win/lose screen was shown.

The algorithm in `tryFillGap(boolean preferField)` attempts to expand neighboring cells left/right/up/down in four passes (left expand right, right expand left, top expand down, bottom expand up). If no matching pair found, it returns false. Previously `fillGap()` threw; now it falls through.

Reference: forge-gui-desktop/src/main/java/forge/gui/framework/SRearrangingUtil.java:315-321

Test case:
```java
// Simulate a layout where source cell is isolated with no matching left/right/top/bottom pair
DragCell src = new DragCell();
src.setBounds(100,100,200,200);
FView.SINGLETON_INSTANCE.getDragCells().clear();
// No neighboring cells share edges with src
SRearrangingUtil.fillGap(src);
// Should not throw UnsupportedOperationException
```
Impact: Prevents crash on game end; gap remains cosmetic. Acceptable because caller removes cell regardless.

#### 2. cellSrc static field can be stale if fillGap called concurrently

Description:
`SRearrangingUtil` uses static fields `cellSrc`, `cellTarget`, `cellNew`, `srcSelectedDoc`, `dropzone`, etc. `fillGap(DragCell sourceCell0)` sets static `cellSrc = sourceCell0` then calls private `fillGap()`. If two UI threads trigger rearrangement simultaneously (e.g., during game end cleanup and user resize), static state may be overwritten between set and use.

Reference: SRearrangingUtil.java:469-472

Test case:
```java
// Thread A
new Thread(() -> SRearrangingUtil.fillGap(cellA)).start();
// Thread B
new Thread(() -> SRearrangingUtil.fillGap(cellB)).start();
// cellSrc may be set to cellB before fillGap() reads it in Thread A
```
Impact: Potential race condition leading to resizing wrong cell. Low probability but not impossible during game end where closeAll() iterates docked zones.

#### 3. cellsToResize.contains check uses reference equality

Description:
In `tryFillGap`, when collecting cells to resize, code does `if (!cellsToResize.contains(cell)) { cellsToResize.add(cell); }`. `DragCell` does not override `equals`, so contains uses reference equality. If same logical cell appears via different reference (e.g., from different list), duplicate may be added, leading to double `setBounds` calls. Not a crash but may cause inconsistent bounds.

Reference: SRearrangingUtil.java:345, 424, 444

Test case:
```java
DragCell a = new DragCell();
DragCell b = a; // same ref
List<DragCell> list = new ArrayList<>();
list.add(a);
list.contains(b); // true
// Different instance with same bounds would not be detected
```
Impact: Minor; unlikely to cause crash.

#### 4. Unfillable gap leaves empty space in layout

Description:
By silently falling through, layout may retain an empty rectangular gap where a removed cell existed. The cell is removed from `FView` via `removeDragCell`, but neighboring cells are not resized. This may cause visual artifact: blank area in UI persists until next layout save/restore.

Reference: FloatingZone.closeAll() lines 300-301, 371-372, 396

Test case:
```java
// Create layout with three cells in row: A | B | C
// Remove B where left/right neighbors don't share top/bottom corners (e.g., different heights)
// fillGap(B) returns false, B removed, gap remains between A and C
```
Impact: Backwards compatible (no crash). Cosmetic degradation acceptable per PR description.

#### 5. No logging for unfillable gap

Description:
Previously exception would surface in logs; now silent fall-through provides no diagnostic. If layout corruption grows, debugging becomes harder.

Reference: SRearrangingUtil.java:320

Test case:
```java
// Trigger unfillable gap
SRearrangingUtil.fillGap(isolatedCell);
// No log entry, no metric
```
Impact: Observability reduced.

#### 6. fillGap called from multiple contexts with different expectations

Description:
`fillGap` is called from:
- VMatchUI.populate() line 264 (empty cell removal)
- VMatchUI cleanup line 112
- FloatingZone.closeAll() line 300
- FloatingZone.undockZone() line 371
- FloatingZone.closeWindow() line 396
- ACEditorBase line 371

All callers assume `fillGap` either succeeds or throws. Now it silently does nothing. For editor context, silent failure may hide layout bugs.

Reference: search results

Test case:
```java
// In deck editor, remove a panel with no neighbor
SRearrangingUtil.fillGap(cell);
// Editor layout leaves gap, user sees blank space
```
Impact: Consistent behavior across contexts, but may mask bugs.

#### 7. tryFillGap prefers FIELD cells but may skip valid resize

Description:
`fillGap()` first tries `tryFillGap(true)` preferring FIELD cells, then `tryFillGap(false)`. If first pass finds a matching pair but cellsToResize does not contain a FIELD doc, `containsField` returns false and the pass returns false, falling through to second pass. This is intentional per commit. However, if FIELD preference causes first pass to reject a valid resize that would fill gap, second pass may still succeed. Good.

Reference: SRearrangingUtil.java:318-319

Test case:
```java
// Gap between two non-field cells, but a FIELD cell is also adjacent
// preferField true will reject resize of non-field pair, second pass succeeds
```
Impact: Correct behavior.

### Performance and security
#### 8. O(n) scan of all DragCells per fillGap call

Description:
`tryFillGap` iterates over `FView.SINGLETON_INSTANCE.getDragCells()` four times per call, each scanning entire list. During game end, `closeAll()` iterates over dockedZones and calls `fillGap` per removed cell. For N zones, complexity ~O(N²). Acceptable for typical game (≤8 players, few zones).

Reference: SRearrangingUtil.java:337-348, 363-...

Test case:
```java
// 100 docked zones, each fillGap scans 200 cells -> 20k iterations
```
Impact: Negligible.

#### 9. No security impact

Description:
Layout manipulation is client-side UI only. No network or file system exposure.

## Verdict
Request Changes

The PR prevents a crash at game end by suppressing UnsupportedOperationException when a layout gap cannot be filled. While the crash fix is warranted, the change silently swallows layout errors without logging, and relies on static mutable state which is susceptible to race conditions during concurrent UI updates. The code should at minimum log a warning when an unfillable gap is encountered, and consider making fillGap stateless to avoid static field hazards.

## Recommended Fixes
1. Add warning log when gap cannot be filled: `System.out.println("Warning: unfillable gap for cell " + cellSrc + ", layout may contain empty space")` or use Forge logger.
2. Make fillGap instance method or pass cellSrc as parameter through call chain to avoid static mutable state.
3. Consider returning boolean from fillGap(DragCell) to allow callers to decide on fallback behavior.
4. Add unit test for isolated cell scenario to ensure no exception thrown and cell removed cleanly.

---

MostCromulent:
Each suggestion checked against the branch. References are to `forge-gui-desktop/src/main/java/forge/`.

|---|---|---|
| 1 | Warn when a gap cannot be filled | The unfillable gap is a routine step in teardown rather than an error state — `closeAll()` removes docked zone cells one at a time, and several command-zone cells appear whenever AI commanders' zones are docked, so intermediate layouts regularly leave a gap the grow-a-neighbor pass cannot cover. It is also transient: `closeAll()` never calls `SLayoutIO.saveLayout` (`view/arcane/FloatingZone.java:288-309`) and runs only from `finishGame()`, as the match UI gives way to the win/lose screen, and from `initialize()`, before a new match populates the layout from file. Nothing persists and nothing renders, so a warning would fire on ordinary game ends with no matching symptom. `gui/framework` has no logger either; its only `System.out` is commented out (`SLayoutIO.java:473`). |
| 2 | Make `fillGap` stateless to avoid static fields | The static fields predate this PR and are untouched by a one-line diff. All seven call sites run on Swing UI paths. Threading the state through the call chain would rewrite code this change does not modify. |
| 3 | Return a boolean so callers can choose a fallback | No caller has a fallback to choose. Every call site removes the cell immediately afterward, which is what makes the unfilled gap harmless. |
| 4 | Add a unit test for the isolated-cell case | The scenario needs a populated `FView` singleton and live Swing layout. The desktop module verifies UI behaviour by running it rather than by unit test, and adding the first such harness for a one-line change is disproportionate. |

Worth noting for the record: `undockZone` and `removeDocked` also call `fillGap` and do save the layout afterwards (`FloatingZone.java:384` and `:408`). Those paths previously threw, so this change turns a crash into a cosmetic gap there as well.

No changes applied.

---

PR #10898, opened 2026-06-05
Make phase label buttons more button-y

MostCromulent:
Many users don't realise phase labels are buttons rather than baked-in UI elements, hence constant bug reports about phases not being activated.

This PR makes a few subtle changes to make the design less flat and try and better signal their button-y nature. 

- The default cursor changes to a hand pointer when hovered over a phase label
- Adds a light bevel fill, gradient, and highlight edge to signal enabled/disabled button state
- Hovering over a button previews the changed state

Skin safe and makes no changes to color scheme.

This isn't the whole solution but hopefully it helps a bit. 

 | Before | After | 
  |:------:|:-----:|
  | Before | After |

---


tool4ever (approved):
LGTM

PR #10900, opened 2026-06-06
Keep focus on prompt during opponent-hand reveal

MostCromulent:
Follow-up to #10763.

#10763 stopped the floating-zone search filter from stealing focus by gating its focus grab on the prompt's OK/Cancel buttons being enabled. 

That covered normal play, but opponent-initiated hand reveal still grabbed focus: it shows the floating zone and the OK/EndTurn confirm as two separate EDT tasks, with the zone shown first — so when the zone checked whether a prompt button was actionable, the OK button hadn't been enabled yet, and focus went to the filter.

This establishes the confirm prompt before showing the zone, so the OK button is already enabled when the zone decides where focus goes. Focus stays on the prompt; the filter stays clickable.

### EDIT: Driveby - dim unrevealed cards

Tacked a small visual tweak onto this PR since it lives in the same reveal flow: when an opponent reveals a card, the floating zone shows their whole hand, rendering blank card backs for the cards that remain unrevealed. 

Those backs are now dimmed — reusing the same darkening already applied to unselectable cards during selection — so the revealed card(s) stand out instead of getting lost among identical backs. The dimming only kicks in when at least one card in the zone is viewable, so an all-hidden zone (e.g. peeking at an opponent's full hand) is left unchanged.

---


jamincollins:
## Findings

#### 1. InputConfirm set before zone show may deadlock if EDT posts input after show
Description: In PlayerControllerHuman.reveal() the InputConfirm is created, inputQueue.setInput(inp) is called, then FThreads.invokeInEdtNowOrLater shows the floating zone. The InputConfirm.showAndWait() was changed to inp.awaitLatchRelease() after setInput, so the latch is awaited on the calling thread while EDT may need that thread to process UI events that release the latch. If the InputConfirm's showMessage() triggers UI updates that require EDT, and EDT is blocked waiting for the caller to finish, a deadlock can arise.
Test case:
```java
// Simulate calling reveal from non-EDT thread
Thread t = new Thread(() -> controller.reveal(...));
t.start();
// If EDT posts runnable that requires input to be processed before latch release, both threads block
```
Impact: UI freeze on hand reveal.

#### 2. Dimmed flag set on CardPanel never cleared on refresh reuse
Description: FloatingZone.doRefresh() sets panel.setDimmed(anyViewable && !mayView). CardPanel instances are reused via getCardPanel(cardId). If a panel is reused for a different card in a later refresh, the dimmed flag from previous card persists until explicitly overwritten. The current code sets dimmed for each panel in the same loop, but if the panel list shrinks (filter reduces cards) and panels are not cleared, leftover dimmed state may remain on panels that are later removed.
Test case:
```java
// First refresh: card A not viewable -> dimmed=true
// Second refresh: same panel reused for card B viewable -> dimmed set to false correctly
// However if panel is removed from list but not disposed, dimmed stays true in memory
CardPanel p = zone.getCardPanel(id);
// reuse without explicit setDimmed
```
Impact: Visual inconsistency where cards appear dimmed incorrectly.

#### 3. anyViewable computed with short-circuit OR but applied after loop
Description: anyViewable |= mayView(card) is accumulated while building panels. The dimmed decision uses anyViewable which includes the current card's viewability. This is correct for dimming unviewable when at least one is viewable, but if all cards are unviewable, anyViewable remains false and dimmed stays false, which matches spec. No bug.

#### 4. Ghost overlay drawn after normal draw but without clipping
Description: drawGhostOverlay fills a rounded rect over the whole card. If the card is tapped and offset applied, the overlay uses cardXOffset/cardYOffset without offset correction. The normal dim overlay uses offset for tapped cards.
Test case:
```java
CardPanel panel = new CardPanel(...);
panel.setTapped(true);
panel.paint(g);
// ghost overlay may misalign by 1px
```
Impact: Visual glitch on tapped ghosts.

#### 5. getMatchUI().mayView(panel.getCard()) may NPE if panel.getCard() returns null
Description: FloatingZone.doRefresh() calls panel.setDimmed(anyViewable && !getMatchUI().mayView(panel.getCard())). CardPanel.setCard may be called with null? In CardPanel constructor, card may be null? If card is null, mayView will NPE inside mayView logic when calling canBeShownToAny.
Test case:
```java
CardPanel panel = new CardPanel(matchUI, null);
panel.setDimmed(anyViewable && !matchUI.mayView(panel.getCard())); // NPE
```
Impact: Crash on empty zone.

#### 6. InputConfirm created with message fm that may be null
Description: reveal() builds fm via MessageUtil.formatMessage(message, getLocalPlayerView(), owner). If message is blank, fallback to default, but formatMessage could return null if localizer missing key. Passing null to InputConfirm constructor leads to null message shown.
Test case:
```java
String fm = null;
InputConfirm inp = new InputConfirm(this, fm, "OK", "End", true);
```
Impact: Null pointer in showMessage.

#### 7. dimmed flag interacts with nonselectable dimming causing double darkening
Description: CardPanel paints dim overlay when nonselectable || dimmed. If card is both nonselectable and dimmed, overlay is painted once with alpha 0.6. If future code adds additive dim, intensity may be wrong. Not critical now.

#### 8. setDimmed changes visual appearance of cards that were previously selectable
Description: Changing CardPanel rendering adds dim for hidden cards. Existing UI tests may assume blank backs are fully visible. No API change, visual only.

#### 9. Ghost tint color constant hard-coded
Description: GHOST_TINT = new Color(90,120,175,110) is hard-coded. No user preference. Could clash with themes.

#### 10. InputConfirm setInput before zone show changes timing of prompt visibility
Description: Setting input before showing zone ensures isInputButtonEnabled returns true when FloatingZone.showWindow() checks. However, InputConfirm.showMessage() is not called until latch release? Actually InputConfirm extends InputSyncronizedBase which sets input on showAndWait. Now input is set earlier, so prompt buttons may be enabled before zone is visible, causing focus steal. Verified via code.

#### 11. FloatingZone.doRefresh called from setVisible which is on EDT, but mayView may be expensive
Description: mayView iterates over local players and checks canBeShownToAny, potentially costly for large hands. Called per card per refresh.

### Performance and security
#### 12. Rebuilding cardPanels list each refresh allocates new ArrayList each time
Description: doRefresh creates new ArrayList<CardPanel> each call. Acceptable for small zones, but for large library reveals could allocate heavily.

## Verdict
Request Changes
The PR fixes focus stealing on opponent hand reveal and adds dimming for hidden cards, but introduces potential deadlock risk with InputConfirm ordering, null safety gaps, and visual glitch on tapped ghost overlays. The changes need guardrails and tests.

## Recommended Fixes
1. Move inputQueue.setInput(inp) to happen after the EDT runnable has completed, or use a latch that is released on EDT after zone is visible to avoid deadlock.
2. Null-check panel.getCard() before calling mayView, and clear dimmed flag when panel is reused or disposed.
3. Align ghost overlay drawing with tapped offset used in normal dim overlay.
4. Add unit test for mayView null card and for anyViewable false case.

---

MostCromulent:
Each suggestion checked against the branch.

|---|---|---|
| 1 | Move `setInput` after the EDT runnable, or release the latch on the EDT once the zone is visible | This would reintroduce the bug the PR fixes. `showAndWait()` is defined as `getController().getInputQueue().setInput(this); awaitLatchRelease();` (`InputSyncronizedBase.java:37-40`), so the change splits one method into its own two calls and moves only their position relative to the EDT runnable. No thread changes role: the game thread blocks on the latch as before, and the EDT releases it on the button click. `awaitLatchRelease()` asserts it is not running on the EDT (line 22), so an EDT caller would already fail on master. Setting the input before the zone is shown is what keeps `FloatingZone.showWindow()` from taking focus off the prompt. |
| 2 | Null-check `panel.getCard()`, and clear `dimmed` when a panel is reused | Neither case arises. Every panel in `cardPanels` is either constructed as `new CardPanel(getMatchUI(), card)` or given `setCard(card)` in the loop directly above, with `card` drawn from `getCards()` (`FloatingZone.java:646-660`), so the dim loop never sees a null card. That loop also assigns `dimmed` on every panel in the list on each refresh, so a reused panel is always reassigned rather than inheriting the previous card's value. |
| 3 | Align the ghost overlay with the tapped offset used by the dim overlay | `drawGhostOverlay` is not part of this change. The `dimmed` flag joins the existing `nonselectable` condition on the same overlay rather than adding a second drawing path, so no new alignment relationship is created. |
| 4 | Add unit tests for a null card and for `anyViewable` false | The null card is unreachable as covered under 2, and `anyViewable` false yields no dimming, which is the pre-existing appearance. Exercising `doRefresh` also needs a live `CMatchUI`, a `FloatingZone` window and Swing panels; the desktop module verifies this kind of behaviour by running it. |

No changes applied.

---

Issue #10890, opened 2026-06-05
OrCost


Hanmac:
[Titania, Rugged Rumbler](https://scryfall.com/card/msh/235/titania-rugged-rumbler) has complex Ward cost that is A or B.

> Ward—Discard a card or pay {2}. (Whenever this creature becomes the target of a spell or ability an opponent controls, counter it unless that player discards a card or pays {2}.)

An **easy** way would be to make Ward support two different costs via generic Choice,
but `Ward:CostA:CostB` is already used as `Ward:Cost:Desc`, so that might not work.
See "[Minthara, Merciless Soul](https://scryfall.com/card/clb/286/minthara-merciless-soul)"
>`K:Ward:X:where X is the number of experience counters you have`

A **better** handle this would be that we need a Cost like `Or<Discard<>/2>` 
That would help us with other places too.

Hanmac:
@tool4ever we might need to rework Minthara, because her rules text has changed:

>Minthara has ward {X}, where X is the number of experience counters you have.

So it's not a Keyword anymore, but a Static Ability giving a Keyword?

If done, then we can change Ward it to allow `Ward:CostA:CostB`

Hanmac:
OrCost has a bigger problem with #309 that the complex Cost needs to be resolved before SetCost from Trinisphere applies in case part of it is ManaCost

PR #10917, opened 2026-06-07
MSH: extend Ward costs

Hanmac:
related to #10890

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
+++ b/forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
@@ -1954,41 +1954,35 @@ public class CardFactoryUtil {
                 t.setOverridingAbility(sa);
                 inst.addTrigger(t);
             }
-        } else if (keyword.startsWith("Ward") && inst instanceof KeywordWithCost withCost) {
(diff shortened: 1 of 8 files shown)


forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -1954,19 +1954,45 @@ public static void addTriggerAbility(final KeywordInterface inst, final Card car
...
+            } else {
+                boolean skipFirst = false;
+                List<AbilitySub> subs = Lists.newArrayList();
+                for (String costStr : keyword.split(":")) {
+                    if (!skipFirst) {
+                        skipFirst = true;
+                        continue;
+                    }
+
+                    final Cost cost = new Cost(costStr, false);
```
tool4ever: Just cast it to `Ward` and get the costs from there?

Hanmac: i need the unparsed CostString for the UnlessCost, so i used a `Map<String, Cost>` instead

tool4ever (approved):
guess we're stuck with GenericChoice for now

Hanmac:
> guess we're stuck with GenericChoice for now

for now, OrCost would be a bigger can of worms

PR #10925, opened 2026-06-08
Updated 'Copy Collection to Clipboard' menu item in the inventory view

mrstarbuck007:
Outputs the entire card collection, minus any cards in Auto-Sell, to the clipboard in Moxfield CSV format for easy importing.


forge-gui-mobile/src/forge/adventure/scene/AdventureDeckEditor.java
```diff
@@ -511,7 +511,11 @@ public void buildDeckMenu(FPopupMenu menu) {
             super.buildDeckMenu(menu);
             if (!(parentScreen instanceof AdventureDeckEditor adventureEditor) || adventureEditor.getAutoSellPage() == null)
                 return;
-            menu.addItem(new FMenuItem(Forge.getLocalizer().getMessage("btnCopyCollectionToClipboard"), Forge.hdbuttons ? FSkinImage.HDEXPORT : FSkinImage.BLANK, e1 -> FDeckViewer.copyCollectionToClipboard(AdventurePlayer.current().getCards())));
+
+            ItemPool<PaperCard> autoSellCards = AdventurePlayer.current().getAutoSellCards();
+            CardPool playedCards = AdventurePlayer.current().getCards().getFilteredPool(card -> !autoSellCards.contains(card));
+            menu.addItem(new FMenuItem(Forge.getLocalizer().getMessage("btnCopyCollectionToClipboard"), Forge.hdbuttons ? FSkinImage.HDEXPORT : FSkinImage.BLANK, e1 -> FDeckViewer.copyCollectionToClipboard(playedCards)));
+
```
Jetz72: Should build `playedCards` when clicking the button rather than when opening the menul

forge-gui-mobile/src/forge/deck/FDeckViewer.java
```diff
@@ -51,14 +55,44 @@ public static void copyDeckToClipboard(Deck deck) {
...
+            String cardName = card.getCardName();
+            String cardEdition = card.getEdition();
+            String accountedKey = cardName + '\t' + cardEdition;
+            if (!accountedMap.containsKey(accountedKey) && !card.isVeryBasicLand()) {
+                String regexCardName = regexQuote.matcher(cardName).replaceAll("\"\"");
+                regexCardName = regexA.matcher(regexCardName).replaceAll("a");
+                regexCardName = regexE.matcher(regexCardName).replaceAll("e");
+                regexCardName = regexI.matcher(regexCardName).replaceAll("i");
+                regexCardName = regexO.matcher(regexCardName).replaceAll("o");
+                regexCardName = regexU.matcher(regexCardName).replaceAll("u");
```
Jetz72: We do have `StringUtils::stripAccents`, but why wouldn't we want to export the oracle name?

Jetz72 (commented):
Seems like it'd be easier and more accurate to output the edition code and collector number rather than going through and grouping them by set.

forge-gui-mobile/src/forge/deck/FDeckViewer.java
mrstarbuck007: The Moxfield csv import can't handle accented characters so we need to replace them with the ASCII equivalents.

tool4ever:
How about we don't try optimizing generic features for site-specific quirks? that seems like a neverending mess

mrstarbuck007:
Because otherwise the feature is not very useful?

forge-gui-mobile/src/forge/deck/FDeckViewer.java
Jetz72: Strikes me as something that'd ideally be fixed on their end, rather than making our end less correct and less consistent in order to have compatibility with an arbitrary platform. Have you tried reaching out to them?

PR #10930, opened 2026-06-09
Fix macro replay after auto-yield input flow changes

Madwand99:
This PR updates macro replay to work with the newer input flow introduced around auto-yield changes in #10606.

After those changes, the macro system could no longer assume that recorded actions would be replayed against the same simple prompt sequence as before. In practice, replay could stall or drift when prompts appeared between recorded actions, especially around:

- pass priority across phase/stack changes
- simultaneous triggered ability ordering
- target/list selection prompts
- optional confirm/decline prompts
- mana payment prompts and recorded mana-source selections

Without adapting to those input-flow changes, recorded macros can fail even when the original sequence is still legal and repeatable.

The main change is that replay now treats recorded actions as choices that may need to wait for, skip past, or satisfy the current GUI input prompt. `RecordActionsMacroSystem` keeps the macro sequencing/retry logic, while `MacroActionReplayer` handles GUI-specific replay of individual recorded actions against the current `Input` type.

This work was split off from the Macro Window UI work (#10746) so the replay compatibility fixes can be reviewed on their own.


forge-game/src/main/java/forge/game/player/actions/PassPriorityAction.java
```diff
@@ -20,8 +20,52 @@ public boolean wasStackEmpty() {
...
+    }
+
+    @Override
+    public boolean clearsPostStackOrderWait() {
+        return true;
+    }
+
+    @Override
+    public PassPriorityAction asPassPriorityAction() {
+        return this;
```
tool4ever: bloat method, might aswell stick with instanceof

tool4ever:
I'm not convinced trying to solve this with somewhat convoluted heuristics is a good direction:
can't you simply suppress yield suggestions or anything else that messes replay up during it? I can't really think of a reasonable use case that's worth attempting to include auto-pass as part of a macro sequence...

forge-game/src/main/java/forge/game/player/actions/PassPriorityAction.java
Madwand99: Agreed, fixed!

Madwand99:
I've changed the PR to suppress auto-yield behavior during recording/replay. This just leaves the numerous miscellaneous improvements to replay handling.

tool4ever:
Still in my backlog but low priority
kind of still want to spend some time thinking about the general architecture

Madwand99:
> Still in my backlog but low priority kind of still want to spend some time thinking about the general architecture

For now, I've greatly simplified this branch to mostly just the "disable auto-yield" function. If this simplified version is accepted, I can work on the more complex improvement after that.

jamincollins:
## Findings

#### 1. Macro active check prevents auto-yield during macro replay in InputPassPriority.showMessage
The change adds `!getController().isMacroActive()` guard before checking yield controller state in InputPassPriority.showMessage(). This prevents auto-yield logic from suppressing priority prompts during macro recording/replay.

The logic is correct: during macro replay, we want to replay the exact sequence of user actions, not skip prompts based on auto-yield preferences. The guard ensures priority prompts are shown even if auto-yield would normally suppress them.

Test case:
```java
// Verify macro active prevents yield suppression
PlayerControllerHuman controller = new PlayerControllerHuman(game, player, lobbyPlayer);
controller.macros().setRecording(true);
InputPassPriority input = new InputPassPriority(controller);
yieldController.setBoolPref(FPref.YIELD_SUPPRESS_AFTER_END, true);
yieldController.setDidYieldJustEnd(true);

input.showMessage(); // Should show normal prompt, not suppressed due to yield
// Expected: showNormalPrompt() called, not early return from yield suppression
```

Impact: Correct. Prevents macro desync where replayed game skips prompts that were present during recording.

#### 2. Missing null safety for macros in isMacroActive()
The isMacroActive() method checks `macros != null` before calling methods, but the method is called from multiple places without ensuring macros is initialized. If macros is null, isMacroActive() returns false, which is safe.

Test case:
```java
PlayerControllerHuman controller = new PlayerControllerHuman(game, player, lobbyPlayer);
// macros not initialized
boolean active = controller.isMacroActive(); // Should return false, not NPE
assert !active;
```

Impact: Low risk. Null check exists, but method could be called before macros initialization in edge cases.

#### 3. isUiSetToSkipPhase returns false during macro replay
The change makes isUiSetToSkipPhase return false when macro is active. This prevents phase skipping during macro replay.

Potential edge case: If macros is null, isMacroActive() returns false, so isUiSetToSkipPhase proceeds normally. This is acceptable fallback behavior.

Test case:
```java
PlayerControllerHuman controller = new PlayerControllerHuman(game, player, lobbyPlayer);
controller.macros().setReplaying(true);
// During macro replay, phase skip UI settings should not apply
assert !controller.isUiSetToSkipPhase(playerView, PhaseType.COMBAT);
```

Impact: Correct behavior. Ensures macro replay follows exact phase sequence.

#### 4. confirmTrigger auto-yield decision bypass during macro
confirmTrigger now skips AutoYieldStore lookup when macro is active. This prevents recorded auto-yield decisions from interfering with macro replay.

Edge case: If macro is active but macros is null, isMacroActive() returns false, so auto-yield logic would still run. This could cause desync.

Test case:
```java
PlayerControllerHuman controller = new PlayerControllerHuman(game, player, lobbyPlayer);
// Simulate macros null scenario
controller.macros = null;
WrappedAbility wrapper = new WrappedAbility();
boolean result = controller.confirmTrigger(wrapper); // Should not throw NPE
```

Impact: Moderate. Null macros edge case could cause inconsistent behavior.

#### 5. recordConfirm now captures confirm dialog results
PlayerControllerHuman.confirmTrigger now calls recordConfirm for both Libgdx and standard paths, ensuring macro recording captures confirm dialog outcomes.

This is backwards compatible: previously confirm dialogs weren't recorded for macro replay, causing replay to use default decisions. Now decisions are recorded.

Test case:
```java
PlayerControllerHuman controller = new PlayerControllerHuman(game, player, lobbyPlayer);
controller.macros().setRecording(true);
WrappedAbility wrapper = mock(WrappedAbility.class);
when(wrapper.getCardView()).thenReturn(cardView);
// Trigger confirm dialog
boolean result = controller.confirmTrigger(wrapper);
// Verify ConfirmAction recorded
ConfirmAction action = controller.macros().getRememberedActions().getLast();
assert action.getCardView() == cardView;
assert action.getResult() == result;
```

Impact: Improves macro fidelity. Breaking change in behavior but intended.

#### 6. mayAutoPass disabled during macro
mayAutoPass() now returns false when macro is active, preventing auto-pass during macro replay.

This is correct for data integrity: macro replay must not auto-pass based on current game state, must replay recorded actions.

Test case:
```java
PlayerControllerHuman controller = new PlayerControllerHuman(game, player, lobbyPlayer);
controller.macros().setReplaying(true);
yieldController.setAutoYield(true);
assert !controller.mayAutoPass(); // Auto-pass disabled during replay
```

Impact: Correct. Maintains replay fidelity.

#### 7. Interaction between isMacroActive and yieldController.shouldAutoYield
The guard `!isMacroActive() && yieldController.shouldAutoYield()` in declareAttackers and checkStackForAutoYield prevents auto-yield from skipping inputs during macro.

Side effect: If macro recording starts mid-game, auto-yield may suddenly stop working, potentially causing unexpected prompts. However, this is desirable for recording fidelity.

Test case:
```java
PlayerControllerHuman controller = new PlayerControllerHuman(game, player, lobbyPlayer);
yieldController.setAutoYield(true);
// Start recording mid-game
controller.macros().setRecording(true);
// Auto-yield should be disabled for subsequent actions
assert !controller.mayAutoPass();
```

Impact: Acceptable. Recording fidelity takes precedence over convenience.

#### 8. setTriggerDecision early return during macro
setTriggerDecision returns early if macro is active, preventing auto-yield decisions from being applied during replay.

Side effect: Trigger decisions made during recording are not applied during replay, which is correct. However, if macro recording is started after some triggers have been auto-yielded, the recorded decisions may not match replay behavior.

Test case:
```java
PlayerControllerHuman controller = new PlayerControllerHuman(game, player, lobbyPlayer);
controller.macros().setReplaying(true);
controller.setTriggerDecision("key", AutoYieldStore.TriggerDecision.ACCEPT, false);
// Should not select button OK during replay
```

Impact: Correct. Prevents replay from modifying UI state based on auto-yield.

### Performance and security
#### 9. No performance impact
All changes are simple boolean checks. isMacroActive() is O(1). No performance concerns.

#### 10. Security: macros() method exposure
The macros() method returns IMacroSystem which could be manipulated. However, this is internal API and no security issue.

## Verdict
Request Changes

The PR correctly addresses macro replay desync issues caused by auto-yield input flow changes. The isMacroActive guards prevent auto-yield logic from interfering with macro recording/replay, and recordConfirm captures confirm dialog results for fidelity. However, null safety for macros initialization needs verification, and the interaction between macro state changes mid-game could cause unexpected behavior. The changes appear correct but need testing for edge cases where macros is null or macro state changes during gameplay.

## Recommended Fixes
1. Add defensive null check for macros in isMacroActive() - already present but verify initialization order
2. Ensure macros is initialized before any isMacroActive() checks are performed
3. Add unit tests for macro active state transitions mid-game
4. Document behavior change for confirm dialog recording

---

forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -3892,6 +3904,9 @@ public void setDisableAutoTriggers(final boolean disable) {
     }
 
     public boolean isUiSetToSkipPhase(final PlayerView turnPlayer, final PhaseType phase) {
+        if (isMacroActive()) {
```
tool4ever: Can't we let the user still take advantage of the phase stops while recording and only when replaying we do it manually?

Madwand99: Agreed. Recording now honors the player’s phase stops. When a phase is skipped during recording, the macro records the implicit priority pass; during replay, phase-stop settings are ignored so the recorded passes control advancement. Stack auto-yields remain suppressed during recording to avoid unrecorded decisions. I tested replay with the Draw stop both enabled and disabled during recording.

tool4ever (approved):

PR #10931, opened 2026-06-09
Add Reed Richards

tehdiplomat:


Fulgur14 (approved):

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -1248,6 +1249,12 @@ private CardCollectionView doDraw(Map<Player, CardCollection> revealed, SpellAbi
                     runParams.putAll(params);
                 }
 
+                if (sa == null) {
+                    // if this draw is not caused by a SA, it must be the normal draw at the beginning of the turn, so update the flag for that
+                    numExtraDrawsThisTurn++;
+                    runParams.put(AbilityKey.ExtraDraws, numExtraDrawsThisTurn);
```
tool4ever: you're doing a replacement but modifying the params for the trigger :P

I'll make a fix

tool4ever (dismissed):
I couldn't find a way to avoid extra count

forge-game/src/main/java/forge/game/player/Player.java
```diff
@@ -1238,8 +1240,12 @@ private CardCollectionView doDraw(Map<Player, CardCollection> revealed, SpellAbi
                 setLastDrawnCard(c);
                 c.setDrawnThisTurn(true);
                 numDrawnThisTurn++;
+                numExtraDrawnThisTurn++;
                 if (game.getPhaseHandler().is(PhaseType.DRAW)) {
                     numDrawnThisDrawStep++;
+                    if (numDrawnThisDrawStep == 1) {
+                        numExtraDrawnThisTurn--;
```
tehdiplomat, line 1247: I think this is wrong. What if you draw during a different players draw step?

tehdiplomat: Also if you skip your "normal" draw to Pursuit of Knowledge, but then extra draw during your draw step. @tool4ever

tool4ever: that's why I'm checking for it in ReplaceDraw again
not sure if that can be consolidated differently

tehdiplomat: Whats wrong with checking `sa == null` like I originally had?

forge-game/src/main/java/forge/game/replacement/ReplaceDraw.java
```diff
@@ -47,18 +46,25 @@ public ReplaceDraw(final Map<String, String> params, final Card host, final bool
...
-            final Game game = getHostCard().getGame();
+            if (p.numDrawnThisDrawStep() == 0 && ownDraw) {
+                return false;
+            }
+        }
 
-            final Player p = (Player)runParams.get(AbilityKey.Affected);
-            if (p.numDrawnThisDrawStep() == 0 && game.getPhaseHandler().is(PhaseType.DRAW, p)) {
+        if (hasParam("FirstExtraCardDrawnThisTurn")) {
+            if ((int) runParams.get(AbilityKey.ExtraDraws) != (ownDraw ? 1 : 0)) {
```
tehdiplomat, line 67: I don't understand what `ownDraw` has to do with extra draws at all?

forge-game/src/main/java/forge/game/player/Player.java
tool4ever: technically if there was a "draw two instead" replacement the cause should stay null, even if that's currently not handled yet

forge-game/src/main/java/forge/game/replacement/ReplaceDraw.java
tool4ever: that's the tailoring for this param since then the second is the "first extra" instead

tehdiplomat: But as I mentioned up above, if you replace your draw step draw with any type of replacement effect that doesn't draw cards, any future draw step draws are still "extra" draws and should not be considered your normal draw. Extra draws are just extra draws.

forge-game/src/main/java/forge/game/player/Player.java
tehdiplomat: So the example is:

1) Blood Scrivener + Normal Draw step draw on an empty hand?

this ends up  drawing normal draw + N extra cards, the first of which should be replaceable by Reed?

forge-game/src/main/java/forge/game/replacement/ReplaceDraw.java
Fulgur14: Technically, it doesn't care about "normal" draw -- it cares about the first draw in the draw step. That will usually be the normal draw, but in some situations, like if the draw is replaced with Words of Wisdom and there's Howling Mine out, your first draw in your draw step will be different.

tehdiplomat: Ahh i guess so.

tool4ever (approved):

PR #10937, opened 2026-06-09
Add forge-scryfall-uuid-map CLI tool for patching edition files

phughk:
## Problem

To use Scryfall's rate-free CDN for card images, we need a UUID for every (set, collector-number, language) combination. Those UUIDs aren't in Forge's existing data. Scryfall publishes bulk-data exports with all of them, but they need to be processed into per-set JSON files that the runtime can fetch on demand.

This tool is the offline generator — run it once (or after each Scryfall data update) to produce the files that go in the forge-extras repository.

## What it does

```mermaid
flowchart LR
    A[Scryfall bulk-data\nall_cards JSON] --> B[BulkDataFetcher\ndownloads & streams]
    B --> C[CardStreamParser\nreads UUID + set + cn + lang]
    C --> D["CdnUuidJsonWriter\nwrites cdn_uuid/{set}.json"]
    D --> E[~1,041 set files\nready for forge-extras]
```

Output — one JSON file per Scryfall set code:
```
cdn_uuid/neo.json
{
  "1":   {"pt":"…uuid…","de":"…uuid…","ja":"…uuid…","en":"…uuid…"},
  "2":   {"en":"…uuid…"},
  "A-40":{"en":["front-uuid","back-uuid"]}
}
```

DFCs with distinct face UUIDs store an array for the affected language.

## Usage

Run from the repo root. No full Forge build required.

```bash
# Full multi-language output — downloads all_cards from Scryfall (~2.5 GB)
mvn -pl forge-scryfall-uuid-map exec:java \
  -Dexec.args="--output-dir /path/to/forge-extras/cdn_uuid"

# English-only, faster — downloads default_cards (~100 MB)
mvn -pl forge-scryfall-uuid-map exec:java \
  -Dexec.args="--output-dir /path/to/forge-extras/cdn_uuid --default-cards"

# Use a pre-downloaded bulk file (auto-detected if named all-cards-*.json or default-cards-*.json)
mvn -pl forge-scryfall-uuid-map exec:java \
  -Dexec.args="--output-dir /path/to/forge-extras/cdn_uuid --bulk-file all-cards-20260608.json"
```

The default (`all_cards`) includes every language variant and produces a complete multi-language UUID map. `--default-cards` is faster to regenerate but only covers English — the runtime falls back to the Scryfall API for non-English cards when a UUID is absent.

## New module

`forge-scryfall-uuid-map/` — standalone Maven module, not included in the game distribution. Added to the root `pom.xml` modules list.

| Class | Role |
|---|---|
[...]

PR #10928, opened 2026-06-08
Fix bulk card image downloader to use Scryfall CDN

phughk:
# Card image downloader: CDN cache + real rate limiting

## Summary

Image downloads went through Scryfall's rate-limited API one card at a time.
A single 429 froze *all* downloads for a flat 5 minutes.
This adds a local cache of Scryfall's CDN image URLs, built via bulk data or per-set search, so most downloads skip the API and hit the unthrottled CDN directly.
Rate limiting itself is also fixed: correct per-endpoint pacing, and honoring `Retry-After` instead of a blanket cooldown.

## Key changes

- `CdnUuidCache` — local per-set cache: card/lang/face → CDN image UUID. Read-only on the gameplay path; a miss queues a background sync instead of blocking.
- `ScryfallBulkDataSync` — one-shot download of Scryfall's full catalog (~350MB) to populate the cache for every set at once.
- `ScryfallSetSync` — per-set fallback via `/cards/search`, for when only a few sets are missing.
- `ScryfallRateLimiter` — replaces the old ad-hoc pacing/cooldown code; matches Scryfall's real per-endpoint limits and respects `Retry-After`.
- `ImageFetcher`, `GuiDownloadFilteredCardImages`, `GuiDownloadService` — prefer CDN URLs when cached, warm the cache before a bulk run, wait out cooldowns instead of skipping.
- `DialogDownloadCardImages` (desktop) / `CardImageBrowserScreen` (mobile) — new "sync bulk data" and "clear CDN cache" buttons.
- `FView`, `Forge`, `SplashScreen` — first-run prompt offering the bulk sync.
- New tests (`CdnUuidCacheTest`, `ScryfallSetSyncTest`), new `gson` dependency, new translation strings.

## Component interaction

```mermaid
flowchart LR
    subgraph UI
        Dialog[DialogDownloadCardImages /<br/>CardImageBrowserScreen]
    end

    subgraph "Gameplay image path"
        IF[ImageFetcher<br/>Swing/LibGDX impls]
    end

    subgraph "Bulk download path"
        GDF[GuiDownloadFilteredCardImages]
        GDS[GuiDownloadService]
    end

    subgraph "Sync sources"
        SetSync[ScryfallSetSync<br/>per-set /cards/search]
        BulkSync[ScryfallBulkDataSync<br/>full catalog file]
    end

    Cache[(CdnUuidCache<br/>per-set .json.gz)]
    RL{{ScryfallRateLimiter}}

    Dialog -->|Download button| GDF
[...]

PR #10938, opened 2026-06-09
Add scryfall_uuid extra-params to edition files

phughk:
## Problem

PR #10928 (the runtime CDN fetcher) needs a UUID for each card to construct non-rate-limited CDN URLs. Those UUIDs must be bundled with the game so they are available offline and without any API calls at startup.

## What this PR contains

115,230 small JSON files under `forge-gui/res/cdn_uuid/`:

```
res/cdn_uuid/
├── 6ed/
│   ├── 1.json    {"de":"…","en":"…","fr":"…",…}
│   ├── 2.json
│   └── …
├── ltr/
│   └── …
└── … (1,041 sets)
```

Each file maps language codes to the Scryfall UUID for that printing. DFCs store `[frontUuid, backUuid]` when the faces differ.

## How it was generated

Using the CLI tool from PR #10937 against Scryfall's `default-cards` bulk-data snapshot.

## Size note

The files are plain JSON, typically ~100–300 bytes each. Total uncompressed ~20 MB; they compress well and are included in the existing `res/` directory that ships with every Forge distribution.

## Depends on

PR #10928 — the runtime code that reads these files. Without it, the files are inert.


tehdiplomat (changes requested):
I'll have comments later, just want to make sure this doesn't get merged for now

phughk:
Closing and moving to forge extras

PR #10945, opened 2026-06-10
Avoid AI wheel effects into draw punishers

Rauglothgor:
## Summary

This is a narrow AI heuristic for voluntary wheel-style mass draw effects when an opponent already controls an active draw-punisher or draw-conversion permanent.

The change is intentionally scoped to wheel execution paths instead of the global playability chain:

- `DiscardAi` for discard-then-draw wheels such as Wheel of Fortune style effects
- `ChangeZoneAllAi` for Timetwister-style shuffle/draw effects
- a shared `ComputerUtilCard` helper that estimates large self-draws and checks active battlefield draw-punisher triggers

It avoids the broader simulator/caching direction from #10559 and the high-level `AiController` hook concern discussed on #10576.

## Behavior

The AI now declines large voluntary mass-draw lines when an opponent's battlefield has an active trigger that materially punishes or converts the AI drawing cards, such as damage/life-loss draw punishers or token conversion.

The AI still allows the line when the draw is small, no active opponent draw-punisher exists, or the AI's own draw-punisher damage appears immediately lethal.

## Validation

Passed locally with temporary WSL Maven/JDK tooling:

```text
```

Observed result: `BUILD SUCCESS`, with 0 checkstyle violations.

## Notes

No tests are included here. This is meant to stay as a small upstream review slice after recent maintainer feedback about avoiding test-only wiring for AI heuristics unless the test catches a meaningful integration regression.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
@@ -76,6 +76,8 @@ import forge.util.MyRandom;
 import forge.util.TextUtil;
 
 public class ComputerUtilCard {
+    private static final int DANGEROUS_MASS_DRAW_COUNT = 3;
+
     public static Card getMostExpensivePermanentAI(final CardCollectionView list, final SpellAbility spell, final boolean targeted) {
         CardCollectionView all = list;
         if (targeted) {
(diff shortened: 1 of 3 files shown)


tool4ever:
Let's just keep working with the original PR here too

Issue #6913, opened 2025-01-28
AI always chooses the highest MV card..no matter what


P-rus:
This has existed for a while now and has yet to be addressed.  It encompasses multiple situations and cannot be quantified in simple terms.   I will explain in detail below to clarify:

- When searching, AI always chooses the highest possible MV in the deck, regardless of available mana
- if ramping multiple lands, the same type is always chosen, no matter the need
- if targeting non-creatures with removal, the highest MV  is chosen, even if it doesn't affect their board (i.e destroying O-ring targeting an opponent's permanent)
- If targeting creatures, the highest MV or P/T is chosen, whether for removal or modifying (i.e counters are always put on one creature)
- If multiple targets can be chosen due to a copied spell, the same card is chosen (i.e cloned recursion cards that hit simultaneously and/or copied removal effects)
- AI ignores 'for each opponent' clause in cards that target multiple permanents and instead target whichever happen to be the highest MV on the battlefield (i.e Grasp of Fate type spells)
- AI prioritizes MV over board state (i.e number of permanents or total board presence per opponent doesn't matter)
- If there is a legal target, the removal is always used, no matter the MV (i.e Swords to Plowshares on the highest MV creature when mana becomes available)
- Will not board wipe if it has a board presence more than an indeterminate MV, even if there is a far greater presence on opponent's side (i.e Nevinnyral's Disc is not played optimally and often is allowed to be removed)
- Board wipe effects are always played if AI has no board presence, even if there are no viable targets (i.e Cleansing Meditation )

This vastly affects the type of gameplay that goes on in FSM.  There can be no 'casual' styled game play and you cannot test sub-optimal decks that fit with your playgroup's style.  The AI always removes, no matter what.  The AI can be predicted, no matter what.  

**Suggestions for improving AI performance:**

- Have the AI prioritize starred cards ( favorited ) allowing the player to customize AI behavior
- Modify AI behavior on removal effects to be more like what was done to Sakura Tribe-Elder ( great job on that by the way )
- Fix the 'for each opponent' clause 
- Have the AI include total opponent board presence in the calculation when using targeted removal

Thanks again for everyone's hard work.

MorihiroS:
Requests about same problems:
#4282 
#5896

tool4ever:
With above PR I'm going to close this one:
- this is really a mix of too many different AI areas
- also some have seen significant progress
- others like _Cleansing Meditation_ are a special case due to their script structure
[...]

Issue #4282, opened 2023-12-03
AI searches for itself/discards to player cards with high cmc in early game


MorihiroS:
**Describe the bug**
In early game AI searches for itself or discards to player cards with high cmc like Avacyn, Angel of Hope or Iona, Shield of Emeria. But it doesn't make sense

**To Reproduce**
1. AI casts Thoughtseize and targets player
2. AI sees 3 lands, Noble Hierarch, Serra Paragon, Touch the Spirit Realm and Avacyn, Angel of Hope
3. AI chooses Avacyn, Angel of Hope to discard
or
1. AI casts Eladamri's Call
2. AI hasn't 8 lands or Elvish piper on the battlefield
3. AI finds Avacyn, Angel of Hope

**Expected behavior**
AI should at least count his lands/opponent's lands to choose which card it should find/discard

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-12.01

**Screenshots**

![though1](https://github.com/Card-Forge/forge/assets/115870280/80cd6a49-95bf-4c7f-80ae-280aa1de4041)
![though2](https://github.com/Card-Forge/forge/assets/115870280/422357c8-fe4f-4e50-8292-d76ab16f535c)
![though3](https://github.com/Card-Forge/forge/assets/115870280/e4f73cac-5cc9-45fc-ae88-303a25ded380)
![though4](https://github.com/Card-Forge/forge/assets/115870280/e62253a0-194a-4888-a27b-e1a5a20e2f4b)

MorihiroS:
@Hanmac, need keep label, please

MorihiroS:
@Agetian, but AI don't choose cards with high cmc from player's hands by trigger of Thought-Knot Seer:

![though](https://github.com/user-attachments/assets/ed39cd8f-b359-4e8f-8bd2-fd86c7ad83c9)

MorihiroS:
@Agetian, can you add the logic of Thought-Knot Seer's trigger to "search nonland card for AI" and "choose opponent's card to discard"? Tutors and cards like Thoughtseize are almost useless cards for the AI, because of which it loses to itself frequently...

PR #10947, opened 2026-06-10
Prefer near-term threats for AI card choices

Rauglothgor:
## Summary

This is a narrow first slice for #6913 and #4282. It reduces two common highest-MV-only AI choices without changing the broader targeting system:

- Thoughtseize-style opponent hand discard now prefers a near-term playable threat before falling back to the highest-MV card.
- Creature-to-hand tutor choices now prefer castable or soon-castable creatures before falling back to the largest creature.

The goal is to avoid early-game choices such as taking or tutoring Avacyn/Iona purely because they have high mana value when lower-MV cards are the actual near-term threat.

## Behavior

For single-card discard from an opponent's hand, key cards still win first. If no registered key card applies, the AI checks nonland cards that appear playable soon based on current mana plus a likely land drop, then chooses the best threat from that subset.

For creature tutors to hand, the AI now checks creatures it can cast with available mana sources, then creatures close to castable after a likely land drop, before falling back to the previous best-creature choice.

## Scope

This deliberately does not attempt to solve every item in #6913. It does not change Pithing Needle naming, multi-target copied-spell choices, board wipes, or for-each-opponent targeting. Those should remain separate patches because they touch different AI paths.

## Validation

Passed locally with temporary WSL Maven/JDK tooling:

```text
```

Observed result: `BUILD SUCCESS`, with 0 checkstyle violations.

## Notes

No tests are included here. This is meant as a small behavior slice rather than broad AI rewiring.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
@@ -605,6 +605,18 @@ public class ChangeZoneAi extends SpellAbilityAi {
             return null;
         }
 
+        if (ai.getTurn() <= 3) {
+            int manaSources = ComputerUtilMana.getAvailableManaEstimate(ai, false);
+            if (CardLists.count(ai.getCardsIn(ZoneType.Hand), CardPredicates.LANDS_PRODUCING_MANA) > 0) {
+                manaSources++;
+            }
+            final int nearTermMana = manaSources + 1;
+            CardCollection nearTerm = CardLists.filter(list, c -> c.getCMC() <= nearTermMana);
+            if (!nearTerm.isEmpty()) {
+                return ComputerUtilCard.getBestCreatureAI(nearTerm);
+            }
+        }
+
         // not urgent, get the largest creature possible
         // TODO checkETBEffects
         return ComputerUtilCard.getBestCreatureAI(list);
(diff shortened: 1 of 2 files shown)


MorihiroS:
@Rauglothgor Thank you for taking care of this. This is one of the biggest AI problems in Forge (along with playing lands tapped)!

Rauglothgor:
Thanks. I kept this PR limited to the near-term target-priority cases from the linked issues, so it should stay reviewable while still covering a visible AI weakness.

forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
```diff
@@ -605,6 +605,22 @@ private static Card chooseCreature(final Player ai, CardCollection list) {
             return null;
         }
 
+        CardCollection castable = CardLists.filter(list,
```
tool4ever: - prefiltering the whole list with an expensive method is terrible for performance
- but there's really no point in reproducing almost the same logic already checked above anyway...if you want to improve early tempo could just add a condition for turn number or something like that

Rauglothgor: Updated in 347bae89f0b. I removed the full `hasEnoughManaSourcesToCast` prefilter in `chooseCreature`; the tutor preference now only uses a cheap CMC filter during the AI's first three turns, leaving the existing life-in-danger castability path unchanged.

Validated with `mvn -pl forge-ai -am -DskipTests compile` under WSL/JDK 17.

tool4ever (approved):

PR #10948, opened 2026-06-10
Add optional CommanderBracket.app integration for bracket estimation

Madwand99:
This adds an optional CommanderBracket.app integration for richer Commander bracket estimates while keeping Forge’s local bracket calculation as the default fallback.

Key changes:

- Adds a new Advanced preference, default off: “Use CommanderBracket.app for bracket calculations”.
- Adds a CommanderBracket service/API client with:
  - public CommanderBracket.app endpoint support
  - Forge/x.y User-Agent
  - single-worker rate limiting
  - request de-duplication and high-priority promotion
  - session caching and failure cooldowns
- Stores API bracket metadata directly in deck files:
  - DeckHash=(sha256)
  - CommanderBracket=(1-5)
- Uses saved/session API brackets in the deck browser and match tooltip when available.
- Keeps local Forge bracket calculation available when the API is disabled, unavailable, or still pending.
- Adds clickable “Powered by Commander Bracket” attribution.
- Updates bracket detail panels to refresh asynchronously when high-priority API details return.
- Adds localized UI strings for all available Forge languages.
- Adds a small shared JsonUtil for simple JSON parsing/escaping used by the API integration.

CommanderBracket

Thanks to Codex for help with code and keith@commanderbracket.app for help with API access!


tool4ever (commented):
Sorry, I liked your offline implementation but I'm not really a fan of such logic to simply connect with a third-party API, besides needing all this massive code hardly being reasonable trade-off
- only to maybe get a more accurate value
- save a few mouse clicks for the user to just use his browser instead

(And Scryfall alone already causes enough headache)

Madwand99:
> Sorry, I liked your offline implementation but I'm not really a fan of such logic to simply connect with a third-party API, besides needing all this massive code hardly being reasonable trade-off

That’s fair, and I understand the concern. I agree that Forge should be cautious about third-party services, especially given the ongoing Scryfall maintenance burden.

The reason I think this case may still be worth considering is that the offline implementation and the CommanderBracket.app result are solving different problems. Forge’s local calculation is intentionally a minimum bracket estimate. It is useful as a floor, but it cannot give the kind of holistic bracket estimate Commander players usually want: deck speed, tutors, combo density, game changers, archetype context, and narrative reasoning.

We could try to port that logic into Forge, but that would likely be much larger than this PR and would also mean maintaining the model ourselves as Commander bracket guidance evolves. This PR is intended as a compromise: keep Forge’s lightweight offline estimate, while optionally allowing users to ask CommanderBracket.app for a better estimate when they want one.

I also tried to limit the blast radius:

- the feature is opt-in and defaults off
- Forge still works fully offline
- local bracket calculation remains the fallback
- results are cached/saved for user decks
- requests are rate-limited

So I don’t see this as replacing Forge logic with a web dependency. I see it as giving users an optional bridge to a specialized bracket estimator, while keeping Forge’s existing offline behavior intact.

The real question may be whether Forge wants a more accurate Commander bracket estimate at all. If yes, then I think either we maintain a much larger model in Forge, or we use an optional external integration like this. My feeling is that this PR is the smaller and more maintainable version of that feature, but I’m open to reducing the scope further if there are specific parts that feel too invasive.

Madwand99:
I completed another architecture and cleanup pass based on the earlier review.

The CommanderBracket integration is now divided more clearly between service orchestration, API communication, and response interpretation. JSON parsing is shared with `DeckUrlLoader`, allowing its duplicate parser to be removed. The latest cleanup removed roughly 230 production lines.

I also tightened several behavioral details:

- API requests stop when the preference is disabled.
- `429` retries are bounded and respect `Retry-After`.
- Invalid or contradictory API results are rejected or sanitized.
- Session and saved bracket values follow an explicit precedence.
- Match tooltips reuse available results without initiating requests.
- Deck-manager updates target relevant views and only perform a full refresh when bracket sorting requires it.
- Metadata-only saves preserve the original card order.
- Generated decks and Forge precons are excluded from background persistence and requests while remaining available for explicit high-priority analysis.

Short rationale comments now document the less obvious queue, caching, hashing, persistence, and Swing layout decisions.

jamincollins:
## Findings
#### 1. Deck hash computation ignores card order and may collide across decks with same cards in different order
Description: `toCanonicalCommanderBracketDecklist` sorts lines case-insensitively but uses `String.CASE_INSENSITIVE_ORDER` which is not locale-independent and does not guarantee stable ordering across JVM versions. More critically, the canonical list uses `"C 1 "` and `"M "` prefixes that do not encode card set/edition information; two cards with same name in different sets will hash identically.
Test case:
```java
Deck d1 = new Deck();
d1.addCard(new PaperCard("Lightning Bolt")); // Set A
Deck d2 = new Deck();
d2.addCard(new PaperCard("Lightning Bolt")); // Set B
String h1 = CommanderBracketService.hashDecklist(toCanonicalDecklist(d1));
String h2 = CommanderBracketService.hashDecklist(toCanonicalDecklist(d2));
assert !h1.equals(h2); // fails: same hash, different cards
```
Impact: Different decks could share API cache entries leading to wrong bracket estimates.

#### 2. saveMetadata writes DeckHash/CommanderBracket without validating hash matches current deck state
Description: `DeckStorage.saveMetadata` inserts hash/bracket lines without recomputing hash from current deck contents. If deck changed on disk but hash not updated, stale bracket persists.
Test case:
```java
Deck deck = loadDeck("mydeck.dck");
deck.setCommanderBracket("oldhash", 3);
storage.saveMetadata(deck);
// modify deck cards
storage.saveMetadata(deck); // hash still old, bracket still 3
```
Impact: Backwards compatibility break; bracket may no longer match deck.

#### 3. CommanderBracketService.getBracket returns 1 for generated decks, suppressing API
Description: Generated decks return 1 without checking if API is enabled or if deck is Commander. This may mask errors.
Test case:
```java
DeckProxy proxy = DeckProxy.forGeneratedDeck(commanderDeck);
int b = CommanderBracketService.getBracket(proxy);
assert b == 1; // always
```
Impact: False bracket for generated Commander decks.

#### 4. DeckContext.create with null deck uses CommanderBracketCalculator.calculate(null) which may throw NPE
Description: `CommanderBracketCalculator.calculate(null)` may assume non-null deck and dereference.
Test case:
```java
DeckContext ctx = DeckContext.create(null);
CommanderBracketCalculator.Result r = ctx.getLocalResult(); // NPE?
```
Impact: Crash in tooltip rendering when deck is null.

#### 5. JsonUtil.parse does not allow trailing whitespace after root value; API response may include newline
Description: Parser enforces `parser.index != parser.text.length()` throws IOException on trailing newline.
Test case:
```java
JsonUtil.parse("{\"a\":1}\n"); // IOException: Trailing JSON data
```
Impact: Valid HTTP responses rejected.

#### 6. CommanderBracketApiClient firstCommanderName returns null if commander line contains extra spaces
Description: Regex `^\\d+\\s+` removes quantity but line may contain leading spaces.
Test case:
```java
String decklist = "// Commander\n 1  Animar";
String name = firstCommanderName(decklist); // returns "1  Animar" not "Animar"
```
Impact: API request with malformed commander name.

#### 7. DeckFileHeader.parseCommanderBracket accepts bracket 0? No, validates 1-5. But `setCommanderBracket` allows null hash with non-null bracket
Description: `setCommanderBracket(String hash, Integer bracket)` does not validate hash non-null when bracket non-null.
Test case:
```java
deck.setCommanderBracket(null, 3); // allowed
```
Impact: Deck file may contain CommanderBracket without DeckHash.

#### 8. DeckStorage.saveMetadata inserts lines after NAME/COMMENT but before section header; may duplicate
Description: Loop removes existing hash/bracket lines but continues after insertion point change.
Test case:
```java
List<String> lines = ["[metadata]", "Name=Test", "DeckHash=old", "CommanderBracket=3", "[main]"];
saveMetadata(deck); // may leave duplicate
```
Impact: Duplicate metadata lines break parser.

#### 9. CommanderBracketService cacheRemoteResult saves metadata even when API disabled
Description: `cacheRemoteResult` calls `deck.setCommanderBracket` unconditionally then `saveDeckMetadata` only if proxy can save. But API disabled path may still call cache from stale session.
Test case:
```java
set pref UI_USE_COMMANDER_BRACKET_API false;
CommanderBracketService.getExplanation(deck, proxy); // may cache from session
```
Impact: Metadata written even when user opted out.

#### 10. API client uses single worker thread; priority queue compareTo uses `priority.compareTo` where enum order is declaration order HIGH < LOW? 
Description: `Priority.HIGH, LOW` enum ordinal HIGH=0, LOW=1. `compareTo` returns negative for HIGH vs LOW, so queue orders HIGH first? PriorityQueue uses natural order, smaller ordinal first. OK. But promotion of queued request may duplicate in-flight set.
Test case:
```java
enqueue(hash, HIGH); // inFlight add succeeds
enqueue(hash, LOW); // inFlight contains, promotes queued request
```
Impact: Potential duplicate processing.

### Performance and security
#### 11. CommanderBracketApiClient posts decklist without size limit; large decks may exceed URL length? POST body but no size check.
Description: Decklist sent as JSON string without truncation; could be MBs.
Test case:
```java
String huge = "1 Card\n".repeat(100000);
postDeck(huge, hash); // OOM or timeout
```
Impact: DoS via large deck.

#### 12. JsonUtil escape does not escape control characters beyond 0x20; may produce invalid JSON for exotic card names.
Test case:
```java
String name = "Card\u001fName";
String escaped = JsonUtil.escape(name); // produces \u001f? Not handled
```
Impact: API request malformed.

## Verdict
Request Changes
The PR introduces useful API integration but has multiple safety and correctness gaps: hash collisions due to missing set info, metadata persistence without validation, null handling, JSON parser strictness, and potential duplicate metadata writes. Fixes needed before merge.

## Recommended Fixes
1. Include set/edition in canonical decklist for hash.
2. Recompute hash before saving metadata; validate hash/bracket pair.
3. Guard null deck in DeckContext.
4. Allow trailing whitespace in JsonUtil parser.
5. Trim commander name parsing robustly.
6. Validate setCommanderBracket requires non-null hash when bracket set.
7. Ensure saveMetadata removes duplicates atomically.
8. Respect API disabled flag in cache path.
9. Add size limit for decklist before POST.
10. Improve JsonUtil escape to handle all control chars.

---

PR #10929, opened 2026-06-08
Enable deck sleeves on desktop

MostCromulent:
Screenshot 2026-06-09 075350

Screenshot 2026-06-09 075357

Screenshot 2026-06-09 075526

Closes Card-Forge/forge#8893.

## Summary

This wires up and enables sleeves on desktop: a lobby sleeve picker, the owner's sleeve drawn as the back of any card the viewer can't see, and a remote client's sleeve shown on the host.

## Implementation

- Most of the backend infrastructure for this was already present from [42f4126](https://github.com/Card-Forge/forge/commit/42f4126aff07cbfa58e45d2abfb0dd2dba317b30) ("Prepare Sleeve for Desktop") but commented out or unused — this PR enables and connects it.
- Lobby: a sleeve label beside the avatar (left-click opens the picker, right-click randomizes).
- Routed both the avatar and sleeve pickers through the label's own click command rather than an added mouse listener. A second adapter on a hoverable label loses clicks once a slot is disabled and re-enabled, which is why a network client could never edit its avatar or sleeve; this also fixes that latent avatar bug.
- Hidden-card back: ImageCache resolves the owner's sleeve for the t:hidden token and falls back to the standard card back when no sleeve is set.
- VLobby syncs the sleeve index so a client's selection shows on the host.

## Testing completed

- Desktop host with a desktop client: the client can pick its avatar and sleeve and both sync to the host's lobby.
- Cross-platform desktop + mobile multiplayer: sleeve selection works and syncs in both directions.
- In-match hidden cards render the owner's sleeve; with no sleeve chosen the standard card back renders.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10952, opened 2026-06-10
Improve lobby sleeve icon downscaling

MostCromulent:
Minor polish for https://github.com/Card-Forge/forge/pull/10929

The lobby panel drew the full 360x500 sleeve source straight into a 58x80 label, a ~6x single-pass reduction that looked aliased and rough. Instead, scale via area-averaging resize() at 2x the label size. This looks much smoother at small icon size and avoids aliasing.

The scaled image is cached so the downscale runs once rather than on every repaint.

---


tool4ever (changes requested):
This seems wrong since FLabel already has some own logic for it:
maybe the current `iconScaleFactor` would be a better approach?

MostCromulent:
Tested a few different approaches using FLabel; it seems none of the existing paths are adapted for a downscale of this size and all resulted in same aliasing problem. 

I thought most elegant solution was to just incorporate logic into the FLabel background icon painter so that where you have a very large downscale it applies the area-averaging. AI audited the call sites and currently the sleeves are the only case that hits this threshold. This should fix issue for any future cases though.

jamincollins:
## Findings

#### 1. Pre-scale cache uses Image.SCALE_SMOOTH which may be aliasing-prone for large downscales
Description:
The code caches a pre-scaled image via `new ImageIcon(img.getScaledInstance(tw, th, Image.SCALE_SMOOTH)).getImage()`. SCALE_SMOOTH is a hint, not area-averaging. The comment says "pre-scale it once with area-averaging", but Image.getScaledInstance with SCALE_SMOOTH does not guarantee area averaging; it typically uses bilinear filtering which can still alias on extreme reductions. For a 360x500 source to 2x draw size (~116x160), the reduction factor is ~3x, which may still alias.

Reference: forge-gui-desktop/src/main/java/forge/toolbox/FLabel.java:611

Test case:
```java
// Verify that scaledBg is created with SCALE_SMOOTH, not an area-averaging algorithm
FLabel label = new FLabel.Builder().iconInBackground(true).build();
label.setIcon(new ImageIcon(new BufferedImage(360,500,BufferedImage.TYPE_INT_ARGB)));
// Inspect scaledBg via reflection - it's created via getScaledInstance with SCALE_SMOOTH
```
Impact: Visual quality may be less than promised; comment claims area-averaging but implementation uses generic smooth scaling.

#### 2. Threshold 3x is arbitrary and may miss aliasing at 2.5x reduction
Description:
The condition `(iw > sw * 3 || ih > sh * 3)` triggers pre-scaling only when source exceeds 3x draw size. Aliasing can occur below 3x, especially with high-contrast sleeve borders. The 3x cutoff is undocumented.

Reference: forge-gui-desktop/src/main/java/forge/toolbox/FLabel.java:608

Test case:
```java
// Source 250x350, draw 80x100 => ratio 3.125 triggers, ratio 2.9 would not
// Both may alias
```
Impact: Inconsistent visual quality across sleeve sizes.

#### 3. img.getWidth(null) can return -1 for unloaded images
Description:
In setIcon(), iw and ih are obtained via `img.getWidth(null)` and `img.getHeight(null)`. For an image that hasn't been fully loaded, these can return -1. The paintContent check `sw > 0 && sh > 0 && (iw > sw * 3 || ih > sh * 3)` would then compare -1 > positive, which is false, so no pre-scale. However later `g.drawImage(src, ... 0, 0, srcW, srcH, null)` with srcW = -1 may cause rendering exception.

Reference: forge-gui-desktop/src/main/java/forge/toolbox/FLabel.java:500, 608, 621

Test case:
```java
Image img = new URL("http://example.com/missing.png").openStream() ... // fails to load
label.setIcon(new ImageIcon(img));
label.paintComponent(...); // img width -1
```
Impact: Potential NullPointer or IllegalArgumentException during paint.

#### 4. scaledBg cache grows without bounds on resize
Description:
The cache condition `if (scaledBg == null || scaledBgW < tw || scaledBgH < th)` only grows the cache. If label is resized larger then smaller, the oversized cached image is retained forever for that FLabel instance. Since FLabel is reused for lobby slots, memory can accumulate.

Reference: forge-gui-desktop/src/main/java/forge/toolbox/FLabel.java:610

Test case:
```java
FLabel label = ...; // initial size 58x80 -> cache 116x160
label.setSize(200,300); // tw=400,th=600 -> new cache
label.setSize(58,80); // cache remains 400x600, never shrinks
```
Impact: Memory leak for long-running sessions with dynamic resizing.

#### 5. scaledBg not cleared when icon changes to smaller image
Description:
setIcon() sets `scaledBg = null` which is good. However if setIcon is called with same Image instance but different scaling due to resize, scaledBgW/H may be larger than needed and not regenerated until size increases.

Reference: forge-gui-desktop/src/main/java/forge/toolbox/FLabel.java:497

Test case:
```java
label.setIcon(bigIcon);
label.setSize(200,300); // creates 400x600 cache
label.setIcon(bigIcon); // same reference, scaledBg cleared
label.setSize(58,80); // creates 116x160 cache - ok
```
Impact: Acceptable due to nulling, but repeated setIcon with same icon may cause unnecessary recreation.

#### 6. Removing PlayerPanel pre-scale workaround changes visual output
Description:
Commit ed8911d42c3 removed the PlayerPanel.setSleeveIcon() helper that pre-scaled via SkinImage.resize(58*2,80*2). The PR now relies solely on FLabel pre-scale. The old workaround used SkinImage.resize which may use different interpolation than Image.SCALE_SMOOTH. Visual difference may be noticeable.

Reference: forge-gui-desktop/src/main/java/forge/screens/home/PlayerPanel.java (removed lines)

Test case:
```java
// Compare rendered sleeve before/after
// Old: SkinImage.resize -> ?
// New: Image.SCALE_SMOOTH + BICUBIC
```
Impact: Backwards compatibility of visual appearance changed; may be intentional but undocumented.

#### 7. iconScaleFactor interacts with pre-scale calculation
Description:
sw is computed as `(int)(sh * iar)` where sh = (int)(h * iconScaleFactor). With iconScaleFactor 0.99, sw/sh are slightly smaller. The pre-scale threshold uses sw*3, so slight changes in iconScaleFactor shift threshold. If iconScaleFactor changes at runtime, cached scaledBg may be too small for new draw size.

Reference: forge-gui-desktop/src/main/java/forge/toolbox/FLabel.java:596-597, 608

Test case:
```java
label.setIcon(...);
label.setSize(100,100);
// cache created for sw~99
label.setIconScaleFactor(0.5f); // sw changes, cache may be too large, still used
```
Impact: Potential blur if cache oversized.

#### 8. scaledBg shared across icon changes if same instance reused
Description:
scaledBg is an instance field of FLabel. If setIcon(null) is called, img becomes null but scaledBg is cleared in setIcon only when iconInBackground||iconScaleAuto. If icon is set to null, img=null, but scaledBg is cleared. Good. However if setIcon is called with a new icon but same Image object reference, scaledBg is nulled, so OK.

Reference: forge-gui-desktop/src/main/java/forge/toolbox/FLabel.java:497

Test case:
```java
Image img = ...;
label.setIcon(new ImageIcon(img));
label.setIcon(new ImageIcon(img)); // scaledBg cleared, recreated
```
Impact: Minor performance cost.

#### 9. No synchronization for scaledBg creation
Description:
paintContent can be called on EDT only, but if resetIcon() triggers on resize timer, it may be on EDT. However scaledBg creation via getScaledInstance may be expensive and block EDT. No guard against concurrent creation.

Reference: forge-gui-desktop/src/main/java/forge/toolbox/FLabel.java:611

Test case:
```java
// Rapid resize triggers multiple paint calls
// Each may create scaledBg concurrently? EDT serializes, but still blocks UI
```
Impact: UI stutter on first paint of large icon.

### Performance and security
#### 10. Image.SCALE_SMOOTH can be slow for large images
Description:
Creating scaled instance via getScaledInstance with SCALE_SMOOTH may be synchronous and CPU heavy for 360x500 -> 116x160. This happens on EDT during paint, potentially causing frame drops.

Reference: forge-gui-desktop/src/main/java/forge/toolbox/FLabel.java:611

Test case:
```java
Image scaled = new ImageIcon(big).getImage().getScaledInstance(...);
```
Impact: Performance regression on first paint.

#### 11. No disposal of old scaledBg
Description:
When scaledBg is replaced, old Image is not disposed. Java Image may be GC'd, but native resources may linger.

Reference: forge-gui-desktop/src/main/java/forge/toolbox/FLabel.java:611

Test case:
```java
// Repeated icon changes create many Images without disposal
```
Impact: Potential native memory leak.

## Verdict
Request Changes
The pre-scale approach improves visual quality but introduces several issues: the comment claims area-averaging but code uses SCALE_SMOOTH, cache grows without bounds, unloaded images may cause negative dimensions, and pre-scaling occurs on EDT without background loading. The 3x threshold is arbitrary and may miss aliasing. Memory usage and EDT blocking need mitigation.

## Recommended Fixes
1. Replace Image.SCALE_SMOOTH with explicit area-averaging via Graphics2D with RenderingHints.VALUE_INTERPOLATION_BICUBIC or use BufferedImage scaling with explicit algorithm.
2. Add null/negative dimension checks before using iw/ih and srcW/srcH; guard against unloaded images.
3. Implement cache size cap or allow shrinking cache when label size decreases significantly.
4. Move scaled image creation off EDT or pre-create on setIcon.
5. Log warning when iw/ih negative.
6. Consider making threshold configurable and documented.
7. Dispose previous scaledBg before replacing.

---

MostCromulent:
Each suggestion checked against the branch. References are to `forge-gui-desktop/src/main/java/forge/toolbox/FLabel.java` at the head of this PR.

|---|---|---|
| 1 | Replace `SCALE_SMOOTH` with explicit area-averaging | Already area-averaging — `Image.getScaledInstance` maps `SCALE_SMOOTH` to `AreaAveragingScaleFilter` (`java.desktop/java/awt/Image.java`), and `BufferedImage` does not override it. The suggested bicubic pass is the single-pass downscale this PR avoids, so it would revert the change. |
| 2 | Guard null / negative `iw`, `ih` | Unreachable. With `iw = -1` the threshold `iw > sw * 3` is false and the code falls through to the `drawImage` call master already makes. A null icon sets `iar = 0` (line 508), so `sw` is 0 and the `sw > 0` guard short-circuits before `img` is dereferenced. |
| 3 | Cap or shrink the cache | Correct that it only grows, but that is deliberate: exact matching would rebuild at every intermediate width during a drag-resize. Not reachable today — every `iconInBackground` label has a fixed pixel size, including the sleeve at `"width 58px, height 80px"` (`PlayerPanel.java:141`) and the overlay logo (`SOverlayUtils.java:56`). A `scaledBgW > tw * 2` band is the form to add if a growable constraint ever appears. |
| 4 | Scale off the EDT or in `setIcon` | `setIcon` also runs on the EDT, so this relocates the cost and pays it for labels never painted. The pass covers ~180k source pixels once per size, not per paint. |
| 5 | Warn on negative dimensions | Pre-existing path, and the pre-scale branch is skipped there, so this PR adds no failure mode to report. |
| 6 | Make the threshold configurable | Below 3x the single bicubic pass is the intended output, not a fallback. The constant marks where two-pass scaling starts to pay; the inline comment gives the reason. |
| 7 | Dispose the previous `scaledBg` | `java.awt.Image` has no `dispose()`. The nearest call, `flush()`, discards the accelerated copy Java2D may cache but explicitly leaves an offscreen image's pixels untouched, so it would not release the raster this is aimed at. Any accelerated surface belongs to the image's `SurfaceManager` and is freed through `sun.java2d.Disposer` once the image becomes unreachable — which reassigning the field makes it. |

No changes applied.

---

PR #10954, opened 2026-06-11
Teamwork cost

### LINKED ISSUE #10855: MSH: Teamwork
> Teamwork N (As an additional cost to cast this spell, you may tap any number of creatures you control with total power N or more.)

Optional Additional Cost

on the spoilered cards, it works similar to Entwine
> Choose one. If this spell was cast using teamwork, choose both instead.

But because of Agent Maria Hill, it needs to be an extra Cost like Waterbend
> Whenever Agent Maria Hill becomes tapped to pay a teamwork cost

[Virtual Assistant](https://www.magicspoiler.com/mtg-spoiler/virtual-assistant/) has:
> Whenever you cast a spell using teamwork, ...

which should be easy if marked as optional cost

tehdiplomat:
Added basic Teamwork Optional Cost and two cards as examples.

Didn't add
- Trigger when casting a spell with a Teamwork cost (Virtual Assistant)
- Trigger when using this creature to pay for a Teamwork Cost (Agent Maria Hill)

### DIFF
--- a/forge-game/src/main/java/forge/game/GameActionUtil.java
+++ b/forge-game/src/main/java/forge/game/GameActionUtil.java
@@ -472,7 +472,6 @@ public final class GameActionUtil {
                 costs.add(new OptionalCostValue(OptionalCost.Entwine, cost));
             } else if (keyword.startsWith("Teamwork")) {
                 String[] k = keyword.split(":");
-                // tapXType has a special check for withTotalPower, and NEEDS it to be "+withTotalPowerGE"
                 String costString = "Teamwork<" + k[1] + ">";
                 final Cost cost = new Cost(costString, false);
                 costs.add(new OptionalCostValue(OptionalCost.Teamwork, cost));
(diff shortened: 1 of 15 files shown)


Hanmac:
I already got an idea how to handle the Maria Hill one

Using Cost Payment Stack

tool4ever (dismissed):

Hanmac:
I can't test it right now, but `Agent Maria Hill` should be codable using new `Teamwork` flag

forge-game/src/main/java/forge/game/GameActionUtil.java
```diff
@@ -470,6 +470,12 @@ public static List<OptionalCostValue> getOptionalCostValues(final SpellAbility s
                 String[] k = keyword.split(":");
                 final Cost cost = new Cost(k[1], false);
                 costs.add(new OptionalCostValue(OptionalCost.Entwine, cost));
+            } else if (keyword.startsWith("Teamwork")) {
+                String[] k = keyword.split(":");
+                // tapXType has a special check for withTotalPower, and NEEDS it to be "+withTotalPowerGE"
```
tehdiplomat: This comment doesn't make sense anymore since we created a new Cost type

tool4ever (approved):

PR #10961, opened 2026-06-11
MSC: add TriggerConnives

### LINKED ISSUE #10960: MSC: Connive Trigger
Some cards have Connive trigger:
`Whenever a creature you control connives`

Hanmac:
Closes #10960

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/ConniveEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/ConniveEffect.java
@@ -13,6 +13,7 @@ import forge.game.player.PlayerCollection;
 import forge.game.replacement.ReplacementResult;
 import forge.game.replacement.ReplacementType;
 import forge.game.spellability.SpellAbility;
+import forge.game.trigger.TriggerType;
 import forge.game.zone.ZoneType;
 import forge.util.Lang;
 import forge.util.Localizer;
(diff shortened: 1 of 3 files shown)


tool4ever (dismissed):

forge-game/src/main/java/forge/game/ability/effects/ConniveEffect.java
tool4ever: it still needs to trigger

forge-game/src/main/java/forge/game/ability/effects/ConniveEffect.java
```diff
@@ -78,28 +79,28 @@ public void resolve(SpellAbility sa) {
                 game.getTriggerHandler().collectTriggerForWaiting();
 
                 CardCollection hand = new CardCollection(p.getCardsIn(ZoneType.Hand));
-                if (hand.isEmpty() || !p.canDiscardBy(sa, true)) {
-                    continue;
+                if (!hand.isEmpty() && p.canDiscardBy(sa, true)) {
```
tool4ever, line 83: looks like this was never fully correct:
we actually need to run another `triggerChangesZoneAll` here first and then refresh `zoneMovements`

Hanmac: Hm, I don't exaclty understand what you mean?
`final CardZoneTable zoneMovements = AbilityKey.addCardZoneTableParams(moveParams, sa);`
after the draw cards?

Feel free to edit it like you think

tool4ever: it is draw, then discard - so two different boardstates
I can add it later

Hanmac: like this?

tool4ever (approved):

PR #10967, opened 2026-06-11
MSC: Typhoid Mary, Fractured (support likely needed)

dracontes:
Taking the relevant Discord discussion into account, I took it that I should hew closely to existing practice for `Charm` effects for proper display instead of using `Branch`. As `CheckSVar`s aren't used for these lines (and I tried to no avail) the best tack I came across was use `Count$Compare` for `CharmNum$` to limit selection for each of two concatenated `Charm` lines. 

The current outcome is that the first line works fine if I haven't discarded a card, but if I do discard a card, the trigger goes on the stack with no effect. I take it that means the `Count$Compare` is working in limiting selection for the first line but somehow the second `Charm` line is being ignored.

FWIW, the only other card to have a functional `SubAbility$` associated with a `Charm` line is [Bind to Secrecy](https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/b/bind_to_secrecy.txt).

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/typhoid_mary_fractured.txt
+++ b/forge-gui/res/cardsfolder/upcoming/typhoid_mary_fractured.txt
@@ -1,0 +1,13 @@
+Name:Typhoid Mary, Fractured
+ManaCost:1 B R
+Types:Legendary Creature Mutant Villain
+PT:3/3
+T:Mode$ Attacks | ValidCard$ Card.Self | Execute$ TrigCharm | TriggerZones$ Battlefield | TriggerDescription$ Whenever NICKNAME attacks, ABILITY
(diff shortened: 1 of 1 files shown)


tool4ever:
the easiest solution would probably be if AtRandom supported Count SVars too 🤔

dracontes:
> the easiest solution would probably be if AtRandom supported Count SVars too 🤔

FWIW, considered the situation a bit further: splitting the trigger condition into two upstream lines leading into their corresponding `Charm` effects while labeling one trigger condition as `Secondary$` does seem to work as expected. I'm not sure how formally undesirable that tack is, though. (New script appended for convenience.)

[typhoid_mary_fractured.txt](https://github.com/user-attachments/files/28853322/typhoid_mary_fractured.txt)

Hanmac:
@tool4ever should we do Random as Variable Parameter?

tool4ever:
yea I meant something like that, I guess the syntax doesn't matter too much here

PR #10978, opened 2026-06-12
Fix #9973: AI never uses mutagen tokens

### LINKED ISSUE #9973: AI never cracks mutagen tokens
In sealed play, the CPU AI never chooses to activate the mutagen tokens . Not sure how easy this is to solve. 

Certainly in 1v1 game play there are times when you'd want to not crack mutagen tokens immediately (Like for Krang, Master Mind or Donatello, Turtle Techie), but the CPU seems to never crack them. 

It may be very difficult to get the CPU to determine whether it should wait (as in the aforementioned cases), but if that's not possible, it still seems like cracking them when they can is better than never using them at all.

Windows, official release 2.0.11

ItsameGui:
Even if the AI has a sure fire way to win, it'd
never use the mutagen tokens to give itself a
better position.
Now, the AI recognizes that it can indeed use the
tokens for its' advantage and win.


forge-gui-desktop/src/test/java/forge/ai/AITest.java
```diff
@@ -65,8 +65,16 @@ protected Game initAndCreateGame() {
         return game;
     }
 
-    protected Game initAndCreateThreePlayerGame() {
-        initAndCreateGame();
+    protected Game initAndCreateGame() {
+        if (!initialized) {
+            GuiBase.setInterface(new HeadlessGuiDesktop());
+            FModel.initialize(null, preferences -> {
```
tool4ever, line 71: what's going on here...?

PR #10979, opened 2026-06-12
Group Tokens battlefield mode and smart large number formatting (k/M/B)

Zastinian:
This PR improves the battlefield organization and visual clarity, especially during late-game scenarios involving large amounts of tokens.
It introduces a new grouping mode and a professional formatting system for card stacks.


jamincollins:
## Findings
#### 1. Group Tokens mode groups tokens only, but compatibility check includes tapped/damage
The PlayArea.collectAllTokens method uses groupTokens flag to decide whether to require tapped/damage match. When group_tokens mode is active, groupTokens is true, so tokens are grouped only if tapped and damage match. This is correct for tokens as per MTG rules - tokens with different damage should not group. However, the flag also activates for group_creatures and group_all modes, which is intentional. The issue is the updateGroupScope logic sets groupTokens true for group_creatures and group_all, causing tokens to be grouped with strict tapped/damage matching even in those modes. The original code had groupTokensAndCreatures for creatures, but groupTokens now includes creatures in group_creatures mode via PlayArea.collectAllCreatures which uses groupTokensAndCreatures. This seems consistent.

Test case:
```java
// Verify group_tokens mode only groups tokens
PlayArea area = new PlayArea(...);
FModel.getPreferences().setPref(FPref.UI_GROUP_PERMANENTS, "group_tokens");
area.updateGroupScope();
// Should group tokens but not creatures
assert area.isGroupTokens() == true;
assert area.isGroupTokensAndCreatures() == false;
```

Impact: Minor confusion about naming - groupTokens flag is true for all grouping modes, not just tokens.

#### 2. formatGroupCount does not handle negative counts
formatGroupCount in CardPanel and CardRenderer assumes count >= 0. Negative count would produce negative formatted string like "-1k". Group count should never be negative, but no guard exists.

Test case:
```java
String result = CardPanel.formatGroupCount(-1);
// Returns "-1" which is nonsensical
```

Impact: UI could display negative counts if bug exists elsewhere.

#### 3. formatLargeValue replaces ".0" globally, removing decimal from integer values
formatLargeValue uses replace(".0", "") which will also replace ".0" in the middle of a number string, potentially corrupting values like "10.05" -> "10.5". Also replace(".00", "") then replace(".0", "") order may cause issues.

Test case:
```java
String v = formatLargeValue(10.05); // Should be "10.05", but replace logic may produce "10.05"
// Actually formatLargeValue uses String.format which produces "10.05"
// replace(".0","") won't affect "10.05"
// But for 10.0 -> format "10.0" -> replace(".0","") -> "10"
// Good. For 100.0 -> "100" -> replace -> "100"
// For 10.10 -> "10.10" -> replace(".0","") -> "10.10" unchanged
// Potential issue with 1.01 -> "1.01" ok
```

Impact: Low risk but replace logic is fragile.

#### 4. formatGroupCount integer division loses precision
formatGroupCount divides count by 1000.0, 1000000.0, etc. For count=999, returns "999". For count=1000, returns formatLargeValue(1.0)+"k" -> "1k". Good. For count=1499, returns "1.5k". Good. Edge case at boundary 999999 -> formatLargeValue(999.999) -> val >=100 -> "1000k"? Wait 999999/1000=999.999, val>=100 -> format to "1000". Returns "1000k" which is 1M. Should be "1000k" is confusing but technically correct? Better to cap at 999.999k -> 1000k. Could be improved.

Test case:
```java
CardPanel.formatGroupCount(999999); // Returns "1000k" which is misleading
```

Impact: UX confusion at boundaries.

#### 5. groupCount badge hit detection assumes badge size
isBadgeHit in CardPanel uses hardcoded badgeWidth=cardWidth/3 and badgeHeight=cardHeight/6. Actual badge size in drawGroupCountBadge is computed dynamically based on text width. Mismatch causes hit detection to be inaccurate.

Test case:
```java
// Click on badge area with long formatted count like "1.5M"
// Actual badge width larger than assumed, hit detection misses
```

Impact: Users may not be able to click badge.

#### 6. Default preference changed to group_tokens in commit c3113ed7787 then reverted in 805dfe47c98
Commit c3113ed7787 sets UI_GROUP_PERMANENTS default to "group_tokens". Commit 805dfe47c98 immediately reverts to "default". This creates confusion and suggests the PR author changed default then realized it was wrong. The merge commit brings both changes, but final value is "default" in pr-10979 branch. However, the commit history shows a flip-flop that may cause users upgrading from master to pr-10979 to see default change unexpectedly if they had never set preference.

Test case:
```java
// New install with pr-10979 should get default, not group_tokens
// But commit history shows default was temporarily group_tokens
```

Impact: No functional issue now, but historical confusion.

#### 7. Mobile groupCount set to 0 for non-head panels
In VCardDisplayArea.addCards, groupCount is set to totalCount+1 for head panel, 0 for others. CardRenderer.drawCardWithOverlays likely expects groupCount>=1 to draw badge. Setting to 0 for non-head panels means badge not drawn. Good. But if stack is built incorrectly, head panel may get wrong count.

Impact: Correct behavior.

#### 8. groupTokens flag used for both tokens and creatures in group_creatures mode
PlayArea.updateGroupScope sets groupTokens=true for group_creatures mode. Then collectAllTokens uses groupTokens for tapped/damage check. That's fine. But collectAllCreatures uses groupTokensAndCreatures. The naming is confusing but functional.

Test case:
```java
// In group_creatures mode, tokens should group with tapped/damage match
// Creatures group with tapped/damage/text match
// Implementation seems correct
```

Impact: Maintainability issue.

#### 9. Mobile VField.isGroupingEnabled returns true for group_all regardless of card type
For group_all, returns true for all cards. This means all permanents are grouped on mobile. Correct.

#### 10. Desktop PlayArea grouping flag includes groupTokens which is true for all modes except default/stack
grouping = groupTokens || groupTokensAndCreatures || groupAll simplifies to groupTokens || groupTokensAndCreatures || groupAll. Since groupTokens includes group_creatures and group_all, grouping is essentially true for group_tokens, group_creatures, group_all. Good.

### Performance and Security
#### 11. formatGroupCount and formatLargeValue called per card paint
These are static methods called during paint for each card panel with groupCount>=2. For large boards with many tokens, this could be called many times per frame. String formatting is relatively cheap but could be optimized with caching.

Test case:
```java
// 1000 tokens, each paint triggers formatGroupCount -> 1000 String.format calls per frame
```

Impact: Minor performance cost, likely acceptable.

#### 12. No input validation on preference values
UI_GROUP_PERMANENTS preference can be set to any string via config. updateGroupScope compares with equals, so invalid value falls through to all flags false. Safe.

Impact: None.

## Verdict
Request Changes

The PR introduces Group Tokens battlefield mode and large number formatting, but has several edge case issues: formatLargeValue replace logic is fragile, formatGroupCount boundary values produce confusing output like "1000k", badge hit detection size mismatch, and groupCount can be negative. The default preference flip-flop is confusing. The core functionality appears sound but needs hardening.

## Recommended Fixes
1. Add guard for negative counts in formatGroupCount and return "0" or String.valueOf(count) for negative
2. Fix formatLargeValue to use DecimalFormat or proper trimming instead of replace(".0","")
3. Adjust isBadgeHit to compute badge size same way as drawGroupCountBadge
4. Cap formatGroupCount to avoid "1000k" -> convert to "1M" when >=1000k
5. Add unit tests for formatGroupCount boundaries
6. Document groupTokens flag naming confusion
7. Consider caching formatted group counts per panel to avoid repeated String.format

---

Issue #8268, opened 2025-07-31
Support switching AI profile via Deck file


tool4ever:
Deck authors should be able to define the profile that best fits the archetype.

Maybe even be able to override specific settings.

I imagine that would also be great for enemy flavor in adventure:
e. g. writing some suicidal boss that loves to drop boardwipes whenever he can :)

PR #10982, opened 2026-06-13
Support deck-specified AI profiles

Rauglothgor:
﻿## Problem

Deck authors currently cannot bind a deck to the AI profile that best fits that deck or scenario. This is a narrow slice of #8268.

## Minimal Fix

This adds a deck-file AI profile hint through the existing `AiHints` metadata field:

```text
AiHints=AiProfile$Reckless
```

`PlayerControllerAi.setupAutoProfile(Deck)` reads the existing `AiHints` entry with `getAiHint("AiProfile")` and applies a nonblank deck profile to AI lobby players. When a deck profile is applied, profile rotation is disabled for that AI player so the deck file remains deterministic.

The serializer test covers the deck-file metadata round trip and confirms an existing `AiHints` entry is preserved alongside `AiProfile`.

## Validation

Ran locally with temporary Maven/JDK tooling:

```text
mvn -pl forge-gui-desktop -am -Dtest=forge.deck.DeckAiProfileTest -Dsurefire.failIfNoSpecifiedTests=false test
```

Observed result:

- `DeckAiProfileTest`: `Tests run: 1, Failures: 0, Errors: 0, Skipped: 0`
- `forge-ai` compile: `BUILD SUCCESS`
- Maven checkstyle: `0 Checkstyle violations`

## AI Assistance Disclosure

Codex assisted with implementation and validation. I reviewed the diff and kept the scope limited to the deck-profile behavior requested in #8268.

## Known Limits

- No UI changes.
- No per-setting AI profile overrides.
- No AI heuristic changes.
- Profile names are passed through to the existing AI profile loading path.

Refs #8268.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
+++ b/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
@@ -54,6 +54,8 @@ import java.util.stream.Collectors;
  * Handles phase skips for now.
  */
 public class PlayerControllerAi extends PlayerController {
+    private static final String AI_PROFILE_HINT = "AiProfile";
+
     private final AiController brains;
 
     private boolean pilotsNonAggroDeck = false;
@@ -70,6 +72,11 @@ public class PlayerControllerAi extends PlayerController {
 
     public void setupAutoProfile(Deck deck) {
         pilotsNonAggroDeck = deck.getName().contains("Control") || deck.getAverageCMC() > 3;
+        final String deckAiProfile = deck.getAiHint(AI_PROFILE_HINT);
+        if (StringUtils.isNotBlank(deckAiProfile) && lobbyPlayer instanceof LobbyPlayerAi aiPlayer) {
+            aiPlayer.setRotateProfileEachGame(false);
+            aiPlayer.setAiProfile(deckAiProfile);
+        }
     }
 
     public void setUseSimulation(boolean value) {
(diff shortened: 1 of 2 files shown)


forge-core/src/main/java/forge/deck/Deck.java
```diff
@@ -648,6 +650,14 @@ public void setAiHint(String hintType, String hintValue) {
         aiHints.add(hintType + "$" + hintValue.trim());
     }
 
+    public String getAiProfile() {
+        return getAiHint(AI_PROFILE_HINT);
+    }
+
+    public void setAiProfile(String profileName) {
```
tool4ever: lets not add methods purely for tests since this is basically redundant to the Archetype part anyway

tool4ever:
Hmn, just realized `EnemyData` already supports profile selection :/
So this might need to be refactored together to avoid duplicate logic
But could be tricky since there's also the *.json deck variants...?
So maybe the deck files should actually be on a per property level instead to support a minor hierarchy of sorts 🤔

Rauglothgor:
Thanks, I missed the existing `EnemyData.ai` path.

I removed the extra `Deck` helper methods from this PR, but I agree the remaining change still risks creating a second profile-selection path instead of sharing a clearer hierarchy with adventure enemy JSON and deck variants. I am going to close this narrow PR rather than push a speculative refactor here.

If there is a preferred shape for deck-level AI overrides later, I can revisit it from that design instead of layering this on top.

PR #10989, opened 2026-06-13
Implement gauntlet options for booster draft

### LINKED ISSUE #10568: Booster Draft: Allow playing up to X rounds in a gauntlet

When you play cube/booster draft, your match options are currently:

 - Play a single opponent in a single round
 - Play multiple opponents (at once) in a single round
 - Play all opponents in a gauntlet

Since a draft pod is 8 players, playing all opponents means playing 7 rounds. But 7 rounds is a bit too excessive (length-wise) for a single booster draft. I'd like to be able to play 3-4 rounds, like an actual booster draft at an LGS.

Provide a 4th match option which is to play against single opponents in a gauntlet, but be able to choose the number of rounds (min: 1, max: pod size - 1)

There is none. It's either 1 round or 7 rounds, nothing in between (to my knowledge).

**Additional context**

ItsameGui:
- Add options dialog to set gauntlet rounds and match length
- Persist selected values between sessions
- Include initial support for multiple rounds and help dialogs

Closes Card-Forge#10568

Co-authored by @PedroCalhau <pedro.d.andrade@tecnico.ulisboa.pt>


forge-gui/res/languages/en-US.properties
```diff
@@ -3557,7 +3564,10 @@ lblLoadingGameResources=Loading game resources...
 lblBackupRestore=Backup and Restore
 lblBackupRestoreDescription=Backup or Restore Classic game mode data to/from Downloads folder
 lblExportLogs=Export Logs
-lblExportLogsDescription=Export forge.log files and network debug logs to Downloads folder as a zip
+lblExportLogsDescription=Export forge.log files an
+View PedroCalhau's full-sized avatar
+Pedro Andrade PedroCalhau
+d network debug logs to Downloads folder as a zip
```
MostCromulent: Seems like an editing error has slipped in here.

forge-gui/res/languages/en-US.properties
```diff
@@ -3574,27 +3584,3 @@ lblDataMigrationMsg=Data Migration completed!\nPlease check your Inventory and E
...
-lblCommanderBracketReasonGameChangersOne=1 to 3 Game Changers make the deck at least Bracket 3.
-lblCommanderBracketReasonMassLandDenial=Mass Land Denial makes the deck at least Bracket 4.
-lblCommanderBracketReasonChainedExtraTurn=Chained Extra Turn cards make the deck at least Bracket 4.
-lblCommanderBracketReasonExtraTurnsFour=4 or more Extra Turn cards make the deck at least Bracket 4.
-lblCommanderBracketReasonExtraTurnsThree=3 or more Extra Turn cards make the deck at least Bracket 3.
-lblCommanderBracketReasonExtraTurnsTwo=2 Extra Turn cards make the deck at least Bracket 2.
-lblCommanderBracketReasonExtraTurnsFew=Fewer than 2 Extra Turn cards do not raise the bracket.
-lblCommanderBracketReasonLateGameCombo=Late Game 2-card combos make the deck at least Bracket 3.
-lblCommanderBracketReasonEarlyGameCombo=Early Game 2-card combos make the deck at least Bracket 4.
-lblAutoSellVariantsCommander=Auto-sell variants of owned cards in Commander
```
MostCromulent, line 3600: Also editing error?

forge-gui/res/languages/en-US.properties
```diff
@@ -1617,7 +1625,6 @@ lblFileNotFound=File not found
 lblPutCardInWhichPlayerZone=Put card in {0} for which player?
 lblPutCardInWhichPlayerBattlefield=Put card on the battlefield for which player?
 lblPutCardInWhichPlayerPlayOrStack=Put card on the stack / in play for which player?
-lblPutTokenInWhichPlayerBattlefield=Put token on the battlefield for which player?
```
MostCromulent, line 1620: Error?

forge-gui-desktop/src/test/java/forge/screens/home/sanctioned/CSubmenuDraftTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import forge.deck.Deck;
+import forge.deck.DeckGroup;
+
+/**
+ * Unit tests for CSubmenuDraft private parsing logic.
+ */
```
MostCromulent: Seems pretty unlikely this needs a unit test - AI coding agents have a tendency to add these unnecessarily to try and prove their wiring works, we warn against including these [here](https://github.com/Card-Forge/forge/blob/master/CONTRIBUTING.md#using-ai-coding-agents).

ItsameGui:
I've made some changes taking into account what you mentioned, would it be possible to review the code again?

jamincollins:
# Review for PR #10989: Implement gauntlet options for booster draft

## Findings

#### 1. CGauntletOptionsDialog.parseBoundedInt silently swallows parse errors with title change only
Description: In CGauntletOptionsDialog.showDialog() the OK button parses rounds and games with parseBoundedInt. parseBoundedInt calls Integer.parseInt which throws NumberFormatException on non-numeric input. The catch block only updates dialog title to "Invalid number" but does not prevent result[0] from being left as null? Actually result[0] is only set after successful parse. The dialog remains open, but user receives no persistent feedback and can re-click OK repeatedly. The title change is transient and may be missed.

Test case:
```java
// Simulate user entering non-numeric rounds
String rawValue = "abc";
CGauntletOptionsDialog dialog = CGauntletOptionsDialog.SINGLETON_INSTANCE;
// parseBoundedInt will throw NumberFormatException, caught, title set to Invalid number
// No validation feedback in UI beyond title, dialog stays open
// Impact: poor UX, could cause confusion, especially for non-English locales
```
Impact: Low UX degradation, no functional break but violates feedback principle.

#### 2. Games per match parsed with bounds 1-9 but UI only offers 1/3/5
Description: VGauntletOptionsDialog initializes cbMatchLength with {"1","3","5"} but CGauntletOptionsDialog.parseBoundedInt validates games with minimum 1 maximum 9 fallback 3. The dialog allows any value via direct UI manipulation? Combo box restricts, but parseBoundedInt permits up to 9. If code path changes to allow free text, invalid values like 2,4,7 would be accepted though UI never shows them. Also CSubmenuDraft.initialize() only persists gamesPerMatch and reloads binder, but does not persist rounds from dialog.

Test case:
```java
CGauntletOptionsDialog.GauntletOptions opts = new CGauntletOptionsDialog.GauntletOptions(4, 7);
// parseBoundedInt accepts 7 because max=9
// No UI path to select 7, but programmatic use could persist 7
// Preferences UI_MATCHES_PER_GAME will store 7, causing mismatch with binder expectations
```
Impact: Inconsistent validation range, potential hidden state.

#### 3. Gauntlet rounds preference persisted but not read in dialog OK handler
Description: CGauntletOptionsDialog.showDialog() parses rounds and games, then persistDefaults(rounds, games). The rounds value is persisted, but CSubmenuDraft.initialize() only persists gamesPerMatch from dialog, not rounds. Actually dialog persists both. However startGauntlet uses getConfiguredGauntletRounds which reads UI_GAUNTLET_ROUNDS preference. The dialog OK handler does persist rounds via persistDefaults, so okay. But CSubmenuDraft.initialize() listener for btnGauntletOptions only updates UI_MATCHES_PER_GAME and binder, it does not update any rounds UI. Since rounds is only used for gauntlet launch, it's fine. But note dialog shows rounds field, persists it, but user never sees confirmation.

Test case:
```java
// User sets rounds=10 via dialog, persists to prefs
// Later launches gauntlet, getConfiguredGauntletRounds caps to maxRounds
// No feedback that rounds was capped
```
Impact: Silent capping may surprise user.

#### 4. startGauntlet rounds calculation mixes configured and duelType for limited gauntlet
Description: startGauntlet computes:
```java
final int rounds = view.isLimitedGauntletSelected()
    ? Math.min(parseGauntletRounds(duelType, opponentDecks), configuredRounds)
    : configuredRounds;
```
parseGauntletRounds throws IllegalStateException if rounds <1 or >maxRounds. configuredRounds is clamped to [1,maxRounds]. Math.min of parsed duelType rounds and configured rounds means user can select more rounds via combo than preference, but actual rounds limited to smaller of the two. If user prefers 8 rounds but combo offers 3, rounds becomes 3. If user prefers 2 rounds but combo offers 7, rounds becomes 2. This is intentional? Might be confusing.

Test case:
```java
// maxRounds=7, configuredRounds=4, duelType="7"
// rounds = Math.min(7,4) = 4
// User selected 7 but gets 4
```
Impact: Potential UX confusion about which value takes precedence.

#### 5. getConfiguredGauntletRounds returns Math.min(4, maxRounds) on NumberFormatException, but preference may be empty string
Description: getConfiguredGauntletRounds parses savedRounds via Integer.parseInt. If pref is empty string, NumberFormatException thrown, fallback Math.min(4, maxRounds). safeDefault in dialog uses fallback for empty. But preference may be corrupted to "0" or negative. parseBoundedInt validates on save, but existing prefs from older versions may contain invalid values. parseInt succeeds for "0", then Math.max(1, Math.min(0,max)) => 1. So negative would be caught by parseBoundedInt on save, but old data could have negative? Integer.parseInt succeeds for "-5". Math.max(1, Math.min(-5,max)) => 1. So safe.

Test case:
```java
FModel.getPreferences().setPref(FPref.UI_GAUNTLET_ROUNDS, "-5");
// getConfiguredGauntletRounds -> parsed -5 -> Math.max(1, Math.min(-5,max)) =>1
// No crash, but silent correction
```
Impact: Silent correction acceptable.

#### 6. fillOpponentComboBox does not clear previous items before adding new ones
Description: fillOpponentComboBox is called in update() and after deck selection. It uses combo.addItem() without combo.removeAllItems(). If user switches deck, combo accumulates items from previous deck. Also renderer placeholder logic checks value == null or empty, but combo may retain old items.

Test case:
```java
// Select deck A with 3 AI decks -> combo items: 1,2,3
// Select deck B with 7 AI decks -> combo items: 1,2,3,1,2,3,4,5,6,7
// Duplicate entries cause incorrect parsing
```
Impact: Functional bug, could lead to wrong opponent selection or NumberFormatException.

#### 7. parseGauntletRounds throws IllegalStateException for non-numeric duelType
Description: parseGauntletRounds does Integer.parseInt(duelType) without try/catch. If combo contains non-numeric item (e.g., "Gauntlet" for full gauntlet), and isLimitedGauntletSelected is true, duelType will be numeric. But if UI bug causes non-numeric, exception propagates to user.

Test case:
```java
String duelType = "Gauntlet";
int rounds = Integer.parseInt(duelType); // NumberFormatException
```
Impact: Unhandled exception possible.

#### 8. CGauntletOptionsDialog.showDialog creates new VGauntletOptionsDialog each time, no disposal on cancel?
Description: showDialog creates new view each invocation. On cancel, view.dispose() called, result[0]=null. On OK success, view.dispose() called. Good. However if user opens dialog repeatedly, multiple instances created, okay.

Impact: Minor memory churn.

#### 9. UI_GAUNTLET_ROUNDS default value set to "4" in ForgePreferences, but dialog default rounds is 4, MAX 99
Description: Preference default 4 matches dialog DEFAULT_ROUNDS 4. Good. However existing users with no preference will get 4 rounds. Acceptable.

Test case: New user launches draft gauntlet -> rounds=4.

#### 10. CSubmenuDraftTest.java removed in commit d4279ca6495
Description: Test file CSubmenuDraftTest.java was removed. Tests for parseGauntletRounds valid, too low, too high, non-numeric were present. Removal eliminates regression protection for parsing logic.

Impact: Loss of test coverage.

#### 11. Games per match preference updated via dialog but gamesInMatch binder loads immediately, may cause UI flicker
Description: CSubmenuDraft.initialize() listener persists UI_MATCHES_PER_GAME and calls view.getGamesInMatchBinder().load(). Binder loads from preferences. Works.

Impact: None.

#### 12. GauntletMini.launch receives rounds but games per match not passed
Description: startGauntlet calls FModel.getGauntletMini().launch(rounds, humanDeck.getDeck(), gameType). GauntletMini likely uses UI_MATCHES_PER_GAME preference internally. Not passed explicitly. If preference not updated before launch, rounds may use old games per match. Dialog updates preference before launch? Dialog is opened via button, user clicks OK, preference persisted, dialog closed, then user clicks Start. So preference updated before start. Acceptable.

Test case:
```java
// User opens options, sets games=5, OK persists, closes
// Clicks Start -> games per match uses persisted value
```
Impact: Works as designed.

### Performance and Security
#### 13. parseBoundedInt uses Integer.parseInt on user-supplied text without length limit
Description: JTextField allows arbitrarily long input. Integer.parseInt on 10k digits may cause CPU spike. Not a security issue in desktop app, but could cause DoS via UI.

Impact: Low.

## Verdict
Request Changes

The PR adds useful gauntlet options but introduces several UX and correctness issues: combo box not cleared on deck change, silent capping of rounds, inconsistent validation ranges, and loss of unit tests. The dialog feedback on invalid input is minimal and the rounds/games per match precedence is confusing.

## Recommended Fixes
1. Clear opponent combo box before populating in fillOpponentComboBox to avoid duplicate items.
2. Add explicit validation feedback in CGauntletOptionsDialog: disable OK until input valid, or show error dialog instead of title change.
3. Align games per match validation range with UI options (1/3/5) or extend UI to allow 1-9.
4. Restore CSubmenuDraftTest.java or add equivalent tests for parseGauntletRounds and getConfiguredGauntletRounds.
5. Ensure rounds preference is updated via dialog and provide user feedback when rounds is capped by max opponents.
6. Consider making rounds parsing in parseGauntletRounds safe with try/catch to avoid unhandled NumberFormatException.
7. Document precedence rules for limited gauntlet rounds vs configured rounds.

---

PR #10990, opened 2026-06-13
fix Viv Vision, Teen Synthezoid

dyegoaurelio:
The Cybernetic Senses trigger never drew a card, even when Viv's power was 4 or greater.


tool4ever (approved):
thanks

PR #10993, opened 2026-06-13
Teamwork properties

tehdiplomat:


Hanmac:
@tehdiplomat why do you make a new Instance, than use the function instead of using a static function like before?

tehdiplomat:
I hate all of our places that are just Static functions. It feels so sloppy.

Hanmac:
> I hate all of our places that are just Static functions. It feels so sloppy.

But that just creates new objects without good reason 🤔 

@tool4ever your opinion on static vs non static?

tool4ever:
yea spawning lots of new objects on each valid check doesn't feel right either

maybe just convert them into Interfaces? then each Entity class can access its property check as a real instance method

Hanmac (approved):

Issue #7583, opened 2025-05-16
Allow user to select which card displays as the face of a deck


Jetz72:
Both desktop and mobile have logic for determining which card should be shown on a deck's "box", when rendering decks in image view:

https://github.com/Card-Forge/forge/blob/349129f88f00c611e6b75b70aa9fce6065fa2861/forge-gui-desktop/src/main/java/forge/itemmanager/views/ImageView.java#L1176-L1179

https://github.com/Card-Forge/forge/blob/349129f88f00c611e6b75b70aa9fce6065fa2861/forge-gui-mobile/src/forge/itemmanager/views/ImageView.java#L1141-L1146

It should be possible for a user to be able to override this manually. 

A good first step would be consolidating those two bits of logic for picking a face card within `DeckProxy`. The chosen face card could easily be stored as a field in `Deck`, serialized as one of the parameters at the top of a deck file, and used to override the automatically selected face. Then we'd just need a way for the user to choose or clear the face card in each deck editor. The dropdown that appears when right-clicking a card would make a sensible place for this.

Additional considerations - if cards not in the deck should be usable as a deck's face; if there should be an additional option to set the back face of a DFC as the deck's face; if yes, will that extend to weirder alt-face types like specialize and meld?

Magius74:
I would like to +1 this request and also take it a step further and ask that a custom image be an option as well. I will point out that the Precons in Quest mode already allow this by simply entering an image file name in the [metadata] section under "Image=" line. Could we achieve the same functionality by allowing support for a [metadata] section in Constructed deck files as well??

PR #10994, opened 2026-06-14
Use any card's art as a deck sleeve

MostCromulent:
Screenshot 2026-06-14 094055

Screenshot 2026-06-14 094120

Screenshot 2026-06-14 112138

## What it does

Select any card's art as a deck sleeve. You pick a card from a dialog; chosen art renders in the lobby and in-match exactly where the built-in sleeves do, styled to match them, on both desktop and mobile.

## How it works

A player's sleeve is stored as either a built-in sleeve index or a card image key, carried on `LobbyPlayer`/`Player`/`PlayerView` and exposed as a `TrackableProperty` so opponents see the choice; it travels over the wire in `UpdateLobbyPlayerEvent`.

Two preferences persist the choice across sessions: `UI_SLEEVE_ART_LIBRARY` holds the saved "Card Art Sleeves" library, and `UI_SLEEVE_ART_KEYS` holds the per-lobby-slot selection — sitting parallel to the existing `UI_SLEEVES` (built-in index per slot), so a slot using card art stores its key here while choosing a built-in sleeve clears it. Both are stored as comma-joined, URL-safe-base64-encoded image keys so card identifiers stay valid in the `KEY=VALUE` preference format.

The art itself comes from Scryfall's `art_crop`:

- When a card-art sleeve is needed, the existing Scryfall image fetcher is reused but **forced to the `art_crop` variant** (Scryfall's tightly-cropped artwork), regardless of the user's `UI_CARD_ART_FORMAT` preference. It's gated behind the online-image-fetcher preference and skips cards with no usable art_crop (custom cards, or those missing an artist / collector number).
- The download is cached under a dedicated sleeves cache (`CACHE_SLEEVE_PICS_DIR`), keyed by a hash of the image key with an `.artcrop.jpg` filename — chosen so the fetcher's Scryfall URL rewriting leaves it as the art_crop image.
- At render time the cached art_crop is centre-cropped to the built-in sleeve aspect and framed. The fetch is async — until the art lands, the built-in sleeve shows in its place.

## Safety

The only thing sent over the network is a **card image key** (a card identifier) — never a URL or an image file. Each client independently derives the Scryfall `art_crop` from that key through its own preference-gated fetch path, so a player cannot point opponents at an arbitrary URL or push image bytes to them; an unrecognised key simply falls back to the built-in sleeve.

---


MostCromulent:
Since this just uses a card identifier to fetch the relevant scryfall art, it opens up some fun design space for future development: e.g. Adventure Mode enemies could have an optional field to use a specific card's art as their sleeve / player decks could designate sleeve art as metadata.

paulsnoops:
image

Would it be possible to be able to choose how the image is displayed on the sleeve?

MostCromulent:
> Would it be possible to be able to choose how the image is displayed on the sleeve?

@paulsnoops Now you can!

Screenshot 2026-06-17 070250

Agetian:
Looks very nice! :)

tool4ever (commented):
@MostCromulent
did you consider #7583? (technically not identical but sort of the same workflow)
storing it per deck would give greater flexibility while at the same time avoiding more extra changes for netplay 🤔

tool4ever:
Hmn yea I think this does neatly combine both variants so you can have one global selection and your pet decks are individualized on their own like in paper :+1:

tool4ever:
@MostCromulent
doesn't work for me on desktop, preview is always empty and then it doesn't change:
grafik

maybe it's because I have the auto downloader turned off, though then it's still weird it lets me select normally 🤔

tool4ever (approved):

Issue #2174, opened 2022-12-27
StaticAbilityNumLoyaltyAct: corner case leads to illegal maximum raise


tool4ever:
Activations allowed by _The Chain Veil_ need to be used up somehow for correct interaction with _Oath of Teferi_:

- If you have a PW affected by both it results in a total of 3

- However if you activate your PW twice via the Veil's effect before you play the Oath the maximum used for checking should stay at 2 (but currently you will get a third out of it)

PR #10997, opened 2026-06-14
MSH: refactor additional activations

### LINKED ISSUE #10923: MSH: Power-Up activated an additional time
Similar to [Elvish Refueler](https://scryfall.com/card/dft/161/elvish-refueler)

>During your turn, as long as you haven’t activated an exhaust ability this turn, you may activate exhaust abilities as though they haven’t been activated.

https://github.com/Card-Forge/forge/blob/fffd2b7a08377010cf16e858094cd19942f94357/forge-game/src/main/java/forge/game/spellability/SpellAbilityRestriction.java#L513-L526

Wonder Man, Hollywood Hero:
> Each power-up ability of permanents you control can be activated an additional time.

But might be similar, but not enough? 
while doing that, should we touch the Boast special case too?

Hanmac:
Closes #10923


forge-gui/res/cardsfolder/c/comet_stellar_pup.txt
```diff
@@ -13,6 +13,6 @@ SVar:DBRemoveCounters:DB$ RemoveCounter | CounterType$ LOYALTY | CounterNum$ 2 |
 SVar:DBCleanup:DB$ Cleanup | ClearRemembered$ True
 SVar:CometGo:DB$ PutCounter | CounterType$ LOYALTY | SubAbility$ DBEffect | SpellDescription$ 6 — [+1], and you may activate CARDNAME's loyalty ability two more times this turn.
 SVar:DBEffect:DB$ Effect | StaticAbilities$ LoyaltyAbs | RememberObjects$ Self | ExileOnMoved$ Battlefield
-SVar:LoyaltyAbs:Mode$ NumLoyaltyAct | ValidCard$ Card.EffectSource | Additional$ 2 | OnlySourceAbs$ True | Description$ You may activate EFFECTSOURCE's loyalty ability two more times this turn.
+SVar:LoyaltyAbs:Mode$ Activations | ValidCard$ Card.EffectSource | ValidSA$ Activated.Loyalty | ValidPlayer$ You | Additional$ 2 | Description$ You may activate EFFECTSOURCE's loyalty ability two more times this turn.
```
tool4ever: Logic for `OnlySourceAbs$` missing:
> In some unusual cases, Comet may gain loyalty abilities other than the one he normally has. In those cases, the effect of rolling a 6 applies only to his normal ability.

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -354,9 +355,6 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
 
     private String overlayText = null;
 
-    private int planeswalkerAbilityActivated;
-    private boolean planeswalkerActivationLimitUsed;
```
tool4ever, line 358: this flag was needed since it's a bit more complicated
right now you'd allow this again: #2174

Hanmac: I can't find a rules for this explaining that the Chain Veil should be used up first

I asked the (previous?) Rules Manager again: 
https://bsky.app/profile/hanmac10.bsky.social/post/3moar765uec2y

tool4ever: Pretty sure I researched it back then, here's another source anyway:
https://www.reddit.com/r/magicTCG/comments/1bnq4rz/interaction_between_chain_veil_oath_of_teferi/

Hanmac: i saw other answers saying you get three activations:
https://magicjudge.tumblr.com/post/172667030966/how-would-the-new-oath-of-teferi-work-with-the

otherwise this ruling can't be true, because it would collide with itself
>  Each additional time The Chain Veil’s last ability resolves will allow you to activate a loyalty ability of each planeswalker you control an additional time. For example, if you activate The Chain Veil’s last ability, untap it, then activate it again, you can activate a loyalty ability of a planeswalker you control three times that turn.

tool4ever: no contradiction there:
you can get three - but only if Oath is around to count as second one

if it ETB after you've activated a Planeswalker twice it can't give you a third

forge-gui/res/cardsfolder/c/comet_stellar_pup.txt
Hanmac: ugh i can't add a new "EffectSource" to SpellAbilityProperty like this:
`sa.equals(source.getEffectSourceAbility().getRootAbility())`

because ActivatedAbilities are always copied on the Stack with new IDs, so the EffectSourceAbility (and its Root) will always be different

tool4ever: I assume that's why `getOriginalAbility()` was also used

forge-game/src/main/java/forge/game/card/Card.java
Hanmac: Also asked current rules Manager there:
https://bsky.app/profile/hanmac10.bsky.social/post/3mogib2na7c26

Hanmac: Reverted PW abilities "for now" until I get Rules confirmation.

tool4ever (approved):

forge-game/src/main/java/forge/game/card/Card.java
Hanmac: > no contradiction there:
> you can get three - but only if Oath is around to count as second one

i got this answer now:

>ringdingdingo 13:24:38
> the order doesn't matter

and the Wizards Rules Manager aren't answering me :(

Issue #10886, opened 2026-06-05
Refactoring counted Map<T, Integer> as Multiset<T>


Hanmac:
Guava Multiset already handles adding/removing Amount for us, so we should not need to do that ourselves.

Big uses of `, Integer>` i have seen and could be changed:
*  `ItemPool<T> implements Iterable<Entry<T, Integer>>` (and it uses `Map.Entry`)
* `Map<CounterType, Integer>` in GameEntity (and Subclasses) #11121
* `ActivationTable` is a `Table<SpellAbility, Optional<StaticAbility>, List<Player>>`, but the Value part should be a `Multiset<Player>` instead #10998

What probably shouldn't be changed:
* Damage noticed on a Card using `Map<Card, Integer>`

undecided:
* `TokenCreateTable extends ForwardingTable<Player, Card, Integer>`
in the end, the prototype is copied and giving a new ID each

PR #10998, opened 2026-06-14
ActivationTable: use Multiset for Player values

Hanmac:
Part of #10886

i noticed we doesn't need a "sorted" list for Activations by Player, so i used Multiset instead

### DIFF
--- a/forge-game/src/main/java/forge/game/card/ActivationTable.java
+++ b/forge-game/src/main/java/forge/game/card/ActivationTable.java
@@ -2,21 +2,21 @@ package forge.game.card;
 
 import com.google.common.collect.ForwardingTable;
 import com.google.common.collect.HashBasedTable;
-import com.google.common.collect.Lists;
+import com.google.common.collect.HashMultiset;
+import com.google.common.collect.Multiset;
 import com.google.common.collect.Table;
 import forge.game.player.Player;
 import forge.game.spellability.SpellAbility;
 import forge.game.staticability.StaticAbility;
 
-import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 
-public class ActivationTable extends ForwardingTable<SpellAbility, Optional<StaticAbility>, List<Player>> {
-    Table<SpellAbility, Optional<StaticAbility>, List<Player>> dataTable = HashBasedTable.create();
+public class ActivationTable extends ForwardingTable<SpellAbility, Optional<StaticAbility>, Multiset<Player>> {
+    Table<SpellAbility, Optional<StaticAbility>, Multiset<Player>> dataTable = HashBasedTable.create();
 
     @Override
-    protected Table<SpellAbility, Optional<StaticAbility>, List<Player>> delegate() {
+    protected Table<SpellAbility, Optional<StaticAbility>, Multiset<Player>> delegate() {
         return dataTable;
     }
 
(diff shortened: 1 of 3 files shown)


Agetian (approved):

forge-game/src/main/java/forge/game/card/ActivationTable.java
```diff
@@ -40,27 +40,24 @@ public void add(SpellAbility sa) {
         if (original != null) {
             Optional<StaticAbility> st = Optional.ofNullable(root.getGrantorStatic());
 
-            List<Player> activators = get(original, st);
-            if (activators == null) {
-                activators = Lists.newArrayList();
-            }
+            Multiset<Player> activators = Objects.requireNonNullElse(get(original, st), HashMultiset.create());
```
tool4ever, line 43: @Hanmac 
not sure such cleanup is worth it since you're now always creating extra empty Set...?
maybe writing as `Supplier` is less overhead 🤔

Hanmac: It was creating an empty List before 🤷

tool4ever: yes but only inside `activators == null`

now you create it *every* time because it's passed as method param...

I can make a small PR later

Hanmac: Or we could store Empty in a constant? (i didn't found one in guava)

tool4ever: it needs to be mutable new instance anyway

Hanmac: @tool4ever not at this part, this one can be immutable (because it is only read from the trigger part, not written)

tool4ever: in theory the whole `activators` storing could be avoided if `getOriginal` was only applied to the entries while accessing the data - but might be too roundabout?

Hanmac: I was wrong about activators there and mutable

the read-only part is below

just that guava Table doesn't have a "ComputeIfAbsent"

forge-game/src/main/java/forge/game/card/ActivationTable.java
```diff
@@ -40,27 +40,24 @@ public void add(SpellAbility sa) {
...
+    public Multiset<Player> getActivators(SpellAbility sa) {
         SpellAbility root = sa.getRootAbility();
         SpellAbility original = getOriginal(sa);
         Optional<StaticAbility> st = Optional.ofNullable(root.getGrantorStatic());
 
         if (contains(original, st)) {
             return get(original, st);
         }
-        return Lists.newArrayList();
+        return HashMultiset.create();
```
Hanmac, line 61: @tool4ever i meant this one

This could be a constant?

tool4ever: could try and only return a parent type
seems like it doesn't extend `Set` though

Hanmac: hmm `getAbilityResolvedThisTurnActivators` could return Collection instead?

Then use [Iterables.frequency](https://github.com/google/guava/blob/cb0bfa6f4f801941bf3649d1ad497c85d1dd0cf5/guava/src/com/google/common/collect/Iterables.java#L354) instead of `Collections.frequency`
(it has a check for Multiset and Set)

PR #11003, opened 2026-06-14
Add missing doXMath on some variables

Eradev:
Add missing doXMath on some variables so we can do stuff like `/LimitMax.1`.

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -2717,7 +2717,7 @@ public class AbilityUtils {
         }
 
         if (sq[0].startsWith("PlanarDiceSpecialActionThisTurn")) {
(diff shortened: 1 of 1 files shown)


Hanmac (approved):

tool4ever:
AbilityUtils alone currently has an "amazing" 228 `doXMath` calls...
Refactoring that number down might be a better fix instead instead of making that situation worse for a custom card 🤷‍♂️

Jetz72:
Would it work to just shift the logic of the `xCount` method into a private one and turn the original into a wrapper that always applies `doXCount`?

Jetz72:
I'd vote for going ahead with tweaks like this unless a `doXMath` cleanup/refactor is near the top of someone's to-do list. As long as it makes the script engine more consistent, it'll arguably make that kind of refactor more straightforward when we do get to it.

Agetian (approved):

PR #11006, opened 2026-06-14
Fix multiplayer crash when a remote client controls an opponent

MostCromulent:
## Summary

Control-an-opponent effects (e.g. Secret of Bloodbending) crash the game in multiplayer when the controlling player is a remote client. Reproduction: in a network game, the remote client casts Secret of Bloodbending targeting the host; the game thread dies with `IllegalArgumentException` at `AbstractGuiGame.setCurrentPlayer` as soon as control takes effect on the host's turn.

Root cause: the controlled player's priority runs through the controller's shared `InputProxy`, so `InputProxy.update` calls `setCurrentPlayer` on the controller's GUI. On the host, that GUI's `FControlGameEventHandler` registers the controlled player as local when `GameEventPlayerControl` fires, so the call succeeds. The server-side `RemoteClientGuiGame` instead uses a `GameEventForwarder` that only relays events to the client and never updates its own `gameControllers` map — so the controlled player is never registered server-side and `setCurrentPlayer` throws.

The fix mirrors the host-side bookkeeping in `RemoteClientGuiGame`: on each `GameEventPlayerControl`, register or unregister the controlled player in the proxy's `gameControllers`, identifying the controller via the controlled player's mind-slave master and `isLocalPlayer` (not by name) so only the actual controller's proxy registers the player.

---

### DIFF
--- a/forge-gui/src/main/java/forge/gamemodes/net/server/RemoteClientGuiGame.java
+++ b/forge-gui/src/main/java/forge/gamemodes/net/server/RemoteClientGuiGame.java
@@ -5,6 +5,7 @@ import forge.ai.GameState;
 import forge.deck.CardPool;
 import forge.game.GameEntityView;
 import forge.game.event.GameEvent;
+import forge.game.event.GameEventPlayerControl;
 import forge.game.GameView;
 import forge.game.card.CardView;
 import forge.game.phase.PhaseType;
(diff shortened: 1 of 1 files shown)


tool4ever (approved):
thanks, logic looks valid

but I found more ugly mobile port branching related to the exception detection: f239755

PR #11007, opened 2026-06-15
Card-script param declarations (batch 1: framework classes)

MostCromulent:
Batch 1 of the per-class card-script parameter declarations proposed in https://github.com/Card-Forge/forge/pull/10984#issuecomment-4703234427.

## Summary

This batch covers the framework/base classes rather than individual effects. These classes hold the parameters shared across nearly every ability — costs, conditions, activation/target restrictions, descriptions, `Defined`, sub-ability keys, and AI hints — so their params are all optional and together form the base set that every effect inherits. Effect classes (later batches) will declare their own params on top of this base and additionally mark which are required.

Each class now exposes a `public static final String[] OPTIONAL_PARAMS` listing the params it reads, making the parameters a class consumes self-documenting.

## How the params were chosen

The list for each class was produced programmatically, not hand-picked: a regex scanner reads each class's source for the literal forms a param read takes — `getParam`/`hasParam`/`getParamOrDefault` calls, the defined-resolution helpers (`getCards`, `getDefinedPlayersOrTargeted`, …), map lookups, and positional helpers like `addToCombat` — and each class declares exactly that set. The same matcher backs the drift test, so the lists are mechanically reproducible rather than a one-off snapshot.

Claude then adversarially reviewed the scanner's rules and its output against the source: this is what caught false positives — e.g. defined-*values* like `"Self"` and `"Targeted"` being mistaken for param names — and tightened the rules. `SpellAbilityAi` is the one hand-curated list: as a cross-cutting reader it touches params owned by many effects, so it declares only its own `AI*` knobs.

## The drift test

`CardScriptParamDeclarationTest` fails if a class reads a param it doesn't declare, or declares one that appears nowhere in its source — so the declarations can't silently fall out of sync with the code. It is opt-in per class, so later batches can land a few classes at a time.

---

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/AbilityFactory.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityFactory.java
@@ -45,7 +45,7 @@ import java.util.stream.Collectors;
  * @author Forge
  * @version $Id$
  */
-public final class AbilityFactory {
+public final class AbilityFactory implements IHasForgeParams {
     public static final String[] OPTIONAL_PARAMS = {
         "BidSubAbility", "CantChooseSubAbility", "Choices", "ChooseNumberSubAbility",
         "ChooseSubAbility", "ChosenPile", "Cost", "Execute", "FallbackAbility",
--- a/forge-game/src/main/java/forge/game/ability/AbilityFactory.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityFactory.java
@@ -46,6 +46,17 @@ import java.util.stream.Collectors;
  * @version $Id$
  */
 public final class AbilityFactory {
+    public static final String[] OPTIONAL_PARAMS = {
+        "BidSubAbility", "CantChooseSubAbility", "Choices", "ChooseNumberSubAbility",
+        "ChooseSubAbility", "ChosenPile", "Cost", "Execute", "FallbackAbility",
+        "FalseSubAbility", "GiftAbility", "GuessCorrect", "GuessWrong", "HeadsSubAbility",
+        "Highest", "LoseSubAbility", "Lowest", "MatchedAbility", "NonBasicSpell",
+        "NotLowest", "Origin", "OtherwiseSubAbility", "PreventionSubAbility",
+        "RegenerationAbility", "RepeatSubAbility", "ResultSubAbilities", "ReturnAbility",
+        "SpellDescription", "SubAbility", "TailsSubAbility", "TrueSubAbility",
+        "UnchosenPile", "UnmatchedAbility", "ValidTgts", "VoteSubAbility",
+        "VoteTiedAbility", "WinSubAbility",
+    };
 
     public static final List<String> additionalAbilityKeys = Lists.newArrayList(
             "WinSubAbility", "OtherwiseSubAbility", // Clash
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -50,6 +50,13 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 public class AbilityUtils {
+    public static final String[] OPTIONAL_PARAMS = {
+        "AbilityCount", "AnnounceMax", "Destination", "ETB", "ForgetOtherTargets",
+        "IncludeAllComponentCards", "LockInText", "RememberCostMana", "RememberTargets",
+        "Triggered", "UnlessColor", "UnlessCost", "UnlessPayer", "UnlessResolveSubs",
+        "UnlessSwitched", "UnlessUpTo", "XMax", "XMin",
+    };
+
     private final static ImmutableList<String> cmpList = ImmutableList.of("LT", "LE", "EQ", "GE", "GT", "NE");
 
     // should the three getDefined functions be merged into one? Or better to
--- a/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
+++ b/forge-game/src/main/java/forge/game/ability/AbilityUtils.java
@@ -49,7 +49,7 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
-public class AbilityUtils {
+public class AbilityUtils implements IHasForgeParams {
     public static final String[] OPTIONAL_PARAMS = {
         "AbilityCount", "AnnounceMax", "Destination", "ETB", "ForgetOtherTargets",
         "IncludeAllComponentCards", "LockInText", "RememberCostMana", "RememberTargets",
--- a/forge-game/src/main/java/forge/game/cost/Cost.java
+++ b/forge-game/src/main/java/forge/game/cost/Cost.java
@@ -21,6 +21,7 @@ import com.google.common.collect.Lists;
 import forge.card.CardType;
 import forge.card.mana.ManaCost;
 import forge.game.CardTraitBase;
+import forge.game.ability.IHasForgeParams;
 import forge.game.card.Card;
 import forge.game.card.CounterEnumType;
 import forge.game.card.CounterType;
@@ -46,7 +47,7 @@ import java.util.List;
  * @author Forge
  * @version $Id$
  */
-public class Cost implements Serializable {
+public class Cost implements Serializable, IHasForgeParams {
     public static final String[] OPTIONAL_PARAMS = {
         "AffectedZone", "Amount", "Announce", "Collected", "CollectedCards", "Color",
         "Cost", "Exiled", "ExiledCards", "FirstForetell", "ForEachShard", "Foraged",
--- a/forge-game/src/main/java/forge/game/cost/Cost.java
+++ b/forge-game/src/main/java/forge/game/cost/Cost.java
@@ -47,6 +47,15 @@ import java.util.List;
  * @version $Id$
  */
 public class Cost implements Serializable {
+    public static final String[] OPTIONAL_PARAMS = {
+        "AffectedZone", "Amount", "Announce", "Collected", "CollectedCards", "Color",
+        "Cost", "Exiled", "ExiledCards", "FirstForetell", "ForEachShard", "Foraged",
+        "ForagedCards", "IgnoreGeneric", "MinMana", "ModeCost", "OnlyFirstSpell",
+        "RaiseCost", "RaiseTo", "ReduceAmount", "ReduceCost", "Relative",
+        "SpellDescription", "TapCreaturesForMana", "Type", "UnlessValidTarget", "UpTo",
+        "ValidCard", "ValidSpell", "ValidTarget",
+    };
+
     /**
      * Serializables need a version ID.
      */
--- a/forge-game/src/main/java/forge/game/spellability/SpellAbility.java
+++ b/forge-game/src/main/java/forge/game/spellability/SpellAbility.java
@@ -74,6 +74,18 @@ import forge.game.zone.ZoneType;
  * @version $Id$
  */
 public abstract class SpellAbility extends CardTraitBase implements ISpellAbility, IIdentifiable, Comparable<SpellAbility> {
+    public static final String[] OPTIONAL_PARAMS = {
+        "AlternateCost", "Amount", "Announce", "Boast", "CantCopy", "CloakUp", "CostDesc",
+        "CumulativeUpkeep", "DisguiseUp", "DividedAsYouChoose", "Exhaust", "Hidden",
+        "IsCurse", "ManaRestriction", "ManifestUp", "MaxTotalTargetCMC",
+        "MaxTotalTargetPower", "MorphUp", "Origin", "Planeswalker", "PowerUp",
+        "PrecostDesc", "SpellDescription", "TargetType", "TargetingPlayer",
+        "TargetingPlayerControls", "TargetsWithControllerProperty",
+        "TargetsWithDefinedController", "TargetsWithRelatedProperty",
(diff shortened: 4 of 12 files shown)


forge-game/src/test/java/forge/game/ability/CardScriptParamDeclarationTest.java
```diff
@@ -0,0 +1,256 @@
...
+        "\\b\\w*[Kk]ey\\w*\\s*=\\s*\"([A-Za-z][A-Za-z0-9]+)\"");
+    private static final Pattern ADD_TO_COMBAT = Pattern.compile("\\baddToCombat\\(([^)]*)\\)");
+    private static final Pattern STRING_LITERAL = Pattern.compile("\"([A-Za-z][A-Za-z0-9]*)\"");
+
+    // The ability declarator prefixes are not params
+    private static final Set<String> MARKERS = Set.of("AB", "SP", "ST", "DB");
+
+    // Framework classes that own the shared base layers (their OPTIONAL_PARAMS form the base)
+    private static final String[] FRAMEWORK = {
+        "forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java",
```
tool4ever: hardcoding paths is too ugly imo
maybe we can do an `IHasForgeParams` with `getRequired/OptionalParams` or something?

MostCromulent: > Evaluating the IHasForgeParams suggestion:
> 
> The interface and the hardcoded-path concern are largely orthogonal. The path list exists for discovery — finding which classes declare params. An interface doesn’t make classes discoverable; enumerating its implementors still needs either a classpath/reflection scan or a hardcoded list. What actually removes the list is discovering declarers by walking the source tree, and that’s independent of how a declaration is stored — static field, annotation, or interface method. So adopting the interface on its own wouldn’t drop the paths.
> 
> As the mechanism for a build-time check specifically, the interface also has two frictions. Two of the param owners, AbilityUtils and AbilityFactory, are stateless utility classes that are never instantiated (AbilityFactory is final); their declarations can’t live on an instance method, so they’d stay as static fields — leaving two declaration forms to find and read. And reading declarations through the interface needs instances: the abstract bases (SpellAbilityEffect, SpellAbility, CardTraitBase) can’t be constructed, so a build check would end up reading the returned constant or the source anyway.
> 
> Where the interface is clearly the right tool is a runtime-queryable contract — the engine asking a live SpellAbility which params it accepts. There instance methods fit, inheritance covers the abstract bases, and the utilities legitimately don’t participate. If instead the intent is just to drop the hardcoded paths, source-tree-walk discovery does that directly (the test already walks the effects directory; extending it to the framework classes removes the list), with no new dependency, and it crosses the forge-game/forge-ai boundary because it reads files rather than reflecting on classes (forge-game doesn’t depend on forge-ai).
> 
> So the deciding question is whether these params need to be queryable from the engine at runtime, or only validated in a build check. The first justifies the interface; the second is solved by discovery alone, independent of declaration form.

MostCromulent: Might have found a way around this after making Claude do another look.

MostCromulent: After much prodding of Claude, I think https://github.com/Card-Forge/forge/pull/11007/commits/a5fffaf32bdbe9a4074323f2bf256e0f20d288eb might be closer to what you're after?

forge-game/src/test/java/forge/game/ability/CardScriptParamDeclarationTest.java
```diff
@@ -0,0 +1,325 @@
...
+ * Declarers are found by scanning the classpath for implementors of {@link IHasForgeParams}: a
+ * class is covered iff it implements that interface and lists its params, so there is no path or
+ * package list to maintain. Listing a param field without implementing the interface fails the
+ * test, so discovery and declaration cannot drift apart.
+ */
+public class CardScriptParamDeclarationTest {
+
+    // The literal forms a card-script parameter read takes in source
+    private static final Pattern GETPARAM = Pattern.compile(
+        "(?<![A-Za-z0-9_])(?:getParam|hasParam|getParamOrDefault|getParamOrDefaultBoolean)\\(\\s*\"([^\"]+)\"");
```
tool4ever, line 42: `getParamOrDefaultBoolean` doesn't exist?

forge-game/src/main/java/forge/game/ability/AbilityUtils.java
```diff
@@ -49,7 +49,14 @@
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
-public class AbilityUtils {
+public class AbilityUtils implements IHasForgeParams {
+    public static final String[] OPTIONAL_PARAMS = {
+        "AbilityCount", "AnnounceMax", "Destination", "ETB", "ForgetOtherTargets",
```
tool4ever, line 54: Destination is effect specific?

forge-game/src/main/java/forge/game/ability/AbilityFactory.java
```diff
@@ -45,7 +45,18 @@
  * @author Forge
  * @version $Id$
  */
-public final class AbilityFactory {
+public final class AbilityFactory implements IHasForgeParams {
+    public static final String[] OPTIONAL_PARAMS = {
```
tool4ever, line 49: I'd still like to see `additionalAbilityKeys` unioned here less manually

forge-game/src/main/java/forge/game/cost/Cost.java
```diff
@@ -46,7 +47,16 @@
  * @author Forge
  * @version $Id$
  */
-public class Cost implements Serializable {
+public class Cost implements Serializable, IHasForgeParams {
+    public static final String[] OPTIONAL_PARAMS = {
+        "AffectedZone", "Amount", "Announce", "Collected", "CollectedCards", "Color",
```
tool4ever, line 52: would like to see the Hash* fields queried instead if possible

forge-game/src/main/java/forge/game/spellability/SpellAbility.java
```diff
@@ -74,6 +74,18 @@
  * @version $Id$
  */
 public abstract class SpellAbility extends CardTraitBase implements ISpellAbility, IIdentifiable, Comparable<SpellAbility> {
+    public static final String[] OPTIONAL_PARAMS = {
+        "AlternateCost", "Amount", "Announce", "Boast", "CantCopy", "CloakUp", "CostDesc",
+        "CumulativeUpkeep", "DisguiseUp", "DividedAsYouChoose", "Exhaust", "Hidden",
```
tool4ever, line 79: sometimes, like with `CumulativeUpkeep` they are only used internally as helper

so for this case if it'd show up in a script it'd actually be wrong

conceptually I'm not sure yet if we need a third category for params that are technically still implemented but very exotic/deprecated instead 🤔

tool4ever:
In general direction seems good.

> So the deciding question is whether these params need to be queryable from the engine at runtime, or only validated in a build check. The first justifies the interface; the second is solved by discovery alone, independent of declaration form.

I want runtime so the Card Workshop can eventually also make partial use of these. 💡

For now I'd still like to see the scope more clearly defined:

1. I guess we're looking at covering syntax for a pretty good amount of our standard.
Is there anything we can't detect easily which can at least be noted as "skipped" - consider options for confidence interval + statistics available?

2. We'll also have some simpler semantics checks (e.g. unused SVar)
Is this one still easier with Python so we'll have a clean responsibility split - or would it now only be responsible for the PR commenting part?

3. What parts of 1+2 run via the normal test suite and what only gets included when CI?
Should we support different failure modes (Warning vs. Error) depending on
a) execution context
b) type of logic error
c) type of script

4. Before doing more batches devise strategy with the fields in `CardScriptParser`
- Imo they should be refactored together to avoid redundancy
- This might mean MVP of these two PR first

5. if the parent PR tests more of the syntax via Java does it get much slower? Should it be limited only to `/upcoming`?

6. Is there still a use case for keeping the corpus discovery approach? E.g. it might be too messy to also manually define each Params accepted values (+ we're far from being consistent in including "Defined" or "Valid" in their names)
And sometimes only "True" is valid and others like `CombatDamage` are actually tri-state...

---

So the future order could be this:
1. Java test: IHasForgeParams alignment
2. Java test: check for valid API
3. Java test: Linter
4. Python via CI: compare with Scryfall
5. Github workflow: post findings from 3.+4. as comments

Might actually be worth a markdown file explaining these workflow steps 🤔

jamincollins:
## Findings
#### 1. IHasForgeParams reflection accessor uses public field lookup which may miss private or protected declarations
Description:
IHasForgeParams.getOptionalParams() uses getClass().getField("OPTIONAL_PARAMS") which only finds public fields. If any declarer defines OPTIONAL_PARAMS with non-public visibility, the accessor silently returns empty array, causing CardScriptParamDeclarationTest to think the class declares no params and potentially miss reads.
Test case:
```java
public class PrivateDeclarer implements IHasForgeParams {
    static final String[] OPTIONAL_PARAMS = {"Secret"};
}
String[] params = new PrivateDeclarer().getOptionalParams();
assert params.length == 0; // field not public, getField throws NoSuchFieldException
```
Impact: Silent loss of declarations for non-public fields, leading to false negatives in drift detection.

#### 2. CardScriptParamDeclarationTest literal read detection misses dynamic param names
Description:
The test uses regex on source literals to find reads like getParam("X"). If code uses a constant or concatenates a param name, the read is invisible to the test, so a param can be removed from OPTIONAL_PARAMS without triggering a failure.
Test case:
```java
private static final String PARAM = "HiddenParam";
sa.getParam(PARAM);
```
Impact: Drift detection is incomplete; undeclared dynamic reads will pass.

#### 3. Required param groups are not validated for duplicates within a group
Description:
requiredGroups parses REQUIRED_PARAMS as String[][] but does not check for duplicate entries within the same inner array. Duplicate entries are allowed by the parser and would cause ambiguous one-of semantics.
Test case:
```java
public static final String[][] REQUIRED_PARAMS = {{"A","A","B"}};
```
Impact: Potential confusion in validation logic; test does not flag duplicates inside groups.

#### 4. getOptionalParams catches both NoSuchFieldException and IllegalAccessException and returns empty array
Description:
When field access fails due to IllegalAccessException (e.g., SecurityManager), the method returns empty array without logging. This masks configuration errors.
Test case:
```java
// Under SecurityManager that blocks reflection
String[] params = new SomeClass().getOptionalParams();
```
Impact: Silent degradation of param discovery.

#### 5. SpellAbility OPTIONAL_PARAMS may overlap with CardTraitBase OPTIONAL_PARAMS
Description:
SpellAbility extends CardTraitBase which already declares many params. SpellAbility adds its own OPTIONAL_PARAMS. The test checks effects must not re-declare base params, but SpellAbility is not an effect (EFFECTS_DIR check). Overlap between base and subclass params is not prevented, risking duplicate declarations across inheritance.
Test case:
```java
// Both CardTraitBase and SpellAbility declare "Amount"
```
Impact: Ambiguity about which declaration is authoritative; may cause drift test false positives/negatives.

#### 6. READ_GATE_EXEMPT only contains SpellAbilityAi
Description:
The exemption set is hard-coded to SpellAbilityAi. If another cross-cutting reader is added later, undeclared reads will be incorrectly flagged.
Test case:
Add new utility class that reads params owned by others.
Impact: Future maintenance burden; test will require update.

### Side Effects & Interactions with Existing Systems
#### 7. CardScriptParamDeclarationTest scans classpath at test runtime
Description:
The test uses ClassPath.from(cl).getTopLevelClassesRecursive(SCAN_PACKAGE) which loads classes. If a class has static initializers with side effects, test execution may trigger unintended game initialization.
Test case:
A declarer with static block that accesses DB.
Impact: Test flakiness / side effects.

#### 8. IHasForgeParams.getOptionalParams uses reflection on every call
Description:
Reflection lookup per call is expensive if used in hot paths. While test-only use may be fine, future consumers may call it frequently.
Test case:
Loop 1M calls to getOptionalParams.
Impact: Performance degradation.

### Performance and Security
#### 9. CardScriptParamDeclarationTest reads entire source files into memory for regex scans
Description:
The test reads whole files via Files.readAllBytes and runs multiple regex passes. For large codebases this can be memory intensive.
Test case:
Large source file 1MB.
Impact: Increased memory usage during test.

#### 10. Regex patterns for param reads may over-match
Description:
Patterns like MAPGET match any .get("A-Za-z...") which may capture non-param strings, leading to false positive undeclared reads.
Test case:
```java
map.get("SomeNonParamKey");
```
Impact: Noise in error reporting.

## Verdict
Request Changes
The PR introduces a valuable drift detection mechanism but contains several correctness and maintainability gaps: reflection visibility assumptions, incomplete read detection, missing duplicate checks, and hard-coded exemptions. These should be addressed before merging.

## Recommended Fixes
1. Change IHasForgeParams.getOptionalParams to use getDeclaredField with setAccessible(true) and walk superclass chain to support non-public fields.
2. Add duplicate detection within REQUIRED_PARAMS groups.
3. Log or fail fast on IllegalAccessException instead of silently returning empty.
4. Document that dynamic param reads are not covered by the test and consider a whitelist for known dynamic usages.
5. Add an explicit check for overlap between subclass and superclass OPTIONAL_PARAMS and define inheritance semantics.
6. Make READ_GATE_EXEMPT configurable or derive it from a marker annotation.

---

MostCromulent:
Closing in favour of #12030, which combines this PR with #10984 into one MVP. The `IHasForgeParams` declarations and batch 1 carry over, with the regex drift test replaced by a bytecode scan. The review questions in this thread are answered point by point in the new PR's description.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #11015, opened 2026-06-16
Remember replacement effect ordering

### LINKED ISSUE #5182: Request: Replacement Trigger Order Memory
I play Hardened Scales, Kami of Whispered Hopes, Doubling Season, Branching Evolution and more alongside Hakbal, when Hakbal triggers I get frustrated by having to select which order in which to proc these effects every time. Lots of Explore means lots of triggers.

Select an order, So long as there's no new things use that order for auto-yielding that proc

Automatically deciding the procs in additive then multiplicative order but that might not work for all people in all cases. The saving the replacement order to auto-yield is a feature with wider impacts

**Additional context**
Also with the explore and putting into a graveyard, if you don't put something to the graveyard you likely don't want to do that for the remainder of the turn so it may be worth saving that decision until the end of turn too.

Madwand99:
This extends the existing “remember this order” behavior from #10683 for simultaneous ability ordering to replacement effects.

Changes include:

- Applies the ordering dialog to replacement effects when multiple replacement effects are available.
- Allows “Always use this order” to remember replacement-effect ordering for the current game.
- Reuses the existing saved-order reset flow so replacement-effect order memory is cleared alongside ability order memory.
- Handles stale or invalid remembered choices defensively by falling back to the normal selection path.
- Updates desktop/mobile/adventure-facing localization text for the new replacement-effect ordering prompts.

Thanks to Codex for help with code!


forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -2102,22 +2102,104 @@ public ReplacementEffect chooseSingleReplacementEffect(final List<ReplacementEff
...
+        } else {
+            for (final ReplacementEffect replacementEffect : possibleReplacers) {
+                orderedREVs.add(replacementEffect.getView());
+            }
+        }
+
+        final boolean reordering = savedOrder != null;
+        final IGuiGame.OrderResult<ReplacementEffectView> orderResult = getGui().order(
+                localizer.getMessage(reordering ? "lblReorderReplacementEffects" : "lblSelectOrderForReplacementEffects"),
+                localizer.getMessage("lblApplyFirst"), 0, 0,
```
tool4ever: should probably allow just choosing one, otherwise you're forcing more than backed by the rules
(which would be especially useless in case the RE are mutually exclusive anyway)

Madwand99: Good call! Fixed.

tool4ever:
@Hanmac
you think it's clear enough?

Hanmac (dismissed):

Hanmac:
@tool4ever the `List<ReplacementEffectView>` -> `List<ReplacementEffect>` part could be cleaned up, but that maybe later

MostCromulent (approved):
Will leave code review to the experts, but in concept looks good to me.

tool4ever (approved):

PR #11031, opened 2026-06-17
Fix GameEventPlayerStatsChanged carrying ALL cards

tool4ever:
There should be no reason that player keywords still cause cards to update.
Besides `MatchController#refreshCardDetails` hasn't even used that param for a long time anyway.

@MostCromulent
maybe Claude can check I didn't miss something?

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/DamagePreventEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/DamagePreventEffect.java
@@ -168,7 +168,7 @@ public class DamagePreventEffect extends SpellAbilityEffect {
 
         o.getView().updatePreventNextDamage(o);
         if (o instanceof Player) {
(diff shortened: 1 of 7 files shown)


Hanmac (approved):

MostCromulent:
@tool4ever Claude says all clear:

> I reviewed PR # 11031 and it's correct and safe to merge. The refreshCardDetails param really is
>   unused on both platforms — desktop's CMatchUI never overrides it (no-op in AbstractGuiGame) and
>   mobile's MatchController ignores its cards argument outright — so the precomputed allCards
>   collection was pure dead weight. All call sites are accounted for (the old boolean constructors
>   are removed, so any miss would fail to compile), allCards()/updateCards() are referenced nowhere
>   else, and the FCollection<PlayerView> type is sound and serializable for the network-forwarded
>   event. Player keyword/detail display still refreshes via the retained processPlayer(p, 
>   livesUpdate) path.

PR #11037, opened 2026-06-18
Move GameState to game package; some cleanup

Jetz72:
Was doing some experimentation with the GameState class and it seemed like something that'd make more sense to have in the game module than the AI one. It doesn't appear dependent on anything else in the AI module, so I've moved it.

It also had a vestigial abstract `getPaperCard` method, which was implemented the same way everywhere and - as far as I can tell - hasn't been used since a PR in 2017 replaced its only invocation with something else. Removed that method. It was also the only abstract method remaining in the class, so for now I made the class concrete and removed the anonymous implementations. We could go the other way and re-insert `getPaperCard` as the supplier for the card in `processCardsForZone`, but I'm not sure what we'd ever override it with and this way seems cleaner.

Changed the return type of `chooseCardsToDiscardToMaximumHandSize` to match the ones from `chooseCardsToDiscardFrom` and `chooseCardsToDiscardUnlessType`.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
+++ b/forge-ai/src/main/java/forge/ai/PlayerControllerAi.java
@@ -877,7 +877,7 @@ public class PlayerControllerAi extends PlayerController {
     }
 
     @Override
-    public CardCollection chooseCardsToDiscardToMaximumHandSize(int numDiscard) {
+    public CardCollectionView chooseCardsToDiscardToMaximumHandSize(int numDiscard) {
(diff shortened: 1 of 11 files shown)


Hanmac (approved):

Jetz72:
(Cleaned up some wording, and also noticed I hit "assign" instead of "request review". I probably shouldn't write PR submissions when I'm 2 minutes away from falling asleep at my desk.)

tool4ever (approved):

PR #11044, opened 2026-06-18
Add deck URL loading to the deck importer

MostCromulent:
Screenshot 2026-06-19 080637

## Summary

Follow-up to #10570, which added Moxfield/Archidekt deck URL loading to the deck chooser. Per [this comment](https://github.com/Card-Forge/forge/pull/10570#issuecomment-4646969017), this wires the same loader into the ordinary deck importer flow.

A URL field and Load button sit above the Card List in the importer. Pasting a URL fetches the decklist text and populates the import box, so the existing live preview and art/date/format filters apply just as with pasted text.

---


MostCromulent:
Polite bump for https://github.com/Card-Forge/forge/pull/10517 while I'm here. 

That should solve the recurring issue of users not knowing how to import decks to commander format, which I would think will be even more common with the URL import in place.

forge-gui/src/main/java/forge/deck/DeckImportController.java
```diff
@@ -194,6 +195,10 @@ public void importBannedAndRestrictedCards(boolean includeBannedAndRestricted){
 
     public boolean importBannedAndRestrictedCards(){ return this.includeBnRInDeck; }
 
+    public String getDeckTextFromUrl(final String deckUrl) throws IOException {
+        return DeckUrlLoader.loadImportText(deckUrl);
```
tool4ever: no one line helpers that grow the stack for no value

PR #11045, opened 2026-06-19
Add TappedOut and MTGGoldfish deck URL loading

Madwand99:
This expands deck URL loading beyond Moxfield and Archidekt with support for TappedOut and MTGGoldfish.

Thanks to Codex and #10247 for help with code and inspiration.


forge-gui/src/main/java/forge/deck/DeckUrlLoader.java
```diff
@@ -131,6 +142,18 @@ static String getDeckName(final Map<?, ?> root, final String deckId, final Strin
         return requestedName;
     }
 
+    private static void deleteRenamedSourceDecks(final StorageImmediatelySerialized<Deck> storage, final Deck deck) throws IOException {
+        final List<String> oldNames = new ArrayList<>();
+        for (final Deck savedDeck : storage) {
+            if (!deck.getName().equals(savedDeck.getName()) && isSameSourceDeck(deck.getSourceUrl(), savedDeck.getSourceUrl())) {
```
tool4ever: are you silently deleting decks user might have modified further locally?

Madwand99: I was -- I removed this behavior. The intent was to handle provider-side deck renames cleanly. For example, TappedOut can give us a better deck title than the slug-derived name, so on reload we wanted to avoid leaving both the old saved URL deck name and the refreshed provider title around for the same source deck. But this is probably too aggressive: same source URL does not mean the local saved deck is still an unmodified import.

tool4ever: yea sure, both use cases are valid
you should just ask the user first or something

Madwand99: For more complex handling of cases like this, I'm basically waiting for my Deck Browser PR to be accepted. There's a lot of functionality that can be implemented once that's available. For now it's probably best to keep things as simple as possible.

tool4ever (approved):

PR #11050, opened 2026-06-19
ItemColumn: nullsFirst on function

Hanmac:
On the case when the function returns null instead of a comparable


tool4ever (approved):
don't think I'll have time to test before next snap but I have no idea if a chain of 4 comparators is really better anyway 🤷‍♂️

PR #11055, opened 2026-06-20
Script fixes

### LINKED ISSUE #11054: Bugs with new Namor and Alicia
Namor Atlantean King, when attacking a player with higher life it grants other creatures +2.  The effect DOES trigger  but it doesn't grant the bonus.

Alicia Masters Skilled Sculpter, should be 0/4 showing as 4/4

tool4ever:
Closes #11054


Hanmac (approved):

Hanmac:
@tool4ever apropos `forEachCardInGame`, I found some more that uses `new Visitor`

And while looking at it, I thought about replacing this with `Stream.concat` and use `Predicate` instead?

PR #11091, opened 2026-06-27
Change mana restriction CantCast to be more generic

Eradev:
Change the mana restriction check CantCast to be more generic to support custom cards such as:

image


tool4ever:
syntax wise it's still a bit ugly:
it'd really be nice if we could just write `!Spell,Spell.Artifact` instead...
but that means solving the TODO, I feel like we had a discussion for that somewhere already? 🤔

Hanmac:
@tool4ever should we revive this MR?

tool4ever:
not sure we need more stale PR in the backlog

PR #11093, opened 2026-06-27
Replace Scryfall's API with custom API using Scryfall data

Eradev:
We've seen increased reports of throttling by Scryfall recently.

This replaces Scryfall's API with my own which uses Scryfall's bulk data to generate a lookup table to their CDN.

* Bypass Scryfall's API for lookup (Only refreshes the DB once per day at around 2:00 AM EST);
* Cloudflare-cached for speed;
* Only implements [/cards/named](https://scryfall.com/docs/api/cards/named) and [/cards/{set}/{num}](https://scryfall.com/docs/api/cards/collector);
* Only return images.

I can give access to the repo if maintainers want access.


forge-core/src/main/java/forge/util/ImageUtil.java
```diff
@@ -257,7 +257,7 @@ public static String getScryfallDownloadUrl(PaperCard cp, String face, String se
             cardCollectorNumber = cardCollectorNumber.substring(0, cardCollectorNumber.length() - 1);
         }
 
-        return String.format("%s/%s/%s?format=image&version=%s%s", editionCode, encodeUtf8(cardCollectorNumber),
+        return String.format("%s/%s/%s?version=%s%s", editionCode, encodeUtf8(cardCollectorNumber),
```
tehdiplomat: Since you are accepting get params, does `format=image` not just get ignored by your webserver?

Eradev: Yes it is ignored.

tehdiplomat: So I feel like we shouldn't be making these line changes. Maybe just have a UI element called "Use Scryfall Mirror" that passes in the new URL instead of scryfalls?

Eradev: Added it as an option.

tool4ever (commented):
thanks though I'd prefer a more official approach via #10928

Eradev:
I just wanted to avoid all the uncessary files, and users needing to clear cached stuff, especially with all the new spoilers coming out months before the release.

jamincollins:
## Findings
#### 1. Alternate API base URL selection uses preference at call time
The PR adds `ImageFetcher.getCardArtApiBaseUrl()` which reads `UI_USE_ALTERNATE_CARD_ART_API` on each call and returns either `URL_PIC_CARDFORGE_NET_API_DOWNLOAD` or `URL_PIC_SCRYFALL_DOWNLOAD`. `buildCardArtApiUrl` concatenates the chosen base with Scryfall-style paths produced by `ImageUtil.getScryfallDownloadUrl`. 
Test case:
```java
// Simulate preference toggle
ForgePreferences prefs = FModel.getPreferences();
prefs.setPref(ForgePreferences.FPref.UI_USE_ALTERNATE_CARD_ART_API, "true");
String url = ImageFetcher.buildCardArtApiUrl("a/b/c.jpg");
// Expected: url starts with https://card-forge.net/api/v1/cards/
assert url.startsWith("https://card-forge.net/api/v1/cards/");
```
Impact: The path format is assumed identical between Scryfall and the custom API. If card-forge.net does not accept the same `/cards/{set}/{collector}/{...}` paths or the `named?fuzzy=` query used in `GuiDownloadPicturesHQ`, image downloads will silently fail. The PR does not verify API compatibility.

#### 2. Named search query uses Scryfall endpoint semantics on alternate API
`GuiDownloadPicturesHQ` builds `scryfallurl = ImageFetcher.getCardArtApiBaseUrl() + "named?fuzzy=" + cardname` with additional `set`, `face`, and `format=image` params. Scryfall's named endpoint is specific to Scryfall; card-forge.net may not implement the same query parameters.
Test case:
```java
String base = ImageFetcher.getCardArtApiBaseUrl();
String url = base + "named?fuzzy=Lightning Bolt&format=image";
// If alternate API is enabled, this URL will be issued to card-forge.net
// Verify HTTP 200 and image content; otherwise fallback is required.
```
Impact: Users with alternate API enabled will receive 404/empty responses for manual HQ downloads, degrading the download feature.

#### 3. Cooldown logic applies to alternate API URLs in LibGDXImageFetcher
`LibGDXImageFetcher` checks `if (scryfallCooldownTime != null && urlToDownload.startsWith(getCardArtApiBaseUrl()))` before skipping downloads. The cooldown is set only when Scryfall returns 429, but the check now gates the alternate API as well because `getCardArtApiBaseUrl` returns the alternate base when enabled.
Test case:
```java
// Simulate Scryfall 429 cooldown
ImageFetcher.noteScryfallRateLimited(); // sets scryfallCooldownTime
ForgePreferences prefs = FModel.getPreferences();
prefs.setPref(ForgePreferences.FPref.UI_USE_ALTERNATE_CARD_ART_API, "true");
String altUrl = "https://card-forge.net/api/v1/cards/abc.jpg";
// LibGDXImageFetcher will skip this URL while cooldown is active
```
Impact: A Scryfall rate-limit event incorrectly blocks downloads from the alternate API, causing unnecessary missing images.

#### 4. Preference read in static method may race with initialization
`getCardArtApiBaseUrl` is static and reads `FModel.getPreferences()` each call. If called during early startup before preferences are loaded, `getPrefBoolean` may return default or throw.
Test case:
```java
// Call during static init before FModel is ready
String base = ImageFetcher.getCardArtApiBaseUrl();
```
Impact: Potential NPE or defaulting to Scryfall unexpectedly, leading to inconsistent URL selection across the session.

#### 5. Scryfall pacing and cooldown helpers remain tied to Scryfall URL constant
Master branch contains `isScryfall(url)` which checks `URL_PIC_SCRYFALL_DOWNLOAD`. The PR does not update these helpers. When alternate API is enabled, `isScryfall` returns false for all URLs, so `paceScryfall` and `inScryfallCooldown` never throttle the alternate API.
Test case:
```java
String alt = "https://card-forge.net/api/v1/cards/xyz.jpg";
boolean isScry = ImageFetcher.isScryfall(alt); // false
ImageFetcher.paceScryfall(alt); // no-op
```
Impact: If card-forge.net has its own rate limits, Forge will hammer it without pacing, risking bans or errors. Conversely, Scryfall pacing is lost if the URL is rewritten to alternate base.

#### 6. UI preference default is false, but no migration or documentation
`UI_USE_ALTERNATE_CARD_ART_API` defaults to `"false"`. Existing users keep Scryfall behavior, which is correct. However, no migration path or user-facing warning exists if card-forge.net is down.
Impact: Silent fallback missing; users may enable the option and see missing images with no indication why.

#### 7. SwingImageFetcher and LibGDXImageFetcher replace URL constant checks with dynamic base
Both fetchers changed `urlToDownload.startsWith(ForgeConstants.URL_PIC_SCRYFALL_DOWNLOAD)` to `urlToDownload.startsWith(getCardArtApiBaseUrl())`. This affects fullborder filename logic. If a cached file path was created under Scryfall base, switching preference mid-session can cause mismatched destination paths.
Test case:
```java
// User toggles preference after images cached
// Old dest path: .../cache/.../abc.full.jpg from Scryfall URL
// New URL base changes, but dest path logic still uses same filename
```
Impact: Potential duplicate downloads or cache misses when toggling the option.

#### 8. GuiDownloadFilteredCardImages now uses buildCardArtApiUrl for all filtered images
Previously Scryfall base was hardcoded; now it uses the dynamic base. If the alternate API mirrors Scryfall data but uses different language codes or file extensions, filtered downloads may break.
Impact: Inconsistent image sets between users on different APIs.

### Performance and security
#### 9. No validation of alternate API URL scheme or response
The PR trusts `https://card-forge.net/api/v1/cards/` without validating TLS, response codes, or content-type. No timeout or retry differentiation for the alternate host.
Test case:
```java
// Simulate alternate API returning HTML error page
// ImageFetcher will attempt to write HTML to .jpg file
```
Impact: Corrupted cache files, wasted bandwidth, and potential exposure to malicious content if the host is compromised.

## Verdict
Request Changes
The feature introduces an alternate image source but assumes API compatibility without verification, leaks Scryfall-specific cooldown logic to the alternate host, and lacks fallback or validation. The PR also diverges from master’s Scryfall rate-limiting changes, creating merge risk.

## Recommended Fixes
1. Verify card-forge.net API contract: confirm it accepts the same path and query parameters as Scryfall, and implement a fallback to Scryfall on non-200 responses.
2. Decouple cooldown/pacing from base URL: make `isScryfall` and pacing logic configurable per host, or disable cooldown for the alternate API.
3. Add unit tests for `getCardArtApiBaseUrl` and `buildCardArtApiUrl` with preference toggles and null safety.
4. Guard `getCardArtApiBaseUrl` against uninitialized preferences and cache the base URL after first successful read.
5. Document the alternate API option and add a health check / error message when downloads fail due to the alternate host.

---

Issue #10922, opened 2026-06-08
AI doesn't pay 2 life when The Kingpin of Crime ability triggers


MorihiroS:
**Describe the bug**
AI has creatures with toughness greater than their power but doesn't pay 2 life for attack when ability of The Kingpin of Crime triggers

**To Reproduce**
1. AI has The Kingpin of Crime and 2 other creatures with toughness greater than their power
2. Ability of The Kingpin of Crime triggers
3. AI doesn't pay 2 life
4. AI doesn't attack

**Expected behavior**
AI should pay 2 life when The Kingpin of Crime ability triggers if
a) it has more then 8 hp;
b) it might attack with success

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version: 2.0.13-SNAPSHOT-06.07

**Screenshots**

<img width="1536" height="764" alt="Image" src="https://github.com/user-attachments/assets/8e692089-380e-4b34-be10-d28186690a30" />

MorihiroS:
@tool4ever It seems to me that the problem was exactly the opposite—the AI wasn't paying 2 life. Why close the issue then, if the request is about something else entirely?

tool4ever:
it really doesn't matter, marking can be done either way

MorihiroS:
@tool4ever AI still uses Exort and still synergizes with Doran, the Siege Tower and similar cards. Why take away the AI’s ability to play it in a synergistic deck, where it’s 1/5 with Exort, is a thousand-year-old mystery to me. Please reopen the issue. And it's better to roll back the “improvement”; I don't see a big problem here.

<img width="1540" height="786" alt="Image" src="https://github.com/user-attachments/assets/4ca6d07c-2661-48a5-952b-efd93c413839" />
<img width="1539" height="781" alt="Image" src="https://github.com/user-attachments/assets/7e61eea7-c49a-44d3-a077-f08e6d8db2be" />

tool4ever:
just because YOU would be fine with giving AI an only half usable card doesn't mean it should be the standard play experience for all users 🤦‍♂️

and the very fact you opened this issue because the ability wasn't being used already shows we can't expect such difference to be clearly understood currently anyway

(FWIW AI did still cast them in the past but still too many people ignored the warning and then complained 🤷‍♂️)

MorihiroS:
> just because YOU would be fine with giving AI an only half usable card doesn't mean it should be the standard play experience for all users 🤦‍♂️

[...]

PR #11113, opened 2026-06-29
Remove The Kingpin of Crime from AI decks

Rauglothgor:
﻿## Problem
The Kingpin of Crime changes combat damage after attackers are already declared. The previous AI heuristic tried to pay for the trigger at that point, but that is not safe as a narrow fix because attack selection has already used the current power/toughness assumptions.

Related to #10922.

## Minimal Fix
Add `AI:RemoveDeck:All` so Forge does not include The Kingpin of Crime in AI decks until combat planning can account for this effect before attacks are chosen.

## Validation
- `git diff --check upstream/master --`
- Confirmed the PR diff is one card-script line: `AI:RemoveDeck:All`.

## Known Limits
This does not implement a general AI plan for toughness-as-combat-damage effects. It only prevents the unsupported card from being selected for AI decks.

## AI Assistance Disclosure
I used AI assistance to inspect the review feedback, revise the branch, and run validation. I reviewed the final one-line diff before updating the PR.

### DIFF
--- a/forge-gui/res/cardsfolder/t/the_kingpin_of_crime.txt
+++ b/forge-gui/res/cardsfolder/t/the_kingpin_of_crime.txt
@@ -6,5 +6,5 @@ K:Extort
 T:Mode$ AttackersDeclared | AttackingPlayer$ You | Execute$ TrigEffect | TriggerZones$ Battlefield | TriggerDescription$ Whenever you attack, you may pay 2 life. If you do, until end of turn, creatures you control with toughness greater than their power assign combat damage equal to their toughness rather than their power.
 SVar:TrigEffect:AB$ Effect | Cost$ PayLife<2> | StaticAbilities$ CombatDamageToughness
 SVar:CombatDamageToughness:Mode$ CombatDamageToughness | ValidCard$ Creature.YouCtrl+powerLTtoughness | Description$ Until end of turn, creatures you control with toughness greater than their power assign combat damage equal to their toughness rather than their power.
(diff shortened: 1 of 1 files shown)


tool4ever (changes requested):
doesn't really convince me:
whole AI attack calculation is based on current toughness to decide if creature can survive + other factors related to power

making such a big switch after can easily have AI screw itself over royally instead

I doubt this can be wired together in a reasonable way 🤔

otherwise this card just seems like a prime candidate for `RemoveDeck:All`

Rauglothgor:
﻿Thanks, that makes sense. I dropped the `EffectAi` logic entirely because the attack has already been chosen by the time this trigger resolves, so that approach is not safe as a narrow fix.

I revised the PR to only add `AI:RemoveDeck:All` to The Kingpin of Crime, matching your suggested fallback. The current diff is just that one card-script line.

Validation: `git diff --check upstream/master --`

tool4ever (approved):

MorihiroS:
>The previous AI heuristic tried to pay for the trigger at that point, but that is not safe as a narrow fix because attack selection has already used the current power/toughness assumptions.

@Rauglothgor Actually, AI didn't even try to pay 2 hp, and I probably missed the moment when it started doing that...

MorihiroS:
I tested it on version 2.0.14-SNAPSHOT-06.29; AI completely ignores the trigger, but exort works. I really don't understand why such a strange solution was necessary.

PR #11115, opened 2026-06-29
Improve death-drain creature evaluation

### LINKED ISSUE #4171: AI would rather kill Serrated Scorpion than Zulaport Cutthroat/Blood Artist
**Describe the bug**
AI would rather kill Serrated Scorpion than Zulaport Cutthroat/Blood Artist. But abilities of Zulaport Cutthroat and Blood Artist are more dangerous.

**To Reproduce**
1. Player casts Zulaport Cutthroat and Serrated Scorpion
2. AI tries to destroy Serrated Scorpion

**Expected behavior**
AI should consider abilities of Zulaport Cutthroat/Blood Artist/Cruel Celebrant as slightly more dangerous, than single-use Serrated Scorpion.

**Desktop**
OS: Windows 10
Java Version: 1.8.0_202-b08
Forge Version: 1.6.59-SNAPSHOT-11.0X

**Screenshots**

zulaport

Rauglothgor:
﻿## Problem
When choosing creature removal targets, the AI can prefer a one-shot death trigger such as Serrated Scorpion over repeatable death-drain engines such as Zulaport Cutthroat, Blood Artist, and Cruel Celebrant.

Fixes #4171.

## Minimal Fix
Creature evaluation now adds a small value bump for dies triggers that:
- move from battlefield to graveyard,
- include `Other` in `ValidCard`, and
- execute a `LoseLife` ability.

This keeps self-only death triggers such as Serrated Scorpion from receiving the same repeatable-engine value.

## Validation
- `JAVA_HOME=/home/rauglothgor/.local/codex-tools/jdk17 /home/rauglothgor/.local/codex-tools/apache-maven-3.9.9/bin/mvn -pl forge-ai -am -DskipTests compile`
- Build succeeded with 0 checkstyle violations.

## Known Limits
This is not a general trigger evaluator. It only covers the narrow repeatable death-drain case described in the issue.

## AI Assistance Disclosure
I used AI assistance to inspect the issue, trace the target-evaluation path, make the narrow change, and run validation. I reviewed the final diff and validation before opening this PR.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/CreatureEvaluator.java
+++ b/forge-ai/src/main/java/forge/ai/CreatureEvaluator.java
@@ -333,11 +333,31 @@ public class CreatureEvaluator implements Function<Card, Integer> {
                 continue;
             }
             final SpellAbility ab = t.ensureAbility();
-            if (ab != null && ApiType.LoseLife.equals(ab.getApi())) {
+            if (isRepeatableOpponentDrain(ab)) {
                 value += 30;
             }
         }
         return value;
+    }
+
+    private boolean isRepeatableOpponentDrain(final SpellAbility ab) {
+        if (ab == null || !ApiType.LoseLife.equals(ab.getApi()) || !hasGainLifeSubAbility(ab)) {
+            return false;
+        }
+        final String defined = ab.getParamOrDefault("Defined", "");
+        final String validTgts = ab.getParamOrDefault("ValidTgts", "");
+        return defined.contains("Opponent") || "Player".equals(validTgts) || validTgts.contains("Opponent");
+    }
+
+    private boolean hasGainLifeSubAbility(final SpellAbility ab) {
+        SpellAbility sub = ab.getSubAbility();
+        while (sub != null) {
+            if (ApiType.GainLife.equals(sub.getApi())) {
+                return true;
+            }
+            sub = sub.getSubAbility();
+        }
+        return false;
     }
 
     protected int addValue(int value, String text) {
--- a/forge-ai/src/main/java/forge/ai/CreatureEvaluator.java
+++ b/forge-ai/src/main/java/forge/ai/CreatureEvaluator.java
@@ -175,6 +175,7 @@ public class CreatureEvaluator implements Function<Card, Integer> {
         if (ComputerUtilCard.hasActiveUndyingOrPersist(c)) {
             value += addValue(30, "revive");
         }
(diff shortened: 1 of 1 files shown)


forge-ai/src/main/java/forge/ai/CreatureEvaluator.java
```diff
@@ -318,6 +319,27 @@ private int evaluateSpellAbility(SpellAbility sa) {
...
+            }
+            if (!"Battlefield".equals(t.getParamOrDefault("Origin", ""))
+                    || !"Graveyard".equals(t.getParamOrDefault("Destination", ""))) {
+                continue;
+            }
+            if (!t.getParamOrDefault("ValidCard", "").contains("Other")) {
+                continue;
+            }
+            final SpellAbility ab = t.ensureAbility();
+            if (ab != null && ApiType.LoseLife.equals(ab.getApi())) {
```
tool4ever: a lot of these AI issues are lacking consideration for the bigger picture:
in this case limiting it to the one observed API seems pretty arbitrary since the same argument could easily apply to all others too

but while usually somewhat rarer those triggers could potentially also be a downside for the controller instead (at least partially) 🤔

So I'm not sure this won't lead to too many false positives...

Rauglothgor: Good point. I narrowed this so it no longer scores every dies-triggered LoseLife ability: it now requires the repeatable Other death trigger, outward-facing life loss (Defined opponent or player/opponent target), and a GainLife subability. That keeps it limited to the Zulaport/Blood Artist/Cruel Celebrant pattern from #4171 and avoids self-loss/downside triggers. Validated with mvn -pl forge-ai -am -DskipTests compile.

tool4ever: still way too hardcoded:
including "Other" ETB/LTB triggers might be good but only if score doesn't get biased from one specific case alone 👎

(and otherwise the issue author will just spam more requests with scenarios with identical "isRepeatableX" variants)

Rauglothgor:
Thanks, that makes sense. I do not want to turn this into another hardcoded CreatureEvaluator special case, so I am closing this PR rather than expanding it in the wrong direction.

If I revisit #4171, I will come back with a broader trigger-scoring approach that handles repeatable ETB/LTB payoffs and downside cases consistently instead of adding another isRepeatableX variant.

PR #11121, opened 2026-06-30
CounterType: replace Map with Multiset

Hanmac:
Part of #10886

replaces `Map<CounterType, Integer>` with `Multiset<CounterType>`


forge-game/src/main/java/forge/game/card/CardView.java
```diff
@@ -343,6 +323,7 @@ public boolean hasSamePT(CardView otherCard) {
         return true;
     }
     void updateCounters(Card c) {
+        super.updateCounters(c);
         set(TrackableProperty.Counters, c.getCounters());
```
tool4ever: the 2 calls here are redundant via the super

tool4ever (approved):

Issue #8957, opened 2025-10-19
AI casts Exude Toxin with X = 0


SunFarts:
**Describe the bug**
I raised this issue in #8836 among a couple of other things. I should've separated ea. of those things into its own separate issue. All apologies! I believe only the issue w/ the AI misusing a couple of lands was addressed. The issue is outlined @ the beginning of the linked thread & the rest of the thread can likely be ignored.

**To Reproduce**
1. AI is able to cast Scavenger Regent's Adventure for more than X=0 & prob should based on your board state.
2. You have a board state the AI could wreck using Scavenger Regent's Adventure.
3. Observe carefully.

**Expected behavior**
Toxins are harmful/fatal rather than amusing/annoying puffs of flatulence.

**Desktop (please complete the following information):**
 - OS: Win 10
 - 10.04 snap but it was prob on a previous albeit recent snap

**Additional context**
I saw another thread opened recently where AI is having difficulty using certain X spells. It seems like a really hairy problem because there are so many different X spells doing so many different things; there's no simple, catch-all logic that can have AI using X spells well in all circumstances. Is it really something where ea. X spell needs its own associated logic coded specifically for it??

From what I've seen, it looks like the AI does a before & after evaluation when analyzing combat outcomes. Perhaps something similar could be applied here as it pertains to X spells?? @ the least, if AI sees no significant benefit from using Scavenger Regent's Adventure @ X=0, it wouldn't do it. Altho I just thought of the situation where a player can use the Adventure repeatedly to prevent themselves from decking . . .

aryst0krat:
You should probably _actually_ describe the problem in your report rather than being coy about it, for future reference. This isn't a creative writing assignment. You're just making the job slightly harder.

Jetz72:
I appreciate the prose but maybe in the future keep the titles straightforward so we don't have to cross reference it with the reproduction steps and a Scryfall search before we can figure out what the actual issue you're reporting is.

SunFarts:
@aryst0krat I did actually describe the problem quite clearly in the original post; all you had to do was click the # link I gave right up front in this thread. The problem here seems to be I raised several issues in that original post & this has caused some confusion. I already apologized for that & will refrain from doing that in the future.

[...]

PR #11124, opened 2026-06-30
Avoid X=0 PumpAll AI choices

Rauglothgor:
## Problem

Issue #8957 reports the AI casting Exude Toxin with X = 0. That Omen is scripted as a `PumpAll` effect with `NumAtt$ -X`, `NumDef$ -X`, and `SVar:X:Count$xPaid`.

## Minimal Fix

For `PumpAll` abilities that use paid X in `NumAtt` or `NumDef`, set the root X value before calculating the pump amount. If the AI cannot pay a positive X, decline the play instead of evaluating the effect as X = 0.

## Validation

- `mvn -pl forge-ai -am -DskipTests compile`

## Known Limits

This is limited to `PumpAll` effects whose power or toughness change depends on paid X. It does not attempt to solve every X-spell decision path.

## AI Assistance Disclosure

I used AI assistance to inspect the issue/code path and draft this narrow change. I reviewed the diff and ran the validation above.

Refs #8957

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/PumpAllAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/PumpAllAi.java
@@ -63,8 +63,14 @@ public class PumpAllAi extends PumpAiBase {
             }
         }
 
-        final int power = AbilityUtils.calculateAmount(source, sa.getParam("NumAtt"), sa);
-        final int defense = AbilityUtils.calculateAmount(source, sa.getParam("NumDef"), sa);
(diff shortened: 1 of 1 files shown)


Agetian:
The approach seems valid to me. Not sure if any code reuse is possible off the top of my head (feels like we may have ways of setting paid X elsewhere, but can't check the code base atm and don't remember if there's a comprehensive one that goes all the way to root SA...), but overall looks workable to me. What do you think @Hanmac @tool4ever ?

Agetian (approved):

PR #11131, opened 2026-07-01
Fix auto-payment for reusable filter-land mana abilities

Madwand99:
This fixes a mana-planning blind spot where Forge ignored mana abilities that had mana in their activation cost. That meant reusable filter lands like Flooded Grove were not considered by available-action detection or auto-payment, so a hand containing Rampant Growth could be treated as unplayable even with a valid line through Plains -> Flooded Grove. This also prevents auto-yield from fast-forwarding when you have viable spells to play!

Changes:

- Allow AI/auto-payment mana discovery to consider mana-costed mana abilities when their activation costs use reusable resources, such as filter lands.
- Prefer multi-mana producers for colored costs when generic mana remains, so a filter land can produce mana that pays both the colored requirement and leftover generic cost.


forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
```diff
@@ -1781,10 +1791,11 @@ public static int determineLeftoverMana(final SpellAbility sa, final Player play
...
         final List<SpellAbility> res = new ArrayList<>();
         for (final SpellAbility a : c.getManaAbilities()) {
-            // if a mana ability has a mana cost the AI will miscalculate
-            // if there is a parent ability the AI can't use it
+            // if there is a parent ability the AI can't use it; mana costs are only
+            // considered when the activation cost is reusable, like filter lands.
             final Cost cost = a.getPayCosts();
-            if (cost.hasManaCost() || (a.getApi() != ApiType.Mana && a.getApi() != ApiType.ManaReflected)) {
+            if ((a.getApi() != ApiType.Mana && a.getApi() != ApiType.ManaReflected)
+                    || (cost.hasManaCost() && !cost.isReusuableResource())) {
```
tool4ever: no, there is zero logic to allow AI to calculate how it needs to spend mana for such filtering to work 👎

see #8455 for more details

Madwand99: Interesting! Did not know about this. So, I pushed a fix for autopayment/playability planning for simple mana abilities with mana activation costs, such as filter lands and Signets.

The mana planner now allows reusable tap mana abilities with exactly one mana in their activation cost, but explicitly plans/pays that activation cost before using the ability. It also prevents those mana-costed mana abilities from paying for each other recursively, so they are not treated as free mana. Hopefully this works as a fix, seems to pass all the tests I've thrown at it so far.

forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
```diff
@@ -169,6 +171,14 @@ private static void sortManaAbilities(final ListMultimap<ManaCostShard, SpellAbi
             }
 
             newAbilities.sort((ability1, ability2) -> {
+                if (!shard.isGeneric() && cost.getConvertedManaCost() > 1) {
+                    int amount1 = ability1.totalAmountOfManaGenerated(sa, true);
+                    int amount2 = ability2.totalAmountOfManaGenerated(sa, true);
+                    if (amount1 != amount2 && (amount1 > 1 || amount2 > 1)) {
+                        return amount2 - amount1;
```
tool4ever: besides the risk of messing with the comparator transitivity this leads to such dumb choices:
Image

Madwand99: Yeah, good point. I added a ton more test cases. Code should be able to handle a lot more cases now.

forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
```diff
@@ -1650,7 +1778,7 @@ private static ListMultimap<Integer, SpellAbility> groupSourcesByManaColor(final
                             Set<String> reflectedColors = CardUtil.getReflectableManaColors(m);
                             // find possible colors
                             for (byte color : MagicColor.WUBRG) {
-                                if (mp.canProduce(MagicColor.toShortString(color), tail) || reflectedColors.contains(MagicColor.toLongString(color))) {
+                                if ("Chosen".equals(origin) || mp.canProduce(MagicColor.toShortString(color), tail) || reflectedColors.contains(MagicColor.toLongString(color))) {
```
tool4ever: why does Chosen suddenly mean any color...?!

Madwand99: It was part of an attempt to make Nykthos work. I decided that was complex enough to be it's own PR and stop that attempt, but this artifact remains. I've refactored this reference.

forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
```diff
@@ -1044,6 +1163,15 @@ private static boolean canPayShardWithSpellAbility(ManaCostShard toPay, Player a
             return false;
         }
 
+        if ("Chosen".equals(m.getOrigProduced())) {
```
tool4ever: and what about this...?

Madwand99: Fixed! (and did a cleanup pass)

forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
```diff
@@ -1028,36 +1139,46 @@ private static boolean canPayShardWithSpellAbility(ManaCostShard toPay, Player a
...
                     return true;
                 }
             }
             return false;
         }
 
-        if (!sa.allowsPayingWithShard(sourceCard, MagicColor.fromName(m.getOrigProduced()))) {
-            return false;
+        for (String s : m.mana(ma).split(" ")) {
+            if (canPayShardWithProducedMana(toPay, ai, sa, sourceCard, s, xManaCostPaidByColor)) {
```
tool4ever: this isn't just clean up, you're making subtle logic changes again!

revert this last part again and make extra PR with a clear scenario it would fix

Madwand99: Ack! Sorry about that. Reverted and removed the Chosen block you were originally asking about.

tool4ever:
Sorry, I don't really think this can become mergeable with reasonable effort and this seems a bit too much for your AI to handle :/
I fail to see any real algorithm in your code or rather it's just: "try to pay for the filter land and hope for the best"

This already becomes evident with one of my first tests, here's AI trying to pay for *Vigor*:
grafik

but in good news looking at the code did lead to some other cleanup

Madwand99:
> Sorry, I don't really think this can become mergeable with reasonable effort and this seems a bit too much for your AI to handle :/ I fail to see any real algorithm in your code or rather it's just: "try to pay for the filter land and hope for the best"

So it's possible a more general algorithm may become possible with more work, I can investigate this. However: the current state of the code (without a patch like this one) is not acceptable. The auto-yield will just yield my entire turn if it thinks I have nothing to play, which it does if I have a filter land! This is just not acceptable! It is possible that this code is currently not able to handle every case (and it wasn't intended to) but it 1000% better than what existed before.

Madwand99:
I'm going to close this branch for now, and plan out a more algorithmic approach most likely using a best-first search mana payment planner.

PR #11138, opened 2026-07-02
MagicStack: store CastThisTurn as SpellAbility

Hanmac:
Part of #4700 and related to #10465

This makes castThisTurn into a `List<SpellAbility>` instead of a `List<Card>`

while the Card -> castSA still can be done via `CardPredicates.castSA` and other stuff,
`getActivationPlayer` might be different from `getController`?

also it's an easier access to `getMayPlay()`


tool4ever (approved):

Hanmac:
i wanted to change `getThisTurnCast` into a SpellAbility Valid Check, but that might need to be bigger changes for later

PR #11139, opened 2026-07-02
#11065 Targets for each player fix

### LINKED ISSUE #11065: Nils, Discipline Enforcer
**Describe the bug**
When Nils, Discipline Enforcer's ability triggers at end step, the AI can put all counters on its own creatures instead of on one creature for each player.

**To Reproduce**
1. Have AI cast Nils, Discipline Enforcer
2. go to end step with AI having multiple creatures
3. see that they put counters on their own creatures instead of 'up to 1 for each player'

**Expected behavior**
Nils, Discipline Enforcer should allow you to put a +1/+1 counter on one target for each player

**Screenshots**
the AI at end of turn put a counter on both Nils and Stuffy Doll 
Image

 - Version [e.g. 2.0.13-snapshot-06.22

steven-robert-eddy:
## Problem

Issue #11065 reports that AI targeting for Nils, Discipline Enforcer does not correctly respect `isForEachPlayer()` restrictions, allowing targets to be chosen incorrectly instead of selecting one valid permanent per player.

## Solution

Added logic to handle `isForEachPlayer()` targeting restrictions.

- **What changed?**
  - Added a new conditional block that:
    1. Iterates through each player in the game.
    2. Gets targetable cards controlled by that player.
    3. For opponent creatures, selects the best creature using `getBestAI()`.
    4. For AI-controlled creatures, selects the worst creature using `getWorstPermanentAI()` to minimize downside.
    5. One thing to note here is I am not exactly sure if this is the right implementation of which creature to target, please let me know what is expected here
    6. Adds exactly one selected target per player to the target list.

- **Why this approach?**
  - Explicitly iterating over each player ensures the AI respects the "for each player" targeting restriction.
  - Existing AI heuristics are reused to maximize value against opponents while minimizing negative effects on the AI's own permanents.

- **Why not alternatives?**
  - Random target selection would ignore strategic evaluation.
  - Selecting from a single combined pool could place multiple targets on one player's permanents, violating the intended targeting behavior.

## Validation

- `mvn -pl forge-ai -am -DskipTests compile`

- **Manual testing:**
  - AI with multiple creatures targets exactly one creature per player.
  - Multiplayer games correctly target one creature for each opponent.

## Known Limits

- Limited to `CountersPutAi` handling of `isForEachPlayer()` targeting.
- Applies only to abilities using the `isForEachPlayer()` restriction.
- Only affects targeting of battlefield permanents that satisfy the ability's targeting requirements.

## AI Assistance Disclosure

I used AI assistance to help analyze the issue, review the implementation, and draft this pull request description. I verified the implementation and performed the validation listed above.

Closes #11065

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
@@ -826,19 +826,17 @@ public class CountersPutAi extends CountersAi {
             if (sa.getTargetRestrictions().isForEachPlayer()) {
                 sa.resetTargets();
                 for (Player p : ai.getGame().getPlayers()) {
-                    CardCollection targetable = CardLists.getTargetableCards(p.getCardsIn(ZoneType.Battlefield), sa);
-                    if (!targetable.isEmpty()) {
-                        Card choice;
-                        if (p.isOpponentOf(ai)) {
-                            // prefer best opponent target (most relevant creature)
-                            choice = ComputerUtilCard.getBestAI(targetable);
-                        } else {
-                            // for self choose a low-value own permanent
-                            choice = ComputerUtilCard.getWorstPermanentAI(targetable, true, true, true, false);
-                        }
-                        if (choice != null) {
-                            sa.getTargets().add(choice);
-                        }
+                    CardCollection targetable = CardLists.canSubsequentlyTarget(new CardCollection(p.getCardsIn(ZoneType.Battlefield)), sa);
+                    Card choice;
+                    if (p.isOpponentOf(ai)) {
+                        // prefer best opponent target (most relevant creature)
+                        choice = ComputerUtilCard.getBestCreatureAI(targetable);
+                    } else {
+                        // for self choose a low-value own permanent
+                        choice = ComputerUtilCard.getWorstCreatureAI(targetable);
+                    }
+                    if (choice != null) {
+                        sa.getTargets().add(choice);
                     }
                 }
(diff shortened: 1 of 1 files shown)


forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
```diff
@@ -827,6 +822,28 @@ protected AiAbilityDecision doTriggerNoCost(Player ai, SpellAbility sa, boolean
...
             sa.resetTargets();
 
+            // Handle TargetsForEachPlayer: choose up to one target controlled by each player
+            if (sa.getTargetRestrictions().isForEachPlayer()) {
+                sa.resetTargets();
+                for (Player p : ai.getGame().getPlayers()) {
+                    CardCollection targetable = CardLists.getTargetableCards(p.getCardsIn(ZoneType.Battlefield), sa);
+                    if (!targetable.isEmpty()) {
+                        Card choice;
+                        if (p.isOpponentOf(ai)) {
```
steven-robert-eddy: I am not sure if this is the correct target for each player creature, please let me know if it is incorrect

tool4ever: just use `CardLists.canSubsequentlyTarget` in the targeting loop instead of extra logic block

tool4ever: doesn't look like your AI understood what I meant, I suggest you get to know this class a bit better first if possible 🤔

PR #11142, opened 2026-07-02
Compleated: bypass Multiset can't have negative amount

Hanmac:
With checking for Compleated and Loyalty counters, it bypasses that the amount can be negative for a short period of time


tool4ever (approved):
should work for now 🤔

Issue #8806, opened 2025-09-29
Better usable Desc for ValidTgts


Hanmac:
> @tool4ever for better Desc on `TgtPrompt` and `ValidTgts`, i think we should add something like `ValidTgtsDesc`.
> that works like in-between them:
> 
> * ValidTgts: `Creature.YouCtrl`
> * ValidTgtsDesc: `creature you control`
> * TgtPrompt: `Select target creature you control`
> 
> so in theory, we might get rid of TgtPrompt for most of the cases. 

 _Originally posted by @Hanmac in [#8646](https://github.com/Card-Forge/forge/issues/8646#issuecomment-3343405243)_

Devs: i'm looking for a better name than `ValidTgtsDesc` that hopefully doesn't conflict with other Names

The Goal is to use it later in GUI for: "CardName is not a ValidTgtsDesc"

@Eradev later, i probably need some help to check all 11k+ instances of TgtPrompt in the CardScripts and if they can be removed for new Parameter

tehdiplomat:
Why not just change all of the prompts to not have `Select target` and that just gets preceded in that target input?

Hanmac:
@tehdiplomat i'm not on code right now, i need to check how many of the `TgtPrompt` doesn't have/start with `Select target`

Some numbers:
* `TgtPrompt$ ` = 11k
* `TgtPrompt$ Select ` = 9.7k
* `TgtPrompt$ Select target` = 9k

that means that 1.3k doesn't start with Select, and 0.7k start with Select but doesn't use `Select target`

Jetz72:
The probably-too-ambitious dream solution would be a way to generate the target prompt text dynamically from the other parameters. Probably would involve some series of nested templates that get filled with text that reflects the number of targets and conditions in the Valid syntax. That'd eliminate the need for manual prompt text, with some increase in maintenance costs. It would make localization of prompt text more difficult in that it'd have to be implemented programmatically but less difficult in that we wouldn't need to map thousands of individual strings.

tehdiplomat:
Thats what it originally was :D Then there were so many corner cases that i think it just became common place to always add it.

Jetz72:
If the code for it is still there, then it might be worth looking into how many `TgtPrompt`s we could eliminate by fleshing it out a bit more.

Hanmac:
@tehdiplomat why not have the TgtPrompt use "Select target" as Prefix:

Special case "Any":

* the Valid: "Any"
[...]

PR #11148, opened 2026-07-04
PutCounterAll: update cards with ValidCardsDesc

Hanmac:
This adds ValidCardsDesc to PutCounterAll to make better StackDescription
(Spell and Activated Abilities)


forge-gui/res/cardsfolder/a/ajani_the_greathearted.txt
```diff
@@ -6,7 +6,7 @@ S:Mode$ Continuous | Affected$ Creature.YouCtrl | AddKeyword$ Vigilance | Descri
 SVar:NonStackingEffect:True
 SVar:PlayMain1:TRUE
 A:AB$ GainLife | Cost$ AddCounter<1/LOYALTY> | Planeswalker$ True | LifeAmount$ 3 | SpellDescription$ You gain 3 life.
-A:AB$ PutCounterAll | Cost$ SubCounter<2/LOYALTY> | ValidCards$ Creature.YouCtrl | CounterType$ P1P1 | CounterNum$ 1 | Planeswalker$ True | ValidCards2$ Planeswalker.YouCtrl+StrictlyOther | CounterType2$ LOYALTY | AILogic$ OwnCreatsAndOtherPWs | SpellDescription$ Put a +1/+1 counter on each creature you control and a loyalty counter on each other planeswalker you control.
+A:AB$ PutCounterAll | Cost$ SubCounter<2/LOYALTY> | ValidCards$ Creature.YouCtrl | ValidCardsDesc$ creature you control | ValidCardsDesc2$ other planeswalker you control | CounterType$ P1P1 | CounterNum$ 1 | Planeswalker$ True | ValidCards2$ Planeswalker.YouCtrl+StrictlyOther | CounterType2$ LOYALTY | AILogic$ OwnCreatsAndOtherPWs | SpellDescription$ Put a +1/+1 counter on each creature you control and a loyalty counter on each other planeswalker you control.
```
tool4ever, line 7: this feels wrong

should just use `StackDescription$ SpellDescription` instead of having to manually put these ugly extra params everywhere for literally the same outcome

Hanmac: but it is more flexible

tool4ever:
a) I still think it's pointless teaching scripters to use something like `ValidCardsDesc$ creature you control | ValidCardsDesc2$ other planeswalker you control` instead of `StackDescription$ SpellDescription` for everything 👎

b) how many of these cards are actually dynamic?
the very few could also just use `StackDescription$ REP X counters_{n:X}...`

Hanmac:
> a) I still think it's pointless teaching scripters to use something like `ValidCardsDesc$ creature you control | ValidCardsDesc2$ other planeswalker you control` instead of `StackDescription$ SpellDescription` for everything 👎

It's the same as using `ValidTgtsDesc` for the TargetDescription #8806 

> b) how many of these cards are actually dynamic?

You mean from the Valid Desc or from the Amount? For Amount, there are these examples:
* Wild Onslaught for example does 1 or 2 counters
* Vraska, Regal Gorgon does `For each creature card in your graveyard`

tool4ever:
> It's the same as using ValidTgtsDesc for the TargetDescription https://github.com/Card-Forge/forge/issues/8806

But for target prompt we don't really have another script param we can just reuse...

> You mean from the Valid Desc or from the Amount?

doesn't really matter, `SpellAbilityEffect#tokenizeString` can handle both

Hanmac:
But this is more ugly
> StackDescription$ REP X counters_{n:X}

tool4ever:
well ok, then just use the special params for these cases?

Hanmac:
> well ok, then just use the special params for these cases?

it is already used for this:
https://github.com/Card-Forge/forge/blob/f996041a28caeff7ec17ffe1f45beeecac843919/forge-game/src/main/java/forge/game/ability/effects/CountersPutAllEffect.java#L20-L28

tehdiplomat:
Yea i think I agree with the statement that this Desc field is kinda useless when StackDescription$ SpellDescription is cleaner. i.e. in cases where there's no variability. But i dont necessarily mind having the buildable description for those cases.

Hanmac:
Imo extra Desc is still more clean than SpellDescription

PR #11154, opened 2026-07-05
MagicStack: use Multimap for Spells Cast

Hanmac:
Still WIP

I'm refactoring how "ThisTurnCast" stuff is stored.
If it is done as ListMultimap, using Player as Map-Key, then we don't need extra counting.

and if we store ThisGameCast too, then we can clean up some other stuff.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/SetStateAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/SetStateAi.java
@@ -138,7 +138,7 @@ public class SetStateAi extends SpellAbilityAi {
             if (card.getState(CardStateName.Original).hasKeyword(Keyword.HIDDEN_AGENDA)
                     && card.isInZone(ZoneType.Command)) {
                 final String chosenName = card.getNamedCard();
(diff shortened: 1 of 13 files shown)


Hanmac:
@tool4ever @MostCromulent 
I want to find a way to copy the castThisTurn/LastTurn/ThisGame in GameCopier and GameSnapshot

for this i probably need to copy the LKI again to replace the oldPlayer with newPlayer (and maybe the host cards too?)

Hanmac:
@tool4ever your comment?

tool4ever:
A single card really shouldn't cause unlimited holding of references to all spells cast for the whole game:
that just creates extra memory pressure which isn't a worthy trade-off here

Hanmac:
@tool4ever it's similar to #10585 for #4700 
to handle `mayPlayed` better in the future.
(instead of having mayPlay on the Static which is problematic, have it on the MagicStack)

`Card->castSA->mayPlay` is less clean than `Spell->mayPlay`

Issue #3410, opened 2023-07-05
AI: Progressive slowdown when running complex predictions in a stalled battlefield state


schnautzr:
When Endless Swarm is in play, there is a risk of the game slowing to a standstill the longer it remains in play. The AI appears to be considering whether to activate mana abilities at times when there is nothing to pay for.

![image](https://github.com/Card-Forge/forge/assets/10385459/d312edc0-c108-4d18-8cd8-2537ebde7c8c)

schnautzr:
@Agetian

Agetian:
I'll definitely take a look at this one soon, I'll see what I can do.
This is definitely not specific to Endless Swarm in particular, you will see similar behavior with lots of token decks since at a certain threshold (dependent, of course, on the player's PC specs, especially RAM, but also CPU speed), the game will progressively start slowing down and will eventually grind down to a halt (or become unplayably bad). I've seen this sort of behavior with token decks revolving around multiple doublers such as Doubling Season and lots of token generators. The unfortunate thing is, the AI in its current design sort of has to run many checks for every single card object present on the battlefield. What seems obvious to the human player ("hey, the AI can't do anything here, what could it possibly be thinking about?") may not be so obvious for the AI algorithm which has to run checks for static abilities, presence of activated abilities that may be possible to activate, some predictions for combat, and many other things, some of which involve going over the entire battlefield and examining every card object. That said, there is, of course, room for optimization here, and I'll see what exactly is taking the AI so long in this particular case :) If I find any room for improvement, I'll handle it :)

At any rate, definitely not the case for an AI flag here - if we follow this logic, we would eventually come to a conclusion that we simply need to ban any form of token generation for the AI at all :D

Do you have a battlefield state where the AI is already slow for you? I might run a profiler to determine what's going on exactly.

schnautzr:
No, but I did take a screenshot:

![image](https://github.com/Card-Forge/forge/assets/10385459/ce597dd4-9400-42d2-80ee-22c55d13f449)

Agetian:
Alright, I'll see if I can reproduce it from that state :)
Can you check if this game state is slow for you? This runs adequately normal for me, with no slowdown. If it won't be slow for you either, it'll indicate that what you're experiencing is a more complex interaction than just having an Endless Swarm out and/or the presence of Elderscale Wurm on the human side. 

[ai_endless_swarm.txt](https://github.com/Card-Forge/forge/files/11953865/ai_endless_swarm.txt)

[...]

PR #11160, opened 2026-07-05
AI perf follow-ups: fog check caching, tap/untap scan zones, chump block re-evaluation

Blackvipe99:
Follow-up to #11157, continuing the AI slowdown work from #3410. Three changes, all targeting full-game card scans and redundant recomputation in the AI's decision hot paths.

### 1. Cache `isPreventCombatDamageThisTurn()` per game timestamp (`ReplacementHandler`)

Even zone-restricted (#11157), the fog check still runs once per attacker per hypothetical combat per candidate spell ability. The set of active fog-type effects only changes together with events that advance the game timestamp (zone changes — fog effect cards entering/leaving the command zone — and ability-granting effects), so the result is now cached and recomputed only when `game.getTimestamp()` changes. Fields are `volatile`; a benign race recomputes the same value. **Reviewer note:** this assumes any event that adds/removes a matching replacement effect also advances the game timestamp — true for the Effect-card lifecycle that fog effects use, but please sanity-check against edge cases I may not know about.

### 2. Skip hidden zones when scanning for Tap/Untap replacement effects (`ReplacementHandler.getReplacementList`)

Profiling AI timeouts (captured eval-thread stacks at the moment of timeout) showed the dominant remaining stall was not combat prediction but `Card.canTap`/`canUntap` → `cantHappenCheck` → `getReplacementList` → `forEachCardInGame`, reached thousands of times per AI decision via `ComputerUtilMana.groupSourcesByManaColor` (per mana source per cost check) and `CreatureEvaluator` (per creature evaluation). Sample stack:

```
at forge.game.card.Card.updateReplacementEffects(Card.java:7044)
at forge.game.replacement.ReplacementHandler.lambda$getReplacementList$0(ReplacementHandler.java:119)
at forge.game.Game.forEachCardInGame(Game.java:751)
at forge.game.replacement.ReplacementHandler.getReplacementList(ReplacementHandler.java:101)
at forge.game.replacement.ReplacementHandler.cantHappenCheck(ReplacementHandler.java:156)
at forge.game.card.Card.canTap(Card.java:4679)
at forge.game.cost.CostTap.canPay(CostTap.java:69)
...
at forge.ai.ComputerUtilMana.groupSourcesByManaColor(ComputerUtilMana.java:1573)
```

For `ReplacementType.Tap`/`Untap` events the scan now skips cards in zones outside `STATIC_ABILITIES_SOURCE_ZONES` — no tap/untap-replacing effect functions from a library or hand, and the existing `zonesCheck` already rejected them after visiting (and after paying `getReplacementEffects()`'s rebuild cost per card). In multiplayer Commander this cuts each of those scans by the ~350+ cards sitting in libraries.

### 3. Avoid redundant `lifeInDanger` re-evaluation in `makeChumpBlocks` (`AiBlockController`)

`makeChumpBlocks` re-ran `lifeInDanger` (a full combat damage prediction over all attackers) at every recursion level — once per attacker — making block assignment O(attackers² × board). The recursion now only re-evaluates when a blocker was actually assigned since the last check; skipping an attacker leaves the combat unchanged, so the previous result still holds. Decisions are identical by construction.

### Testing

* Full build + headless sims: constructed 1v1 and 4-player FFA with token-swarm decks and a fog stall deck — games complete normally, fog effects verified applying to combat damage, no exceptions.
* Benchmark: 4-player Commander mirror of a Krenko, Mob Boss goblin deck (worst-case wide boards), `MATCH_AI_TIMEOUT=3`: before these changes the AI hit the timeout during the match; after, zero timeouts across repeated full games.
* Real 5-player Commander play with these patches: no AI decision exceeded the 3s timeout over a full game session.

### DIFF
--- a/forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
+++ b/forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
@@ -42,6 +42,7 @@ import forge.game.ability.ApiType;
 import forge.game.player.Player;
 import forge.game.player.PlayerCollection;
 import forge.game.spellability.AbilitySub;
+import forge.game.spellability.Spell;
 import forge.game.spellability.SpellAbility;
 import forge.game.zone.Zone;
 import forge.game.zone.ZoneType;
@@ -102,12 +103,15 @@ public class ReplacementHandler {
             Card c = preList.get(crd);
             Zone cardZone = game.getZoneOf(c);
 
-            // tap/untap events can only be replaced by effects active from open zones;
-            // don't scan libraries and hands for them - this is a major hot path, as
-            // canTap/canUntap run a cantHappenCheck per mana source per AI cost check
-            if ((event == ReplacementType.Tap || event == ReplacementType.Untap)
+            // all tap/untap replacements are active from the battlefield or the command zone
+            // (e.g. Ood Sphere); skip other zones - this is a major hot path, as canTap/canUntap
+            // run a cantHappenCheck per mana source per AI cost check
+            // (performance mode only, in case a custom card wants one active from elsewhere)
+            if (Spell.isPerformanceMode()
+                    && (event == ReplacementType.Tap || event == ReplacementType.Untap)
                     && cardZone != null
-                    && !ZoneType.STATIC_ABILITIES_SOURCE_ZONES.contains(cardZone.getZoneType())) {
+                    && cardZone.getZoneType() != ZoneType.Battlefield
+                    && cardZone.getZoneType() != ZoneType.Command) {
                 return true;
             }
 
--- a/forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
+++ b/forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
@@ -919,20 +919,6 @@ public class ReplacementHandler {
      * @return true if there is some resolved fog effect
      */
     public final boolean isPreventCombatDamageThisTurn() {
-        // cache per game timestamp: the set of active fog effects only changes together
-        // with events that advance it (zone changes, ability grants), and this check
-        // sits in the AI's combat prediction hot path
-        final long ts = game.getTimestamp();
-        if (preventCombatDamageTimestamp != ts) {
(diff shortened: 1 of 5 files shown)


forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
```diff
@@ -910,6 +919,20 @@ public boolean visit(Card c) {
      * @return true if there is some resolved fog effect
      */
     public final boolean isPreventCombatDamageThisTurn() {
+        // cache per game timestamp: the set of active fog effects only changes together
```
tool4ever: use `AiCache.getCached`
(might need to move some methods to AI module so engine stays free of dependency)

forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
```diff
@@ -102,6 +102,15 @@ public List<ReplacementEffect> getReplacementList(final ReplacementType event, f
             Card c = preList.get(crd);
             Zone cardZone = game.getZoneOf(c);
 
+            // tap/untap events can only be replaced by effects active from open zones;
+            // don't scan libraries and hands for them - this is a major hot path, as
+            // canTap/canUntap run a cantHappenCheck per mana source per AI cost check
+            if ((event == ReplacementType.Tap || event == ReplacementType.Untap)
```
tool4ever: let's just guard it by `Spell.performanceMode` on the off chance some custom card needs it to work:
then you can also make the constraints even tighter since only *Ood Sphere* should care for it?

Blackvipe99: Done in 0c9c6da — the skip now only applies when performance mode is on, and the allowed zones are tightened to Battlefield + Command (grepped cardsfolder: every `Event$ Tap/Untap` RE is `ActiveZones$ Battlefield` except Ood Sphere's goad effect and the Edge of Malacol / Imprison This Insolent Wretch planechase/archenemy ones, which all sit in Command).

tool4ever (approved):

PR #11164, opened 2026-07-05
Serialize commander bracket to fix network client game-start crash

### LINKED ISSUE #11135: Multiplayer game doesnt start due to nullpointer
**Describe the bug**
When trying to enter multiplayer game from the multiplayer lobby as player(not host), the host starts the game but my screen remains in the lobby (i hear the game sounds like drawing cards).
When checking the game logs, i see a nullpointer exception because  getMatch() is null.

**To Reproduce**
1. Go to 'Multiplayer Lobby'
2. Click on 'Ready'
3.Wait for host to start

**Expected behavior**
The game should remain stuck on the lobby, in my case only for me and not the other friends in the lobby

**Screenshots**

 - OS: [Windows 11]
 - Version [2.0.13-SNAPSHOT-06.22]

**Additional context**
Earlier in the day i played fine in multiplayer, then tonigth i cant start any multiplayer game.

MostCromulent:
Closes Card-Forge/forge#11135.

## Summary

Fixes a crash that stranded non-host players on the lobby screen when a network Commander game started.

## Bug report

The host started the game, but the client stayed in the lobby while the match ran on the server (card sounds audible, no board). The client log showed a `NullPointerException` during `openView`:

    NullPointerException: Cannot invoke "forge.game.Match.getPlayers()"
        because the return value of "forge.game.GameView.getMatch()" is null
      at forge.game.GameView.getDeck(GameView.java:256)
      at forge.screens.match.views.VField.getCommanderBracketTooltipLine(VField.java:457)
      at forge.screens.match.views.VField.getPlayerDetailsHtml(VField.java:423)
      at forge.screens.match.views.VField.updateDetails(VField.java:418)
      at forge.screens.match.views.VField.<init>(VField.java:162)
      at forge.screens.match.CMatchUI.initMatch(CMatchUI.java:388)
      at forge.screens.match.CMatchUI.openView(CMatchUI.java:1225)

## Root cause

The commander-bracket tooltip added in #10803 calls `GameView.getDeck(player)`, which dereferences `GameView`'s `transient`, server-only `match` field — null on every network client by design (registered decks live only on the host). The tooltip NPEs as the client builds its `VField`s, aborting match-screen setup. It only triggers when the client's `DECKGEN_MAXIMUM_COMMANDER_BRACKET` is 1–4; the default short-circuits before the deck is touched.

## Fix

The host computes each player's bracket at game start and publishes it as a serialized `PlayerView` property, which the tooltip reads instead of the deck. `VField` no longer calls `getDeck`, so the crash is gone and the bracket displays on clients in parity with the host.

## Why this is architecturally consistent

- Mirrors how `GameView` already serializes its other match-derived data (title, games-played, match-over) as `TrackableProperty` values, instead of reaching through the transient `match` field.
- Only the resulting integer crosses the wire — no decklist — so no hidden information is exposed. Computed host-side, where `CommanderBracketCalculator` (a `forge-gui` class `forge-game` can't depend on) lives.

---

### DIFF
--- a/forge-game/src/main/java/forge/game/GameView.java
+++ b/forge-game/src/main/java/forge/game/GameView.java
@@ -253,6 +253,9 @@ public class GameView extends TrackableObject {
     }
 
     public Deck getDeck(final PlayerView pv) {
+        if (getMatch() == null) { // transient, null on network clients after deserialization
+            return null;
+        }
         for (final RegisteredPlayer rp : getMatch().getPlayers()) {
             if (pv.isLobbyPlayer(rp.getPlayer())) {
                 return rp.getDeck();
--- a/forge-game/src/main/java/forge/game/player/PlayerView.java
+++ b/forge-game/src/main/java/forge/game/player/PlayerView.java
@@ -180,6 +180,13 @@ public class PlayerView extends GameEntityView {
         set(TrackableProperty.Life, p.getLife());
     }
 
+    public int getCommanderBracket() {
+        return get(TrackableProperty.CommanderBracket);
+    }
+    public void setCommanderBracket(final int bracket) {
+        set(TrackableProperty.CommanderBracket, bracket);
+    }
+
     public boolean getIsExtraTurn() {
         return get(TrackableProperty.IsExtraTurn);
     }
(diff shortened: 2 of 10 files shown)


tool4ever:
looks like the same crash happens if remote client calls `viewDeckList()` from Game Menu

can't we better access GameLobby since both sides should already be able to reach decks this way?

Hanmac (dismissed):

tool4ever (approved):

PR #11167, opened 2026-07-06
Add Randomizer & Archipelago functionality to Forge Adventure

BramTeurlings:
### Randomizer/Archipelago
`NewGameScene.java`
- New randomizer option
- Added descriptions for each mode
- UI Also works in portrait mode/mobile
- Made the options scrollable, they are dynamically built in code now instead of statically from the layout file

`forge/adventure/archipelago`
- Contains all new files for Archipelago
- `ArchipelagoData.java` > Responsible for managing & persisting game data & common logic relevant for either randomizer
- `LocalRandomizer.java` > Responsible for managing game logic related to the local randomizer mode
- `ArchipelagoRandomizer.java` > Responsible for managing game logic related to the archipelago randomizer mode
- `Archipelago.java` > Manages the network connection to the Archipelago server

Generic tweaks in various forge-adventure files
- Most changes will start with a check for the current Archipelago mode, if the mode is disabled, there will be no change in behaviour from the base-game
- I have removed the Leather Boots as starting equipment in Archipelago mode since it auto-fills one of the user's slots with an item and that makes it less exciting for them to find an item for that slot later since they would be losing the benefits from the Leather Boots. Instead, I have simply increased the base movement of all entities in Archipelago mode to speed up the game and clear that boot slot.
- A new Archipelago icon/sprite was added to the game files, the sprite was made in paint by one of our contributors and the image is free to use from Archipelago's GitHub.
- An Archipelago connection screen was added, it can be seen in the Adventure mode's main menu but only if the currently loaded save file has Archipelago mode enabled.
- World generation was altered so that important POI's for Archipelago checks always fall fully within the region they are supposed to spawn in, not on a island of the correct colour surrounded by a neighbouring colour as that could break randomizer logic. This change only applies in Archipelago mode.

Overhauled input system > https://github.com/Card-Forge/forge/pull/11168
 - The base-game's input system currently suffers from multiple input methods fighting for control. One particular issue that prompted me to overhaul it was the angle the player walks in when inputting a diagonal direction on keyboard. It's not 45 degrees like you would expect, it's actually close to 22.5 degrees because one of the input methods correctly reads 45 degrees but the logic is executed a second time on one of the pressed keys (not both) resulting in a combined angle of 22.5 degrees. Further issues include one of the pressed keys being ignored until re-pressed if pressed too quickly in conjunction with other movement keys.
 - The new input system defines a clear priority between keyboard/mouse (or touch)/controller inputs so they no longer fight for control.
[...]

PR #11168, opened 2026-07-06
Feature/11167: Revamped input handling + better road detection

BramTeurlings:
Split off from: [Archipelago PR](https://github.com/Card-Forge/forge/pull/11167)

### Overhauled input system

- The base-game's input system currently suffers from multiple input methods fighting for control. One particular issue that prompted me to overhaul it was the angle the player walks in when inputting a diagonal direction on keyboard. It's not 45 degrees like you would expect, it's actually close to 22.5 degrees because one of the input methods correctly reads 45 degrees but the logic is executed a second time on one of the pressed keys (not both) resulting in a combined angle of 22.5 degrees. Further issues include one of the pressed keys being ignored until re-pressed if pressed too quickly in conjunction with other movement keys.
- The new input system defines a clear priority between keyboard/mouse (or touch)/controller inputs so they no longer fight for control.
- The new input system stores each of its directions/vectors in a separate value so no unintentional overlapping of directions occur
- Fixed a crash where the console would attempt to fetch a String for a list on an index which doesn't exist. Very easy to trigger after making a new game, just open the console and press arrow down and arrow up.

image

- Made the console clear the text field when pressing the down arrow without a more recent command in the console's history. This more closely mimics what Linux/Windows terminals do.

### Better road detection
- `forge-gui-mobile/src/forge/adventure/stage/WorldStage.java#L269` > This shifts the detection point for which biome the player is walking on to the centre of the sprite instead of the left-most pixel. The vertical axis remains unchanged at the bottom.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/stage/GameStage.java
+++ b/forge-gui-mobile/src/forge/adventure/stage/GameStage.java
@@ -252,7 +252,6 @@ public abstract class GameStage extends Stage {
         showDialog();
     }
 
-
     public boolean axisMoved(Controller controller, int axisIndex, float value) {
         if (MapStage.getInstance().isDialogOnlyInput() || isPaused()) {
             return true;
@@ -267,7 +266,6 @@ public abstract class GameStage extends Stage {
         Fly
 
     }
-
 
     HashMap<PlayerModification, Float> currentModifications = new HashMap<>();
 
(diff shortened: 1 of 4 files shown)


forge-gui-mobile/src/forge/adventure/stage/GameStage.java
```diff
@@ -371,28 +370,75 @@ public final void act(float delta) {
...
+                    keyboardInput.y = 1;
+                    break;
+                }
+            }
+            for (int key : KeyBinding.Down.getBindings()) {
+                if (Gdx.input.isKeyPressed(key)) {
+                    keyboardInput.y = -1;
+                    break;
+                }
+            }
```
Jetz72: Looping over a `KeyBinding`'s binding list to test it against `isKeyPressed` seems like it'd make sense to have as a helper method on the `KeyBinding` itself. Kinda surprised that isn't what the `isPressed` method on that enum is for.

Also, tiny suggestion: maybe make it so pressing both keys simultaneously cancels out the movement rather than preferring one direction arbitrarily?

BramTeurlings: Those are great observations. I've added a new helper method `isPressed()` to `KeyBinding.java` as you suggested. I also removed the `getBindings()` method I added earlier as it is not longer necessary or used.

I've now made the `keyboardInput` vectors cumulative such that opposing directions pressed indeed cancel each other out.

[a631d4f](https://github.com/Card-Forge/forge/pull/11168/commits/a631d4fc4e41468d24d025fa1abd9215a03b102d)

Jetz72 (approved):

tehdiplomat:
Reverting just in case this broke android builds.

tehdiplomat:
@BramTeurlings FYI, this broke adventure mode on Android devices from being able to enter towns or start battles.

BramTeurlings:
Ah, I've seen this bug and fixed it on Windows but I wasn't yet able to figure out how to build for Android. I'll see if I can figure that out and re-open a PR with a fix. Sorry for the inconvenience. 
Does anyone happen to know what the easiest way is to build Android locally btw?

PR #11170, opened 2026-07-06
Feature/11167: Addition of cost for each equipment in Shandalar

BramTeurlings:
Split off from: [Archipelago PR](https://github.com/Card-Forge/forge/pull/11167)

### Addition of cost for each piece of equipment

- The Archipelago randomizer will also randomize equipment available for purchase in shops from the total pool available in the Shandalar world, instead of randomizing the prices of item's whose cost is missing, I instead added a cost to each of them that I think would be reasonable based on other items in the game and the item's power level.

### DIFF
--- a/forge-gui/res/adventure/common/world/items.json
+++ b/forge-gui/res/adventure/common/world/items.json
@@ -365,6 +365,7 @@
     "description": "Draft a Chandra themed card that can be cast with mana of any color",
     "equipmentSlot": "Left",
     "iconName": "ChandrasTome",
+    "cost": 8000,
     "effect": {
       "startBattleWithCardInCommandZone": [
         "Chandra's Tome"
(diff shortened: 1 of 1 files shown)


tehdiplomat:
These equipments had costs removed because they were deemed too powerful, but straight removing them from the list might break saves for people who already have them.

BramTeurlings:
> These equipments had costs removed because they were deemed too powerful, but straight removing them from the list might break saves for people who already have them.

The costs for these items is not used anywhere in the code apart from inside of the [future randomizer pull request](https://github.com/Card-Forge/forge/pull/11167) where they can be enabled using a setting in Archipelago's `.yaml` config file only. I asked around on the Discord and was told that adding pricing for these items would not cause any issues. I agree that removing them is not a good solution so instead I suggest we give them a use inside of the randomizer and don't reference them anywhere else.

Furthermore, if this item *were* to be put in a shop, the game would assign it the default cost of 1000 as seen here. Arguably, having a cost closer to what the item's power level actually should be is a safer solution than only removing the cost.
image


Does that work for you or should we add some type of additional parameter that marks this items as "unused" or something similar?

tehdiplomat (approved):

PR #11173, opened 2026-07-07
Autotap - Add support for filter abilities

Eradev:
## What's new?

The auto-tapper is now an intelligent planner. This should hopefully avoid obvious mistakes and allow a better use of utility lands.

* Prioritize permanent sources rather than disposable ones (e.g. Lotus Petal).
* Prioritize colorless mana for generic mana, unless you have a {C} heavy hand. (e.g. Eldrazi)
* Added support for filter abilities with generic cost (e.g. {1}: Add {R}.), and filter lands (e.g. Cascade Bluffs). Prevents chaining if not needed, and should avoid obvious pitfalls (e.g. uselessly activate Initiates of the Ebon Hand).
* Added support for Mana Flare and Utopia Sprawl effects.
* Added support for SVar:AIManaReserve:True for utility lands. This makes them less likely to be chosen by the auto-tapper. (Need to add it to more card rules)
* Castability probe: Detection of remaining cards in hand/command zones to select which assets to use first to try and maximize plays. This is toggleable for performance in options (Desktop and mobile). Default is on.

* Added multiple tests for edge cases.

## Debugging

You can add the VM option `-Dforge.debugManaPayment=true` to help visualize what it attempts to use, or `-Dforge.debugManaPayment=true` for an easier to read plan (recommended).

Let me know if there are edge cases it misses.


tool4ever (changes requested):
Hopefully this'll work better than the last AI attempt 🤔

Anyway this is huge and the highlighting part is already a separate feature: #5112

Please extract that into its own PR first, it'll certainly be useful for testing the rest of this 👍

Eradev:
> Hopefully this'll work better than the last AI attempt 🤔
> 
> Anyway this is huge and the highlighting part is already a separate feature: #5112
> 
> Please extract that into its own PR first, it'll certainly be useful for testing the rest of this 👍

Moved to #11176

forge-gui-desktop/src/test/java/forge/ai/controller/AutoPaymentTest.java
```diff
@@ -143,4 +192,567 @@ public void testKeepColorsOpen() {
...
+    // {R}{W} with Mountain + Plains + Signet should consolidate onto one basic + Signet (rule A):
+    // exactly one basic tapped, Signet used, one basic left untapped.
+    @Test
+    public void consolidatesTwoColoredShardsOntoSignet() {
+        Game game = initAndCreateGame();
+        Player p = game.getPlayers().get(1);
+
+        addCard("Mountain", p);
+        addCard("Plains", p);
+        addCard("Boros Signet", p);
```
tool4ever, line 334: after we're done with most code concerns need to remove `AI:RemoveDeck:All` from all applicable cards so I can batch fuzz random generated decks for crashes

forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
```diff
@@ -590,15 +1645,23 @@ public static String predictManafromSpellAbility(SpellAbility saPayment, Player
 
     // returns null if unpayable
     private static List<Mana> payManaCost(final ManaCostBeingPaid cost, final SpellAbility sa, final Player ai, final boolean test, boolean checkPlayable, boolean effect) {
+        final int depth = manaPaymentDepth.get() + 1;
+        manaPaymentDepth.set(depth);
+        final boolean outermost = depth == 1;
+        try {
```
tool4ever, line 5074: this will need a cleaner solution than this massive wrap for which you're trying to ignore the code style

at the same time a `ThreadLocal` seems overkill and only risks wrong usage of such non obviously "constrained in concurrency" class

why not simply some overload for a nested entrypoint instead?

forge-gui/res/cardsfolder/c/celestial_colonnade.txt
```diff
@@ -5,4 +5,5 @@ R:Event$ Moved | ValidCard$ Card.Self | Destination$ Battlefield | ReplacementRe
 SVar:ETBTapped:DB$ Tap | Defined$ Self | ETB$ True
 A:AB$ Mana | Cost$ T | Produced$ Combo W U | SpellDescription$ Add {W} or {U}.
 A:AB$ Animate | Cost$ 3 W U | Defined$ Self | Power$ 4 | Toughness$ 4 | Types$ Creature,Elemental | Colors$ White,Blue | OverwriteColors$ True | Keywords$ Flying & Vigilance | SpellDescription$ Until end of turn, this land becomes a 4/4 white and blue Elemental creature with flying and vigilance. It's still a land.
+SVar:AIManaReserve:True
```
tool4ever, line 8: please try to solve only one problem with this PR
this size is completely unreasonable besides manlands being something that shouldn't need manual tagging

forge-ai/src/main/java/forge/ai/simulation/SpellAbilityPicker.java
```diff
@@ -359,6 +359,10 @@ public Score evaluateSa(final SimulationController controller, PhaseType phase,
             simulator.setInterceptor(choicesIterator);
             // I feel like something here is making a wrong assumption about what the target is
             lastScore = simulator.simulateSpellAbility(sa);
+            final int castBonus = ComputerUtilMana.estimateCastBonusForSpell(sa, player);
```
tool4ever, line 362: are you having a LLM go wild?

this is completely out of scope

forge-gui/res/cardsfolder/c/celestial_colonnade.txt
Eradev: I will separate them more cleanly.

PR #11176, opened 2026-07-07
Autotap - Highlight affected cards

Eradev:
Added a highlight around cards that will be used for auto-tap.

Related issue: https://github.com/Card-Forge/forge/issues/5112


forge-gui/src/main/java/forge/gamemodes/match/input/InputPayMana.java
```diff
@@ -441,6 +445,37 @@ protected void onManaAbilityPaid() {} // some inputs overload it
...
+        if (!canPayManaCost) {
+            getController().getGui().clearAutoTapPreviewCards();
+            return;
+        }
+
+        final ManaCostBeingPaid costCopy = new ManaCostBeingPaid(manaCost);
+        Evaluator<CardCollection> proc = new Evaluator<CardCollection>() {
+            @Override
+            public CardCollection evaluate() {
+                return ComputerUtilMana.getManaSourcesToPayCost(costCopy, saPaidFor, player);
```
tool4ever: seems rather unfortunate we'd have to call this method in addition to `canPayManaCost`

imo you could just refactor the latter so it always returns a CardCollection on success and null otherwise, then you can reuse it here for free 🤔

forge-gui/src/main/java/forge/gamemodes/match/AbstractGuiGame.java
```diff
@@ -359,6 +359,23 @@ public boolean isWeaklySelectable(final CardView card) {
         return weaklySelectableCards.contains(card);
     }
 
+    private final Set<CardView> autoTapPreviewCards = Sets.newHashSet();
```
tool4ever: a) I don't like how this duplicates all this extra tracking for the fourth time
ideally we're able to simply reuse WeaklySelectable or even Highlighted - does it actually look reasonable if the "actionable" cards are displayed at the same time? 🤔

or otherwise I wouldn't mind something like a `MultiSet` approach for `WeaklySelectable` so we can provide the GUI with more context on how "strong" the selectability actually is, which might come in handy for other features down the road too 💡

b) imo a preference and/or hotkey seems needed anyway since not every user will want free AI help on all the time

tool4ever (approved):

PR #11190, opened 2026-07-09
iOS support: JvmDowngrader build pipeline (MobiVM) + platform fixes

shoeless:
## Summary

Adds working **iOS support** for Forge. Instead of hand-backporting modern Java to MobiVM's Java‑7‑era runtime, it transforms the built jars so **unmodified upstream Java runs on iOS**:

1. **JvmDowngrader (`-c 52`)** rewrites Java 9–17 bytecode (records, `List.of`, `String.repeat`, switch expressions, string‑concat indy, …) down to Java 8.
2. **MobiVmBridge** (a small ASM tool) remaps the Java‑8 library surface MobiVM lacks onto backports: `streamsupport` (`java.util.function`/`stream`/`Optional`), relocated ThreeTen as `java.time`, a hand‑built `java.nio.file` supply, and `forge.compat`.
3. **MobiVmLinkAudit** is the merge gate — it resolves every `java`/`javax` member reference in the final jars against the real runtime + supplies, so an upstream merge that introduces a new Java‑8 gap is caught at build time rather than as a device crash.

The pipeline lives under `forge-gui-ios/pipeline/` (`ios-pipeline.sh [classpath|sim|device]`) and bootstraps its third‑party jars automatically.

## iOS platform fixes

- **Audio** — enable music + sound effects; play in silent mode; fix mono‑MP3‑as‑static; fix a classic‑mode music crackle by routing effects through AVAudioPlayer (single audio clock domain) and suspending the otherwise‑unused OpenAL engine.
- **Rendering** — fix device‑only red/blue color corruption at the source (`skipPngCrush` per‑resource in `robovm.xml`).
- **Adventure mode** — now runs on iOS (gamepad dep, scene null‑guards, TenPatch force‑link, Box2D natives).
- **Input** — keyboard fixes (delete key + duplicate characters, delivered via `keyTyped` on iOS).
- **Online/multiplayer** — bundle JvmDowngrader `util/*` runtime helpers; guard `ProcessHandle`/UPnP for MobiVM gaps.
- **Serialization** — work around RoboVM's AOT lambda plugin dropping `FLAG_SERIALIZABLE` (was crashing games with interacting static abilities) + `ServiceLoader` entry remapping (fixes time‑zone data registration).

Shared/upstream source is otherwise unchanged except a few genuinely‑iOS platform hooks (adventure resource path when `Gdx.app.getType()==iOS`, a `TextRenderer` `BreakIterator` fallback, and an `IOSAdapter` audio‑format override).

## Testing

Device‑verified on physical iPads (iOS 16 and iOS 26): boots, plays full games vs. AI, adventure mode, audio, and online hosting (online hosting is buggy will fix in upcoming PR). Also runs in the iOS Simulator (same RoboVM build).

## Notes for reviewers

- Developer‑specific config (bundle id, code‑signing identity, device UDIDs) is **not** tracked — it lives in an untracked repo‑root `.env`; `robovm.properties` is generated from `robovm.properties.template` on first pipeline run.
- This is a large, self‑contained addition (build pipeline + iOS backend); happy to split it if that helps review.


tool4ever:
Nice work, obviously rather huge but hardly unexpected.

A bit unfortunate there's still a few workarounds needed in the neighbour modules but tbh Android is probably just as guilty of that too 😆

Anyway three main points we need to think about:
a) User delivery - not sure this would be worth the effort to maintain if jailbroken device was a requirement...
would this also work if we provide instructions for something like *Sideloadly*?

b) CI/CD - can your pipeline be integrated into our normal GitHub actions? if it's build time is way longer than the other platforms maybe only as release though

c) Not having Sentry is a real loss, any chance to get that working somehow?

tool4ever:
I'm gonna do a manual review after you're done with those Copilot findings 👍

Would you be able to support this port further (probably users will find more bugs) in case we decide to merge it?

tool4ever:
Normal `test-build` still fails on your branch:
> Error:  Failed to execute goal on project forge-gui-ios: Could not resolve dependencies for project forge:forge-gui-ios:jar:2.0.14-SNAPSHOT
Error:  dependency: forge.stubs:java-function-stubs:jar:1.0 (compile)
Error:  	Could not find artifact forge.stubs:java-function-stubs:jar:1.0 in central (https://repo.maven.apache.org/maven2)
Error:  dependency: xyz.wagyourtail.jvmdowngrader:jvmdowngrader-java-api:jar:1.3.6-mobivm (compile)
Error:  	Could not find artifact xyz.wagyourtail.jvmdowngrader:jvmdowngrader-java-api:jar:1.3.6-mobivm in central (https://repo.maven.apache.org/maven2)
Error:  dependency: forge.stubs:java-time-supply:jar:1.0 (compile)
Error:  	Could not find artifact forge.stubs:java-time-supply:jar:1.0 in central (https://repo.maven.apache.org/maven2)
Error:  dependency: forge.stubs:java-nio-supply:jar:1.0 (compile)
Error:  	Could not find artifact forge.stubs:java-nio-supply:jar:1.0 in central (https://repo.maven.apache.org/maven2)

I suppose because a different workflow wouldn't find any cached artifacts?
is there a maven trick to ignore them?
since for normal development we don't need to see their shading anyway?

shoeless:
Thanks — catching up on your review, and I fixed the `test-build` failure you flagged.

**The `test-build` failure:** good catch — the iOS module shouldn't be in the default reactor at all. It depends on artifacts the iOS pipeline generates into the local repo (the streamsupport / JvmDowngrader / relocated-ThreeTen supplies) that aren't on Maven Central, so a plain `mvn install` — and your `test-build` — can't resolve them. I've moved `forge-gui-ios` out of the default `<modules>` into a non-default `ios` profile:

```xml
<profiles>
  <profile>
    <id>ios</id>
    <modules>
      <module>forge-gui-ios</module>
    </modules>
  </profile>
</profiles>
```

So `mvn install` / `test-build` no longer touch it (and never see the shaded supplies — exactly as you said, normal development doesn't need them). The iOS build is unaffected: the pipeline builds the module directly (`cd forge-gui-ios`), never through the root reactor, and the compat-gate workflow calls the pipeline rather than a reactor build. Anyone who wants it in a reactor build can `mvn -Pios ...`. Pushed.

---

On your three original points:

**a) User delivery — good news: no jailbreak is needed at any point.** Three working paths, in increasing order of convenience:

- **Sideloadly / AltStore**: they take an *unsigned* IPA and re-sign it with the user's own (free) Apple ID before installing. I've added a `Build iOS IPA` workflow (macOS runner, manual dispatch) that produces exactly that unsigned IPA as an artifact, so release IPAs can ship alongside the desktop/Android builds. Free-account sideloads expire after 7 days (re-sideload to refresh); a $99 developer account extends that to a year.
- **TestFlight**: works today — I've been distributing test builds to internal testers via TestFlight for months (build + upload is scripted). Needs an Apple developer account on the publishing side, but installs are one-tap for users and last 90 days.
- **Dev-signed direct install** for anyone building from source (Xcode free account).

**b) CI/CD — done, two tiers so the runtime cost lands where you suggested:**

- `ios-compat-gate.yml` (this PR): runs on **ubuntu** in a few minutes on any PR touching Java/poms. It executes the JvmDowngrader transform + link audit *without any Apple tooling* and fails when a change uses an API that would crash on iOS. This is the important one day-to-day: nobody needs a Mac to avoid breaking iOS.
- `ios-build-ipa.yml`: the full RoboVM AOT build on a macOS runner (~30-60 min, so manual-dispatch / release-time only per your suggestion), uploading the unsigned IPA artifact from (a). The RoboVM AOT cache is persisted between runs, which cuts rebuild times substantially.

**c) Sentry — done and device-verified.** The reason it was stubbed isn't a platform incompatibility: the `io.sentry` classes simply weren't bundled/linked into the iOS build, so every telemetry call from shared code threw `NoClassDefFoundError` (and killed the game thread on some choose-card paths — that's why it became a no-op stub rather than left broken). The real sentry-java (8.21.1, the same artifact desktop ships) now goes through the downgrade pipeline cleanly, `io.sentry.**` is force-linked, and the iOS launcher initializes it exactly like the desktop launcher (same DSN; events only send when the user's `USE_SENTRY` preference allows).

Verified on hardware (iPad, iOS 26): the previously-crashing choose-cards flows (Cultivate/Genesis Hydra library searches) work with the real SDK linked, and init/capture/transport all function. I have iOS reporting to the mobile project (2), matching Android, rather than desktop's project 3 — let me know if that's not how you'd want mobile events routed.

A couple of things I ran into while verifying delivery — quite possibly me holding it wrong, so I'd value your steer on how you want this set up:

1. **`sentry.asgardsrealm.net` was returning CF error 1033** ("tunnel not connected") for me across a few networks/clients over several hours, so my test events made it to the Cloudflare edge but no further. Might just be a transient on the ingest side, or I could be hitting the wrong host — is that the endpoint mobile should be using?
2. **The checked-in `sentry.properties` DSNs use `http://…:9000`**, whereas the `Main.java` fallback uses the plain-443 `https://…` form. I went with the `https` form for iOS (matching what desktop actually runs, since the properties file isn't packaged), but I wasn't sure which is canonical now — happy to switch iOS to whatever you'd prefer, and there's a small `…//3` double-slash in the desktop fallback if that's worth tidying.

iOS is wired to the mobile project's DSN (`https://…/2`, same form as Android's manifest) and verified end-to-end up to the edge, so events should flow once the ingest side is reachable — just point me at the right target if I've picked wrong.

---

On the **"workarounds in the neighbour modules"** aside: addressed. iOS-specific behavior in `forge-gui-mobile` is now explicitly gated behind a `GuiBase.isIOS()` flag (the exact mirror of the existing Android launcher flag) instead of ApplicationType sniffing, and in the process a few places where iOS-motivated changes would have silently altered Android/desktop behavior (file-path resolution, Scryfall fetch strategy, font texture filtering) were scoped back to iOS-only. The genuinely platform-neutral hardening (null guards, BreakIterator fallback) stays shared.

shoeless:
Addressed all the Copilot findings (plus one bonus simplification) — pushed as two commits:

**Copilot fixes:**
- **`FSkin.getFileHandle` writable paths** — correct catch. The iOS `internal()` reroute now applies only to paths under `ASSETS_DIR` (the app bundle); cache/Documents paths keep `absolute()`. This also turns out to make the generated `.fnt` font cache effective on iOS (it lives under the cache dir, so the lookup was silently missing and fonts were regenerating from TTF every launch) and un-breaks downloaded skins + planechase images there.
- **`LibGDXImageFetcher` wrong-exception log** — fixed (`t.getMessage()`); note the `e`-vs-`t` mixup actually predates this PR, the diff just touched the line.
- **`HostedMatch` per-game `System.gc()`** — now gated to iOS only. It's a deliberate mitigation for iOS's per-process memory (jetsam) ceiling on multi-game Commander matches; desktop/Android no longer pay the stall.
- **`OSLogPrintStream`** — both Unicode issues fixed: `print(String)` now buffers UTF-8 bytes instead of truncating chars, and `flushBuffer()` decodes explicitly as UTF-8.
- **`NetworkLogWriter` / `FServerManager` `catch (Throwable)`** — narrowed to `catch (LinkageError | Exception)`: still degrades gracefully when MobiVM lacks a class (`NoClassDefFoundError` is a `LinkageError`), no longer swallows `OutOfMemoryError` and friends.

**Bonus:** the Scryfall "can't follow 302 redirects" workaround in `LibGDXImageFetcher` turned out to be obsolete — an on-simulator probe showed RoboVM's HTTP stack follows the `api.scryfall.com` → `cards.scryfall.io` redirect fine (200, image bytes, correct final URL), and after deleting the ~120-line JSON-API resolution path, card images verifiably download end-to-end through plain redirect-following (fresh fetches → valid jpgs on disk → rendering). The file is now within ~20 lines of master (connection timeouts, a 0-byte-download guard, and the widened catches).

Verifying that also led me to the *real* image-fetch failure the workaround was mis-attributed to, and it turned out to be a bigger fish — a **file-descriptor exhaustion (EMFILE)** bug that I've now fixed (separate commit). iOS gives each process a 256-fd soft limit; the app's classpath jars get pinned open by ordinary classloader resource probes (~177 fds), a loaded match holds ~70 more, and the game-start image-fetch burst opens dozens of sockets on top. Once it tips past 256 everything fails at once, and none of it announces "EMFILE": lazy texture opens throw `GdxRuntimeException: Couldn't load file` (blank card frames / broken match board while the game keeps running), and `getaddrinfo` returns `Unable to resolve host` (it needs an fd) — which is why the Scryfall fetches looked like a DNS problem. The only honest signal was the nested `Caused by: … open failed: EMFILE` in the crash log. Fix is a one-liner in spirit: raise `RLIMIT_NOFILE` at launch (to `min(hard, OPEN_MAX)`). Verified in the simulator — a full match load plus wiped-cache fetch burst now runs with correctly rendered card frames and working downloads. This one was masking itself as several unrelated bugs (blank board, DNS failures, even audio glitches), so it's a good one to have out of the way.

shoeless:
> Would you be able to support this port further (probably users will find more bugs) in case we decide to merge it?

Yes — happy to. This port is what I play on weekly across two iPads.

tool4ever:
RE Sentry:
- sorry should have mentioned it's definitely down right now
- I think a new project probably makes sense, not sure we need to create it in the web GUI first though before changing it in code
- I'll investigate our definitive DSN format, not a block here

forge-gui-ios/robovm.xml
```diff
@@ -30,21 +56,79 @@
...
+    <pattern>com.thoughtworks.xstream.**</pattern>
+    <pattern>java.io.ObjectInputStream</pattern>
+    <pattern>java.io.ObjectOutputStream</pattern>
+    <pattern>java.io.ObjectStreamClass</pattern>
+    <pattern>java.io.ObjectStreamField</pattern>
+    <!-- NOTE: java.util.function is NOT available in RoboVM - use forge.util.function backports instead -->
+    <!-- No-op Sentry stubs (sentry-stub module): real io.sentry SDK is excluded on iOS, so
+         shared Forge telemetry calls (BugReporter/PlayerControllerHuman/Forge/etc.) resolve to
+         these. Force-linked because they're reached only via precompiled bytecode references. -->
+    <pattern>io.sentry.**</pattern>
```
tool4ever: sentry is below now already, this entry is outdated?

forge-gui-mobile/src/forge/assets/FSkin.java
```diff
@@ -31,6 +31,22 @@ public class FSkin {
...
+    /**
+     * FileHandle for a bundled resource. iOS must use internal() with a relative path (the app
+     * bundle lives behind a /var symlink that absolute() fails to realpath inside the sandbox);
+     * every other platform keeps the original absolute() behavior — Android's assets are
+     * extracted to storage and are NOT reachable via internal() APK paths.
+     * Only paths under ASSETS_DIR (the read-only bundle) are rerouted: writable locations
+     * (cache/Documents — downloaded skins, generated .fnt files, planechase pics) keep
+     * absolute(), which works fine in the iOS sandbox.
+     */
+    static FileHandle getFileHandle(String path) { //package-visible: FSkinFont shares this
```
tool4ever: is this method redundant to `Assets#getFileHandle`?

forge-gui-ios/robovm.xml
shoeless: Good catch — removed it, along with its stale "no-op Sentry stubs / real SDK excluded" comment (untrue since Sentry was enabled; the correct entry lives lower in the file). While there I also dropped the neighboring "java.util.function is NOT available — use forge.util.function backports" note: that's a hand-backport-era relic — the jvmdg bridge remaps `java.util.function` → streamsupport, so unmodified upstream uses it freely (155 files) and `forge.util.function` is used in 0. Pushed.

forge-gui-mobile/src/forge/assets/FSkin.java
shoeless: It was — consolidated into a single `Assets.getFileHandle`. I kept it in `Assets` (next to `HybridFileHandleResolver`, since file resolution is an Assets concern and it respects the existing `FSkin → Assets` direction), with the `startsWith(ASSETS_DIR)` guard so writable cache/Documents paths keep `absolute()` on iOS. `FSkin`/`FSkinFont` now call it; both the FSkin copy and the private Assets copy are gone. Pushed.

shoeless:
@tool4ever , thanks for the feedback so far. Anything I can do to simplify the review?

tool4ever:
I'm fine with the code quality, we're just trying to test on a real device too
Might be next weekend before I have time myself though

shoeless:
Sounds good. I would like help verifying this on an android device. I run this on two physical iPads: an iPad (10th gen) on iOS 26.2 and an older iPad on iOS 16.7, so it's covered across a wide OS range. Full games vs. AI, adventure mode, and online multiplayer all work, and I've had this exact branch on-device (audio, card rendering, the memory fixes) throughout.

forge-gui-mobile/src/forge/sound/AudioClip.java
```diff
@@ -6,26 +6,42 @@
...
+    // independent CoreAudio clients whose render cycles beat against each other, producing
+    // a slow periodic crackle in the co-summed music (~14-32s, un-fixable by rate-matching
+    // alone). Playing effects in the SAME AVAudioPlayer/mediaserverd domain as the music
+    // collapses everything to one clock and eliminates the beat; the now-unused OpenAL
+    // engine is suspended in Main (forge-gui-ios). Trade-off: a same-type effect can't
+    // overlap itself (one player per type) - re-triggers while still playing are dropped
+    // rather than truncating the sound. Evaluated per-instance (not a class-load-time
+    // constant) so it can never latch the wrong value and strand effects on the suspended
+    // OpenAL engine.
+    private final boolean ios;
```
Jetz72, line 41: Rather than a separate branch for iOS in every single method, maybe just create a new `IAudioClip` implementation and instantiate one of those instead on iOS?

forge-gui/src/main/java/forge/gui/GuiBase.java
```diff
@@ -19,6 +20,11 @@ public class GuiBase {
 
     public static void setIsAndroid(boolean value) { isAndroidport = value; }
     public static boolean isAndroid() { return isAndroidport; }
+
+    /** Set by the iOS launcher (forge.ios.Main), mirroring the Android flag above, so shared
+     *  modules can branch on the platform without sniffing libGDX's ApplicationType. */
+    public static void setIsIOS(boolean value) { isIOSport = value; }
+    public static boolean isIOS() { return isIOSport; }
```
Jetz72, line 27: Should probably have a shared `isMobile` method as well. We'll probably need to do a pass over the existing uses of `isAndroid` and see if they need to apply to iOS too.

forge-gui-mobile/src/forge/Forge.java
```diff
@@ -653,13 +667,30 @@ public static void setBackScreen(final FScreen screen0, boolean replace) {
         Dscreens.addFirst(front);
     }
 
+    // iOS apps must not programmatically terminate (App Store guideline / HIG),
+    // so the iOS device adapter's exit()/restart() are no-ops. That left the
+    // normal exit flow stuck on the ClosingScreen, forcing the user to swipe the
+    // app away and relaunch. On iOS, return to the main menu instead so the app
+    // stays usable: adventure -> classic home (switchToClassic), classic -> home.
```
Jetz72, line 674: If the app is trying to restart to apply certain settings (e.g. switching Adventure planes), this would prevent it from reloading as it expects. I don't expect Forge is going to be on the app-store in the foreseeable future, so would it be fine to just let the app exit?

tool4ever (approved):
Also worked on my iPhone with Sideloadly 🥳

I'm slightly optimistic that with the changes in the other modules mostly isolated no regressions will be caused, but I'll watch our Sentry for it.

Two minor follow-ups I found:
- App Icon seems a bit weird, instead of the Forge one?
- ingame bottom buttons seemed too huge, not sure it's that way on Android?

thanks!

forge-gui-mobile/src/forge/assets/ImageCache.java
```diff
@@ -353,29 +506,90 @@ private Texture loadAsset(String imageKey, File file, boolean others) {
...
+        String fileName = toRelativePath(absolutePath);
+
+        // iOS only: downloaded images (in Documents/cache) bypass the AssetManager —
+        // read the bytes and create the Pixmap directly (libGDX iOS can't load a
+        // FileHandle from Documents). Other platforms keep the AssetManager path.
+        boolean isDownloadedImage = GuiBase.isIOS() && absolutePath.contains("/Documents/cache");
+        Texture directTexture = null;
+
+        if (isDownloadedImage) {
+            try {
```
Jetz72, line 521: Would it make sense to move this alternate caching behavior into the asset manager, instead of bypassing it and reimplementing its responsibilities in a bunch of platform-specific branches in this file? The logic across this file gets a bit hard to follow.

Jetz72 (commented):
Bit late on these notes, but might be worth considering for any follow-up.

tool4ever:
I also wonder if [CodenameOne](https://github.com/codenameone/CodenameOne) would eventually be a less hacky alternative...

forge-gui/src/main/java/forge/gui/GuiBase.java
liminalwarmth: Opened #11382 adding `GuiBase.isMobile()` — it converts the two call sites that already spell out `isAndroid() || isIOS()`. Left the remaining `isAndroid()` uses alone for now: each one needs a judgment call on whether it should apply to iOS (and ideally on-device verification), so that pass seems better done incrementally.

Eradev:
This PR downgraded the card textures for every OS. I don't know if you only wanted to do so for iOS, but users have complained about a degraded experience.

tool4ever:
@Eradev
If it makes sense and @shoeless doesn't turn up soon feel free to liberate the other platforms for now? thanks

shoeless:
@Eradev , @tool4ever , will post a quick fix in a moment then tackle some of the related refactoring above in the coming days.

forge-gui-mobile/src/forge/sound/AudioClip.java
shoeless: Done — `AudioClip` is back to a pure OpenAL `IAudioClip` implementation, and the iOS AVAudioPlayer path moved verbatim into a second implementation, `MusicAudioClip`. The `createClip` factories pick the implementation per call, so there are no more per-method `isIOS` branches. (#11556)

forge-gui/src/main/java/forge/gui/GuiBase.java
shoeless: `GuiBase.isMobile()` landed in #11382. On top of it, #11556 converts the sites where Android and iOS want identical behavior (adventure data paths, fallback-skin textures, hardware-cursor skip, hover detection, PID logging). The sites where `isAndroid()`/`isIOS()` gate genuine platform machinery (APK updates, audio backend, orientation) intentionally keep the specific check. The remaining behavior-changing sites to audit are tracked in #11384.

forge-gui-mobile/src/forge/Forge.java
shoeless: Agreed — the return-to-main-menu intercept meant restart-required settings (Adventure plane, skin, language, card-DB toggles) could never take effect, since those are process-lifetime singletons. #11556 makes iOS `exit()`/`restart()` terminate the process and let the user relaunch. Forge iOS is sideload/TestFlight distributed, so Apple's no-programmatic-exit HIG concern doesn't apply.

forge-gui-mobile/src/forge/assets/ImageCache.java
shoeless: Done in #11556. The downloaded-image path no longer bypasses the AssetManager. A small custom `TextureData` (`CardTextureData`) does the byte read + downscale and is handed to the manager via `TextureParameter.textureData` — the stock `TextureLoader` honours a preset `textureData` and ignores the FileHandle, so the read stays a plain java.io byte read (iOS can't reliably load a FileHandle from Documents) while the AssetManager owns the whole lifecycle: get/load/unload/dispose/context-loss reload/memory-tracking.

That let me delete the three parallel structures this comment was about — `pixmapCache`, the `downloadedTextureCache` LRU, and `textureToPath` — plus the `isDownloadedImage` bypass branch and the `toRelativePath` rewriting. `loadAsset`/`getAsset` now take the same manager path on every platform; the one residual iOS check merely selects the decode parameter (a per-platform decode concern, not a re-implementation of the manager's responsibilities). No custom loader and no `setLoader`, so skins/UI/fonts and other textures are untouched.

PR #11192, opened 2026-07-09
Bind docked zones to live PlayerView to fix stale cards on new match

### LINKED ISSUE #11111: When "Start New Match" is selected, information from the previous game is carried over.
**Describe the bug**
After a game ends, when I press Start New Match to begin the next game, the cards used in the previous game are still displayed in windows such as Playable Zone Cards.

**To Reproduce**
1. Go to Game Result.
2. Click to "Start New Match"
3. See "Playable Zone Cards" or other window.

**Expected behavior**
Clear the informations.

**Screenshots**
Image

 - OS: Windows 10
 - Browser Firefox
 - Version forge-installer-2.0.14-SNAPSHOT-06.28

MostCromulent:
Closes Card-Forge/forge#11111.

## Summary

Fixes docked zone tabs (e.g. "Playable Zone Cards") showing the previous game's cards after "Start New Match".

## Bug report

With a zone view docked as a tab (rather than floating), starting a new match left the previous game's cards in the tab. They never cleared — persisting through the whole new game, even after drawing and playing. Floating zone windows were unaffected.

## Root cause

At each game start, `CMatchUI.updatePlayerControl` rebuilds the docked tabs via `FloatingZone.registerZoneDocs`, binding each `VZone` to a `PlayerView` from `getLocalPlayers()` (the `gameControllers` key set). On restart the desktop reuses the same `CMatchUI` without resetting controller state (only mobile calls `resetForNewMatch`), and the new game installs fresh `PlayerView` instances. Because `TrackableObject` equality is by id, re-registering the local player leaves the `HashMap`'s original **stale key** in place — so `getLocalPlayers()` returns an orphaned `PlayerView` from the prior game that never receives updates. Floating windows dodge this by being rebuilt lazily on click against the live view.

## Fix

Source the local players from the live game view (`getGameView().getPlayers()` filtered by `isLocalPlayer`) instead of `getLocalPlayers()`, so docked zones bind to the live `PlayerView` and refresh correctly.

## Alternative considered

Resetting the reused controller state each game (mirroring mobile's `resetForNewMatch`) would fix the root cause more broadly. It was not taken here because it lives in shared lifecycle code (`AbstractGuiGame`/`HostedMatch`) exercised by desktop, mobile, network host, and client — a much larger blast radius. This PR takes the lower-risk desktop-scoped fix.

---

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
+++ b/forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
@@ -940,7 +940,14 @@ public final class CMatchUI
     @Override
     public void updatePlayerControl() {
         initHandViews();
-        FloatingZone.registerZoneDocs(this, getLocalPlayers());
+        // Use live PlayerViews; getLocalPlayers() can hand back an orphaned key from the prior game (equality is by id)
+        final List<PlayerView> localPlayers = new ArrayList<>();
+        for (final PlayerView p : getGameView().getPlayers()) {
+            if (isLocalPlayer(p)) {
+                localPlayers.add(p);
+            }
+        }
+        FloatingZone.registerZoneDocs(this, localPlayers);
         SLayoutIO.loadLayout(null);
         FloatingZone.pruneUnparentedDocks();
         view.populate();
--- a/forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
+++ b/forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
@@ -940,14 +940,7 @@ public final class CMatchUI
     @Override
     public void updatePlayerControl() {
         initHandViews();
-        // Use live PlayerViews; getLocalPlayers() can hand back an orphaned key from the prior game (equality is by id)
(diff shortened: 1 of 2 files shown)


forge-gui-desktop/src/main/java/forge/screens/match/CMatchUI.java
```diff
@@ -940,7 +940,14 @@ public void updateTurn(final PlayerView player) {
     @Override
     public void updatePlayerControl() {
         initHandViews();
-        FloatingZone.registerZoneDocs(this, getLocalPlayers());
+        // Use live PlayerViews; getLocalPlayers() can hand back an orphaned key from the prior game (equality is by id)
+        final List<PlayerView> localPlayers = new ArrayList<>();
```
tool4ever: this should be done with a generic fix imo since this pattern seems used in multiple places, e.g. `updateDrawOffer`

so this could lead to more subtle bugs unless AI can think of a reason why we want state from game 1 in game 1+X this way...?

tool4ever (approved):

PR #11193, opened 2026-07-09
Change decks directory option

GiovanniMenzano:
Adds an option for changing the default decks directory in the desktop version.

The new option is available under Deck editor options. When a user selects a new deck directory, Forge saves the setting and informs the user that a restart is required for the change to take effect.
If the current deck directory contains files, the function also asks the user if they want to migrate existing decks to the newly selected directory. The operation copies files recursively, skips files that already exist in the destination, and leaves the old decks directory untouched.


GiovanniMenzano:
For the record, I made it mostly because I wanted to share my custom decks with my girlfriend, so I made a shared cloud folder and used the function to migrate all my decks there. She did the same and now she gets the updates every time I create or modify something.
But I also made it to be able to export easily.
I'm also thinking about in app folders to help with organization.

PR #11198, opened 2026-07-10
Add bounded mana payment planner for costed mana abilities

Madwand99:
Adds a bounded best-first mana payment planner for cases the greedy mana payer cannot reliably handle, especially signets, filter lands, and restricted mana abilities.
The planner searches virtual payment states without tapping real permanents, tracks floating mana and used sources, then replays the selected mana abilities once a valid plan is found. ComputerUtilMana keeps the existing greedy path for normal payments and only escalates to the planner when needed, including cases where greedy payment would leave enough total mana but strand important colors for other spells in hand.
This also updates mana color choice handling so planned combo/chosen mana outputs are preserved when the chosen mana abilities are replayed.

Thanks to Codex for help with coding and many, many tests.


tool4ever:
Thanks though there's now also a "human contributed" attempt in progress: #11173

If that leads nowhere either I'll check out this new approach, though I probably won't be a big fan if this separate system duplicates things too much - but we'll see

Madwand99:
> Thanks though there's now also a "human contributed" attempt in progress: #11173
> 
> If that leads nowhere either I'll check out this new approach, though I probably won't be a big fan if this separate system duplicates things too much - but we'll see

I notice that many of the tests on #11173 don't pass their code, and one of them is an invalid test. I'll be checking those tests to add some to my own test suite. So far I have 55 tests, with more on the way. Should I push these to this PR? Due to your own comments I have been keeping most tests to myself.

Madwand99:
## Comparison with PR #11173

### Size

Production changes are measured against each branch's merge base.

| | This PR | PR #11173 |
|---|---:|---:|
| Production LOC | +1,102 / -45 | +4,778 / -427 |
| Net production LOC | +1,057 | +4,351 |
| Architecture | Separate planner class | `CastabilityProbe` plus primarily `ComputerUtilMana` |

This PR is substantially smaller: its latest branch now adds 4,351 net production lines, including 4,360 additions in `ComputerUtilMana`.

PR #11173 also includes +2,026 / -1 committed test lines. Our tests remain local for now but can be made available on request.

### Test Results

The current local suite has 137 behavioral payment tests plus 11 planner-internal fallback/state-cap tests.

| Test group | This PR | PR #11173 |
|---|---:|---:|
| Original `AutoPaymentTest` cases (64) | 64 passed / 0 failed | 52 passed / 12 failed |
| Imported comparison cases (73) | 73 passed / 0 failed | 69 passed / 4 failed |
| Behavioral total (137) | **137 passed / 0 failed** | **121 passed / 16 failed** |
| Planner fallback/state-cap cases (11) | 11 passed / 0 failed | Not applicable |

The 11 fallback/state-cap tests cannot compile against PR #11173 because it does not contain the `ManaPaymentPlanner` API. They therefore should not count as failures for that PR.

### Shared Strengths

Both implementations pass tests covering:

- Basic filter-land and Signet activation costs
- Hidden Grotto correctly failing to cast Vigor
- Lotus Petal activating a Signet when needed
- Cascading Cataracts
- Restricted creature mana
- Many Treasure/disposable-resource cases
- Equivalent-source scalability cases
- The Yuna and Goldspan performance scenarios
- The Weatherseed Treaty filter-land scenario
- Preserving Gaea's Cradle when exact single mana is available elsewhere
- Avoiding a redundant second Signet

### Differences

PR #11173 fails 12 of the original behavioral tests. These include:

- Chaining two Signets when the chain is necessary
- Evendo's variable mana
- Castle Garenbrig's additional creature-only mana
- Cabal Coffers
- Nykthos castability and simulated payment
- A filter-land line where its extra mana can pay generic cost
- Choosing needed colors from any-color sources
- Preferring colorless mana over a colored land for generic cost
- Reporting floating mana and floating activation-mana sources
- Keeping non-tap, mana-costed abilities out of the simple mana-source list

Notably, PR #11173 still avoids a redundant second Signet, but cannot chain two Signets when that is the only valid payment. This PR passes both cases.

The remaining four PR #11173 failures are intentional source-policy differences in the imported comparison tests:

- Preserve a white-capable source needed by a known white spell in hand
- Preserve the only `{C}` source needed by Thought-Knot Seer
- Use an off-color rock for generic mana to preserve a dual land
- Preserve Lotus Petal when reusable mana can complete the payment

Those expectations were revised locally; they should not be treated as neutral regressions in PR #11173.

### Architecture

This PR uses a bounded best-first search with virtual mana state, memoization, source equivalence, a hard state cap, and legacy fallback. It searches without mutating game state and returns a replayable payment plan.

PR #11173 adds nested and recursive payment handling around the existing payment logic, with a separate `CastabilityProbe`. Its current implementation is concentrated primarily in `ComputerUtilMana`; it does not provide the same explicit planner, exhaustion, fallback, or replay-plan boundaries.

tool4ever:
thanks for comparing

tests will in fact be needed however **148 is a completely unreasonable amount** 👎 👎 👎

they need to be written around code knowledge so we're not just setting up lots of arbitrary scenarios that just cover the same logic

it would have been better to wait and then iterate on each case / deciding if initial support is worth it

Madwand99:
> thanks for comparing
> 
> tests will in fact be needed however **148 is a completely unreasonable amount** 👎 👎 👎
> 
> they need to be written around code knowledge so we're not just setting up lots of arbitrary scenarios that just cover the same logic
> 
> it would have been better to wait and then iterate on each case / deciding if initial support is worth it

Almost every test covers different logic and different cards. The tests are critical for defining what the code can do. We can certainly cut down on the final tests included, but for now the more tests we have, the more cases are covered. Picking a subset of test to keep will be quite challenging as almost any test left out will mean some kinds of cards aren't going to be covered.

Madwand99:
The latest update closes correctness gaps found during review:
- Chosen mana: Unchosen Chosen abilities are no longer treated as producing every color unless the ability actually chooses a color before producing mana.
- Triggered combo mana: Multiple TapsForMana triggers are composed correctly, including full amounts and “different colors” restrictions. Candidate output remains bounded.
- Search limits: All planner searches now share one state budget and explicitly report exhaustion, preventing secondary searches from bypassing the cap.
- Safe replay: Before mutating game state, the planner verifies that activation mana still matches the planned payment. Failed or mismatched replay now restores activation costs, mana abilities, temporary choices, and the original mana pool before legacy fallback runs.
- Planning isolation: Mana choices are snapshotted and restored so speculative planning cannot leak mutations into the game.

jamincollins:
## Findings
### Logic correctness
#### 1. shouldUsePlanner may short-circuit on activation cost with non-reusable resource
ManaPaymentPlanner.shouldUsePlanner checks activationCost.hasManaCost() && activationCost.hasTapCost() && (activationCost.isReusuableResource() || manaSa.getManaPart().isComboMana()). If activationCost.isReusuableResource() is true but activation cost has no mana, condition fails. Test case needed to verify signet with reusable resource but activation cost only tap.

Test case:
```java
// Verify shouldUsePlanner returns true for signet with reusable resource
ManaCostBeingPaid cost = new ManaCostBeingPaid(new ManaCost("G"));
SpellAbility sa = mockSpellAbilityWithCost(cost);
Player ai = mockPlayerWithSignet();
assert ManaPaymentPlanner.shouldUsePlanner(cost, sa, ai, true);
```
Impact: Planner may be incorrectly disabled for certain reusable mana sources, causing fallback to greedy payer.

#### 2. specifyComboFromActivePayment mutates ColorSet during iteration
In specifyComboFromActivePayment, colors is mutated via colors = ColorSet.fromMask(colors.getColor() & ~choice) inside loop with different flag. This creates a new ColorSet each iteration but the loop uses same variable name, could cause confusion. Also remaining.ai_payMana is called with MagicColor.toShortString(choice) which may not match remaining needs.

Test case:
```java
// Test different=true with multiple colors
Map<Byte, Integer> combo = ManaPaymentPlanner.specifyComboFromActivePayment(sa, ColorSet.fromMask(MagicColor.RED|MagicColor.BLUE), 2, true);
// Expect one red, one blue, not two of same
```
Impact: May produce invalid color combo for X spells with different flag, breaking rule compliance.

### Edge cases
#### 3. findPlan returns null leading to silent fallback
In ComputerUtilMana.findPlan branch, plan = ManaPaymentPlanner.findPlan(...). If plan is null, plannerExhausted stays false, and later greedyPayment == null && shouldUsePlanner && !plannerExhausted returns null, causing cost not paid. No fallback to greedy.

Test case:
```java
// Plan returns null for cost with only Phyrexian mana
ManaCostBeingPaid cost = new ManaCostBeingPaid(new ManaCost("P"));
assert ManaPaymentPlanner.findPlan(cost, sa, ai, true) == null;
```
Impact: AI may be unable to pay costs with exotic mana when planner returns null, losing plays.

#### 4. activePayment ThreadLocal not cleaned on exception
ManaPaymentPlanner.Plan.pay sets activePayment, try-finally removes it. If exception occurs before set, finally removes activePayment which may clear unrelated context. Also rollback may leave mana abilities on stack.

Test case:
```java
// Simulate exception during pay
ManaPaymentPlanner.Plan plan = ...;
try { plan.pay(...); } catch (Exception e) { /* activePayment leaked? */ }
```
Impact: ThreadLocal leak can cause specifyComboFromActivePayment to use stale context, leading to wrong color choices.

### Data integrity
#### 5. source equivalence keys ignore card state
getSourceEquivalenceKeys likely uses card identity only, not game state (tapped, counters). Two identical cards in different states may be treated equivalent, causing planner to reuse invalid sources.

Test case:
```java
// Two signets, one tapped
Card signet1 = tappedSignet;
Card signet2 = untappedSignet;
assert !plannerEquivalenceKeysEqual(signet1, signet2);
```
Impact: Planner may plan using tapped source, leading to replay failure and rollback.

#### 6. VirtualPool.remainingCost mutates ManaCostBeingPaid
VirtualPool.remainingCost creates new ManaCostBeingPaid(cost) then calls remaining.payMana(m, ai.getManaPool()) which mutates remaining. If cost is reused elsewhere, side effects occur.

Test case:
```java
ManaCostBeingPaid cost = new ManaCostBeingPaid(...);
VirtualPool pool = new VirtualPool(manaList);
pool.remainingCost(cost, paidFor, ai);
// cost now modified?
```
Impact: Shared cost objects corrupted, causing incorrect planning decisions.

### Side effects
#### 7. Planner exhaustion marks EXHAUSTED but still returns Plan.EXHAUSTED
Plan.isExhausted() checks actions == EXHAUSTED_ACTIONS. However findPlan returns Plan.EXHAUSTED when candidates == null, and findBetterConsolidatingPlan may also return null. Caller treats null differently than exhausted, leading to inconsistent fallback.

Test case:
```java
Plan plan = ManaPaymentPlanner.findPlan(cost, sa, ai, true);
assert plan != null && plan.isExhausted();
```
Impact: Exhaustion handling inconsistent; may skip legacy fallback when it should be allowed.

#### 8. shouldUntapStorageLand removed without replacement
PlayerControllerAi removed shouldUntapStorageLand method and StorageLand case. AI may now never untap storage lands, hurting mana generation for storage land decks.

Test case:
```java
PlayerControllerAi ai = new PlayerControllerAi(player);
assert ai.shouldUntapStorageLand(card) == false; // method removed
```
Impact: Regression in storage land usage, AI may lose mana advantage.

### Performance / Security
#### 9. MAX_STATES default 10000 may be too low for complex boards
ManaPaymentPlanner.maxStates = 10000. Complex multi-signet boards may exhaust state search early, causing suboptimal plans.

Test case:
```java
ManaPaymentPlanner.maxStates = 10000;
// Board with 12 signets + 5 filters
Plan plan = ManaPaymentPlanner.findPlan(cost, sa, ai, true);
assert plan != null && !plan.isExhausted();
```
Impact: Performance degradation on complex boards, AI may miss winning lines.

## Verdict
Request Changes
Planner adds important bounded mana payment support but several logic, edge case, and data integrity issues need fixing before merge. Critical concerns around ThreadLocal cleanup, null plan handling, and source equivalence.

## Recommended Fixes
1. Ensure specifyComboFromActivePayment does not mutate caller ColorSet; use local copy.
2. Handle null plan explicitly in ComputerUtilMana to fall back to greedy payer.
3. Guard activePayment ThreadLocal against stale context and ensure cleanup on all paths.
4. Include card state in source equivalence keys.
5. Defensive copy of ManaCostBeingPaid before mutation in VirtualPool.
6. Unify exhausted vs null handling; document semantics.
7. Re-evaluate removal of shouldUntapStorageLand or provide replacement heuristic.

---

PR #11199, opened 2026-07-10
Support Monstrosity spell ability property

Ellios77:
Abilities with the Monstrosity script parameter can be identified directly, matching the existing pattern for Boast and other ability types. Enables valid strings such as Activated.Monstrosity to match.

Example:
S:Mode$ ReduceCost | ValidCard$ Card | ValidSpell$ Activated.Monstrosity | Activator$ You | Amount$ 1 | Description$ Monstrosity abilities you activate cost {1} less to activate.


Hanmac (approved):
I allow it

PR #11219, opened 2026-07-12
[Fix] Nick Fury script fix and Dig filter fix

alexbrooke96:
Nick fury was remembering all cards he'd ever seen and not putting all unselected cards to bottom of deck

Dig w/ filter wasnt showing peaked cards if no valid targets

Tested locally, works as expected now.


tool4ever:
can't really reproduce + effect order wrong now anyway
I guess you created this without checking my previous commit that fixed it?

PR #11221, opened 2026-07-13
[Fix] entity counter respects type

alexbrooke96:
CounterAddedAll triggers fired for entities that only received counters of other types, because filterTable kept zero-count entries. Filter them out, handle CounterType$ Any (null) correctly, and add a regression test.


Hanmac (approved):

tool4ever (approved):
thanks

PR #11266, opened 2026-07-17
Fix turn transition after cleanup-step loss

delebedev:
Forge's Puzzle mode can make a player lose during cleanup when the turn limit expires. The player receives a loss outcome immediately, but remains in the turn order until the game-over condition is processed. Next-turn selection could therefore keep returning the same lost player and hang.

| | Before | After |
|---|---|---|
| Pending loss | Select the next player first | Process the game outcome first |
| Duel | Assign the lost player and continue the phase transition | End without starting another phase |
| Multiplayer | Lost player remains in the rotation | Remove the lost player and select a live player |

Game outcomes are also checked again if next-player selection itself causes a loss.

### DIFF
--- a/forge-game/src/main/java/forge/game/phase/PhaseHandler.java
+++ b/forge-game/src/main/java/forge/game/phase/PhaseHandler.java
@@ -512,7 +512,11 @@ public class PhaseHandler implements java.io.Serializable, IHasForgeLog {
                     game.onCleanupPhase();
                     // set previous player
                     playerPreviousTurn = this.getPlayerTurn();
-                    setPlayerTurn(handleNextTurn());
+                    Player next = handleNextTurn();
+                    if (game.isGameOver()) {
+                        return;
+                    }
+                    setPlayerTurn(next);
 
                     // start effects for next turn (do this first for ControlPlayer)
                     game.getCleanup().executeUntil();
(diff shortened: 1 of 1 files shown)


tool4ever:
can't reproduce, which puzzle is being fixed here?

delebedev:
I first ran into this while doing some headless work with Forge's puzzle runner. It was difficult to see what was happening from the headless symptom alone, so I figured Codex could help me turn it into a clear desktop reproduction.

We found that the bundled **Pauper Puzzles #25 - Good Money After Bad** puzzle reaches the same path:

1. Kill `Humbler of Mortals` with `Disfigure`, then `Lightning Bolt`.
2. End the human turn.
3. Pass the AI upkeep and declare-attackers prompts.
4. Let `Blastoderm` attack and the puzzle reach its turn-limit loss during AI cleanup.

| Before | After |
|---|---|
| Forge remains on "Waiting for..." | The normal Puzzle Goal loss result appears |
| 36.8-178.5% CPU; 80.2% mean | 0.3-2.3% CPU; 0.7% mean |
| Game thread stays in `getNextActivePlayer()` | Game thread returns to the idle worker pool |

- [Baseline video (1.5x playback)](https://pub-ee18c3c0efd64ad5967c2972fae3edd3.r2.dev/1784408629-forge-turn-limit-baseline-1.5x.mp4)
- [Patched video (1.5x playback)](https://pub-ee18c3c0efd64ad5967c2972fae3edd3.r2.dev/1784408629-forge-turn-limit-patched-1.5x.mp4)

Both videos use the same sequence, with CPU samples and thread dumps collected during each run.

tool4ever:
Thanks, I've turned it into a more rules-based fix here: #11274

PR #11273, opened 2026-07-18
Improve multiplayer AI attack targeting with TTK and softmax selection

Madwand99:
This improves multiplayer combat targeting so the AI is less likely to repeatedly focus a single player based on small or incidental score differences.

Changes:
- Replace the quadratic low-life bonus with a bounded, combat-only time-to-kill (TTK) estimate.
- Factor both incoming danger and the AI’s ability to finish an opponent into threat evaluation.
- Use softmax sampling for the primary attack defender, keeping strong threat differences meaningful while varying choices among similarly threatening opponents.
- Keep ranked fallback defenders so the AI can attack an open opponent when its preferred target is not profitable to attack.
- Ensure TTK block simulations do not use another player’s hidden information.
- TTK is intentionally conservative and bounded to ten combat steps. It currently models repeated combat damage only; poison, commander damage, burn, lifegain, and evolving board states remain future extensions.
- Added a generic softmax function


forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -202,24 +209,19 @@ public static Player choosePreferredDefenderPlayer(Player ai) {
...
             int score = ComputerUtil.evaluateBoardPosition(ai, opp);
-            // round away slightly so a single land drop doesn't mean players with earlier turn order are predictably attacked
-            score = (int) Math.ceil(score / 10) * 10;
-            int lowLifeThreshold = Math.min(20, opp.getStartingLife());
-            if (life > 0 && life < lowLifeThreshold) {
-                // TODO commander damage
-                int lifeDeficit = lowLifeThreshold - life;
-                score += lifeDeficit * lifeDeficit;
-            }
+            score += ComputerUtil.getCombatTtkScore(ComputerUtil.estimateCombatTurnsToKill(ai, opp));
```
tool4ever: My main thoughts for now:
a) this scoring change makes it only look at attack power from AI, where previously being at low life was already enough (assuming other opponents will also consider the same defender)
b) How sure are we that it's really a superior strategy to get rid of the weakest player instead of e.g. slowing the  strongest down?
c) I have some doubts looking more than one combat into the future this way won't just lead to other wrong conclusions (attackers/blockers trading or just the board state completely changing)

Madwand99: Yeah, some valid concerns there that I also share. It is tricky to get this right. I pushed a change that should help: I’ve kept the existing threat evaluation as the primary signal and changed outgoing TTK into a short-term tactical modifier rather than a general low-life replacement: it only scores kills within one or two combats, so it no longer relies on long static-board projections.
The finishing bonus is also scaled down exponentially when that opponent’s underlying threat is below the table leader’s. This means a low-threat player who is easy to eliminate can help break a close decision, but should not displace a clearly more dangerous opponent that the AI ought to be slowing down. I did not reintroduce a separate low-life bonus, since that would largely double up on the same elimination pressure.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -3176,6 +3174,45 @@ public static boolean aiLifeInDanger(Player ai, boolean serious, int payment) {
...
+     * Estimates how many identical, combat-only attacks {@code attacker} needs to defeat
+     * {@code defender}. The estimate assumes the defender blocks optimally and the attacker
+     * attacks with every creature that can attack next turn.
+     */
+    public static int estimateCombatTurnsToKill(final Player attacker, final Player defender) {
+        if (attacker == null || defender == null || !attacker.isOpponentOf(defender)
+                || defender.getLife() <= 0 || !defender.canLoseLife() || defender.cantLoseForZeroOrLessLife()) {
+            return Integer.MAX_VALUE;
+        }
+        return AiCache.getCached("estimateCombatTurnsToKill",
```
tool4ever, line 3186: too much duplicated code:
this method should just be another variant reusing `aiLifeInDanger`

also need to think about passing all opps together for a better average?

Madwand99: I’ve refactored TTK to reuse predictNextCombatsRemainingLife; it now only converts the shared combat prediction’s damage result into a number of combats. I kept an internal mode for this use case so it can calculate exact remaining life and avoid using another player’s hidden information.

For passing all opponents: do you mean modelling aggregate table pressure on a defender? The current helper evaluates each opponent’s combat separately and returns the worst individual outcome, rather than an average or cumulative result, so I kept the single-attacker call for now. An explicit aggregate-pressure model could be useful, but think it needs separate design rather than simply changing the opponent list.

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -3210,18 +3244,18 @@ public static int predictNextCombatsRemainingLife(Player ai, boolean serious, bo
...
             // TODO for performance skip ahead to safer blocking approach (though probably only when not in checkDiff mode as that could lead to inflated prediction)
             block.assignBlockersForCombat(combat, excludedBlockers);
 
             // TODO predict other, noncombat sources of damage and add them to the "payment" variable.
             // examples : Black Vise, The Rack, known direct damage spells in enemy hand, etc
             // If added, might need a parameter to define whether we want to check all threats or combat threats.
 
-            if (serious && ComputerUtilCombat.lifeInSeriousDanger(ai, combat, payment)) {
+            if (checkDanger && serious && ComputerUtilCombat.lifeInSeriousDanger(ai, combat, payment)) {
                 return Integer.MIN_VALUE;
```
tool4ever, line 3258: you should still just check this, it means only 1 turn because of poison or commander

forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -3176,10 +3174,46 @@ public static boolean aiLifeInDanger(Player ai, boolean serious, int payment) {
...
+        return (COMBAT_TTK_HORIZON - turnsToKill + 1) * COMBAT_TTK_SCORE_PER_TURN;
+    }
+
+    private static int estimateCombatTurnsToKillChanged(final Player attacker, final Player defender) {
+        final int remainingLife = predictNextCombatsRemainingLife(defender, false, true, 0, null, List.of(attacker), true, false);
+        if (remainingLife == Integer.MAX_VALUE) {
+            return Integer.MAX_VALUE;
+        }
+        final int damage = defender.getLife() - remainingLife;
+        return damage > 0 ? (defender.getLife() + damage - 1) / damage : Integer.MAX_VALUE;
```
tool4ever, line 3208: imo we're also losing accuracy here if AI is at very low life so all opponents will basically lead to TTK=1

but it should still matter which opponent causes lower negatives

forge-ai/src/main/java/forge/ai/ComputerUtil.java
tool4ever: now you no longer need the extra parameter, just use serious...

forge-ai/src/main/java/forge/ai/ComputerUtil.java
tool4ever: hmn I think when I clean it up I'll just leave a TODO for now to check if somehow all opponents are part of the same team which would mean cumulative is the right choice

seems like too much of a corner case to complicate things here

forge-ai/src/main/java/forge/ai/ComputerUtil.java
Madwand99: Good call. Done!

forge-ai/src/main/java/forge/ai/ComputerUtil.java
Madwand99: I think TTK should intentionally collapse all terminal next-combat outcomes to 1: once a player would lose, excess normal damage is not more immediately relevant, and comparing negative life would also mis-rank lethal poison or commander damage that may not reduce life at all. If we need to distinguish between multiple one-turn lethal threats, I think that belongs in the broader board-threat evaluation rather than the TTK value itself.

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -67,6 +67,9 @@
  */
 public class AiAttackController {
 
+    // A 75-point lead gives a defender roughly a 73% chance of being selected over one alternative.
+    private static final double DEFENDER_SCORE_TEMPERATURE = 75d;
```
tool4ever: these static fields belong to `AiProps`

Madwand99: Done!

forge-ai/src/main/java/forge/ai/AiAttackController.java
```diff
@@ -202,24 +206,27 @@ public static Player choosePreferredDefenderPlayer(Player ai) {
...
-                // TODO commander damage
-                int lifeDeficit = lowLifeThreshold - life;
-                score += lifeDeficit * lifeDeficit;
-            }
+            final int threatScore = ComputerUtil.evaluateBoardPosition(ai, opp);
+            threatScores.put(opp, threatScore);
+        }
+        int highestThreat = Collections.max(threatScores.values());
+
+        final int temperature = AiProfileUtil.getIntProperty(ai, AiProps.MULTIPLAYER_DEFENDER_SOFTMAX_TEMPERATURE);
```
tool4ever, line 225: I wouldn't mind a more dynamic approach to determine the score interval that counts as equal threat
but allowing AI to make potentially game losing attacks (even with a rather low probability) just seems wrong

Madwand99: The small chance for even low-threat opponents to be attacked is intentional. From a recent comment on the Discord: "...Also, humans will take attacks of opportunity. As in, attacking the least threatening guy because they can't get in on the biggest threat. I don't see the AI doing that at all. I will sit there with no creatures out for several turns, just land go. And I wont ever get attacked, they just plow into each other. "

No one should become immune to being attacked just because they sandbag. The small chance of being attacked prevents this. Some randomness added to AI decision making prevents them from becoming unnaturally predictable and makes the game more fun.

tool4ever:
I need to think about if much more complexity for threat scoring is the right approach or if actually supporting AI spreading its damage to multiple players would solve this way more naturally (though maybe also a bit tricky to get right)

Madwand99:
I think your recent PR #11331 is complementary rather than a replacement for this PR. It introduces target variety through a dynamic equal-threat window, but it retains the existing threat signal, including the unconditional low-life bonus. This PR improves that signal with combat TTK, limits finishing pressure when a weaker player is serving as a check on a clearer leader, and uses a graded selection probability plus a fallback to an open target when the preferred defender produces no attack.
In particular, master still has a cliff in its decision: opponents below the dynamic score cutoff are never considered, while every eligible opponent is equally likely. That means it can still concentrate all AI attacks on one player whenever that player is sufficiently ahead, and it can still mistake low life for an urgent reason to finish a player even when another opponent is the more immediate danger.
This PR makes the preference proportional rather than binary. An opponent’s ability to kill the AI soon is reflected through combat TTK, while the AI’s ability to finish an opponent is treated as a limited opportunity bonus, not an overriding priority. The finishing bonus is reduced when that low-life player is much less threatening than the leader, preserving them as a possible counterweight rather than having every AI eliminate them.
Softmax then gives near-equal alternatives a meaningful chance without declaring all lower-ranked players impossible targets. Finally, when the preferred target is too well defended to produce a profitable attack, the AI tries its next-ranked target, allowing it to take the available attack opportunity instead of repeatedly declining combat.

In other words: we still need this PR.

jamincollins:
## Findings

#### 1. Softmax temperature can select negative temperature causing IllegalArgumentException
Description:
Aggregates.itemWithSoftmax throws IllegalArgumentException when temperature <= 0 or !Double.isFinite(temperature). AiProps.MULTIPLAYER_DEFENDER_SOFTMAX_TEMPERATURE is read via AiProfileUtil.getIntProperty which returns int. If a user configures a negative value or zero, temperature <=0 triggers exception during defender selection in multiplayer. The code does not guard against invalid config values before calling softmax.
Test case:
```java
// Simulate invalid config
int temperature = -1; // user sets negative
Player ai = ...;
Player opp1 = ...;
Player opp2 = ...;
List<Player> preferences = List.of(opp1, opp2);
Map<Player,Integer> scores = Map.of(opp1,10, opp2,12);
// This will throw IllegalArgumentException from Aggregates.itemWithSoftmax
Player selected = Aggregates.itemWithSoftmax(preferences, scores::get, temperature);
```
Impact: AI crashes during combat declaration in multiplayer games with misconfigured property, leading to unplayable game.

#### 2. getFinishingBonus uses threatScore - highestThreat which can be negative, making exp negative, reducing bonus unexpectedly
Description:
getFinishingBonus computes `Math.exp((threatScore - highestThreat) / temperature)`. When threatScore < highestThreat, exponent is negative, resulting in factor <1, which reduces TTK bonus for weaker opponents. The comment says "Preserve weaker opponents as potential checks on a clear leader", but the math actually diminishes their bonus. This may invert intent: weaker opponents get less TTK influence.
Test case:
```java
int highestThreat = 100;
int threatScore = 80;
int temperature = 75;
double factor = Math.exp((threatScore - highestThreat) / (double) temperature);
// factor ~ 0.74, so weaker opponent gets reduced TTK bonus
int ttkScore = 40;
int bonus = (int)Math.round(ttkScore * factor); // ~30, reduced
```
Impact: AI may undervalue weaker opponents when it should preserve them as checks, potentially over-committing to strongest threat.

#### 3. estimateCombatTurnsToKillChanged returns damage>0 division but can divide by zero
Description:
estimateCombatTurnsToKillChanged computes damage = defender.getLife() - remainingLife. Then returns `(defender.getLife() + damage -1) / damage`. If damage == 0, the denominator is zero, causing ArithmeticException. damage==0 occurs when remainingLife == defender.getLife(), i.e., combat predicts no damage. The method does not guard against damage ==0 before division.
Test case:
```java
Player attacker = ...;
Player defender = ...;
defender.setLife(20);
// predictNextCombatsRemainingLife returns 20 (no damage)
int remainingLife = 20;
int damage = defender.getLife() - remainingLife; // 0
// division by zero
int turns = (defender.getLife() + damage -1) / damage;
```
Impact: AI crashes during threat evaluation when attacker cannot damage defender.

#### 4. aiAggression condition parentheses changed semantics
Description:
Original code:
`&& (ComputerUtilMana.getAvailableManaEstimate(ai) > 0 || tradeIfTappedOut)`
`&& (ComputerUtilMana.getAvailableManaEstimate(defendingOpponent) == 0 || MyRandom.percentTrue(extraChanceIfOppHasMana))`
Changed to:
`&& (ComputerUtilMana.getAvailableManaEstimate(ai) > 0) || tradeIfTappedOut`
`&& (ComputerUtilMana.getAvailableManaEstimate(defendingOpponent) == 0) || MyRandom.percentTrue(extraChanceIfOppHasMana)`
This removes grouping, making `|| tradeIfTappedOut` apply to whole conjunction due to operator precedence, potentially allowing aggression 4 when mana conditions fail if tradeIfTappedOut true. Similar for opponent mana.
Test case:
```java
boolean tradeIfTappedOut = true;
int aiMana = 0;
int oppMana = 5;
boolean extraChance = false;
// Original requires aiMana>0||tradeIfTappedOut AND oppMana==0||extraChance
// New: (aiMana>0)||tradeIfTappedOut is true, but overall && with opponent part becomes (true || tradeIfTappedOut) && (false || false) => false
// Actually precedence: && binds tighter than ||, so expression becomes:
// (A && B) || tradeIfTappedOut && (C || D) ... complex
// Need to evaluate actual Java precedence: && before ||
// So aiMana>0 || tradeIfTappedOut is evaluated first? No, && has higher precedence than ||.
// Expression: P && Q || R && S... This is ambiguous.
```
Impact: AI aggression logic changed, may attack in situations where it shouldn't, risking loss of tricks or mana.

#### 5. defenderPreferences list may be empty leading to IndexOutOfBoundsException
Description:
AiAttackController constructor sets `defenderPreferences = getPreferredDefenderPlayers(ai, true)` and then `defendingOpponent = defenderPreferences.get(0)`. If ai.getOpponents() is empty, getPreferredDefenderPlayers returns empty list. Accessing get(0) throws IndexOutOfBoundsException.
Test case:
```java
Player ai = new Player(); // no opponents
AiAttackController ctrl = new AiAttackController(ai);
// getPreferredDefenderPlayers returns empty list
// defendingOpponent = defenderPreferences.get(0) -> IndexOutOfBoundsException
```
Impact: AI crashes in 1-player or test scenarios with no opponents.

#### 6. Aggregates.itemWithSoftmax returns last candidate on underflow, may be biased
Description:
When roll remains >=0 after loop, method returns candidates.get(candidates.size() - 1). With floating point rounding, this path can be taken more often than probabilistic expectation, biasing selection to last candidate.
Test case:
```java
List<String> candidates = List.of("A","B","C");
double temperature = 1e-9; // very small, weights huge
// totalWeight may overflow to Infinity, roll becomes NaN, loop never subtracts below zero, returns last
```
Impact: Selection bias and potential non-determinism.

#### 7. chooseNextDefender increments index without bounds check before refreshCombatants
Description:
chooseNextDefender does `while (++defenderPreferenceIndex < defenderPreferences.size())`. If defenderPreferenceIndex starts at -1, first increment to 0 is fine. However if defenderPreferences.size() is 0, loop never executes, but defendingOpponent remains previous. If list size is 1, index increments to 0 then 1, loop exits, defendingOpponent not updated. No reset of index on new combat, leading to stale index across multiple declareAttackers calls.
Test case:
```java
AiAttackController ctrl = new AiAttackController(ai);
// first declareAttackers, defenders size 1
ctrl.chooseNextDefender(); // index becomes 0, returns false
// second declareAttackers, index is now 0, ++ makes 1, loop exits
```
Impact: AI may skip defender rotation or retain stale defender.

#### 8. Change of AiCardMemory.MemorySet from TRICK_ATTACKERS to MANDATORY_ATTACKERS changes behavior
Description:
Line changed from TRICK_ATTACKERS to MANDATORY_ATTACKERS. These sets have different semantics; mandatory attackers may be fewer or different. Existing AI behavior relying on trick attackers for combat declaration may change, potentially reducing attack volume.
Test case:
```java
// Previously, AI remembered cards marked as trick attackers
// Now uses mandatory attackers only
// If a card is in TRICK_ATTACKERS but not MANDATORY_ATTACKERS, it will no longer be auto-declared
```
Impact: Backwards compatibility break, AI may under-attack.

#### 9. AiProps.MULTIPLAYER_DEFENDER_SOFTMAX_TEMPERATURE renamed from MULTIPLAYER_TARGETING_SOFTMAX_TEMPERATURE
Description:
Code references MULTIPLAYER_DEFENDER_SOFTMAX_TEMPERATURE but old property name may exist in user configs. If user had custom config for MULTIPLAYER_TARGETING_SOFTMAX_TEMPERATURE, it will be ignored, falling back to default 75, changing behavior silently.
Impact: User configs silently ignored.

#### 10. handlePlayingSpellAbility signature change from Consumer to Runnable breaks compatibility
Description:
Signature changed from `Consumer<SpellAbility>` to `Runnable`. Callers passing Consumer will no longer compile; existing code in codebase may rely on lambda receiving SpellAbility parameter. This is an API break.
Test case:
```java
ComputerUtil.handlePlayingSpellAbility(ai, sa, s -> s.setChosenTargets(...));
// Compilation error: Runnable.run() takes no arguments
```
Impact: Build break for dependent code; API incompatibility.

#### 11. setActivatingPlayer added in playStack/playNoStack may cause duplicate activation
Description:
playStack and playNoStack now call `sa.setActivatingPlayer(ai)` before cost check. If caller already set activating player, this may override with same or different value, potentially affecting triggers that check activation timing.
Test case:
```java
SpellAbility sa = ...;
sa.setActivatingPlayer(otherPlayer);
ComputerUtil.playStack(sa, ai, game);
// Activating player overwritten to ai
```
Impact: Trigger logic may misfire.

#### 12. Aggregates.itemWithSoftmax uses MyRandom.getRandom().nextDouble()*totalWeight, but totalWeight may overflow to Infinity
Description:
If temperature is very small and values differ greatly, exp((value-max)/temperature) may overflow to Infinity for max value, totalWeight becomes Infinity, roll becomes NaN, loop never finds roll<0, returns last candidate.
Impact: Softmax selection breaks for extreme temperature values.

### Performance and security

#### 13. itemWithSoftmax builds full List and values List for each call, O(n) memory allocation
Description:
AiAttackController.getPreferredDefenderPlayers called each combat declaration, with 2-8 opponents. Allocation overhead is minor but repeated in loops with timeout simulation. No caching.
Impact: Minor GC pressure, but acceptable.

#### 14. estimateCombatTurnsToKill uses AiCache with identity keys, may retain references to Player objects preventing GC
Description:
AiCache.getCached with List.of(AiCache::identity, AiCache::identity) caches result per Player instance identity. In long games, Player objects persist anyway, but cache may grow unbounded if players recreated.
Impact: Potential memory leak in long sessions.

## Verdict
Request Changes
The PR introduces TTK-based threat scoring and softmax defender selection for multiplayer AI, which is conceptually sound, but contains several correctness and safety issues: potential IndexOutOfBounds when no opponents, division by zero in TTK estimation, altered boolean precedence in aiAggression logic, and API signature break for handlePlayingSpellAbility. These need fixing before merge.

## Recommended Fixes
1. Guard defenderPreferences empty case and return early if no opponents.
2. Add damage>0 check before division in estimateCombatTurnsToKillChanged, return Integer.MAX_VALUE if damage<=0.
3. Restore parentheses in aiAggression condition to preserve original semantics.
4. Validate temperature config >0 before calling softmax, fallback to default.
5. Revert handlePlayingSpellAbility signature change or provide overload for backward compatibility.
6. Reset defenderPreferenceIndex when combat round starts.
7. Add unit tests for softmax selection distribution and TTK edge cases.

---

PR #11283, opened 2026-07-19
HOB cards from the preview panel

Fulgur14:


forge-gui/res/cardsfolder/upcoming/gandalf_goblins_bane_flameshape.txt
tool4ever, line 1: Wrong filename

tool4ever (approved):

Issue #11260, opened 2026-07-17
Move nextId statics into Game methods


Hanmac:
Move functions like Card::nextId into Game methods:

https://github.com/Card-Forge/forge/blob/e53ed40b202a1208db523d60373da35c34ca5556/forge-game/src/main/java/forge/game/Game.java#L67-L68

with that, the limit isn't global anymore

I don't know yet how to store the ID in Game object
maybe as a `Map<Class, Int>`?

Maybe refactor/remove GameEntityCache? it is only used for playerCache?
https://github.com/Card-Forge/forge/blob/e53ed40b202a1208db523d60373da35c34ca5556/forge-game/src/main/java/forge/game/Game.java#L262-L278

it seems GameEntityCache can be replaced by GameEntityViewMap?

Tyrathalis:
Thanks, this should also help a lot with AI/simulation use cases. Deterministic replay requires resetting all counters between games, currently handled via reflection.

Regarding scope, it looks like SpellAbility, SpellAbilityStackInstance, StaticAbility, ReplacementEffect, Trigger (and CombatView on the view side, with negative ids) all have the same pattern. Game.maxId covers Card/GameEntity ids.

Since #11203, GameCopier/GameSnapshot now preserve object IDs, so we should make sure the per-Game counter is part of the snapshotted state, otherwise we could see collisions after copying. Regarding Map<Class, Int> vs alternatives, I think a single shared per-Game counter would also work, and would keep IDs unique across types, which could be nice for debugging. Either way, the requirements that matter for determinism should just be, one, that counters travel wit the Game on snapshot, and two, that same construction order guarantees same IDs.

Happy to verify any branch with our fork-fidelity harness, forkcheck, which should properly exercise these sorts of issues.

Hanmac:
One Problem is once again, `getCardForUi`
for these ones, "host.getGame" returns null,
so i can't use `game.getNextIdFor(Trigger.class)`

Tyrathalis:
For the getCardForUi path, those cards are built with fromPaperCard(pc, null) and never join a Game. I don't think their IDs participate in game state or determinism at all, they should only need local uniqueness for view maps. So, the null-game case should be able to just fall back to a separate static counter without needing a Game threaded through.

[...]

PR #11285, opened 2026-07-19
Determinism hooks: per-thread MyRandom seeding + order-stable zone preparation

Tyrathalis:
Draft for the determinism conversation with @Hanmac (#11260) and the manabrew project (Discord #ai-plotting) — two small, separately revertable changes that make seeded simulation deterministic across concurrent games and across JVMs. Re-authors the determinism patch carried in witchesofthehill/forge (`d658cbc757`, credited as co-author) with two refinements. Intentionally excludes per-Game ID counters (#11260 is the better design and is maintainer-led) and `ImageKeys.clearCaches()` (useful, but not determinism — can be its own PR).

**1. `MyRandom`: the random source becomes an `InheritableThreadLocal`.** `setRandom()` — already documented as the deterministic-simulation hook — now seeds the *calling thread's* stream, so concurrent headless games in one JVM can't contaminate each other's sequences. Refinement over the manabrew patch: inheritable, so threads spawned from a seeded thread (AI evaluation helpers) continue the parent's stream rather than silently falling back to an unseeded one. Single-threaded behavior is unchanged. Design question for review: `InheritableThreadLocal` vs plain `ThreadLocal` — we chose inheritable so seeding survives helper-thread spawns; happy to hear if there's a spawn pattern where inheritance is wrong.

**2. `Match.preparePlayerZone`: iterate the deck section in sorted order.** `CardPool`'s backing map doesn't guarantee iteration order, so for a fixed seed the pre-shuffle library order, the random-foil roll sequence, and the card-id assignment order can all vary across JVMs. Refinement over sorting the finished list: sorting the *iteration* also fixes the RNG-consumption and id-assignment order, making zone preparation a pure function of (deck, seed) — the id-assignment half complements #11260's same-construction-order ⇒ same-ids invariant. Note for anyone with seed-pinned game corpora: behavior-invariant but trajectory-changing for a given seed.

**Tests** (3 new, all green; full forge-gui-desktop suite green, 289 tests): seeded-stream isolation between sibling threads; child-thread stream inheritance; library preparation invariant under CardPool insertion order and stable cross-JVM (sorted pre-shuffle order asserted directly).

**Fork-fidelity verification**: we ran this branch through our forkcheck harness (mid-game fork → digest-identical state → deterministic seeded completion under a model-driven bridge, the gate built for #11203): twin determinism 40/40, 0 static mismatches, 0 transport failures — no fidelity regression.

Portions authored with an AI assistant (Claude), reviewed by a human.

### DIFF
--- a/forge-core/src/main/java/forge/util/MyRandom.java
+++ b/forge-core/src/main/java/forge/util/MyRandom.java
@@ -30,15 +30,27 @@ import java.util.Random;
  * @version $Id$
  */
 public class MyRandom {
-    /** Constant <code>random</code>. */
-    private static Random random = new SecureRandom();
+    /**
+     * Per-thread random source. Thread-local so concurrent games in one JVM
+     * (headless simulation workers, parallel AI harnesses) that seed their
+     * stream via {@link #setRandom(Random)} cannot contaminate each other's
+     * sequences. Inheritable so threads spawned from a seeded thread (AI
+     * evaluation helpers) continue the parent's stream instead of silently
+     * falling back to an unseeded one. Single-threaded behavior is unchanged.
+     */
+    private static final ThreadLocal<Random> random = new InheritableThreadLocal<Random>() {
+        @Override
+        protected Random initialValue() {
+            return new SecureRandom();
+        }
+    };
 
     /**
      * <p>
      * percentTrue.<br>
      * If percent is like 30, then 30% of the time it will be true.
      * </p>
-     * 
+     *
      * @param percent an int.
      * @return a boolean.
      */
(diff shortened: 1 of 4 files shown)


Hanmac:
I wonder if `InheritableThreadLocal` is needed or `ThreadLocal.withInitial` might be enough 

I need to see if `InheritableThreadLocal.withInitial` exist  too

Tyrathalis:
I don't think InheritableThreadLocal.withInitial exists. ThreadLocal.withInitial does exist, but there isn't an Inheritable version, which is why the patch uses the anonymous-subclass initialValue() form.

It looks like there's currently a tradeoff here with ThreadUtil's pooled threads. Inheritance binds at thread creation, and cached/work-stealing pool threads are reused across games, so a pool thread either never sees the seed (plain) or can carry a stale game's Random (inheritable, if the thread was created mid-game). We went with making it inheritable, since that covers dedicated new Thread spawns and first-use pool growth from a seeded thread, and because our harness (and manabrew's) consumes randomness on the seeding thread anyway. But we could definitely switch to ThreadLocal.withInitial if you prefer it for maintainability reasons, just let me know.

In the longer run, it probably makes sense to move this in-Game, into a game.getRandom(), along the lines of #11260's per-Game id counters. I didn't want to scope-creep this PR, but I'm happy to move in that direction if preferred.

PR #11292, opened 2026-07-20
Joven and Chandler, merged

dracontes:
Dealing with this sooner than later. Also getting the hang of the improved variant system. Seems to work fine. I'd note that I can't get the cards disambiguated for the "Add Card to [Zone]" Dev tools but that is also true for the flavor name variants.


tool4ever:
no, as already noted by Jetz this doesn't even handle legality
also see discussion in #8520

👎

dracontes:
> no, as already noted by Jetz this doesn't even handle legality also see discussion in #8520
> 
> 👎

Ah, duly noted then.

PR #11295, opened 2026-07-20
Raise the image cache size so a large zone view fits in it

liamiak:
Fixes the freeze when opening a large zone view — most visibly the library during a tutor
search. Sorting, closing and reopening it froze again each time.

## Cause

A card with a local image file occupies two `ImageCache` entries: the decoded original under its
image key, and the scaled copy the panel paints under `key#WxH`. A large deck is effectively
singleton — my biggest are 555 cards over 546 distinct names — so one library view needs ~1,100
entries against a cap of 400. It evicts everything it just loaded and rebuilds on every refresh.

400 was never a considered choice for this workload; it is what `UI_IMAGE_CACHE_MAXIMUM` has
always defaulted to. Forge writes every preference to the profile, so raising the declared
default would move only profiles that have never been launched. This treats exactly 400 as
unset and uses 1500. Any other value was hand-edited — the pref has no settings UI — and is
left alone.

## Measured

500-card library, 477 of them with a local image file, driven through the same
`ImageCache.getImage` entry point `CardPanel` uses, on the EDT. Three runs each, ms:

| | master | this |
|---|---|---|
| first open | 3624 / 5224 / 5970 | 3575 / 3492 / 3541 |
| everything after first render | 13797 / 22490 / 21128 | 2494 / 2622 / 2676 |

Heap 503MB → 561MB on this workload. Master is heavy-tailed here and this is not; holding the
view is the difference.

## Soft values

Real card images decode to ~1.27MB, so 1500 entries is a ~1.9GB ceiling against the shipped
`-Xmx4096m` — entry count alone is the wrong bound at this size. Soft values make memory
pressure the thing that ultimately evicts, which is what this class' javadoc already claims it
does. `forge-gui-mobile`'s `ImageCache` scales its own capacity to device RAM for the same
reason.

## What this does not do

**Cards with no local image file get nothing from this.** `isPlaceholder` (`ImageCache.java:464`)
suppresses the scaled-copy `_CACHE.put` at :604 whenever the online fetcher is enabled, so their
rendered faces are discarded and redrawn on every refresh no matter what the cap is — measured
at ~6.2s per pass for 305 such cards. That is a separate bug and this PR does not touch it.

1500 fits one large library view with ~400 spare. It does not fit two.

No test: this is a timing and memory property of a cache, and a timing assertion in CI would be
flaky.

### DIFF
--- a/forge-core/src/main/java/forge/ImageKeys.java
+++ b/forge-core/src/main/java/forge/ImageKeys.java
@@ -9,6 +9,7 @@ import org.apache.commons.lang3.StringUtils;
 
 import java.io.File;
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 
 public final class ImageKeys {
     public static final String CARD_PREFIX           = "c:";
@@ -50,7 +51,7 @@ public final class ImageKeys {
     private static Map<String, Boolean> editionImageLookup = new HashMap<>();
 
     private static Map<String, Set<String>> editionAlias = new HashMap<>();
-    private static Set<String> toFind = new HashSet<>();
+    private static Set<String> toFind = ConcurrentHashMap.newKeySet();
 
     private static boolean isLibGDXPort = false;
 
@@ -91,14 +92,18 @@ public final class ImageKeys {
         return tokenKey.substring(ImageKeys.TOKEN_PREFIX.length());
     }
 
-    private static final Map<String, File> cachedCards = new HashMap<>(50000);
-    public static HashSet<String> missingCards = new HashSet<>();
+    private static final Map<String, File> cachedCards = new ConcurrentHashMap<>(50000);
+    public static Set<String> missingCards = ConcurrentHashMap.newKeySet();
     public static void clearMissingCards() {
         missingCards.clear();
     }
     public static File getCachedCardsFile(String key) {
         return cachedCards.get(key);
     }
+    // cachedCards, missingCards and toFind are concurrent collections so background image
+    // loaders can resolve files while the EDT does the same; a racing duplicate findFile
+    // probe is harmless (both threads store the same result). The slow disk probing runs
+    // without holding any lock so it can never stall the EDT.
     public static File getImageFile(String key) {
         if (StringUtils.isEmpty(key))
             return null;
@@ -412,7 +417,7 @@ public final class ImageKeys {
     public static boolean hasImage(PaperCard pc) {
         return hasImage(pc, false);
     }
-    public static boolean hasImage(PaperCard pc, boolean update) {
+    public static synchronized boolean hasImage(PaperCard pc, boolean update) {
         Boolean editionHasImage = editionImageLookup.get(pc.getEdition());
         if (editionHasImage == null) {
             String setFolder = getSetFolder(pc.getEdition());
--- a/forge-gui-desktop/src/main/java/forge/CachedCardImage.java
+++ b/forge-gui-desktop/src/main/java/forge/CachedCardImage.java
@@ -26,7 +26,12 @@ public abstract class CachedCardImage implements ImageFetcher.Callback {
             if (image == null) {
                 String key = card.getCurrentState().getImageKey(viewers);
                 Logger.debug("Fetch due to missing key: " + key + " for " + card);
-                fetcher.fetchImage(key, this);
+                fetcher.fetchImage(key, () -> {
+                    // The cache may hold placeholder renders scaled from before the
+                    // download completed; drop them so the real image shows.
+                    ImageCache.clearGeneratedVariants(key);
+                    onImageFetched();
+                });
             }
         }
     }
--- a/forge-gui-desktop/src/main/java/forge/CachedCardImage.java
+++ b/forge-gui-desktop/src/main/java/forge/CachedCardImage.java
@@ -13,6 +13,7 @@ public abstract class CachedCardImage implements ImageFetcher.Callback {
     final Iterable<PlayerView> viewers;
     final int width;
     final int height;
+    boolean fetchQueued;
 
     static final SwingImageFetcher fetcher = new SwingImageFetcher();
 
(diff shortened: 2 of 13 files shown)


tool4ever:
thanks, interesting ideas but rather large:
might take a while until a Dev with GUI interest has time to comb through it

liamiak:
Force-pushed a rewrite of this branch — "rather large" was fair, and I wanted to re-review
the whole thing with Opus 5.0 and restructure it for ease of evaluation rather than leave it
sitting as one big lump. Rebased onto current master (nothing there has touched these files).

It is now five commits, each one mechanism, each building and passing the desktop test suite
on its own. The description above is rewritten around measurements rather than assertions.

Instrumenting it changed the PR rather than just documenting it. I had assumed the cost was
image decode plus cache eviction thrash. It is neither: **~77% of a 6.2s reopen was
re-rendering card faces for cards with no local image** — ~284 of them at ~16ms, redrawn on
every refresh because placeholder renders were explicitly never cached ("the rendering speed
seems to be fast enough", which does not survive being multiplied by 284). The quadratic panel
bookkeeping I had led with turned out to be ~30ms of the 6156.

That also let me delete things rather than add them. Gone since you last looked:

- the new `ZONE_CARD_IMAGE_SIZE` preference and its `ForgePreferences` entry, along with the
  `prefs.save()` it did from inside a layout path — once a drawn card face is cached
  independently of display size, preloading no longer needs to know what size a view will ask
  for, so the preference had nothing left to do
- a whole-cache scan per downloaded image, replaced by a multimap of the affected keys
- a static font cache and a couple of other micro-optimisations that were not worth the
  questions they invited

Net effect is one fewer file touched than the version you commented on, and no new user
preference anywhere in the PR.

Two things I would rather flag than have you find:

- **Commit 3 (async loading) is not measurable on the machine I tested on.** ~57% of that
  collection's images are missing locally, and placeholder rendering has to stay on the EDT
  because `FCardImageRenderer` has mutable statics — so moving decode off it changes little
  there. The case it helps is a complete local collection, where the same numbers imply
  seconds saved, but that is inference from measurement rather than a measurement. If you
  would rather see it land separately, or not at all, that is a reasonable call.
- **All the numbers come from one machine and one collection state**, so the 77% is not a
  universal figure.

tool4ever:
good call, that will surely help

tool4ever:
Also worth noting Nr. 5 is related to a previous attempt: #7838

liamiak:
Thanks — that pointer led somewhere I didn't expect.

Reading #7838, the concern tehdiplomat raised there turns out not to be specific to that PR. Desktop's on-demand fetcher has never handled rate limiting at all: `SwingImageFetcher` downloads via `ImageIO.read(URL)`, which discards the HTTP status, so a 429 is indistinguishable from any other failed download and it keeps asking. Mobile got proper 429 handling and a cooldown in 678aae47 back in February, and the `scryfallCooldownTime` field it uses is declared on the shared `ImageFetcher` base class — only the mobile half was ever wired up. Desktop also keeps no gap between requests, while `GuiDownloadService` already keeps 100ms, and fetches go onto a work-stealing pool so a deck editor grid of missing images asks for all of them at once. I've opened #11524 for that; it's independent of this PR.

On the constraint itself, I went looking and found `docs/Card-Images.md`. It says "you can no longer predownload card images", and three lines later "You can do small pre-caching by loading your decks in the deck editor prior to playing to download just those images." As I read it, commit 5 sits on the second line rather than the first: same deck, same volume, same `fetchImage` path as opening the deck editor, and gated by the same `UI_ENABLE_ONLINE_IMAGE_FETCHER` preference that governs passive auto-download everywhere else. What differs is that it triggers itself at match start rather than waiting for you to open the editor. With #11524 in, it also inherits the pacing and the cooldown.

That said, automatic versus user-initiated is your call, not mine — if you'd rather pre-caching stayed something the player asks for, say so and I'll drop commit 5. The measured win in this PR is the rendering work; the fetching commit is the least evidenced part of it.

liamiak:
Force-pushed this down to one commit, 13 lines. Your first note — that it was large and would need a GUI dev to comb through — turned out to be the most useful thing said about it.

Re-measuring against current master (nothing upstream has touched these files in 379 commits, and it rebased clean), I had the causation wrong. I told you ~77% of a reopen was re-rendering card faces and that caching those renders was the fix. The re-rendering was real, but it was a symptom: the 400-entry cap cannot hold even one 500-card view, so the scaled copies were evicted and the faces had to be redrawn. Raise the cap, the scaled copies survive, and the re-rendering stops on its own.

Everything after the first render goes 17,890ms → 4,959ms from the cap alone. Caching the renders as well takes it to 1,722ms — but holds one full 488x680 image per card, about +550MB of heap against +40MB for this. That is a poor trade for the one remaining case, and it needs a bounded design before I would send it.

The rest is parked rather than dropped. The render cache needs that memory question answered; commits 3 and 5 I still have no measurement for, which is also the honest answer to the pre-caching question I asked you and you never needed to answer. If any of them are worth reviving, I would rather bring each back on its own with its own number.

kevlahnota:
IIRC before I adapat assetsmanager to mobile version, forge is using a loading cache which has 700 capacity and it introduce black textures since the mobile device cant handle large texture cache. Though it may sound good on modern mobile devices, the images are now sourced from scryfall and it has large dimensions which eats up memory. I just think the 1500 capacity is good for desktop with large RAM and good CPU for rendering it but not good specially for mobile devices. Desktop without large RAM may also suffer with large texture cache. Anyway did you test this on mobile device?

liamiak:
Re-measured this and the description above had the configuration wrong, so I have rewritten it.

The 28 Aug push was a rebase only, and I have just force-pushed once more to fix a comment in
the source that claimed every card takes two cache entries — only cards with a local image
file do. The change itself is the same 13 lines described on 23 Aug.

**I told you the measurement used an install with no downloaded card images, "the worst case".**
That is in fact the one configuration where this change does nothing. `isPlaceholder`
(`ImageCache.java:464`) suppresses the scaled-copy `_CACHE.put` at :604 whenever the online
fetcher is enabled, so a card with no local image file re-renders its face on every refresh
whatever the cap is. Measured: 305 such cards cost ~6.2s per pass, and cap 401 vs 1500 differ by
4% — noise. Turning the fetcher off on the same zone drops reopen to 0-1ms, which is how that
path ever looked fast.

**The case this actually fixes is a populated collection**, where each card takes two entries.
477 of 500 with local images, three runs each: everything after the first render goes 13.8-22.5s
to 2.5-2.7s, heap 503 → 561MB. First open is no worse.

The placeholder path looks like a separate bug — every card face redrawn on every zone refresh,
for anyone without a full local collection. Happy to open an issue, or to leave it if you would
rather look first.

One note on the soft values, since it is the line with no timing behind it: real card images
decode to ~1.27MB, so 1500 entries is a ~1.9GB ceiling against the shipped `-Xmx4096m`. I kept
them as the bound rather than the entry count, which is what `forge-gui-mobile`'s `ImageCache`
already does with device RAM. If you would rather the cap were derived from
`Runtime.maxMemory()` the same way, that is a small change and lands near 1500 at 4GB anyway.

Issue #11058, opened 2026-06-22
Java Stream + forEachCardInGame


Hanmac:
The Visitor interface used for `Game::forEachCardInGame` could/should probably be reworked using JavaStream.
Either as Predicate, Consumer or Collector.

For this to clean up, classes like Zone/PlayerZone should return a (filtered) Stream of their cards instead of CardCollection objects.

I need to check out if `Stream.concat` or `Stream.Builder` is better for this.

Hanmac:
[Stream.concat](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/stream/Stream.html#concat(java.util.stream.Stream,java.util.stream.Stream)) has some warnings about Concatting multiple streams (in our case, multiple zones)
it also suggests `flatMap` like this:
```java
Stream<T> concat = Stream.of(s1, s2, s3, s4).flatMap(s -> s);
```
but wasn't there a problem with such method on Android? or did i confuse it with another?

Hanmac:
PlayerZoneBattlefield storing MeldedCards doesn't make much sense
that should probably be stored in Game or Player object?

Jetz72:
> it also suggests `flatMap` [...] but wasn't there a problem with such method on Android? or did i confuse it with another?

We currently use `flatMap` in several places, so that was probably a different one.

> PlayerZoneBattlefield storing MeldedCards doesn't make much sense
> that should probably be stored in Game or Player object?

We have a `ZoneType.merged` for mutated permanents. Could use that to generally hold the components of permanents represented by multiple cards.

Hanmac:
@tool4ever 
InboundToken are used because a created Token or MakeCard might have its own ETB ReplacementEffect,
that the ReplacementHandler might not see without it when using `forEachCardInGame`

If the logic of the ReplacementHandler is updated to look at more at just `forEachCardInGame`,
we might remove InboundToken again

Hanmac:
Sometimes, cards are filtered by a preList
https://github.com/Card-Forge/forge/blob/fbab58912ecf368405059cf11df5aa62ec1c7874/forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java#L102-L103
https://github.com/Card-Forge/forge/blob/fbab58912ecf368405059cf11df5aa62ec1c7874/forge-game/src/main/java/forge/game/GameAction.java#L1098-L1100

this should probably be handled using `distinct`, so that the preList ones are handled first

PR #11301, opened 2026-07-20
Speed up combat legality checks on heavy boards

liamiak:
## Problem

On a heavy board, working out attacks or blocks takes tens of seconds. Profiling
pointed at two redundancies in the shared combat legality code — both recompute
results that cannot differ, so removing them changes no decision.

Measured on a synthetic board of 40 creatures per side, this branch against master
on the same machine:

| | master | this branch |
|---|---|---|
| declare attackers | 17.7 s | **8.7 s** |
| declare blockers | 16.3 s | **6.8 s** |

Smaller boards improve proportionally (at 20 a side: 7.8 s → 5.2 s and 1.3 s → 0.8 s).

These are in `forge-game`, not the AI, so they also speed up the equivalent
legality checks during human play.

## Cause 1 — every static ability check copied the board

`Game.getCardsIn(Iterable<ZoneType>)` allocates a fresh `CardCollection` and copies
every card from every listed zone on each call. Every static-ability helper reaches
for `ZoneType.STATIC_ABILITIES_SOURCE_ZONES` through it — `getBlockCost`,
`StaticAbilityCantAttackBlock.cantBlockBy`,
`StaticAbilityAssignCombatDamageAsUnblocked`, and others. Combat evaluation calls
those thousands of times between two zone changes, so the same list is rebuilt over
and over.

`Game` now caches that list against a counter bumped by every zone mutation.

Two notes for reviewers:

- The counter is bumped at each point in `Zone` that mutates `cardList`, **not** in
  `Zone.onChanged()`, because `PlayerZone` overrides that without calling `super`.
  Missing a mutation would serve a stale list, so all of `add`, `remove`,
  `setCards`, `removeAllCards`, `reorder`, `shuffle` and `sort` bump it.
- The returned view is shared rather than freshly built, so unlike
  `getCardsIn(Iterable)` it must only be read. That is documented on the method; the
  three call sites converted here only iterate it.

## Cause 2 — block requirements were checked after block costs

`CombatUtil.mustBlockAnAttacker` tested `getBlockCost(...)` — which walks every
static ability in play — before `attackerLureSatisfied(...)`, which is a handful of
keyword tests. Both guard the same `continue`, so an attacker that places no
requirement on this blocker was skipped either way, but only after paying for a full
scan of the board. Since that is the overwhelmingly common case, swapping the two
guards lets it skip out first.

`getBlockCost` also allocates its `Cost` lazily now, as it returns `null` unless
something in play actually adds a cost to blocking.

## Decisions are unchanged

This is a performance change only, and was checked as such:

- The AI declares the same blockers at every board size before and after.
- `CombatUtil.canBlock(attacker, blocker, combat)` is called the same number of
  times, so control flow is unchanged — only its cost drops (892 µs → 128 µs per
  call at 40 a side).
## Tests are a separate commit — take or leave them

The fix is the first commit; the tests are the second and touch nothing else, so
they can be dropped without rebasing anything.

They exist because `mustBlockAnAttacker` — the method cause 2 reorders — is only
reached when an attacker places a requirement on a blocker, so no existing test
touches it: a board without lure or "must be blocked if able" attackers runs
straight past. Two cases, kept minimal per the contributing guidance on not
growing the suite unnecessarily.

The argument for them is about what comes next rather than this PR: the method is
an inviting optimisation target, since its result depends only on the blocker, so
caching it per blocker looks obviously safe until you notice it also depends on
which blocks have already been assigned. If you would rather not carry them, drop
the second commit — the reorder stands on its own, as both guards end in the same
`continue` and the same attackers pass either way.

Full `forge-gui-desktop` suite: 288 tests, the only failure being
`CardRankerTest.testRank`, which fails intermittently on master for unrelated
reasons (a 1 s timeout that is really measuring a one-time data load) and is
addressed separately in #11299.

## Relationship to #10507

No file overlap with #10507, which is in `Card` / `CardState` /
`KeywordCollection` / `AiAttackController`. This branch is in `Game`, `Zone`,
`CombatUtil` and two static-ability classes, so the two should be independent — and
the largest remaining hot spot I measured, `CardState.getStaticAbilities()`
allocating per call, is that PR's ground rather than this one's.

---

Assisted by an AI coding agent (Claude Opus 4.8); co-author credit is in the commit
per the contributing guidelines. Happy to share the profiling harness and benchmark
used for the numbers above if that is useful — they are left out here since the
benchmark takes ~100 s and asserts nothing, so it is a tool rather than a test.

### DIFF
--- a/forge-game/src/main/java/forge/game/Game.java
+++ b/forge-game/src/main/java/forge/game/Game.java
@@ -630,6 +630,34 @@ public class Game {
         return cards;
     }
 
+    // Every static ability check asks for the cards that could be the source of one, and
+    // building that list copies most of the game. Combat evaluation does that thousands of
+    // times between two zone changes, so cache the list against a counter that every zone
+    // mutation bumps.
+    private int zoneVersion;
+    private CardCollectionView staticAbilitySourceCards;
+    private int staticAbilitySourceVersion = -1;
+
+    public synchronized void bumpZoneVersion() {
+        zoneVersion++;
+    }
+
+    /**
+     * The cards that may be the source of a static ability, i.e. those in
+     * {@link ZoneType#STATIC_ABILITIES_SOURCE_ZONES}.
+     * <p>
+     * Unlike {@link #getCardsIn(Iterable)} the returned view is shared rather than freshly
+     * built, so callers must only read it: modifying it corrupts the cache, and it is
+     * stale after the next zone change.
+     */
+    public synchronized CardCollectionView getStaticAbilitySourceCards() {
+        if (staticAbilitySourceCards == null || staticAbilitySourceVersion != zoneVersion) {
+            staticAbilitySourceCards = getCardsIn(ZoneType.STATIC_ABILITIES_SOURCE_ZONES);
+            staticAbilitySourceVersion = zoneVersion;
+        }
+        return staticAbilitySourceCards;
+    }
+
     public CardCollectionView getCardsInOwnedBy(final Iterable<ZoneType> zones, Player p) {
         CardCollection cards = new CardCollection();
         for (final ZoneType z : zones) {
--- a/forge-game/src/main/java/forge/game/combat/CombatUtil.java
+++ b/forge-game/src/main/java/forge/game/combat/CombatUtil.java
@@ -332,22 +332,22 @@ public class CombatUtil {
     }
 
     public static Cost getBlockCost(Game game, Card blocker, Card attacker) {
-        Cost blockCost = new Cost(ManaCost.ZERO, true);
+        // Allocated lazily: hardly any board has an effect that adds a cost to blocking,
+        // and this runs per blocker/attacker pair while blocks are being worked out.
+        Cost blockCost = null;
         // Sort abilities to apply them in proper order
-        boolean noCost = true;
-        for (Card card : game.getCardsIn(ZoneType.STATIC_ABILITIES_SOURCE_ZONES)) {
+        for (Card card : game.getStaticAbilitySourceCards()) {
             for (final StaticAbility stAb : card.getStaticAbilities()) {
                 Cost c1 = stAb.getBlockCost(blocker, attacker);
                 if (c1 != null) {
+                    if (blockCost == null) {
+                        blockCost = new Cost(ManaCost.ZERO, true);
+                    }
                     blockCost.add(c1);
-                    noCost = false;
                 }
             }
         }
 
-        if (noCost) {
-            return null;
-        }
         return blockCost;
     }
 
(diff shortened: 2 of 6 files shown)


forge-game/src/main/java/forge/game/combat/CombatUtil.java
```diff
@@ -753,11 +753,17 @@ public static boolean mustBlockAnAttacker(final Card blocker, final Combat comba
...
         final Player defender = blocker.getController();
         for (final Card attacker : attackers) {
-            if (getBlockCost(blocker.getGame(), blocker, attacker) != null) {
+            // Check the block requirement before the block cost. Both skip this attacker,
+            // but getBlockCost walks every static ability in play while
+            // attackerLureSatisfied is a handful of keyword tests - and the common case,
+            // an attacker that places no requirement on this blocker, is skipped by the
+            // latter. Testing the cost first made every combat pay for a full scan of the
+            // board per attacker, per blocker.
+            if (attackerLureSatisfied(attacker, blocker, combat.getBlockers(attacker))) {
```
tool4ever, line 762: the relevant keywords checked with this aren't that common either

but more importantly they aren't "real" keywords and will eventually be turned into static abilities anyway

forge-game/src/main/java/forge/game/Game.java
```diff
@@ -630,6 +630,34 @@ public CardCollectionView getCardsIn(final Iterable<ZoneType> zones) {
         return cards;
     }
 
+    // Every static ability check asks for the cards that could be the source of one, and
+    // building that list copies most of the game. Combat evaluation does that thousands of
+    // times between two zone changes, so cache the list against a counter that every zone
+    // mutation bumps.
+    private int zoneVersion;
```
tool4ever, line 637: seems a bit messy, imo the right approach is outlined in #11058

we should have methods that let us iterate arbitrarily without too much overhead

tbh I'm still surprised putting them into CardCollections costs that much peformance

liamiak:
Thanks for taking a look, and for the pointer to #11058 — agreed that lazy iteration is the better end state than what I did here.

On the surprise that `CardCollection` allocation costs this much: it surprised me too, so I profiled it before and after on a synthetic board of 40 creatures per side (temporary instrumentation, not in the PR):

| | before | after |
|---|---|---|
| `CombatUtil.canBlock(attacker, blocker, combat)` | 892 µs/call | 128 µs/call |
| declare attackers | 17.7 s | 8.7 s |
| declare blockers | 16.3 s | 6.8 s |

The call count of `canBlock` is identical before and after, so it's purely the per-call cost dropping. The reason it bites so hard isn't one allocation being expensive — it's the multiplier: combat evaluation calls the static-ability helpers thousands of times between two zone changes, and `getCardsIn(STATIC_ABILITIES_SOURCE_ZONES)` rebuilds a full board copy on each. (There's a second one a layer down too — `CardState.getStaticAbilities()` allocates a fresh `FCollection` and runs an update pass per call — which I didn't touch here.)

So the underlying problem is exactly the one #11058 is aimed at. Given that, would you prefer:

- I **drop the cache** from this PR and keep just the `mustBlockAnAttacker` guard reorder (that half is independent — both guards end in the same `continue`, so the same attackers pass either way, and it's a clean win on its own), and the allocation problem gets fixed properly via #11058-style streaming; **or**
- keep the version-counted cache as an interim measure until #11058 lands.

Happy either way — I'd rather this match where you want the codebase to go than land as-is. And on the lure keywords: fair, I only reordered around them, so that stays correct whether they're keywords or statics later.

tool4ever:
imo both areas should just be fixed properly

liamiak:
Closing this in favor of #11314, which subsumes it. That PR's `hasPossibleBlocker` guard is the same "skip block assignment when nothing can legally block" early-out this PR added, and its `isUnblockedCombatAlwaysSafe` covers the broader "taking the attack unblocked is provably safe" case more thoroughly (with proper guards for must-block, commander lethal, poison, and danger thresholds). No need to carry two versions of the same optimization — thanks @Madwand99.

PR #11297, opened 2026-07-20
The Misty Mountains Cold (HOB)

Fulgur14:

PR #11302, opened 2026-07-20
Assign 15 cards that were falling back to the UNKNOWN set

liamiak:
## What this does

Startup currently logs a warning for every card whose script exists but that no
edition file lists, and drops it into the UNKNOWN set:

```
The card A-Goldspan Dragon was not assigned to any set. Adding it to UNKNOWN set... to fix see res/editions/ folder.
```

Fifteen of those are simply missing entries. This adds them — 15 lines across 10
edition files, no code.

## The fourteen Alchemy rebalances

Each of these has a `[rebalanced]` section in the set it belongs to, and entries
for its siblings, but was itself never listed. Each new line mirrors the collector
number, rarity and artist of the original printing, exactly as the surrounding
entries do:

| set | added |
|---|---|
| Kaldheim | `A-41` Alrund's Epiphany, `A-139` Goldspan Dragon, `A-169` Esika's Chariot, `A-255` Faceless Haven |
| Zendikar Rising | `A-24` Luminarch Aspirant, `A-232` Omnath, Locus of Creation |
| Alchemy Horizons: Baldur's Gate | `A-217` Monster Manual, `A-232` Baba Lysaga, Night Witch |
| Strixhaven | `A-41` Divide by Zero |
| Throne of Eldraine | `A-125` Fires of Invention |
| Kamigawa: Neon Dynasty | `A-152` Kumano Faces Kakkazan |
| Innistrad: Crimson Vow | `A-63` Hullbreaker Horror |
| Innistrad: Midnight Hunt | `A-112` The Meathook Massacre |
| Modern Horizons 2 | `A-145` Unholy Heat |

## The Misty Mountains Cold

Added in #11297 — the card script landed but the entry in `The Hobbit.txt` did
not, so it was falling through to UNKNOWN. Added as `104 R ... @Rovina Cai` per
Scryfall.

## Result

Before: 20 "not assigned to any set" warnings at startup. After: 5.

The remaining five are a separate question and are deliberately **not** touched
here — see the comment below.

### DIFF
--- a/forge-gui/res/editions/Alchemy Horizons Baldur's Gate.txt
+++ b/forge-gui/res/editions/Alchemy Horizons Baldur's Gate.txt
@@ -321,7 +321,9 @@ A-208 C A-Druidic Ritual @Vincent Christiaens
 A-209 R A-Earthquake Dragon @Johan Grenier
 A-210 U A-Emerald Dragon @Diego Gisbert
 A-215 U A-Jade Orb of Dragonkind @Olena Richards
+A-217 R A-Monster Manual @David Gaillet
 A-223 U A-Split the Spoils @Edgar Sánchez Hidalgo
+A-232 R A-Baba Lysaga, Night Witch @Slawomir Maniak
 A-239 U A-Krydle of Baldur's Gate @Bryan Sola
 A-243 M A-Minsc & Boo, Timeless Heroes @Andreas Zafiratos
 A-259 C A-Lantern of Revealing @Eytan Zana
--- a/forge-gui/res/editions/Innistrad Crimson Vow.txt
+++ b/forge-gui/res/editions/Innistrad Crimson Vow.txt
@@ -438,6 +438,7 @@ A-48 C A-Binding Geist @Campbell White
 A-52 U A-Cobbled Lancer @Igor Kieryluk
 A-58 R A-Dreamshackle Geist @Andreas Zafiratos
 A-62 U A-Gutter Skulker @Viko Menezes
+A-63 R A-Hullbreaker Horror @Svetlin Velinov
 A-66 C A-Lantern Bearer @Zoltan Boros
 A-69 U A-Mischievous Catgeist @Denman Rooke
 A-81 C A-Stitched Assistant @Andrey Kuzinskiy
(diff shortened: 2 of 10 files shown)


liamiak:
### The five remaining warnings — needs a maintainer decision

After this PR, five cards still fall through to UNKNOWN:

```
Blinding Radiance
Goblin Bruiser
Ogre Painbringer
Titanic Pelagosaur
Treetop Recluse
```

I did not touch these, because I do not think an edition entry is the right fix
and the correct one is a content decision rather than a data one.

All five arrived in the same 2019 commit, 3c1b1068 *"Add balance of ANA cards as
found at https://scryfall.com/sets/ana"*, alongside about a dozen others. The rest
of that batch have edition entries; these five never got one, and have been
warning ever since.

The reason they can't get one: **they no longer exist.** Querying Scryfall today:

- `set:ana` returns 46 cards, all lands — which matches
  `res/editions/Arena New Player Experience.txt`, which contains only basic lands,
  Encroaching Wastes and the snow-covered lands.
- Searching each of the five by exact name returns 404. They are not in ANA, and
  not in any other set.

So they look like Arena New Player Experience tutorial cards that WotC has since
removed, and Scryfall delisted along with them. There is no set to assign them to.

Options as I see them:

1. **Delete the five card scripts.** They are cards that no longer exist in any
   product, so nothing can reference them legitimately. This silences the warnings
   properly rather than papering over them.
2. **Leave them and suppress the warning** — e.g. via the `filtered` set in
   `CardDb`, though that currently only carries funny/un-set cards when
   `loadNonLegalCards` is off, so it would be a slight repurposing.
3. **Leave them as they are**, accepting five permanent startup warnings.

I have deliberately not done (1) here since removing card content felt like your
call rather than mine, but I am happy to push it as a follow-up if that is the
direction you would like — it is a five-file deletion.

paulsnoops:
These cards no longer exist and were intentionally removed from edition files but we kept the scripts so they don't break decks they already existed in and so people still have the option to use them, we don't want them back in the original editions

paulsnoops:
The Misty Mountains Cold just wasn't added to the edition yet because Scryfall didn't have it when I updated the file. I'll be doing the daily spoiler update today anyway

P.s during spoiler season the scripts and editions don't always line up, we do daily updates so a bit of patience is needed

PR #11308, opened 2026-07-21
Update deadly_cover_up.txt

BigCrunch22:
Oracle Text of Deadly Cover-Up
---
As an additional cost to cast this spell, you may collect evidence 6.

Destroy all creatures. If evidence was collected, exile a card from an opponent’s graveyard. Then search its owner’s graveyard, hand, and library for any number of cards with that name and exile them. **_That player shuffles, then draws a card for each card exiled from their hand this way._**

---

I noticed when I was manually checking for Oracle text and descriptions for cards that start with the letter D is that this card is missing an effect since its Oracle text was incomplete. With assistance from Claude, the missing effect from the card was added so that it functions how it should.


tool4ever (approved):
thanks

Hanmac (approved):

PR #11309, opened 2026-07-21
DestoryEffect: filter for canBeDestroyed first

### LINKED ISSUE #11291: Inspirit, Flagship Vessel: Indestructible artifacts can sometimes be destroyed by  "Destroy" sweepers
I have been playing the Counter Intelligence EoE precon in the Realm of Legends. When the AI overloads a Vandalblast, all my artifacts are destroyed and go to graveyard even when they should be Indestructible 

1. Cast and fully station Inspirit to give all your artifacts (exept Inspirit) Hexproof and Indestructible
2. Have the AI cast Vandalblast overloaded. (There is another black "destroy" boardwipe that I observed this issue with, but I didn't catch the name because the game ended from triggers before I thought of recording it)
3. Artifacts that were on board before Inspirit may remain but artifacts cast after Inspirit will be destroyed

It seems like the intended result per the rules would be that all artifacts are destroyed simultaneously, so Inspirit would be the only artifact that goes to the graveyard.

I think this happens because the Forge game engine does actions sequentially in the order the things entered the battlefield, so it destroys Inspirit and then no longer considers the other artifacts "indestructible" when it gets around to destroying them.

Proposed solution: one of
 - rework "destroy" sweepers as a double pass, first pass determines if the affected permanent will actually be destroyed and the second pass actually destroys the doomed permanents
 - add and remove keywords as a SBA so they don't get updated in the middle of something resolving (probably both harder and less useful)
 - something else?

In-game logs:
Image

Hanmac:
Closes #11291

that should filter for `canBeDestroyed` while still respecting `AlwaysRemember`

the ones inside `internalDestroy` are still for the ReplacementEffects

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/DestroyEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/DestroyEffect.java
@@ -2,13 +2,16 @@ package forge.game.ability.effects;
 
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import forge.game.Game;
 import forge.game.GameActionUtil;
 import forge.game.ability.AbilityKey;
 import forge.game.ability.SpellAbilityEffect;
 import forge.game.card.Card;
+import forge.game.card.CardCollection;
 import forge.game.card.CardCollectionView;
+import forge.game.card.CardLists;
 import forge.game.card.CardUtil;
 import forge.game.card.CardZoneTable;
 import forge.game.spellability.SpellAbility;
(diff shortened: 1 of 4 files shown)


forge-game/src/main/java/forge/game/ability/effects/DestroyEffect.java
tool4ever, line 119: I wonder if we need logic for the inverse case too: something would grant Indestructible from other zone 🤔
(because `GameAction.destroy` checks for it again)

Hanmac: @tool4ever `Erebos's Titan` with
> As long as your opponents control no creatures, this creature has indestructible. (Damage and effects that say “destroy” don’t destroy it.)

Could cause it to become indestructible if the other creatures are destroyed first?

Hanmac:
@tool4ever maybe move this part of Indestructible about `Destroy` by Effect from there:
https://github.com/Card-Forge/forge/blob/57fb7b37862aab8ff4cc9fa9335a709dfd5dac3b/forge-game/src/main/java/forge/game/GameAction.java#L2136-L2148

From the `canBeDestroyed` into a `CantHappen` Layer for the bottom part?

Indestructible by StateBasedAction can still be handled there:
https://github.com/Card-Forge/forge/blob/57fb7b37862aab8ff4cc9fa9335a709dfd5dac3b/forge-game/src/main/java/forge/game/GameAction.java#L1459-L1462

Being part of ReplacementEffect might help us with the "Look-Back" logic?

I also wonder why `canBeDestroyed` looks for `(isCreature() && getNetToughness() <= 0)`?
Probably some leftovers

tool4ever:
or just check only if sa = null

Hanmac:
@tool4ever i will try to find a way to do that the next days

My gut is telling me it would help if Indestructible would use CantHappen ReplacementEffects
and then we need to check LastStateBattlefield for RE's so when a card is destroyed, it doesn't cause the StaticAbilities to mess with the Replacement Effects. (See Erebos's Titan)

Hanmac:
@tool4ever i updated Indestructible to a ReplacementEffect
(only used by SA?)

now we need to find a way to make it use LastStateBattlefield for "Erebos's Titan"

PR #11317, opened 2026-07-21
Added support for igd v2 gateways and multi-device upnp discovery

GiovanniMenzano:
Fix automatic UPnP port forwarding on IGD v2 gateways and on routers that advertise port arguments as 'ui4' instead of the usual 'ui2'.

jUPnP's default 'PortMappingListener' has two compatibility limitations:

1) it uses an exact type comparison against 'InternetGatewayDevice:1'. Since the device version is part of that type, an 'InternetGatewayDevice:2' gateway is rejected before its connection services are inspected, despite implementing the IGD v1 functionality needed for port mapping;

2) PortMappingAdd and PortMappingDelete callbacks pass port numbers as 'UnsignedIntegerTwoBytes' values, matching the standard UPnP 'ui2' port datatype. For some reasons, some routers advertise those action arguments as 'ui4' instead.
When this happens, jUPnP rejects the already-typed 'ui2' value because it does not match the datatype declared by the router, resulting in an InvalidValueException even though the numeric port is valid.

This change uses version-compatible IGD discovery and supplies port numbers as decimal strings, allowing jUPnP to convert each value to the datatype actually advertised by the router.


GiovanniMenzano:
The original failure was reproduced on a network containing a Netgear RAX10 access point and a D-Link DVA-5592 gateway. I also made a final test on the gateway to confirm that the dynamic mapping is created and removed correctly.

tool4ever:
Hmn, I'm not sure this wouldn't be fixed better in the upstream library...? 🤔

GiovanniMenzano:
You're absolutely right. At first, I thought that opening a PR, have it accepted and released in a library like jupnp might take a long time. Also, to be completely honest, I feel like I shouldn't be near a networking library used for so many projects 😅
Anyway, the fix obviously belongs upstream, so I'll try and see what I can do.
Do you think I should keep this PR open while the jupnp change is being reviewed, and after the new version is released, replace the current implementation with a dependency update?

GiovanniMenzano:
I did the fix in the library, hoping the maintainer takes a look at it soon. For reference:
https://github.com/jupnp/jupnp/pull/367

GiovanniMenzano:
fix merged, waiting for the release 😁

GiovanniMenzano:
Hi @tool4ever, release 3.0.5 has been published. I just pushed the change

tool4ever (approved):
nice work

GiovanniMenzano:
thanks!

PR #11318, opened 2026-07-22
AI: hoist loop-invariant life-loss check out of mana-source listing

liamiak:
### What

`ComputerUtilMana.getAvailableManaSources` decides, per mana source, whether a source that deals damage to you when tapped is safe to use. The guard for that is:

```java
if (ai.canLoseLife() && !ai.cantLoseForZeroOrLessLife()) { ... }
```

Neither call depends on the individual source, but each is a full-board scan:
- `canLoseLife()` → `StaticAbilityCantGainLosePayLife.anyCantLoseLife` walks static abilities
- `cantLoseForZeroOrLessLife()` → `ReplacementHandler.cantHappenCheck(GameLoss, ...)` runs a whole-game replacement check

So on a large board this evaluates the pair **once per mana source, per mana-feasibility check** — and mana feasibility (`ComputerUtilCost.canPayCost` → `groupSourcesByManaColor` → `getAvailableManaSources`) is one of the hottest AI paths, re-run for every candidate spell on every priority pass.

This hoists the expression to a single evaluation per call.

### Why it's safe

Both methods are pure reads, and nothing inside the loop mutates the player's life-loss state, so the value is loop-invariant — source selection is unchanged.

### Measured

Profiling a 4-player AI-vs-AI sim on a large multicolour deck, the per-source "deals lethal on tap" portion of the mana-source listing dropped from **~7.4s to ~0.01s** (≈2,280µs → ≈3µs per source), with identical results and games completing normally.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
@@ -1357,6 +1357,10 @@ public class ComputerUtilMana {
         // 2. Search for mana sources that have a certain number of abilities
         // 3. Use lands that produce any color many
         // 4. all other sources (creature, costs, drawback, etc.)
+        // Whether tapping a source could kill us depends only on the player, not on the individual
+        // source, but canLoseLife()/cantLoseForZeroOrLessLife() each run a full-board scan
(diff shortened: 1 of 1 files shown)


tool4ever (approved):
good find, thanks

Hanmac (approved):

PR #11323, opened 2026-07-22
Stream card zones

Hanmac:
Part of #11058

doesn't replace `Game::forEachCardInGame` yet and also doesn't handle special places like InboundToken

### DIFF
--- a/forge-game/src/main/java/forge/game/Game.java
+++ b/forge-game/src/main/java/forge/game/Game.java
@@ -56,6 +56,7 @@ import org.tinylog.TaggedLogger;
 
 import java.util.*;
 import java.util.function.Predicate;
+import java.util.stream.Stream;
 
 /**
  * Represents the state of a <i>single game</i>, a new instance is created for each game.
@@ -789,6 +790,11 @@ public class Game {
         };
         forEachCardInGame(visitor);
         return all;
+    }
+
+    public final Stream<Card> streamCardsIn(final Iterable<ZoneType> zones) {
+        PlayerCollection players = getPlayers();
+        return StreamUtil.stream().flatMap(z -> z == ZoneType.Stack ? getStackZone().streamCards() : players.flatMap(p -> p.streamCardsIn(z)));
     }
 
     public final GameAction getAction() {
--- a/forge-game/src/main/java/forge/game/Game.java
+++ b/forge-game/src/main/java/forge/game/Game.java
@@ -794,7 +794,7 @@ public class Game {
 
     public final Stream<Card> streamCardsIn(final Iterable<ZoneType> zones) {
         PlayerCollection players = getPlayers();
-        return StreamUtil.stream().flatMap(z -> z == ZoneType.Stack ? getStackZone().streamCards() : players.flatMap(p -> p.streamCardsIn(z)));
+        return StreamUtil.stream(zones).flatMap(z -> z == ZoneType.Stack ? getStackZone().streamCards() : players.stream().flatMap(p -> p.streamCardsIn(z)));
     }
 
     public final GameAction getAction() {
--- a/forge-game/src/main/java/forge/game/player/Player.java
+++ b/forge-game/src/main/java/forge/game/player/Player.java
@@ -63,6 +63,7 @@ import java.util.*;
 import java.util.Map.Entry;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * <p>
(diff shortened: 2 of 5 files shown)


tool4ever:
hmn, I think I need to try and see if they would work / be faster, just need to find some time

jamincollins:
## Findings
#### 1. Flashback zone handling returns activatable cards instead of zone contents
Player.streamCardsIn(ZoneType zoneType, boolean filterOutPhasedOut) treats ZoneType.Flashback as a special case and returns getCardsActivatableInExternalZones(true).stream(). This returns cards the player can activate from graveyard, exile, library, command, sideboard and opponents zones, not cards actually in a Flashback zone. Flashback is not a real ZoneType storage zone; it is a mechanic. The stream therefore does not represent zone contents and violates the method contract of "stream cards in zone".

Test case:
```java
Player p = game.getPlayer(0);
Stream<Card> flashbackStream = p.streamCardsIn(ZoneType.Flashback);
List<Card> cards = flashbackStream.toList();
// cards contains activatable cards from graveyard/exile/etc, not empty Flashback zone
// Expected: empty stream or cards from actual Flashback zone if it existed
```
Impact: Callers expecting zone contents will receive unrelated activatable cards, leading to incorrect game logic.

#### 2. Zone base streamCards ignores filter parameter
Zone.streamCards(final boolean filter) returns cardList.stream() unconditionally. The filter flag is ignored in the base implementation. Only PlayerZoneBattlefield overrides to filter phased out cards when filter is true. Non-Battlefield zones therefore return phased out cards even when filterOutPhasedOut=true is requested.

Test case:
```java
Zone library = player.getZone(ZoneType.Library);
Card phased = new Card();
phased.setPhasedOut(true);
library.add(phased);
long count = library.streamCards(true).count();
// count == 1, phased out card is included
// Expected: 0 when filter=true
```
Impact: Data integrity issue; phased out cards leak into streams from hand, library, graveyard, etc.

#### 3. Null zone returns empty stream but Flashback bypasses null check
Player.streamCardsIn returns Stream.of() when getZone(zoneType) returns null. Good null safety. However Flashback case returns activatable cards without checking if zone exists, and filterOutPhasedOut parameter is ignored for Flashback.

Test case:
```java
Player p = new Player(game);
Stream<Card> s = p.streamCardsIn(ZoneType.Hand); // getZone may return null for invalid zone
s.toList(); // returns empty, no NPE
// Flashback case never checks zone null
```
Impact: Inconsistent handling; Flashback special case could throw if getCardsActivatableInExternalZones depends on zones being initialized.

#### 4. Stack owner filter uses equals which may be reference based
Player.streamCardsIn for ZoneType.Stack filters by c.getOwner().equals(this). If Card owner reference is different instance but equal by id, equals may fail depending on implementation. Safer to compare identity or use getPlayerId.

Test case:
```java
Card c = new Card();
c.setOwner(otherPlayer);
game.getStackZone().add(c);
Stream<Card> s = player.streamCardsIn(ZoneType.Stack);
s.findAny().isPresent(); // may be false positive/negative
```
Impact: Potential missed cards in stack streaming for correct player.

#### 5. Stream reuse and multiple enumeration risk
All streamCardsIn methods return fresh streams. Callers may assume stream can be reused. Lazy streams over mutable card lists can lead to ConcurrentModificationException if cards move between zones during iteration.

Test case:
```java
Stream<Card> s1 = player.streamCardsIn(ZoneType.Battlefield);
Stream<Card> s2 = player.streamCardsIn(ZoneType.Battlefield);
s1.forEach(c -> game.moveCard(c, ZoneType.Graveyard));
s2.toList(); // may throw or return inconsistent snapshot
```
Impact: Non-deterministic behavior; streams should be documented as point-in-time snapshots.

#### 6. Game.streamCardsIn for Stack does not filter phased out
Game.streamCardsIn returns getStackZone().streamCards() for Stack zone without owner filter and without considering phased out. Stack cards are never phased out in MTG rules, so acceptable, but inconsistent with Player version which filters by owner.

Test case:
```java
Stream<Card> gameStack = game.streamCardsIn(List.of(ZoneType.Stack));
gameStack.count(); // includes all players' stack cards
```
Impact: Correct for game view but documentation unclear.

#### 7. PlayerZoneBattlefield.streamCards double streams
PlayerZoneBattlefield.streamCards(boolean filter) calls super.streamCards(filter) which returns cardList.stream(), then applies filter if needed. This creates an intermediate stream. Fine, but super implementation ignores filter, so filtering always happens in subclass.

Test case:
```java
PlayerZoneBattlefield bf = new PlayerZoneBattlefield(...);
bf.add(phasedCard);
bf.streamCards(true).count(); // works via override
```
Impact: Minor inefficiency; acceptable.

### Performance and security
#### 8. FlatMap over players and zones creates O(N*M) streams
Game.streamCardsIn flattens zones via StreamUtil.stream(zones).flatMap(z -> ... players.stream().flatMap(...)). For many players and zones this creates nested streams with repeated player iteration.

Test case:
```java
List<ZoneType> zones = Arrays.asList(ZoneType.Hand, ZoneType.Library, ZoneType.Graveyard);
game.streamCardsIn(zones).count(); // iterates players three times
```
Impact: Performance degradation in games with many players/zones; consider caching or pre-filtering.

## Verdict
Request Changes
The stream card zones implementation introduces useful API but contains semantic mismatches for Flashback handling and ignores filter parameter in base Zone.streamCards. Phased out cards may leak from non-Battlefield zones, and Flashback special case returns activatable cards instead of zone contents. These correctness issues need fixing before merge.

## Recommended Fixes
1. Remove Flashback special case from Player.streamCardsIn or document clearly that it returns activatable cards, not zone contents. Prefer returning empty stream for non-existent zones.
2. Make Zone.streamCards(filter) respect filter parameter, at least filtering phased out cards in base implementation or provide default filter.
3. Ensure Card owner equality uses identity or player ID comparison for stack filtering.
4. Document that returned streams are lazy views and should be consumed immediately or collected to list for stability.
5. Consider optimizing Game.streamCardsIn to avoid repeated player iteration per zone.

---

PR #11337, opened 2026-07-23
Fix AI implementation of Toxic Deluge and Flowstone Slide

liamiak:
Both cards carry `AI:RemoveDeck:All` — the AI is not allowed to play them at all.

## What a reviewer can check

`PumpAllAiTest` has five assertions and all five fail without this change. Two go through
`canPlaySa` for the X the AI chooses. Three play a real turn with `playUntilNextTurn` and check
the board it left behind: that the sweep resolved, that Toxic Deluge paid only the 2 life the
kill needed, and that the X it picked spared its own 3/3s while killing the opponent's 2/2s.

## Why the flag removal cannot ship on its own

With only the script changes on current master, the AI casts both spells for X=0 and puts them
in the graveyard having killed nothing:

```
Flowstone Slide  endedIn=Graveyard  bearsSurviving=3  aiLife=20
Toxic Deluge     endedIn=Graveyard  bearsSurviving=3  aiLife=20
```

`AI:RemoveDeck:All` was load-bearing — it was protecting the AI from a card it would waste. The
flag comes off only because the X selection goes in.

## The cause

Their X is announced as a cost and nothing ever chose it: `PumpAll` has an AI class, so
`AiController`'s generic X handling is skipped, and Toxic Deluge's X lives in a `PayLife<X>`
cost that path would not have seen anyway. Toxic Deluge was also missing `IsCurse$ True`, so
`PumpAllAi` never reached its sweeper branch and instead asked whether to pump the AI's own
creatures.

## The fix

`PumpAllAi` picks X for a -X/-X sweep when the ability announces its own X, scoring by what each
value destroys. Candidates are the toughnesses actually on the battlefield rather than every
point up to the maximum, so the AI does not walk a 40 point life total one step at a time.

Rather than reimplementing combat reasoning, the two questions it needs go to the code that
already answers them. Whether the swing simply wins goes to the attack planner, which knows
about fog, evasion and `cantWin`. Whether the AI lives through the counterattack goes to
`predictNextCombatsRemainingLife`, which models the blocks it will make — asked only when the
sweep is not lethal and a cheap pessimistic bound says the survivors could get there.

Each needed one small addition, in its own commit: a counterpart to `removeBlocker` for our own
creatures, and a way to tell combat prediction which attackers a wipe removes.

### Multiplayer: which boards a sweep is weighed against

`checkApiLogic` weighs a sweep against `getStrongestOpponent()`. For choosing X that undercounts:
a mass -X/-X lands on every opponent's board at once, so the X worth paying for is the one
measured against all of them. The X selection here takes the full list.

The pre-existing card-value comparison keeps the narrower list deliberately. Its margin

```java
(evaluateCreatureList(comp) + 200) < evaluateCreatureList(human)
```

is calibrated against a single opponent and does not rescale to a sum. With `human` summed over
the table the test becomes "am I behind the combined board", which the leader usually is.
Measured in a three-player game with the wider list feeding it: an AI holding the single best
board, four Hill Giants against three and three, sweeps and destroys its own four. With the
narrower list it declines and develops.

So the split is on purpose. The margin looks worth revisiting on its own terms, but that is not
this PR's to change.

Best read commit by commit — a shared predicate, the two enabling additions, the fix, then
tests. 363 tests, 0 checkstyle violations.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -187,6 +187,15 @@ public class AiAttackController {
         this.blockers.remove(blocker);
     }
 
+    /**
+     * Counterpart to removeBlocker, for asking what an attack looks like without one of our own
+     * creatures - one that is about to die to something we are considering casting, say.
+     */
+    public void removeAttacker(Card attacker) {
+        this.myList.remove(attacker);
+        this.attackers.remove(attacker);
+    }
+
     private boolean canAttackWrapper(final Card attacker, final GameEntity defender) {
         if (nextTurn) {
             return CombatUtil.canAttackNextTurn(attacker, defender);
--- a/forge-ai/src/main/java/forge/ai/ComputerUtil.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtil.java
@@ -3179,6 +3179,13 @@ public class ComputerUtil {
         return predictNextCombatsRemainingLife(ai, serious, checkDiff, payment, excludedBlockers, ai.getOpponents());
     }
     public static int predictNextCombatsRemainingLife(Player ai, boolean serious, boolean checkDiff, int payment, final CardCollection excludedBlockers, final List<Player> opps) {
+        return predictNextCombatsRemainingLife(ai, serious, checkDiff, payment, excludedBlockers, opps, null);
+    }
+    /**
+     * @param excludedAttackers creatures to leave out of the opponents' attack - ones we are about
+     *            to kill, say, so that a board wipe is not weighed against the board it removes.
+     */
+    public static int predictNextCombatsRemainingLife(Player ai, boolean serious, boolean checkDiff, int payment, final CardCollection excludedBlockers, final List<Player> opps, final CardCollection excludedAttackers) {
         // life won't change
         int remainingLife = Integer.MAX_VALUE;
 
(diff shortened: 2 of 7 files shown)


liamiak:
Force-pushed on 23 Aug with no note, so: the AI code is unchanged from what was here in August. The push was a rebase onto current master plus three tests.

The two existing ones go through `canPlaySa` and assert the X the AI chooses. The new three play a real turn with `playUntilNextTurn` and check the board left behind — that the sweep resolved, that Toxic Deluge paid only the 2 life the kill needed, and that the X it picked spared its own 3/3s while killing the opponent's 2/2s.

That last one is the assertion I would most want a second opinion on: it is the only one that pins the *choice* of X rather than that a sweep happened at all.

The four AI files and both card scripts are byte-identical across the rebase.

tool4ever:
don't really have time for a full review right now
but the logic looks pretty expensive

liamiak:
Fair — measured it, and you were right about where to look.

Instrumented in one JVM, warm: the gate cost about 6ms per evaluation on a contested board, and about 4ms of that was a single `ComputerUtilCost.setMaxXValue` call. The combat predictions I had bounded with `SURVIVAL_CHECKS` never ran at all, so the part I capped was not the cost and the part I treated as setup was.

Worse, that call ran first, before anything established the sweep could do something — so on a board with nothing worth killing it was paid for nothing, every turn the card sat in hand.

Pushed a fix: the candidate scan is now bounded arithmetically, and the solver call waits until the scoring has found an X worth paying for. It only ever narrows the candidates, so which X gets chosen is unchanged. Contested boards measure the same as before; on an empty opposing board the gate goes from ~4ms to ~0.02ms and makes no solver call at all.

    contested board       6.27 / 5.87 ms  ->  6.21 / 5.39 ms
    empty opposing board  4.17 / 3.36 ms  ->  0.030 / 0.022 ms

PR #11339, opened 2026-07-23
Fix auto-tap highlight never showing

autumnmyst:
The auto-tap highlight never appeared because `getManaSourcesToPayCost` read its cards from a `List<Mana>` that a `test=true` payment never populates (test mode only predicts the mana produced and records the chosen abilities) so it always returned an empty set and no card reached the strength-2 threshold the highlight requires. This threads the chosen mana abilities out of `payManaCost` so their host cards are reported, and draws the glow after the actionable highlight so it stays visible when black borders are disabled, where that highlight is an outset that otherwise covers it.

EDIT: now after rebase this PR only changes the visuals for more clarity by making the boarder highlight the same yellow as the glow for visual clarity.

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
+++ b/forge-gui-desktop/src/main/java/forge/view/arcane/CardPanel.java
@@ -322,13 +322,16 @@ public class CardPanel extends SkinnedPanel implements CardContainer, IDisposabl
         final int cornerSize = noBorderPref && !cardImgHasAlpha ? 0 : Math.max(4, Math.round(cardWidth * CardPanel.ROUNDED_CORNER_SIZE));
         final int offset = isTapped() && (!noBorderPref || cardImgHasAlpha) ? 1 : 0;
 
-        // Yellow glow for cards that Auto would tap to pay (weak-selectable strength >= 2)
-        if (isPreferenceEnabled(FPref.UI_SHOW_AUTOTAP_PREVIEW) && matchUI.getWeakSelectableStrength(getCard()) >= 2) {
-            for (int layer = 2; layer >= 1; layer--) {
-                g2d.setColor(new Color(1f, 1f, 0f, 0.14f * layer));
-                final int n = Math.max(1, Math.round(layer * cardWidth * CardPanel.SELECTED_BORDER_SIZE));
-                g2d.fillRoundRect(cardXOffset - n, (cardYOffset - n) + offset, cardWidth + (n * 2), cardHeight + (n * 2), cornerSize + n, cornerSize + n);
-            }
+        // Yellow glow for cards that Auto would tap to pay (weak-selectable strength >= 2).
+        // Drawn after the actionable highlight (see below) so it shows over it, except when a
+        // chosen/hover/flash/zone frame outranks it - then it goes underneath, here.
+        final boolean autoTapGlow = isPreferenceEnabled(FPref.UI_SHOW_AUTOTAP_PREVIEW)
+                && matchUI.getWeakSelectableStrength(getCard()) >= 2;
+        final boolean priorityFrame = matchUI.isHighlighted(getCard()) || isSelected
+                || (hasFlash && getCard() != null && matchUI.mayView(getCard()))
+                || zoneBannerColor != null;
+        if (autoTapGlow && priorityFrame) {
+            drawAutoTapGlow(g2d, cornerSize, offset);
         }
 
         // Magenta outline for when card is chosen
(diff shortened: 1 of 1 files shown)


forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
```diff
@@ -812,6 +834,10 @@ private static List<Mana> payManaCost(final ManaCostBeingPaid cost, final SpellA
             resetPayment(paymentList);
         }
 
+        if (paymentListOut != null) {
+            paymentListOut.addAll(paymentList);
```
tool4ever: thanks for the catch, can actually just switch the return value out which I've done now

if you rebase this class should be able to be dropped from the diff

then we can check the GUI part

jamincollins:
## Findings
#### 1. Auto-tap glow drawn under priority frames may obscure actionable highlight
The new priorityFrame condition includes isHighlighted, isSelected, hasFlash with mayView, and zoneBannerColor. When priorityFrame is true, drawAutoTapGlow is called before the magenta/green/cyan/zone frames (line 329-330), placing the yellow glow underneath those frames. This matches intent ("underneath" for higher priority). However, the glow's solid yellow border (Color.YELLOW at line 427) is drawn with the same border size as the actionable highlight frames (SELECTED_BORDER_SIZE). The actionable highlight frames are drawn after the black fill and before the inner white border (lines 335-354). The new glow when priorityFrame is false is drawn after the inner border drawing (line 415-417), which is after the actionable highlight frames have been drawn. This order ensures glow over actionable highlight, but the glow's outer fade (alpha 0.28) may bleed into the card interior due to fillRoundRect covering the whole card plus border, potentially altering card art perception. No MTG rule violation, but visual correctness depends on alpha blending order.

Test case:
```java
// Simulate paintComponent ordering
boolean autoTapGlow = true;
boolean priorityFrame = false;
if (autoTapGlow && priorityFrame) drawAutoTapGlow(); // before frames
// frames drawn...
if (autoTapGlow && !priorityFrame) drawAutoTapGlow(); // after frames
// Verify glow is on top of highlight when priorityFrame false
```
Impact: Visual stacking may be inconsistent if priorityFrame logic changes.

#### 2. Null card with hasFlash true causes NPE in priorityFrame
priorityFrame = matchUI.isHighlighted(getCard()) || isSelected || (hasFlash && getCard() != null && matchUI.mayView(getCard())) || zoneBannerColor != null;

The hasFlash branch checks getCard() != null before mayView, good. However, matchUI.isHighlighted(getCard()) at line 327 is called without null guard. If getCard() returns null (empty panel), isHighlighted may throw NPE inside implementation. CardPanel.getCard() can be null during initialization or after disposal.

Test case:
```java
CardPanel panel = new CardPanel();
panel.setCard(null);
panel.paintComponent(g); // should not NPE on isHighlighted
```
Impact: Potential NPE on empty panels.

#### 3. cardWidth zero causes division by zero or negative rects
drawAutoTapGlow uses Math.round(cardWidth * SELECTED_BORDER_SIZE). If cardWidth is zero (uninitialized layout), n = Math.max(1, ...) ensures minimum 1, but outer radius calculation uses same cardWidth. fillRoundRect with negative width/height may occur if cardWidth is zero and offset calculations produce negative extents. Also cornerSize + outer may overflow int for very large cards.

Test case:
```java
int cardWidth = 0;
int outer = Math.max(1, Math.round(2 * cardWidth * 0.05f)); // 1
g2d.fillRoundRect(-1, 0, 2, 2, ...); // draws tiny rect
```
Impact: Minor rendering glitch, not crash.

#### 4. Glow alpha increased from 0.14*layer to 0.28 outer fade
Old code drew two layers with alpha 0.28 and 0.14 (layer 2 and 1). New code draws one outer fade with alpha 0.28 and solid yellow border. Visual appearance changed: previous glow was translucent double-layer; new glow is more opaque solid border with fade. Users with UI_SHOW_AUTOTAP_PREVIEW enabled may perceive stronger highlight, potentially masking card borders. Backwards compatibility of visual style broken.

Test case:
```java
// Compare old vs new alpha
Color oldLayer2 = new Color(1f,1f,0f,0.28f);
Color oldLayer1 = new Color(1f,1f,0f,0.14f);
Color newOuter = new Color(1f,1f,0f,0.28f);
Color newBorder = Color.YELLOW; // alpha 1.0
```
Impact: UI regression, user preference expectation changed.

#### 5. drawAutoTapGlow called twice if priorityFrame changes mid-paint
autoTapGlow and priorityFrame are computed once at line 324-328, then used twice. If matchUI.isHighlighted changes during paint (unlikely), state consistent. More concerning: drawAutoTapGlow uses cardWidth from outer scope captured at paint start; if card resizes mid-paint, dimensions stale. No side effect.

Test case:
```java
// Simulate repaint with changing selection
boolean initial = matchUI.getWeakSelectableStrength(card) >= 2;
// selection changes after autoTapGlow computed but before draw
```
Impact: Rare flicker.

### Performance and security
#### 6. No performance impact, but repeated Math.round per paint
drawAutoTapGlow computes outer and n each paint. Math.round and Math.max are cheap. No security issue.

## Verdict
Comment
The change improves visibility of auto-tap glow over actionable highlight, but introduces null safety risk for isHighlighted(getCard()) and alters visual intensity of glow. Recommend adding null guard and preserving previous alpha levels.

## Recommended Fixes
1. Guard matchUI.isHighlighted(getCard()) with getCard() != null.
2. Consider making glow alpha configurable to preserve prior subtlety.
3. Add unit test for CardPanel paint order with auto-tap + highlight combos.

---

PR #11346, opened 2026-07-23
Add Adventure settings option to start directly in Adventure mode

### LINKED ISSUE #11344: Options Setting to Prefer Adventure Mode On App Start
I primarily use Forge in Adventure Mode and sometimes play with a controller. Because the Adventure Mode menu supports Controller and it doesn't seem like the other mode does, it creates an annoying step in between where I have to navigate to Adventure Mode first using the mouse.

It would be nice to have an optional setting in Adventure Mode that could be checked to allow Adventure Mode to be the default mode on app startup, since you can still easily shift back to the other mode from that menu. 

I considered adding controller support to the primary menu, but it doesn't seem like that makes a lot of sense. Allowing the user to optionally start in the mode that they prefer from the adventure mode options menu felt like the optimal approach here. 

**Additional context**
n/a

liminalwarmth:
Resolves #11344: adds an option inside Adventure Mode's settings to start Forge directly in Adventure Mode.

### Why

Adventure Mode fully supports controller navigation, but the mode-selector splash screen shown at startup does not. Players who primarily play Adventure with a controller must reach for the mouse on every launch just to get into the mode they always play. Forge already has a preference for this (`UI_SELECTOR_MODE`, exposed as "Selector Mode" in the classic mobile settings page), but it is not reachable from within Adventure Mode itself.

### What

Adds a **"Start in Adventure mode (applies on next launch)"** checkbox to Adventure's Settings scene, right below the World selector:

- Checked: sets `UI_SELECTOR_MODE` to `Adventure`, so the next launch boots straight into the Adventure main menu (fully controller-navigable).
- Unchecked: restores `Default`, bringing back the mode-selector splash screen — so switching back to other modes stays easy, per the issue discussion.
- The checkbox re-reads the preference on scene enter, so it stays in sync if the mode is changed from the classic Selector Mode setting.
- `Forge.selector` is updated immediately, matching how the classic `SettingsPage` applies the same preference.

No new preference is introduced — this only exposes the existing `UI_SELECTOR_MODE` preference where controller players need it.

### Changes

- `forge-gui-mobile/src/forge/adventure/scene/SettingsScene.java`: new checkbox wired to `UI_SELECTOR_MODE`; the boolean `addSettingField` overload now returns its `CheckBox` so the setting can be refreshed on enter.
- `forge-gui/res/languages/en-US.properties`: new `lblStartInAdventureMode` string.

Primary feature was written with Codex 5.6 Sol with guidance to scope it as small as possible, code reviewed in several rounds by Sol and Claude Fable, and manually tested (by me on MacOS) before opening the PR.

### Testing

- `mvn -pl forge-gui-mobile -am compile` passes on this branch.
- Feature validated in play on a macOS build carrying this change: direct boot into Adventure with the box checked, selector restored when unchecked, checkbox state stays in sync with the classic Selector Mode setting.

Screenshot 2026-07-23 at 11 30 10 AM

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/scene/SettingsScene.java
+++ b/forge-gui-mobile/src/forge/adventure/scene/SettingsScene.java
@@ -35,6 +35,7 @@ public class SettingsScene extends UIScene {
     TextraButton backButton;
     TextraButton newPlane;
     ScrollPane scrollPane;
+    CheckBox startInAdventure;
 
     SelectBox<String> selectSourcePlane;
     TextField newPlaneName;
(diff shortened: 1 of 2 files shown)


liminalwarmth:
If this is considered unnecessary because the option technically already existed in the app's top-level options menu, no sweat. I just didn't even realize that it was there until after I had added this feature, because I never dug into that part of the app or realized that that option was available since it didn't exist on the settings I had been looking at.

liminalwarmth:
I think I actually would prefer to handle this by mirroring the selector mode dropdown from the standard settings menu rather than adding a new checkbox. Closing this for now, and I'm going to come back with a different approach.

PR #11347, opened 2026-07-23
Add macOS borderless fullscreen support for Adventure

### LINKED ISSUE #11342: Adventure Mode Borderless Full Screen Support for 2nd Monitor
I'm currently playing on Mac OS and have been frustrated by the behavior in full screen mode that the game client minimizes to the task bar any time I click out to my second monitor. 

It would be nice for Adventure Mode to introduce a borderless full-screen option below the standard full-screen option for modern dual-monitor setups, where you can play in full screen on one monitor and have a browser or other program up on the second monitor and move fluidly between them. Ideally, the audio would also continue while the Forge app loses focus. 

There are not really alternatives to this, since every alternative would functionally be the same as a full-screen borderless mode, but I suppose the full-screen implementation could be modified to work like a borderless mode as a possibility. 

**Additional context**
I have a fix for this on my local fork (which required testing a few different approaches to get exactly the right functionality where command-tabbing and screen priority work correctly) but it's for Mac OS only, as I haven't tested it on Linux or Windows (they are omitted by design and simply don't include the new full screen borderless option). I plan to PR just this solution scoped to MacOS but I'm unsure what the policy is for single-platform settings.

liminalwarmth:
Resolves #11342

Implements the feature request in [Card-Forge/forge#11342](https://github.com/Card-Forge/forge/issues/11342): a borderless fullscreen option for Adventure that remains full-monitor while allowing normal use of other applications.

### Why

The existing GLFW fullscreen mode automatically iconifies when it loses focus, so using an application on a second monitor minimizes Adventure. Disabling auto-iconify on that exclusive-fullscreen window prevents the minimization, but leaves the game above other applications and does not provide normal macOS minimize behavior.

### What

Adds a **Borderless Fullscreen** option (shown on macOS only until validated elsewhere) beside the existing Fullscreen setting:

- Adventure opens as a monitor-sized, undecorated normal window instead of an exclusive-fullscreen window.
- Cmd-Tab and applications on a second monitor work without minimizing or resizing Adventure.
- The macOS menu bar and Dock stay hidden while the borderless window is visible.
- Cmd-M minimizes normally, and restoring the window preserves its full-monitor geometry.

Per review feedback, the change to keep music playing while unfocused has been removed from this PR; it will be proposed separately as a preference handled in SoundSystem.

The window setup itself is platform-neutral; the option and launcher behavior are gated with `OperatingSystem.isMac()` because macOS is the only platform this has been validated on so far. The setting is hidden on other platforms, a stale preference value is ignored there, and the existing fullscreen path remains unchanged for Windows, Linux, and Android.

A small, lazily initialized AppKit bridge applies the presentation options and native minimize action without adding a dependency or loading macOS symbols on other platforms.

### Changes

- `forge-gui-mobile-dev/src/forge/app/GameLauncher.java`: normal full-monitor window setup, macOS presentation lifecycle, and minimize action.
- `forge-gui-mobile/src/forge/adventure/scene/SettingsScene.java`: macOS-gated Adventure setting, mutually exclusive with Fullscreen.
- `forge-gui-mobile/src/forge/screens/settings/SettingsPage.java`: matching macOS-gated option in the classic mobile settings page.
- `SettingData`, `ForgePreferences`, and `en-US.properties`: persist and label the new option.

Primary feature was written with Codex 5.6 Sol with guidance to scope it as small as possible, code reviewed in several rounds by Sol and Claude Fable, and manually tested (by me on MacOS) before opening the PR.

### Testing

- `mvn -pl forge-gui-mobile-dev -am test -DskipTests=false` passes (3 tests).
- `mvn -q -pl forge-gui-mobile-dev -am package -DskipTests` passes.
- Checkstyle reports 0 violations.
- Validated in play on macOS: menu bar and Dock hiding, Cmd-Tab, second-monitor application use, Cmd-M minimize and restore, stable full-monitor geometry, and mouse release after focus changes.


forge-gui-mobile-dev/src/forge/app/GameLauncher.java
```diff
@@ -138,7 +169,7 @@ public void focusGained() {
             @Override
             public void focusLost() {
                 super.focusLost();
-                SoundSystem.instance.setWindowFocus(false);
+                SoundSystem.instance.setWindowFocus(macOSBorderlessFullScreen);
```
Jetz72: If you want to disable the behavior of stopping music when focus is lost, that should probably be a separate PR, and be gated behind a separate preference rather than tied to a window mode. It also probably should be handled in the SoundSystem and not on the window listener.

forge-gui-mobile-dev/src/forge/app/GameLauncher.java
```diff
@@ -147,6 +178,74 @@ public void focusLost() {
...
+            }
+        }
+
+        private static void setPresentationOptions(long options) {
+            long application = JNI.invokePPJ(NS_APPLICATION, SHARED_APPLICATION, OBJC_MSG_SEND);
+            if (options != DEFAULT) {
+                ensureMinimizeMenu(application);
+            }
+            JNI.invokePPPV(application, SET_PRESENTATION_OPTIONS, options, OBJC_MSG_SEND);
+        }
```
Jetz72, line 238: This class of platform-specific constants and native interface invocations makes me think that whatever limitation you're aiming to overcome here would better be resolved in LWJGL or GLFW.

Jetz72 (commented):
Feels like the AI took a few nuances of circumstance and implementation that came up in the feature request and made them into fundamental parts of the design. As though the MacOS exclusivity was a core part of the feature and not a simple gating of something that hasn't been verified on other platforms.

forge-gui-mobile-dev/src/forge/app/GameLauncher.java
liminalwarmth: I was originally using a GLFW approach, but on Mac OS I was running into issues sliding other windows over it for some reason and having it properly behave like a normal FS borderless window. GLFW was the Windows and Linux approach when I had those in, but I pulled them out for this cut since I couldn't test on those platforms. I'll take a look at LWJGL.

liminalwarmth:
> As though the MacOS exclusivity was a core part of the feature and not a simple gating of something that hasn't been verified on other platforms.

That's probably Sol being extremely literal when I told it to focus on Mac OS because I wasn't able to test on Linux and Windows. Let me know if there are other places you're specifically seeing that and I can take another look.

Edit: To clarify, that was actually me specifically requesting MacOS only to try to keep the PR smaller. I could add back in the other OS if desired? But can't verify them easily.

forge-gui-mobile-dev/src/forge/app/GameLauncher.java
liminalwarmth: > ObjCRuntime, JNI, LibSystem are all LWJGL's own bundled macOS bindings — no new dependency was added. What GLFW lacks is any API for app-level NSApplication.presentationOptions (hide/auto-hide menu bar and Dock for a normal-level window). GLFW's ownborderless path on macOS puts the window at a level above the menu bar, which is exactly what you found broken: with auto-iconify off it floats above everything and you can't use a second monitor. Your history confirms GLFW's built-in behavior is the limitation, not a workaround you skipped.

liminalwarmth: One of my design goals here was to support something like alt-tabbing to a web browser on a single screen behind the game window or using keyboard commands to minimize the app or show the desktop while still allowing it to keep focus, and not do things like pop the dock in the normal course of gameplay while the mouse is at the bottom of the screen. I had to ask it to change the approach we tried initially which was failing at those things. It says these are LWJGL's bundled AppKit bindings.

forge-gui-mobile-dev/src/forge/app/GameLauncher.java
liminalwarmth: Extracted; I'll open a separate PR for that as requested.

liminalwarmth:
Okay; extracted the sound piece for a separate PR, made it less MacOS-oriented in how things are named and presented (including the settings option), and explained the LWJGL binding approach.

If you wanted this expanded to both Windows and Linux so it covers the multi-platform desktop cases, that's not too much trouble. I just can't test them directly. Happy to make the adjustment or let someone else try that!

forge-gui-mobile-dev/src/forge/app/GameLauncher.java
Jetz72: I get the vision, I'm just surprised none of the underlying libraries support this properly. Looking at the division of responsibilities though, maybe libGDX would be the more appropriate home for a solution like this? Unless we're utilizing its window/application configurations incorrectly...

liminalwarmth: Yeah, agreed. Unfortunately, I'm not personally familiar with any of the idiosyncrasies of these libraries, so I can only tell you the testing angles I came at it from, and I can tell you what Fable and Sol tell me when I relay that technical feedback and do a little digging. Here are the tight highlights from some research about this (edits mine, comments from Fable):

> "Agreed on the division of responsibilities [Jetz suggested]; the blocker is that libGDX delegates windowing to GLFW, which has no presentation-options API (glfw#2246), and libGDX would need these same ObjC calls (see libgdx#6646/#7546)."
> - **You're (most likely) not misusing libGDX.** libGDX's tracker confirms this is a gap: #6646 — borderless fullscreen support request (https://github.com/libgdx/libgdx/issues/6646) and #7546 — macOS Dock stays interactable over fullscreen (https://github.com/libgdx/libgdx/issues/7546) are this same class of problem.
> - **The issue is a layer below libGDX, because libGDX delegates all desktop windowing to GLFW (via LWJGL).** So "the underlying libraries should support this" really points at GLFW, and there's a defensible reason they don't: presentationOptions is application-global state, not per-window state, and GLFW historically treats app-level policy as the application's business.

I asked Fable to check if there was an alternative approach that would still achieve the design goals on MacOS (no taskbar popup, minimizable window, etc) but work in a way more consistent with the directions you've proposed and got this:

> "Short answer: no — for the full set of behaviors you want, NSApplication.presentationOptions is the macOS mechanism, and every path to it looks like the code you already have. The alternatives each sacrifice one of your stated goals."

That sounds reasonable to me based on what I saw when testing, but I'm not familiar enough with this problem space or these libraries to be able to personally validate that or suggest a different pattern. I think on Linux and Windows GLFW might handle this better since Fable had paths for both in the original code that kept that approach? So if you wanted to support it for all three, I think it's just macOS that would be the problem and need special handling. The only other alternatives that were feasible would be making corrections or suggesting additions to the libraries themselves, which feels like probably the correct long-term solution but is unlikely to help Forge support this anytime soon. 

However, no worries if you want to reject this PR because the pattern feels wrong for the repo standards! I have it fixed in my local build, which solves my immediate problem. I'll paste the full explanation I got from Fable below as a starting point in case anyone wants to dig deeper on this.

Thanks for taking the time to review it and discuss regardless of your decision!

<details>

<summary><i>Fable's full response (click to expand)</i></summary>

>   Short answer: no — for the full set of behaviors you want, 
>   NSApplication.presentationOptions is the macOS mechanism, and every path to it
>   looks like the code you already have. The alternatives each sacrifice one of
>   your stated goals. Here's the rundown:
> 
>   What you want is a specific combination: a screen-covering window at normal
>   window level (so Cmd-Tab layering, second monitor, and show-desktop all work),
>   with the menu bar and Dock auto-hidden so they don't pop during gameplay.
>   Going through the alternatives:                                          
> 
>   - macOS native fullscreen (toggleFullScreen, the green-button mode): hides
>   menu bar and Dock properly, but moves the game to its own Space. Cmd-Tabbing
>   to a browser swipes you to a different Space instead of layering windows on
>   one screen — fails your core goal.
>   - GLFW's own borderless-fullscreen path: puts the window at a level above the
>   menu bar. That's the broken behavior you already hit — floats over everything,
>   second monitor unusable.
>   - Accessory activation policy (the proposal in glfw#2246
>   (https://github.com/glfw/glfw/issues/2246)): hides the menu bar and Dock icon
>   by making the app an "accessory" — but accessory apps are excluded from
>   Cmd-Tab entirely, which breaks app switching. And it's still just an open
>   feature request, not shipped API.
>   - Window sized below the menu bar (no presentation options at all): menu bar
>   stays visible and the Dock still pops when your mouse hits the screen bottom —
>   fails the "don't interrupt gameplay" requirement.
>   - Display capture / kiosk modes: strictly more exclusive than what you have;
>   fails everything about coexisting with other windows.
> 
>   So the design space isn't "which mechanism" — it's settled — it's which layer
>   makes the ObjC call. Three options, and this maps directly onto Jetz72's
>   concern:
> 
>   1. Keep it in Forge (current PR): the pragmatic choice. It's already isolated
>   in one MacOSPresentation class using LWJGL's bundled bindings, documented as
>   removable.
>   2. Contribute it to libGDX's Lwjgl3 backend: the right long-term home. libGDX
>   shipped generic borderless fullscreen in 1.11 (#6646
>   (https://github.com/libgdx/libgdx/issues/6646) is closed), but its macOS
>   behavior still has exactly these gaps (#7546
>   (https://github.com/libgdx/libgdx/issues/7546) — Dock stays interactable). A
>   small config option or macOS helper there would likely be welcomed, and
>   Forge's shim gets deleted when it lands. The catch is release latency — you'd
>   keep the shim until a libGDX release with the fix reaches Forge anyway.
>   3. Contribute it to GLFW: architecturally purest, but presentation options are
>   app-global state that GLFW has historically treated as out of scope, and the
>   pipeline (GLFW → LWJGL → libGDX → Forge) is years long.
>   
>   Sources: glfw#2246 — Hide taskbar/Dock icon and menu bar        
>   (https://github.com/glfw/glfw/issues/2246), libgdx#6646 — borderless
>   fullscreen support (https://github.com/libgdx/libgdx/issues/6646), libgdx#7546
>   — macOS Dock over fullscreen (https://github.com/libgdx/libgdx/issues/7546),
>   GLFW window guide (https://www.glfw.org/docs/3.3/window_guide.html),
>   NSApplication.PresentationOptions (https://developer.apple.com/documentation/a
>   ppkit/nsapplication/presentationoptions)

</details>

liminalwarmth: If you'd be okay with this approach but would prefer for it to be solved as a universal option across operating systems (with a temporary stopgap solution for just MacOS), I'm happy to add the other two paths back in. Just let me know. As I said earlier, I just can't manually test them and would need help there.

liminalwarmth:
Rebased onto current master to resolve the conflict in `GameLauncher.java` — the focus handlers now use the `Forge.setWindowFocus` calls introduced on master, so this PR no longer touches the focus/audio lines at all. No functional changes to the borderless fullscreen feature itself.

liminalwarmth:
Reworked the settings UI for this feature: instead of two mutually-exclusive checkboxes (Fullscreen + Borderless Fullscreen) that had to uncheck each other, there's now a single **Display Mode** dropdown — Windowed / Fullscreen / Borderless Fullscreen (the borderless option still offered on macOS only). It's a cleaner UI pattern since the three states are genuinely exclusive, and it removes the cross-checkbox syncing logic in both the Adventure settings scene and the mobile settings page.

The underlying prefs are unchanged (`UI_FULLSCREEN_MODE` + `UI_BORDERLESS_FULLSCREEN_MODE`), so nothing else in the PR is affected. `CustomSelectSetting` gains a small overridable `getSelectedValue()` hook so a select can derive its displayed value from more than one pref.

forge-gui-mobile/src/forge/adventure/scene/SettingsScene.java
```diff
@@ -31,6 +33,10 @@
  * Scene to handle settings of the base forge and adventure mode
  */
 public class SettingsScene extends UIScene {
+    private static final String DISPLAY_MODE_WINDOWED = "Windowed";
+    private static final String DISPLAY_MODE_FULLSCREEN = "Fullscreen";
+    private static final String DISPLAY_MODE_BORDERLESS = "Borderless Fullscreen";
```
Jetz72, line 38: The options text should probably be retrieved from the localized strings.

forge-gui-mobile-dev/src/forge/app/GameLauncher.java
Jetz72: Ideal resolution here would be Fable's second suggestion - contributing a fix for borderless window behavior to the LibGDX repo. I don't know their policy on LLM contributions, though. You might start by writing up an issue and seeing what they have to say about it?

I'd want a second opinion on whether it's worth merging in a fairly bulky workaround on our end pending an outcome there. @tehdiplomat, @tool4ever, thoughts on including this? - (full block: https://github.com/Card-Forge/forge/pull/11347/changes#diff-9a4cb10ab4c9403e05af8432e9c9eebd5a125ac3f2d8d8c4d230580cf6b7c131R182-R263 )

If we can get it to the point where borderless is properly handled by LibGDX, I'd be more comfortable proceeding without the platform gating.

tool4ever: hmn it's nice AI ended up finding a way but such low level with needing all these function pointers?
I worry about which macOS range even matches all these signatures and if the API ever changes we have no way to see the error in our IDE... :/

liminalwarmth: The reason I didn't suggest going that direction is that my understanding was that you would actually need to also get changes in GLFW and Fable thought that that might be a potentially very long process (even assuming that a rando LLM PR would be actioned/welcome there absent a larger contextual push--I feel much less confident about pointing my agents at that library for changes since I can't manually test the results of the upstream changes or second-order impacts). Since I don't really work with those libraries, I didn't feel comfortable requesting specific changes to them personally because I'm only poking more directly at the code in this repo right now.

Again, this is a fairly niche (personal) complaint that seems to only affect playing in desktop mode on MacOS _if_ you want to do other things while playing without having the game screen minimized or care about the bottom bar popping over it. I think it's probably also just fine to close this PR for now and note it as an issue if the current approach feels too janky, which would be totally understandable.

forge-gui-mobile/src/forge/adventure/scene/SettingsScene.java
liminalwarmth: Makes sense. Happy to fix this if you guys decide you want to move forward with this approach per the conversation above.

jamincollins:
## Findings

#### 1. MacOSPresentation.setBorderlessVisible may be invoked on non-main thread without dispatch guard
The `setBorderlessVisible` method checks `JNI.invokeI(PTHREAD_MAIN)` and falls back to `dispatch_sync_f` on the main queue. However `iconified(boolean)` and `created(Lwjgl3Window)` are LWJGL callbacks which may be invoked from the render thread, not the main AppKit thread. Dispatching via `dispatch_sync_f` from the render thread will block until the main thread processes, potentially causing deadlock if the main thread is waiting on the render thread. Additionally the `MacOSPresentation` inner class uses `EnumerationMutationHandler` as a callback type purely for signature compatibility; the handler is never cleaned up, leading to potential native resource leak on repeated toggles.

Test case:
```java
// Simulate rapid iconify events from render thread while main thread holds lock
Thread render = new Thread(() -> {
    for (int i = 0; i < 100; i++) {
        MacOSPresentation.setBorderlessVisible(i % 2 == 0);
    }
});
render.start();
render.join();
// Verify no deadlock and NSApplication presentation options remain consistent
```
Impact: Potential UI freeze on macOS when toggling borderless mode.

#### 2. Display mode dropdown does not preserve fullscreen state when migrating from old settings
`SettingsScene.currentDisplayMode()` returns `DISPLAY_MODE_BORDERLESS` if `OperatingSystem.isMac() && settings.borderlessFullScreen`, otherwise falls back to `settings.fullScreen`. However `SettingsPage.getSelectedValue()` for the custom select setting checks preference `UI_BORDERLESS_FULLSCREEN_MODE` first, then falls back to `UI_FULLSCREEN_MODE`. On first launch after upgrade, `SettingData.borderlessFullScreen` defaults to `false` (primitive boolean) while preference may have been set by migration code elsewhere. The two sources of truth diverge: `SettingData` is persisted via JSON, while preferences are stored via `FModel.getPreferences()`. If JSON contains stale `fullScreen=true` and preference `UI_BORDERLESS_FULLSCREEN_MODE=true` from a manual edit, `currentDisplayMode()` will return BORDERLESS while the preference UI shows Windowed. No migration step synchronizes JSON and preferences on load.

Test case:
```java
SettingData sd = Config.instance().getSettingData();
sd.fullScreen = true;
sd.borderlessFullScreen = false;
FModel.getPreferences().setPref(FPref.UI_BORDERLESS_FULLSCREEN_MODE, true);
FModel.getPreferences().setPref(FPref.UI_FULLSCREEN_MODE, false);
// currentDisplayMode() returns WINDOWED per SettingsScene logic
// SettingsPage.getSelectedValue() returns BORDERLESS per preference check
```
Impact: Inconsistent UI state and possible loss of user preference.

#### 3. Monitor virtual coordinates can be negative on multi-monitor setups
`GameLauncher` obtains primary monitor via `Lwjgl3ApplicationConfiguration.getPrimaryMonitor()` and uses `monitor.virtualX` / `monitor.virtualY` for window position. On macOS multi-monitor setups, the primary monitor's virtual origin is not guaranteed to be (0,0); it can be negative. Setting window position with negative coordinates may place the borderless window off-screen on some configurations, making it unrecoverable without killing the process.

Test case:
```java
Monitor monitor = Lwjgl3ApplicationConfiguration.getPrimaryMonitor();
int vx = monitor.virtualX;
int vy = monitor.virtualY;
// On a left-mounted secondary display, vx may be -1920
config.setWindowPosition(vx, vy);
// Window appears off-screen
```
Impact: Window lost off-screen, user cannot interact.

#### 4. SettingsScene select box initialization uses `currentDisplayMode()` which reads SettingData before preferences are loaded
`currentDisplayMode()` reads `Config.instance().getSettingData()` at construction time. If `Config` hasn't finished loading JSON settings, `borderlessFullScreen` defaults to `false`, causing the dropdown to initialise to Windowed even though preferences already contain Borderless. No null guard for `Config.instance()` null.

Test case:
```java
// Config.instance().getSettingData() returns default SettingData with borderlessFullScreen=false
// Preference UI later updates to true after async load
// Dropdown flashes Windowed -> Borderless
```
Impact: Flicker and confusing UX.

#### 5. New SettingData field `borderlessFullScreen` added without default serialization handling
`SettingData` is deserialized from JSON adventure config files. Adding a primitive boolean field without `@JsonProperty` default means older config files missing the field will deserialize `borderlessFullScreen` as `false` (Java default). That's fine for new installs, but existing adventure saves that store `fullScreen=true` will interpret as fullscreen, not borderless. However the preference `UI_BORDERLESS_FULLSCREEN_MODE` defaults to `"false"` string. If a user upgrades with an existing config that has `fullScreen=true`, switching to Borderless mode will write both `fullScreen=false` and `borderlessFullScreen=true`, but reverting to Fullscreen will leave `borderlessFullScreen=true` stale, causing `currentDisplayMode()` to return Borderless on macOS despite user selecting Fullscreen.

Test case:
```java
SettingData sd = new SettingData();
sd.fullScreen = true; // loaded from old config
sd.borderlessFullScreen = false; // default
// User selects Fullscreen via UI -> sets sd.fullScreen=true, sd.borderlessFullScreen=true (bug)
// Later user selects Fullscreen again -> sd.fullScreen=true, sd.borderlessFullScreen=false? No, code sets borderlessFullScreen=false only when mode != BORDERLESS
```
Impact: Stale borderless flag persists.

#### 6. MacOSPresentation.ensureMinimizeMenu creates new NSMenu each time presentation is enabled
`ensureMinimizeMenu` checks `WINDOWS_MENU` existence via `JNI.invokePPJ(application, WINDOWS_MENU, ...)`. If the menu already exists from a previous run, it returns early. However if the application is relaunched with a different menu configuration, the old menu may still be attached, leading to duplicate "Window" menu items on subsequent `setBorderlessVisible(true)` calls. The code never removes previously added menu items.

Test case:
```java
MacOSPresentation.setBorderlessVisible(true);
MacOSPresentation.setBorderlessVisible(false);
MacOSPresentation.setBorderlessVisible(true);
// Menu item count increases
```
Impact: Duplicate menu entries, UI clutter.

#### 7. Borderless fullscreen mode disables window decorations but does not disable OS-level window management
`config.setDecorated(false)` removes window frame, but the window remains resizable via OS-level drag? On macOS, undecorated windows still receive mouse events at screen edges. The code sets `config.setResizable(false)` only for windowed mode, not for borderless mode. Borderless mode uses `setWindowedMode` with full screen dimensions, but resizable flag defaults to true. User can resize window, breaking the full-screen illusion.

Test case:
```java
boolean borderlessFullScreen = true;
config.setWindowedMode(displayMode.width, displayMode.height);
config.setDecorated(false);
// config.setResizable() not called -> defaults to true
```
Impact: User can resize "fullscreen" window.

### Performance and security
#### 8. JNI calls performed on every iconify event without caching
`MacOSPresentation.setBorderlessVisible` performs `JNI.invokeI(PTHREAD_MAIN)` and potentially `dispatch_sync_f` on every iconify event. Iconify events can fire rapidly (e.g., user minimizes/restores repeatedly). Each call performs Objective-C runtime lookups via `ObjCRuntime` and allocates `MemoryStack` in `ensureMinimizeMenu`. No caching of selector handles or application object pointer.

Test case:
```java
for (int i=0;i<1000;i++) {
    MacOSPresentation.setBorderlessVisible(i%2==0);
}
// Measure time spent in JNI
```
Impact: UI stutter on repeated minimize.

#### 9. Hard-coded display mode strings risk localization mismatch
`DISPLAY_MODE_WINDOWED`, `DISPLAY_MODE_FULLSCREEN`, `DISPLAY_MODE_BORDERLESS` are hard-coded English strings used for comparison (`equals(o)`). If the UI is localized, the combo box options may be displayed via localizer but the comparison remains English, causing selection to never match and settings to reset on every UI open.

Test case:
```java
// Localizer returns "Fenêtre" for Windowed in French
// Combo box displays localized string
// valueChanged receives "Fenêtre"
// DISPLAY_MODE_WINDOWED.equals("Fenêtre") -> false
// Settings incorrectly set to false
```
Impact: Settings lost on non-English locales.

## Verdict
Request Changes

The macOS borderless fullscreen implementation introduces valuable functionality but contains several correctness and safety issues around thread dispatching, multi-monitor coordinates, settings synchronization, and localization. The JNI bridge for NSApplication presentation options is fragile and may deadlock. Settings state diverges between `SettingData` JSON and `ForgePreferences`, risking inconsistent UI. Multi-monitor negative virtual coordinates and missing resize guard further undermine reliability. Address these before merging.

## Recommended Fixes
1. Ensure `MacOSPresentation.setBorderlessVisible` runs on the AppKit main thread using `dispatch_async` rather than `dispatch_sync_f` to avoid deadlock, and cache `NSApplication` pointer and selector handles.
2. Synchronize `SettingData.borderlessFullScreen` and `UI_BORDERLESS_FULLSCREEN_MODE` on load/save, with migration logic that clears stale borderless flag when fullscreen mode changes.
3. Clamp `monitor.virtualX`/`virtualY` to >=0 or query display bounds to ensure window is on-screen on multi-monitor setups.
4. Initialize display mode dropdown from preferences first, falling back to `SettingData`, and guard against `Config` null.
5. Disable resizing for borderless mode: `config.setResizable(false)` in borderless branch.
6. Use localized resource keys for display mode comparison, not hard-coded English strings.
7. Ensure `ensureMinimizeMenu` is idempotent and removes duplicate menu items on re-enable.

---

PR #11350, opened 2026-07-23
Adventure mode: Read starting equipment from difficulty config

vanja-ivancevic:
- Read the starting-equipment tooltip from the selected difficulty configuration
- Fix Old Border incorrectly showing Leather Boots, which the plane does not grant
- Show `None` when no starting equipment is configured


Agetian:
Looks good to me if it works in standard Adventure and Realm of Legends without additional config updates :)

Agetian (approved):

PR #11354, opened 2026-07-23
Implement Hobbit Storied / an enduring story (#11282)

### LINKED ISSUE #11282: Hobbit: Storied
> Storied (If you control three or more artifacts, legendaries, and/or Sagas, you have an enduring story for the rest of the game.)

similar to Ascend and the City's blessing

liamiak:
Closes #11282.

> Storied (If you control three or more artifacts, legendaries, and/or Sagas, you have an enduring story for the rest of the game.)

As the issue says, this is the same shape as Ascend and the city's blessing, so it mirrors that mechanic rather than generalising it.

**Two separate commits**, so the second can be dropped or relocated independently:

1. the mechanic (engine)
2. the two previewed card scripts

## 1. The mechanic

| piece | mirrors |
|---|---|
| `Keyword.STORIED`, `ApiType.EnduringStory` | `ASCEND`, `Ascend` |
| `CardFactoryUtil` wiring | Ascend's always-on static trigger, gated `EnduringStory$ False` so it stops firing once gained, with `IsPresent … PresentCompare$ GE3` |
| the trigger's overriding ability, built inline in `CardFactoryUtil` | `AscendEffect`, minus the `ApiType` and effect class — nothing scripts this API, so it needs none |
| `Player.enduringStoryEffect` + `hasEnduringStory`/`setEnduringStory` (+ snapshot copy) | `blessingEffect`/`hasBlessing`/`setBlessing` |
| check syntax | the blessing's four entry points |

Cards can test it exactly as they test the blessing: `EnduringStory$` on a trait, `Activation$ EnduringStory` on an ability, **`Condition$ EnduringStory` on a continuous static** (the form Tendershoot Dryad and 8 other cards use for the blessing), and `Player.hasEnduringStory` as a property.

A permanent that is more than one of artifact/legendary/Saga still counts **once** — there's a test pinning that, since the union `IsPresent` is the part most likely to be subtly wrong.

Scoped to permanents, as Ascend's keyword wiring is. No previewed Storied spells are known; if the set has any they'd need the separate spell-resolution path Ascend uses.

## 2. The card scripts

`Bifur, Melodic Rider` (147) and `Thorin Oakenshield` (165) — both already listed in `editions/The Hobbit.txt` with these numbers and artists, neither scripted yet. They go in `cardsfolder/upcoming/` beside the other Hobbit cards.

They reuse existing precedents: the two-trigger "enters or attacks" pattern from Aetherwind Basker, `Mode$ Panharmonicon | ValidCard$ Dwarf.YouCtrl` for Bifur's extra Dwarf trigger, and `AddKeyword$ Ward:1` for Thorin.

They also earn their place: both have "As long as you have an enduring story, …" as a continuous static, which is how I found that `StaticAbility.checkConditions` needed its `EnduringStory` entry — the mechanic was incomplete without it.

## Verified

10 new tests, full desktop suite green (296). Highlights: below/at the three-permanent threshold; artifacts, legendaries and Sagas counting together; a legendary artifact counting once rather than twice; the story persisting after you drop back below three; opponents not getting it; and Thorin's ward appearing **only** once his controller reaches three qualifying permanents.

## Notes

- **Happy to rework this as a generalised player-designation** (blessing/monarch/initiative/story sharing one mechanism) if you'd prefer that to mirroring — I went with the mirror to keep the diff tight and avoid touching working City's Blessing code, but it's your call.
- **Happy to move the two card scripts** to the Card Scripts: The Hobbit (HOB) project instead, if they'd rather live there than in this PR. Dropping commit 2 leaves the mechanic standalone.
- The designation card has no image key (there's `ImageKeys.BLESSING_IMAGE` for the blessing; I didn't invent a story equivalent).

---

Written with Claude Code (Opus 4.8), per the AI coding agent guidance in CONTRIBUTING.md; also recorded as a commit co-author.


forge-game/src/main/java/forge/game/ability/effects/EnduringStoryEffect.java
```diff
@@ -0,0 +1,56 @@
...
+
+public class EnduringStoryEffect extends SpellAbilityEffect {
+
+    /** Artifacts, legendaries and/or Sagas - a permanent that is more than one of these still counts once. */
+    public static int countStoried(final Player p) {
+        return CardLists.count(p.getCardsIn(ZoneType.Battlefield), EnduringStoryEffect::isStoried);
+    }
+
+    private static boolean isStoried(final Card c) {
+        return c.isArtifact() || c.getType().isLegendary() || c.getType().hasSubtype("Saga");
```
tool4ever: `isHistoric`

liamiak:
Claude: Good call — switched `countStoried` to `isHistoric()` and the trigger to `Permanent.YouCtrl+Historic`. Thanks!

Liamiak: I feel silly for missing this one - thanks for pointing us to the historical function!

Hanmac:
i don't know if we need it as StaticTrigger (with its own Effect) or if we should do it as Static Action

tool4ever:
well, the mechanics article doesn't make it sound like it would be a SBA

liamiak:
Worth noting Ascend uses both shapes, one per card type — they're not alternatives:

- **Permanents** get the state trigger (`Mode$ Always | Static$ True | Blessing$ False | …`) with `DB$ Ascend`, which uses the stack.
- **Instants and sorceries** don't get a trigger at all. `AbilityUtils.resolvePreAbilities` checks the keyword and the permanent count inline and calls `setBlessing` directly, deliberately before condition checks so the spell can then satisfy its own `Condition$ Blessing`.

I mirrored only the first half, so as it stands a Storied instant or sorcery would do nothing. I didn't because I've not seen cards for it yet, and didn't want to add speculative code, though that may not be the preferred path.

On the SBA question: agreed it isn't one. Ascend on a permanent is a state-*triggered* ability (CR 702.131) and uses the stack, which is what `Static$ True` gives; a state-based action wouldn't. So the permanent path reads correctly as a trigger to me, and the spell path is the genuinely action-shaped one.

Hanmac:
@tool4ever what do you think?
Should AscendEffect and this new Effect be combined into a PlayerAttributeEffect?

tool4ever:
Hmn, is there even a point to keep such internal effect around?
We could just construct it directly in `CardFactoryUtil` 🤔

Hanmac:
> Hmn, is there even a point to keep such internal effect around? We could just construct it directly in `CardFactoryUtil` 🤔

Ai might not like it if the Static Trigger has no API Effect 
You can try to make it without extra Effect

Hanmac:
@tool4ever @liamiak can we get this MR ready?

tool4ever (dismissed):

Hanmac:
@tool4ever 
i still would prefer it without extra Effect, maybe that can be changed later?

liamiak:
Done — the internal effect is gone and the ability is built directly in `CardFactoryUtil`, as
@tool4ever suggested. Net 39 lines lighter: no `EnduringStoryEffect`, no `ApiType` entry. Your
premise checked out, incidentally — `$ Ascend` appears in zero card scripts and `EnduringStory`
only ever as a `Condition$`, so both classes only ever served their own keyword's trigger.

@Hanmac your warning was half right, and it is worth writing down what the inline form has to
handle.

**The AI does need a way in.** `AiController.doTrigger` dispatches on `ApiType`, and without one
there is a single fallback:

```java
if (sa.getPayCosts() == Cost.Zero && !sa.usesTargeting()) {
    // For non-converted triggers (such as Cumulative Upkeep) that don't have costs or targets to worry about
    return true;
}
```

That compares against the `Cost.Zero` *instance*. Passing `Cost.Zero` to the constructor is not
enough, because the trigger is copied onto each card and `SpellAbility.copy` does
`clone.setPayCosts(getPayCosts().copy())`, so the copy holds an equal-but-different cost.
`Cost` has no value equality either — I measured both, and `equals` is false as well — so
`getPayCosts` is pinned to the singleton in the override.

**The part that actually broke it was not the AI at all.** `getOriginalHost()` is null on an
inline ability, so mirroring Ascend's `getOriginalHost().getSetCode()` threw inside `resolve()` and
the exception was swallowed. The symptom was identical to the AI refusing to run the trigger, which
is what sent me down the wrong path first. It takes the set code from `getHostCard()` now.

296 tests, 0 failures.

One thing I noticed but deliberately left alone: our trigger reports `mandatory=true` into
`doTrigger`, and that flag is ignored on the API-less path — a mandatory trigger the AI cannot
evaluate is silently skipped rather than run. Storied does not need that changed, since the
`Cost.Zero` route works, so it is not in this PR. Happy to look at it separately if you think the
AI should honour mandatory there.

forge-game/src/main/java/forge/game/card/CardFactoryUtil.java
```diff
@@ -638,6 +638,35 @@ public static void addTriggerAbility(final KeywordInterface inst, final Card car
...
+                        + " | TriggerDescription$ Storied (" + inst.getReminderText() + ")";
+
+                final Trigger trigger = TriggerHandler.parseTrigger(trig, card, intrinsic);
+                trigger.setOverridingAbility(new AbilityStatic(card, Cost.Zero, null) {
+                    @Override
+                    public Cost getPayCosts() {
+                        // AiController.doTrigger only runs an API-less trigger when it recognises a
+                        // free one, and it compares against the Cost.Zero instance. Copying the
+                        // trigger onto each card would otherwise replace it with an equal copy.
+                        return Cost.Zero;
```
tool4ever: seems like we could just optimize SpellAbility.copy() to reuse the instance if it's the static one?

Hanmac: imo the `== Cost.Zero` checks needs to be fixed
better check `isFree` instead?

tool4ever: yea that probably too

tool4ever (approved):

PR #11359, opened 2026-07-24
fix fates reversal mana cost

coder-5:
afr-A-102-a-fates-reversal


Jetz72:
The actual card costs 1B - https://scryfall.com/card/afr/102/fates-reversal

The Alchemy print costs B, and we have a separate script for that - https://github.com/Card-Forge/forge/blob/master/forge-gui/res/cardsfolder/rebalanced/a-fates_reversal.txt

PR #11360, opened 2026-07-24
Make AI-vs-AI games deterministic at a fixed seed

liamiak:
## Problem

AI-vs-AI games were **nondeterministic across runs even with a fixed `MyRandom` seed** — the same game replayed differently from one JVM launch to the next (on a reanimator test deck, ~40% of seeds diverged). This makes bugs hard to reproduce, blocks deterministic regression testing, and makes before/after performance comparison unreliable.

## Root cause

Order-significant iteration over collections whose key type uses **identity-based `hashCode`**, which the JVM randomizes per launch. This covers enums (`ManaCostShard`), and objects whose `hashCode` folds in a `Class` object's identity hash (e.g. `Trigger.hashCode() = Objects.hash(Trigger.class, id)`).

Diagnosed with `-XX:+UnlockExperimentalVMOptions -XX:hashCode=3` (deterministic identity hashes): under that flag replays were stable, under default hashing they diverged — confirming the cause is identity-hash-ordered iteration, not the RNG. Each site below was then localized by log-diffing the first point two runs diverge.

## Fixes (each independently revertable)

1. **Simultaneous trigger order** — `TriggerWaiting` collected fired triggers into a `HashMap<Trigger,Player>` and returned `keySet()`; `MagicStack` uses that as the order simultaneous triggers go on the stack. → `LinkedHashMap` (the setter already receives an ordered list). *First divergence seen: Solitude's evoke-sacrifice vs ETB-exile triggers swapping.*

2. **Mana shard order** — `ManaCostBeingPaid.unpaidShards` (`HashMap`) and `ComputerUtilMana`'s shard multimap (HashMap-backed) are keyed by the `ManaCostShard` enum and drive which source taps for a color. → `EnumMap` / `MultimapBuilder.enumKeys`. This also honors the ordering both the enum and the field's own comment already document ("least ways to pay first", "color before generic"). *First divergence: tapping Hallowed Fountain vs Godless Shrine for {W}.*

3. **Sacrifice candidate order** — the `+WithDifferentNames` sacrifice path filtered candidates through a `HashSet<Card>` and copied it back, scrambling the tie-break order. → `LinkedHashSet`.

4. **Attack tie-breaking** — `AttackConstraints.getLegalAttackers` picks the fewest-violation attack via `stream().min()` (first on tie) over a map filled from `legalAttackers.asSet()` (a `HashSet` keyed by attack-config maps). → iterate the ordered `FCollection`. Only affects boards with attack requirements (goad/lure/must-attack) that tie.

## Validation

- Full `forge-gui-desktop` suite: **286 tests, 0 failures**.
- **Determinism:** two full 5-seed passes (default hashing) are now byte-identical pass-to-pass on a reanimator deck, a Goblins deck, and a 510-card pile — where ~40% diverged before. Independently corroborated by pinning identity hashes with `-XX:hashCode=3`.

## Scope / honesty

I audited the highest-traffic decision paths — spell/ability selection, targeting, discard, mana, sacrifice, and combat — and the core selection logic (`Aggregates.itemWithMax` over zone-ordered lists, list-based discard, seeded `MyRandom` sampling) was already deterministic. This PR fixes the concrete sources that manifest across these decks plus the suite; I don't claim *every* path is covered (clone/copy, pile-splitting, and sim-mode copies weren't exhaustively swept). Each change is behavior-preserving apart from making a previously-arbitrary order deterministic.

---
🤖 Implemented with the assistance of Claude Code (Opus). Happy to split into per-fix PRs if that's easier to review.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtil.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtil.java
@@ -536,7 +536,9 @@ public class ComputerUtil {
 
         CardCollection typeList = CardLists.getValidCards(ai.getCardsIn(ZoneType.Battlefield), type.split(";"), source.getController(), source, ability);
         if (differentNames) {
-            final Set<Card> uniqueNameCards = Sets.newHashSet();
+            // LinkedHashSet: this list is copied back into typeList and drives which cards get
+            // sacrificed on a scoring tie, so its order must not depend on Card identity hashCode.
+            final Set<Card> uniqueNameCards = Sets.newLinkedHashSet();
             for (final Card card : typeList) {
                 // CR 201.2b Those objects have different names only if each of them has at least one name and no two objects in that group have a name in common
                 if (!card.hasNoName()) {
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCost.java
@@ -385,7 +385,10 @@ public class ComputerUtilCost {
 
                 CardCollection typeList = CardLists.getValidCards(ai.getCardsIn(ZoneType.Battlefield), type.split(";"), source.getController(), source, sourceAbility);
                 if (differentNames) {
-                    final Set<Card> uniqueNameCards = Sets.newHashSet();
+                    // LinkedHashSet: this list is copied back into typeList and drives which cards
+                    // get sacrificed on a scoring tie, so its order must not depend on Card
+                    // identity hashCode (which varies per JVM run).
+                    final Set<Card> uniqueNameCards = Sets.newLinkedHashSet();
                     for (final Card card : typeList) {
                         // CR 201.2b Those objects have different names only if each of them has at least one name and no two objects in that group have a name in common
                         if (!card.hasNoName()) {
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
@@ -5,6 +5,7 @@ import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
+import com.google.common.collect.MultimapBuilder;
 import forge.ai.AiCardMemory.MemorySet;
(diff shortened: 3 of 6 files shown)


Jetz72:
Wonder how viable it'd be to give cards a proper `hashCode`. Could base it on ID and game, but I think different instances are also created when moving between zones, maybe a timestamp would also be needed? There's also the game state copying and rollback logic to consider... 

Also, the comments are nice to contextualize the lines of code in relation to the PR, but parts of them are way too specific to fit in with the codebase.

tool4ever (approved):
yea this will help on bug hunts
I'll just trim the comments a bit with next cleanup

liamiak:
Fair comment on the comments; I'll try to be more succinct in the future.

On proper card hashCodes: viable, but narrower in effect than it first looks. Two things I hit while digging:

- `gameTimestamp` is mutable (`setGameTimestamp`), so it can't go into a hashCode — a key whose hash changes while it's in a map becomes unfindable. `id + game` is stable for an object's lifetime, but ids aren't unique: `CardCopyService` reuses the original's id for LKI and game-state copies, so a card and its copies would share a bucket. Harmless with identity `equals`, though value-hash + identity-equals is unusual enough to want a comment saying it's deliberate.
- It also wouldn't have covered most of this PR. Of the four fixes, only `AttackConstraints` is keyed on anything involving Card; the other three are keyed on `Trigger` and on the `ManaCostShard` enum.

The narrower systemic culprit is that `Trigger`/`SpellAbility`/`StaticAbility`/`ReplacementEffect` all hash as `Objects.hash(SomeClass.class, id)` — and `SomeClass.class.hashCode()` is the Class object's identity hash, randomized per JVM. Dropping that component makes all four deterministic, one line each. Pinning identity hashes with `-XX:hashCode=3` stabilised replays, which is what pointed at it. Enum-keyed maps need `EnumMap` either way.

One caveat on the whole idea: stable hashing makes iteration *deterministically* arbitrary — great for reproducibility, but not the same as correct order. `TriggerWaiting` is already handed an ordered list (CR 603.3b lets the controller order simultaneous triggers), and `ManaCostShard`'s declaration order is documented in the enum as "least ways to be paid for first."

Happy to take the `.class` change and/or a Card hashCode as a follow-up PR if that's useful.

PR #11363, opened 2026-07-24
Add option to keep background music playing when the window loses focus

liminalwarmth:
Adds a **Pause Music in Background** option (default on, preserving current behavior) that lets players keep background music playing when the Forge window loses focus.

This was originally part of #11347, where borderless fullscreen kept the music playing while unfocused as a side effect of the window mode. Per [review feedback there](https://github.com/Card-Forge/forge/pull/11347#discussion_r3646299749), it is better suited as a stand-alone sound option: a separate preference, independent of window mode, with the policy handled in `SoundSystem` rather than the window listener.

### What

- New preference `UI_PAUSE_MUSIC_ON_FOCUS_LOSS`, defaulting to `true` (music pauses when unfocused, exactly as today).
- `SoundSystem.updatePlayPause()` consults the preference; the window listener continues to report raw focus state and is unchanged.
- Checkbox beside the existing volume controls in the mobile settings page and the Adventure settings scene. It is hidden on Android, where the desktop focus events never fire (Android music pausing is handled by the activity lifecycle, unaffected by this change).
- Not platform-gated otherwise — the logic is pure Java and applies to any desktop window mode, windowed or fullscreen.

### Testing

- `mvn -pl forge-gui-mobile-dev -am test -DskipTests=false` passes (3 tests).
- `mvn -B -pl forge-gui-mobile-dev -am package -DskipTests` passes; checkstyle reports 0 violations.
- Default-on preference preserves the existing pause-on-unfocus behavior; the new code path only runs when a player opts out.


Jetz72 (dismissed):
I suppose that `isAndroid` check might need to be updated as of 20 minutes ago, but so will many others across the project.

liminalwarmth:
> I suppose that `isAndroid` check might need to be updated as of 20 minutes ago, but so will many others across the project.

Hey, cool, I was following that iOS support thread. Might have to crack out the iPad. This is an easy change, one sec.

liminalwarmth:
Good timing 😄 — updated the guards here to `!isAndroid() && !isIOS()` and rebased onto master (the branch predated #11190, so `isIOS()` didn't exist on it yet). Also opened #11382 adding a `GuiBase.isMobile()` helper for the broader cleanup you mentioned.

Jetz72 (approved):

PR #11373, opened 2026-07-25
AI: make the untap-for-ramp logic fire, and point it at the mana source

liamiak:
`AILogic$ PoolExtraMana` is meant to let the AI tap a Voyaging Satyr to untap a land and reach a
bigger spell. Two things stop it working.

**The gate counts the wrong mana.** `doPoolExtraManaLogic` asks `hasEnoughManaSourcesToCast`, which
counts potential sources including tapped ones, and then asks whether the cost minus the freed mana
is payable using only sources it can actually use. Two notions of available mana in one test, so the
case the card exists for — plenty of lands, tapped out, untapping one reaches the spell — is exactly
the case it declines. Measured, Voyaging Satyr with a `{2}{G}` in hand:

| board | before | after |
|---|---|---|
| 2 untapped, 0 tapped | untaps a Forest | same |
| 2 untapped, 1 tapped | nothing | untaps a Forest |
| 2 untapped, 2 tapped | nothing | untaps a Forest |
| 1 untapped, 2 tapped | nothing | nothing |
| 3 untapped, 0 tapped | casts it | same |

**The decision never reaches the target chooser.** `untapPrefTargeting` falls through to
`getMostExpensivePermanentAI`. On a land-only untapper you can't tell, because every legal target is
a land; on one that untaps any permanent it takes the biggest creature instead of the land. It now
asks `getBestLandAI` when the reason is mana.

With both fixed the same AILogic covers cards that untap any permanent, so this drops
`AI:RemoveDeck:All` from Kiora's Follower, Unbender Tine, Vizier of Tumbling Sands, Tidewater Minion
and Krosan Restorer, and adds the logic to Argothian Elder and Ley Weaver, which are unflagged and
untap two lands today whether or not it helps. Aphetto Alchemist stays flagged: it untaps artifacts
and creatures, never lands.

The opponent's-turn clause untapped unconditionally. It now checks that we hold something castable
at that timing, which is what its own comment always said it was for.

**Cost.** The gate swap makes one of these cards' main-2 decision about 22% dearer — 10.6ms to
12.9ms for a whole `chooseSpellAbilityToPlay` on a 40-permanent board, because `canPayManaCost` is
1.6ms against `hasEnoughManaSourcesToCast` at 1.1ms. The opponent's-turn check is deliberately kept
off the mana solver for the same reason: 4.1ms against 4.2ms before. One board, one machine.

Four tests, three of which fail without the change. Krosan Restorer's threshold ability and Tidewater
Minion's `{4}` Debuff were checked separately, since the flag was suppressing those too. Vizier's
cycling trigger I reasoned about rather than measured: it is an untargeted free untap, so it should
not be a loss.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/UntapAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/UntapAi.java
@@ -2,6 +2,7 @@ package forge.ai.ability;
 
 import java.util.List;
 import java.util.Map;
+import java.util.function.Predicate;
 
 import forge.ai.AiAbilityDecision;
 import forge.ai.AiPlayDecision;
@@ -228,7 +229,7 @@ public class UntapAi extends SpellAbilityAi {
                 if (choice == null) {
                     if ("PoolExtraMana".equals(sa.getParam("AILogic"))) {
                         // untapping was decided on to reach a spell, so untap what makes the mana
-                        choice = ComputerUtilCard.getBestLandAI(CardLists.filter(untapList, CardPredicates.LANDS_PRODUCING_MANA));
+                        choice = ComputerUtilCard.getBestAI(CardLists.filter(untapList, CardPredicates.PRODUCES_MANA));
                     } else if (CardLists.getNotType(untapList, "Creature").isEmpty()) {
                         choice = ComputerUtilCard.getBestCreatureAI(untapList); // if only creatures take the best
                     } else if (!sa.getPayCosts().hasManaCost() || sa.isTrigger()
@@ -437,27 +438,20 @@ public class UntapAi extends SpellAbilityAi {
 
         // TODO: currently limited to Main 2, somehow improve to let the AI use this SA at other time?
         if (ph.is(PhaseType.MAIN2, ai) && untapReachesASpell(ai, playable, untappingCards.size())) {
-            CardCollection manaLandsTapped = CardLists.filter(ai.getCardsIn(ZoneType.Battlefield),
-                    CardPredicates.LANDS_PRODUCING_MANA, CardPredicates.TAPPED);
-            manaLandsTapped = CardLists.getValidCards(manaLandsTapped, sa.getParam("ValidTgts"), ai, source, null);
-
-            if (!manaLandsTapped.isEmpty()) {
-                // already have a tapped land, so agree to proceed with untapping it
+            if (!targetableManaSources(ai, sa, source, CardPredicates.TAPPED).isEmpty()) {
+                // already have a tapped source, so agree to proceed with untapping it
                 return true;
             }
 
-            // pool one additional mana by tapping a land to try to ramp to something
-            CardCollection manaLands = CardLists.filter(ai.getCardsIn(ZoneType.Battlefield),
-                    CardPredicates.LANDS_PRODUCING_MANA, CardPredicates.CAN_TAP);
-            manaLands = CardLists.getValidCards(manaLands, sa.getParam("ValidTgts"), ai, source, null);
+            // pool one additional mana by tapping a source to try to ramp to something
+            CardCollection manaSources = targetableManaSources(ai, sa, source, CardPredicates.CAN_TAP);
 
-            if (manaLands.isEmpty()) {
+            if (manaSources.isEmpty()) {
                 // nothing to untap
                 return false;
             }
 
-            Card landToPool = manaLands.getFirst();
-            SpellAbility manaAb = landToPool.getManaAbilities().getFirst();
+            SpellAbility manaAb = manaSources.getFirst().getManaAbilities().getFirst();
 
             ComputerUtil.playNoStack(ai, manaAb, game, false);
 
@@ -466,10 +460,18 @@ public class UntapAi extends SpellAbilityAi {
 
         // past declare blockers during the opponent's turn and right before our turn, untapping is
         // worth it for something we can hold up, e.g. a removal spell or burn spell we can't pay for
-        // yet - the lands untap on their own next turn, so it buys nothing beyond that
+        // yet - the sources untap on their own next turn, so it buys nothing beyond that
         return ph.getNextTurn() == ai
                 && (ph.is(PhaseType.COMBAT_DECLARE_BLOCKERS) || ph.getPhase().isAfter(PhaseType.COMBAT_DECLARE_BLOCKERS))
                 && holdsSomethingToCast(ai, inHand);
+    }
+
+    /** The mana sources in the given state that this untap could legally take as its target. */
+    private static CardCollection targetableManaSources(final Player ai, final SpellAbility sa,
+            final Card source, final Predicate<Card> state) {
+        CardCollection sources = CardLists.filter(ai.getCardsIn(ZoneType.Battlefield),
+                CardPredicates.PRODUCES_MANA, state);
+        return CardLists.getValidCards(sources, sa.getParam("ValidTgts"), ai, source, null);
     }
 
     /**
--- a/forge-ai/src/main/java/forge/ai/ability/UntapAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/UntapAi.java
@@ -39,7 +39,9 @@ public class UntapAi extends SpellAbilityAi {
     @Override
     protected boolean checkAiLogic(final Player ai, final SpellAbility sa, final String aiLogic) {
         if ("PoolExtraMana".equals(aiLogic)) {
-            return doPoolExtraManaLogic(ai, sa);
+            // mana is the reason this logic exists, but it must not veto the untaps that are worth
+            // making on their own, e.g. a Time Vault that cannot untap itself
+            return doPoolExtraManaLogic(ai, sa) || hasPriorityUntapTarget(ai, sa);
         }
         if ("PreventCombatDamage".equals(aiLogic)) {
             return doPreventCombatDamageLogic(ai, sa);
@@ -464,6 +466,21 @@ public class UntapAi extends SpellAbilityAi {
         return ph.getNextTurn() == ai
                 && (ph.is(PhaseType.COMBAT_DECLARE_BLOCKERS) || ph.getPhase().isAfter(PhaseType.COMBAT_DECLARE_BLOCKERS))
                 && holdsSomethingToCast(ai, inHand);
+    }
+
+    /**
+     * True when something we could target is worth untapping for its own sake, whatever the mana
+     * situation. The targeting pass applies the full filters, so this only has to be permissive
+     * enough not to lose those cards before it runs.
+     */
+    private static boolean hasPriorityUntapTarget(final Player ai, final SpellAbility sa) {
+        // tapped-ness is a field read and targetability is not, so narrow on it first
+        CardCollection tapped = CardLists.filter(ai.getYourTeam().getCardsIn(ZoneType.Battlefield),
+                CardPredicates.TAPPED);
+        if (tapped.isEmpty()) {
+            return false;
+        }
+        return detectPriorityUntapTargets(CardLists.getTargetableCards(tapped, sa)) != null;
     }
 
     /** The mana sources in the given state that this untap could legally take as its target. */
--- a/forge-ai/src/main/java/forge/ai/ability/UntapAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/UntapAi.java
@@ -226,7 +226,10 @@ public class UntapAi extends SpellAbilityAi {
                 choice = detectPriorityUntapTargets(untapList);
 
                 if (choice == null) {
-                    if (CardLists.getNotType(untapList, "Creature").isEmpty()) {
+                    if ("PoolExtraMana".equals(sa.getParam("AILogic"))) {
+                        // untapping was decided on to reach a spell, so untap what makes the mana
+                        choice = ComputerUtilCard.getBestLandAI(CardLists.filter(untapList, CardPredicates.LANDS_PRODUCING_MANA));
+                    } else if (CardLists.getNotType(untapList, "Creature").isEmpty()) {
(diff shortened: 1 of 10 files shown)


forge-ai/src/main/java/forge/ai/ability/UntapAi.java
```diff
@@ -347,6 +355,78 @@ public Card chooseSingleCard(Player ai, SpellAbility sa, Iterable<Card> list, bo
...
+    }
+
+    /**
+     * The tapped mana source worth untapping for ramp, or null. Untap a source if and only if it
+     * makes a difference to what we can play: some nonland spell in hand is uncastable now but
+     * becomes castable once that specific source is untapped. Verified with the real mana-payment
+     * check ({@link ComputerUtilMana#canPayManaCost}) against the source temporarily untapped, so it
+     * accounts for the actual cost - colours included - not just mana value.
+     */
+    private static Card pickRampUntap(final Player ai, final CardCollection untapList) {
```
tool4ever: code looks too unclean
also just redundant to `doPoolExtraManaLogic`...?

liamiak: Good calls, both — addressed in the last three commits.

- **Unclean:** the smell was a board poke — it flipped a card's tapped state to probe affordability and restored it in a `finally`. Replaced with a small shared helper, `untappingEnablesACast`, that models the freed mana by shaving generic off a *copy* of the cost (`ManaCostBeingPaid#decreaseShard`) and asking `canPayManaCost` against usable sources only. No tap/untap of real permanents.
- **Redundant:** you were right that it overlapped `doPoolExtraManaLogic`, so I routed the dorks through the same helper. That also surfaced a latent bug on the dork side: its old `!hasEnoughManaSourcesToCast` gate counts *tapped* sources, so when you're tapped out but source-rich it decided it had "enough" and declined to ramp — exactly the spot where untapping a land is what lets the spell resolve. The helper asks "can I pay with usable sources now, and would one more mana close the gap?" instead, which is a strict superset of the old firing (nothing that ramped before stops), and the whole `forge.ai` suite stays green.

Split into three commits — helper + generic path, then the dork routing/fix, then an optional test — so the fix and the test are separable.

liamiak: You were right on both counts, and following the second one to its end changed the whole PR.

`doPoolExtraManaLogic` wasn't only being duplicated — it was broken. It gated on `hasEnoughManaSourcesToCast`, which counts tapped sources, so the case it exists for (plenty of lands, tapped out, untapping one reaches the spell) was exactly the case it declined. Fixing that and the target choice makes the existing AILogic cover the cards I'd been hand-writing policy for, so the new methods are gone: +76/-41 in `UntapAi` instead of +169/-52, with card scripts doing the rest.

Force-pushed, so the thread above points at code that no longer exists.

forge-gui/res/cardsfolder/k/kioras_follower.txt
```diff
@@ -2,6 +2,5 @@ Name:Kiora's Follower
 ManaCost:G U
 Types:Creature Merfolk
 PT:2/2
-A:AB$ Untap | Cost$ T | ValidTgts$ Permanent.Other | TgtPrompt$ Select another target permanent. | SpellDescription$ Untap another target permanent.
-AI:RemoveDeck:All
+A:AB$ Untap | Cost$ T | ValidTgts$ Permanent.Other | TgtPrompt$ Select another target permanent. | AILogic$ PoolExtraMana | SpellDescription$ Untap another target permanent.
```
tool4ever, line 5: not sure ignoring the other card types is better in general

also I don't really like how the logic just guesses AI wants to pay with the extra shard, I need to think if it makes sense to synchronize it with ComputerUtilMana somewhow

liamiak: You were right, and it's worse than it looked — thanks for pushing on it.

Two separate things were narrowing these cards.

**"Mana source" was spelled "land".** The target choice filtered on `LANDS_PRODUCING_MANA`, so a tapped Llanowar Elves or Signet was invisible even though reusing one is the same ramp. Now `CardPredicates.PRODUCES_MANA` with `getBestAI`, which `detectPriorityUntapTargets` a few lines up already uses for mixed lists.

**The bigger one: the AILogic was vetoing untaps that are worth making anyway.** `checkAiLogic` runs before `checkApiLogic`, so when `doPoolExtraManaLogic` declined, `detectPriorityUntapTargets` never ran at all. Measured: with `AILogic$ PoolExtraMana` a Kiora's Follower will *not* untap a tapped Time Vault; without the tag it untaps it and takes the extra turn. The gate is now additive — `doPoolExtraManaLogic(...) || hasPriorityUntapTarget(...)` — so mana is *a* reason to untap rather than the only one.

Timing note: my first version of that check cost about a fifth of whole `chooseSpellAbilityToPlay` on a wide board, because `getTargetableCards` ran over everything. Filtering on tapped first brings it back to the noise floor.

The eleven land-only carriers are unaffected — their `ValidTgts` re-imposes `Land`, and on an all-land list the new predicate and `getBestAI` select exactly what the old ones did.

On the shard guess: that arithmetic and its TODO are already on master, I only moved them — not defending it, and happy to leave it untouched while you think about the `ComputerUtilMana` side. One data point in case it helps: since it only shaves `GENERIC`, a `{G}` one-drop is never predicted; I had to use a two-drop to make the gate fire in a test.

tool4ever: Still undecided but some notes:
- we would save the runtime cost of two extra payability checks per SA
  - since priority order for these pool abilities should be rather low most likely other SA costs have been checked before them in that priority window
    - might also make it easier to eventually remove the limit for permanent spells since we know AI considered the API playable (though some safety checks still run after `canPayCost`)
- but also lose the knowledge that running `checkApiLogic` in isolation gives us a "fully aware" decision
  - then again it's not really a hard contract, might find a way to cover both
  - there is already some precedent of API checks mixing things via `AiCardMemory`
  - still need to compare with related `ManaAi#doManaRitualLogic`

liamiak: I went and did the `doManaRitualLogic` comparison — the premise splits between the two.

`ManaRitual` is evaluated after the hand: `saComparator` exempts `ApiType.Mana` from the "cast 0 mana cost spells first (might be a Mox)" hoist, so the AI has genuinely costed the hand first (measured order: Sengir Vampire, Nightmare, Bog Wraith, Hill Giant, *then* Dark Ritual).

The untappers are hoisted to the **front** of the window — a `{T}` ability has total mana 0 and isn't `ApiType.Mana`. Measured, the untap is the only SA evaluated, so there's no prior knowledge yet to reuse.

Prototyped both directions (25 permanents, 5-card hand, configs interleaved in one JVM):

**(a) order + memory** — add `PoolExtraMana` to that `ApiType.Mana` exception; record `CantAfford`/`CantAffordX` hosts in a new `AiCardMemory` set cleared next to `HELD_MANA_SOURCES_FOR_NEXT_SPELL`; the scan only *narrows* on it — the payability check still decides, and an empty set falls back to the full scan, so `checkApiLogic` in isolation still computes its own answer. Gate 3.8 → 1.0 ms, 10 → 3 solver calls (reachable); 3.3 → 1.7 ms, 10 → 5 (unreachable).

**(b) prefilter** — no ordering change: skip a spell before any solver call when `cmc > getAvailableManaEstimate(ai, true) + extraGeneric`, i.e. `ManaRitual`'s arithmetic as a pre-filter for this exact check. Gate → 1.7 ms / 4 calls and 0.7 ms / 2 calls. Doesn't touch the isolation question at all; the cost is that the estimate is colour-blind and can under-count, so it can miss a ramp.

`UntapAiTest` passes 7/7 under either as default; full suite 362/0/6, same as baseline. Whole-phase timing I can't call — two runs disagreed on the sign.

Happy to land either separately from this PR, or to leave it entirely while you think about the `ComputerUtilMana` side. (b) is five lines and independent of everything you're weighing; (a) is the one that gives each logic the half it's missing.

tool4ever: thanks, admittedly I did not expect `saComparator` to also move all 0 cmc abilities to the front 🤦‍♂️

I will check if that's really the best default behaviour

tool4ever: notably it's a 14 year old change which includes more than its comment claims :/

- getting a better option for an activated ability by casting a permanent first seems smart
- but AI would then also prefer using mana for *Assassinate* even if *King's Assassin* is available too

so which order is more useful seems more like a highly situational 50/50 instead of an obvious heuristic?
though maybe the downside of wasting mana is too high :thinking:

liamiak: Tested the ordering question rather than argue it, and it goes against my own option (a). Same board as `rampsWhenTappedOutButSourceRich`, but with both a Centaur Courser `{2}{G}` (needs the untap) and a Grizzly Bears `{1}{G}` (doesn't) in hand:

| | untapped? | cast |
|---|---|---|
| today | yes | Centaur Courser |
| (a) untapper last | **no** | Grizzly Bears - Courser stranded |
| (b) prefilter | yes | Centaur Courser |

Evaluated after the hand, the AI takes the affordable 2-drop and the untap never happens. So the hoist is doing real work for this logic - please drop (a). (b) is the five-liner that doesn't touch ordering at all. None of the seven `UntapAiTest` cases holds both an affordable and an unaffordable spell, which is why this never surfaced.

For the wider question: `AIActivateLast$ True` is checked in `compareEvaluator` *above* the 0-cost hoist, so cards that lose under the default can already opt out - only Vivien of the Arkbow uses it today. And the priority loop (`PhaseHandler` ~1049) keeps asking until the player passes, so activating a free ability first doesn't forfeit the spell; it's cast on the next pass.

tool4ever: the problem with such tests is that they're pretty arbitrary:
if AI had another one drop available you could instead have support to go wide which is another viable strategy (more bodies to overwhelm opponent defense)

@tehdiplomat
your opinion on always using 0 cmc activated abilities first?

liamiak: Your #11642 looks like it answered this: the 0 cmc hoist stayed, the comment moved to "use 0 cmc abilities first (might be a Mox)" which is what it actually does, and going wide is now a TODO alongside it. I am reading that as settled unless you meant otherwise.

On that basis option (a) is dropped - the ordering change and the memory set with it - which is also where the tests you called arbitrary were pointing, so those go too. That leaves (b), the five-line prefilter that does not touch ordering at all.

liamiak:
Separately, and independent of the ordering question above: there is a colour form of the reach check ready, which settles the TODO in `untapReachesASpell`.

Untapping a source hands back that source's colour, not a generic mana, so a spell held up on coloured pips never shows as reachable - `{G}{G}` has no generic part for `decreaseShard` to reduce. With one untapped and one tapped Forest and Strangleroot Geist in hand, the AI keeps its untapper and casts nothing. +42/-9 with a test that fails without it, and it costs 19 extra solver probes over a twelve turn game, so about 8ms.

I have not pushed it here, because it delegates to `ComputerUtilCost.isPayableWith`, which #11504 adds - the same helper that PR now uses to decide what a land would turn on. Pushing it would drag that PR's contents into this one. So it waits on #11504 landing, or I can inline a local copy here if you would rather this one did not wait.

forge-gui/res/cardsfolder/k/kioras_follower.txt
tool4ever: nah I just updated it so the comment doesn't give the wrong impression on a quick glance

still looking into more strategic input...

liamiak: Human liamiak reasoning here:

1: There's chicken and egg between good mana and good play - Some of the behavior that flows out of this really flows out of the Forge AI preferring higher CMC effects over e.g. efficiency (like Assassinate vs King's Assassin). That's an edge case driven by value simplification preferences elsewhere in Forge AI, and it shouldn't stop us from making mana work better.

2: On the test: You're right from a play perspective CMC 2+1 vs CMC 3 to play can be a viable strategy. My mental heuristic when playing less-known decks (absent reason to expect board wipe, and in not-highly-tuned formats) is you're better off casting bigger spells, as smaller spells can fit in later around bigger spells for more efficient mana utilization). However, from a test perspective, you're putting yourself at a turn disadvantage to play a 2 CMC card vice a 3 because you can't use your own ramp.  The "right" play in CMC 3 vs 2+1 is a separate call, but right now there can't even be a choice - you just get CMC 2, which is generally worse.

tool4ever: worth noting that "ManaRitual" already gets a positive priority boost

liamiak: Human again: Doesn't that boost net it to 0 after the -9 from every Mana-API ability?

Looking at all this makes me realize Forge is doing mana eval backwards vs my internal heuristics - iterating on ordered spell/ability vs looking at resources first and then planning until the next turn against them, and then we're doing a lot of crunching to try to make up for that lack, across multiple PRs from many folks. Might be too big a change, though.  Trying to get my thoughts in order, but mana/payment eval seems core regardless.

tool4ever: ugh yes indeed that check should really be `isManaAbility` because something like *Deathrite Shaman* can't be activated while paying anyway - I'll change that, probably made AI screw up a fair amount 🤦‍♂️
thanks for pointing to this obvious part, sometimes it's these minor things that get missed which can already help a lot

yea it's all fairly complicated which is why I'm looking so hard to streamline and reuse things

PR #11375, opened 2026-07-25
Add Russian language support (ru-RU)

tCynik:
Adds Russian localization for the main UI:
- ru-RU.properties (translated interface strings)
- LangRussian (pluralization rules for Russian)
- Fix for Cyrillic glyphs not rendering (blank UI in ru-RU)

---
Note: this contribution was substantially coded with the assistance of Claude Code (Anthropic).

### DIFF
--- a/forge-core/src/main/java/forge/util/Lang.java
+++ b/forge-core/src/main/java/forge/util/Lang.java
@@ -42,6 +42,7 @@ public abstract class Lang {
             case "ja" -> new LangJapanese();
             case "ko" -> new LangKorean();
             case "fr" -> new LangFrench();
+            case "ru" -> new LangRussian();
             default -> new LangEnglish();
         });
     }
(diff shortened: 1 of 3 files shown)


Agetian (approved):

Agetian:
Looks good for the initial implementation. Not sure if Planeswalker should be kept as is (english Planeswalker since even on Russian cards, that's what they say), or we can use the transliterated Планесвокер, I think it works either way :) Can be merged and later improved as needed.

tCynik:
> Looks good for the initial implementation. Not sure if Planeswalker should be kept as is (english Planeswalker since even on Russian cards, that's what they say), or we can use the transliterated Планесвокер, I think it works either way :) Can be merged and later improved as needed.

Thanks!
I think so too, "Планесвокер" it's definitely wrong word, next PR i'll fix one. In official MTG localization i've seen it like a "Плейнсволкер". 
Next trying to implement adventure mode localization...

PR #11376, opened 2026-07-25
Discard: fix discardedMap

Hanmac:
This fixes the Connive/Recruit problem with Immutable Map

instead, create a new Map for Discarded cards

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
@@ -896,7 +896,7 @@ public abstract class SpellAbilityEffect {
         for (Map.Entry<Player, CardCollectionView> e : toDiscardMap.entrySet()) {
             discardedBefore.put(e.getKey(), Lists.newArrayList(e.getKey().getDiscardedThisTurn()));
             final CardCollection discardedByPlayer = new CardCollection();
-            for (Card card : Lists.newArrayList(e.getValue())) { // without copying will get concurrent modification exception
+            for (Card card : e.getValue()) {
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/ability/SpellAbilityEffect.java
```diff
@@ -890,30 +890,29 @@ public void run() {
...
+    protected static void discard(SpellAbility sa, final boolean effect, Map<Player, CardCollectionView> toDiscardMap, Map<AbilityKey, Object> params) {
         Map<Player, List<Card>> discardedBefore = Maps.newHashMap();
-        for (Player p : discarders) {
-            discardedBefore.put(p, Lists.newArrayList(p.getDiscardedThisTurn()));
+        Map<Player, CardCollectionView> discardedMap = Maps.newLinkedHashMap();
+        for (Map.Entry<Player, CardCollectionView> e : toDiscardMap.entrySet()) {
+            discardedBefore.put(e.getKey(), Lists.newArrayList(e.getKey().getDiscardedThisTurn()));
             final CardCollection discardedByPlayer = new CardCollection();
-            for (Card card : Lists.newArrayList(discardedMap.get(p))) { // without copying will get concurrent modification exception
+            for (Card card : Lists.newArrayList(e.getValue())) { // without copying will get concurrent modification exception
```
tool4ever: now the extra List isn't needed at least?

Hanmac:
I'm more confused on how a Null might get in there, but i left the check there

tool4ever (approved):
if only we had some Map interface that doesn't offer mutable methods to enforce it at compile time :)

PR #11379, opened 2026-07-25
AI: don't pay life on an optional trigger when it isn't safe (+ unflag Bringer of the Black Dawn)

liamiak:
## The bug

The AI's activated-ability path gates cost payment for *wisdom* — `willPayCosts` →
`checkLifeCost` ("don't drop below 4 life"). The **trigger** path, in `SpellAbilityAi.doTrigger`,
only checked cost *affordability* (`canPayCost`: "do I have 2 life?" → yes at 5). So any optional
"you may pay N life: <effect>" trigger would pay life it couldn't spare, every trigger, even into
lethal range — the AI bleeds itself out. This hits pay-life **tutors** (Bringer of the Black Dawn),
**reanimation** (Master of Death, Lorcan), and the classic pay-life **draws** (Crossway
Troublemakers, Erebos Bleak-Hearted).

## The fix

One life-safety check in `SpellAbilityAi.doTrigger`, right beside the existing affordability check —
so it covers every API's optional pay-life trigger in a single place, mirroring the activated path.
Mandatory triggers still fire; costless triggers are untouched. (`UnlessCost$ PayLife` cards such as
Gix use a different mechanism, `willPayUnlessCost`, and are intentionally out of scope.)

## Bringer of the Black Dawn

Its `AI:RemoveDeck:All` flag dates to a 2018 blanket pass and was, in effect, masking this
death-spiral (it paid 2 life to tutor even at 5 life). With the gate in place the card is safe, so
the second commit removes the flag. Its tutor-target selection already works (fetches the best card
when the AI can spare the life).

## Testing

A focused regression test asserts both directions on Bringer's trigger — fires when the AI can spare
the life, declines when paying would drop it below the safety threshold. Full `forge-gui-desktop`
suite stays green (291).

### DIFF
--- a/forge-ai/src/main/java/forge/ai/SpellAbilityAi.java
+++ b/forge-ai/src/main/java/forge/ai/SpellAbilityAi.java
@@ -179,13 +179,13 @@ public abstract class SpellAbilityAi {
             return false;
         }
 
-        // An optional trigger must respect the AI's life safety, not just affordability: the
-        // activated path gates pay-life via willPayCosts/checkLifeCost, but the trigger path only
-        // checked canPayCost above. Without this, any "you may pay N life: <effect>" trigger (e.g.
-        // Bringer of the Black Dawn's tutor, Master of Death's return) would pay life it cannot
-        // spare, even into lethal range. Applies to every API; mandatory triggers still fire.
+        // An optional trigger must respect the AI's cost wisdom, not just affordability: canPayCost
+        // above only checks the cost *can* be paid, while the activated path also runs willPayCosts.
+        // Reuse it here so each API applies its own judgment - e.g. Bringer of the Black Dawn's tutor
+        // won't pay 2 life it cannot spare, and discard/sacrifice/counter costs are weighed too,
+        // instead of a single hardcoded life check. Mandatory triggers still fire.
         if (!mandatory && sa.getPayCosts() != null
-                && !ComputerUtilCost.checkLifeCost(aiPlayer, sa.getPayCosts(), sa.getHostCard(), 4, sa)) {
+                && !willPayCosts(aiPlayer, sa, sa.getPayCosts(), sa.getHostCard())) {
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


forge-ai/src/main/java/forge/ai/SpellAbilityAi.java
```diff
@@ -179,6 +179,16 @@ public final boolean doTrigger(final Player aiPlayer, final SpellAbility sa, fin
             return false;
         }
 
+        // An optional trigger must respect the AI's life safety, not just affordability: the
+        // activated path gates pay-life via willPayCosts/checkLifeCost, but the trigger path only
+        // checked canPayCost above. Without this, any "you may pay N life: <effect>" trigger (e.g.
+        // Bringer of the Black Dawn's tutor, Master of Death's return) would pay life it cannot
+        // spare, even into lethal range. Applies to every API; mandatory triggers still fire.
+        if (!mandatory && sa.getPayCosts() != null
+                && !ComputerUtilCost.checkLifeCost(aiPlayer, sa.getPayCosts(), sa.getHostCard(), 4, sa)) {
```
tool4ever: too broad (all API) and at the same time too specific (only one cost)

should research if you can't just use the already mentioned `willPayCosts` path without side effects

liamiak: Liamiak: Well, you can't beat experience; I should have poked around a bit more on this one.  I'll try to do so next time, and also tried to update the AI framework to prompt it to as well.  Reworked.

Claude: Done. Swapped the hardcoded `checkLifeCost` for `willPayCosts(aiPlayer, sa, sa.getPayCosts(), sa.getHostCard())`, so instead of one blanket life check across every API it dispatches to each API's own override and weighs all cost types (life, discard, sacrifice, counters) — the same judgment the activated path already applies.

On side effects: checked all seven overrides. Six are pure `ComputerUtilCost.check*` calls. The only mutation is `LifeLoseAi`'s `setMaxXValue`, which early-returns unless the cost has a `Count$xPaid` X — so no fixed pay-life trigger like Bringer reaches it — and it's the same call the activated path already makes. The full `forge.ai` suite plus the Bringer life-gate test stay green.

tool4ever (approved):

PR #11388, opened 2026-07-25
AI: let it activate connives that do not target

liamiak:
The AI never activates a connive that doesn't target. `ConniveAi.canPlay` ended with

```java
if (!sa.getTargets().isEmpty() && sa.isTargetNumberValid())
```

An untargeted connive — Hypnotic Grifter, M.O.D.O.K. — has no targets to choose, so the loop
above adds none and that check can never pass. The ability isn't judged unattractive, it is
unreachable. A triggered connive on the same card fires, because `doTriggerNoCost` decides
differently.

The target requirement is now skipped only where the ability does not use targeting at all.
Anything that does target still has to have chosen something, so an "up to one target"
connive is not activated for zero effect.

## How this was found

Chasing a crash report from a Connive discard, since fixed upstream by #11376. While writing
a regression test for that, the AI turned out never to activate the ability at all, so a
gameplay-level test passed against fully broken code. That is the bug left over once #11376
landed.

## Testing

- `mvn -pl forge-gui-desktop -am test`: 288 tests, 0 failures.
- The regression test was verified to fail without this change — the AI never activates the
  ability, so nothing is drawn or discarded — and to pass with it. It also caught the #11376
  crash independently while that was still unfixed here.

The second commit is that test and can be dropped if you would rather not carry it. It plays
a game until the AI connives of its own accord and checks the result, rather than calling the
changed method directly.


forge-ai/src/main/java/forge/ai/ability/ConniveAi.java
```diff
@@ -63,7 +63,13 @@ protected AiAbilityDecision canPlay(Player ai, SpellAbility sa) {
...
             }
         }
-        if (!sa.getTargets().isEmpty() && sa.isTargetNumberValid()) {
+        // No requirement that something was targeted: a connive that does not target at all
+        // (Hypnotic Grifter, M.O.D.O.K.) leaves the loop above with no targets, and demanding
+        // one made those abilities unplayable rather than untargeted. isTargetNumberValid
+        // already answers "are this ability's targeting requirements met", including when
+        // there are none - which is how doTriggerNoCost below decides, so the two paths now
+        // agree on what a valid connive looks like.
+        if (sa.isTargetNumberValid()) {
```
tool4ever: a) this whole PR is too noisy, basically only need a diff for this line for trivial logic
b) change is wrong in case it's "up to one target" - then AI might waste resources for zero effect so need to check `usesTargeting()` instead

liamiak: Both fair — force-pushed.

**(b)** You're right, and it isn't hypothetical: two `DB$ Connive` are `TargetMin$ 0 | TargetMax$ 1`, and one `SP$ Connive` is `TargetMin$ X`, which is the same hole whenever X is 0. Now:

```java
if (!sa.usesTargeting() || (!sa.getTargets().isEmpty() && sa.isTargetNumberValid())) {
```

so anything that targets still has to have chosen something.

**(a)** Comment dropped — the change is that one line. The test is still commit 2 if you'd rather not carry it.

tool4ever (approved):

PR #11389, opened 2026-07-25
LandTraitChanges: small improvements

Hanmac:
Part of #11366

small improvement for LandTraitChanges

sadly Java doesn't let me cast `List<LandTraitChanges>` to both `Iterable<IKeywordsChange>` and `Iterable<ICardTraitChanges>`, so i need `List.copyOf`?

no other changes on the Iterable Lists yet.
Might need to test it first if i can change them into `Stream<>`?

### DIFF
--- a/forge-game/src/main/java/forge/game/card/Card.java
+++ b/forge-game/src/main/java/forge/game/card/Card.java
@@ -3416,6 +3416,9 @@ public class Card extends GameEntity implements Comparable<Card>, IHasSVars, ITr
     }
 
     public boolean hasRemoveIntrinsic() {
+        if (changedCardTypes.isEmpty()) {
+            return false;
+        }
         // only Layer 4 are affected, and it's never intrinsic
         return changedCardTypes.values().stream().anyMatch(ICardChangedType::isRemoveLandTypes);
     }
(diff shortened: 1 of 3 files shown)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -3416,6 +3416,9 @@ public final boolean hasSpellAbility(final int id) {
     }
 
     public boolean hasRemoveIntrinsic() {
+        if (changedCardTypes.isEmpty()) {
+            return false;
+        }
         // only Layer 4 are affected, and it's never intrinsic
         return changedCardTypes.values().stream().anyMatch(ICardChangedType::isRemoveLandTypes);
```
tool4ever, line 3423: AI claimed this line should be optimized too 🤔

Hanmac: But I like the Stream Form better 😅

Hanmac: @liamiak from my tests right now, it isn't much faster to use the for-loop (and its more ugly)

tool4ever:
I also don't understand why it complains without `List.copyOf` 🤔
might try a few things later

Hanmac:
> I also don't understand why it complains without `List.copyOf` 🤔 might try a few things later

See these StackOverflow posts:
https://stackoverflow.com/questions/44374403/cast-from-lista-to-listb-when-b-extends-a
https://stackoverflow.com/questions/2745265/is-listdog-a-subclass-of-listanimal-why-are-java-generics-not-implicitly-po

The problem might not be, if I changed `getChangedCardKeywordsList` and `getChangedCardTraitsList` into `Stream<>`.
I might try a few things later too

tool4ever (approved):
how about this? 🤔

PR #11390, opened 2026-07-25
Let the AI use Hex Parasite

liamiak:
Hex Parasite is flagged `AI:RemoveDeck:All`. Unusable due to one gate:

## The gate

`CountersRemoveAi`'s `CounterType$ Any` branch made four targets unreachable — persist, undying, stripping opposing creatures, and "take good counter(s) off opponent" — due to the X cost amount. Hex Parasite is the only card affected: three others pair `CounterType$ Any` with an `X`, but only this one defines it as `Count$xPaid`.

Lifting the gate cannot change a fixed-amount card: the block ran exactly when `xPay` was false, so the only case added is `xPay` true. An X cost also needs to pay for the counters its target has, not the maximum `setMaxXValue` left on the ability. `takeTarget` does both, and all six branches now use it — Dark Depths and planeswalker reduce to it, each bounded by `amount`.

## Measured

`CountersRemoveAi`'s decision, six lands available, against master:

| Board | Before | After |
|---|---|---|
| Opposing creature, 3x `+1/+1` | `TargetingFailed` | `WillPlay`, X=3 |
| Own persist creature, 1x `-1/-1` | `TargetingFailed` | `WillPlay`, X=1 |
| Own Dark Depths, 3x ice | `WillPlay`, X=3 | unchanged |
| Nothing with counters | declines | declines |

Played out to the board rather than stopping at the decision: the Depths break leaves Marit Lage, and the planeswalker case takes an opponent's Jace from 3 loyalty to dead.

## Two misses the new reach exposed

**The fallback commits to one card.** It takes `getBestAI`'s single choice, then walks that card's counter types for a non-negative one; a creature carrying only a `-1/-1` counter ends the branch, with a Chalice of the Void sitting untouched beside it. It now drops that candidate and tries the next. This is the one part that reaches fixed-amount cards, and only where the branch previously did nothing at all.

**`{B/P}` is weighed against nothing.** It is paid with 2 life whenever no black source is available, and the AI would go from 3 life to 1 to strip a single counter. Phyrexian mana is a life cost, so it now takes the floor the AI already keeps for those: `remainingLife` 4, the value every `checkLifeCost` caller passes, overridable with the existing `AILifeThreshold`. Only abilities with no `AIPhyrexianPayment` param are affected, and all five cards that set it keep their behaviour. Hex Parasite declines at 5 life, pays at 20 to kill a planeswalker, and never pays while it can still make black.

## Testing

376 tests, 0 failures, 6 skipped, playing the phase out through `AiController`. Removing `CountersRemoveAi`'s change, the `AI:RemoveDeck:All` line, the fallback loop, or the life floor each fails a test.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
@@ -728,6 +728,13 @@ public class ComputerUtilMana {
                             break; // no one to finish with the gut shot
                         }
                     }
+                } else {
+                    // otherwise treat it as the life cost it is, with the floor checkLifeCost uses
+                    int remainingLife = sa.hasParam("AILifeThreshold")
+                            ? Integer.parseInt(sa.getParam("AILifeThreshold")) : 4;
+                    if (ai.getLife() - phyLifeToPay < remainingLife && !ai.cantLoseForZeroOrLessLife()) {
+                        break;
+                    }
                 }
 
                 if (toPay.isPhyrexian()) {
--- a/forge-ai/src/main/java/forge/ai/ability/CountersRemoveAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/CountersRemoveAi.java
@@ -153,11 +153,7 @@ public class CountersRemoveAi extends SpellAbilityAi {
                     Card depth = depthsList.getFirst();
                     int ice = depth.getCounters(iceType);
                     if (amount >= ice) {
-                        sa.getTargets().add(depth);
-                        if (xPay) {
-                            sa.setXManaCostPaid(ice);
-                        }
-                        return new AiAbilityDecision(100, AiPlayDecision.WillPlay);
+                        return takeTarget(sa, depth, xPay, amount, ice);
                     }
                 }
             }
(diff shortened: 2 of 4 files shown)


liamiak:
Rebased onto master and simplified. The add-target / set-X / accept step was repeated in all four branches and is now one `takeTarget` helper, which also moved out from between `checkApiLogic` and the javadoc that documents it. The `Math.max(1, ...)` around the X came out: `setMaxXValue` returning 0 already bails with `CantAffordX`, and every branch requires the target to hold at least one of the counter it reads, so it could never bind.

+162/-42 down to +148/-42. The measured table in the description is now taken against master rather than partly inferred, and the four rules were additionally diffed with the amount forced fixed, to cover the path no card in the pool reaches by probe.

liamiak:
Force-pushed on 25 Aug with no note — apologies for the silence. Rebased onto today's master; 376 tests, 0 failures, 6 skipped.

Two changes since, both from running the newly reachable branches out to the board rather than stopping at the decision.

The fallback branch takes `getBestAI`'s single choice and walks that one card's counter types for a non-negative one. If it has none — a creature carrying only a `-1/-1` counter — the branch falls through and nothing is activated, with a Chalice of the Void carrying two charge counters sitting next to it. It now drops that candidate and tries the next. This is the one part that reaches fixed-amount `CounterType$ Any` cards as well, though only where the branch previously did nothing.

The other replaces the `AIPhyrexianPayment$ Never` I had put on the card. `Never` fixed the "3 life down to 1 to strip one counter" case by never paying at all, which also gave up breaking its own Dark Depths and finishing a planeswalker. Phyrexian mana is a life cost, so `ComputerUtilMana` now applies the floor the AI already keeps for those — `remainingLife` 4, the value every `checkLifeCost` caller passes, with `AILifeThreshold` as the per-ability override. It sits in an `else`, so the five cards that set `AIPhyrexianPayment` are untouched. Hex Parasite declines at 5 life and pays 2 at 20 to kill a planeswalker; measured, it never pays while it can still make black.

Both are separate commits and droppable alone.

PR #11387, opened 2026-07-25
Adventure: reorganize settings scene into tabs with grouped sections

liminalwarmth:
Resolves #11386.

Reorganizes the Adventure settings scene from a single ~40-row scrolling list into four tabs with labeled section headers, so every option sits under a heading that says what it affects.

I went through all the options and tried to reorganize them into logical headers for Adventure mode so the options menu in Adventure is easier to navigate and read. Aside from the reorganization, the new tabs, and one checkbox being flipped to positive phrasing (item cracking, see below), no functional changes are included in this PR.

Authored with Claude Fable and Sol 5.6, categories and naming consistency manually checked and tweaked before submission, and I manually tested on macOS desktop with multiple rounds of code reviews until no issues remain flagged by the agents.

Screenshot 2026-07-25 at 1 05 55 PM

## Tabs and sections

- **General** — General Settings (World, Landscape Mode on Android) · Sound Options (Music Volume, Sound Volume) · Card Pool Options (Alchemy Variants, Card Variants from All Sets, Unknown Cards, Non-Legal Cards, Archetype Decks)
- **Display** — Graphic Options (Video Mode, Fullscreen, Day/Night Background Image, Battlefield Texture Filtering, Auto Cache Size, Dispose Textures) · Card Display (Border Mask, Card Animations, Auto-tap Highlight, Era-Matched Token Art, Download Missing Card Art) · Shop Display (Reward/Shop Card Display Ratio, Tooltip Ratio, Shop Item Names, Not-For-Sale Overlay) · Card Overlays (Card Name, Mana Cost, Perpetual Mana Cost, Power/Toughness, Card ID, Ability Icon)
- **Gameplay** — Adventure Options (Item Cracking, Equipment Loadouts, Hidden Enemy Tracking) · Match Options (Win/Loss Overlay, Ante ×3) · Auto Selling (Auto-Sell prompt, Commander variants)
- **Input** — Vibration Options (intensity + the five vibrate-on-event toggles)

The guiding rules: everything that affects what's on screen lives on the Display tab (including Video Mode and Fullscreen), everything that shapes the card pool lives in one Card Pool Options section on the General tab, and the auto-sell pair gets its own small section on Gameplay rather than hiding among adventure-mode toggles.

## Implementation notes

- Every existing setting keeps its exact widget, preference wiring, and platform gating — they are only regrouped. (The recent Android un-gating of Unknown/Non-Legal Cards from #11364 is preserved.)
[...]

PR #11392, opened 2026-07-25
Add rotating Adventure battle backgrounds

vanja-ivancevic:
- Rotates through any number of JPG, JPEG, or PNG battle backgrounds in a folder and avoids immediate repeats when alternatives are available.
- Uses one HTTPS URL in `<plane>/skin/battle-backgrounds.txt`. That URL points to a plain-text index whose non-comment lines are relative image paths.
- Maps every indexed path in two matching places. For example, `enemy/lich/lich_01.jpg` is fetched relative to the index URL and stored under `<plane cache>/skin/battle_backgrounds/enemy/lich/lich_01.jpg`.
- Downloads only missing images when the Adventure plane starts, removes cached images no longer listed, and uses the last valid cached index during temporary host failures.
- Generates stable path-safe folders from canonical point-of-interest names, enemy names, and TMX map paths, so publishing new art normally requires only adding files and index entries on the host.
- Checks folders in this exact priority order, stopping at the first folder containing images: (1) TMX enemy `battleBackground`, (2) `encounter/<poi>/<enemy>`, (3) enemy archetype `battleBackground`, (4) `enemy/<enemy>`, (5) TMX map `battleBackground`, (6) `map/<poi>/<map>`, (7) `map/<map>`, (8) point-of-interest `battleBackground`, (9) `poi/<poi>`, (10) `<category>/<biome>`, and (11) `<category>`.
- For each candidate folder, checks the plane cache, local plane files, then common Adventure files. If no rotation folder contains an image, the existing plane, common, current-skin, and default-skin single-image fallbacks remain unchanged.
- Keeps Old Border artwork on the Old Border Shandalar website instead of adding it to the Forge distribution, while leaving the same system available to other Adventure planes and mods.


tool4ever (commented):
I like the idea, though this adds yet another ~20MB of extra assets to download for every user

Since these seem rather optional imo it's time to think about how they can be downloaded manually like some of the other parts

vanja-ivancevic:
I agree. If other modders want to add backgrounds, the bloat will continue
to increase. Additionally, this is 20MB of data that vanilla Shandalar
players won't even use.

One approach could be making it optional in the settings via an
auto-download feature, handling backgrounds similarly to how we handle card
art, like hosting them separately for users to download.

Personally, I would like to have them activated by default for Old Border
Shandalar, as it enhances the experience. Perhaps when a user switches to
and loads that specific adventure mode, the assets could download in the
background. Most users likely wouldn't mind an extra 20MB of cached art,
especially since we already cache card art this way.

Maybe we can have the game download all images in given folders on a
server, that way modders can add/remove bgs without updates too…

Do you have any preferences?

tool4ever:
Well some of the lists for the ingame downloaders already point at your domain anyway
so it might make sense to supply them from there
not sure if they can be called via adventure mode but ideally each plane would have its own list with links for its optional assets 🤔
depending on which location you choose some overlap from #11387 is possible

liminalwarmth:
This is cool, and I would definitely use this feature. If you did package it into the new tabs interface I'm proposing, display would be a logical place to stick it, since there are other settings there that are similar, as tool4forever pointed out. 

The tabs definitely become more useful as more GUI options get added, as I'm finding with the additional enhancements I'm doing on local that I haven't pushed upstream until there's more discussion about some of the existing pending changes.

More of an advanced feature, but it would also be neat if plane authors could assign (and distribute) background overrides for specific combat encounters or creature archetypes (if this doesn't already exist).

vanja-ivancevic:
1. I really like your suggestion of tabbing the settings. We have way too many settings right now in one list.
2. I was thinking about have these custom BGs just as a default setting for the plane's modder to design as they want, with fallbacks in place, as they are sort of a core visual game experience of plane. I don't fully see the need of enabling or disabling more backgrounds. As such, I would suggest to keep it out of settings for now.
3. I 100% agree with your suggestion to have custom BGs for certain POIs and even enemies. I will look into what the best way to do this would be.

forge-gui-mobile/src/forge/assets/FSkinTexture.java
```diff
@@ -130,18 +143,123 @@ private boolean isAdventureBackground() {
...
             return null;
         }
 
+        String adventureCacheDirectory = GuiBase.getAdventureCacheDirectory();
+        if (adventureCacheDirectory != null && !adventureCacheDirectory.isEmpty()) {
+            FileHandle cachedFile = Gdx.files.absolute(adventureCacheDirectory + ForgeConstants.SKIN_DIR + filename);
+            if (cachedFile.exists()) {
+                return cachedFile;
+            }
+        }
```
Jetz72, line 306: @shoeless Will this work with the new iOS support? I saw in that PR that several instances of `Gdx.files.absolute` got swapped out.

Jetz72 (commented):
Still iffy about third party content URLs for assets being included in the main repo. It's not unprecedented, but it's also not unprecedented that such a site proves unsustainable. Ideally we could someday shift to a system where entire Adventure planes can easily be downloaded and installed as complete packages, and then there's no risk of the user getting an incomplete experience. But for now I think this implementation is alright, pending confirmation on iOS handling.

forge-gui-mobile/src/forge/assets/FSkinTexture.java
vanja-ivancevic: I double-checked this with AI-assisted research against Apple’s filesystem guidance, LibGDX 1.13.5, and Forge’s existing iOS handling. `absolute() `is supported on iOS and is appropriate for this downloaded cache path, which is outside the read-only app bundle.
The latest revision routes all three lookups through Forge’s existing `Assets.getFileHandle()` helper. Bundled `ASSETS_DIR` resources therefore use `internal() `on iOS, while writable cache paths remain absolute. The iOS compatibility audit also passes. Thanks for flagging this.

kevlahnota:
I'm fine with these but maybe a configurable option in the settings, a user can enable this and input any repository for the image source they want.

vanja-ivancevic:
> I'm fine with these but maybe a configurable option in the settings, a user can enable this and input any repository for the image source they want.

We can add an “Enable Extra Battle Backgrounds” setting, enabled by default. When enabled, users can choose either the active plane’s default source or a custom HTTPS background index URL. Disabling it would stop remote syncing and use Forge’s original battle backgrounds and fallbacks. Previously downloaded files would remain cached but would be ignored until the option is enabled again. What do you think?

kevlahnota:
> > I'm fine with these but maybe a configurable option in the settings, a user can enable this and input any repository for the image source they want.
> 
> We can add an “Enable Extra Battle Backgrounds” setting, enabled by default. When enabled, users can choose either the active plane’s default source or a custom HTTPS background index URL. Disabling it would stop remote syncing and use Forge’s original battle backgrounds and fallbacks. Previously downloaded files would remain cached but would be ignored until the option is enabled again. What do you think?

As long as its configurable and defaults to original battle background by default since most android user device is not modern enough to handle lots of big textures. Anyway does old savefile works without the definition of battlebackground data additions?

vanja-ivancevic:
That makes sense for mobile devices. Would a platform-specific default be acceptable?

On Android and iOS, **Extra Battle Backgrounds** would default to disabled, preserving the original Forge backgrounds. On desktop, it would default to enabled with **Plane Default** selected. For planes such as Shandalar Old Border, the supplied backgrounds are intended to be part of the plane's experience, and most players are unlikely to discover and enable a new setting themselves.

The option would remain configurable on every platform, including support for a custom HTTPS index URL. The downloaded files are cached, but they are not all loaded as textures at once. Forge only loads the selected battle background for rendering.

I also checked old-save compatibility. The settings themselves are separate from save files, and absent enemy, map, encounter, or background definitions safely use the existing fallbacks. I did find that `PointOfInterestData` can be serialized inside saved quests, so I preserved its previous `serialVersionUID`. I verified this by loading an old serialized quest containing a target POI; it restored successfully, with the new `battleBackground` field defaulting to `null`.

PR #11393, opened 2026-07-25
Let the AI use Thespian's Stage

liamiak:
Thespian's Stage (`{2}, {T}`: becomes a copy of target land, keeping this ability) is `AI:RemoveDeck:All`, and `CloneAi` says why — its default target branch is `return false` under *"Those can just use AI:RemoveDeck:All until this can do a reasonably good job of picking a good target."*

## Picking a land

New `AILogic$ CloneBestLand` takes the highest `evaluateLand` score that beats what the Stage already is (156).

A land is scored under whoever controls it *now*, while the copy always arrives under ours, so reading an opponent's board raw misjudges in both directions: Cabal Coffers is 6 for them and 306 for us, and Gaea's Cradle behind their creatures is 706 for them and 6 for us. So an opponent's land is judged on an LKI copy moved to our side, the same shape `ComputerUtilCard.filterOutFatalCopies` now uses for the creature case. Copying their land is much of why this card is good.

That replaces the `getBestLandToPlayAI` ranking from the previous push, which scores each land under its current controller.

The only land skipped is a legendary we already control, since copying that just means sacrificing one of the two.

The bar is deliberately conservative: the copy keeps the clone ability, so a target only has to beat a plain land. It passes on marginal upgrades like Wasteland (116) rather than risk a bad one.

The copy gets a statics pass so our own effects reach it: under our Prismatic Omen an opponent's Swamp is worth 118 rather than 106. It does not strip *their* statics from the copy, which is a gap this shares with the pattern it follows.

Timing is `AtOppEOT`, unchanged, for the reasons in the comment above. Removing the flag also lets a colourless-only land into AI deckbuilding, a fair trade.

## Testing

`mvn -pl forge-gui-desktop -am test`: 364 tests, 0 failures.

Four tests, each pinned by a mutation only it catches — the phase gate by `waitsForTheTurnBeforeOurs`, the legendary skip by `copiesTheBetterLandItControls`, the score bar and the re-control by `ignoresALandThatOnlyScoresOnTheirBoard`, and including opponents' lands at all by `copiesAnOpponentsLandToo`.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
@@ -810,10 +810,6 @@ public class ComputerUtilCard {
         return eval;
     }
 
-    public static int evaluateLand(final Card c) {
-        return landEvaluator.apply(c);
-    }
-
     public static int evaluatePermanentList(final CardCollectionView list) {
         int value = 0;
         for (int i = 0; i < list.size(); i++) {
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
@@ -810,6 +810,10 @@ public class ComputerUtilCard {
         return eval;
     }
 
+    public static int evaluateLand(final Card c) {
+        return landEvaluator.apply(c);
+    }
+
     public static int evaluatePermanentList(final CardCollectionView list) {
         int value = 0;
         for (int i = 0; i < list.size(); i++) {
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCard.java
@@ -784,6 +784,10 @@ public class ComputerUtilCard {
         return eval;
     }
 
+    public static int evaluateLand(final Card c) {
+        return landEvaluator.apply(c);
+    }
+
     public static int evaluatePermanentList(final CardCollectionView list) {
         int value = 0;
         for (int i = 0; i < list.size(); i++) {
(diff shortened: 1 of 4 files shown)


liamiak:
Force-pushed a timing change: the Stage is now taken at the end of the turn before ours, rather than in our own main phase.

Two reasons. The upgrade is permanent, so there is nothing to gain by taking it early, and holding the mana until then leaves a possible threat open. The second one I had missed: the cost taps the Stage, so activating in our own main phase costs us the copied land's mana for that whole turn, while at the opponent's end step it unlocks again in our untap step.

That is exactly the base class's `AtOppEOT` rule, so the override delegates to it rather than restating the condition. Added a third test asserting the AI declines during its own main phase.

liamiak:
Rebased onto master and simplified. The land ranking now goes through `ComputerUtilCard.getBestLandToPlayAI`, which is the same `evaluateLand` ranking and had no callers, rather than repeating it here — so this branch reads like the `CloneBestCreature` one above it, and `forge.ai.simulation` stays out of the ability AIs.

Verified as a no-op: 0 divergences over every land and pair from a 15-land pool under both controllers. The three tests are unchanged in number but now cover the controller and legendary filters too, which mutation testing showed were previously untested.

liamiak:
Reworked the target selection: opponents' lands are scored now rather than skipped, since copying their land is much of why the Stage is good.

I had excluded them because `evaluateLand` reads a land under whoever controls it now — their Cabal Coffers scores 6, their Gaea's Cradle 706, and neither is what we would actually get. That is fixable rather than a reason to look away: an opponent's land is judged on an LKI copy moved to our side, the same shape `ComputerUtilCard.filterOutFatalCopies` now uses for the creature case. Coffers 6 → 306, Cradle 706 → 6.

That undoes the `getBestLandToPlayAI` reuse from the last push — it ranks each land under its current controller, which is the thing being corrected — so the ranking is a small loop instead.

The statics pass on the copy made no measured difference across a twelve-land pool; it is there to match the existing pattern. Tests are four rather than five, after mutation testing showed one of them was redundant.

liamiak:
Rebased. The creature side of this landed separately as #11675, so that half is gone and what is left is just the land case: `CloneBestLand` picks the land the Stage is best off becoming, scoring an opponent's land as a copy under our control.

Also dropped a `ComputerUtilCard.evaluateLand` wrapper I had added - `GameStateEvaluator.evaluateLand` is already public and is what it called - so this no longer touches `ComputerUtilCard` at all. Three files now.

Tests unchanged, 4/4.

PR #11394, opened 2026-07-25
Only interrupt auto-pass on targeting when an opponent is the caster

autumnmyst:
The "interrupt when targeted" yield option currently fires for any spell or ability that targets you or your permanents, including your own. Breaking a yield for something you just cast yourself, which isn't very helpful since you are actively doing it yourself.

This scopes that interrupt to opponents' spells and abilities, matching the other opponent-driven interrupts in `onSpellAbilityCast`.


tool4ever:
Worth noting that this isn't limited to spells, maybe a more dynamic combination system would be interesting in the future

PR #11396, opened 2026-07-26
Let the AI use Venser, the Sojourner (and two other mass-unblockable cards)

liamiak:
Venser, the Sojourner is `AI:RemoveDeck:All`. Two of his three abilities already work — the flag was covering for one dead one, and that one is not his alone.

## What was dead

The **-1** ("Creatures can't be blocked this turn") returned `CantPlayAi` on every board, including ones where it wins on the spot — `EffectAi`'s no-`AILogic` branch is a bare `return CantPlayAi`, so nothing here narrows existing behaviour.

`AILogic$ UnblockableForLethal` holds it until attacking with everything that can attack would take an opponent to zero with nothing able to block — the state the ability itself creates, so blockers are not counted.

| AI power | Opponent life | Before | After |
|---|---|---|---|
| 18 | 18 | `CantPlayAi` | activates |
| 18 | 19 | `CantPlayAi` | declines |
| 18 | 18, opponent holds a Fog | `CantPlayAi` | declines |

Same tag on the other two cards where this is an optional choice, and both argue for it harder than Venser: Jace, Arcane Strategist spends seven loyalty, Tezzeret's Gatebreaker sacrifices itself. The rest are mandatory triggers, static, or wrong for a lethal-only gate — Glaring Spotlight also grants hexproof, and Reverse the Polarity is a Charm mode CharmAi reaches another way.

## The check

It lives in `ComputerUtilCombat` beside `sumDamageIfUnblocked`, with the guards `doAssault()` already has: `ai.cantWin()`, `opp.cantLoseForZeroOrLessLife()` and `hasAFogEffect`, fog last since it walks several zones.

## Testing

368 tests, 0 failures, 6 skipped. The tests play the phase out through `AiController` rather than calling the ability AI directly, and the lethal case runs to end of combat — the opponent is dead and all three attackers survive, so it asserts the ability won the game rather than that a loyalty counter moved. Removing either half of the fix fails it.

## Not included: War Cadence

`RestrictBlocking` asks a similar question with its own loop, and an earlier revision of this PR pointed it at the same helper. Measured end to end, the AI pays `{X}{R}` at declare-attackers and then declares no attackers, so the tax buys nothing. What it would need is upstream of the ability: `AiAttackController` plans the attack in the loop commented `// presumes the Human will block`, which already carries `// TODO should be limited to how much getBlockCost the opp can pay`. Until the attacker can price blocks, that branch can only spend mana, so `RestrictBlocking` is left untouched here.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
@@ -251,6 +251,32 @@ public class ComputerUtilCombat {
         return sum;
     }
 
+    /**
+     * Whether swinging with everything that can attack would kill this opponent outright if none
+     * of it could be blocked.
+     */
+    public static boolean unblockedAttackIsLethal(final Player ai, final Player opp, final Iterable<Card> candidates) {
+        if (ai.cantWin() || opp.cantLoseForZeroOrLessLife()) {
+            return false;
+        }
+        final CardCollection attackers = CardLists.filter(candidates, c -> CombatUtil.canAttack(c, opp));
+        if (attackers.isEmpty() || sumDamageIfUnblocked(attackers, opp) < opp.getLife()) {
+            return false;
+        }
+        // last, because it walks the battlefield, external zones and card memory
+        return !ComputerUtil.hasAFogEffect(opp, ai, true);
+    }
+
+    /** As above, against any one of the AI's opponents. */
+    public static boolean unblockedAttackIsLethal(final Player ai) {
+        for (final Player opp : ai.getOpponents()) {
+            if (unblockedAttackIsLethal(ai, opp, ai.getCreaturesInPlay())) {
+                return true;
+            }
+        }
+        return false;
+    }
+
     // Returns the number of poison counters unblocked attackers would deal
     /**
      * <p>
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilCombat.java
@@ -251,6 +251,34 @@ public class ComputerUtilCombat {
         return sum;
     }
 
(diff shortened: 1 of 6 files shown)


liamiak:
Rebased onto master and cut this down. The lethal check now lives in `ComputerUtilCombat` next to `sumDamageIfUnblocked` rather than as a private method here, and picks up the two guards `AiAttackController.doAssault()` already has — `ai.cantWin()` and an opponent Fog — so the AI no longer spends Venser's loyalty, or sacrifices Tezzeret's Gatebreaker, into a Fog.

+98/-4, of which 61 is the test. Description updated to match.

liamiak:
Force-pushed on 25 Aug with no note — apologies for the silence. Rebased onto today's master, and the PR is smaller than it was.

The three-card change behaves the same, but the test proves more: `goesUnblockableForALethalSwing` now plays through to MAIN2 and asserts the opponent is dead and all three attackers survived, rather than that a loyalty counter moved. Checked in a three-player game too — the targeted opponent dies, the third player is untouched. 368 tests, 0 failures, 6 skipped.

I've dropped the War Cadence commit that went up in that push. Measured: the AI pays `{X}{R}` for the tax at declare-attackers and then declares zero attackers, and the same board with War Cadence removed gives the same zero attackers and the same opponent at 18 life — so the ability changed nothing but the mana. Inferred, not instrumented: the reason is `AiAttackController` planning that attack in the loop commented "presumes the Human will block", where `// TODO should be limited to how much getBlockCost the opp can pay` already sits — the tax is invisible to the decision it was bought for. Pointing the branch at the declared attackers instead stops the waste but leaves it permanently inert, so `RestrictBlocking` is untouched and identical to master here.

That leaves the block-cost TODO as the real piece of work. Happy to take it separately if it's wanted.

PR #11400, opened 2026-07-26
Display edition code in rewards screen/Oracle view

Eradev:
Display the edition code instead of the anvil when the card has no picture in the Rewards screen, and Oracle view. Basically replicates what was being shown in the deck editor.

image

### DIFF
--- a/forge-gui-mobile/src/forge/card/CardImageRenderer.java
+++ b/forge-gui-mobile/src/forge/card/CardImageRenderer.java
@@ -115,6 +115,10 @@ public class CardImageRenderer {
     }
 
     public static void drawCardImage(Graphics g, CardView card, boolean altState, float x, float y, float w, float h, CardStackPosition pos, boolean useCardBGTexture, boolean noText, boolean isChoiceList, boolean showArtist, boolean showArtBox) {
+        drawCardImage(g, card, altState, x, y, w, h, pos, useCardBGTexture, noText, isChoiceList, showArtist, showArtBox, false);
+    }
+
+    public static void drawCardImage(Graphics g, CardView card, boolean altState, float x, float y, float w, float h, CardStackPosition pos, boolean useCardBGTexture, boolean noText, boolean isChoiceList, boolean showArtist, boolean showArtBox, boolean useEditionLabel) {
         updateStaticFields(w, h);
 
         float blackBorderThickness = w * BLACK_BORDER_THICKNESS_RATIO;
@@ -226,7 +230,7 @@ public class CardImageRenderer {
             y += textBoxHeight;
 
             //draw type line
-            drawTypeLine(g, state, canShow, headerColors, x, y, w, typeBoxHeight, noText, false, false);
+            drawTypeLine(g, state, canShow, headerColors, x, y, w, typeBoxHeight, noText, false, false, useEditionLabel || !showArtBox);
             y += typeBoxHeight;
         } else if (isClass) {
             //draw text box
@@ -235,7 +239,7 @@ public class CardImageRenderer {
             y += textBoxHeight;
 
             //draw type line
-            drawTypeLine(g, state, canShow, headerColors, x, y, w, typeBoxHeight, noText, false, false);
+            drawTypeLine(g, state, canShow, headerColors, x, y, w, typeBoxHeight, noText, false, false, useEditionLabel || !showArtBox);
             y += typeBoxHeight;
         } else if (isDungeon) {
             if (!drawDungeon) {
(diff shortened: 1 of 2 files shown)


jamincollins:
## Findings
#### 1. Edition label shown for face-down cards using UNKNOWN_CODE
The drawTypeLine method now passes `CardEdition.UNKNOWN_CODE` when `canShow` is false, resulting in the literal "???" displayed on face-down cards in reward view. The reward rendering path always uses `useEditionLabel=true`, so face-down cards will show "???" instead of hiding the set area.
Test case:
```java
CardStateView state = new CardStateView();
state.setSetCode("LEA");
boolean canShow = false;
String set = canShow ? state.getSetCode() : CardEdition.UNKNOWN_CODE;
// set == "???"
```
Impact: Visual artifact on face-down reward cards; Oracle view may show misleading set code.

#### 2. Null set code with useEditionLabel true leads to empty width adjustment
When `useEditionLabel` is true and `state.getSetCode()` returns null, the code checks `!StringUtils.isEmpty(set)` but does not guard against null before width calculation. StringUtils.isEmpty handles null, so no NPE, but width reduction is skipped while the rarity icon path would have still reserved space. Inconsistent layout width between label and icon modes.
Test case:
```java
String set = null;
if (!StringUtils.isEmpty(set)) { // false, no drawSetLabel
    // w not reduced
}
// w remains full, type text may overflow
```
Impact: Layout jitter when set code is missing; type line may clip.

#### 3. drawCardImage overload chain forces useEditionLabel false for existing callers
Existing callers of the 5-arg drawCardImage overload are redirected to the new 6-arg version with `useEditionLabel` defaulting to false. However, the 5-arg overload with 7 parameters (showArtist, showArtBox) previously defaulted to false. The change is compatible but the `useEditionLabel || !showArtBox` condition in drawTypeLine means that when `showArtBox` is false, edition label is forced on even for non-reward paths, potentially changing rendering for Oracle view and other UI that relied on rarity icons.
Test case:
```java
drawCardImage(g, card, false, x, y, w, h, pos, true, false, false, true, false);
// showArtBox=false -> useEditionLabel || !showArtBox == true
// edition label rendered where rarity icon expected
```
Impact: Backwards compatibility break for callers expecting rarity icons when showArtBox is false.

#### 4. RewardActor initialization render uses edition label unnecessarily
The initialization render at line 761 draws off-screen with `useEditionLabel=true`. This forces edition label initialization even when the final render might not need it, increasing first-render cost. More importantly, the off-screen render uses reward.getCard() but passes reward.getCard() again, not the card parameter, leading to potential mismatch if reward card differs from card parameter.
Test case:
```java
// renderPlaceholder(g, card, alternate, displayArt)
// init render uses reward.getCard(), not card
CardImageRenderer.drawCardImage(g, CardView.getCardForUi(reward.getCard()), false, ...)
```
Impact: Initialization may render wrong card; unnecessary edition label computation.

### Performance and security
#### 5. getSetWidth called per frame for edition label without caching
When `useEditionLabel` is true, `CardRenderer.getSetWidth(TYPE_FONT, set)` is called every frame to compute label width. Font metric lookup is cheap but repeated for every card in reward grid. No caching of set width per set code. Could cause minor frame drops with many rewards.
Test case:
```java
for each reward card:
    float setWidth = CardRenderer.getSetWidth(TYPE_FONT, set); // repeated
```
Impact: Performance degradation in large reward screens.

## Verdict
Request Changes
The PR introduces edition code display in rewards/Oracle view but contains several logic and compatibility issues: face-down cards show "???", layout width inconsistency with null set codes, forced edition labels when showArtBox is false altering existing UI, potential initialization mismatch in RewardActor, and per-frame set width calculation. These need addressing before merge.

## Recommended Fixes
1. Guard edition label rendering behind canShow && !faceDown, skip drawing and width reduction for face-down cards to avoid "???" artifact.
2. Ensure width adjustment is symmetric between edition label and rarity icon paths; reserve space for label even when set is empty to prevent layout shift.
3. Change `useEditionLabel || !showArtBox` to explicit `useEditionLabel` to preserve existing rarity icon behavior when showArtBox is false.
4. Use the `card` parameter for initialization render in RewardActor.renderPlaceholder, not reward.getCard().
5. Cache set width per set code/font size to avoid repeated metric calculation.

---

kevlahnota (dismissed):

forge-gui-mobile/src/forge/card/CardImageRenderer.java
```diff
@@ -515,24 +519,39 @@ private static void drawTypeLine(Graphics g, CardStateView state, boolean canSho
...
+                    g.drawImage(FSkinImage.SET_SPECIAL, x + w + iconPadding, y + (h - iconSize) / 2, iconSize, iconSize);
+                } else if (state.getRarity() == CardRarity.MythicRare) {
+                    g.drawImage(FSkinImage.SET_MYTHIC, x + w + iconPadding, y + (h - iconSize) / 2, iconSize, iconSize);
+                } else if (state.getRarity() == CardRarity.Rare) {
+                    g.drawImage(FSkinImage.SET_RARE, x + w + iconPadding, y + (h - iconSize) / 2, iconSize, iconSize);
+                } else if (state.getRarity() == CardRarity.Uncommon) {
+                    g.drawImage(FSkinImage.SET_UNCOMMON, x + w + iconPadding, y + (h - iconSize) / 2, iconSize, iconSize);
+                } else {
+                    g.drawImage(FSkinImage.SET_COMMON, x + w + iconPadding, y + (h - iconSize) / 2, iconSize, iconSize);
+                }
```
Hanmac: Can this `state.getRarity` to FSkinImage be simplified via `image = match(state.getRarity())` ?

Hanmac (approved):

PR #11402, opened 2026-07-26
Only report a library search when the searcher owns the library

### LINKED ISSUE #10626: Opposition Agent Activates when Searching an Opponent's Library
Tested on Secrets of Strixhaven Stable on Desktop

Certain cards (Acquire, Eternal Dominion, Inevitable Betrayal, even Surgical Extraction) allow the active player to search an opponent's library. When doing this, the active player's Opposition Agent will incorrectly activate, exiling the card to its effect, rather than completing the effect that triggered the search. (Save for Surgical Extraction, which successfully re-exiles the cards after)

This also triggers when an opponent has an Opposition Agent and you search another player's library, despite Opposition Agent keying off opponents searching 'their' libraries.

liamiak:
Fixes #10626.

## The bug

Opposition Agent reads "While an opponent is searching **their** library, they exile each card they
find." The replacement was gated on a flag set for *any* library search — no matter whose library,
or who was looking.

`changeHiddenOriginResolve` already separates the two: `decider` does the looking, `player` owns the
library. The other half of the same card, `ControlOpponentsSearchingLibrary`, is already gated on
`decider.equals(player)`, and so is `incLibrarySearched`. This was the one place that ignored it.

## Naming

`TriggerSearchedLibrary` already has this exact predicate, spelled `SearchOwnLibrary$`, and all four
"whenever an opponent searches their library" cards use it. `AbilityKey.FoundSearchingLibrary` had
one reader and one card script, both wanting the own-library meaning, so it is renamed to match
rather than joined by a near-duplicate — no second parameter, no new key. Worth flagging: any
out-of-tree card script using `FoundSearchingLibrary$` would need the new spelling.

## Turtle Tracks

"Any number of target players may each search their library" — the targets search, but the script
named no fetcher, so the effect fell back to the caster and then retargeted. Four defects on master,
all reproduced:

- NPE at `ChangeZoneEffect:924` when it resolves with no targets, which `TargetMin$ 0` allows
- only the first target ever searches
- the caster picks the land instead of the target
- "whenever an opponent searches their library" never fires — Archivist of Oghma stays silent where
  Diabolic Tutor on the same board triggers it

`DefinedPlayer$ Targeted` fixes all four. It is bundled here because without it this change would
regress the card.

## Coverage

Every `Chooser$` in cardsfolder: 278 uses, 240 on ChangeZone, 65 with a Library origin, plus 165
`Origin$ All` lines (none carrying a `Chooser$` or a player target). Seven of the 65 carry
`NoLooking$ True` — Ecological Appreciation among them — so `searchedLibrary` never becomes true and
the new condition is never reached; I ran the card, outcome unchanged. Guided Passage is the one
script naming a chooser who does not search; master already exiles all three cards to an opposing
Agent, this sends them to hand.

The condition only ANDs onto the old one, so it can turn the flag off but never on.

## Testing

Five cases, one per branch. `mvn -pl forge-gui-desktop -am test`: 358 tests, 0 failures, 0 checkstyle
violations. Fails on master with `expected:<Battlefield> but was:<Exile>`.


forge-game/src/main/java/forge/game/ability/effects/ChangeZoneEffect.java
```diff
@@ -917,6 +917,7 @@ private void changeHiddenOriginResolve(final SpellAbility sa) {
                 decider = player;
             }
 
+            final Player fetcher = player;
```
tool4ever: that seems like limiting the fix for no reason...why not simply check against `decider`?

liamiak: You're right, and it goes further than limiting the fix — thanks, this was the better thread to pull.

`decider` is already the engine's notion of the searching player: it gates `LimitSearchLibrary`,
it's the subject of `canSearchLibraryWith`, and the `SearchedLibrary` trigger reports it as the
searcher with `player` as the target. Most to the point, `decider.equals(player)` is already the
own-library gate for `getControlledWhileSearching` three lines below — the *other half of
Opposition Agent*. I wrote a second, worse version of a predicate that was already sitting in the
same method. Switched to `decider` and dropped the captured fetcher.

Two things fell out of that.

It fixes the cases I'd written off as card-script problems. Cranial Extraction, Surgical
Extraction and Haunting Echoes script as `DefinedPlayer$ <victim> | Chooser$ You`, so `decider`
separates them without touching the scripts. Across every library search carrying a `Chooser` —
52 of them — that's 38 cards, including Extirpate, Slaughter Games, Memoricide, Unmoored Ego and
the Bribery-likes. The "Not fixed here" section is gone.

It also fixes a regression I'd introduced. Fertilid's Favor, Fertilid, Varragoth and Auditore
Ambush are "target player searches their library" with no `DefinedPlayer`, so the fetcher defaults
to the caster while `Chooser$ Targeted` names the real searcher. Comparing against the fetcher
suppressed the Agent there — turning it off against exactly the effects it's meant to punish.
Confirmed rather than assumed: that case fails on the previous commit with
`expected:<Exile> but was:<Battlefield>`.

I checked the counter-case too — Intuition and Gifts Ungiven, where an opponent chooses from cards
you found in your own library. Those sub-abilities carry `NoLooking$ True`, so `searchedLibrary` is
already false before this line and they're unaffected.

Test rewritten to one case per shape (own tutor, retargeted search, named victim, granted search);
287 tests, 0 failures.

tool4ever (dismissed):
while the card needs a bigger rework anyway this minor fix should still be viable

liamiak:
(Non-AI) liamiak here: I've been working on scaffolding and process, and was letting this one lie until they got more mature. I realize I'm coming at this from a different angle, and apologize for any frustration and/or friction that it may cause.   

On to AI response:  

@tool4ever — you were right to push back. I found a card this broke, though not Ecological Appreciation.

Ecological Appreciation carries `NoLooking$ True`, and `searchedLibrary` is gated on exactly that, so the changed line is never reached there; I ran the card and the result is the same before and after. Six of the seven `Chooser`-but-not-searcher library scripts are protected the same way. The seventh is Guided Passage, where master is already wrong: nobody searches, yet all three cards go to an opposing Opposition Agent.

The real break was elsewhere — scripts with no `Chooser$` **and** no `DefinedPlayer$` that target a player, where `decider` stays the caster while `player` retargets. There are 18; 17 read "search target player's library". The exception is Turtle Tracks, and my change sent its land to the battlefield where master correctly exiled it.

Naming the fetcher (`DefinedPlayer$ Targeted`) fixes that, and three other things that are wrong on master today: it NPEs at `ChangeZoneEffect:924` with zero targets, only the first target searches, and the caster picks the land. It also never fired "whenever an opponent searches their library" — Archivist of Oghma stays silent where Diabolic Tutor on the same board triggers it.

That last one pointed at the cleaner split you asked for: `TriggerSearchedLibrary` already has this predicate as `SearchOwnLibrary$`, used by all four such cards. The trigger side had the split; the replacement side didn't. Rather than add a second parameter, I renamed `AbilityKey.FoundSearchingLibrary` — one reader, one card script, both wanting the own-library meaning — so both sides now use one name. If they ever make a card that wants "found while searching *any* library", that would want a second parameter rather than this rename.

PR #11403, opened 2026-07-26
Ignore controller inputs when unfocused

### LINKED ISSUE #11329: Option to disable controller input
If the game is open, even if it is not focused and I am on another virtual desktop, my controller inputs will perform unintended actions in the game.

Ideally there would be 3 levels to choose from:
1. Controller input disabled
2. Controller input enabled while card forge has focus
3. Controller input enabled even without focus

But even just adding option 1 would work well for my use case.

1. Closing the game when tabbing away to a different task that requires a controller, then reopening it when returning.
2. Pausing `javaw.exe` in task manager when tabbing away and unpausing it on return.

These workarounds work, but shouldn't be necessary. I would like to leave the game running and should be able to, especially since I don't actually intend to play Card Forge with a controller.

Eradev:
I feel this was an oversight, and shouldn't have been allowed in the first place. Therefore no options were added.

Fixes #11329

### DIFF
--- a/forge-gui-mobile/src/forge/Forge.java
+++ b/forge-gui-mobile/src/forge/Forge.java
@@ -303,11 +303,6 @@ public class Forge implements ApplicationListener {
         }
     }
 
-    /** Gamepad backends deliver input without OS focus gating; reject when unfocused. */
-    public static boolean shouldAcceptControllerInput() {
-        return hasWindowFocus;
-    }
-
     private static void haltControllerInput() {
         if (!isMobileAdventureMode) {
             return;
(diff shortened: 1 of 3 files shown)


forge-gui-mobile/src/forge/Forge.java
```diff
@@ -287,6 +289,33 @@ public static boolean hasGamepad() {
...
+        }
+        hasWindowFocus = focused;
+        if (!focused) {
+            haltControllerInput();
+        }
+    }
+
+    /** Gamepad backends deliver input without OS focus gating; reject when unfocused. */
+    public static boolean shouldAcceptControllerInput() {
+        return hasWindowFocus;
```
tool4ever: why this redundant method?

tool4ever (approved):

PR #11408, opened 2026-07-26
Some Card Fixes

BigCrunch22:
Added fixes for the following cards:
---

- **Incite:** Attach the _**SubAbility$ DBEffect**_ that was not included in the script.
- **Phantasmal Extraction:** Changed the _**RememberAllPumped$**_ to _**RememberPumped$**_ since the former doesn't exist based on Claude's analysis.
- **Reparations:** Included the _**OptionalDecider$ You**_ to the script to make drawing from the card's effect optional.
- **Vault 13: Dweller's Journey:** Attach the _**SubAbility$ DBChangeZoneAll**_ that was not included on the script.

### DIFF
--- a/forge-gui/res/cardsfolder/i/incite.txt
+++ b/forge-gui/res/cardsfolder/i/incite.txt
@@ -1,7 +1,7 @@
 Name:Incite
 ManaCost:R
 Types:Instant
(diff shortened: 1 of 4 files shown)


tool4ever (approved):
good finds, thanks

PR #11411, opened 2026-07-26
fix for Bullseye, Death Dealer unable to activate ability sometimes

### LINKED ISSUE #11405: Bullseye, Death Dealer unable to activate ability sometimes
**Describe the bug**
Bullseye's ability needs either an artifact on the battlefield under its contoller's control or a nonland card in hand, but he doesn't let you activate the ability unless there is an artifact on the battlefield. If there is, you get the choice, but without it, it won't even let you activate it.

**To Reproduce**
See attached game State. Bullseye unable to activate ability. Add a treasure token to the battlefield and suddenly he's able to activate the ability even able to choose the alternate cost of discard

[Bulleye_state.txt](https://github.com/user-attachments/files/30382686/Bulleye_state.txt)

**Expected behavior**
With mana open and valid targets to discard, Bullseye should be able to activate his ability even without artifacts on the battlefield 

Windows 2.0.14-SNAPSHOT-07.16

coder-5:
closes #11405

### DIFF
--- a/forge-game/src/main/java/forge/game/spellability/SpellAbility.java
+++ b/forge-game/src/main/java/forge/game/spellability/SpellAbility.java
@@ -456,6 +456,13 @@ public abstract class SpellAbility extends CardTraitBase implements ISpellAbilit
                 return true;
             }
         }
+        if (isActivatedAbility() && hasParam("AlternateCost")) {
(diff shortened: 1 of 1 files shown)


tool4ever (approved):
thanks

PR #11416, opened 2026-07-26
Fix NPE when restoring a snapshot taken before a card was created

cuinhellcat:
### Problem

`GameSnapshot.copyGameState()` copies the snapshot's cards back into the running game, but it never removes cards that the running game has and the snapshot does not — tokens, copies and effect cards that were created after the snapshot was taken.

The loop that restores attachments afterwards iterates the *current* battlefield and looks every card up in the snapshot:

```java
for (Card newCard : toGame.getCardsIn(ZoneType.Battlefield)) {
    Card fromCard = fromGame.findById(newCard.getId());

    if (fromCard.isAttachedToEntity()) {   // NPE: fromCard is null for cards the snapshot never had
```

For a card the snapshot does not know, `findById` returns `null` and the next line throws.

Interestingly `assignGameState` already anticipates such cards a few lines further down — it prints `"Missing card " + c` and continues — but by then `copyGameState` has already thrown.

### When it shows up

Not in the current use of snapshots: rolling back a cancelled cast (`GameActionUtil.rollbackAbility`) creates no new objects, so every card on the battlefield is also in the snapshot. Any restore that reaches back past the creation of a token or a copy throws instead — which is what happens as soon as snapshots are used to step further back than the action in progress.

### Fix

Remove those cards from the game as part of the restore. Returning to a state in which they did not exist is exactly what a restore means, and it also gives the `"Missing card"` branch below nothing left to report.

### Notes

* Built against `master`; `forge-game` compiles clean.
* No test added, per the note in CONTRIBUTING about agent-added tests. I do have a regression test for this (it reproduces the NPE reliably by adding a card after the snapshot and restoring) and am happy to include it if you would like it.
* This came out of a personal multi-step undo experiment: https://github.com/cuinhellcat/mtg-forge-rewind — the fix is independent of that and useful on its own, so it is offered separately.

### AI disclosure

### DIFF
--- a/forge-game/src/main/java/forge/game/GameSnapshot.java
+++ b/forge-game/src/main/java/forge/game/GameSnapshot.java
@@ -14,6 +14,7 @@ import forge.game.spellability.SpellAbility;
 import forge.game.spellability.SpellAbilityStackInstance;
 import forge.game.trigger.TriggerType;
 import forge.game.zone.PlayerZoneBattlefield;
+import forge.game.zone.Zone;
 import forge.game.zone.ZoneType;
 
 import java.util.Collections;
@@ -332,6 +333,19 @@ public class GameSnapshot {
         Collections.sort(unorderedEntities);
         for(UnorderedEntities ue : unorderedEntities) {
             setCardInCopiedGame(toGame, ue.toPlayer, ue.fromCard, ue.newCard, ue.fromType, ue.zonePosition);
+        }
+
+        // Cards that exist in the game being restored but not in the snapshot: tokens,
+        // copies and effect cards created after it was taken. There is no earlier state to
+        // put them back into, so they leave the game. Without this the loop below looks
+        // them up in the snapshot and dereferences the null it gets back.
+        for (Card extraCard : toGame.getCardsInGame()) {
+            if (fromGame.findById(extraCard.getId()) == null) {
+                Zone zone = extraCard.getZone();
+                if (zone != null) {
+                    zone.remove(extraCard);
+                }
+            }
         }
 
         // This loop happens later to make sure all cards are in the correct zone first


forge-game/src/main/java/forge/game/GameSnapshot.java
```diff
@@ -334,6 +335,19 @@ public void copyGameState(Game fromGame, Game toGame) {
             setCardInCopiedGame(toGame, ue.toPlayer, ue.fromCard, ue.newCard, ue.fromType, ue.zonePosition);
         }
 
+        // Cards that exist in the game being restored but not in the snapshot: tokens,
+        // copies and effect cards created after it was taken. There is no earlier state to
+        // put them back into, so they leave the game. Without this the loop below looks
+        // them up in the snapshot and dereferences the null it gets back.
+        for (Card extraCard : toGame.getCardsInGame()) {
```
tehdiplomat, line 342: Does this break things like meld?

cuinhellcat: No — melding mints no new card. `MeldEffect` changes the primary to its Meld state and moves the secondary into `PlayerZoneBattlefield.meldedCards`; both keep their ids, and `forEachCardInGame` walks `getMeldedCards()`, so both halves are in any snapshot taken before or during the meld. This loop only drops ids the snapshot does not have, so neither half is ever a candidate.

I checked rather than assumed, with and without this patch, and the results are identical either way: snapshot → meld → restore keeps both halves, and a snapshot taken while melded restores the primary in its Meld state with the secondary still in `getMeldedCards()`.

What the check did turn up is unrelated to this PR but reproduces on master unchanged: a restore adds the melded secondary back into the battlefield's card list without ever calling `removeFromMelded`, so afterwards it is both melded and loose on the battlefield. Happy to look at that separately if it is worth a report — and happy to share the throwaway test either way.

(The answer above is claude code. I am from germany, I am obviously vibe coding ... so I don't know if the above is true or not. If this kind of "help" is not appreciated, just tell me. I don't want to cause trouble or steal anyones time. Just want to help)

cuinhellcat: Correction to my answer above, and thanks for the nudge — the honest version is: this does not break meld, but meld is already broken.

Snapshots mishandle it in both directions on unmodified master. A snapshot taken while melded restores the back half loose on the battlefield, turning one permanent into two; a snapshot taken before the meld leaves it in `getMeldedCards()` while also putting it on the battlefield, with the front half still pointing at it. `copyGameState` places cards by zone type alone, and the melded collection is not a zone type.

Fix and both reproductions: #11421. Independent of this PR — different lines, and it reproduces without this one applied.

tool4ever:
Hallo 👋

FYI there's already a PR related to your experiment: #10294
but it's probably dead at this point :/
(mostly just need a clean GUI connection with existing Undo)
imo code from Claude is better in quality than other agents, but things get a bit messy if human controller can't understand at least either Java or some software architecture in general to push back if it hallucinates:
then you'd essentially just be our copy-paster and bill payer 😛

cuinhellcat:
> Hallo 👋
> 
> FYI there's already a PR related to your experiment: #10294 but it's probably dead at this point :/ (mostly just need a clean GUI connection with existing Undo) imo code from Claude is better in quality than other agents, but things get a bit messy if human controller can't understand at least either Java or some software architecture in general to push back if it hallucinates: then you'd essentially just be our copy-paster and bill payer 😛

I have no problem in being just a bill payer (I have this subscription and I can use some of my contingent for this project.)

As long as copy pasting helps, I will do it.

But I totally understand your point. I will switch to testing, what claude code did and only post here, if tests work on my end... If I cause trouble, I will immediately step back. Just tell me to stop, I will do so.

tool4ever (approved):
LGTM

PR #11418, opened 2026-07-27
Clean up the ability when a cast is rolled back via snapshot

cuinhellcat:
Addresses #10049 and, most likely, #8762.

### Problem

`GameActionUtil.rollbackAbility` returns as soon as the snapshot has been restored:

```java
if (game.restoreGameState()) {
    // If we're able to restore the whole game state when rolling back an ability don't try to manually roll back
    System.out.println("Restored state from snapshot! ...");
    return;
}
```

That is right for the cards, their zones and the mana pool — `copyManaPool` already puts the mana back, so `payment.refundPayment()` on top of it would hand out the mana twice. It is not right for everything else the early return skips:

* `ability.setXManaCostPaid(null)`, `setSpendPhyrexianMana(false)`, `clearPipsToReduce()`, `setPaidLife(0)` and the `Announce` SVars
* `ability.clearTargets()`, the Charm mode chain, `resetOnceResolved()`
* `oldCard.setCastSA(null)` and `setCastFrom(null)`
* `game.getStack().clearFrozen()` and `getTriggerHandler().clearWaitingTriggers()`

A `SpellAbility` is not part of a snapshot, and neither is the frozen stack. So after cancelling a cast with undo restore enabled, the card keeps the announced value and targets from the attempt that was called off, still believes it is being cast, and the stack stays frozen.

That is what the two reports look like from the outside:

* **#10049** — March of the Otherworldly Light can no longer be cast after cancelling with an invalid number. Its additional cost is announced, and the announced value is exactly what survives.
* **#8762** — cancelling a cast sometimes freezes the game. A stack left frozen with waiting triggers does that, and it explains why it only happens sometimes: `Hanmac` already pointed at `GameSnapshot` in that thread, and the reporter was told to turn undo restore off.

### Fix

Do the ability-level cleanup in both paths — still no refund, still no zone handling. The values reset in the snapshot path are the same ones the manual path resets, pulled into two small helpers so the paths cannot drift apart.

### Reproduction

At engine level: enable snapshots, stash, then set `xManaCostPaid`, `castSA`, `castFrom` and freeze the stack the way a cast does, and call `rollbackAbility`. Before the change the announced X, the cast markers and the frozen stack all survive; after it they are gone. I could not drive the actual GUI cancel headlessly, so I have not confirmed the two reports end to end — the mechanism matches what they describe.

### Notes

* No test added, per the note in CONTRIBUTING about agent-added tests. The reproduction above exists as one (fails before, passes after) if you want it.

### AI disclosure

### DIFF
--- a/forge-game/src/main/java/forge/game/GameActionUtil.java
+++ b/forge-game/src/main/java/forge/game/GameActionUtil.java
@@ -942,6 +942,20 @@ public final class GameActionUtil {
             // If we're able to restore the whole game state when rolling back an ability don't try to manually roll back
             System.out.println("Restored state from snapshot! Rolled back: " + ability.getHostCard().getName() + " - " + ability.getActivatingPlayer());
 
+            // The snapshot covers the cards, their zones and the mana pool, so the manual
+            // rollback below and its refund would double up on it. The ability is not part of
+            // the snapshot though, and neither is the frozen stack: without the cleanup an
+            // announced value or a chosen target survives into the next attempt to cast the
+            // card, and the stack stays frozen.
+            Card restored = game.getCardState(oldCard, null);
+            if (restored != null) {
+                restored.setCastSA(null);
+                restored.setCastFrom(null);
+            }
+            resetAnnouncedCosts(ability);
+            resetChosenTargets(ability);
+            game.getStack().clearFrozen();
+            game.getTriggerHandler().clearWaitingTriggers();
             return;
         }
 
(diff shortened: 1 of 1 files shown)


tool4ever:
Hmn?
SpellAbilities live on Cards, so they should be fresh either way...

cuinhellcat:
Different directions of the snapshot.

Taking one builds new cards, so their abilities are fresh. Restoring one builds nothing: `copyGameState` looks each card up by id and keeps the object it finds. Same card, same `SpellAbility` instances, including whatever the cancelled cast wrote onto them.

Unmodified master — stash, set `xManaCostPaid`, restore, look the card up again:

```
same card object?    true
same ability object? true
X on that ability:   3
```

The manual path already resets exactly this, a few lines below the early return: `setXManaCostPaid(null)`, `clearTargets()`, `resetOnceResolved()`, `clearFrozen()`, `clearWaitingTriggers()`. If abilities came back fresh, those resets would be pointless there too.

Concretely: announce X=4 on Fireball, cancel at the mana prompt. With undo restore on, the card is back in hand still carrying X=4 and its target, and the stack stays frozen — #10049 and #8762 from the outside.

Refund and zone handling stay skipped, since the snapshot covers those.

cuinhellcat:
I'm off Forge for the time being, so I won't be taking this further. Happy for it to be closed — the reasoning is all in the thread above if anyone wants it later.

PR #11419, opened 2026-07-27
Snapshot cards into the zone they are actually in, not the controller's

### LINKED ISSUE #9297: Vraska, the Silencer on a manifested sorcery
**Describe the bug**
A sorcery was manifested face down as a 2/2 creature and it died while Vraska, the Silencer was in play. I paid to have the sorcery-creature returned to play under my control as a treasure and the game locked.

**To Reproduce**
1 Have Vraska, the Silencer in play
2 Have a sorcery manifested face down as a 2/2 creature.
3 Kill the 2/2
4 pay 1 to return it to play.

**Expected behavior**
I'm not sure if it should be able to return a sorcery or not.

**Screenshots**

Windows 10

**Additional context**

Adding Vraska, the Silencer (16): Whenever a nontoken creature an opponent controls dies, you may pay {1}. If you do, return that card to the battlefield tapped under your control. It's a Treasure artifact with "{T}, Sacrifice this artifact: Add one mana of any color," and it loses all other card types. [Zone Changer: Face-down card (H0)] to stack
Game-6 > java.lang.IndexOutOfBoundsException: Index: 9, Size: 7
	at java.base/java.util.ArrayList.rangeCheckForAdd(ArrayList.java:756)
	at java.base/java.util.ArrayList.add(ArrayList.java:481)
	at forge.util.collect.FCollection.insert(FCollection.java:431)
	at forge.util.collect.FCollection.add(FCollection.java:417)
	at forge.game.zone.Zone.add(Zone.java:139)
	at forge.game.zone.Zone.add(Zone.java:89)
	at forge.game.zone.Zone.add(Zone.java:86)
	at forge.game.GameSnapshot.copyGameState(GameSnapshot.java:321)
	at forge.game.GameSnapshot.assignGameState(GameSnapshot.java:79)
	at forge.game.GameSnapshot.makeCopy(GameSnapshot.java:52)
	at forge.game.GameSnapshot.makeCopy(GameSnapshot.java:38)
	at forge.game.Game.stashGameState(Game.java:195)
	at forge.game.phase.PhaseHandler.mainLoopStep(PhaseHandler.java:1058)
	at forge.game.phase.PhaseHandler.mainGameLoop(PhaseHandler.java:1039)
	at forge.game.phase.PhaseHandler.startFirstTurn(PhaseHandler.java:1033)
	at forge.game.GameAction.startGame(GameAction.java:2333)
	at forge.game.Match.startGame(Match.java:91)
	at forge.gamemodes.match.HostedMatch.lambda$startGame$1(HostedMatch.java:259)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)

cuinhellcat:
### Problem

`copyGameState` decides which player's zone a card goes into by asking the card for its controller:

```java
Player toPlayer = findBy(toGame, fromCard.getController());
```

That is only right on the battlefield. A graveyard, hand or library belongs to the card's **owner**. When control has changed — a stolen creature, or a card returned to play under someone else's control — the card is filed under the wrong player, and restoring the snapshot moves it into that player's graveyard for good. Cards silently change hands across an undo.

### Fix

Take the player from the zone the card is actually in: the controller on the battlefield, the owner everywhere else, with a fallback to the controller for the stack, which has no player. Card copies made for a stored game likewise keep their original owner rather than inheriting the controller.

### Reproduction

Put nine cards in one player's graveyard plus a tenth whose controller is the opponent, snapshot and restore: before the change the tenth ends up in the opponent's graveyard, after it stays where it belongs.

### Relationship to #9297

This came out of looking into #9297, and I want to be precise about what it does and does not do. That report's stack trace is the same mismatch seen from a different angle:

```
java.lang.IndexOutOfBoundsException: Index: 9, Size: 7
    at forge.util.collect.FCollection.insert(FCollection.java:431)
    at forge.game.zone.Zone.add(Zone.java:139)
    at forge.game.GameSnapshot.copyGameState(GameSnapshot.java:321)
```

The card kept the index it had in its owner's graveyard and was inserted at that index into the controller's shorter one. But `setCardInCopiedGame` no longer inserts by index on master, and I could not get master to throw that exception. So this PR does not claim to close #9297 — it fixes the misplacement the crash grew out of, which is a bug on its own.

If you would also like `Zone.add` to clamp an out-of-range index defensively (it already guards the empty-zone case a few lines up), that is a two-line addition and I am happy to add it.

### Notes

* No test added, per the note in CONTRIBUTING about agent-added tests; I have the reproduction above as one if you want it.

### AI disclosure

### DIFF
--- a/forge-game/src/main/java/forge/game/GameSnapshot.java
+++ b/forge-game/src/main/java/forge/game/GameSnapshot.java
@@ -298,7 +298,12 @@ public class GameSnapshot {
 
         for(Card fromCard : fromGame.getCardsInGame()) {
             Card newCard = toGame.findById(fromCard.getId());
-            Player toPlayer = findBy(toGame, fromCard.getController());
+            // Zones belong to a player: the controller's battlefield, but the owner's
+            // graveyard, hand and library. Going by the controller alone files a card whose
+            // control has changed under the wrong player, so a stolen creature that died
+            // reappears in the thief's graveyard when the snapshot is used.
+            Player fromZonePlayer = fromCard.getZone().getPlayer();
+            Player toPlayer = findBy(toGame, fromZonePlayer != null ? fromZonePlayer : fromCard.getController());
             ZoneType fromType = fromCard.getZone().getZoneType();
             int zonePosition = 0;
             if (ZoneType.ORDERED_ZONES.contains(fromType)) {
@@ -308,8 +313,9 @@ public class GameSnapshot {
             }
 
             if (newCard == null) {
-                // Storing a game uses this path...
-                newCard = createCardCopy(toGame, toPlayer, fromCard);
+                // Storing a game uses this path... the copy keeps the original owner, which
+                // is not the same player as the one whose zone it currently sits in.
+                newCard = createCardCopy(toGame, findBy(toGame, fromCard.getOwner()), fromCard);
             } else {
                 ZoneType type = newCard.getZone().getZoneType();
                 if (type != fromType) {


forge-game/src/main/java/forge/game/GameSnapshot.java
```diff
@@ -298,7 +298,12 @@ public void copyGameState(Game fromGame, Game toGame) {
 
         for(Card fromCard : fromGame.getCardsInGame()) {
             Card newCard = toGame.findById(fromCard.getId());
-            Player toPlayer = findBy(toGame, fromCard.getController());
+            // Zones belong to a player: the controller's battlefield, but the owner's
+            // graveyard, hand and library. Going by the controller alone files a card whose
+            // control has changed under the wrong player, so a stolen creature that died
+            // reappears in the thief's graveyard when the snapshot is used.
+            Player fromZonePlayer = fromCard.getZone().getPlayer();
+            Player toPlayer = findBy(toGame, fromZonePlayer != null ? fromZonePlayer : fromCard.getController());
```
tool4ever, line 306: technically `fromCard.getZone().getPlayer() == fromCard.getController()` shouldn't be different either way, since it falls back to owner
are you saying it's somehow not getting updated on LTB case?

cuinhellcat: You are right: they cannot differ.

My reasoning was a stolen creature dying. The temporary controller is only cleared while the card is in play — `doLoseControl` returns early otherwise — so I expected a card sitting in its owner's graveyard while still naming the thief as controller. Testing that says otherwise:

```
same object? false   id before=1  id after=1
after dying: zone=Graveyard  zonePlayer=p2  controller=p2  owner=p2
```

Moving to the graveyard hands back a new card object, and a fresh object has no temporary controllers, so the controller falls back to the owner — exactly as you said.

The reproduction I had only "worked" because it called `setController` on a card already in a graveyard, which the engine never does. So this PR fixes nothing real. Closing it.

PR #11420, opened 2026-07-27
Restore the foretold state when a snapshot is restored

### LINKED ISSUE #7844: bug: cancelled foretell card is no longer foretold
**Describe the bug**
If you attempt to cast a card with Foretell but cancel when asked to pay the mana, the card returns to exile face-up and can no longer be cast with Foretell.

**To Reproduce**
1. Foretell a card with that ability, e.g. `Doomskar Oracle`
2. On your next turn, attempt to cast the card with Foretell from your exile zone, but do not pay its mana cost.
3. Instead, hit "Cancel"
4. Observe that the card in your exile zone is now face-up, and no longer provides the option to cast it using Foretell.

**Expected behavior**
The card exiled with Foretell should remain face-down with the Foretell emblem and still be castable with the Foretell ability.

**Platform Information**
This happens only in the desktop version; mobile works fine in this respect (but Foretell does have other issues for another ticket).

cuinhellcat:
Fixes #7844.

### Problem

Cancel the cast of a foretold card and it comes back to exile without its foretold state, so it can no longer be cast with Foretell.

`setCardInCopiedGame` restores a card's tapped, face-down, manifested, sickness and state-name, but the two lines for the foretold flags are commented out:

```java
newCard.setSickness(fromCard.hasSickness());
//newCard.setForetold(fromCard.isForetold());
//newCard.setForetoldCostByEffect(fromCard.isForetoldCostByEffect());
newCard.setState(fromCard.getCurrentStateName(), false);
```

They have been comments since the method was introduced in "Trying to fix gateway dialog glitching" — never active, so this is not re-enabling something that was switched off for a reason. Nothing else restores that state, while `CardCopyService.copyCard` already carries both flags into the snapshot, so the information is there and simply not read back.

### Reproduction

Take a snapshot with a face-down foretold card in exile, turn it face up onto the stack the way casting does, restore. The card returns to exile face down, and `isForetold()` is false.

### Notes

* `isForetold()` reads the field for a card in exile and derives it from the cast ability elsewhere, so `setForetold(fromCard.isForetold())` restores the value that matters and is a no-op for cards that are not in exile.
* No test added, per the note in CONTRIBUTING about agent-added tests. I have one that fails before the change and passes after (the reproduction above); happy to include it if you want it.

### AI disclosure

### DIFF
--- a/forge-game/src/main/java/forge/game/GameSnapshot.java
+++ b/forge-game/src/main/java/forge/game/GameSnapshot.java
@@ -389,8 +389,8 @@ public class GameSnapshot {
         newCard.setFaceDown(fromCard.isFaceDown());
         newCard.setManifested(fromCard.getManifestedSA());
         newCard.setSickness(fromCard.hasSickness());
-        //newCard.setForetold(fromCard.isForetold());
-        //newCard.setForetoldCostByEffect(fromCard.isForetoldCostByEffect());
+        newCard.setForetold(fromCard.isForetold());
+        newCard.setForetoldCostByEffect(fromCard.isForetoldCostByEffect());
         newCard.setState(fromCard.getCurrentStateName(), false);
     }
 


cuinhellcat:
Closing this in favour of #10294, which already restores the foretold flags as part of its wider pass over the snapshot state. No point spending review time on the same two lines twice.

If #10294 does not move, this is a one-line change to bring back — say the word and I will reopen.

PR #11422, opened 2026-07-27
Give a stolen card back when a snapshot is restored

cuinhellcat:
### Problem

Nothing in `GameSnapshot` restores who controls a card. A creature that changed hands after the snapshot was taken stays with its new controller when the state is put back.

Reported from play: cast Claim the Firstborn, then roll back. The spell returns to hand, the creature stays on the thief's side.

The zone handling misses it too:

```java
ZoneType type = newCard.getZone().getZoneType();
if (type != fromType) { ... remove ... }
```

A card already in the game is only taken out of its current zone when the zone *type* differs — and a move between two players' battlefields keeps the type. So the creature is added to the owner's battlefield while remaining in the thief's, ending up in both at once.

### Fix

Restore the controller alongside the rest of the card state, and take the card out of whatever zone it currently sits in unless that is already the zone it is going into.

### Reproduction

Snapshot with a creature under its owner's control, give control to the opponent with a zone correction, restore: before the change control stays with the opponent and the creature is in both battlefields; after it, the creature is back under its owner with a single copy on the board.

### Notes

* Touches the same method as #11419, though not the same lines. If you take both, #11419 fixes which player's zone a card is filed into and this one fixes leaving the old zone behind; they are independent but related.
* No test added, per the note in CONTRIBUTING about agent-added tests. The reproduction above exists as one (fails before, passes after) if you want it.

### AI disclosure

### DIFF
--- a/forge-game/src/main/java/forge/game/GameSnapshot.java
+++ b/forge-game/src/main/java/forge/game/GameSnapshot.java
@@ -14,6 +14,7 @@ import forge.game.spellability.SpellAbility;
 import forge.game.spellability.SpellAbilityStackInstance;
 import forge.game.trigger.TriggerType;
 import forge.game.zone.PlayerZoneBattlefield;
+import forge.game.zone.Zone;
 import forge.game.zone.ZoneType;
 
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


cuinhellcat:
Closing this in favour of #10294, which already restores the controller as part of its wider pass over the snapshot state.

One piece of this PR is not in #10294, so I do not want it to get lost: a card already in the game is only removed from its current zone when the zone *type* differs, and a move between two players' battlefields keeps the type. A stolen creature therefore ends up in both battlefields at once, independently of who the card names as controller. If that is worth having on its own I am happy to open it as a separate one-hunk PR.

Found while playing: cast Claim the Firstborn, roll back, creature stays on the thief's side.

tool4ever:
seems relevant enough so we can just discuss it here, that's way easier than on some already too complex other one:
- I think the zone part is clearly right
- I don't know about the controller one, how does this even work if it's just from a temporary effect which was set up with a different timestamp? Wouldn't the GameCommand meant to remove the entry just fail?

cuinhellcat:
Sorry for the slow reply, and thanks for looking at it.

I'm off Forge for the time being — my time is going elsewhere — so I'd rather hand this back than leave it sitting in your queue.

Worth flagging: #10294 is closed now. I withdrew this PR and #11420 in its favour, so both pieces are unowned again — the zone hunk here, and the one-line foretold restore in #11420.

On the controller question: you're right to doubt it. I can't check it properly without going back into the code, so treat that hunk as unverified rather than take my word for it.

Closing this. The zone part is one change and reproduces with Claim the Firstborn plus a rollback, if anyone wants to pick it up.

PR #11401, opened 2026-07-26
Fix commander map in GameSnapshot

Eradev:
Fixes a bug reported on Discord where the game would crash with Experimental Undo on and a mutating Planeswalker.

https://discord.com/channels/267367946135928833/1530448986111086642

PR #11426, opened 2026-07-27
Revert "Check meld "zone" when generating snapshots"

tool4ever:
Reverts Card-Forge/forge#9572

We cannot have this method simply return the secondary meld halfs all the time.
It's used everywhere to obtain the real boardstate, `ReplacementHandler` etc.
So there's a medium risk we'd be using traits that don't exist in the game currently ⚠️

This changes it so it's more aligned with the logic introduced by #11401

Together with #11421 everything should then be fine....


Hanmac:
Somewhat related to #11058 too

Hanmac:
I wish to (re)move the Battlefield.meldedCards property too, but that might be something for later

Hanmac (approved):

PR #11427, opened 2026-07-27
Fix #11375: Russian is blocked on Android/iOS and unreadable on Desktop

tCynik:
#11375 merged Russian support, but it doesn't actually work:

- **Android/iOS**: switching to Russian is blocked by a "please download CJK font" prompt with
  nothing to download - the required font (`Roboto-Bold`) isn't in the downloadable list, even
  though it already ships with the app.
- **Desktop**: the switch succeeds, but nearly all UI text (menus/buttons/labels) renders as
  tofu - the desktop skin's base font has no Cyrillic glyphs, and Russian never wired up the
  font-fallback check that Chinese/Japanese/Korean already use for this exact situation.

Two independent commits, one per platform - see commit messages for full root-cause detail.

### Fix

- `FSkinFont` / `SettingsPage` (mobile): resolve the required font from the app's own bundled
  assets instead of only the downloaded-fonts cache, and auto-apply it instead of blocking.
- `LangRussian` (desktop): add `canDisplayCheck()` so the existing font-fallback mechanism
  (already used by zh/ja/ko) fires for Russian too.

### Test plan

- [x] Builds clean (Android + Desktop via Maven).
- [x] Android 17 (API 37), clean-room install: Russian applies immediately, no font prompt.
- [x] Windows desktop: confirmed broken before commit 2 (tofu outside card text), fixed after,
  including in a live match.
- [ ] Not verified on iOS (no device/build environment available).

---

Note: this contribution was substantially coded with the assistance of Claude Code (Anthropic).

### DIFF
--- a/forge-core/src/main/java/forge/util/lang/LangRussian.java
+++ b/forge-core/src/main/java/forge/util/lang/LangRussian.java
@@ -37,4 +37,15 @@ public class LangRussian extends Lang {
         return "Roboto-Bold";
     }
 
+    // Desktop (GuiUtils.newFont) uses this character to test whether font1.ttf can render
+    // the active locale's script; on failure it falls back to the system's default Swing
+    // font instead of font1.ttf. Every CJK Lang subclass already overrides this (see
+    // LangChinese/LangJapanese/LangKorean) - Russian was missing it, so font1.ttf's "can
+    // display ' '" trivially passed and desktop kept rendering Cyrillic UI text with a font
+    // that has no Cyrillic glyphs at all, leaving most non-card text unreadable there.
+    @Override
+    public char canDisplayCheck() {
+        return 'Р';
+    }
+
 }
--- a/forge-core/src/main/java/forge/util/lang/LangRussian.java
+++ b/forge-core/src/main/java/forge/util/lang/LangRussian.java
@@ -37,12 +37,8 @@ public class LangRussian extends Lang {
         return "Roboto-Bold";
     }
 
-    // Desktop (GuiUtils.newFont) uses this character to test whether font1.ttf can render
-    // the active locale's script; on failure it falls back to the system's default Swing
-    // font instead of font1.ttf. Every CJK Lang subclass already overrides this (see
-    // LangChinese/LangJapanese/LangKorean) - Russian was missing it, so font1.ttf's "can
-    // display ' '" trivially passed and desktop kept rendering Cyrillic UI text with a font
-    // that has no Cyrillic glyphs at all, leaving most non-card text unreadable there.
+    // desktop's font1.ttf-can-render-script probe (GuiUtils.newFont); unset here like
+    // the CJK langs, it silently passed and left Cyrillic UI text unreadable
     @Override
     public char canDisplayCheck() {
         return 'Р';
--- a/forge-gui-mobile/src/forge/assets/FSkinFont.java
+++ b/forge-gui-mobile/src/forge/assets/FSkinFont.java
@@ -435,7 +435,7 @@ public class FSkinFont {
         //not found generate
         if (useCjkFont && !Forge.forcedEnglishonCJKMissing) {
             String ttfName = Forge.CJK_Font;
-            FileHandle ttfFile = Gdx.files.absolute(ForgeConstants.FONTS_DIR + ttfName + ".ttf");
+            FileHandle ttfFile = resolveCjkFontFile(ttfName);
             if (ttfFile != null && ttfFile.exists()) {
                 generateFont(ttfFile, fontName, fontSize);
             }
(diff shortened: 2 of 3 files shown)


forge-gui-mobile/src/forge/assets/FSkinFont.java
```diff
@@ -521,13 +521,29 @@ public static Iterable<String> getAllCJKFonts() {
         final Array<String> allCJKFonts = new Array<>();
 
         allCJKFonts.add("None");
-        final FileHandle dir = Gdx.files.absolute(ForgeConstants.FONTS_DIR);
-        for (FileHandle fontFile : dir.list()) {
-            String fontName = fontFile.name();
-            if (!fontName.endsWith(".ttf")) { continue; }
-            allCJKFonts.add(fontName.replace(".ttf", ""));
+        for (String dirPath : new String[]{ForgeConstants.FONTS_DIR, ForgeConstants.COMMON_FONTS_DIR}) {
```
Hanmac: Maybe use `List.of(ForgeConstants.FONTS_DIR, ForgeConstants.COMMON_FONTS_DIR)` to make it more readable?

tCynik: ok, i've improved readability a bit )

Hanmac (dismissed):

Hanmac:
@tool4ever what do you think?

forge-core/src/main/java/forge/util/lang/LangRussian.java
```diff
@@ -37,4 +37,15 @@ public String getFontFile() {
         return "Roboto-Bold";
     }
 
+    // Desktop (GuiUtils.newFont) uses this character to test whether font1.ttf can render
+    // the active locale's script; on failure it falls back to the system's default Swing
+    // font instead of font1.ttf. Every CJK Lang subclass already overrides this (see
+    // LangChinese/LangJapanese/LangKorean) - Russian was missing it, so font1.ttf's "can
+    // display ' '" trivially passed and desktop kept rendering Cyrillic UI text with a font
+    // that has no Cyrillic glyphs at all, leaving most non-card text unreadable there.
```
tool4ever: I don't think we need 6 comment lines explaining AI thoughts

tCynik: oops, not good.
Fixed all comments, thanks )

jamincollins:
## Findings

#### 1. canDisplayCheck returns char but GuiUtils checks canDisplay on Font
Description: LangRussian now overrides canDisplayCheck() to return 'Р'. GuiUtils.newFont() uses ttf.canDisplay(lang.canDisplayCheck()) to decide if font supports language. The check assumes canDisplayCheck() returns a representative glyph. For Russian, returning 'Р' is fine, but the logic relies on ttf.canDisplay which may return true for partial coverage, leading to false positive.
Test case:
```java
Lang lang = Lang.initInstance("ru-RU");
Font ttf = Font.createFont(Font.TRUETYPE_FONT, new File("font1.ttf"));
boolean can = ttf.canDisplay(lang.canDisplayCheck());
// can may be true for 'Р' even if many Cyrillic glyphs missing
System.out.println("Can display check: " + can);
```
Impact: Desktop may still use font1.ttf if 'Р' present, causing unreadable UI for other Cyrillic characters.

#### 2. SettingsPage auto-sets UI_CJK_FONT for bundled fonts without user consent
Description: SettingsPage.valueChanged() now checks if required font exists via resolveCjkFontFile, and if exists, silently sets prefs.UI_CJK_FONT to requiredFont and saves. This bypasses user choice and may set Russian font preference even when user previously chose another.
Test case:
```java
ForgePreferences prefs = FModel.getPreferences();
prefs.setPref(FPref.UI_CJK_FONT, ""); // empty
// user changes language to ru-RU
// code auto-sets prefs to Roboto-Bold without dialog
String font = prefs.getPref(FPref.UI_CJK_FONT);
assert font.equals("Roboto-Bold");
```
Impact: Unexpected preference mutation; may cause font loading conflicts.

#### 3. resolveCjkFontFile fallback may load wrong font for CJK languages
Description: resolveCjkFontFile first checks FONTS_DIR then COMMON_FONTS_DIR. If a user has a file named Roboto-Bold.ttf in FONTS_DIR (unlikely but possible), it will be used instead of bundled. No validation of font coverage.
Test case:
```java
FileHandle fh = FSkinFont.resolveCjkFontFile("Roboto-Bold");
System.out.println(fh.path());
// could point to user-downloaded file with no Cyrillic
```
Impact: Potential font mismatch.

#### 1. Lang.getFontFile() returns "Roboto-Bold" for Russian, but getFontFile() is used for CJK detection
Description: Code treats any non-null getFontFile() as CJK needing custom font. Russian now triggers CJK path, but Russian font is bundled. If Forge.locale changes to null, Lang.initInstance may fallback to default English, causing null pointer in getFontFile? initInstance handles null.
Test case:
```java
Lang lang = Lang.initInstance(null);
String f = lang.getFontFile(); // should be null
```
Impact: Potential NPE if initInstance returns null.

#### 2. FSkinFont.preloadAll uses getFontFile != null to decide MAX_FONT_SIZE
Description: Russian now returns non-null font file, so MAX_FONT_SIZE becomes MAX_FONT_SIZE_MANY_GLYPHS (36). This reduces font size range for Russian users, possibly causing layout issues.
Test case:
```java
FSkinFont.preloadAll("ru-RU");
int max = FSkinFont.MAX_FONT_SIZE; // now 36 instead of 72
```
Impact: Reduced font size range for Russian.

#### 3. SettingsPage valueChanged does not handle requiredFont == null case explicitly
Description: Code checks if requiredFont != null then exists. If requiredFont null, no check, language change proceeds. OK but path not tested.

#### 1. Existing preferences with UI_CJK_FONT empty for Russian users will be mutated on next language change
Description: Prior to PR, changing to Russian blocked with dialog. Now it auto-sets preference. Existing installations with empty pref will change behavior silently.
Test case:
```java
// user has UI_CJK_FONT empty, switches language to ru-RU
// preference now set to Roboto-Bold
```
Impact: Backwards compatibility change.

#### 2. FSkinFont.getAllCJKFonts now scans COMMON_FONTS_DIR, exposing bundled fonts to CJK selector
Description: Users may now select Roboto-Bold as CJK font, which is not intended for CJK. Could cause incorrect font selection.
Test case:
```java
List<String> fonts = FSkinFont.getAllCJKFonts();
// contains Roboto-Bold
```
Impact: UI confusion.

#### 1. Desktop font selection now uses canDisplayCheck, but mobile uses resolveCjkFontFile
Description: Inconsistent handling between desktop and mobile. Desktop still relies on canDisplay probe; mobile uses file existence. Could diverge.
Test case:
```java
// Desktop: canDisplayCheck 'Р' may still pass for font1.ttf if 'Р' present
// Mobile: resolveCjkFontFile finds Roboto-Bold, uses it
```
Impact: Different rendering per platform.

#### 2. LangRussian canDisplayCheck returns 'Р' but default Lang returns ' '
Description: For languages without override, canDisplayCheck returns space, which most fonts can display, so GuiUtils will not fallback to system font. This is existing behavior.

### Performance and security
#### 1. FSkinFont.getAllCJKFonts scans two directories each time settings page opens
Description: No caching; may cause I/O on UI thread.
Test case:
```java
long t0 = System.nanoTime();
FSkinFont.getAllCJKFonts();
long dt = System.nanoTime() - t0;
```
Impact: Minor UI lag.

## Verdict
Request Changes
The PR fixes Russian blocking and unreadable desktop text by routing Russian through CJK font pathway and adding canDisplayCheck, but introduces silent preference mutation, inconsistent platform handling, and exposes bundled fonts in CJK selector. Needs refinement to avoid auto-setting prefs and to separate Russian from true CJK logic.

## Recommended Fixes
1. Separate Russian font handling from CJK pathway: introduce isCyrillic() flag or distinct getFontFile for bundled vs downloadable fonts, avoid treating Russian as CJK.
2. Do not auto-set UI_CJK_FONT silently; prompt user or pre-populate preference during first run.
3. Restrict getAllCJKFonts to user-downloaded fonts only; keep bundled fonts out of selector.
4. Verify canDisplayCheck character choice covers Cyrillic block adequately; consider checking multiple glyphs.
5. Add unit tests for LangRussian.canDisplayCheck and SettingsPage language change flow.

---

Hanmac (approved):

PR #11429, opened 2026-07-27
Fix multiplayer controller setup during AI simulation

Madwand99:
## Summary

- install scoped simulation controllers for every non-preserved player while resolving a copied game stack
- preserve the simulated AI controller and restore all temporary controller overrides afterward
- pass the copied AI player consistently from simulation call sites
- add multiplayer regressions covering complete stack resolution, deterministic value-aware sacrifice choices, stable scoring, and visible spell selection

## Motivation

`GameSimulator.resolveStack()` currently installs a simulation controller for only one selected opponent. In multiplayer simulations, effects that require another opponent to make a choice can therefore consult that copied player's match controller instead. This can make simulation results incomplete or dependent on an unsuitable controller.

The new tests use Innocent Blood in a three-player game. On unmodified master both tests fail when the second opponent's copied match controller is asked to choose a sacrifice. With this change, every opponent receives a scoped simulation controller, both choose their least valuable creature, repeated simulations produce the same score, and `SpellAbilityPicker` selects the beneficial play.

## Scope

This PR changes only simulator controller setup. It does not add play-selection heuristics, safety checks, caching, threat scoring, or board-state scoring rules. It is intended as the first small correctness step following the discussion in #10559 and #10576.

## Tests

- `GameSimulatorMultiplayerTest`: 2 passed with this change; 2 failed on unmodified master at the expected copied-controller assertion
- `GameSimulatorMultiplayerTest`, `SpellAbilityPickerSimulationTest`, and `AiSacrificeDecisionTest`: 140 passed, 0 failed

### DIFF
--- a/forge-ai/src/main/java/forge/ai/simulation/GameSimulator.java
+++ b/forge-ai/src/main/java/forge/ai/simulation/GameSimulator.java
@@ -53,7 +53,7 @@ public class GameSimulator {
             debugLines = origLines;
             Game copyOrigGame = copier.makeCopy();
             Player copyOrigAiPlayer = copyOrigGame.getPlayers().get(1);
-            resolveStack(copyOrigGame, copyOrigGame.getPlayers().get(0));
+            resolveStack(copyOrigGame, copyOrigAiPlayer);
             origScore = eval.getScoreForGameState(copyOrigGame, copyOrigAiPlayer);
         }
 
@@ -225,9 +225,7 @@ public class GameSimulator {
         }
 
         if (resolve) {
-            // TODO: Support multiple opponents.
-            Player opponent = aiPlayer.getWeakestOpponent();
-            resolveStack(simGame, opponent);
+            resolveStack(simGame, aiPlayer);
         }
 
         // TODO: If this is during combat, before blockers are declared,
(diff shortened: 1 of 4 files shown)


forge-ai/src/main/java/forge/ai/simulation/GameSimulator.java
```diff
@@ -260,11 +258,9 @@ public Score simulateSpellAbility(SpellAbility origSa, GameStateEvaluator eval,
         return score;
     }
 
-    public static void resolveStack(final Game game, final Player opponent) {
-        // TODO: This needs to set an AI controller for all opponents, in case of multiplayer.
```
tool4ever, line 264: Imo your tests only fail because AI are skipped, as already iterated on here:
https://github.com/Card-Forge/forge/pull/10567#discussion_r3173970779

we can think about doing them anyway, though only if I see a real bug to avoid extra costs
(+ we're trying to avoid PlayerController methods depending too much on outside information)

Madwand99: Yep, you are right. This is tricky... closing this draft.

PR #11436, opened 2026-07-28
AI: validate heuristic plays with depth-zero simulation

Madwand99:
## Summary

Adds a lightweight one-play safety check for the heuristic AI.

After the existing heuristics select a spell, ability, or land play, the checker simulates only that proposed action and compares the resulting board-state score with the current score. The play is rejected when the simulation shows that the AI would lose or worsen its position.

This supplements the existing heuristics rather than replacing them.

## Behavior

The safety check covers:

- Normal spell and ability selection.
- Land plays.
- Optional spells cast without paying their mana cost.
- Actions offered while a parent ability is resolving.
- Trying another candidate after the preferred candidate is rejected.

Mandatory casts are never vetoed. The existing simulation AI is also left unchanged.

Unsupported situations, such as responding to a nonempty stack, fail open and preserve the heuristic decision.

## Simulation Changes

- Adds a configurable recursion depth to `SimulationController`.
- Uses depth zero for the safety check, ensuring only the proposed action is evaluated.
- Preserves free-cast and play-effect metadata when mapping abilities into the copied game.
- Leaves the existing simulation depth of three unchanged for normal simulation AI.

This change does not add exception suppression, caching, multiplayer controller replacement, or multi-action candidate ranking.

## Motivation

Card interactions can make otherwise reasonable heuristic decisions harmful in ways that are difficult to encode individually. Examples include:

- Wheel effects giving an opposing Xyris a large number of tokens.
- Draw-trigger chains causing the AI to kill itself.
- Destroying a creature into Grave Pact and losing a much more valuable creature.
- Playing a land into lethal Zo-Zu damage.

A single simulated action allows the normal game engine and board-state evaluator to account for these interactions without adding card-specific heuristics.

## Tests

Adds 12 focused regression tests covering:

- Lethal and nonlethal Xyris interactions.
- Optional and mandatory free casts.
- Resolving and non-resolving stack states.
- Beneficial plays that must not be over-blocked.
- Candidate fallback after an unsafe play.
- Equal-score actions.
- Depth-zero simulation.
- Favorable and unfavorable Grave Pact exchanges.
- Lethal land plays into Zo-Zu.

Thanks to Codex for help on code and tests!


forge-ai/src/main/java/forge/ai/simulation/GameSimulator.java
```diff
@@ -141,6 +141,12 @@ private SpellAbility findSaInSimGame(final SpellAbility sa) {
             result = saMatcher(GameActionUtil.getAlternativeCosts(cSa, aiPlayer, true), desc);
         }
 
+        if (result != null && sa.hasParam("WithoutManaCost") && !result.hasParam("WithoutManaCost")) {
```
tool4ever, line 160: looks pretty clean

I just don't understand this part

Madwand99: findSaInSimGame() reconstructs the action from the copied card’s abilities. For cascade/discover-style casts, the incoming ability is a temporary copyWithNoManaCost() ability that does not exist on the copied card, so description matching can return the normal printed spell instead. This restores the free-cast form and CastFromPlayEffect flag so simulation does not try to pay mana or fail to cast. testAiDeclinesOptionalFreeWheelIntoXyris covers this path.

Madwand99:
I’ve made a broader follow-up to address simulation fidelity issues uncovered while testing the safety check.
The copied ability now preserves optional costs, kicker counts, X values, selected modes, targets, and divided allocations. The evaluation also handles phased permanents, mandatory delayed returns, multiplayer teams, and explicit costs more accurately. Without these changes, the checker could reject valid plays because the simulation was evaluating a different choice or treating a temporary absence as a permanent loss.
The checker remains depth-zero and fails open when it cannot produce a reliable result, so existing heuristics still make the primary decision.

tool4ever:
ugh, no that has just turned it into a giant mess again, I'm not reviewing that 😩

1. I'll revert last commit
2. I'll add opt-in Hybrid mode that uses this
3. Then afterwards you can do follow-up PRs for *each* problem to improve it

Madwand99:
If you prefer. I'll never understand the desire to split a PR up like this personally. Without the latest commit, the AI will just not play many cards because it can't simulate their effects properly.

tool4ever:
Appreciate you trying to get everything right directly...
but the effort to review whole PR doesn't just grow linearly if you double its size with 300+ lines in a single commit
that's especially true for your AI which:
- rarely leaves code comments
- doesn't split changes up better
- has little regard for technical debt instead of adding logic at inappropriate places
- starts introducing weird bloat the larger the PR

are you really expecting me to sort through that to find the logic parts that belong together?

so I'm only trying to make sure some of your changes actually make it into production

tool4ever:
> I’ve made a broader follow-up to address simulation fidelity issues uncovered while testing the safety check. The copied ability now preserves optional costs, kicker counts, X values, selected modes, targets, and divided allocations. The evaluation also handles phased permanents, mandatory delayed returns, multiplayer teams, and explicit costs more accurately. Without these changes, the checker could reject valid plays because the simulation was evaluating a different choice or treating a temporary absence as a permanent loss. The checker remains depth-zero and fails open when it cannot produce a reliable result, so existing heuristics still make the primary decision.

sounds like valid concerns but the first part seems like a general problem between the copying, so it will need a cleanly isolated helper instead

tool4ever:
Ok, added GUI support (at least on Desktop)
tbh not my cleanest work either but it's still somewhat experimental anyway

tool4ever (approved):

PR #11440, opened 2026-07-29
Speed up simulation AI evaluation

omraj21:
Two changes to the cost of scoring a game state in the simulation AI. Neither
changes which move the AI picks.

**Cache `AiDeckStatistics` instead of rebuilding the deck per evaluation**

`GameStateEvaluator.getScoreForGameStateImpl` calls `AiDeckStatistics.fromPlayer`
on every scored state to feed `evalManaBase`. For a player with a registered deck
that goes through `fromDeck`, which calls `Card.fromPaperCard` for every card in
the decklist, so scoring one state rebuilt ~60-100 `Card` objects.

The statistics only depend on the decklist (CMC, colour pips, land count), which
does not change during a game. Simulation copies share the same `Deck` instance
(`GameCopier.clonePlayer` passes `p.getDeck()` straight through), so keying the
existing `AiCache` on the deck also hits for every simulated state rather than
just the root one.

Measured on a 60-card deck with a small board, averaged over 2000 calls:

| | before | after |
|---|---|---|
| `AiDeckStatistics.fromPlayer` | 0.1513 ms | 0.0030 ms |

That took it from ~10% of a `GameStateEvaluator` evaluation to ~0.2%.

The empty-deck path used by tests derives stats from the player's current cards,
which *do* change during a game, so it is deliberately left uncached. `AiCache`
is already cleared once per AI decision in `AiController.chooseSpellAbilityToPlay`.

**Only verify the game copy score when assertions are enabled**

`GameSimulator`'s constructor called `ensureGameCopyScoreMatches` on every
simulator built with no target phase. That re-scores the copied game with the
evaluator's debug mode switched on, so it costs a second full evaluation plus the
debug string building it implies, purely to assert the copy is faithful.

It is a developer self-check rather than gameplay logic, so it is now gated on
assertions being enabled. Maven Surefire enables assertions by default (verified
with a probe test), so every test still exercises the check and a game copy bug
still fails loudly in CI, while shipped builds skip the duplicated work.

Verified: forge.ai.** test suite, 245 tests, 0 failures. Full `mvn -U -B clean test`
across all 12 modules passes.

---
Written with the help of GitHub Copilot CLI; the commits carry a `Co-authored-by`
trailer for it.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiDeckStatistics.java
+++ b/forge-ai/src/main/java/forge/ai/AiDeckStatistics.java
@@ -117,7 +117,11 @@ public class AiDeckStatistics {
             return fromCards(cardlist);
         }
 
-        return fromDeck(deck, player);
+        // These statistics are a pure function of the decklist, which doesn't change during a game.
+        // Simulation copies share the same Deck instance (see GameCopier.clonePlayer), so caching on
+        // it also avoids rebuilding every card of the deck for each simulated state that gets scored.
+        return AiCache.getCached("aiDeckStatistics", () -> fromDeck(deck, player),
+                List.of(AiCache::identity), deck);
     }
 
 }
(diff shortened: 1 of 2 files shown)


Hanmac:
@tool4ever imo the MR should be split.

The part with AiDeckStatistics looks okay

The part with assertions check looks confusing?

tool4ever (approved):
both okay for me, the comment explains it clearly

I'll have to think about ways to give AiCache different scopes for the future

PR #11441, opened 2026-07-29
Stop the AI paying real costs for abilities that do nothing

omraj21:
Two abilities were mapped to `AlwaysPlayAi`, whose entire logic is
`return WillPlay` with no look at the board. In both cases the engine can resolve
the ability to a no-op (or worse), while the activation cost sacrifices a permanent.

**Monarch**

`GameAction.becomeMonarch` returns immediately when the chosen player is already
the monarch, so activating one of these abilities while holding the crown does
nothing. Throne of the High City taps and sacrifices itself for {4}, King
Solomon's Frogs exiles itself, and Tchaka, Venerable King exiles itself from the
graveyard, so the AI was throwing a permanent away for no effect. With the old
mapping the added test shows Throne of the High City going from one copy on the
battlefield to zero.

`BecomeMonarchAi` declines only when every player the ability would crown is
already the monarch. Targeted versions are deliberately left alone: cards like
Jared Carthalion, True Heir and Garland, Royal Kidnapper hand the crown to an
opponent on purpose to turn on another ability.

Initiative is intentionally not changed - `GameAction.takeInitiative` documents
that "You can take the initiative even if you already have it" and still runs the
trigger, so always-play is correct there.

**Rooms**

`UnlockDoor` with `Mode$ LockOrUnlock` is wrong in two ways. When every door of
the targeted Room is already unlocked, `UnlockDoorEffect`'s `case 0` branch can
only call `lockRoom`, so the ability shuts off one of the AI's own Rooms. Keys to
the House pays {3}, taps and sacrifices itself for that ability.

When exactly one door is locked, the effect offers both doors and locks the one
that is still open. That choice went through `SpellAbilityAi.chooseCardState`,
whose base implementation prints a "default implementation is used" warning and
returns the first element, so which door the AI got was down to state ordering.

`UnlockDoorAi` declines when no reachable Room has a locked door, and picks a
locked door when asked to choose. `Mode$ Unlock` and `Mode$ ThisDoor` can only
ever unlock, so they keep the previous behaviour. Affects Keys to the House and
Marina Vendrell.

Each fix comes with a test that fails on the old mapping.

Verified: forge.ai.** test suite, 251 tests, 0 failures. Full `mvn -U -B clean test`
across all 12 modules passes.

---
Written with the help of GitHub Copilot CLI; the commits carry a `Co-authored-by`
trailer for it.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/SpellApiToAi.java
+++ b/forge-ai/src/main/java/forge/ai/SpellApiToAi.java
@@ -34,7 +34,7 @@ public enum SpellApiToAi {
             .put(ApiType.AssembleContraption, AssembleContraptionAi.class)
             .put(ApiType.AssignGroup, AssignGroupAi.class)
             .put(ApiType.Balance, BalanceAi.class)
-            .put(ApiType.BecomeMonarch, AlwaysPlayAi.class)
+            .put(ApiType.BecomeMonarch, BecomeMonarchAi.class)
             .put(ApiType.BecomesBlocked, BecomesBlockedAi.class)
             .put(ApiType.BidLife, BidLifeAi.class)
             .put(ApiType.BlankLine, AlwaysPlayAi.class)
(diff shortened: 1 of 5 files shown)


Hanmac:
Probably should do separate PRs for Monarch/Unlock

omraj21:
Splitting this into two PRs so each fix can be reviewed on its own, since they are unrelated beyond both being AlwaysPlayAi mappings:

- #11446 - Stop the AI crowning itself when it is already the monarch (BecomeMonarch)
- #11447 - Stop the AI locking its own Room for no reason (UnlockDoor)

Same commits, no content changes apart from correcting the test counts now that each branch is measured on its own.

PR #11446, opened 2026-07-29
Stop the AI crowning itself when it is already the monarch

omraj21:
`BecomeMonarch` is mapped to `AlwaysPlayAi`, whose entire logic is `return WillPlay` — it never looks at the board.

`GameAction.becomeMonarch` returns immediately when the chosen player is already the monarch:

```java
public void becomeMonarch(final Player p, final String set) {
    final Player previous = game.getMonarch();
    if (p == null || p.equals(previous)) {
        return;
    }
    ...
```

So activating one of these abilities while already holding the crown does nothing at all.

That is not merely a wasted activation, because several of these abilities cost a permanent:

| card | cost |
|---|---|
| Throne of the High City | `{4}, {T}, Sacrifice` |
| King Solomon's Frogs | `{3}, {T}, Exile` |
| Tchaka, Venerable King | `{3}, Exile from graveyard` |

The AI was throwing a permanent away for no effect. The added test shows Throne of the High City going from one copy on the battlefield to zero under the old mapping.

## The fix

`BecomeMonarchAi` declines only when **every** player the ability would crown is already the monarch.

Targeted versions are deliberately left alone. Cards like Jared Carthalion, True Heir and Garland, Royal Kidnapper hand the crown to an *opponent* on purpose in order to turn on another ability, so always-play is still right for them.

## Initiative is intentionally not changed

The initiative mechanic looks identical but is not. `GameAction.takeInitiative` explicitly documents:

```java
// You can take the initiative even if you already have it
```

and still runs the `TakesInitiative` trigger in that case, so `AlwaysPlayAi` remains correct for `TakeInitiative`.

## Verification

Three tests, each of which fails on the old mapping or covers a case that must keep working:

- `doesNotCrownItselfTwice` — the regression; fails with `expected:<1> but was:<0>` without the fix
- `takesTheCrownFromAnOpponent` — still activates when an opponent holds the crown
- `claimsAnUnclaimedCrown` — still activates when nobody is the monarch

`forge.ai.**` test suite: 248 tests, 0 failures. Full `mvn -U -B clean test` across all 12 modules passes.

---
*This PR was previously part of #11441, which bundled it with an unrelated `UnlockDoor` fix. Split out so each can be reviewed on its own.*

### DIFF
--- a/forge-ai/src/main/java/forge/ai/SpellApiToAi.java
+++ b/forge-ai/src/main/java/forge/ai/SpellApiToAi.java
@@ -34,7 +34,7 @@ public enum SpellApiToAi {
             .put(ApiType.AssembleContraption, AssembleContraptionAi.class)
             .put(ApiType.AssignGroup, AssignGroupAi.class)
             .put(ApiType.Balance, BalanceAi.class)
-            .put(ApiType.BecomeMonarch, AlwaysPlayAi.class)
+            .put(ApiType.BecomeMonarch, BecomeMonarchAi.class)
             .put(ApiType.BecomesBlocked, BecomesBlockedAi.class)
             .put(ApiType.BidLife, BidLifeAi.class)
             .put(ApiType.BlankLine, AlwaysPlayAi.class)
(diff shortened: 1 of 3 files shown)


Agetian (approved):

Agetian:
Looks good to me

PR #11448, opened 2026-07-29
Stop the AI seeking a card its library cannot supply

omraj21:
`Seek` is mapped to `AlwaysPlayAi`, whose entire logic is `return WillPlay` — it never looks at the board.

`SeekEffect` builds a pool from the library filtered by the seek type, and when that pool is empty it simply skips the type:

```java
if (pool.isEmpty()) {
    if (notify.length() != 0) notify.append("\r\n");
    notify.append(Localizer.getInstance().getMessage("lblSeekFailed", seekType));
    continue; // can't find if nothing to seek
}
```

The cost has already been paid by that point. Runecarved Obelisk pays it with the card itself:

```
A:AB$ Seek | Cost$ T Sac<1/CARDNAME> | Type$ Card.cmcEQY | ...
SVar:Y:Count$ValidLibrary Card.YouOwn+cmcLEX$GreatestCardManaCost
SVar:X:Count$CardCounters.CHARGE
```

With no charge counters nothing in the library can match, so the AI sacrificed the artifact for nothing. The added test shows it going from one copy on the battlefield to zero under the old mapping.

## The fix

`SeekAi` mirrors how `SeekEffect` builds its pool (including the `DefinedCards` variant) and declines when no seek type can find anything.

## Cost-dependent types are deliberately left alone

Spawning Pod seeks `Creature.cmcEQX` where `SVar:X:Sacrificed$CardManaCost/Plus.1` — the filter depends on **which creature the cost sacrifices**, so it cannot be resolved before the cost is paid. Guessing there would make the AI stop using the card entirely, which would be a worse bug than the one being fixed.

So any type whose `X`/`Y`/`Z` resolves to a `Sacrificed$` SVar is treated as findable, preserving the old behaviour. I verified this by running Spawning Pod activations with and without the change and getting identical results.

## Verification

- `doesNotSacrificeItselfSeekingAnEmptyPool` — the regression; fails with `Runecarved Obelisk should not have been sacrificed expected:<1> but was:<0>` without the fix
- `seeksWhenTheLibraryCanAnswer` — with 6 charge counters it still sacrifices and draws, so the guard does not over-decline

`forge.ai.**` test suite: 247 tests, 0 failures. Full `mvn -U -B clean test` across all 12 modules passes.

---

### DIFF
--- a/forge-ai/src/main/java/forge/ai/SpellApiToAi.java
+++ b/forge-ai/src/main/java/forge/ai/SpellApiToAi.java
@@ -182,7 +182,7 @@ public enum SpellApiToAi {
             .put(ApiType.Sacrifice, SacrificeAi.class)
             .put(ApiType.SacrificeAll, SacrificeAllAi.class)
             .put(ApiType.Scry, ScryAi.class)
-            .put(ApiType.Seek, AlwaysPlayAi.class)
+            .put(ApiType.Seek, SeekAi.class)
             .put(ApiType.SetInMotion, AlwaysPlayAi.class)
             .put(ApiType.SetLife, LifeSetAi.class)
             .put(ApiType.SetState, SetStateAi.class)
--- a/forge-ai/src/main/java/forge/ai/ability/SeekAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/SeekAi.java
@@ -1,0 +1,91 @@
+package forge.ai.ability;
+
+import java.util.Arrays;
+import java.util.List;
+
+import com.google.common.collect.Lists;
+
+import forge.ai.AiAbilityDecision;
+import forge.ai.AiPlayDecision;
+import forge.ai.SpellAbilityAi;
+import forge.game.ability.AbilityUtils;
+import forge.game.card.Card;
+import forge.game.card.CardCollection;
+import forge.game.card.CardLists;
+import forge.game.player.Player;
+import forge.game.spellability.SpellAbility;
+import forge.game.zone.ZoneType;
+
+/**
+ * Seek picks at random from the cards in the library that match its type, and SeekEffect skips a
+ * type whose pool is empty after the cost has already been paid. Runecarved Obelisk taps and
+ * sacrifices itself to seek, so seeking a type the library cannot supply throws the artifact away
+ * for nothing.
+ */
+public class SeekAi extends SpellAbilityAi {
+
+    @Override
+    protected AiAbilityDecision canPlay(Player aiPlayer, SpellAbility sa) {
+        for (Player seeker : getSeekers(aiPlayer, sa)) {
+            for (String seekType : getSeekTypes(sa)) {
+                if (canFindSomething(seeker, seekType, sa)) {
+                    return new AiAbilityDecision(100, AiPlayDecision.WillPlay);
+                }
+            }
+        }
+        return new AiAbilityDecision(0, AiPlayDecision.DoesntImpactGame);
+    }
+
+    private static List<Player> getSeekers(Player aiPlayer, SpellAbility sa) {
+        if (sa.usesTargeting()) {
+            // targets aren't chosen yet at this point, so assume we can aim at ourselves
+            return Lists.newArrayList(aiPlayer);
+        }
+        List<Player> defined = Lists.newArrayList();
+        for (String def : sa.getParamOrDefault("Defined", "You").split(" & ")) {
+            defined.addAll(AbilityUtils.getDefinedPlayers(sa.getHostCard(), def, sa));
+        }
+        return defined.isEmpty() ? Lists.newArrayList(aiPlayer) : defined;
+    }
+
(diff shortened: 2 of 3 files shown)


Agetian (approved):

jamincollins:
## Findings
#### 1. dependsOnCostPayment heuristic is fragile for indirect SVar references
SeekAi.dependsOnCostPayment checks `sa.getSVar(svar).startsWith("Sacrificed$")`. For Spawning Pod, `SVar:X` is defined as `Sacrificed$CardManaCost/Plus.1`. The SVar value starts with `Sacrificed$`, so the heuristic works. However, if the SVar is defined as a calculation that references `Sacrificed$` indirectly, e.g. `SVar:X:Count$Card$...` where the count itself depends on sacrificed card, the startswith check will miss it. The heuristic only inspects the raw SVar string, not evaluated value.
Test case:
```java
SpellAbility sa = new SpellAbility();
sa.setSVar("X", "Sacrificed$CardManaCost+1");
String seekType = "Creature.cmcEQX";
boolean depends = SeekAi.dependsOnCostPayment(seekType, sa); // true
// If SVar defined as "Sacrificed$CardManaCost" via another SVar, startswith may fail
```
Impact: AI may incorrectly assume findable for types that actually depend on cost, leading to premature decline of viable seeks.

#### 2. getSeekers assumes self-target when usesTargeting
When `sa.usesTargeting()` is true, SeekAi returns only `aiPlayer` as seeker, assuming targets aren't chosen yet. For Seek abilities that target an opponent's library, e.g., `Seek` with `TargetType=Player`, the AI will evaluate its own library instead of the intended target. No check for `TargetType` or actual target selection.
Test case:
```java
SpellAbility sa = new SpellAbility();
sa.setUsesTargeting(true);
sa.setParam("TargetType", "Player");
// getSeekers returns aiPlayer, ignoring actual target
```
Impact: Incorrect pool evaluation for targeted seeks.

#### 3. pool.isEmpty check before DefinedCards evaluation
canFindSomething creates pool via `AbilityUtils.getDefinedCards` when `DefinedCards` param exists. If DefinedCards resolves to empty collection, pool.isEmpty returns false and AI declines. However SeekEffect may still have fallback to library? The code mirrors SeekEffect but SeekEffect's DefinedCards path is used directly; if defined cards empty, SeekEffect will skip that type. Acceptable but edge case where DefinedCards points to a zone that is temporarily empty but will be populated later.

#### 4. seekType.equals("Card") short circuits without validation
If seekType is "Card", canFindSomething returns true if pool not empty. No validation of card legality (e.g., card may be in library but cannot be sought due to other restrictions). SeekEffect does random pick from pool, so any card is fine. OK.

#### 5. Mapping Seek to SeekAi changes AI behavior globally
Previously `ApiType.Seek` mapped to `AlwaysPlayAi`. Now SeekAi may decline activations where pool is empty. This is intended, but cards like `Seek the Beast` with no cost may still be declined if library lacks matching type, which is correct. However, cards with `Seek` as part of a larger ability chain may have side effects when declined. Backwards compatibility is changed intentionally.

#### 6. getSeekTypes splits on comma without trimming
`sa.getParam("Types").split(",")` yields raw strings with possible whitespace. `CardLists.getValidCards` may treat `" Creature"` differently from `"Creature"`. No trim applied.
Test case:
```java
sa.setParam("Types", " Creature, Land ");
List<String> types = SeekAi.getSeekTypes(sa);
// types = [" Creature", " Land "]
```
Impact: Potential false negative when types contain whitespace.

#### 7. SVar lookup in dependsOnCostPayment may trigger side effects
`sa.getSVar(svar)` may invoke text change effects and parsing on each canPlay evaluation. Called for each seek type per seeker. Could cause repeated expensive evaluation during AI decision loops.

#### 8. canPlay returns WillPlay on first findable type
If seek ability has multiple types (e.g., `Types=Creature,Land`), AI will play if any type findable. SeekEffect processes types in order and picks first findable pool. Matching behavior is fine.

### Performance and security
#### 9. CardLists.getValidCards called repeatedly
For each seek type, pool is filtered via `CardLists.getValidCards`. No caching of pool validity per seeker. With large libraries, repeated filtering may be costly. Acceptable for current scale.

## Verdict
Comment
The PR correctly prevents AI from sacrificing Runecarved Obelisk when library cannot supply a card, and tests verify both empty and non-empty cases. The dependsOnCostPayment heuristic is reasonable for Spawning Pod but fragile for indirect SVar references, and getSeekers assumes self-target for targeting seeks. Whitespace handling in Types param could cause false negatives. Overall change is beneficial with minor refinements needed.

## Recommended Fixes
1. Trim seek types after split to avoid whitespace false negatives.
2. Make dependsOnCostPayment recursively check SVar values for Sacrificed$ substring, not just startswith.
3. Review getSeekers to handle targeted seeks where target player is known or can be inferred.
4. Cache pool validity per seeker to avoid repeated filtering.

---

PR #11456, opened 2026-07-29
Detect attack-punisher enchantments by trigger instead of by name

omraj21:
`AiAttackController` only recognised **Revenge of Ravens**, by name, when working out whether attacking would cost the AI life:

```java
// TODO: detect Revenge of Ravens by the trigger instead of by name
revengeOfRavens = !CardLists.filter(player.getCardsIn(ZoneType.Battlefield),
        CardPredicates.nameEquals("Revenge of Ravens")).isEmpty();
```

Several cards are printed with the same trigger, and the AI walked straight into all of them:

| card | trigger |
|---|---|
| Hissing Miasma | Whenever a creature attacks you, its controller loses 1 life |
| Blood Reckoning | same, for creatures you don't control |
| Marchesa's Decree | same, alongside the monarch ability |

This does what the TODO asks and looks for the trigger: an `Attacks` trigger on the defending player's battlefield whose effect is `LoseLife` aimed at `TriggeredAttackerController`.

With the old name check, the added test shows the AI sending all three attackers into lethal life loss:

```
holdsBackAgainstHissingMiasma  AI should not attack into lethal life loss expected:<0> but was:<3>
holdsBackAgainstBloodReckoning AI should not attack into lethal life loss expected:<0> but was:<3>
```

## What deliberately does not count

Cards that punish the attacking *creature* rather than its controller — Circle of Flame dealing 1 damage to the attacker, Raking Canopy, Palazzo Archers — are excluded. The creature pays that cost, not the player, and normal combat evaluation already accounts for it. There's a test asserting the AI still attacks into Circle of Flame, so the check can't quietly widen to cards it shouldn't cover.

## Two smaller corrections in the same logic

- The life total is compared against the actual life lost per attacker rather than assuming 1, so a card draining 2 each is handled correctly.
- The "not worth attacking" filter uses that same number instead of a hardcoded `getNetCombatDamage() <= 1`.

Amounts that aren't plain numbers are skipped rather than guessed at, since overestimating here would make the AI refuse to attack for no reason.

## Verification

5 tests: no-punisher control, Revenge of Ravens (the case the name check already covered), Hissing Miasma, Blood Reckoning, and the Circle of Flame exclusion. Two of them fail on the old code.

Full `mvn -U -B clean test`: 12 modules, 308 tests, 0 failures, run three times.

---

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiAttackController.java
+++ b/forge-ai/src/main/java/forge/ai/AiAttackController.java
@@ -20,6 +20,7 @@ package forge.ai;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import org.apache.commons.lang3.StringUtils;
 import forge.ai.ability.AnimateAi;
 import forge.game.GameEntity;
 import forge.game.ability.AbilityUtils;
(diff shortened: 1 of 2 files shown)


jamincollins:
## Findings
#### 1. Attacked param not validated against defender
The `lifeLostPerAttacker` method at `forge-ai/src/main/java/forge/ai/AiAttackController.java:1791-1798` iterates over all triggers with `TriggerType.Attacks` and `hasParam("Attacked")` on the defending player's battlefield and sums their life loss without verifying that the trigger's `Attacked` param actually refers to the current defender.

The trigger is only relevant when the attacked entity matches the defender we are evaluating. For example, `Revenge of Ravens` has `Attacked$ You,Planeswalker.YouCtrl`. If the AI is evaluating an attack against one of the defending player's creatures, the trigger would not fire, but the code still counts it.

Test case:
```java
// Setup: defending player controls Revenge of Ravens
// AI considers attacking a creature controlled by defender
GameEntity defender = creatureControlledByDefender;
int loss = AiAttackController.lifeLostPerAttacker(defender);
// Expected: 0, because the Attacks trigger requires Attacked$ You
// Actual: 1, because the code sums the trigger regardless of Attacked target
Assert.assertEquals("Life loss should be zero when attacking a creature", 0, loss);
```
Impact: The AI may incorrectly hold back from attacking creatures when the punisher only applies to attacks against the player/planeswalker, leading to overly conservative play.

#### 2. Defender type mismatch causes false positive life loss
When `defender` is a `Card` instance, `lifeLostPerAttacker` resolves the defending player via `card.getController()` at line 1784. The subsequent trigger scan does not check whether the trigger's `Attacked` param would actually match the card versus the player.

Test case:
```java
// Defending player has Hissing Miasma
// AI evaluates attack against defending player's creature
Player defendingPlayer = game.getPlayers().get(0);
Card defenderCreature = addCard("Grizzly Bears", defendingPlayer);
int loss = AiAttackController.lifeLostPerAttacker(defenderCreature);
// Hissing Miasma triggers on Attacked$ You, not Attacked$ Card
// Current code returns 1, should return 0
Assert.assertEquals("Creature attacks should not trigger player-targeted punisher", 0, loss);
```
Impact: Null safety is fine, but logical boundary is wrong: the AI treats player-targeted punishers as applicable to creature attacks.

#### 3. Queue mutation side effect in doRevengeOfRavensAttackLogic
`doRevengeOfRavensAttackLogic` at `forge-ai/src/main/java/forge/ai/AiAttackController.java:1741-1767` removes attackers from the `attackersLeft` queue in place via `attackersLeft.removeAll(attUnsafe)`. The queue is passed in from the caller and may be reused for subsequent defender evaluations.

Test case:
```java
Queue<Card> attackers = new LinkedList<>(allAttackers);
boolean first = controller.doRevengeOfRavensAttackLogic(playerDefender, attackers, 0, null);
// attackers is now mutated
boolean second = controller.doRevengeOfRavensAttackLogic(otherDefender, attackers, 0, null);
// Second evaluation uses already-filtered list, not original set
```
Impact: Backwards compatibility risk: if the method is called multiple times per combat phase with the same queue reference, earlier filtering leaks into later decisions.

#### 4. Multiple triggers summed without per-trigger Attacked validation
The current implementation sums `lifeLossFromTriggerEffect` for every matching trigger. If a player controls both a player-targeted punisher and a planeswalker-targeted punisher, the sum is correct for attacks against the player, but for attacks against a planeswalker the player-targeted trigger should not count.

Test case:
```java
// Defending player has Revenge of Ravens (Attacked$ You,Planeswalker.YouCtrl) and a custom trigger
// Custom trigger: Attacks, Attacked$ You, LoseLife 2
// AI evaluates attack against defending player's planeswalker
// Both triggers match hasParam("Attacked") and sum to 3
// Correct sum for planeswalker attack should be 1+2=3? Actually both trigger.
// Hmm.
```
The more concrete issue is that triggers with different Attacked targets are summed indiscriminately.

Impact: The heuristic may overestimate life loss when multiple punishers with different targeted scopes coexist.

### Performance and security
#### 5. Repeated battlefield scan on each attack evaluation
`lifeLostPerAttacker` scans the entire defending player's battlefield and iterates over all triggers for each defender evaluation. In multiplayer games with large boards, this is executed for each potential defender and each attack iteration.

Test case:
```java
// 4-player game, each player has 30 cards on battlefield with 2 triggers each
// Attack evaluation for 10 attackers against 3 defenders
// Operations: 3 defenders * 4 players * 30 cards * 2 triggers = 720 trigger checks per decision
```
Impact: No security risk, but performance scales O(N*M) per evaluation. Caching the per-player life loss per turn would reduce repeated work.

## Verdict
Request Changes

The trigger-based detection is a correct improvement over name-based checking, but the implementation counts triggers without verifying that the trigger's `Attacked` param matches the defender being evaluated. This can cause false positives when attacking creatures or when triggers target different entities. The queue mutation side effect also risks leaking state across evaluations.

## Recommended Fixes
1. Validate the trigger's `Attacked` param against the defender before summing life loss. For a `GameEntity` defender, only count triggers whose `Attacked` param resolves to that entity.
2. Avoid mutating the input `attackersLeft` queue; work on a copy or return the filtered set.
3. Cache `lifeLostPerAttacker` per defending player per combat phase to avoid repeated scans.
4. Add tests for defender-type mismatches, e.g., attacking a creature while a player-targeted punisher is present.

---

tool4ever (approved):

PR #11457, opened 2026-07-29
Rate limit and bound chat from remote clients

Tyrathalis:
## Summary

Chat from a remote client is accepted before login and rebroadcast to every peer, and nothing bounds how fast those messages arrive. A client sending chat in a loop costs the host's CPU, everyone's bandwidth and everyone's scrollback at once.

This adds a per-client token bucket, and makes the existing length bound on the rebroadcast configurable.

## Behaviour

- Token bucket sized for a person typing: 10 message burst, one refilled per second. Over-rate messages are dropped rather than queued, since queuing is what the sender wants.
- The rebroadcast length cap becomes a property instead of `LogSafe`'s default.

Both are policy rather than correctness, so both are tunable and both can be switched off outright:

| property | default | off |
|---|---|---|
| `forge.net.chatBurst` | 10 | `0` |
| `forge.net.chatRefillMillis` | 1000 | — |
| `forge.net.maxChatLength` | 512 | `0` |

These sit alongside the existing `forge.net.heartbeatTimeout`. The defaults are a guess at "generous for a human" — if they are wrong for how people actually use the lobby, they are one property away rather than a rebuild.

Switching the cap off still strips control characters from the rebroadcast: that part is a correctness property, not an anti-spam measure, so only the truncation is optional.

Values are read per call rather than into `static final` fields, because `Integer.getInteger` in a static initialiser is fixed at class-load and surefire shares one JVM, so a test could not otherwise vary them.

## Deliberately not included

A general per-message rate limit covering all inbound traffic. That risks throttling legitimate high-rate game traffic, and there is no measured baseline here for what a busy turn sends — guessing a number and stalling a real game is how this kind of change gets reverted.

## Tests

`ChatRateLimitTest`, 2 unit tests, no server and no socket: the bucket empties and refills, and the disable switch never refuses.

---

Portions authored with an AI assistant (Claude), reviewed by a human.

### DIFF
--- a/forge-gui-desktop/src/test/java/forge/net/ChatRateLimitTest.java
+++ b/forge-gui-desktop/src/test/java/forge/net/ChatRateLimitTest.java
@@ -1,0 +1,53 @@
+package forge.net;
+
+import forge.gamemodes.net.server.RemoteClient;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+/**
+ * Chat is accepted before login and rebroadcast to every peer, so a client
+ * that sends it in a loop costs the host and everyone else. The limit is a
+ * token bucket sized for a person typing, and switchable off.
+ */
+public class ChatRateLimitTest {
+
+    /** Far above any legitimate burst, low enough to fail fast. */
+    private static final int DRAIN_CEILING = 10_000;
+
+    @Test(timeOut = 30_000)
+    public void testChatIsRateLimitedAndRefills() throws Exception {
+        final RemoteClient client = new RemoteClient(null);
+
+        // Bounded, not "while (allowChatMessage())". An unbounded drain loop
+        // terminates only if the limiter works, so with the limiter removed --
+        // the case this exists to detect -- it spins forever and the run hangs
+        // instead of failing. A hang reads as "still running", which is worse.
+        int drained = 0;
+        while (drained < DRAIN_CEILING && client.allowChatMessage()) {
+            drained++;
+        }
+        Assert.assertTrue(drained > 0, "A burst should be allowed - people do type");
+        Assert.assertTrue(drained < DRAIN_CEILING,
+                "Bucket never emptied after " + DRAIN_CEILING + " messages; not rate limiting");
+
+        // One token per refill interval, so this cannot be shortened much.
+        Thread.sleep(1100);
+        Assert.assertTrue(client.allowChatMessage(),
+                "Allowance must come back - a rate limit that never refills is a mute");
+    }
+
+    @Test(timeOut = 30_000)
+    public void testLimitCanBeSwitchedOff() {
+        System.setProperty("forge.net.chatBurst", "0");
+        try {
+            final RemoteClient client = new RemoteClient(null);
+            for (int i = 0; i < DRAIN_CEILING; i++) {
+                Assert.assertTrue(client.allowChatMessage(),
+                        "Disabled limit must never refuse (refused at " + i + ")");
+            }
+        } finally {
+            System.clearProperty("forge.net.chatBurst");
+        }
+    }
+
+}
--- a/forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
+++ b/forge-gui/src/main/java/forge/gamemodes/net/server/FServerManager.java
@@ -71,6 +71,15 @@ public final class FServerManager implements IHasForgeLog {
 
     static final int HEARTBEAT_TIMEOUT_SECONDS = Integer.getInteger("forge.net.heartbeatTimeout", 45);
 
+    /**
+     * Chat is rebroadcast to every peer, so one very long message costs
+     * everyone. Read per call for the same reason as the burst in
+     * {@link RemoteClient}; zero or less switches truncation off.
+     */
+    private static int maxChatLength() {
+        return Integer.getInteger("forge.net.maxChatLength", 512);
+    }
+
     private static final int OUTBOUND_BUFFER_LOW_WATER = 64 * 1024;
     private static final int OUTBOUND_BUFFER_HIGH_WATER = 1024 * 1024;
     private static final int RECONNECT_TIMEOUT_SECONDS = 300;
@@ -1008,10 +1017,17 @@ public final class FServerManager implements IHasForgeLog {
                 if (client == null) {
                     return;
                 }
+                if (!client.allowChatMessage()) {
+                    return; // Over its rate; dropped rather than queued
+                }
                 // Strip control characters before echoing to other players: a
                 // carriage return lets one player paint fake system lines in
-                // everyone else's chat pane.
-                final String text = LogSafe.forDisplay(raw);
+                // everyone else's chat pane. Stripping is unconditional; only
+                // the length cap is policy, so zero or less strips without
+                // truncating rather than skipping the scrub.
+                final String text = maxChatLength() > 0
+                        ? LogSafe.forDisplay(raw, maxChatLength())
+                        : LogSafe.forDisplay(raw, Integer.MAX_VALUE);
                 String username = client.getUsername();
                 // Append (Host) indicator for the host player
                 if (client.getIndex() == 0) {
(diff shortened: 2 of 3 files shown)


MostCromulent:
No problem in principle, but what was the situation that led to the need for this?

Tyrathalis:
No inciting incident. I was looking at ways clients could overload each other (for 2df8aaa) and noticed there were no spam protections. Spam isn't much of an issue with no public rooms or matchmaking, but it seemed like an easy fix, so here it is if you're interested. The limits are off-switchable if the defaults turn out wrong for how people actually use the lobby.

MostCromulent:
@tool4ever I don't have a hard view against this, but it does seem to be solving a problem that doesn't actually exist in practice.

tool4ever:
ok thanks, yea this could become more interesting if lobbies were more broadly available like on some forks
though I'm not sure they're even using this chat system in that case 🤔
will consider it a bit more when my backlog is a bit less full

jamincollins:
## Findings

#### 1. Token bucket initialization uses class-load time value
Description: RemoteClient.chatTokens is initialized with `chatBurst()` which reads the system property at construction time. However, chatBurst() is read per call in allowChatMessage(), so the initial value may be stale if the property changes after construction. More importantly, chatLastRefill is initialized with `System.currentTimeMillis()` at construction, but the first call to allowChatMessage() will compute elapsed based on construction time, which is correct. However, if a client object is created long before first message (e.g., during server startup), the initial burst tokens are available immediately, which is fine. The real issue: if chatBurst() returns <=0 (limit disabled), chatTokens is initialized to 0 or negative, but allowChatMessage() returns true immediately without checking tokens. The initialization `private double chatTokens = chatBurst();` means when limit is disabled, chatTokens starts negative/zero, but since burst<=0 short-circuits, it's harmless. Still, the initialization couples construction-time config to runtime config.

Test case:
```java
System.setProperty("forge.net.chatBurst", "5");
RemoteClient c1 = new RemoteClient(null);
// Simulate property change after construction
System.setProperty("forge.net.chatBurst", "20");
// First allowChatMessage reads burst=20, but chatTokens was initialized to 5
// This means the first few calls will be limited by old burst value
```
Impact: Configuration changes at runtime are not fully reflected for existing clients until tokens refill to the new burst cap. For a server that changes properties via JMX/reload, this causes inconsistent limits.

#### 2. chatLastRefill not updated when elapsed == 0
Description: In allowChatMessage(), elapsed is computed as `now - chatLastRefill`. If elapsed == 0 (same millisecond), the refill block is skipped, so chatLastRefill remains unchanged. This is fine for a single call, but under high concurrency with same-millisecond calls, tokens may be consumed faster than intended because refill only happens when elapsed > 0. The synchronized method protects token decrement, but the refill calculation uses `chatTokens + elapsed / refill`. With elapsed == 0, no refill occurs, which is correct for same-millisecond calls. However, if system clock moves backwards, elapsed becomes negative, refill block is skipped, and chatLastRefill stays old, causing future elapsed to be larger (potentially huge). No handling for clock skew.

Test case:
```java
RemoteClient client = new RemoteClient(null);
// Drain burst
for(int i=0;i<10;i++) client.allowChatMessage();
// Simulate clock going backwards by 1 hour via mock or reflection
// Next call will see negative elapsed, skip refill, keep old chatLastRefill
// After clock corrects, elapsed is huge, granting massive burst
```
Impact: Clock skew can cause token bucket to grant huge bursts, defeating rate limit.

#### 3. maxChatLength zero/negative disables truncation but still strips controls
Description: FServerManager checks `maxChatLength() > 0` to decide whether to truncate. If maxChatLength <= 0, it passes `Integer.MAX_VALUE` to LogSafe.forDisplay, which still strips control characters via scrub. This is intentional per commit message. However, LogSafe.scrub uses `StringUtils.truncate(text, maxLength)` where maxLength is Integer.MAX_VALUE. StringUtils.truncate will still copy the entire string, which may be extremely large. If a malicious client sends a 100MB chat message, the server will allocate a 100MB String, strip controls, then broadcast to all peers. The length cap is policy, but the control stripping still processes the entire message, allowing a memory exhaustion attack.

Test case:
```java
String huge = "a".repeat(100_000_000);
LogSafe.forDisplay(huge, Integer.MAX_VALUE); // allocates 100MB string
```
Impact: Disabling truncation via property does not prevent OOM from huge messages.

#### 4. swapChannel does not reset token bucket state
Description: RemoteClient.swapChannel() is called on reconnect (FServerManager line ~1095). It updates channel and replies, but does not reset chatTokens or chatLastRefill. This means a reconnecting client retains its previous token bucket state, which could be partially drained. If a client was rate-limited and then reconnects, it immediately gets the remaining tokens, possibly allowing a burst that the limit intended to prevent. Conversely, if the client reconnects after a long pause, the elapsed calculation will refill based on chatLastRefill from before disconnect, potentially granting a large burst immediately after reconnect.

Test case:
```java
RemoteClient client = new RemoteClient(ch1);
for(int i=0;i<10;i++) client.allowChatMessage(); // drain
client.swapChannel(ch2);
// Tokens are still ~0, chatLastRefill is old
// Wait 1 second, allowChatMessage() refills correctly
// But if reconnect happens immediately, no tokens available
```
Impact: Reconnect behavior is inconsistent with user expectation; rate limit state should be per-connection or reset on reconnect for fairness.

#### 5. allowChatMessage() synchronized on instance may become contention point
Description: allowChatMessage() is synchronized on the RemoteClient instance. Message handling in FServerManager.MessageHandler is on Netty event loop thread. If multiple threads process messages for the same client (e.g., pipelining), synchronization is fine. However, the synchronized block is entered for every chat message, and the method reads system properties via Integer.getInteger() on every call. Integer.getInteger() is synchronized internally and reads System properties, which is relatively expensive. Combined with synchronized block, this may cause contention under high chat volume.

Test case:
```java
// Simulate 1000 concurrent chat messages
ExecutorService pool = Executors.newFixedThreadPool(100);
RemoteClient client = new RemoteClient(null);
for(int i=0;i<1000;i++) pool.submit(() -> client.allowChatMessage());
```
Impact: Latency spikes under chat flood; rate limiting intended to reduce load may add overhead.

### Performance and security
#### 6. Token bucket refill calculation uses double division
Description: `chatTokens + (double) elapsed / refill` uses double arithmetic for token accounting. This is fine for precision, but the subsequent check `if (chatTokens < 1)` and decrement `chatTokens -= 1` uses double comparison. With very small elapsed values, tokens may become fractional, and repeated decrements could accumulate floating point error, potentially allowing one extra message over time. More importantly, the refill uses `Math.min(burst, ...)`, so tokens never exceed burst. However, burst is int, refill is int, elapsed is long, division yields double. No issue for correctness, but performance could be improved with integer math.

Test case:
```java
// With refill=1000ms, elapsed=1ms, tokens increase by 0.001 each call
// After 1000 calls within 1ms, tokens may be slightly > burst due to rounding
```
Impact: Minimal, but potential for slight over-allowance.

#### 7. Chat rate limit applied before authentication
Description: FServerManager.MessageHandler checks `client.allowChatMessage()` for every MessageEvent, regardless of whether the client is logged in. The commit message says "Chat is accepted from a remote client before it has logged in". This means a client can send messages before login, consuming tokens. That's intentional per design. However, the rate limit is per RemoteClient instance, and clients are created when channel connects. A malicious client can open many connections, each with its own token bucket, bypassing the per-client limit. No global rate limit exists.

Test case:
```java
// Open 1000 TCP connections, send 10 messages each per second
// Total 10,000 messages/sec to server
```
Impact: DDoS via connection multiplication; rate limit is per-client not per-IP.

## Verdict
Request Changes

The PR adds a reasonable token-bucket rate limit for chat and configurable length bound, but several edge cases undermine the protection: token bucket state survives reconnects, clock skew can grant massive bursts, disabling length truncation allows OOM via huge messages, and the per-client limit is bypassable via connection multiplication. The implementation also couples initialization to construction time and uses expensive synchronized property reads.

## Recommended Fixes
1. Reset chatTokens and chatLastRefill in swapChannel() or document that reconnect preserves state. Consider resetting to full burst on reconnect for fairness.
2. Guard against clock skew by using monotonic time (System.nanoTime()) for elapsed calculation instead of System.currentTimeMillis().
3. Enforce a hard maximum message size before allocating the String for processing, even when truncation is disabled. Reject messages > some absolute cap (e.g., 1MB) at the Netty decoder level.
4. Cache chatBurst() and chatRefillMillis() values per client at construction or use volatile reads with a refresh interval, to avoid repeated Integer.getInteger() calls.
5. Consider per-IP rate limiting in addition to per-client to mitigate connection multiplication attacks.
6. Use integer arithmetic for token bucket to avoid floating point drift.

---

PR #11458, opened 2026-07-29
Prefer winning sooner and losing later

omraj21:
`GameStateEvaluator.getScoreForGameOver` returned a flat `Integer.MAX_VALUE` for a win and `Integer.MIN_VALUE` for a loss:

```java
private Score getScoreForGameOver(Game game, Player aiPlayer) {
    if (game.getOutcome().getWinningTeam() == aiPlayer.getTeam() || ...) {
        return new Score(Integer.MAX_VALUE);
    }
    return new Score(Integer.MIN_VALUE);
}
```

So every winning line scored identically and the simulation AI had no reason to take the faster one. The same applies in reverse when losing: a line that dies in six turns scored the same as one that dies right now, so nothing pushed the AI towards the delay that gives it a chance to draw out of trouble.

This discounts both ends by the turn number.

## Keeping the two existing meanings of the sentinels

Both extremes carry meaning elsewhere, so neither could simply be shifted:

**`SimulationController.shouldRecurse`** stopped searching on `== Integer.MAX_VALUE`, which a discounted win would no longer equal — that would have quietly turned off the "we already found a win, stop searching" optimisation. It now asks `GameStateEvaluator.isWinning(...)` and still stops on any win, whatever turn it lands on.

**`GameSimulator`** returns `Integer.MIN_VALUE` to mean *"this line could not be simulated"*, and `OnePlaySafetyChecker` treats that value as acceptable rather than fatal. A real loss is therefore scored one above the bottom of the range so a loss can never be mistaken for a failed simulation.

## Verification

5 tests, three of which fail with the old flat scoring:

- `winsSoonerScoreHigher` — *a win on turn 2 should beat a win on turn 12*
- `lossesLaterScoreHigher` — *dying on turn 12 should beat dying on turn 2*
- `aLossIsNotMistakenForAFailedSimulation` — *a real loss must score above the failed-simulation sentinel*

plus `everyWinBeatsEveryLoss` and `winsStillStopTheSearch` guarding the two invariants above.

Full `mvn -U -B clean test`: 12 modules, 308 tests, 0 failures. I saw one unrelated intermittent failure across ten full-suite runs on this branch; the new tests are pure arithmetic on the turn counter and passed 5/5 in isolation, and I could not reproduce the blip again on either this branch or master.

---


forge-ai/src/main/java/forge/ai/simulation/GameStateEvaluator.java
```diff
@@ -68,13 +68,30 @@ private static String cardToString(Card c) {
...
+    }
+
     private Score getScoreForGameOver(Game game, Player aiPlayer) {
+        // a game long enough to overflow the discount is not one we can reason about anyway
+        int turn = Math.min(Math.max(game.getPhaseHandler().getTurn(), 0), MAX_DISCOUNTED_TURNS);
         if (game.getOutcome().getWinningTeam() == aiPlayer.getTeam() ||
                 game.getOutcome().isWinner(aiPlayer.getRegisteredPlayer())) {
-            return new Score(Integer.MAX_VALUE);
+            // prefer winning sooner
+            return new Score(Integer.MAX_VALUE - turn * GAME_OVER_TURN_PENALTY);
```
tool4ever, line 90: that's interesting but since when do we look multiple turns in the future right now anyway?
shouldn't this consider recursion depth for the less complex combos more likely to succeed?

tool4ever (approved):

Issue #2167, opened 2022-12-25
Simulation: usage of addExtraKeywordCost can break targeting


tool4ever:
The problem is it creates a new SA the interceptor in `ComputerUtil.handlePlayingSpellAbility()` doesn't use.

Here's an example where the AI will refuse to kill you unless you remove its conspire pay options first:
[simtest.txt](https://github.com/Card-Forge/forge/files/10300652/simtest.txt)

PR #11460, opened 2026-07-30
Improve spell-choice fidelity in AI simulation

Madwand99:
## Summary

Improve `GameSimulator` so the ability resolved in the copied game preserves the choices already made for the original ability.

Previously, simulation could evaluate a different version of a spell by losing its optional costs, keyword-cost count, announced X value, selected modes, targets, or divided allocations.

## Changes

- Extract spell-choice translation into a focused `SpellAbilityChoiceCopier` helper.
- Preserve:
  - optional costs such as kicker
  - multikicker and replicate counts
  - announced X values
  - single and multiple selected modes
  - targets and divided allocations
  - free-cast state
- Make AI controllers respect preselected keyword-cost amounts.
- Pass the final post-cost/post-mode ability to deferred targeting.

The last point matters for abilities such as Replicate: adding the keyword cost can create a replacement ability, so targets must be applied to that final ability rather than the discarded original.

## Tests

Added focused simulation regressions for kicker, multikicker, Replicate with targets, X spells, modal spells, and divided allocations.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ComputerUtil.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtil.java
@@ -62,6 +62,7 @@ import forge.util.collect.FCollection;
 import org.apache.commons.lang3.StringUtils;
 
 import java.util.*;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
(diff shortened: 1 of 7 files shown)


forge-ai/src/main/java/forge/ai/ComputerUtil.java
```diff
@@ -111,7 +113,7 @@ public static boolean handlePlayingSpellAbility(final Player ai, SpellAbility sa
             return false;
         }
         if (chooseTargets != null) {
-            chooseTargets.run();
+            chooseTargets.accept(sa);
```
tool4ever, line 115: I believe this refactor might also fix #2167

if so nice work, I'll try testing later

tool4ever (approved):
thanks

PR #11462, opened 2026-07-30
Improve multiplayer and phasing evaluation in AI simulation

Madwand99:
## Summary

This improves `GameStateEvaluator` fidelity for multiplayer teams, phasing, and eliminated players.

Previously, the evaluator assumed every player other than the AI was an opponent. This caused allied cards, life, and permanents to reduce the AI's score. The evaluator now classifies resources using `isOpponentOf()` while preserving the existing weighting in normal one-on-one games.

Phased-out permanents are also included in long-term board and mana-base evaluation. Since they will phase back in, temporarily phasing a valuable permanent should not be scored as permanently losing it.

Finally, an AI player that has already lost now receives the terminal minimum score even when a multiplayer game continues.

## Tests

Added focused coverage confirming that:

- Allied permanents, life, and cards improve the team score.
- Phased permanents retain their long-term board and mana-base value.
- An eliminated player receives the minimum score in a continuing multiplayer game.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/simulation/GameStateEvaluator.java
+++ b/forge-ai/src/main/java/forge/ai/simulation/GameStateEvaluator.java
@@ -119,7 +119,7 @@ public class GameStateEvaluator {
 
     private Score getScoreForGameStateImpl(Game game, Player aiPlayer) {
         // TODO: try and reuse evaluateBoardPosition
-        int handScore = 0;
+        int score = 0;
         int aiCards = 0;
         int teammateCards = 0;
         int opponentCards = 0;
@@ -131,14 +131,14 @@ public class GameStateEvaluator {
         for (Player player : game.getPlayers()) {
             int cards = player.getCardsIn(ZoneType.Hand).size();
             if (player.isOpponentOf(aiPlayer)) {
-                handScore -= 4 * cards;
+                score -= 4 * cards;
                 opponentCards += cards;
                 opponentLife += player.getLife();
                 debugPrint("  Opponent " + (++opponents) + " life: -" + player.getLife());
             } else {
                 int fullValueCards = player.isUnlimitedHandSize()
                         ? cards : min(cards, player.getMaxHandSize());
-                handScore += cards + 4 * fullValueCards;
+                score += cards + 4 * fullValueCards;
                 teamLife += player.getLife();
                 teamPlayers++;
                 if (player == aiPlayer) {
@@ -156,18 +156,17 @@ public class GameStateEvaluator {
         }
         debugPrint("Their cards in hand: " + opponentCards);
         // TODO weight cards in hand more if opponent has discard or if we have looting or can bluff a trick
-        int score = handScore + 2 * teamLife / teamPlayers;
+        score += 2 * teamLife / teamPlayers;
         if (opponents > 0) {
             score -= 2 * opponentLife / opponents;
         }
 
         // evaluate mana base quality
         AiDeckStatistics statistics = AiDeckStatistics.fromPlayer(aiPlayer);
-        int availableManaScore = evalManaBase(game, aiPlayer, statistics);
+        int availableManaScore = evalManaBase(aiPlayer, statistics, false);
         int strategicManaScore = aiPlayer.getCardsIn(ZoneType.Battlefield, false).stream()
-                .anyMatch(c -> c.isPhasedOut() && phasesInNormally(c)
-                        && !c.getManaAbilities().isEmpty())
-                ? evalManaBase(game, aiPlayer, statistics, true) : availableManaScore;
+                .anyMatch(c -> c.isPhasedOut() && phasesInNormally(c) && !c.getManaAbilities().isEmpty())
+                ? evalManaBase(aiPlayer, statistics, true) : availableManaScore;
         int availableScore = score + availableManaScore;
         score += strategicManaScore;
         // TODO deal with opponents. Do we want to use perfect information to evaluate their manabase?
@@ -186,10 +185,10 @@ public class GameStateEvaluator {
                 continue;
             }
             int value = evalCard(game, aiPlayer, c);
-            int availableValue = phasedOut ? 0 : value;
+            int availableValue = value;
             // To make the AI hold-off on playing creatures before MAIN2 if they give no other benefits,
             // keep track of the score while treating summon sick creatures as having a value of 0.
-            if (gamePhase.isBefore(PhaseType.MAIN2) && c.isSick() && c.getController() == aiPlayer) {
+            if (phasedOut || (gamePhase.isBefore(PhaseType.MAIN2) && c.isSick() && c.getController() == aiPlayer)) {
                 availableValue = 0;
             }
             String str = cardToString(c);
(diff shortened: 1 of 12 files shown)


forge-ai/src/main/java/forge/ai/simulation/GameStateEvaluator.java
```diff
@@ -112,37 +115,48 @@ public Score getScoreForGameState(Game game, Player aiPlayer) {
...
+        int handScore = 0;
         int myCards = 0;
-        int theirCards = 0;
-        for (Card c : game.getCardsIn(ZoneType.Hand)) {
-            if (c.getController() == aiPlayer) {
-                myCards++;
+        int alliedCards = 0;
+        int opponentCards = 0;
+        int alliedLife = 0;
+        int alliedPlayers = 0;
```
tool4ever: same int twice?

forge-ai/src/main/java/forge/ai/simulation/GameStateEvaluator.java
```diff
@@ -112,37 +115,48 @@ public Score getScoreForGameState(Game game, Player aiPlayer) {
...
+            if (player.isOpponentOf(aiPlayer)) {
+                handScore -= 4 * cards;
+                opponentCards += cards;
+                opponentLife += player.getLife();
+                debugPrint("  Opponent " + (++opponents) + " life: -" + player.getLife());
             } else {
-                theirCards++;
+                int fullValueCards = player.isUnlimitedHandSize()
+                        ? cards : min(cards, player.getMaxHandSize());
+                handScore += cards + 4 * fullValueCards;
```
tool4ever: base score randomly changed

forge-ai/src/main/java/forge/ai/simulation/GameStateEvaluator.java
```diff
@@ -96,6 +96,9 @@ private Score getScoreForGameOver(Game game, Player aiPlayer) {
     }
 
     public Score getScoreForGameState(Game game, Player aiPlayer) {
+        if (aiPlayer.hasLost()) {
+            return new Score(Integer.MIN_VALUE);
+        }
         if (game.isGameOver()) {
             return getScoreForGameOver(game, aiPlayer);
```
tool4ever: I think this needs an `isInGame()` check instead so it also uses the more dynamic score just implemented

forge-ai/src/main/java/forge/ai/simulation/GameStateEvaluator.java
```diff
@@ -157,7 +171,7 @@ private Score getScoreForGameStateImpl(Game game, Player aiPlayer) {
 
         int summonSickScore = score;
         PhaseType gamePhase = game.getPhaseHandler().getPhase();
-        for (Card c : game.getCardsIn(ZoneType.Battlefield)) {
+        for (Card c : game.getCardsIncludePhasingIn(ZoneType.Battlefield)) {
```
tool4ever, line 180: you can't just leave these at the same score though

and even if you would half it AI might still thinking using *Oubliette* to save own creature would be a good play so I'm not sure this part really works out...

forge-ai/src/main/java/forge/ai/simulation/GameStateEvaluator.java
Madwand99: The base weighting for the evaluating player is unchanged. cards + 4 * min(cards, maxHandSize) is equivalent to the previous five points per retained card plus one point per excess card. This form applies the existing weighting independently to each friendly player so teammate hands are included in team-game evaluation.

forge-ai/src/main/java/forge/ai/simulation/GameStateEvaluator.java
Madwand99: I've renamed the variables to hopefully make this more clear.

forge-ai/src/main/java/forge/ai/simulation/GameStateEvaluator.java
Madwand99: Agreed. I changed this to check !aiPlayer.isInGame() after the normal game.isGameOver() handling.
This lets an eliminated player in an ongoing multiplayer game receive the same turn-adjusted terminal score as other finished players. It also handles a player that has already won while the multiplayer game is still resolving. Importantly, a real loss no longer returns Integer.MIN_VALUE, since that value is reserved as the failed-simulation sentinel.
I added regressions for both an eliminated player and a finished winner while the overall multiplayer game remains open.

forge-ai/src/main/java/forge/ai/simulation/GameStateEvaluator.java
Madwand99: Agreed that phased-out permanents should not simply receive the same immediate score as active permanents, especially for effects such as Oubliette. An arbitrary partial value would have the opposite problem, though: normal phasing effects such as Clever Concealment would appear to permanently lose the protected cards.
I changed the existing secondary score into a more general availableValue:
value is the strategic score and includes permanents guaranteed to phase in normally.
availableValue excludes phased-out permanents because they cannot currently provide mana, activate abilities, or participate in combat.
Permanents marked WontPhaseInNormal, as with Oubliette, are excluded from both scores.
SpellAbilityPicker uses the available score for its existing “hold this play for later” check, while simulation comparison retains the strategic score. The effect cache is only used when both score deltas agree, avoiding reuse of a result whose immediate and strategic consequences differ.
The tests now prove both sides: Clever Concealment remains acceptable, while using Oubliette on the AI’s own creature is rejected.

forge-ai/src/main/java/forge/ai/simulation/SimulationController.java
```diff
@@ -250,17 +251,21 @@ public void possiblyCacheResult(Score score, SpellAbility sa) {
...
+            if (scoreDelta <= 0 && scoreDelta == availableScoreDelta
+                    && d.targets != null) {
                 // FIXME: Support more than one target in this logic.
                 GameObject[] hostAndTarget = currentHostAndTarget;
                 if (currentHostAndTarget != null) {
                     GameStateEvaluator evaluator = new GameStateEvaluator();
                     Player player = sa.getActivatingPlayer();
                     int cardScore = evaluator.evalCard(player.getGame(), player, (Card) hostAndTarget[2]);
-                    effectCache.add(new CachedEffect(hostAndTarget[0], sa, hostAndTarget[1], cardScore, scoreDelta));
+                    effectCache.add(new CachedEffect(hostAndTarget[0], sa, hostAndTarget[1],
```
tool4ever: can you please higher your line limit preference?

these random changes on top don't make it easier

Madwand99: Not sure this is a preference I can change but I can try to be more vigilant about it in the future. Fixed! And sorry about the added complexity, the phasing handling improvements were non-trivial.

forge-ai/src/main/java/forge/ai/simulation/GameStateEvaluator.java
tool4ever: yea this is almost too much together again...though I can understand how one thing can lead to another 😬

will do some manual checks in a bit

tool4ever (approved):

PR #11485, opened 2026-08-01
Let the AI pay for converge and sunburst when X is the colour knob

liamiak:
On a converge or sunburst card whose colour count lives in a second SVar, `X` has one job: buy colours. The AI was not announcing it, so those cards were cast for the minimum.

`ComputerUtilCost.setMaxXValue` ends with `sa.setXManaCostPaid(x)`, leaving X at the maximum. `PermanentAi` then measured its converge baseline *there*:

```java
final int xPay = ComputerUtilCost.setMaxXValue(sa, ai, false);   // X is now the max
if (source.hasConverge()) {
    int nColors = ComputerUtilMana.getConvergeCount(sa, ai);     // measured at max X
    for (int i = 1; i <= xPay; i++) {
        sa.setXManaCostPaid(i);
        int newColors = ComputerUtilMana.getConvergeCount(sa, ai);
        if (newColors > nColors) { nColors = newColors; }
        else { sa.setXManaCostPaid(i - 1); break; }
    }
}
```

The baseline is already the best colour count available, so the first step of the walk always compared worse and X collapsed to **0**. `TokenAi` had a different miss: its X block is guarded by `"X".equals(tokenAmount)`, so a card counting colours in `Y` never reached any X handling at all.

Both want the same thing, so it is now named once - `ComputerUtilMana.setXForBestConverge` - and `PermanentAi` shrinks by 10 lines calling it.

## Effect

All five cards in the pool whose cost has X and for which `Card.hasConverge()` is true, with five differently coloured lands untapped:

| | X before | X after | |
|---|---|---|---|
| Engineered Explosives | 0 | **5** | 0 charge counters, so it could only ever destroy 0-drops |
| Chamber Sentry | 0 | **5** | entered as a 0/0 and died |
| Skyrider Elf | 0 | **3** | entered as a 0/0 and died |
| Sweep the Skies | unannounced | **4** | converge 1 -> 5, so five thopters rather than one |
| Prismatic Ending | unannounced | unannounced | not fixed, see below |

**Only Engineered Explosives is flagged.** The other three that this fixes are in AI decks today - two of them entering as 0/0s and dying.

Edge cases behave: five Wastes gives X=0, and Plains/Plains/Island gives X=2 rather than paying for a colour already covered.

## Engineered Explosives unflagged

Casting it for X=0 was the only thing wrong with it. The activated half needs nothing: `DestroyAll` routes it through the same `DestroyAllAi.doMassRemovalLogic` as Wrath of God, and it matched Wrath's decision on every board I tried. Two charge counters, everything at cmc 2 so both cards hit the same permanents:

| AI creatures | opponent creatures | Engineered Explosives | Wrath of God |
|---|---|---|---|
| 161 | 307 | declines | declines |
| 161 | 468 | plays | plays |
| 322 | 629 | plays | plays |

It does wipe at a cost to itself when the swing justifies it, on the shared `CREATURE_EVAL_THRESHOLD` - not only when it risks nothing.

The flag was also self-fulfilling: `AiController.getSpellAbilityToPlayImpl` strips every ability of a flagged card before evaluating anything, so "the AI never plays it" held regardless of whether its AI worked.

## Prismatic Ending, not fixed

Announcing X takes its converge from 1 to 5 and it still declines, so X is necessary but not sufficient there. Its condition is `ConditionPresent$ Permanent.nonLand+cmcLEY`, and `Count$Converge` reads the colours *actually* paid:

```java
if (sq[0].contains("Converge")) {
    SpellAbility castSA = c.getCastSA();
    return doXMath(castSA == null ? 0 : castSA.getPayingColors().countColors(), expr, c, ctb);
}
```

Before the spell is cast `getCastSA()` is null, so Y is 0 and `cmcLEY` matches nothing above a 0-drop. The AI's `getConvergeCount` predicts converge by simulating payment, but the condition path uses the rules-side count; bridging the two is a separate piece of work and I have left the card flagged.

`DamageAllAi:30` and `DestroyAllAi:64` have the same `X`-only assumption, but no card in the pool currently reaches them through it.

## Testing

Two tests. Engineered Explosives plays a real turn and lands with five charge counters - that one has to go through the full `AiController` path, since it is the unflagging that needs proving. Sweep the Skies asserts the announced X instead, because `TokenAi` gates token spells behind a random roll and a played-turn assertion on it is flaky; both assertions were run repeatedly each way.

Each test fails without its own half of the change. 337 tests, 0 failures, checkstyle clean.

### DIFF
--- a/forge-gui/res/cardsfolder/e/engineered_explosives.txt
+++ b/forge-gui/res/cardsfolder/e/engineered_explosives.txt
@@ -6,6 +6,5 @@ A:AB$ DestroyAll | Cost$ 2 Sac<1/CARDNAME> | ValidCards$ Permanent.nonLand+cmcEQ
 SVar:X:Count$xPaid
 SVar:Y:Count$CardCounters.CHARGE
 SVar:NonStackingEffect:True
-AI:RemoveDeck:All
 DeckHints:Ability$Proliferate
 Oracle:Sunburst (This enters with a charge counter on it for each color of mana spent to cast it.)\n{2}, Sacrifice Engineered Explosives: Destroy each nonland permanent with mana value equal to the number of charge counters on Engineered Explosives.
--- a/forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
+++ b/forge-ai/src/main/java/forge/ai/ComputerUtilMana.java
@@ -85,6 +85,25 @@ public class ComputerUtilMana {
         return 0;
     }
 
+    /**
+     * Announce X on a converge or sunburst card, where its only job is to buy colors: the least X
+     * that still reaches the most of them. Returns the number of colors that will be paid.
+     */
+    public static int setXForBestConverge(final SpellAbility sa, final Player ai, final int maxX) {
+        int bestX = 0;
+        int bestColors = 0;
+        for (int i = 0; i <= maxX; i++) {
+            sa.setXManaCostPaid(i);
+            int colors = getConvergeCount(sa, ai);
+            if (colors > bestColors) {
+                bestColors = colors;
+                bestX = i;
+            }
+        }
+        sa.setXManaCostPaid(bestX);
+        return bestColors;
+    }
+
     // Does not check if mana sources can be used right now, just checks for potential chance.
     public static boolean hasEnoughManaSourcesToCast(final SpellAbility sa, final Player ai) {
         if (ai == null || sa == null)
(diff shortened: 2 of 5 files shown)


forge-gui/res/cardsfolder/e/engineered_explosives.txt
```diff
@@ -6,6 +6,5 @@ A:AB$ DestroyAll | Cost$ 2 Sac<1/CARDNAME> | ValidCards$ Permanent.nonLand+cmcEQ
 SVar:X:Count$xPaid
 SVar:Y:Count$CardCounters.CHARGE
 SVar:NonStackingEffect:True
-AI:RemoveDeck:All
```
tool4ever, line 9: I don't think the reasoning for this was zero X since that's another regression and more like AI unable to coordinate which one is actually a smart choice

liamiak:
You're right — closing into #11486.

Engineered Explosives needs board-aware counter selection rather than most-colours: its counters
pick which mana value the wipe hits, so five is usually the worst live choice. It stays flagged.
#11486 carries the other three cards and unifies the duplicated simulation you flagged there.

PR #11487, opened 2026-08-02
Don't count mana sources against the cost when reserving them

omraj21:
`AiController.reserveManaSources` asks `getManaSourcesToPayCost` for the sources it should hold back for a spell it wants to cast later, then checked how many cards it got against the mana value of the spell:

```java
// This is a simplification, since one mana source can produce more than one mana,
// but should work in most circumstances to ensure safety in whatever the AI is using this for.
if (manaSources.size() >= cost.getConvertedManaCost()) {
```

The comment undersells it — **the check never rejected anything.** Planning a payment drains the `ManaCostBeingPaid` that was passed in, so by the time the comparison runs `getConvertedManaCost()` has already dropped to `0` and the condition is `size() >= 0`.

I confirmed that with a probe before changing anything. On a board of Sol Ring + 2 Mountains against a mana value 4 spell:

```
planSize=3  costCMC=0  canPay=true  -> reserved=true
```

Three cards for a mana value four spell — correct, because Sol Ring makes two mana, but it only passed because the guard had already degenerated. Had the comparison worked as written, it would have refused a perfectly good reservation.

## The count isn't needed either way

`getManaSourcesToPayCost` returns the sources of a payment it has **already worked out**, and `null` when the cost can't be paid at all:

```java
final List<SpellAbility> payment = payManaCost(cost, sa, ai, true, true, effect);
if (payment == null) {
    return null;
}
```

So reaching that point already means the cost is covered, and counting cards against mana value can only ever be wrong when a source makes more than one mana.

## One real behaviour change

For chained spells the sources needed by the other spell are removed from the list. If that left nothing, the AI still reserved an empty set and reported success. It now returns `false`, so the caller knows the second spell has no sources of its own to hold back. The added `doesNotReserveSourcesTheChainedSpellNeeds` test fails on the old code.

## Verification

3 tests: the multi-mana source case, the genuinely-unpayable case, and the chained-spell case above.

Full `mvn -U -B clean test`: 12 modules, 339 tests, 0 failures, run three times.

---

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiController.java
+++ b/forge-ai/src/main/java/forge/ai/AiController.java
@@ -801,19 +801,12 @@ public class AiController {
             CardCollection exceptSources = ComputerUtilMana.getManaSourcesToPayCost(
                     ComputerUtilMana.calculateManaCost(exceptForThisSa.getPayCosts(), exceptForThisSa, player, true, 0, false),
                     exceptForThisSa, player, false);
-            if (exceptSources != null && !exceptSources.isEmpty()) {
-                // the other spell needs these too, so they can't be promised to both
-                if (manaSources.removeAll(exceptSources) && manaSources.isEmpty()) {
-                    return false;
-                }
+            // the other spell needs these too, so they usually can't be promised to both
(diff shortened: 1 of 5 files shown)


tool4ever (approved):
I doubt this reservation check is even fully safe in all cases but I don't have time to argue with AI right now 🤷‍♂️

PR #11488, opened 2026-08-02
Equip loop fix

Eradev:
Found the AI to be looping with Equip activations.

I was able to reproduce it with a test using the following cards in play and no mana in the pool:

Nahiri, Storm of Stone (Equip cost -1)
Kor Halberd (Equip 1 -> 0)
Nahiri, Forged in Fury (Target)


tool4ever:
Can't reproduce:
grafik

Eradev:
My bad. My main was dirty.

PR #11491, opened 2026-08-02
AI: don't let a RemoveCounterAll sub-ability veto its own parent

liamiak:
`ApiType.RemoveCounterAll` mapped to `CannotPlayAi`, which refuses in `chkDrawback` as well as
`canPlay`. `chkDrawbackWithSubs` bails on any unwilling sub-ability, so a `RemoveCounterAll` sub
took its parent down with it.

## What a reviewer can check

Alaundo the Seer's `{T}` ends by ticking time counters on the other cards it owns in exile, so
the whole ability is unreachable — and it carries no `AI:RemoveDeck`, so the AI has it in decks
with a dead activated ability. Playing a turn out:

```
master:  alaundoTapped=false  exiled=0  timeCounters=0
this:    alaundoTapped=true   exiled=1  timeCounters=1
```

`RemoveCounterAllDrawbackTest` asserts that and fails without the change.

## Can this make the AI choose something bad?

`canPlay` still refuses, exactly as `CannotPlayAi` did, so nothing gains a new main ability. Only
Aether Snap and Thief of Blood use this API as a main ability and both stay flagged.

The three live cards where removing all counters is a real cost rather than bookkeeping — Blood
Hound, Hammer Jammer and Inhumaniac, each stripping their own `+1/+1` counters — reach it from
mandatory triggers, so the AI is never asked either way.

`chkDrawback` is left to the base class rather than overridden. No `RemoveCounterAll` ability in
any of the 19 scripts targets, so the base default is willing; if one ever does, the base class
checks its candidates instead of assuming.

## Scope

19 cards reach this API and 5 carry `AI:RemoveDeck`, so 14 are in AI decks today. Where it is a
sub-ability it is a consequence of the parent rather than a decision: clear the FATE counters just
checked, tick TIME counters on suspended cards, reset your own `+1/+1`s, remove a `-1/-1` from
your own creature. 8 of the 13 unflagged cards reach it from a trigger — that split is read from
the scripts rather than measured, and only Alaundo is measured end to end.

## Not in this PR

Oblivion Stone is the card that led me here — its wipe ends by clearing the fate counters it just
checked — but unflagging it also needs `AILogic$ AtOppEOT` and an `AITgts` filter so the AI seeds
counters at the opponent's end step rather than spending its own tap. That is a separate
judgement, and better sent on its own once this lands.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/SpellApiToAi.java
+++ b/forge-ai/src/main/java/forge/ai/SpellApiToAi.java
@@ -158,7 +158,7 @@ public enum SpellApiToAi {
             .put(ApiType.Regenerate, RegenerateAi.class)
             .put(ApiType.Regeneration, AlwaysPlayAi.class)
             .put(ApiType.RemoveCounter, CountersRemoveAi.class)
-            .put(ApiType.RemoveCounterAll, CannotPlayAi.class)
+            .put(ApiType.RemoveCounterAll, CountersRemoveAllAi.class)
             .put(ApiType.RemoveFromCombat, RemoveFromCombatAi.class)
             .put(ApiType.RemoveFromGame, AlwaysPlayAi.class)
             .put(ApiType.RemoveFromMatch, AlwaysPlayAi.class)
--- a/forge-ai/src/main/java/forge/ai/ability/CountersRemoveAllAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/CountersRemoveAllAi.java
@@ -1,0 +1,21 @@
+package forge.ai.ability;
+
+import forge.ai.AiAbilityDecision;
+import forge.ai.AiPlayDecision;
+import forge.ai.SpellAbilityAi;
+import forge.game.player.Player;
+import forge.game.spellability.SpellAbility;
+
+/**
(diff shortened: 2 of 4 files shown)


liamiak:
Force-pushed on 23 Aug with no note, so here is what changed — it is smaller than the version the branch name suggests.

The Oblivion Stone commit is gone. Unflagging that card also needs `AILogic$ AtOppEOT` and an `AITgts` filter so the AI seeds fate counters at the opponent's end step rather than spending its own tap, and that is a separate judgement. What is left is the API mapping only: `RemoveCounterAll` moves off `CannotPlayAi` to a `CountersRemoveAllAi` whose `canPlay` still refuses exactly as before, so nothing gains a new main ability.

The earlier version also overrode `chkDrawback` to return `WillPlay`. That is dropped too — the base class checks candidates rather than assuming willingness. No `RemoveCounterAll` in any of the 19 scripts targets today, so the base default is willing anyway, but it stops being an assumption.

The test moved with it, from Oblivion Stone to Alaundo the Seer. Alaundo carries no `AI:RemoveDeck`, so it is a dead activated ability in AI decks today — the bug, rather than a card the AI is already flagged out of.

Rebased onto master; 3 files, +70/-1.

PR #11492, opened 2026-08-02
Make Adventure action animations frame-aware

vanja-ivancevic:
- Many vanilla Adventure sprites already contain detailed action sequences that fixed delays cut short; frame-aware timing lets those authored frames finish.
- Vanilla Shade attack: the current 0.8-second delay displays 4 of 16 frames, while frame-aware playback displays all 16 over 3.2 seconds.

  Vanilla Shade attack before and after frame-aware timing

- Vanilla Wandering Giant death: the current 0.3-second delay displays only the opening of its 24-frame sequence, while frame-aware playback displays all 24 frames over 4.8 seconds.

  Vanilla Wandering Giant death before and after frame-aware timing

- Replace fixed Adventure encounter and result delays with durations derived from each sprite's frame count at the existing 0.2-second frame rate.
- Play Attack, Hit, and Death animations once, hold their final frame, and restart explicitly requested actions.
- Preserve existing fallback timing and looping behavior for atlases without the requested action regions.
- Cap action-animation waits at five seconds so unusually large atlases cannot create excessive delays.
- Wait for the longer player or enemy action so neither animation is cut short.

I understand that it could be irritating to wait, but it's definitely closer to what the sprite artists intended to have.

### DIFF
--- a/forge-gui-mobile/src/forge/adventure/character/CharacterSprite.java
+++ b/forge-gui-mobile/src/forge/adventure/character/CharacterSprite.java
@@ -16,6 +16,8 @@ import java.util.HashMap;
  */
 
 public class CharacterSprite extends MapActor {
+    private static final float DEFAULT_ANIMATION_FRAME_DURATION = 0.2f;
+    private static final float MAX_DEATH_ANIMATION_DURATION = 3f;
     private static final float MAX_ACTION_ANIMATION_DURATION = 5f;
     private final HashMap<AnimationTypes, HashMap<AnimationDirections, Animation<TextureRegion>>> animations = new HashMap<>();
     float timer;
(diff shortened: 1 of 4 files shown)


tehdiplomat:
This idea is fine, but is there a maximum seconds we want to allow? What if something has 100 frames (because someone thought it would be funny)

vanja-ivancevic:
The longest vanilla action sequence I found is the Wandering Giant’s 24-frame Death animation, which lasts 4.8 seconds at the existing 0.2-second frame rate. We could cap action-animation waits at 5 seconds, which preserves all current authored sequences while preventing unusually large atlases from creating excessive delays.

kevlahnota:
@vanja-ivancevic can we make the deathanimation play at a faster rate if it exceeds 3 seconds and make it play at 3 seconds max?

vanja-ivancevic:
> @vanja-ivancevic can we make the deathanimation play at a faster rate if it exceeds 3 seconds and make it play at 3 seconds max?

Implemented. Death animations keep the normal 0.2-second frame rate when they are 3 seconds or shorter. Longer Death animations now scale their frame rate proportionally so every authored frame plays within a 3-second maximum. For example, the Wandering Giant’s 24 frames now play at 0.125 seconds per frame.

kevlahnota (approved):
works fine ty

PR #11493, opened 2026-08-02
Improve hybrid AI evaluation of costs and delayed effects

Madwand99:
## Summary

This improves the depth-zero safety check used by Hybrid AI when a play has explicit costs or mandatory delayed consequences.

Previously, the checker compared the final result directly against the original board state. This could reject reasonable plays because discarded, sacrificed, or temporarily removed cards appeared to be permanent losses. It also could not evaluate next-turn delayed draws because those triggers were not exposed until cleanup.

The checker now:

- Uses the post-cost game state as the comparison baseline, leaving cost approval to the existing AI heuristics.
- Resolves the proposed play and its resulting stack before evaluating it.
- Evaluates newly created mandatory delayed returns to hand or battlefield.
- Evaluates concrete next-turn delayed draws, allowing their benefits and associated draw-punisher triggers to affect the result.
- Ignores optional and `UnlessCost` delayed effects rather than inventing a choice for the AI.
- Continues to treat land plays separately so enters-the-battlefield drawbacks are compared against the pre-play position.

`TriggerHandler` now retains pending `NextTurn` triggers in a snapshot-accessible collection. Their normal game timing is unchanged: they still become active at cleanup exactly as before.

Regression coverage includes:

- Explicit discard costs such as Phantasmagorian.
- Nine-Lives Familiar returning from the graveyard.
- Ignorant Bliss and Memory Jar restoring temporarily removed hands.
- Lat-Nam's Legacy receiving its next-turn cards.
- Urza's Bauble not hiding immediate lethal Disciple of the Vault damage.
- Urza's Bauble being rejected when its delayed draw is lethal through Nekusar.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/simulation/OnePlaySafetyChecker.java
+++ b/forge-ai/src/main/java/forge/ai/simulation/OnePlaySafetyChecker.java
@@ -1,10 +1,19 @@
 package forge.ai.simulation;
 
 import forge.ai.simulation.GameStateEvaluator.Score;
+import forge.game.Game;
+import forge.game.ability.AbilityUtils;
+import forge.game.ability.ApiType;
 import forge.game.card.Card;
 import forge.game.player.Player;
 import forge.game.spellability.SpellAbility;
+import forge.game.trigger.Trigger;
+import forge.game.trigger.TriggerType;
 import forge.game.zone.ZoneType;
+
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 public final class OnePlaySafetyChecker {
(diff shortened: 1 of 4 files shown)


tool4ever:
Getting a bit too hacky for my taste again 😩
might be a while before I have time to come up with more detailed feedback, some other PR have higher priority for now

Madwand99:
> Getting a bit too hacky for my taste again 😩

The main "hack" here is that I evaluate delayed effects (such as the delayed draw from Lat-Nam's Legacy) as if it is happening immediately. There might be a better way to approximate this. Open to ideas! Or we could leave this approximation out and just be content with the safety checker just denying certain cards.

PR #11499, opened 2026-08-03
Let the AI use Goblin Welder

liamiak:
Goblin Welder was `AI:RemoveDeck:All`, which is not only a deckbuilding flag:
`AiController.chooseSpellAbilityToPlay` strips every ability of a flagged card from the candidate
list, so on master the AI never activates it and no `PumpAi` code runs.

Lifting the flag alone is worse than leaving it. It targets both the artifact sacrificed and the
graveyard card replacing it; with no `AILogic` the generic pump targeting takes the most expensive
artifact on the board regardless of owner. With only the flag removed:

| Board (Welder plus) | Flag lifted, no AILogic |
|---|---|
| ours: Chromatic Star, Sol Ring; Wurmcoil in the yard | sacrificed the **Sol Ring** |
| ours: Sol Ring; only an Ornithopter in the yard | still welded — **Sol Ring → Ornithopter** |
| theirs: Wurmcoil, Blightsteel in their yard | sacrificed **their** Wurmcoil, handing them the Colossus and two tokens |

`AILogic$ Weld` picks the pair as one trade, delegating each choice to what already exists:

- **ours, going away** — `getCardPreference(… "SacCost" …)`, else `getWorstAI`. That keeps Sol Ring
  over Chromatic Star, which carries `SVar:SacMe:1`.
- **theirs, going away** — `getBestRemovalTargetAI`; welding against an opponent is removal.
- **coming back** — `pumpMandatoryTarget`: best from our graveyard, cheapest from theirs. The
  sub-ability needs the tag too — `pumpTgtAI` resets targets before dispatching, and the generic path
  would take their *best*.
- **which trade** — the largest mana-value delta, positive being the whole go/no-go test.
  `evaluatePermanentList` values a non-creature permanent at `CMC + 1`; `ZoneExchangeAi` gates its
  swap the same way. A point off an opponent counts as a point to us, as `GameStateEvaluator` does
  for board presence.
- **timing** — the shared `AtOppEOT` rule.

Only this card carries the `AILogic`, so nothing else through `PumpAi` changes.
`AI:RemoveDeck:Random` goes too, since `AI_CAN_PLAY` needs both flags clear;
`DeckNeeds:Type$Artifact` replaces it.

Reasoned, not measured: the opponent's end step over acting earlier. Known limit: one candidate pair
per player, not every combination.

Six tests, second commit, droppable alone: two fail without the fix, two caught earlier versions of
this logic, two guard the "no trade" branch. Full desktop suite: 363 tests, 0 failures, 6 skipped.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/PumpAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/PumpAi.java
@@ -80,6 +80,9 @@ public class PumpAi extends PumpAiBase {
             if (ph.getPhase().isAfter(PhaseType.COMBAT_FIRST_STRIKE_DAMAGE) || !ph.inCombat()) {
                 return false;
             }
+        } else if (logic.equals("Weld")) {
+            // nothing is gained by welding early, so hold it to the last moment before our turn
+            return super.checkPhaseRestrictions(ai, sa, ph, "AtOppEOT");
         }
         return super.checkPhaseRestrictions(ai, sa, ph);
     }
@@ -233,6 +236,8 @@ public class PumpAi extends PumpAiBase {
         } else if (aiLogic.startsWith("Donate")) {
             // Donate step 1 - try to target an opponent, preferably one who does not have a donate target yet
             return SpecialCardAi.Donate.considerTargetingOpponent(ai, sa);
+        } else if ("Weld".equals(aiLogic)) {
+            return doWeldLogic(ai, sa);
         } else if (aiLogic.equals("InfernoOfTheStarMounts")) {
             int numRedMana = ComputerUtilMana.determineLeftoverMana(new SpellAbility.EmptySa(source), ai, "R", false);
             int currentPower = source.getNetPower();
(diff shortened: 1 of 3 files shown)


liamiak:
Rebased onto master and reworked. Two things changed the shape of this:

- `AI:RemoveDeck:All` turns out to gate the ability at runtime, not only at deckbuilding — `AiController.chooseSpellAbilityToPlay` strips a flagged card's abilities from the candidate list, so no `PumpAi` code was ever reached. That makes removing the flag on its own a regression rather than a no-op, and the description now shows what the generic targeting does with the flag lifted and no `AILogic`.
- The pair taken is now the one with the largest mana value delta rather than the first profitable one found, so a bigger swing against an opponent is no longer passed over for a smaller upgrade of our own.

`AI:RemoveDeck:Random` goes as well, since `AI_CAN_PLAY` needs both flags clear; `DeckNeeds:Type$Artifact` replaces it. Six tests, two of which fail without the fix.

liamiak:
Rebased onto master — `PumpAi.checkPhaseRestrictions` was reworked since (#11739, #11742) and the weld path sits above that change. Full desktop suite: 371 tests, 0 failures, 6 skipped.

I also ran this in real games rather than only unit tests: 20 headless mirror matches with a mono-red artifact deck. The AI chose to weld 49 times — 22 upgrading its own board (Ichor Wellspring → Sundering Titan, Pyrite Spellbomb → Wurmcoil Engine), 27 against opponents (their Blightsteel Colossus → their Mind Stone). No exceptions.

That run found a bug in my own logic, now fixed in the first commit: the sub-ability's targets weren't cleared on entry and the chosen pair was re-derived after the scan, so 2 of 30 decisions committed to a different card than the one they were scored on. It now keeps the pair it scored — 0 divergences in 49.

Cost: the `AtOppEOT` gate keeps it off the per-decision path — 100 calls across 20 games, ~0.1s of AI time per game (per-call cost varied 8–28ms between runs).

Limit the run also exposed: the go/no-go is mana value, so a cheap-but-strong artifact is undervalued (it will trade a Sol Ring for a Mind Stone) and a token's 0 counts as 0. That's Forge's shared permanent valuation rather than anything specific to this card — happy to address it here if you'd prefer, but I didn't want to change the global model in this PR.

PR #11500, opened 2026-08-03
MBC cards, 3rd August, Part 2

Fulgur14:


tool4ever (approved):

forge-gui/res/cardsfolder/upcoming/kuroki_thief_of_talents.txt
```diff
@@ -0,0 +1,12 @@
+Name:Kuroki, Thief of Talents
```
Fulgur14, line 1: I feel like this one might need some extra support, I couldn't find anything similar; it has basically a branch based on whether the UnlessCost is paid or not. This way of writing it could fail in some incredibly unlikely situations (if the opponent has no cards in hand and somehow manages to not add any cards with the draw 4 due to replacement effects or empty library).

tool4ever: if his library is empty he can't pay anyway...

Fulgur14: There's still Words of Wilding, Obstinate Familiar, etc.

tool4ever: well you could switch to `RememberRevealedPlayer`

Fulgur14: Thanks :)

PR #11510, opened 2026-08-04
Implemented cards from the PUNK set

BigCrunch22:
Implented the following cards that didn't require code changes from the _**Black Lotus Unknown Planechase set**_:
---

- **Asmoranomardicadaistinaculdacar's Kitchen**
- **Bicycle Rack**
- **Black Lotus Lounge**
- **Circus of the Sun + _Token_**
- **City Hall**
- **Elvish Impersonation Contest**
- **Ghirapur Grand Prix**
- **Jalira's Show  + _Token_**
- **Mojave Desert**
- **No Way Out**
- **Omenpath Instability + _Tokens_**
- **Oteclán**
- **Preston's Stage + _Token_**
- **Pursued by Something**
- **Raiders' Allegiance**
- **Shy Town**
- **Sky Deck**
- **Tarnation**
- **That's Enough Slices + _Token_**
- **The Bean**
- **The Food Court**
- **The Windy City**

_Implemented with assistance from Claude._


forge-gui/res/cardsfolder/w/welcome_to_valley.txt
```diff
@@ -0,0 +1,8 @@
...
....Echidna,Creature.Elephant,Creature.Elk,Creature.Ferret,Creature.Fish,Creature.Fox,Creature.Frog,Creature.Giraffe,Creature.Goat,Creature.Hamster,Creature.Hedgehog,Creature.Hippo,Creature.Horse,Creature.Hyena,Creature.Insect,Creature.Jackal,Creature.Jellyfish,Creature.Kangaroo,Creature.Leech,Creature.Lemur,Creature.Llama,Creature.Lizard,Creature.Lobster,Creature.Mammoth,Creature.Mite,Creature.Mole,Creature.Mongoose,Creature.Monkey,Creature.Mouse,Creature.Nautilus,Creature.Octopus,Creature.Otter,Creature.Ox,Creature.Oyster,Creature.Pangolin,Creature.Platypus,Creature.Porcupine,Creature.Possum,Creature.Rabbit,Creature.Raccoon,Creature.Rat,Creature.Rhino,Creature.Sable,Creature.Salamander,Creature.Scorpion,Creature.Seal,Creature.Shark,Creature.Sheep,Creature.Skunk,Creature.Sloth,Creature.Slug,Creature.Snail,Creature.Snake,Creature.Spider,Creature.Sponge,Creature.Squid,Creature.Squirrel,Creature.Starfish,Creature.Trilobite,Creature.Turtle,Creature.Walrus,Creature.Weasel,Creature.Whale,Creature.Wolf,Creature.Wolverine,Creature.Wombat,Creature.Worm | ValidActivatingPlayer$ You | TriggerZones$ Command | Execute$ TrigDraw | TriggerDescription$ Whenever you cast an animal spell, draw a card.
```
Jetz72: Problem with implementing this loosely-defined batch is keeping it updated. If Wizards comes along and adds a new one, it's unlikely anybody is going to remember to come back to this list and update it.

BigCrunch22: Yeah that is what I am worried about this implementation.

BigCrunch22:
- Added The Sphere, which required code changes (**AmassAi** and **AmassEffect**)

_It's my first time doing a Java code-related change, so criticism is very much appreciated. Performed with assistance from Claude._

forge-game/src/main/java/forge/game/card/CounterEnumType.java
```diff
@@ -129,7 +129,7 @@ public enum CounterEnumType implements CounterType {
 
     DOOM("DOOM", 255, 104, 118, CounterAiCategory.Negative),
 
-    DREAD ("DREAD", 205, 170, 240, CounterAiCategory.Positive),
+    DREAD ("DREAD", 205, 170, 240, CounterAiCategory.Neutral),
```
Jetz72: I'm not sure we want to change the AI logic for a black-border counter over something this obscure. The logic shouldn't even matter for counters that go on a Plane, since as far as I know there's no way to interact with them.

Jetz72:
Implementing a "color" parameter for Amass is a bit of a grey area, since the keyword on [the card](https://scryfall.com/card/punk/PLA037/the-sphere) isn't using that color as a parameter the way it does with the creature type - e.g.. it doesn't "Amass Red Fans 5". The reminder text seems to indicate that Fans are just an exception to the existing Amass behavior. 

So at its most literal, it'd be done by checking if the type is "Fans" in the Amass effect and if so, making a red army instead. But that isn't particularly flexible. Your way is flexible, but introducing a keyword parameter might be a bit much for a single, non-legal, non-Oracle, non-traditional card that also doesn't use a keyword parameter, and is from a set that often sees functional errata with reprints. I'd suggest holding off on adding color-Amass to the engine until Wizards gives us a few more examples.

Jetz72 (dismissed):

BigCrunch22:
That's bad timing. Just noticed a bit of inconsistency.

Jetz72:
Failing to compile because Point counters already exist as a counter type it seems. Have you tested these?

BigCrunch22:
> Failing to compile because Point counters already exist as a counter type it seems. Have you tested these?

I have done some testing with the cards. I just missed the entry on the point counter on the enums. That's my bad.

BigCrunch22:
Performed a much more thorough analysis on each of the cards and their interactions. They should work as they are read right now. There are some cards to note though:

- Event Horizon - Due to Wastes' characteristics, I made it so that each non-basic cards changes their names to Wastes while its on the board and gets every characteristic of said land (sans Basic).
- Preston's Stage - The chaos Flash effect affects the stack as well since there is an assumption that the chaos effect should work with the "perform a magic trick" condition.

Jetz72:
> * Event Horizon - Due to Wastes' characteristics, I made it so that each non-basic cards changes their names to Wastes while its on the board and gets every characteristic of said land (sans Basic).

Should probably be done via a parameter to `CloneEffect` - something like `CopyFromChosenName` but specifying the name directly rather than needing the named-card being defined by a prior step.

> * Preston's Stage - The chaos Flash effect affects the stack as well since there is an assumption that the chaos effect should work with the "perform a magic trick" condition.

Hmm. So the issue is that "creature cards in your hand have Flash" fails to interact with "whenever you cast a spell with Flash"? I guess casting spells isn't included in any last-known-info or look-back-in-time rules. I think the adaptation is reasonable, but I'd leave a comment in the script explaining the disparity.

BigCrunch22:
> > * Event Horizon - Due to Wastes' characteristics, I made it so that each non-basic cards changes their names to Wastes while its on the board and gets every characteristic of said land (sans Basic).
> 
> Should probably be done via a parameter to `CloneEffect` - something like `CopyFromChosenName` but specifying the name directly rather than needing the named-card being defined by a prior step.
>
Consulted Claude for this. It seems CopyFromChosenName wouldn't quite work since its resolution is one and done and won't go the moment it planeswalks away. It gave me a suggestion of what can be done related to StaticAbilityContinuous.java, and it gave a structure similar to what is currently there but tied to the engine cloning its effect (since it's not really a clone in the sense that it will copy Wastes as is, just like Blood Moon doesn't make the non-basic lands named Mountain and have the Basic supertype). 

I'm not really keen on doing an engine change since it's not really strictly necessary for the effect to work. If this is still not on par even with that context in mind, I might just remove it.

Jetz72:
Hmm, yeah. "Non-basic lands are Wastes" is a bit ambiguous in what it means rules-wise. My first impression was that since Wastes isn't a type, the next meaning behind that would be that they become copies of the card Wastes. But yeah I don't think my suggestion quite works since it needs to be an ongoing static ability. More intuitive but less adherent-to-the-rules might be to just have it emulate Blood Moon's consequences - remove all land types and all other abilities of the card, then add `{T}: Add {C}.`. I think that'd be possible with existing script tools?

BigCrunch22:
> Hmm, yeah. "Non-basic lands are Wastes" is a bit ambiguous in what it means rules-wise. My first impression was that since Wastes isn't a type, the next meaning behind that would be that they become copies of the card Wastes. But yeah I don't think my suggestion quite works since it needs to be an ongoing static ability. More intuitive but less adherent-to-the-rules might be to just have it emulate Blood Moon's consequences - remove all land types and all other abilities of the card, then add `{T}: Add {C}.`. I think that'd be possible with existing script tools?

It already does that I believe. Also, I rename the cards to Wastes since a card like [Walker of the Wastes](https://scryfall.com/card/ogw/10/walker-of-the-wastes) synergizes with cards named "Wastes".

forge-gui/res/cardsfolder/g/ghirapur_grand_prix.txt
```diff
@@ -0,0 +1,11 @@
+Name:Ghirapur Grand Prix
+ManaCost:no cost
+Types:Plane Avishkar
+T:Mode$ PlaneswalkedTo | ValidCard$ Card.Self | Execute$ TrigEngines | TriggerDescription$ When you planeswalk here, all players start their engines! (If you have no speed, it starts at 1. It increases once on each of your turns when an opponent loses life. Max speed is 4.)
+SVar:TrigEngines:DB$ ChangeSpeed | Mode$ Increase | Defined$ Player
```
Jetz72: Won't this increase speed for players that already have a speed greater than zero?

forge-gui/res/cardsfolder/c/circus_of_the_sun.txt
```diff
@@ -0,0 +1,17 @@
...
+SVar:DBDraw:DB$ Draw | NumCards$ XPower | ConditionDefined$ RememberedLKI | ConditionPresent$ Card | ConditionCompare$ GE1 | SubAbility$ DBCleanupDraw
+SVar:XPower:RememberedLKI$CardPower
+SVar:DBCleanupDraw:DB$ Cleanup | ClearRemembered$ True
+T:Mode$ ChaosEnsues | TriggerZones$ Command | Execute$ TrigCounter | TriggerDescription$ Whenever Chaos ensues, put a flying counter on up to one target creature. It becomes a Performer in addition to its other types.
+SVar:TrigCounter:DB$ PutCounter | ValidTgts$ Creature | TargetMin$ 0 | TargetMax$ 1 | CounterType$ Flying | CounterNum$ 1 | SubAbility$ DBAnimate
+SVar:DBAnimate:DB$ Animate | Defined$ Targeted | Types$ Performer | Duration$ Permanent
+Oracle:Whenever you planeswalk here or at the beginning of your upkeep, create two 1/1 red performer creature tokens with flying and haste.\nWhenever one or more creatures you control with flying deal combat damage to a player, you may return it to your hand. If you do, draw cards equal to its power.\nWhenever Chaos ensues, put a flying counter on up to one target creature. It becomes a Performer in addition to its other types.
```
Jetz72, line 18: I think Circus of the Sun can't work as written without errata. "Whenever one or more creatures you control with flying deal combat damage to a player, you may return it to your hand. If you do, draw cards equal to its power." Besides the ambiguous "it" in reference to "one or more creatures", the wording also permits you to return a card you control but do not own to your hidden zone.

Jetz72 (commented):
I'm a little on the fence about renaming affected cards to "Wastes" for synergy, with no errata, rulings, or precedent to suggest the card is intended to work that way. Affected cards becoming actual copies of "Wastes" would be one literal interpretation of "Non-basic lands are Wastes". Scrubbing types, abilities, and adding the tap for C ability isn't as literal, but it's precedented by the net outcome of the effect the wording clearly references. 

I think maybe we just set that script aside for now? I can toss it into [tier 4 of the un-cards list](https://github.com/Card-Forge/forge/wiki/Un%E2%80%90cards,-Playtest-Cards,-and-Other-Funny-Cards#tier-4), with the other ones that have unsettled rules questions. Sometimes circumstances change and those gain better rules precedent (e.g. Land with an Adventure).

Was checking over my notes from an update to that list that I'd been working on, I noticed a couple things I'd written about some other planes here -

While not on Scryfall, The Sphere has been reprinted twice since the one you had implemented here with red fans. Once in [2025](https://mtg.wiki/page/Planechase_Unknown/2025_list_of_event_cards#Las_Vegas_2025_cards), and again in [2026](https://mtg.wiki/page/Planechase_Unknown/2026_list_of_event_cards#Las_Vegas_2026_cards). Notably, both have very different rules, and would be possible to implement. The issue is that Scryfall doesn't have an entry for it and the reprints have no official collector numbers, meaning there isn't a clean way to separate it from the image that Scryfall has with the original text.

The Windy City and The Bean were also heavily altered for their second printings. In those cases though we have both the Scryfall entry and a separate collector number, so it would technically be possible to support both through functional variants (see Garbage Elemental for instance). Can leave that off for now, though.

forge-gui/res/cardsfolder/c/circus_of_the_sun.txt
BigCrunch22: Here is the implementation I worked with when creating said effect (which would be the most likely interpretation of said effect):

_Whenever one or more creatures you control with flying deal combat damage to a player, you may return it to your hand. If you do, draw cards equal to its power._

becomes

_**Whenever one or more creatures you control with flying deal combat damage to a player, you may return any number of them to their owners' hands. If you do, draw cards equal to the total power of the creatures returned this way.**_

Should I reword the Oracle text to use that kind of phrasing even though it might not be the "official" wording of said effect?

Jetz72: Hmm... I think leave the Oracle text, but also put a note in the card script. Perhaps someday if we ever add support for the un-cards that put cards in another player's hidden zones, and if it isn't errata'd in a second printing by then, we could indeed tweak it to work as written.

Jetz72 (approved):

PR #11511, opened 2026-08-04
RoL: Mapfix; added arzakon card to arzakon deck

VigilantSera:


forge-gui/res/adventure/Realm of Legends/decks/legends/kaysa.dck
```diff
@@ -1,7 +1,7 @@
 [metadata]
-Name=Autumn Willow
+Name=Kaysa
```
tehdiplomat, line 2: Will changing this deck break an active save?

VigilantSera: I tested it on an existing save and it works fine. The old deck name is no longer referenced anywhere and you can't save inside the map.

tehdiplomat (approved):

PR #11524, opened 2026-08-05
Honour Scryfall rate limiting when fetching card images on desktop

liamiak:
The mobile fetcher checks the HTTP status when downloading a card image and, on a 429, backs off
Scryfall for five minutes. Desktop never got that - it calls `ImageIO.read(URL)`, which throws the
response code away, so a rate limit is indistinguishable from any other failed download and we
carry on asking. The cooldown field it would use is already declared on the shared base class;
only the mobile half was ever wired up, in 678aae47 back in February.

`SwingImageFetcher` now reads through a `URLConnection`, sends the User-Agent and Accept headers
Scryfall asks for, and records the cooldown on a 429. The cooldown logic moves up to `ImageFetcher`
so both platforms share one copy rather than mobile carrying its own.

Desktop also had no spacing between requests. Downloads are submitted to a work stealing pool, so a
screen of missing images - a deck editor grid, for instance - asks for all of them at once.
`paceScryfall` keeps a 100ms gap, the same interval `GuiDownloadService` already keeps for its bulk
downloads. Mobile already sleeps between attempts in its own loop, so it is left alone.

The pool this sleeps in has exactly two users, `ImageKeys` set lookup and `ImageFetcher` itself, so
pacing delays other image work and nothing else.

Not unit tested: the cooldown is static mutable state shared across the JVM, and setting it from a
test would leak a five minute Scryfall cooldown into the rest of the suite. Verified by build on
both the desktop and mobile modules, and the desktop suite - 351 tests, 0 failures.

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/util/SwingImageFetcher.java
+++ b/forge-gui-desktop/src/main/java/forge/util/SwingImageFetcher.java
@@ -7,8 +7,10 @@ import javax.swing.*;
 import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.net.URLConnection;
 
 public class SwingImageFetcher extends ImageFetcher {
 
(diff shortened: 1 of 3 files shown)


tool4ever:
thanks, finally respecting their limits will be good 👍

tool4ever (approved):

PR #11544, opened 2026-08-06
Restore generic tap-locking Aura AI

vanja-ivancevic:
- Restore automatic target selection for Auras that prevent enchanted permanents from untapping.
- Let the generic logic introduced in #7361 run when cards do not declare `AILogic`.
- Fix Merseine and other affected Auras without card-specific script hints.


tool4ever:
this shouldn't be needed, looks like the superior logic got broken via #7361 and should be fixed instead

vanja-ivancevic:
> this shouldn't be needed, looks like the superior logic got broken via #7361 and should be fixed instead

You’re right, thanks for pointing me to #7361. I’ve replaced the Merseine-specific workaround with a general fix and verified through a scripted game simulation that the AI now casts and resolves Merseine correctly.

tool4ever (approved):

Issue #11384, opened 2026-07-25
iOS follow-up: checklist of isAndroid() call sites to review for isMobile() migration


liminalwarmth:
With iOS support merged (#11190), `GuiBase.isAndroid()` no longer implies "not desktop" — iOS currently takes the **desktop** path at every `isAndroid()` call site. As @Jetz72 noted in [review](https://github.com/Card-Forge/forge/pull/11190#discussion_r3647553697), the existing uses need a pass to see which should apply to iOS too. #11382 adds a `GuiBase.isMobile()` helper (and converts the two sites that already spelled out `isAndroid() || isIOS()`: `Config.resPath()`, `FSkin.useFallbackDir()`).

This issue is a survey of the remaining `isAndroid()` call sites in shared code as a migration checklist. **Caveat:** `isMobile()` differs from `isAndroid()` exactly on iOS, so every conversion below is a live iOS behavior change — the current desktop-path behavior is what was device-verified in #11190, so each flip should ideally be checked on an iOS device/simulator before landing.

## Candidates (semantics look like "any touch/mobile platform")

Touch input vs. mouse pointer:
- [ ] `Forge.java:303` — `hasKeyboard()`: whether the platform has a physical keyboard (feeds `hasExternalInput()`)
- [ ] `Forge.java:376` — always-on gamepad listener vs. only-when-connected (battery/perf concession for touch devices)
- [ ] `Forge.java:475` — skips setting a desktop mouse cursor image
- [ ] `Forge.java:1516` — skips `mouseMoved`/hover-listener handling (has a "todo: mouse listener for android?" comment)
- [ ] `toolbox/FDisplayObject.java:131,136` — precise card-hover detection and hover bookkeeping gated to desktop ("TODO: mouse detection on android?")
- [ ] `screens/match/MatchScreen.java:404` — hover-based card magnifier gated to desktop-or-gamepad (`!isAndroid() || hasGamepad()`)
- [ ] `adventure/util/RewardActor.java:183,488,626,728,892,993` — repeated `isAndroid() || hasGamepad()` selecting hold-to-view tooltip (touch/gamepad) vs. hover tooltip (mouse)
- [ ] `adventure/util/RewardActor.java:1388-1391` — hold-tooltip centered on screen (touch/gamepad) vs. anchored at hover position (mouse)
- [ ] `adventure/util/KeyBinding.java:59` — hides keyboard-key glyph labels when no controller is connected (no physical keyboard to label)

[...]

PR #11556, opened 2026-08-07
iOS: follow-ups to #11190 (review feedback + fixes)

shoeless:
Follow-ups to the iOS support PR #11190 — the review responses Jetz72 asked for, plus a few bugs found during device/sim testing. Sim-verified on iPad Air (iOS 18.4), iPhone 16, and an Android tablet emulator.

## Review responses (#11190)

- **Audio: split the iOS backend into its own `IAudioClip`** (Jetz72). `AudioClip` is a pure OpenAL implementation again; the iOS AVAudioPlayer path moves verbatim into a second implementation, `MusicAudioClip`, chosen by the `createClip` factories — no more per-method `isIOS` branches.
- **`GuiBase.isMobile()` conversions** (Jetz72). Builds on the helper from #11382, converting the shared sites where Android and iOS want identical behavior (adventure data paths, fallback-skin textures, hardware-cursor skip, hover detection, PID logging). Platform-machinery sites keep their specific check; the behavior-changing sites to audit are tracked in #11384.
- **iOS `exit()`/`restart()` terminate the process** (Jetz72). The return-to-main-menu intercept meant restart-required settings (Adventure plane, skin, language) could never take effect. iOS is sideload/TestFlight distributed, so the app now exits and the user relaunches.
- **Downloaded card images load through the AssetManager** (Jetz72). The iOS bypass and its parallel caches are gone: a small `CardTextureData` does the byte read + downscale and is handed to the manager via `TextureParameter.textureData` (the stock `TextureLoader` honours a preset `textureData` and ignores the FileHandle). Deletes `pixmapCache`, the `downloadedTextureCache` LRU, `textureToPath`, the `isDownloadedImage` branch, and the `toRelativePath` rewriting. No custom loader / `setLoader`, so non-card textures are untouched.

## Bug fixes found in testing

- **Card textures render full-quality RGBA8888** — drops the RGB565 downgrade (sim peak ~1.9 GB on a 4 GB device, unchanged).
- **App icon** shipped via an `Assets.xcassets` catalog — the IPA previously had no icon data (a contributor-global `*.png` gitignore silently dropped it); adds a `.gitignore` negation so it can't recur.
- **Oversized iPhone prompt buttons** fixed via `HdpiMode.Pixels` (points-vs-pixels density mismatch). Phones only; iPad stays on `Logical`.
- **App version display** — the iOS `getVersionString()` was a hardcoded `"0.0"`; now reads the bundle version via `NSBundle` (shows `2.0.14 (build N)`).
- **`NetworkLogWriter`** — narrow a `catch (Throwable)` to `Exception` so it doesn't swallow `OutOfMemoryError` (Copilot).

## Testing
Built and driven through a game on the iPad Air 13" sim (iOS 18.4), iPhone 16 sim, and an Android tablet emulator: cards render, borders correct, version string correct, buttons sized correctly, no regressions to non-card textures.


shoeless:
Added three more fixes on top of your cleanup: the Xalan output serializer force-link (every `XmlUtil.saveDocument` call failed on iOS — achievements plus the card/deck/item-view pref writers), the net-lobby fix for #11190's per-keystroke CHANGE regression (chat sent one message per typed character; team selection echo-cascaded with 3+ players — verified with three simulators over loopback), and an iOS hardware-keyboard backspace fix (one press deleted two characters: the queued keyDown plus the UITextField delegate's keyTyped both ran).

tool4ever (approved):
thanks, good timing since we just had a full release
I will make a build with this and add it so some more people can test it

PR #11560, opened 2026-08-08
Add language support to ImageKey

churrufli:
This PR adds language support to ImageKey. This is a draft and I'm opening it for early feedback before continuing with the implementation.

### DIFF
--- a/forge-gui/res/editions/Commander 2018.txt
+++ b/forge-gui/res/editions/Commander 2018.txt
@@ -4,6 +4,7 @@ Date=2018-08-10
 Name=Commander 2018
 Type=Commander
 ScryfallCode=C18
+CardLang=en,es,fr,de,it,pt,ja
 
 [cards]
 1 R Boreas Charger @Christine Choi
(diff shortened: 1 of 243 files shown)


forge-gui/res/editions/Commander 2018.txt
```diff
@@ -139,7 +140,7 @@ ScryfallCode=C18
 131 R Bear Umbra @Howard Lyon
 132 R Boon Satyr @Wesley Burt
 133 C Borderland Explorer @Lucas Graciano
-134 R Budoka Gardener @Kev Walker
+134 R Budoka Gardener @Kev Walker ${"lang": "en"}
```
tool4ever, line 143: Is this a Scryfall error?
https://gatherer.wizards.com/C18/es-es/134/budoka-gardener

churrufli: It seems so: https://scryfall.com/card/c18/134/budoka-gardener

I wonder if Is there any Gatherer Wizards bulk data available that I could use to check whether there are other similar errors in Scryfall? Or perhaps these kinds of errors aren't particularly relevant?

Also, what do you think about adding the languages to each card line when they don't completely match "CardLang"? Or perhaps using something like "missing languages" for cards that are missing some of the languages?

paulsnoops:
I have to be honest, we had a similar thing with another PR regarding images adding bloat to the edition files before and I'm not really a fan of it. They're already a lot to maintain and this just adds to the effort. Personally, I would prefer we approach this in a different way.

PR #11572, opened 2026-08-09
Add plain-text card art language index and language-aware unique card selection

churrufli:
This replaces the approach from #11507 and follows up on the discussion started in #11560.
Instead of a generated bitmask index, this stores confirmed non-English printings in a plain text file (`forge-gui/res/languages/card-art-languages.txt`, one `SET/CN=lang,lang` line per printing), generated by `forge-gui/tools/updateCardArtLanguages.py`. No bitmask, no changes to edition files,

- New Preferences options: "Card Image Language" and "Prefer My Language for Unique Cards" (off by default, zero behavior change unless the user opts in).
- `ImageFetcher` / `GuiDownloadFilteredCardImages` try the preferred language first when confirmed, falling back to today's default otherwise.
- Same as the card name translation files (`res/languages/cardnames-*.txt`), this new file will need to be kept up to date over time. I'm happy to integrate its generation into whatever process/script this project already uses to update those. Either way, the result is plain text and editable by hand if needed.

One personal note to close: some time ago I helped with Forge's internationalization, (in the cardforge.org git) among other things. I still believe that opening the door to more languages is always a good thing for this project. I hope this gets some consideration, thank you all.


churrufli:
Closing this in favor of #10928 — it already covers the language-availability data I was trying to build, and I'd rather not maintain a parallel system. I'll follow that PR and revisit the preferred-language setting and language-aware unique cards once it lands.

PR #11581, opened 2026-08-09
test(net): guard that hidden-zone reveals reach the client delta

shoeless:
## What

Adds `RevealSyncDeltaTest` — a deterministic, in-process regression test that verifies a hidden-zone card reveal reaches a networked client through the real delta-sync path.

The test:
1. Builds an exact board with the existing `AITest` helpers (no AI, no sockets, ~10s).
2. Puts one card in a player's hidden **Library** and confirms they cannot see it.
3. Adds **Future Sight** to their battlefield; its continuous `MayLookAt$ Player` static reveals the top card (verified against `CardView.canBeShownTo`).
4. Runs the real `DeltaSyncManager.collectDeltas` and asserts the top card's `PlayerMayLook` is present in the delta the client receives.

## Why

`DeltaSyncManager.walkAndCollect` recurses into **hidden-zone** (`Library`/`Hand`) `TrackableCollection` properties and collects each nested `CardView`'s dirty props — which is what lets a player see their own revealed top card (Future Sight, Garruk's Horde, etc.) on a networked client. That behavior was **not covered by any existing test**:

- `DeltaSyncUnitTest` only checks `DeltaPacket` byte-size accounting — it never runs `collectDeltas` on a real `GameView`.
- `NetworkPlayIntegrationTest` checks visibility for **public** zones only (Battlefield/Graveyard/Exile) and is stress-gated / non-deterministic.

If a future change stopped walking hidden-zone collections, the owning player would silently stop seeing their own revealed top card — this test would catch it. It also documents a reusable pattern for "does host change X reach the client delta?" tests.

Test-only change; no engine/runtime code touched.

### DIFF
--- a/forge-gui-desktop/src/test/java/forge/net/RevealSyncDeltaTest.java
+++ b/forge-gui-desktop/src/test/java/forge/net/RevealSyncDeltaTest.java
@@ -1,0 +1,102 @@
+package forge.net;
+
+import forge.ai.AITest;
+import forge.game.Game;
+import forge.game.GameView;
+import forge.game.card.CardView;
+import forge.game.player.Player;
+import forge.game.player.PlayerView;
+import forge.game.zone.ZoneType;
+import forge.gamemodes.net.DeltaPacket;
+import forge.gamemodes.net.server.DeltaSyncManager;
+import forge.trackable.TrackableProperty;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.util.Map;
+
+/**
+ * Reusable harness for verifying that host-side game-state changes actually reach a
+ * networked client through the real delta-sync path ({@link DeltaSyncManager}).
+ *
+ * <p>Pattern for tests of this type:
+ * <ol>
+ *   <li>Build an exact board with {@link AITest} helpers (deterministic — no AI, no sockets).</li>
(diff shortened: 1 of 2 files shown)


tool4ever (commented):
imo seems too arbitrary to test that deltas would just start ignoring some zones...
besides `NetworkChecksumUtil` would quickly catch it anyway

shoeless:
I checked: `PlayerMayLook` is checksum-eligible, but `collectChecksumObjects` only walked Battlefield/Hand/Graveyard/Exile/Command — Library never entered a checksum, so a dropped reveal on a library card (the case here: Future Sight / top-of-library effects) couldn't be caught at runtime, and covered-zone detection is sampled rather than prompt.

Added a second commit that includes Library in the checksum walk so runtime does catch it — feel free to drop that commit if it's unwanted, and equally fine closing the test if you'd rather rely on the checksum alone. @MostCromulent — was leaving Library out of the checksum walk deliberate (perf/scope), or just not needed at the time?

MostCromulent:
> @MostCromulent — was leaving Library out of the checksum walk deliberate (perf/scope), or just not needed at the time?

I'm pretty sure this was just an oversight rather than a deliberate design choice.

tool4ever:
focus was on public zones since walking libraries with little dynamic properties seems rather pointless

tool4ever:
closing since it only adds overhead

PR #11582, opened 2026-08-09
Sync missing keys from en-US.properties to all other language files

churrufli:


Jetz72 (approved):
They fall back anyway, but this at least makes life easier for anyone doing translations.

PR #11583, opened 2026-08-10
Fix Universes Within double triggers (CardFace variant list aliasing)

shoeless:
## Problem

`CardFace.assignMissingFieldsToVariant` assigned the base face's keyword / ability / static-ability / trigger / replacement list **references** into a flavor-name variant instead of copying them:

```java
if (variant.keywords == null) variant.keywords = this.keywords;   // aliases the base list
else variant.keywords.addAll(0, this.keywords);
```

When a second variant then ran `variant.keywords.addAll(0, this.keywords)` — where `variant.keywords` **is** `this.keywords` — it appended the list to itself, **doubling every keyword/ability/trigger/replacement**. On Universes Within / flavor-name cards this made triggers fire twice (e.g. **Blanka**, **The Howling Abomination**) and, because the base list is shared, **permanently corrupted the base `CardRules` for every printing**.

## Fix

Copy the base lists into the variant (`new ArrayList<>(...)`) instead of aliasing, and guard the `addAll` with a `!= this.x` self-append check. Root-cause fix — supersedes the earlier per-card script band-aid.

Single file, `forge-core` only. Compiles clean.

### DIFF
--- a/forge-core/src/main/java/forge/card/CardFace.java
+++ b/forge-core/src/main/java/forge/card/CardFace.java
@@ -223,20 +223,25 @@ final class CardFace implements ICardFace, Cloneable {
         if("".equals(variant.initialLoyalty)) variant.initialLoyalty = this.initialLoyalty;
         if("".equals(variant.defense)) variant.defense = this.defense;
 
-        if(variant.keywords == null) variant.keywords = this.keywords;
-        else variant.keywords.addAll(0, this.keywords);
+        // COPY the base lists into the variant (don't alias): assigning the base list reference and
+        // then addAll'ing into it on a SECOND variant call does this.list.addAll(0, this.list),
+        // doubling every keyword/ability/trigger/replacement on Universes Within / flavor-name cards
+        // (Blanka, The Howling Abomination fired twice) and permanently corrupting the shared base
+        // CardRules for every printing. The `!= this.x` guard also prevents a self-append.
+        if(variant.keywords == null) variant.keywords = new ArrayList<>(this.keywords);
+        else if(variant.keywords != this.keywords) variant.keywords.addAll(0, this.keywords);
 
-        if(variant.abilities == null) variant.abilities = this.abilities;
-        else variant.abilities.addAll(0, this.abilities);
+        if(variant.abilities == null) variant.abilities = new ArrayList<>(this.abilities);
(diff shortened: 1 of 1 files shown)


forge-core/src/main/java/forge/card/CardFace.java
```diff
@@ -223,20 +223,25 @@ void assignMissingFieldsToVariant(CardFace variant) {
         if("".equals(variant.initialLoyalty)) variant.initialLoyalty = this.initialLoyalty;
         if("".equals(variant.defense)) variant.defense = this.defense;
 
-        if(variant.keywords == null) variant.keywords = this.keywords;
-        else variant.keywords.addAll(0, this.keywords);
+        // COPY the base lists into the variant (don't alias): assigning the base list reference and
+        // then addAll'ing into it on a SECOND variant call does this.list.addAll(0, this.list),
+        // doubling every keyword/ability/trigger/replacement on Universes Within / flavor-name cards
+        // (Blanka, The Howling Abomination fired twice) and permanently corrupting the shared base
+        // CardRules for every printing. The `!= this.x` guard also prevents a self-append.
```
Jetz72, line 230: Good catch, but can we leave off the five-line comment describing a bug that will no longer exist once we've merged this?

tool4ever: this still makes no sense to me:
`assignMissingFieldsToVariant` should only be called once for each face...
so why would it matter that it can point to its parent fields? 🤷‍♂️

Jetz72: Actually yea, now that I think about it at some time other than 5 in the morning, this feels like a problem that could only be caused by faulty logic elsewhere. Possibly some part of `findOrCreateVariantForFlavorName` not working right with in-script flavor names + flavor name references in the edition file? 

If anything, I think `assignMissingFieldsToVariant` should be throwing an assertion error if it's somehow called twice.

shoeless:
Closing this one — you're both right that no in-tree path hits it, and standalone it lacks motivation. It's actually a prerequisite for a serialized-CardRules startup cache I've been preparing (re-finalizing restored rules is exactly the "second call" scenario), so I'll fold it into that PR — along with the assertion Jetz72 suggested, so a second merge fails loudly instead of silently doubling.

PR #11584, opened 2026-08-10
Fix CardType.getSortedSubTypes multi-thread init race (TimSort crash)

shoeless:
## Problem

`CardType.getSortedSubTypes()` lazily builds a `static` list, but it assigned the field **first** and then sorted it **in place**:

```java
sortedSubTypes = Lists.newArrayList();
sortedSubTypes.addAll(...);   // many
Collections.sort(sortedSubTypes);   // mutates the already-published field
```

Two threads racing the first-time init (parallel AI evaluation, headless multi-thread sims) could observe the half-built field and mutate it mid-sort, tripping TimSort's `IllegalArgumentException: Comparison method violates its general contract` and crashing the game/eval (seen via `CostExile` / type-text paths).

## Fix

Build and sort a **local** list, publish it only once fully sorted, and `synchronized` the initializer (field made `volatile`). Behavior-neutral — the returned sorted list is identical; only the first-init race is removed. Found via a multi-threaded DeckBattler gauntlet.

Single file, `forge-core` only.

### DIFF
--- a/forge-core/src/main/java/forge/card/CardType.java
+++ b/forge-core/src/main/java/forge/card/CardType.java
@@ -899,18 +899,20 @@ public final class CardType implements Comparable<CardType>, CardTypeView {
     private static List<String> sortedSubTypes;
     public static List<String> getSortedSubTypes() {
         if (sortedSubTypes == null) {
-            sortedSubTypes = Lists.newArrayList();
-            sortedSubTypes.addAll(Constant.BASIC_TYPES);
-            sortedSubTypes.addAll(Constant.LAND_TYPES);
-            sortedSubTypes.addAll(Constant.CREATURE_TYPES);
-            sortedSubTypes.addAll(Constant.SPELL_TYPES);
-            sortedSubTypes.addAll(Constant.ENCHANTMENT_TYPES);
-            sortedSubTypes.addAll(Constant.ARTIFACT_TYPES);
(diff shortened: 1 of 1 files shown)


tool4ever (changes requested):
trying to make everything thread safe (manually) just complicates things
especially when the scenarios to cause it seem rather experimental 🤔
since this method isn't even used that heavily and the sorting rarely matters some better optimization should be possible

Hanmac:
@tool4ever also we might look into this method, because right now, it can contain duplicates
(because of types like "Spacecraft")

shoeless:
Fair — reworked without any manual locking. The getter now builds a `TreeSet` (sorted + de-duplicated in one step) and publishes `ImmutableList.copyOf(tmp)` to a plain field. The race dies from construction order rather than synchronization: the list is complete before the field is assigned, and ImmutableList's all-final-field internals make even a racy first read safe — worst case two threads build identical copies and one wins. Net −1 line vs the old code.

The TreeSet also covers @Hanmac's point: Spacecraft (listed in two sections) no longer shows up twice in the Advanced Search subtype list.

It stays lazy because the type lists are populated at runtime (FModel init + custom editions) — an eager initializer would capture empty sets. All callers only read the list, so the immutability can't bite anyone.

tool4ever (approved):

Hanmac (approved):

Hanmac (approved):

PR #11585, opened 2026-08-10
Fix library cards showing as card backs for guest in multiplayer

shoeless:
## Problem

`CardView.mayPlayerLook` looked up the may-look `PlayerView` collection with `indexOf` (reference equality):

```java
return col != null && col.indexOf(pv) != -1;
```

In network play the client's local `PlayerView`s are distinct objects from the deserialized server ones, so the lookup always missed. Cards a player should be able to see — their own revealed library top (Future Sight/Garruk's Horde), opponent-hand reveals, etc. — rendered as **card backs** for the guest.

## Fix

Compare by `getId()` instead of object reference. Single file, `forge-game` only.

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardView.java
+++ b/forge-game/src/main/java/forge/game/card/CardView.java
@@ -622,8 +622,12 @@ public class CardView extends GameEntityView {
     }
     public boolean mayPlayerLook(PlayerView pv) {
         TrackableCollection<PlayerView> col = get(TrackableProperty.PlayerMayLook);
-        // TODO don't use contains as it only queries the backing HashSet which is problematic for netplay because of unsynchronized player ids
-        return col != null && col.indexOf(pv) != -1;
+        if (col == null) { return false; }
+        // compare by id, not reference: MP client/server PlayerViews are distinct objects
+        for (PlayerView p : col) {
+            if (p.getId() == pv.getId()) { return true; }
+        }
+        return false;
     }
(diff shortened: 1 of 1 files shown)


forge-game/src/main/java/forge/game/card/CardView.java
```diff
@@ -622,8 +622,12 @@ void setMayPlayPlayers(Iterable<Player> list) {
     }
     public boolean mayPlayerLook(PlayerView pv) {
         TrackableCollection<PlayerView> col = get(TrackableProperty.PlayerMayLook);
-        // TODO don't use contains as it only queries the backing HashSet which is problematic for netplay because of unsynchronized player ids
-        return col != null && col.indexOf(pv) != -1;
+        if (col == null) { return false; }
```
tool4ever, line 625: can't reproduce

shoeless:
Confirmed — redundant on master, closing. `indexOf` already compares by id (`TrackableObject.equals` is final: same class + same id — your 1ae39dd270f from 2022), so this explicit loop is semantically identical to current behavior. Traced the history: I originally hit the card-backs symptom on a branch based on January master — before #9642 replaced the client sync path with delta sync — and it doesn't reproduce on the current stack, as you found.

One adjacent real find while verifying: `canBeShownTo` compares `mindSlaveMaster == viewer` by reference (~lines 697/719) — that one does break for a deserialized viewer in netplay mind-control scenarios. Happy to PR it separately if there's interest.

PR #11586, opened 2026-08-10
CardDb: make loadCard idempotent (fix latent duplicate-PaperCard bug)

shoeless:
## Problem

`CardDb.loadCard` is not idempotent: `addSetCard` unconditionally increments the art index and appends a fresh `PaperCard`. A redundant call for a `(name, set)` printing that's already loaded therefore **duplicates** that `PaperCard` in `allCardsByName`.

## Fix

Skip when that exact `(name, set)` printing already exists, via a new side-effect-free `hasPrintingInSet` check. Guarding on card-name presence alone would break the "add a printing from a not-yet-loaded set" path (`StaticData.getOrLoadCommonCard`), where the card can already exist under other sets while the requested set's printing is still missing — so the guard is keyed on the specific edition.

## Why "any printing present" is a safe guard despite multi-printing sets

A set **can** contain multiple printings of the same name (`CardEdition.getCardInSet` returns a `List<EditionEntry>` — basic-land arts, showcase variants; that's what `artIndex` is for). The guard is still sound because `loadCard` is **all-or-nothing per `(name, set)`**: `addFromSetByName` loops over every `EditionEntry` of that name in the set and `addSetCard`s each one. `addSetCard` has exactly two callers — that loop, and the eager full-set path in `initialize` — so no code path ever adds a *subset* of a name's printings for a set. Any-printing-present therefore implies the `(name, set)` load already ran, art variants included. (Entries excluded by the functional-variant filter are excluded on every call equally, so the invariant holds.)

Single file, `forge-core` only.

### DIFF
--- a/forge-core/src/main/java/forge/card/CardDb.java
+++ b/forge-core/src/main/java/forge/card/CardDb.java
@@ -445,6 +445,12 @@ public final class CardDb implements ICardDatabase, IDeckGenPool {
                 return;
             }
         }
+        // addSetCard appends unconditionally, so skip if this card's printing from this set
+        // is already loaded (presence in other sets isn't enough — loadCard also adds new-set printings)
+        CardEdition guardEd = editions.get(setCode);
+        if (guardEd != null && !guardEd.equals(CardEdition.UNKNOWN) && hasPrintingInSet(cr, guardEd)) {
+            return;
+        }
         boolean reIndexNecessary = false;
         CardEdition ed = editions.get(setCode);
         if (ed == null || ed.equals(CardEdition.UNKNOWN)) {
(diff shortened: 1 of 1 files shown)


Hanmac:
artIndex is on my list to be removed anyway 

I just need to find a better way to rename the image files #7771

shoeless:
Good to know — this guard doesn't lean on artIndex, only on loadCard adding all of a (name, set)'s printings in one call, so it should carry over cleanly (or simplify) when artIndex goes away with #7771.

Pushed a small hardening while re-checking it: the existing-printing check now also matches the card's rules name, since a lookup name can be shared across different cards (CopyFaceFrom alt faces, flavor names) — no stock data hits that today, but custom content could.

To be clear, the duplication itself isn't artIndex's doing — `allCardsByName` is an ArrayList-backed multimap and nothing in the add path checks for an existing entry, so a redundant call appends regardless; artIndex only determines how the duplicates are labeled. The guard becomes droppable once that rework makes the add path uniqueness-enforcing.

forge-core/src/main/java/forge/card/CardDb.java
```diff
@@ -462,6 +468,21 @@ public void loadCard(String cardName, String setCode, CardRules cr) {
         }
     }
 
+    /** True if a printing of this card from edition is already loaded (no side effects, unlike getCardFromSet). */
+    private boolean hasPrintingInSet(String cardName, CardRules cr, CardEdition edition) {
+        String code1 = edition.getCode(), code2 = edition.getCode2();
+        for (PaperCard pc : getAllCards(cardName)) {
+            if (!pc.getRules().getName().equalsIgnoreCase(cr.getName())) {
+                continue; // name shared with a different card (alt face / flavor name)
+            }
```
Jetz72: Iterating over `getAllCards(cr)` will get you all printings of the given CardRules.

shoeless: Much cleaner — switched to `getAllCards(cr)`, which drops the name-equality check entirely (the rules-keyed lookup can't cross cards). Thanks!

PR #11588, opened 2026-08-10
iOS pipeline: self-recovering simulator builds + docs

shoeless:
## Problem

RoboVM 2.3.24's `robovm:ipad-sim` mojo launches the app after the AOT link and then blocks in `SimLauncherProcess.waitFor` until the app quits — and on Apple Silicon it picks its own simulator (its device-type table predates arm64 sims). Net effect: **every scripted simulator build hung at the launch step** and needed the wedged maven process killed by hand.

## Fix

All the pipeline needs from the mojo is `target/robovm.tmp/config.xml`, written once the AOT link finishes. So `sim` mode now:

1. runs the mojo in its own process group with output to a log,
2. waits for `config.xml` to appear,
3. stops the process group right before the launch step it would wedge in,
4. rebundles with the standalone `AppCompiler` via the existing `assemble_arm64_sim_app` (the AOT cache is content-hashed — nothing recompiles).

Also adds `pipeline/README.md` documenting the transform stages (JvmDowngrader → MobiVmBridge → MobiVmLinkAudit) and the CI `audit` mode, plus comment-only clarifications in the two pipeline helpers.

## Verification

Ran `pipeline/ios-pipeline.sh classpath && pipeline/ios-pipeline.sh sim` on this branch against current master: the build completed **build → install → launch on the simulator fully unattended** — previously the same invocation required manually detecting and killing the wedged mojo on every single build.

Build-tooling only; no app/runtime code touched.

### DIFF
--- a/forge-gui-ios/pipeline/ios-pipeline.sh
+++ b/forge-gui-ios/pipeline/ios-pipeline.sh
@@ -451,11 +451,33 @@ sim() {
     trap 'mv -f "$OSLOG_LIB.committed" "$OSLOG_LIB" 2>/dev/null || true' EXIT
     build_oslog sim
     echo "=== robovm ipad-sim build ($SIM_ARCH) ==="
-    # On Apple Silicon (arch=arm64-simulator) this fails at the mojo's device
-    # selection AFTER producing the binary + config.xml — expected; we assemble
-    # the .app ourselves below. On Intel (x86_64) it produces the .app directly.
-    (cd "$ROOT/forge-gui-ios" && mvn robovm:ipad-sim --settings "$SETTINGS" \
-        -Dmaven.repo.local="$CLONE" -Drobovm.arch="$SIM_ARCH" -DskipTests 2>&1 | tail -8) || true
+    if [ "$SIM_ARCH" = "arm64-simulator" ]; then
+        # The mojo blocks in its own app launch after the AOT link (and picks its own
+        # simulator). All we need from it is config.xml, written once the link is done:
+        # stop it there and let assemble_arm64_sim_app rebundle with the AppCompiler
+        # (the AOT cache is content-hashed, so nothing recompiles).
+        local TMPD="$ROOT/forge-gui-ios/target/robovm.tmp" mvnpid
+        # not under robovm.tmp: the mojo wipes that directory after mvn has opened the log
+        local MVNLOG="$ROOT/forge-gui-ios/target/ipad-sim.log"
+        mkdir -p "$TMPD"
+        rm -f "$TMPD/config.xml"
+        set -m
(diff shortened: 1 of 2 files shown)


forge-gui-ios/pipeline/README.md
```diff
@@ -0,0 +1,84 @@
+# iOS modern-Java pipeline
```
tool4ever: I already moved this to `Docs/Development/iOS-Builds.md`

forge-gui-ios/pipeline/ios-pipeline.sh
```diff
@@ -512,7 +556,11 @@ EOF
     echo "INSTALLED"
 }
 
-# CI modes audit: the full transform + link-gate WITHOUT any Apple tooling.
+# ------------------------------------------------------------------ CI modes
```
tool4ever: redundant changes (in this whole file)

forge-gui-ios/pipeline/src/MobiVmBridge.java
```diff
@@ -358,11 +358,12 @@ static byte[] rewriteClass(InputStream is) throws Exception {
      * only FLAG_MARKERS(2) and FLAG_BRIDGES(4) of LambdaMetafactory.altMetafactory
      * — FLAG_SERIALIZABLE(1) is silently discarded, so a serializable lambda
      * comes out NOT implementing Serializable and the compiler-emitted
-     * `checkcast java/io/Serializable` throws at runtime.
-     * Convert the ignored bit into an explicit marker interface,which RoboVM
-     * demonstrably honors. Bit 1 is left set: RoboVM ignores it either way and
-     * flags=7 + explicit marker is the (spec-valid) historical ECJ encoding,
-     * so the transformed jars still run on real JVMs.
+     * `checkcast java/io/Serializable` throws at runtime (first hit: jgrapht
```
tool4ever: don't need this extra gossip again

tool4ever:
By the way, I assume the code block in "ios/Main.java" related to `cardcache.bin` is obsolete?

shoeless:
Trimmed to just the functional change — the README and comment churn are gone (the doc you moved to `docs/Development/iOS-Builds.md` covers it). Two corrections from re-testing: the early-stop is now gated to Apple Silicon — on Intel the mojo bundles the .app during its launch phase, so stopping at config.xml would have broken those builds — and the wait tests `-s` with a settle sleep, since the mojo opens config.xml before serializing into it.

And correct about `cardcache.bin` — nothing in the repo writes those files (they're left over from a card-cache experiment on my iOS branch); dropped the block in this PR.

tool4ever (approved):

PR #11592, opened 2026-08-10
Fix unselectable tokens after game 1 of a match

MostCromulent:
## Summary

In a networked match, the joining player could not select tokens for blocking or targeting in any game after the first. The network layer now rebuilds its per-game state at each new game, restoring token selection and correcting related state that also survived past a game boundary.

## Bug report

Reported from a Legacy match on the mobile/adventure Classic UI: neither player's tokens could be selected for blocking or targeting. It affected only the joining player — the host could select tokens in the same games — and only showed up in matches played with sideboards. An earlier symptom was an Orcish Bowmasters trigger that could target one token but not the others.

The host's log shows the failure directly, with the warning immediately preceding each click the client sent:

    21:41:08 [WARN ] TrackableSerializer: Could not resolve IdRef(tag=0, id=122) from Tracker
    21:41:08 [INFO ] GameProtocolHandler: Received: GuiGameEvent 2258: selectCard (3 args)
    21:41:09 [WARN ] TrackableSerializer: Could not resolve IdRef(tag=0, id=121) from Tracker
    21:41:09 [INFO ] GameProtocolHandler: Received: GuiGameEvent 2259: selectCard (3 args)

Ids 118/119/121/122 are Spirit Tokens, confirmed by the declared-attackers entry in the same log. Game 1 produced zero of these warnings; game 2 produced 72.

## Root cause

Each `Game` owns its own `Tracker` (`Game.java:144`) and restarts card ids at 1, and `HostedMatch` starts every game of a match by calling `setGameView` on the same GUI instance. Several pieces of network state were latched to the first call instead of following the game:

- `RemoteClientGuiGame.codecTrackerSet` installed the codec `Tracker` once, so from game 2 the server decoded client `IdRef`s against game 1's tracker.
- `initialSyncSent` was cleared only on reconnect, so game 2's first update was a delta computed against a baseline the client no longer had.
- `GameClientHandler.tracker` was likewise created once per connection.
- `AbstractGuiGame.setGameView` merges rather than replaces whenever both views are non-null. The host forces replacement by passing `null` first, but the server swallows that half, so the client merged the new game into the finished one.

When the client clicked a token, its `CardView` was encoded as an `IdRef` and resolved on the server against the stale tracker. Tokens are hit hardest because their ids are allocated last and so are least likely to exist in the previous game's tracker; ordinary cards reuse ids the previous game also held, so they resolve to a stale view whose id still leads `findByView` to the correct card. That is why only tokens appeared broken.

## Fix

Key all per-game network state to game identity rather than to connection lifetime.

- `RemoteClientGuiGame` tracks the `Tracker` it last synced against; a different one means a new game, which re-applies the codec tracker and calls `resetDeltaSync` (the existing reconnect reset, renamed, since it now serves both callers).
- `RemoteClient.applyCodecTracker` performs the pipeline swap on the channel's event loop.
- `AbstractGuiGame.setGameView` takes the direct-assign path when the incoming view belongs to a different game, not only when a side is null.
- `GameClientHandler` keys its `Tracker` to the incoming `GameView` id.

## Why this is architecturally consistent

- `GameView.getId()` is `Game.getId()`, from a monotonic counter — unique across games, stable within one, and preserved through serialization, so both sides key on the same signal.
- The host path is unchanged: `HostedMatch` already brackets every switch with `setGameView(null)`, so the id comparison is a no-op there and only affects the remote client, which never receives that clear.
- Replacing the client's view graph orphans nothing — both GUIs already rebuild their card panels at `openView`, mobile via `CardAreaPanel.resetForNewGame()` and desktop via `initMatch` plus `FloatingZone.closeAll()`. The identity-preservation rationale for merging applies within a game, where ids denote the same cards.
- The codec swap moves to the event loop because subgame transitions call `setGameView` after `openView` on a live connection. The decoder's tracker is already `volatile`, so this is ordering, not visibility.
- The tracker is deliberately not cleared on the `null` half of a transition, so a late message from the outgoing game still resolves. `resetDeltaSync` rewinds the delta sequence, which the client reads only for logging.

## Testing completed

Verified with a throwaway TestNG harness, run against this branch and against master. The harness is not part of this PR.

- A token created in game 2 resolves through the server's codec tracker here; on master it resolves to `null` — the value `TrackableSerializer.resolve` hands to `selectCard`.
- The codec tracker points at the current game's `Tracker` after a switch; on master it still points at the finished game's.
- A view from a different game replaces the existing one rather than merging into it.

All three fail on master and pass here.

---


forge-gui/src/main/java/forge/gamemodes/net/server/RemoteClientGuiGame.java
```diff
@@ -295,12 +296,14 @@ public void sendFullState() {
...
-        // setGameView is called before openView, and the client can't respond
-        // until after openView — so the encoder/decoder are ready in time.
-        if (!codecTrackerSet && gameView != null && gameView.getTracker() != null) {
-            client.setCodecTracker(gameView.getTracker(), syncManager.getConsumerId());
-            codecTrackerSet = true;
+        // Each game owns a Tracker and restarts card ids at 1, so a new one means the
+        // delta baseline no longer applies and client IdRefs must resolve against it.
+        // A null view is a transient clear between games; leave both pointing at the
+        // outgoing game so its in-flight messages still resolve.
+        if (gameView != null && gameView.getTracker() != lastSyncedTracker) {
```
tool4ever: what about `RemoteClientGuiGame.openView`?
it should always send a full packet each game anyway?
but maybe the order is wrong?

tool4ever (approved):

PR #11593, opened 2026-08-11
Update cardnames translations

churrufli:
Update cardnames translations files. Adds updateCardnameTranslations.py (forge-gui/tools/), which fills in missing/empty entries in cardnames-{es-ES,de-DE,fr-FR,it-IT,pt-BR}.txt using SF official printed name/type/text - without touching anything already translated.

Compares Forge's card/face names (from res/cardsfolder) against each translation file, fills only what's missing or empty, and respects the existing -patch.txt override convention. --dry-run and --lang flags included.


forge-gui/tools/updateCardnameTranslations.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# Fills in missing/empty entries in cardnames-*.txt from Scryfall's official
```
tool4ever, line 4: what's the difference to `cardnamesTranslations.py`?

churrufli: cardnamesTranslations.py regenerates the whole file every run, this one only fills in what's missing/empty and leaves the rest untouched - smaller, reviewable diffs.
That's actually the main reason I wrote it this way rather than reusing cardnamesTranslations.py directly: I didn't want to risk overwriting any translation someone already fixed by hand just because a full regen ran again.
Would you rather I update the existing cardnamesTranslations.py to work this way too, or skip touching/add any script in tools/ and just submit the translations this one generated?

tool4ever (approved):

PR #11606, opened 2026-08-13
LibGDX Update 1.14.2

kevlahnota:


kevlahnota:
@shoeless can you check why the ios gateway check is failing for the missing java-time-supply.jar?
image

Hanmac (approved):

PR #11612, opened 2026-08-14
Validate pr3

PtrckAraujo:


MostCromulent:
@PtrckAraujo Seems like this may have been intended for your own fork rather than Forge master?

PtrckAraujo:
Sorry for the issue. This was caused by an agent misbehaving. I’ll look into it to prevent it from happening again.

PR #11619, opened 2026-08-14
Move hardcoded Adventure Mode strings to language files

churrufli:
Some UI strings were still hardcoded in Java. This moves them into the .properties files, following the same pattern used elsewhere in the module.

Added all keys to en-US.properties.
Added Spanish translations to es-ES.properties.
Added the English text as placeholders to the other 8 languages.
No behavior changes apart from the new Spanish translations.

This only covers strings that was hardcoded in Java. The larger amount of narrative text in res/adventure/ is a huge separate task.


forge-gui-mobile/src/forge/adventure/scene/NewGameScene.java
```diff
@@ -464,98 +464,98 @@ private void showModeHelp() {
...
-                        summaryText.append("On your currently selected difficulty, Easy, you will receive three jumpstart packs of your chosen color.");
+                        summaryText.append(Forge.getLocalizer().getMessage("advDiffEasyStandard"));
                         break;
                     case "Normal":
-                        summaryText.append("On your currently selected difficulty, Normal, you will receive two jumpstart packs of your chosen color and one of an allied color.");
+                        summaryText.append(Forge.getLocalizer().getMessage("advDiffNormalStandard"));
                         break;
                     case "Hard":
-                        summaryText.append("On your currently selected difficulty, Hard, you will receive one jumpstart pack of your chosen color and one of an allied color.");
+                        summaryText.append(Forge.getLocalizer().getMessage("advDiffHardStandard"));
```
Jetz72: In long methods that need many localizer calls, I'd suggest calling `Forge.getLocalizer()` once and stashing it in a variable.

churrufli: Okay, I’ll take a look at it (next week, I’m on vacation right now). Thanks!

churrufli: Addressed the review comment about calling `Forge.getLocalizer()` once and reusing the reference in methods with many localizer calls.

Applied this pattern to every method in this PR's diff that had 3+ calls to `Forge.getLocalizer()`:

- `NewGameScene.showModeHelp()` (30 calls)
- `SettingsScene` constructor (57 calls), `copyNewPlane()` (4), `restartForge()` (3)
- `AdventureQuestController.getQuestNPCResponse()` (6), `randomItemName()` (10)
- `DuelScene.enter()` (5), `showAnteCardPopup()` (3)
- `EventScene` constructor (3)
- `AdventureEventData.getDescription()` (4)

Each of these now declares `Localizer localizer = Forge.getLocalizer();` once at the top of the method and reuses `localizer` for all subsequent `.getMessage(...)` calls. No behavior changes, added the `forge.util.Localizer` import where needed. Left one-off/two-off calls elsewhere untouched since the pattern doesn't add much value there.

Jetz72 (approved):

PR #11621, opened 2026-08-15
Adventure Mode: Internationalization groundwork + Spanish Shandalar localization

churrufli:
(edited: 04/09/2026)
## Summary
Lays the groundwork for translating Adventure Mode and delivers a complete Spanish (es-ES) translation of Shandalar, the default/flagship world:
* Moves a small number of hardcoded English UI strings in Adventure Mode Java files to the existing `Localizer`/`.properties` system.
* Adds full Spanish translation for Shandalar's narrative (quests-es-ES.json, shops-es-ES.json). Uses the existing per-language file mechanism in Config.getFile()—previously unused—to transparently load the Spanish version when UI language is es-ES, with automatic fallback to English. No changes to English source files. The pattern is now ready for other languages/worlds without extra engineering.
* Adds a translation mechanism for NPC/enemy dialog embedded directly in `.tmx` map files. Each dialog node in `DialogData` can carry an optional `loctext`/`locname` key, placed inside the node itself rather than derived from its position in the tree, so reordering, inserting, or removing dialog options never desynchronizes an existing translation. If a key is missing or untranslated, the node falls back straight to the English text already written in the `.tmx`, so partial translation coverage can never break anything and English-only devs never have to touch localization at all.
* Adventure translations are loaded explicitly by Adventure Mode code (selecting a plane, loading a save, starting a new game), never automatically by the core `Localizer`, which stays fully unaware of Adventure Mode. Each plane keeps its own, independent translation bundle. (As required).
* Every translated file for a plane (dialog, quests, shops) now lives together in one place: `forge-gui/res/adventure/<World>/languages/`. No central directory gets cluttered as more languages are added. (As required).
* A small, curated set of purely generic navigation labels (e.g. "Continue", "Leave") share a single key across nodes, cutting the translation file's size substantially; everything else keeps an independent key by default to avoid any risk of two unrelated lines silently affecting each other. (As required).
* Adds `adventure_i18n.py` (`forge-gui/tools/`) plus a guide (`README-adventure_translations.md`) covering the full process for starting a translation on any world, from scratch or building on existing work. (As required).
* All `.tmx` files containing dialog across Shandalar's common maps are translated to es-ES, giving Shandalar full narrative coverage in Spanish.
## Scope
Only Shandalar (the default/flagship world, `common/`) is covered.
## About me
I'm churrufli, from Spain, and I've known Forge since 2017 (check SlightlyMagic for it). I contributed (back in the pre-AI era) to, among other things, the Forge multi-language UI, Net Decks and Net Deck Archive, deck extraction scripts, Skins, and Windows tools for Forge (now discontinued). Currently focused mainly on Forge's internationalization.
## About the contribution
I used Claude Opus 5 as support in files: `Localizer.java`, `Config.java`, and the Python script to explore and evaluate different implementation approaches. The design and key decisions were mine, but a portion of the code in those files was generated and then adapted by me. I originally drafted this text in Spanish and had it translated, so I apologize for any stylistic roughness.


Jetz72:
How well does this translation logic hold up in the event the original map or dialog tree is altered, and the person altering it forgets or doesn't know to adjust the translation data accordingly? How easy would it be for us to accidentally create a significant error in a translation that goes unnoticed for months or years?

churrufli:
@Jetz72 The keys are positional, so you're completely right. Reordering a dialog tree silently resolves to the wrong text, with nothing to catch it. I'd rather rework it than patch around it: using Localizer and putting the key inside the node so it survives reordering by design. Give me some time and I'll come back with a revised proposal. Thanks for the review.

churrufli:
@Jetz72 I've got this implemented now. The core idea is still that devs who only touch English don't change anything about their current workflow.

Each dialog node can carry an optional key (loctext/locname) pointing to its translation. If it doesn't exist, the game just falls back to the English text that's already in the .tmx, no extra steps needed. So if someone creates a new NPC and writes its dialogue in English without thinking about localization at all, it simply shows up in English until it gets translated.

Translations also live in separate files (adventure-<language>.properties), outside the .tmx files and outside the general UI .properties, so a dev editing maps never steps on a translator's work, and vice versa.

forge-gui/res/adventure/common/maps/map/maze/maze_2.tmx
```diff
@@ -44,11 +44,18 @@
...
+			{
+				&quot;deleteMapObject&quot;: 58
+			},
+			{
+				&quot;deleteMapObject&quot;: 59
+			},
+			{
+				&quot;deleteMapObject&quot;: -1
+			}
+		],
```
Jetz72: Replacing the formatting kinda makes it difficult to see what changed.

forge-gui/res/languages/adventure-es-ES.properties
```diff
@@ -0,0 +1,1169 @@
...
+adv.common_maps_map_main_story_explore_library_of_varsil_4.33.d.n1.name=(Continuar)
+adv.common_maps_map_main_story_explore_library_of_varsil_4.35.d.n0.text=Ante ti se alza uno de los constructos más grandes que has visto jamás. A diferencia de los gólems de abajo, su forma parece menos inspirada en otros humanoides y más una silueta única y propia.
+adv.common_maps_map_main_story_explore_library_of_varsil_4.35.d.n1.text=Detrás de la máquina ves los restos de muchos libros arrancados de las estanterías de abajo, ahora hechos pedazos.
+adv.common_maps_map_main_story_explore_library_of_varsil_4.35.d.n2.name=(Continuar)
+adv.common_maps_map_main_story_explore_library_of_varsil_4.35.d.n3.text=(Nota: Este encuentro es difícil y opcional)
+adv.common_maps_map_main_story_explore_library_of_varsil_4.35.d.n4.name=(Continuar)
+adv.common_maps_map_main_story_explore_library_of_varsil_4.35.d.n5.name=(Continuar)
```
Jetz72: It looks like a handful of common strings (e.g. "Continue") are repeated often. These could probably share a key?

forge-gui/res/adventure/Shandalar/languages/quests-es-ES.json
Jetz72, line 1: Might dump these files in a `/languages/` folder or something so they don't start cluttering up the directory as more languages are added.

Jetz72 (commented):
This version keeps the process of introducing new English content straightforward, but I'm trying to picture what that'd look like for anyone doing a new translation. If there's already an existing translation in a language the translator already speaks (and they don't mind generational loss), the easiest way would be to duplicate the existing file and translate that line-by-line. If not, then you have to go through the adventure content to track down all the English text and keys (or add them if it's the first time a plane is being translated) before the text can be translated.

Any thoughts on that process at this stage?

forge-core/src/main/java/forge/util/Localizer.java
```diff
@@ -167,12 +171,27 @@ public void setLanguage(final String languageRegionID, final String languagesDir
...
         }
     }
 
+    private String lookup(final String key, final boolean forceEnglish) {
+        if (!forceEnglish && adventureBundle != null) {
+            try {
+                return adventureBundle.getString(key);
+            } catch (final MissingResourceException ignored) {}
+        }
+        return (forceEnglish ? englishBundle : resourceBundle).getString(key);
```
tool4ever, line 204: not sure I like this idea of having to scan adventure assets for other game modes

also seems like this wouldn't work well with other planes?

forge-gui/res/languages/adventure-es-ES.properties
churrufli: Resolved. A curated set of generic navigation labels (Continue, Leave...) now share a real key; the .properties file dropped from 1,169 to 749 lines.

forge-gui/res/adventure/Shandalar/languages/quests-es-ES.json
churrufli: resolved. Each world now has a single <World>/languages/ folder.

forge-core/src/main/java/forge/util/Localizer.java
churrufli: 1.- resolved. Loaded explicitly only from Adventure Mode code, never automatically.
2.- Fixed — each plane now loads its own translation bundle explicitly, from that plane's own languages/ folder. A plane with no translation for the active language simply falls back to English.

churrufli:
@Jetz72 @tool4ever Thanks so much for the thorough review, really appreciate the feedback. I've made the requested changes; whenever you get a chance to take another look, it'd mean a lot. Thank you! 

I also added adventure_i18n.py and a guide (README-adventure_translations.md) in forge-gui/tools/, covering the full process for starting a new translation from scratch.

forge-gui-mobile/src/forge/adventure/util/Config.java
```diff
@@ -203,6 +207,19 @@ public String getCommonFilePath(String path) {
         return commonPrefix + path;
     }
 
+    private String langFilePath(String fullPath, String rootPrefix) {
+        //Translated files live in ONE "languages" folder at the plane's own
+        //root (e.g. ".../Shandalar/languages/"), not scattered next to each
+        //individual original file, so there's always a single predictable
+        //place to find a plane's translations regardless of where in its
+        //own folder structure the original file happens to live.
+        //e.g. "world/quests.json" -> "<planeRoot>/languages/quests-es-ES.json"
```
Jetz72: I assume you're using an LLM based on, among other things, comments like these. If so - disclosure of its involvement should be included in the PR. 

Either way, comments documenting the rationale or details behind the implementation should generally be kept concise unless something complicated or unintuitive needs explaining.

Jetz72:
Seems good to me. Just needs the merge conflict resolved, unless anyone else has any objections.

Jetz72 (dismissed):

Jetz72 (approved):

Issue #11409, opened 2026-07-26
CardTrait caching


Hanmac:
Card Traits are updated multiple times via

getChangedCardTraitsList and their functions like `applyReplacementEffect`

inside, ICardTraitChanges apply their stuff to all of the four categories. (+1 for Keywords?)

combine the apply function into one

and then store the result as cache like KeywordCache

Hanmac:
@tool4ever @liamiak @MostCromulent no MR yet, just brainstorm for now

What I might do is adding an Interface, that can be implemented into CardState and CardTraitCache.
With that, if the card has no trait changes, and doesn't have a basic land type change (I'm looking at you, blood moon)
then the methods can just return the original Collections from CardState.

(plus leftState and rightState for split cards)

tool4ever:
- there are multiple AI-based PR in progress with slightly different approaches
- ideally we can isolate the caches more cleanly somehow, then again interfaces can't have instance fields 🤔

Hanmac:
> there are multiple AI-based PR in progress with slightly different approaches

That's why i made this Issue to talk about the different approaches and what the best way would be.
  
> then again interfaces can't have instance fields

The Cache itself will be a Record class of course, just sharing an Interface with CardState

Hanmac:
Also, the cache might only apply before special rules apply

see 
https://github.com/Card-Forge/forge/blob/38ce75b27ac41b690c06eb80e51351342b466dff/forge-game/src/main/java/forge/game/card/CardState.java#L755-L771

and
https://github.com/Card-Forge/forge/blob/38ce75b27ac41b690c06eb80e51351342b466dff/forge-game/src/main/java/forge/game/card/Card.java#L7060-L7065

So for the RulesHost check to apply, the Cache might need to be before these are added.

And these:
https://github.com/Card-Forge/forge/blob/38ce75b27ac41b690c06eb80e51351342b466dff/forge-game/src/main/java/forge/game/card/CardState.java#L733-L751
might be added to LandTraitChanges (with changed Name?), but then it can't be a record anymore? 🤔 
I mean, yes, the abilities should only be added if the Type is right, but there is nothing that turns something into a Planeswalker, Battle or Saga?

tool4ever:
hmn? I can always clone a Planeswalker creature...

Hanmac:
> hmn? I can always clone a Planeswalker creature...

Yes, but that will create a new CardState that has always been a Planeswalker (creature)
[...]

PR #11625, opened 2026-08-15
CardState: cache LandMana and copy them for LKI

Hanmac:
Part of #11409

in case of it looses their intrinsic type, the abilities  are not created.

it helps for later cleanup


tool4ever:
I wish we could avoid copying for LKI because most times nothing in the game should need it, but right now `updateStateForView()` probably messes things up

tool4ever:
anything specific to test for that might break from this?

Hanmac:
Not that I know of

I changes slightly if a Saga somehow enter without Saga sub type

tool4ever (approved):

PR #11630, opened 2026-08-16
Fix variant colors

### LINKED ISSUE #11629: Joshua Rosfield incorrectly treated as colorless
**Describe the bug**
The card Joshua Rosfield (Rograkh, Son of Rogahh) is being treated as colorless by the game engine despite its color indicator of red. Detail view shows it with a colorless frame instead of a red frame, so it may be a data issue.

**To Reproduce**
Have Joshua Rosfield on the battlefield with a Ring of Valkas equipped during your upkeep. Note that after the triggered ability of the ring resolves, no +1/+1 counter is placed on the creature.

**Expected behavior**
The creature gets a +1/+1 counter as appropriate for a red creature.

tool4ever:
Closes #11629


Hanmac (approved):

Jetz72:
Hmmm. I think I wrote that that way to account for if a variant specified a mana cost with a different set of colors than the original card. By default color derives from mana cost, so if neither the original nor the variant specify a color indicator, it'd make sense for the variant's color to derive from its own cost. 

I suppose we could make the variant's color derive from its cost only if it specifies an override for the original's cost, and otherwise inherit the original's color.

tool4ever:
oh yea good point, I can try and tweak that later

PR #11631, opened 2026-08-16
CardState: move IntrinsicSpell into LandTraitChanges

Hanmac:
Part of #11409 

kinda cleanup before Cache are added

i try to make each of them "before changes" - "changes" - "after rules" more uniform
See this comment: https://github.com/Card-Forge/forge/issues/11409#issuecomment-5307524964

### DIFF
--- a/forge-game/src/main/java/forge/game/card/CardState.java
+++ b/forge-game/src/main/java/forge/game/card/CardState.java
@@ -456,49 +456,47 @@ public class CardState implements GameObject, IHasSVars, ITranslatable {
     }
 
     public final FCollectionView<SpellAbility> getSpellAbilities() {
-        FCollection<SpellAbility> newCol = new FCollection<>();
-        updateSpellAbilities(newCol);
-        newCol.addAll(abilities);
-        card.updateSpellAbilities(newCol, this);
-        return newCol;
+        FCollection<SpellAbility> result = new FCollection<>(abilities);
+        if (getStateName().equals(CardStateName.Original)) {
+            if (getCard().hasState(CardStateName.LeftSplit))
+                result.addAll(getCard().getState(CardStateName.LeftSplit).abilities);
+            if (getCard().hasState(CardStateName.RightSplit))
+                result.addAll(getCard().getState(CardStateName.RightSplit).abilities);
+        }
+        card.updateSpellAbilities(result, this);
+        return result;
     }
     public final FCollectionView<SpellAbility> getManaAbilities() {
-        FCollection<SpellAbility> newCol = new FCollection<>();
-        updateSpellAbilities(newCol);
-        newCol.addAll(abilities);
-        card.updateSpellAbilities(newCol, this);
-        newCol.removeIf(Predicate.not(SpellAbility::isManaAbility));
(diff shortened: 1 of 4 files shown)


forge-game/src/main/java/forge/game/card/CardState.java
```diff
@@ -456,49 +456,47 @@ public final boolean removeIntrinsicKeyword(final Keyword k) {
...
+        return result;
     }
     public final FCollectionView<SpellAbility> getManaAbilities() {
-        FCollection<SpellAbility> newCol = new FCollection<>();
-        updateSpellAbilities(newCol);
-        newCol.addAll(abilities);
-        card.updateSpellAbilities(newCol, this);
-        newCol.removeIf(Predicate.not(SpellAbility::isManaAbility));
-        return newCol;
+        FCollection<SpellAbility> result = new FCollection<>(abilities);
```
tool4ever: these two methods should reuse `getSpellAbilities()`

tool4ever (commented):
feels like the logic flow gets a bit harder to follow now 🤔

can't we do something a bit more radical and try to avoid all `rulesHost` stuff:
just have `getReplacementList()` add them on demand?

a bit like `getHiddenStaticAbilities()` but the big bonus is we might be able to reuse a single instance of `adventureRep/omen...` for the whole game 🏎️
(because all of them can just use `ReplacedCard` instead of `Self`)

Hanmac:
> feels like the logic flow gets a bit harder to follow now 🤔
> 
> can't we do something a bit more radical and try to avoid all `rulesHost` stuff: just have `getReplacementList()` add them on demand?
> 
> a bit like `getHiddenStaticAbilities()` but the big bonus is we might be able to reuse a single instance of `adventureRep/omen...` for the whole game 🏎️ (because all of them can just use `ReplacedCard` instead of `Self`)

Maybe for later MRs.

The Counter Based ones are still excluded by the rulesHost check:
https://github.com/Card-Forge/forge/blob/1dfbcc87c85b90b37b390a22045541be0f09a59c/forge-game/src/main/java/forge/game/card/Card.java#L7063-L7073



i also wanted to move these special ones:
https://github.com/Card-Forge/forge/blob/1dfbcc87c85b90b37b390a22045541be0f09a59c/forge-game/src/main/java/forge/game/card/Card.java#L3467-L3478

into `getAllPossibleAbilities`, there:
https://github.com/Card-Forge/forge/blob/1dfbcc87c85b90b37b390a22045541be0f09a59c/forge-game/src/main/java/forge/game/card/Card.java#L7472-L7480

so we don't need to exclude them anymore when checking for `hasNoAbilities`:

https://github.com/Card-Forge/forge/blob/1dfbcc87c85b90b37b390a22045541be0f09a59c/forge-game/src/main/java/forge/game/card/Card.java#L3443-L3449

but then the Tests does fail, because it can't find them anymore:
https://github.com/Card-Forge/forge/blob/44a5826757811633a9f20319837d05382ea99ed6/forge-gui-desktop/src/test/java/forge/ai/simulation/GameSimulationTest.java#L202-L204

PR #11633, opened 2026-08-17
Implement Beam me up and Marooned's restriction on it (#11589)

### LINKED ISSUE #11589: TRK: Beam me up
> Beam me up <Cost> (You may cast this card from your graveyard for <Cost> if you also return a creature you control to its owner's hand. Then Exile this Card)

New Keyword like Flashback variant.

See similar things about cards gaining Flashback.

liamiak:
Two TRK cards, and they only make sense together.

> **Open Communications** `{U}` Sorcery — Draw a card. Beam me up `{2}{U}` *(You may cast this card from your graveyard for `{2}{U}` if you also return a creature you control to its owner's hand. Then exile this spell.)*
>
> **Marooned** `{U}` Aura — … Enchanted permanent doesn't untap during its controller's untap step **and can't be beamed up.**

Beam me up is Flashback with one extra mandatory cost, so it reuses `getGraveyardSpellByKeyword` — already shared by Flashback, Harmonize and Mayhem — and the same "exile as it leaves the stack" replacement Flashback installs.

The return is appended to the cost in `GameActionUtil` rather than written into each script's keyword cost. Harmonize already adds its extra cost there, and keeping it out of the script means the printed cost stays `{2}{U}` instead of growing a `Return` clause.

Marooned's clause has no keyword on the card, so it travels as a granted keyword string, the way "doesn't untap during your untap step" does. The restriction then rides on the appended cost's valid string:

```
Return<1/Creature.YouCtrl+canBeBeamedUp/creature you control>
```

A permanent that can't be beamed up simply isn't a legal way to pay, and both the human and AI paths honour it for free — `CostReturn` already filters by that string.

**The second commit is separable.** The exile replacement had become a third verbatim copy of Flashback's and Harmonize's; folding them into one branch turns the engine change into net **+7 lines**. Drop that commit and the mechanic still stands.

Three things worth knowing:

- Wording is from Scryfall spoilers — TRK is dated 2026-11-13. These are the only two cards in the set touching the mechanic as currently spoiled.
- I could not test the exile *firing*. Adding the SA to the stack isn't a real cast, so no "leaves the stack" event occurs — shipped Flashback fails the same scaffolding identically. The test asserts the replacement is installed and points at the right spell property instead.
- Harmonize had no test at all, and the Flashback test that exists covers `GameActionUtil` rather than this code — so the fold brings the test that guards it. I checked both new assertions fail when broken: deleting Marooned's granted keyword, and swapping the spell property.

## Testing

Full suite 357 tests, 0 failures, checkstyle clean.


forge-gui/res/cardsfolder/upcoming/marooned.txt
```diff
@@ -0,0 +1,10 @@
+Name:Marooned
+ManaCost:U
+Types:Enchantment Aura
+K:Enchant:Artifact,Creature:artifact or creature
+SVar:AttachAILogic:Curse
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigTap | TriggerDescription$ When this Aura enters, you may pay {3}. If you do, tap enchanted permanent.
+SVar:TrigTap:AB$ Tap | Defined$ Enchanted | Cost$ 3
+R:Event$ Untap | ActiveZones$ Battlefield | ValidCard$ Permanent.EnchantedBy | ValidStepTurnToController$ You | Layer$ CantHappen | Description$ Enchanted permanent doesn't untap during its controller's untap step and can't be beamed up.
+S:Mode$ Continuous | Affected$ Permanent.EnchantedBy | AddKeyword$ CARDNAME can't be beamed up.
```
tool4ever: this would need to be a hidden Keyword

unless @Hanmac dislikes it too much?

Hanmac: i would prefer a static if able

liamiak: Gone with the static. `StaticAbilityMode.CantBeBeamedUp` plus a `StaticAbilityCantBeBeamedUp` modelled on `StaticAbilityCantRegenerate`, so Marooned's line is now

```
S:Mode$ CantBeBeamedUp | ValidCard$ Permanent.EnchantedBy | Secondary$ True
```

matching the other aura-borne `Cant` statics. Agreed it's the better shape - the affected permanent doesn't have to carry invented keyword text, and any card can grant it. The cost's valid string is unchanged, so `CardProperty` was the only caller that had to move.

@tool4ever your `AB$ Tap | Cost$ 3` is kept - I rebased onto it rather than pushing over it, and checked it parses to a Tap ability with cost {3} on `Defined$ Enchanted`.

One gap I'd rather flag than paper over: nothing here tests whether the AI *should* beam something up. It will surface the option through `getOriginalAndAltCostAbilities`, but returning a creature to draw one card is a poor trade in most spots and I have not measured how it judges that. Happy to look if you want it in this PR, otherwise it seems like a thing to see in practice first - I would rather not pre-emptively flag the card out of AI decks on a guess.

Full suite 357/0/6, checkstyle clean.

tool4ever: yea there are other more important AI tasks anyway

tool4ever (approved):

PR #11634, opened 2026-08-17
Remove extra text renders for Prepare Cards

kevlahnota:


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -3117,7 +3117,7 @@ public String getAbilityText(final CardState state) {
             sb.append("\r\n");
         }
 
-        if (hasState(CardStateName.PreparedSpell) && state.getStateName().equals(CardStateName.Original)) {
+        /*if (hasState(CardStateName.PreparedSpell) && state.getStateName().equals(CardStateName.Original)) {
```
tool4ever: we can't just remove this from detail text panel

maybe better check why it's no problem for adventure?

kevlahnota: the detail panel IIRC the code should be on the composetext method if the card has prepared spell though i'll update this pr

Hanmac: @tool4ever why have you been against this to move from `getAbilityText` to `composeCardText`?

tool4ever: @Hanmac
this is old, see #11635

PR #11635, opened 2026-08-17
Move prepare card text details to CardDetailUtil

kevlahnota:


forge-gui/src/main/java/forge/gui/card/CardDetailUtil.java
```diff
@@ -656,6 +656,15 @@ public static String composeCardText(final CardStateView state, final GameView g
             area.append("\n\n");
             area.append("Owner: ").append(card.getOwner().toString());
         }
+
+        if (card.hasPreparedSpell()) {
+            area.append("\n\n");
+            area.append("Prepared — "
```
tool4ever: spacing and lblPrepared got dropped

tool4ever (commented):
Seems it doesn't work with Clones but no big deal I guess

Hanmac (approved):

forge-gui/src/main/java/forge/gui/card/CardDetailUtil.java
```diff
@@ -656,6 +656,15 @@ public static String composeCardText(final CardStateView state, final GameView g
             area.append("\n\n");
             area.append("Owner: ").append(card.getOwner().toString());
         }
+
+        if (card.hasPreparedSpell()) {
+            area.append("\n\n");
+            area.append(Localizer.getInstance().getMessage("lblPrepared") + " — "
+                + card.getAlternateStateName() + " "
```
tool4ever: what's the point of `getAlternateStateName()` instead of `getAlternateState().getName()`?

kevlahnota: CardStateName is lightweight, CardState is heavy object IIRC

tool4ever: don't understand
every redundant TrackableProperty just increases network traffic

Hanmac: its a rename of `BackSideName`, but we might remove this Property?

kevlahnota: It can be removed, I just think serializing and deserializing during netplay a CardState object is heavier.

Hanmac: @kevlahnota AlternateState is serialized anyway, so we don't need extra AlternateStateName

Hanmac (dismissed):

Hanmac (approved):

PR #11640, opened 2026-08-19
Fix UnsupportedCharsetException: IBM437 crash on iOS during Zip extraction

voguelike:
### Summary
Fixes an unhandled `java.nio.charset.UnsupportedCharsetException: IBM437` crash on iOS (RoboVM) when downloading and unzipping Net Decks in `GuiDownloadZipService`.

### Problem
RoboVM strips legacy DOS charsets (such as `IBM437`) on iOS to reduce binary size. Calling `Charset.forName("IBM437")` during ZIP extraction throws an unhandled `UnsupportedCharsetException`, causing Net Deck downloads to fail instantly:

```
java.nio.charset.UnsupportedCharsetException: IBM437
	at java.nio.charset.Charset.forName(Charset.java:313)
	at forge.gui.download.GuiDownloadZipService.extract(GuiDownloadZipService.java:174)
	at forge.gui.download.GuiDownloadZipService.downloadAndUnzip(GuiDownloadZipService.java:76)
```

### Fix
- Wrap `Charset.forName("IBM437")` in a `try-catch` block targeting `UnsupportedCharsetException`.
- Fall back gracefully to `StandardCharsets.UTF_8` on environments lacking legacy charset tables (e.g., iOS).
- Retain existing desktop & legacy Android behavior without regressions.

> [!NOTE]
> *Co-authored with assistance from Antigravity AI agent.*

### DIFF
--- a/forge-gui/src/main/java/forge/gui/download/GuiDownloadZipService.java
+++ b/forge-gui/src/main/java/forge/gui/download/GuiDownloadZipService.java
@@ -171,13 +171,21 @@ public class GuiDownloadZipService extends GuiDownloadService {
                 }
             }
 
-            final Charset charset = Charset.forName("IBM437");
+            Charset charset;
+            try {
+                charset = Charset.forName("IBM437");
+            } catch (java.nio.charset.UnsupportedCharsetException e) {
+                // Fallback for environments like iOS/RoboVM that lack legacy charsets
+                charset = java.nio.charset.StandardCharsets.UTF_8;
+            }
+
             ZipFile zipFile;
             try {
                 zipFile = new ZipFile(zipFilename, charset);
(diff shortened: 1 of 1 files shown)


Hanmac:
@kevlahnota can you check if we need `IBM437` anyway?

tool4ever (approved):
this makes sense, ensuring all our zip files are updated to UTF-8 instead would certainly not be a great alternative :D

kevlahnota:
> @kevlahnota can you check if we need `IBM437` anyway?

IIRC it was needed since some decks has IBM437 encoding on their file name.

kevlahnota (approved):

Hanmac (approved):
Approve for now, can remove IBM437 Decks later?

PR #11654, opened 2026-08-22
move isTurnFaceUp to getAllPossibleAbilities

Hanmac:
Test case doesn't work yet, and i don't know how/why

@liamiak or @MostCromulent maybe the AIs know why

### DIFF
--- a/forge-ai/src/main/java/forge/ai/simulation/GameCopier.java
+++ b/forge-ai/src/main/java/forge/ai/simulation/GameCopier.java
@@ -494,21 +494,21 @@ public class GameCopier {
         }
     }
 
-    public GameObject find(GameObject o) {
+    public <T extends GameObject> T find(T o) {
         if (origGame.EXPERIMENTAL_RESTORE_SNAPSHOT) {
-            return snapshot.find(o);
+            return (T) snapshot.find(o);
         }
 
-        GameObject result = null;
+        T result = null;
         if (o instanceof Card) {
-            result = cardMap.get(o);
+            result = (T) cardMap.get(o);
             if (result != null) {
                 return result;
             } else {
                 System.out.println("Couldn't map " + o + "/" + System.identityHashCode(o));
             }
         } else if (o instanceof Player) {
-            result = playerMap.get(o);
+            result = (T) playerMap.get(o);
             if (result != null)
                 return result;
         }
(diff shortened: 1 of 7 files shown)


liamiak:
The `assertNotNull` on line 203 passes — the SA is found, so the `Card.java` move itself is fine. What fails is line 207 (`expected:<18> but was:<20>`), and stderr from the same run has the reason:

```
Simulation: SA not found! Morph — Reveal a black card from your hand (Turn this face up any time for its morph cost.) / class forge.game.ability.StaticAbilityApiBased
```

`GameSimulator.findSaInSimGame` maps a real-game SA onto the copied sim card by matching descriptions against the *cached* list:

```java
FCollectionView<SpellAbility> candidates = hostCard.getSpellAbilities();
SpellAbility result = saMatcher(candidates, desc);
```

`updateSpellAbilities` used to inject the turn-face-up SAs into the FaceDown state's cached list, so that lookup hit. `getAllPossibleAbilities` is computed and never reaches the cache, so now it misses, `simulateSpellAbility` returns `Score(MIN_VALUE)`, and the sim game is untouched.

Adding a fallback makes it green (`Tests run: 1, Failures: 0`):

```java
SpellAbility result = saMatcher(candidates, desc);
if (result == null) {
    result = saMatcher(hostCard.getAllPossibleAbilities(aiPlayer, false), desc);
}
```

Worth noting this isn't only a test problem. `SpellAbilityPicker.getCandidateSpellsAndAbilities` → `ComputerUtilAbility.getSpellAbilities` → `getAllPossibleAbilities` on real-game cards, and those SAs go straight to `simulator.simulateSpellAbility`. So as it stands the simulation AI offers itself the morph-up ability and then always scores it `MIN_VALUE` — it can never choose to unmorph. The test is catching that, not being awkward.

I checked the other consumers and they're unaffected: `InputPassPriority`, `PlayerControllerHuman`, `ActivateAbilityEffect` and the non-simulation AI paths all go through `getAllPossibleAbilities` already. `findSaInSimGame` is the only reverse lookup reading the raw cached list, so it's the only thing the move breaks.

🤖 Diagnosed with the assistance of Claude Code (Opus 5).

Hanmac:
@tool4ever can you look at this too?

tool4ever:
root cause is ugly and somehow related to how the SA descriptions are rebuilt 🤔

tool4ever (dismissed):
should be mergeable for now

Hanmac:
@tool4ever what is this about removing mayPlayPlayers?

Hanmac:
@tool4ever the main reason for this MR, in `hasNoAbilities`:
https://github.com/Card-Forge/forge/blob/c94a8f4c9cee5b153c49bfe1c120e180d89d4aaa/forge-game/src/main/java/forge/game/card/Card.java#L3442-L3445
Can this be removed?

tool4ever:
> @tool4ever what is this about removing mayPlayPlayers?

just dead code confusing me while debugging

> Can this be removed?

done

tool4ever (approved):

PR #11657, opened 2026-08-22
Implement Aggressive and Equipment swap keywords

BigCrunch22:
- **Bear with Set's Mechanic** (Aggressive)
- **Arcanum Things** (Equipment swap)

_This is my first time implementing engine changes, so feedback is appreciated. I would have continued with the implementation of other cards, but I think implementing engine-related changes one card at a time is better, so these two get sent out first._

**_Made with assistance from Claude_**


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -2668,7 +2668,7 @@ private String keywordsToText(final Collection<KeywordInterface> keywords) {
                     }
                     sbLong.append(type).append(" offering");
                     sbLong.append(" (").append(inst.getReminderText()).append(")");
-                } else if (keyword.startsWith("Equip") || keyword.startsWith("Fortify")
+                } else if (keyword.startsWith("Equipment swap") ||keyword.startsWith("Equip") || keyword.startsWith("Fortify")
```
squee1968: nitpicking, but there should be a space between || and keyword.startsWith("Equip")

PR #11663, opened 2026-08-23
cards: use SubCounter if able

### LINKED ISSUE #11023: Some RemoveAnyCounter to SubCounter
Some with fixed Card like `RemoveAnyCounter<1/Any/CARDNAME`
might be able to be replaced by `SubCounter<1/Any/CARDNAME` ?

RemoveAnyCounter might be only for cases where Card is not defined?

Hanmac:
Closes #11023

the GUI is slightly better for this one


Hanmac:
@tool4ever @Agetian the GUI is slightly better, but i'm still not happy with this.

For example, with `Reaping Willow` if it only has one counter type,
it still does ask me to choose how much I want to remove, even if the amount is fixed.

PR #11672, opened 2026-08-24
Improve AI handling of storage lands

vanja-ivancevic:
- We are preparing Fallen Empires Adventure dungeons whose AI decks need the original storage-land cycle, which is currently excluded because the AI always untaps these lands before they can accumulate mana.
- Add a small optional-untap policy that counts other mana-producing lands and storage counters, then untaps only when that total bridges the mana-value gap to a card in hand.
- Keep the untap decision bounded to one battlefield count and one hand scan. It does not invoke spell selection, playability evaluation, mana payment, or ritual simulation.
- Leave counter spending to Forge's existing `ManaRitualBattery` handling once the land is untapped.
- Enable Bottomless Vault, Dwarven Hold, Hollow Trees, Icatian Store, and Sand Silos for AI decks.
- This is deliberately a mana-value heuristic. Color requirements, alternative costs, and coordination between multiple storage lands remain outside its scope.


Agetian (dismissed):

tool4ever (changes requested):
I think when I checked this for anything obviously crazy earlier it looked almost reasonable, now after all these force-pushes it seems to have turned into yet another giant AI mess with lots of expensive recursion 😩

Here's just one example of spaghetti code no one will ever be able to maintain:


this needs a clean architecture and less bloat

vanja-ivancevic:
Thank you very much for your comment. That's really valuable. I'm very inexperienced with this stuff, and appreciate the learning opportunity!

tool4ever (approved):
thanks

vanja-ivancevic:
I've reworked this from the architectural boundary outward. The untap decision now performs one count of other mana-producing lands and one mana-value scan of the hand, with no calls into spell selection, playability evaluation, mana payment, or ritual simulation. `ComputerUtilMana` and `ManaAi` are unchanged from `master`; the final implementation is 10 production lines plus the card-script updates and two focused tests.

PR #11675, opened 2026-08-24
Fix AI cloning creatures that shrink under its own control

liamiak:
`CloneAi` ranks candidates with `evaluateCreature`, which reads live power and toughness. For a characteristic-defining creature those are computed under whoever controls it now, so an opponent's Nightmare scores as the 8/8 it is for them rather than the 0/0 it would be for us.

With Cryptoplasm out and the opponent holding a Nightmare and eight Swamps, the AI copies it, becomes a 0/0 and dies on the spot. Six cards reach this with nothing keeping them out of AI decks: **Artisan of Forms, Cephalid Facetaker, Cryptoplasm, Crystalline Resonance, Hulkling and Protean Thaumaturge**. The one flagged card on the path, Mindlink Mech, only ever sees creatures that crewed it.

## The check

#11690 already answers this for token copies: take an LKI copy, put it under our control, recheck statics on that card alone, and see whether a toughness-setting CDA leaves it alive. That is now `ComputerUtilCard.filterOutFatalCopies`, called from both `CopyPermanentAi` and `CloneAi`.

An earlier version of this PR claimed the value could not be worked out ahead of the fact and priced candidates with a one-play simulation instead. That was wrong on both halves — the LKI route works, and the simulation was gated behind `usesHybridSimulation`/`usesFullSimulation`, so on the default profile it skipped every CDA candidate rather than pricing one.

| Board | master | with this |
|---|---|---|
| Their Nightmare, only target | copies it, Cryptoplasm dies | keeps ours |
| Their Nightmare + a Serra Angel | copies Nightmare, dies | Serra Angel |
| Their Nightmare, we hold Swamps | Nightmare 8/8 | unchanged |

Known gap: cards choosing through `Choices$` rather than targeting — Vesuvan Doppelganger — resolve via `CloneEffect` and `chooseSingleEntityForEffect`, a path this does not touch.

Three tests, all driving `AiController` on the default profile; two fail without the fix. Full desktop suite: 360 tests, 0 failures.


tool4ever:
The idea is nice, but over-engineered in typical AI fashion :P

See my related implementation here: #11690
When that's merged feel free to turn it into a generic helper for such clone scenarios too

liamiak:
Done — thanks, your version is better than mine on two counts, not just smaller.

Lifted the check out of `CopyPermanentAi` into `ComputerUtilCard.filterOutFatalCopies`, and both call it now. `CopyPermanentAi` keeps its own `SetToughness` guard, since that one is about the copy effect overriding toughness rather than the candidate. Source went from 101 lines to 30, and the `GameSimulator` trigger-matching hunk went with the simulation it existed for.

The old version also had a hole I had not spotted: it was gated on `usesHybridSimulation() || usesFullSimulation()`, and on the default profile that is false, so it skipped every CDA candidate rather than pricing one. Measured on a board where their Nightmare would be a live 8/8 for us — old code left Cryptoplasm a 2/2, yours copies it correctly.

What is left on top of master is one behaviour: don't clone into a body that dies under our control. Their 8/8 Nightmare while we hold Islands is a 0/0 for us, and `getBestCreatureAI` ranks it top because it evaluates under *their* board.

| Board | master | with this |
|---|---|---|
| Their Nightmare, only target | copies it, Cryptoplasm dies | keeps ours |
| Their Nightmare + a Serra Angel | copies Nightmare, dies | Serra Angel |
| Their Nightmare, we hold Swamps | Nightmare 8/8 | unchanged |

Three tests, all through `AiController` on the default profile. 360 in the module, 0 failures.

tool4ever (approved):

PR #11676, opened 2026-08-24
Improve High Tide AI usage

vanja-ivancevic:
- We are preparing Fallen Empires Adventure dungeons whose authored AI decks use High Tide, which is currently excluded by `AI:RemoveDeck:All`.
- Add dedicated `HighTide` handling for its effect-based spell during the AI controller main phases, delegating payoff selection to the existing mana-ritual heuristic.
- Keep the card script schema-truthful. The AI derives the estimated net gain directly as the number of untapped Island lands minus the Island used to cast High Tide. This includes nonbasic lands with the Island subtype.
- Replace `AI:RemoveDeck:All` with `AI:RemoveDeck:Random`, allowing authored AI decks to use High Tide while keeping it out of generated decks that may lack sufficient Islands or worthwhile fixed-cost payoffs. This is especially important because High Tide also benefits opposing Islands.
- The existing ritual heuristic holds High Tide without a useful fixed-cost payoff and casts it when the additional mana unlocks one. X-cost spell planning remains outside this change.
- Verified with normal AI scenarios: with three Islands, Tundra, and Time Spiral, High Tide is selected first; without a payoff it is held; during an opponent main phase it is held. Existing Dark Ritual selection and mana-resolution scenarios also pass.

### DIFF
--- a/forge-gui/res/cardsfolder/h/high_tide.txt
+++ b/forge-gui/res/cardsfolder/h/high_tide.txt
@@ -1,8 +1,8 @@
 Name:High Tide
 ManaCost:U
 Types:Instant
-A:SP$ Effect | Triggers$ IslandTrigger | SpellDescription$ Until end of turn, whenever a player taps an Island for mana, that player adds an additional {U}.
+A:SP$ Effect | Triggers$ IslandTrigger | AILogic$ HighTide | AINoRecursiveCheck$ True | SpellDescription$ Until end of turn, whenever a player taps an Island for mana, that player adds an additional {U}.
 SVar:IslandTrigger:Mode$ TapsForMana | ValidCard$ Island | Execute$ TrigMana | Static$ True | TriggerDescription$ Whenever a player taps an Island for mana, that player adds an additional {U}.
 SVar:TrigMana:DB$ Mana | Produced$ U | Amount$ 1 | Defined$ TriggeredActivator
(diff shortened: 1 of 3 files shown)


forge-gui/res/cardsfolder/h/high_tide.txt
```diff
@@ -1,8 +1,9 @@
 Name:High Tide
 ManaCost:U
 Types:Instant
-A:SP$ Effect | Triggers$ IslandTrigger | SpellDescription$ Until end of turn, whenever a player taps an Island for mana, that player adds an additional {U}.
+A:SP$ Effect | Triggers$ IslandTrigger | Produced$ U | Amount$ X | AILogic$ ManaRitual | AINoRecursiveCheck$ True | SpellDescription$ Until end of turn, whenever a player taps an Island for mana, that player adds an additional {U}.
```
tool4ever: no fake params on scripts

vanja-ivancevic: Fixed, thanks. I removed the fake `Produced` and `Amount` parameters and the supporting `X` SVar. High Tide now uses a dedicated `AILogic`, and the ritual evaluator derives its untapped-Island net gain internally. The High Tide scenarios and existing Dark Ritual selection and mana-resolution checks pass.

PR #8464, opened 2025-08-20
Support named abilities

Eradev:
# Add support to name abilities

This is mostly for custom cards. It'll allow scripters to group them and target them as they don't have access to stuff like `kicked` or `madness`.

## Use cases

This will support things like being able to target specific SA using a custom name. (Flash on Meditate abilities in this example)

```text
Name:Plo Koon
ManaCost:3 W W
Types:Legendary Creature KelDor Jedi
PT:4/4

S:Mode$ CastWithFlash | ValidSA$ Activated.NamedAbilityMeditate | Caster$ You | Description$ You may activate meditate abilities any time you could cast an instant.

A:AB$ ChangeZone | Named$ Meditate | Cost$ 1 W | ActivationZone$ Battlefield | SorcerySpeed$ True | Origin$ Battlefield | Destination$ Hand | Defined$ Self | SpellDescription$ Meditate (Return this creature to its owner's hand. Meditate only as a sorcery.)

Oracle:You may activate meditate abilities any time you could cast an instant.\nMeditate 1W (Return this creature to its owner's hand. Meditate only as a sorcery.)
```

Restrict trigger to only if was triggered by a specific type of SA (Only Scry when Meditating and not being bounced), and reduce cost for a specific type of ability.

```text
Name:Jedi Training
ManaCost:U
Types:Enchantment

S:Mode$ ReduceCost | ValidCard$ Card | ValidSpell$ Activated.NamedAbilityMeditate | Activator$ You | Amount$ 1 | Description$ Meditate abilities you activate cost {1} less to activate.

T:Mode$ ChangesZone | Origin$ Battlefield | Destination$ Hand | TriggerZones$ Battlefield | ValidCard$ Creature.Jedi+YouCtrl | Condition$ FromNamedAbilityMeditate | SubAbility$ DBScry | TriggerDescription$ Whenever a Jedi creature you control meditates, scry 1.
SVar:DBScry:DB$ Scry | ScryNum$ 1

Oracle:Meditate abilities you activate cost {1} less to activate.\nWhenever a Jedi creature you control meditates, scry 1.
```

It will also allow for cards to check if a card was cast using an certain ability using `Count$FromNamedAbility<name>.<true>.<false>`:

```text
Name:Chronic Traitor
ManaCost:2 B
Types:Creature Human Rogue
PT:2/1

[...]

PR #11700, opened 2026-08-26
StaticAbilityManaRestriction: better ability for Myr Superion

### LINKED ISSUE #11689: `Myr Superion` and similar cause problems with noAbility
@tool4ever `Myr Superion` isn't related to this problem

`A:SP$ PermanentCreature | Cost$ 2 | ManaRestriction$ Creature.inZoneBattlefield`
this also would have been a BasicSpell.

And its Text
`Text:Spend only mana produced by creatures to cast this spell.`
should have been a Static Method in the first place.

_Originally posted by @Hanmac in https://github.com/Card-Forge/forge/pull/11682#discussion_r3852219992_

IMO to clean this up, most of these should use a StaticAbility instead of changing `SP$ Permanent`

Hanmac:
Closes #11689


tool4ever (commented):
no sorry, I think that's a terrible idea:
a) just introduces a secondary way since you can't even remove the ManaRestriction$
b) more importantly we're spending hours thinking about CardTraits caching to speed up AI turns - this just introduces another expensive check in a place like `allowsPayingWithShard` which will get called multiple times per mana source and per SA every priority window...all that for a tiny refactor of 3 cards?! 👎

Hanmac:
@tool4ever 
You caused the problem with "Text" on NoAbilities

tool4ever:
Yea whatever, I can assure you the old logic was way worse before my refactor here: eb6d168
I worked with that was available at that time and I certainly didn't introduce these hardscripted spell versions: a80396a
🤷‍♂️

Hanmac:
@tool4ever this should reduce the amount of StaticAbility calls

> a) just introduces a secondary way since you can't even remove the ManaRestriction$

This is because these 3 are done as Static Ability, while the other are part of Activated Ability or Alternative Cost.
Different things so they need to be implemented differently.

tool4ever:
Hmn, I need to think if we can find a way to support *Throne of Eldraine* too?
Like mapping the affected to a specific ability, might be other cases like that? 🤔
maybe reuse logic of #8464?

Hanmac:
> _Throne of Eldraine_
> Draw two cards. Spend only mana of the chosen color to activate this ability.

This is part as ONE Ability, so I don't think we should split it.
This matters if something else gets this activated ability.

forge-game/src/main/java/forge/game/staticability/StaticAbilityManaRestriction.java
```diff
@@ -0,0 +1,33 @@
+package forge.game.staticability;
+
+import forge.game.card.Card;
+import forge.game.spellability.SpellAbility;
+
+public class StaticAbilityManaRestriction {
+    public static boolean manaRestriction(final SpellAbility sa, final Card source)  {
+        for (final StaticAbility stAb : sa.getHostCard().getStaticAbilities()) {
```
tool4ever, line 12: either needs a comment explaining the reasoning

or if you prefer we ignore custom cards do it like #11328 wanted:
new `intrinsicOnly` field in `StaticAbilityMode` since there are probably a few others, e.g. `CantBeCopied`?

Hanmac: if its own Card anyway, should i remove the check for `ValidSA$ Spell.Self` too?
(with isSpell check in SpellAbility?)

tool4ever: that would make it a bit clearer when reading the script that it doesn't work on other cards now anyway

tool4ever:
I still want to check some alternatives, maybe storing it inside `ManaConversionMatrix` would be better since that gets filled for each SA anyway...

Hanmac:
Maybe not like ManaConversionMatrix, but caching would be nice to have 🤔

tool4ever:
I'm more neutral to this now, will try to do some profiling tonight

tool4ever (approved):
while my results are inconclusive, I guess it gets rid of the Text fields at least :/

PR #11709, opened 2026-08-27
CDN Multi-language card image downloads (continues #10928)

churrufli:
## Summary

Continues the work from #10928 (CDN UUID cache / card image downloader). Adds language support for card image downloads: preferred-language selection, an optional multi-language index download, and text/i18n cleanup across the related UI.

### Language preferences
- New options in Preferences (desktop and mobile): preferred language for downloaded card images, and a checkbox to prioritize that language when choosing which printing to show as the "unique card."
- Unique-art selection respects the chosen language, with automatic fallback to English when a card doesn't exist in that language.
- Both single-card and bulk image downloads use this same preferred language to resolve each card's URL.

### Multi-language index download
- New "Download Index" button next to a language dropdown, which downloads Scryfall's `all_cards` bulk file filtered to English + the chosen language (avoids caching the other ~16 languages that won't be used).
- The existing English-only button (`default_cards`, ~75 MB) is unchanged in behavior.
- The automatic per-set sync (triggered when downloading images if any sets aren't cached yet) also respects the preferred language, querying `lang:en OR lang:<language>` instead of English only.
- `image_status` filtering (Scryfall): cards marked `missing` or `placeholder` are discarded instead of being cached as if they were real art (this mostly affects localized languages, where this status is more common).

### Bug fixes
- Set codes that collide with Windows reserved device names (e.g. Conflux's `CON`) no longer break cache writes on that OS.
- Mobile download screen wrapped in an `FScrollPane`: content was getting cut off at certain resolutions/orientations, leaving some controls unreachable.

### Text and i18n
- Buttons and messages show the approximate real size of each download.
- All new keys translated across all 10 Forge locales.
- Fixed two mobile Preferences options that had hardcoded English text bypassing the localization system.

## No behavior change by default
Anyone who doesn't touch the new language options keeps exactly the same behavior as in #10928: English-only sync, same fast bulk-download button.

Assisted by Claude Opus 5


tool4ever:
thanks, I suppose this iteration is the best of "both" worlds
Hopefully another dev will have time to test drive/code review, I have way too many PR waiting on me (+ a bit burned out on such massive AI PR)

forge-gui-mobile/src/forge/Forge.java
```diff
@@ -360,6 +363,25 @@ public static void openHomeScreen(int index, FScreen lastMatch) {
             System.out.println(fScreen.toString());*/
     }
 
+    /**
+     * First run (or cache cleared): offer the one-time bulk CDN sync so later card image
+     * downloads resolve instantly instead of one Scryfall set at a time. Call only once a real
+     * home/adventure screen is open -- not while the mode-selector splash is still showing.
+     */
+    public static void maybePromptForBulkCdnSync() {
+        if (CdnUuidCache.hasAnyCachedSets()) {
```
kevlahnota: If you select not now or download now but the user didn't download the cached sets, on next startup it will ask again if the user has not any cached sets saved though maybe create a temp file after the user press any buttons so on next startup use that temp file if it exists then just dont show the message again. so far it works fine on my test.

kevlahnota (changes requested):

kevlahnota (approved):

PR #11712, opened 2026-08-27
Improve AI handling of reveal-count effects

vanja-ivancevic:
- Evaluate any-number self-reveal effects from a detached LKI host containing the cards the AI can reveal, without modifying the live card or remembered state.
- Reuse Forge's normal `AbilityUtils` and SVar evaluation for the full subability chain, including Amount, Twice, and Chroma expressions.
- Add a narrow per-API evaluation-host hook at the controller's existing alternate-host boundary, with host and cast-state restoration in `try`/`finally`.
- Decline the effect when no eligible cards can be revealed, allow the affected cards in curated AI decks while retaining their random-deck exclusion, and identify Scent of Nightshade as a curse for targeting.
- This prepares the reveal-count cards needed by a Fallen Empires Adventure dungeon while improving the same mechanics generally.
- Verified with focused game-state scenarios for Scent of Cinder, Phosphorescent Feast's Chroma expression, an empty reveal set, and Cinder Seer's activated ability; the live host was unchanged after every evaluation.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/RevealAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/RevealAi.java
@@ -7,7 +7,6 @@ import forge.ai.PlayerControllerAi;
 import forge.game.ability.AbilityUtils;
 import forge.game.card.Card;
 import forge.game.card.CardCollection;
-import forge.game.card.CardCopyService;
 import forge.game.card.CardLists;
 import forge.game.cost.Cost;
 import forge.game.player.Player;
@@ -19,8 +18,12 @@ public class RevealAi extends RevealAiBase {
 
     @Override
     protected AiAbilityDecision checkApiLogic(final Player ai, final SpellAbility sa) {
-        if (isRememberedSelfRevealAnyNumber(sa) && getRevealableCards(ai, sa).isEmpty()) {
-            return new AiAbilityDecision(0, AiPlayDecision.CantPlayAi);
+        if (isRememberedSelfRevealAnyNumber(sa)) {
+            CardCollection revealable = getRevealableCards(ai, sa);
+            if (revealable.isEmpty()) {
+                return new AiAbilityDecision(0, AiPlayDecision.MissingNeededCards);
+            }
+            setAiEvaluationHost(sa, revealable);
         }
 
         if (!revealHandTargetAI(ai, sa, false)) {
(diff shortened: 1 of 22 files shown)


forge-ai/src/main/java/forge/ai/ability/RevealAi.java
```diff
@@ -26,6 +34,39 @@ protected AiAbilityDecision checkApiLogic(final Player ai, final SpellAbility sa
...
+        final SpellAbility root = ab.getRootAbility();
+        if (ab != root.getSubAbility() || !isRememberedSelfRevealAnyNumber(root)) {
+            return super.chkDrawbackWithSubs(ai, ab);
+        }
+
+        final Card host = root.getHostCard();
+        // The AI reveals every valid card, so project that count while evaluating the dependent effect.
+        final CardCollection projected = getRevealableCards(ai, root);
+        projected.removeIf(host::isRemembered);
+        host.addRemembered(projected);
```
tool4ever: AI isn't allowed to modify engine state like this

vanja-ivancevic: Fixed in 7ffc0dde890. The AI now projects remembered-count SVars on a detached SpellAbility copy and transfers only the selected target back, so it no longer changes the live card’s remembered list or view state.

tool4ever: Still seems a bit hacky...
Furthermore I'm not sure more complicated cases `SVar:X:Remembered$ChromaSource.Green/Twice` would get handled by this?

I recommend exploring some more generic approach that focuses on your original remembering but works with LKI hosts.
This is already done partly for engine needs here: https://github.com/Card-Forge/forge/blob/0eb3a69b025a3180dadf436fe4e7eda1171a02fa/forge-ai/src/main/java/forge/ai/AiController.java#L830

If you find a clever way to extend it you'd get the reset for free there, probably per API based? 🤔

vanja-ivancevic: Thanks, that was the right direction. I removed the SVar scanning and first-subability copy. `RevealAi` now provides a detached LKI host with the projected remembered cards through a small API hook, and `AiController` evaluates and restores it at the existing alternate-host boundary under `try`/`finally`. Normal `AbilityUtils` processing now handles the full chain, including `Remembered$ChromaSource.Green/Twice`, while the live card remains unchanged. I verified the ordinary count, Chroma, zero-eligible-card, and activated Seer cases.

vanja-ivancevic:
Post-revision self-review: I traced this hook through `canPlayAndPayFor` and checked every current any-number self-reveal script it can match. It creates one detached LKI host per evaluation, does not interpret SVars or mutate live card/game state, and leaves the normal API path to evaluate ordinary counts, operators, and Chroma. Existing matches such as Rosheen, Roaring Prophet and Sacellum Godspeaker use the same reveal-to-remembered semantics. The twelve scripts changed here remain `AI:RemoveDeck:Random`, so they are available to curated AI decks such as Adventure NPC decks but remain excluded from random deck generation. The four focused scenarios and a clean `forge-ai` reactor build pass.

tool4ever (approved):

PR #11716, opened 2026-08-27
Implement assimilate and Borg Queen, Perfection Manifest

### LINKED ISSUE #11240: StarTrek: Assimilate
> When Borg Queen enters, assimilate target creature card from an opponent’s graveyard. (Put it onto the battlefield under your control with a +1/+1 counter. It’s a Borg artifact creature and loses all other creature types.)

Depending on the Zone it targets, it should be either:
* ChangeZone + Control Gain + Counter + Animate
* Control Gain + Counter + Animate

might need cleaner rules or more examples

liamiak:
Closes #11240.

**Two commits, and the second one is droppable.**

**1 — the card.** Uses the existing `StaticEffect$` on ChangeZone, which `GameAction.setupStaticEffect`
applies before the card enters, so the types are already in place for ETB triggers. Card script and
tests only, no engine change - the same shape as *Ashiok, Nightmare Weaver* and *Bronzehide Lion*.

```
| GainControl$ True | WithCountersType$ P1P1 | StaticEffect$ Animate
SVar:Animate:Mode$ Continuous | Affected$ Card.IsRemembered | AddType$ Artifact & Creature & Borg | RemoveCreatureTypes$ True
```

**2 — an `Assimilate` API**, per @Hanmac: the keyword's rules in one place so the two forms cannot
drift. A card in a graveyard or exile is moved and the type change rides that zone change; a
permanent already on the battlefield only changes hands and is animated where it stands. The card
script becomes `DB$ Assimilate | TgtZone$ Graveyard | ValidTgts$ Creature.OppOwn`.

That half is speculation, and I would rather label it than bury it. Scryfall has exactly one card
with "assimilate" in its oracle text and it only ever reaches a graveyard, so the battlefield branch
has no card behind it - it is covered by a test that drives the API directly. **The commits are
separable: drop the second and the card still works through `StaticEffect$`.** I ran the first
commit on its own to check that rather than assume it.

Entering already typed is measured, not assumed. With Reckless Fireweaver out ("whenever an artifact
you control enters"), assimilating deals **2** damage - one for the Queen, one for what she
assimilated. Applied after the move it deals 1.

`Borg` is already in `TypeLists.txt` and the edition file already lists 197. The type change adds
Artifact and Creature but does not remove card types, so an assimilated Nyx-Fleece Ram stays an
enchantment.

Eight tests. The seven from the first commit are untouched by the second and pass against both
implementations, including the one that checks a creature on their battlefield stays out of reach.

Full desktop suite: 368 tests, 0 failures, 6 skipped. Checkstyle clean.

### DIFF
--- a/forge-gui/res/cardsfolder/upcoming/borg_queen_perfection_manifest.txt
+++ b/forge-gui/res/cardsfolder/upcoming/borg_queen_perfection_manifest.txt
@@ -1,0 +1,10 @@
+Name:Borg Queen, Perfection Manifest
+ManaCost:4 B B
+Types:Legendary Artifact Creature Borg Noble
+PT:1/4
+S:Mode$ Continuous | Affected$ Creature.Artifact+YouCtrl | AddPower$ 2 | Description$ Artifact creatures you control get +2/+0.
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigAssimilate | TriggerDescription$ When NICKNAME enters, assimilate target creature card from an opponent's graveyard. (Put it onto the battlefield under your control with a +1/+1 counter. It's a Borg artifact creature and loses all other creature types.)
+SVar:TrigAssimilate:DB$ ChangeZone | Origin$ Graveyard | Destination$ Battlefield | ValidTgts$ Creature.OppOwn | TgtPrompt$ Select target creature card in an opponent's graveyard | GainControl$ True | WithCountersType$ P1P1 | StaticEffect$ Animate
+SVar:Animate:Mode$ Continuous | Affected$ Card.IsRemembered | AddType$ Artifact & Creature & Borg | RemoveCreatureTypes$ True
+DeckHas:Ability$Graveyard|Counters
+Oracle:Artifact creatures you control get +2/+0.\nWhen Borg Queen enters, assimilate target creature card from an opponent's graveyard. (Put it onto the battlefield under your control with a +1/+1 counter. It's a Borg artifact creature and loses all other creature types.)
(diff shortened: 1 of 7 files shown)


Hanmac:
@liamiak i was waiting for this one because my gut is telling me that the Effect changes depending on if the Target is in the Graveyard or in on the Battlefield

So it should probably capsuled in a Assimilate Effect


@tool4ever we had ÀnimateSubAbility` before but it got replaced with some other effect?
But i can't remember how exactly

forge-gui/res/cardsfolder/upcoming/borg_queen_perfection_manifest.txt
```diff
@@ -0,0 +1,10 @@
+Name:Borg Queen, Perfection Manifest
+ManaCost:4 B B
+Types:Legendary Artifact Creature Borg Noble
+PT:1/4
+S:Mode$ Continuous | Affected$ Creature.Artifact+YouCtrl | AddPower$ 2 | Description$ Artifact creatures you control get +2/+0.
+T:Mode$ ChangesZone | Origin$ Any | Destination$ Battlefield | ValidCard$ Card.Self | Execute$ TrigAssimilate | TriggerDescription$ When NICKNAME enters, assimilate target creature card from an opponent's graveyard. (Put it onto the battlefield under your control with a +1/+1 counter. It's a Borg artifact creature and loses all other creature types.)
+SVar:TrigAssimilate:DB$ ChangeZone | Origin$ Graveyard | Destination$ Battlefield | ValidTgts$ Creature.OppOwn | TgtPrompt$ Select target creature card in an opponent's graveyard | GainControl$ True | WithCountersType$ P1P1 | AnimateSubAbility$ DBAssimilate
+SVar:DBAssimilate:DB$ Animate | Defined$ Remembered | Types$ Artifact,Creature,Borg | RemoveCreatureTypes$ True | Duration$ Permanent
```
tool4ever: it's definitely weird that AI reimplemented the obsolete variant instead of StaticEffect$

liamiak: Right, thanks — `StaticEffect$` is the one, and the hook is gone. *Ashiok, Nightmare Weaver* and *Bronzehide Lion* were the examples I should have found; `setupStaticEffect` already runs before the card enters, which is the whole thing I was trying to build. The Fireweaver check still shows 2 damage rather than 1, so nothing was lost by dropping it.

liamiak:
@Hanmac — you were both right about `AnimateSubAbility`. That is gone; the first commit uses `StaticEffect$`.

On encapsulating it: the second commit does, as an `Assimilate` API covering both forms — graveyard or exile moves the card, a permanent on the battlefield changes hands and is animated where it stands.

Recording for the record that right now, Scryfall has exactly one card with the word in its oracle text and it only reaches a graveyard, so the battlefield branch has no card behind it and is only covered by a test driving the API directly. The two commits are separable — drop the second and the card still works. If you know of other assimilate cards coming, that would decide if we need this.

PR #11719, opened 2026-08-27
Fix adventure quest dialogs repeating after load

Nerdslayer:
## Description

Quest-level prologue and epilogue dialogs were shown again after loading an Adventure save because their `prologueDisplayed` and `epilogueDisplayed` flags were marked `transient`. Java serialization therefore reset both flags to `false` on every load.

Persist these flags so dialogs that have already been displayed remain dismissed across save/load cycles.

### Why the explicit `serialVersionUID`?

`AdventureQuestData` previously relied on Java's automatically generated serialization UID. Changing these fields from transient to persistent changes that generated UID, which would make existing Adventure saves fail to deserialize with an `InvalidClassException`.

The explicit value `-859190699184917225L` is the UID calculated from the class before this change. Preserving that exact value keeps existing save files compatible; it is not a newly chosen or arbitrary identifier.

## Testing

- Ran `mvn -pl forge-gui-mobile -am test`
  - 3 tests passed, 0 failures
- Verified a serialization round trip preserves the displayed-dialog state
- Confirmed the fix in-game using an existing save

AI assisted with development.


Jetz72:
> The explicit value `-859190699184917225L` is the UID calculated from the class before this change. Preserving that exact value keeps existing save files compatible; it is not a newly chosen or arbitrary identifier.

I believe when a similar issue came up previously, someone mentioned that the auto-calculated serial UIDs can vary between desktop and Android. What's going to happen if it doesn't match?

Nerdslayer:
@Jetz72 Good catch. If it didn't match then old saves would fail to load. I'm not positive the Android UID will be the same so I'll change the implementation to not rely on that.

Nerdslayer:
Ok, I believe this is a better solution. I revised the implementation so AdventureQuestData no longer pins the desktop-calculated UID or adds serialized fields. The private transient fields are excluded from both the serialized layout and automatic UID calculation, so removing them preserves each platform’s previous UID. Displayed state is now stored in DialogData, which already has an explicit cross-platform serialVersionUID = 1L. I also confirmed that AdventureQuestData retains its original calculated desktop UID so old saves will load just fine.

Jetz72 (approved):
Seems alright for now. The automatic UIDs are a problem that should be addressed at some point, but that'll need to be handled delicately.

PR #11720, opened 2026-08-28
Implement Host keyword

BigCrunch22:
- Immersturm Battlefield (Host)

_Engine changes to make the keyword work is included. Implementation feedback is appreciated._

**_Made with assistance from Claude_**


tool4ever:
@Hanmac
I think if we want to pollute the engine with such semi-official, often more of an inside joke and one-shot "mechanics" we should have a clean way to separate them from official/rules-backed first, if only via a CardFactory subclass?

Hanmac:
It also might collide with Host/Augment from Un-Set

Jetz72:
The Host in Host/Augment is a supertype, not a keyword, so I think that won't be an issue. It's also not implemented but I would like to get to it someday.

One engine feature we might consider would be the ability to define simple keywords in resource files instead of the engine. There are a huge number of them that just resolve to a regular ability anyway, so it seems totally do-able to write them as such using our script syntax and some parameter placeholders. That'd be a useful tool for custom sets and for supporting these one-off playtest keywords without cluttering the engine further.

PR #11725, opened 2026-08-28
Improve self-sacrificing mana ritual AI

vanja-ivancevic:
- Correct the mana-ritual source estimate when activating the ritual sacrifices its own source, so that source is not also counted as remaining mana.
- Local A/B testing reproduced the issue with Black Lotus. The previous logic sacrificed Lotus while still one mana short of Juggernaut, Gilded Lotus, and Triskelion, leaving the payoff stranded in hand.
- With the correction, the AI preserves Lotus in all three unreachable scenarios while still pursuing reachable Juggernaut and Gilded Lotus lines.
- A resolved control scenario confirmed that the previous logic put Black Lotus into the graveyard without casting the intended payoff. Existing automatic-payment and AI safety checks also passed.
- This revision leaves the Fallen Empires sacrifice-land cycle and its AI deck-removal hints unchanged.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/ManaAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/ManaAi.java
@@ -147,6 +147,9 @@ public class ManaAi extends SpellAbilityAi {
 
         CardCollection manaSources = ComputerUtilMana.getAvailableManaSources(ai, true);
         int numManaSrcs = manaSources.size();
+        if (manaSources.contains(host) && ComputerUtilCost.isSacrificeSelfCost(sa.getRootAbility().getPayCosts())) {
+            numManaSrcs--;
+        }
         int manaReceived = sa.hasParam("Amount") ? AbilityUtils.calculateAmount(host, sa.getParam("Amount"), sa) : 1;
         manaReceived *= sa.getParam("Produced").split(" ").length;
 


forge-gui/res/cardsfolder/s/svyelunite_temple.txt
```diff
@@ -4,6 +4,6 @@ Types:Land
 R:Event$ Moved | ValidCard$ Card.Self | Destination$ Battlefield | ReplacementResult$ Updated | ReplaceWith$ ETBTapped | Description$ CARDNAME enters tapped.
 SVar:ETBTapped:DB$ Tap | Defined$ Self | ETB$ True
 A:AB$ Mana | Cost$ T | Produced$ U | SpellDescription$ Add {U}.
-A:AB$ Mana | Cost$ T Sac<1/CARDNAME> | Produced$ U | Amount$ 2 | SpellDescription$ Add {U}{U}.
-AI:RemoveDeck:All
+A:AB$ Mana | Cost$ T Sac<1/CARDNAME> | Produced$ U | Amount$ 2 | AILogic$ ManaRitual | AINoRecursiveCheck$ True | SpellDescription$ Add {U}{U}.
```
tool4ever: - this logic wasn't really aimed at lands...how likely is AI to screw up its manabase?
- why are you changing removal only for a single one of these?

vanja-ivancevic: Good points. I tested the land case directly. With Svyelunite Temple as the AI only land and Coral Merfolk in hand, the legacy AI selected and resolved the sacrifice ability, putting Temple in the graveyard while the Merfolk remained in hand. That confirms ManaRitual is not a safe general policy for this land cycle. The single-card removal change was driven only by the immediate Adventure deck, which is not a defensible reason to treat one member of the five-card cycle differently. I have reverted both Svyelunite Temple script changes, so this PR now contains only the independently tested generic self-sacrificing source-count correction and leaves the entire land cycle unchanged.

vanja-ivancevic:
Thanks, I agree that the Fallen Empires lands should be handled as a complete cycle rather than by applying the existing ritual logic to Svyelunite Temple alone. I have kept this PR focused on correcting the source count for existing Black Lotus-style self-sacrificing mana rituals. I will submit the normal-payment handling and consistent AI deck-status change for all five Fallen Empires sacrifice lands as a separate PR.

tool4ever (approved):

PR #11726, opened 2026-08-28
Clean up

tool4ever:


forge-ai/src/main/java/forge/ai/ability/CountersPutAi.java
```diff
@@ -777,8 +777,6 @@ protected AiAbilityDecision doTriggerNoCost(Player ai, SpellAbility sa, boolean
                     && sa.hasSVar(amountStr) && sa.getSVar(amountStr).equals("Count$xPaid")) {
                 // Spend all remaining mana to add X counters (eg. Hero of Leina Tower)
                 int payX = ComputerUtilCost.setMaxXValue(sa, ai, true);
-
-                root.setXManaCostPaid(payX);
```
Hanmac, line 781: @tool4ever why? this causes payX variable to be unused?

tool4ever: thanks, forgot to clean it

the idea is to have the util method handle it all

PR #11732, opened 2026-08-29
Replace weak Master Blue Wizard deck with tempo deck

Nerdslayer:
I battled a Master Blue Wizard and was shocked at how weak the deck it was using was, which is bluewizard_hard_control2.dck. It played weak commons like Arcane Investigator that might not make the cut in a casual draft deck. The other Master Blue Wizard decks, such as the artifact deck that contains all power nine, are on a completely different level. I replaced this deck with a Delver tempo style one that is much more in line with the other decks in terms of power level and coherent strategy.


Simisays:
👍 looks good to me

PR #11733, opened 2026-08-29
Make Liliana Master Black Wizard deck more focused

Nerdslayer:
The Master Black Wizard Liliana deck is weak and unfocused. It is not in line with the power levels of the other Master Black Wizard decks. This removes a lot of the singletons, a lot of which did not fit the deck's strategy, and adds more Liliana planeswalkers.


Simisays:
👍 looks good to me !

PR #11734, opened 2026-08-29
Allow sorcery-speed tap pumps before combat

delebedev:
`PumpAi` currently rejects every tap-cost Pump during its controller's precombat main phase before reaching the later sorcery-speed allowance. This leaves abilities such as Basilisk Gate unusable in the only window where their combat bonus matters.

This lets sorcery-speed abilities continue through the existing targeting, value, and cost checks. The precombat hold remains unchanged for instant-speed tap abilities, and opponent-turn timing is unchanged.

In a Basilisk Gate exact-lethal state, the current path declined the activation in 30 fresh runs; with this change it activated and won in 30 fresh runs. A one-life-higher control remained unwinnable in 30 fresh runs.

I used Codex to help author this change.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/PumpAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/PumpAi.java
@@ -89,7 +89,8 @@ public class PumpAi extends PumpAiBase {
         final Game game = ai.getGame();
         boolean main1Preferred = "Main1IfAble".equals(sa.getParam("AILogic")) && ph.is(PhaseType.MAIN1, ai);
         if (game.getStack().isEmpty() && sa.getPayCosts().hasTapCost()) {
-            if (ph.getPhase().isBefore(PhaseType.COMBAT_DECLARE_ATTACKERS) && ph.isPlayerTurn(ai)) {
+            if (ph.getPhase().isBefore(PhaseType.COMBAT_DECLARE_ATTACKERS) && ph.isPlayerTurn(ai)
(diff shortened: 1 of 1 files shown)


forge-ai/src/main/java/forge/ai/ability/PumpAi.java
```diff
@@ -89,7 +89,8 @@ protected boolean checkPhaseRestrictions(final Player ai, final SpellAbility sa,
         final Game game = ai.getGame();
         boolean main1Preferred = "Main1IfAble".equals(sa.getParam("AILogic")) && ph.is(PhaseType.MAIN1, ai);
         if (game.getStack().isEmpty() && sa.getPayCosts().hasTapCost()) {
-            if (ph.getPhase().isBefore(PhaseType.COMBAT_DECLARE_ATTACKERS) && ph.isPlayerTurn(ai)) {
+            if (ph.getPhase().isBefore(PhaseType.COMBAT_DECLARE_ATTACKERS) && ph.isPlayerTurn(ai)
+                    && !isSorcerySpeed(sa, ai)) {
```
tool4ever, line 93: oops looks like I made that check a bit too greedy, good find

though that part was originally meant to preserve creatures for combat: 4d15a3d
so I suggest going in that direction instead since it's too difficult for AI to judge if Pump is more valuable than acting as attacker/blocker

tool4ever:
Better logic provided via #11739

PR #11736, opened 2026-08-29
Prefer card printings in the selected language in Adventure Mode

Nerdslayer:
I noticed that in Adventure mode I would sometimes encounter cards that are in Japanese. This is especially an issue since no raw card text is available in this UI so I have no way of knowing what the card does. 

To fix this I made it so automatic card selection now prefers printings matching Forge’s selected UI language if available, then English, then any available language. Explicit edition requests remain authoritative.

Adds regression coverage for selected-language preference, English fallback, any-language fallback, and explicit editions. Just let me know if you think the added regression tests are excessive and I can remove them.

Sol assisted in development.

### DIFF
--- a/forge-core/src/main/java/forge/StaticData.java
+++ b/forge-core/src/main/java/forge/StaticData.java
@@ -67,6 +67,11 @@ public class StaticData {
         this(cardReader, null, customCardReader, null, editionFolder, customEditionsFolder, blockDataFolder, "", cardArtPreference, enableUnknownCards, loadNonLegalCards, false, false);
     }
     public StaticData(CardStorageReader cardReader, CardStorageReader tokenReader, CardStorageReader customCardReader, CardStorageReader customTokenReader, String editionFolder, String customEditionsFolder, String blockDataFolder, String setLookupFolder, String cardArtPreference, boolean enableUnknownCards, boolean loadNonLegalCards, boolean allowCustomCardsInDecksConformance, boolean enableSmartCardArtSelection) {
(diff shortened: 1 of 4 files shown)


tool4ever:
Seems redundant to #11709, just without logic to really support printings in multiple languages

Nerdslayer:
@tool4ever Thanks, I didn't know about that PR and there is some overlap. This PR only uses edition-level CardLang, so its multilingual support is incomplete.

The specific problem I was addressing is a foreign-only edition such as PMDA Earthquake being selected automatically while Forge is using English. 

I do not believe the other PR will solve this issue because when English is selected:
- Language-aware unique-card selection is disabled because buildPreferredLanguageAvailability() returns null.
- The downloader uses each edition’s configured CardLang, rather than forcing "en".
- Most normal editions already have CardLang=en, so they still download English images.
- A Japanese-only edition such as PMDA has CardLang=ja, so it remains Japanese.

So #11709 supports choosing non-English languages and downloading multilingual variants, but English is treated as “use the edition default,” not “prefer an English printing.” Even enabling its “prefer language for unique cards” checkbox does not change that, because the predicate is explicitly disabled for English.

This is consistent with #11709’s stated goal of “no behavior change by default,” but it means it does not fix the issue I am trying to address.

If I'm incorrect in how it works and the other PR would result in English card art being preferred when English is selected, I would be happy to just close this PR. Otherwise, perhaps I could add this English preference functionality on top of #11709's new system.

PR #11741, opened 2026-08-31
Fix Greensleeves ability definition

monkyman64:
The minus 3 ability for Greensleeves is not defined as a planeswalker ability, and as such can be activated multiple times per turn.


tool4ever (approved):
thanks

PR #11752, opened 2026-09-01
AI hybrid simulation: don't simulate a subability

bassdr:
## Bug

`CharmAi.chooseOptionsAi` calls `canPlaySa()` on each mode to score it
([CharmAi.java:124](https://github.com/Card-Forge/forge/blob/master/forge-ai/src/main/java/forge/ai/ability/CharmAi.java#L124)),
and `canPlaySa` ends at `saSideEffects`, which in hybrid mode calls
`OnePlaySafetyChecker.isAcceptable`. `GameSimulator.findSaInSimGame` matches by description against
the host card's abilities, and a mode is in neither list, so it returns null:

```
Simulation: SA not found! Pinecone Strike deals 3 damage to target creature. / class forge.game.spellability.AbilitySub
```

6 hybrid-sim games, deck with 2x `Pinecone Strike`: **121 misses, 100% `AbilitySub`.**

## Fix

Skip subabilities in `isAcceptable`. `DelayedTriggerAi.java:28` already guards its own call the
same way.

**The AI decision is unchanged**: a miss scores `Integer.MIN_VALUE`, which `isAcceptable` already
returns `true` for. This drops a wasted full game copy per mode, and the stderr noise. 121 -> 0 on
the same seed.

Hybrid only — full simulation routes modes through `SpellAbilityPicker.chooseModeForAbility` and
never passes one to `simulateSpellAbility`.

## Not fixed here

`isAcceptable` returning `true` on `MIN_VALUE` means any line the simulator cannot reproduce
passes the safety check silently. Happy to look at that separately if it is worth an issue.

Written with Claude Code, per CONTRIBUTING.md. No tests added.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/simulation/OnePlaySafetyChecker.java
+++ b/forge-ai/src/main/java/forge/ai/simulation/OnePlaySafetyChecker.java
@@ -3,6 +3,7 @@ package forge.ai.simulation;
 import forge.ai.simulation.GameStateEvaluator.Score;
 import forge.game.card.Card;
 import forge.game.player.Player;
+import forge.game.spellability.AbilitySub;
 import forge.game.spellability.SpellAbility;
 import forge.game.zone.ZoneType;
 
@@ -13,7 +14,8 @@ public final class OnePlaySafetyChecker {
         // Forge keeps the parent ability on the stack while it resolves. Score actions offered
         // during that resolution incrementally; priority responses need two full stack-resolution
         // branches and are not supported yet.
-        if (sa == null || CHECKING.get()
+        // A subability is never played on its own, so there is nothing here to rule unsafe.
+        if (sa == null || sa instanceof AbilitySub || CHECKING.get()
                 || (!player.getGame().getStack().isEmpty() && !player.getGame().getStack().isResolving())) {
             return true;
         }


tool4ever:
this sounds a bit hallucinated with the mention of `OnePlaySafetyChecker` (Hybrid mode) first and then full later?

also those methods aren't even meant to be used on subabilities

bassdr:
You are right on both counts. Rewrote the PR.

**Subabilities:** the source is `CharmAi.chooseOptionsAi` calling `canPlaySa()` on each mode, which reaches `OnePlaySafetyChecker` via `saSideEffects`. `DelayedTriggerAi.java:28` already guards its own call with `instanceof AbilitySub`; `CharmAi` does not. Fix is now that guard in `isAcceptable` rather than teaching `GameSimulator` to resolve modes. 6 games, deck with 2x `Pinecone Strike`: 121 misses, 100% `AbilitySub`, 0 after.

**Hybrid vs full:** my fault, the old description mixed them. It is hybrid only — full simulation goes through `SpellAbilityPicker.chooseModeForAbility` and never passes a mode to `simulateSpellAbility`. The last line claiming modal spells were "scored unplayable" was wrong.

The AI decision does not change either way, since `isAcceptable` already returns `true` on `MIN_VALUE`. What is left is a wasted game copy per mode and the stderr noise.

Separate question, tell me if it deserves an issue: `isAcceptable` treating `MIN_VALUE` as acceptable means any line the simulator cannot reproduce passes the safety check silently.

tool4ever:
Aah now I understand
I will just update CharmAi, it probably should enter the checks at a lower path anyway
thanks!

and simulation is still not mature enough to spam with errors

bassdr:
Agreed, lower is the right place. Closing this.

Two more unguarded sites if you want them in the same pass, both on `ebf1965737`:

- `CharmAi:286` — the second `canPlaySa(sub)` loop, same `AbilitySub sub`
- `ImmediateTriggerAi:88` and `DelayedTriggerAi:166` — `sa.getAdditionalAbility("Execute")`, unguarded, while `ImmediateTriggerAi:26` and `DelayedTriggerAi:28` guard the identical expression in the sibling method

Only `CharmAi:124` showed up in my logs, so the other three are read from the source, not measured.

PR #11758, opened 2026-09-01
RoL: Fix map crash on plot-critical dungeon

VigilantSera:
Remove alara pack shop which suddenly broke, preventing this plot-critical map from loading


Jetz72 (approved):
Still intending to remove the shop despite changing the `CON` codes?

VigilantSera:
You know I tried adding the shop back and it still causes a crash even after the CFX fix. The log's error message refers to an issue with the ALA set, not the CFX one, not sure why. So I'll leave it out.

Jetz72:
I assumed those showed up because the shop was this one:
https://github.com/Card-Forge/forge/blob/985c69738ec312cee848b5955f2ac1187eb197bd/forge-gui/res/adventure/Realm%20of%20Legends/world/shops.json#L940-L954

But may be best to set it aside until we have a clearer picture.

PR #11759, opened 2026-09-01
Initialize stack item views before tracking

delebedev:
`StackItemView` attaches its tracker before setting constructor properties. When the tracker is frozen, those initial values are deferred. A later `clearDelayed()` can leave the new view empty and make stack-event consumers fail while reading fields such as the activating player.

Initialize the view before attaching the tracker. Later updates retain normal freeze behavior.

I used Codex to help diagnose and author this change.


tool4ever:
this sounds rather fabricated 🤔
please explain how you could produce such an outcome ingame...

delebedev:
You are right, I thought I was onto something around sim runs, but can't reproduce easily. Will reopen if I find plausible way.

PR #11760, opened 2026-09-01
Desktop: default the screen scale when headless

### LINKED ISSUE #11754: `forge sim` cannot start on a headless JVM: GuiDesktop initialises the screen before the mode argument is read
Batch simulation needs no display, but Forge installs the desktop GUI before it reads the mode
argument, so `sim` cannot start on a headless JVM.

[`Main.java:57`](https://github.com/Card-Forge/forge/blob/master/forge-gui-desktop/src/main/java/forge/view/Main.java#L57)
runs `GuiBase.setInterface(new GuiDesktop())`; `case "sim":` is at line 76. `GuiDesktop`
initialises screen scaling in a **static field initialiser** (line 359 → line 353,
`GraphicsEnvironment.getLocalGraphicsEnvironment().getDefaultScreenDevice()`), so line 57 touches
AWT before line 76 discovers the run is batch.

Repro: on a JDK without AWT (Gentoo `dev-java/openjdk[headless-awt]`, no `libawt_xawt.so`) or with
no display, `java -jar forge-gui-desktop-*-jar-with-dependencies.jar sim -d a.dck b.dck -n 10`
fails before any game starts. `xvfb-run` does not help when the library is absent rather than the
display.

`SimulateMatch` has no GUI dependency — only `ForgeConstants` needs an `IGuiBase`, and only for
`getAssetsDir()`. Installing a `java.lang.reflect.Proxy` implementing `IGuiBase` and calling
`SimulateMatch.simulate()` directly works, which is evidence nothing about `sim` needs the desktop
GUI.

Fix: read the mode argument before installing the interface, or make the screen-scale
initialisation lazy rather than a static initialiser.

Unrelated but worth a docs note: startup exceptions go to Sentry, so a failure here prints nothing
and exits 1. `-Dsentry.dsn=` is what surfaces the stack trace.

Forge master `e9cc3207f4`, also 2.0.14. Linux, OpenJDK 25.0.4 (headless-awt).

bassdr:
Fixes #11754.

`GraphicsEnvironment.isHeadless()` returns false whenever a display exists, so the GUI path is
unchanged; verified `isHeadless=false` and the real scale on a headful JVM.

Verified with no `DISPLAY`, stock jar, no other changes:

```
$ java -Dsentry.dsn= -jar forge-gui-desktop-*-jar-with-dependencies.jar sim -d a.dck b.dck -n 2
Simulation mode
Match Result: Ai(1)-A: 1 Ai(2)-B: 0
Match Result: Ai(1)-A: 1 Ai(2)-B: 1
```

Before: `ExceptionInInitializerError` at `Main.java:57`, no games.

Nothing else in the `sim` path needed a display, which was the open question in the issue.

Written with Claude Code, per CONTRIBUTING.md. No tests added.

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/GuiDesktop.java
+++ b/forge-gui-desktop/src/main/java/forge/GuiDesktop.java
@@ -350,6 +350,11 @@ public class GuiDesktop implements IGuiBase {
     }
 
     private static float initializeScreenScale() {
+        // Main installs GuiDesktop before it reads the mode argument, so `sim` and `parse` load
+        // this class with no display.
+        if (GraphicsEnvironment.isHeadless()) {
+            return 1.0f;
+        }
         GraphicsConfiguration gc = GraphicsEnvironment.getLocalGraphicsEnvironment().getDefaultScreenDevice().getDefaultConfiguration();
         AffineTransform at = gc.getDefaultTransform();
         double scaleX = at.getScaleX();


tool4ever:
thanks, though it seems another Dev wants to have the fix scoped a bit wider: #11761

PR #11761, opened 2026-09-01
Make sim mode run without a display

### LINKED ISSUE #11754: `forge sim` cannot start on a headless JVM: GuiDesktop initialises the screen before the mode argument is read

jamincollins:
## The problem

`sim` is documented as the way to run AI matches on headless servers, but it has been unable to start without a display since 48f3ae13154 (2022-04-16) — and it failed with **no output at all**:

```console
$ DISPLAY= java -jar forge.jar sim -d deck1 deck2 -n 1 -q
$ echo $?
1
# nothing on stdout, nothing on stderr
```

## Root cause

Two defects compound:

**1. An eager display touch during class load.** `GuiDesktop` resolved the screen scale from a static initializer, so merely *loading* the class called `GraphicsEnvironment.getDefaultScreenDevice()` and threw `HeadlessException`. `Main` installs the GUI interface before it parses argv, so `sim` died before it could read its own arguments.

**2. The crash was invisible.** Sentry's uncaught-exception handler chains to whatever default handler is already installed — and nothing installed one this early, so the exception was reported and then discarded without ever reaching stderr.

## The fix

- **`GuiDesktop`** — the screen scale is resolved on first use (via a holder, so the read stays lock-free on the paint path) and falls back to `1.0` when there is no display. Loading the class no longer touches the display.
- **`Main`** — sets `java.awt.headless` for the console modes before any AWT class loads, *unless the user set it explicitly*, so a desktop `sim` run exercises the same path a server does. Installs a printing uncaught-exception handler before `Sentry.init`, covering the startup window up to `ExceptionHandler.registerErrorHandling()` (which replaces the default handler; past that point `BugReporter` prints before it reports).
- **Dialog entry points** — degrade to a return value instead of throwing. This matters most for `BugReportDialog`: a crash mid-run built a dialog that itself threw, and that secondary exception *replaced* the crash being reported, so a failed batch run lost the actual bug.

One behavioural note worth flagging: `showOptionDialog` returns `-1` ("closed without choosing", which `FOptionPane` already returns) rather than the default option. `SOptionPane.showConfirmDialog` reads index 0 as *Yes* with `defaultYes=true` in all four overloads, so returning the default would have silently approved every confirmation — including destructive ones such as overwriting a saved game state.

## Testing

Verified on a host with no `DISPLAY`: a full game completes and exits 0, as do 2-player, 3-player, best-of-N (`-m`), Commander, Planechase, and a 4-player bracket tournament. A `-verbose:class` trace of a complete `sim` run shows no display-dependent AWT class is ever initialised. Under Xvfb the GUI is unchanged — output is byte-identical before and after apart from timestamps — and HiDPI still reports `2.0` with `-Dsun.java2d.uiScale=2`.

Full suite: 365 tests, 0 failures.

## On the new test

`CONTRIBUTING.md` asks contributors to avoid adding unnecessary tests, so let me make the case explicitly rather than leave it implied. This regression survived ~18k commits without anyone noticing, which is the "future integration regression" case that guidance carves out. `HeadlessStartupTest` forks a JVM with `-Djava.awt.headless=true`, because `GraphicsEnvironment` caches headlessness on first use and CI runs the suite under Xvfb with a real `DISPLAY` — so an in-process check would exercise the headful path and prove nothing.

It has five methods. Four fork a JVM and assert real behaviour. The fifth, `commandLineModesAreRecognized`, is a plain change-detector over a three-literal list — **I'm happy to drop it if you'd rather keep the suite lean.**

## Not addressed

Pre-existing and out of scope, but found while working here and worth separate issues:

- All 10 shipped `forge-gui/res/defaults/gauntlet/*.dat` files fail to deserialize (`GauntletIO.loadGauntlet` returns null with a `ConversionException`, surfacing at `CSubmenuGauntletContests.initialize`).
- `parse` mode is non-functional: it never initialises `Lang`, so `CardFace` NPEs. It is display-independent — it fails identically under Xvfb.

---

🤖 Written with [Claude Code](https://claude.com/claude-code) (Claude Opus 5), per the AI-agent disclosure in `CONTRIBUTING.md`. Also recorded as a `Co-Authored-By` trailer on the commit.

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/view/Main.java
+++ b/forge-gui-desktop/src/main/java/forge/view/Main.java
@@ -33,6 +33,29 @@ public final class Main {
      * Main entry point for Forge
      */
     public static void main(final String[] args) {
+        // Command line modes must never touch the display: they run on servers, in CI, and over ssh.
+        // This has to happen before any AWT class loads, so it also has to happen before Sentry.init.
+        final String mode = args.length > 0 ? args[0].toLowerCase() : "";
+        if (isCommandLineMode(mode) && System.getProperty("java.awt.headless") == null) {
+            // Only when the user hasn't decided: setting this here would otherwise beat an explicit
+            // -Djava.awt.headless=false on the command line, since it runs before GraphicsEnvironment
+            // caches its answer.
+            System.setProperty("java.awt.headless", "true");
+        }
+
+        // Sentry chains to whatever default handler is already installed, but if none is, an uncaught
+        // exception is reported and then discarded without ever reaching stderr. Install a printing
+        // handler first so failures during the startup below are visible rather than a bare exit code.
+        //
+        // Scope: ExceptionHandler.registerErrorHandling() replaces the default handler outright, so
+        // this one only covers the window up to that call — which is where GuiBase.setInterface runs.
+        // After it, visibility depends on BugReporter, which prints before it reports.
+        Thread.setDefaultUncaughtExceptionHandler((t, ex) -> {
+            System.err.println("Uncaught exception in thread " + t.getName() + ":");
+            ex.printStackTrace();
+            System.err.flush();
+        });
+
         Sentry.init(options -> {
             options.setEnableExternalConfiguration(true);
             options.setRelease(BuildInfo.getVersionString());
@@ -70,8 +93,6 @@ public final class Main {
         }
 
         // command line startup here
-        String mode = args[0].toLowerCase();
-
         switch (mode) {
             case "sim":
                 SimulateMatch.simulate(args);
@@ -86,11 +107,21 @@ public final class Main {
                 break;
 
             default:
-                System.out.println("Unknown mode.\nKnown mode is 'sim', 'parse' ");
+                System.out.println("Unknown mode.\nKnown modes are 'sim', 'parse', 'server'");
                 break;
         }
 
         System.exit(0);
+    }
+
+    /**
+     * Modes that run to completion on the console and never open a window, and so should force
+     * headless AWT. Keep in sync with the named cases of the switch in {@link #main(String[])}.
+     * The switch's {@code default} arm also stays on the console, but an unrecognised argument is
+     * not a mode and only prints usage, so it is deliberately not listed here.
+     */
+    public static boolean isCommandLineMode(final String mode) {
+        return "sim".equals(mode) || "parse".equals(mode) || "server".equals(mode);
     }
 
     @SuppressWarnings("deprecation")
--- a/forge-gui-desktop/src/main/java/forge/view/Main.java
+++ b/forge-gui-desktop/src/main/java/forge/view/Main.java
@@ -47,9 +47,15 @@ public final class Main {
         // exception is reported and then discarded without ever reaching stderr. Install a printing
         // handler first so failures during the startup below are visible rather than a bare exit code.
         //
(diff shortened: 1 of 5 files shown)


forge-gui-desktop/src/main/java/forge/view/Main.java
```diff
@@ -33,6 +33,29 @@ public final class Main {
...
+        }
+
+        // Sentry chains to whatever default handler is already installed, but if none is, an uncaught
+        // exception is reported and then discarded without ever reaching stderr. Install a printing
+        // handler first so failures during the startup below are visible rather than a bare exit code.
+        //
+        // Scope: ExceptionHandler.registerErrorHandling() replaces the default handler outright, so
+        // this one only covers the window up to that call — which is where GuiBase.setInterface runs.
+        // After it, visibility depends on BugReporter, which prints before it reports.
+        Thread.setDefaultUncaughtExceptionHandler((t, ex) -> {
```
tool4ever, line 59: can't we just move `registerErrorHandling()` up to here instead...?

jamincollins: Tried that — it NPEs. `registerErrorHandling()` reads `ForgeConstants.LOG_FILE`, whose class init resolves `ASSETS_DIR` through `GuiBase.getInterface()`, so it needs the interface already set:

```
ExceptionInInitializerError
  caused by: NPE: Cannot invoke "IGuiBase.getAssetsDir()" because GuiBase.getInterface() is null
  at ExceptionHandler.registerErrorHandling(ExceptionHandler.java:117)
```

And the window that needs covering includes `setInterface(new GuiDesktop())` itself, which is where the headless crash was. Comment in a9466096724 now says so.

tool4ever (approved):
LGTM

PR #11765, opened 2026-09-02
Set threads to daemon to gracefully exit the game while disposing assets

kevlahnota:


kevlahnota:
@Agetian ~I can repro the bug running classic then single match then exit the game, with the system.exit(0) it produce the exit code -1073740791, but using daemon threads it exits normally however running adventure then match and exit seems the app is still running in the background.~ EDIT.. fixed the freeze on exit, the timer on awaitnexinput needs to be daemon also. But this needs testing if it will not introduce new issues.

Jetz72:
If something is hanging when the process exits, wouldn't we want to find out what's hanging and why, and terminate it properly rather than forcefully kill every thread? 

Do we have any IO operations on affected threads (e.g. saving adventure data or deck lists)?

kevlahnota:
> If something is hanging when the process exits, wouldn't we want to find out what's hanging and why, and terminate it properly rather than forcefully kill every thread?
> 
> Do we have any IO operations on affected threads (e.g. saving adventure data or deck lists)?

No idea really since there's no exception, on classic mode and adventure mode, just open the match then close the app to trigger the bug, it does close the app but it stays on the task manager forever unless you force the system.exit, it will throw some exit codes number, though i only tested mobile gui

Jetz72:
Hmm. Can't reproduce it on my setup. Might vary by platform? Did you try running it with a debugger, triggering the issue, and then hitting the suspend button? It should let you examine what each thread is doing - it does on IntelliJ at least.

kevlahnota:
> Hmm. Can't reproduce it on my setup. Might vary by platform? Did you try running it with a debugger, triggering the issue, and then hitting the suspend button? It should let you examine what each thread is doing - it does on IntelliJ at least.

to repro, comment out system.exit(0) then when you are in classic or in adventure mode open a match with opponent then close the app, it will call gdx.app.exit() and then wait if it close succesfully, here on windows the java app still resides on task manager even the app window is already closed. on intellij it doesnt stop without this patch. if you use system.exit(0), inteliij throws the exit code -1073740791 (0xC0000409) instead of exit code 0, seems it kills the jvm abruptly (instead of finishing the thread)

tool4ever:
This is tricky:
- I think in some cases like that GUI input timer it's clear they should be daemon since they stay parked by design otherwise
- `ImageFetcher` schedules via `ServicePool` so it's unaffected here
- the end result should be roughly the same, since System.exit would just kill them too so this may actually provide a little bit more grace time
- ideally we had some cooperative architecture so each child can safely terminate on its own but that's obviously a bigger project

tool4ever (approved):

Jetz72:
Ah, commenting out the `System.exit` did it. Looks like the game thread is stuck waiting for `InputSynchronizedBase.cdlDone`, i.e. for the player to do something. Probably wouldn't hurt to have something inform the game that it's been cancelled and that it should no longer be able to block for input. Maybe cause all human players to concede, or just flag the game as over?

PR #11780, opened 2026-09-03
Architecture to skip GUI updates when simulating

tool4ever:
This skips one of the more expensive methods when there is no point in calculating it.
Since View classes might still be useful to interface with the engine in some cases this can be restricted further for developers as needed.

---

First commit also made me discover a nasty engine racing crash:
- when *Spark Double* copies a Planeswalker `loyaltyRep` is null at first
- `ReplacementHandler` now creates another LKI copy for the lookahead and only that will produce a new ETB counter replacement
- this leads to an endless loop (since the real Affected has none it's a new ID each round)

The only reason it didn't fail on master was because `getAbilityText()` was randomly calling `getReplacementEffects()` first to force the caching!

### DIFF
--- a/forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
+++ b/forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
@@ -83,8 +83,8 @@ public class ReplacementHandler {
 
             // CR 614.12 ETB replacements look at what the card would be on the battlefield
             affectedCard = (Card) runParams.get(AbilityKey.Affected);
-            // must force cache these ETB replacements so the LKI copies them
-            affectedCard.getCurrentState().getLandTraitChanges().get(0).applyReplacementEffect(Lists.newArrayList());
+            // must force cache the CardState ETB replacements so the LKI copies them
+            affectedCard.getReplacementEffects();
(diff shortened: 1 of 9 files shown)


forge-game/src/main/java/forge/game/replacement/ReplacementHandler.java
```diff
@@ -83,6 +83,8 @@ public List<ReplacementEffect> getReplacementList(final ReplacementType event, f
 
             // CR 614.12 ETB replacements look at what the card would be on the battlefield
             affectedCard = (Card) runParams.get(AbilityKey.Affected);
+            // must force cache these ETB replacements so the LKI copies them
+            affectedCard.getCurrentState().getLandTraitChanges().get(0).applyReplacementEffect(Lists.newArrayList());
```
Hanmac: Imo looks ugly 

Why not call `getReplacementEffects`?

(Until I implement better trait cache)

tool4ever: sure, that will work too

Hanmac (approved):

tool4ever:
We might be able to extend it to LKI later too, but I still need to check in which cases they could end up for display 🤔

PR #11782, opened 2026-09-03
Fix iOS crash when importing decks from clipboard

NickDeVeau:
## Summary

- avoid the iOS/RoboVM crash caused by named capture groups in the deck import regexes
- preserve named-group behavior on standard JVMs while falling back to indexed capture groups on regex engines that reject the syntax
- cover every importer pattern and a representative sideboard/card line with a portability regression test

## Root cause

Opening **Import from Clipboard** initializes `DeckRecognizer`. RoboVM's regex implementation rejects Java named capture syntax such as `(?<pre>...)`, which raised a `PatternSyntaxException` during static initialization and terminated the iOS app with an `ExceptionInInitializerError`.

## Testing

- focused Maven regression test passed for `DeckRecognizerPortablePatternTest`
- iOS IPA built successfully in GitHub Actions
- installed and verified on a physical iPad running iPadOS 18.7.3
- confirmed the original clipboard import crashed before the change and imports successfully afterward

## Contribution note

The issue was reproduced and the fix was validated on-device by the contributor. OpenAI Codex assisted with diagnosis, implementation, test creation, and PR preparation.

### DIFF
--- a/forge-core/src/main/java/forge/deck/DeckRecognizer.java
+++ b/forge-core/src/main/java/forge/deck/DeckRecognizer.java
@@ -33,6 +33,7 @@ import org.apache.commons.lang3.tuple.Pair;
 import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 
 /**
  * <p>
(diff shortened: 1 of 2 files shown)


tool4ever:
This seems to be the same cause as #8976
so I assume it could crash on Android too

While this does add some overhead I wouldn't mind having this transformation done via code, as simply refactoring this class into indexes too might be a more messy alternative?

tool4ever (approved):

PR #11785, opened 2026-09-04
Charge the Sanctum and Compel Brutality (FRA)

Fulgur14:


Fulgur14:
I think Compel Brutality will need some new AILogic for its second mode.

tool4ever (approved):
I think the second mode should work already
if you want to just put `AITgts$` on it so it always selects yours with the highest loyalty

Fulgur14:
We'll try like this for now, I think :)

PR #11786, opened 2026-09-04
Crash Fix when restarting on Android

kevlahnota:


forge-gui-mobile/src/forge/assets/Assets.java
```diff
@@ -129,54 +130,57 @@ private Assets() {
...
-            borders.clear();
-        if (deckbox != null)
-            deckbox.clear();
-        if (cursor != null)
-            cursor.clear();
-        if (fonts != null)
-            fonts.clear();
-        Forge.safeDispose(manager);
+        if (!isDisposed) {
+            isDisposed = true;
```
Hanmac, line 134: @kevlahnota probably should have done an early return ala:

```java
 if (isDisposed) {
    return;
 }
```

PR #11789, opened 2026-09-05
Allow shared fetches when other players have no cards to retrieve

delebedev:
An empty opposing graveyard can veto Exhume even when the AI has a creature to return. Ignore empty participants in shared library/graveyard retrieval while retaining the AI’s own candidate and battlefield-entry checks.

I used Codex to help author this change.

### DIFF
--- a/forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
+++ b/forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
@@ -333,6 +333,11 @@ public class ChangeZoneAi extends SpellAbilityAi {
             ComputerUtilCost.setMaxXValue(sa, ai, sa.isTrigger());
         }
 
+        final boolean allowEmptyDefinedPlayers = isMandatoryCreatureGraveyardReturn(sa, origin, type);
+        boolean emptyDefinedPlayer = false;
+        boolean nonEmptyDefinedPlayer = false;
+        boolean ownCanReturn = false;
+
         for (final Player p : pDefined) {
             CardCollectionView list = p.getCardsIn(origin);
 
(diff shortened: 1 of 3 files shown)


delebedev:
~~Marking this as draft while I iterate on a more general fix for empty participants in mandatory zone changes.~~

forge-gui-desktop/src/test/java/forge/ai/ability/ChangeZoneAiTest.java
```diff
@@ -54,4 +54,133 @@ public void testXTargetCountIgnoresTheAisOwnPermanents() {
                     t.getController().equals(ai));
         }
     }
+
+    @Test
+    public void testExhumeWithEmptyOpponentGraveyard() {
```
delebedev, line 61: happy to drop the tests, they only demonstrate that logic works for mentioned cards

tool4ever: they are actually harmful when *Weird Harvest* is still marked `RemoveDeck:All` :/

forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
```diff
@@ -333,6 +333,11 @@ private static AiAbilityDecision hiddenOriginCanPlayAI(final Player ai, final Sp
             ComputerUtilCost.setMaxXValue(sa, ai, sa.isTrigger());
         }
 
+        final boolean allowEmptyDefinedPlayers = canIgnoreEmptyDefinedPlayers(ai, sa, origin,
```
tool4ever: check if *Exhume* still needs the `NeedsToPlayVar` hint

forge-gui-desktop/src/test/java/forge/ai/ability/ChangeZoneAiTest.java
delebedev: dropped tests

forge-ai/src/main/java/forge/ai/ability/ChangeZoneAi.java
delebedev: Removed NeedsToPlayVar and the unused X definition. 
AI still rejects empty and noncreature-only own graveyards without it in my tests.

tool4ever (approved):

PR #11790, opened 2026-09-05
AI: don't block with creatures that die before they deal damage

### LINKED ISSUE #11695: Computer Doesn't Understand Engulfing Slagwurm
**Describe the bug**
The AI does not have the proper rules for dealing with "when blocked" triggers. Engulfing Slagwurm destroys all creatures blocking or blocked by it. When you attack with an engulfing slagwurm or similar creature the AI blocks like normal combat will still occur and loses much more than required to block it. The issue appears worse when the wurm gains trample, where the AI tries to block the wurm with every creature on its board.

**To Reproduce**
1. Start a battle with a "Baru, Wurmspeaker" deck or any deck with slagwurm
2. Play and give engulfing slagwurm trample
3. Attack and witness the AI blocks it with their entire board.

**Expected behavior**
I would expect the AI to either use chump blockers when the wurm does not have trample, or only block with indestructible creatures when it does.

 - Device: Google Pixel 10
 - OS: Android 17
 - Version: Forge v2.0.15-SNAPSHOT-08.20

liamiak:
Fixes #11695.

A creature that destroys its blockers on block — Engulfing Slagwurm, Cockatrice, Dead-Iron Sledge etc. — kills them before the combat damage step. Those blockers never deal their damage, and against trample they never soak any either. Three block routines assumed otherwise:

- `makeGangBlocks` summed damage from blockers that never deal it, so it formed a gang that could not kill the attacker. Three Craw Wurms vs an Engulfing Slagwurm: the AI blocks with two at any life total, expecting to lose one and kill the wurm. It loses both, the wurm lives and gains 8 life.
- `makeChumpBlocks` chumped a trampling one whose damage went through in full anyway. Defender at 12: blocking costs a creature, prevents 0 damage and gives the attacker 2 life.
- `makeGangNonLethalBlocks` made its "neither of which will die" double block with two creatures that both die. Menace attacker vs two Wall of Stone (0/8, which do survive its damage): both walls destroyed, attacker takes nothing.

No new heuristic — each now asks the question the neighbouring code already asks. `reinforceBlockersToKill` and the first-strike gang loop both refuse a blocker that `canDestroyBlockerBeforeFirstStrike`, and `shieldDamage` already returns 0 for one. `canDestroyAttacker` and `canDestroyBlocker` carry the same guard internally, which is why every pairwise block decision was already right; the three gaps are exactly the places that reason across several blockers, bypassing those predicates.

The hunks need to land together: fixing the chump alone leaves the AI in danger, and it escalates to a three-creature gang instead of one wasted chump.

Cost, single JVM with the arms interleaved, 8 attackers / 12 blockers / 24 trigger permanents: **+0.6%** on `declareBlockersFor` for a board with no blocker-destroyer, and **−29.9%** with one, since pruning doomed blockers collapses the nested double/triple block search. The check runs after the cheap value tests, so it only sees candidates that already passed them.

Three tests in `BasicBlockTests`. Each fails on master with the numbers above; the control assertions in them — the AI still chumps to survive at 9 life, and still makes the menace block against an attacker with no such trigger — pass on master too.

Not covered: `makeMultiChumpBlocks` has the same blind spot, but only at the trample + menace + destroyer intersection, and I have no case for it. Triggers whose conditions need a triggering context (No Quarter's `ValidBlocker$ LessPowerThanAttacker`) are invisible to `combatTriggerWillTrigger` regardless of this change.

https://claude.ai/code/session_01NAz6n9nxxxN2K7gYzqUaU5

### DIFF
--- a/forge-ai/src/main/java/forge/ai/AiBlockController.java
+++ b/forge-ai/src/main/java/forge/ai/AiBlockController.java
@@ -444,7 +444,10 @@ public class AiBlockController {
                         && !ComputerUtilCombat.dealsFirstStrikeDamage(c, false, combat)) {
                     return false;
                 }
-                return lifeInDanger || wouldLikeToRandomlyTrade(attacker, c, combat) || ComputerUtilCard.evaluateCreature(c) + diff < ComputerUtilCard.evaluateCreature(attacker);
+                // a blocker destroyed before it deals its damage can't help the gang kill the attacker
+                return (lifeInDanger || wouldLikeToRandomlyTrade(attacker, c, combat)
+                        || ComputerUtilCard.evaluateCreature(c) + diff < ComputerUtilCard.evaluateCreature(attacker))
+                        && !ComputerUtilCombat.canDestroyBlockerBeforeFirstStrike(c, attacker, false);
             });
             if (usableBlockers.size() < 2) {
                 return;
@@ -563,8 +566,10 @@ public class AiBlockController {
             final List<Card> blockGang = new ArrayList<>();
             int absorbedDamage; // The amount of damage needed to kill the first blocker
 
-            List<Card> usableBlockers = CardLists.filter(blockers, c -> c.getNetToughness() > attacker.getNetCombatDamage() // performance shortcut
-                    || c.getNetToughness() + ComputerUtilCombat.predictToughnessBonusOfBlocker(attacker, c, true) > attacker.getNetCombatDamage());
+            List<Card> usableBlockers = CardLists.filter(blockers, c -> (c.getNetToughness() > attacker.getNetCombatDamage() // performance shortcut
+                    || c.getNetToughness() + ComputerUtilCombat.predictToughnessBonusOfBlocker(attacker, c, true) > attacker.getNetCombatDamage())
+                    // surviving the damage is no use if something destroys it before the damage step
+                    && !ComputerUtilCombat.canDestroyBlockerBeforeFirstStrike(c, attacker, false));
             if (usableBlockers.size() < 2) {
                 return;
             }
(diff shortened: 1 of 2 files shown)


forge-ai/src/main/java/forge/ai/AiBlockController.java
```diff
@@ -675,6 +680,11 @@ private void makeChumpBlocks(final Combat combat, List<Card> attackers, boolean
 
         boolean blocked = false;
         List<Card> chumpBlockers = getPossibleBlockers(combat, attacker, blockersLeft, true);
+        if (attacker.hasKeyword(Keyword.TRAMPLE)) {
```
tool4ever: you're filtering too early, see the related logic below which might select another attacker

liamiak: Good catch — it filtered the candidate list before the logic below that consumes it. Moved: the check is per-candidate at the point of decision now, which matches what `reinforceBlockersAgainstTrample` already does a few lines down. `shieldDamage == 0` sets `damageAbsorbed` to 0, so it falls out of the `getNetCombatDamage() > damageAbsorbed` case the redirect already handles rather than sitting in front of it.

Following the same thread: `makeGangNonLethalBlocks` was hand-rolling "neither of which will die" as a toughness test, and `getSafeBlockers` already answers that through `canDestroyBlocker` — so that hunk is now one line calling it. Worth flagging that this also changes the deathtouch case, where the old toughness test counted a big blocker as a survivor. Happy to drop that half if you'd rather keep this PR narrow. Thanks for the pointer.

tool4ever: I actually went back more to your previous approach since it was easier for clean up 😊

tool4ever (approved):

PR #11792, opened 2026-09-05
Realm of legends 1.075: New Dungeon: Crypt of Broken Rules

VigilantSera:


VigilantSera:
There is one edit to common: the weasel enemy spirtemap has been wrong for some time, but since I'm using it again and found the fix, I included it here.

PR #11793, opened 2026-09-05
Implement Chira, All In

BigCrunch22:
# Chira, All In — Engine Support

## Summary

Implements the previously-unimplemented card **Chira, All In**. Its trigger condition covers five different kinds of random selection — discarding, targeting, choosing a mode, flipping a coin, rolling a die — each of which is already a general action in the engine rather than something specific to this card. Each is wired to a `*All`-style trigger event that fires once per occurrence of the action, however many cards/targets/modes it covers, with an optional `Random$` param on the trigger script narrowing it to the random case. Three of the five clauses (targeting — in two different forms — and mode choice) needed real engine changes, covered below with the CR rules that shape them; the other two reuse existing triggers unchanged.

## Card

> **Chira, All In** — {3}{R} — Legendary Creature — Human Rogue — 3/3
> Whenever you discard one or more cards at random, select one or more targets at random, choose one or more modes at random, flip one or more coins, or roll one or more dice, put a +1/+1 counter on Chira and create a tapped Treasure token.

## Targets chosen at random

Chira's target clause has to cover two structurally different actions that can each end in "a target was assigned at random": choosing a spell or ability's *own* targets, and later *redirecting* an already-targeted spell's targets to something else. Both report to the same `TriggerType.TargetChosenAll` event, so a card watching for it doesn't need to know which path produced it.

### Choosing targets — `TargetSelection.java`

CR 601.2c / 602.2b treat choosing a spell or ability's targets as one step of casting or activating it, however many targets that step ends up picking. `fireTargetChosenTrigger(boolean random)` reflects that directly: it's guarded to fire at most once per targeting sequence, and is called from every place targeting can finish — the random-target branch, both "only one legal target exists" auto-target shortcuts, the general recursive completion check for ordinary manual multi-target selection, and both exit points of the separate stack-targeting loop (e.g. a counterspell choosing what to counter). Only the random-target branch passes `random = true`, which is what lets `Random$ True` on a trigger script narrow it to just that case.

### Redirecting targets — `ChangeTargetsEffect.java`

Redirecting an already-cast spell's targets is a separate action from choosing them, with its own, more constrained rule:

> **CR 115.7a** If an effect allows a player to "change the target(s)" of a spell or ability, each target can be changed only to another legal target. If a target can't be changed to another legal target, the original target is unchanged... If all the targets aren't changed to other legal targets, none of them are changed.

That all-or-nothing clause applies per spell being redirected, not per individual target — and a single resolution of `ChangeTargetsEffect`'s `RandomTarget` mode can be redirecting several spells at once, each with its own chain of sub-ability targets. So for each targeted spell, a random replacement is staged for its whole chain before anything is written to the game; if even one sub-target in that chain has no legal candidate, the staged picks for that spell are discarded and its targets are left exactly as they were, and the effect moves on to the next targeted spell rather than treating one exhausted candidate pool as a reason to stop redirecting everything else.

Because one resolution can end up reassigning several targets this way (e.g. a spell and its copy both being redirected), every replacement actually committed is collected and reported to `TargetChosenAll` as a single batch once resolution finishes — the same fire-once intent as `TargetSelection`, just via an explicitly collected list instead of a live read, since here the targets belong to other spells, not this ability's own. That report is issued from a `finally` block rather than a plain statement after the loop, since the sibling `ChangeSingleTarget` mode (CR 115.7b, "change *a* target") can still return out of the method early on no targets found, and this guarantees that return can never strand a batch already collected before it runs.

## Modes chosen at random — `TriggerFacesDilemma`, `CharmEffect.java`

CR 700.2 is specific about what counts as a mode:

> **CR 700.2** A spell or ability is modal if it has two or more options in a bulleted list preceded by instructions for a player to choose a number of those options, such as "Choose one —." Each of those options is a mode.

That's a narrower category than "any ability with a choice in it." Forge's `DB$/AB$ Charm` is built specifically for that bulleted "Choose one/two/three —" structure — every sampled card using it (793 in the cardsfolder) has that structure in its Oracle text — which makes it the right mechanism to key a "modes chosen" trigger off of.

`TriggerType.FacesDilemma` — the "you faced a dilemma," i.e. chose mode(s), trigger already used by Seven of Nine — fires from `CharmEffect.makeChoices()` whenever a mode is actually chosen.

`CharmEffect.makeChoices()` had a real, pre-existing bug independent of Chira: its random-mode branch (`Random$ True`) returned before ever reaching the code that fires `FacesDilemma`, so no random-Charm card — Chira aside — could trigger off its own mode choice. Fixed by firing `FacesDilemma` (with `AbilityKey.Random = true`) from that branch too, once modes are actually chosen. `TriggerFacesDilemma` gained an optional `Random$` filter param, matching the same pattern as `TriggerDiscardedAll`/`TriggerTargetChosenAll`, so a card can watch for "any mode chosen" or "a mode chosen at random" as needed.

This fix applies to both human- and AI-controlled players: `CharmEffect.makeChoices()` is the one method both `PlaySpellAbility.playAbility` (human) and `PlayerControllerAi.prepareSingleSa` (AI) call to resolve a Charm-typed ability's modes, random or not.

## Cards discarded / coins flipped / dice rolled

The remaining three clauses reuse existing, already-general triggers as-is:

- `DiscardedAll` already fires for any discard and accepts an optional `Random$` param to narrow to random discards, which Chira's script uses.
- `FlippedCoinOnce` and `RolledDieOnce` fire once per flip/roll instruction regardless of how many coins or dice were involved. Flipping and rolling are inherently random, so neither needs a `Random$` gate.

## Card script — `chira_all_in.txt`

Five triggers off the same Oracle text, one per clause, `Secondary$ True` on all but the first so only one line of rules text is shown. Unchanged by the `FacesDilemma` fix — this line was already fed only by the `Charm` path:

```
T:Mode$ DiscardedAll    | TriggerZones$ Battlefield | ValidPlayer$ You | Random$ True | Execute$ TrigCounterTreasure | TriggerDescription$ ...
T:Mode$ TargetChosenAll | TriggerZones$ Battlefield | ValidPlayer$ You | Random$ True | Execute$ TrigCounterTreasure | Secondary$ True | TriggerDescription$ ...
T:Mode$ FacesDilemma    | TriggerZones$ Battlefield | ValidPlayer$ You | Random$ True | Execute$ TrigCounterTreasure | Secondary$ True | TriggerDescription$ ...
T:Mode$ FlippedCoinOnce | TriggerZones$ Battlefield | ValidPlayer$ You |               Execute$ TrigCounterTreasure | Secondary$ True | TriggerDescription$ ...
T:Mode$ RolledDieOnce   | TriggerZones$ Battlefield | ValidPlayer$ You |               Execute$ TrigCounterTreasure | Secondary$ True | TriggerDescription$ ...
SVar:TrigCounterTreasure:DB$ PutCounter | Defined$ Self | CounterType$ P1P1 | CounterNum$ 1 | SubAbility$ DBTreasure
SVar:DBTreasure:DB$ Token | TokenTapped$ True | TokenScript$ c_a_treasure_sac | TokenOwner$ You
```

_**Code changes made with heavy assistance from Claude**_


forge-game/src/main/java/forge/game/ability/effects/ChangeTargetsEffect.java
```diff
@@ -47,108 +50,147 @@ public void resolve(SpellAbility sa) {
...
 
-        for (final SpellAbility tgtSA : sas) {
-            SpellAbilityStackInstance si = stack.getInstanceMatchingSpellAbilityID(tgtSA);
-            if (si == null) {
-                // If there isn't a Stack Instance, there isn't really a target
-                continue;
-            }
+        // Targets RandomTarget reassigns are collected here and reported as one batch below,
+        // not one trigger per target changed.
+        final List<GameEntity> randomlyChosenTargets = new ArrayList<>();
```
tool4ever: this looks rather messy - I assume *Chef's Kiss* would be a card you're aiming at here

However without an official ruling I'd argue that this is not what happens as in general the CR will break such actions down more:
>608.2f. Some spells and abilities include actions taken on multiple players and/or objects. In most cases, each such action is processed simultaneously. If the action can't be processed simultaneously, it's instead processed considering each affected player or object individually. [...]

this will benefit us since you can move the logic closer to where `BecomesTargetOnce` fires, perhaps even reuse that type too 🤔

BigCrunch22: _**Here is the answer I got from Claude addressing reusing BecomesTargetOnce**_

---

### Why not `BecomesTargetOnce`?

Short answer: it answers a different question.

**What it does:** `BecomesTargetOnce` fires from `SpellAbilityStackInstance#updateTarget()` and reports which *objects* newly became targets. It has no notion of *who* made that choice or *whether it was random* — it doesn't need one, since it backs cards like "whenever this becomes the target of a spell."

**What Chira needs:** the opposite framing — *did a player select target(s) at random*. That requires:
- the chooser (`ValidPlayer$ You`)
- a `Random` flag

Neither exists on `BecomesTargetOnce` or its run params. Adding them would change what that trigger means for every card that already uses it, not just this one.

tool4ever: hmn, it doesn't hurt to introduce Random into it (just like you did with FacesDilemma)
also `AbilityKey.Cause` is already available to check if "You" did it
(technically Chooser$ could mess with that but it's not used together right now so a small code comment asserting that constraint is ok)
but yea a mix of subabilities where some target non-randomly might screw things up? I might think about this one a bit more

tool4ever: Ok, I see no real reason not to proceed with my suggested simplification:
a) MagicStacks `BecomesTargetOnce` can just check for any `TargetsAtRandom$` sub (and maybe for non-empty too even if that also seems to no exist right now?) - while that meaning will be slightly different from your version just leave a comment (it can always be reworked later if needed)
b) SpellAbilityStackInstaces `updateTarget` can just receive it via an extra param - though it should be noted that the logic seems to be different here already since it can execute these triggers for multiple subabilities which seems wrong imo, either you address that directly here or I'll just try to take care of it afterwards
c) I doubt we'll get a more specific variant of this trigger soon

tool4ever (approved):

PR #11797, opened 2026-09-06
Implements 3 Non-Legal die roll cards

Rensen3:
**Implemented the following cards:**
- GO TO JAIL
- Genevieve, Conniving Dragon
- Squirrel-Powered Scheme

includes engine changes to allow replacement effects to change die roll results

I tried to implement as many missing die roll related cards that I could, thinking that since we now have black-bordered cards that do something similar there wouldn't be a problem, but turns out most of them are hardcoded.

_I want to go over each missing die roll related card that I didn't implement and the reasons why:_
 - Krark's Other Thumb: Doubling the dice rolls and ignoring half is easy enough but that's not what the card actually does. Needs its own logic to roll dice in pairs. Gets extra complicated when involving more cards that cause rerolls like Barbarian Class or multiple of this cards.
 - Scooh: il-defined rules
 - Clam-I-Am: Would need to be hardcoded (reroll only on 3s. Possible multiple times if you keep getting 3s)
 - Goblin Blastronaunts: Triggers after the spell/ability resolves so there is nothing on the stack to copy
 - Goblin Bookie: Similar to Monitor Monitor but that one is hardcoded
 - Goblin Girder Gang: Art Matters
 - Pippa, Duchess of Dice: Similar to Monitor Monitor but that one is hardcoded
 - Snickering Squirrel: Similar to Xenosquirrels/Nightshift but those are hardcoded
 - Socketed Sprocketer: Similar to Vedalken Squirrel-Whacker but that one is hardcoded. Storing results appears on Centaur of Attention.
 - The Big Idea: Needs its own roll additional die logic
 - Wall of Fortune: Similar to Monitor Monitor but that one is hardcoded
 - Ricochet: Similar to Chaos Dragon but remembering the lowest player and rerolling if there is a tie
 - Free-Range Chicken: Similar to Long List of the Ents and Volo, Itinerant Scholar but noting die roll results rather than creature types. Maybe have StoreSVar support lists.

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/RollDiceEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/RollDiceEffect.java
@@ -11,6 +11,7 @@ import forge.game.event.GameEventRollDie;
 import forge.game.player.Player;
 import forge.game.player.PlayerCollection;
 import forge.game.player.PlayerController;
+import forge.game.replacement.ReplacementResult;
 import forge.game.replacement.ReplacementType;
 import forge.game.spellability.SpellAbility;
 import forge.game.trigger.TriggerType;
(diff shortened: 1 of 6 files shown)


forge-game/src/main/java/forge/game/ability/effects/RollDiceEffect.java
```diff
@@ -228,6 +229,15 @@ private static int rollDiceForPlayer(SpellAbility sa, Player player, int amount,
             resultsList.add(new DieRollResult(unmodified, unmodified));
         }
 
+        //replacement effects that affect die results
+        for (DieRollResult roll: resultsList) {
+            repParams.put(AbilityKey.DiceResult, roll.getModifiedValue());
+            if (Objects.requireNonNull(player.getGame().getReplacementHandler().run(ReplacementType.RollDice, repParams)) == ReplacementResult.Updated) {
```
tool4ever, line 235: you can't just run a replacement a second time like that

PR #11798, opened 2026-09-06
Sound: take clips from the software mixer instead of one device line each

bassdr:
`AudioSystem.getLine` returns a `DirectAudioDevice` line per `Clip`, and `SoundSystem.loadedClips`
caches one `AudioClip` per `SoundEffectType` for the life of the process. So a session holds an open
device line for every distinct effect it has ever played — before any pooling, and at zero
concurrency. On PulseAudio/PipeWire each line is a separate application stream.

Once the device stops handing out lines, `Clip.open()` parks in `DirectAudioDevice.nOpen` while
holding the mixer's monitor, and the game thread is stuck there. That is the freeze in this issue's
original report and in the two dumps posted on it.

`SoftMixingMixer` mixes in Java and opens the device once. It is not a registered `MixerProvider`,
so it needs direct construction and `--add-exports java.desktop/com.sun.media.sound=ALL-UNNAMED`
(added to the desktop launcher args). Without the flag `softMixer` is null and behaviour is
unchanged.

Tested on Linux/PipeWire, 90s of effects at ~3/s over the 39 files in `res/sound`:

| | before | after |
|---|---|---|
| PipeWire streams | one per distinct effect, 137 nodes, then hang | **1, flat** |
| open device lines | 54 | 0 |

All 39 effects open on it and keep `MASTER_GAIN`, so the volume control is unaffected.

**A `Clip` is never released, which the other two commits fix.** Java's garbage collector does not
cover this: a `Clip` holds its resources until `close()`, collected or not.

- *Between matches:* `dispose()` assigned null to a loop variable and cleared the byte[] cache,
  leaving every pooled clip open. It is reachable only from `invalidateSoundCache()`.
- *During one:* `getIdleClip()` adds a clip whenever the others are busy and nothing removes it, so
  each pool grows to its peak concurrency and keeps a decoded copy of the sound per entry. Idle
  clips are now closed after 10s, one kept per effect so a single play still costs nothing.

Same 39 files, driven to the pool's 16-clip cap: 594 clips and 53 MB held with neither fix; 253 at
the peak and **39 after 20s quiet, 13 MB**, and `dispose()` now returns to 0. Replaying every
effect after a reap works and reuses the kept clip.

The line cap and open timeout that were on this branch before are dropped: they bounded the
symptom, and the mixer removes the condition.


forge-gui-desktop/src/main/java/forge/sound/AudioClip.java
```diff
@@ -116,12 +126,13 @@ private ClipWrapper getIdleClip() {
     }
 
     private ClipWrapper addClip() {
-        if (clips.size() < maxSize && !failed) {
+        if (clips.size() < maxSize && !failed && openClips.get() < MAX_TOTAL_CLIPS) {
```
tool4ever, line 129: are we even storing more than one clip this way? this sounds rather speculative again

forge-gui/src/main/java/forge/sound/SoundSystem.java
```diff
@@ -25,18 +30,41 @@ public class SoundSystem {
...
+     * Sound effects are requested from the game thread, and opening an audio line can block in the
+     * driver for as long as the device stays busy -- which stalls the match. Playback is
+     * fire-and-forget, so it runs here instead, and requests the device cannot keep up with are
+     * dropped rather than queued behind it.
+     */
+    private static final int MAX_QUEUED_SOUNDS = 32;
+    private static final ExecutorService soundQueue = new ThreadPoolExecutor(1, 1, 0L, TimeUnit.MILLISECONDS,
+            new ArrayBlockingQueue<>(MAX_QUEUED_SOUNDS),
+            r -> {
+                Thread t = new Thread(r, "Forge sound");
```
tool4ever: hmn the AltSystem is already thread-based

I'm not sure if it makes sense to mix both more together or if the additional stuff it does is still needed these days

also this would affect GDX/Android too 🤔

bassdr:
Thanks — the second point is right and I've dropped that part.

**On the threading:** you're correct that this doesn't belong here. `SoundSystem` is in the shared
`forge-gui` module and is called from the mobile port (`GuiMobile`, `AdventurePlayer`), so my
executor would have shipped to GDX/Android as well, and `AltSoundSystem` already does threaded
playback. I should have checked which module I was editing before adding a second threading model
beside an existing one. It's removed — the branch no longer touches how playback is dispatched.

**On "are we even storing more than one clip this way":** fair challenge, and I under-argued it.
The hang I saw had **138 open audio lines** with **39 files** in `res/sound/`. If the pool only
ever held one clip per file the ceiling would be 39, so it was averaging ~3.5 per file and the
`maxSize = 16` path is genuinely being exercised. Music was disabled in that session, so those are
effects, not the music path. That count is from `jstack` on the real hang, not from a harness.

I'll be straight about what I can and can't show, though: I have **one** hang, on one machine
(Linux/PipeWire, JDK 25). Whether a device blocks rather than failing the open cleanly is
driver-dependent, so I can't offer a reproduction you can run. If you'd rather read this as a
resource-leak report than a freeze report, that's a fair reading — the leak itself is visible in
the source regardless of what any particular device does with it.

**One thing worth knowing before deciding the shape of a fix:** `AudioClip.dispose()` is only
reachable from `invalidateSoundCache()`, which fires when the audio menu enumerates sound sets. It
is never called during a match. So making `dispose()` work fixes cleanup at menu time and does
nothing about accumulation while you play — which is why I added the line count rather than only
fixing `dispose()`. If you'd prefer to bound it a different way (close each clip when it finishes
instead of pooling 16, or lower `maxSize`), that's entirely reasonable and I'd be glad to redo it
that way, or to leave it with you.

What's left in the branch is now: `dispose()` actually closing clips, a count of open lines so the
total can be capped, and `hasResource()` no longer constructing a clip to answer whether a file
exists.

bbabnikk:
I caught the freeze red-handed with two jstack dumps 10s apart on a stock 2.0.14 install (Java 26, second Commander game vs AI). The game thread is parked in the identical  native frame in both dumps, CPU blocked: 

  "Game-2" ... runnable                                                                                                           
    at DirectAudioDevice.nOpen(Native Method)
    at DirectAudioDevice$DirectDL.implOpen(...)
    at DirectClip.implOpen(...) / .open(...)
    at AudioClip$ClipWrapper.createClip(AudioClip.java:215)
    at AudioClip.<init>(AudioClip.java:72)
    at SoundSystem.fetchResource(SoundSystem.java:63)
    at SoundSystem.play(SoundSystem.java:144)
   ...                                                                                     
    at MagicStack.resolveStack(MagicStack.java:628) 

Same session also had **135 leaked `Direct Clip` threads**, so the unreleased-lines mechanism in your comment checks out. 

Against this evidence, three gaps remain in this PR: 

**1. A wedged device still freezes the game thread.** The cap makes exhaustion unlikely, but the game thread still performs a blocking `Clip.open()` on every cache miss (`new AudioClip` → `addClip`) and every pool growth (`getIdleClip` → `addClip`). If the device wedges for any non-exhaustion reason (ALSA suspend, BT reconnect, PipeWire hiccup), that call parks in `nOpen` forever with no timeout, the exact stack above. A cap can't prevent that; only getting the open off the game thread (or bounding the wait) can. Suggestion: submit the `new ClipWrapper(...)` construction to a shared daemon executor and `future.get(3, SECONDS)`, falling back to `Dummy` on timeout, so the worst case is a few seconds of silence instead of a dead game. 

**2. Open failure still propagates into the game loop.** When the device is out of lines rather than wedged, `createClip` catches `LineUnavailableException`, prints, then throws `MissingResourceException`  which travels up through `play()` → `receiveEvent()` → `EventBus.post()` (direct executor, same thread) → `Game.fireEvent()` → `resolveStack`. So the failure mode just changes from freeze to game-thread death. Suggestion: catch in `addClip` and latch to `Dummy` (permanently for missing files, with a short retry cooldown for transient device errors), so audio degrades to silence instead of killing the match. 

**3. The 64-cap is a one-way ratchet.** Nothing closes a line until app exit, so once 64 opens have accumulated, every genuinely-new concurrent need returns `Dummy` for the rest of the session, even though most of those 64 lines are sitting idle. Suggestion: reap surplus idle clips in `getIdleClip` (e.g. keep 2 open per file, `close()` the rest). Steady state then stays small enough that the cap never binds, and the two mechanisms compose instead of overlapping. Needs a `closed` guard in `start`/`loop`/`stop` (all under the existing `synchronized (this)`) so a reaped clip can't be started concurrently. 

None of this diminishes the PR,  the `hasResource` file-check fix in particular closes a leak source I hadn't spotted (a clip opened per existence check).

tool4ever:
I don't see how `fetchResource` leaks anything...?
it will merely prefill the cache 🤷‍♂️

bbabnikk:
"leak" was the wrong word for it. Since hasResource false only happens when the file is missing (no line opened), and a true answer is always followed by playing that same cached clip, it prefills at most what playback would open anyway.                                                                 

The change is still worth keeping, just on different grounds:

- The old version can throw instead of answering. On a cache miss with an exhausted device, new AudioClip → createClip throws MissingResourceException (after LineUnavailableException) out of what looks like a pure boolean query running on the game thread. The file check returns false cleanly.
- The old version can block instead of answering. Answering existence currently costs a file search + MP3 decode + Clip.open() on the game thread during event visualization, the same nOpen park from the dumps above, reachable without any playback. The file check can't block. 

So: not a leak fix, but game-thread protection, which is the same thesis as points 1-2. Points 1-3 otherwise stand as written.

bassdr:
I still get the crash. I have to disable audio completely as a workaround, as forge will exceed my pipewire entry count 
Not sure if it matters, but I disable background music and leave only sound effects.

All in all, this pr did not fix it. And latest master still has the bug (will crash after 3 games for sure on master).

I also tried both audio pipelines in the settings. The alternate still leaks entries, but a lot less.

bassdr:
Last version was tested on linux, it works. I now see only one audio stream from forge in pipewire.

forge-gui-desktop/src/main/java/forge/sound/AudioClip.java
```diff
@@ -49,6 +55,55 @@
...
+     * The JDK's software mixer: it mixes clips itself and opens a single device line, where
+     * AudioSystem.getLine returns one DirectAudioDevice line per clip. It is not registered as a
+     * MixerProvider, so it has to be constructed directly, and is null without
+     * --add-exports java.desktop/com.sun.media.sound=ALL-UNNAMED.
+     */
+    private static final Mixer softMixer = openSoftMixer();
+
+    private static Mixer openSoftMixer() {
+        try {
+            Mixer mixer = (Mixer) Class.forName("com.sun.media.sound.SoftMixingMixer")
```
tool4ever, line 68: I don't think using an internal Java class is a great idea, surely the way you've phrased this problem is common enough that there's a less hacky approach available...

bassdr: Yep, found one, on it, it's a small refacto though.

PR #11805, opened 2026-09-07
Refactor: CantGainControl and LethalDamageByPower off hidden keywords (#3307)

calaespi:
## Summary
- Replace `AddHiddenKeyword` for control protection and lethal-by-power with proper static ability handling (`StaticAbilityCantGainControl` + Continuous `LethalDamageByPower$`).
- Update Guardian Beast and Zilortha scripts; remove keyword checks from `ControlGainEffect` / `Card.canBeControlledBy` / `Card.getLethal`.
- Narrower follow-up to #9703 / issue #3307: Lure/`MustBeBlockedByAll`, BounceAtUntap, and unrelated card/test churn are intentionally out of scope for a smaller mergeable step.

## Test plan
- [x] `mvn -pl forge-game -am compile -DskipTests` (checkstyle clean)
- [x] `mvn -pl forge-gui-desktop -am test -Dtest=HiddenKeywordsRefactorTest -Dsurefire.failIfNoSpecifiedTests=false` (3/3 pass)
- [ ] Manual smoke: Guardian Beast untapped vs tapped control steal
- [ ] Manual smoke: Zilortha lethal damage uses power

Replaces the broader stale PR #9703.

Made with [Cursor](https://cursor.com)


forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -6870,7 +6871,13 @@ public final boolean canBeTargetedBy(final SpellAbility sa) {
     }
 
     public final boolean canBeControlledBy(final Player newController) {
-        return newController.isInGame() && !(hasKeyword("Other players can't gain control of CARDNAME.") && !getController().equals(newController));
+        if (!newController.isInGame()) {
+            return false;
+        }
+        if (getController().equals(newController)) {
+            return true;
```
tool4ever: why was this logic inverted?

forge-gui/res/cardsfolder/z/zilortha_strength_incarnate.txt
```diff
@@ -3,5 +3,5 @@ ManaCost:3 R G
 Types:Legendary Creature Dinosaur
 PT:7/3
 K:Trample
-S:Mode$ Continuous | Affected$ Creature.YouCtrl | AddHiddenKeyword$ Lethal damage dealt to CARDNAME is determined by its power rather than its toughness. | Description$ Lethal damage dealt to creatures you control is determined by their power rather than their toughness.
+S:Mode$ Continuous | Affected$ Creature.YouCtrl | LethalDamageByPower$ True | Description$ Lethal damage dealt to creatures you control is determined by their power rather than their toughness.
```
tool4ever, line 6: `Card.updateKeywords()` has a comment for this, needs to be investigated

forge-game/src/main/java/forge/game/card/Card.java
calaespi: Good catch — it wasn't meant to change behavior. I had rewritten the De Morgan form into early returns, which made the `equals` check look inverted.

Restored the original shape with a 1:1 swap:

`return newController.isInGame() && !(StaticAbilityCantGainControl.cantGainControl(this) && !getController().equals(newController));`

forge-gui/res/cardsfolder/z/zilortha_strength_incarnate.txt
calaespi: Investigated: `updateKeywords()` called `updateLethalDamage()` because Zilortha previously granted a hidden keyword, so keyword refresh kept the UI lethal value in sync.

With `LethalDamageByPower$` on Continuous (RULES layer), that path no longer runs. Fixed by refreshing the view inside `addLethalDamageByPower` / `removeLethalDamageByPower`, and clarified the remaining `updateKeywords()` comment (keywords/P/T can still affect displayed lethal).

Hanmac:
the `doLoseControl` corner cases is probably something i need to ask Judges again.

like what happens when temporal control gain effects expire, while they are affected by it :thinking:

tool4ever (approved):
@Hanmac 
it seems a bit weird but on Arena *Zidane, Tantalus Thief* does make you a treasure when the control of his first trigger ends
based on that it should probably apply too

Hanmac:
@tool4ever I mean 
You take control of something until EoT
Then you would loose control, but you don't.

This kinda messes with the "Timestamp->Player" Map.

It probably needs to make a new Timestamp, so it doesn't collide with the layers 🤔

tool4ever:
ugh yea, notably it is already asymmetrical:
if I steal something with one of the real Layer 2 implementations, e.g. `Confiscate` and then destroy the aura it will currently move back to the previous controller :/

I think the Map entry probably always needs to be removed when the duration ends, so maybe the check belongs inside `controllerChangeZoneCorrection` instead...?
However this means `getController()` would have to return the zone owner where the card currently sits (if there is one).

Either way seems complicated enough that I'd also like a clear judge answer first, but since this PR only helps to clean up the hidden KW I'll just merge for now until an eventual revisit

PR #11809, opened 2026-09-07
AI: count "any color" and "combo" sources as producing colors

liamiak:
`ComputerUtilCost.getAvailableManaColors` collected the raw `Produced$` script string. Every caller runs that through `ColorSet.fromNames`, which resolves colour names only - and `"Any"` and `"Combo W U"` are neither, so both resolved to nothing:

| `Produced$` | cards | counted before |
|---|---|---|
| `Any` | 486 | no |
| `Combo ...` | 530 | no |
| bare `W`/`U`/`B`/`R`/`G`/`C` | 1324 | yes |

Shocklands, filter lands, triomes and Birds of Paradise are all in the first two rows, so four AI decisions saw a board of them as producing no colours - none of them about fetching land: `ComputerUtil:417` (Dream Halls marks every coloured spell "wrong colour"), `ManaAi:196`, `TapAi:106`, `SpecialCardAi:1428`.

The fix asks the card what it can produce rather than what its script says. `canProduceSameManaTypeWith` already walked the mana abilities through `CardUtil.canProduce`; that loop moves to `Card.getProducibleColors` so both callers share it. Resolving the colours also keeps `Colorless` distinct from "no colours", which the raw strings could not - and which cannot be preserved downstream, since `MagicColor.COLORLESS` is `0` and a `ColorSet` has no room for it.

**Kept out of the engine.** The walk is eight lines duplicated from the loop inside `canProduceSameManaTypeWith`, rather than a shared helper on `Card`. That is a real cost and worth naming, but it keeps this an AI-side fix with **zero engine diff**: `canProduceSameManaTypeWith` is untouched, no activating player is filled in anywhere, and `getComboColors` still declines its `NotedColors` branch - so **Paliano, the High City** reads exactly as it does on master. Paliano is the only card reaching that branch, so nothing changes on the rules side at all.

**Cost.** Per scan, 12-land board, best of 5x200 across three JVMs: 14-15us before, 20-22us after. Three callers asked this once per candidate for an answer that cannot change inside their loop, so they now ask once, and skip it entirely when there is no candidate to test - never more scans than before.

**Tests.** `AnyColorManaSourceTest`: an any-colour source must offer white for Swords to Plowshares; a `{C}` source must report colourless and no colour, while an any-colour source must report every colour and *not* colourless; Plains must offer white where Island must not. The first two fail on master, the third passes either way.

`mvn -pl forge-gui-desktop -am test`: 453 tests, 0 failures, 6 skipped. Checkstyle clean.

Split out of #11504, which needs this and is otherwise a heuristic change.

https://claude.ai/code/session_01NAz6n9nxxxN2K7gYzqUaU5


forge-ai/src/main/java/forge/ai/ability/TapAi.java
```diff
@@ -96,14 +96,17 @@ public boolean willPayUnlessCost(Player payer, SpellAbility sa, Cost cost, boole
                     Integer amount = lifeCost.convertAmount();
                     if (payer.getLife() > (amount + 1) && payer.canPayLife(amount, true, sa)) {
                         final int landsize = payer.getLandsInPlay().size() + 1;
+                        // the same sources answer this for every card in hand, so ask once
+                        final byte availableColors = ColorSet.fromNames(
```
tool4ever, line 107: should check if hand is empty first

forge-game/src/main/java/forge/game/card/Card.java
```diff
@@ -3350,19 +3350,40 @@ public final boolean canProduceColorMana(final Set<String> colors) {
...
     }
 
+    /** Every color this card could produce, walking its mana abilities once. */
+    public final Set<String> getProducibleColors() {
+        Set<String> colors = new HashSet<>();
+        for (final SpellAbility ab : getManaAbilities()) {
+            // canProduce is far more expensive with no activating player set, so fill it in for
+            // the duration of the check and put it back afterwards - this is a read, and
+            // setActivatingPlayer trickles down to sub-abilities
+            final Player fillIn = ab.getActivatingPlayer() == null ? getController() : null;
```
tool4ever: how is it more expensive?

forge-ai/src/main/java/forge/ai/ability/TapAi.java
liamiak: Good catch. The CMC filter below means even a non-empty hand can skip every iteration, so an empty-hand check on its own would still leave the wasted scan - filtering the hand to the matching CMC first covers both:

```java
final CardCollection playable = CardLists.filter(
        payer.getCardsIn(ZoneType.Hand), CardPredicates.hasCMC(landsize));
if (playable.isEmpty()) {
    return false;
}
```

The same hoist in `ManaAi` and `SpecialCardAi` had the same hole, so both now materialise the candidate list and skip the scan when it is empty. Pushed - never more scans than master in any case now.

forge-game/src/main/java/forge/game/card/Card.java
liamiak: Fair - "more expensive" doesn't say why, and it isn't the main reason anyway.

The real one is that the answer is wrong without a player: `getComboColors` gives up on its `NotedColors` branch, so the source reports no colours at all. One card uses it - Paliano, the High City - but now that `getAvailableManaColors` routes through this method, that wrong answer would reach four AI decisions rather than just the `canProduceSameManaTypeWith` card-script property.

The cost point is secondary and I should have named the mechanism: `metConditions()` is reached for every mana ability (`conditions` is never null), and `SpellAbilityCondition.areMet` falls back to the controller *and* logs a line on every check. That already happens on master - `canProduceSameManaTypeWith(Steam Vents)` prints 14 lines, Birds of Paradise 8 - so filling the player in silences an existing warning per check rather than avoiding a new one.

So it is six lines of save/restore for one card read correctly, plus the log noise going quiet. Comment reworded to say that. If you would rather not carry the save/restore, I will drop it and take the Paliano regression back.

Hanmac:
While touching ManaColors and Produced, see this ticket too: #10750

I wanted to replace ManaProduced with a Multiset:
> "B B B", should be refactored into a map `{B => 3}`

But that is probably still too much work to do.

forge-ai/src/main/java/forge/ai/ability/TapAi.java
```diff
@@ -96,14 +97,19 @@ public boolean willPayUnlessCost(Player payer, SpellAbility sa, Cost cost, boole
...
+                            return false;
+                        }
+                        // the same sources answer this for every card in hand, so ask once
+                        final byte availableColors = ColorSet.fromNames(
+                                ComputerUtilCost.getAvailableManaColors(payer, source)).getColor();
+                        for (Card c : playable) {
                             // Check if the AI intends to play the card and if it can pay for it with the mana it has
                             boolean willPlay = ComputerUtil.hasReasonToPlayCardThisTurn(payer, c);
-                            boolean canPay = c.getManaCost().canBePaidWithAvailable(ColorSet.fromNames(ComputerUtilCost.getAvailableManaColors(payer, source)).getColor());
+                            boolean canPay = c.getManaCost().canBePaidWithAvailable(availableColors);
```
tool4ever: this is still terribly written, evaluating both expressions when one could already short-circuit

forge-game/src/main/java/forge/game/card/Card.java
tool4ever: well the reality is not every single engine method is expected to be tested in isolation and if we start down this road everything will end up messy

no error gets logged ingame actually and AI fixes do not belong in the engine
so just remove it please

liamiak: Removed, and taken further - `getProducibleColors` is out of the engine entirely. `canProduceSameManaTypeWith` is untouched and the walk now sits in `ComputerUtilCost`, so the PR has zero engine diff. That is eight lines duplicated from the engine loop, which I would rather name than hide, but it keeps the AI fix on the AI side, which is what you are asking for.

On the warning: you are right, and I checked how I got it wrong. The 14 lines I quoted are real in `AITest`, but `getAllPossibleAbilities` sets the activating player persistently and every priority pass calls it - so in a real game it is 0. I was measuring the harness and describing the product.

forge-ai/src/main/java/forge/ai/ability/TapAi.java
liamiak: Fixed - `canBePaidWithAvailable` is the bitmask test, so it gates the call to `hasReasonToPlayCardThisTurn` now rather than both being computed up front.

liamiak:
Thanks @Hanmac - that ticket is more relevant than I first read it. This PR's body documents "colourless cannot survive `ColorSet`, since `MagicColor.COLORLESS` is `0`" as a hard constraint, and your Byte -> `MagicColor` note is exactly the fix for it; `AnyColorManaSourceTest.colorlessIsReportedButIsNotAColor` pins the current behaviour, so it is either evidence for that change or a test to update when it lands. Nothing here should get in its way.

If a first pass on the event-coalescing part (#10812) would help I am happy to take it, but you mentioned wanting to do the refactor yourself - say the word either way.

forge-game/src/main/java/forge/game/card/Card.java
tool4ever: nah you didn't have to move the whole method since you were able to use it to reduce engine code too

liamiak: Put it back as you said - `getProducibleColors` is in `Card` again and `canProduceSameManaTypeWith` calls it, so there's one copy of the walk instead of the two I'd left.

On the save/restore: I measured instead of assuming this time. Across 20 AI games the walk visited 491,691 mana abilities and none had a null activating player, 6,810 of them combo-mana - `getAllPossibleAbilities` sets it on the same SA objects every priority pass. So Paliano loses nothing. That's the precondition rather than Paliano end-to-end; it's draft-only and its branch also bails without draft notes.

liamiak:
One thing the last round turned up that I'd missed: routing the AI's colour survey through the engine walk made it resolve `ManaReflected` sources, and that is expensive - a battlefield `getValidCards` scan per ability, a fresh `FCollection` from `getSpellAbilities()` per match, then recursion into other reflecting permanents. Timed both arms in one JVM over identical input: 11.0 us before, 2.70 ms after on a board with 4 Reflecting Pool + 4 Exotic Orchard. Ablating just that branch puts it at 27.8 us.

So the AI survey now skips them. That matches master, which got nothing from those cards anyway - `origProduced` defaults to `"1"` for `AB$ ManaReflected` and `ColorSet.fromNames` drops it. The engine keeps the full walk.

Also rebased onto current master while I was in here.

forge-ai/src/main/java/forge/ai/ability/ManaAi.java
```diff
@@ -190,10 +190,14 @@ public static boolean doManaRitualLogic(Player ai, SpellAbility sa, boolean from
         CardCollection cardList = new CardCollection();
         // TODO check other zones
         List<SpellAbility> all = ComputerUtilAbility.getSpellAbilities(ai.getCardsIn(ZoneType.Hand), ai);
-        for (final SpellAbility testSa : ComputerUtilAbility.getOriginalAndAltCostAbilities(all, ai)) {
+        final List<SpellAbility> testable = ComputerUtilAbility.getOriginalAndAltCostAbilities(all, ai);
```
tool4ever, line 193: if you're worried about reflection performance I'd much rather you drop the additional `getOriginalAndAltCostAbilities` transformation
it adds very little value compared to a method named `getAvailableManaColors` ignoring your options from reflected sources

liamiak: You're right - reverted. The survey resolves reflected sources again and the `includeReflected` parameter is gone.

I'd measured the skip's speed and never its accuracy, which was the actual question. On a board with four Reflecting Pool and four Exotic Orchard it dropped a colour in 30 of 3,773 surveys, and the colour mask differed every time - so the AI could decline a spell it could pay for. Not worth microseconds.

Being upfront about what that costs: on that same board the survey goes from ~11us to ~3.3ms, roughly 3s per game of extra AI time. That's a ceiling from a deliberately hostile deck with eight reflecting permanents at once. On a board with none it's 8us against 24us, and 1,196 of the 14,044 shipped decks contain any of the 47 `ManaReflected` cards at all.

Profiling where that time goes was more interesting than I expected - it isn't the reflected walk. 94% is `CardLists.getValidCards`, at ~8.3us per card to decide `Land.YouCtrl`; `getSpellAbilities` and `getTriggers` together are 3%. Over four games the reflected path asked **four** distinct valid-questions - `{Land.YouCtrl, Land.OppCtrl}` x two players - and asked them 44,297 times.

So if this cost is worth attacking, `isValid`/`getValidCards` looks like the place, and it would pay off for every caller rather than just this path. Happy to open that separately.

One cheap thing available meanwhile: a reflected source can only contribute the colours in its `ColorOrType` range, so when the other sources already cover that range it can be skipped without changing the answer - 47% of them, 29% of the time, zero divergence over 3,773 surveys. I've left it out to keep this PR narrow; say the word if you'd like it in.

On `getOriginalAndAltCostAbilities`: it's already on master and sits in the per-ability loop rather than in this one call, so dropping it wouldn't offset the above - but I can look at it on its own merits if you like.

tool4ever: right, I still have to look at rituals as another candidate for my #11673 framework anyway

tool4ever (approved):

PR #11814, opened 2026-09-08
Fix translated ability text landing on the wrong ability

### LINKED ISSUE #11795: Freed from the Real cannot untap
Describe the bug
"Freed from the Real" does not correctly untap the enchanted creature.

When "Freed from the Real" is attached to "Zaxara, the Exemplary", activating the ability that should untap the enchanted creature instead resolves as “Tap Zaxara, the Exemplary”.

Because of this, the normal interaction between "Zaxara, the Exemplary" and "Freed from the Real" cannot be used.

Relevant game log:

"Resolve Stack: Libero dalla Realtà (100) - Tap Zaxara, l'Esemplare (101)."

"Add To Stack: Torro ha attivato Freed from the Real"

"Resolve Stack: Libero dalla Realtà (100) - Tap Zaxara, l'Esemplare (101)."

"Add To Stack: Torro ha attivato Freed from the Real"

---

To Reproduce

1. Put "Zaxara, the Exemplary" onto the battlefield.
2. Cast "Freed from the Real" targeting Zaxara.
3. Tap Zaxara to activate its mana ability.
4. Activate the "{U}: Untap enchanted creature" ability of "Freed from the Real".
5. Instead of untapping Zaxara, Forge resolves the ability as "Tap Zaxara, the Exemplary".

The issue can be reproduced repeatedly.

---

Expected behavior
Paying "{U}" to activate the untap ability of "Freed from the Real" should untap the enchanted creature.

With Zaxara enchanted, the expected sequence is:

1. Tap Zaxara to add two mana.
2. Pay "{U}" with "Freed from the Real".
3. Zaxara becomes untapped.
4. Zaxara can then be tapped again.

---

Screenshots
I can provide screenshots if needed.

The game log also clearly shows the incorrect resolution:

"Freed from the Real - Tap Zaxara, the Exemplary"

instead of untapping it.

---

Desktop (please complete the following information):

- OS: N/A
- Browser: N/A
- Version: N/A

---

Smartphone (please complete the following information):

- Device: Android smartphone
- OS: Android 16
- Version: 2.0.15

---

Additional context
This does not appear to be an issue with Zaxara itself or with untapping creatures in general.

"Thousand-Year Elixir" is able to correctly untap "Zaxara, the Exemplary" in Forge.

The problem appears to be specifically related to the untap ability of "Freed from the Real".

I tested this during a Commander game on Android.

liamiak:
Issue #11795: on a non-English client, Freed from the Real's `{U}: Untap enchanted creature` resolves as *Tap*. The engine is fine - both abilities are offered, and resolving the untap ability untaps. They are labelled wrongly: in it-IT the **Tap** ability reads `{U}: STAPpa la creatura incantata` ("untap"), so there is no way to pick the right one.

Three changes, each separately testable.

**1. Realign Aura oracles.** `buildOracleMapping` pairs English oracle lines with translated ones by index, and translated Aura oracles routinely arrive a line short, having dropped the leading `Enchant ...` line - so every later line wears its neighbour's translation. Mana symbols and stat changes survive translation, so where a line carries one they say which pairing is right; where none does, the leading line is the one usually missing. Over the shipped data (224,595 rows, eight languages) this realigns 1,439 rows and declines 4 whose symbols say the current pairing is already right.

**2. Give every English line an entry.** Where a translation is short for some other reason, every English line gets an entry and uncovered ones carry no text, so they fall back to English rather than matching a neighbour. Measured with the `SpellDescription`s the card scripts pass to the matcher: 11,747 checked, 84 change, all to English, and every one had been sharing its entry with another ability.

**3. The cache.** `translatedCaches` was keyed on ability text alone, so the first card to resolve a line fixed that answer for every card sharing it. Now keyed per card, falling back to a shared cache when a card has no mapping of its own. Both are `ConcurrentHashMap` - the path is reached from the game thread and the EDT.

No data files are changed; an earlier revision patched six oracle rows by hand and the realignment makes them redundant.

**Tests.** `CardTranslationTest`, one method per part, each failing with only the others applied. Full suite: 711 tests, 0 failures, 6 skipped. Checkstyle clean.

Known remaining: 419 rows are short by more than one line (Shiv's Embrace in es/it/pt arrives as one unsplit line), and Fiery Mantle it/pt have an empty translated oracle.

Fixes #11795

https://claude.ai/code/session_0177x5FTgdfzYB9BRDFzaN1Y


forge-core/src/main/java/forge/util/CardTranslation.java
```diff
@@ -345,19 +387,33 @@ public static String translateSingleDescriptionText(String descText, ITranslatab
         if (descText == null)
             return "";
         if (!needsTranslation()) return descText;
-        if (translatedCaches.containsKey(descText)) return translatedCaches.get(descText);
+        // the answer comes out of one card's oracle mapping, so it can only be reused for that
+        // card - two cards sharing an ability need not word its translation the same way
+        Map <String, String> cardCache =
+                translatedCaches.computeIfAbsent(card.getTranslationKey(), k -> new HashMap<>());
+        if (cardCache.containsKey(descText)) return cardCache.get(descText);
```
tool4ever, line 406: it probably makes sense to try and always get the correct wording if it's available for consistency

but you shouldn't break the possibility of obtaining it from a different card if it's somehow in the cache already
(which matters a lot when the ability moves cards from clone effects with name change or something)

forge-core/src/main/java/forge/util/CardTranslation.java
```diff
@@ -312,9 +312,51 @@ public static void buildOracleMapping(String faceName, String oracleText, String
...
+     * - a card whose abilities read identically cannot be played. Drop any entry that more than
+     * one oracle line lays claim to; those lines fall back to English, which is at least unique.
+     */
+    private static void dropAmbiguousEntries(List <Pair <String, String> > mapping,
+            String [] splitOracleText, String faceName) {
+        if (mapping.size() >= splitOracleText.length) {
+            return; // every line has its own entry, so no two can share one
+        }
+        List <String> oracleLines = new ArrayList<>();
+        for (String line : splitOracleText) {
```
tool4ever: isn't this mostly included in mapping already, is doing the rest the best way?

forge-core/src/main/java/forge/util/CardTranslation.java
```diff
@@ -312,9 +312,51 @@ public static void buildOracleMapping(String faceName, String oracleText, String
...
+    /**
+     * The lines above are paired by position, so a translated oracle with fewer lines than the
+     * English one pairs each line with its neighbour's translation. Ability text is then matched
+     * back by edit distance, which lets two different abilities land on the same translated line
+     * - a card whose abilities read identically cannot be played. Drop any entry that more than
+     * one oracle line lays claim to; those lines fall back to English, which is at least unique.
+     */
+    private static void dropAmbiguousEntries(List <Pair <String, String> > mapping,
+            String [] splitOracleText, String faceName) {
+        if (mapping.size() >= splitOracleText.length) {
```
tool4ever: Considered different strategy where we expand the mapping size to maximum of either oracle linebreaks?

that would allow you to english fallback for the actually missing lines?

as long as `getLevenshteinDistance` can tell them apart but this would also mean calling it only on load is usually enough instead of ingame too?

might increase memory from additional strings stored then though? 🤔

liamiak: Took it. Every English line gets an entry; the ones the translation does not cover carry no text and fall back to English at lookup. That deletes `dropAmbiguousEntries` outright - which answers your other question too - net -26 lines on the file.

Measured over all eight languages with the `SpellDescription`s from the card scripts rather than oracle lines. That difference matters, and it is what I got wrong first time round: 11,747 descriptions, 84 change, all of them to English, and every one had been sharing its entry with another ability. Nothing that had sole claim on a translation loses it.

Shiv's Embrace is a good example of why the line-level view misleads. Its German row is missing the `Enchant creature` line, so both its abilities resolved to one entry and read identically - a collision all along, but querying with oracle lines does not show it. They now read differently; the `+2/+2` line still shows the wrong German, which is the data defect rather than the pairing.

On running the matcher at load only: I do not think it can move, and not for the reason I would have guessed. About 28% of scripted descriptions have no oracle line at all - Abomination's delayed-trigger text, sub-ability fragments, `{c:Targeted}` templates - and only 64% of faces resolve cleanly. It is not recovering a link that got dropped; the two sets genuinely do not correspond.

One thing you did not ask for: both caches are `ConcurrentHashMap` now. The path is reached from the game thread and the EDT, and `computeIfAbsent` on a plain `HashMap` is not safe under that. Happy to split it out if you would rather.

forge-core/src/main/java/forge/util/CardTranslation.java
liamiak: Moot now - adopting the padding above removed the method this was about, and the re-derivation with it.

forge-core/src/main/java/forge/util/CardTranslation.java
liamiak: Two-tier now. The per-card cache answers first; a card with no mapping of its own - the clone that took a different name - falls back to a shared cache, which only ever stores real translations. So the per-card answer stays authoritative without losing the cross-card reuse you asked me not to break.

liamiak:
Two corrections, both from the same find.

**The six data rows were in the wrong file.** They edited `res/languages/cardnames-*.txt`, which `tools/cardnamesTranslations.py` regenerates from Scryfall - there are "Update cardnames translations" commits doing exactly that. The durable place is `tools/cardnames-<lang>-patch.txt`, applied on top of the regenerated data; es-ES already carries 3,988 rows there. Pushed: the patch rows are added and the shipped-file edits kept, so it works before the next regeneration and survives it.

**The class is much bigger, and has an exact cause.** Aura rows missing their leading `Enchant ...` line - checked by content, since being one line short does not say *which* line went (Arrest keeps `Encantar criatura.` in Spanish and loses the effect line instead):

| de-DE | es-ES | fr-FR | it-IT | pt-BR | zh-CN | ko-KR | ja-JP |
|---|---|---|---|---|---|---|---|
| 337 | 310 | 153 | 144 | 104 | 41 | 14 | **0** |

1,103 rows over 411 cards. Those six are exactly the languages the generator covers. Old-border Auras printed `Enchant creature` as the type line, so Scryfall's `printed_text` never carries it - and ja-JP scores zero because `JapaneseTranslations.main()` is commented out, so that file is not built from this source at all.

So the smallest durable fix is probably in the generator - inject the line when the English oracle has one and `printed_text` does not - rather than 1,100 patch rows. I have a data-only test that detects the class (reads the data files, touches no engine) and could come along as a guard.

Happy to do any, all or none of that - it is your call, and I did not want to open a 1,100-row diff without asking.

forge-core/src/main/java/forge/util/CardTranslation.java
```diff
@@ -345,19 +355,42 @@ public static String translateSingleDescriptionText(String descText, ITranslatab
...
-        translatedCaches.put(descText, result);
+        cardCache.put(descText, result);
+        if (!result.equals(descText)) {
+            sharedCache.putIfAbsent(descText, result);
+        }
         return result;
     }
 
     private static String translateSingleIngameText(String descText, List <Pair <String, String> > mapping) {
+        int candidateIndex = matchingEntry(descText, mapping);
```
tool4ever: wouldn't it be better now to have an exit condition in the loop this calls the first time `getRight()` is null so it stops comparing?

forge-core/src/main/java/forge/util/CardTranslation.java
tool4ever: ok, more flexible idea:
if we end up having two matches with identical `distance` just do another pass for 100% to decide the tie (`equals()`)?

tool4ever:
I actually prefer if we can at least partly account for it with our Java logic so we can work with the original Scryfall data to some degree instead of only relying on a manual patching step 👍

forge-core/src/main/java/forge/util/CardTranslation.java
tool4ever: actually thinking about it some more this seems rather unlikely

forge-core/src/main/java/forge/util/CardTranslation.java
liamiak: Tried it, and it reintroduces the collision - with that exit in place Pemmin's Aura in de-DE goes back to showing two abilities as the same string.

The null entries are a suffix: a line gets one exactly when the translation ran out. Their only job is to absorb a match. Matching a null entry and matching nothing both return the English text, so the entry exists purely to stop an uncovered line matching a *translated* neighbour's entry - exit at the first null and the absorber is gone, so the line matches the neighbour again.

On the cost: `translateSingleDescriptionText` caches per card and text, so this loop runs once per distinct description per card, not per render.

liamiak:
Took the Java route rather than the patch rows - the six data rows are gone, and the pairing is realigned at load instead.

When a translated oracle is short by exactly one line and the English first line is `Enchant ...`, both pairings are scored by the parts that survive translation (`{U}`, `+2/+2`); positional is kept only if it wins outright. Over the shipped data - 224,595 rows, all eight languages - that realigns **1,439** rows and declines **4**: Favorable Destiny, Fiery Mantle (de, es) and Rancor it-IT, which are short a line for other reasons (reordered lines, or a retained Enchant line with two merged). Nothing outside that shape is touched.

Two corrections to what I told you earlier:

- The old-border type-line explanation accounts for about 8% of the class - 46 of the 593 rows the symbols can adjudicate. Most affected rows have a perfectly normal type line and lost the line for some other reason, so I gave you a cause that isn't the main driver.
- I said 1,103 rows; that was a content check. Counted as rows short by exactly one line it is 1,443, of which the rule moves 1,439.

I also checked whether a second signal could adjudicate the rows carrying no symbols - total-length ratios overlap completely between the two shapes, so there isn't one. Those rows take the shift, which is right in 593 of 597 cases where it can be checked.

Verified against the original report: with the realignment, it-IT reads `TAPpa` on the Tap ability and `STAPpa` on Untap, and resolving each does what it says. Rebased onto master.

tool4ever:
Sigh, I need to be careful with how I word my responses when an eager LLM is involved 😅

I mostly meant the original part of just avoiding wrong descriptions via english fallback

Oh well, now that the scope creep happened I might as well ask:
are there other related cases and can `leadingLineOffset` be generalized (or just future-proofed) a bit further without too much work?

liamiak:
Fair — the general version didn't survive the data, but it pointed at the right change.

**Other cases.** Of 220,383 card/language rows, 96% pair one-to-one; 4,811 are one line short, 1,658 shorter still, 2,435 have a *longer* translation.

And a correction: the lines aren't dropped so much as **folded into the neighbour** — de-DE reads `Fliegend; Echo` as one line — or, for Sagas and Classes, omitted outright (381 rows, 291 ko-KR). `Enchant` is only 505 of the 1,358 leading cases the symbols can place; the rest are Flying (147), Devoid (60), Trample (42).

**Generalising.** Letting the symbols place the line anywhere isn't worth having: of 20 hand-checked rows it changed, 11 improved and 9 regressed. Symbols disagree for reasons other than misalignment — fr-FR writes `{1}` for `{C}`.

The failures had a shape, so instead:

- the gate: `startsWith("Enchant ")` becomes that *or* a keyword line. Enchant stays explicit, since the word test rejects `Enchant creature with another Aura attached to it`.
- the veto: counting agreements only passes 216 rows the symbols contradict. Counting agreements less disagreements, ignoring translated lines carrying none of their own, brings that to 13.

Hand-checked on 60 rows across seven languages, sampled where the two rules disagree: 22/60 before, 52/60 after. Realigns 3,005 rows against 1,127 — though 1,649 move on the gate alone with no symbols to check them, proportionally what the rule already does (520 of 1,127).

`isKeywordLine` is a guess: no trailing period, five words or fewer once reminder text is stripped, since forge-core can't see `forge.game.keyword`.

New test: Opal-Eye, Konda's Yojimbo is no Aura - German folds its Defender line into Bushido, and without the change its `{T}` ability shows `{1}{W}`.

Rebased onto your inlining commit - say the word if you'd like `isKeywordLine` folded in the same way.

tool4ever (approved):

liamiak:
Took the point from `Remove obsolete noise` one commit further up: `pairingScore` took the index of the unmatched line but only ever received two values, left over from the any-position version I dropped. Collapsed to a single pass over the translated lines.

Verified by dumping every ability's displayed text for all one-line-short cards in de-DE, ko-KR and zh-CN on both builds - 2,557 rows over 1,880 cards, identical.

One clarification on the numbers above: the corpus counts come from running the rule over the data files, not through the engine. The engine-level evidence is the tests.

PR #11816, opened 2026-09-08
Fix auto-pass network crash and consolidate logic

MostCromulent:
## Summary

A remote client toggling auto-yield can crash the host with a `ConcurrentModificationException`. Investigating that crash turned up a consolidation worth taking on its own: it removes more code than it adds, and it leaves the whole yield path easier to reason about.

The crash happens because one click sends two messages, and they race each other on the host. Removing the race means deleting the client's `passPriority`, which leaves the host to work out for itself whether the toggle should skip the prompt. Four separate pieces of code each held part of that answer, so the consolidation is what makes the fix possible.

## Bug report

Four-player networked Commander game, host running on Android:

    21:51:29 [ERROR] InvocationTargetException in sendYieldUpdate args=[CardAutoYield]: java.util.ConcurrentModificationException
        at forge.game.StaticEffects.clearStaticEffects(StaticEffects.java:46)
        at forge.game.GameAction.checkStaticAbilities(GameAction.java:1088)
        at forge.game.GameActionUtil.getAlternativeCosts(GameActionUtil.java:285)
        at forge.player.PlayerControllerHuman.tryAutoPassNow(PlayerControllerHuman.java:3915)

The game thread logged spell resolution immediately after.

## Root cause

Review of #11718, the open PR that enforces single-thread engine access, asked why advanced yield can change game state when that player already holds priority.

It does test for priority. `applyYieldUpdate` ends by re-checking whether the open prompt should now auto-pass, and that re-check returns unless the player sits at an `InputPassPriority`. The test is only correct while that prompt stays on screen.

The same click also sends a separate `passPriority`, which the host runs on its own thread. That message releases the game thread while the re-check is still inside the availability scan, `AvailableActions.compute`, which runs static abilities over every card and mutates the shared game graph as it goes. Both threads then enter `checkStaticAbilities`.

So one click became two independent actions. Commit `0c56a561e03` introduced the pattern back in 2014, when both calls were local and no network boundary separated them.

## The old architecture

To delete the client's message, the host needs one method that can answer whether to skip the prompt. It skips a prompt for four independent reasons:

| | Reason the host skips the prompt |
|---|---|
| **A** | An explicit yield is active: until end of turn, until the stack clears, or until a phase marker |
| **B** | Auto-pass-with-no-actions is on and the player has no available action |
| **C** | The player set the current phase to skip, and the stack is empty |
| **D** | The top of the stack is an ability the player auto-yields |

No method answers all four in a form a caller can use. `chooseSpellAbilityToPlay` covers them, but as three sequential exits rather than a value it can hand out. Everything else knows **A** and **B** only:

| Caller | Tested for |
|---|---|
| `chooseSpellAbilityToPlay`, building a prompt | All four, as three sequential exits |
| `mayAutoPass()` | A, B |
| `tryAutoPassNow`, when a yield update arrives while a prompt is open | A, B |
| `autoPassCancel`, `InputLockUI`, `allowAwaitNextInput` | A, B |

Spreading the answer across those callers produced three defects.

**Reason C is tested twice, from sources that can disagree.** One test runs when the host builds the prompt. The other hides inside `isAutoPassingNoActions`, the method that otherwise answers reason **B**, so that `tryAutoPassNow` can see **C** at all. That second copy sits behind the auto-pass-with-no-actions preference, and it calls `isUiSetToSkipPhase`, which `RemoteClientGuiGame` always answers `false`. For a network client it therefore never fired.

**Reason D is not tested once a prompt is open.** That is why the client had to send the second message.

**The availability scan for reason B runs in two places.** `applyYieldUpdate` carries a three-way conditional whose only job is to keep some update types away from it.

Passing priority is overloaded too, independently of the four reasons. The code passes priority by clicking OK, but `onOk()` has a second meaning: it accepts a yield suggestion when the prompt shows one. So an internal caller could arm a yield the player never accepted.

## The new architecture

The four reasons are now gathered behind two named questions, each with one definition and one source. This is not tidying. The host cannot answer a toggle that arrives while a prompt is open unless one method knows every reason to skip that prompt. Nothing tested reason **D** (an auto-yielded ability on top of the stack) once a prompt was open.

| Question | Answered by | Asked by |
|---|---|---|
| Is the player auto-passing? | `mayAutoPass()`, reasons A and B | `autoPassCancel`, `InputLockUI`, `allowAwaitNextInput` |
| Should this prompt be skipped? | `mayAutoPass() \|\| skipsPromptForStackOrPhase()`, reasons A to D | `chooseSpellAbilityToPlay`, `tryAutoPassNow` |

`chooseSpellAbilityToPlay` now has one exit and one delay rule, and the availability scan has one caller.

Priority is now passed by calling `InputPassPriority.passPriority()`, so the result no longer depends on what the prompt shows. No client gesture sends a follow-up `passPriority`, so that protocol method has no sender left and this change deletes it.

The two questions keep separate names for a reason. The engine calls `autoPassCancel` for every player at every cleanup, and that call asks the first question, not the second.

## How this fixes the crash

The race is gone. One click sends one message, and the same host step both decides and answers, so nothing releases the game thread mid-decision.

The graph mutation is gone from the path. `tryAutoPassNow` no longer runs the availability scan, which only ever supplied reason **B** (no available action). That scan now runs when the auto-pass-with-no-actions preference changes, and nowhere else.

This does not make all engine access thread-safe. Two update types still reach the scan from a protocol thread through `refreshAvailableActionsForPrompt`: the client state seed, and a change to the auto-pass-with-no-actions preference. That is safe only because the player sits at a priority prompt, which parks the game thread in blocking input. Master reaches the same scan on a wider set of updates, so this narrows the exposure without closing it. #11718 is what turns that condition into an enforced invariant.

## Behaviour changes

All three follow from the consolidation. Once a reason is tested in one place, every caller sees it, including the callers that previously saw only part of the answer. They apply to single-player, hotseat and network games alike.

- **A skip-phase toggle now acts on the prompt in front of you** because reason **C**. Clicking the label for the phase you are stopped in will therefore end it.
- **`YIELD_SKIP_RESOLVE_DELAY` now controls the pause before an auto-yielded ability resolves.** Reason **D** had its own exit with an unconditional sleep. Folding the three exits into one leaves a single delay rule covering all four reasons.
- **Re-enabling auto-yields with an auto-yielded ability on top of the stack now passes priority.** `tryAutoPassNow` can see reason **D** for the first time. Previously nothing passed priority there.

## Also fixed

- A skip-phase toggle at its own prompt did nothing for a network player, and worked for anyone else only when auto-pass-with-no-actions was on. Skip-phase and disable-yields toggles both reached the host but never a local controller, so the same gesture gave two different results depending on transport.
- Macro replay, the dev-mode win cheat and `tryAutoPassNow` could accept a yield suggestion instead of passing priority.
- The host could build a prompt from a stale "has available actions" value, because it skipped the scan for a yield that the same test then retired.

---


forge-gui/src/main/java/forge/player/PlayerControllerHuman.java
```diff
@@ -3899,28 +3888,40 @@ public void applyYieldUpdate(final YieldUpdate update) {
...
         tryAutoPassNow();
     }
 
     @Override
     public void setYieldPref(final FPref pref, final String value) {
-        // Dialog already wrote to FModel; APINA is the only pref whose toggle can flip mayAutoPass for a sitting prompt
-        if (pref == FPref.YIELD_AUTO_PASS_NO_ACTIONS) tryAutoPassNow();
+        // Dialog already wrote to FModel; APINA is the only pref whose toggle can flip mayAutoPass for a
+        // sitting prompt, and the budget field saves on every keystroke, so sweeping for the rest is waste
+        if (pref != FPref.YIELD_AUTO_PASS_NO_ACTIONS) return;
```
tool4ever: this no longer seems to pass if it could

also shouldn't we check that the preference is actually getting enabled?

MostCromulent: https://github.com/Card-Forge/forge/pull/11816/commits/2f67e9dde0b40bdddd30b90006009e44372e7bc2 addresses your second point.

can you explain your first point a bit more?

tool4ever: hmn I can no longer reproduce it, now it always passes if I activate APINA

tool4ever (approved):

PR #11740, opened 2026-08-31
Implement Rulebreaker (+Partial implementation of DeckRule)

BigCrunch22:
<html>
<body>
<html><head></head><body><h1>Rulebreaker mechanic + partial <code>DeckRule</code> deckbuilding framework</h1>
<h2>Summary</h2>
<p>This adds a new custom keyword, <strong>Rulebreaker</strong>, and the engine infrastructure needed to support it: a partial implementation of the <code>DeckRule:</code> line-type framework sketched in <a href="https://github.com/Card-Forge/forge/issues/8066">#8066</a> ("Rework deckbuilding rules into a set of objects"). Rulebreaker cards bend specific deckbuilding restrictions (color identity, deck size) when they're your commander. All 12 cards in the Rulebreaker cycle are implemented and scripted, and Adventure Mode's separate, pre-existing deck-legality enforcement has been checked against them and patched where it wasn't <code>DeckRule</code>-aware.</p>
<h2>Background</h2>
<p>Rulebreaker cards read like:</p>
<blockquote>
<p><strong>Rulebreaker</strong> — A deck with this commander can have artifact creature and Equipment cards of any color identity and any basic land cards.</p>
</blockquote>
<p>This is a deckbuilding-legality effect, not a live-game one — it changes which cards are legal to put in the deck, not anything that happens once a game starts.</p>
<p><a href="https://github.com/Card-Forge/forge/issues/8066">#8066</a> already proposes the right shape for this: a new top-level script line, <code>DeckRule:</code>, parsed independently of <code>K:</code> lines, with rule classes like <code>ColorIdentity</code> and <code>Size</code>. This PR implements that framework — scoped to what Rulebreaker actually needs (<code>ColorIdentity</code> and <code>Size</code>), not the issue's full vision (<code>Copies</code>, <code>Commander</code>, <code>FormatPool</code>, <code>Variants</code>, <code>FormatRestrictions</code> are not attempted here).</p>
<h2>New: the <code>DeckRule</code> framework (<code>forge-core</code>)</h2>
<p>Three new classes in <code>forge.deck</code>, plus a new shared restriction-matching helper in <code>forge.card</code>:</p>
<ul>
[...]

Issue #8066, opened 2025-07-15
Rework deckbuilding rules into a set of objects


Jetz72:
The logic for if a deck or card is legal is cumbersome to apply with flexibility. Most of it is wrapped up in huge methods that test many rules at once and return a single string describing the first violation, or null if there isn't one.
I suggest we redefine deck rules as objects, with methods for testing themselves against a deck list. Each would need, at minimum:
* `boolean isValid(Deck)` - Checks a deck and returns false if the rule is violated.
* `String getDescription()` - Gives a user-facing string describing this rule generally.

But we could also expand their functionality by letting them be more specific, with methods such as:
* `List<String> getErrors(Deck)` - Returns a list of user-facing error messages describing each violation of this rule. Might come with a helper method `getError` that returns them all concatenated together.
* `Map<DeckSection, CardPool> getViolations(Deck)`  - Returns a list of cards that are in violation of this rule. Could be used to do things like highlighting problem cards in a deck editor. Some rules might be violated with this being empty, if a deck requires something that is not there, for instance.

This system would want to account all deckbuilding rules, across all supported formats. Rules would need to be general enough to be flexible but specific enough to be useful. We might use multiple classes, with configurations and exclusions depending on game mode:
* `DeckRuleSize` - A rule that specifies that the total number of cards in one or more sections must fall in a certain range (or be zero, if the section is optional).
* `DeckRuleCopies` - A rule that specifies the upper limit of copies of each card across the whole deck. Maintains an ordered map of predicates (e.g. isBasicLand, isPhenomenon) to quantities. Cards can be tested against these predicates and use the quantity of the first one that matches.
* `DeckRuleCommander` - A rule that specifies the default criteria for a commander, and possibly also a signature spell.
* `DeckRuleColorIdentity` - Rule that all cards in the main deck and sideboard must fall within the color identity of the cards in the command zone.
* `DeckRuleTypes` - Specifies that all cards in the deck or in a given section must or must not fall within a given list of types.
* `DeckRuleVariants` - Forbids nontraditional cards in the main, side, and commander sections.
* `DeckRuleFormatPool` - A rule that forbids cards outside the format's pool, as specified by a list of editions, edition types, and/or individual cards to allow or deny.
* `DeckRuleFormatRestrictions` - Works similar to the above, but is defined by a list of individual cards or predicates that are either banned or limited to a certain quantity.

[...]

PR #11817, opened 2026-09-08
Implement DeckRuleCopies + cache tokenized DeckRule lines per CardFace

BigCrunch22:
## Summary

This implements `DeckRule:Copies`, the per-card copy-count rule class sketched in [#8066](https://github.com/Card-Forge/forge/issues/8066), building on the `DeckRule:` framework and `DeckRuleColorIdentity`/`DeckRuleSize` groundwork from [#11740](https://github.com/Card-Forge/forge/pull/11740). It replaces the two ad hoc mechanisms that previously implemented per-card copy-count exceptions — the `K:DeckLimit:N:...` keyword and a raw, string-matched `K:A deck can have any number of cards named CARDNAME.` keyword — with a single `DeckRule:Copies:Limit$ <Unlimited|n> | Affected$ <branches>` line, and migrates all 15 cards that used either mechanism.

`DeckRule:` line parsing is also restructured into two steps: the expensive part (splitting a raw `DeckRule:` string into a rule class and its params) is now tokenized once per `CardFace` and cached there, shared across every printing of a card; only the cheap part (building the actual typed rule object) still happens per access.

## Background

[#8066](https://github.com/Card-Forge/forge/issues/8066) sketches `DeckRuleCopies` roughly like:

> ```
> # Seven Dwarves
> DeckRule:Copies:Limit$ 7 | Affected$ Name:CARDNAME | Description$ A deck can have up to seven cards named CARDNAME.
>
> # Relentless Rats
> DeckRule:Copies:Limit$ Unlimited | Affected$ Name:CARDNAME | Description$ A deck can have any number of cards named CARDNAME.
> ```

Before this PR, per-card copy limits were implemented two different, unrelated ways:

- **`K:DeckLimit:N:<description text>`** — a pseudo-keyword read via `CardRules.getKeywordMagnitude("DeckLimit")`, with its own special-cased text rendering in `Card.java`.
- **`K:A deck can have any number of cards named CARDNAME.`** — not a real keyword at all, just an unrecognized `K:` line whose literal text `DeckFormat.canHaveAnyNumberOf()` matched with `Iterables.contains(...)`.

Both are deckbuilding-legality effects, not live-game ones — exactly what #8066's `DeckRuleCopies` is meant to formalize, in the same spirit as `DeckRuleColorIdentity`/`DeckRuleSize` from #11740.

## New: `DeckRuleCopies` (`forge-core`)

One new class, `forge.deck.DeckRuleCopies`:

```
DeckRule:Copies:Limit$ <Unlimited|n> | Affected$ <branches>
```

- **`Limit$ <Unlimited|n>`** — the max copies of a matching card allowed in a deck.
- **`Affected$ <branches>`** — required; which cards the limit applies to. Reuses the same comma-separated branch grammar `DeckRuleColorIdentity`'s `Exempt$` uses (`CardRulesPredicates#restrictionBranch`, `public` so `DeckRuleCopies` can call it directly on a single already-isolated branch). That shared grammar has a `named<Name>` token form — e.g. `namedCARDNAME` — that matches on an exact card name via the existing `CardRulesPredicates.name(EQUALS, ...)` helper, as a plain, colon-free alternative to #8066's `Name:CARDNAME` sketch above.
  * `DeckRuleCopies` substitutes the literal token `CARDNAME` with the rule's own bearing face's real name *before* handing the branch to `restrictionBranch()`, and does that substitution per-branch, after `Affected$`'s value has already been split on commas — so a comma inside the resolved name (e.g. "Vazal, the Compleat") can never be mistaken for a second branch.
- Each migrated card keeps its own oracle-accurate `Description$` rather than generic template text — e.g. Gleemox keeps `This card is banned.` and Once More with Feeling keeps its `DCI ruling —` phrasing.

## Parsing & caching: tokenize once per `CardFace`, build once per access

`DeckRule:` lines used to be re-split from raw strings (colon, pipe, dollar-sign parsing) on every single call to `DeckRule.parseAll`, and that result was then cached once per `PaperCard`. Two problems with that: the cache lived on the wrong object (different printings of the same card are different `PaperCard` instances, so each one redid the same tokenizing work from scratch the first time it was touched), and it only ever looked at the card's main face — a `DeckRule:` line on a card's other face was silently ignored.

This restructures it into two layers:

- **`forge.card.DeckRuleLine`** (new) — a plain, immutable `(ruleClass, params)` pair: one tokenized `DeckRule:` line, not yet built into a typed rule object.
- **`ICardRawAbilites`** gains `List<DeckRuleLine> getTokenizedDeckRules()`.
- **`CardFace`** implements it: splits each raw `DeckRule:` string into a `DeckRuleLine` and caches the resulting list in a `private transient` field, computed once and shared across every `PaperCard` printing built from that face. The `Key$ Value | Key2$ Value2` param-splitting logic that used to live in `DeckRule.parseParams` moved here verbatim (renamed `parseDeckRuleParams`).
- **`DeckRule.parseAll(IPaperCard)`** now loops over `card.getAllFaces()` instead of only `card.getRules().getDeckRules()` (the main face), and for each face reads its own cached `getTokenizedDeckRules()` before building typed rule objects (`DeckRuleColorIdentity`, `DeckRuleSize`, `DeckRuleCopies`) from them. Each face's own name is threaded into `DeckRuleCopies` for its own `CARDNAME` resolution, rather than the card's overall name — so a `DeckRule:Copies` line on a specific face of a multi-faced card resolves `CARDNAME` to that face's name, not the combined card name. This part — building the typed objects — is deliberately *not* cached, since it's cheap once the tokenizing is already done, and still needs to fold in per-`PaperCard` data (a commander's marked colors, for `AllowedAdditionalColor$`) that can't be shared across printings.
- **`PaperCard.getDeckRuleList()`** drops its own `private transient List<DeckRule> deckRuleList` cache field entirely and just calls `DeckRule.parseAll(this)` directly on every access — the expensive work it used to exist to avoid now lives (and is shared) one level down, on `CardFace`.
- **`PaperToken.getDeckRuleList()`** is unchanged — still an uncached passthrough to `DeckRule.parseAll(this)`, since tokens are never a deck's own card-pool entry.

## Other engine plumbing (from the earlier round, unchanged here)

**`forge-core`**
- `DeckFormat.java` — `canHaveAnyNumberOf()` / `canHaveSpecificNumberInDeck()` go through a private `getOwnCopiesLimit(IPaperCard)` helper that scans `getDeckRuleList()` for a `DeckRuleCopies` entry. `getMaxCardCopies(PaperCard)` and `getDeckConformanceProblem()` call `getDeckRuleList()` instead of `DeckRule.parseAll(...)` directly.
- `CardRules.java` — `getKeywordMagnitude(String)` was removed; it only ever existed to read `DeckLimit`'s numeric argument.

**`forge-game`**
- `Card.java` — the two `DeckLimit`-specific text branches (creature/permanent ability text, and `abilityTextInstantSorcery()`) are gone. The creature/permanent path already folds each active `DeckRule`'s `Description$` into the card's live text (from #11740); `abilityTextInstantSorcery()` gained that rendering here, since two migrated cards (Once More with Feeling, Dragon's Approach) are sorceries.

**Call sites on the cache**
`ACEditorBase.java` (desktop deck editor), `FDeckEditor.java` (mobile deck editor), and `DuelScene.java` (Adventure mode) all call `card.getDeckRuleList()` rather than `DeckRule.parseAll(card)` directly, for their `DeckRuleColorIdentity`/`DeckRuleSize` handling. None of them touch `DeckRuleCopies` — Copies is only ever reached through `DeckFormat.canHaveAnyNumberOf()` / `canHaveSpecificNumberInDeck()` / `getMaxCardCopies()`, so no Copies-specific UI change is needed anywhere.

## Cards migrated

All 15 cards that used either old mechanism, all using `Affected$ namedCARDNAME`:

| Card | Old mechanism | New `Limit$` |
| --- | --- | --- |
| Relentless Rats | `K:A deck can have any number...` | `Unlimited` |
| Rat Colony | `K:A deck can have any number...` | `Unlimited` |
| Templar Knight | `K:A deck can have any number...` | `Unlimited` |
| Tempest Hawk | `K:A deck can have any number...` | `Unlimited` |
| Hare Apparent | `K:A deck can have any number...` | `Unlimited` |
| Persistent Petitioners | `K:A deck can have any number...` | `Unlimited` |
| Cid, Timeless Artificer | `K:A deck can have any number...` | `Unlimited` |
| Slime Against Humanity | `K:A deck can have any number...` | `Unlimited` |
| Shadowborn Apostle | `K:A deck can have any number...` | `Unlimited` |
| Dragon's Approach | `K:A deck can have any number...` | `Unlimited` |
| Once More with Feeling | `K:DeckLimit:1:...` | `1` |
| Vazal, the Compleat | `K:DeckLimit:1:...` | `1` |
| Nazgûl | `K:DeckLimit:9:...` | `9` |
| Seven Dwarves | `K:DeckLimit:7:...` | `7` |
| Gleemox | `K:DeckLimit:0:...` (ban) | `0` |

## Known limitations

- Gleemox wasn't a named target of this migration, but it used the exact same `K:DeckLimit` mechanism being removed (`DeckLimit:0` — its ban is implemented as a 0-copy limit); it's migrated to `Limit$ 0` here to avoid silently un-banning it. A dedicated `DeckRule:FormatRestrictions` (per #8066) would be the more semantically correct home for this later.
- `Affected$` has no "omit it to mean self" shorthand — every one of the 15 migrated cards has to spell out `Affected$ namedCARDNAME` in full. A future revision could default `Affected$` to the bearing face's own name when omitted, closer to #8066's terser sketch.
- Building the typed `DeckRule` list is no longer cached per `PaperCard` — a hot path calling `getDeckRuleList()` many times against the same instance will re-run the (cheap) object construction each time, though not the (expensive) string tokenizing.
- `DeckRule:Commander`, `DeckRule:FormatPool`, `DeckRule:Variants`, `DeckRule:FormatRestrictions` from #8066's fuller vision are still not implemented — only `Copies`, joining `ColorIdentity`/`Size` from #11740.
- Exhaustively re-checked `cardsfolder` for any other per-card copy-count mechanism beyond `K:DeckLimit`/the raw keyword (broad sweeps for "any number of", "up to N cards named", "Megalegendary", "banned", and every `K:` keyword verb containing "Deck/Limit/Copy") — no others found. A handful of similarly-worded tutor-effect cards (Squadron Hawk, Battalion Foot Soldier, Legion Conquistador, Gathering Throng, Howling Wolf, Nesting Wurm, Skyshroud Sentinel) were checked directly and confirmed unrelated.

***Code changes made with heavy assistance from Claude***

### DIFF
--- a/forge-gui/res/cardsfolder/c/cid_timeless_artificer.txt
+++ b/forge-gui/res/cardsfolder/c/cid_timeless_artificer.txt
@@ -3,7 +3,7 @@ ManaCost:2 W U
 Types:Legendary Creature Human Artificer
 PT:4/4
 K:Cycling:W U
-K:A deck can have any number of cards named CARDNAME.
+DeckRule:Copies:Limit$ Unlimited | Affected$ Name:CARDNAME | Description$ A deck can have any number of cards named CARDNAME.
 S:Mode$ Continuous | Affected$ Creature.Artifact+YouCtrl,Creature.Hero+YouCtrl | AddPower$ X | AddToughness$ X | Description$ Artifact creatures and Heroes you control get +1/+1 for each Artificer you control and each Artificer card in your graveyard.
 SVar:X:Count$ValidBattlefield,Graveyard Artificer.YouCtrl
 DeckHas:Ability$Discard
--- a/forge-gui/res/cardsfolder/c/cid_timeless_artificer.txt
+++ b/forge-gui/res/cardsfolder/c/cid_timeless_artificer.txt
@@ -2,8 +2,8 @@ Name:Cid, Timeless Artificer
 ManaCost:2 W U
 Types:Legendary Creature Human Artificer
 PT:4/4
+DeckRule:Copies:Limit$ Unlimited | Affected$ namedCARDNAME | Description$ A deck can have any number of cards named CARDNAME.
 K:Cycling:W U
-DeckRule:Copies:Limit$ Unlimited | Affected$ Name:CARDNAME | Description$ A deck can have any number of cards named CARDNAME.
 S:Mode$ Continuous | Affected$ Creature.Artifact+YouCtrl,Creature.Hero+YouCtrl | AddPower$ X | AddToughness$ X | Description$ Artifact creatures and Heroes you control get +1/+1 for each Artificer you control and each Artificer card in your graveyard.
 SVar:X:Count$ValidBattlefield,Graveyard Artificer.YouCtrl
 DeckHas:Ability$Discard
--- a/forge-core/src/main/java/forge/card/CardFace.java
+++ b/forge-core/src/main/java/forge/card/CardFace.java
@@ -69,6 +69,43 @@ final class CardFace implements ICardFace, Cloneable {
     // these are raw and unparsed used for Card creation
     @Override public Iterable<String> getKeywords()   { return keywords; }
     @Override public Iterable<String> getDeckRules()  { return deckRules; }
+
+    private transient List<DeckRuleLine> tokenizedDeckRules = null;
+
+    /** This face's own {@code DeckRule:} lines, split into (rule class, params) pairs and cached - computed once per face and shared across every printing, rather than once per {@code PaperCard}. */
+    @Override public List<DeckRuleLine> getTokenizedDeckRules() {
+        if (tokenizedDeckRules == null) {
+            final List<DeckRuleLine> result = new ArrayList<>();
+            for (final String raw : deckRules) {
+                final int colonPos = raw.indexOf(':');
+                final String ruleClass = colonPos > 0 ? raw.substring(0, colonPos) : raw;
+                final String rest = colonPos > 0 ? raw.substring(colonPos + 1) : "";
+                result.add(new DeckRuleLine(ruleClass, parseDeckRuleParams(rest)));
+            }
+            tokenizedDeckRules = result;
+        }
+        return tokenizedDeckRules;
+    }
+
+    /** Splits a {@code Key$ Value | Key2$ Value2} parameter string (see forge.deck.DeckRule, which builds the typed rule objects from this). */
+    private static Map<String, String> parseDeckRuleParams(final String rest) {
+        final Map<String, String> params = new LinkedHashMap<>();
+        for (final String piece : rest.split("\\|")) {
+            final String trimmed = piece.trim();
+            if (trimmed.isEmpty()) {
+                continue;
+            }
+            final int dollarPos = trimmed.indexOf('$');
+            if (dollarPos < 0) {
+                continue;
+            }
+            final String key = trimmed.substring(0, dollarPos).trim();
+            final String value = trimmed.substring(dollarPos + 1).trim();
+            params.put(key, value);
+        }
+        return params;
+    }
+
     @Override public Iterable<String> getAbilities()  { return abilities; }
     @Override public Iterable<String> getStaticAbilities() { return staticAbilities; }
     @Override public Iterable<String> getTriggers()   { return triggers; }
--- a/forge-core/src/main/java/forge/card/CardRules.java
+++ b/forge-core/src/main/java/forge/card/CardRules.java
@@ -906,16 +906,6 @@ public final class CardRules implements ICardCharacteristics {
         return false;
     }
 
-    public Integer getKeywordMagnitude(final String k) {
-        for (final String inst : mainPart.getKeywords()) {
-            final String[] parts = inst.split(":");
-            if (parts[0].equals(k) && StringUtils.isNumeric(parts[1])) {
-                return Integer.valueOf(parts[1]);
-            }
-        }
-        return null;
-    }
-
     public ColorSet getDeckbuildingColors() {
         if (deckbuildingColors == null) {
             byte colors = 0;
--- a/forge-core/src/main/java/forge/card/CardRulesPredicates.java
+++ b/forge-core/src/main/java/forge/card/CardRulesPredicates.java
@@ -328,8 +328,8 @@ public final class CardRulesPredicates {
         return result == null ? card -> false : result;
     }
 
-    /** Tokenizes dot-then-plus, matching {@code Card.isValid()}'s own restriction-string split. */
-    private static Predicate<CardRules> restrictionBranch(final String rawBranch) {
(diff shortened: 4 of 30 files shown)


Hanmac:
Why does `DeckRule.parseAll` need to be parsed again? Shouldn't it be only parsed once and added to PaperCard?

forge-gui/res/cardsfolder/c/cid_timeless_artificer.txt
```diff
@@ -3,7 +3,7 @@ ManaCost:2 W U
 Types:Legendary Creature Human Artificer
 PT:4/4
 K:Cycling:W U
-K:A deck can have any number of cards named CARDNAME.
+DeckRule:Copies:Limit$ Unlimited | Affected$ Name:CARDNAME | Description$ A deck can have any number of cards named CARDNAME.
```
tool4ever: let's avoid the colon usage inside a param again

tool4ever (dismissed):

Hanmac:
@tool4ever what was your comment on doing this in CardRules instead of PaperCard?

tool4ever:
I initially missed that the other was the naked Strings only + I thought PaperCard will kind of lead to duplicated instances for each printing
at the same time maybe it needs to live there if we ever started doing variants involving this element? 🤷‍♂️

Jetz72:
While I do think this is moving the script syntax in the right direction, I'll emphasize that the point of #8066 is to break down large validation methods like `DeckFormat::getDeckConformanceProblem`, which outputs only a single string and whose logic needs to be duplicated if you want any useful data about the nature of the problem. The idea of the `DeckRule` objects is that a game mode or deck format maintains a list of rules that are configured accordingly. When we want to verify a deck is clear of rules violations, we run it through each of the rules imposed by that format or mode, and they can each flag violations in more granular ways while accounting for any odd cards in the list that should adjust their behavior.

Making card scripts more consistent and knocking out another couple pseudo-keywords are definitely benefits. But I think a proper implementation of #8066 will still involve heavy revisions to what is being added here.

Regarding caching rules on the `PaperCard`, I'd vote to keep them on the `CardFace`, with a helper method on the `PaperCard` (and possibly the `CardRules`? Though maybe it's safer without that...) that puts together the complete list of them across the respective `getAllFaces`. Avoids duplicate copies of the rules list for every printing while still allowing variants and alt-faces to include them, for whatever god-forsaken hypothetical set dabbles in that combination of mechanics.

Hanmac:
yeah, imo CardFace or CardRules might be better?

BigCrunch22:
Before I proceed with the truly intended function of https://github.com/Card-Forge/forge/issues/8066, I would like to create the DeckRule classes first.

Issue #11818, opened 2026-09-08
Archenemy: My Laughter Echoes crashes with NullPointerException when setting a scheme in motion again


ttmullins:
**Describe the bug**
`My Laughter Echoes` can crash an Archenemy game with a `NullPointerException` when its ability tries to set the previous scheme in motion again.

The crash occurs because the `Again$ True` path uses `controller.getActiveScheme()`, but the current `Player.setSchemeInMotion(SpellAbility cause, Card scheme)` implementation moves the scheme into the Command zone without updating `activeScheme`.

As a result, `getActiveScheme()` can return `null`, which is then passed into `moveToCommand()` and causes the crash.

The crash stack included:

java.lang.NullPointerException: Cannot invoke "forge.game.card.Card.getName()" because "c" is null
    at forge.game.GameAction.moveTo(GameAction.java:767)
    at forge.game.GameAction.moveTo(GameAction.java:746)
    at forge.game.GameAction.moveToCommand(GameAction.java:944)
    at forge.game.player.Player.setSchemeInMotion(Player.java:292)
    at forge.game.ability.effects.SetInMotionEffect.resolve(SetInMotionEffect.java:27)

**To Reproduce**
1. Start an Archenemy game.
2. Use a scheme deck containing `My Laughter Echoes`.
3. Set `My Laughter Echoes` in motion so it remains active as an ongoing scheme.
4. Later set a non-ongoing scheme in motion.
5. Allow `My Laughter Echoes` to trigger and choose to abandon it.
6. Forge attempts to set the previous scheme in motion again.
7. The game crashes with a `NullPointerException`.

**Expected behavior**

`My Laughter Echoes` should abandon itself and then set the previously active non-ongoing scheme in motion again without crashing.

The active scheme reference should remain valid when `SetInMotion | Again$ True` is resolved.

**Screenshots/Logs**
A crash log and the console output has been attached.

[Forge_console_SchemeCard_CTD_2026-09-07.txt](https://github.com/user-attachments/files/31959208/Forge_console_SchemeCard_CTD_2026-09-07.txt)
[Forge_SchemeCard_CTD_2026-09-07.txt](https://github.com/user-attachments/files/31959209/Forge_SchemeCard_CTD_2026-09-07.txt)

**Desktop (please complete the following information):**
- OS: Windows 11
- Browser: N/A
- Version: Forge 2.0.15-SNAPSHOT / latest daily snapshot

**Additional context**
[...]

PR #11821, opened 2026-09-08
Fix remaining My Laughter Echoes replay issues

ttmullins:
Follow-up to #11818 and #11820.

## Summary

PR #11820 fixed the original `NullPointerException` by restoring the `activeScheme` assignment in `Player.setSchemeInMotion(...)`.

While validating that fix, I found three remaining issues with `My Laughter Echoes` that can still cause incorrect behavior.

### 1. My Laughter Echoes can trigger on itself

The card script currently uses:

`ValidCard$ Card.!OnGoing`

`Ongoing` is a card type/supertype restriction, but in this form it is being treated as a card property restriction. As a result, the filter can incorrectly match My Laughter Echoes itself when it is first set in motion.

That causes the ongoing scheme to create an abandon trigger for its own `SetInMotion` event and can lead to a self-trigger/repeat loop.

The script now uses:

`ValidCard$ !Ongoing`

which correctly excludes ongoing schemes.

### 2. `Again$ True` should replay the exact triggering scheme

`SetInMotionEffect` currently handles `Again$ True` by replaying:

`controller.getActiveScheme()`

However, the `SetInMotion` trigger already stores the exact scheme that caused the trigger in `AbilityKey.Scheme`.

Using the triggering object is more precise than depending solely on the mutable `activeScheme` field, especially if game state changes before the trigger resolves.

The effect now:

- reads the triggering scheme from `AbilityKey.Scheme`,
- resolves it to the current card state,
- falls back to `activeScheme` for compatibility,
- and safely does nothing if neither can be resolved.

### 3. Abandoning an ongoing scheme bypasses normal trigger refresh

`AbandonEffect` moves a scheme directly from the Command zone back to the Scheme Deck with direct zone remove/add calls.

Because this does not go through the normal zone-change path, the card's active trigger registration is not immediately refreshed.

This matters for My Laughter Echoes because its `SetInMotionAgain` subability runs immediately after it is abandoned. Without refreshing the active triggers first, the abandoned scheme can still be registered while the replayed scheme is being set in motion.

`AbandonEffect` now clears and re-registers the abandoned card's triggers immediately after the move. Since My Laughter Echoes is no longer in its `TriggerZones$ Command` zone, its SetInMotion trigger is no longer active during the replay.

## Changes

- Change My Laughter Echoes from `ValidCard$ Card.!OnGoing` to `ValidCard$ !Ongoing`.
- Resolve `Again$ True` using the trigger's `AbilityKey.Scheme`, with `activeScheme` as a compatibility fallback.
- Refresh active triggers immediately after `AbandonEffect` moves a scheme back to the Scheme Deck.

## Result

With the #11820 fix plus these follow-up changes, My Laughter Echoes now:

- enters the Command zone without triggering on itself,
- remains ongoing normally,
- triggers when a later non-ongoing scheme is set in motion,
- abandons itself correctly,
- replays the exact triggering scheme once,
- and does not retrigger itself during that replay.

## Testing

Manually tested with an AI Archenemy:

- My Laughter Echoes was set in motion first and remained in the Command zone.
- It did not trigger on itself.
- On the following turn, the AI set a non-ongoing scheme in motion.
- My Laughter Echoes triggered and was abandoned.
- The triggering non-ongoing scheme was set in motion again exactly once.
- No crash or repeat loop occurred.

Full Windows/Linux build was also previously verified with:

`mvn -U -B clean -P windows-linux install`

Result:

- Full reactor: PASS
- Desktop tests: 450 run, 0 failures, 0 errors, 6 skipped
- Forge Installer: SUCCESS

## AI assistance

This contribution was developed with assistance from ChatGPT (OpenAI) for code analysis, debugging, and implementation guidance. The changes were manually reviewed and functionally tested before submission.

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/AbandonEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/AbandonEffect.java
@@ -34,6 +34,9 @@ public class AbandonEffect extends SpellAbilityEffect {
         controller.getZone(ZoneType.Command).remove(source);
         controller.getZone(ZoneType.SchemeDeck).add(source);
 
+        controller.getGame().getTriggerHandler().clearActiveTriggers(source, null);
+        controller.getGame().getTriggerHandler().registerActiveTrigger(source, false);
+
         final Map<AbilityKey, Object> runParams = AbilityKey.newMap();
         runParams.put(AbilityKey.Scheme, source);
         controller.getGame().getTriggerHandler().runTrigger(TriggerType.Abandoned, runParams, false);
--- a/forge-game/src/main/java/forge/game/ability/effects/SetInMotionEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/SetInMotionEffect.java
@@ -25,20 +25,7 @@ public class SetInMotionEffect extends SpellAbilityEffect {
 
         for (int i = 0; i < repeats; i++) {
             if (again) {
-                Card scheme = null;
-
-                Object triggeredScheme = sa.getRootAbility().getTriggeringObject(AbilityKey.Scheme);
-                if (triggeredScheme instanceof Card) {
-                    scheme = controller.getGame().getCardState((Card) triggeredScheme, null);
-                }
-
-                if (scheme == null) {
-                    scheme = controller.getActiveScheme();
-                }
-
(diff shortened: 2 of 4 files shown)


tool4ever (approved):
thanks

PR #11822, opened 2026-09-08
Fix Archenemy lobby team synchronization

### LINKED ISSUE #11812: Archenemy + Planechase: Heroes can be assigned to different teams and attack each other
**Describe the bug**

When Archenemy and Planechase are enabled together, players assigned as Heroes can end up on different internal teams.

This causes behavior that should not be possible in a normal Archenemy game, such as an AI Hero attacking another Hero instead of treating that player as a teammate.

I reproduced this multiple times in local multiplayer with:

- 1 human Hero
- 1 AI Hero
- 1 AI Archenemy
- Archenemy enabled
- Planechase enabled

The lobby correctly displayed both non-Archenemy players as `Heroes`, but during gameplay the AI Hero was able to declare attacks against the human Hero.

**Possible cause**

The current team assignment logic in:

`forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java`

contains:

```java
final boolean isArchenemy = slot.isArchenemy();
final int team = GameType.Archenemy.equals(currentGameType) && !isArchenemy ? 1 : slot.getTeam();
```

This appears to determine Hero team assignment based on `currentGameType` being exactly `GameType.Archenemy`.

However, Archenemy can be active at the same time as another variant such as Planechase.

If `currentGameType` is Planechase while Archenemy is still present in the active variant set, the Archenemy-specific Hero team assignment is skipped and Forge falls back to:

```java
slot.getTeam()
```

This could leave the Heroes on separate teams even though both are marked as Heroes in the lobby.

**To Reproduce**

1. Open the local Multiplayer lobby.
2. Add three players.
3. Enable `Archenemy`.
4. Enable `Planechase`.
5. Configure:
   - Player 1: Heroes
   - Player 2: Heroes
   - Player 3: Archenemy
6. Start the game.
7. Play until the AI Hero enters combat.
8. The AI Hero may treat the other Hero as an opponent and attack them.

I observed this more than once in separate games.

**Actual behavior**

The AI Hero is able to attack another Hero.

This suggests that the two Heroes are not being treated as members of the same team internally, despite the lobby displaying both players as `Heroes`.

**Expected behavior**

When the Archenemy variant is active:

- Exactly one player should be the Archenemy.
- All Heroes should be on the same team.
- Heroes should not be legal attack targets for other Heroes.

This should remain true when Archenemy is combined with another compatible variant such as Planechase.

**Possible fix**

Instead of checking whether `currentGameType` is exactly Archenemy:

```java
final int team = GameType.Archenemy.equals(currentGameType) && !isArchenemy
        ? 1
        : slot.getTeam();
```

it may be safer to determine team assignment from the active variant set:

```java
final int team = variantTypes.contains(GameType.Archenemy)
        ? (isArchenemy ? 0 : 1)
        : slot.getTeam();
```

I tested this change locally and it prevented the Heroes from being assigned as opponents when Archenemy was combined with Planechase.

I am not claiming this is necessarily the complete upstream fix, but it appears to address the behavior I was seeing.

**Desktop**

- OS: Windows 11
- Forge: Desktop application
- Version: 2.0.15-SNAPSHOT / latest daily snapshot

ttmullins:
Fixes #11812

## Summary

This fixes Archenemy lobby role selection becoming out of sync with the numeric team values stored in `LobbyData`.

The desktop lobby uses a dedicated Archenemy/Hero selector (`aeTeamComboBox`) when the Archenemy variant is active, but `VLobby.getSlot()` ultimately reads `PlayerPanel.getTeam()`.

Previously, `PlayerPanel.getTeam()` always returned the value from the normal team combo box, even while that combo was hidden and the Archenemy/Hero selector was being displayed.

That meant the UI could correctly display:

- Archenemy
- Heroes

while the underlying `LobbySlot.team` values still contained stale or unrelated numeric team assignments.

This caused multiple incorrect behaviors.

With Archenemy + Planechase, Heroes could end up on different internal teams and be treated as opponents, allowing an AI Hero to attack another Hero.

A second reproduction also showed that after switching which player was the Archenemy, both players could end up on the same numeric team. The game would then immediately end at Turn 0 because Forge believed all opposing teams had already lost.

The existing `startGame()` logic partially masked the problem by forcing non-Archenemy players onto team 1 only when `currentGameType` was exactly `GameType.Archenemy`.

That does not work reliably when Archenemy is active alongside another variant such as Planechase, because `currentGameType` may refer to the other active variant even though Archenemy is still present in the applied variant set.

## Changes

- `PlayerPanel.getTeam()` now returns the Archenemy/Hero selector value while the Archenemy variant is active.
- `GameLobby.applyToSlot()` now synchronizes the numeric team whenever it updates a player's Archenemy/Hero role.
- Archenemy is assigned team 0 and Heroes are assigned team 1 in the underlying lobby state.
- `GameLobby.startGame()` now derives Archenemy teams from the active variant set rather than requiring `currentGameType` to be exactly `GameType.Archenemy`.

## Result

While Archenemy is active, Forge now consistently maintains:

- Archenemy = team 0
- Heroes = team 1

This keeps the visible Archenemy/Hero selection, `LobbyData`, and the final `RegisteredPlayer` team assignment synchronized.

It also prevents:

- Heroes from being treated as opponents in Archenemy + Planechase,
- stale hidden team values from affecting gameplay,
- and the immediate Turn 0 game-over seen after switching which player is the Archenemy.

The `startGame()` normalization remains as a defensive check so malformed or stale lobby data cannot produce incorrect teams when the match begins.

## Testing

Manually tested:

- Human Hero vs AI Archenemy.
- Switching the Archenemy role between the human and AI.
- Starting a game after changing the Archenemy assignment.
- Archenemy + Planechase team behavior.
- AI Archenemy gameplay through multiple turns.

The previous Turn 0 game-over no longer occurred, and the AI Archenemy was able to play normally.

Ran:

`mvn -U -B clean -P windows-linux install`

Result:

- Full reactor: PASS
- Desktop tests: 450 run, 0 failures, 0 errors, 6 skipped
- Forge Installer: SUCCESS

## AI assistance

This contribution was developed with assistance from ChatGPT (OpenAI) for code analysis, debugging, and implementation guidance. The changes were manually reviewed and functionally tested before submission.

### DIFF
--- a/forge-gui-desktop/src/main/java/forge/screens/home/PlayerPanel.java
+++ b/forge-gui-desktop/src/main/java/forge/screens/home/PlayerPanel.java
@@ -906,7 +906,9 @@ public class PlayerPanel extends FPanel {
     }
 
     public int getTeam() {
-        return teamComboBox.getSelectedIndex();
+        return lobby.hasVariant(GameType.Archenemy)
+                ? aeTeamComboBox.getSelectedIndex()
+                : teamComboBox.getSelectedIndex();
     }
     public void setTeam(final int team) {
         teamComboBox.suppressActionListeners();
--- a/forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
+++ b/forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
@@ -138,6 +138,7 @@ public abstract class GameLobby implements IHasGameType {
                     lastArchenemy = otherIndex;
                 }
                 otherSlot.setIsArchenemy(becomesArchenemy);
+                otherSlot.setTeam(becomesArchenemy ? 0 : 1);
             }
         }
 
@@ -448,7 +449,9 @@ public abstract class GameLobby implements IHasGameType {
             final int avatar = slot.getAvatarIndex();
             final int sleeve = slot.getSleeveIndex();
             final boolean isArchenemy = slot.isArchenemy();
-            final int team = GameType.Archenemy.equals(currentGameType) && !isArchenemy ? 1 : slot.getTeam();
+            final int team = variantTypes.contains(GameType.Archenemy)
+                    ? (isArchenemy ? 0 : 1)
+                    : slot.getTeam();
             final Set<AIOption> aiOptions = slot.getAiOptions(); // TODO: could AiOptions carry the choice of which AI is selected to play against?
 
             final boolean isAI = slot.getType() == LobbySlotType.AI;
--- a/forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
+++ b/forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
@@ -449,9 +449,7 @@ public abstract class GameLobby implements IHasGameType {
             final int avatar = slot.getAvatarIndex();
             final int sleeve = slot.getSleeveIndex();
             final boolean isArchenemy = slot.isArchenemy();
-            final int team = variantTypes.contains(GameType.Archenemy)
-                    ? (isArchenemy ? 0 : 1)
-                    : slot.getTeam();
+            final int team = slot.getTeam();
             final Set<AIOption> aiOptions = slot.getAiOptions(); // TODO: could AiOptions carry the choice of which AI is selected to play against?
 
             final boolean isAI = slot.getType() == LobbySlotType.AI;
--- a/forge-gui-mobile/src/forge/screens/constructed/PlayerPanel.java
+++ b/forge-gui-mobile/src/forge/screens/constructed/PlayerPanel.java
@@ -1107,7 +1107,9 @@ public class PlayerPanel extends FContainer {
     }
 
     public int getTeam() {
-        return cbTeam.getSelectedIndex();
+        return screen.hasVariant(GameType.Archenemy)
+                ? cbArchenemyTeam.getSelectedIndex()
+                : cbTeam.getSelectedIndex();
     }
     public void setTeam(int team0) {
         applyingTeamFromNetwork = true;


forge-gui-desktop/src/main/java/forge/screens/home/PlayerPanel.java
```diff
@@ -906,7 +906,9 @@ public void setSleeve(final int index, final String artKey, final int artOffset)
     }
 
     public int getTeam() {
-        return teamComboBox.getSelectedIndex();
+        return lobby.hasVariant(GameType.Archenemy)
```
tool4ever, line 909: I think gui-mobile has the same problem?

forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
```diff
@@ -448,7 +449,9 @@ public Runnable startGame() {
             final int avatar = slot.getAvatarIndex();
             final int sleeve = slot.getSleeveIndex();
             final boolean isArchenemy = slot.isArchenemy();
-            final int team = GameType.Archenemy.equals(currentGameType) && !isArchenemy ? 1 : slot.getTeam();
+            final int team = variantTypes.contains(GameType.Archenemy)
```
tool4ever: shouldn't just `getTeam()` be correct at this point now because you fixed it in the slot data?

forge-gui-desktop/src/main/java/forge/screens/home/PlayerPanel.java
ttmullins: You are correct, gui-mobile had the same issue and I have now applied the getTeam() fix there as well.

forge-gui/src/main/java/forge/gamemodes/match/GameLobby.java
ttmullins: Correct again, since applyToSlot() now keeps LobbySlot.team synchronized with the Archenemy role, startGame() can just use slot.getTeam(). I have made the correction.

tool4ever (approved):

Issue #144, opened 2022-04-26
Animate and Pump Effects should use StaticLayer


Hanmac:
for Continuous Effects happening in the right order or maybe be affected by dependency, Animate and Pump might need to use Static Abilities

this might also be needed when talking about linked abilities

If possible, it would be ugly if they would appear as Effects in the Command Zone

Northmoc:
Instead of Command Zone effects possibly information in the Card Detail pane?

Someone mentioned in the Discord putting non-intrinsic stuff first in Card Detail, not sure I agree 100% but it is a thought.

Hanmac:
for example which layers are affected by possible dependency Rules:

* 613.1d Layer 4: Type-changing effects are applied.
Cards like Blood Moon
or Graaz, Unstoppable Juggernaut applying to previous non-creatures that got animated
* 613.1f Layer 6: Ability-adding effects, keyword counters, ability-removing effects, and effects that say an object can’t have an ability are applied.
like Cavalry Master that refers to other creatures with abilities
or stuff that makes the card loose all abilities
*613.4b Layer 7b: Effects that set power and/or toughness to a specific number or value are applied. Effects that refer to the base power and/or toughness of a creature apply in this layer.
Andrios, Roaming Explorer for example, the pump ability otherwise wouldn't make the static ability apply for these cards

for which Layer 4 and unlikely Layer 6 could have ways to cause a dependency Loop

tool4ever:
Related example but without dependency:
https://www.reddit.com/r/mtgrules/comments/i659qf/heliod_suncrowned_interaction_with_dance_of_the/

Hanmac:
> Related example but without dependency: https://www.reddit.com/r/mtgrules/comments/i659qf/heliod_suncrowned_interaction_with_dance_of_the/

that's slightly unrelated on that Pump Effects should use Layers, but more about that the Player might need to be forced to choose relative Timestamps (which might be annoying)

Hanmac:
@tehdiplomat @Northmoc @kevlahnota can i make Effect Objects in the Command Zone, but stop them from getting rendered in the GUI?

tehdiplomat:
I don't think that's on by default, but it probably wouldn't be too bad to add a flag for CardViews to make them invisible (or maybe based off a Match setting so devs can flip it on to see all of the ones active, but for players it would just be off)

tool4ever:
[...]

PR #11826, opened 2026-09-08
Defer subtype sanitizing to end of layer 4 fold

bbabnikk:
Type-granting statics played before a one-shot animation lost the granted subtype: e.g. Arcane Adaptation / Leyline of Transformation out first, then an earthbend activation → the land ended up a Creature with no chosen type.

Root cause: layer 4 entries fold in timestamp order and each step enforced CR 205.3d on the intermediate type, dropping the subtype before the legitimizing Creature type arrived. Paper applies the animation first via dependency (CR 613.8). The fold now validates once on the final type.

- forge-core CardType: sanisfy flag on the CardTypeView bulk ops
- forge-core CardChangedType: no intermediate sanitize in applyChanges

Verification (no test committed per repo guidelines on agent-added tests): compiled the actual forge-core type-system sources (CardType, CardChangedType, StateChangedType, WordChangedType, registries seeded as FModel.loadDynamicGamedata does) and folded timestamp-ordered entries:

- Before fix: grant-before-animation → Land Creature, Sliver missing (bug reproduced); animation-before-grant → OK; Maskwood AddAllCreatureTypes shape → OK (matches the "not always" report: Maskwood worked, Adaptation/Leyline didn't).
- After fix: all three keep the granted type.
- Guards: a grant with no legitimizing type is still dropped (205.3d end-state enforcement intact), subtype removals still apply, printed subtypes (Forest) unaffected.
- Single-arg addAll/removeAll delegate unchanged, so all other callers keep identical behavior; only CardChangedType.applyChanges (sole fold caller, via getTypeWithChanges) defers sanitizing, and the fold's existing end sanitize remains the enforcer.

disclosure: written with an AI coding agent (Muse Code).


Hanmac:
Wrong fix, it should be fixed via Dependency Layers

tool4ever:
that looks harmful, needs real fix by #144 instead 👎

bbabnikk:
Ok, closing. will approach it from the direction of #144 instead.

PR #11763, opened 2026-09-01
enhancement: consistent lazy loading

JacopoMadaluni:
## The problem
A forge engine needs the full cardb in memory (about 400MB) to run a single game.
This is impractical for someone hosting a server, or in my case, for running forge in the browser via GraalVM and web workers.

## The proposed fix
Re enable `LOAD_CARD_SCRIPTS_LAZILY` and make lazy lookups consistent with eager lookup.

Eager mode is unchanged, the following only applies when lazy mode is enabled.

At startup every card script is stream-parsed to build a `{name: file}` index covering all names from all scripts.
This includes CopyFaces, flavour names, double `//` cards. The parsed rules are discarded to keep resident memory light. Builds in about 1s and runs inside a 64MB heap in my case.
A lookup then becomes a matter of resolving a file from the name from the index. This should, as far as I understand, yield identical coverage to eager mode.

A minor util was included, `resetLazyLoadedCards()`, which allows the caller to drop accumulated cards. Custom cards are re-fed afterwards so they survive the reset. I wasn't sure about these so just in case, I'm never clearing them up. I don't fully understand the impact this has on custom cards.

#### ensureAllCardsLoaded 

Theres a bunch of places in which I had to call `ensureAllCardsLoaded`. I don't particularly like this, but theres some nasty cards like [jhoira-of-the-ghitu-avatar](https://scryfall.com/card/pmoa/69/jhoira-of-the-ghitu-avatar) that as far as I understand would need to take a random sorcery from the whole card pool. That would require the whole thing to be in memory. Theres other few cases that require this, but I think they are all warranted.

## Testing

Compiles against master (`forge-core`, `forge-game`, `forge-gui`).
```
$ mvn -U -B clean test
forge-gui-desktop: Tests run: 360, Failures: 0, Errors: 0, Skipped: 6
forge-game:        Tests run: 3,   Failures: 0, Errors: 0, Skipped: 0
```

We are also about to test this live on the manabrew app for stability.
The following cards, made up of edge cases, loads fine under lazy loading:
```
Lim-Dûl's Vault
Fire // Ice
Start // Fire
A-Acererak the Archlich
+2 Mace
Rocket-Powered Goblin Glider
Stitcher's Wings
Mintstrosity
K9, Mark I
Archangel Avacyn
Vhal, Eager Scholar
```

Issue #11183, opened 2026-07-07
~142 tests in forge-gui-desktop silently never run (DeckRecognizerTest and CardDb mock test classes execute 0 tests, locally and in CI)


Blackvipe99:
**Describe the bug**
Several test classes in `forge-gui-desktop` silently execute **zero tests** — both locally and in the `Test build` CI workflow. The build stays green, so this is easy to miss, but ~142 tests that look like coverage aren't actually running:

| class | @Test methods | executed |
|---|---|---|
| `forge.deck.DeckRecognizerTest` | 82 | 0 |
| `forge.card.CardDbCardMockTestCase` | 54 | 0 |
| `forge.card.CardDbLazyCardLoadingCardMockTestCase` | 4 | 0 |
| `forge.card.CardEditionCollectionCardMockTestCase` | 2 | 0 |

~~These files are actively maintained (e.g. `DeckRecognizerTest` was touched by the Scarlet Witch fix for #11145), so tests are presumably being added/updated in the belief that CI runs them.~~ *(edit: struck — the #11145 attribution was wrong, an artifact of file-history tracking in a shallow local clone; see comment below.)*

**To Reproduce**
1. `mvn -pl forge-gui-desktop -am test` → `Tests run: 282` — the surefire report `TEST-TestSuite.xml` contains no `classname="forge.deck.DeckRecognizerTest"` (or the CardDb classes) entries. The same 282 count appears in the latest master `Test build` run (28857278586).
2. Targeted run: `mvn -pl forge-gui-desktop -am test -Dtest=DeckRecognizerTest -Dsurefire.failIfNoSpecifiedTests=false` → build SUCCESS, but the report shows `tests="0" errors="0" skipped="0" failures="0"`. No warning anywhere.

**Expected behavior**
Those tests execute in `mvn test` (or fail loudly if the class can't be instantiated), so regressions in the deck importer / card DB are actually caught.

**Additional context / observations**
- All four dead classes extend `CardMockTestCase` (PowerMock + TestNG). PowerMock itself isn't the whole story: `GameSimulationTest` extends `BaseGameSimulationTest extends CardMockTestCase` and **does** run.
- One visible difference: the dead classes' `@Test` methods are package-private (`void testX()`), while e.g. `GameSimulationTest`'s are `public` — TestNG documents that test methods should be public. (Making two methods public in a quick local experiment did not get them discovered either, so class-level instantiation may also be failing silently.)
[...]

PR #11830, opened 2026-09-09
Fix Testng Powermock silent test skip

leriomaggio:
This PR addresses the issues in #11183.

## What was broken

TestNG 7.10 removed `org.testng.IObjectFactory`. PowerMock 2.0.9 still
references it, so every subclass of `PowerMockTestCase` throws
`NoClassDefFoundError` when TestNG tries to inspect it. TestNG treats a
class it cannot inspect as a class with no tests, so the build stayed green.

The issue estimated 4 classes and ~142 tests.
**It is 163 tests across 8 classes.**
 
~The actual numbers:~
~- **163 tests across 8 classes** were hidden by the PowerMock breakage~
~- **101 test methods** were also package-private, which TestNG ignores~
  ~silently. 89 of those sit inside the same 8 classes; 12 more were in~
  ~otherwise-healthy classes and were lost to visibility alone.~

Affected classes: `CardDbCardMockTestCase`, `DeckRecognizerTest`,
`CardDbLazyCardLoadingCardMockTestCase`, `CardDbPerformanceTests`,
`CardEditionCollectionCardMockTestCase`,
`CardDbWithNoImageCardDbMockTestCase`, `ComprehensiveRulesSection103`,
`ComprehensiveRulesSection104`.

## Changes and Fix

PowerMock was only used for static mocking and private field access.
Mockito 5's inline mock maker does both, so PowerMock is removed entirely.
Losing its per-class classloader means static state now leaks between test
classes running in the same JVM. Static mocks are closed in `@AfterMethod`,
and the `Localizer` singleton is restored to whatever was installed before,
rather than cleared, so the AITest-based classes that set it up once per JVM
keep working.

## Fixes after restoring tests

1. `DeckRecognizer` failed to parse colour letters `w/u/b/r/g/c`, card
   names containing `?`, and collector numbers with lowercase runs such as
   `118†s`.
2. `CardEdition.UNKNOWN.getScryfallCode()` threw NPE, hitting any caller
   that iterates the edition collection.
3. The game simulation harness passed a `null` controller to
   `PlaySpellAbility`. That was safe until #10051 moved the code into
   forge-game, where the controller is dereferenced.
4. `CardDbLazyCardLoadingCardMockTestCase` checked a card was not loaded yet by
  calling `getCard()` and expecting null. PR #11763 made every
  lookup load the card on demand. Those checks use `contains()` instead, 
which reads the loaded-card index without loading. Nothing caught that behaviour change when it merged, because these tests were among the ones not running.

## Test oracle changes

Many tests in CardDB assert against the live card database, so reprints
drift them. Where an oracle could be computed, it now is: latest-printing
editions, art indices and one boundary date. 
Moreover, six tests expected `null` when requesting a card from a set it was never printed in, but this behaviour has changed to fall back to any printing instead. 
The tests now assert non-null.

## Regression guard

`TestSuiteIntegrityTest` fails the build if a test class stops being
collected. ~Two checks, one per failure mode found here: every test class
loads and inspects the way TestNG does it, and every `@Test` method is
public. Both fail on the pre-fix tree.~
It loads every test class and inspects it the way TestNG does, which is the
call that used to throw `NoClassDefFoundError`. It fails on the pre-fix tree.

All tests now are 🟢 (also in CI!!)

## Known hazard 

Test classes now share a JVM and its static state. Anything a test installs into a singleton is visible to every class that runs after it, so a suite that passes per-package in an IDE can still fail under `mvn test`.


Hanmac:
I'm cherry-picking the test-unrelated ones into separate MRs.

As for the "parse colour letters w/u/b/r/g/c" thing, i need to debug that myself later.
Probably would be better to directly return the color enum?

tool4ever:
I feel like they are related since we didn't see them without these tests, too much splitting just hides that logical flow

forge-gui-desktop/src/test/java/forge/FCollectionTest.java
```diff
@@ -45,7 +45,7 @@ void testBadIteratorLogicTwo() {
      * from the collection while looping over it neither throws nor skips an element.
      */
     @Test
-    void testRemoveWhileIterating() {
+    public void testRemoveWhileIterating() {
```
tool4ever: this test was definitely being run already (#11677)
are you calling them differently?

forge-gui-desktop/src/test/java/forge/card/CardDbLazyCardLoadingCardMockTestCase.java
```diff
@@ -100,13 +118,15 @@ public void tesLoadAndGetUnsupportedCardHavingWrongSetCode() {
         String expectedSetCode = "EXO"; // Exodus
         CardRarity expectedCardRarity = CardRarity.Rare;
 
-        PaperCard dominatingLycidCard = this.cardDb.getCard(cardName);
-        assertNull(dominatingLycidCard);
+        // #11763 made every CardDb lookup load the card on demand, so getCard() no longer
```
tool4ever: I don't think we need such multiple lines comments four times

forge-gui-desktop/src/test/java/forge/card/CardDbLazyCardLoadingCardMockTestCase.java
tool4ever: is there a race or is is guaranteed that `initMocks()` from parent executes first? (because otherwise cardDb would be the wrong one?)

forge-gui-desktop/src/test/java/forge/card/CardDbLazyCardLoadingCardMockTestCase.java
```diff
@@ -25,8 +25,15 @@ public void setup() {
...
     protected void initializeStaticData() {
-        StaticData data = CardDatabaseHelper.getStaticDataToPopulateOtherMocks(true);
-        PowerMockito.when(FModel.getMagicDb()).thenReturn(data);
+        // A database of this class's own, loaded lazily, and a fresh one for every method.
+        //
+        // Every test here asserts a card is not yet loaded, loads it, then asserts it is, so a
+        // database shared between methods stops being pristine as soon as the first one runs.
+        // This deliberately does not use the keyed cache CardDatabaseHelper offers the other
+        // CardDb test classes: lazy loading only indexes card names rather than parsing all
+        // 33,000 card scripts, so rebuilding it per method costs well under a second.
```
tool4ever: I'm not sure this doesn't undersell the cost, it still has to parse everything to get the names after all?

at the very least could try to use `resetLazyLoadedCards()`?

forge-gui-desktop/src/test/java/forge/FCollectionTest.java
leriomaggio: You're right, and thanks for taking a second look at it. 
I re-checked TestNG's AnnotationHelper and I realised I misinterpreted the logic of how that works: the `isPublic` check only applies when the class itself is annotated @Test. A method carrying its own @Test is collected whatever its visibility. So I reverted all the changes there, realigning with master so to keep the "noise" in the PR at minimum 🙌 

On that note, I've also removed the matching check from `TestSuiteIntegrityTest`
(will update the PR description accordingly in a moment!)

forge-gui-desktop/src/test/java/forge/card/CardDbLazyCardLoadingCardMockTestCase.java
leriomaggio: Done ✅ It's written in the class now!

forge-gui-desktop/src/test/java/forge/card/CardDbLazyCardLoadingCardMockTestCase.java
leriomaggio: No race: TestNG runs the parent's @BeforeMethod before the child's. But you're right that relying on it isn't obvious, so I've removed the dependency. `cardDb` is assigned in `initializeStaticData()`, which the parent calls once the mock is in place.

forge-gui-desktop/src/test/java/forge/card/CardDbLazyCardLoadingCardMockTestCase.java
leriomaggio: Agreed. Switched to building the database once and calling `resetLazyLoadedCards()` before each test.

leriomaggio:
> I'm cherry-picking the test-unrelated ones into separate MRs.
> 
> As for the "parse colour letters w/u/b/r/g/c" thing, i need to debug that myself later. Probably would be better to directly return the color enum?

Happy either way on returning the `enum` from `getMagicColor`.

One thing worth knowing if the fix goes into `MagicColor.Color.fromName` instead, as I actually thought about it myself and the enum was my first instinct too: `ImageUtil.specFaceToCollectorSuffix` treats a `null` return as "not being a colour", so accepting single letters there would make any single-letter face name produce a collector-number suffix. That's why I normalised in the recogniser rather than in the `enum`.

tool4ever (approved):

PR #11831, opened 2026-09-09
fix(cards): correct two malformed script keys

jczastkiewicz:
Spirit of Resilience writes DBCleanup: where it means SVar:DBCleanup:

The Dawning Archaic writes ODeckHints: for DeckHints:

Fixed


tool4ever (approved):
thanks

PR #11832, opened 2026-09-09
AI: skip unrelated regeneration abilities before checking costs

Shedletsky:
At mtgbattles.com we were running 100,000+ Forge games per day, and a small percentage repeatedly exceeded our three-minute game limit in AI vs AI matches. This bug was the leading cause of those when using only cards from the Unlimited set.

Local debugging found combat prediction evaluating Wall of Bone's regeneration playability and payment while asking whether Dragon Whelp could regenerate. The same unnecessary work appeared for animated Swamps in a Kormus Bell game.

Move the existing target/Defined applicability check in `ComputerUtil.canRegenerate` before `canPlay` and `canPayCost`. Relevant abilities still go through the existing playability, mana, life-payment, and sacrifice checks. This applies to both self-regeneration and targeted regeneration; the engine change is confined to this method.

Replaying the two frozen matchups with **Temurin Java 17.0.20.1+1** shows the impact:

| Seeded matchup | Before, Java 17 | After, Java 17 | Observed improvement |
| --- | ---: | ---: | --- |
| Wall of Bone / Dragon Whelp | 385.2 s | 89.2 s | 4.3× faster; 76.8% less time |
| Wall of Bone / Kormus Bell | >600 s (timed out) | 148.2 s | >4.0× faster (lower bound) |

Both patched games finished naturally. The Dragon Whelp comparison produced **2,184 identical gameplay log lines**, with the same winner and the same **25 regeneration resolutions and 25 destruction replacements** in the same sequence. The speedup did not come from skipping regeneration or changing that game's actions. A run that reaches the time limit is counted as a timeout even if Forge prints an apparent winner.

<details>
<summary>Comparison setup and limits</summary>

- **After:** the exact Java 17 package tested from submitted head `d1e207a56a1fb033463510e7d9c24a6c61d625d9`, which includes `master` through `4bee0abda5`.
- **Before:** that same package with only `ComputerUtil.class` replaced by the version compiled from `4bee0abda5` using the same JDK and compiler options. The source difference is this PR's fix. Every other JAR entry is byte-for-byte identical. Recompiling the patched class with those options also matched the submitted class byte-for-byte.
- Same frozen decks, seats, seeds, resource files, runner, 2 GB JVM heap, and 600-second game/AI-decision limits on both sides. Each replay used a fresh JVM; the four replays ran in sequence without a debugger or stack sampling. The order was before/after for Dragon Whelp and after/before for Kormus Bell.
- One replay per variant per matchup on a shared development workstation, with other development tasks active. These timings demonstrate the problem on these inputs; they are not a statistical estimate of fleet-wide speedup. Earlier Java 17 replays of the patched Kormus Bell game included a timeout at 180 seconds and a natural completion in 230.7 seconds with a longer limit, so this optimization does not guarantee every game finishes within three minutes.

</details>

Validation on the submitted head used **Java 17 for both compilation and execution**:

- Clean full Windows/Linux source build and Checkstyle passed.
- The full default Maven test suite passed: **460 passed, 0 failures/errors, 6 skipped** across the reactor. This includes all 12 new regeneration cases and seven existing automatic-payment/reserved-mana cases. All six skips are existing opt-in network stress tests requiring `-Drun.stress.tests=true`.
- Two real mtgbattles compatibility games passed against the packaged JAR from this commit, checking turn, cast, damage, outcome, and result-line formats.
- The 12 regeneration cases were rerun against both comparison packages on Java 17. All **11 behavior cases pass on both**. The performance regression test fails on the unpatched package as expected (one unrelated playability evaluation instead of zero) and passes on the submitted package.

The [test source is included in this PR](https://github.com/Shedletsky/forge/blob/d1e207a56a1fb033463510e7d9c24a6c61d625d9/forge-gui-desktop/src/test/java/forge/ai/RegenerationPredictionTest.java). It covers unrelated-ability evaluation, self and targeted regeneration, missing/wrong/tapped mana, suppressed abilities, cannot-regenerate effects, opponent prediction, life and sacrifice costs, and actual AI regeneration through lethal combat with Wall of Bone, Living Wall, Uthden Troll, and Will-o'-the-Wisp. Prediction itself must not spend resources or create shields.

Reproduce the build and tests from the repository root with `JAVA_HOME` and `PATH` selecting JDK 17:

```sh
mvn -version
mvn -B clean -P windows-linux -Dmaven.compiler.debug=true -Dmaven.compiler.debuglevel=lines,vars,source install
```

We used **Astra (xhigh) in Codex** to investigate, implement, and test this fix. We are software developers, but not experts in the Forge codebase.

### DIFF
--- a/forge-gui-desktop/src/test/java/forge/ai/RegenerationPredictionTest.java
+++ b/forge-gui-desktop/src/test/java/forge/ai/RegenerationPredictionTest.java
@@ -1,0 +1,170 @@
+package forge.ai;
+
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import forge.game.Game;
+import forge.game.ability.AbilityApiBased;
+import forge.game.ability.ApiType;
+import forge.game.card.Card;
+import forge.game.combat.Combat;
+import forge.game.cost.Cost;
+import forge.game.phase.PhaseType;
+import forge.game.player.Player;
+import forge.game.spellability.SpellAbility;
+import forge.game.zone.ZoneType;
+
+import org.testng.AssertJUnit;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class RegenerationPredictionTest extends AITest {
+
+    @DataProvider(name = "selfRegenerators")
+    public Object[][] selfRegenerators() {
+        return new Object[][] {
+            { "Wall of Bone", "Swamp", "Forest" },
+            { "Will-o'-the-Wisp", "Swamp", "Forest" },
+            { "Uthden Troll", "Mountain", "Forest" },
+            { "Living Wall", "Forest", null }
+        };
+    }
+
+    @Test(dataProvider = "selfRegenerators")
+    public void predictionRespectsManaAndAbilityRestrictions(String name, String manaName, String wrongManaName) {
+        Game game = initAndCreateGame();
+        Player ai = game.getPlayers().get(1);
+        Card creature = addCard(name, ai);
+        Card other = addCard("Dragon Whelp", ai);
+        game.getAction().checkStateEffects(true);
+        AssertJUnit.assertFalse(ComputerUtil.canRegenerate(ai, creature));
+        if (wrongManaName != null) {
+            addCard(wrongManaName, ai);
+            game.getAction().checkStateEffects(true);
+            AssertJUnit.assertFalse("Wrong-colored mana must not pay for regeneration", ComputerUtil.canRegenerate(ai, creature));
+        }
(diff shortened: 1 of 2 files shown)


forge-gui-desktop/src/test/java/forge/ai/RegenerationPredictionTest.java
```diff
@@ -0,0 +1,170 @@
+package forge.ai;
```
tool4ever, line 1: such changes are welcome
just remove this test class, it's way too specific

tool4ever:
Fixed by #11920

PR #11837, opened 2026-09-09
Configurable pod size and picks per pack for network draft

MostCromulent:
Screenshot 2026-09-09 191656

## Summary

Player requested to be able to draft online with four people rather than have bots fill the rest of an eight-seat pod. The host of a network draft can now choose the pod size and how many cards are taken from each pack.

## Fix

Edition files already carry `RecommendedPodSize` and `DoublePick`, and the offline draft reads them silently. Marvel's Spider-Man and Teenage Mutant Ninja Turtles declare a four-seat pod at two cards per pack, Commander Legends declares two cards every pack, and Double Masters two on the first pick of each pack. This surfaces those values to the network host as defaults rather than inventing new configuration.

- Pod size and picks-per-pack controls in the event setup dialog on desktop and mobile. The set's own value is marked and preselected, so anything else reads as a house rule. Seats nobody takes are drafted by AI, which do not play in the match.
- The pick rule moves into a single predicate shared by the offline and network draft paths. It previously ran only in offline drafts.
- The pod size limits the lobby, so seats can never outnumber the pod.

---


Shedletsky:
I've thought about adding similar draft options to Forge for a long time, this is exciting to see.

tool4ever (approved):
LGTM

PR #11839, opened 2026-09-09
Fix ChangeTargetsEffect that affects Chira's random targeting

BigCrunch22:
## What
`ChangeTargetsEffect` set `AbilityKey.Cause` to the `SpellAbility` itself when firing `BecomesTargetOnce`, instead of `sa.getHostCard()`.

## Why it mattered
`SpellAbility.isValid()` only matches spell/ability-shaped restrictions (`Spell`, `Instant`, `Triggered`, etc.) against itself — `Card.*` restrictions always fail there, correctly, since a `SpellAbility` isn't a `Card`. So any `ValidCause$ Card...` check on the trigger silently failed. `MagicStack`'s own `BecomesTargetOnce` firing already passes `getHostCard()`; this call site just didn't match it.

## Fix
```java
runParams.put(AbilityKey.Cause, sa.getHostCard());
```
***Used Claude as assistance for the creation of the code***

### DIFF
--- a/forge-game/src/main/java/forge/game/ability/effects/ChangeTargetsEffect.java
+++ b/forge-game/src/main/java/forge/game/ability/effects/ChangeTargetsEffect.java
@@ -198,7 +198,11 @@ public class ChangeTargetsEffect extends SpellAbilityEffect {
                 Map<AbilityKey, Object> runParams = AbilityKey.newMap();
                 runParams.put(AbilityKey.SourceSA, tgtSA);
                 runParams.put(AbilityKey.Targets, distinctObjects);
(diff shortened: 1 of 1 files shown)


Hanmac:
> SpellAbility doesn't implement GameObject

Your premise is wrong:

`class CardTraitBase implements GameObject`
`class SpellAbility extends CardTraitBase`

BigCrunch22:
Got the description redone after another analysis. Script changes stays the same, so no issues on that part.

tool4ever (approved):

Hanmac:
We probably should unify it later.
Most places that use `AbilityKey.Cause` cast it to SpellAbility.

